PR #2011, opened 2018-07-28
fixed Issue #2001 - support for Tinker periodic boundary box

### LINKED ISSUE #2001: TXYZ Arc file parse fails, doesn't recognize periodic boundary box definition
### Expected behaviour
Parse Tinker Archive file into a Universe

### Actual behaviour
Error on parsing, fails to parse periodic boundary box produced by Tinker on line 2 of each frame. 

103
    10.000000   10.000000   10.000000   90.000000   90.000000   90.000000
     1  OH    -4.258215   -1.973780    0.949736  1232     2    10
     2  CT    -4.263714   -1.766981    2.353863  1063     1     3    11    12
     3  CT    -2.868573   -1.546285    2.893741  1066     2     4     6    13
     4  OS    -2.130847   -2.679206    2.594471  1068     3     5
     5  CT    -0.793726   -2.351497    2.219711  1069     4     7    14    18

[hexanes_arc.zip](https://github.com/MDAnalysis/mdanalysis/files/2215812/hexanes_arc.zip)
``` python
import MDAnalysis as mda

u = mda.Universe(top)

....

```

0.18.0

micaela-matta:
Fixes #2001

 - TXYZ now reads box information when present

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/topology/TXYZParser.py
```diff
@@ -88,9 +89,17 @@ def parse(self, **kwargs):
             names = np.zeros(natoms, dtype=object)
             types = np.zeros(natoms, dtype=np.int)
             bonds = []
+            #Checks whether file contains periodic boundary box info
```
kain88-de: The comment would be better if it explains why this works. Currently, it only explains what the code does. So something along the lines of 


> PBC box information is stored optionally in tinker files at the beginning of a frame. The box is stored as (x, y, z, alpha, beta, gamma). If not box information is stored the second value (space separated) is an atom name. To detect if a box is stored we, therefore, need to check if the second value can be converted to float. This assumes there are no pure numerical atom names.


it is longer but much better understandable for a maintainer later/

package/MDAnalysis/topology/TXYZParser.py
```diff
@@ -43,6 +43,7 @@
 """
 
 from __future__ import absolute_import
```
kain88-de: Since you now use `zip` we also need to import the iterating zip from six. Please add the line `from six.moves import zip` below the __future__ import.

package/MDAnalysis/coordinates/TXYZ.py
```diff
@@ -74,7 +74,15 @@ def __init__(self, filename, **kwargs):
         root, ext = os.path.splitext(self.filename)
         self.xyzfile = util.anyopen(self.filename)
         self._cache = dict()
-
+        with util.openany(self.filename) as inp:
```
kain88-de: also add a comment here *why* the code works.

package/MDAnalysis/topology/TXYZParser.py
```diff
@@ -88,9 +89,17 @@ def parse(self, **kwargs):
             names = np.zeros(natoms, dtype=object)
             types = np.zeros(natoms, dtype=np.int)
             bonds = []
+            #Checks whether file contains periodic boundary box info 
+            fline = inf.readline()
+            try:
+                float(fline.split()[1])
+            except:
```
kain88-de: expect ValueError

testsuite/MDAnalysisTests/datafiles.py
```diff
@@ -85,7 +85,7 @@
     "XTC_sub_sol",
     "XYZ", "XYZ_psf", "XYZ_bz2",
     "XYZ_mini", "XYZ_five", # 3 and 5 atoms xyzs for an easy topology
-    "TXYZ", "ARC",        # Tinker files
+    "TXYZ", "ARC", "ARC_PDB",        # Tinker files
```
richardjgowers: it's called `ARC_PDB` here but `ARC_PBC` everywhere else

orbeckst:
There's not `__future__.zip`:
```python
 from __future__ import absolute_import, zip
```
Makes Travis unhappy.

kain88-de:
Zip is in six.moves and not the future module.

package/MDAnalysis/topology/TXYZParser.py
```diff
@@ -88,9 +91,22 @@ def parse(self, **kwargs):
...
+                # If float conversion failed, we have first atom line
+                pass
+            else:
+                # If previous try succeeded it was a box
+                # so read another line to find the first atom line
+                fline = inf.readline()
             # Can't infinitely read as XYZ files can be multiframe
-            for i in range(natoms):
-                line = inf.readline().split()
+            for i, line in zip(range(natoms), itertools.chain([fline], inf)):
```
richardjgowers: The linter was complaining about this line because it thought we weren't iterating the `range`, but we are, so I changed the linter RC file.

Issue #1185, opened 2017-01-23
Add molecule unwrapping / nopbc utilities


richardjgowers:
A commonly asked question is how to make molecules whole (ie undo PBC transformations.)

The only thing we currently have is [make_whole](https://github.com/MDAnalysis/mdanalysis/blob/develop/package/MDAnalysis/lib/mdamath.py#L270) which is pretty much hidden away anyway.

So things that need doing is:
 * add undo PBC support for orthogonal and triclinic boxes (`make_whole` only does ortho) as some sort of mini library (`mda.lib.pbc_tools` ?)
 * support unwrapping without bonds (can use distance as proportion of box length, not always ideal for v large molecules, but works 90% of time)
 * AtomGroup method that unwraps (will center atoms around the COM by default?) `AtomGroup.unwrap` ?
 * add unwrap kwarg to various methods as standard (similar to how we have pbc as kwarg) (aka #1004)

Have I missed anything?

jbarnoud:
I do not know if it should be in `mda.lib` as it deals with atom groups and we tend to avoid dependencies to other pieces of the library in `lib`. `make_whole` is already an alien on that respect. I would be file with some `mda.pbc` module.

I am all in favor of `AtomGroup.unwrap`. One of its use case would be to deal with odd boxes that gets transformed into triclinic boxes. The unwrap method would be an interesting replacement to `gmx trjconv -ur compact` im my workflow.

I would like a `AtomGroup.make_whole` method that would make each fragment whole. It may get tricky with molecules that are only partially in the `AtomGroup`, though.

It would be awesome to have such library if we get to implement #786.

jbarnoud:
(just mentioning `AtomGroup.wrap` for the sake of completeness.)

kain88-de:
I would be for `mda.core.pbc` and `AtomGroup.pbc.unwrap`. That gives PBC the importance it deserves and having a special `.pbc` in the Atomgroup makes it easy to collect all functions with unwrapping/wrapping and should make them easier to discover for people new to MD simulations.

jbarnoud:
@kain88-de sounds good to me.

orbeckst:
I would like to have the following functionality in this (?) module that I have not seen anywhere else: superimpose by RMSD (align) and then cut out a centered and compact unitcell out of the inifititely periodic (rotated) lattice so that, e.g., density calculations always show a perfectly aligned unitcell (maybe changing in size due to pressure coupling).

[...]

PR #2012, opened 2018-07-28
Issue 1004 unwrap

### LINKED ISSUE #1004: Molecules across periodic boundaries are not properly treated
### Expected behaviour
A residue or atomgroup containing atoms that straddle a periodic boundary should correctly unwrap the collection to one side or the other in order to properly compute the center of mass or moment of inertia.

### Actual behaviour
Although `pbc=True` flags exist for a variety of methods that operate on residues and atom groups, this only maps atom coordinates back into the original periodic simulation box. However, if analyzing a trajectory, such as in my case a collection of molecules, some of which are located across a periodic boundary, this does nothing to solve the issue of correctly computing centers of mass and (importantly for me) the moment of inertia. 

To deal with this, I wrote a few helper functions in my analysis scripts, which I've reproduced below. This is simply one approach; if the bond/topology information is available it would potentially be easier to utilize that information. 

It would be great to see something like this incorporated into MDAnalysis, in a much more efficient manner. For reference, `H` is the Parrinello-Rahman matrix, or `triclinic_dimensions` as referenced in MDAnalysis.

``` python
import MDAnalysis as mda

def roundf(x):
    if x >= 0:
        return np.floor( x + 0.5 )
    else:
        return np.ceil( x - 0.5 )

def apply_min_image(H,x):
    s = np.dot(la.inv(H),x)
    for i in range(0,3):
        s[i] -= isperiodic[i]*roundf(s[i])
    return np.dot(H,s)

def upwrap_molecule(A, H):
    for i in range(1, A.shape[0]):
        A[i,:] = apply_min_image(H, A[i,:] - A[i-1,:]) + A[i-1,:]
```

My approach above is simple: it linearly marches through the coordinates and computes the minimium image distance between an atom and the next one, which will reconstruct a molecule on the side of the first (reference) atom. 

If something like this, or a better approach, can be incorporated into MDAnalysis in a more efficient manner, then great! All I need is really a way to properly take into account periodicity. Currently the COM and inertial tensor are incorrect for molecules whose atoms are across a periodic boundary. 

Thanks!

0.15.0

richardjgowers:
Fixes #1004

 - 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/core/groups.py
+++ b/package/MDAnalysis/core/groups.py
@@ -103,7 +103,7 @@ from numpy.lib.utils import deprecate
 
 from .. import _ANCHOR_UNIVERSES
 from ..lib import util
-from ..lib.util import cached, warn_if_not_unique, unique_int_1d
+from ..lib.util import cached, warn_if_not_unique, unique_int_1d, make_whole
 from ..lib import distances
 from ..lib import transformations
 from ..selections import get_writer as get_selection_writer_for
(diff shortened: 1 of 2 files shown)


orbeckst:
@richardjgowers is this PR still active? 

The issue #1004 is still open and presumably, in the area of PBC/making whole we still have some work to do, at least from reading #1185 . Is this also your sense?

richardjgowers:
@orbeckst yeah this PR is still relevant.  Our current 'pbc' option equates to 'wrap', ie pack everything in before calculating an answer.  What most people need is actually 'unwrap' ie make the molecule whole before giving me an answer.  This PR would try and do that.

PR #2013, opened 2018-07-28
Modified the distance function in RDF

ayushsuhane:
RDF in analysis.py currently uses  distance_array which becomes slow for large number of particles. Although it still uses the same distance array for small number of particles, it can also use KDTree for large number of particles to boost the performance.

PR Checklist
------------
 - [X ] Tests?
 - [X] Docs?
 - [X ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/analysis/rdf.py
+++ b/package/MDAnalysis/analysis/rdf.py
@@ -235,7 +235,6 @@ class InterRDF(AnalysisBase):
 
 class InterRDF_s(AnalysisBase):
     """Site-specific intermolecular pair distribution function
-
     Arguments
     ---------
     u : Universe
@@ -254,48 +253,32 @@ class InterRDF_s(AnalysisBase):
     step : int (optional)
           The step size through the trajectory in frames (default is
           every frame)
-
     Example
     -------
-
     First create the :class:`InterRDF_s` object, by supplying one Universe and
     one list of pairs of AtomGroups, then use the :meth:`~InterRDF_s.run`
     method::
-
       from MDAnalysisTests.datafiles import GRO_MEMPROT, XTC_MEMPROT
       u = mda.Universe(GRO_MEMPROT, XTC_MEMPROT)
-
       s1 = u.select_atoms('name ZND and resid 289')
       s2 = u.select_atoms('(name OD1 or name OD2) and resid 51 and sphzone 5.0 (resid 289)')
       s3 = u.select_atoms('name ZND and (resid 291 or resid 292)')
       s4 = u.select_atoms('(name OD1 or name OD2) and sphzone 5.0 (resid 291)')
       ags = [[s1, s2], [s3, s4]]
-
       rdf = InterRDF_s(u, ags)
       rdf.run()
-
     Results are available through the :attr:`bins` and :attr:`rdf` attributes::
-
       plt.plot(rdf.bins, rdf.rdf[0][0][0])
-
     (Which plots the rdf between the first atom in ``s1`` and the first atom in
     ``s2``)
-
     To generate the *cumulative distribution function* (cdf), use the
     :meth:`~InterRDF_s.get_cdf` method ::
-
       cdf = rdf.get_cdf()
-
     Results are available through the :attr:'cdf' attribute::
-
       plt.plot(rdf.bins, rdf.cdf[0][0][0])
-
     (Which plots the cdf between the first atom in ``s1`` and the first atom in
     ``s2``)
-
-
     .. versionadded:: 0.19.0
-
     """
     def __init__(self, u, ags,
                  nbins=75, range=(0.0, 15.0), density=True, **kwargs):
@@ -320,7 +303,6 @@ class InterRDF_s(AnalysisBase):
         self.count = count_list
         self.edges = edges
         self.bins = 0.5 * (edges[:-1] + edges[1:])
-        self._maxrange = self.rdf_settings['range'][1]
 
         # Need to know average volume
         self.volume = 0.0
@@ -328,13 +310,8 @@ class InterRDF_s(AnalysisBase):
 
     def _single_frame(self):
         for i, (ag1, ag2) in enumerate(self.ags):
-            pairs, dist = distances.capped_distance(ag1.positions,
-                                                   ag2.positions,
-                                                   self._maxrange,
-                                                   box=self.u.dimensions)
-            result = np.ones((len(self.ag1), len(self.ag2)), dtype=np.float64)
-            result *= self._maxrange + 1.0
-            result[tuple(zip(*pairs))] = dist
+            result=distances.distance_array(ag1.positions, ag2.positions,
+                                            box=self.u.dimensions)
             for j in range(ag1.n_atoms):
                 for k in range(ag2.n_atoms):
                     count = np.histogram(result[j, k], **self.rdf_settings)[0]
@@ -373,15 +350,12 @@ class InterRDF_s(AnalysisBase):
 
     def get_cdf(self):
         """Calculate the cumulative distribution functions (CDF) for all sites.
-
         Note that this is the actual count within a given radius, i.e.,
         :math:`N(r)`.
-
         Returns
         -------
               cdf : list
                       list of arrays with the same structure as :attr:`rdf`
-
         """
         # Calculate cumulative distribution function
         # Empty list to restore CDF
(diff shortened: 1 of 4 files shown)


ayushsuhane:
I have tried to keep the flow of the code intact, partly because I am not sure how to handle the exclusion block after the evaluation of pairs and their corresponding distances. If the exclusion block is ``None``, rdf can be directly calculated by binning the distances, but with blocks we have to create the ``n x m`` arrays first. I am not sure whether this will have any advantage. I will try to check the performance first with available data files. Also, the tests are  missing from the InterRDF_s. If it happens that the performance is increased, I will also try to add tests for InterRDF_s

package/MDAnalysis/analysis/rdf.py
```diff
@@ -187,24 +187,32 @@ def _prepare(self):
...
             self._exclusion_mask = None
 
     def _single_frame(self):
-        distances.distance_array(self.g1.positions, self.g2.positions,
-                                 box=self.u.dimensions, result=self._result)
+        pairs, dist = distances.capped_distance(self.g1.positions,
+                                                self.g2.positions,
+                                                self._maxrange,
+                                                box=self.u.dimensions)
+        self._result[tuple(zip(*pairs))] = dist
```
richardjgowers: This looks ugly and slow, but I see that for the exclusion mask feature you need to have the full matrix of distances.  I think it might be better to implement the exclusion mask in a different way to avoid having to have this full matrix representation.  

Because this full matrix scales O(N^2) we should really be avoiding it anyway.  I'll have a think to see if I can come up with a quick way to do exclusion_matrix.   You should be able to go through the indices, check if they're inside an excluded block, and if so ignore that distance.

richardjgowers:
So with exclusion block, you can check the raw indices quite easily.  If exclusion block = (2, 3), then `x={0, 1}, y={0, 2}` is excluded, and `x={2, 3}, y={3, 5}` etc

Or more generally `x={i * lx, (i+1) * lx - 1}, y={i * ly, (i+1) * ly - 1}`

So you can look at the x index, determine it's in the `i`th block, and see if the corresponding y index is also in the same block, if it is, exclude it

ayushsuhane:
Yes, I think this should fix the problem. However, I am not getting any improvement in the timings except use of capped_distance/nsgrid would be able to handle large datasets. 
```
idxA, idxB = pairs[:, 0]//self._exclusion_block[0], pairs[:, 1]//self._exclusion_block[1]
mask = np.where(idxA != idxB)[0]
dist = dist[mask]
```

zemanj:
@ayushsuhane I worked on a faster RDF computation a while ago but it's not production quality yet since I haven't incorporated exclusions yet, and it also doesn't make use of any more sophisticated approaches such as grid search for capped distances. I'd very much appreciate it if you could have a look at what I did in the [lib_distances](https://github.com/zemanj/mdanalysis/tree/lib_distances) branch of my forked repo and tell me what you think.

ayushsuhane:
@zemanj Thanks!! It is amazing. Pardon me for a noob question, but I am trying to understand what is it which makes the function ``_calc_distance_histogram`` faster than ``np.histogram(distance_array(.., .., .., ..), **kwargs)`` apart from a small overhead (hopefully) taken by numpy.histogram .

Correct me if I am wrong, but I thought numpy methods also work on C based computations, and the function you have defined is doing the same but (most probably) without any checks which numpy.histogram most certainly does. Apart from it, there is another difference i.e the usage of ``minimum_image_ortho_lazy`` for distance computations. I am not sure if this assumption brought the performance boost or doing C based computations for generating a histogram. Maybe you can throw some light on it. How much of performance improvement do you achieve with this approach. If it is significant, I guess it can be incorporated in the library.

ayushsuhane:
@richardjgowers So the problem here is the use of large range. If the box size is small i.e. 30-50, searching a range of 15 (do we really need this large range even for smaller boxes?) really becomes a N^2 problem and i dont think there is anything we can do about it. I was testing the performance with (``small.gro``)[https://github.com/ayushsuhane/Benchmarks_Distance/blob/master/Notebooks/small.gro], and which has an orthogonal box of 50. This is why the normal distance array was working quite fast as compared to other  methods. Ofcourse, there is some overhead to calculate the pairs in capped_function as well. So choosing ``capped_function`` over sparse array is always slower, but it has an advantage of dealing with arbitrary size of data (upto a limit though).  What do you suggest? Its the question of performance over flexibility?

richardjgowers:
Search range is 15 for scientific reasons. Normal intermolecular force
field potentials are 9-12, and you want to go a little beyond that. So
can’t really change that.

A box size of 50 is fairly small tbh. So maybe try looking at the larger
system I sent.

ayushsuhane:
image

While the performance of RDF decreased a little bit for small data set (g1, g2), but it seems there is an improvement for larger datasets (the last point in the graph correspond to 10000 particles in both the groups used for RDF calculations)

I also extended the benchmarks to include the number of atoms as a parameter, and this performance should be reflected there as well.

ayushsuhane:
@richardjgowers I think it depends on case to case, which is difficult to quantify. I tried to come up with rules based on one/two cases at hand, but a more exhaustive study needs to be performed. However, I think this should suffice for now. It already shows an improvement in time by a factor of 2-2.5 for large datasets, which is quite huge (from 70 secs to 25 secs for 10000 atoms in both groups)

zemanj:
@ayushsuhane Ok here's some explanation of what I did:

I started working on RDF computations for a project investigating long-range correlations in ionic liquids, which involved systems comprising up to 350,000 ion pairs. The amount of RAM required to store a distance array then amounts to about 912.7 GB, which is slightly more than what is available on my laptop... :wink:
The fact that memory requirements for RDF computations scale like O(N^2) quickly becomes a total showstopper, which IMHO is simply unacceptable. This is where the ``distance_histogram`` methods come into play:
In the methods I implemented, the calculated distances are never written to main memory and actually never leave the CPU cache. While this already makes the routine less memory bound, there is a much bigger advantage: Since most MD simulations employ PBC, the maximum distance considered for RDFs is half the box length, and therefore, only distances within that cut-off are of interest. In the ``distance_histogram`` method, the square of the distances is computed and checked to lie within the given range before taking square roots. This means that in a cubic, homogeneous system, one can save almost half of the square root evaluations.
Further improvements are:
* Coordinate memory layout is changed (That's only O(N)) so that atom coordinates can be processed in blocks of 32 particles, allowing exploitation of SIMD acceleration (SSE/AVX units up to AVX1024)
* Cache line optimization (so far only for 64 byte cache line size)
* Memory locality (make the most out of what's already in the CPU cache)
* Prevention of false sharing (when using OpenMP)
* Faster vectorized minimum image computations

The last point is exactly what is implemented in ``minimum_image_ortho_lazy``. It's called lazy because the routine assumes that distances can never exceed 1.5 times the box length. This is ensured by packing all coordinates in the box prior to RDF computation (again, that's only O(N)).
Within the ``minimum_image_ortho_lazy`` method, vectorized CMOVs are used instead of branching to account for PBC.

My latest benchmarks for groups containing 5000 atoms suggest a speedup factor of about 2.6 for single core, about 7.6 for OpenMP on 4 cores with hyperthreading, and about 18 when used in a trivially parallelized MPI implementation (one process per trajectory frame) on the same machine (overcommitted with 12 MPI processes). The latter is probably something for pmda, though.

Please be aware of the fact that AVX acceleration will only be used when compiled with ``-march=native``, that might affect the timings a bit.

Concerning the checks you mentioned that numpy.histogram performs, con you elaborate on that? I simply check the range and keep numpy compliance by having a closed interval in the last RDF bin (even though I think the latter is complete BS). I use int64_t for the histograms and therefore don't check for overflow since that's extremely unlikely.

There's certainly still room for improvement since my code doesn't scale well beyond 4 cores with OpenMP (but with MPI it does), and it only ever uses brute force neighbor search, which is probably its weakest point.

I'll try to get this thing production ready asap, maybe we can use it to replace the current brute force computations. Comments/criticism/opinions, and especially ideas for improvements and how this could be combined with e.g. grid search methods are very welcome!

richardjgowers:
@zemanj all those changes sound like they're compatible (ie the speedups will be multiplicative) with what @ayushsuhane and @seb-buch has done so far.  We've implemented a domain decomposition method, so rather than comparing all pairs, we spatially bin the particles and only compare nearby particles.  What you've done can be applied to the bin<->bin distance calculation (with some hacking...).

zemanj:
@richardjgowers Let's hope it doesn't end up being too hacky in the end... I'll look into it once this goes in.

richardjgowers:
It’s mostly that nsgrid (merged now into lib) doesn’t make memory in each
domain contiguous, so your nice vector loads aren’t going to work.
Definitely fixable though.

ayushsuhane:
@richardjgowers  Thanks to you for constant feedback and support.

Issue #974, opened 2016-09-07
"around" selection is slow and memory-intensive


mnmelo:
As of `0.15` performing an `around` search can be quite expensive, both compute-wise as memory-wise. This crops up more obviously for systems with many atoms, and where the reference for `around` is itself also large.

This was tested in a system with ~640k atoms, where the reference has 1130 atoms (`select_atoms around 15 name ATNAME`). It took at least 15 seconds. Doubling the system size (and reference too) busts my 16Gb of memory.

From the discussion in #383, and later the addition of OpenMP (#529), I was under the impression our implementation was quite fast. But for comparison, for the same system and machine VMD does the same selection operation with little memory usage spike, and virtually instantaneously. Granted, VMD doesn't do PBC but is it enough to account for this?

I took a look at their code and from what I could understand they implement several optimizations some of which I thought we were already doing since #529:
- gridded search;
- parallelization (threading);
- pre-filtering of the search space (for compact references it's easy to define a parallelepiped that contains it with a padding equal to the cutoff, and you're sure your contacts will all be inside that volume);
- shortcutting (when an atom being searched is found to be in contact, no need to calculate further distances from it to the other reference atoms).

I admit I'm not familiar with the internals of `distance_array` and `contact_matrix`, but my question is: is there room for optimization or am I hitting MDAnalysis' limits here?

kain88-de:
How does the situation look with older versions of MDAnalysis?

I don't doubt that the KDTree is memory intensive, if I remember the algorithm it has to calculate a full distance matrix. @richardjgowers has some prototype code for the gridded search.

mnmelo:
Ah, ok, I remember reading something about @richardjgowers' gridded stuff, and thought that was already in place with OpenMP. I'll check with older versions and post here.

richardjgowers:
So iirc, PBC conditions have to use a `distance_array`, so it will do a `len(system) * len(reference)` search.  I imagine KDTree is much better than this.

The code for the `around` selection is all here if you want to try and optimise it.
https://github.com/MDAnalysis/mdanalysis/blob/develop/package/MDAnalysis/core/Selection.py#L242

And yeah, I played around with gridded search before, [here](https://github.com/richardjgowers/cellgrid) I think it mostly works... I was having some trouble making it competitive as brute force methods for smaller system sizes, probably because of a mistake I made in the Cython code.

richardjgowers:
[...]

PR #2022, opened 2018-08-04
Around keyword can use KDTree for periodic boundary conditions

ayushsuhane:
Around keyword only used _apply_distmat if periodic keyword was active. Both brute force and kdtree routines can be used now for PBC and non-PBC calculations

Fixes  partly #974 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -20,6 +20,8 @@ The rules for this file:
 
 Enhancements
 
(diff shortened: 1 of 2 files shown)


richardjgowers:
Can you rebase, the failures are unrelated

PR #2024, opened 2018-08-04
BUG: Force 64-bit storage for xdr file offsets

### LINKED ISSUE #2021: BUG (windows): seeking over 4 GB on xdr formats
`test_seek_tell_largefile()` in `testsuite/MDAnalysisTests/formats/test_libmdaxdr.py` currently fails on Windows.

Interestingly, the bytes tell value suggests that the seek goes all the way back (?) to the start of the file after 2**32 byte seek, and then another 1000 bytes (so unit tests fails with `1000 == (2**32 + 1000)`)

There are a few documents online indicating that Windows can `_fseeki64`  beyond the end of a file, and the low-level gromacs library we use is apparently supposed to use `_fseeki64` to move the file pointer on Windows.

The unit test passes if we use `2**30` bytes, which may be a helpful clue--are we overflowing something or more likely to be dealing with an OS memory blocking issue of some sort?

tylerjereddy:
* Fixes Issue #2021 with Windows xdr large file
offset handling; should reduce appveyor failure
count from `13` -> `11` if no new issues
recently introduced in dev branch

* on the Windows platform xdr file offsets are now
forced to use a 64-bit storage type so that, for example,
file seeks beyond 4 GB are correctly reported

Fixes #2021 

There may be a better way to do this but `off64_t` wasn't being recognized on my Windows box despite including the type headers.

### DIFF
--- a/package/MDAnalysis/lib/formats/src/xdrfile.c
+++ b/package/MDAnalysis/lib/formats/src/xdrfile.c
@@ -45,6 +45,7 @@
 #endif
 
 #include <sys/types.h>
+#define _FILE_OFFSET_BITS 64
 
 #ifdef _WIN32
 #  include <io.h>
@@ -2518,8 +2519,8 @@ static int xdrstdio_getlong (XDR *, int32_t *);
 static int xdrstdio_putlong (XDR *, int32_t *);
 static int xdrstdio_getbytes (XDR *, char *, unsigned int);
 static int xdrstdio_putbytes (XDR *, char *, unsigned int);
-static off_t xdrstdio_getpos (XDR *);
-static int xdrstdio_setpos (XDR *, off_t, int);
+static int64_t xdrstdio_getpos (XDR *);
+static int xdrstdio_setpos (XDR *, int64_t, int);
 static void xdrstdio_destroy (XDR *);
 
 /*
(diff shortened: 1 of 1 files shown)


tylerjereddy:
There's something alarming about Windows tests in dev branch (the appveyor merge test) -- someone must have merged something without checking the total failure count on Appveyor. I'm not really sure what to do about this, but I can't outpace the rate of breakage from multiple devs if we don't prevent that somehow.

jbarnoud:
From the `ftello` manpage, it seems that `xdrio_getpos` may return a int32 on some platforms. To be on the safe side, maybe it would be good to set `#define _FILE_OFFSET_BITS 64` and make sure we always get an int64? See <https://linux.die.net/man/3/ftello>.

richardjgowers:
If we pytest.mark the current failures as known failures we can then make
appveyor fail on failures

jbarnoud:
Pytest.mark.xfail can even take a condition so the tests can be allowed to fail on windows only. Le 5 août 2018 4:01 PM, Richard Gowers <notifications@github.com> a écrit :If we pytest.mark the current failures as known failures we can then make
appveyor fail on failures

tylerjereddy:
I revised to add the preprocessor flag @jbarnoud requested; I didn't have success removing my type adjustments after adding that flag though (would have been nice if that worked).

I've also rebased and force pushed: the appveyor merge AND branch tests should both show 11 total test failures (down from 13). There should also be 1 xfail & 9 errors on Windows; I'll look into xfail-tagging as discussed here in a separate PR as time permits.

PR #2025, opened 2018-08-05
Use stricter type for pkdtree indices

jbarnoud:
Using consistently np.int64 for the indices used in pkdtree avoids type
errors when using unique_int_1d. This should reduce the number of
failing tests on windows.

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


jbarnoud:
AppVeyor failures go down back to 13.

PR #2028, opened 2018-08-06
faster fragment finding implementation

richardjgowers:
Currently finding fragments is slow, this should be faster!

### DIFF
--- a/benchmarks/benchmarks/ag_methods.py
+++ b/benchmarks/benchmarks/ag_methods.py
@@ -4,7 +4,7 @@ import MDAnalysis
 import numpy as np
 
 try:
-    from MDAnalysisTests.datafiles import GRO
+    from MDAnalysisTests.datafiles import GRO, TPR, XTC
     from MDAnalysis.exceptions import NoDataError
 except:
     pass
@@ -322,3 +322,17 @@ class AtomGroupAttrsBench(object):
         Requires ag of size 2.
         """
         self.ag[:2].bond
+
+
+class FragmentFinding(object):
+    """Test how quickly we find fragments (distinct molecules from bonds)"""
+    params = [(TPR, XTC),  # single large fragment, many small solvents
+              (PSF, DCD),  # single large fragment
+              (TRZ_psf, TRZ)]  # 20ish polymer chains
+    param_names = ['universe']
+
+    def setup(self, universe):
+        self.u = MDAnalysis.Universe(*universe)
+
+    def test_find_fragments(self, universe):
+        frags = self.u.atoms.fragments
--- a/package/MDAnalysis/lib/_cutil.pyx
+++ b/package/MDAnalysis/lib/_cutil.pyx
@@ -357,8 +357,8 @@ def find_fragments(atoms, bondlist):
         todo.insert(atoms_view[i])
     # Process edges into map
     for i in range(bonds_view.shape[0]):
-        a = bondlist[i, 0]
-        b = bondlist[i, 1]
+        a = bonds_view[i, 0]
+        b = bonds_view[i, 1]
         # only include edges if both are known nodes
         if todo.count(a) and todo.count(b):
             bondmap[a].insert(b)
--- a/package/MDAnalysis/lib/_cutil.pyx
+++ b/package/MDAnalysis/lib/_cutil.pyx
@@ -320,20 +320,24 @@ cdef float _norm(float * a):
 @cython.boundscheck(False)
 @cython.wraparound(False)
 def find_fragments(atoms, bondlist):
-    """Calculate distinct fragments from nodes and edges
+    """Calculate distinct fragments from nodes (atom indices) and edges (pairs
+    of atom indices).
 
     Parameters
     ----------
-    atoms : numpy.ndarray
-       array of nodes
-    bonds : numpy.ndarray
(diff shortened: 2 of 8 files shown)


package/MDAnalysis/lib/_cutil.pyx
```diff
@@ -310,4 +313,71 @@ cdef float _norm(float * a):
...
+        # Loop until fragment fully exploredt
+        while not frag_todo.empty():
+            for a in frag_todo:
+                frag_done.insert(a)
+                this_frag.push_back(a)
+                for b in bondmap[a]:
+                    if not frag_done.count(b):
+                        frag_next.insert(b)
+            # update todo list
+            frag_todo.clear()
```
richardjgowers: this is kind of ugly, but I wanted to not change `frag_todo` whilst I was iterating over it, but maybe it's safe to do so?  Ideas welcome

package/MDAnalysis/lib/_cutil.pyx
```diff
@@ -310,4 +313,71 @@ cdef float _norm(float * a):
...
+    frags = []
+
+    while not todo.empty():  # While not all nodes have been done
+        # Start a new fragment
+        frag_todo.clear()
+        frag_done.clear()
+        this_frag.clear()
+        # Grab a start point for next fragment
+        frag_todo.insert(deref(todo.begin()))
+        # Loop until fragment fully exploredt
```
jbarnoud: extra "t" at the end of the comment.

package/MDAnalysis/lib/_cutil.pyx
jbarnoud: Since the swap is done in constant time, I am fine with this approach.

jbarnoud:
The fact that the tests are broken may indicate that I missed something reading the code, though. The good thing is that there is no new windows specific failures,

package/MDAnalysis/lib/_cutil.pyx
```diff
@@ -310,4 +313,71 @@ cdef float _norm(float * a):
...
     for n in range(3):
         result += a[n]*a[n]
-    return sqrt(result)
\ No newline at end of file
+    return sqrt(result)
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+def find_fragments(np.int64_t[:] atoms, int[:, :] bondlist):
```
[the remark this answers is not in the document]
richardjgowers: I actually put this in mdamath instead, but it has docs now

richardjgowers:
@jbarnoud yeah the tests found that I was occasionally having duplicates in fragments, fixed now.  Tests, docs, benchmarks etc done, should be good to go

benchmarks/benchmarks/ag_methods.py
```diff
@@ -319,3 +319,17 @@ def time_bond(self, num_atoms):
         Requires ag of size 2.
         """
         self.ag[:2].bond
+
+
+class FragmentFinding(object):
```
orbeckst: 👍 for added benchmark!

jbarnoud:
AppVeyor is complaining about types. From what I understand, edges should be int64 instead of int32 in the tests.

richardjgowers:
Ok that's weird, the appveyor github status should have been red?

tylerjereddy:
appveyor looks red to me

richardjgowers:
@MDAnalysis/coredevs this is ready

package/MDAnalysis/lib/_cutil.pyx
```diff
@@ -312,3 +316,76 @@ cdef float _norm(float * a):
...
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+def find_fragments(atoms, bondlist):
+    """Calculate distinct fragments from nodes and edges
+
+    Parameters
+    ----------
+    atoms : numpy.ndarray dtype int64
```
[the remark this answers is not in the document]
zemanj: Not sure if this should be tested?

richardjgowers:
@zemanj thanks!  I was a little uncertain about modifying the set as I iterated over it, but this works nicely

zemanj:
@richardjgowers I just made the docstring a little more verbose and made sure that `bondmap` is constructed from `bonds_view` instead of `bondlist`. In practice, this won't change anything but makes sure that *array_like* input also works correctly (now corresponding to the docstring). I also changed the inputs in the respective test to be nested lists instead of arrays for that reason. Using numpy arrays of correct shape and dtype should be covered by the tests in `core/test_fragments.py`.

zemanj:
@richardjgowers BTW do you happen to know if it's possible in Cython to construct a `numpy.ndarray` directly from a `libcpp.set` without converting to a `libcpp.vector` first? I was unable to find anything useful on that topic...

kain88-de:
There will be hardly a way do to this directly. The set uses a hash table
like data structure under the hood. How is the conversion to a vector done?

BTW using a normal for loop to fill a pre allocated numpy array should work
just fine for the conversion.

zemanj:
@kain88-de Thanks, that's pretty much what I expected. I was just wondering whether something equivalent to the range constructor of the `std::vector` in C++ existed for numpy arrays in Cython...

zemanj:
Ok, good to go. Merge?

richardjgowers:
Yup, thanks for the help!

Issue #1988, opened 2018-07-15
Problem with loading XTC offsets in parallel


mkhoshle:
Hi MDAnalysis Developers,

I have an MPI job where I have 48 trajectory segments in the XTC format (newtraj_0.xtc,...newtraj_47.xtc). In order to have 48 segments one can just use a small size xtc file and copy this single segment 48 times to get a list of segments. I did not put all the segments in the repository because the whole size was large (30 GB). I repeat my job several times to collect statistics.
In my MPI job I use 48 processes and each process tries to copy one of the segments (e.g. newtraj_6.xtc for rank ID 6) to another folder. This copying is to avoid each effects. After each rank finish copying, it creates an index file for its own segment by doing print(len(mda.Universe(PSF, newtraj_rankID.xtc).trajectory)). I have a barrier to synchronize all processes.

**Expected behavior**
What I expect to see from my code is that each rank should create its index file and they should be able to successfully use the chain-reader to create the universe. 
Note:
This error happens when I test my code on more than one node. On a single workstation (node), it works fine. The index files are being written independently by each rank. So there should not be any overlap and I also tested the code with rank0 only creating the indices. However, it does not work either. The offset files cannot be corrupted except having a race condition. Is there any method in MDAnalysis like creating the universe, getting n_frames, etc that lead to rewriting the index file?

**Actual behavior**
Some ranks fail due to the following error:
```
Traceback (most recent call last):
  File "/oasis/scratch/comet/mkhoshle/temp_project/MPI-traj-split/main1-ga4py-updated-rmsd-chain-reader.py", line 73, in <module>
    u = mda.Universe(PSF, XTCs)
  File "/home/mkhoshle/miniconda2/envs/ga4py/lib/python2.7/site-packages/MDAnalysis/core/universe.py", line 305, in __init__
    self.load_new(coordinatefile, **kwargs)
  File "/home/mkhoshle/miniconda2/envs/ga4py/lib/python2.7/site-packages/MDAnalysis/core/universe.py", line 535, in load_new
    self.trajectory = reader(filename, **kwargs)
  File "/home/mkhoshle/miniconda2/envs/ga4py/lib/python2.7/site-packages/MDAnalysis/coordinates/chain.py", line 123, in __init__
    for filename in self.filenames]
  File "/home/mkhoshle/miniconda2/envs/ga4py/lib/python2.7/site-packages/MDAnalysis/coordinates/core.py", line 83, in reader
[...]

PR #2033, opened 2018-08-08
Expansion of dihedrals analysis module

hfmull:
 - Added Dihedral and Janin classes to dihedral.py
 - Ramachandran class changed to account for selection errors
 - Added tests to test_dihedral.py to account for changes
 - Added reference plots for Ramachandran and Janin classes

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?
 - [n/a] Issue raised/referenced?


package/MDAnalysis/analysis/dihedrals.py
```diff
@@ -147,6 +235,16 @@ def __init__(self, atomgroup, **kwargs):
 
         phi_sel = [res.phi_selection() for res in residues]
         psi_sel = [res.psi_selection() for res in residues]
+        if any(sel is None for sel in phi_sel):
```
hfmull: This was added because `phi_selection()` and `psi_selection()` do not account for altloc selections. They will return `None` in the middle of the protein if that is the case, so this removes the `None` and the corresponding phi or psi angle. The amount of residues for which this is the case is very small so it seemed easier to just ignore them instead of rewriting `phi_selection()` and `psi_selection()`. If a user is interested in those specific residues, they can use the `Dihedral` class to look at them.

package/MDAnalysis/analysis/dihedrals.py
```diff
@@ -196,3 +294,94 @@ def plot(self, ax=None, **kwargs):
...
+
+        self.ag1 = residues.atoms.select_atoms("name N")
+        self.ag2 = residues.atoms.select_atoms("name CA")
+        self.ag3 = residues.atoms.select_atoms("name CB")
+        self.ag4 = residues.atoms.select_atoms("name CG CG1")
+        self.ag5 = residues.atoms.select_atoms("name CD CD1 OD1 ND1 SD")
+
+        if any(len(self.ag1) != len(ag) for ag in [self.ag2, self.ag3,
+                                                   self.ag4, self.ag5]):
+            raise ValueError("Too many or too few atoms selected. Check for "
```
hfmull: Similar to `Ramachandran`, altloc will make the selections fail, so here the class simply fails. Unlike `Ramachandran`, the user can then select for only on altloc and the analysis will run. Also I've found some topologies that were incomplete and did not have all the atoms it was supposed to so that's why it says 'too few'.

package/MDAnalysis/analysis/dihedrals.py
```diff
@@ -196,3 +294,94 @@ def plot(self, ax=None, **kwargs):
...
+        if not residues.issubset(protein):
+            raise ValueError("Found atoms outside of protein. Only atoms "
+                             "inside of a 'protein' selection can be used to "
+                             "calculate dihedrals.")
+        elif len(remove) != 0:
+            warnings.warn("All ALA, CYS, GLY, PRO, SER, THR, and VAL residues"
+                          " have been removed from the selection.")
+            residues = residues.difference(remove)
+
+        self.ag1 = residues.atoms.select_atoms("name N")
```
hfmull: I decided not to use `chi1_selection()` for this one because it does not capture all of the atoms it should, specifically the residues with multiple branches or non-carbon side chains.

package/MDAnalysis/analysis/dihedrals.py
orbeckst: altloc (and anything else that specifies as "dual topology") is a problem. 

@richardjgowers is there a sensible way how a user can switch between different topologies? The problem is that at the moment, `ag.residues` will contain _all_ atoms, including altloc ones, even if my selection for `ag` excluded them. This is problematic for the `residue.phi_selection()`, `residue.psi_selection()` and `residue.chi1_selection()` methods, which then see groups that are not 4 atoms.

richardjgowers: Currently, not really.  This would have to be something done when the file is parsed.  It sounds like maybe `Residue.phi_selection()` should take an altloc kwarg or something to tell it what to identify?

package/MDAnalysis/analysis/dihedrals.py
richardjgowers: Yes please!

package/MDAnalysis/analysis/dihedrals.py
orbeckst: MMTF switches between topologies... somehow, so we have that.

We could/should use the fact that the `residue.X_selections()` are methods. Perhaps indeed use a default `Residue.phi_selection(selection="altloc ' '")` or similar – with the issue that we also have to catch the case when this selection makes no sense.

richardjgowers: I think mmtf does the topologies as frames, (or rather we read models as frames, like how PDB is abused (or at least one of the various ways..)).  

Maybe naïve question, is altloc just a different position? Or can it be different functionalisation?  If the former, I guess you could write an altloc aware reader/parser which expands out the different possibilities.

hfmull: My understanding is that altloc just refers to different positions, but that those positions correspond to separate atoms that are both present in the topology, so they are treated just like extra atoms that are part of the residue. One of the issues with altloc selection is that there doesn't seem to be a way to select the atoms that have no altloc (at least I haven't figured it out). So to get the atoms with no altloc and those with altloc A, you have to say 'not altloc B', assuming there is only A and B for options. I think there can be more altlocs but I have yet to see any that go into C or higher.

orbeckst: Do you have the PDB ID of a file with altloc handy?

If it has altlocs then we assign a default, which is either `''` (empty string) or `' '` (one space). I don't know if our selection language can deal with it. But there's always the object-oriented/pandas-style approach
```python
ag[ag.altloc = '']
```
where you use boolean indexing. (Assuming that we expose an `.altloc` attribute somewhere... @richardjgowers would know.)

package/MDAnalysis/analysis/dihedrals.py
```diff
@@ -45,39 +45,74 @@
...
+
+   from MDAnalysis.analysis.dihedrals import Dihedral
+   R = Dihedral(ags).run()
+
+The angles can then be accessed with :attr:`Dihedral.angles`.
+
+Ramachandran analysis
+~~~~~~~~~~~~~~~~~~~~~
+
+The :class:`~MDAnalysis.analysis.dihedrals.Ramachandran` class allows for the
```
orbeckst: Please add references to the papers (Ramachandran and Janin) and add a section _References_.

package/MDAnalysis/analysis/dihedrals.py
hfmull: I've been using PDB ID 19hc, but that method returns an AttributeError and says that `AtomGroup` has no altloc attribute

package/MDAnalysis/analysis/dihedrals.py
```diff
@@ -45,39 +45,74 @@
...
    R = Ramachandran(r).run()
 
+Then it can be plotted using the built-in plotting method :meth:`Ramachandran.plot()`::
+
    import matplotlib.pyplot as plt
    fig, ax = plt.subplots(figsize=plt.figaspect(1))
-   ax.set_title("Ramachandran Plot (AdK)")
    R.plot(ax=ax, color='k', marker='s')
 
-Alternatively, if you wanted to plot the data yourself, the angles themselves
```
orbeckst: Still mention that the data is in `:attr:angles` – most people will want to use it.

richardjgowers:
Iirc it has a capital L?

package/MDAnalysis/analysis/dihedrals.py
```diff
@@ -45,39 +45,74 @@
...
-   fig, ax = plt.subplots(figsize=plt.figaspect(1))
-   ax.axis([-180,180,-180,180])
-   ax.axhline(0, color='k', lw=1)
-   ax.axvline(0, color='k', lw=1)
-   ax.set(xticks=range(-180,181,60), yticks=range(-180,181,60),
-          xlabel=r"$\phi$ (deg)", ylabel=r"$\psi$ (deg)")
-   for ts in R.angles:
-       ax.scatter(ts[:,0], ts[:,1], color='k', marker='s')
+Reference plots can be added to the axes for both the Ramachandran and Janin
+classes using the kwarg ``ref=True``. These were made using data obtained from
```
orbeckst: Cool!

Can you generate example plots for Ramachandran and Janin so that we can (eventually) add them to the docs? I can't quite remember how best to do it... I did it for the [MDA tutorial](https://www.mdanalysis.org/MDAnalysisTutorial/trajectories.html). 

Just create a directory `package/doc/sphinx/source/images`, add them there, and we'll figure out the rest later.

package/MDAnalysis/analysis/dihedrals.py
```diff
@@ -45,39 +45,74 @@
...
+----------------
+
+.. autoclass:: Dihedral
+   :members:
+   :inherited-members:
+
+   .. attribute:: angles
+
+       Contains the time steps of the angles for each atomgroup in the list as
+       an n_frames×len(atomgroups) :class:`numpy.ndarray` with content
```
orbeckst: ooooh... proper typography!

orbeckst: Put the variables in double-back ticks
```..rest
``n_frames×len(atomgroups)``
```
so that they get code formatting.

(Still ❤️ ing the ×)

package/MDAnalysis/analysis/dihedrals.py
```diff
@@ -89,6 +124,16 @@
        an n_frames×n_residues×2 :class:`numpy.ndarray` with content
        ``[[[phi, psi], [residue 2], ...], [time step 2], ...]``.
 
+.. autoclass:: Janin
+   :members:
+   :inherited-members:
+
+   .. attribute:: angles
+
+       Contains the time steps of the phi and psi angles for each residue as
```
orbeckst: chi1 and chi2

Btw, you can use proper Greek characters with LaTeX-like formatting
```reST
contains :math:`\chi_` and :math:`\chi_2` angles
```

package/MDAnalysis/analysis/dihedrals.py
```diff
@@ -101,13 +146,63 @@
 import MDAnalysis as mda
 from MDAnalysis.analysis.base import AnalysisBase
 from MDAnalysis.lib.distances import calc_dihedrals
+from MDAnalysisTests.datafiles import Rama_ref, Janin_ref
```
orbeckst: We can't reference MDAnalysisTests in MDAnalysis because people might not have the tests installed.

Instead, for right now, create a directory ``analysis/data`` and deposit your files there.

You will also need to make sure that `package/setup.py` actually includes your data files when you do a
```
python setup.py sdist
```

Have a look at the `testsuite/setup.py` for how to include datafiles https://github.com/MDAnalysis/mdanalysis/blob/10fb62d95f52e7b6b77870a53a35e7adbc477658/testsuite/setup.py#L183 with the `package_data` keyword.

package/MDAnalysis/analysis/dihedrals.py
orbeckst: Add a comment to the code, summarizing what you explained above.

package/MDAnalysis/analysis/dihedrals.py
orbeckst: Please also add your comment above as a comment in the code.

It will be helpful to anyone else coming after you.

package/MDAnalysis/analysis/dihedrals.py
```diff
@@ -193,6 +303,111 @@ def plot(self, ax=None, **kwargs):
...
+        ax : :class:`matplotlib.axes.Axes`
+             Axes with the plot, either `ax` or the current axes.
+
+        """
+        if ax is None:
+            ax = plt.gca()
+        ax.axis([0, 360, 0, 360])
+        ax.axhline(180, color='k', lw=1)
+        ax.axvline(180, color='k', lw=1)
+        ax.set(xticks=range(0,361,60), yticks=range(0,361,60),
```
orbeckst: PEP8 formatting: space after comma

testsuite/MDAnalysisTests/datafiles.py
```diff
@@ -156,7 +156,10 @@
     "legacy_DCD_c36_coords", # frames 1 and 4 read in for tip125_tric_C36.dcd using legacy DCD reader
     "GSD",
     "GRO_MEMPROT", "XTC_MEMPROT", # YiiP transporter in POPE:POPG lipids with Na+, Cl-, Zn2+ dummy model without water
-    "DihedralsArray", "GLYDihedralsArray" # phi and psi angles for testing Ramachandran class
+    "DihedralArray", "DihedralsArray", # time series of single dihedral
+    "RamaArray", "GLYRamaArray", # time series of phi/psi angles
+    "JaninArray", "LYSJaninArray", # time series of chi1/chi2 angles
+    "Rama_ref", "Janin_ref" # data for allowed and marginally allowed regions
```
orbeckst: Needs to go into MDAnalysis (see above).

testsuite/MDAnalysisTests/datafiles.py
```diff
@@ -417,8 +420,15 @@
...
-DihedralsArray = resource_filename(__name__, 'data/adk_oplsaa_dihedrals.npy')
-GLYDihedralsArray = resource_filename(__name__, 'data/adk_oplsaa_GLY_dihedrals.npy')
+DihedralArray = resource_filename(__name__, 'data/adk_oplsaa_dihedral.npy')
+DihedralsArray = resource_filename(__name__, 'data/adk_oplsaa_dihedral_list.npy')
+RamaArray = resource_filename(__name__, 'data/adk_oplsaa_rama.npy')
+GLYRamaArray = resource_filename(__name__, 'data/adk_oplsaa_GLY_rama.npy')
+JaninArray = resource_filename(__name__, 'data/adk_oplsaa_janin.npy')
+LYSJaninArray = resource_filename(__name__, 'data/adk_oplsaa_LYS_janin.npy')
+
+Rama_ref = resource_filename(__name__, 'data/rama_ref_data.npy')
```
orbeckst: Needs to go into MDAnalysis (see above).

testsuite/MDAnalysisTests/datafiles.py
```diff
@@ -417,8 +420,15 @@
...
+Janin_ref = resource_filename(__name__, 'data/janin_ref_data.npy')
```
orbeckst: Needs to go into MDAnalysis (see above).

orbeckst: Can you also prepare references for each amino acid separately? I think that this would be more interesting than all sidechains averaged. The peptide backbone is not strongly dependent on the chemical identity of the residue (GLY, PRO are exceptions) but the Janin plots should depend strongly on the nature of the sidechain.

testsuite/MDAnalysisTests/analysis/test_dihedrals.py
```diff
@@ -71,7 +109,61 @@ def test_protein_ends(self, universe):
         with pytest.warns(UserWarning):
             rama = Ramachandran(universe.select_atoms("protein")).run()
 
+    def test_None_removal(self):
+        with pytest.warns(UserWarning):
+            u = mda.coordinates.MMTF.fetch_mmtf('19hc')
```
orbeckst: What happens if `fetch_mmtf()` cannot download the PDB? You should either XFAIL the test if this happens or include a gzipped 19hc.pdb.gz as a data file.

testsuite/MDAnalysisTests/analysis/test_dihedrals.py
```diff
@@ -71,7 +109,61 @@ def test_protein_ends(self, universe):
...
+
+class TestJanin(object):
+
+    @pytest.fixture()
+    def universe(self):
+        return mda.Universe(GRO, XTC)
+
+    def test_janin(self, universe):
+        janin = Janin(universe.select_atoms("protein")).run()
+        test_janin = np.load(JaninArray)
```
orbeckst: make the `test_janin` a class-level fixture so that it only gets loaded once:
```python


class ....:

   @pytest.fixture
   def janin_ref_array(self):
         return np.load(JaninArray)

   def test_janin(self, universe, janin_ref_array):
         janin = Janin(universe.select_atoms("protein")).run()
         assert_almost_equal(janin.angles, janin_ref_array, 4, ...)
```

testsuite/MDAnalysisTests/analysis/test_dihedrals.py
```diff
@@ -40,7 +78,7 @@ def universe(self):
 
     def test_ramachandran(self, universe):
         rama = Ramachandran(universe.select_atoms("protein")).run()
-        test_rama = np.load(DihedralsArray)
+        test_rama = np.load(RamaArray)
```
orbeckst: make it a fixture (see comments below for `JaninArray` – this is the common way to treat "reference data" in pytest

testsuite/MDAnalysisTests/analysis/test_dihedrals.py
```diff
@@ -71,7 +109,61 @@ def test_protein_ends(self, universe):
...
+
+    @pytest.fixture()
+    def universe(self):
+        return mda.Universe(GRO, XTC)
+
+    def test_janin(self, universe):
+        janin = Janin(universe.select_atoms("protein")).run()
+        test_janin = np.load(JaninArray)
+
+        assert_almost_equal(janin.angles, test_janin, 4,
```
orbeckst: Is `decimal=4` sufficient to let the tests pass on Linux _and_ OSX?

orbeckst: Add a one line comment for the justification for reducing the test precision.

package/MDAnalysis/analysis/dihedrals.py
```diff
@@ -199,27 +211,28 @@ def _conclude(self):
...
 
 class Ramachandran(AnalysisBase):
-    """Calculate phi and psi dihedral angles of selected residues.
+    """Calculate :math:`\phi` and :math:`\psi` dihedral angles of selected residues.
 
-    Phi and psi angles will be calculated for each residue corresponding to
-    `atomgroup` for each time step in the trajectory. A :class:`~MDAnalysis.ResidueGroup`
-    is generated from `atomgroup` which is compared to the protein to determine
-    if it is a legitimate selection.
+    :math:`\phi` and :math:`\psi` angles will be calculated for each residue \
```
orbeckst: backslash not needed at end of line

package/MDAnalysis/analysis/dihedrals.py
```diff
@@ -120,19 +123,28 @@ class which takes a list of `atomgroups`, this class only needs a list of
...
    .. attribute:: angles
 
-       Contains the time steps of the phi and psi angles for each residue as
-       an n_frames×n_residues×2 :class:`numpy.ndarray` with content
-       ``[[[chi1, chi2], [residue 2], ...], [time step 2], ...]``.
+       Contains the time steps of the :math:`\chi_1` and :math:`\chi_2` angles
+       for each residue as an ``n_frames×n_residues×2`` :class:`numpy.ndarray`
+       with content ``[[[chi1, chi2], [residue 2], ...], [time step 2], ...]``.
+
+References
```
orbeckst: good

package/MDAnalysis/analysis/dihedrals.py
orbeckst: @richardjgowers was right
```python
>>> import MDAnalysis as mda
>>> u = mda.fetch_mmtf('19hc')
>>> u.atoms.altLocs
array(['', '', '', ..., '', '', ''], dtype=object)
>>> set(u.atoms.altLocs)
{'', 'A', 'B'}
# select no A or B
>>> u.atoms[u.atoms.altLocs == '']
<AtomGroup with 5944 atoms>
>>> u.atoms
<AtomGroup with 6098 atoms>
```
The default altLoc indicator is the empty string.

package/MDAnalysis/analysis/data/filenames.py
```diff
@@ -0,0 +1,35 @@
...
+# D. L. Dotson, J. Domanski, S. Buchoux, I. M. Kenney, and O. Beckstein.
+# MDAnalysis: A Python package for the rapid analysis of molecular dynamics
+# simulations. In S. Benthall and S. Rostrup editors, Proceedings of the 15th
+# Python in Science Conference, pages 102-109, Austin, TX, 2016. SciPy.
+#
+# N. Michaud-Agrawal, E. J. Denning, T. B. Woolf, and O. Beckstein.
+# MDAnalysis: A Toolkit for the Analysis of Molecular Dynamics Simulations.
+# J. Comput. Chem. 32 (2011), 2319--2327, doi:10.1002/jcc.21787
+#
+
```
orbeckst: Please add brief documentation:
```rest
Analysis data files
------------------

.. data:: Rama_ref
   Reference Ramachandran histogram for :class:`MDAnalysis.analysis.dihedrals.Ramachandran`. 
   The data were calculated on a data set of 500 PDB structures taken from [Lovell2003]_.

.. data:: Janin_ref
   ...
```

You are starting this data module so you should set a good example for documentation in the future.

orbeckst: Insert the docs for `analysis.docs.data.filenames` into the sphinx docs as you did for dihedrals. Or ping me an I will add it when you have done the docs inside filenames.py and dihedrals.py.

package/MDAnalysis/analysis/dihedrals.py
```diff
@@ -45,49 +45,106 @@
...
-   fig, ax = plt.subplots(figsize=plt.figaspect(1))
-   ax.axis([-180,180,-180,180])
-   ax.axhline(0, color='k', lw=1)
-   ax.axvline(0, color='k', lw=1)
-   ax.set(xticks=range(-180,181,60), yticks=range(-180,181,60),
-          xlabel=r"$\phi$ (deg)", ylabel=r"$\psi$ (deg)")
-   for ts in R.angles:
-       ax.scatter(ts[:,0], ts[:,1], color='k', marker='s')
+Reference plots can be added to the axes for both the Ramachandran and Janin
+classes using the kwarg ``ref=True``. These were made using data obtained from
```
orbeckst: Replace "These were made using" with
```rest
The Ramachandran reference data (:data:`~MDAnalysis.analysis.data.filenames.Rama_ref`) and 
Janin reference data (:data:`~MDAnalysis.analysis.data.filenames.Janin_ref`) were made ...
```

It's good to put in cross links between the docs.

package/MDAnalysis/analysis/data/filenames.py
```diff
@@ -0,0 +1,35 @@
+# -*- Mode: python; tab-width: 4; indent-tabs-mode:nil; coding:utf-8 -*-
```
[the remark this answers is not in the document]
orbeckst: The absence of the file is the reason for the test failures.

package/MDAnalysis/analysis/dihedrals.py
```diff
@@ -45,49 +45,106 @@
...
+
    import matplotlib.pyplot as plt
    fig, ax = plt.subplots(figsize=plt.figaspect(1))
-   ax.set_title("Ramachandran Plot (AdK)")
    R.plot(ax=ax, color='k', marker='s')
 
-Alternatively, if you wanted to plot the data yourself, the angles themselves
-can be accessed using :attr:`Ramachandran.angles`::
+The Janin class works in the same way, only needing a list of residues. To plot
+the data yourself, the angles can be accessed using :attr:`Ramachandran.angles`.
```
orbeckst: or :attr:`Janin.angles`

testsuite/MDAnalysisTests/datafiles.py
```diff
@@ -85,7 +85,7 @@
     "XTC_sub_sol",
     "XYZ", "XYZ_psf", "XYZ_bz2",
     "XYZ_mini", "XYZ_five", # 3 and 5 atoms xyzs for an easy topology
-    "TXYZ", "ARC", "ARC_PBC",        # Tinker files
```
orbeckst: I am not sure why ARC_PBC was removed. This might have been a merge accident because it is present in current develop. *Please add it back!* The line should be
```python
 "TXYZ", "ARC", "ARC_PBC",        # Tinker files
```

testsuite/MDAnalysisTests/datafiles.py
```diff
@@ -294,7 +297,6 @@
 XYZ_five = resource_filename(__name__, 'data/five.xyz')
 TXYZ = resource_filename(__name__, 'data/coordinates/test.txyz')
 ARC = resource_filename(__name__, 'data/coordinates/test.arc')
-ARC_PBC = resource_filename(__name__, 'data/coordinates/new_hexane.arc')
```
orbeckst: Add this line back in
```python
ARC_PBC = resource_filename(__name__, 'data/coordinates/new_hexane.arc')
```

testsuite/MDAnalysisTests/datafiles.py
orbeckst: Make sure that you pull from your branch – maybe you didn't get the merge commit?

package/MDAnalysis/analysis/dihedrals.py
orbeckst: I raised #2044. @hfmull if you can add anything to the issue report (e.g., example PDB ID where the current chi1_angles() gives wrong results) then that would be great!

orbeckst:
One test failed but not because of anything you did (but more likely #1988). I restarted the test and hopefully everything will light up "passing".

I am approving the PR and then it can be merged whenever Travis is happy.

Excellent work, @hfmull !

orbeckst:
FYI: @hfmull wrote a short report on his REU project https://figshare.com/articles/Technical_Report_SPIDAL_Summer_REU_2018_Dihedral_Analysis_in_MDAnalysis/6957296

orbeckst:
@hfmull FYI: I benchmarked your *Dihedral* analysis class against the naive approach in Gist https://gist.github.com/orbeckst/26081375f3ea3152f08bbcc90c14c5eb and your approach is between 80 and 100 times faster for a typical protein. Good job!

PR #2034, opened 2018-08-09
BUG: improve Windows PDB file handling

tylerjereddy:
* patch Windows file PDB handling behavior
such that 9 unit tests that were either
failing or erroring (currently xfailed
for Windows) now pass

* this should reduce appveyor
xfails from `21` to `12`, with no
errors or real failures

* I'm admittedly a little uncomfortable
with all the tweaks needed for PDB
file handling on Windows, but the seeking
around really isn't that portable as
previously discussed; my main concern
is that I'm micro-adjusting such
that unit tests pass--hopefully our tests
are sufficiently diverse that I can't
overspecialize the windows handling
such that it works on our tests
but not in the wild

### DIFF
--- a/package/MDAnalysis/coordinates/PDB.py
+++ b/package/MDAnalysis/coordinates/PDB.py
@@ -316,6 +316,11 @@ class PDBReader(base.ReaderBase):
             models.append(0)
         if len(crysts) == len(models):
             offsets = [min(a, b) for a, b in zip(models, crysts)]
+            if len(offsets) == 1 and offsets[0] < 0:
+                    err_msg = '''This should ONLY happen on
+                                 Windows; see PR#2034.'''
+                    assert os.name == 'nt', err_msg
+                    offsets[0] = 0
         else:
             offsets = models
         # Position of the start of each frame
--- a/testsuite/MDAnalysisTests/core/test_atomselections.py
+++ b/testsuite/MDAnalysisTests/core/test_atomselections.py
@@ -344,9 +344,6 @@ class TestSelectionsAMBER(object):
         assert_equal(sel.names, ['HH31', 'HH32', 'HH33', 'HB1', 'HB2', 'HB3'])
 
 
-@pytest.mark.xfail(os.name == 'nt',
-                   strict=True,
-                   reason="Not supported on Windows yet.")
 class TestSelectionsNAMD(object):
     @pytest.fixture()
     def universe(self):
--- a/testsuite/MDAnalysisTests/utils/test_modelling.py
+++ b/testsuite/MDAnalysisTests/utils/test_modelling.py
@@ -141,9 +141,6 @@ def u_water():
 
 
 class TestMerge(object):
-    @pytest.mark.xfail(os.name == 'nt',
-                       strict=True,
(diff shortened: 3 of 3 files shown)


tylerjereddy:
Tests pass but codecov doesn't like the patch diff since the new code path is only explored on appveyor tests, which don't yet have codecov reporting.

Perhaps I can open a separate PR to add codecov reporting to appveyor, when time permits.

orbeckst:
There's a blog post on [CodeCov Test Coverage Integration](https://www.appveyor.com/blog/2017/03/17/codecov/) – my understanding of Codecov is that it merges reports so it would be good to get coverage of the sum of our different code paths.

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -316,6 +316,11 @@ def __init__(self, filename, **kwargs):
             models.append(0)
         if len(crysts) == len(models):
             offsets = [min(a, b) for a, b in zip(models, crysts)]
+            if len(offsets) == 1:
```
orbeckst: If this is a Windows-specific hack, should we be even more specific and say
```python
if len(offsets) == 1 and offsets[0] < 0:
   assert os.name == 'nt', "This should ONLY happen on Windows, see PR #2034"
   # happens with windows_overbyte in some cases
   offsets[0] = 0
```

Btw, you can put both statements in the same `if` because the `and` is short circuited: even `offsets = []` will work because `offsets[0]` is never evaluated.

richardjgowers:
I still don't really understand how Windows doesn't have bytes, but I don't have a computer I could play around with this on.  But yeah, PDB isn't the only Reader using that method, so it's likely that this bug still exists elsewhere.

One thing we could do would be to rewrite the seeking to use line counts (which *should* be the same on Windows).  I just think this will be slower, but at least it might be 100% reliable.

tylerjereddy:
Ok, revised accordingly; I'll open a separate issue for adding codecov on Appveyor / windows so I don't forget.

jbarnoud:
@tylerjereddy Out of curiosity, did you look at where the PDB reader was yielding negative offsets? Could it come from use of unicode strings as if they where bytes, or from python's handling of end of lines?

The former (unicode strings) seems to be implicitly assumed: from what I understand, the band aid only applies if `CRYST1` is set for each model.

What about decoding `line` before the `len` in https://github.com/MDAnalysis/mdanalysis/blob/c3de7a4b612f793508b9213578f52d2175432c6a/package/MDAnalysis/coordinates/PDB.py#L298

tylerjereddy:
@jbarnoud The negative probably happens because of the previous bandaid to fix other windows stuff with `windows_overbyte`; there is a PR or issue somewhere that links to Python standard lib docs where Windows behavior for the seek() is described as unreliable.

tylerjereddy:
I can't remember if it is because the file is opened in text mode & then seeking is used or something like that--if there's a dev consensus to explore another approach I'm happy to do that.

tylerjereddy:
I think this can be safely closed -- as I understand it we fully support PDB files on Windows, especially now that we are not concerned with Python 2.7.

Issue #782, opened 2016-03-18
remove flag registry


orbeckst:
**tl;dr:** _Get rid of MDAnalysis.core.flags and replace with the lightest solution possible_

The _flag registry_ in [MDAnalysis.core.flags](http://pythonhosted.org/MDAnalysis/documentation_pages/core/init.html#flags) (in `MDAnalysis/core/__init__.py`)  determines global behavior. In principle, these flags can be changed by the user but in practice this seems to be rarely done. In any case, changes are not persistent and have to be changed for a new python session manually by setting the flag.
# Action: remove MDAnalysis.core.flags

As discussed in #315 we decided that the flag registry in the current form does not serve a useful purpose and is too complicated for what it accomplishes. It should be reduced to the smallest and simplest data structure that is necessary, such as a simple dictionary or variables at the module level.

In particular there is currently no expectation that the user should be able to change any parameters in MDAnalysis that change the global behavior (such as changing the base units) because it adds additional assumptions to code and reduces portability.
# Impact

Flags are used throughout the code to set defaults, e.g., in the readers and writers and some of the selections. All of this has to be changed.

Issues are marked with label [Component-Configuration](https://github.com/MDAnalysis/mdanalysis/labels/Component-Configuration). In particular
- [x] #759 Periodic/Nonperiodic keywords to selections
- [ ] #104 Managing default selection keywords
- [ ] #596 Using physical units in AtomGroups

richardjgowers:
Hopefully we can just replace where flags offered options with a kwarg to the function

orbeckst:
Yes, I hope so, too, but I didn't want to be overly prescriptive without properly reviewing the code in detail.

jbarnoud:
Do we want to go through a transition phase with deprecation warnings? Or do we want to brutally remove the flags?

I would go for the deprecation warnings. For a version or two, the flag replacement could be mirrors to the flags by default. The flags mechanism could then remain available, but with a deprecation warning. I do not know if these flags are actually used; I never used them, but they control some critical mechanisms such as using periodic boundary conditions or not.

orbeckst:
On 18 Mar, 2016, at 17:16, Jonathan Barnoud wrote:

> Do we want to go through a transition phase with deprecation warnings? Or do we want to brutally remove the flags?
> 
> I would go for the deprecation warnings. For a version or two, the flag replacement could be mirrors to the flags by default. The flags mechanism could then remain available, but with a deprecation warning.

[...]

Issue #1721, opened 2017-12-05
asv: refine tests with PBC/no-PBC


orbeckst:
Follow-up on PR  #1695, which included initial tests for distance based calculations. The comments noted a few issues
* use of core.flags https://github.com/MDAnalysis/mdanalysis/pull/1695#discussion_r148338963
* clear separation of tests with and without PBC https://github.com/MDAnalysis/mdanalysis/pull/1695#issuecomment-341405236

The asv benchmarks, specifically in [benchmarks/analysis/distances.py](https://github.com/MDAnalysis/mdanalysis/blob/develop/benchmarks/benchmarks/analysis/distances.py) should be refined to address the above comments.

soaebhasan12:
Hi @orbeckst — I'd like to work on this as my GSoC 2026 starter contribution.

I've read PR #1695 and both review comments: the `core.flags` dependency and the lack of clean PBC/no-PBC separation in the timing methods. My plan:

1. Add `params = [True, False]` and `param_names = ['pbc']` to `DistancesBench` (and any related benchmark classes) in `benchmarks/benchmarks/analysis/distances.py`, so ASV records separate timings for the with-PBC and without-PBC code paths.
2. Replace the `core.flags` usage — pass `box=self.box` directly when `pbc=True` and `box=None` when `pbc=False`, which is the current idiomatic API.
3. Run `asv run --quick --no-pull` locally to confirm both parameter variants execute cleanly before opening the PR.

I have ASV set up locally and have read through the existing benchmark files. I am applying for GSoC 2026 Project 3 and this is my intended starter contribution. I will open the PR within the next day or two.

Please let me know if the approach above looks right or if there's anything specific you'd like handled differently — I want to make sure the implementation matches what you had in mind before writing the code.

orbeckst:
Hello @soaebhasan12 , you're welcome to contribute to MDAnalysis. But I want to let you know that you're too late for GSOC 2026. As we wrote in https://www.mdanalysis.org/2026/02/19/gsoc2026/ we require a multi-step process and the deadline to enter the process has passed (the first step was a submitting a pre-proposal and the deadline was March 9, 2026).

soaebhasan12:
Sir, i request you please review my proposal at least one time. I will be
very thankful to you...

soaebhasan12:
this is my proposal google drive link (docx file)
[...]

PR #2035, opened 2018-08-09
Using Capped distance in distance selections

### LINKED ISSUE #974: "around" selection is slow and memory-intensive

ayushsuhane:
Fixes #974 , part of #782

Increases the performance of 'around' selection using NSgrid.

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [X ] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -21,6 +21,9 @@ The rules for this file:
 
 Enhancements
 
+  * Replaced multiple apply methods in distance based selections with
+    lib.distances.capped_distance for automatic selection of
+    distance evaluation method. (PR #2035)
   * Added return_distances argument in lib.distances.capped_distances
     to evaluate and return distances only when required. Modified
     the optimization rules in lib.distances._determine_method for
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -21,7 +21,8 @@ The rules for this file:
 
 Enhancements
 
-  * Replaced multiple apply methods in distance based selections with
+  * Replaced multiple apply (_apply_distmat, _apply_kdtree)
+    methods in distance based selections with
     lib.distances.capped_distance for automatic selection of
     distance evaluation method. (PR #2035)
   * Added return_distances argument in lib.distances.capped_distances
--- a/package/MDAnalysis/core/selection.py
+++ b/package/MDAnalysis/core/selection.py
@@ -239,11 +239,6 @@ class DistanceSelection(Selection):
      - _apply_distmat
     """
     def __init__(self):
-        
-        if flags['use_KDTree_routines'] in (True, 'fast', 'always'):
-            self.apply = self._apply_KDTree
-        else:
-            self.apply = self._apply_distmat
 
         self.periodic = flags['use_periodic_selections']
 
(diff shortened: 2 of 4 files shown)


ayushsuhane:
As benchmark for 'around' selection is already present, I compared the three methods (exactly same functions) for that specific case. It seems that the timings have improved by an order of magnitude.
(https://github.com/ayushsuhane/Benchmarks_Distance/blob/master/Notebooks/AroundBenchmark.ipynb)[https://github.com/ayushsuhane/Benchmarks_Distance/blob/master/Notebooks/AroundBenchmark.ipynb]
However, bear in mind, the improvement is for this case and not for every case. 

@orbeckst mentioned adding benchmarks. While this will be reflected if we run the benchmarks, should I add more cases? Maybe PBC/nonPBC as in here (Issue #1721 ). But due to lack of time, I guess I will pick it up after GSOC period is over.

ayushsuhane:
I am not sure how to add another method here mostly because of flags. Any suggestions to get around the problem without breaking the tests (and more specifically code). I thought its better to discuss here, how do we want to proceed. Do we want to keep earlier code and optimize the methods or do we want to replace with one or two using flags/keywords. As we are on the topic, I don't think using capped function for optimization (in its current form)  is suitable here, because it does extra computations of pairs/distances which we don't need in this case.

orbeckst:
This is looking promising already. 

We want to get rid of the flags #782 so there might not remain a way for the user to choose a method. Bearing this in mind, would you recommend to switch over to NSGrid (and take the performance hit in some cases)?

orbeckst:
> benchmarks: While this will be reflected if we run the benchmarks, should I add more cases? Maybe PBC/nonPBC as in here (Issue #1721 ). 

PBC/non-PBC would be excellent.

richardjgowers:
@ayushsuhane I think ideally we'd reuse `capped_distance_array`, then we could simplify this greatly.  The only reason I can think not to is that it calculates indices **and** distances, so maybe could be faster.  Is this why you didn't use that?

richardjgowers:
Ah just reread, you could add a `return_distances` kwargs (default True) to `capped_distance` and then it should be identical to this right?  It doesn't make sense to have so much duplicated code.

ayushsuhane:
@richardjgowers Yes. While this is the case with Periodic KDTree and distance array, it would not be a problem if we are using only capped_distance(.., .., method='nsgrid') as it returns all the distances and pairs, which serves the purpose here. 

However, yes, adding return distance to kwargs is a good idea. I will look into it

ayushsuhane:
@orbeckst  For this special case of around selections, it wouldn't hurt to take a hit and replacing with nsgrid. But I think a better approach would be to use capped_distance and optimize there. So that if anyone plan to change/modify the method later, one has to change only the capped_function, and not worry about its dependencies.

That being said, there are certain drawbacks in the capped_distance, which I wrote with different assumptions and needs to be changed. For instance, ``_bruteforce_capped`` checks distance with every atoms individually in a loop, which is a bad idea for a numpy array but it can handle arbitrary size of atoms. Now that we wouldn't want to use brute  force for arbitrary size and limit it only till its the fastest, it would be better to change it. Similarly for ``_pkdtree_capped`` where we have used for loops to get individual distances and pairs, which can/should be avoided.

ayushsuhane:
@richardjgowers One more thing. It seems importing any library inside the function, for instance ``from . import PeriodicKDTree`` inside ``_pkdtree_capped`` takes a large amount of time  which is reflected baldly in benchmarks. Any suggestions how can we avoid it. Also, I guess its a bad practice to use local imports, isn't it? I thought we just did it for the time being, but if we are looking for optimizations, it shouldn't really be there

ayushsuhane:
It seems importing doesn't take any significant time. My bad.

ayushsuhane:
I think after #2041, we can move to ``capped_distance`` directly. Here is an example of choosing the best method for  around selection based on the rules specified in capped_distance.

image

However, I have one concern in changing the codebase of around. It seems tests are written for different flags and methods for every class i.e. `` apply_kdtree``, ``apply_distmat``. So there are essentially two things, one is removal of flags, and another is using a single method instead of multiple ``_apply`` methods. What do you think should be done first. Any suggestions how should I go about it?

richardjgowers:
@ayushsuhane I think you can just rip the flags (& associated tests) out.  If someone chose `kdtree` in the flags and we don't respect that, they'll still get the same distances out, so it's not going to hurt anyone.

richardjgowers:
@ayushsuhane this needs a quick rebase right?

richardjgowers:
Looks solid, iirc the coverage for selections was complete, so if tests pass this is good to go

ayushsuhane:
@richardjgowers Not sure what is happening here. Can you take a look? I tried restarting the build but the appveyor still fails due to encore.

zemanj:
@ayushsuhane This looks like yet another random number issue to me... I just had the same test failing on my linux box, but it passed in a subsequent run. The failing test uses ``analysis.encore.bootstrap.get_ensemble_bootstrap_samples()`` with only 10 samples, so probably that's the culprit. Maybe just give it another try?

zemanj:
@orbeckst Ok I restarted the Travis build for this PR, but now my noob question: In how far is that related to AppVeyor?

zemanj:
@richardjgowers Ok I gotta remember that shortcut :wink:

richardjgowers:
@zemanj it’s unrelated :) I didn’t immediately see the rerun appveyor button on mobile so I turned the PR off then on again....

zemanj:
That's what i thought... I don't see any option to rerun a build on AppVeyor either. Not on mobile, though.

PR #2042, opened 2018-08-12
TST: add codecov to appveyor CI run

### LINKED ISSUE #2037: TST: include appveyor in codecov report
It would be helpful to include the code paths explored from Windows tests in the final codecov report by sending coverage info from Appveyor -- see #2034 for an example where this would help avoid CI coverage issues.

tylerjereddy:
Fixes #2037.

I tried to do something similar to the codecov reporting we do with Travis; the idea is to explore code paths that may be unique to  Windows for test reports -- see, for example #2034, where CI fails because we don't flush through the windows code paths with codecov yet.

This may take a few iterations  to get right with the CI I suppose.

I think the preferred way to invoke codecov is actually a little different form this now, but I figured I may as well just do what we do on Travis & see if that works.


orbeckst:
@tylerjereddy feel free to rebase+merge yourself. I am leaving the PR open in case you want to add something to it or someone else happens to spot something that I overlooked.

orbeckst:
I haven't heard anything else so let's roll with it.

PR #2048, opened 2018-08-14
Simplified code in `lib.distances` a bit

### LINKED ISSUE #2046: MDAnalysis.lib.distances needs rework
The documentation of the module ``MDAnalysis.lib.distances`` has several issues, whereof one has already been addressed by @xiki-tempula in issue #2004. I also see some points where code duplication could be reduced to make the code more DRY.

### Documentation issues:
* [ ] Probably deprecated title: *Fast distance array computation*
  IMHO that seems a bit misleading. Even though the module contains methods to compute distance arrays, it also includes methods for coordinate transformation, PBC handling, and for the calculation of angles and dihedrals.
* [x] The same issue applies to the doc header, which doesn't describe the module's contents sufficiently since it only refers to distance arrays. (Fixed via PR #2070)
* [x] The methods listed as ``.. autofunction: ...`` in the header appear as duplicates in the docs. Furthermore, the way the ``autofunction`` strings are formatted may not be very useful (nested square brackets without default values for kwargs). (Fixed via PR #2070)
* [x] Many docstrings violate the 80 character line length limit. (Fixed via PR #2070)
* [x] The methods ``augment_coordinates`` and ``undo_augment``, which are imported from ``MDAnalysis.lib._augment``, are missing from the docs. (Fixed via PR #2062)
* [x] Many docstrings have issues with spelling mistakes, missing or improperly referenced parameters and methods, wrong or missing data types, and potentially misleading explanations. That includes but is not limited to issue #2004. (Fixed via PR #2070)
* [x] In general, the docstrings could be much more standardized, there's no need to have different descriptions for parameters such as *box* which always have the same purpose or meaning. (Fixed via PR #2070)

### Code issues:
* [x] Many functions use the ``_box_check()`` helper function to determine the type of simulation box supplied. Thereafter, the box coordinates are transformed to the memory layout expected by the subsequently called low-level C functions. This transformation should be incorporated into ``_check_box()`` to avoid code duplication. (Fixed via PR #2048)
* [x] The docstrings of most functions state that the box must be supplied in the format ``[lx, ly, lz, alpha, beta, gamma]`` as returned by `Timestep.dimensions```. The ``_box_check`` function doesn't reflect that. So ~either the requirements for the box should be less strict or~ ``_box_check`` should be stricter in that respect. (Fixed via PR #2048)
* [x] Not only checking but also creating the results array (if required) should take place in the ``_check_result_array()`` helper function. (Fixed via PR #2048)
* [x] Many functions now incorporate automatic dtype conversion, so the dtype check in the ``_check_array()`` helper function is now redundant and can be removed in these cases. (Fixed via PR #2048 by means of a new `@check_coords()` decorator)
* [x] PR #2048 introduced a subtle bug so that in certain situations, some functions change their input coordinate arrays in-place. (Fixed via PR #2083)
* [x] Many functions choke on empty input coordinates arrays (i.e., with ``shape=(0, 3)``). (Fixed via PR #2083)
* [x] Depending on the employed search method, the results of `capped_distance()` and `self_capped_distance()` are not always numpy arrays. (Fixed via PR #2083)
* [x] If no pairs are found, `_bruteforce_capped_self()` correctly returns empty pairs but unfortunately also non-empty mumbo-jumbo distances. (Fixed via PR #2083)
* [x] `_bruteforce_capped()` crashes if all input coordinates are the same and `box` is ``None``. (Fixed via PR #2083)
* [x] `lib.distances._check_box()` should be moved to `lib.util` (with underscore removed). (Fixed via PR #2114)
* [x] Not all methods for `*capped_distance()` have the same cut-off criteria (sometimes `distances < max_cutoff` and sometimes `distances <= max_cutoff`). (Fixed via PR #2114)
* [x] The different methods for `*capped_distance()` do not always return the same number of pairs. (EDIT: only pathological cases, **won't fix**)
* [x] `lib.nsgrid` calculates distances in single precision, whereas all functions in lib.distances use double precision. (Fixed via PR #2114)
* [x] `lib.nsgrid.PBCBox` uses an arbitrary constant `EPSILON=1e-5` to determine whether a box is triclinic. This a) does not correspond to the behavior of other functions, and b) fails if a box angle is, e.g., 90.00001 degrees (or higher or negative). (Fixed via PR #2114)
* [x] `lib/include/calc_distances.h` contains a lot of duplicated code (functions differing in PBC type only). (EDIT: unifying functions for different PBC types impacts performance, **won't fix**)
* [ ] OpenMP parallelization in `lib/include/calc_distances.h` often suffers from [false sharing](https://en.wikipedia.org/wiki/False_sharing).
* [ ] In `lib.distances.distance_array()`, the inner loop (the one over the `configuration` coordinate array) should be parallelized instead of the outer one (going over the `reference` coordinate array), since often, one seeks to know the number of "neighbors" with respect to some reference atom.
* [ ] Tidy the namespace, make C functions invisible, currently we have everything visible

### TODO suggestions:
* [x] Documentation:
  * [x] Ensure that parameters referenced in docstrings correspond to function signatures. (Fixed via PRs #2048 and #2070)
  * [x] Properly format references to other methods so that links are generated. Add ~``.. seealso :``~ ``See Also`` sections where applicable. (Fixed via PR #2070)
  * [x] Use italics only to reference function parameters, use teletype font for variables and values in docstrings. (Fixed via PR #2070)
  * [x] Use logical instead of visual markup (e.g., backticks instead of asterisks for function parameters). (Fixed via PR #2070)
  * [x] Enforce 80 character lines (Fixed via PR #2070)
  * [x] Standardize phrasing of repeatedly occurring parameter descriptions (Fixed via PR #2070)
  * [x] Always state correct type of parameters (e.g. ``numpy.ndarray`` if a numpy array is expected) (Fixed via PRs #2048, #2062, and #2070)
  * [x] Mark optional parameters by adding *, optional* after the type (e.g. ``box : array_like, optional``) (Fixed via PR #2070)
  * [x] Do not give default values for kwargs in brackets since these appear in the auto-generated function signature. (Fixed via PR #2070)
  * [ ] Module Header:
    * [ ] Change title to something more descriptive (e.g., "fast geometry computations" or similar)
    * [x] Improve module description to better correspond to its contents (Fixed via PR #2070)
    * [x] Remove ~``.. autofunction`` lines~ `:members:` from `distances.rst` causing duplicates
* Code:
  * [x] Move *box* memory layout transformations to ``_check_box`` and ~either loosen requirements for its expected format or~ make checks stricter. (Fixed via PR #2048)
  * [x] Fix bugs introduced in PR #2048 and add tests assuring input arrays always remain unmodified. (Fixed via PR #2083)
  * [x] For completeness, make sure that all functions also work with empty coordinate arrays (``shape=(0, 3)``) and add corresponding test cases. (Fixed via PR #2083)
  * [x] Ensure `capped_distance()` and `self_capped_distance()` always return numpy arrays and add corresponding tests. (Fixed via PR #2083)
  * [x] Ensure `_bruteforce_capped_self()` returns empty distances if no pairs are found. Add corresponding tests for *all* `*capped_distance()` functions. (Fixed via PR #2083)
  * [x] Fix `_bruteforce_capped()` for the case when all input coordinates are the same and `box` is ``None``. (Fixed via PR #2083)
  * [x] Move `lib.distances._check_box()` to `lib.util` (with underscore removed). (Fixed via PR #2114)
  * [x] Ensure that all methods used in `capped_distance()` have the same cut-off criterion (`distances <= max_cutoff`). (Fixed via PR #2114)
  * [x] Use double precision for distance computation in `lib.nsgrid`. (Fixed via PR #2114)
  * [x] Fix box type detection in `lib.nsgrid` and some functions in `lib.distances`. (Fixed via PR #2114)
  * [ ] Optimize parallelism in `lib/include/calc_distances.h` for 64 byte cachelines (avoid false sharing).
  * [ ] Parallelize inner instead of outer loop in `lib.distances.distance_array()`.
  
That's quite a lot of things to do, but I've already started working on most of the points.
There are still issues to be discussed, especially the module's title ~and description and how to proceed with the requirements for boxes~.

### Current version of MDAnalysis:
~0.18.1-dev~ 0.19.1-dev

zemanj:
Fixes (partially) #2046, this is just the beginning of a series of related PRs.

### Changes made in this Pull Request:
* Renamed `_box_check()` -> `_check_box()` to be consistent with naming of other helper functions.
* `_check_box()` now also returns the box in the format expected by low-level C functions in `lib.c_distances`, which reduces code duplications.
* Removed now obsolete box type `'tri_vecs_bad'` from `_check_box()`.
* Renamed `_check_array()` -> `_check_coord_array()` to better reflect its purpose.
* Renamed `_check_results_array()` -> `_check_result_array()` since the checked parameter is always named `result`.
* Moved coord array shape checks to separate functions.
* `_check_result_array()` now returns a numpy array of correct shape and dtype if `result` is `None`, which simplifies the code of several functions.
* Improved docstrings of `transform_RtoS()` and `transform_StoR()` and added input coordinate shape check. Adjusted `test_augment.py` accordingly.
* Added shape check for input coordinates of `calc_distance()`, `calc_angle()`, and `calc_dihedral()`.

### Question:
* Before I can continue to fix the rest of the docs, I'd like to hear further opinions on how to proceed with the requirements for the box arrays. Should the documented requirements for the box format be less strict or should we rather make the checks in `_check_box()` stricter? 

PR Checklist
------------
 - [ ] Tests? (~Do we need additional tests?~ Additional tests needed)
 - [x] Docs? (More to do)
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/lib/distances.py
+++ b/package/MDAnalysis/lib/distances.py
@@ -128,96 +128,169 @@ from .c_distances import (calc_distance_array,
 from .c_distances_openmp import OPENMP_ENABLED as USED_OPENMP
 
 
-def _box_check(box):
-    """Take a box input and deduce what type of system it represents based
-    on the shape of the array and whether all angles are 90.
+def _check_box(box):
+    """Take a box input and deduce what type of system it represents based on
+    the shape of the array and whether all angles are 90 degrees.
 
     Parameters
     ----------
-    box : array
-        Box information of unknown format.
+    box : numpy.ndarray
+        Array with dtype ``numpy.float32`` containing box information of unknown
+        format.
 
     Returns
     -------
     boxtype : str
-        * ``ortho`` orthogonal box
-        * ``tri_vecs`` triclinic box vectors
-        * ``tri_box`` triclinic box lengths and angles
+        * ``'ortho'`` orthogonal box
+        * ``'tri_vecs'`` triclinic box vectors
+
+    checked_box : numpy.ndarray
+        Array of dtype ``numpy.float32`` containing box information:
+        * If `boxtype` is ``'ortho'``, `cecked_box` will have the shape ``(3,)``
+          containing the x-, y-, and z-dimensions of the orthogonal box.
+        * If  `boxtype` is ``'tri_vecs'``, `cecked_box` will have the shape
+          ``(3, 3)`` containing the triclinic box vectors in a lower triangular
+          matrix as returned by
+          :meth:`~MDAnalysis.lib.mdamath.triclinic_vectors`.
 
     Raises
     ------
     TypeError
-        If box is not float32.
+        If the dtype of `box` is not ``numpy.float32``.
     ValueError
-        If box type not detected.
+        If box type cannot be detected.
+
+
+    .. seealso: :meth:`~MDAnalysis.lib.mdamath.triclinic_vectors`
+    .. versionchanged: 0.19.0
+       * Added automatic conversion of input to :class:`numpy.ndarray` with
+         dtype ``numpy.float32``.
+       * Now also returns the box in the format expected by low-level functions
+         in :mod:`~MDAnalysis.lib.c_distances`.
+       * Removed obsolete `boxtype` ``tri_vecs_bad``.
     """
     if box.dtype != np.float32:
-        raise TypeError("Box must be of type float32")
-
-    boxtype = 'unknown'
-
-    if box.shape == (3,):
-        boxtype = 'ortho'
-    elif box.shape == (3, 3):
-        if np.all([box[0][1] == 0.0,  # Checks that tri box is properly formatted
-                      box[0][2] == 0.0,
-                      box[1][2] == 0.0]):
-            boxtype = 'tri_vecs'
-        else:
-            boxtype = 'tri_vecs_bad'
-    elif box.shape == (6,):
+        raise TypeError("Invalid box dtype. Must be numpy.float32, got {0}."
+                        "".format(box.dtype))
+    if box.shape == (6,):
         if np.all(box[3:] == 90.):
-            boxtype = 'ortho'
+            return 'ortho', box
         else:
-            boxtype = 'tri_box'
+            return 'tri_vecs', triclinic_vectors(box)
+    elif box.shape == (3,):
+        return 'ortho', box
+    elif box.shape == (3, 3):
+        # Check if triclinic box vectors are properly formatted:
+        if np.all([box[0][1] == 0.0, box[0][2] == 0.0, box[1][2] == 0.0]):
+            return 'tri_box', box
+        else:
+            return 'tri_box', triclinic_vectors(triclinic_box(box[0], box[1],
+                                                              box[2]))
+    raise ValueError("Box input not recognized, must be an array of box "
+                     "dimensions.")
 
-    if boxtype == 'unknown':
-        raise ValueError("box input not recognised"
-                         ", must be an array of box dimensions")
 
-    return boxtype
+def _check_coord_array(coords, desc):
+    """Check if an array is a valid coordinate array.
 
+    The `coords` array must meet the following requirements:
+      * Must have a shape of ``(n, 3)``.
+      * Its dtype must be ``numpy.float32``.
 
-def _check_array(coords, desc):
-    """Check an array is a valid array of coordinates
+    Parameters
+    ----------
+    coords : numpy.ndarray
+        The coordinate array to check for validity.
+    desc : str
+        Name of the coordinate array. Only used in error messages.
 
-    Must be:
-       (n,3) in shape
-       float32 data
+    Raises
+    ------
+    ValueError
+        If `coords` has a wrong shape.
+    TypeError
+        If the dtype of `coords` is not ``numpy.float32``.
     """
-    if (coords.ndim != 2 or coords.shape[1] != 3):
-        raise ValueError("{0} must be a sequence of 3 dimensional coordinates"
-                         "".format(desc))
-    _check_array_dtype(coords, desc)
-
-
-def _check_array_dtype(coords, desc):
-    """Check whether an array contains values of dtype: np.float32 or not"""
+    _check_coord_array_shape_2d(coords, desc)
     if coords.dtype != np.float32:
-        raise TypeError("{0} must be of type float32".format(desc))
+        raise TypeError("{0} must be of type numpy.float32".format(desc))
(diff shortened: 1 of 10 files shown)


richardjgowers:
Looks like the Mac tests are failing on the last decimal place of precision.  I think this is expected with a float32 triclinic box.

package/MDAnalysis/lib/distances.py
```diff
@@ -128,96 +128,184 @@ def _run(funcname, args=None, kwargs=None, backend="serial"):
...
+        If the dtype of `coords` is not ``numpy.float32``.
     """
-    if (coords.ndim != 2 or coords.shape[1] != 3):
-        raise ValueError("{0} must be a sequence of 3 dimensional coordinates"
-                         "".format(desc))
-    _check_array_dtype(coords, desc)
+    _check_coord_array_shape_2d(coords, desc)
+    if coords.dtype != np.float32:
+        raise TypeError("{0} must be of type numpy.float32".format(desc))
+# The following two lines would break a lot of tests. WHY?!
```
richardjgowers: What tests fail here?  Is it that some coords arrays are often non C contiguous?

richardjgowers:
@zemanj these changes look good, thanks!

package/MDAnalysis/lib/distances.py
zemanj: Ignore it, that's just a personal note I forgot to delete. Requiring coordinate arrays to be C-contiguous doesn't really make sense here since all the functions have calls to `np.astype(...)` which will return a continuous copy anyway.

richardjgowers:
@zemanj ok if you get the precision on the macos tests working we can merge this.  I'd rather chip away at 2046 in small parts to keep the PRs smaller.  I added a WIP tag, remove it once you're happy for us to pull the trigger on this PR.

zemanj:
@richardjgowers Thanks for taking a look! I'll add some more tests to increase coverage tomorrow and decrease the precision in the tests for `float32` to 6 decimals.

richardjgowers:
@zemanj yeah let's tidy box inputs to the first option.  If the docs are requesting the `.dimensions` format even better, as it's technically not an API break....

orbeckst:
I'm also in favor of tidy inputs. 

If we never documented the weaker requirements then you don't add a versionchanged marker to the docs but we should still have a note in CHANGELOG under changes saying that tidy box input is now consistently enforced. 

--
Oliver Beckstein
email: orbeckst@gmail.com

zemanj:
@richardjgowers Thanks for the merge, CHANGELOG and some additional tests are still missing, though.

PR #2036, opened 2018-08-10
additional PSA tests

orbeckst:
Changes made in this Pull Request:
- test functions that require seaborn (skip if cannot be imported)
- updated warning for plot_nearest_neighbors() to use run_pairs_analysis(neighbors=True)

PR Checklist
------------
 - [x] Tests?
 - n/a Docs?
 - n/a CHANGELOG updated?
 - n/a Issue raised/referenced?

Issue #1745, opened 2017-12-18
remove `save()` methods from analysis classes


orbeckst:
Many classes in MDAnalysis.analysis have a `save()` method as a convenience to persist calculated data such as a timeseries to disk. However, storing data structures to disk in a portable manner is difficult and typical approaches such as using Python pickle are not portable (not even between Python 2 and Python 3). Instead of having to worry about the details of storing data, we should just let the user decide how to store data (e.g., convert to pandas dataframes and let pandas handle all this, or write as a custom text format).

(This issue came up in discussion on PR #1744 (for issue #1743).) 

## Proposal

Remove `save()` methods from analysis classes if the method is only a convenience functions. If the saved files can be read in again by the class, e.g., for re-analysis, then a more detailed case-by-case evaluation should be performed.

EDIT: Generally accepted, I rewrote the list of cases below.

## approach

1. survey the list of save methods below: decide which ones should be removed and note here
2. add a *deprecation notice for 0.17.1* (once this is done, update the Milestone to 1.0 instead of closing)
   - docs (add a `.. deprecated:: 0.17.1` saying that the `save()` functionality will be removed in 1.0 because users generally know best how they want to persist the data.
   - `DeprecationWarning`
3. add docs that explain which attributes might be worthwhile to save to disk and show an example for how to to this (basically, adding the code from the `save()` method to the docs)
4. remove in 1.0

## cases
With `git grep 'save\w*(self'`): ~Add a checkmark for any method that should be removed~.

### To be removed 
Check methods once they have been removed.

 - [x] MDAnalysis/analysis/align.py: def save(self, rmsdfile): #2486
 - [x] MDAnalysis/analysis/contacts.py: def save(self, outfile): #2487
 - [x] MDAnalysis/analysis/diffusionmap.py: def save(self, filename): #2487
 - [x] MDAnalysis/analysis/hbonds/hbond_analysis.py: def save_table(self, filename="hbond_table.pickle"): #2494
 - [x] MDAnalysis/analysis/hbonds/hbond_autocorrel.py: def save_results(self, filename='hbond_autocorrel'): #2494
 - [x] MDAnalysis/analysis/hbonds/wbridge_analysis.py: def save_table(self, filename="wbridge_table.pickle"):
 - [x] MDAnalysis/analysis/hole.py: def save(self, filename="hole.pickle"): #2487
 - [x] MDAnalysis/analysis/lineardensity.py: def save(self, description='', form='txt'): #2487
[...]

PR #2049, opened 2018-08-14
more PSA tests

### LINKED ISSUE #1006: increase code coverage in PSA
Hi all,
One issue with code coverage in PSA is the amount of code dedicated to plotting the results of the path similarity analysis. I'd like to get people's opinions on whether this code should exist to as is, or if we should consider moving it to an example notebook outside of the class. In my limited experience with the analysis modules, it seems like the place for demonstrating how to aggregate and plot the data should be documentation, and not the actual code. There are around 300/1925 lines dedicated to plotting in the PSA module.

orbeckst:
Fixes #1006 

 - add more tests for code in analysis.psa (on the heels of merged PR #2036 )
 - NOTE: load/save fails --- maybe a bug??

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/testsuite/MDAnalysisTests/analysis/test_psa.py
+++ b/testsuite/MDAnalysisTests/analysis/test_psa.py
@@ -134,6 +134,54 @@ class TestPSAnalysis(object):
         err_msg = "Frechet distances did not increase after path reversal"
         assert frech_matrix[1,2] >= frech_matrix[0,1], err_msg
 
+    def test_get_num_paths(self, psa):
+        assert psa.get_num_paths() == 3
+
+    def test_get_paths(self, psa):
+        paths = psa.get_paths()
+        assert len(paths) == 3
+        assert isinstance(paths, list)
+
+    def test_psa_pairs_ValueError(self, psa):
+        with pytest.raises(ValueError):
+            psa.psa_pairs
+
+    def test_psa_pairs(self, psa):
+        psa.run_pairs_analysis()
+        assert len(psa.psa_pairs) == 3
+
+    def test_hausdorff_pairs_ValueError(self, psa):
+        with pytest.raises(ValueError):
+            psa.hausdorff_pairs
+
+    def test_hausdorff_pairs(self, psa):
+        psa.run_pairs_analysis(hausdorff_pairs=True)
+        assert len(psa.hausdorff_pairs) == 3
+
+    def test_nearest_neighbors_ValueError(self, psa):
+        with pytest.raises(ValueError):
+            psa.nearest_neighbors
+
+    def test_nearest_neighbors(self, psa):
+        psa.run_pairs_analysis(neighbors=True)
+        assert len(psa.nearest_neighbors) == 3
+
+    @pytest.mark.xfail
+    def test_load(self, psa):
+        """Test that the automatically saved files can be loaded"""
+        expected_path_names = psa.path_names[:]
+        expected_paths = [p.copy() for p in psa.paths]
+        psa.save_paths()
+        psa.load()
+        assert psa.path_names == expected_path_names
+        # manually compare paths because
+        #         assert_almost_equal(psa.paths, expected_paths, decimal=6)
+        # raises a ValueError in the assertion code itself
+        assert len(psa.paths) == len(expected_paths)
+        for ipath, (observed, expected) in enumerate(zip(psa.paths, expected_paths)):
+            assert_almost_equal(observed, expected, decimal=6,
+                                err_msg="loaded path {} does not agree with input".format(ipath))
+
     def test_dendrogram_produced(self, plot_data):
         """Test whether Dendrogram dictionary object was produced"""
         err_msg = "Dendrogram dictionary object was not produced"
--- a/package/MDAnalysis/analysis/psa.py
+++ b/package/MDAnalysis/analysis/psa.py
@@ -213,7 +213,7 @@ Classes, methods, and functions
 from __future__ import division, absolute_import, print_function
 
 import six
-from six.moves import range, cPickle
+from six.moves import range, cPickle, zip
 from six import string_types
 
 import os
@@ -1352,10 +1352,10 @@ class PSAnalysis(object):
 
         # Set default directory names for storing topology/reference structures,
         # fitted trajectories, paths, distance matrices, and plots
-        self.datadirs = {'fitted_trajs' : '/fitted_trajs',
-                         'paths' : '/paths',
-                         'distance_matrices' : '/distance_matrices',
-                         'plots' : '/plots'}
+        self.datadirs = {'fitted_trajs' : 'fitted_trajs',
+                         'paths' : 'paths',
+                         'distance_matrices' : 'distance_matrices',
+                         'plots' : 'plots'}
         for dir_name, directory in six.iteritems(self.datadirs):
             try:
                 full_dir_name = os.path.join(self.targetdir, dir_name)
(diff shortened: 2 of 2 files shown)


orbeckst:

testsuite/MDAnalysisTests/analysis/test_psa.py
```diff
@@ -134,6 +134,54 @@ def test_reversal_frechet(self, frech_matrix):
...
+
+    def test_nearest_neighbors_ValueError(self, psa):
+        with pytest.raises(ValueError):
+            psa.nearest_neighbors
+
+    def test_nearest_neighbors(self, psa):
+        psa.run_pairs_analysis(neighbors=True)
+        assert len(psa.nearest_neighbors) == 3
+
+    @pytest.mark.xfail
```
orbeckst: I think I marked it as failure because I don't know why they fail... this needs to be investigated.

orbeckst: @sseyler do you have an idea why this test might be failing?

sseyler: @orbeckst Sorry, I haven't been checking my GitHub messages very often since I've moved to BitBucket. What's the status of the test?

orbeckst: Nothing changed since I posted the message. I'd like you to look at the code https://github.com/MDAnalysis/mdanalysis/pull/2049/files#diff-d7b0118128e4a5cf1dad1bda8bde2f12R170 and tell me if my assumption that the input paths and the loaded paths should be the same is flawed. Or if not, why would this test fail?

As part of #1745 you wanted to keep these saved paths (https://github.com/MDAnalysis/mdanalysis/issues/1745#issuecomment-352904322), so now we have to test that they are actually written correctly.

sseyler: I just manually ran through the same failed test but had no problem saving and loading the trajectories; both `assert`s passed. I'm not sure what the issue is exactly, but I noticed that the paths that are actually loaded by `PSAnalysis.load()` are actually *not* the fitted paths as stated in the method's docs.

It looks like I need to do several things:
- [ ] Adjust the `PSAnalysis.load()` method to use a keyword to optionally load the original and/or fitted paths. The original paths will probably be what I'll have loaded by default and the user will have an option to load fitted trajectories (unless you're opposed to this for any reason).
- [ ] Adjust the `PSAnalysis.save_paths()` method to use a keyword to optionally save the original and/or fitted paths. The default option should match the default option that will be used for `PSAnalysis.load()`, although it could also just load the original paths by default and then, by default, also attempt to load any fitted paths it finds.
- [ ] Decide whether 
- [ ] Possibly change `PSAnalysis.load()` to `PSAnalysis.load_paths()`. This breaks backwards compatibility, but is consistent with `PSAnalysis.save_paths()`, is clearer about what's actually being loaded, and eliminates ambiguity given that the `PSAnalysis.save()` method is being removed.

orbeckst: Thanks for the analysis.

If you want to change the API then _now_ is a good time. First we would add deprecations (0.19.1 would be ideal) and then change and remove in 0.20.0 (or 1.0?)

As a general rule, simple methods/functions that are doing one thing are better than ones that do wildly different ones.

orbeckst:
Discussion regarding `test_load()` from https://github.com/MDAnalysis/mdanalysis/pull/2049#discussion_r227519989:

> I just manually ran through the same failed test but had no problem saving and loading the trajectories; both asserts passed.

When I run the test locally (and not allow xfail)
```
pytest --disable-warnings  MDAnalysisTests/analysis/test_psa.py::TestPSAnalysis::test_load
```
I get
```
MDAnalysisTests/analysis/test_psa.py F                                                                                        [100%]

============================================================= FAILURES ==============================================================
_____________________________________________________ TestPSAnalysis.test_load ______________________________________________________

self = <MDAnalysisTests.analysis.test_psa.TestPSAnalysis object at 0x1175f4358>
psa = <MDAnalysis.analysis.psa.PSAnalysis object at 0x117d84ac8>

    def test_load(self, psa):
        """Test that the automatically saved files can be loaded"""
        expected_path_names = psa.path_names[:]
        expected_paths = [p.copy() for p in psa.paths]
        psa.save_paths()
        psa.load()
>       assert psa.path_names == expected_path_names
E       AssertionError: assert ['/private/va...h_psa003.npy'] == ['/private/var...itted003.npy']
E         At index 0 diff: '/private/var/folders/ds/z7mbf7nx1gn31308g31t9tcc0000gp/T/pytest-of-oliver/pytest-11/test_load0/paths/path_psa001.npy' != '/private/var/folders/ds/z7mbf7nx1gn31308g31t9tcc0000gp/T/pytest-of-oliver/pytest-11/test_load0/paths/fitted001.npy'
E         Use -v to get the full diff

MDAnalysisTests/analysis/test_psa.py:176: AssertionError
============================================== 1 failed, 358 warnings in 1.86 seconds ===============================================
```

This looks like what you said, namely that different paths are stored. 
- path_psa001.npy
- fitted001.npy

@sseyler for right now, can you tell me what the test should be testing so that we know the current behavior?
1. Should I test for a different set of names to appear? If so, where do I get the names?
2. Or is this a bug and need to be fixed? In which case I will add a comment and keep the xfail in place.

I'd like to merge the PR and then we can open a new issue for the points that you raised.

Thanks!

orbeckst:
@sseyler did you update https://github.com/MDAnalysis/mdanalysis/pull/2049#discussion_r227519989 – if so, sorry, I didn't notice, I only get a ping for new comments, not for edits.

orbeckst:
@sseyler can you rewrite the test that is currently marked `@xfail` so that it properly tests the `psa.load()` and `psa.save_path()` functionality?

orbeckst:
(And then remove the `@xfail`...)

testsuite/MDAnalysisTests/analysis/test_psa.py
sseyler: The most basic issue with the test's failing was a problem with inconsistent naming conventions in the `Path` and `PSAnalysis` classes for default filenames for writing `Path` coordinates as .npy files. This problem was addressed by dealing with the most pertinent issue, which was the handling of the default (base) filename used for saving `Path` coordinates that are the result of rmsd fitting by `AlignTraj`. The semi-permanent solution I came up with was to define two default instance variables for the `PSAnalysis` class:

```python
self.default_fitted_path_prefix = 'psafit_'
self.default_path_basename = 'path'
```

specifying the prefix and filename to default to when saving trajectories ("DCD," "TRR," etc.) and the (`numpy`) coordinates of `Path` objects ("npz," "npy"). The user can still specify a "prefix" and a "postfix" to the default filename (or choose their own filename, to which the prefix will be prepended and the postfix will be appended) in the `PSAnalysis.generate_paths()` method, which will then be passed to `Path.run()` which in turn calls `Path.fit_to_reference()`. `Path.fit_to_reference()` appends the user's selected postfix to the filename to get an updated filename, but leaves the prefix separately specified. Then, `Path.fit_to_reference()` constructs an `AlignTraj` object, passing `prefix` and `filename` as parameters to the constructor; by default, `prefix` is set to "psafit_" as a default value in `Path.fit_to_reference()` in order to be consistent with the default prefix used by `AlignTraj` ("rmsfit_"). The point of leaving "psa" in the prefix is to differentiate the source of the need for rmsd fitting from PSA from other possibilities. Also, `PSAnalysis.generate_paths()` method appends a 3-digit number to each trajectory filename, after the user-selected postfix is appended, to differentiate the (multiple) trajectories (i.e., paths) that were loaded into `PSAnalysis`.

orbeckst: @sseyler thanks for the explanations. Do you have a PR for these changes? You can either 

1. do a PR against this branch or alternatively, 
2. create this branch in your fork, add your code to the branch, and  submit a new PR with my old commits and your new ones. I would then close this PR.

sseyler: @orbeckst I set up a PR (#2239) against my fork of this branch (more-psa-tests).

orbeckst:
This PR has been superseded by PR #2239.

PR #2052, opened 2018-08-15
Adding bond/angle reading to PARM7 TOPParser.py

IAlibay:
As discussed on the MDnalysis group, work in progress to update the AMBER TOPParser to read in bond/angle/dihedral information.

### Changes made in this Pull Request:
 - TOPParser.py now reads bonds, angles, dihedrals and impropers
 - Unused `atom_per` renamed to `num_per_record` and now reflects the number of elements per record in a given PARM7 section (TOPParser.py)
 - New helper functions `parse_chunks` and `parse_dihedrals`  (TOPParser.py)

### To do:
 - Add more stringent tests for TOPParser for non-bond related sections. Possibly defer this to another PR?

PR Checklist
------------
 - [x] Tests? (only basic tests added currently)
 - [x] Docs?
 - [x] CHANGELOG updated?


richardjgowers:
@IAlibay thanks, this looks really good.  Only thing I can think of is that it would be good to have tests to check the identity of a bond, angle and dihedral.  Especially if they are 0 based, so eg check that bond (0, 1) exists.  From what I can tell you're concatenating the various amber bond types together?  Maybe check that a known bond from each section exists correctly.

package/MDAnalysis/topology/TOPParser.py
```diff
@@ -308,3 +519,59 @@ def parsesection_mapper(self, atoms_per, numlines, mapper):
...
+        guaranteed.
+        """
+        improp = []
+        dihed = []
+        for i in itertools.chain(diha, dihh):
+            if i[3] < 0:
+                improp.append(i[:3]+(abs(i[3]),))
+            elif i[2] < 0:
+                vals = i[:2] + (abs(i[2]),) + i[3:]
+                if vals in dihed:
```
richardjgowers: nitpicky, but here you could just have `if not vals in dihed:` and not have the `continue` branch

IAlibay: That does make a lot cleaner thanks :)

Unfortunately, on profiling my original proposed changes to the TOPParser, and it seems that there is a very large overhead associated with this particular if statement (switching to `if not vals in dihed` unfortunately does not improve the issue). When loading a ~ 1 million atom system (AMBER benchmark's STMV system), I see a 50x decrease in performance, taking ~ 1850 seconds to create a universe vs ~ 36 seconds without the if statement.

The original purpose of this particular line was to prevent duplicate dihedral entries which would arise from multiterm dihedrals. However, I am no longer sure if it is actually necessary. Looking more closely at the PSF reader, it seems to not attempt to remove extra multiterm dihedrals either. Additionally, it appears that duplicate dihedrals are removed somewhere in `topologyobjects.py` (although I am not 100% exactly where it is happening). 

I was therefore wondering if it would be reasonable to allow for duplicate dihedrals on read, or if this might cause some odd behaviour somewhere?

richardjgowers: hmm, you could make dihed and improp a set, it's hard to beat the performance of Python's inbuilt set.

Can you link the STMV system you mentioned?

IAlibay: @richardjgowers 
Thanks, as per your suggestion I ended up doing a sorted set conversion in 308bb68 which ended up with much lower overheads ( ~ 1 s extra compared to no duplicate removal for the creation of the STMV system universe).

I did also get rid of the np.int64 conversion for the bond reader (under parse_bonded), as it had a rather high cost and (hopefully I've not missed anything) as far as I can tell the topology attribute for bonds doesn't require np.ints.

Unfortunately the code in this PR does still double the cost of the TOPParser (STMV universe creation goes from ~ 14 s to ~ 28 s) but I think that is unavoidable.

With regards to the STMV system, it can be found as part of the AMBER benchmark suite (under PME/STMV_production_NPT_4fs): http://ambermd.org/Amber16_Benchmark_Suite.tar.bz2

IAlibay:
@richardjgowers Thanks, more thorough bond identity tests have been added with 708df0e.
With regards to bond types, unfortunately the PARM7 format doesn't differentiate between different types of bonds, angles, impropers or dihedrals (unlike TPR files for example).

richardjgowers:
@IAlibay this is ready to merge, I'm just waiting for your other PR so your author credits are in place.  If you add them to this branch I can merge this now.

IAlibay:
@richardjgowers Thanks ``CHANGELOG`` and ``AUTHORS`` have been updated accordingly in bb1139f

IAlibay:
@richardjgowers Unfortunately, I have just encountered edge case where the TOPParser will fail when one of the parsed topology sections is empty (i.e. there are no dihedrals without hydrogens).

Would it be possible to wait on merging until I fix this?

richardjgowers:
Hey, yep, thanks for letting me know :)

IAlibay:
@richardjgowers 995fe65 should address the empty topology section issue mentioned above. The code is currently set to check if an empty section has an empty line and if so, get an extra line. Unfortunately the AMBER documentation on this kind of case isn't too clear, so I left in the possibility that there might be no empty line at all.

Also along with some new tests for TOPParser, I have included a new topology which is detailed as ``Amber ncrst with positions/forces/velocities``. This description is used because this file will be used as part of an upcoming PR which will add an AMBER NCRST parser.

orbeckst:

testsuite/MDAnalysisTests/topology/test_top.py
```diff
@@ -45,35 +51,310 @@ def test_attr_size(self, top):
...
+    atom_i_bond_values = ((1, 4), (4, 5))
+    atom_zero_angle_values = ((0, 1, 2), (0, 1, 3), (0, 1, 4))
+    atom_i_angle_values = ((0, 1, 4), (1, 4, 5), (2, 1, 4), (3, 1, 4))
+    atom_zero_dihedral_values = ((0, 1, 4, 5),)
+    atom_i_dihedral_values = ((0, 1, 4, 5), (2, 1, 4, 5), (3, 1, 4, 5))
+    atom_zero_improper_values = ()
+    atom_i_improper_values = ()
+
+
+class CheckErrors(object):
```
richardjgowers: @IAlibay I don't think these tests will be run.  pytest looks for classes which are named `class TestSomething` or functions/methods called `test_thing`

Each of these three files should be an individual test, so the class should look like:

```python
class TestErrors(object):
    def test_missing_version(self):
        with pytest.raises etc

    def test_missing_title(self):
        etc

IAlibay: Sorry about that, looks like I need to have a more careful read of the pytest docs. It should be fixed in 07adc38

Issue #1979, opened 2018-07-10
deprecate start/stop/step in AnalysisBase.__init__()


orbeckst:
In preparation for #1463 where we want analysis classes to use `start`, `stop`, `step` in the `run()` method only, we need to deprecate the use in 

- [x] `analysis.base.AnalysisBase.__init__`: https://github.com/MDAnalysis/mdanalysis/blob/1a1b2f1442d52d1216b53b32c4c42c3b47368d38/package/MDAnalysis/analysis/base.py#L92
- custom analysis classes (`git grep 'start=.*stop='`)
   - [x] density.py
   - [x] diffusionmap.py (clean-up only, already uses AnalysisBase)
   - [x] hbonds/hbond_analysis.py
   - [x] hbonds/wbridge_analysis.py
   - [x] rms.py (removed the *deprecation* for start/stop/step in `run()`!)

Issue #1463, opened 2017-07-10
update analysis.base.AnalysisBase to conform to API


orbeckst:
In #719 we decided on the details of the analysis interface ("Bauhaus" style). The description for implementation is on the wiki [MDAnalysis.analysis user interface](https://github.com/MDAnalysis/mdanalysis/wiki/MDAnalysis.analysis-user-interface) but we also decided to change a few additional things that are directly defined in the base class and automatically apply to all analysis code: The following needs to be changed in [AnalysisBase](http://www.mdanalysis.org/mdanalysis/documentation_pages/analysis/base.html#MDAnalysis.analysis.base.AnalysisBase):

1. `start`, `stop`, `step` are kwargs of `AnalysisBase.run()` but _not_ of `__init__()`.
2. `verbose` is a kwarg of `__init__` where it sets the default; `verbose` is *also a kwarg of `run()`* where it overrides the default value for the duration of `run` (this allows the user to mute the frame-by-frame calculation but still have verbose output during pre- and post-processing).

Because **1** changes the API, we need to deprecate `start`, `stop`, `step` in `__init__` and **remove them in 1.0.0**. Until then it needs to be possible to set the iteration range in `__init__` and kwargs in `run()` will override the `__init__` values.

orbeckst:
@kain88-de can you have a look and see how much work it is to get this done for 0.17.0?

kain88-de:
From my memory. 

A problem is that we often use information from the start stop step arguments in the init function. This means all sub classes have to be checked after we add a deprecation to the base class. 

Verbosity is likely easier. There might be a corner case in the PCA analysis. 

I’m not sure if there are some surprises in the decorators like base_analysis. 

With writing tests that’s at least a couple of hours of work.

orbeckst:
@richardjgowers can we get deprecations into 0.19.0 #1971 ?

In particular, the `start`/`stop`/`step` in `run()` is reversing a decision that we made much earlier when we tried to move them from `run` to `__init__`. We have depreciation code that says "start/stop/step will be removed from run() in 0.17.0" :-p.

richardjgowers:
@orbeckst I don't understand what needs doing for this, do some classes have start/stop/step in run?

orbeckst:
Classes *should* have start/stop/step in `run()` but not in `__init__`.

At least one (rms.RMSD.run()) has it already in `run()` but it is deprecated. It should be un-deprecated because it will become the norm.

[...]

PR #2055, opened 2018-08-22
removed quiet in favour of verbose

### LINKED ISSUE #1975: remove deprecated `quiet` keyword
We deprecated `quiet=True` in favor of `verbose=False` a long time ago and we announced that this feature would be removed in 0.17.0. However, `quiet` is still accepted in various parts (analysis module).

The `quiet` kwarg and the related code and tests should be finally removed.

richardjgowers:
Fixes #1975 #1979 
Starts #1463

 - 

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -78,7 +78,7 @@ Enhancements
     analysis module (PR #1997, PR #2033)
   * Added the analysis.data module for reference data used in analysis (PR #2033)
   * Added analysis.dihedrals with Ramachandran class to analysis module (PR #1997)
-  * Added augment functionality to create relevant images of particles 
+  * Added augment functionality to create relevant images of particles
   * Most functions in `MDanalysis.lib.distances` previously only accepting
     arrays of coordinates now also accept single coordinates as input (PR #2048,
     Issues #1262 #1938)
@@ -109,6 +109,8 @@ Fixes
   * Fixed SphericalLayer and SphericalZone selections with pbc=True.  Previously
     these shifted all atoms to inside the primary unit cell when calculating the
     center of the reference group (Issue #1795)
+  * PCA analysis now uses start frame as reference frame rather than 0th frame
+    (PR #2055)
 
 Changes
   * TopologyAttrs are now statically typed (Issue #1876)
@@ -130,8 +132,15 @@ Changes
     enforced to be of the form ``[lx, ly, lz, alpha, beta, gamma]`` as required
     by the docs (Issue #2046, PR #2048)
   * Added periodic keyword to select_atoms (#759)
+  * PCA.transform now requires that PCA.run() has already been called
+    otherwise a ValueError is raised (PR #2055)
+  * The quiet keyword has been removed and replaced with verbose (Issue #1975
+    PR #2055)
 
 Deprecations
+  * start/stop/step are deprecated in the initialization of Analysis classes.
+    These parameters should instead be given to the run() method of the class.
+    (Issue #1463 #1979 PR #2055)
   * Almost all "save()", "save_results()", "save_table()" methods in
     analysis classes were deprecated and will be removed in 1.0 (Issue
     #1972 and #1745)
--- a/package/MDAnalysis/analysis/base.py
+++ b/package/MDAnalysis/analysis/base.py
@@ -32,11 +32,12 @@ import six
 from six.moves import range, zip
 import inspect
 import logging
+import warnings
 
 import numpy as np
 from MDAnalysis import coordinates
 from MDAnalysis.core.groups import AtomGroup
-from MDAnalysis.lib.log import ProgressMeter, _set_verbose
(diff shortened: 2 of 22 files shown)


orbeckst:
That fails with some `_set_verbose` still being used somewhere.

orbeckst:

orbeckst:
@richardjgowers I rebase this branch against develop and force-pushed. Be careful with your next `git pull`.

richardjgowers:
Ok I still need to

 - [ ] check if/where we refer to start/stop/step in docs
 - [ ] check existing usage of start/stop/step in tests etc
 - [ ] add tests for the new deprecations
 - [ ] check I got verbose right

package/MDAnalysis/analysis/base.py
```diff
@@ -89,25 +90,28 @@ def _conclude(self):
...
+        self._verbose = verbose
+        # do deprecated kwargs
+        # remove in 1.0
+        deps = []
+        for arg in ['start', 'stop', 'step']:
+            if arg in kwargs and not kwargs[arg] is None:
+                deps.append(arg)
+                setattr(self, arg, kwargs[arg])
+        if deps:
+            raise ValueError(deps)
```
richardjgowers: I've made this a VE so that I can find all deprecated usages in the testsuite, will remove later..

package/MDAnalysis/analysis/pca.py
```diff
@@ -181,46 +181,39 @@ def __init__(self, universe, select='all', align=False, mean=None,
...
-        self._atoms = self._u.select_atoms(select)
+
+        self._calculated = False
         self.n_components = n_components
+        self._select = select
+        self._mean = mean
+
+    def _prepare(self):
+        # access start index
+        self._u.trajectory[self.start]
```
richardjgowers: I'm not sure if this is right.  Previously this used the 0th frame, but it makes more sense to use the `self.start` frame right?

package/MDAnalysis/analysis/pca.py
```diff
@@ -305,7 +298,7 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
         pca_space : array, shape (number of frames, number of components)
         """
         if not self._calculated:
-            self.run()
+            self.run(start=start, stop=stop, step=step)
```
richardjgowers: this seems to cause errors.  Are `run` and `transform` meant to be allowed to use different frames? @kain88-de

package/MDAnalysis/analysis/pca.py
orbeckst: Yes, it makes more sense to me to use `start`. Should be documented, though.

package/MDAnalysis/analysis/pca.py
orbeckst: Yes. You may first compute the transformation and then apply it to an arbitrary trajectory to project out the motions corresponding to the PCs.

I can't think of a clean way to separate them here so instead of running `self.run()` rather raise the error and let the user do it explicitly.  Also makes for much clearer scripts on the user side.

Needs to be documented in the CHANGELOG and with a 
```reST
.. versionchanged 0.19.0:: 
   :meth:`transform` does not automatically :meth:`run` the PCA if needed; instead 
   it raises :exc:`ValueError` to remind the user to perform :meth:`run` first.
```

richardjgowers:
@orbeckst this is hopefully finished

package/CHANGELOG
```diff
@@ -127,6 +129,7 @@ Changes
     enforced to be of the form ``[lx, ly, lz, alpha, beta, gamma]`` as required
     by the docs (Issue #2046, PR #2048)
   * Added periodic keyword to select_atoms (#759)
+  * PCA.transform now requires that PCA.run has already been called (PR #2055)
 
 Deprecations
   * Almost all "save()", "save_results()", "save_table()" methods in
```
orbeckst: add start/stop/step deprecations here

package/CHANGELOG
```diff
@@ -127,6 +129,7 @@ Changes
     enforced to be of the form ``[lx, ly, lz, alpha, beta, gamma]`` as required
     by the docs (Issue #2046, PR #2048)
   * Added periodic keyword to select_atoms (#759)
+  * PCA.transform now requires that PCA.run has already been called (PR #2055)
```
orbeckst: add start/stop/step in `run()` here

orbeckst: Add removal of quiet here

orbeckst: Add: PCA.transform now requires that `run` has been called before, otherwise a`ValueError` is raised.

orbeckst:
@richardjgowers all that is needed are doc/CHANGELOG additions. Then you'll have fixed two issues! (`<carrot dangle/>`)

richardjgowers:
@orbeckst gepöken

PR #2056, opened 2018-08-22
deprecate flags

### LINKED ISSUE #759: Periodic/Nonperiodic keywords to selections
So currently we have some [geometric selections](http://pythonhosted.org/MDAnalysis/documentation_pages/selections.html#geometric) which allow you to select atoms based on absolute and relative positions.  These can work with and without periodic boundary conditions, both are useful at different times.  

The use of PBC is currently controlled by flags, which I think we're moving towards removing (#315) so instead a keyword in the selection could be used.

```
u.select_atoms('A around B distance C selection D')

u.select_atoms('A point B x y z C distance D')
```

So with the above selections, there are positions **A** -> **D** where we could put the keyword.  

Position **A** would be tricky as the keyword precedes the token, but it could be done.  (`'periodic around 7.0 protein'`)

**D** is messy for the `around` case as generally anything after **C** gets made into another selection.

I think **C** reads best.. (`'around 7.0 periodic protein'`)

So tl;dr
- add `periodic` keyword?
- what should the default be? (probably same as currently?)
- where to put the `periodic` keyword?

richardjgowers:
Fixes #759 
Starts #782

 - add periodic kwarg to select_atoms

PR Checklist
------------
 - [ ] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -120,12 +120,16 @@ Changes
     (Issue #1894)
   * removed unused function MDAnalysis.lb.mdamath.one_to_many_pointers()
     (issue #2010)
+  * Added periodic keyword to select_atoms (#759)
 
 Deprecations
   * almost all "save()", "save_results()", "save_table()" methods in
     analysis classes were deprecated and will be removed in 1.0 (Issue
     #1972 and #1745)
-
+  * deprecated use of core.flags.  For use_pbc, the pbc keyword should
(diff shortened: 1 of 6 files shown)


jbarnoud:
Doesn't this silently reverse the default behavior for the equivalent of the 'use_pbc' flag?

jbarnoud:
Just for the record, the `convert_lengths` flag is broken, but it was broken already.

PR #2062, opened 2018-09-03
Improved docstrings in lib/_augment.pyx

zemanj:
Fixes (partially) #2046, this is the second of a series of related PRs, following PR #2048.

 - Improved documentation of `MDAnalysis.lib.distances.augment_coordinates()` and `MDAnalysis.lib.distances.undo_augment()` to better explain what the methods are actually doing.
- Added corresponding `.. autofunction` in `lib/distances.py` so that the signatures of both functions are now properly displayed in the html docs.
- Corrected a mistake from PR #2048 so that `distance_array()` now also returns an array for single coordinate input.
- Added missing CHANGELOG entries for PR #2048.

Apparently, there is a problem with Windows/Linux (i.e., CRLF vs. LF) line endings resulting in an incorrect `git diff` output for `lib/_augment.pyx`, which now looks as if I changed every line in the file, even though I only changed the docstrings. With `git diff -b` the changes are displayed correctly, though. On github, just change the diff settings to ignore whitespace.

PR Checklist
------------
 - [ ] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -80,6 +80,8 @@ Enhancements
   * Added dihedrals.py with Dihedral, Ramachandran, and Janin classes to
     analysis module (PR #1997, PR #2033)
   * Added the analysis.data module for reference data used in analysis (PR #2033)
+  * Most functions in `MDanalysis.lib.distances` previously only accepting
+    arrays of coordinates now also accept single coordinates as input (PR #2048)
 
 Fixes
   * rewind in the SingleFrameReader now reads the frame from the file (Issue #1929)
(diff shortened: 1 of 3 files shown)


zemanj:
@ayushsuhane Since you wrote the augment routines (PR #1977), could you take a quick look at the new docstrings? Just to make sure I didn't add any bogus. :wink:

package/CHANGELOG
```diff
@@ -80,7 +80,8 @@ Enhancements
   * Added dihedrals.py with Dihedral, Ramachandran, and Janin classes to
     analysis module (PR #1997, PR #2033)
   * Added the analysis.data module for reference data used in analysis (PR #2033)
-  * Added dihedrals.py with Ramachandran class to analysis module
```
richardjgowers: @zemanj I think you accidentally killed this entry

zemanj: I thought that's already included three lines above... am I wrong?

richardjgowers: ah it's weird it's in the log twice then, yeah ok

zemanj:
Mmmh... nobody up to reading some good ol' poetry? :smile:

ayushsuhane:
@zemanj  Looks good. The docs are more clear now. Thanks!! 👍

Issue #2066, opened 2018-09-08
Python 3.4 Travis fail due to unmet dependencies


orbeckst:
The Python 3.4 tests are all failing now:

```
UnsatisfiableError: The following specifications were found to be in conflict:
  - numpy=1.15 -> *[track_features=blas_openblas]
  - numpy=1.15 -> python[version='>=3.7,<3.8.0a0'] -> tk[version='>=8.6.7,<8.7.0a0']
  - python=3.4
Use "conda info <package>" to see the dependencies for each package.
```

Examples:
- https://travis-ci.com/MDAnalysis/mdanalysis/jobs/144464939
- https://travis-ci.org/MDAnalysis/mdanalysis/jobs/425782569

Do we need to pin numpy or should we deactivate 3.4 tests?

cc: @richardjgowers @jbarnoud @tylerjereddy

zemanj:
We could pin the numpy version here as yet another band-aid. Though, what we really want is something like "take the highest possible version that is compatible with this python version". Is there a sane way to achieve this without doing nasty things like parsing the output of `conda info`?

richardjgowers:
I thought figuring this stuff out was something conda was meant to figure out for itself.  Maybe we're not giving conda enough information initially to make these decisions.  Pinning on the 3.4 build is an easy fix which wil allow PRs to continue though

orbeckst:
We will have to do *something* soon because this holds up merging of all PRs.

> I thought figuring this stuff out was something conda was meant to figure out for itself. Maybe we're not giving conda enough information initially to make these decisions.

Yes, I though so, too. Perhaps this is actually a bug in one of the build recipes (maybe @kain88-de sees immediately what's wrong) or in the ci-helper script (???) but that doesn't help us right now.

What is the last version of numpy that works for us in 3.4? We then pin that version, [using ci-helper's NUMPY_VERSION](https://github.com/astropy/ci-helpers/blob/master/README.md)
```bash
PYTHON_VERSION=3.4 NUMPY_VERSION=1.14 CODECOV="true" SETUP_CMD="${PYTEST_FLAGS} --cov=MDAnalysis" 
```
(or whatever is needed)

kain88-de:
This looks like a conda error. Numpy 1.15 has not been build for python3.4. This is coming up now because the 1.15 version has been published 6 days ago.

[...]

PR #2068, opened 2018-09-10
 Fix SETTLE in TPR parser 

jbarnoud:
Support for SETTLE interactions in the TPR parser was added in #1953.
Though, the support was only added for the older versions of the TPR
format, causing errors with the newest files. In practice, when reading
the bonds on a recent TPR file containing SETTLE constraints, the
indices to the atoms involved in the bonds would be shifted. This is
because, the older TPR files use a different format to describe SETTLE
constraints.

This commit should fix the issue by supporting both the old and the new
format for SETTLE.

PR Checklist
------------
 - [X] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/testsuite/MDAnalysisTests/topology/test_tprparser.py
+++ b/testsuite/MDAnalysisTests/topology/test_tprparser.py
@@ -233,3 +233,5 @@ def bonds_water(request):
 def test_settle(bonds_water):
     # There are 101 water molecule with 2 bonds each
     assert len(bonds_water) == 202
+    # The last index corresponds to the last water atom
+    assert bonds_water[-1][1] == 2262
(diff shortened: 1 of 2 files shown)


orbeckst:

richardjgowers:
changelog update needed? It's important to note if this changes previous behaviour (ie this bug was silent before) or if it allows tprs with settle bonds (parser would choke on these before)

jbarnoud:
The PR fixes a bug introduced in the dev branch. As the bug was not released yet, I do not think it is worth telling in the changelog that it is fixed :)

testsuite/MDAnalysisTests/topology/test_tprparser.py
```diff
@@ -233,3 +233,5 @@ def bonds_water(request):
 def test_settle(bonds_water):
     # There are 101 water molecule with 2 bonds each
     assert len(bonds_water) == 202
+    # The last index corresponds to the last water atom
```
orbeckst: Does this test a new TPR file?

jbarnoud: Yes it does. It tests almost all versions of gromacs from 4.0.0 to 2018.

PR #2070, opened 2018-09-11
improved docstrings in lib/distances.py

zemanj:
Fixes (partially) #2046, this is the third of a series of related PRs, following PRs #2048 and #2062.

- Improved documentation of all functions in `MDAnalysis.lib.distances`:
  - Ensured that parameters referenced in docstrings correspond to function signatures.
  - Properly formated references to other methods so that links are generated.
  - Now uses logical instead of visual markup.
  - Enforced 80 character lines.
  - Standardized phrasing of repeatedly occurring parameter descriptions.
  - Now uses correct type specifications for parameters and return values.
  - Marked optional parameters by adding *, optional* after the type.
- Removed all `.. autofunction` entries that caused duplicates in the documentation.
- Improved module description to better correspond to its contents.
- `box` parameter of `capped_distance()` and `self_capped_distance()` now also accept `array_like` input (now corresponding to the docs like all other functions accepting box input).

PR Checklist
------------
 - [ ] Tests?
 - [x] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/lib/distances.py
+++ b/package/MDAnalysis/lib/distances.py
@@ -77,8 +77,7 @@ from numpy.lib.utils import deprecate
 from .util import check_coords
 from .mdamath import triclinic_vectors, triclinic_box
 from ._augment import augment_coordinates, undo_augment
-
-
+from .nsgrid import FastNS
 
 # hack to select backend with backend=<backend> kwarg. Note that
 # the cython parallel code (prange) in parallel.distances is
@@ -502,13 +501,6 @@ def _determine_method(reference, configuration, max_cutoff, min_cutoff=None,
     -------
     function : callable
         The function implementing the guessed (or deliberatly chosen) method.
-
-    Note
-    ----
-    Currently implemented methods are present in the ``methods`` dictionary
-        bruteforce : returns ``_bruteforce_capped``
-        PKDtree : return ``_pkdtree_capped``
-        NSGrid : return ``_nsgrid_capped``
     """
     methods = {'bruteforce': _bruteforce_capped,
                'pkdtree': _pkdtree_capped,
@@ -669,7 +661,7 @@ def _pkdtree_capped(reference, configuration, max_cutoff, min_cutoff=None,
         coordinates ``reference[pairs[k, 0]]`` and
         ``configuration[pairs[k, 1]]``.
     """
-    from .pkdtree import PeriodicKDTree
+    from .pkdtree import PeriodicKDTree  # must be here to avoid circular import
 
     kdtree = PeriodicKDTree(box=box)
     cut = max_cutoff if box is not None else None
@@ -745,8 +737,6 @@ def _nsgrid_capped(reference, configuration, max_cutoff, min_cutoff=None,
         coordinates ``reference[pairs[k, 0]]`` and
         ``configuration[pairs[k, 1]]``.
     """
-    from .nsgrid import FastNS
-
     if box is None:
         # create a pseudobox
         # define the max range
@@ -898,13 +888,6 @@ def _determine_method_self(reference, max_cutoff, min_cutoff=None, box=None,
     -------
     function : callable
         The function implementing the guessed (or deliberatly chosen) method.
-
-    Note
-    ----
-    Currently implemented methods are present in the ``methods`` dictionary
-        bruteforce : returns ``_bruteforce_capped_self``
-        PKDtree : return ``_pkdtree_capped_self``
-        NSGrid : return ``_nsgrid_capped_self``
     """
     methods = {'bruteforce': _bruteforce_capped_self,
                'pkdtree': _pkdtree_capped_self,
@@ -1035,7 +1018,7 @@ def _pkdtree_capped_self(reference, max_cutoff, min_cutoff=None, box=None):
         distance between the coordinates ``reference[pairs[k, 0]]`` and
         ``reference[pairs[k, 1]]``.
     """
-    from .pkdtree import PeriodicKDTree
+    from .pkdtree import PeriodicKDTree  # must be here to avoid circular import
 
     pairs, distance = [], []
     kdtree = PeriodicKDTree(box=box)
@@ -1092,8 +1075,6 @@ def _nsgrid_capped_self(reference, max_cutoff, min_cutoff=None, box=None):
         distance between the coordinates ``reference[pairs[k, 0]]`` and
         ``reference[pairs[k, 1]]``.
     """
-    from .nsgrid import FastNS
-
     reference = np.asarray(reference, dtype=np.float32)
     if reference.shape == (3, ) or len(reference) == 1:
         return [], []
--- a/package/MDAnalysis/lib/distances.py
+++ b/package/MDAnalysis/lib/distances.py
@@ -262,9 +262,8 @@ def distance_array(reference, configuration, box=None, result=None,
         ``numpy.float64``.
         Avoids creating the array which saves time when the function
         is called repeatedly.
-    backend : str, optional
-        Select the type of acceleration; ``'serial'`` is always available.
-        Another possibility is ``'OpenMP'``.
+    backend : {'serial', 'OpenMP'}, optional
+        Keyword selecting the type of acceleration.
 
     Returns
     -------
@@ -328,9 +327,8 @@ def self_distance_array(reference, box=None, result=None, backend="serial"):
         Preallocated result array which must have the shape ``(n*(n-1)/2,)`` and
         dtype ``numpy.float64``. Avoids creating the array which saves time when
         the function is called repeatedly.
-    backend : str, optional
-        Select the type of acceleration; ``'serial'`` is always available.
-        Another possibility is ``'OpenMP'``.
+    backend : {'serial', 'OpenMP'}, optional
+        Keyword selecting the type of acceleration.
 
     Returns
     -------
@@ -409,9 +407,9 @@ def capped_distance(reference, configuration, max_cutoff, min_cutoff=None,
         triclinic and must be provided in the same format as returned by
         :attr:`MDAnalysis.coordinates.base.Timestep.dimensions`:\n
         ``[lx, ly, lz, alpha, beta, gamma]``.
-    method : str, optional
-        Keyword to override the automatic guessing ofthe employed search method.
-        Can be ``'bruteforce'``, ``'nsgrid'``,  or ``'pkdtree'``.
+    method : {'bruteforce', 'nsgrid', 'pkdtree'}, optional
+        Keyword to override the automatic guessing of the employed search
+        method.
     return_distances : bool, optional
         If set to ``True``, distances will also be returned.
 
@@ -496,9 +494,9 @@ def _determine_method(reference, configuration, max_cutoff, min_cutoff=None,
         triclinic and must be provided in the same format as returned by
         :attr:`MDAnalysis.coordinates.base.Timestep.dimensions`:\n
         ``[lx, ly, lz, alpha, beta, gamma]``.
-    method : str, optional
-        Keyword to override the automatic guessing ofthe employed search method.
-        Can be ``'bruteforce'``, ``'nsgrid'``, or ``'pkdtree'``.
+    method : {'bruteforce', 'nsgrid', 'pkdtree'}, optional
+        Keyword to override the automatic guessing of the employed search
+        method.
 
     Returns
     -------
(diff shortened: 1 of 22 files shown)


package/MDAnalysis/lib/distances.py
```diff
@@ -367,75 +362,82 @@ def self_distance_array(reference, box=None, result=None, backend="serial"):
...
-        Function returns distances if set to 'True'
+        Maximum cutoff distance between the reference and configuration.
+    min_cutoff : float, optional
+        Minimum cutoff distance between reference and configuration.
+    box : array_like, optional
+        The unitcell dimensions of the system, which can be orthogonal or
+        triclinic and must be provided in the same format as returned by
+        :attr:`MDAnalysis.coordinates.base.Timestep.dimensions`:\n
+        ``[lx, ly, lz, alpha, beta, gamma]``.
+    method : str, optional
```
orbeckst: You can provide the selection with
```reST
method : {"bruteforce", "nsgrid", "pkdtree", None}
     Keyword to override the automatic guessing ofthe employed search method.
     ``None`` chooses an optimal method.
```
(see https://numpydoc.readthedocs.io/en/latest/format.html#sections under _4. Parameters_)

package/MDAnalysis/lib/distances.py
```diff
@@ -367,75 +362,82 @@ def self_distance_array(reference, box=None, result=None, backend="serial"):
...
     Note
     -----
-    Currently only supports brute force and Periodic KDtree
+    Currently supports brute force, grid-based, and periodic KDtree search
+    methods.
 
-    .. SeeAlso:: :func:'MDAnalysis.lib.distances.distance_array'
-    .. SeeAlso:: :func:'MDAnalysis.lib.pkdtree.PeriodicKDTree.search'
-    .. SeeAlso:: :class:'MDAnalysis.lib.nsgrid.FastNS.search'
+    .. seealso:: :meth:`~MDAnalysis.lib.distances.distance_array`
```
orbeckst: This should use NUmpyDoc 
```reST
See Also
--------
MDAnalysis.lib.distances.distance_array
MDAnalysis.lib.pkdtree.PeriodicKDTree.search
MDAnalysis.lib.nsgrid.FastNS.search
```
see https://numpydoc.readthedocs.io/en/latest/format.html#sections, _11. See Also_

package/MDAnalysis/lib/distances.py
```diff
@@ -525,64 +523,134 @@ def _determine_method(reference, configuration, max_cutoff, min_cutoff=None,
...
+        ``i``-th coordinate in `reference` and the ``j``-th coordinate in
+        `configuration`.
+    distances : numpy.ndarray, optional
+        Distances corresponding to each pair of indices. Only returned if
+        `return_distances` is ``True``. ``distances[k]`` corresponds to the
+        ``k``-th pair returned in `pairs` and gives the distance between the
+        coordinates ``reference[pairs[k, 0]]`` and
+        ``configuration[pairs[k, 1]]``.
     """
     from .pkdtree import PeriodicKDTree
```
orbeckst: We rather have all imports at the top of the file instead of methods. If it has to be inside a method (eg to avoid circular imports) then it should be commented. See https://github.com/MDAnalysis/mdanalysis/wiki/Style-Guide#importing-modules

orbeckst: (Yes, I know, this is not your code...)

package/MDAnalysis/lib/distances.py
```diff
@@ -592,55 +660,73 @@ def _pkdtree_capped(reference, configuration, max_cutoff, min_cutoff=None,
...
+        ``i``-th coordinate in `reference` and the ``j``-th coordinate in
+        `configuration`.
+    distances : numpy.ndarray, optional
+        Distances corresponding to each pair of indices. Only returned if
+        `return_distances` is ``True``. ``distances[k]`` corresponds to the
+        ``k``-th pair returned in `pairs` and gives the distance between the
+        coordinates ``reference[pairs[k, 0]]`` and
+        ``configuration[pairs[k, 1]]``.
     """
     from .nsgrid import FastNS
```
orbeckst: See above: would rather import at top of module.

package/MDAnalysis/lib/distances.py
```diff
@@ -689,120 +775,124 @@ def _nsgrid_capped(reference, configuration, max_cutoff, min_cutoff=None,
...
-        in the function [None]
+        Maximum cutoff distance between `reference` coordinates.
+    min_cutoff : float, optional
+        Minimum cutoff distance between `reference` coordinates.
+    box : array_like, optional
+        The unitcell dimensions of the system, which can be orthogonal or
+        triclinic and must be provided in the same format as returned by
+        :attr:`MDAnalysis.coordinates.base.Timestep.dimensions`:\n
+        ``[lx, ly, lz, alpha, beta, gamma]``.
+    method : str, optional
```
orbeckst: explicitly list method names (see above)

package/MDAnalysis/lib/distances.py
```diff
@@ -689,120 +775,124 @@ def _nsgrid_capped(reference, configuration, max_cutoff, min_cutoff=None,
...
     Note
     -----
-    Currently only supports brute force, Periodic KDtree and Grid Search
+    Currently supports brute force, grid-based, and periodic KDtree search
+    methods.
 
-    .. SeeAlso:: :func:'MDAnalysis.lib.distances.self_distance_array'
-    .. SeeAlso:: :func:'MDAnalysis.lib.pkdtree.PeriodicKDTree.search'
-    .. SeeAlso:: :func:'MDAnalysis.lib.nsgrid.FastNS.self_search'
+    .. seealso:: :meth:`~MDAnalysis.lib.distances.self_distance_array`
```
orbeckst: Use `See Also` numpy doc. (See above)

package/MDAnalysis/lib/distances.py
```diff
@@ -689,120 +775,124 @@ def _nsgrid_capped(reference, configuration, max_cutoff, min_cutoff=None,
...
-        in the function [None]
+        Maximum cutoff distance between `reference` coordinates.
+    min_cutoff : float, optional
+        Minimum cutoff distance between `reference` coordinates.
+    box : numpy.ndarray
+        The unitcell dimensions of the system, which can be orthogonal or
+        triclinic and must be provided in the same format as returned by
+        :attr:`MDAnalysis.coordinates.base.Timestep.dimensions`:\n
+        ``[lx, ly, lz, alpha, beta, gamma]``.
+    method : str, optional
```
orbeckst: document options explicitly

package/MDAnalysis/lib/distances.py
```diff
@@ -1334,15 +1551,33 @@ def calc_distance(a, b, box=None):
...
+        The unitcell dimensions of the system, which can be orthogonal or
+        triclinic and must be provided in the same format as returned by
+        :attr:`MDAnalysis.coordinates.base.Timestep.dimensions`:\n
+        ``[lx, ly, lz, alpha, beta, gamma]``.
+
+    Returns
+    -------
+    angle : float
+        The angle between the vectors `b`:math:`\\rightarrow`\ `a` and
+        `b`:math:`\\rightarrow`\ `c`.
```
orbeckst: Maybe add a 
```reST
See Also
--------
calc_angles : calculate multiple angles efficiently
```

package/MDAnalysis/lib/distances.py
```diff
@@ -1351,14 +1586,35 @@ def calc_angle(a, b, c, box=None):
...
+        The unitcell dimensions of the system, which can be orthogonal or
+        triclinic and must be provided in the same format as returned by
+        :attr:`MDAnalysis.coordinates.base.Timestep.dimensions`:\n
+        ``[lx, ly, lz, alpha, beta, gamma]``.
+
+    Returns
+    -------
+    dihedral : float
+        The dihedral angle between the planes spanned by the coordinates
+        (`a`, `b`, `c`) and (`b`, `c`, `d`).
```
orbeckst: Maybe add a 
```reST
See Also
--------
calc_dihedrals : calculate multiple dihedrals efficiently
```

package/MDAnalysis/lib/distances.py
```diff
@@ -1318,14 +1523,26 @@ def apply_PBC(coords, box, backend="serial"):
...
-        simulation box, if given periodic boundary conditions will be applied
+        The unitcell dimensions of the system, which can be orthogonal or
+        triclinic and must be provided in the same format as returned by
+        :attr:`MDAnalysis.coordinates.base.Timestep.dimensions`:\n
+        ``[lx, ly, lz, alpha, beta, gamma]``.
+
+    Returns
+    -------
+    distance : float
+        The distance between positions `a` and `b`.
```
orbeckst: Maybe add a 
```reST
See Also
--------
calc_distances : calculate multiple distances efficiently
```

package/MDAnalysis/lib/distances.py
```diff
@@ -52,20 +52,8 @@
 
 Functions
 ---------
-
-.. autofunction:: distance_array(reference, configuration [, box [, result [, backend]]])
```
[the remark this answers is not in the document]
zemanj: Removing the `:members:` from `distances.rst` and adding back the `.. autofunction ::` entries works, so I'll do that.

package/MDAnalysis/lib/distances.py
zemanj: The numpydoc docstring guide also says
> If it is not necessary to specify a keyword argument, use `optional`.

And furthermore (what you referred to)
> When a parameter can only assume one of a fixed set of values, those values can be listed in braces, with the default appearing first.

Listing the default `None` (which should come first, btw) doesn't really make sense to me, that would be like describing an optional kwarg as, for example, `x : int or None`.

I think a proper way of combining the two guidelines should look like this:
```
method : {'bruteforce', 'nsgrid', 'pkdtree'}, optional
    Keyword to override the automatic guessing of the employed search method.
```
The default is given in the function signature anyway.

What do you think?

package/MDAnalysis/lib/distances.py
```diff
@@ -220,55 +208,59 @@ def _check_result_array(result, shape):
...
-        ly, lz, alpha, beta, gamma]``.
-    result : numpy.ndarray(dtype=numpy.float64), optional
-        Preallocated result array which must have the
-        shape ``(len(ref), len(conf))`` and ``dtype=numpy.float64``.
+        Configuration coordinate array of shape ``(3,)`` or ``(m, 3)`` (dtype is
+        arbitrary, will be converted to ``numpy.float32`` internally).
+    box : array_like, optional
+        The unitcell dimensions of the system, which can be orthogonal or
+        triclinic and must be provided in the same format as returned by
+        :attr:`MDAnalysis.coordinates.base.Timestep.dimensions`:\n
```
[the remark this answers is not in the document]
zemanj: Luckily, it does! :smile:

package/MDAnalysis/lib/distances.py
zemanj: Right, will do.

package/MDAnalysis/lib/distances.py
zemanj: I'll move the imports to the top if they don't have any side effects (which they shouldn't).

package/MDAnalysis/lib/distances.py
zemanj: That is something I deliberately omitted. Since I introduced the new `@check_coords()` decorator, both `calc_angle()` and `calc_angles()` accept single coordinates as well as coordinate arrays, and return either an array or a single float, depending on the input.
The only difference is that `calc_angle()` returns the angle(s) in degrees while the result of `calc_angles()` is in radians. IMHO we should deprecate `calc_angle()`.
See also my comments on `calc_distance()` and `calc_dihedral()`.

package/MDAnalysis/lib/distances.py
zemanj: Both functions accept both single coordinates and coordinate arrays. There is virtually no difference left between `calc_distance()` and `calc_distances()`. We might want to deprecate `calc_distance()`.

package/MDAnalysis/lib/distances.py
zemanj: Same as above. The only difference between `calc_dihedral()` and `calc_dihedrals()` is that the former returns degrees and the latter radians. We should probably deprecate `calc_dihedral()`.

zemanj:
Was 18.1 an official release? If not, we could simply remove the now obsolete functions `calc_distance()`, `calc_angle()` and `calc_dihedral()` instead of deprecating them.
Either way, in places where the functions are used, they could be replaced according to
* `calc_distance(...)` :arrow_right: `calc_distances(...)`
* `calc_angle(...)` :arrow_right: `np.rad2deg(calc_angles(...))`
* `calc_dihedral(...)` :arrow_right: `np.rad2deg(calc_dihedrals(...))`

richardjgowers:
@zemanj if you want to deprecate `calc_angle` etc you can just remove them completely, they've never been released.  It might be cleaner to have fewer functions

zemanj:
@richardjgowers I opened an issue (#2072) regarding the obsolete functions. Thanks for the info!
@orbeckst Thank you for the quick and very constructive review! :+1:

PR #2073, opened 2018-09-13
removed obsolete functions from lib.distances

### LINKED ISSUE #2072: obsolete functions in lib.distances
**Problem description**

There are three functions in `lib.distances` that were introduced in version 0.18.1 but became obsolete since PR #2048:
* `calc_distance()`
* `calc_angle()`
* `calc_dihedral()`

These functions were intended for use with single coordinates (numpy arrays with shape `(3,)`) but are no longer needed since the functions `calc_bonds()`, `calc_angles()`, and `calc_dihedrals()` now also accept single coordinates as input (and, likewise, return a single distance / angle / dihedral in that case).

To be precise, we now have the following equivalences:
* `calc_distance(...)` :left_right_arrow: `calc_bonds(...)`
* `calc_angle(...)` :left_right_arrow: `np.rad2deg(calc_angles(...))`
* `calc_dihedral(...)` :left_right_arrow: `np.rad2deg(calc_dihedrals(...))`

The obsolete functions are used in the following files:
* `analysis/hbonds/hbond_analysis.py`
* `analysis/hbonds/wbridge_analysis.py`
* `core/topologyobjects.py`

**Possible solutions**

1. Leave everything as is (but maybe update the docstrings).
2. Mark the three functions as deprecated and replace their occurrences in the code according to the equivalent functions as listed above.
3. Remove the functions and replace them (as in solution 2)

Personally, I'm *not* in favor of solution 1.
The most important question for me is whether 0.18.1 was an official release, or in other words: Can we choose solution 3 or do we need solution 2?

0.18.1-dev

zemanj:
Fixes #2072

 - removed the now obsolete functions `calc_distance`, `calc_angle`, and `calc_dihedral` from `lib.distances`
- replaced all of their occurrences in the code by suitable calls to `calc_bonds`, `calc_angles`, and `calc_dihedrals`, respectively
- transformed the corresponding tests into tests of `calc_bonds`, `calc_angles`, and `calc_dihedrals` for single coordinate input
- updated the docstrings of `calc_bonds`, `calc_angles`, and `calc_dihedrals` to better reflect the possibility of single coordinate input (also related to issue #2046)

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


zemanj:
Regarding my latest commit: The precision used to be 3 decimals in the tests for `calc_angle` and `calc_dihedral`. When I applied these tests to `calc_angles` and `calc_dihedrals`, I increased the precision to 5 decimals, but that fails in one of the tests for numpy 1.10.4. I therefore decreased the precision again to 4 decimals.

zemanj:
Ok this seems good to go.

testsuite/MDAnalysisTests/utils/test_distances.py
```diff
@@ -383,7 +383,6 @@ def triclinic_box():
     @pytest.fixture()
     def positions():
```
richardjgowers: I’m not sure how utils/test distances became a thing, but we should be merging it into lib/ rather than vice versa

zemanj: True that, will do.

zemanj:
@richardjgowers The tests are still quite an unstructured mess though...

zemanj:
@richardjgowers Thanks for reviewing!

PR #2074, opened 2018-09-14
fixes #1795

### LINKED ISSUE #1795: different results of select_atoms in 0.17.0 and 0.16.2
I call select_atoms method with following selection: '(resname WAT or resname HOH) and (sphzone 6.0 (resnum 99 147 231 261 289))'

### Expected behaviour

I expect to get exactly the same results in 0.16.2 and 0.17.0 - 5 water molecules.

### Actual behaviour

Version 0.16.2 gives 5 water molecules.
Version 0.17.0 gives 31 water molecules, completely different molecules.

richardjgowers:
Fixes #1795

 - SphericalZone and SphericalLayer no longer shift atoms to inside
primary unit cell when calculating center of reference group

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -106,6 +106,9 @@ Fixes
   * Modifying coordinates by assignation is consistently persistent when using
     the memory reader (Issue #2018)
   * Allow import of WaterBridgeAnalysis from analysis.hbonds (#2064)
+  * Fixed SphericalLayer and SphericalZone selections with pbc=True.  Previously
+    these shifted all atoms to inside the primary unit cell when calculating the
+    center of the reference group (Issue #1795)
 
 Changes
   * TopologyAttrs are now statically typed (Issue #1876)
--- a/package/MDAnalysis/core/selection.py
+++ b/package/MDAnalysis/core/selection.py
@@ -288,8 +288,7 @@ class SphericalLayerSelection(DistanceSelection):
         sel = self.sel.apply(group)
         box = self.validate_dimensions(group.dimensions)
         periodic = box is not None
-        ref = sel.center_of_geometry(pbc=periodic).reshape(1, 3).astype(
-            np.float32)
+        ref = sel.center_of_geometry().reshape(1, 3).astype(np.float32)
         pairs = distances.capped_distance(ref, group.positions, self.exRadius,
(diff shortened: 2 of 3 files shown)


richardjgowers:
I think this finally fixes #1795.  `center_of_geometry(pbc=True)` doesn't really make sense for these selections, if anything we need `center(unwrap=True)`  ie #2012 but that's not merged yet.  The test that is added is an unwrapped system, so shouldn't change before/after this change.

tylerjereddy:
The Travis failure is likely unrelated -- we should maybe go ahead and do a crude pinning fix to the NumPy version there I suppose so we don't block PRs & then rebase here maybe.

richardjgowers:
@tylerjereddy the centroid is still used, but it always assumes that all atoms are in the correct image to calculate this. WRT your sketch, if you had protein/X atoms in either side of the box this selection would fail because it would calculate the centroid as roughly the center of the box.  This is exactly what unwrap would fix.

In general `pbc=True` needs to get replaced with `wrap=True` (ie consider all atoms in the primary unit cell), and `unwrap=True` (make sure bonds aren't broken across periodic boundaries) which is what #2012 is about.

zemanj:
> In general `pbc=True` needs to get replaced with `wrap=True` (ie consider all atoms in the primary unit cell), and `unwrap=True` (make sure bonds aren't broken across periodic boundaries) which is what #2012 is about.

I agree with @richardjgowers that this can only be fixed for the general case (atom coordinates in arbitrary images) by first wrapping and then unwrapping the system. But even then, PBC have to be taken into account when calculating the COG of the reference group (and afterwards when computing distances, but that's already handled correctly I think).

If I'm not mistaken, this PR currently only works correctly for already unwrapped systems without PBC.

richardjgowers:
@zemanj yep we need proper unwrap support.  But this PR fixes the regression that happened when we started to use `pbc=True` inside the COG calculation.

orbeckst:
Could someone take the responsibility and merge? (I haven't really followed the conversation in depth but nothing has happened over the last 2d and the status is green so naively I'd think one could hit the button...)

Issue #2077, opened 2018-09-18
DummyReader should be convertable to a MemoryReader


kain88-de:
**Is your feature request related to a problem? Please describe.**

I would like to create a dummy universe with a trajectory attached. I can create the dummy universe. I do not know how I load a numpy array as a trajectory into it.

This would work nicely as a feature with our new trajectory writer shortcut. 

**Describe the solution you'd like**

```python
u = mda.Universe.empty(2, trajectory=True)
# shape is (n_frames, n_atoms, 3)
u.trajectory.add_coordinates(np.ones((100, 2, 3))
```

**Describe alternatives you've considered**
I could load a file from the test trajectories.

orbeckst:
Can you do a `u.transfer_to_memory()`?

kain88-de:
If iteration won't work then `transfer_to_memory` won't work either. It
iterates over  a trajectory.

richardjgowers:
Related to #2076 we should just nuke dummyreader and replace with memory

PR #1537, opened 2017-07-20
[WIP] u.transfer_to_memory also transfers the box

jbarnoud:
This commit makes the memory reader deal with a non-constant box
dimensions and makes `Universe.transfer_to_memory copy the box in
addition to the positions.

Until now, only one box was stored in the memory reader, making it
impractical for NPT systems.

Fixes #1041

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?
 - [X] Issue raised/referenced?

Issue #1041, opened 2016-10-20
transfer_to_memory doesn't copy dynamics dimension information


kain88-de:
### Expected behaviour

If I load a trajectory from an NPT simulation into memory I would expect that the memory reader also copies the dimension information for every frame
### Actual behaviour

Only the information from the first frame is used.

orbeckst:
I consider this a bug and because `ts.dimensions` is a top level attribute of `Timestep` it should also be copied.

However, it raises the question if the MemoryReader should also be aware of other per-timestep information – aren't we storing these data nowadays in a `ts.data` dict?

kain88-de:
The data dict is reader specific. I'm OK if that isn't copied. But other available top level attributes should be copied if available.

jbarnoud:
Do we have an equivalent of `Reader.timeseries` for the box shape? Fixing this is not difficult, but I have to iterate through the trajectory even when the faster `timeseries` is available.

kain88-de:
We don't have an equivalent. But libdcd reader does return it as well. I wouldn't mind adding a method that can also read the box information for all frames and returns them as an array. We should think if we want to add this to `timeseries` though to avoid iterating several times over a possibly large trajectory.

orbeckst:
I don't think we have but we should. Ideally you'd also get the additional time step information such as lambda or the more complicated stuff in some other formats such as trz (or tng?). Basically, it would be good to start with a data structure that allows some flexibility depending on available per frame data. E.g. a dataframe like array with named fields. We could start with a slow Python implementation in the base class and upgrade readers as we go along. 

Ideally you'd want to do coordinates and other timestep data together, of course. 

--
Oliver Beckstein
email: orbeckst@gmail.com

jbarnoud:
If libdcd has it, then it would make sense to expose it. Not much of a priority, but convenient at least in that context.

kain88-de:
Btw xarray sounds like the right package for a data format that such a universal time series should return.

orbeckst:
@jbarnoud if you fix this issue then we have dealt with all [defect for 0.17](https://github.com/MDAnalysis/mdanalysis/issues?utf8=%E2%9C%93&q=is%3Aopen%20is%3Aissue%20label%3Adefect%20milestone%3A0.17.0%20) issues.

orbeckst:
[...]

PR #2080, opened 2018-09-19
MemoryReader facelift

### LINKED ISSUE #2076: Iterating or frame selection on a dummy reader should give a sensible error.
**Expected behavior**

Iterating or frame selection on a dummy reader should give a sensible error.

**Actual behavior**
```python
<ipython-input-13-d91d7163bc22> in <module>()
      2 u.trajectory.n_frames
      3 u.add_TopologyAttr("masses")
----> 4 u.trajectory[0]

~/anaconda3/lib/python3.6/site-packages/MDAnalysis/coordinates/base.py in __getitem__(self, frame)
   1319         if isinstance(frame, numbers.Integral):
   1320             frame = apply_limits(frame)
-> 1321             return self._read_frame_with_aux(frame)
   1322         elif isinstance(frame, (list, np.ndarray)):
   1323             if isinstance(frame[0], (bool, np.bool_)):

~/anaconda3/lib/python3.6/site-packages/MDAnalysis/coordinates/base.py in _read_frame_with_aux(self, frame)
   1359         """Move to *frame*, updating ts with trajectory and auxiliary data."""
   1360         ts = self._read_frame(frame)
-> 1361         for aux in self.aux_list:
   1362             ts = self._auxs[aux].update_ts(ts)
   1363         return ts

~/anaconda3/lib/python3.6/site-packages/MDAnalysis/coordinates/base.py in aux_list(self)
   1542     def aux_list(self):
   1543         """ Lists the names of added auxiliary data. """
-> 1544         return self._auxs.keys()
   1545 
   1546     def _check_for_aux(self, auxname):

AttributeError: 'DummyReader' object has no attribute '_auxs'


```

``` python
u = mda.Universe.empty(2, trajectory=True)
u.trajectory.n_frames
u.add_TopologyAttr("masses")
u.trajectory[0]
```

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) 0.18.0
- Which operating system? Linux

richardjgowers:
Fixes #2076 #2077 #1041 #2081 
Replaces #1537

 - `MemoryReader` now has velocities and forces (which hopefully work identically to positions)
 - `MemoryReader` now has variable dimensions (ie NPT conditions)
 - `Universe.transfer_to_memory` takes advantage of all of the above
 - `Universe.empty` now uses a MemoryReader (now it does the above)

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


richardjgowers:
I'm a little confused what the expectations are for the MemoryReader wrt modifying attributes of Timesteps.  From what I can tell, it tries to give **views** not **copies** of the underlying arrays.  This means that if you edit the positions, change frame, your changes are saved (ie go back to that frame and you've still got the modified positions rather than the original).  Is this correct?

Therefore, does it make sense to make `.dimensions` **always** an array of dimensions so that it can always follow this pattern?

orbeckst:
Yes, I'd say the expectation is that everything is permanent. In this regard, MemoryReader is a very different beast from all the other Readers. But I think it's required so that MemoryReader can become a swiss-army knife for trajectory manipulation.

I *think* that the new transformations are also permanent for MemoryReader (@jbarnoud @davidercruz ?)

jbarnoud:
The transformations are indeed permanent on the memory reader. In practice, the transformation is applied on all frames in one go, and then purged from the list of transformation to not apply the same transformation again when changing frame.

The expectation with the memory reader is that changes done to the coordinates are kept, on the contrary to the other readers. Having the box following the same pattern is what I would expect.

package/MDAnalysis/coordinates/dummy.py
```diff
@@ -1,61 +0,0 @@
...
-"""
-from __future__ import absolute_import
-
-
-import numpy as np
-
-from .base import SingleFrameReaderBase
-
-
-class DummyReader(SingleFrameReaderBase):
```
jbarnoud: Shouldn't this be deprecated first?

richardjgowers: It was only used internally, and I think it's so buggy (it doesn't iterate?) I doubt anyone found any good use for it.

package/MDAnalysis/coordinates/memory.py
```diff
@@ -346,15 +400,26 @@ def parse_n_atoms(filename, order='fac', **kwargs):
...
             order=self.stored_order,
-            dimensions=self.ts.dimensions,
+            dimensions=dims,
+            velocities=vels,
+            forces=fors,
             dt=self.ts.dt,
             filename=self.filename,
         )
         new[self.ts.frame]
-        new.ts = self.ts.copy()
```
richardjgowers: this line was problematic as `base.Timestep.copy()` isn't respectful of MemoryReader behaviour, see #2081.  I don't think removing it causes any problems, as `new[self.ts.frame]` should populate the Timestep with everything needed

package/MDAnalysis/core/universe.py
```diff
@@ -627,26 +636,42 @@ def transfer_to_memory(self, start=None, stop=None, step=None,
         from ..coordinates.memory import MemoryReader
 
         if not isinstance(self.trajectory, MemoryReader):
-            # Try to extract coordinates using Timeseries object
-            # This is significantly faster, but only implemented for certain
-            # trajectory file formats
-            try:
-                coordinates = self.trajectory.timeseries(
```
orbeckst: Is it true that we do not have dedicated C-level timeseries anymore and thus the `self.trajectory.timeseries` was essentially just doing what happened in the `except` block? Or are there still cases where the `timeseries` approach would be beneficial?

package/MDAnalysis/core/universe.py
```diff
@@ -627,26 +636,42 @@ def transfer_to_memory(self, start=None, stop=None, step=None,
...
-                coordinates = np.array(coordinates)
+            n_frames = len(range(
+                *self.trajectory.check_slice_indices(start, stop, step)
+            ))
+            n_atoms = len(self.atoms)
+            pm_format = '{step}/{numsteps} frames copied to memory (frame {frame})'
+            pm = ProgressMeter(n_frames, interval=1,
+                               verbose=verbose, format=pm_format)
+            coordinates = np.zeros((n_frames, n_atoms, 3), dtype=np.float32)
+            # TODO: This assumes constant...
```
orbeckst: I don't understand this comment.

package/MDAnalysis/core/universe.py
```diff
@@ -627,26 +636,42 @@ def transfer_to_memory(self, start=None, stop=None, step=None,
...
+            ts = self.trajectory.ts
+            has_vels = ts.has_velocities
+            has_fors = ts.has_forces
+            has_dims = ts.dimensions is not None
+
+            if has_vels:
```
orbeckst: ternary operator for readability?
```python
velocities = np.zeros_like(coordinates) if has_vels or None
```
Does Python short-circuit, i.e., is the creation of `np.zeros_like(coordinates)` skipped if not needed?

orbeckst: Python (at least 3.5.3) properly short-circuits and does not execute the function in the skipped part:
```python
In [1]: def dome():
   ...:     print("did it")
   ...:     return 10
   ...:

In [2]: x = dome() if True else 0
did it

In [3]: x = dome() if False else 0
```

testsuite/MDAnalysisTests/coordinates/test_memory.py
```diff
@@ -200,3 +200,166 @@ def test_position_assignation(self, reader):
...
+        u.trajectory = mr_reader
+
+        return u
+
+    @pytest.mark.parametrize('attr', ['positions', 'velocities', 'forces', 'dimensions'])
+    def test_copying(self, mr_reader, attr):
+        mr2 = mr_reader.copy()
+
+        ts = mr2.ts
+        setattr(ts, attr, 7)
```
orbeckst: Remind me how this works: Are you setting e.g. `ts.positions` to 7? All positions in the array or are you replacing the array with the integer "7"?

If it is the latter, shouldn't we doing the former?

testsuite/MDAnalysisTests/coordinates/test_memory.py
```diff
@@ -200,3 +200,166 @@ def test_position_assignation(self, reader):
...
+        # move the Reader around, forcing updates of ts
+        ts = mr2[1]
+        ts = mr2[0]
+
+        assert_equal(getattr(ts, attr), 7)
```
orbeckst: array comparison? But the arrays should be float32 so you really ought to use `assert_almost_equal`.

package/MDAnalysis/core/universe.py
richardjgowers: Only DCDReader ever had a fast timeseries (and still does).  It didn't return dimensions information though, which is needed.  We could change `DCD.timeseries` to have `DCD.timeseries(...., dimensions=True)` to then return both coordinates and dimensions.

Then maybe the `base.Reader` could implement the entire `else:` branch below (I started this in #1400 once...)

So yeah maybe I should try and use the fast DCD routine here still since it still exists, then manually grab the dimensions..

package/MDAnalysis/coordinates/memory.py
```diff
@@ -360,17 +360,16 @@ def __init__(self, coordinate_array, order='fac',
...
-                raise ValueError("Provided dimensions array has shape {}. "
-                                 "This must be a array of shape (6,) or "
-                                 "(n_frames, 6)".format(dimensions.shape))
-            self.dimensions_array = dimensions
+            dimensions = np.zeros((self.n_frames, 6), dtype=np.float64)
+        elif dimensions.shape == (6,):
+            # single box, tile this to trajectory length
+            # allows modifying the box of some frames
+            dimensions = np.tile(dimensions, (self.n_frames, 1))
+        elif dimensions.shape != (self.n_frames, 6):
```
orbeckst: Does this need to be protected against `AttributeError` if anything is passed through that is not a numpy array?

package/MDAnalysis/core/universe.py
```diff
@@ -425,17 +429,19 @@ def empty(cls, n_atoms, n_residues=None, n_segments=None,
...
-        residue_segindex : numpy.array, optional
+        atom_resindex : array like, optional
+          mapping of atoms to residues, e.g. with 6 atoms,
+          `atom_resindex=[0, 0, 1, 1, 2, 2]` would put 2 atoms
+          into each of 3 residues.
+        residue_segindex : array like, optional
           mapping of residues to segments
         trajectory : bool, optional
-          if True, attaches a dummy reader to the Universe, therefore
+          if True, attaches a MemoryReader to the Universe, therefore
```
orbeckst: Make this (and others) a link
```reST
:class:`~MDAnalysis.coordinates.memory.MemoryReader`
```

package/MDAnalysis/coordinates/memory.py
```diff
@@ -360,17 +360,16 @@ def __init__(self, coordinate_array, order='fac',
...
-            if dimensions.shape == (6,):
-                # single box, tile this to trajectory length
-                # allows modifying the box of some frames
-                dimensions = np.tile(dimensions, (self.n_frames, 1))
-            if dimensions.shape != (self.n_frames, 6):
-                raise ValueError("Provided dimensions array has shape {}. "
-                                 "This must be a array of shape (6,) or "
-                                 "(n_frames, 6)".format(dimensions.shape))
-            self.dimensions_array = dimensions
+            dimensions = np.zeros((self.n_frames, 6), dtype=np.float64)
```
orbeckst: Did you decide to always have dimensions as an array, even if it contains zeros? (I assume this is more consistent with how `ts.dimensions` works, and it allows setting the box if required – more flexibility!) :+1:

package/MDAnalysis/coordinates/memory.py
```diff
@@ -326,8 +326,12 @@ def __init__(self, coordinate_array, order='fac',
             self.coordinate_array.shape[self.stored_order.find('a')]
 
         if velocities is not None:
-            if velocities.ndim == 2:
-                velocities = velocities[np.newaxis, :, :]
+            try:
+                if velocities.ndim == 2:
+                    velocities = velocities[np.newaxis, :, :]
+            except AttributeError:
```
zemanj: This just checks whether the `velocities` object has an attribute `ndim`. For type checking, better use `isinstance()`. But since that's not really "pythonic" anyway, I'd prefer calling `numpy.asarray(velocities)` before doing attribute checks.

package/MDAnalysis/coordinates/memory.py
```diff
@@ -338,8 +342,12 @@ def __init__(self, coordinate_array, order='fac',
             self.velocity_array = None
 
         if forces is not None:
-            if forces.ndim == 2:
-                forces = forces[np.newaxis, :, :]
+            try:
+                if forces.ndim == 2:
+                    forces = forces[np.newaxis, :, :]
+            except AttributeError:
```
zemanj: As above,I'd prefer using `numpy.asarray(forces)`.

package/MDAnalysis/coordinates/memory.py
```diff
@@ -361,14 +369,19 @@ def __init__(self, coordinate_array, order='fac',
...
+            try:
+                if dimensions.shape == (6,):
+                    # single box, tile this to trajectory length
+                    # allows modifying the box of some frames
+                    dimensions = np.tile(dimensions, (self.n_frames, 1))
+                elif dimensions.shape != (self.n_frames, 6):
+                    raise ValueError("Provided dimensions array has shape {}. "
+                                     "This must be a array of shape (6,) or "
+                                     "(n_frames, 6)".format(dimensions.shape))
+            except AttributeError:
```
zemanj: Again, better use `numpy.asarray(dimensions)`

richardjgowers:
Ok changed to use `np.asarray`, should be good to go hopefully

richardjgowers:
@MDAnalysis/coredevs if this gets reviewed I can finish up 0.19.0

wouterboomsma:
Thanks for notifying me, @orbeckst. I haven't followed recent development in MDAnalysis very closely, and therefore don't have anything intelligent to add to what has already been discussed. Good to see that MemoryReader is still being used. Thanks for putting in the effort in bringing this up to date @richardjgowers.

PR #2085, opened 2018-09-30
WIP, TST: add Python 2.7 ARMv8 testing to MDA CI.

### LINKED ISSUE #1989: Adding ARM arch to CI
The [recently-announced](http://blog.shippable.com/shippable-arm-packet-deliver-native-ci-cd-for-arm-architecture) free availability of ARM nodes for open source CI may eventually be of interest to us as we continue to expand platform / architecture support.

tylerjereddy:
Fixes #1989. Will probably need to adjust our repo settings to get the triggers working on this PR.

Python 2.7 only at the moment because of issues running our test suite with Python 3.7, and poor availability of 3.6 on shippable for the somewhat-exotic architecture (see related NumPy / Cython discussions for that..). Maybe when 3.7 is running on our x64 linux ci we can revisit for arm (works fine for other projects).

I think we'll need to set up a webhook somehow since I don't seem to have the option to do it for MDA through shippable itself, assuming we actually want to do this. Not sure why I couldn't just do it, since it worked for NumPy & I have lower privileges there unless we have some special block on auto-hook generation from services.

Results are [all green](https://app.shippable.com/github/tylerjereddy/mdanalysis/runs/37/1/console) on my fork with this feature branch. Good to double check that test skip counts match reasonably close to i.e., 2.7 on travis linux.

If the MDAnalysis project gets a separate S3 amazon cache with a different account, then we'll have to comment out subsections of the new `yml` file so that dependencies can be cached progressively before timeout. If we're lucky, the cache will be preserved and the build / test can run in about 20 minutes. Builds can be really slow on ARM for things like SciPy, so caching is really important.

There are 96 cores available on the nodes and we can use them all--helps for testing, but I'm not sure if our building mda itself has been optimized for parallel compile (tricky business usually with compile order / distutils constraints).

### DIFF
--- a/shippable.yml
+++ b/shippable.yml
@@ -1,0 +1,64 @@
+branches:
+    only:
+       - mda_arm_ci_shippable
+
+language: python
+
+python:
+    # NOTE: Docker options in pre_ci_boot section 
+    # are not allowed on restricted shared node pools.
+    # so haven't had luck getting Python 3.6 working
+    # on ARM; try 2.7 for now
+    - 2.7
+    # NOTE: some pretty major issues with our test infrastructure
+    # / dependencies on 3.7 it seems, so hold off on that for now
+
+runtime:
+    # use the free open source pool of nodes
+    # only for ARM platform
+    nodePool: shippable_shared_aarch64
+
+build:
+    ci:
+    # install dependencies
+    - sudo apt-get install libfreetype6-dev pkg-config libpng12-dev build-essential
+    - sudo apt-get install gcc gfortran libblas-dev liblapack-dev
+
+    - pip install --upgrade pip pytest
+    - pip install cython --cache-dir=/root/.cache/pip/wheels/$SHIPPABLE_PYTHON_VERSION
+    - pip install numpy --cache-dir=/root/.cache/pip/wheels/$SHIPPABLE_PYTHON_VERSION
+    - pip install scipy --cache-dir=/root/.cache/pip/wheels/$SHIPPABLE_PYTHON_VERSION
+    - pip install matplotlib --cache-dir=/root/.cache/pip/wheels/$SHIPPABLE_PYTHON_VERSION
+    - pip install mmtf-python --cache-dir=/root/.cache/pip/wheels/$SHIPPABLE_PYTHON_VERSION
+    - pip install biopython --cache-dir=/root/.cache/pip/wheels/$SHIPPABLE_PYTHON_VERSION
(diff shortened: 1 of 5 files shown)


shippable.yml
```diff
@@ -0,0 +1,64 @@
...
+    nodePool: shippable_shared_aarch64
+
+build:
+    ci:
+    # install dependencies
+    - sudo apt-get install libfreetype6-dev pkg-config libpng12-dev build-essential
+    - sudo apt-get install gcc gfortran libblas-dev liblapack-dev
+
+    - pip install --upgrade pip pytest
+    - pip install cython --cache-dir=/root/.cache/pip/wheels/$SHIPPABLE_PYTHON_VERSION
```
tylerjereddy: I suspect we may be able to remove this cache path code duplication, but messing with the cache is time-consuming if you accidentally reset it!

shippable.yml
```diff
@@ -0,0 +1,64 @@
...
+    - pip install joblib --cache-dir=/root/.cache/pip/wheels/$SHIPPABLE_PYTHON_VERSION
+    - pip install networkx --cache-dir=/root/.cache/pip/wheels/$SHIPPABLE_PYTHON_VERSION
+    - pip install hypothesis --cache-dir=/root/.cache/pip/wheels/$SHIPPABLE_PYTHON_VERSION
+
+    # build and test MDAnalysis
+    - cd package
+    - python setup.py develop --no-deps
+
+    - cd ../testsuite
+    - python setup.py develop --no-deps
```
tylerjereddy: do we benefit from a `-j` parallel distutils build? NumPy was modified to take modest advantage of this, but not sure about mda.

testsuite/MDAnalysisTests/analysis/test_encore.py
```diff
@@ -437,6 +437,8 @@ def test_clustering_two_ensembles(self, ens1, ens2):
                                                           "results: {0}".format(cluster_collection)
 
     
+    @pytest.mark.skipif(os.uname()[4] == 'aarch64',
+                        reason="currently fails on ARMv8 arch")
```
tylerjereddy: I've not had time to investigate the small number of ARMv8 / shippable-specific failures; compromise was to add skipifs for now. Not sure if xfail is preferred.

IAlibay:
@tylerjereddy given #2956 is this good to close?

tylerjereddy:
Yep, closing

PR #2090, opened 2018-10-04
made GSD windows support optional, now raises Error

### LINKED ISSUE #1923: GSD & Windows
See #1914 

TODO: Add nice handling of GSD under Windows

richardjgowers:
Fixes #1923

 - now optionally import GSD modules

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/coordinates/GSD.py
+++ b/package/MDAnalysis/coordinates/GSD.py
@@ -47,9 +47,11 @@ Classes
 from __future__ import absolute_import, division
 
 import numpy as np
-from . import base
-import gsd.hoomd
+import os
+if os.name != 'nt':
+    import gsd.hoomd
 
+from . import base
 
 class GSDReader(base.ReaderBase):
     """Reader for the GSD format.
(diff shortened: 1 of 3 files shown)


richardjgowers:
@tylerjereddy I don't have a windows box, but I think this should work.. some tests will need to be marked as xfail for windows (if they're not already)

package/MDAnalysis/coordinates/__init__.py
```diff
@@ -746,6 +747,13 @@ class can choose an appropriate reader automatically.
 from . import TXYZ
 from . import memory
 from . import MMTF
-from . import GSD
 from . import null
 from . import dummy
+
+if not os.name == 'nt':
```
tylerjereddy: You probably want `!=` here

package/MDAnalysis/topology/__init__.py
```diff
@@ -301,5 +303,12 @@
 from . import DLPolyParser
 from . import HoomdXMLParser
 from . import MMTFParser
-from . import GSDParser
 from . import MinimalParser
+
+if not os.name == 'nt':
```
tylerjereddy: Probably want `!=` here

richardjgowers:
Ok my last fix wasn't any good as it changed the namespace (ie `MDAnalysis.coordinates.GSD.GSDReader` wouldn't exist) so it might break scripts rather than fail gracefully as intended.  This fix is even more fixier than before.

tylerjereddy:
Ok, in it goes. I think we can safely ignore codecov patch diff coverage being a little off target here.

This should mean we can safely remove `gsd` installs to Windows builds, which were only acting as "mock" imports really.

PR #2093, opened 2018-10-08
WIP, TST: remove gsd & add Python 2.7 to appveyor config

tylerjereddy:
IIRC the need to "mock import" `gsd` was preventing us from building & testing MDAnalysis with Python 2.7 on Windows (`gsd` could be built but not actually used in 3.6, but not even built in 2.7 I think).

This follows up on #2090 to see if Appveyor is happy with Python 2.7 builds now.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -31,6 +31,7 @@ environment:
 
         - PYTHON_VERSION: 2.7
           PYTHON_ARCH: 64
+          MSVC_VERSION: "Visual Studio 10 Win64"
 
 init:
     - ps: Write-Host ${env:PYTHON} ${env:PYTHON_VERSION} ${env:PYTHON_ARCH}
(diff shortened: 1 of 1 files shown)


tylerjereddy:
I guess that was a tad optimistic

.appveyor.yml
```diff
@@ -29,6 +29,10 @@ environment:
           PYTHON_ARCH: 64
           MSVC_VERSION: "Visual Studio 10 Win64"
 
+        - PYTHON_VERSION: 2.7
+          PYTHON_ARCH: 64
+          MSVC_VERSION: "Visual Studio 10 Win64"
```
tylerjereddy: what motivated this change? if we want to support Python 2.7 on Windows we need to use older compilers and enforce C89 standard in our C source files I think. Maybe more trouble than it is worth though.

richardjgowers: I was just playing spot the difference between the two builds and hoping it was a typo.  I didn't realise we had fundamental problems like that.

kain88-de:
Did we decide to drop 2.7 for windows? We are not going for the windows build with 2.7 on conda-forge

richardjgowers:
@kain88-de it won't work atm as far as I know, if someone did a PR that fixed it then sure, so it's more a known issue than we've dropped 2.7 support on Windows

tylerjereddy:
We'd need to revert to C89 standard to get compilation working here so closing based on previous discussions.

PR #2100, opened 2018-10-11
Merge testsuite CHANGELOG/AUTHORS into main

### LINKED ISSUE #2050: Merge testsuite/CHANGELOG and testsuite/AUTHORS into package/
@orbeckst has [elected me as Frodo](https://github.com/MDAnalysis/mdanalysis/issues/2040#issuecomment-413027930).  We currently track changes to testsuite in a separate CHANGELOG, which is just confusing.  I'd rather just have these changes in `package/CHANGELOG`, maybe a separate tests section.  The only problem I see is keeping the old changes, some sort of weird script might work, but tbh `testsuite/CHANGELOG` is small enough that it's quicker to do it manually.

richardjgowers:
Fixes #2050

 - merged testsuite AUTHORS/CHANGELOG into package
 - replaced testsuite versions with symlink to package

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?


richardjgowers:
So this puts `AUTHORS` and `CHANGELOG` into the repo root.  This might mean they are included in a sdist of the package though... which might kill the installation process..

orbeckst:
>  This might mean they are included in a sdist of the package though

Did you mean "they are **not** included in a sdist..."?

Can you add `../AUTHORS` to `MANIFEST.in`? Or does a symlink work?

richardjgowers:
Hmm nearly there, the MANIFEST for testsuite doesn't work with a relative path:

```
copying ../package/AUTHORS -> MDAnalysisTests-0.19.1.dev0/../package
copying ../package/CHANGELOG -> MDAnalysisTests-0.19.1.dev0/../package
```

testsuite/MDAnalysisTests/__init__.py
```diff
@@ -98,11 +98,7 @@
 
 # keep in sync with RELEASE in setup.py
 __version__ = "0.19.1-dev"
-try:
-    from MDAnalysisTests.authors import __authors__
-except ImportError:
-    logger.info('Could not find authors.py, __authors__ will be empty.')
-    __authors__ = []
+from MDAnalysis import __authors__
```
richardjgowers: we never really had to dynamically generate this twice, since tests rely on main package

orbeckst:
Fake it with symlinks?

Or hack in an copy, similar to the shenanigans for the AUTHORS?

Oliver

richardjgowers:
Ok this is puzzling, importing `__authors__` from the main package breaks duecredit?

https://travis-ci.org/MDAnalysis/mdanalysis/builds/440685406?utm_source=github_status&utm_medium=notification

testsuite/MDAnalysisTests/__init__.py
```diff
@@ -98,7 +98,7 @@
 
 # keep in sync with RELEASE in setup.py
 __version__ = "0.19.1-dev"
-from MDAnalysis import __authors__
+#from MDAnalysis import __authors__
```
orbeckst: I think you cannot import MDAnalysis before the actual tests... at least I vaguely remember that we didn't allow it because of coverage.  IIRC, we do not import anything in MDAnalysis except when running tests. On these grounds alone, this line should not be here.

I am not sure why it would break the duecredit tests although I also remember that we're doing some fairly fiddly things to make sure that duecredit is either present or not.

orbeckst: Yes, I know, it's not in https://github.com/MDAnalysis/mdanalysis/wiki/UnitTests#requirements-for-tests-in-mdanalysistests ...

orbeckst:
@richardjgowers why does it fail on Windows
```
================================== FAILURES ===================================
___________________________ test_testsuite_authors ____________________________
    def test_testsuite_authors():
        from MDAnalysisTests import __authors__
>       assert len(__authors__) > 0, 'Could not find the list of authors'
E       AssertionError: Could not find the list of authors
E       assert 0 > 0
E        +  where 0 = len([])
utils\test_authors.py:33: AssertionError
----------- coverage: platform win32, python 3.6.6-final-0 -----------
```

orbeckst:
Is it because it cannot write the author file 
```
setup.py:133: UserWarning: Cannot write the list of authors. EOF before the list of authors
  '{}'.format(e))
```
and then it's not available?

Why does it fail to write AUTHORS on Windows but not on Linux/macOS? Or has it something to do with AppVeyor vs Travis environment??

richardjgowers:
The new `testsuite/AUTHORS` isn't created on Windows.  Maybe the symlink doesn't work?

orbeckst:
How do you provide `testsuite/AUTHORS` at the moment? Is this a symlink? If yes, that might be problematic on Windows (@tylerjereddy might know more).

orbeckst:
Oops, cross post.

orbeckst:
Could we do something like
```python
for authorpath in ['AUTHORS', '../package/AUTHORS']:
   if os.path.exists(authorpath):
       dynamic_authorlist(authorpath)
       break
else:
    raise IOError(errno.ENOENT, "AUTHORS not found.")
```

richardjgowers:
Can we not just remove AUTHORS from the testsuite?

orbeckst:
Is AUTHORS packaged in the distribution? (I think the answer is yes, judging from MANIFEST but can you confirm?)

orbeckst:
We don't use AUTHORS anywhere else in testsuite itself so if it gets packaged then I agree with removing it.

Any other opinions @MDAnalysis/coredevs ?

testsuite/setup.py
```diff
@@ -128,8 +128,9 @@ def dynamic_author_list():
 if __name__ == '__main__':
     try:
         dynamic_author_list()
-    except (OSError, IOError):
-        warnings.warn('Cannot write the list of authors.')
+    except (OSError, IOError) as e:
```
orbeckst: remove creation of `MDAnalysisTests/authors.py` and remove the code that imports `authors.py` in `__init__.py`.

richardjgowers:
@orbeckst relative links in MANIFEST don't work: 
https://github.com/MDAnalysis/mdanalysis/pull/2100#issuecomment-429155838

I can customise what `setup.py sdist` does so it copies the AUTHORS from `../package/` to the cwd... not sure if this is a bad idea though.  (I have to fix it to work from any location for a start)

WRT conda packaging, the conda build just uses whatever we put on pypi, so as long as pip works, conda works

testsuite/setup.py
```diff
@@ -42,79 +42,24 @@
...
-'''
-        print(header, file=outfile)
-
-        # Write the list of authors as a python list
-        template = u'__authors__ = [\n{}\n]'
-        author_string = u',\n'.join(u'    u"{}"'.format(name)
-                                    for name in authors)
-        print(template.format(author_string), file=outfile)
+class MDA_SDist(sdist.sdist):
+    def run(self):
```
orbeckst: Can you make it so that it tries to get a fresh AUTHORS from ../package when it can. But if this fails, it uses an existing AUTHORS (e.g., one that we packaged into the sdist tar ball!) In this way, you can sdist from an unpacked tar ball – which normally works.

If you copied AUTHORS then you should remove it. Otherwise you should keep it.

testsuite/setup.py
```diff
@@ -42,79 +42,24 @@
...
+        try:
+            print('copying AUTHORS')
+            shutil.copyfile('../package/AUTHORS', './AUTHORS')
+            super(MDA_SDist, self).run()
+        finally:
+            print('removing AUTHORS')
+            os.remove('AUTHORS')
```
orbeckst: Only remove if you successfully copied.

testsuite/setup.py
```diff
@@ -228,4 +168,5 @@ def dynamic_author_list():
           # had 'KeyError' as zipped egg (2MB savings are not worth the
           # trouble)
           zip_safe=False,
+          cmdclass={'sdist': MDA_SDist},
```
orbeckst: No pretty but not tolerable ;-).

richardjgowers:
Yeah ok I’ll tidy it up

richardjgowers:
@orbeckst this is finished, just the appveyor timed out, how do you restart those builds?

orbeckst:
When I log in to AppVeyor with my GitHub credentials I see a RE-BUILD PR button.

I am hitting it now; I don't know if I am the only one who sees the button (@tylerjereddy ?)

testsuite/MDAnalysisTests/__init__.py
```diff
@@ -98,11 +98,6 @@
 
 # keep in sync with RELEASE in setup.py
 __version__ = "0.19.1-dev"
-try:
```
jbarnoud: Since MDAnalysis is a dependency of MDAnalysisTests, we could have `__authors__ = MDAnalysis.__authors__`.

richardjgowers: @jbarnoud I tried this a few commits back, it messed up duecredit, and I think it also messes with coverage:
 https://github.com/MDAnalysis/mdanalysis/blob/b52e961d992abd43911e1cfb97b2468d92670e92/testsuite/MDAnalysisTests/__init__.py#L103

PR #2114, opened 2018-10-19
Issue 2046 lib distances rework continued

zemanj:
Fixes (partially) #2046. This is the fifth of a series of related PRs following PRs #2048, #2062, #2070, and #2083.

 - Moved `lib.distances._check_box()` to `lib.util.check_box()`
 - Ensured that all methods used in `capped_distance()` have the same cut-off criterion (`distances <= max_cutoff`). According to the `scipy.spatial.cKDTree.query_ball_tree` [docs](https://docs.scipy.org/doc/scipy-0.19.1/reference/generated/scipy.spatial.cKDTree.query_ball_tree.html) and according to some tests I ran, this should also be the case for the pKDTree method. In practice, this is rather pathological (comparing floating point values for equality, you know...) but should still be consistent.
- Ensured that none of the functions in `lib.distances` and `lib.nsgrid` make arbitrary assumptions about when a box is orthogonal. Since the number 90 has an exact float representation, we should refrain from defining arbitrary margins around that value. This is definitely a case where comparing floating point values for equality makes perfect sense (as it has always been done by `lib.mdamath.triclinic_vectors` and the box checks in `lib.distances` (now in `lib.utils`)).
 - Just like all other distance-related functions, `lib.nsgrid`now uses double precision for distance computation.
 - Prefixed all non-user-level classes in `lib.nsgrid` with an underscore.
 - Adjusted all affected tests accordingly.
 - Fixed up half-broken docs of `lib.nsgrid`.
 - Minor doc fix-ups in `lib.utils` I stumbled accross.

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/lib/nsgrid.pyx
+++ b/package/MDAnalysis/lib/nsgrid.pyx
@@ -65,11 +65,12 @@ neighboring cells only. Care must be taken to ensure that `cellsize` is
 greater than the desired search distance, otherwise all of the neighbours might
 not reflect in the results.
 
-
 .. [#] a pair correspond to two particles that are considered as neighbors .
 
-
 .. versionadded:: 0.19.0
+
+Classes
+-------
 """
 
 # Used to handle memory allocation
@@ -314,10 +315,9 @@ cdef class _PBCBox(object):
 cdef class NSResults(object):
     """Class to store the results
 
-    All the required outputs from :class:`FastNS` is stored in the
-    instance of this class. All the methods of :class:`FastNS` returns
-    an instance of this class, which can be used to  generate the desired
-    results on demand.
+    All outputs from :class:`FastNS` are stored in an instance of this class.
+    All methods of :class:`FastNS` return an instance of this class, which can
+    be used to  generate the desired results on demand.
     """
 
     cdef readonly real cutoff
@@ -403,7 +403,7 @@ cdef class NSResults(object):
 
         See Also
         --------
-        MDAnalysis.lib.nsgrid.NSResults.get_pairs
+        :meth:`~NSResults.get_pairs`
 
         """
 
@@ -457,15 +457,15 @@ cdef class NSResults(object):
             size ``m`` where m is the number of neighbours of
             query atom[i].
 
-        .. code-block:: python
+            .. code-block:: python
 
                 results = NSResults()
                 indices = results.get_indices()
 
-        ``indices[i]`` will output a list of neighboring
-        atoms of ``atom[i]`` from query atoms ``atom``.
-        ``indices[i][j]`` will give the atom-id of initial coordinates
-        such that ``initial_atom[indices[i][j]]`` is a neighbor of ``atom[i]``
+            ``indices[i]`` will be a list of neighboring atoms of
+            ``atom[i]`` from query atoms ``atom``. ``indices[i][j]`` will give
+            the atom-id of initial coordinates such that
+            ``initial_atom[indices[i][j]]`` is a neighbor of ``atom[i]``.
 
         """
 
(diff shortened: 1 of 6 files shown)


testsuite/MDAnalysisTests/lib/test_util.py
```diff
@@ -1668,3 +1668,55 @@ def test_dedent_docstring(text):
     doc = util.dedent_docstring(text)
     for line in doc.splitlines():
         assert line == line.lstrip()
+
+
+class TestCheckBox(object):
```
zemanj: I just copied this over from `lib/test_distances.py`. The code itself remained unchanged.

zemanj: I just copied this over from `lib/test_distances.py`. The code itself remained unchanged.

zemanj:
Sorry for the comment mess. GitHub is extremely unresponsive for me today. It took half an hour until I could finally see that all my seemingly unsuccessful attempts to add comments did in fact work. :angry:

zemanj:
Sorry for the comment mess. GitHub is extremely unresponsive for me today. It took half an hour until I could finally see that all my seemingly unsuccessful attempts to add comments did in fact work. :angry:
Hope you can see the comments, they only appear for me on every fifth page reload or so.

package/MDAnalysis/lib/nsgrid.pyx
```diff
@@ -88,21 +89,22 @@ DEF EPSILON = 1e-5
 
 ctypedef np.int_t ns_int
 ctypedef np.float32_t real
+ctypedef np.float64_t dreal
```
richardjgowers: Is there a performance different for moving to double precision?  (probably yes?) If so, maybe we should open an issue about precision in our distance calculations.  Especially as for this Cython it's easy to used a fused type (aka template)

richardjgowers:
@zemanj I'll find some time to go over this in detail, but it looks good at a glance

package/MDAnalysis/lib/nsgrid.pyx
zemanj: Good call, I'll run some benchmarks.

zemanj:
@richardjgowers I tested the performance impact of switching from single to double precision distance computation in `lib.nsgrid` with the following code:

```Python
>>> import MDAnalysis as mda
>>> from MDAnalysisTests.datafiles import TRR, TPR
>>> from MDAnalysis.lib.distances import capped_distance, self_capped_distance
>>> u = mda.Universe(TPR, TRR)
>>> water_oxygen = u.select_atoms('resname SOL and name OW')
>>> print(water_oxygen.n_atoms)
11084
>>> print(len(u.trajectory))
10

>>> %%timeit
>>> for ts in u.trajectory:
>>>     pairs, dists = capped_distance(water_oxygen.positions,
...                                   water_oxygen.positions,
...                                   15, box=ts.dimensions,
...                                   method='nsgrid')
>>> %%timeit
>>> for ts in u.trajectory:
>>>     pairs, dists = self_capped_distance(water_oxygen.positions,
...                                         15, box=ts.dimensions,
...                                         method='nsgrid')
```
Output for capped_distance:
`1 loop, best of 3: 24.3 s per loop` (single precision)
`1 loop, best of 3: 25.6 s per loop` (double precision)
:arrow_right: Performance loss: ~5%

Output for self_capped_distance:
`1 loop, best of 3: 21.1 s per loop` (single precision)
`1 loop, best of 3: 21.9 s per loop` (double precision)
:arrow_right: Performance loss: ~4%

So I'd say that the performance impact is rather negligible.

richardjgowers:
Looks like it needs a rebase onto develop

zemanj:
@richardjgowers I think the small performance impact indicates one or both of the following:
* the algorithm is strongly memory-bound
* the implementation might have some inefficiencies

I guess it's a bit of both since I already identified some rather crude thingies in `nsgrid.pyx` (such as the way neighboring cells are determined). However, I'll need to do some profiling before I can tell more.

zemanj:
I recently noticed an increasing amount of CI hiccups where jobs fail due to errors unrelated to the actual PR. Usually, those are I/O-related errors. I started collecting these cases and will open an issue once I gathered enough information.

PR #735, opened 2016-02-23
Analysis - Dielectric constant module

mattihappy:
Hello everybody,

I wrote a simple class to compute the trajectory average value of the static (isotropic) dielectric constant for systems simulated in tin foil boundary conditions (see [reference](http://www.tandfonline.com/doi/abs/10.1080/00268978300102721)).

As of now, I kept it very simple as it could be a good test bench/example for the consensus reached in the discussion about the analysis user interface in #719.

If any of you is familiar with the computation of the dielectric constant, let me know if you get consistent and meaningful results from the module (I run some tests and it appears to do so).

If everything works fine, I will add test cases.
### Example

``` python
import MDAnalysis as mda
from MDAnalysisTests.datafiles import DCD, PSF
from dielectric import DielectricConstant

universe  = mda.Universe(PSF,DCD)
sel_string = 'all'
selection = universe.select_atoms(sel_string)

diel = DielectricConstant(selection, temperature=300.)
diel.run()
```

PR #2118, opened 2018-10-23
Continued dielectric analysis module

PicoCentauri:
This PR continues the work of [PR#735](https://github.com/MDAnalysis/mdanalysis/pull/735). 

Adds the ability to calculate the static dielectric constant of a system. Furthermore also the directional dependent static constant can be calculated (useful if i.e. an electric field is applied). The code was compared to the gromaces `gmx dipole` module, giving the same results.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


testsuite/MDAnalysisTests/utils/test_units.py
```diff
@@ -54,7 +54,7 @@ def test_unicode(self):
 
     def test_unicode_encoding_with_symbol(self):
         try:
-            assert_equal(units.lengthUnit_factor[u"Å"], 1.0)
+            assert_equal(units.lengthUnit_factor[u"Å"], 1.0)
```
richardjgowers: ok this is confusing.  I tried changing this to the Å my keyboard has and it can't commit as it doesn't see any changes...

PicoCentauri: I just tried to reset the old utf character. Maybe it works....

richardjgowers:
@PicoCentauri just an FYI, once this is "ready" it will have to wait to get merged until we've got a 0.19.1 release out.  As this is a new feature, it has to go into a 0.20.0 release

package/MDAnalysis/analysis/dielectric.py
```diff
@@ -0,0 +1,146 @@
...
+          The step size through the trajectory in frames [0]
+    temperature : float
+        Temperature (Kelvin) at which the system has been simulated [298.15]
+    verbose : bool (optional)
+          Show detailed progress of the calculation if set to ``True``; the
+          default is ``False``.
+        
+    Attributes
+    ----------
+    results : dict
```
jbarnoud: I would like to see the units each value is in.

jbarnoud:
It looks like something went wrong with your merge/rebase.

package/MDAnalysis/analysis/dielectric.py
PicoCentauri: done

package/MDAnalysis/analysis/dielectric.py
```diff
@@ -0,0 +1,146 @@
...
+    then use the :meth:`run` method::
+    
+      diel = DielectricConstant(selection)
+      diel.run()
+
+    The static dielectric constant of the system will be returned
+    within the `results` variable of the object.
+    
+    
+    .. versionadded:: 0.19.1
```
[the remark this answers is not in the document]
PicoCentauri: done

package/MDAnalysis/analysis/dielectric.py
```diff
@@ -0,0 +1,146 @@
...
+                        "eps": np.zeros(3),
+                        "eps_mean": 0}
+
+    def _prepare(self):
+        error = "ERROR: Total charge of the selection is not zero."
+        assert_almost_equal(0, np.sum(self.charges), err_msg=error)
+        
+    def _single_frame(self):
+        # Make molecules whole
+        for frag in self.selection.fragments:
```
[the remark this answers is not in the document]
PicoCentauri: good idea! I added an option for making the wholes whole.

package/MDAnalysis/analysis/dielectric.py
```diff
@@ -0,0 +1,146 @@
...
+A tool to compute the average dipole moment and the static dielectric constant
+
+.. math::
+
+   \varepsilon = 1 + \frac{\langle M^2 \rangle - \langle M \rangle^2}
+                            {3 \varepsilon_ 0 V k_B T}
+
+for a system simulated in tin foil boundary conditions. For a distance
+
+References
```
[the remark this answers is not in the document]
PicoCentauri: I never worked with duecredrit but I added this according to your suggestions. seems to look fine.

package/MDAnalysis/analysis/dielectric.py
```diff
@@ -0,0 +1,146 @@
...
+from MDAnalysis.lib.mdamath import make_whole
+
+class DielectricConstant(AnalysisBase):
+    r"""Dielectric constant
+
+    Parameters
+    ----------
+    selection : AtomGroup
+          any atomgroup
+    start : int
```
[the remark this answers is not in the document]
PicoCentauri: done

PicoCentauri:
Sorry for the screwed rebase. It should be fine now.

package/MDAnalysis/analysis/dielectric.py
```diff
@@ -0,0 +1,148 @@
...
+                        "eps": np.zeros(3),
+                        "eps_mean": 0}
+
+    def _prepare(self):
+        error = "ERROR: Total charge of the selection is not zero."
+        assert_almost_equal(0, np.sum(self.charges), err_msg=error)
+
+    def _single_frame(self):
+        # Make molecules whole
+        if make_whole:
```
jbarnoud: The test cannot the `False`. Here, `make_whole` refers to `MDAnalysis.lib.mdamath.make_whole` and a function always test as `True`. It should be `self.make_whole`, which should be defined in the `__init__`.

The fact that id did not make the tests fail indicate that it is not covered by the tests.

package/MDAnalysis/analysis/dielectric.py
```diff
@@ -0,0 +1,148 @@
...
+import numpy as np
+from numpy.testing import assert_almost_equal
+
+from MDAnalysis.units import constants, convert
+from MDAnalysis.analysis.base import AnalysisBase
+from MDAnalysis.lib.mdamath import make_whole
+from MDAnalysis.due import due, Doi
+
+due.cite(Doi("10.1080/00268978300102721"),
+description="Dielectric analysis",
```
jbarnoud: The following lines (until the parenthesis that closes `due.cite`) should be aligned with `Doi` in agreement to PEP8.

package/MDAnalysis/analysis/dielectric.py
```diff
@@ -0,0 +1,148 @@
...
+
+    .. versionadded:: 0.20.0
+
+    """
+    def __init__(self, selection, temperature=300, make_whole=True, **kwargs):
+        super(DielectricConstant, self).__init__(selection.universe.trajectory,
+                                                 **kwargs)
+        self.selection = selection
+        self.temperature = temperature
+        self.charges = selection.charges
```
jbarnoud: If there is no charge defined in the topology, this will likely raise a not so useful `AttributeError`. It would be nice to catch it, and raise a `NoDataError` with a friendly message, instead.

package/MDAnalysis/analysis/dielectric.py
```diff
@@ -0,0 +1,148 @@
...
+
+.. math::
+
+   \varepsilon = 1 + \frac{\langle M^2 \rangle - \langle M \rangle^2}
+                            {3 \varepsilon_ 0 V k_B T}
+
+for a system simulated in tin foil boundary conditions.
+"""
+
+from __future__ import division
```
jbarnoud: You need to also import `absolute_import` from `__future__` to make pylint happy.

package/MDAnalysis/analysis/dielectric.py
PicoCentauri: I fixed this and added a test. But I didn't found a nice test file where this actually makes a different for the results.

package/MDAnalysis/analysis/dielectric.py
PicoCentauri: done

package/MDAnalysis/analysis/dielectric.py
PicoCentauri: Good idea. I added this and also a check if the total charge is non zero

package/MDAnalysis/analysis/dielectric.py
PicoCentauri: done

jbarnoud:
The test file has the molecule whole already. For the test to be useful, 
you need to break the molecules trhough PBC. You can try to wrap the 
molecules in the box first and see if they are broken. If they are not, 
you will have to translate them and wrap them in the box so that they break.

If you use the memory reader, you can do these operations in MDAnalysis 
at the beginning of the test.

zemanj:
There is a fundamental problem in the code: It only checks if the entire selections total charge is zero.
This check is certainly required but not sufficient. If the system has PBC and contains molecules carrying a net charge, the analysis will yield wrong results.

* In case of a system containing net-neutral molecules only, the system's itinerant dipole moment can be expressed as the sum of molecular dipole moments. That's correctly dealt with by making molecules whole in every time step.
* However, in cases where ionic species are present, there exists a generally non-vanishing cumulatve *inter*molecular dipole moment, which unfortunately depends on the choice of the origin and is therefore undefined.
In such a cases, one has to calculate the relative dielectric permittivity from the itinerant dipole moment's temporal derivative, a.k.a. current (sum over charge times velocity instead of over charge times position).

**tl;dr**: It has to be checked that *every fragment* in the selection is charge neutral.
If not, one can either try to compute the dielectric permittivity using velocities (if present) or just throw an error with an appropriate error message.

PicoCentauri:
@jbarnoud I was also unhappy with the test. Thanks for the hint. I will break some molecules and adjust the test.

@zemanj Thanks for pointing this out. I forgot to add this check. If you agree I will now only add a check for each fragment. But of course we can implement the term arising from current later.

zemanj:
@PicoCentauri Sure, implementing routines for calculating the permittivity of systems with ionic species can be done later. Just be aware of the fact that checking all `selection.fragments` for charge neutrality won't suffice since the selection might include the fragments only partially.

In that respect, it would be cool to have an optional `compound` parameter for `*Group.total_charge()` enabling it to compute the total charge per segment/residue/molecule/fragment in the spirit of `*Group.center()`, i.e., taking into account only atoms *belonging to the group*.
Then, you'd simply have to check for `np.allclose(selection.total_charge(compound='fragments'), 0.0)`.

PicoCentauri:
> it would be cool to have an optional compound parameter for *Group.total_charge()

Yes this seems very handy. I will look at your recent [commit](https://github.com/MDAnalysis/mdanalysis/commit/a2db387843874b669a63baa56cbb125fc6274486) and see what I can do.

testsuite/MDAnalysisTests/analysis/test_dielectric.py
```diff
@@ -0,0 +1,64 @@
...
+        # cut molecules apart
+        ag.universe.transfer_to_memory()
+        for ts in ag.universe.trajectory:
+            ag.wrap()
+    
+        eps_broken = DielectricConstant(ag, make_whole=False).run()
+        assert_almost_equal(eps_broken.results['eps_mean'], 721.711, decimal=1)
+        
+        eps_whole = DielectricConstant(ag, make_whole=True).run()
+        assert_almost_equal(eps_whole.results['eps_mean'], 3.874, decimal=1)
```
PicoCentauri: This test should work with and without making the fragments whole. But it seems that the method `group.unwrap()` for making fragments whole does not work here - at least when applied once. By running
```
        eps_whole = DielectricConstant(ag, make_whole=True)
        eps_whole.run()
        eps_whole.run()
```
you get a result close to the desired one. Does somebody have an idea?

zemanj: @PicoCentauri You set up the `results` dict in the class initializer instead of the `_prepare()` method.
Thus, if you execute the `run()` method a second time, the `results` will already be filled with data from the previous run.
Initialize or reset the `results` dict in the `_prepare()` method and you should be fine.

package/MDAnalysis/units.py
```diff
@@ -199,6 +199,8 @@
     'N_Avogadro': 6.02214129e+23,          # mol**-1
     'elementary_charge': 1.602176565e-19,  # As
     'calorie': 4.184,                      # J
+    'Boltzman_constant': 8.314462159e-3,   # KJ (mol K)**-1
+    'electric_constant': 5.526350e-3,      # As (Angstroms Volts)**-1
```
richardjgowers: Is there a reason we can’t just import from SciPy.constants now? Less code to maintain. In fact I’d sooner just import from scipy in the actual module rather than us maintain our own list of constants

testsuite/MDAnalysisTests/analysis/test_dielectric.py
richardjgowers: These two expected results should be different test cases

package/MDAnalysis/units.py
jbarnoud: Until very recently, scipy was an optional dependency. I guess it would be fine now.

PicoCentauri: I’m also preferring using the scipy constants directly. The question for this PR is: Should we keep it like this for now and update it later?
Should we introduce scipy.constants for the new constants here? Or first adjust the units.py?

jbarnoud: I would say that changing how the constants are handled is out of scope for this PR.

richardjgowers:
Sure it’d be nice to not further the (IMO) (post scipy being reqd) bad idea

testsuite/MDAnalysisTests/analysis/test_dielectric.py
PicoCentauri: I split the test case and also improved the creation of the results dictionary.

PicoCentauri:
Do we still want to merge this PR? I'm open for changes or adjustments and it would be a bit sad throw our work away...

jbarnoud (approved):

jbarnoud:
@PicoCentauri @richardjgowers @zemanj This PR looks like it has been sitting there for no particular reason. Is there anything blocking it going through?

package/MDAnalysis/analysis/dielectric.py
```diff
@@ -0,0 +1,168 @@
...
+# MDAnalysis: A Toolkit for the Analysis of Molecular Dynamics Simulations.
+# J. Comput. Chem. 32 (2011), 2319--2327, doi:10.1002/jcc.21787
+#
+
+r"""
+Dielectric --- :mod:`MDAnalysis.analysis.dielectric`
+===========================================================
+
+:Authors: Mattia Felice Palermo, Philip Loche
+:Year: 2021
```
IAlibay:
```suggestion
:Year: 2022
```

package/MDAnalysis/analysis/dielectric.py
```diff
@@ -0,0 +1,168 @@
...
+
+      diel = DielectricConstant(universe.atoms)
+      diel.run()
+      print(diel.results)
+
+    The static dielectric constant of the provided atomgroup is saved
+    within the :class:`~MDAnalysis.analysis.base.Results` attribute.
+
+
+    .. versionadded:: 2.0.0
```
IAlibay:
```suggestion
    .. versionadded:: 2.1.0
```

PicoCentauri:
No, I think it is ready.

IAlibay (changes requested):
couple small things

jbarnoud:
@IAlibay this needs you approval to go through.

IAlibay:
So quick pre-approval question here - once we merge this code in, who holds the copyrights to this code? I.e. if we do go ahead with a re-license in the near future, do we have a means of contacting @mattiafelice-palermo ?

testsuite/MDAnalysisTests/analysis/test_dielectric.py
```diff
@@ -0,0 +1,75 @@
+# -*- Mode: python; tab-width: 4; indent-tabs-mode:nil; coding:utf-8 -*-
+# vim: tabstop=4 expandtab shiftwidth=4 softtabstop=4 fileencoding=utf-8
+#
+# MDAnalysis --- https://www.mdanalysis.org
+# Copyright (c) 2006-2017 The MDAnalysis Development Team and contributors
```
IAlibay, line 5: Should the copyright notice span the correct range of years? Not sure what our policy on this has been.

PicoCentauri: I also don't know. But at least we could put 2022.

IAlibay: We've been super inconsistent here 😕 I guess it doesn't matter so much?

We mostly need a decision on https://github.com/MDAnalysis/mdanalysis/pull/2118#issuecomment-1032863726

mattiafelice-palermo: Dear @IAlibay, I apologize for being unresponsive but I have not been active in the field of computational chemistry for several years. Let me know what kind of approval you need in order to proceed with the PR.

jbarnoud: Hi @mattiafelice-palermo !

We are currently exploring if it is desirable and feasible to change the license of MDAnalysis to something more permissive. We will soon poll the contributors, but we need a way to reach them all.

We need a way to reach you for that poll or need to know if you are OK with your contribution to MDAnalysis to be relicensed to a more permissive open source license. If you are not OK with that, it is not a blocker, we'll just add that on the things to consider.

Cheers.

mattiafelice-palermo: @jbarnoud Thanks for the thorough explanation! You can reach me through GitHub, or if an email address is needed, I can PM it to you. I am more than ok with relicensing the code to a more permissive license :)

Issue #2125, opened 2018-10-25
PDB support on Windows


richardjgowers:
Opening this so there's a common place to debug this issue... will edit as things crystallise

Multi frame PDB reading doesn't work on Windows.  It seems to be because of how we use (byte?) file offsets to seek around the files.  

It's a little confusing as we do similar tricks for the xyz format and that seems to work fine, so maybe it's the fine details of how we implemented it in PDB...

richardjgowers:
[windows_written.pdb.zip.zip](https://github.com/MDAnalysis/mdanalysis/files/2521999/windows_written.pdb.zip.zip)

PR #2126, opened 2018-10-25
removed support for PDB on Windows...

richardjgowers:
Starts #2125

 - makes PDB reading fail on Windows (for now)

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/testsuite/MDAnalysisTests/core/test_atomselections.py
+++ b/testsuite/MDAnalysisTests/core/test_atomselections.py
@@ -344,13 +344,10 @@ class TestSelectionsAMBER(object):
         assert_equal(sel.names, ['HH31', 'HH32', 'HH33', 'HB1', 'HB2', 'HB3'])
 
 
-@pytest.mark.xfail(os.name == 'nt',
-                   strict=True,
-                   reason="Not supported on Windows yet.")
 class TestSelectionsNAMD(object):
     @pytest.fixture()
     def universe(self):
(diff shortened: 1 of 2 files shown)


testsuite/MDAnalysisTests/core/test_atomselections.py
```diff
@@ -344,13 +344,10 @@ def test_type(self, universe):
...
 
 
-@pytest.mark.xfail(os.name == 'nt',
-                   strict=True,
-                   reason="Not supported on Windows yet.")
 class TestSelectionsNAMD(object):
     @pytest.fixture()
     def universe(self):
-        return MDAnalysis.Universe(PSF_NAMD, PDB_NAMD)
+        return MDAnalysis.Universe(PSF_NAMD)
```
richardjgowers: These selections only use topology, so we only need a PSF-universe

richardjgowers:
I think once/if this merges, then everything "accessible" from Windows should work...

orbeckst:
This PR can only be an initial step towards getting this to work. We cannot ship a WIndows build that cannot work with PDB files.

richardjgowers:
@orbeckst ngl just uses a single PDB to grab topology, (ie it serialises a Universe via a PDB save/load) then coords are slurped straight from our Reader, so a multiframe PDB traj wouldn't work

I'm playing with a windows VM now... maybe the PDB thing is an easy fix....

PR #2127, opened 2018-10-26
PDB & Windows support.

### LINKED ISSUE #2125: PDB support on Windows

### LINKED ISSUE #2128: PDBReader fails to read PDB file written from Windows
Related to #2125 

PDB files written from Windows can't be read again, this gives:

```
IndexError                                Traceback (most recent call last)
<ipython-input-2-b3a4175a284e> in <module>()
----> 1 u = mda.Universe('new.pdb')

~/code/mdanalysis/package/MDAnalysis/core/universe.py in __init__(self, *args, **kwargs)
    333                 coordinatefile = None
    334 
--> 335             self.load_new(coordinatefile, **kwargs)
    336             # parse transformations
    337             trans_arg = kwargs.pop('transformations', None)

~/code/mdanalysis/package/MDAnalysis/core/universe.py in load_new(self, filename, format, in_memory, **kwargs)
    600         kwargs['n_atoms'] = self.atoms.n_atoms
    601 
--> 602         self.trajectory = reader(filename, **kwargs)
    603         if self.trajectory.n_atoms != len(self.atoms):
    604             raise ValueError("The topology and {form} trajectory files don't"

~/code/mdanalysis/package/MDAnalysis/coordinates/PDB.py in __init__(self, filename, **kwargs)
    314         self.n_frames = len(offsets)
    315 
--> 316         self._read_frame(0)
    317 
    318     @staticmethod

~/code/mdanalysis/package/MDAnalysis/coordinates/PDB.py in _read_frame(self, frame)
    403                 try:
    404                     # does an implicit str -> float conversion
--> 405                     occupancy[pos] = line[54:60]
    406                 except ValueError:
    407                     # Be tolerant for ill-formated or empty occupancies

IndexError: index 10 is out of bounds for axis 0 with size 10
```

Even away from Windows.

- Which version are you using? 0.19.1

richardjgowers:
Fixes #2125
Fixes #2128

 - PDBReader now finds offsets in binary mode, fixes reading PDB files with DOS line endings

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


richardjgowers:
Ok so I think PDB on Windows mostly works, the only remaining test failures are to do with **writing** PDB files.  Other failures look fairly minor.

I'm starting to think that the errors related to writing are how newlines are written....

jbarnoud:
IIRC, the magic you removed was there because python was eating the windows new line creating a discrepancy between what we were doing on strings and what was actually happening in the file. How is this solved now?

richardjgowers:
It’s not solved, but this narrows down the problem to only files written
from Windows, previously no files worked. I think I can fix the Windows
ascii issue in a later PR

richardjgowers:
@MDAnalysis/coredevs Ok this gets reading PDB files written from a unix system (ie most of them in the testsuite and the wild?) working on Windows. Round trip reading is a known issue - #2128 

I'd like to get this merged so we can release a 0.19.1 with Windows support

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -275,27 +275,16 @@ def __init__(self, filename, **kwargs):
...
         while line:
             # need to use readline so tell gives end of line
             # (rather than end of current chunk)
             line = pdbfile.readline()
-            # the tell() text opaque number striding is not portable
-            # to windows so compensate accordingly
-            if os.name == 'nt':
-                # for a text file mode these aren't necessarily
-                # bytes; var name for simplicity only
-                windows_overbyte += 1
```
kain88-de: doesn't this break files we wrote with windows? @tylerjereddy do you remember why you included those lines?

richardjgowers: Yeah this accounted for the extra newline char.  This reverts @tylerjereddy 's fix which fixed Windows-written files but broke Unix-written files (which includes the PDB itself).  [See here](https://github.com/MDAnalysis/mdanalysis/issues/2128#issuecomment-434490576) but I think there's a universal fix possible, this patch just reverts this change so that a 0.19.1 can read most(?) PDB files

testsuite/MDAnalysisTests/coordinates/test_pdb.py
```diff
@@ -512,6 +500,9 @@ def tearDown(self):
         del self.universe, self.multiverse, self.universe2
         del self.tmpdir
 
+    @pytest.mark.xfail(os.name == 'nt',
```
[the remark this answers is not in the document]
richardjgowers: Depends if our PDBWriter is actually incorrect to do this?  I think it's more our hacky ascii offset-seeking thing (which is v. cool) is choking on different newlines.

richardjgowers:
Ok this fixes reading DOS ending files, but it breaks `NamedStream` usage.... NamedStreams have a string buffer, we want a bytes buffer initially...

richardjgowers:
And now there's a nasty hack that hopefully gets everything green.

But I don't think this issue is really a PDB issue, rather it's an all multiframe ascii file issue... unless we're doing PDB file seeking differently, I'll poke around some more to confirm

package/MDAnalysis/coordinates/PDB.py
tylerjereddy: The Python 2.7.4 [standard docs for `file.tell()`](https://docs.python.org/2/library/stdtypes.html#file.tell) indicate:

> On Windows, `tell()` can return illegal values (after an fgets()) when reading files with Unix-style line-endings. Use binary mode ('rb') to circumvent this problem.

If we didn't use `anyopen` my impression is that we could fix the problem.

My comment indicating that the results are **opaque** numbers and not necessarily the number of bytes since the start of the file is discussed in some detail in [a SO post](https://stackoverflow.com/a/15935038/2942522).

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -272,41 +273,34 @@ def __init__(self, filename, **kwargs):
         models = []
         crysts = []
 
-        pdbfile = self._pdbfile = util.anyopen(filename)
+        # hack for streamIO
+        if isinstance(filename, util.NamedStream) and isinstance(filename.stream, cStringIO):
+            filename.stream = io.BytesIO(filename.stream.getvalue().encode())
+
+        pdbfile = self._pdbfile = util.anyopen(filename, 'rb')
```
tylerjereddy: [`anyopen` docs](https://www.mdanalysis.org/docs/documentation_pages/lib/util.html?highlight=anyopen#MDAnalysis.lib.util.anyopen) indicate that

> only the first letter of `mode` is used and thus any additional modifiers are silently ignored

I took a quick look at the source and we do seem to pass through the `mode`, but there's a bunch of handling code in there that may not really use it after all--are we sure this is using binary properly?

tylerjereddy:
The 2.7 failures look related; I added a review note above that may partly explain why 2.7 is a bit of a pain--that stack overflow post I mention talks about 2.x vs 3.x stuff too. It is a little confusing unfortunately.

If `anyopen` really does support binary mode reading it needs a doc update.

richardjgowers:
Ok cool, everything works and we now supports Windows.

orbeckst:
AWESOME!

--
Oliver Beckstein
email: orbeckst@gmail.com

tylerjereddy:
So, basically we don't care about 2.7, and `anyopen` docs are wrong so we can use binary mode to solve the issue?

richardjgowers:
@tylerjereddy I think this should work under 2.7 too, but I'll add a DOS-file to the testsuite to make sure (ie read a DOS file in linux & py2.7).  And yeah `anyopen` specs need tightening up because we're using `rb` now

richardjgowers:
[WINfiles.zip](https://github.com/MDAnalysis/mdanalysis/files/2546389/WINfiles.zip)

DOS endings ho!

richardjgowers:
@orbeckst ok added tests with files from Windows, so even on Linux we now test we can read these files.  Closer to that beer...

richardjgowers:
poke @orbeckst

orbeckst:
Hooray! 

(Btw, what about the other PDB-like formats, PQR, PDBQT, ... ? – Do we need to raise issues for them, too, or are these fine because we never try to seek them?)

richardjgowers:
Yeah I added tests for all ascii multi frame formats. Pdbqt etc are one and
done so no seeking

PR #2129, opened 2018-10-28
Fixed NumPy deprecations

### LINKED ISSUE #2122: replace deprecated np.matrix operations with operations on np.arrays
**Expected behavior**

No warnings regarding numpy use should be issued when using MDAnalysis.

**Actual behavior**

For example:
```
.../mdanalysis/package/MDAnalysis/analysis/align.py:288: PendingDeprecationWarning: the matrix subclass is not the recommended way to represent matrices or deal with linear algebra (see https://docs.scipy.org/doc/numpy/user/numpy-for-matlab-users.html). Please adjust your code to use regular ndarray.
```

```
pytest  MDAnalysisTests/analysis/test_align.py
```

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) 0.19.0-dev
- Which version of Python (`python -V`)? Python 3.6.6

### LINKED ISSUE #2123: replace deprecated numpy sequence slices with tuple slices
**Expected behavior**

No warnings regarding numpy use should be issued when using MDAnalysis.

**Actual behavior**

For example:
```
... mdanalysis/package/MDAnalysis/coordinates/memory.py:485: FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.
  ts._replace_positions_array(self.coordinate_array[basic_slice])
```

```
pytest  MDAnalysisTests/analysis/test_align.py
```

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) 0.19.0-dev
- Which version of Python (`python -V`)? Python 3.6.6

PicoCentauri:
Fixes #2122

Fixes #2123 

Fixed some NumPy depreciation warnings including those mentioned in issue #2122 and #2123.

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/analysis/align.py
+++ b/package/MDAnalysis/analysis/align.py
@@ -285,7 +285,7 @@ def rotation_matrix(a, b, weights=None):
     # so that R acts **to the left** and can be broadcasted; we're saving
     # one transpose. [orbeckst])
     rmsd = qcp.CalcRMSDRotationalMatrix(a, b, N, rot, weights)
-    return np.matrix(rot.reshape(3, 3)), rmsd
+    return np.array(rot.reshape(3, 3)), rmsd
 
 
 def _fit_to(mobile_coordinates, ref_coordinates, mobile_atoms,
--- a/package/MDAnalysis/analysis/align.py
+++ b/package/MDAnalysis/analysis/align.py
@@ -285,7 +285,7 @@ def rotation_matrix(a, b, weights=None):
     # so that R acts **to the left** and can be broadcasted; we're saving
     # one transpose. [orbeckst])
     rmsd = qcp.CalcRMSDRotationalMatrix(a, b, N, rot, weights)
-    return np.array(rot.reshape(3, 3)), rmsd
+    return rot.reshape(3, 3), rmsd
 
 
 def _fit_to(mobile_coordinates, ref_coordinates, mobile_atoms,
--- a/package/MDAnalysis/analysis/helanal.py
+++ b/package/MDAnalysis/analysis/helanal.py
@@ -749,15 +749,14 @@ def rotation_angle(helix_vector, axis_vector, rotation_vector):
 def vector_of_best_fit(origins):
     origins = np.asarray(origins)
     centroids = center(origins)
-    M = np.matrix(origins - centroids)
-    A = M.transpose() * M
+    M = np.array(origins - centroids)
+    A = np.dot(M.transpose(), M)
     u, s, vh = np.linalg.linalg.svd(A)
-    vector = vh[0].tolist()[0]
+    vector = vh[0]
     #Correct vector to face towards first residues
     rough_helix = origins[0] - centroids
     agreement = mdamath.angle(rough_helix, vector)
     if not (-np.pi / 2 < agreement < np.pi / 2):
-        vector = vh[0] * -1
-        vector = vector.tolist()[0]
+        vector *= -1
     best_fit_tilt = mdamath.angle(vector, [0, 0, 1])
     return vector, best_fit_tilt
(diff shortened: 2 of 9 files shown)


package/MDAnalysis/analysis/rms.py
```diff
@@ -587,7 +587,7 @@ def _prepare(self):
             #    (x~: selected coordinates, x: all coordinates)
             # Final transformed traj coordinates: x' = (x-x~_com)*R + ref_com
             self._rot = np.zeros(9, dtype=np.float64)  # allocate space
-            self._R = np.matrix(self._rot.reshape(3, 3))
+            self._R = np.array(self._rot.reshape(3, 3))
```
PicoCentauri: Should we note the change in the docstring?

package/MDAnalysis/analysis/align.py
```diff
@@ -285,7 +285,7 @@ def rotation_matrix(a, b, weights=None):
     # so that R acts **to the left** and can be broadcasted; we're saving
     # one transpose. [orbeckst])
     rmsd = qcp.CalcRMSDRotationalMatrix(a, b, N, rot, weights)
-    return np.matrix(rot.reshape(3, 3)), rmsd
+    return np.array(rot.reshape(3, 3)), rmsd
```
richardjgowers: if `rot` can do `.reshape` it's probably already an array? so just return `rot.reshape`

package/MDAnalysis/analysis/helanal.py
```diff
@@ -749,15 +749,14 @@ def rotation_angle(helix_vector, axis_vector, rotation_vector):
 def vector_of_best_fit(origins):
     origins = np.asarray(origins)
     centroids = center(origins)
-    M = np.matrix(origins - centroids)
-    A = M.transpose() * M
+    M = np.array(origins - centroids)
```
richardjgowers: same here, probably don't need the `np.array` call if the subtraction already works

package/MDAnalysis/analysis/rms.py
```diff
@@ -101,7 +101,7 @@
    ax.plot(time, rmsd[5], 'b--', label="NMP")
    ax.legend(loc="best")
    ax.set_xlabel("time (ps)")
-   ax.set_ylabel(r"RMSD ($\AA$)")
+   ax.set_ylabel(r"RMSD ($\\\\AA$)")
```
richardjgowers: ?

package/MDAnalysis/analysis/rms.py
richardjgowers: I don't think so, this change should mostly be invisible

Not sure you need to `np.array` call

package/MDAnalysis/coordinates/base.py
```diff
@@ -501,24 +501,23 @@ def copy_slice(self, sel):
            copy official attributes of the Timestep.
 
         """
-        # Detect the size of the Timestep by doing a dummy slice
```
richardjgowers: need this comment

package/MDAnalysis/analysis/rms.py
PicoCentauri: done

package/MDAnalysis/analysis/rms.py
PicoCentauri: Only one backslash throws a warning, so two are needed. 4 is definitely two much ;).

package/MDAnalysis/analysis/align.py
PicoCentauri: done

package/MDAnalysis/analysis/helanal.py
PicoCentauri: done

package/MDAnalysis/coordinates/base.py
PicoCentauri: done

richardjgowers:
Thanks @PicoCentauri another one down!

PR #2136, opened 2018-11-07
Fix #2132

### LINKED ISSUE #2132: capped_distance array: kernel exploding for a specific frame in this trajectory
**Expected behavior**

capped_distance should return something for each step of the trajectory 

**Actual behavior**

when iterating over the trajectory, the kernel dies as the 8th frame is being processed - so initially I suspected a memory issue.
however, when running only the 8th frame, the same thing (see below) happens. 
I have used other analysis functions on this trajectory including distance_array, so it seems specific to the recent implementation of capped_distance.

``` python
import MDAnalysis as mda
from MDAnalysis.lib import distances

u = mda.Universe(TOP, TRJ)
u.trajectory[7]
A=u.atoms.select_atoms("type os oh")
H=u.atoms.select_atoms("type HW")
#this works
distances.distance_array(H.positions, A.positions, box=u.dimensions)
#this doesn't
capped_distance(H.positions, A.positions, max_cutoff=3.0, box=u.dimensions)

```

- 0.19.0
- Python 3.6.6 :: Anaconda custom (64-bit)
- Mac OS X Mojave

[TOP](https://github.com/MDAnalysis/mdanalysis/files/2548915/PEG.prmtop.zip)
[TRJ](https://github.com/MDAnalysis/mdanalysis/files/2548810/PEG_03_prod.nc.zip)

zemanj:
Fixes #2132 

 - _NSGrid.coord2cellid() used to fail for coords close to box boundaries.
By defining the grid's cellsize as double precision, this is now fixed.
See Issue #2132 for details.
- increasing the grid cutoff is now only done when `force == False`.
This has never been an issue and is just for consistency.

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/testsuite/MDAnalysisTests/lib/test_nsgrid.py
+++ b/testsuite/MDAnalysisTests/lib/test_nsgrid.py
@@ -223,4 +223,4 @@ def test_nsgrid_probe_close_to_box_boundary():
     cutoff = 3.0
     conf = np.zeros((1, 3), dtype=np.float32)
     searcher = nsgrid.FastNS(cutoff, conf, box)
-    pairs, dists = searcher.search(ref)
+    results = searcher.search(ref)
--- a/testsuite/MDAnalysisTests/lib/test_nsgrid.py
+++ b/testsuite/MDAnalysisTests/lib/test_nsgrid.py
@@ -47,7 +47,6 @@ def run_grid_search(u, ref_id, cutoff=3):
 
     return searcher.search(searchcoords)
 
-
 def test_pbc_box():
     """Check that PBC box accepts only well-formated boxes"""
     pbc = True
@@ -214,3 +213,14 @@ def test_nsgrid_selfsearch(box, result):
         searchresults = searcher.self_search()
     pairs = searchresults.get_pairs()
     assert_equal(len(pairs)//2, result)
+
+def test_nsgrid_probe_close_to_box_boundary():
+    # FastNS.search used to segfault with this box, cutoff and reference
+    # coordinate prior to PR #2136, so we ensure that this remains fixed.
+    # See Issue #2132 for further information.
+    ref = np.array([[55.783722, 44.190044, -54.16671]], dtype=np.float32)
+    box = np.array([53.785854, 43.951054, 57.17597, 90., 90., 90.], dtype=np.float32)
+    cutoff = 3.0
+    conf = np.zeros((1, 3), dtype=np.float32)
+    searcher = nsgrid.FastNS(cutoff, conf, box)
+    pairs, dists = searcher.search(ref)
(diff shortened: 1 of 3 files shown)


zemanj:
@richardjgowers Good idea. Could you please verify that this currently fails:
```Python
import MDAnalysis as mda
from MDAnalysis.lib import distances
import numpy as np

ref = np.array([[55.783722, 44.190044, -54.16671]], dtype=np.float32)
conf = np.zeros((1, 3), dtype=np.float32)
box = np.array([53.785854, 43.951054, 57.17597, 90., 90., 90.], dtype=np.float32)
pairs, dist = distances.capped_distance(ref, conf, max_cutoff=1.0, box=box, method='nsgrid')
```

richardjgowers:
@zemanj that didn't work (ie it worked, therefore didn't work :)), not sure why.  This did though:

```python
In [4]: from MDAnalysis.lib import nsgrid

In [5]: fstns = nsgrid.FastNS(3.0, conf, box)

In [6]: fstns
Out[6]: <MDAnalysis.lib.nsgrid.FastNS at 0x10f5f0048>

In [7]: fstns.search(ref)
Segmentation fault: 11
```

zemanj:
For me both approaches fail. I'll use the one directly using FastNS for testing then.

testsuite/MDAnalysisTests/lib/test_nsgrid.py
```diff
@@ -214,3 +213,14 @@ def test_nsgrid_selfsearch(box, result):
...
+def test_nsgrid_probe_close_to_box_boundary():
+    # FastNS.search used to segfault with this box, cutoff and reference
+    # coordinate prior to PR #2136, so we ensure that this remains fixed.
+    # See Issue #2132 for further information.
+    ref = np.array([[55.783722, 44.190044, -54.16671]], dtype=np.float32)
+    box = np.array([53.785854, 43.951054, 57.17597, 90., 90., 90.], dtype=np.float32)
+    cutoff = 3.0
+    conf = np.zeros((1, 3), dtype=np.float32)
+    searcher = nsgrid.FastNS(cutoff, conf, box)
+    results = searcher.search(ref)
```
richardjgowers: I'd prefer if there was an `assert` somewhere in the test, even if it's just a regression test, ie check the distance is whatever comes out now.  Currently this could be anything

zemanj: I usually prefer to keep tests minimal but you're right, some assertions won't hurt here.

zemanj:
While we're there: Is there a good reason for `FastNS.search()` setting up a separate `FastNS.searchgrid` which is enforced to be identical to the already existing `FastNS.grid`? I can't think of one.

zemanj:
@richardjgowers Feel free to squash-merge if you're ok with this.

richardjgowers:
@zemanj the grids have to match because you iterate over positions in one grid and search in the other grid?  It looks like it's used after construction

richardjgowers:
thanks @zemanj !

zemanj:
@richardjgowers Of course the search grid has to be identical, and that's exactly what makes it obsolete. It's solely used to determine the search coords' cell ID, which one can simply get from `self.grid.coord2cellid(searchcoords_bbox[i])`. Anyway, it probably doesn't matter since the overhead cannot be more than O(n).

PR #2140, opened 2018-11-15
added experimental pickling support to Universe

richardjgowers:
This adds a basic form of pickling for a Universe, everything seems to work in the snippet below

```python
import dask
from dask import distributed

import MDAnalysis as mda
from MDAnalysisTests.datafiles import GRO, XTC

c = distributed.Client()

u = mda.Universe(GRO, XTC)

ag = u.atoms[:5]

def get_pos(u, ag, frame):    
    u.trajectory[frame]
    
    return ag.center_of_mass()

futures = []
for i in range(10):
    futures.append(dask.delayed(get_pos)(u, ag, i))

L = c.compute(dask.delayed(futures))

ref = [get_pos(u, ag, i) for i in range(10)]
```

WRT what sort of schedulers this should work for, I think this is only safe for distributed or multiprocessing, but not thread (as you can't share a Universe **and** play with frames)

@VOD555 I think this means you can send AtomGroups directly (no more remembering `.ix` tricks), and you don't reread the topology file (ie faster) because that half of the Universe is pickled.

### DIFF
--- a/package/MDAnalysis/coordinates/base.py
+++ b/package/MDAnalysis/coordinates/base.py
@@ -2078,6 +2078,20 @@ class ReaderBase(ProtoReader):
 
         self._ts_kwargs = ts_kwargs
 
+    @classmethod
+    def _unpickle_Reader(cls, filename, timestep, auxs, trans):
+        new_R = cls(filename)
+        new_R.add_transformations(trans)
+        for auxname, auxdata in auxs.items():
+            new_R.add_aux(auxname, auxdata)
+
+    def __reduce__(self):
+        return (self._unpickle_Reader,
+                (self.filename, self.ts, self._auxs, self._transformations))
+
+    def __len__(self):
+        return self.n_frames
+
     def copy(self):
         """Return independent copy of this Reader.
 
--- a/package/MDAnalysis/coordinates/base.py
+++ b/package/MDAnalysis/coordinates/base.py
@@ -2078,17 +2078,6 @@ class ReaderBase(ProtoReader):
 
         self._ts_kwargs = ts_kwargs
 
-    @classmethod
-    def _unpickle_Reader(cls, filename, timestep, auxs, trans):
-        new_R = cls(filename)
-        new_R.add_transformations(trans)
-        for auxname, auxdata in auxs.items():
-            new_R.add_aux(auxname, auxdata)
-
-    def __reduce__(self):
-        return (self._unpickle_Reader,
-                (self.filename, self.ts, self._auxs, self._transformations))
-
     def __len__(self):
         return self.n_frames
 
--- a/package/MDAnalysis/coordinates/base.py
+++ b/package/MDAnalysis/coordinates/base.py
@@ -208,7 +208,7 @@ from .. import units
 from ..auxiliary.base import AuxReader
 from ..auxiliary.core import auxreader
 from ..core import flags
-from ..lib.util import asiterable, Namespace
+from ..lib.util import asiterable, Namespace, anyopen
 
 
 class Timestep(object):
@@ -374,6 +374,15 @@ class Timestep(object):
             ts.forces = forces
 
         return ts
+
+    def __getstate__(self):
+        state = self.__dict__.copy()
+        state.pop('_reader', None)
+
+        return state
+
+    def __setstate__(self, state):
+        self.__dict__.update(state)
 
     def _init_unitcell(self):
         """Create custom datastructure for :attr:`_unitcell`."""
@@ -2029,6 +2038,33 @@ class ProtoReader(six.with_metaclass(_Readermeta, IOBase)):
         return ts
 
 
+class _AsciiPickle(object):
+    # mixin to allow ascii files to pickle it up with the pros
+    def __getstate__(self):
+        # Shallow copy of state of self
+        # shallow ie don't recursively copy all objects,
+        # just copy the references that __dict__ holds
+        stuff = self.__dict__.copy()
+        # don't pass the file handle over
+        del stuff['_f']
+        # instead pass enough metadata to reconstruct
+        stuff['_pickle_fn'] = self._f.name
+        # TODO: what other state does Reader hold?
+        # TODO: reconstruct file handle position
+        return stuff
+
+    def __setstate__(self, state):
+        self.__dict__.update(state)
+        self._f = anyopen(self._pickle_fn)
+
+        del self._pickle_fn
+
+class _BAsciiPickle(_AsciiPickle):
+    def __setstate__(self, state):
+        self.__dict__.update(state)
+        self._f = anyopen(self._pickle_fn, 'rb')
+
+        del self._pickle_fn
 
 class ReaderBase(ProtoReader):
     """Base class for trajectory readers that extends :class:`ProtoReader` with a
--- a/package/MDAnalysis/coordinates/chain.py
+++ b/package/MDAnalysis/coordinates/chain.py
@@ -353,6 +353,16 @@ class ChainReader(base.ProtoReader):
         self.ts = None
         self.rewind()
 
+    def __getstate__(self):
+        state = self.__dict__.copy()
+        state.pop('_ChainReader__chained_trajectories_iter', None)
+
+        return state
+
+    def __setstate__(self, state):
+        self.__dict__.update(state)
+        self.__chained_trajectories_iter = self._chained_iterator()
+
     def _get_local_frame(self, k):
         """Find trajectory index and trajectory frame for chained frame `k`.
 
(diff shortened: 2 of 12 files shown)


richardjgowers:
poke @VOD555 if this works then it makes pmda a hell of a lot simpler

package/MDAnalysis/coordinates/base.py
```diff
@@ -2078,6 +2078,20 @@ def __init__(self, filename, convert_units=None, **kwargs):
 
         self._ts_kwargs = ts_kwargs
 
+    @classmethod
+    def _unpickle_Reader(cls, filename, timestep, auxs, trans):
```
orbeckst: You are not using `timestep` here. Is this "for future use"? Then please document.

testsuite/MDAnalysisTests/core/test_universe.py
```diff
@@ -283,10 +283,14 @@ def test_load_multiple_args(self):
...
+    def test_pickle(self):
         u = mda.Universe(PSF, DCD)
-        with pytest.raises(NotImplementedError):
-            cPickle.dumps(u, protocol = cPickle.HIGHEST_PROTOCOL)
+
+        s = cPickle.dumps(u, protocol = cPickle.HIGHEST_PROTOCOL)
+
+        new_u = cPickle.loads(s)
+
+        assert_equal(u.atoms.names, new_u.atoms.names)
```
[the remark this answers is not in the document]
orbeckst: Also test something on the coordinate site.

testsuite/MDAnalysisTests/core/test_universe.py
```diff
@@ -283,10 +283,14 @@ def test_load_multiple_args(self):
         assert_equal(len(u.atoms), 3341, "Loading universe failed somehow")
         assert_equal(u.trajectory.n_frames, 2 * ref.trajectory.n_frames)
 
-    def test_pickle_raises_NotImplementedError(self):
+    def test_pickle(self):
```
orbeckst: Should also test that
- aux
- transformations

are properly pickled.

package/MDAnalysis/coordinates/base.py
richardjgowers: Ah no, I just forgot to use it.  There's potentially some state in `ts`, (ie if someone edits positions this is temporarily stored there)

richardjgowers:
@orbeckst I'm happy to push ahead and finish this once I get some confirmation it works for pmda too, else it's just a useless fix

package/MDAnalysis/core/universe.py
```diff
@@ -767,11 +767,15 @@ def __repr__(self):
...
+    def _unpickle_U(cls, top, traj, anchor):
+        u = cls(top, anchor_name=anchor)
+        u.load_new(traj)
+
+        return u
 
-    def __setstate__(self, state):
-        raise NotImplementedError
+    def __reduce__(self):
+        return (self._unpickle_U, (self._topology, self.trajectory.filename, self.anchor_name))
```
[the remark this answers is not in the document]
richardjgowers: Turns out a `Reader` just knows how to serialise itself (including file handles)

package/MDAnalysis/coordinates/base.py
richardjgowers: Reader just seems to work magically....

package/MDAnalysis/core/universe.py
```diff
@@ -738,7 +738,7 @@ def _gen_anchor_hash(self):
                 return self._anchor_uuid
             except AttributeError:
                 # store this so we can later recall it if needed
-                self._anchor_uuid = uuid.uuid4()
+                self._anchor_uuid = str(uuid.uuid4())
```
richardjgowers: uuid's didn't seem to survive the serialisation well, but if I just take a string it's all good (and still uuidesque)

package/MDAnalysis/core/universe.py
```diff
@@ -768,11 +768,22 @@ def __repr__(self):
...
             n_atoms=len(self.atoms))
 
-    def __getstate__(self):
-        raise NotImplementedError
+    @classmethod
+    def _unpickle_U(cls, top, traj, anchor):
+        """Special method used by __reduce__ to deserialise a Universe"""
+        # top is a Topology object at this point, but Universe can handle that
+        u = cls(top)
+        u.anchor_name = anchor
```
richardjgowers: this is the magic that lets future atomgroups find their 'home' universe

richardjgowers:
Ok looks like multiprocessing works, except for ascii format trajectories.  I think this is close to working and I'd like to put it in (for a project I have brewing) so @MDAnalysis/coredevs try and break this (while I fix up ascii formats)

orbeckst:
@VOD555 can you have a look at this work in the context of PMDA?

orbeckst:
These are surprisingly few changes.

orbeckst:
As I said in the review "More testing and documentation" (I think you also need CHANGELOG etc). It should be clearly documented what works and what doesn't. If not everything works then label it as "experimental". 

And it would be nice to fail with meaningful exceptions – if possible.

package/MDAnalysis/coordinates/chain.py
```diff
@@ -353,6 +353,16 @@ def __init__(self, filenames, skip=1, dt=None, continuous=False, **kwargs):
         self.ts = None
         self.rewind()
 
+    def __getstate__(self):
+        state = self.__dict__.copy()
+        state.pop('_ChainReader__chained_trajectories_iter', None)
```
richardjgowers: So I can't really reproduce the state of this generator cleanly (or at least in a simple way).  Tbh I've never used chainreader or really touched the code, so maybe I should just slap a 'NotImplementedError' on this and let someone who understands this fix it in the future.

package/MDAnalysis/coordinates/base.py
```diff
@@ -2029,6 +2038,33 @@ def _apply_transformations(self, ts):
...
+        # TODO: reconstruct file handle position
+        return stuff
+
+    def __setstate__(self, state):
+        self.__dict__.update(state)
+        self._f = anyopen(self._pickle_fn)
+
+        del self._pickle_fn
+
+class _BAsciiPickle(_AsciiPickle):
```
richardjgowers: This is kinda ugly, but to unpickle the file handle you need to remember the mode it was opened in. This isn't currently possible afaik

richardjgowers:
I did some hacking, all readers now pickle, to the best of my knowledge..... Lots of ugly hacks everywhere though

package/MDAnalysis/coordinates/chain.py
orbeckst: I think what you're doing is ok — like all the others, it resets to frame 0 on unpickling. 

Once pickle support is upgraded to also restore the frame index then moving to the appropriate frame should just use the usual `trajectory[frame]` and then let ChainReader take care of how to do it.

If you somehow managed to pickle `ChainReader.readers` in https://github.com/MDAnalysis/mdanalysis/blob/6d6b86202a16bff69fad3d8017aca19eed76578c/package/MDAnalysis/coordinates/chain.py#L265-L266 then it should work. Otherwise, we need a special getstate/setstate pair that blanks `readers` and rebuilds them from `ChainReader.filenames` (although we should actually store the original `filenames` because they can be a list of tuples — that's an addition to ChainReader itself).

testsuite/MDAnalysisTests/parallelism/test_multiprocessing.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+@pytest.fixture(params=[
+    (PSF, DCD),
+    (GRO, XTC),
+    (PDB_multiframe,),
+    (XYZ,),
+])
+def u(request):
+    if len(request.param) == 1:
```
orbeckst: maybe just
```python
def u(request):
   return mda.Universe(*request.param)
```?

Comment (orbeckst) [reply testsuite/MDAnalysisTests/parallelism/test_multiprocessing.py]: Then that would also work for ChainReader shenanigans
```python
u(TOPOL, TRJ1, TRJ2)
u(TOPOL, [TRJ1, TRJ2])
u(TOPOL, [(TRJ1, "pdb"), (TRJ2, "xtc"), TRJ3])
```

orbeckst:
@yuxuanzhuang as discussed on the GSoC call, put up your updated PR from your fork and then we can close this one as superseded. (Mention PR #2140 in the text of your PR so that they link nicely to each other.)

orbeckst:
Oops – sorry, you just did it 16 minutes ago – my page hadn't been refreshed.

I am closing this PR now (superseded by PR #2704)

PR #2141, opened 2018-11-15
polymer PersistenceLength attention

richardjgowers:
 - added polymer.sort_backbone
 - updates to polymer.PersistenceLength docs

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/analysis/polymer.py
+++ b/package/MDAnalysis/analysis/polymer.py
@@ -27,10 +27,37 @@ Polymer analysis --- :mod:`MDAnalysis.analysis.polymer`
 
 
 :Author: Richard J. Gowers
-:Year: 2015
+:Year: 2015, 2018
 :Copyright: GNU Public License v3
 
 This module contains various commonly used tools in analysing polymers.
+
+
+Persistence length worked example
+---------------------------------
+
+This example shows how to find the persistence length of a polymer
+using MDAnalysis.
+::
+
+  from MDAnalysis.tests.datafiles import TRZ_psf, TRZ
+  import MDAnalysis as mda
+  from MDAnalysis.analysis import polymer
+  u = mda.Universe(TRZ_psf, TRZ)
+  # this system is a pure polymer melt of polyamide,
+  # so we can select the chains by using the .fragments attribute
+  chains = u.atoms.fragments
+  # select only the backbone atoms for each chain
+  backbones = [chain.select_atoms('not name O* H') for chain in chains]
+  # sort the chains, removing any non-backbone atoms
+  sorted_backbones = [polymer.sort_backbone(bb) for bb in backbones]
+  lp = polymer.PersistenceLength(sorted_backbones)
+  # Run the analysis, this will average over all polymer chains
+  # and all timesteps in trajectory
+  lp = lp.run()
+  print('The persistence length is: {}'.format(lp.pl))
+  # always check the visualisation of this:
+  lp.plot()
 
 """
 from __future__ import division, absolute_import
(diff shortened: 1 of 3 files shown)


package/MDAnalysis/analysis/polymer.py
```diff
@@ -27,61 +27,158 @@
...
+
+    Returns
+    -------
+    sorted_backbone : AtomGroup
+      backbone in order, so `sorted_backbone[i]` is bonded to
+      `sorted_backbone[i - 1]` and `sorted_backbone[i + 1]`
+
+
+    .. versionadded:: 0.20.0
+    """
```
zemanj: Cosmetics: replace `AtomGroup` by `:class:` links to `~MDAnalysis.core.groups.AtomGroup` throughout the docstrings (if you have too much time :wink:)

package/MDAnalysis/analysis/polymer.py
```diff
@@ -27,61 +27,158 @@
...
+    if not len(backbone.fragments) == 1:
```
zemanj: Use `if backbone.n_fragments != 1` instead of `len(backbone.fragments)` once #2182 goes through (almost 7 times faster)

package/MDAnalysis/analysis/polymer.py
```diff
@@ -27,61 +27,158 @@
...
+        raise ValueError("Multiple fragments found in backbone.  "
+                         "backbone must be a single contiguous AtomGroup")
+    if max(len(at.bonded_atoms & backbone) for at in backbone) > 2:
+        # find which atom has too many bonds for easier debug
+        branches = [at for at in backbone
+                    if len(at.bonded_atoms & backbone) > 2]
```
zemanj: Here you do twice the work required. Get rid of line 99, simply compute the branches list and then do
```Python
if branches:
    raise ValueError(...)
```

package/MDAnalysis/analysis/polymer.py
```diff
@@ -27,61 +27,158 @@
...
+        raise ValueError(
+            "backbone is not linear.  "
+            "The following atoms have more than two bonds in backbone: "
+            "{}".format(','.join(str(a) for a in branches)))
+
+    caps = [atom for atom in backbone
+           if len(atom.bonded_atoms & backbone) == 1]
```
zemanj: `caps` could be empty in case of a ring structure. Needs check + test case

package/MDAnalysis/analysis/polymer.py
```diff
@@ -27,61 +27,158 @@
...
+    return sorted_backbone
+
+
 class PersistenceLength(AnalysisBase):
     r"""Calculate the persistence length for polymer chains
 
     The persistence length is the length at which two points on the polymer
-    chain become decorrelated.
+    chain become decorrelated.  This is determined by first measuring the
+    autocorrelation (:math:`C(n)`) of two bond vectors (:math:`\mathbf{a}`)
```
zemanj: ... two bond vectors (`\mathbf{a}_i`, `\mathbf{a}_{i+n}`)

package/MDAnalysis/analysis/polymer.py
```diff
@@ -27,61 +27,158 @@
...
+    .. math::
+
+       C(n) \approx \exp\left( - \frac{n l_B}{l_P} \right)
+
+    where :math:`l_B` is the average bond length, and :math:`l_P` is the
+    persistence length.
 
+    Parameters
+    ----------
+    atomgroups : list
```
zemanj: shouldn't that be an `iterable`? Also works with tuples or object arrays i think.

testsuite/MDAnalysisTests/analysis/test_persistencelength.py
```diff
@@ -62,31 +62,25 @@ def test_ag_ValueError(self, u):
...
 
     def test_plot_ax_return(self, p_run):
         '''Ensure that a matplotlib axis object is
         returned when plot() is called.'''
-        p_run.perform_fit()
         actual = p_run.plot()
         expected = matplotlib.axes.Axes
         assert isinstance(actual, expected)
 
-    def test_raise_NoDataError(self, p):
```
zemanj: Keep this test for `_perform_fit()` to make CodeCov happy

testsuite/MDAnalysisTests/analysis/test_persistencelength.py
```diff
@@ -62,31 +62,25 @@ def test_ag_ValueError(self, u):
...
+    def test_lb(self, p_run):
         assert_almost_equal(p_run.lb, 1.485, 3)
 
     def test_fit(self, p_run):
-        p_run.perform_fit()
-
         assert_almost_equal(p_run.lp, 6.504, 3)
         assert len(p_run.fit) == len(p_run.results)
 
     def test_plot_ax_return(self, p_run):
```
zemanj: Add case where an axis `ax` is passed:
```Python
assert p_run.plot(ax=ax) is ax
```

package/MDAnalysis/analysis/polymer.py
```diff
@@ -27,61 +27,158 @@
...
+    sorted_backbone : AtomGroup
+      backbone in order, so `sorted_backbone[i]` is bonded to
+      `sorted_backbone[i - 1]` and `sorted_backbone[i + 1]`
+
+
+    .. versionadded:: 0.20.0
+    """
+    if not len(backbone.fragments) == 1:
+        raise ValueError("Multiple fragments found in backbone.  "
+                         "backbone must be a single contiguous AtomGroup")
```
zemanj: Could also be zero fragments if `backbone` is empty. Maybe write
`"{} fragments found...".format(backbone.n_fragments)`

package/MDAnalysis/analysis/polymer.py
richardjgowers: I'm always mixed on this.  I view most docs via `sort_backbone?` in a jupyter notebook, so the docs being in sphinx style makes them less readable..

zemanj: Your call, it's just cosmetics for the html docs. Anyway, I see your point.

richardjgowers: Turns out we both win, sphinx creates the link even if we just write `AtomGroup`

zemanj: Cool! I'm a bit unsure if that works only because of line 67:
```Python
from ..core.groups import requires, AtomGroup
```
Since AtomGroup is included in the module's namespace, sphinx has no trouble resolving the link.
I'm afraid that this will be different for classes not present in the local namespace.

richardjgowers:
@zemanj I think that's everything addressed, over to you

PR #2142, opened 2018-11-18
Added more vdw radii

### LINKED ISSUE #2138: Topology table of vdw radii is pretty sparse
There is a dictionary of vdw radii in `MDAnalysis.topology.tables.vdwradii`.  It currently contains:

{'C': 1.5, 'F': 1.2, 'H': 0.4, 'N': 1.1, 'O': 1.05, 'S': 1.6, 'P': 1.6}

It should really cover more..

PicoCentauri:
Fixes #2138

Added more vdw radii in a raw table and convert them using the `kv2dict` function.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/topology/tables.py
+++ b/package/MDAnalysis/topology/tables.py
@@ -30,6 +30,7 @@ tables are dictionaries that are indexed by the element.
 
 .. autodata:: atomelements
 .. autodata:: masses
+.. autodata:: vdwradii
 
 The original raw data are stored as multi-line strings that are
 translated into dictionaries with :func:`kv2dict`. In the future,
@@ -42,6 +43,7 @@ The raw tables are stored in the strings
 
 .. autodata:: TABLE_ATOMELEMENTS
 .. autodata:: TABLE_MASSES
+.. autodata:: TABLE_VDWRADII
 """
 from __future__ import absolute_import
 from ._elements import NUMBER_TO_ELEMENT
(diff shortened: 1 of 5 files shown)


package/MDAnalysis/topology/tables.py
```diff
@@ -296,17 +298,76 @@ def kv2dict(s, convertor=str):
...
+#     J. Phys. Chem. 68 (3): 441-451. doi:10.1021/j100785a001.
+# [2] Rowland and Taylor (1996). "Intermolecular Nonbonded Contact Distances in Organic Crystal Structures:
+#                                 Comparison with Distances Expected from van der Waals Radii".
+#     J. Phys. Chem., 1996, 100 (18), 7384.7391. doi:10.1021/jp953141+.
+# [3] Mantina, et al. (2009). "Consistent van der Waals Radii for the Whole Main Group".
+#     J. Phys. Chem. A, 2009, 113 (19), 5806-5812. doi:10.1021/jp8111556.
+#------------ -----------
+# atomtype    r_vdw
+#------------ -----------
+H    1.10
```
richardjgowers: this value used to be 0.4

PicoCentauri:
I also noticed these changes, but I first wanted to use the current literature values. In principle we could also introduce such a factor for [guessing the bonds](https://github.com/MDAnalysis/mdanalysis/blob/93c48f6061efc43ed2c9f3c042c931883c8465dd/package/MDAnalysis/topology/guessers.py#L139)...

richardjgowers:
[We've already got a factor](https://github.com/MDAnalysis/mdanalysis/blob/93c48f6061efc43ed2c9f3c042c931883c8465dd/package/MDAnalysis/topology/guessers.py#L158) (which is from VMD I think) but with the new radii we'll need to find a new factor.  So with H being much smaller we'll have to go larger, but maybe this will cause trouble with other elements..

tylerjereddy:
See also: http://redmine.gromacs.org/issues/2094#change-14100

I guess MDA isn't really used for solvation directly like GROMACS is, but obviously changing VDW radii can have a pretty massive effect for some calculations. When GROMACS changed their tables a few years ago I lost a few million particles in some virus simulation hydrations.

orbeckst:
If the new radii + fudge factor gives better guesses than before then that's encouraging and provides motivation to change it. I am generally in favor of changing the radii. I don't know if these radii are outside of our bond guessers at all.

Perhaps you could add a more tests for the included systems where we have topology (bond) information. Basically, anything that has a 
- TPR
- PSF
- PRMTOP
- PDB with CONECT records

richardjgowers:
@PicoCentauri thanks for the new tests, looks like you've got this working.  Where did 0.55 come from? Looks like the gromacs default is 0.57, then we can just say we're in line with that?  Also can you add a quick note to CHANGELOG to say that guess_bonds is different & hopefully better

PicoCentauri:
> Where did 0.55 come from? Looks like the gromacs default is 0.57, then we can just say we're in line with that?

I started with the gromacs values, but I found the best agreement for a slightly smaller value. 0.56 also works for the two test cases, but for 0.57 the guesser will find too many bonds.

richardjgowers:
@PicoCentauri this is ready to merge, it just needs a rebase if you can do that?

PicoCentauri:
@richardjgowers done.

richardjgowers:
@PicoCentauri awesome, thank you!

PR #2156, opened 2018-12-04
TST: add Azure Windows CI

tylerjereddy:
Run 3 versions of Python in Windows tests in Azure CI.

10 free parallel jobs means this service is gaining popularity with NumPy / SciPy / Numba, etc. & can do Mac and Linux too.

I think there are a few Python 3.5 failures for MDA on Windows that need attention.

It may take a bit of effort to get the CI hooks working properly -- I went ahead and added the marketplace app for the repo, but there's also the online web config where I'll have to add some other devs etc. if we move forward with this.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,42 @@
+trigger:
+  # start a new build for every push
+  batch: False
+  branches:
+    include:
+      - develop
+jobs:
+- job: Windows
+  pool:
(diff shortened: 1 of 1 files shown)


tylerjereddy:
Here's the build I got on my fork: https://dev.azure.com/tylerjereddy/mdanalysis/_build/results?buildId=449

PR #2157, opened 2018-12-04
WIP, TST: add Azure Windows CI

tylerjereddy:
Supersede #2156 -- try to avoid chicken & egg `.yml` file problem by triggering Azure builds on this specific main repo branch for now instead of a branch from my fork.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,42 @@
+trigger:
+  # start a new build for every push
+  batch: False
+  branches:
+    include:
+      - develop
+jobs:
+- job: Windows
+  pool:
+    vmIMage: 'VS2017-Win2016'
+  strategy:
+    maxParallel: 6
+    matrix:
+        Python35-64bit:
+          PYTHON_VERSION: '3.5'
+          PYTHON_ARCH: 'x64'
+        Python36-64bit:
+          PYTHON_VERSION: '3.6'
+          PYTHON_ARCH: 'x64'
+        Python37-64bit:
+          PYTHON_VERSION: '3.7'
+          PYTHON_ARCH: 'x64'
+  steps:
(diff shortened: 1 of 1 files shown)


tylerjereddy:
Ok, you can find the Azure build in the CI list in this PR now -- if / when merged we may have to adjust the web client settings a little, although the `trigger` directive may work automatically with develop specified once merged.

As noted in the PR this supercedes, Python 3.5 Windows has a few failures apparently, but 3.6/ 3.7 should work just fine.

I think maybe coverage upload needs to be added to those jobs if desired, and perhaps using a few more cores, though I think we've had pytest-xdist issues lately?

tylerjereddy:
Also, let me know and I'll add/ hand over the web client / account by adding admins with desired emails / Microsoft Azure accounts.

richardjgowers:
@tylerjereddy not in a negative way, but what's the benefit of azure here over appveyor? Also travis windows looks like it's maybe on the horizon (https://blog.travis-ci.com/2018-10-11-windows-early-release) which is maybe worth thinking of? Would be nice to have our CI under one roof?

tylerjereddy:
It is mostly the 10 free parallel jobs and using Mac, OS X, Win platforms simultaneously I suppose. And maybe the open sourced commands / directives for the CI itself. Travis has been having lots of issues lately for some projects & Appveyor can take forever if you expand to test multiple configs in a matrix.

Feel free to close if this isn't of interest -- plenty of other stuff I can chip away. The ARMv8 CI stuff in my other PR may be more interesting with various ARMv8 supercomputers popping up.

tylerjereddy:
I think this PR is obsolete now since we've adopted Azure already. The benefits are fairly clear with the parallel jobs though, I think. We can expand it as testing requirements grow later..

PR #2159, opened 2018-12-04
changed encore test to be xfail for now

richardjgowers:
band aid for #2158 so we can still get PRs done in the meantime...

PR #2161, opened 2018-12-05
Update TPR version for Gromacs 2019

jbarnoud:
Update the TPX version and add new TPR files for the tests using gromacs
2019-beta3.

PR Checklist
------------
 - [X] Tests?
 - [X] Docs?
 - [x] CHANGELOG updated?


jbarnoud:
This gromacs update felt way to easy: I basically just had to add the new TPX version to the list of supported versions for the tests to pass. In the long run, we probably need tests with more exotic MDP options.

tylerjereddy:
Looks pretty clean -- a rebase once #2159 is merged should resolve the CI failures.

I guess as long as people report issues reading new formats & we add unit tests when fixing those issues it should get better over time, hopefully.

package/MDAnalysis/topology/tpr/setting.py
```diff
@@ -41,7 +41,7 @@
 from six.moves import range
 
 #: Gromacs TPR file format versions that can be read by the TPRParser.
-SUPPORTED_VERSIONS = (58, 73, 83, 100, 103, 110, 112)
+SUPPORTED_VERSIONS = (58, 73, 83, 100, 103, 110, 112, 116)
```
richardjgowers: @jbarnoud how risky is it to allow versions greater than this, but issue a warning that this is currently speculative support?  Ie how much does the tpr format change version to version?

testsuite/MDAnalysisTests/topology/test_tprparser.py
```diff
@@ -84,7 +85,7 @@ class TestTPRGromacsVersions(TPRAttrs):
     @pytest.fixture(params=[TPR400, TPR402, TPR403, TPR404, TPR405, TPR406,
                             TPR407, TPR450, TPR451, TPR452, TPR453, TPR454,
                             TPR455, TPR502, TPR504, TPR505, TPR510, TPR2016,
-                            TPR2018,])
+                            TPR2018, TPR2019B3])
```
richardjgowers: #effort

jbarnoud: ?

richardjgowers: It's just funny how easy it is to add the new tests

package/MDAnalysis/topology/tpr/setting.py
jbarnoud: The last two versions (2018 and 2019) did work with virtually no change. I am very uneasy about it, though, as the release note lists many features added and removed.

testsuite/MDAnalysisTests/topology/test_tprparser.py
jbarnoud: Yes! And it is especially nice since with nose I had to copy paste the whole class, which was super ugly!

package/MDAnalysis/topology/tpr/setting.py
orbeckst: Let's not allow greater versions. Much better to fail reliably than to work unreliably.

PR #2165, opened 2018-12-11
Capped fix

### LINKED ISSUE #2164: Strange segfault in capped_distance
I'm getting a strange intermittent segfault when using `capped_distance`.  This is basically searching for all pairs within 5.0 of different AtomGroups.  Running the code below will segfault for different pairs each time as well....

```python
import MDAnalysis as mda
import itertools

u = mda.Universe('single.dcd')

nfrags = 200
apf = len(u.atoms) // nfrags
frags = [u.atoms[apf * i:apf * (i+1)] for i in range(nfrags)]


for i, j in itertools.combinations(range(nfrags), 2):
    frag_i, frag_j = frags[i], frags[j]

    print(frag_i.indices.min(), frag_j.indices.min())
    capped_distance(frag_i.positions, frag_j.positions, 5.0)
```

richardjgowers:
Fixes #2164 but I don't understand why....

### DIFF
--- a/package/MDAnalysis/lib/distances.py
+++ b/package/MDAnalysis/lib/distances.py
@@ -691,20 +691,21 @@ def _nsgrid_capped(reference, configuration, max_cutoff, min_cutoff=None,
             # and supply the pseudobox
             # along with only one set of coordinates
             pseudobox = np.zeros(6, dtype=np.float32)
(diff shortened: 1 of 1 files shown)


richardjgowers:
hmm doesn't actually fix it..

PR #2169, opened 2019-01-02
Fixes the segmentation Fault in Capped distance

### LINKED ISSUE #2164: Strange segfault in capped_distance

ayushsuhane:
Fixes #2164 

 - Modified the NSGrid to handle the case when probe exactly touches the boundary of the box. 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


package/MDAnalysis/lib/nsgrid.pyx
```diff
@@ -901,6 +901,8 @@ cdef class FastNS(object):
                                     if probe[m] > self.box.c_pbcbox.box[m][m]:
                                         check = False
                                         break
+                                    if probe[m] == self.box.c_pbcbox.box[m][m]:
```
orbeckst: I am always wary of testing equality for floats. Is this a case when it is justified to do so?

Or should this also do a comparison within a certain epsilon?

package/MDAnalysis/lib/nsgrid.pyx
```diff
@@ -901,6 +901,8 @@ cdef class FastNS(object):
...
+                                        probe[m] -= 0.01*self.grid.cellsize[m]
```
orbeckst: What's the justification for 0.01? Does the exact value matter? Could it be made smaller?

package/MDAnalysis/lib/nsgrid.pyx
```diff
@@ -979,9 +981,11 @@ cdef class FastNS(object):
                                     if probe[m] < 0:
                                         check = False
                                         break
-                                    elif probe[m] >= self.box.c_pbcbox.box[m][m]:
+                                    elif probe[m] > self.box.c_pbcbox.box[m][m]:
                                         check = False
                                         break
+                                    if probe[m] == self.box.c_pbcbox.box[m][m]:
```
orbeckst: see comments above

richardjgowers:
@ayushsuhane thanks for the fix, but I think this is fixing the issue in the wrong locatio - `coord2cellid` shouldn't return invalid cellids ever. See #2175

package/MDAnalysis/lib/nsgrid.pyx
ayushsuhane: Yes @orbeckst, you are right. It is not a good idea to compare the equality of floats. While debugging, I found the error here which seemed to get resolved by this correction. But I agree, it wasn't robust enough.

ayushsuhane: As far as the algorithm is concerned, this specific patch tries to find the neighboring cell by moving in all 3 directions (27 combinations using probe). The problem here was that probe was moving out of the box dimensions (not quite, very close to the edge of the box).

ayushsuhane:
Thanks @richardjgowers #2175 does the correct treatment to not allow invalid cell ids. However, I figured the problem was specifically due to the artificial box creation. Earlier, the box was created with 1.2 times the  (minimum - maximum) dimensions in either directions. But it implicitly assumed that the cutoff would be 10% of the box size. I think it was the reason the error was not reproducible. Now I changed the box creation in a way that the new box's max dimension would be atleast "cutoff" distance away from  the nearest particle. Hope it gives away the right picture, and also the right solution. (Time shall tell :P)

richardjgowers:
Ok finally got round to checking, this works a charm, thanks @ayushsuhane (and @orbeckst for the poke)

PR #2176, opened 2019-01-12
[WIP] Generate doc stubs for transplanted methods

### LINKED ISSUE #1845: docs for many AtomGroup methods missing
### Expected behaviour
radius_of_gyration(), principal_axes() (see #1828 ), center_of_mass() and most other "compute something on an AtomGroup" (that require e.g., mass) should be in the docs for AtomGroup.

### Actual behaviour
They are not in the docs anymore (at least since 0.17.0 but probably since #363 / 0.16.0)

Look at https://www.mdanalysis.org/mdanalysis/documentation_pages/core/groups.html#MDAnalysis.core.groups.AtomGroup

I think the problem is that we only attached certain methods to AtomGroup when specific TopologyAttrs are present. The bare AtomGroup does not contain masses so many of the most used methods are invisible to the user (and a lot of our docs is broken, e.g., in the tutorial).

develop, but also published 0.17.0

jbarnoud:
The `transplant_stub.py` script introspects the groups and topology
attributes to write files in `documentation_pages/core` that contain the
documentation for the transplanted methods.

For the stubs to be picked up by sphinx, the docstring of the class to
document must contain

    .. include:: XXX.txt

where "XXX" is the name of the class.

A stub contains a table of the methods, their short descriptions, and
what topology attribute they require.

Fixes #1845

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,6 @@ package/doc/html/
 MDAnalysis.log
 # Ignore the authors.py files as they are generated files
 authors.py
+
+# Ignore generated documentation stubs
+package/doc/sphinx/source/documentation_pages/core/*.txt
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,6 +54,7 @@ matrix:
            BUILD_CMD="cd ${TRAVIS_BUILD_DIR}/package && python setup.py build_ext --inplace"
            INSTALL_HOLE="false"
            PIP_DEPENDENCIES="${PIP_DEPENDENCIES} sphinx-sitemap"
+           CONDA_DEPENDENCIES="${CONDA_DEPENDENCIES} tabulate"
 
     - env: NAME="Lint"
            PYLINTRC="${TRAVIS_BUILD_DIR}/package/.pylintrc"
--- a/package/MDAnalysis/core/groups.py
+++ b/package/MDAnalysis/core/groups.py
@@ -435,6 +435,9 @@ class GroupBase(_MutableBase):
     |                               |            | that are part of ``s`` or  |
     |                               |            | ``t`` but not both         |
     +-------------------------------+------------+----------------------------+
+
+    .. include:: GroupBase.txt
+
     """
     def __init__(self, *args):
         try:
@@ -1781,6 +1784,8 @@ class AtomGroup(GroupBase):
     new :class:`AtomGroup` for multiple matches.  This makes it difficult to use
     the feature consistently in scripts.
 
+    .. include:: AtomGroup.txt
+
 
     See Also
     --------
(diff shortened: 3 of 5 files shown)


jbarnoud:
It looks like this:

screenshot from 2019-01-12 19-35-02

-----

screenshot from 2019-01-12 19-35-32

zemanj:
That looks like a pretty pretty hacky but probably versatile solution.

The only thing I don't like about it is that the function signatures are the ones of the original methods and not the transplanted ones. This will most likely confuse users.
For example, `fragments` ends up being a *property* of `AtomGroup`, so it should appear without the parentheses. Likewise, methods such as `center_of_mass()` should not show up with `group` as their first argument.

A rather ugly and certainly more tedious solution is to manually insert the transplanted methods in the `.. autoclass::` sections of the respective module.

Example for `AtomGroup`:

```Python
"""
.. autoclass:: AtomGroup
   :members:
   :inherited-members:

   .. autodata::
      MDAnalysis.core.topologyattrs.Bonds.fragments
      :annotation:

   .. autofunction::
      MDAnalysis.core.topologyattrs.Masses.center_of_mass(pbc=None, \
      compound='group')
"""
```
The empty `:annotation:` section is necessary to suppress the nonsensical bogus annotation which is otherwise auto-generated by sphinx. Using `.. autoattribute::` instead of `.. autodata::` unfortunately doesn't yield the expected result.

orbeckst:
> Currently, the problem I have is that I do not know what to do with the references. If I keep the citations as is, sphinx complains about duplicates. If I remove them, then sphinx complains about missing citations. The only solution I can see is to introspect the modules doc to build a list of the citations, so I can remove only the reference that are already defined; though, I'd rather avoid doing that kind of book keeping.

Can you give an example of what the problematic case looks like?

orbeckst:
Would it help if we decide that the only place where we have references is in [documentation_pages/references.rst](https://github.com/MDAnalysis/mdanalysis/blob/develop/package/doc/sphinx/source/documentation_pages/references.rst)? We can add it to the [Documentation Guide](https://github.com/MDAnalysis/mdanalysis/wiki/WritingDocumentation) as a requirement. It would basically mean that within the docs you have to properly cite
1. check if citations is in `references.rst`
2. add if necessary
3. use the citation key

Or perhaps we look into something like [mcmtroffaes/sphinxcontrib-bibtex](https://github.com/mcmtroffaes/sphinxcontrib-bibtex), at the cost of making doc writing a bit more convoluted.

jbarnoud:
Superseded by #2418.

Issue #1926, opened 2018-05-31
RecordType TopologyAttr doesn't copy


kain88-de:
### Expected behaviour
`u.copy()` always works

### Actual behaviour

```python
TypeError                                 Traceback (most recent call last)
<ipython-input-5-ec0d294eeb2a> in <module>()
      1 u = mda.Universe(PDB)
----> 2 ref = u.copy()

~/anaconda3/lib/python3.6/site-packages/MDAnalysis/core/universe.py in copy(self)
    337     def copy(self):
    338         """Return an independent copy of this Universe"""
--> 339         new = self.__class__(self._topology.copy())
    340         new.trajectory = self.trajectory.copy()
    341         return new

~/anaconda3/lib/python3.6/site-packages/MDAnalysis/core/universe.py in __init__(self, *args, **kwargs)
    300 
    301             # generate and populate Universe version of each class
--> 302             self._generate_from_topology()
    303 
    304             # Load coordinates

~/anaconda3/lib/python3.6/site-packages/MDAnalysis/core/universe.py in _generate_from_topology(self)
    348         # Put Group level stuff from topology into class
    349         for attr in self._topology.attrs:
--> 350             self._process_attr(attr)
    351 
    352         # Generate atoms, residues and segments.

~/anaconda3/lib/python3.6/site-packages/MDAnalysis/core/universe.py in _process_attr(self, attr)
    848                   'segment': self._topology.n_segments}
    849         logger.debug("_process_attr: Adding {0} to topology".format(attr))
--> 850         if (attr.per_object is not None and len(attr) != n_dict[attr.per_object]):
    851             raise ValueError('Length of {attr} does not'
    852                              ' match number of {obj}s.\n'

~/anaconda3/lib/python3.6/site-packages/MDAnalysis/core/topologyattrs.py in __len__(self)
    262     def __len__(self):
    263         """Length of the TopologyAttr at its intrinsic level."""
--> 264         return len(self.values)
    265 
    266     def __getitem__(self, group):

TypeError: len() of unsized object
```

### Code to reproduce the behaviour

``` python
import MDAnalysis as mda
from MDAnalysisTests.datafiles import PDB

u = mda.Universe(PDB)
u.copy()

```

### Currently version of MDAnalysis:
(run `python -c "import MDAnalysis as mda; print(mda.__version__)"`)
dev

richardjgowers:
[...]

PR #2179, opened 2019-01-19
Fixed copying record_type attribute

### LINKED ISSUE #2178: Universe.copy() error with PDB file
**Expected behavior**

I expect to have a new, independent copy of the universe. It works 

**Actual behavior**

I receive the following error.

```
<ipython-input-423-f4aece42e018> in <module>
----> 1 y = x.copy()

/opt/homebrew/anaconda3/envs/mdanalysis/lib/python3.7/site-packages/MDAnalysis/core/universe.py in copy(self)
    354     def copy(self):
    355         """Return an independent copy of this Universe"""
--> 356         new = self.__class__(self._topology.copy())
    357         new.trajectory = self.trajectory.copy()
    358         return new

/opt/homebrew/anaconda3/envs/mdanalysis/lib/python3.7/site-packages/MDAnalysis/core/universe.py in __init__(self, *args, **kwargs)
    310 
    311             # generate and populate Universe version of each class
--> 312             self._generate_from_topology()
    313 
    314             # Load coordinates

/opt/homebrew/anaconda3/envs/mdanalysis/lib/python3.7/site-packages/MDAnalysis/core/universe.py in _generate_from_topology(self)
    365         # Put Group level stuff from topology into class
    366         for attr in self._topology.attrs:
--> 367             self._process_attr(attr)
    368 
    369         # Generate atoms, residues and segments.

/opt/homebrew/anaconda3/envs/mdanalysis/lib/python3.7/site-packages/MDAnalysis/core/universe.py in _process_attr(self, attr)
    887                   'segment': self._topology.n_segments}
    888         logger.debug("_process_attr: Adding {0} to topology".format(attr))
--> 889         if (attr.per_object is not None and len(attr) != n_dict[attr.per_object]):
    890             raise ValueError('Length of {attr} does not'
    891                              ' match number of {obj}s.\n'

/opt/homebrew/anaconda3/envs/mdanalysis/lib/python3.7/site-packages/MDAnalysis/core/topologyattrs.py in __len__(self)
    262     def __len__(self):
    263         """Length of the TopologyAttr at its intrinsic level."""
--> 264         return len(self.values)
    265 
    266     def __getitem__(self, group):

TypeError: len() of unsized object
```

``` python
import MDAnalysis as mda
from MDAnalysis.tests.datafiles import PDB, XTC

u = mda.Universe(PDB)
v = u.copy()
# or
u = mda.Universe(PDB, XTC)
v = u.copy()
```

0.19.2 and Python 3.7 (but it also fails in 2.7)

richardjgowers:
Fixes #2178 #1926 

 - can now initialise record_type attr off boolean values

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -22,6 +22,9 @@ Changes
   * added official support for Python 3.7 (PR #1963)
   * stopped official support of Python 3.4 (#2066, PR #2174)
 
+Enhancements
+  * added record_type to select_atoms language (PR #2179)
+
 Fixes
   * fixed gcc support in MacOS (Issue #2162, PR #2163)
   * fixed error when reading bonds/angles/dihedrals from gsd file (Issue #2152,
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -37,6 +37,8 @@ Fixes
   * Fixed Empty select_atoms string, now raises warning and returns empty
     AtomGroup (Issue #2112)
   * Read TPR files from Gromacs 2019 (PR #2161)
+  * Fixed copying Universes created from PDB files (record_type attribute
+    was problematic) (Issue #2178 #1926, PR #2179)
 
 
 11/06/18 richardjgowers
--- a/package/MDAnalysis/core/groups.py
+++ b/package/MDAnalysis/core/groups.py
@@ -2232,6 +2232,7 @@ class AtomGroup(GroupBase):
             record_type *record_type*
                 for selecting either ATOM or HETATM from PDB-like files.
                 e.g. ``select_atoms('name CA and not record_type HETATM')``
+
         **Boolean**
 
             not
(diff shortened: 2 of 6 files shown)


richardjgowers:
@jbarnoud yeah it's fragile, so maybe could do something like that.  Downsides would be it would be an O(N) string operation and PDB parsing is already slow.  My argument would be that this stuff isn't really used by users but by the readers, so less checking is ok?

jbarnoud:
@richardjgowers The `where` is already O(N) so it would not change the overall complexity. It would make it a bit slower, though; but will it be any significant? If it is, or really thought to be, I am fine with skipping the check. In this case, though, it may be worth having a word about it in the doc.

richardjgowers:
@jbarnoud or I could even just store this as another `dtype=object` array, so whatever you put in you get back out.  Not sure why I encoded this as binary tbh... memory footprint reasons?

richardjgowers:
@jbarnoud ok I simplified it, it's just a string field now

jbarnoud:
Travis blocks because of #2180, but I guess Travis will block everything because of that issue.

PR #2182, opened 2019-01-22
New compounds 'molecules' and 'fragments' for COM and COG computations

zemanj:
 - `*Group.center_of_mass()`, `*Group.center_of_geometry()`, and `*Group.center()` are now able to compute centers (of mass/geometry/whatever) per molecule or per fragment by setting the optional `compound` kwarg to `'molecules'` or `'fragments'`, respectively.
- In order to facilitate the per-fragment computation, the format of the items stored in `Universe._fragdict` now contain tuples of the form `(fragnum, fragment)` instead of just `fragment`.
- The fragnum(s) of an `Atom` /  `AtomGroup` can be accessed via the `Atom.fragnum` / `AtomGroup.fragnums` attribute (similar to `molnum` / `molnums`).

Tests and changelog entries are yet to come.

PR Checklist
------------
 - [ ] Tests?
 - [x] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/core/groups.py
+++ b/package/MDAnalysis/core/groups.py
@@ -707,7 +707,7 @@ class GroupBase(_MutableBase):
 
 
         .. versionchanged:: 0.19.0 Added `compound` parameter
-        .. versionchanged:: 0.20.0 Added `'molecules'` and `'fragments'`
+        .. versionchanged:: 0.20.0 Added ``'molecules'`` and ``'fragments'``
             compounds
         """
 
@@ -788,12 +788,12 @@ class GroupBase(_MutableBase):
 
     @warn_if_not_unique
     def center_of_geometry(self, pbc=None, compound='group'):
-        """Center of geometry (also known as centroid) of the group.
+        """Center of geometry of (compounds of) the group.
 
         Computes the center of geometry (a.k.a. centroid) of
         :class:`Atoms<Atom>` in the group. Centers of geometry per
-        :class:`Residue` or per :class:`Segment` can be obtained by setting the
-        `compound` parameter accordingly.
+        :class:`Residue`, :class:`Segment`, molecule, or fragment can be
+        obtained by setting the `compound` parameter accordingly.
 
         Parameters
         ----------
@@ -1827,11 +1827,12 @@ class AtomGroup(GroupBase):
         # REMOVE in 1.0
         #
         # is this a known attribute failure?
-        if attr in ('fragments',):  # TODO: Generalise this to cover many attributes
+        if attr in ('fragments', 'fragnums'):  # TODO: Generalise this to cover many attributes
             # eg:
             # if attr in _ATTR_ERRORS:
             # raise NDE(_ATTR_ERRORS[attr])
-            raise NoDataError("AtomGroup has no fragments; this requires Bonds")
+            raise NoDataError("AtomGroup has no {}; this requires Bonds"
+                              "".format(attr))
         elif hasattr(self.universe._topology, 'names'):
             # Ugly hack to make multiple __getattr__s work
             try:
@@ -3123,8 +3124,9 @@ class Atom(ComponentBase):
     """
     def __getattr__(self, attr):
         """Try and catch known attributes and give better error message"""
-        if attr in ('fragment',):
-            raise NoDataError("Atom has no fragment data, this requires Bonds")
+        if attr in ('fragment', 'fragnum'):
+            raise NoDataError("Atom has no {} data, this requires Bonds"
+                              "".format(attr))
         else:
             raise AttributeError("{cls} has no attribute {attr}".format(
                 cls=self.__class__.__name__, attr=attr))
--- a/package/MDAnalysis/core/groups.py
+++ b/package/MDAnalysis/core/groups.py
@@ -652,8 +652,9 @@ class GroupBase(_MutableBase):
         """Weighted center of (compounds of) the group
 
         Computes the weighted center of :class:`Atoms<Atom>` in the group.
-        Weighted centers per :class:`Residue` or per :class:`Segment` can be
-        obtained by setting the `compound` parameter accordingly.
+        Weighted centers per :class:`Residue`, :class:`Segment`, molecule, or
+        fragment can be obtained by setting the `compound` parameter
+        accordingly.
 
         Parameters
         ----------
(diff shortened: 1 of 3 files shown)


zemanj:
If anyone has an idea how to store/access unique fragment indices more efficiently than by changing the format of the `Universe._fragdict` items, please let me know.

package/MDAnalysis/core/universe.py
```diff
@@ -1012,9 +1014,9 @@ def _fragdict(self):
         frags = tuple([AtomGroup(np.sort(ix), self) for ix in frag_indices])
 
         fragdict = {}
-        for f in frags:
+        for i, f in enumerate(frags):
             for a in f:
-                fragdict[a] = f
+                fragdict[a] = (i, f)
```
richardjgowers: if we do a `namedtuple` here, we can do `fragdict[self].ix` / `fragdict[self].fragment` later

package/MDAnalysis/core/topologyattrs.py
```diff
@@ -1673,29 +1678,99 @@ class Bonds(_Connection):
...
+        :class:`Atoms<MDAnalysis.core.groups.Atom>` bonded to this
+        :class:`~MDAnalysis.core.groups.Atom`."""
         idx = [b.partner(self).index for b in self.bonds]
         return self.universe.atoms[idx]
 
     transplants[Atom].append(
         ('bonded_atoms', property(bonded_atoms, None, None,
                                   bonded_atoms.__doc__)))
 
+    def fragnum(self):
```
richardjgowers: not a huge fan of this, but I see why it's needed

maybe fragindices to denote it as something we've defined, in line with resindices

package/MDAnalysis/core/universe.py
zemanj: Cool! That sounds much more readable.

package/MDAnalysis/core/topologyattrs.py
zemanj: Sure, no problem. Just out of interest: why aren't 'molnums' called 'molindices'?

richardjgowers: molnum is directly read from a tpr file, ie it's gromacs' opinion of molecules

zemanj: Thx, now that makes sense.

package/MDAnalysis/core/universe.py
```diff
@@ -1012,9 +1017,10 @@ def _fragdict(self):
         frags = tuple([AtomGroup(np.sort(ix), self) for ix in frag_indices])
 
         fragdict = {}
-        for f in frags:
+        fraginfo = collections.namedtuple('fraginfo', 'ix, fragment')
+        for i, f in enumerate(frags):
             for a in f:
-                fragdict[a] = f
+                fragdict[a.ix] = fraginfo(i, f)
```
richardjgowers: Atoms should already just be hashing via their `.ix` (and Universe)

package/MDAnalysis/core/groups.py
```diff
@@ -715,27 +730,41 @@ def center(self, weights, pbc=None, compound='group'):
...
 
-        if compound.lower() == 'group':
+        comp = compound.lower()
+        if comp == 'group':
             if pbc:
                 coords = atoms.pack_into_box(inplace=False)
             else:
                 coords = atoms.positions
+            # Weights don't make sense if we have less than 2 atoms:
+            if len(self) < 2:
```
richardjgowers: if we're going to put in optimisations for `len == 1` things we need tests too

package/MDAnalysis/core/universe.py
zemanj: `Atom` is derived from `ComponentBase`, so only their `ix` (not their `universe`) is used for hasing.
The reason why I decided to change the `_fragdict` keys to directly use atom indices instead of atoms is performance:

Test system: AtomGroup consisting of 500 fragments with 25 atoms per fragment

Old timings for first access / subsequent access:
* `ag.fragments`: 195 ms / 75 ms
* `ag.fragindices`: 140 ms / 50 ms

New timings for first access / subsequent access:
* `ag.fragments`: 140 ms / 20 ms  --> speedup 1.39 / 3.75
* `ag.fragindices`: 130 ms / 2.7 ms --> speedup 1.08 / 18.5

Since `Universe._fragdict` is cashed and usually (if not always) static, this makes computations involving fragments and fragindices *a lot* faster.

package/MDAnalysis/core/groups.py
zemanj: What about zero weights? This happens if dummy atoms are present.
I personally don't have a problem with defining the weighted center of a single atom as its position, *regardless* of a zero weight.

The same is *not* true when computing the weighted center of a group of atoms and the sum of the weights is zero.

package/MDAnalysis/core/universe.py
richardjgowers: Hm, that's odd.  So `Atom.__hash__` is just this:

https://github.com/MDAnalysis/mdanalysis/blob/develop/package/MDAnalysis/core/groups.py#L3028

Is the extra hash call causing the slowdown?

zemanj: I guess so... in my test system that saves 50,000 function calls (`Atom.__hash__` and `hash(self.ix)` for each `Atom`), and these are expensive in Python. Not sure if it's the only reason, though.

zemanj: I think I got it now. The former routine to construct `AtomGroup.fragments` used to be this:
```Python
return tuple(sorted(
            set(a.fragment for a in self),
key=lambda x: x[0].index
))
```
The implicit loop `a.fragment for a in self` needs to create an `Atom` instance for each atom in the group, followed by a call to `Atom.fragment`.

That is much more expensive than the new routine, where the implicit loop is
```
fragdict[aix].fragment for aix in self.ix
```
This only requires instantiating the `self.ix` numpy array, and its members are primitive types.

package/MDAnalysis/core/groups.py
zemanj: The behavior is now that the result vector is `nan` if `weights.sum() == 0`. This is now consistently enforced regardless of group size (also for groups with single atoms). I hope I have it all covered now.

richardjgowers: So it sounds like you settled on `np.nan` for zero sum weights? Seems right

package/MDAnalysis/core/universe.py
richardjgowers: Ok cool, thanks for the performance tweak!

zemanj:
@richardjgowers Thanks to @bsipocz the [ci-helpers fix](https://github.com/astropy/ci-helpers/pull/383) is already in place, so it's nice to know that this passes even without #2184 merged. Good to go, I guess.

zemanj:
Ok now it's *really* good to go from my side.

zemanj:
@richardjgowers ty for reviewing!

PR #2038, opened 2018-08-10
Added unwrap/wrap transformations

davidercruz:
Adds a wrap transformation based on atoms.wrap and a unwrap transformation based on @richardjgowers  's _cutil.make_whole
Replaces PR #2020 

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?
 - [ ] Issue raised/referenced?

PR #2189, opened 2019-01-30
*Group.unwrap

zemanj:
Fixes (in part) #1004, #1185

 - Added `*Group.unwrap(self, box=None, compound='fragments', reference='com',
               inplace=True)`

Since I'd really like to have this feature and there wasn't too much progress going on lately in PRs #2012 and #2038, I decided to go ahead and get this going. Using the method in other methods such as `*Group.center*()` and the like should be straightforward but is not yet covered here.

Since in most cases, boxes are orthorhombic and the majority of molecules are not split accross boundaries, I implemented a short-cut so that only those fragments that are potentially broken are made whole. This yielded a speed-up of a factor between 4 and 5 compared to a simple loop blindly calling `make_whole(fragment)` on every fragment in the group (as it's implemented in PR #2038 at the moment).

Tests and changelog entry are yet to come.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


kain88-de:
You can try a DNA as an example. In GROMACS the two DNA strands are modeled as individual fragments. An unwrap algorithm should however still keep both strands together. With Gromacs that required several passes over the trajectory. I will check if I can share an example configuration.

zemanj:
We still have the problem that using `make_whole()` on periodic molecules (think, for example, of an infinitely periodic graphene sheet or carbon nanotube) is likely to yield unwanted results.
But maybe... if people try unwrapping periodic molecules, we should give them what they deserve. :grin:

Another issue is that `find_fragments()` does not correctly handle dummy particles since fore some reason these don't show up as bonded atoms. For example, unwrapping TIP4P water from a gromacs topology / trajectory fails.

zemanj:
@kain88-de Regarding the DNA example: Right now my implementation does not keep groups of individual compounds together, since this would additionally require making the COMs or COGs of the compounds whole. Such a functionality should either go into a separate method or it should be accessible via an optional kwarg. Furthermore, this would indeed require a method which is able to compute the necessary shifts from the (naturally non-bonded) COMs or COGs of the respective compounds.

jbarnoud:
Si this is supposed to supersede #2012, right? Is everything from #2012 in here already?

zemanj:
@jbarnoud That's a bit hard to judge since #2012 still has a large number of TODOs and as far as I could see several lines of code which need fixing. Nevertheless, I'm pretty sure my implementation already covers most of what #2012 is aiming at.

I didn't fully get what the `choose_image()` method is supposed to be good for; IMHO unwrapping the centers of compounds (see my [comment above](https://github.com/MDAnalysis/mdanalysis/pull/2189#issuecomment-458867881)) to keep them as closely together as possibe cannot be achieved by solely arranging the compounds' centers around a specific position in the cell.
Maybe @richardjgowers can shed some light on what exactly the `choose_image()` function accomplishes?
The "keep clusters of compounds intact" functionality is not yet included in this PR, though (but will be very soon).
What I found problematic in #2012 is that the `unwrap()` method effectively acts on atoms which aren't part of the group (`group.fragments >= group`!). This is handled properly in this PR (I hope).

package/MDAnalysis/core/topologyattrs.py
```diff
@@ -1799,6 +1800,137 @@ def n_fragments(self):
...
                                  n_fragments.__doc__)))
 
+    def unwrap(self, box=None, compound='fragments', reference='com',
+               inplace=True):
+        """Move atoms in this group so that bonds within the
+        group's compounds aren't split accross periodic boundaries.
+
+        Parameters
+        ----------
+        box : array_like or None, optional
```
richardjgowers: maybe just force this to always be `self.dimensions`?  Thinking of how to reduce the complexity of this

package/MDAnalysis/core/topologyattrs.py
```diff
@@ -1799,6 +1800,137 @@ def n_fragments(self):
...
+            If ``True``, coordinates are modified in place.
+
+        Returns
+        -------
+        coords : numpy.ndarray
+            Unwrapped atom coordinate array of shape ``(n, 3)``.
+
+        Note
+        ----
+        * This is the inverse transformation of packing atoms into the primary
```
richardjgowers: There's some ascii in make_whole which you can steal for this. Maybe also a seealso for pack_into_box (which maybe we can rename wrap?)

package/MDAnalysis/core/topologyattrs.py
```diff
@@ -1799,6 +1800,137 @@ def n_fragments(self):
...
+          unit cell, preventing bonds from being "broken" accross periodic
+          boundaries.
+        * Be aware of the fact that only atoms *belonging to the group* will
```
richardjgowers: probably deserves an example of how to select the fragments atoms in an atomgroup belong to

package/MDAnalysis/core/topologyattrs.py
```diff
@@ -1799,6 +1800,137 @@ def n_fragments(self):
...
+          be unwrapped! If you want entire molecules to be unwrapped, make sure
+          that all atoms of these molecules belong the group.
+
+        See Also
+        --------
+        :func:`MDAnalysis.lib.mdamath.make_whole`
```
richardjgowers: & pack_into_box & applyPBC

package/MDAnalysis/core/topologyattrs.py
```diff
@@ -1799,6 +1800,137 @@ def n_fragments(self):
...
+
+        atoms = self.atoms
+        unique_atoms = atoms.unique
+
+        comp = compound.lower()
+        if comp == 'group':
+            if ortho:
+                positions = unique_atoms.positions
+                # Only unwrap if necessary:
+                if np.any(np.abs(positions - positions[0]) > hbox):
```
richardjgowers: this optimisation is cute, but maybe deserves to live inside of `make_whole`

package/MDAnalysis/core/topologyattrs.py
```diff
@@ -1799,6 +1800,137 @@ def n_fragments(self):
...
+            compound_indices = unique_atoms.fragindices
+        elif comp == 'residues':
+            compound_indices = unique_atoms.resindices
+        elif comp == 'segments':
+            compound_indices = unique_atoms.segindices
+        elif comp == 'molecules':
+            try:
+                compound_indices = unique_atoms.molnums
+            except AttributeError:
+                raise NoDataError("Cannot use compound='molecules': "
```
richardjgowers: the existing attributeerror should suffice rather than adding another layer here

package/MDAnalysis/core/topologyattrs.py
```diff
@@ -1799,6 +1800,137 @@ def n_fragments(self):
...
+                                  "No molecule information in topology.")
+        else:
+            raise ValueError("Unrecognized compound definition: {}\nPlease use"
+                             " one of 'group', 'residues', 'segments', "
+                             "'molecules', or 'fragments'.".format(compound))
+
+        # Now process every compound:
+        unique_compound_indices = unique_int_1d(compound_indices)
```
richardjgowers: I don't like the structuring here, you've got `comp == group` being returned before this, then this says it's where the work is done.  Maybe instead:
```python
if not comp in ALLOWED:
   ValueError:
elif comp == 'group'
    positions = simple unpack
else:
    # find compound indices
    if comp == 'residue':
        # etc
    elif comp == 'fragment':
        # etc
    positions = loop over compounds

return positions
```

So there's a single exit point?

richardjgowers:
@zemanj `choose_image` is meant to translate an AtomGroup in boxlengths (ie not really moving it in terms of PBC) to minimise its distance from a given reference point.  Its to reduce code duplication as it appears elsewhere too.

package/MDAnalysis/core/topologyattrs.py
zemanj: `*Group.wrap()` already exists and extends the functionality of `*Group.pack_into_box()`. I'm in favor of deprecating `pack_into_box()`.

package/MDAnalysis/core/topologyattrs.py
zemanj: I put this here intentionally since the keyword is called `'molecules'` but the `AttributeError` would be raised for `molnums`, which might be confusing. Same as in `*Group.center()`, btw.

package/MDAnalysis/core/topologyattrs.py
zemanj: `*Group.wrap()` has the `box` kwarg as well, so I wanted to stay consistent there...

package/MDAnalysis/core/topologyattrs.py
zemanj: Is there a more elegant way than constructing a group containing all atoms of all fragments in `ag` than `ag_all_frag_atoms = sum(ag.fragments)`?

package/MDAnalysis/core/topologyattrs.py
zemanj: Very true! That way, the optimization will always be applied.

package/MDAnalysis/core/topologyattrs.py
zemanj: I just realized that `pack_into_box()` can do out-of-place packing, while `wrap()` always acts inplace. That should be changed. Furthermore, `wrap()` potentially acts on atoms outside of the calling group. That's dangerous and should be changed as well.

package/MDAnalysis/core/topologyattrs.py
zemanj: Ok what I said doesn't make sense at all since `box` is never really used except for the optimization check, which is potentially inconsistent with what `make_whole()` does and will go there anyway. I'll get rid of the `box` kwarg.

richardjgowers:
@zemanj this is doing the image shift: https://github.com/MDAnalysis/mdanalysis/blob/develop/package/MDAnalysis/core/groups.py#L1251

But more generally I think `AtomGroup.choose_image` is useful enough to have it solo.  Can be a future PR to limit the scope of this one though

package/MDAnalysis/core/topologyattrs.py
richardjgowers: that's fine, internally it should just be going through to `AtomGropu(np.concatenate([ag.ix for ag in ag.fragments]))` without too many layers

package/MDAnalysis/core/topologyattrs.py
richardjgowers: I just couldn't think of a time when you would want to specify a box but not set it as `u.dimensions`

package/MDAnalysis/core/topologyattrs.py
richardjgowers: also if it's done in the Cython level you can exit the `any()` loop early, I think this will do the whole calculation then evaluate the boolean values

zemanj: For an atomgroup consisting of 500 fragments with 25 atoms per fragment, i measured the following runtimes:
* without optimization: 154 ms,
* optimization in `make_whole()`: 65 ms
* optimization in Python: 66.5 ms

Turns out the speed-up is not that great anymore. This simple loop runs in 22 ms on the same system:
```Python
def group_make_whole(ag):
    box = ag.dimensions
    fragments = ag.fragments
    if np.any(box[3:] != 90.0):
        for f in fragments:
            mda.lib.mdamath.make_whole(f)
    else:
        hbox = 0.5 * box[:3]
        for f in fragments:
            pos = f.positions
            if np.any(np.abs(pos - pos[0]) > hbox):
                mda.lib.mdamath.make_whole(f)
```
Seems to be the rest of the machinery that slows down `group.unwrap()`.

zemanj:
@richardjgowers Requests are addressed, I'll write the tests tomorrow.

richardjgowers:
@zemanj ok cool, this is looking really good.  Once we've merged this it should be fairly easy to
 - make `center_of_mass()` just call `self.unwrap(compound='group')` to get coordinates to operate on.
 - make the transformation thing just call this method of Groups

zemanj:
Even though unwrapping requires `Bonds`, I moved `*Group.unwrap()` from `core.topologyattrs.Bonds` to `core.groups.GroupBase` for two reasons:
1. `unwrap()` should also be accessible from `ResidueGroup` and `SegmentGroup`
2. If the topology has no `Bonds`, the error (`NoDataError`) and corresponding message should be coherent and sensible for all kinds of `*Groups`.

One can only accomplish both by either adding a custom `__getattr__()` not only for `AtomGroup` but also for the other ones, or by moving the `unwrap()` method to `GroupBase` and have it check for the `Bonds` attribute of its atoms. I chose the latter possibility since `AtomGroup.__getattr__()` already has a comment stating it's deprecated (although I don't really know why).

zemanj:
I changed the dtype of `molnums` from `int` to `np.int64` to make appveyor happy.

zemanj:
There are at least two minor issues I still need to take care of, so this is *not* yet ready for review.

package/MDAnalysis/core/groups.py
```diff
@@ -1253,6 +1256,160 @@ def wrap(self, compound="atoms", center="com", box=None):
...
+        :func:`~MDAnalysis.lib.mdamath.make_whole`,
+        :meth:`~MDAnalysis.AtomGroup.wrap`,
+        :meth:`~MDAnalysis.AtomGroup.pack_into_box`,
+        :func:`~MDanalysis.lib.distances.apply_PBC`
+
+
+        .. versionadded:: 0.20.0
+        """
+        atoms = self.atoms
+        # bail out early if no bonds in topology:
```
richardjgowers: there's a `requires` decorator that can do this type of thing (and create uniform error messages)

package/MDAnalysis/lib/_cutil.pyx
```diff
@@ -211,6 +231,17 @@ def make_whole(atomgroup, reference_atom=None):
             ortho = False
 
     if ortho:
+        # If atomgroup is already unwrapped, bail out
+        is_unwrapped = True
+        for i in range(1, natoms):
+            for j in range(3):
+                if abs(oldpos[i, j] - oldpos[0, j]) >= half_box[j]:
```
richardjgowers: I'd doublecheck that this `abs` call is producing the right C code, iirc you can sometimes get Python's

package/MDAnalysis/core/groups.py
zemanj: @richardjgowers I know. The `@requires()` decorator only works on AtomGroups, that's why it's not used here.

package/MDAnalysis/core/groups.py
```diff
@@ -1253,6 +1256,160 @@ def wrap(self, compound="atoms", center="com", box=None):
             if not all(s == 0.0):
                 o.atoms.translate(s)
 
+    def unwrap(self, compound='fragments', reference='com', inplace=True):
+        """Move atoms of this group so that bonds within the
+        group's compounds aren't split accross periodic boundaries.
```
tylerjereddy:
```suggestion
        group's compounds aren't split across periodic boundaries.
```

package/MDAnalysis/lib/_cutil.pyx
tylerjereddy: I guess you could try `from libc.math cimport fabs `, but it may not make much difference.

zemanj: Should be `fabs` anyway

package/MDAnalysis/core/groups.py
zemanj: One could extend the `@requires` decorator to check for `GroupBase` instances instead of `AtomGroup` and then check the group's `.atoms` member for the given attribute(s). However, I don't like this solution since it
1. doesn't allow to check for `ResiduGroup` or `SegmentGroup` attributes and
2. more importantly, isn't lightweight anymore since it involves object instantiation when accessing the `.atoms` member.

The check I implemented doesn't instantiate additional objects and its impact on performance should therefore be negligible.

zemanj:
Ok all checks passed. I'll just add changelog entries and this is good to go.
@MDAnalysis/coredevs pls review

package/MDAnalysis/core/groups.py
```diff
@@ -1253,6 +1262,160 @@ def wrap(self, compound="atoms", center="com", box=None):
...
+                positions += target - refpos
+        # We need to split the group into compounds:
+        else:
+            if comp == 'fragments':
+                compound_indices = unique_atoms.fragindices
+            elif comp == 'residues':
+                compound_indices = unique_atoms.resindices
+            elif comp == 'segments':
+                compound_indices = unique_atoms.segindices
+            elif comp == 'molecules':
```
tylerjereddy: The codecov report suggests there's no case where control flow gets here and `comp != 'molecules'`. Indeed, there's already a `ValueError` guard above against unknown `comp` values, so you could just use `else` here safely & avoid the extraneous check / partial statement coverage.

zemanj: Ok let's crank the diff coverage up to 100%

tylerjereddy: The commit history suggests you've already attacked similar cases elsewhere, and it is just barely worth doing, so that's up to you I think.

zemanj: Done. Let's see what CodeCov thinks about it... :crossed_fingers:

zemanj:
@tylerjereddy 
> Does this mean MDAnalysis will soon be able to process my virus trajectories (rhombic dodecahedron with superstructure split all over the place requiring all kinds of trjconv processing & translations)?

If you mean that your system's periodic unit cell is a rhombic dodecahedron, then the answer is no.
Otherwise, that depends on what you want to achieve. This PR so far only addresses the problem of making *bonded* structures (i.e., single molecules) whole.
If you have *non-bonded* structures such as molecule clusters or complexes in the system and need to keep those intact across PBC, that will require cluster unwrapping. Cluster unwrapping is about to come in a follow-up PR very soon.

richardjgowers:
@zemanj I think you can represent a rhombic dodec as a triclinic cell (I think I read somewhere a triclinic cell can represent **any** 3d tiling shape...)

@tylerjereddy I think this will make your unsplitting dreams come true... but if you could check before we merge that would be great.  Am I right that you can use a triclinic representation?

package/MDAnalysis/core/groups.py
richardjgowers: Hmm, really we could have something like `assert 'bonds' in u.topology` which works... I'll raise an issue.  I just like error messages being uniform, so hitting eg 'no masses' always looks similar

package/MDAnalysis/lib/_cutil.pyx
```diff
@@ -211,6 +236,17 @@ def make_whole(atomgroup, reference_atom=None):
             ortho = False
 
     if ortho:
+        # If atomgroup is already unwrapped, bail out
+        is_unwrapped = True
+        for i in range(1, natoms):
+            for j in range(3):
+                if fabs(oldpos[i, j] - oldpos[0, j]) >= half_box[j]:
```
richardjgowers: `fabs`ulous

testsuite/MDAnalysisTests/core/test_unwrap.py
```diff
@@ -0,0 +1,800 @@
...
+from numpy.testing import assert_almost_equal, assert_array_equal
+import pytest
+
+import MDAnalysis as mda
+from MDAnalysis.core import topologyattrs
+from MDAnalysis import NoDataError
+from MDAnalysis.lib.mdamath import triclinic_box
+
+
+def unwrap_test_universe(have_bonds=True, have_masses=True, have_molnums=True,
```
richardjgowers: niiiiice

zemanj: Yeah writing tests took longer than implementing the feature itself... The cool thing about this is that the minimal test universe makes the test lightning fast and (hopefully) covers all relevant situations. I also designed it with cluster unwrapping in mind.

zemanj:
@richardjgowers @tylerjereddy 
> I think you can represent a rhombic dodec as a triclinic cell (I think I read somewhere a triclinic cell can represent **any** 3d tiling shape...)

Not sure about any space-filling 3d shape (sounds plausible, though), but definitely true for a rhombic dodecahedron (gromacs is doing it that way).

tylerjereddy:
I probably won't get a chance to test on my system before the merge, but sounds like these are solid steps in that direction.

I guess when I do get around to it, I can open an issue / PR if there's something special that needs tweaking or adding in.

zemanj:
@richardjgowers thanks! :+1:

PR #2192, opened 2019-01-31
sum method for atomgroups

PicoCentauri:
 -  Added `*Group.sum()` method
 - Added compound parameter to `total_mass` and `total_charge`

Following the discussion of PR #2118 this PR adds a sum method to all groups. Furthermore a compound parameter to `*Group.total_charge()` and `*Group.total_mass()` was added similar to the `center_of_mass()`  method. 

I'm not sure if the name `sum`  is confusing but this was the best thing, which came to my mind.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/core/groups.py
+++ b/package/MDAnalysis/core/groups.py
@@ -859,50 +859,50 @@ class GroupBase(_MutableBase):
         """Accumulates the attribute associated with (compounds of) the group.
 
         Accumulates the attribute of :class:`Atoms<Atom>` in the group.
-        The accumulation is done per :class:`Residue`, :class:`Segment`, molecule, 
+        The accumulation per :class:`Residue`, :class:`Segment`, molecule, 
         or fragment can be obtained by setting the `compound` parameter
-        accordingly. By default the method sums up all elements, but 
-        any numpy function that takes a multi-dimensional array of the same 
+        accordingly. By default, the method sums up all attributes per compound, but 
+        any function that takes a multi-dimensional array of the same 
         dtype and returns a scalar can be used. For multi-dimensional input 
-        arrays the accumulation is performed along the first axes.
+        arrays, the accumulation is performed along the first axes.
 
         Parameters
         ----------
-        attribute : str, numpy.ndarray
-            Attribute or :class:`numpy.ndarray` to accumulate.
-            If a :class:`numpy.ndarray` is provided which first dimensions 
+        attribute : str or array_like
+            Attribute or array of values to accumulate.
+            If a :class:`numpy.ndarray` is provided, its first dimension
             must have the same length as the total number of atoms in 
-            the :class:`AtomGroup`.
-        function : function, optional
+            the group.
+        function : callable, optional
             Numpy function to perform. It must take an array of the same dtype 
             and return a scalar.
         compound : {'group', 'segments', 'residues', 'molecules', 'fragments'}, optional
-            If ``'group'``, the sum over all elements associated
+            If ``'group'``, the sum over all attributes associated
             with atoms in the group will
-            be returned as a single value. Else, the sum for all elements
-            of each :class:`Segment`, :class:`Residue`, molecule, or fragment
-            will be returned as an 1d array.
-            Note that, in any case, *only* the positions of :class:`Atoms<Atom>`
+            be returned as a single value. Otherwise, the accumulation of the attributes
+            per :class:`Segment`, :class:`Residue`, molecule, or fragment
+            will be returned as a 1d array.
+            Note that, in any case, *only* the :class:`Atoms<Atom>`
             *belonging to the group* will be taken into account.
 
         Returns
         -------
         float or numpy.ndarray
-            Acuumulation of the attribute or provided numpy.ndarray.
-            If `compound` was set to ``'group'``, the first dimension 
-            of the input is contracted to a single value.
-            If `compound` was set to ``'segments'``, ``'residues'``, 
+            Acuumulation of the `attribute`.
+            If `compound` is set to ``'group'``, the first dimension 
+            of the `attribute` array will be contracted to a single value.
+            If `compound` is set to ``'segments'``, ``'residues'``, 
             ``'molecules'``, or ``'fragments'``, the length of 
-            the first dimension is the number of compounds. In all cases the
+            the length of the first dimension will correspond to the number of compounds. In all cases, the
             other dimensions of the return array are of the original shape.
 
         Raises
         ------
         ValueError
             If provided array does not have the same length 
-            as the number of atoms in the :class:`AtomGroup`:.
+            as the number of atoms in the group.
         ~MDAnalysis.exceptions.NoDataError
-            If :class:`AtomGroup`: does not have the given attribute.
+            If the given attribute is missing.
         ValueError
             If `compound` is not one of ``'group'``, ``'segments'``,
             ``'residues'``, ``'molecules'``, or ``'fragments'``.
@@ -924,7 +924,7 @@ class GroupBase(_MutableBase):
             >>> sel = u.select_atoms('name CA')
             >>> sel.accumulate('masses', compound='residues')
     
-        To find the maximal charge per fragment of a given :class:`AtomGroup`::
+        To find the maximum atomic charge per fragment of a given :class:`AtomGroup`::
         
             >>> sel.accumulate('charges', compound="fragments", function=np.max)
 
@@ -934,7 +934,7 @@ class GroupBase(_MutableBase):
 
         atoms = self.atoms
 
-        if type(attribute) in (np.ndarray, tuple, list):
+        if isinstance(attribute, (np.ndarray, tuple, list)):
             attribute_values = np.asarray(attribute)
             if len(attribute_values) != len(atoms):
                 raise ValueError("The input array length {} does not match "
--- a/package/MDAnalysis/core/groups.py
+++ b/package/MDAnalysis/core/groups.py
@@ -861,10 +861,10 @@ class GroupBase(_MutableBase):
         Accumulates the attribute of :class:`Atoms<Atom>` in the group.
         The accumulation per :class:`Residue`, :class:`Segment`, molecule, 
         or fragment can be obtained by setting the `compound` parameter
-        accordingly. By default, the method sums up all attributes per compound, but 
-        any function that takes a multi-dimensional array of the same 
-        dtype and returns a scalar can be used. For multi-dimensional input 
-        arrays, the accumulation is performed along the first axes.
+        accordingly. By default, the method sums up all attributes per compound, 
+        but any function that takes a array and returns a acuumulation over a 
+        over a given axis can be used. For multi-dimensional input arrays, 
+        the accumulation is performed along the first axes.
 
         Parameters
         ----------
@@ -874,8 +874,10 @@ class GroupBase(_MutableBase):
             must have the same length as the total number of atoms in 
             the group.
         function : callable, optional
-            Numpy function to perform. It must take an array of the same dtype 
-            and return a scalar.
+            The function performing the accumulation. It must take the array of
+            attribute values to accumulate as its only positional argument and
+            accept an (optional) keyword argument ``axis`` allowing to specify
+            the axis along which the accumulation is performed.
         compound : {'group', 'segments', 'residues', 'molecules', 'fragments'}, optional
             If ``'group'``, the sum over all attributes associated
             with atoms in the group will
@@ -901,8 +903,6 @@ class GroupBase(_MutableBase):
         ValueError
             If provided array does not have the same length 
             as the number of atoms in the group.
-        ~MDAnalysis.exceptions.NoDataError
-            If the given attribute is missing.
         ValueError
             If `compound` is not one of ``'group'``, ``'segments'``,
             ``'residues'``, ``'molecules'``, or ``'fragments'``.
@@ -937,16 +937,11 @@ class GroupBase(_MutableBase):
         if isinstance(attribute, (np.ndarray, tuple, list)):
             attribute_values = np.asarray(attribute)
             if len(attribute_values) != len(atoms):
-                raise ValueError("The input array length {} does not match "
-                                 "the required length {} based on "
-                                 "the group size.".format(len(attribute_values),
-                                                          len(atoms)))
+                raise ValueError("The input array length ({}) does not match "
+                                 "the number of atoms ({}) in the group."
+                                 "".format(len(attribute_values), len(atoms)))
         else:
-            try:
-                attribute_values = getattr(atoms, attribute)
-            except AttributeError:
-                raise NoDataError("AtomGroup does not have '{}' "
-                                  "attribute.".format(attribute))
+            attribute_values = getattr(atoms, attribute)
 
         comp = compound.lower()
 
--- a/package/MDAnalysis/core/groups.py
+++ b/package/MDAnalysis/core/groups.py
@@ -862,17 +862,18 @@ class GroupBase(_MutableBase):
         The accumulation is done per :class:`Residue`, :class:`Segment`, molecule, 
         or fragment can be obtained by setting the `compound` parameter
         accordingly. By default the method sums up all elements, but 
-        any numpy function that takes a (n,) array of the same dtype and 
+        any numpy function that takes any multi-dimensional array of the same dtype and 
         returns a scalar can be used.
 
         Parameters
         ----------
         attribute : str, numpy.ndarray
             Attribute or :class:`numpy.ndarray` to accumulate.
-            If a :class:`numpy.ndarray` is provided it must have the same length 
-            as the total number of atoms in the :class:`AtomGroup`.
+            If a :class:`numpy.ndarray` is provided which first dimensions 
+            must have the same length as the total number of atoms in 
+            the :class:`AtomGroup`.
         function : function, optional
-            Numpy function to perform. It must take a (n,) array of the same dtype 
+            Numpy function to perform. It must take an array of the same dtype 
             and return a scalar.
         compound : {'group', 'segments', 'residues', 'molecules', 'fragments'}, optional
             If ``'group'``, the sum over all elements associated
@@ -933,21 +934,22 @@ class GroupBase(_MutableBase):
 
         if type(attribute) in (np.ndarray, tuple, list):
             attribute_values = np.array(attribute)
-            if attribute_values.shape != (len(atoms),):
-                raise ValueError("The input array dimensions {} does not match "
-                                 "the required dimension ({},) based on "
-                                 "the group size.".format(attribute_values.shape,
+            if len(attribute_values) != len(atoms):
+                raise ValueError("The input array length {} does not match "
+                                 "the required length {} based on "
+                                 "the group size.".format(len(attribute_values),
                                                           len(atoms)))
         else:
             try:
                 attribute_values = getattr(atoms, attribute)
             except AttributeError:
-                raise NoDataError("AtomGroup does not have '{}' attribute.".format(attribute))
+                raise NoDataError("AtomGroup does not have '{}' "
+                                  "attribute.".format(attribute))
 
         comp = compound.lower()
 
         if comp == 'group':
-            return function(attribute_values)
+            return function(attribute_values, axis=0)
         elif comp == 'residues':
             compound_indices = atoms.resindices
         elif comp == 'segments':
@@ -969,6 +971,8 @@ class GroupBase(_MutableBase):
                              " one of 'group', 'residues', 'segments', "
                              "'molecules', or 'fragments'.".format(compound))
 
+        higher_dimemsions = list(attribute_values.shape[1:])
+
         # Sort attribute values by compound
         sort_indices = np.argsort(compound_indices)
         compound_indices = compound_indices[sort_indices]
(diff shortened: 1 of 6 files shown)


package/MDAnalysis/core/groups.py
```diff
@@ -850,6 +850,115 @@ def center_of_geometry(self, pbc=None, compound='group'):
...
+
+        Examples
+        --------
+
+        To find the total charge of a given :class:`AtomGroup`::
+
+            >>> sel = u.select_atoms('prop mass > 4.0')
+            >>> sel.sum(sel.charges)
+
+        To find the total mass per residue of all CA :class:`Atoms<Atom>`::
```
richardjgowers: I'm confused what this will return.  Is it the mass of CA atoms aggregated per residue?  So if each residue only had one CA it would be identical across all?

package/MDAnalysis/core/groups.py
```diff
@@ -850,6 +850,115 @@ def center_of_geometry(self, pbc=None, compound='group'):
...
+            contain molecule information (molnums) or if `compound` is
+            ``'fragments'`` but the topology doesn't contain bonds.
+
+        Examples
+        --------
+
+        To find the total charge of a given :class:`AtomGroup`::
+
+            >>> sel = u.select_atoms('prop mass > 4.0')
+            >>> sel.sum(sel.charges)
```
richardjgowers: This seems clunky.  How tricky would it be making `sel.sum('charges')` work?  But really `sel.charges.sum()` seems most preferable

richardjgowers:
I think I get what this is doing, but maybe it would be better to achieve it using some sort of `groupby` feature instead.

So `sel.groupby('residues').charges`, `sel.groupby('fragment').total_mass()`

So `sel.groupby('residues')` goes to a ResidueGroup `sel.groupby('fragment')` goes to an iterable of AtomGroups

@MDAnalysis/coredevs thoughts?

zemanj:
I generally like the idea of having the opportunity of summing up group properties per compound.
However, I'd prefer it if the `*Group.sum()` method accepted the name of a property (as a string) instead of an array containing the values.
I don't know if one really needs the flexibility of passing arbitrary values. You could have both if you check for the type of the `elements` argument (which I would rename to `property`) and use `getattr()` in case of a string and otherwise, try to use the values as is.

What you're missing right now is a check for the length of the elements array, which has to  match the number of atoms.

And there's a typo in the return statements of `total_mass()` and `total_charge()` (it must read `elements`, not `weights`).

zemanj:
@richardjgowers I'm afraid `*Group.groupby()` returns a dict containing groups as values, which is dead slow because of all the object instantiations.

jbarnoud:
I do not like the name. I do not like that `sum(group)` does not do the same thing as `group.sum()`. I would not mind a `*Group.sum_per_compound` method, though (or any similar name).

For now, I solve this kind of problems like:

```python
total_charges = np.array([compound.sum() for compound in sel.split('molecules')])
```

Though, what I really need most of the time is:

```python
all_centers = np.array([compound.center_of_mass() for compound in sel.split('molecules')])
```

package/MDAnalysis/core/groups.py
PicoCentauri: > Is it the mass of CA atoms aggregated per residue?

Yes, accectly.

> So if each residue only had one CA it would be identical across all?

Yes. In the case for only one residue this is a bit senseless. But could be useful if you have for example a geometric selection...

package/MDAnalysis/core/groups.py
PicoCentauri: `sel.charges.sum()` is nice, but could this brake some scripts where people use the numpy attribites of the `sum` method?

PicoCentauri:
@zemanj 
> I'd prefer it if the *Group.sum() method accepted the name of a property (as a string) instead of an array containing the values.

This should be no problem to change.

@jbarnoud 
Both are nice pieces of code, and do exactly the same as `toal_charge('molecules')`  and `center_of_mass('molecules')` but probably not all users are aware of this. I think it would be nice to have a similar basis for calculating grouped centers (`center_of_mass` etc) and grouped sums of properties (`total_mass` etc). One could kick out the sum method in every group and just add some logic to the` total_charge` and `total_mass` methods?

zemanj:
@jbarnoud I just checked the runtimes of your approach against `center_of_mass(compound='molecules')`:
```Python
In [2]: import numpy as np
In [3]: u = mda.Universe("topol.tpr", "traj_comp.xtc");
In [4]: ag = u.select_atoms("moltype BMIM")
In [5]: ag
Out[5]: <AtomGroup with 12500 atoms>
In [6]: len(np.unique(ag.molnums))
Out[6]: 500
In [7]: %timeit np.array([compound.center_of_mass(pbc=True) for compound in ag.split('molecule')])
10 loops, best of 3: 33.1 ms per loop
In [8]: %timeit ag.center_of_mass(pbc=True, compound='molecules')
1000 loops, best of 3: 1.23 ms per loop
```
Your approach takes about **27** times longer than `ag.center_of_mass(compound='molecules')`.
The difference in runtime is 31.87 ms, which amounts to **53 min** when analyzing 100,000 frames.
That certainly justifies a faster implementation!

zemanj:
@jbarnoud
> I do not like the name. I do not like that `sum(group)` does not do the same thing as `group.sum()`. I would not mind a `*Group.sum_per_compound` method, though (or any similar name).

I agree that the name `sum()` is sub-optimal. But why not follow a more flexible approach and do something like `*Group.accumulate_properties(property, compound, accumulator=np.sum)`?
That way, one could pass arbitrary functions as accumulators.

orbeckst:
Would this be a `Group.reduce(..., func=np.sum)`? Using `min`, `max`, `median`, `std`, `avg` will make it easy to do all kinds of interesting things.

For the most common ones I would still be tempted to create shorthands `Group.sum()`, `Group.avg()` etc – makes TAB completion and interactive exploration much more intuitive.

orbeckst:
(Sorry, I ignored the discussion regarding the appropriateness of the name `sum` – I agree with https://github.com/MDAnalysis/mdanalysis/pull/2192#issuecomment-459407520 that `sum(group)` should do the same as `group.sum()`.)

jbarnoud:
@zemanj Thank you for the benckmark. I did not notice the compound argument in `center_of_mass`. Since the argument already exists, it makes perfect sense to have it everywhere it is relevant; `total_mass` and `total_charges` are perfect candidates here.

I like the accumulator/reduce idea. Because the proposed function can already work with any array, it opens very interesting possibilities. I believe the function would work with any array assuming that the first dimension matches the number of atoms; am I right here?

PicoCentauri:
To sum up the discussion:
The idea is to have a function like `*group.accumulate(attribute, op=np.sum, compound='group')`, that accumulates the given attribute (mass, charge) of the group according to the given operation?

PicoCentauri:
I just adjusted the method and also did a first performance test comparing it to list comprehension.

```python
In [9]: import MDAnalysis as mda 
   ...: from MDAnalysisTests.datafiles import PSF, DCD 
   ...: u = mda.Universe(PSF,DCD) 
   ...: %timeit u.atoms.accumulate("masses",compound="residues", op=np.sum) 
   ...: %timeit np.array([np.sum(compound.masses) for compound in u.atoms.split('residue')])                                                                                                             

1.41 ms ± 33.6 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
4.04 ms ± 30.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

package/MDAnalysis/core/groups.py
```diff
@@ -850,6 +850,128 @@ def center_of_geometry(self, pbc=None, compound='group'):
...
+        
+        Accumulates the attribute of :class:`Atoms<Atom>` in the group.
+        The accumulation is done per :class:`Residue`, :class:`Segment`, molecule, 
+        or fragment can be obtained by setting the `compound` parameter
+        accordingly. By default the method sums up all elements, but 
+        any numpy operation applying to numpy.ndarray can be used.
+
+        Parameters
+        ----------
+        attribute : str
```
jbarnoud: I liked that you could give an arbitrary array (assuming it has the right size). In the ideal case, I would like to be able to provide either the name, or an arbitrary array.

package/MDAnalysis/core/groups.py
```diff
@@ -850,6 +850,128 @@ def center_of_geometry(self, pbc=None, compound='group'):
...
+            with atoms in the group will
+            be returned as a single value. Else, the sum for all elements
+            of each :class:`Segment`, :class:`Residue`, molecule, or fragment
+            will be returned as an 1d array.
+            Note that, in any case, *only* the positions of :class:`Atoms<Atom>`
+            *belonging to the group* will be taken into account.
+
+        Returns
+        -------
+        sum : numpy.ndarray
```
jbarnoud: You do not have to provide a name for the return value. Here I would avoid doing it as the name is actually wrong:

```
Returns
-------
numpy.ndarray
    The description
```

package/MDAnalysis/core/groups.py
```diff
@@ -850,6 +850,128 @@ def center_of_geometry(self, pbc=None, compound='group'):
...
+        n_compounds = len(unique_compound_indices)
+        unique_compound_sizes = unique_int_1d(compound_sizes)
+        # Allocate output array:
+        accumulation = np.zeros(n_compounds)
+        # Compute sums per compound for each compound size:
+        for compound_size in unique_compound_sizes:
+            compound_mask = compound_sizes == compound_size
+            _compound_indices = unique_compound_indices[compound_mask]
+            atoms_mask = np.in1d(compound_indices, _compound_indices)
+            _summands = attribute_values[atoms_mask].reshape((-1, compound_size))
```
jbarnoud: What does "summands" mean as a variable name?

package/MDAnalysis/core/groups.py
```diff
@@ -850,6 +850,128 @@ def center_of_geometry(self, pbc=None, compound='group'):
...
+        The accumulation is done per :class:`Residue`, :class:`Segment`, molecule, 
+        or fragment can be obtained by setting the `compound` parameter
+        accordingly. By default the method sums up all elements, but 
+        any numpy operation applying to numpy.ndarray can be used.
+
+        Parameters
+        ----------
+        attribute : str
+            Attribute to accumulate.
+        op : operation, optional
```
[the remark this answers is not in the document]
PicoCentauri: Okay I will improve the docs. Another constraint to the function currently is that it also should be fine with the `axis=1` option.

package/MDAnalysis/core/groups.py
PicoCentauri: Yes. This should be feasible.

package/MDAnalysis/core/groups.py
PicoCentauri: These are the elements to which the accumulation is applied. I will rename them to elements if this is more clear.

PicoCentauri:
I added some tests and also the possibility to pass multi-dimensional arrays.

zemanj:
@PicoCentauri According to CodeCov, additional tests for `compound='molecules'` and for universes without a) `molnums` and b) `bonds` are needed. Apart from that, this looks really good :+1:

PicoCentauri:
@zemanj Is there anything open here? Otherwise would be nice if we could merge this so that I can continue the work on PR #2118.

zemanj:
@PicoCentauri Thanks for the reminder, I'll check it out later today.

package/MDAnalysis/core/groups.py
```diff
@@ -850,6 +850,150 @@ def center_of_geometry(self, pbc=None, compound='group'):
...
+            If a :class:`numpy.ndarray` is provided which first dimensions 
+            must have the same length as the total number of atoms in 
+            the :class:`AtomGroup`.
+        function : function, optional
+            Numpy function to perform. It must take an array of the same dtype 
+            and return a scalar.
+        compound : {'group', 'segments', 'residues', 'molecules', 'fragments'}, optional
+            If ``'group'``, the sum over all elements associated
+            with atoms in the group will
+            be returned as a single value. Else, the sum for all elements
```
zemanj:
```suggestion
            be returned as a single value. Otherwise, the accumulation of the attributes
```

package/MDAnalysis/core/groups.py
```diff
@@ -850,6 +850,150 @@ def center_of_geometry(self, pbc=None, compound='group'):
...
+            of each :class:`Segment`, :class:`Residue`, molecule, or fragment
```
zemanj:
```suggestion
            per :class:`Segment`, :class:`Residue`, molecule, or fragment
```

package/MDAnalysis/core/groups.py
```diff
@@ -850,6 +850,150 @@ def center_of_geometry(self, pbc=None, compound='group'):
...
+        arrays the accumulation is performed along the first axes.
+
+        Parameters
+        ----------
+        attribute : str, numpy.ndarray
+            Attribute or :class:`numpy.ndarray` to accumulate.
+            If a :class:`numpy.ndarray` is provided which first dimensions 
+            must have the same length as the total number of atoms in 
+            the :class:`AtomGroup`.
+        function : function, optional
```
zemanj:
```suggestion
        function : callable, optional
```

package/MDAnalysis/core/groups.py
```diff
@@ -850,6 +850,150 @@ def center_of_geometry(self, pbc=None, compound='group'):
...
+            Numpy function to perform. It must take an array of the same dtype 
+            and return a scalar.
```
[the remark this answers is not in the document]
PicoCentauri: Sorry for the bad docu here. You are right. Its not restricted to numpy functions and the dtype is also not important. The only thing it necessary is an axis argument as in the numpy functions. Currently I see no other smart way to implement this...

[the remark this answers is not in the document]
zemanj: I don't see any better way either, so I think fixing the docs is sufficient. Something along these lines should do the trick:
```Python
"""
...
function : callable, optional
    The function performing the accumulation. It must take the array of
    attribute values to accumulate as its only positional argument and
    accept an (optional) keyword argument ``axis`` allowing to specify
    the axis along which the accumulation is performed.
...
"""
```
But maybe you have a more concise phrasing in mind.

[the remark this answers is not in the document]
PicoCentauri: Sounds nice!

package/MDAnalysis/core/groups.py
```diff
@@ -850,6 +850,150 @@ def center_of_geometry(self, pbc=None, compound='group'):
...
+                compound_indices = atoms.fragindices
+            except NoDataError:
+                raise NoDataError("Cannot use compound='fragments': "
+                                  "No bond information in topology.")
+        else:
+            raise ValueError("Unrecognized compound definition: {}\nPlease use"
+                             " one of 'group', 'residues', 'segments', "
+                             "'molecules', or 'fragments'.".format(compound))
+
+        higher_dimemsions = list(attribute_values.shape[1:])
```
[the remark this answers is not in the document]
PicoCentauri: Cheers :blush:

package/MDAnalysis/core/groups.py
```diff
@@ -850,6 +850,150 @@ def center_of_geometry(self, pbc=None, compound='group'):
...
+                raise ValueError("The input array length {} does not match "
+                                 "the required length {} based on "
+                                 "the group size.".format(len(attribute_values),
+                                                          len(atoms)))
+        else:
+            try:
+                attribute_values = getattr(atoms, attribute)
+            except AttributeError:
+                raise NoDataError("AtomGroup does not have '{}' "
+                                  "attribute.".format(attribute))
```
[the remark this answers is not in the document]
PicoCentauri: Yes why not. Following this I can also remove this from the `Raises` section of the docstring and the specific test, right?

testsuite/MDAnalysisTests/core/test_atomgroup.py
```diff
@@ -488,6 +488,64 @@ def test_center_wrong_shape(self, ag):
         with pytest.raises(ValueError):
             ag.center(weights)
 
+class TestAccumulate(object):
```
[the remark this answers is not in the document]
PicoCentauri: Thanks.

You are right. I will ad them. Should be doable using the awesome multiple fixtures of pytest. I would rather move all the tests to `core/test_groups.py`. In my opinion this is more consistent. Why should we create a new test file only for accumulation and not for the other group attributes?

The `shape` is implicitly tested in the `test_accumulate_array_attribute_compounds` function. This checks the full array and would raise a shapes missmatch. But if you want I can also add a specific shape test?

zemanj:
I'd suggest to remove it from the docstring but change the tests to check for an AttributeError. Technically, the test isn't required, but it might help us to keep track of this behavior if we change something in AtomGroup.__getattr__() or related.

zemanj:
@PicoCentauri
> The `shape` is implicitly tested in the `test_accumulate_array_attribute_compounds` function. This checks the full array and would raise a shapes missmatch. But if you want I can also add a specific shape test?

Looks like I missed that one. Would you mind changing the tested array's shape to `(len(ag), 2, 5)`? That will guarantee that the "remaining" shape won't simply be flattened.

testsuite/MDAnalysisTests/core/test_accumulate.py
```diff
@@ -0,0 +1,165 @@
...
+from numpy.testing import assert_equal, assert_almost_equal
+
+import MDAnalysis as mda
+from MDAnalysis.exceptions import DuplicateWarning, NoDataError
+from MDAnalysisTests.datafiles import PSF, DCD, TPR_xvf, TRR_xvf, GRO
+import pytest
+
+universe = mda.Universe(PSF, DCD)
+universe_molfrag = mda.Universe(TPR_xvf, TRR_xvf)
+levellist = ('atoms', 'residues', 'segments')
```
PicoCentauri: I added the tests for segments; However, I am not happy with putting the decorator `@pytest.mark.parametrize('level', levellist)` in front of **every** function in the class but unfortunately I haven't found a better solution. Maybe somebody who knows pytest better finds a smarter way...

richardjgowers: You can place the decorator at the class level, but then **every** test in the class needs to use it

```python
@pytest.mark.parametrize('this', [1, 2])
class TestThing(object):
    def test_add(self, this):
        something
```

Or you can make a fixture have a loop in it using params

```python
@pytest.fixture(params=[1, 2])
def thing(request):
    return request.param[0]

def test_this(thing):
    assert thing == 1 or thing == 2
```

testsuite/MDAnalysisTests/core/test_accumulate.py
```diff
@@ -0,0 +1,165 @@
...
+
+import numpy as np
+from numpy.testing import assert_equal, assert_almost_equal
+
+import MDAnalysis as mda
+from MDAnalysis.exceptions import DuplicateWarning, NoDataError
+from MDAnalysisTests.datafiles import PSF, DCD, TPR_xvf, TRR_xvf, GRO
+import pytest
+
+universe = mda.Universe(PSF, DCD)
```
richardjgowers: you can't have these objects live here.  This might mean you have to expand a few test functions below too unfortunately

testsuite/MDAnalysisTests/core/test_accumulate.py
```diff
@@ -0,0 +1,165 @@
...
+universe_molfrag = mda.Universe(TPR_xvf, TRR_xvf)
+levellist = ('atoms', 'residues', 'segments')
+
+
+class TestAccumulate(object):
+    """Tests the functionality of *Group.accumulate()."""
+
+    @pytest.mark.parametrize('level', levellist)
+    def test_accumulate_str_attribute(self, level):
```
richardjgowers: so here you could do something like:

```python
@pytest.fixture(params['atoms', 'residues' 'segments'])
def group(request):
    u = mda.Universe(..., ...)
    return getattr(u, request.params[0])

def test_accumulate_str_attribute(group):
    assert_almost_equal(group.accumulate(), np.sum())
```
and the fixture params make the test expand to cover all 3 types of Group

testsuite/MDAnalysisTests/core/test_accumulate.py
PicoCentauri: Nice. Thanks your first suggestion will do the job here.

testsuite/MDAnalysisTests/core/test_accumulate.py
PicoCentauri: Okay... :worried:

testsuite/MDAnalysisTests/core/test_accumulate.py
```diff
@@ -0,0 +1,144 @@
...
+    def test_accumulate_str_attribute(self, group):
+        assert_almost_equal(group.accumulate("masses"), np.sum(group.atoms.masses))
+
+    def test_accumulate_different_func(self, group):
+        assert_almost_equal(
+            group.accumulate("masses", function=np.prod),
+            np.prod(group.atoms.masses))
+
+    @pytest.mark.parametrize('universe, name, compound',
+                             ((mda.Universe(PSF, DCD), 'resids', 'residues'),
```
richardjgowers: this has the same problem, the scope of the Universe objects here will be the entire test run.  The advantage of fixtures is the object only lives as long as the test function

zemanj: I just replaced the DCD / TPR universes with the infamous `UnWrapUniverse` for the compound tests. Creating `UnWrapUniverse`s is so fast that fixtures won't improve anything here.
Hope you're ok with this.

zemanj:
@PicoCentauri I found a couple of typos in the docstrings which I missed in my review. I ddn't want to bug you with this again, so I fixed it myself. I also changed the `attribute` instance check to test for string instances first, which should be the normal use case I think.
Furthermore, I used the `UnWrapUniverse` from `MDAnalysisTests.core.util` for the compound tests. That saves a couple of seconds and doesn't require fixtures.

PicoCentauri:
@zemanj Thanks for fixing the typos. I'm happy with the tests and the method.

richardjgowers:
Awesome, more commits on the board for @PicoCentauri !

PicoCentauri:
Awesome. Thanks for the help and for merging!

PR #2197, opened 2019-02-01
Fixes Issue #2193

### LINKED ISSUE #2193: MMTF parser expects optional fields
**Expected behavior**

When importing MMTF files, MDAnalysis only requires fields which are labeled as "Required" in the [MMTF spec](https://github.com/rcsb/mmtf/blob/v1.0/spec.md#fields).

**Actual behavior**

MDAnalysis fails with omitted optional fields. E.g. a file with missing `altLocList`:

```python
>>> u = MDAnalysis.Universe('no-alt-loc.mmtf')
  File ".../MDAnalysis/core/universe.py", line 312, in __init__
    self._generate_from_topology()
  File ".../MDAnalysis/core/universe.py", line 367, in _generate_from_topology
    self._process_attr(attr)
  File ".../MDAnalysis/core/universe.py", line 896, in _process_attr
    m=len(attr)))
ValueError: Length of altLocs does not match number of atoms.
Expect: 660 Have: 0
```

``` python
import MDAnalysis
import msgpack

with open('1ubq.mmtf', 'rb') as f:
    d = msgpack.unpack(f, encoding='utf-8')

del d['altLocList']

with open('no-alt-loc.mmtf', 'wb') as f:
    msgpack.pack(d, f, use_bin_type=True)

u = MDAnalysis.Universe('no-alt-loc.mmtf')
```

```bash
# python -c "import MDAnalysis as mda; print(mda.__version__)"
0.19.2
# python -V
Python 3.7.2
# uname -a
Linux 4.20.4-arch1-1-ARCH #1 SMP PREEMPT Wed Jan 23 00:12:22 UTC 2019 x86_64 GNU/Linux
```

richardjgowers:
properly respect optional mmtf fields

Fixes #2193

 - 

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


richardjgowers:
@speleo3 I think this should fix your issue.  Until it gets released, you should be able to copy/paste the updated class and import it locally and MDAnalysis will then use the new/patched definition. (defining the class updates the registry in MDAnalysis)


@MDAnalysis/coredevs review please

orbeckst:
@richardjgowers do you know why codecov reports a drop in total coverage of –2.57% even though 100% of the diff are tested??

richardjgowers:
I think it’s just codecov screwing up, I’m only touching mmtf here. The
“new” coverage is 89.something which is our current coverage too

package/MDAnalysis/coordinates/MMTF.py
```diff
@@ -77,7 +79,9 @@ def _read_first_frame(self):
         ts._pos[:, 0] = top.x_coord_list
         ts._pos[:, 1] = top.y_coord_list
         ts._pos[:, 2] = top.z_coord_list
-        ts.dimensions = top.unit_cell
+        if not top.unit_cell is None:
```
orbeckst: Is  `ts.dimensions` that is not defined my MMTF set to something sensible? Do we not require Timestep to have unitcell defined? Will bad things happen if a ts does not have it?

package/MDAnalysis/topology/MMTFParser.py
```diff
@@ -173,38 +175,47 @@ def iter_atoms(field):
...
-        altlocs = AltLocs(np.array([val.replace('\x00', '').strip()
-                                    for val in mtop.alt_loc_list], dtype=object))
-        atomids = Atomids(np.asarray(mtop.atom_id_list, dtype=np.int64))
-        bfactors = Bfactors(np.asarray(mtop.b_factor_list, dtype=np.float64))
-        charges = Charges(np.array(list(iter_atoms('formalChargeList')), dtype=np.float64))
-        names = Atomnames(np.array(list(iter_atoms('atomNameList')), dtype=object))
-        occupancies = Occupancies(np.asarray(mtop.occupancy_list, dtype=np.float64))
-        types = Atomtypes(np.array(list(iter_atoms('elementList')), dtype=object))
-
+        charges = Charges(list(iter_atoms('formalChargeList')))
```
orbeckst: This looks much cleaner than before :-).

testsuite/MDAnalysisTests/topology/test_mmtf.py
```diff
@@ -33,6 +33,20 @@ class TestMMTFParser_gz(TestMMTFParser):
     expected_n_segments = 4
 
 
+class TestMMTFSkinny(ParserBase):
```
orbeckst: Does it test `ts.dimensions`?

Does it test what happens if we try to access an optional topology attribute?

speleo3:
Fantastic, thanks! If you add `if mtop.bond_atom_list is not None` in [MMTFParser.py#L250](https://github.com/MDAnalysis/mdanalysis/blob/e5be7043cfa166610bbaefd410a4dbae7f5ed602/package/MDAnalysis/topology/MMTFParser.py#L250), then it will also read [3NJW-onlyrequired.mmtf](https://github.com/rcsb/mmtf/blob/a667fddad94dbb596ef38d67deef0c1d97cb943a/test-suite/mmtf/3NJW-onlyrequired.mmtf) from rcsb/mmtf#39

package/MDAnalysis/coordinates/MMTF.py
richardjgowers: So it gets set to `None` now.  So calling things like `distance_array(a, b, box=u.dimensions)` will proceed silently without PBC, which is maybe wrong as the user intended there to be a box?  Could instead set to [0, 0, 0] which would make distance calculations scream (hopefully)

testsuite/MDAnalysisTests/topology/test_mmtf.py
richardjgowers: dimensions tests now in coordinates/ & added tests for missing MMTF attributes

package/MDAnalysis/coordinates/MMTF.py
zemanj: @richardjgowers You're right, `None` will disable PBC. In `lib.distances`, all-zeros dimensions will run into `lib.util.check_box()` and will currently be interpreted as a triclinic cell since not all angles are 90 degrees. That will trigger calls to the respective triclinic C implementation, which will commonly result in a no-op for most of the routines. So, either way, things *will* go wrong.

A quick-and-dirty solution to the problem might be making `lib.util.check_box()` throw an error for all-zero boxes.

What we *really* need is a standardized and consistently enforced way to discriminate between no PBC and missing information!

zemanj: BTW that also affects PR #2201

richardjgowers: Tbh the really nice solution is to have 0 dimensions to be non periodic, so we could have semi periodic PBC...

Maybe `u.dimensions` should raise `AttributeError` if there's no box, it's essentially a missing attribute

zemanj: >Maybe `u.dimensions` should raise `AttributeError` if there's no box, it's essentially a missing attribute

I like this idea. That's completely in line with #2199 btw.
>Tbh the really nice solution is to have 0 dimensions to be non periodic, so we could have semi periodic PBC...

Semi-periodic PBC sounds awesome, but has to be taken with a grain of salt. We cannot do this efficiently with the current state of the PBC-related C-level implementations since that would require introducing branching in hot loops. I'm afraid that hoping for branch prediction to keep us up to speed won't work out here because branching will most likely prevent compiler optimizations such as loop auto-vectorization an the like.

zemanj: Forget the last part of what I just wrote. Semi-periodic PBC should already work.

zemanj: Semi-periodic boxes cannot be indicated by putting a zero value in the non-periodic direction. In general, that only works for true 2d systems. For PBC wrapping this might also work in 3d (not sure about triclinic boxes, though), but it might break pretty much anything else in semi-periodic 3d syatems, such as box volume calculation and so on.

orbeckst:
The Python 2.7 has an error
```
==================================== ERRORS ====================================
____________ ERROR collecting MDAnalysisTests/topology/test_mmtf.py ____________
../../../miniconda/envs/test/lib/python2.7/site-packages/_pytest/python.py:500: in _importtestmodule
    mod = self.fspath.pyimport(ensuresyspath=importmode)
../../../miniconda/envs/test/lib/python2.7/site-packages/py/_path/local.py:668: in pyimport
    __import__(modname)
../../../miniconda/envs/test/lib/python2.7/site-packages/duecredit/injections/injector.py:285: in __import
    mod = self._orig_import(name, *args, **kwargs)
E     File "/home/travis/build/MDAnalysis/mdanalysis/testsuite/MDAnalysisTests/topology/test_mmtf.py", line 53
E       sorted({*TestMMTFParser.expected_attrs}
E               ^
E   SyntaxError: invalid syntax
```

package/MDAnalysis/coordinates/MMTF.py
orbeckst: So what's the conclusion for here? Wait for `AttributeError` to show up in strange places or use `None`?

We could guess a bounding box if we really wanted to but this seems too much magic under the hood (so I am _not_ in favor of guessing).

orbeckst: Sorry, shoul've looked at new diff: `None`.

testsuite/MDAnalysisTests/topology/test_mmtf.py
```diff
@@ -46,6 +46,32 @@ class TestMMTFSkinny(ParserBase):
     expected_n_residues = 134
     expected_n_segments = 2
 
+    @pytest.mark.parametrize('attr',
+                             # for all attributes often in MMTF,
+                             # check that we get expected error on access
+                             # (sort so pytest gets reliable order)
+                             sorted({*TestMMTFParser.expected_attrs}
```
orbeckst: This does not work in Python 2

richardjgowers:
@MDAnalysis/coredevs can someone dunk this for me?

PR #2201, opened 2019-02-08
Pbc fixes

### LINKED ISSUE #2200: box conversions and PBC corner cases
**Expected behavior**

* When transforming (valid) triclinic dimensions from a `[lx, ly, lz, alpha, beta, gamma]` box representation into a matrix representation using `lib.mdamath.triclinic_vectors()` and then back to the previous one using `lib.mdamath.triclinic_box()`, the original representation should be retained as exactly as possible.
* Furthermore, related to the above, if the box has one 90° angle (or more), there should be at least one (or more) entries in the lower triangle of the box matrix that are exactly zero.
* Also related (to some extent): Wrapped atom coordinates should reside on the side of the box that is closer to zero in the respective direction. For example, an atom originally residing in the box corner furthest away from the origin should reside in the origin after wrapping (i.e., applying PBC).

**Actual behavior**

* Triclinic boxes, when transformed to a matrix representation and back, do not exactly retain their original angles and vector lengths.
* Triclinic box matrices with one ore more 90° angles do not have zeros in their lower triangle.
* Coordinates stay at the "far" edges of the box after wrapping (also happens with orthorhombic boxes).

``` python
import MDAnalysis as mda
import numpy as np

from MDAnalysis.lib.mdamath import triclinic_vectors, triclinic_box
from MDAnalysis.lib.distances import apply_PBC

print("\n\nTriclinic box matrix\n--------------------")
# triclinic box with two 90° angles and one 70° angle:
box = np.array([10.1, 10.1, 10.1, 90.0, 90.0, 70.0], dtype=np.float32)
print("Triclinic box:\n", box)
print("Triclinic vectors:\n", triclinic_vectors(box))


print("\n\nTriclinic box transform cycle\n-----------------------------")
# triclinic box with 70° angles:
box = np.array([10.1, 10.1, 10.1, 70.0, 70.0, 70.0], dtype=np.float32)
# transform box to matrix representation and back:
print("Box before transform cycle:\n", box)
print("Box after transform cycle:\n", triclinic_box(*triclinic_vectors(box)))


print("\n\nOrthorhombic wrap\n-----------------")
# cubic box with edge length 10.1:
box = np.array([10.1, 10.1, 10.1, 90.0, 90.0, 90.0], dtype=np.float32)
# suppose an atom lies in the corner furthest away from the origin:
box_corner = box[:3]
# try to wrap coordinates:
print("Box:\n", box)
print("Coordinates before wrapping:\n", box_corner)
print("Coordinates after wrapping:\n", apply_PBC(box_corner, box))


print("\n\nTriclinic wrap\n--------------")
# triclinic box:
box = np.array([10.1, 10.1, 10.1, 70.0, 70.0, 70.0], dtype=np.float32)
# suppose an atom lies in the corner furthest away from the origin:
box_corner = triclinic_vectors(box).sum(axis=0)
# try to wrap coordinates:
print("Box:\n", box)
print("Coordinates before wrapping:\n", box_corner)
print("Coordinates after wrapping:\n", apply_PBC(box_corner, box))
```

**Output**
```
Triclinic box matrix
--------------------
Triclinic box:
 [10.1 10.1 10.1 90.  90.  70. ]
Triclinic vectors:
 [[ 1.0100000e+01  0.0000000e+00  0.0000000e+00]
 [ 3.4544036e+00  9.4908962e+00  0.0000000e+00]
 [-4.4148504e-07 -3.0913114e-07  1.0100000e+01]]  # [2, 0] and [2, 1] should be zero


Triclinic box transform cycle
-----------------------------
Box before transform cycle:
 [10.1 10.1 10.1 70.  70.  70. ]
Box after transform cycle:
 [10.1      10.100001 10.1      69.99999  69.99999  69.99999 ]  # should be as above


Orthorhombic wrap
-----------------
Box:
 [10.1 10.1 10.1 90.  90.  90. ]
Coordinates before wrapping:
 [10.1 10.1 10.1]
Coordinates after wrapping:
 [10.1 10.1 10.1]  # should be all zeros


Triclinic wrap
--------------
Box:
 [10.1 10.1 10.1 70.  70.  70. ]
Coordinates before wrapping:
 [17.008808 11.909696  9.177501]
Coordinates after wrapping:
 [10.1  0.   0. ]  # should be all zeros
```

0.19.3-dev
Python 3.6
Ubuntu 18.04

zemanj:
Fixes #2200 

 - Computations in `triclinic_vectors()` are performed in double precision and all-zeros are returned for invalid boxes; 90° angles are handled exactly. Method is now more exact *and* faster. The function now also has an optional `dtype` kwarg specifying the dtype of the returned matrix.
- Likewise, computations in `triclinic_box()` are performed in double precision.
- Faster and more exact volume calculation in `lib.mdamath.box_volume()`.
- Coordinate wrapping has been enhanced on the C-level by introducing a ~two~three-stage method which first tries to do a single shift (which suffices in 99.9 % of real-life cases), and if that is not enough, falls back to the usual multi-box shift approach, followed by an additional single-box shift if required. This means that coordinates are now *guaranteed* to lie within the box but below the upper box boundaries after wrapping.
I noticed a tiny speed-up of 5% for ortho wrapping and a 25% gain for triclinic wrapping.

The output of the code posted in #2200 now looks as follows:
```
Triclinic box matrix
--------------------
Triclinic box:
 [10.1 10.1 10.1 90.  90.  70. ]
Triclinic vectors:
 [[10.1        0.         0.       ]
 [ 3.4544036  9.490896   0.       ]
 [ 0.         0.        10.1      ]]  # note the zeros in the last row


Triclinic box transform cycle
-----------------------------
Box before transform cycle:
 [10.1 10.1 10.1 70.  70.  70. ]
Box after transform cycle:
 [10.1 10.1 10.1 70.  70.  70. ]  # box is exactly retained


Orthorhombic wrap
-----------------
Box:
 [10.1 10.1 10.1 90.  90.  90. ]
Coordinates before wrapping:
 [10.1 10.1 10.1]
Coordinates after wrapping:
 [0. 0. 0.]  # coordinate is shifted to the origin


Triclinic wrap
--------------
Box:
 [10.1 10.1 10.1 70.  70.  70. ]
Coordinates before wrapping:
 [17.008808 11.909696  9.177501]
Coordinates after wrapping:
 [0. 0. 0.]  # coordinate is shifted to the origin
```

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/lib/mdamath.py
```diff
@@ -158,96 +160,241 @@ def _angle(a, b):
...
-    return np.linalg.det(triclinic_vectors(dimensions))
+    dim = np.asarray(dimensions, dtype=np.float64)
+    lx, ly, lz, alpha, beta, gamma = dim
+    if alpha == beta == gamma == 90.0 and lx > 0 and ly > 0 and lz > 0:
+        # valid orthogonal box, volume is the product of edge lengths:
+        volume = lx * ly * lz
+    else:
+        # triclinic or invalid box, volume is determinant of box matrix
+        # (invalid boxes are handled by triclinic_vectors):
+        volume = sarrus_det(triclinic_vectors(dim))
```
tylerjereddy: If the boxes are always convex, I wonder how this performance compares with the volume calculation availble from Qhull via ConvexHull in SciPy

zemanj: @tylerjereddy 
> If the boxes are always convex

How would you construct a non-convex 3d shape from 3 linearly independent vectors?
> I wonder how this performance compares with the volume calculation available from Qhull via ConvexHull in SciPy

`scipy.spatial.ConvexHull` requires points as input, so you won't get around computing the triclinic box matrix from dimensions. From this matrix you then have to construct the remaining four corner points of the box (plus origin) and put those in a numpy array... What I want to say is that the determinant will probably be already computed before `ConvexHull` is even called.

tylerjereddy: It is just a thought -- it seems the convexity assumption is solid as I expected, but you're likely right that there are too many operations in between.

The time required to test the two approaches probably isn't worth it, but intuition isn't always right for complexity analysis. If there was something faster in Qhull that just needed to be exposed in Cython I could always look into it.

tylerjereddy:
If we don't have `asv` benchmarks for the performance enhancements here that may be something to think about moving forward. I would guess we don't since some pbc machinery is newer & the numbers aren't cited using an asv comparison here, but could be wrong.

package/MDAnalysis/lib/mdamath.py
zemanj: I just noticed that computing the determinant here doesn't make sense since `triclinic_vectors()` is guaranteed to return a lower triangular matrix. Therefore, the volume is simply the product of the diagonal.
`numpy.diag(matrix).prod()` is not an option since that takes 3.4 µs.
`matrix[0, 0] * matrix[1, 1] * matrix[2, 2]` takes 384 ns, and scipy *cannot* beat that simply due to the time the function call alone will take.
Not to mentiaon that this is an unnecessary micro-optimization anyway.

The reason for the Sarrus determinant is a different one; In the end I want to be able to compute large batches of 3x3 matrix inverses in a fast way, e.g., for computing the angular velocity vector from the moment of inertia and the angular momentum per molecule. `numpy.linalg.inv()` used to be horribly slow for that.

tylerjereddy: hmm, some of that may depend on the linear algebra backend being used (MKL / openblas?). Anyway, it sounds like you're thinking about this very carefully which is great. Maybe I can follow up with some `asv` benchmarks at some point so we can be sure there's no performance regression on your hard work.

package/MDAnalysis/lib/_cutil.pyx
```diff
@@ -186,14 +187,14 @@ def make_whole(atomgroup, reference_atom=None, inplace=True):
         are returned as a numpy array.
     """
     cdef intset refpoints, todo, done
-    cdef int i, j, nloops, ref, atom, other, natoms
+    cdef np.intp_t i, j, nloops, ref, atom, other, natoms
```
richardjgowers: I don't this is going to break anything, but fwiw the `intmap` above uses `int` types, whereas this becomes a `ssize_t` type.  I think `np.intc` is the better match for `int` in C

https://docs.scipy.org/doc/numpy/user/basics.types.html#array-types-and-conversions-between-types

Can change the `map`/`set` objects up top to use the indexing integers and everyone is happy?

package/MDAnalysis/lib/include/calc_distances.h
```diff
@@ -43,93 +43,249 @@ static void minimum_image(double *x, float *box, float *inverse_box)
...
+                        // check if multi-box shifts are required:
+                        if (crd[0] >= ubound) {
+                            msr = 1;
+                        }
+                    }
+                }
+            }
+        }
+        // multi-box shifts required?
+        if (msr) {
```
richardjgowers: this is nice, I'm guessing branch prediction will help a lot too

package/MDAnalysis/lib/_cutil.pyx
zemanj: I just changed this to get rid of compiler warnings. `ssize_t` will translate to `int32_t` on 32 bit systems and to `int64_t` on 64 bit machines. It cannot break anything since it is downcast when assigning it to an intmap.
`np.intc` is a pretty bad choice for counter variables since it translates to `int` in C.  That again is very often identical to `int32_t` but the standard only guarantees at least 16 bits!

Speaking of integer widths, there is something quite odd in MDAnalysis, which may become a problem in the (probably *very* distant) future: Atom indices are stored as `np.int64` and bond indices, which again refer to atoms, are stored as `np.int32`. That doesn;t really make sense, right?

richardjgowers: Yeah sounds like both should be 32 or 64. I doubt we're going to need 64 bit ints to represent them? Or better to have 64bit ints so we can quickly index using atom indices (common)?

zemanj: I think 32bits should be way enough to accomodate even giant system sizes. 2 ** 31 - 1 is pretty large after all.
IIRC numpy uses `np.intp` for indexing, so one should choose that for index arrays. (unless someone runs MDA on a 16 bit machine... :stuck_out_tongue_winking_eye:)

zemanj: I just tried to correct all the types in `_cutil.pyx` but it turns out to be in fact impossible. The reason is exactly the difference between the atom index representation in `AtomGroup._ix` (`np.int64`) and the bond indicex representation (`np.int32`).

I wanted to find out which type is actually used to store the bond indices and worked my way through the entirely messed up `Bonds`/`_Connection`/`TopologyAttr`/`TopologyGroup` zoo:

`type(u.atoms.bonds)` says it's a `TopologyGroup`, so that's where I started.
`TopologyGroup.indices` of an empty `TopologyGroup` returns an empty `np.int32` array even though its `_bix` member is default-initialized with `np.array([])`, which defaults to `np.float64`.
A non-empty `TopologyGroup` will return `TopologyGroup._bix` when `TopologyGroup.indices` is accessed. `TopologyGroup._bix` has the dtype of whatever is passed as its `bondidx` argument in `TopologyGroup.__init__()`.
Such a `TopologyGroup` object is returned from a `_Connection` object (`Bonds` is one) in its `_Connection.get_atoms()` method. This method initializes the `TopologyGroup` with an `np.int32` index array which is compiled from `_Connection`'s cached `_bondDict` member. The indices stored in `_Connection._bondDict` have the dtype of whatever was used to initialize `_Connection.values` from the `values` argument of `_Connection.__init__()`.
If I got it right, `_Connection` objects usually come into life when a `Universe` is created, and are effectively instantiated by either a topology parser,  in `core.Universe.Merge()`, or in `AtomGroup.guess_bonds()`.
I really didn't want to check all topology parsers for the kinds of bond objects they return (quickly grepping the parsers suggests mostly `np.int32` but also sometimes `np.int64` index arrays for atoms and/or bonds), but I could figure out that `core.Universe.Merge()` instantiates bonds with a list of `(np.int64, int)` tuples (yes, the second item is a Python int) as the `values` argument.
`AtomGroup.guess_bonds()` boils down to initializing bonds with a list of `(np.int64, np.int64)` tuples (atom indices) in `topology.guessers.guess_bonds()`.

So, in the end, I guess it's (mostly, but, as we could see, not always) up to the topology parser to choose the data type used to store the *actual* bond indices.

Now, what happens if one calls `atoms.bonds.to_indices()`? That's where the journey became quite interesting:
`atoms.bonds` invokes the "injected" getter method `TopologyAttr.__getitem__()`, which then invokes `self.get_atoms(atoms)`, which is implemented in  `_Connection.get_atoms(atoms)`. In there, a `TopologyObject` is instantiated with an `np.int32` index array created from whatever resides in `self._bondDict` (discussed above).
Now, since `atoms.bonds` boils down to an *always newly created* `TopologyObject`, calling `atoms.bonds.to_indices()` means actually calling that object's `to_indices()` method, which again calls its own `indices` property, which then returns unique elements of the very same index array it had just been instantiated with.

I wouldn't even dare to mention all the overhead taking place in the various methods and initializers on the way. I always wondered why working with these things is so damn slow. Now I know.

zemanj: Long story short, changing the bond index dtype to `np.int64` or, ideally, to `np.intp` is a piece of cake. However, changing the dtype of atom indices to `np.intp` will break a lot of low-level thingies implemented on the Cython- or C-level.

zemanj:
I added more tests for `triclinic_box()` and `triclinic_vectors()`.

Furthermore, I added an additional single-box shift after any possibly occuring multi-box shift. That way, it is now *guaranteed* that coordinates will always reside within the box and below the upper box boundaries after wrapping.
This only has a performance impact in exotic cases, e.g., if a trajectory is entirely unwrapped. For grid-based methods, this means that overflow checks are no longer required when sorting particles in a grid of the size of the unit cell.

package/MDAnalysis/lib/_cutil.pyx
richardjgowers: @zemanj I'm happy to admit the workings of bonds can be improved, it was one of the first things I ever contributed to MDA :p And when I rewrote how topologies worked I wasn't really concentrating on getting bonds "nice".  That might be why there's a confusing 2 layer feel for how they work, there's essentially the `TopologyGroup` system I did first, then this got translated into a `TopologyAttribute` quite lazily (as I was putting out fires in other places).

The `dtype` of a TopologyAttribute should* be defined in the class (there's an `asarray` call to enforce this), so it can be singly defined there, then parsers should be hands off wrt defining this.

I think maybe we can squash some of the bond functionality (eg `TopologyDict`, maybe an "everything" `TopologyGroup` a la `u.atoms`) into the `_Connection` class.

* there's going to be a lot of shoulds that aren't followed, but that's just required cleanup

zemanj: @richardjgowers I am sorry, my judgement was probably a bit too harsh. I was just a bit frustrated as you might have imagined. :wink: 
I think there's always going to be room for improvement, and that certainly also applies to my own contributions.

In the future, we should make sure to use (global) index arrays instead of lists/tuples/dicts of objects whenever that's feasible, i.e., get as close as possible to using primitive types. Of course, one should also have maintainability in mind, so there might be some trade-offs to consider.

zemanj:
There were still cases where coordinates might have ended up at the upper box boundary, for example when a coordinate was just below zero, say, at `x = -1.0e-7`and got shifted by one box. I fixed that and added corresponding test cases.

zemanj:
@MDAnalysis/coredevs This is ready for review!

zemanj:
@richardjgowers Thank you for reviewing and fixing the CHANGELOG typos!

PR #2204, opened 2019-02-13
Wrap fixes

### LINKED ISSUE #2203: *Group.wrap() acts on non-member atoms
**Expected behavior**

`*Group.wrap()` should always act exclusively on atoms which actually belong to the calling group, regardless of the chosen `compound`.

**Actual behavior**

`*Group.wrap()` always acts on whole `segments`/`residues`/`molecules`/`fragments` if the method's `compound` parameter is chosen accordingly, even if the group contains only a single atom.

``` python
import numpy as np
import MDAnalysis as mda
u = mda.Universe(PSF, DCD)
# set appropriate box:
u.dimensions = [10, 10, 10, 90, 90, 90]
# select residue 0:
rg = u.residues[[0]]
# make an AtomGroup containing only the first atom of residue 0:
ag = rg.atoms[[0]]
# store the positions of the remaining atoms of residue 0:
remaining = rg.atoms[1:]
remaining_positions = remaining.positions
# wrap AtomGroup by residue:
ag.wrap(compound='residues')
# check if the other atoms remained at their previous positions:
assert np.all(remaining_positions == remaining.positions)  # raises AssertionError!
```

- MDAnalysis version 0.19.3-dev
- Python 3.6
- Ubuntu 18.04

zemanj:
Fixes #2203 

 - `*Group.wrap()` is now guaranteed to only work on atoms actually belonging to the calling group.
 - `*Group.wrap()` now returns wrapped positions and can optionally work out-of-place
 - Improved performance of `AtomGroup` creation from `ResidueGroup`s and `SegmentGroup`s via `ResidueGroup.atoms` or `SegmentGroup.atoms`, respectively. This speeds up all methods using these mechanisms.
 
Furthermore, `*Group.pack_into_box()` now calls `*Group.wrap()` with default arguments. In principle, `*Group.pack_into_box()` could (and maybe should?) be deprecated.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


zemanj:
**Performance report**

Test code:
```python
import MDAnalysis as mda
import numpy as np
from timeit import timeit

u = mda.Universe("topol.tpr", "traj_comp.xtc")
ag, rg, sg = u.atoms, u.residues, u.segments

print(repr(ag))
print(repr(rg))
print(repr(sg))

# populate caches
_ = ag.pack_into_box(inplace=False)
_ = rg.pack_into_box(inplace=False)
_ = sg.pack_into_box(inplace=False)

t_ag = timeit(lambda: ag.pack_into_box(inplace=False), number=10000) * 100
t_rg = timeit(lambda: rg.pack_into_box(inplace=False), number=10000) * 100
t_sg = timeit(lambda: sg.pack_into_box(inplace=False), number=10000) * 100

print("AtomGroup:    {:6.1f} µs".format(t_ag))
print("ResidueGroup: {:6.1f} µs".format(t_rg))
print("SegmentGroup: {:6.1f} µs".format(t_sg))
```

Output before this PR:
```
<AtomGroup with 15000 atoms>
<ResidueGroup with 2000 residues>
<SegmentGroup with 2 segments>
AtomGroup:     208.5 µs
ResidueGroup: 1910.3 µs
SegmentGroup:  806.0 µs
```
Output after this PR:
```
<AtomGroup with 15000 atoms>
<ResidueGroup with 2000 residues>
<SegmentGroup with 2 segments>
AtomGroup:     209.1 µs
ResidueGroup:  554.4 µs
SegmentGroup:  537.9 µs
```

Speed-up factors:
AtomGroup: **1.00** (basically no change)
ResidueGroup:  **3.45**
SegmentGroup:  **1.5**

zemanj:
@MDAnalysis/coredevs This is ready, please comment / review

package/MDAnalysis/core/topology.py
```diff
@@ -324,9 +325,9 @@ def segments2residues_1d(self, six):
 
         """
         try:
-            return np.concatenate([self._SR[s] for s in six])
-        except TypeError:
-            return self._SR[six].copy()
+            return np.concatenate(self._SR[six])
+        except ValueError:  # six is not iterable, length 1 or empty
+            return self._SR[six].astype(np.int64, copy=True)
```
richardjgowers: @zemanj is this changing because instead of trying to iterate an integer and getting a `TypeError`, we instead pass an integer to `concatenate` and get a `ValueError`?  And then we don't need the list comprehension because we're just fancy indexing `_SR`?

The `astype` call seems unnecessary, so why not just `np.copy()` if we want to force a copy?

package/MDAnalysis/core/groups.py
```diff
@@ -3075,11 +3162,8 @@ def atoms(self):
         are further ordered by :class:`Segment` in the :class:`SegmentGroup`.
         Duplicates are *not* removed.
         """
-        # If indices is an empty list np.concatenate will fail (Issue #1999).
-        try:
-            ag = self.universe.atoms[np.concatenate(self.indices)]
-        except ValueError:
-            ag = self.universe.atoms[self.indices]
+        u = self.universe
+        ag = u.atoms[u._topology.tt.segments2atoms_1d(self._ix)]
```
richardjgowers: This looks ugly, but I know it's because of the design.  I think when I wrote this I liked having the translation table (`tt`) separate so one day we could optimise it....

package/MDAnalysis/core/topology.py
zemanj: The `astype()` is necessary if `six` is empty: `TransTable._SR.dtype.type` is `numpy.object_`, and the array returned by `TransTable.residues2atoms_1d()` should always be that of atom indices (even if it's empty). Unfortunately, `_SR[[]]` has the same dtype as `_SR`, so the type conversion is for that corner case.
I have to admit that the performance of `ndarray.copy()` seems to be roughly twice as fast as that of `ndarray.astype(..., copy=True)` but that doesn't really matter since the exception handling happening before the call is about a hundred times slower.

While looking at this again, I discovered that `TransTable._SR` is effectively constructed from `TransTable._AR`, which has dtype `np.intp`. Even though we probably don't really support 32 bit architectures so that `np.intp` will always end up being `np.int64`, we should be consistent here and use `np.intp`.

zemanj: > is this changing because instead of trying to iterate an integer and getting a `TypeError`, we instead pass an integer to `concatenate` and get a `ValueError`? And then we don't need the list comprehension because we're just fancy indexing `_SR`?

Exactly.

zemanj: My code comment is wrong, length 1 iterables are unproblematic.

zemanj:
@richardjgowers I rebased against develop, corrected the erraneous code comments, and changed the dtype conversion to `np.intp`.

zemanj:
@richardjgowers ty for reviewing!

Issue #2191, opened 2019-01-30
Import of Test Examples changes matplotlib backend


uliano:
**Expected behavior**

Importing  example datafiles should not affect matplotlib backend as with 'agg' bad things happen (cfr.  [7])

**Actual behavior**

```
uliano@dewey:~$ ipython
Python 3.7.1 | packaged by conda-forge | (default, Nov 13 2018, 10:30:07) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.2.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import matplotlib                                                       
In [2]: matplotlib.get_backend()                                                
Out[2]: 'Qt5Agg'
In [3]: from MDAnalysis.tests.datafiles import PSF,DCD                          
In [4]: matplotlib.get_backend()                                                
Out[4]: 'agg'
In [5]: import matplotlib.pyplot as plt                                         
In [6]: plt.plot([1,2])                                                         
Out[6]: [<matplotlib.lines.Line2D at 0xb18bdd860>]
In [7]: plt.show()                                                              
/opt/anaconda3/lib/python3.7/site-packages/matplotlib/figure.py:445: UserWarning: Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.
  % get_backend())
```

**Currently version of MDAnalysis**

```
uliano@dewey:~$ python -c "import MDAnalysis as mda; print(mda.__version__)"
0.19.2
uliano@dewey:~$ python -V
Python 3.7.1
uliano@dewey:~$ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.14.2
BuildVersion:	18C54
```

orbeckst:
Fair enough, but honestly I have been struggling to understand how to do this differently because matplotlib does not allow us to change the backend later. How can we ensure that our tests run with the agg backend and not with any of the GUI ones?

richardjgowers:
We could do a pytest fixture that switches the backend to/from, something like:

```python
@pytest.fixture
def mpl():
    curr = matplotlib.get_backend()
    try:
        matplotlib.use('agg')
        yield
    finally:
        matplotlib.use(curr)

def test_plot(mpl):
    # test with backend temporarily changed
```

But matplotlib complains about the backend being set after pyplot import, so there's another step I'm missing.  Maybe a forced reload of the package?

jbarnoud:
[...]

PR #2205, opened 2019-02-14
Correcting Documentation in contacts.py

NinadBhat:
Fixes #

Found a minor mistake in the Docstring in contacts.py. Should I create an Issue for this?

 - 

PR Checklist
------------
 - [ ] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/CHANGELOG
```diff
@@ -75,6 +75,7 @@ Fixes
     *exluding* its upper boundaries) (Issue #2200, PR #2201)
   * *Group.wrap() is now guaranteed to only work on atoms actually belonging to
     the calling group. (Issue #2203, PR #2204)
+  * Fixed documentation in analysis/contacts.py
```
zemanj: Since the changelog is intended for users, it's better to specify the module instead of the file:
```suggestion
  * Fixed documentation of MDAnalysis.analysis.contacts
```

package/MDAnalysis/analysis/contacts.py
```diff
@@ -63,6 +63,7 @@
     from MDAnalysis.analysis import contacts
     from MDAnalysis.tests.datafiles import PSF,DCD
     import matplotlib.pyplot as plt
+    import numpy as np
```
zemanj: The example currently fails because of Issue #2191. Thus, in the meantime, you need to import `matplotlib.pyplot` *before* importing `MDAnalysis.tests.datafiles` for it to work. To keep the MDAnalysis imports together, I'd suggest the following order:
```python
import numpy as np
import MDAnalysis as mda
from MDAnalysis.analysis import contacts
from MDAnalysis.tests.datafiles import PSF,DCD
```

zemanj:
@MDAnalysis/coredevs CodeCov seems to be screwing up. Any idea why?

jbarnoud:
@zemanj I do not know why codecov chokes, though there is no reason for the coverage of this patch to be relevant. I think we can ignore the "failure".

@NinadBhat I do not see your github handle in the authors of the version, You can add yourself there, and in the AUTHORS file if you are not there already.

tylerjereddy:
NumPy & SciPy both mostly disabled allowing codecov to fail PRs because it is just so problematic / bizarre sometimes, and doesn't work as documented when you try to change things. It would be great if the ecosystem had some better competition in this area, but I think coveralls (?) has it own (possibly worse) issues. Nice to have the entry in the CI list so you can still click through & check the cov though.

I think we also banned the bot in those projects -- people don't like the email / message spam & rarely find it helpful.

zemanj:
@NinadBhat Thanks for keeping an eye on the docs! Such things slip through the cracks quite easily.

PR #2213, opened 2019-02-28
Sp box

### LINKED ISSUE #2190: Inconsistent dtype for universe.dimensions
**Expected behavior**

The `dtype` of the dimensions for a standard (not in-memory) universe should be consistent with an in-memory universe. This is in particular important for the low-lewel C functions like the `make_whole` function which will fail for the in-memory trajectory, since it requires a [float](https://github.com/MDAnalysis/mdanalysis/blob/992bd738f5cd6c1a16e8e55fc82cb3f442bf239c/package/MDAnalysis/lib/_cutil.pyx#L181-L183).

**Actual behavior**

The dimension of the standard universe representation is a `float` and the in-memory representation is a `double`.

``` python
import MDAnalysis as mda

u = mda.Universe(PSF, DCD)
u2 = mda.Universe(PSF, DCD, in_memory=True)

print(u.dimensions.dtype)
print(u2.dimensions.dtype)
```

I'm using Python 3.6 and the dev version of MDAnalysis on MacOS

zemanj:
Fixes #2190

 - Enforced `dtype=numpy.float32` for `Universe.dimensions` and `Timestep.dimensions`

I searched the code base and found the same lines as indicated by @jbarnoud and @PicoCentauri in the discussion of #2190 (see comments of [@jbarnoud](https://github.com/MDAnalysis/mdanalysis/issues/2190#issuecomment-458902209) and [@PicoCentauri](https://github.com/MDAnalysis/mdanalysis/issues/2190#issuecomment-458908655)).

No calls of `numpy.astype()` to make @richardjgowers happy. :wink:

Checks for valid boxes as outlined in my [comment](https://github.com/MDAnalysis/mdanalysis/issues/2190#issuecomment-462355339) are not included since that requires a more in-depth discussion about what defines a valid box (semi-periodic systems, negative angles, etc.).

I don't think we need to adjust the docs, right?

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -87,6 +87,8 @@ Fixes
   * *Group.wrap() is now guaranteed to only work on atoms actually belonging to
     the calling group. (Issue #2203, PR #2204)
   * Fixed documentation of MDAnalysis.analysis.contacts
+  * Enforced consistent dtype (numpy.float32) of Universe.dimensions and
+    Timestep.dimensions for all Universes and Readers (Issue #2190, PR #2213)
 
 Deprecations
   * analysis.polymer.PersistenceLength no longer requires the perform_fit()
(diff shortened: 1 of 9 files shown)


zemanj:
@richardjgowers Thanks, that was record-breaking fast! :+1:

PR #2215, opened 2019-03-01
Use capped_distance in analysis.distances.contact_matrix  

### LINKED ISSUE #2102: Change analysis.distances.contact_matrix to use capped_distance
So currently we have `contact_matrix`:
https://github.com/MDAnalysis/mdanalysis/blob/develop/package/MDAnalysis/analysis/distances.py#L59

Which finds all contacts within a certain distance and returns a sometimes sparse boolean array

This should use `lib.distances.self_capped_distance` so that it doesn't do the full `0.5 * N * (N-1)` search

Todo:
 - [ ] verify that this is faster
 - [ ] add benchmark to `benchmarks/` (will also be visible in the existing leaflet benchmark)
 - [ ] change `contact_matrix` to use `self_capped_distance`

NinadBhat:
Fixes #2102 

 Replaced distance array in analysis.distances.contact_matrix with capped_distance

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/analysis/distances.py
+++ b/package/MDAnalysis/analysis/distances.py
@@ -40,8 +40,8 @@ See Also
 
 from __future__ import absolute_import
 
-__all__ = ['distance_array', 'contact_matrix',
-           'dist', 'between']
+__all__ = ['distance_array', 'self_distance_array',
+           'contact_matrix', 'dist', 'between']
 
 import numpy as np
 import scipy.sparse
--- a/package/MDAnalysis/analysis/distances.py
+++ b/package/MDAnalysis/analysis/distances.py
@@ -40,13 +40,13 @@ See Also
 
 from __future__ import absolute_import
 
-__all__ = ['contact_matrix', 'capped_distance'
+__all__ = ['distance_array', 'contact_matrix',
            'dist', 'between']
 
 import numpy as np
 import scipy.sparse
 
-from MDAnalysis.lib.distances import capped_distance
+from MDAnalysis.lib.distances import capped_distance, distance_array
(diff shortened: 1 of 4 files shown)


zemanj:
Should make no difference

package/MDAnalysis/analysis/distances.py
```diff
@@ -40,13 +40,13 @@
 
 from __future__ import absolute_import
 
-__all__ = ['distance_array', 'self_distance_array',
-           'contact_matrix', 'dist', 'between']
+__all__ = ['contact_matrix', 'capped_distance'
+           'dist', 'between']
```
richardjgowers: So the history here is that `self_distance_array` and `distance_array` used to live in this module, so when we moved them we had to maintain the API (else we get bugs like you found where `analysis.contacts` imported from here.).

So because this might break someones code if they're doing `from analysis.distances import distance_array` this change needs reverting.

NinadBhat:
@richardjgowers I am not sure why the tests are failing now.
It's showing

>  Could not find a version that satisfies the requirement clustalw==2.1.* (from versions: )
> No matching distribution found for clustalw==2.1.*

NinadBhat:
>  Could not find a version that satisfies the requirement clustalw==2.1.* (from versions: )> No matching distribution found for clustalw==2.1.*

richardjgowers:
Don’t worry about that, it’s not related.
  

  
  

  
  
>

richardjgowers:
Awesome, thanks again @NinadBhat !

PR #2221, opened 2019-03-16
Writes compressed output of given format

### LINKED ISSUE #2216: Write compressed ascii outputs
We can currently read ascii files that are compressed (eg `mda.Universe('this.pdb.gz')`).  It would be cool if `u.atoms.write('new.pdb.gz')` also worked (ie wrote a pdb file that was compressed with gz).

I'm not 100% sure how this should work, but one idea is to add a `GZWriter` which acts as a wrapper, calling the "real" format Writer and then compressing whatever was created.  So calling chain would be 

- `u.atoms.write('something.pdb.gz')` makes a `GZWriter`
- `GZWriter` makes a `PDBWriter`
  - maybe `GZWriter` can get the child Writer to write to a stream to lessen I/O operations?
- `PDBWriter` does its thing (oblivious to whats happening)
- `GZWriter` then compresses the PDB file
  - (or compresses the stream and writes that?)
- ????
- Profit

Other things to consider
- check that compression makes sense (ie only ascii formats, rather than letting people try and compress a dcd)
- making sure errors from child writers are properly passed along (ie the wrapper is transparent where needed)
- other compression formats? (and generic `CompressionWriter`?)
- testssss

Fenil3510:
Fixes #2216 

 - Preferring `file_format` argument over `filename` in `get_writer_for` .
 - The compressed form of atomgroups can be written as `ag.write(filename.writer.gz)`
 - Tests are written

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?


Fenil3510:
Not sure why codecov is failing

Fenil3510:
I guess the test is already written for this in Class `TestGetWriterFor` in test_util.py, heres the test

```
def test_compressed_extension(self):
        for ext in ('.gz', '.bz2'):
            fn = 'test.gro' + ext
            writer = mda.coordinates.core.get_writer_for(filename=fn)
            assert writer == mda.coordinates.GRO.GROWriter
            # Make sure ``get_writer_for`` works with compressed file file names
def test_compressed_extension_fail(self):
        for ext in ('.gz', '.bz2'):
            fn = 'test.unk' + ext
            # Make sure ``get_writer_for`` fails if an unknown format is compressed
            with pytest.raises(TypeError):
                mda.coordinates.core.get_writer_for(filename=fn)
```

jbarnoud:
Thank you for your contribution!

Codecov complains because you added code which is not tested. You need to add a test that calls `ag.write` with a file name ending on `.gz` to go through the lines you added.

From what I see, your change would indeed allow to write a gziped file. Though, as you noticed, `get_writer_for` is already dealing with compressed file extensions. Therefore, I think a better solution would be to get rid of the redundant format logic in `write` rather than adding to it like you did.

@richardjgowers, it seems you wrote the lines around https://github.com/MDAnalysis/mdanalysis/blob/04056e35ccaee5343718b532ff88995fc000e1d9/package/MDAnalysis/core/groups.py#L3086. Do you remember why you extracted the format there rather than leaving `get_writer_for` deal with it?

Fenil3510:
@jbarnoud, ohh ok, I'll add such a test. from what I understood `get_writer_for` gets the relevant writer( PDB etc.), based on the file extension that is stored in `format` which is provided as an argument to it. I think format logic would be necessary as it needs to handle three cases
1. Usual File Extension (Which has a writer)
2. GZipped File
3. No File extension
`format` I guess is a single variable for selecting between these. But if there is something I'm missing, I'm willing to change the code accordingly to make remove redundancy, I'll add tests after the changes are confirmed.

jbarnoud:
@Fenil3510 The format discovery logic has to exist, indeed. My point is that it is already implemented in `get_writer_for` that is called in `ag.write`. In its current version, `ag.write` overwrite the format discovery logic and explicitly ask for a file format, instead of letting `get_writer_for` figure out what format to use.

The user should still be able to ask for a specific format by providing the `format` keyword argument to `ag.write`, though. But just passing the `format` argument untouched should be enough.

Fenil3510:
Oh @jbarnoud you are right. `get_writer_for` does handle compression thing with function calls `get_ext` and `check_compressed_format`, making the thing I did redundant. We can just pass format to `writer(filename, format = None)` and its done. I guess there is no need for `format` variable then. 
I'll make these changes.

jbarnoud:
This is what I had in mind.

But, you are breaking the `file_format` keyword argument of `ag.write`. This keyword argument should allow to bypass the format detection, and it is rather handy from time to time. The default value is "PDB", which is confusing because we actually default to guessing... The default should probably be `None`. This change should be documented, and we need to test what happens when the format cannot be guessed.

Fenil3510:
> But, you are breaking the `file_format` keyword argument of `ag.write`. This keyword argument should allow to bypass the format detection, and it is rather handy from time to time. 

@jbarnoud yes that makes sense. But when the user has given argument `filename.extenstion` and if the extension is valid then the `file_format` keyword would be redundant and preference would be given to the former to identify the format. This is what is happening currently if I'm right. So giving preference to `file_format` over extraction from `filename` is what you mean? ie. reversing the preference. 

The issue then will be that neither `file_format` nor extraction from `filename` can guess the format, in that case, we could use a default value like PDB?

jbarnoud:
Look at the signature of `get_writer_for`. https://github.com/MDAnalysis/mdanalysis/blob/a0bc631699f7619a12aa2ecda1e3fa06113c9507/package/MDAnalysis/core/_get_readers.py#L198

It takes an optional format. The guessing is only done if that format is not explicitly provided. https://github.com/MDAnalysis/mdanalysis/blob/a0bc631699f7619a12aa2ecda1e3fa06113c9507/package/MDAnalysis/core/_get_readers.py#L221

This is good practice because, if the user tells the function to do something, then the function should do what the user asked.

What you did in your last commit is to disregard an explicit format that the user may have passed with the `file_format` argument of `ag.write`.

In the ideal scenario for `ag.write`:
* the format is guessed by default
* if a format is explicitly provided, then that format is used

Fenil3510:
Ok, @jbarnoud I understand what you say but my confusion is in `get_writer_for` vs `get_selection_writer_for`.
Supposedly user provides `ag.write` with arguments `filename(without extension) , file_format = GRO` so this calls `get_writer_for` which throws an error (current implmentation) so `get_selection_writer_for` gets the control and writes in `GRO` format.  So if it is done such that `get_writer_for` handles all possible scenarios, `get_selection_writer_for` would be redundant, wouldn't it? Or is `get_selection_writer_for` for some different purpose?

jbarnoud:
`get_selection_writer_for` is about getting a selection writer. It is a different things, but `ag.write` can write either a structure (writer) or a selection (selection writer).

Fenil3510:
> In the ideal scenario for `ag.write`:
> 
> * the format is guessed by default
> * if a format is explicitly provided, then that format is used
Ok, thanks for the clarification about get_selection_writer. I will make changes as quoted above.

Fenil3510:
@jbarnoud a doubt, if the user writes `ag.write(filename.pdb.gz, file_format = 'GRO')` this results in a conflict in the things I'm about to change. Do these cases need to be take care of?  If `file_format` is to be given preference as you said then it must be saved as `filename.gro.gz` ?

jbarnoud:
As a general rule, it is very frustrating for users when a software tries to outsmart them. Indeed, the example you give presents a contradiction, but the request is clear: the user wants to write a GRO file named "filename.pdb.gz"; therefore it is what we should generate.

richardjgowers:
`format` is meant to overrule everything, the filename is just a (often very good) hint at the format.

I would have thought that you couldn't compress the file frame by frame, but instead you'd need to see the entire file before compressing.  It would be useful to see a test that writes a `'test.pdb.gz'`, then a `'test.pdb'` which gets compressed and check the relative file sizes? Also just a simple round trip test, so make a `test.pdb.gz`, create a new Universe from this file and check that positions & names match.

Fenil3510:
> `format` is meant to overrule everything, the filename is just a (often very good) hint at the format.
@richardjgowers as per the current code `format` extracted from `filename` overrules `file_format` argument, should I keep it that way? or Do as @jbarnoud suggests before writing the test?  
> Also just a simple round trip test, so make a `test.pdb.gz`, create a new Universe from this file and check that positions & names match.
Yes, I will write this test once changes are approved

Fenil3510:
> As a general rule, it is very frustrating for users when a software tries to outsmart them. Indeed, the example you give presents a contradiction, but the request is clear: the user wants to write a GRO file named "filename.pdb.gz"; therefore it is what we should generate.

Noted @jbarnoud, thanks

Fenil3510:
> `format` is meant to overrule everything, the filename is just a (often very good) hint at the format.
> 
>Also just a simple round trip test, so make a `test.pdb.gz`, create a new Universe from this file and check that positions & names match.

Ok @richardjgowers . I'll write the test as suggested.

Fenil3510:
@jbarnoud @richardjgowers I've made a fix. But this is by no means robust. There will be a bit of an issue in the compression part. How would the user indicate that a compressed format is needed? This would be via extraction `filename` itself, right?. The `file_format` argument is only to specify writers according to the current docstring. Should we extend `file_format` to take arguments like `pdb.gz` and `gro.gz`?

Fenil3510:
The build is failing as `get_selection_writer_for` is expecting file_format to be a default one(previously 'PDB'). As there is no default_format now, should I change `get_selection_writer_for` to use PDB if `file_format` is None or let it throw error?

orbeckst:
I haven't reviewed the code but I note that although the tests pass now, the coverage is still low. Do you still have to write tests?

Fenil3510:
@orbeckst, yes haven't added the tests yet. Will do as soon as changes are reviewed. There are some issues that need to be addressed that I've mentioned in the previous comments before tests be written.

Fenil3510:
@orbeckst @richardjgowers @jbarnoud.
The changes support
- Preferring `file_format` over `filename` in `get_writer_for` .
- The compressed form can be written as `filename.pdb.gz`
- Currently tested(locally) only for `PDB` and `GRO` for compression.  This work for every writer in MDAnalysis, right? 
- Tests to be written for compressed files.

orbeckst:
@Fenil3510 write the tests for what you have – people don't always have time to do reviews, and when they can make time then it's better that everything is already there, even if that means that you have to change things after the review.

orbeckst:
Do you have specific questions that you need answered?

Fenil3510:
> @Fenil3510 write the tests for what you have – people don't always have time to do reviews, and when they can make time then it's better that everything is already there, even if that means that you have to change things after the review.

I'm sorry, will follow it. I'm writing tests.

Fenil3510:
> It would be useful to see a test that writes a 'test.pdb.gz', then a 'test.pdb' which gets compressed and check the relative file sizes? Also just a simple round trip test, so make a test.pdb.gz, create a new Universe from this file and check that positions & names match.
Tests were written as per this and from what I understood. Any other tests expected?

orbeckst:
The failure of the doc build https://travis-ci.com/MDAnalysis/mdanalysis/jobs/189731062 is not your fault, I just had the same thing in PR #2231 . Maybe my patch there will fix this.

If you haven't added your name yet to AUTHORS then do that, too. 

Also add an entry to CHANGELOG.

package/MDAnalysis/coordinates/GRO.py
```diff
@@ -326,7 +326,7 @@ def __init__(self, filename, convert_units=None, n_atoms=None, **kwargs):
                w.write(u.atoms)
 
         """
-        self.filename = util.filename(filename, ext='gro')
+        self.filename = util.filename(filename, ext='gro', keep = True)
```
orbeckst: PEP8; no space around `=` in args.

package/MDAnalysis/core/groups.py
```diff
@@ -3054,8 +3054,7 @@ def write(self, filename=None, file_format="PDB",
         if filename is None:
             trjname, ext = os.path.splitext(os.path.basename(trj.filename))
             filename = filenamefmt.format(trjname=trjname, frame=trj.frame)
-        filename = util.filename(filename, ext=file_format.lower(), keep=True)
-
+        filename = util.filename(filename, ext= file_format or 'PDB', keep=True)
```
orbeckst: PEP8 !

orbeckst: Did you test that you can omit `.lower()`??

orbeckst: I don't like `file_format or 'PDB'`. Better be explicit so that only `None` really gets defaulted:
```python
file_format if file_format is not None else 'PDB'
```

package/MDAnalysis/core/groups.py
```diff
@@ -3106,7 +3098,7 @@ def write(self, filename=None, file_format="PDB",
         try:
             # here `file_format` is only used as default,
             # anything pulled off `filename` will be used preferentially
-            writer = get_selection_writer_for(filename, file_format)
+            writer = get_selection_writer_for(filename, file_format or 'PDB')
```
orbeckst: I don't like `file_format or 'PDB'`. Better be explicit so that only `None` really gets defaulted:
```python
file_format if file_format is not None else 'PDB'
```

testsuite/MDAnalysisTests/core/test_atomgroup.py
```diff
@@ -167,7 +167,27 @@ def test_write_frame_none(self, u, tmpdir, extension):
             u.atoms.positions[None, ...], new_positions, decimal=2
         )
 
+    @pytest.mark.parametrize('extension', ('xtc', 'dcd', 'pdb', 'xyz'))
```
orbeckst: Se should capitalized extensions, too.

testsuite/MDAnalysisTests/core/test_atomgroup.py
```diff
@@ -167,7 +167,27 @@ def test_write_frame_none(self, u, tmpdir, extension):
...
     def test_write_frames_all(self, u, tmpdir):
+        for ext in ('.gz', '.bz2'):
+            destination = str(tmpdir / 'test.dcd') + ext
+            u.atoms.write(destination, frames='all')
+            u_new = mda.Universe(destination)
+            ref_positions = np.stack([ts.positions for ts in u.trajectory])
+            new_positions = np.stack([ts.positions for ts in u_new.trajectory])
+            assert_array_almost_equal(new_positions, ref_positions)
+
+    def test_compressed_rite_frames_all(self, u, tmpdir):
```
orbeckst: "rite"?? Do you mean "right"? or "write"?

testsuite/MDAnalysisTests/core/test_atomgroup.py
```diff
@@ -167,7 +167,27 @@ def test_write_frame_none(self, u, tmpdir, extension):
...
 
+    @pytest.mark.parametrize('extension', ('xtc', 'dcd', 'pdb', 'xyz'))
+    def test_compressed_write_frame_none(self, u, tmpdir, extension):
+        for ext in ('.gz', '.bz2'):
+            destination = str(tmpdir / 'test.' + extension + ext)
+            u.atoms.write(destination, frames=None)
+            u_new = mda.Universe(destination)
+            new_positions = np.stack([ts.positions for ts in u_new.trajectory])
+            assert_array_almost_equal(
+                u.atoms.positions[None, ...], new_positions, decimal=2
```
orbeckst: `decimal=2` is a low bar, can't we increase the precision, at least for anything that is not XTC?

orbeckst: (XTC is a lossy format, so XTC compared to original might indeed only be accurate to 2 decimals)

orbeckst: I assume that you used what we already had in the tests... then ignore my comments here.

orbeckst:
@jbarnoud can you please also have a quick peek? Thanks.

orbeckst:
I merged the latest develop into your branch, which should fix the doc build failure (via 4f8dc1723f3fdcccd0d66d1c7ac0e93620d417fe).

testsuite/MDAnalysisTests/core/test_atomgroup.py
Fenil3510: So sorry. Did it in a hurry.

testsuite/MDAnalysisTests/core/test_atomgroup.py
Fenil3510: Yes, I've used from what is in the previous tests.

package/MDAnalysis/core/groups.py
Fenil3510: > Did you test that you can omit `.lower()`??

If file_format is None. `.lower()` will throw an error, so I have shifted it inside `util.filename` function.

package/MDAnalysis/coordinates/GRO.py
Fenil3510: Are we also restricting the number of characters per line to 80/79 as per pep8?

testsuite/MDAnalysisTests/core/test_atomgroup.py
Fenil3510: I'm afraid I don't understand. Should I capitalize all the extensions?. I've used the parameters from the previous test already written.

orbeckst:
oops, I killed travis #2233 ... fixing (so that the tests ACTUALLY run)

testsuite/MDAnalysisTests/core/test_atomgroup.py
Fenil3510: I've added a test for a capitalized extension 'PDB'.

Fenil3510:
> oops, I killed travis #2233 ... fixing (so that the tests ACTUALLY run)

Would need to commit again for Travis to build? As I see it is fixed now.

orbeckst:
Merged develop to get the travis fix from PR #2234.

package/MDAnalysis/coordinates/GRO.py
orbeckst: See the [Style Guide](https://github.com/MDAnalysis/mdanalysis/wiki/Style-Guide): yes, we try to keep within 79 characters. (Not all our code rigorously adheres to it, but that's what we're striving for)

testsuite/MDAnalysisTests/core/test_atomgroup.py
orbeckst: Sorry, clearly I should proof-read what I am typing... yes, test for capitalized extensions was what I was looking for.

orbeckst:
Congratulations @Fenil3510 to your first commit. Thank you for your effort.

orbeckst:
@Fenil3510 I sent you a GitHub invite so that you can be part of [@MDAnalysis/contributors](https://github.com/orgs/MDAnalysis/teams/contributors/members).

Fenil3510:
Happy to be a part of the team!

Issue #2181, opened 2019-01-21
doc example for crafting hbond autocorrelation groups


orbeckst:
**Expected behavior**

Creating the input atomgroups (donors, hydrogens, acceptors) for [hbond autocorrel](https://www.mdanalysis.org/docs/documentation_pages/analysis/hbond_autocorrel.html) should be easy/self explanatory.

**Actual behavior**

It is rather difficult with recurring questions on the mailing list (eg https://groups.google.com/d/msg/mdnalysis-discussion/-gwmJANsSXM/g-oD1byEFAAJ and issue tracker eg #1563), especially when it comes to water: It is not clear to most users how to [generate donor and hydrogen atomgroups that correspond to each other](https://www.mdanalysis.org/docs/documentation_pages/analysis/hbond_autocorrel.html#input)
> Three AtomGroup selections representing the hydrogens, donors and acceptors that you wish to analyse. Note that the hydrogens and donors selections must be aligned, that is hydrogens[0] and donors[0] must represent a bonded pair. If a single donor therefore has two hydrogens, it must feature twice in the donors AtomGroup.

**Proposed action**
Add an example to the docs that shows how to use the module with one of our test systems (e.g., with something from [MDAnalysisData](https://www.mdanalysis.org/MDAnalysisData/)) and in particular, water.

**Currently version of MDAnalysis**

All recent (~0.18 - 0.19) versions of MDA (but really since 0.9).

orbeckst:
@richardjgowers – I think you answered this question before. Can you find your answer(s) from the mailinglist or issue tracker and link them here? This would be a good starting point for the doc fix.

orbeckst:
Related:

> Further I noted there are two utilities for hydrogen bond lifetime calculation.
> https://www.mdanalysis.org/docs/documentation_pages/analysis/hbond_autocorrel.html
> https://www.mdanalysis.org/docs/documentation_pages/analysis/waterdynamics.html
> 
> Is there any difference in between them. Both are based on time-correlation defination of hydrogen bonds?


You need to read the papers in detail. I don't know the answer. The waterdynamics module is primarily for water and calculates a whole bunch of properties in a consistent manner. The hbond_autocorrel module is very flexible in what input it can take and has been used for, e.g., polymers. However, I wish I had a better and more concise answer for you – if the original authors (@richardjgowers  and @alejob ) read this then it would be great if they could come up with a better answer that we can include in the docs.

orbeckst:
[...]

PR #1995, opened 2018-07-18
SurvivalProbability - Readability, performance, and algorithm changes

bieniekmateusz:
Small changes:
- Performance (16 minutes to 60 seconds): 
-- issue due to the unnecessary function overhead,  
-- iterating over the dataset only once (RAM memory optimisation)
- More thorough documentation of parameters and additional sanity checks in __init__
- Removed temporarily the progress meter which only worked for loading the simulation dataset, confusing the reader
- Allows the user to access the distribution for each tau through sp.sp_timeseries
- Returns the taus along with the timeseries (opening access for dt). Not returning tau = 0 

Validity Test Cases Changes:
- Replaced the arbitrary test cases with defined datasets and predictable taus
- Added a test case where no atoms IDs are found. In this case, we return tau = NaN. This is because 0 means that the initial molecular leaves after a given tau. However, if there is no initial molecules in the first place, this would be the wrong conclusion. 

Algorithmic Changes made in this Pull Request - Request for Contribution from the Original Author: 
 - We modified the algorithm to be applicable to other atom groups (e.g. ions) besides water. Particularly, we changed the way the algorithm behaves in the case where there is no molecules found for the reference frame. Previously, one would expect the reference frame to always find some water molecules. If this was not the case, each tau was diluted - divided by (Nt + 1). Because it was unlikely to happen with any reasonable water selection, this case was not important. However, let us consider using SP the survival of ions in a given area, with frames where no ions are found in the first place (reference frame). For such frames, diluting the taus means that the ions leave earlier - whereas in fact, these ions might rarely diffuse into the selected area. The rare diffusion into the selected area should not affect how long they stay (ie their SurvivalProbability). Please let us know if we missed anything. 

Questions:
 - Since this algorithm is applicable to any mobile atoms in a system, we suggest this package is moved out of waterdynamics into a more general package. 

Future considerations:
- Adding progress bar
- Adding dt to allow sample the dataset with a faster shift (and possibly discuss data correlations for larger taus)

Edit: Acknowledgement - This work was done together with @p-j-smith, but we submitted most of it on my account. 

PR Checklist
------------
 - [X ] Tests?
 - [X] Docs?
 - [ ] CHANGELOG updated?
 - [ ] Issue raised/referenced?

Issue #1175, opened 2017-01-18
Restructure documentation for user-friendliness


dotsdl:
The [MDAnalysis documentation](http://pythonhosted.org/MDAnalysis/) is structured in such a way that it contains scant information on the [user interface](http://pythonhosted.org/MDAnalysis/documentation_pages/overview.html) and otherwise gives detail on the internal components of MDAnalysis that users should never really have to touch directly ([parsers](http://pythonhosted.org/MDAnalysis/documentation_pages/topology_modules.html), [readers and writers](http://pythonhosted.org/MDAnalysis/documentation_pages/coordinates_modules.html)). We've seen usage on the mailing list from users in which they try e.g. to directly use a `Reader` object to parse a trajectory, and the docs don't make it clear that this isn't really the way to go about things.

I think the docs should be restructured to have three parts. The first part should read more like the [MDAnalysis Tutorial](http://www.mdanalysis.org/MDAnalysisTutorial/), giving a friendly view as to *how* one uses `MDAnalysis` in practice. The second part should be an API reference, with the full sphinx autodoc output for the user interface components showed in part one. The last part should be the developer docs, which would include the reference docs for internal components, such as parsers, readers, and writers so that they are *somewhere*.

Thoughts?

richardjgowers:
I liked the idea of a tiered approach, first there should be an explanation of *how* it works, and only after an explanation of *why*.

richardjgowers:
So thinking about this, the only problem is that there is potential for the `rst` and `py` docs to diverge.

So one option would be to later implement doc testing on our first part/tutorial so that it always works as intended

http://doc.pytest.org/en/latest/doctest.html

kain88-de:
Related I just found out that scikit-learn has a Circle-CI setup that uploads the changes for a document  build. This allows a reviewer to quickly check the result and give comments without building it himself.

https://8425-843222-gh.circle-artifacts.com/0/home/ubuntu/scikit-learn/doc/_build/html/stable/_changed.html

orbeckst:
From the scikit-learn example that @kain88-de linked I found out about http://sphinx-gallery.readthedocs.io , which allows one to make nice docs from small scripts. Maybe worthwhile looking into in conjunction with the proposed examples gallery MDAnalysis/MDAnalysis.github.io#4 .

orbeckst:
Although this could go into a patch level, I am retargeting to 0.17.0 so that we can focus on fixing bugs for 0.16.1.

[...]

PR #2226, opened 2019-03-21
Survival probability: Intermittency, Step, Residues, Overlapping SP Selection Example

bieniekmateusz:
Fixes #
 - "Step" parameter does not load frames which are not used (dependant on the value of tau_max). This is great of sampling long trajectories. 

 - residue-wise SP can be carried out
 - Intermittency is defined as a number of consecutive absences allowed 
 - Intermittency combined with Step lead to consistent behaviour
 - the raw data are made available (helps with testing, allows building upon the values)
 - simplified and extended documentation to include SP around overlapping areas (for example, headgroups in lipids)
 - Test Cases for all new functionalities have been added 

PR Checklist
------------
 - [X] Tests?
 - [X] Docs?
 - [x] CHANGELOG updated?

We generalised the code to be applicable to other situations which is why we would like to ask if we should move it out of waterdynamics.py to its own file (survivalprobability.py in analysis). 

The intermittency, for the sake of simplicity, is defined in terms of consecutive absences. This is calculated in a single pass over the data. 

We've used the new functionalities and plan to publish on the topic to discuss the usefulness of these small changes. 

This request is done on behalf of me and @p-j-smith


bieniekmateusz:
Hi there, I only modified two files, however, I worked on the "develop" branch by mistake. So I tried to rebase the code to bring my commits to the front, which is why the review picks up so many files. 

I'll see if I can move the commits differently to avoid what happened here.

bieniekmateusz:
Hi again, I see that there are some functionalities which could help with organising the commits: git cherry-pick in particular. Let me know if it is necessary and then I'll further look into it.

orbeckst:
Looks like a clean diff, so no need to play more with `git rebase` and friends.

orbeckst:
I merged latest develop into your branch so that the docs build correctly and don't fail your build.

package/MDAnalysis/analysis/waterdynamics.py
```diff
@@ -297,13 +297,50 @@
...
+  sp_timeseries.insert(1, 0)
+
   # plot
   plt.xlabel('Time')
   plt.ylabel('SP')
   plt.title('Survival Probability')
   plt.plot(tau_timeseries, sp_timeseries)
   plt.show()
 
+Another example applies to the situation where you work with many different "residues".
```
orbeckst: Nice example. Good to have it here for now, eventually we would probably move it into a section with examples #1175 .

package/MDAnalysis/analysis/waterdynamics.py
```diff
@@ -297,13 +297,50 @@
...
+Here we calculate the SP of a potassium ion around each lipid in a membrane and
+average the results. In this example, if the SP analysis were run without treating each lipid
+separately, potassium ions may hop from one lipid to another and still be counted as remaining
+in the specified region. That is, the survival probability of the potassium ion around the
+entire membrane will be calculated.
+
+Note, for this example, it is advisable to use .Universe(in_memory=True) to ensure that the
```
orbeckst: use backticks for formatting
```reST
to use `Universe(in_memory=True)` to ensure
```
and don't use a leading do. It's ok to just name the class.

package/MDAnalysis/analysis/waterdynamics.py
```diff
@@ -297,13 +297,50 @@
...
+simulation does not loaded into RAM memory once per lipid::
```
orbeckst: Do you mean "... to ensure that the simulations is not loaded into RAM memory once per lipid"?

package/MDAnalysis/analysis/waterdynamics.py
```diff
@@ -297,13 +297,50 @@
...
+
+  import MDAnalysis
```
orbeckst: We customarily use
```python
import MDAnalysis as mda
```
(or at least that's what we should be doing consistently)

package/MDAnalysis/analysis/waterdynamics.py
```diff
@@ -1231,22 +1266,78 @@ def print(self, verbose, *args):
         elif verbose:
             print(args)
 
-    def run(self, tau_max=20, start=0, stop=None, step=1, verbose=False):
+    def _correct_intermittency(self, intermittency, selected_ids):
         """
-        Computes and returns the survival probability timeseries
+        Pre-process Consecutive Intermittency
+        :param intermittency: the max gap allowed and to be corrected
```
orbeckst: Use numpy/Napoleon formatting, see https://github.com/MDAnalysis/mdanalysis/wiki/WritingDocumentation#numpy-style

package/MDAnalysis/analysis/waterdynamics.py
```diff
@@ -1231,22 +1266,78 @@ def print(self, verbose, *args):
...
+        :param selected_ids: modifies the selecteded IDs in place by adding atoms which left for <= :param intermittency
+        """
+        self.print('Correcting the selected IDs for intermittancy (gaps). ')
```
orbeckst: Was `self.print()` already in the class?

(We should be consistently using logging... but I think that's another PR for this module.)

package/MDAnalysis/analysis/waterdynamics.py
```diff
@@ -1231,22 +1266,78 @@ def print(self, verbose, *args):
         elif verbose:
             print(args)
 
-    def run(self, tau_max=20, start=0, stop=None, step=1, verbose=False):
+    def _correct_intermittency(self, intermittency, selected_ids):
         """
-        Computes and returns the survival probability timeseries
+        Pre-process Consecutive Intermittency
```
orbeckst: Please explain what this function does. I have no idea what "intermittency" means here.

package/MDAnalysis/analysis/waterdynamics.py
```diff
@@ -1231,22 +1266,78 @@ def print(self, verbose, *args):
...
+        :param intermittency: the max gap allowed and to be corrected
+        :param selected_ids: modifies the selecteded IDs in place by adding atoms which left for <= :param intermittency
+        """
+        self.print('Correcting the selected IDs for intermittancy (gaps). ')
+        if intermittency == 0:
+            return
+
+        # If the atom is absent for a number of frames equal or smaller
```
orbeckst: Stuff like this should be in actual docs.

orbeckst: Or more precisely: the comments here are very good, but the docstring should tell a user what the function does (what problem does it solve). The comments should explain the how – the ideas behind the algorithm.

package/MDAnalysis/analysis/waterdynamics.py
```diff
@@ -1231,22 +1266,78 @@ def print(self, verbose, *args):
...
             Zero-based index of the first frame to be analysed
-        stop : int
+        stop : int, optional
             Zero-based index of the last frame to be analysed (inclusive)
-        step : int
-            Jump every `step`'th frame
-        tau_max : int
+        step : int, optional
+            Jump every `step`'th frame. This is compatible but independant of the taus used, and it is good to consider
+            using the  step as tau_max to remove the overlap. Note that step and taus are consistent with intermittency.
```
orbeckst: put all references parameters in single backticks such as
```reST
Jump every `step` -th frame. This is compatible but independant of the taus used, and it is good to consider
using the  `step` as `tau_max` to remove the overlap.
```
Note that you need a space after the backticks.

Build the docs locally and check that your docs look ok, see https://github.com/MDAnalysis/mdanalysis/wiki/WritingDocumentation#generating-the-docs

package/MDAnalysis/analysis/waterdynamics.py
```diff
@@ -1231,22 +1266,78 @@ def print(self, verbose, *args):
...
+        tau_max : int, optional
             Survival probability is calculated for the range :math:`1 <= \tau <= tau_max`
-        verbose : Boolean
-            Overwrite the constructor's verbosity
+        residues : Boolean, optional
+            If true, the analysis will be carried out on the residues (.resids) rather than on atom (.ids).
+            A single atom is sufficient to classify the residue as within the distance.
+        intermittency : int, optional
```
orbeckst: Good explanation.

orbeckst:
> We generalised the code to be applicable to other situations which is why we would like to ask if we should move it out of waterdynamics.py to its own file (survivalprobability.py in analysis).

I like the idea of moving it into its own module. How about doing this as a separate PR once this one is merged? The smaller a PR the better.

orbeckst:
I broke travis #2233 so once PR #2234 is merged we need to again merge develop into this branch, sorry.

package/MDAnalysis/analysis/waterdynamics.py
bieniekmateusz: Corrected.

package/MDAnalysis/analysis/waterdynamics.py
bieniekmateusz: Noted. I'll have a read later.

package/MDAnalysis/analysis/waterdynamics.py
bieniekmateusz: Yes, we've created self.print() in the previous PR. I'll look into logging for the next PR. Thanks

package/MDAnalysis/analysis/waterdynamics.py
bieniekmateusz: Good point. I've moved the actual documentation to the top and introduced minor corrections in the comments.

package/MDAnalysis/analysis/waterdynamics.py
bieniekmateusz: Addressed: moved the comment to the documentation since it explains it well.

bieniekmateusz:
The requested changes have been introduced. I've built the documentation and it appears fine. Cheers

package/MDAnalysis/analysis/waterdynamics.py
```diff
@@ -1273,15 +1273,42 @@ def run(self, tau_max=20, start=0, stop=None, step=1, verbose=False):
...
-            selected_ids.append(set(self.universe.select_atoms(self.selection).ids))
+        if step <= tau_max + 1:
+            # load all frames
+            for ts in self.universe.trajectory[start:stop]:
+                self.print(verbose, "Loading frame:", ts)
+                selected_ids.append(set(self.universe.select_atoms(self.selection).ids))
+        else:
+            # skip frames that will not be used
+            # Example: step 5 and tau 2: L, L, L, S, S, L, L, L, S, S, ... with L = Load, and S = Skip
+            loaded_counter = 0
```
alejob: Why would you like to do this? Maybe I'm missing something, but when you calculate the Survival Probability, the definition says that you should consider only the particles that REMAIN in the selection between time t and t+tau, ie, you should load all the frames between t and t+tau and count in t, then count in t+1, count in t+2..., count in t+tau, counting only the particles that were in the selection in the previous step, so I don't understand why would you like to skip some frames when step is larger than the tau_max + 1. Maybe an other example could help to clarify this.

alejob:
> > We generalised the code to be applicable to other situations which is why we would like to ask if we should move it out of waterdynamics.py to its own file (survivalprobability.py in analysis).
> 
> I like the idea of moving it into its own module. How about doing this as a separate PR once this one is merged? The smaller a PR the better.

It sounds good, but how could we avoid to broke the old code? There is people that has already implemented scripts with waterdynamics.survivalprobability(), so the idea is to avoid broke that code. Maybe some kind of link inside the MDA code.

package/MDAnalysis/analysis/waterdynamics.py
bieniekmateusz: Thanks @alejob 

It is always the case that we load all the frames between t and t+tau. The step foremost allows to remove the overlap of the measurement. I.e. a step larger than tau_max may be required to ensure that subsequent t0s are not correlated. To rephrase, if an atom survives from t to t+tau, then for the next 20 frames, the data for tau=1 will have 20 data points where the atoms survives.

Then there is a performance regarding sampling. If someone would like to scan through a long simulation, skipping frames is a huge performance gain. This data can be also used for discussion for equilibration and stability of the system.

Mat and @p-j-smith

bieniekmateusz:
So moving it to another file would be a new PR in the future. 

We could say that the code is moved and simply point the user to import the new path - if someone tries to use it. Luckily the function itself and how it works has not changed so it should be a minor inconvinience for the users, hopefully.

orbeckst:
Breaking user code is not something that we do lightly. 

In this case it will be easy: we just import the function into water dynamics and everything else will just be able to use it as before. The call signature did not change (?) and all the old tests for the function still pass, don't they?

In any case, that's all stuff we can discuss in a new issue/PR. 

--
Oliver Beckstein
email: orbeckst@gmail.com

orbeckst:
Merged develop to get the travis fix from PR #2234

package/MDAnalysis/analysis/waterdynamics.py
bieniekmateusz: Corrected.

bieniekmateusz:
I think I've corrected everything, please let me know if I missed anything.

package/MDAnalysis/analysis/waterdynamics.py
alejob: Great, now I understand, and this look pretty good. Thanks for your improvement.

bieniekmateusz:
To give you heads up, after this is merged we will shortly send you another PR with a common autocorrelation function between the waterdynamics.HydrogenBondLifetimes and waterdynamics.SurvivalProbability. 

@richardjgowers noted that it would be good to formalise and extract the common autocorrelation code (PR #1995). After careful considerations with @p-j-smith we now have a separate autocorrelation function that has intermittency built into it, and being independent, is easily testable. The test cases will be well-defined in a similar manner to the ones we've used for the SurvivalProbability. We've already verified that waterdynamics.HydrogenBondLifetimes behaves exactly the same way our autocorrelation does (continuous). These changes should to a big extent simplify the module. In the case of HydrogenBondLifetimes it would mean a simple function call replacing the entire class. This will help simplifying the documentation too (PR #2181)

Work done with @p-j-smith.

package/CHANGELOG
```diff
@@ -15,7 +15,7 @@ The rules for this file:
 ------------------------------------------------------------------------------
 mm/dd/yy micaela-matta, xiki-tempula, zemanj, mattwthompson, orbeckst, aliehlen,
          dpadula85, jbarnoud, manuel.nuno.melo, richardjgowers, mattwthompson,
-         ayushsuhane, picocentauri, NinadBhat
+         ayushsuhane, picocentauri, NinadBhat, bieniekmateusz, p-j-smith
```
orbeckst: Are @bieniekmateusz and @p-j-smith in AUTHORS?

orbeckst:
@alejob if you're happy with the PR could you please also add a formal "approve" review? Thanks.

package/CHANGELOG
bieniekmateusz: We, both of us. We're doing pair-programming.

package/MDAnalysis/analysis/waterdynamics.py
```diff
@@ -1231,27 +1265,89 @@ def print(self, verbose, *args):
...
+            If true, the analysis will be carried out on the residues (.resids) rather than on atom (.ids).
+            A single atom is sufficient to classify the residue as within the distance.
+        intermittency : int, optional
+            The maximum number of consecutive frames for which an atom can leave but be counted as present if it returns
+            at the next frame. An intermittency of `0` is equivalent to a continuous survival probability, which does
+            not allow for the leaving and returning of atoms. For example, for `intermittency=2`, any given atom may
+            leave a region of interest for up to two consecutive frames yet be treated as being present at all frames.
+            The default is continuous (0).
+        verbose : Boolean, optional
+            Print the progress to the console
```
alejob: I like this, it gives great flexibility to SP.

orbeckst:
Thanks @alejob !

I resolved the conflict in the CHANGELOG and we'll just let CI confirm that everything is still good.

orbeckst:
Congratulations @bieniekmateusz and @p-j-smith , merged! Thanks for your hard work (and patience).

Thanks @alejob for reviewing.

PR #1862, opened 2018-04-10
Add Chemfiles as a coordinate reader/writer

Luthaf:
Here is a first working version of the integration of [chemfiles](https://github.com/chemfiles/chemfiles/) as a coordinate reader. See #1194 for some initial discussion on this. There are still a few TODO in the code, but I think this is ready for an initial round of review.

~Currently I get three test failures: two related to aux, and one to container format. I don't really understand what is getting tested here, could someone give me a little pointer?~

Using chemfiles might help with:
  - #864: time dependent topology is supported in chemfiles, it could be brought over to MDAnalysis
  - #865: TNG format is implemented
  - #988: chemfiles handles this by sending a warning if too many atoms are in the frame for PDB format, and not outputting the corresponding bonds
  - #1133: VMD style PDB are supported
  - #1966: I am pretty sure this is working, I'll need to double check it 
  - #1753: supported in chemfiles though the `is_hetatm` atomic property
     - [x] TODO: translate recordtype to chemfiles property
  - #2216: already supported for all text formats, for .gz, .xz and .bz2

Unresolved questions:
-----------------------

Chemfiles supports reading/writing files with a non constant number of atoms. Currently I error when reading if the number of atoms changes, and I do nothing and write the new number of atoms when writing. Is this the desired behaviour?

PR Checklist
------------
 - [x] Tests?
    - [x] Formats with no extension association on chemfiles side (LAMMPS Data)
 - [x] Docs?
 - [ ] CHANGELOG updated?
 - [ ] Issue raised/referenced?

Issue #1194, opened 2017-01-27
make bare-bones readers MDAnalysis-independent


orbeckst:
**In the long run** it would be useful to move forward with making trajectory readers MDAnalysis-independent. @kain88-de  has already done this for the Gromacs XDR reader. What I mean by this is that someone else could, in principle, use `lib.formats` and then use it independently from MDA itself.

Why?

Because between MDAnalysis, mdtraj, pytraj, VMD … we are duplicating code and that's a waste of time and bad software engineering. It would be much better if (eventually) we (the community) have bare-bones readers and then everyone else can use them as part of their library. Generally, people liked that idea, judging from my interactions with others at meetings. Certainly, something like this is not happening overnight but we could take a leadership role and start by moving some of our readers/writers into `lib.format`. We would then be ready to move forward if we wanted to.

A [GSOC project](https://github.com/MDAnalysis/mdanalysis/wiki/Google-Summer-Of-Code) related to trajectory formats could be to

1. formally define the api (basically, numpy arrays or array_like)… follow the Gromacs XDR code
2. move some/all readers/writers to lib.format 

See also https://groups.google.com/d/msg/mdnalysis-devel/QlwoL1_Etns/cqAZRrp0FwAJ

dotsdl:
I'm in favor of this. If we can at least get to the point where our Readers (and perhaps eventually Writers) are clearly decoupled from the rest of the library then we can make eventual splitting out easy if it's of interest to other packages. MDAnalysis then can remain as a library focused on a nice user interface as it always has.

dotsdl:
Can help to reduce the maintenance burden of an always-increasing number of Readers on the MDAnalysis devs alone, too.

kain88-de:
Btw this means that we have to be careful with contributions to these code-pieces now. If we want to have a separate library it can't have a copy-left license like the GPL, a BSD license would be more appropriate. We can only do this if we keep close tabs on the contributions.

Actually what we should do then is after the 0.16. release is to create a new python project that starts with the xtc format and give it a liberal license like BSD. It would be a new dependency of MDAnalysis but that would be the case eventually anyway.

orbeckst:
Regarding licensing: the library could certainly have a GPL – that's a choice. It is true that this could potentially decrease adoption but consider that, for instance, mdtraj is LGPL. 

Until we have a spirited discussion about this I agree with you: we need to keep good track of commiters that we could ask for a relicence if necessary.

[...]

Issue #1030, opened 2016-10-14
transfer_to_memory should accept start,stop,step parameters


kain88-de:
### Expected behaviour

I should be able to do `u.transfer_to_memory(start=10, stop=100, step=2)`. That would be nice because the iteration kwargs then are consistent with the kwargs in python and our analysis classes.

E.g. `u.transfer_to_memory(start=4, step=4, stop=40)` would only load 9 frames from the original trajectory.

### Actual behaviour

This results in a Argument Error since the only accepted kwarg is `frame_interval`.

pratyushprakash:
Hi, can I take this up?

kain88-de:
Yes you can submit a PR for this.

utkbansal:
I'd like to work on this. Just to be clear, in `u.transfer_to_memory(start=4, step=4, stop=40)` I just have to add support for `start` and `stop` parameters. Right?

richardjgowers:
@utkbansal rename `frame_interval` to `step` (it's not been released yet, so we don't have to deprecate etc) and add `start` and `stop`.  Make sure that the defaults are all `None` (including `frame_interval`) else backwards iteration etc won't work.  Then do a `git grep 'frame_interval'` and change references to step

utkbansal:
@richardjgowers Thanks! I'm working on it. 😄

PR #2227, opened 2019-03-27
[WIP] Cythonizes GROParser & GROReader very minimally

Fenil3510:
 -  Cythonized version of GRO class is called when parsing ascii file. Not sure if this is the best way to do it. Please suggest changes.

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


orbeckst:
In [your email](https://groups.google.com/d/msg/mdnalysis-discussion/YFXD_ZLzIVg/a_2TrqyGAwAJ) you said
> I have made changes in the GROParser and created a cythonized version(very basic changes). It gives an improvement of 0.05 seconds on average on a 2mb gro file. This is not much given that cythonization is not extensively done. 

The code does not seem to run on the CI (AppVeyor fails) but taking for a moment that you got your changes to build locally, it is interesting that the simplest approach of just wrapping the code in cython does not change much. 

Do you know where the Python code spends most of its time, i.e., what is the performance bottleneck? Did you profile the Python GRO parser (e.g. with [line_profiler](https://github.com/rkern/line_profiler)) to find out? This will give you a better idea of what needs optimizing.

(Also, eventually the PR needs to build on the CI and we need to figure out if we can move the cython code to the lib module - might be tricky for parsers because code in lib is supposed to be independent of MDAnalysis things such as topologies and AtomGroups. But first explore and see if there are some obvious ways to speed up the parser.)

orbeckst:
@Fenil3510 , I remembered that a while ago PR #1862 was started with the goal to leverage the [chemfiles python interface](http://chemfiles.org/chemfiles.py/0.8.0/) for some trajectory formats. They don't support all formats but a [a fair number](http://chemfiles.org/chemfiles/latest/formats.html#list-of-supported-formats), at least as coordinate "Readers/Writers". Not sure if they can be used as topology "parsers", too. 

chemfiles contains a [GRO reader/writer](http://chemfiles.org/chemfiles/latest/devdoc/format.html?highlight=gro#_CPPv4N9chemfiles3GROE) (written in C/C++ but apparently based on VMD's [molfile plugin](http://www.ks.uiuc.edu/Research/vmd/plugins/molfile/)). I don't know how seamlessly chemfiles can interact with MDAnalysis.

We didn't have this in our project description, partly because the PR stalled, partly because it is complicated enough to understand MDAnalysis and we don't necessarily want you to having to spend time learning another library that we ourselves know little about. However, I feel that this should be mentioned so that you have a better idea of what else is going on in MDAnalysis, as potentially related to your project.

If you or anyone else has opinions, please discuss them here.

cc @richardjgowers @jbarnoud

Fenil3510:
> e CI and we need to figure out if we can move the cython code to the lib module

Thanks for the suggestions @orbeckst. Yes, I just wrapped the code in Cython. I'll see why tests are failing. I'll look into line_profiler to find out what's to be done.

richardjgowers:
@orbeckst I think the problem with adopting C++/C for reading ascii is it then becomes very hard (for python devs) to patch.  Currently our stuff bends over backwards to read everything because we can patch it.

@Fenil3510 I think this might be a good place to get ideas for how to do quick ascii i/o:

http://wesmckinney.com/blog/speeding-up-pandass-file-parsers-with-cython/

Then obviously finding the corresponding cython code in the pandas repo (if they still use Cython).

Luthaf:
Hey, chemfiles dev here! So PR #1862 have stalled a bit, but I just finished the release of version 0.9 of chemfiles so I'll try to back to it. It was basically working, just needed a bit of documentation. It might take a bit of time as I am currently writing my thesis ...

> Not sure if they can be used as topology "parsers", too.

It should be easy to convert from chemfiles Topology class to MDAnalysis version. Current version of #1862 only uses the coordinates and throw away the topology, but I could update it to include chemfiles as a topology reader. 

> chemfiles contains a GRO reader/writer (written in C/C++ but apparently based on VMD's molfile plugin). I don't know how seamlessly chemfiles can interact with MDAnalysis.

We actually have our own implementation of GRO format, not based on VMD molfiles. It lives here: https://github.com/chemfiles/chemfiles/blob/master/src/formats/GRO.cpp.

> I think the problem with adopting C++/C for reading ascii is it then becomes very hard (for python devs) to patch. Currently our stuff bends over backwards to read everything because we can patch it.

We try to write readable and easy to modify C++ code in chemfiles, and modern C++ is closer to Python than old school C. But there is always a barrier to overcome when contributing to a project in a different language.

package/MDAnalysis/topology/GROParser.py
```diff
@@ -63,97 +63,4 @@
...
-            Segids(np.array(['SYSTEM'], dtype=object))
-        ]
-
-        top = Topology(n_atoms=n_atoms, n_res=len(new_resids), n_seg=1,
-                       attrs=attrs,
-                       atom_resindex=residx,
-                       residue_segindex=None)
-
-        return top
+from .c_GROParser import GROParser
```
tylerjereddy: You need to build the C extension into the library via a `setup.py` file so that this doesn't cause issues during the build, which is also visible in CI.

For examples of MDA compiled extension building -- `git grep 'MDAExtension'`. I don't think we normally commit the Cython generated code to our repo. Probably best to just remove that file from version control & set up proper building of the extension.

tylerjereddy:
>  finding the corresponding cython code in the pandas repo

+1 to at least check this; `pandas` goes to great lengths to optimize some if its file loaders and trying to optimize text file reading can be a pretty deep rabbit hole.

package/MDAnalysis/topology/GROParser.py
Fenil3510: Oh, I did build it with setup.py but didn't know how to commit it right. Probably that might have caused the build error. The `.so` files are ignored anyway. So ones can you clarify what files to include after `setup.py`. I'm building with `python setup.py build_ext --inplace`.

Fenil3510: Ok, I get it. So I need to make an `MDAExtension` object for it in the `setup.py` file and cythonize list it there itself, right?

tylerjereddy: Yup, CI is all green now, so I'll mark this resolved

Fenil3510:
@tylerjereddy @orbeckst. I profiled the code of GROParser line by line and checked what takes the most time. It was obviously the looping part, so I thought to make the for loop cythonized. I read [here](https://cython-docs2.readthedocs.io/en/latest/src/tutorial/numpy.html) that cythons access to numpy array is much faster for indexing than the regular array. I made the changes accordingly, but I get an improvement of only 40ms on average, ie. actual average time reduces from 710 to 640ms. I still have to dig into `pandas` parsers as suggested. Any other obvious cythonization I'm missing?

package/MDAnalysis/topology/c_GROParser.pyx
```diff
@@ -86,6 +87,12 @@ class GROParser(TopologyReaderBase):
         # resid, resname, name, index, (x,y,z)
         cdef int n_atoms;
         cdef int i;
+        cdef str line
+        cdef int count = 0
+        cdef np.ndarray[np.int32_t] resids
```
Fenil3510: Defining arrays with numpy  C interface is causing the build to fail probably, though the build is successful locally. Any ideas how to deal with this?

package/MDAnalysis/topology/c_GROParser.pyx
```diff
@@ -0,0 +1,168 @@
...
+        # Gro has the following columns
+        # resid, resname, name, index, (x,y,z)
+        cdef int n_atoms;
+        cdef int i;
+        cdef str line
+        cdef int count = 0
+        cdef np.ndarray[np.int32_t] resids
+        cdef np.ndarray[object] resnames
+        cdef np.ndarray[object] names
+        cdef np.ndarray[np.int32_t] indices
```
tylerjereddy: better to use [typed memoryviews](https://cython.readthedocs.io/en/latest/src/userguide/memoryviews.html) here vs. the old buffer syntax

package/MDAnalysis/topology/c_GROParser.pyx
```diff
@@ -0,0 +1,168 @@
...
+        # Fix wrapping of resids (if we ever saw a wrap)
+        if np.any(resids == 0):
+            # find places where resid hit zero again
+            wraps = np.where(resids == 0)[0]
+            # group these places together:
+            # find indices of first 0 in each block of zeroes
+            # 1) find large changes in index, (ie non sequential blocks)
+            diff = np.diff(wraps) != 1
+            # 2) make array of where 0-blocks start
+            starts = np.hstack([wraps[0], wraps[1:][diff]])
```
tylerjereddy: `starts` is initialized here without prior type declaration

Fenil3510:
@tylerjereddy numpy functions `np.any` and `np.where` don't seem to work with cython typed memoryviews. Would have to write custom function if they are to be used.

tylerjereddy:
You can use constructs like `np.asarray(memoryview)` where needed, but if there's an opportunity for creative speedup with custom loops you may want to look into it. Memoryviews usually can work without the GIL unlike the buffer interface, so it may require some careful design to really take advantage.

tylerjereddy:
NumPy also has a C API

Fenil3510:
@tylerjereddy , I see. I will change it to `memoryview` and see if the custom function has any improvement. I was focussing more on the for loop `for line in inf` and that's where most of the time goes (according to line_profiler). So I was thinking of
1. Faster indexing of numpy array (which I expected to speed up after using buffer interface, but didn't)
2. Custom type conversion function(no python interaction) to replace `int(line[0:5])` but it ended being rather costly
3. Can string slicing be improved in any way?

I'd be grateful for any suggestions.

Fenil3510:
The current changes in the `GROParser` ie. `parse` function to be specific. Profiling the `parse` function separately gives an improvement of 2.5 - 3x on any file (files tested 2MB to 17 MB). But the overall reading from `mda.Universe()` shows a negligible gain. This is because parsing takes 6% of the total time. So this is surely not the bottleneck. I chose `GROParser` for getting started. I will be looking into `GROReader` for some time and probably find out the bottleneck
Any suggestions are welcomed.

Fenil3510:
@orbeckst as you said
> Also, eventually the PR needs to build on the CI and we need to figure out if we can move the cython code to the lib module - might be tricky for parsers because code in lib is supposed to be independent of MDAnalysis things such as topologies and AtomGroups.
I was wondering what is the reason to move cython code to `lib` module? Sorry, if the question is very basic.

orbeckst:
When you look at [`lib.formats`](https://github.com/MDAnalysis/mdanalysis/tree/develop/package/MDAnalysis/lib/formats) you see that these _readers_ are written in a way independent from core MDAnalysis code (i.e., no universes, AtomGroups, Topology, etc). In #1194 we aimed for making more of our *readers* low-level and MDAnalysis-agnostic. When I said

> we need to figure out if we can move the cython code to the lib module - might be tricky for parsers because code in lib is supposed to be independent of MDAnalysis things such as topologies and AtomGroups.

I was expressing the desire to do the same thing for _parsers_ but also expressed that I kind of doubted that this could be done easily.

In short: Moving GROParser to `lib.formats` is not high priority. I was interested in knowing if it was even a possibility.

package/MDAnalysis/topology/GROParser.py
orbeckst: I am not an expert in cython extensions so I have to ask this as a comment here: Why keep a stub Python file instead of naming the cython file just GROParser.pyx and building it as `topology.GROParser`?

cc @tylerjereddy @richardjgowers

Fenil3510:
@orbeckst thankyou for the information. Yes, the intermediate Python file can be removed I believe, I had kept it to play safe, now that I've familiarized with Cython modules, looks like it shouldn't cause an issue, I'll make the changes soon.

Fenil3510:
Had some tests failing in `test_nuclinfo.py`, so had to merge develop into this.

Fenil3510:
Hmm, test fails due to a numpy merge [here](https://github.com/numpy/numpy/pull/13359/files). Some changes in pickling.

package/MDAnalysis/topology/GROParser.pyx
```diff
@@ -47,6 +47,7 @@
 from __future__ import absolute_import
 
 import numpy as np
+cimport numpy as np
```
richardjgowers: this is only needed when you have numpy on the left hand side of things, eg `cdef np.ndarray[float] arr`.  You're using memoryviews here (which is better), so you just need the Python import to get numpy functions imported.

package/MDAnalysis/topology/GROParser.pyx
```diff
@@ -84,6 +85,16 @@ def parse(self, **kwargs):
...
         # resid, resname, name, index, (x,y,z)
+        cdef int n_atoms;
+        cdef int i;
+        cdef str line
+        cdef int count = 0
+        cdef int[:] resids
+        cdef object[:] resnames
+        cdef object[:] names
+        cdef int[:] indices
+        cdef long long[:] starts
```
richardjgowers: pretty sure we just need ints here too

richardjgowers:
@Fenil3510 don't worry about those failures.  What are loading times like with these changes?

package/MDAnalysis/topology/GROParser.pyx
Fenil3510: Yes, my bad. Forgot to remove after I shifted to memoryviews.

package/MDAnalysis/topology/GROParser.pyx
Fenil3510: Travis was failing with `int` and expecting it to be `long long` as I remember. I'll double check.

Fenil3510: `np.where` returns a np.int64 but mac integer is 32-bit as the same issure [here](https://stackoverflow.com/questions/32262976/cython-buffer-type-mismatch-expected-int-but-got-long), I've type-casted the arrays inplace with `np.int32` and works with `int` now.

orbeckst:
@Fenil3510 could you please benchmark the parser on its own and full universe creation with the GRO files in the vesicle data set https://www.mdanalysis.org/MDAnalysisData/vesicles.html ?
```
conda install -c conda-forge MDAnalysisData
```
and 
```python
from MDAnalyisData.datasets import fetch_vesicle_lib

dataset = fetch_vesicle_lib()

grofiles = dataset.structure    # should be 3 GRO files
```

These are fairly big systems (see `dataset.labels` and the web page for description). It would ne nice to have a reproducible benchmark that we can also report, e.g., in a blog post or paper.

Fenil3510:
@orbeckst sorry for the delay, have my exams running parallelly. 
Here are the results and benchmarks
1.  This is only the parser (ie. parse function to be specific), shows gains of 2.5x to 3.5x
benchmark_1

2. Gains in making the whole universe are less for smallest file but upto 1.5x on biggest file. This is expected though.
image

Working on cythonizing the readers will definitely help increase gains.

orbeckst:
@Fenil3510 many thanks, very useful and encouraging! 

Good luck with your exams!

Fenil3510:
> It would be nice to have a reproducible benchmark that we can also report, e.g., in a blog post or paper.
I will start working on this after May 9th.

orbeckst:
I merged current develop, hopefully the tests will all pass then.

Fenil3510:
I've been busy lately, I'll get back on this pretty soon.

orbeckst:
Not sure why the old numpy test (**NUMPY_VERSION=1.10.4**) fails, [Job #1030.5](https://travis-ci.com/MDAnalysis/mdanalysis/jobs/205747332)

```
==================================== ERRORS ====================================
_________________ ERROR collecting MDAnalysisTests/test_api.py _________________
ImportError while importing test module '/home/travis/build/MDAnalysis/mdanalysis/testsuite/MDAnalysisTests/test_api.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
testsuite/MDAnalysisTests/test_api.py:29: in <module>
    import MDAnalysis as mda
package/MDAnalysis/__init__.py:184: in <module>
    from .lib import log
package/MDAnalysis/lib/__init__.py:42: in <module>
    from . import pkdtree
package/MDAnalysis/lib/pkdtree.py:35: in <module>
    from scipy.spatial import cKDTree
../../../miniconda/envs/test/lib/python3.5/site-packages/scipy/spatial/__init__.py:99: in <module>
    from .kdtree import *
../../../miniconda/envs/test/lib/python3.5/site-packages/scipy/spatial/kdtree.py:8: in <module>
    import scipy.sparse
../../../miniconda/envs/test/lib/python3.5/site-packages/scipy/sparse/__init__.py:230: in <module>
    from .csr import *
../../../miniconda/envs/test/lib/python3.5/site-packages/scipy/sparse/csr.py:13: in <module>
    from ._sparsetools import (csr_tocsc, csr_tobsr, csr_count_blocks,
E   ImportError: numpy.core.multiarray failed to import
------------------------------- Captured stderr --------------------------------
RuntimeError: module compiled against API version 0xb but this version of numpy is 0xa
```

I am restarting the job... maybe it's a conda glitch???

Fenil3510:
Looks like an issue with Travis. All other PR's are failing the same test.

Fenil3510:
I'd need some help here. I used types in GROReader for some variables and converted it into Cython file(This method worked with GROParser). But too many tests are failing for this commit, while I can build the universe with `MDAnalysis.datasets` with this commit and gains 3-4x gains but tests are failing as I can't access position method of `Atom` class which is weird. I'm unable to trace the exact reason for this, any help would be much appreciated.

richardjgowers:
Hey Fenil   
  

  
Thanks for sticking with this. I think it might be the reader isn’t getting registered when the module imports... can you import mdanalysis and check mda._READERS (or whatever that dict is called....)
  

  
  

  
  
>

package/MDAnalysis/coordinates/GRO.pyx
```diff
@@ -116,6 +116,10 @@
 from ..core import flags
 from ..exceptions import NoDataError
 from ..lib import util
+cimport cython
+@cython.boundscheck(False)  # Deactivate bounds checking
```
richardjgowers: these decorators aren't doing what you are expecting here, they won't apply to the whole file.  I think they're probably just modifying the Timestep object.

You probably need to decorate the class method directly

richardjgowers:
@Fenil3510 the reader class wasn't getting added to `mda._READERS` which is a dict of all the reader classes which is usually done automatically by the class it inherits from... Not sure why, but I've manually added it for now in `coordinates.__init__`.  I did a rebase and force pushed to your branch, but it's "working" now, in that the code is ran and throws an error :)

Fenil3510:
Yeah, it's failing silently even though there is an except to raise `KeyError` if it's not in the registry. Looking into it.

richardjgowers:
hey @Fenil3510 

I played around with writing a Cython XYZReader function here:

https://gist.github.com/richardjgowers/dbec308c364aae8b9ca84e8e22171823

There's no error checking (this would need to be added!) but it seems to go pretty fast if you want some ideas.

Issue #2198, opened 2019-02-05
Custom analysis function for water bridge analysis


xiki-tempula:
In PR #2087, I have introduced custom analysis functions to customise the behaviour of  count_by_time, count_by_type, which hasn't been merged yet. Following the discussion in #2177, it is suggested that it is better to embed the heavy atom information when detecting hydrogen bonds to avoid further complication #1687.

I have changed the internal representation of water bridge analysis to the format suggested in #2177, which now has the form (selection1_index, selection1_heavy_atom_index, selection2_index, selection2_heavy_atom_index, distance, angle). However, this representation creates a little problem to the implementation of Custom analysis function for (e.g.) count_by_type. I wonder if I can have some suggestions as to how to proceed.

Suppose we want to tabulate the water bridges between selection 1 and selection 2 **at residue level**, **distinguishing different orders of water bridge** and **only selecting the water bridges originating from residue id 233**. We will write the analysis function in this format:

```python
def analysis(current, output):
    # decompose the first hydrogen bond.
    s1_index, to_index, (s1_resname, s1_resid, s1_name),
                                    (to_resname, to_resid, to_name), dist, angle = current[0]
    # decompose the last hydrogen bond.
    from_index, s2_index, (from_resname, from_resid, from_name),
                                         (s2_resname, s2_resid, s2_name), dist, angle = current[-1]
    # if the residue id is 233
    if s1_resid == 233:
        # compute order of water bridge
        order = len(current) - 1
        key = (s1_resname, s1_resid, s2_resname, s2_resid, order)
        # The number of this type of water bridge is 1.
        output[key] = 1
w.count_by_type(analysis_func=analysis)
```

The input format is the build in a way similar to timeseries and has the problem of being WET and doesn't include heavy atom information. I'm thinking of replacing the input with the new DRY format (selection1_index, selection1_heavy_atom_index, selection2_index, selection2_heavy_atom_index, distance, angle). However, to achieve the same goal as the previous analysis function, we will need to include universe in the input paramaters and the analysis function might become:

```python
def analysis(current, output, u):
    # decompose the first hydrogen bond.
    (selection1_index, selection1_heavy_atom_index, 
     to_index, to_index, 
     distance, angle) = current[0]
    (from_index, from_heavy_atom_index, 
     selection2_index, selection2_heavy_atom_index, 
     distance, angle) = current[－1]
    # if the residue id is 233
    if u.atoms[selection1_index].resid == 233:
[...]

Issue #2177, opened 2019-01-16
The format of timerseries for hydrogen bond and water bridge analysis


xiki-tempula:
When one is interested in understanding the hydrogen bond from a hydrogen bond donor to a hydrogen bond acceptor, one might want to know the name of the heavy atom corresponding to the hydrogen bond donor. The current practice implemented in hydrogen bond analysis is that when a hydrogen bond is detected only the donor and acceptor were recorded and the heavy atom corresponding the donor is added later, which cause problems such as #1687.

@orbeckst  has suggested that a workaround will be incorporating the heavy atom information directly into the timeseries object so as to save the trouble afterword. I'm also in favour of this option. I have a PR open which adds multi-order water bridge support to the existing water bridge analysis which will, rather unfortunately, break the existing time series attribute of the water bridge analysis.

So the question is how should we reformulate the time series attribute. Currently, the format has this form.

```
[<atom1 index>, <atom2 index>, (atom1 resname, atom1 resid, atom1 name), (atom2 resname, atom2 resid, atom2 name), <distance>, <angle>]
```

In hydrogen bond analysis, atom 1 is the donor and atom 2 is the receptor. In water bridge analysis, since two donors can be joined by a water molecule. The atom1 is the atoms more closely linked to selection 1 and atom2 is the atom more closely linked to selection 2. So if hydrogen bond acceptor α from selection 1 and hydrogen bond donor atom β from selection B is joined by atom γ and δ from a water molecule. The time series will have this form.

```
[<α index>, <γ index>, (α resname, α resid, α name), (γ resname, γ resid, γ name), <distance>, <angle>],
[<δ index>, <β index>, (δ resname, δ resid, δ name), (β resname, β resid, β name), <distance>, <angle>],
```

My suggestion of how we incorporate heavy atom information will be adding the heavy atom name to the last column of the tuple. So that the time series will have the form:

```
[<α index>, <γ index>, (α resname, α resid, α name), (γ resname, γ resid, γ name, **γ heavy atom name**), <distance>, <angle>],
[<δ index>, <β index>, (δ resname, δ resid, δ name), (β resname, β resid, β name, **β heavy atom name**), <distance>, <angle>],
```

For hydrogen-bond analysis, instead of taking the form of where we have donor precede acceptor.

```
[<donor index>, <acceptor index>, (donor resname, donor resid, donor name), (acceptor resname, acceptor resid, acceptor name), <distance>, <angle>]
```

[...]

PR #2087, opened 2018-10-03
Move the water bridge analysis to the new analysis class

xiki-tempula:
Move the water bridge analysis to the new analysis class, add high order water bridge support. 
Rewritte the count_by_type and count_by_time function, add custom anaylsis function support.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?
 - [ ] Issue raised/referenced?

Issue #954, opened 2016-08-23
PCA raises warning with default options


kain88-de:
### Expected behaviour

When I use the default options for `PCA` no warning should be triggered. 

The problem is that inexperienced users will be put off by a warning or think that we messed up the MDAnalysis and it can't be trusted. We should never throw a warning using the defaults. Instead remove the warning and add a note in the docs.
### Actual behaviour

We warn the user about performance issues when the default options are used.
### Code to reproduce the behaviour

``` python
import MDAnalysis as mda
from MDAnalysis.analysis.pca import PCA

u = mda.Universe(top, trj)
pca = PCA(u).run()
....

```
### Currently version of MDAnalysis:

(run `python -c "import MDAnalysis as mda; print(mda.__version__)"`)
0.15.1-dev0

PR #2237, opened 2019-04-05
Hbond analysis

p-j-smith:
A newly rewritten HydrogenBondAnalysis class.
-
Hi, I have written a new version of the current HydrogenBondAnalysis class that simplifies the class itself and especially the run() method. Currently, this is the only method from the original class that is implemented. I decided to write a new class after reading some of the issues surrounding the current hbonds module (#2177  #2181  #2198 ).

I realise this is an extremely popular analysis tool and so understand if there is no desire for a complete rewrite of the class. However, after reading @orbeckst comments in #2087 (regarding duplication of code across the hbonds module) and #2177 (regarding rewriting the module from scratch and returning DRY, tidy data), I thought I would like to at least create a discussion around the issues raised.

Another motivation for rewriting this class is that @bieniekmateusz and I have written a general autocorrelation function (mentioned in PR #2226 ) that can be used to replace the three different implementations currently in waterdynamics.HydrogenBondLifetimes, waterdynamics.SurvivalProbability and hbonds.hbond_autocorrelation, as suggested @richardjgowers in PR #1995 . 

 - 
* Newly rewritten HydrogenBondAnalysis class
* The code has been simplified significantly
* The run() method has been optimised by using the MDAnalysis.lib.distances.capped_distance() method. 
* The capped_distance() method is used to both find D-H pairs and to find D-A pairs within a cutoff.
* This results in a speed-up of 5 times for the small test WaterDCD system, and a speedup of around 55 times for a system containing around 7,000 water molecules.
* Added a regression test for the current HydrogenBondAnalysis class to directly compare behaviour with the one submitted here.
* Basic documentation of the new class.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
+++ b/package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
@@ -1,0 +1,559 @@
+# -*- Mode: python; tab-width: 4; indent-tabs-mode:nil; coding:utf-8 -*-
+# vim: tabstop=4 expandtab shiftwidth=4 softtabstop=4
+#
+# MDAnalysis --- https://www.mdanalysis.org
+# Copyright (c) 2006-2017 The MDAnalysis Development Team and contributors
+# (see the file AUTHORS for the full list of names)
+#
+# Released under the GNU Public Licence, v2 or any higher version
+#
+# Please cite your use of MDAnalysis in published work:
+#
+# R. J. Gowers, M. Linke, J. Barnoud, T. J. E. Reddy, M. N. Melo, S. L. Seyler,
+# D. L. Dotson, J. Domanski, S. Buchoux, I. M. Kenney, and O. Beckstein.
+# MDAnalysis: A Python package for the rapid analysis of molecular dynamics
+# simulations. In S. Benthall and S. Rostrup editors, Proceedings of the 15th
+# Python in Science Conference, pages 102-109, Austin, TX, 2016. SciPy.
+# doi: 10.25080/majora-629e541a-00e
+#
+# N. Michaud-Agrawal, E. J. Denning, T. B. Woolf, and O. Beckstein.
+# MDAnalysis: A Toolkit for the Analysis of Molecular Dynamics Simulations.
+# J. Comput. Chem. 32 (2011), 2319--2327, doi:10.1002/jcc.21787
+#
+
+"""Hydrogen Bond Analysis --- :mod:`MDAnalysis.analysis.hydrogenbonds.hbond_analysis`
+=====================================================================================
+
+:Author: Paul Smith
+:Year: 2019
+:Copyright: GNU Public License v3
+
+This module provides methods to find and analyse hydrogen bonds in a Universe.
+
+The :class:`HydrogenBondAnalysis` class is a new version of the original
+:class:`MDAnalysis.analysis.hbonds.HydrogenBondAnalysis` class from the module
+:mod:`MDAnalysis.analysis.hbonds.hbond_analysis`, which itself was modeled after the `VMD
+HBONDS plugin`_.
+
+.. _`VMD HBONDS plugin`: http://www.ks.uiuc.edu/Research/vmd/plugins/hbonds/
+
+
+Input
+------
+
+Required:
+  - *universe* : an MDAnalysis Universe object
+
+Options:
+  - *donors_sel* [None] : Atom selection for donors. If `None`, then will be identified via the topology.
+  - *hydrogens_sel* [None] : Atom selection for hydrogens. If `None`, then will be identified via charge and mass.
+  - *acceptors_sel* [None] : Atom selection for acceptors. If `None`, then will be identified via charge.
+  - *d_h_cutoff* (Å) [1.2] : Distance cutoff used for finding donor-hydrogen pairs
+  - *d_a_cutoff* (Å) [3.0] : Distance cutoff for hydrogen bonds. This cutoff refers to the D-A distance.
+  - *d_h_a_angle_cutoff* (degrees) [150] : D-H-A angle cutoff for hydrogen bonds.
+  - *update_selections* [True] : If true, will update atom selections at each frame.
+
+
+Output
+------
+
+  - *frame* : frame at which a hydrogen bond was found
+  - *donor id* : atom id of the hydrogen bond donor atom
+  - *hydrogen id* : atom id of the hydrogen bond hydrogen atom
+  - *acceptor id* : atom id of the hydrogen bond acceptor atom
+  - *distance* (Å): length of the hydrogen bond
+  - *angle* (degrees): angle of the hydrogen bond
+
+Hydrogen bond data are returned in a :class:`numpy.ndarray` on a "one line, one observation" basis
+and can be accessed via :attr:`HydrogenBondAnalysis.hbonds`::
+
+    results = [
+        [
+            <frame>,
+            <donor index (0-based)>,
+            <hydrogen index (0-based)>,
+            <acceptor index (0-based)>,
+            <distance>,
+            <angle>
+        ],
+        ...
+    ]
+
+Example use of :class:`HydrogenBondAnalysis`
+--------------------------------------------
+
+The simplest use case is to allow :class:`HydrogenBondAnalysis` to guess the acceptor and hydrogen atoms, and to
+identify donor-hydrogen pairs via the bonding information in the topology::
+
+  import MDAnalysis
+  from MDAnalysis.analysis.hydrogenbonds.hbond_analysis import HydrogenBondAnalysis as HBA
+
+  u = MDAnalysis.Universe(psf, trajectory)
+
+  hbonds = HBA(universe=u)
+  hbonds.run()
+
+It is also possible to specify which hydrogens and acceptors to use in the analysis. For example, to find all hydrogen
+bonds in water::
+
+  import MDAnalysis
+  from MDAnalysis.analysis.hydrogenbonds.hbond_analysis import HydrogenBondAnalysis as HBA
+
+  u = MDAnalysis.Universe(psf, trajectory)
+
+  hbonds = HBA(universe=u, hydrogens_sel='resname TIP3 and name H1 H2', acceptors_sel='resname TIP3 and name OH2')
+  hbonds.run()
+
+Alternatively, :attr:`hydrogens_sel` and :attr:`acceptors_sel` may be generated via the :attr:`guess_hydrogens` and
+:attr:`guess_acceptors`. This selection strings may then be modified prior to calling :attr:`run`, or a subset of
+the universe may be used to guess the atoms. For example, find hydrogens and acceptors belonging to a protein::
+
+  import MDAnalysis
+  from MDAnalysis.analysis.hydrogenbonds.hbond_analysis import HydrogenBondAnalysis as HBA
+
+  u = MDAnalysis.Universe(psf, trajectory)
+
+  hbonds = HBA(universe=u)
+  hbonds.hydrogens_sel = hbonds.guess_hydrogens("protein")
+  hbonds.acceptors_sel = hbonds.guess_acceptors("protein")
+  hbonds.run()
+
+Slightly more complex selection strings are also possible. For example, to find hydrogen bonds involving a protein and
+any water molecules within 10 Å of the protein (which may be useful for subsequently finding the lifetime of
+protein-water hydrogen bonds or finding water-bridging hydrogen bond paths)::
+
+  import MDAnalysis
+  from MDAnalysis.analysis.hydrogenbonds.hbond_analysis import HydrogenBondAnalysis as HBA
+
+  u = MDAnalysis.Universe(psf, trajectory)
+
+  hbonds = HBA(universe=u)
+
+  protein_hydrogens_sel = hbonds.guess_hydrogens("protein")
+  protein_acceptors_sel = hbonds.guess_acceptors("protein")
+
+  water_hydrogens_sel = "resname TIP3 and name H1 H2"
+  water_acceptors_sel = "resname TIP3 and name OH2"
+
+  hbonds.hydrogens_sel = f"({protein_hydrogens_sel}) or ({water_hydrogens_sel} and around 10 not resname TIP3})"
+  hbonds.acceptors_sel = f"({protein_acceptors_sel}) or ({water_acceptors_sel} and around 10 not resname TIP3})"
+  hbonds.run()
+
+It is highly recommended that a topology with bonding information is used to generate the universe, e.g `PSF`, `TPR`, or
+`PRMTOP` files. This is the only method by which it can be guaranteed that donor-hydrogen pairs are correctly identified.
+However, if, for example, a `PDB` file is used instead, a :attr:`donors_sel` may be provided along with a
+:attr:`hydrogens_sel` and the donor-hydrogen pairs will be identified via a distance cutoff, :attr:`d_h_cutoff`::
+
+  import MDAnalysis
+  from MDAnalysis.analysis.hydrogenbonds.hbond_analysis import HydrogenBondAnalysis as HBA
+
+  u = MDAnalysis.Universe(pdb, trajectory)
+
+  hbonds = HBA(
+    universe=u,
+    donors_sel='resname TIP3 and name OH2',
+    hydrogens_sel='resname TIP3 and name H1 H2',
+    acceptors_sel='resname TIP3 and name OH2',
+    d_h_cutoff=1.2
+  )
+  hbonds.run()
+
+The class and its methods
+-------------------------
+
+.. autoclass:: HydrogenBondAnalysis
+   :members:
+
+"""
+from __future__ import absolute_import, division
+
+import numpy as  np
+
+from .. import base
+from MDAnalysis.lib.distances import capped_distance, calc_angles
+
+from ...due import due, Doi
+
+due.cite(Doi("10.1039/C9CP01532A"),
+         description="Hydrogen bond analysis implementation",
+         path="MDAnalysis.analysis.hydrogenbonds.hbond_analysis",
+         cite_module=True)
+
+del Doi
+
+
+class HydrogenBondAnalysis(base.AnalysisBase):
+    """
+    Perform an analysis of hydrogen bonds in a Universe.
+    """
+
+    def __init__(self, universe, donors_sel=None, hydrogens_sel=None, acceptors_sel=None,
+                 d_h_cutoff=1.2, d_a_cutoff=3.0, d_h_a_angle_cutoff=150, update_selections=True):
+        """Set up atom selections and geometric criteria for finding hydrogen bonds in a Universe.
+
+        Parameters
+        ----------
+        universe : Universe
+            MDAnalysis Universe object
+        donors_sel : str
+            Selection string for the hydrogen bond donor atoms. If the universe topology contains bonding information,
+            leave :attr:`donors_sel` as `None` so that donor-hydrogen pairs can be correctly identified.
+        hydrogens_sel :  str
+            Selection string for the hydrogen bond hydrogen atoms. Leave as `None` to guess which hydrogens to use in
+            the analysis using :attr:`guess_hydrogens`. If :attr:`hydrogens_sel` is left as `None`, also leave
+            :attr:`donors_sel` as None so that donor-hydrogen pairs can be correctly identified.
+        acceptors_sel : str
+            Selection string for the hydrogen bond acceptor atoms. Leave as `None` to guess which atoms to use in the
+            analysis using :attr:`guess_acceptors`
+        d_h_cutoff : float (optional)
+            Distance cutoff used for finding donor-hydrogen pairs [1.2]. Only used to find donor-hydrogen pairs if the
+            universe topology does not contain bonding information
+        d_a_cutoff : float (optional)
+            Distance cutoff for hydrogen bonds. This cutoff refers to the D-A distance. [3.0]
+        d_h_a_angle_cutoff: float (optional)
+            D-H-A angle cutoff for hydrogen bonds, in degrees. [150]
+        update_selections: bool (optional)
+            Whether or not to update the acceptor, donor and hydrogen lists at each frame. [True]
+
+        Note
+        ----
+
+        It is highly recommended that a universe topology with bonding information is used, as this is the only way
+        that guarantees the correct identification of donor-hydrogen pairs.
+        """
+
+        self.u = universe
+        self._trajectory = self.u.trajectory
+        self.donors_sel = donors_sel
+        self.hydrogens_sel = hydrogens_sel
+        self.acceptors_sel = acceptors_sel
+        self.d_h_cutoff = d_h_cutoff
+        self.d_a_cutoff = d_a_cutoff
+        self.d_h_a_angle = d_h_a_angle_cutoff
+        self.update_selections = update_selections
+
+    def guess_hydrogens(self, selection='all', max_mass=1.1, min_charge=0.3):
+        """Guesses which hydrogen atoms should be used in the analysis.
+
+        Parameters
+        ----------
+        selection: str (optional)
+            Selection string for atom group from which hydrogens will be identified.
+        max_mass: float (optional)
+            Maximum allowed mass of a hydrogen atom.
+        min_charge: float (optional)
+            Minimum allowed charge of a hydrogen atom.
+
+        Returns
+        -------
+        potential_hydrogens: str
+            String containing the :attr:`resname` and :attr:`name` of all hydrogen atoms potentially capable of forming
+            hydrogen bonds.
+
+        Notes
+        -----
+        This function makes use of atomic masses and atomic charges to identify which atoms are hydrogen atoms that are
+        capable of participating in hydrogen bonding. If an atom has a mass less than :attr:`max_mass` and an atomic
+        charge greater than :attr:`min_charge` then it is considered capable of participating in hydrogen bonds.
+
+        If :attr:`hydrogens_sel` is `None`, this function is called to guess the selection.
+
+        Alternatively, this function may be used to quickly generate a :class:`str` of potential hydrogen atoms involved
+        in hydrogen bonding. This str may then be modified before being used to set the attribute
+        :attr:`hydrogens_sel`.
+        """
+
+        ag = self.u.select_atoms(selection)
+        hydrogens_ag = ag[
+            np.logical_and(
+                ag.masses < max_mass,
+                ag.charges > min_charge
+            )
+        ]
+
+        hydrogens_list = np.unique(
+            [
+                '(resname {} and name {})'.format(r, p) for r, p in zip(hydrogens_ag.resnames, hydrogens_ag.names)
+            ]
+        )
+
+        return " or ".join(hydrogens_list)
+
+    def guess_donors(self, selection='all', max_charge=-0.5):
+        """Guesses which atoms could be considered donors in the analysis. Only use if the universe topology does not
+        contain bonding information, otherwise donor-hydrogen pairs may be incorrectly assigned.
+
+        Parameters
+        ----------
+        selection: str (optional)
+            Selection string for atom group from which donors will be identified.
+        max_charge: float (optional)
+            Maximum allowed charge of a donor atom.
+
+        Returns
+        -------
+        potential_donors: str
+            String containing the :attr:`resname` and :attr:`name` of all atoms that potentially capable of forming
+            hydrogen bonds.
+
+        Notes
+        -----
+        This function makes use of and atomic charges to identify which atoms could be considered donor atoms in the
+        hydrogen bond analysis. If an atom has an atomic charge less than :attr:`max_charge`, and it is within
+        :attr:`d_h_cutoff` of a hydrogen atom, then it is considered capable of participating in hydrogen bonds.
+
+        If :attr:`donors_sel` is `None`, and the universe topology does not have bonding information, this function is
+        called to guess the selection.
+
+        Alternatively, this function may be used to quickly generate a :class:`str` of potential donor atoms involved
+        in hydrogen bonding. This :class:`str` may then be modified before being used to set the attribute
+        :attr:`donors_sel`.
+
+        """
+
+        # We need to know `hydrogens_sel` before we can find donors
+        # Use a new variable `hydrogens_sel` so that we do not set `self.hydrogens_sel` if it is currently `None`
+        if not self.hydrogens_sel:
+            hydrogens_sel = self.guess_hydrogens()
+        else:
+            hydrogens_sel = self.hydrogens_sel
+        hydrogens_ag = self.u.select_atoms(hydrogens_sel)
+
+        ag = hydrogens_ag.residues.atoms.select_atoms(
+            "({donors_sel}) and around {d_h_cutoff} {hydrogens_sel}".format(
+                donors_sel=selection,
+                d_h_cutoff=self.d_h_cutoff,
+                hydrogens_sel=hydrogens_sel
+            )
+        )
+        donors_ag = ag[ag.charges < max_charge]
+        donors_list = np.unique(
+            [
+                '(resname {} and name {})'.format(r, p) for r, p in zip(donors_ag.resnames, donors_ag.names)
+            ]
+        )
+
+        return " or ".join(donors_list)
+
+    def guess_acceptors(self, selection='all', max_charge=-0.5):
+        """Guesses which atoms could be considered acceptors in the analysis.
+
+        Parameters
+        ----------
+        selection: str (optional)
+            Selection string for atom group from which acceptors will be identified.
+        max_charge: float (optional)
+            Maximum allowed charge of an acceptor atom.
+
+        Returns
+        -------
+        potential_acceptors: str
+            String containing the :attr:`resname` and :attr:`name` of all atoms that potentially capable of forming
+            hydrogen bonds.
+
+        Notes
+        -----
+        This function makes use of and atomic charges to identify which atoms could be considered acceptor atoms in the
+        hydrogen bond analysis. If an atom has an atomic charge less than :attr:`max_charge` then it is considered
+        capable of participating in hydrogen bonds.
+
+        If :attr:`acceptors_sel` is `None`, this function is called to guess the selection.
+
+        Alternatively, this function may be used to quickly generate a :class:`str` of potential acceptor atoms involved
+        in hydrogen bonding. This :class:`str` may then be modified before being used to set the attribute
+        :attr:`acceptors_sel`.
+        """
+
+        ag = self.u.select_atoms(selection)
+        acceptors_ag = ag[ag.charges < max_charge]
+        acceptors_list = np.unique(
+            [
+                '(resname {} and name {})'.format(r, p) for r, p in zip(acceptors_ag.resnames, acceptors_ag.names)
+            ]
+        )
+
+        return " or ".join(acceptors_list)
+
+    def _get_dh_pairs(self):
+        """Finds donor-hydrogen pairs.
+
+        Returns
+        -------
+        donors, hydrogens: AtomGroup, AtomGroup
+            AtomGroups corresponding to all donors and all hydrogens. AtomGroups are ordered such that, if zipped, will
+            produce a list of donor-hydrogen pairs.
+        """
+
+        # If donors_sel is not provided, use topology to find d-h pairs
+        if not self.donors_sel:
+
+            if len(self.u.bonds) == 0:
+                raise Exception('Cannot assign donor-hydrogen pairs via topology as no bonded information is present. '
+                                'Please either: load a topology file with bonded information; use the guess_bonds() '
+                                'topology guesser; or set HydrogenBondAnalysis.donors_sel so that a distance cutoff '
+                                'can be used.')
+
+            hydrogens = self.u.select_atoms(self.hydrogens_sel)
+            donors = sum(h.bonded_atoms[0] for h in hydrogens)
+
+        # Otherwise, use d_h_cutoff as a cutoff distance
+        else:
+
+            hydrogens = self.u.select_atoms(self.hydrogens_sel)
+            donors = self.u.select_atoms(self.donors_sel)
+            donors_indices, hydrogen_indices = capped_distance(
+                donors.positions,
+                hydrogens.positions,
+                max_cutoff=self.d_h_cutoff,
+                box=self.u.dimensions,
+                return_distances=False
+            ).T
+
+            donors = donors[donors_indices]
+            hydrogens = hydrogens[hydrogen_indices]
+
+        return donors, hydrogens
+
+    def _prepare(self):
+
+        self.hbonds = [[], [], [], [], [], []]
+        self.frames = np.arange(self.start, self.stop, self.step)
+        self.timesteps = (self.frames * self.u.trajectory.dt) + self.u.trajectory[0].time
+
+        # Set atom selections if they have not been provided
+        if not self.acceptors_sel:
+            self.acceptors_sel = self.guess_acceptors()
+        if not self.hydrogens_sel:
+            self.hydrogens_sel = self.guess_hydrogens()
+
+        # Select atom groups
+        self._acceptors = self.u.select_atoms(self.acceptors_sel, updating=self.update_selections)
+        self._donors, self._hydrogens = self._get_dh_pairs()
+
+    def _single_frame(self):
+
+        box = self._ts.dimensions
+
+        # Update donor-hydrogen pairs if necessary
+        if self.update_selections:
+            self._donors, self._hydrogens = self._get_dh_pairs()
+
+        # find D and A within cutoff distance of one another
+        # min_cutoff = 1.0 as an atom cannot form a hydrogen bond with itself
+        d_a_indices, d_a_distances = capped_distance(
+            self._donors.positions,
+            self._acceptors.positions,
+            max_cutoff=self.d_a_cutoff,
+            min_cutoff=1.0,
+            box=box,
+            return_distances=True,
+        )
+
+        # Remove D-A pairs more than d_a_cutoff away from one another
+        tmp_donors = self._donors[d_a_indices.T[0]]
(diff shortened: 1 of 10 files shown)


orbeckst:
Make the linter happy ([Job #954.8](https://travis-ci.com/MDAnalysis/mdanalysis/jobs/190898450)):
```
************* Module MDAnalysis.analysis.hbond_analysis
package/MDAnalysis/analysis/hbond_analysis.py:79: [W1618(no-absolute-import), ] import missing `from __future__ import absolute_import`
```

orbeckst:
This looks interesting – I only had a very cursory glance, but (1) code clean-up and (2) performance improvements and (3) extensibility are all very good reasons.

There is a question how this will interact with @xiki-tempula 's water-bridging analysis; the issue text already references the pertinent issues/PRs so you're clearly aware of what's been done/discussed so far. I think we need to have a discussion how we want to move forward with the H-bonding module. Having actual code to discuss is great.

I'll add a few more initial comments on the code.

P.S.: @p-j-smith if you add your committer email address to GitHub then the commits will be properly attributed to you.

xiki-tempula:
The code looks really nice, the capped distance is a very neat way of finding the atoms compared with the old neighbour search. I will try to use capped_distance in the water bridge we well. I guess the only problem is that the interface and the outputs are very different from the previous one.

package/MDAnalysis/analysis/hbond_analysis.py
```diff
@@ -0,0 +1,199 @@
+# -*- Mode: python; tab-width: 4; indent-tabs-mode:nil; coding:utf-8 -*-
```
orbeckst: Did you intentionally _not_ put it in `analysis/hbonds`? 

We need to decide how to move forward with H-bonds, see #2238.

package/MDAnalysis/analysis/hbond_analysis.py
```diff
@@ -0,0 +1,199 @@
...
+   :members:
+
+   .. attribute:: hbonds
+
+"""
+
+import numpy as  np
+
+from . import base
+from MDAnalysis.lib.distances import capped_distance
```
orbeckst: you can import both on the same line

package/MDAnalysis/analysis/hbond_analysis.py
```diff
@@ -0,0 +1,199 @@
...
+from MDAnalysis.lib.distances import calc_angles
+
+class HydrogenBondAnalysis(base.AnalysisBase):
```
orbeckst: Hooray – analysis API!

package/MDAnalysis/analysis/hbond_analysis.py
```diff
@@ -0,0 +1,199 @@
...
+    """
+    Perform an analysis of hydrogen bonds in a Universe.
+    """
+
+    def __init__(self, universe, donors_sel, hydrogens_sel, acceptors_sel, d_h_cutoff, d_a_cutoff, d_h_a_angle_cutoff):
+        """
+        Set up atom selections and geometric criteria for finding hydrogen bonds in a Universe.
+
+        :param universe : Universe
```
orbeckst: Will need to follow numpy doc style

package/MDAnalysis/analysis/hbond_analysis.py
```diff
@@ -0,0 +1,199 @@
...
+from . import base
+from MDAnalysis.lib.distances import capped_distance
+from MDAnalysis.lib.distances import calc_angles
+
+class HydrogenBondAnalysis(base.AnalysisBase):
+    """
+    Perform an analysis of hydrogen bonds in a Universe.
+    """
+
+    def __init__(self, universe, donors_sel, hydrogens_sel, acceptors_sel, d_h_cutoff, d_a_cutoff, d_h_a_angle_cutoff):
```
orbeckst: Can the call signature be made the same as for the original class?

package/MDAnalysis/analysis/hbond_analysis.py
```diff
@@ -0,0 +1,199 @@
...
+        self.u = universe
+        self.donors_sel = donors_sel
+        self.hydrogens_sel = hydrogens_sel
+        self.acceptors_sel = acceptors_sel
+        self.d_h_cutoff = d_h_cutoff
+        self.d_a_cutoff = d_a_cutoff
+        self.d_h_a_angle = d_h_a_angle_cutoff
+
+
+    def run(self, start=None, stop=None, step=None):
```
orbeckst: Instead of putting all logic into `run()` you should put create the `_single_frame()` method that works on a single frame. This will make it easier to, for instance, parallelize the code.

package/MDAnalysis/analysis/hbond_analysis.py
```diff
@@ -0,0 +1,199 @@
...
+            Use every `step` frame between `start` (included) and `stop` (excluded). [``None``]
+
+        """
+
+        self.hbonds = [[], [], [], [], [], []]
+        for ts in self.u.trajectory[start:stop:step]:
+
+            box = ts.dimensions
+
+            # Find D-H pairs
```
orbeckst: I very much like that your algorithm does not use loops.

Can you explain in how far your algorithm differs from the one in the current HydrogenBondAnalysis?

package/MDAnalysis/analysis/hbond_analysis.py
```diff
@@ -0,0 +1,199 @@
...
+            all_donors = self.u.select_atoms(self.donors_sel)
```
orbeckst: For static selections you are doing the selection every step. It's probably a small cost compared to the other calculations. (The original HydrogenBondAnalysis class can create updating selections if needed.)

package/MDAnalysis/analysis/hbond_analysis.py
```diff
@@ -0,0 +1,199 @@
...
+            all_hydrogens = self.u.select_atoms(self.hydrogens_sel)
+            donors_indices, hydrogen_indices = capped_distance(
```
orbeckst: This is similar to the distance-based finding of H in the original class but if available we should use topology information.

orbeckst: Furthermore, this is kind of expensive to do every step. For static selections (say, all water molecules...) we should not really do this every step.

package/MDAnalysis/analysis/hbond_analysis.py
```diff
@@ -0,0 +1,199 @@
...
+        self.d_a_cutoff = d_a_cutoff
+        self.d_h_a_angle = d_h_a_angle_cutoff
+
+
+    def run(self, start=None, stop=None, step=None):
+        """
+        Find hydrogen bonds present over the interval defined by start:stop:step.
+
+        Stores the results as
+        :attr:`HydrogenBondAnalysis.hbonds`
```
orbeckst: Define the data format.

testsuite/MDAnalysisTests/analysis/test_hbond_analysis.py
```diff
@@ -0,0 +1,72 @@
...
+
+    @pytest.fixture()
+    def h(self, universe):
+        h = MDAnalysis.analysis.hbond_analysis.HydrogenBondAnalysis(universe, **self.kwargs)
+        h.run()
+        return h
+
+    def test_hbond_analysis(self, h):
+
+        h = h
```
orbeckst: That's weird, does pytest make you do this? You should be able to omit this line.

testsuite/MDAnalysisTests/analysis/test_hbonds.py
```diff
@@ -397,3 +397,81 @@ def test_table_atoms(self, h, normalized_timeseries, reference_table):
...
+    }
+
+    # hackish way to allow looping over self.reference and generating tests
+    _functions = {
+        'mean': np.mean,
+        'std': np.std,
+    }
+
+    def test_timeseries(self, h, normalized_timeseries):
+        h = h
```
orbeckst: ???

testsuite/MDAnalysisTests/analysis/test_hbonds.py
```diff
@@ -397,3 +397,81 @@ def test_table_atoms(self, h, normalized_timeseries, reference_table):
...
+        'selection1': 'all',
+        'selection2': 'all',
+        'detect_hydrogens': "distance",
+        'distance': 3.0,
+        'angle': 120.0,
+        'distance_type': 'heavy',
+        'pbc': True
+    }
+
+    @pytest.fixture()
```
orbeckst: add `scope="class"` so that it only gets run once – works fine if data are only read.

testsuite/MDAnalysisTests/analysis/test_hbonds.py
```diff
@@ -397,3 +397,81 @@ def test_table_atoms(self, h, normalized_timeseries, reference_table):
         # https://github.com/MDAnalysis/mdanalysis/issues/801)
         for name, ref in reference_table.items():
             assert_array_equal(h.table.field(name), ref, err_msg="resname for {0} do not match (Issue #801)")
+
+
+class TestHydrogenBondAnalysisTIP3PHeavyPBC(object):
+    @staticmethod
+    @pytest.fixture()
+    def universe():
```
orbeckst: add `scope="class"` so that it only gets run once – works fine if data are only read.

testsuite/MDAnalysisTests/analysis/test_hbonds.py
```diff
@@ -397,3 +397,81 @@ def test_table_atoms(self, h, normalized_timeseries, reference_table):
...
+                      for item in hframe]
+        return timeseries
+
+    # keys are the names in the h.table
+    reference = {
+        'distance': {'mean': 2.7627309, 'std': 0.0905052},
+        'angle': {'mean': 158.9038039, 'std': 12.0362826},
+    }
+
+    @pytest.fixture()
```
orbeckst: add scope="class" so that it only gets run once – works fine if data are only read.

orbeckst:
More general discussion in #2238 .

testsuite/MDAnalysisTests/analysis/test_hbonds.py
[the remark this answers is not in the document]
p-j-smith: I based this test on the tests used for the current version of the class and I assumed this line was there for a reason, but it's not needed. I've removed this line and generally tidied up the code for the tests.

package/MDAnalysis/analysis/hbond_analysis.py
p-j-smith: For now, just to avoid any confusion (on my part) and to avoid having to think of a new name for the module. But if we do decide to go ahead with this new version then at some point it should probably be moved into analysis/hbonds.

package/MDAnalysis/analysis/hbond_analysis.py
p-j-smith: I've fixed this.

package/MDAnalysis/analysis/hbond_analysis.py
p-j-smith: And I've now figured out how to use it!

package/MDAnalysis/analysis/hbond_analysis.py
p-j-smith: I've changed to doc style.

package/MDAnalysis/analysis/hbond_analysis.py
p-j-smith: I've fixed this.

package/MDAnalysis/analysis/hbond_analysis.py
p-j-smith: That's a good point, I just based this code on the default behaviour of the current class to update selections, but can introduce static selections.

package/MDAnalysis/analysis/hbond_analysis.py
p-j-smith: Yeah, I totally agree. Using the topology by default is a nice idea.

package/MDAnalysis/analysis/hbond_analysis.py
p-j-smith: I've added this.

testsuite/MDAnalysisTests/analysis/test_hbond_analysis.py
p-j-smith: It is weird. I've deleted it.

testsuite/MDAnalysisTests/analysis/test_hbonds.py
p-j-smith: I'm not sure...

testsuite/MDAnalysisTests/analysis/test_hbonds.py
p-j-smith: I've added this

testsuite/MDAnalysisTests/analysis/test_hbonds.py
p-j-smith: I've added this.

testsuite/MDAnalysisTests/analysis/test_hbonds.py
p-j-smith: I've added this.

package/MDAnalysis/analysis/hbond_analysis.py
p-j-smith: So currently the new analysis uses the same algorithm for finding hydrogen bonds as when current class is instantiated with `detect_hydrogens='distance'`, `distance_type='heavy`, `pbc=True`. The test cases I've added show that when using these parameters the results of the two classes are identical.

package/MDAnalysis/analysis/hbond_analysis.py
p-j-smith: It can, but to be honest I find the call signature of the current class very confusing. When I first tried using current analysis tool I was put off by the large number of selections I had to provide and especially by the idea of having to create a new class if I wanted different donor or acceptor types. I actually preferred the idea of writing my own hydrogen bond analysis script, which forms the basis of the new class.

I know there are now `donor` and `acceptor` keywords that can be used to add atom names not already included in the default selections, but there is no `hydrogens` keyword, meaning that a new class must be created if someone is using a different forcefield or if they are using LAMMPS (which has numbered atom types).

I find using only `donors_sel`, `acceptors_sel` and `hydrogens_sel` much cleaner and easier to understand than the various combinations provided by `selection1`, `selection2`, `selection1_type`, `forcefield`, `donors` and `acceptors`. The other issue with the current default selections is that they only take into account protein atom names and not lipid or CGenFF names. 

I'm not sure how you feel about this, but if rewriting the class involves changing how the results are stored/returned then it might also be a good idea to simplify the signature call.

package/MDAnalysis/analysis/hbond_analysis.py
```diff
@@ -0,0 +1,204 @@
...
+            all_donors.positions,
+            all_hydrogens.positions,
+            max_cutoff=self.d_h_cutoff,
+            box=box,
+            return_distances=False
+        ).T
+
+        # Select donors, hydrogens and acceptors
+        donors = all_donors[donors_indices]
+        hydrogens = all_hydrogens[hydrogen_indices]
```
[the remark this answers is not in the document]
p-j-smith: What would be the best way to find hydrogens via the topology? It's not something I'm very familiar with.

package/MDAnalysis/analysis/hbond_analysis.py
```diff
@@ -0,0 +1,204 @@
...
+class from the module mod:`MDAnalysis.analysis.hbonds.hbond_analysis`, which itself was modeled after the `VMD
+HBONDS plugin`_.
+
+.. _`VMD HBONDS plugin`: http://www.ks.uiuc.edu/Research/vmd/plugins/hbonds/
+
+
+Options:
+  - *d_h_cutoff* (Å) [1.2] : Distance cutoff used for finding donor-hydrogen pairs
+  - *d_a_cutoff* (Å) [3.0] : Distance cutoff for hydrogen bonds. This cutoff refers to the D-A distance.
+  - *d_h_a_angle_cutoff* (degrees) [150] : D-H-A angle cutoff for hydrogen bonds.
```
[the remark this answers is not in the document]
p-j-smith: That's a nice idea.

richardjgowers:
https://github.com/MDAnalysis/mdanalysis/blob/develop/package/MDAnalysis/analysis/hbonds/hbond_autocorrel.py#L229
  

  
  

  
  
>

package/MDAnalysis/analysis/hbond_analysis.py
```diff
@@ -0,0 +1,204 @@
...
+        # Find D-H-A angles greater than d_h_a_angle_cutoff
+        d_h_a_angles = np.rad2deg(
+            calc_angles(
+                donors.positions,
+                hydrogens.positions,
+                acceptors.positions,
+                box=box
+            )
+        )
+        hbond_indices = np.where(d_h_a_angles > self.d_h_a_angle)[0]
```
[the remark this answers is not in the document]
p-j-smith: Most probably, I'll change this.

package/MDAnalysis/analysis/hbond_analysis.py
```diff
@@ -0,0 +1,204 @@
...
+    def __init__(self, universe, donors_sel, hydrogens_sel, acceptors_sel, d_h_cutoff, d_a_cutoff, d_h_a_angle_cutoff):
+        """Set up atom selections and geometric criteria for finding hydrogen bonds in a Universe.
+
+        Parameters
+        ----------
+        universe : Universe
+            MDAnalysis Universe object
+        donors_sel : str
+            Selection string for the hydrogen bond donor atoms
+        hydrogens_sel :  str
```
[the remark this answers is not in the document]
p-j-smith: Yeah that makes sense.

package/MDAnalysis/analysis/hbond_analysis.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def __init__(self, universe, donors_sel, hydrogens_sel, acceptors_sel, d_h_cutoff, d_a_cutoff, d_h_a_angle_cutoff):
+        """Set up atom selections and geometric criteria for finding hydrogen bonds in a Universe.
+
+        Parameters
+        ----------
+        universe : Universe
+            MDAnalysis Universe object
+        donors_sel : str
+            Selection string for the hydrogen bond donor atoms
```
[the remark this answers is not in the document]
p-j-smith: Providing atom groups would definitely be nice, it's something to think about.

package/MDAnalysis/analysis/hbond_analysis.py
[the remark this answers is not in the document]
p-j-smith: I was also thinking that if donors_sel is not provided then it might be nice to implement a `guess_donors()` method that will return a list of potential donor atom names (and their hydrogens) based on the partial charges of the atoms. A similar `guess_acceptors()` method could also be implemented. Or, someone could call these methods before run() to help in quickly creating these lists if they have a large number of donors and acceptors in their system.

p-j-smith:
Here's a little bit more information on the new class in response to @orbeckst questions above:

> I initially would like to understand (with a view towards #2238 )
> 
> * How does your algorithm differ from the original one?
> * Does it produce the same results?
> * What are the major differences between your implementation and the new one?
>   
>   * What are _new_ capabilities?
>   * What is _missing_ compared to the old HydrogenBondAnalysis?
> * How much of the original HydrogenBondAnalysis API can be rebuild in this refactored class?

* How does your algorithm differ from the original one?
In general it is quite similar, but as I mentioned above:
   

    >So currently the new analysis uses the same algorithm for finding hydrogen bonds as when the current class is instantiated with `detect_hydrogens='distance'`, `distance_type='heavy`, `pbc=True`. The test cases I've added show that when using these parameters the results of the two classes are identical. 

* Does it produce the same results?
If the current class is used with the parameters I mentioned in the point above, it does indeed produce the same results.

* What are the major differences between your implementation and the new one?
The main differences are that the for loops have been removed and capped_distances() is used to find potential donor-acceptor pairs.

 * What are _new_ capabilities?
The major improvements are in terms of speed, maintainability and extensibility. On a small test system of around 7000 water molecules it is around 55 times faster than the current code, and I expect that the performance boost will increase with system size. The code is much simpler and easier to read, so both maintaining it and add new features will likely be easier. For instance, the general autocorrelation function that @bieniekmateusz and I have written could be called inside a `hbonds.autocorrelation()` method, which will remove duplicate code.

* What is _missing_ compared to the old HydrogenBondAnalysis?
As it is, the current class does not has all of the convenience functions that the old class has, nor does it have a method to generate a pandas dataframe. It also doesn't have a list of default donors and acceptors, but as I mentioned above I think this is not necessarily a bad thing:

    > It can, but to be honest I find the call signature of the current class very confusing. When I first tried using current analysis tool I was put off by the large number of selections I had to provide and especially by the idea of having to create a new class if I wanted different donor or acceptor types. I actually preferred the idea of writing my own hydrogen bond analysis script, which forms the basis of the new class.
    > 
    > I know there are now `donor` and `acceptor` keywords that can be used to add atom names not already included in the default selections, but there is no `hydrogens` keyword, meaning that a new class must be created if someone is using a different forcefield or if they are using LAMMPS (which has numbered atom types).
    > 
    > I find using only `donors_sel`, `acceptors_sel` and `hydrogens_sel` much cleaner and easier to understand than the various combinations provided by `selection1`, `selection2`, `selection1_type`, `forcefield`, `donors` and `acceptors`. The other issue with the current default selections is that they only take into account protein atom names and not lipid or CGenFF names.
    > 
    > I'm not sure how you feel about this, but if rewriting the class involves changing how the results are stored/returned then it might also be a good idea to simplify the signature call.

To avoid inconveniencing current users that rely on the default acceptors and donors, I though we could add methods to guess these:

> I was also thinking that if donors_sel is not provided then it might be nice to implement a guess_donors() method that will return a list of potential donor atom names (and their hydrogens) based on the partial charges of the atoms. A similar guess_acceptors() method could also be implemented. Or, someone could call these methods before run() to help in quickly creating these lists if they have a large number of donors and acceptors in their system

If these changes are too drastic, then I'm sure it is possible to change the signature call to match the old class, but I haven't thought about how do it yet. Also, I have added a `count_by_time()` method and think it should be relatively straight forward to add the other convenience methods.

package/MDAnalysis/analysis/hbond_analysis.py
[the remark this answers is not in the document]
p-j-smith: I just checked this in the test WaterPSF system and, if selecting all donors in the system, it is 10 times faster to select both groups and then use the `capped_distance()` method. However, if the donors and hydrogens selections are more complicated (such as involving `and around 5 resid 7`) then the distance based method is around 40% faster than using the topology. So as the complexity of the selection increases it may become faster to use the topology to find hydrogens, but generally it is faster using `capped_distances()`.

package/MDAnalysis/analysis/hbond_analysis.py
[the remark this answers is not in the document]
bieniekmateusz: This could be neatly solved with Universe.selectAtoms(AG) returning AG. This way all interfaces would automatically accept str and AG selections.

package/MDAnalysis/analysis/hbond_analysis.py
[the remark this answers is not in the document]
bieniekmateusz: Such guess could be based on the mass of the atoms?

package/MDAnalysis/analysis/hbond_analysis.py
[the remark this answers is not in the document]
orbeckst: I thought the same. Raise a separate issue. It looks like a simple change but this is such a fundamental method that it will require some careful testing. If it works consistently then it would be great. I think there was an old issue related to the "use AG where strings are allowed". 

There are some cases where it won't just work, in particular anywhere where strings are used to create "updating AtomGroups" by executing the same selection every frame (I think the waterdyamics module does this and probably elsewhere, too). Anyway, separate issue!

package/MDAnalysis/analysis/hbond_analysis.py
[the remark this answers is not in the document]
orbeckst: In principle, selecting via topology is the only _correct_ way of finding the bonded hydrogens (assuming that the bonds were not guessed). 

Do you have to find the hydrogens at every time step or can you do it for once and for all?

p-j-smith:
Apologies for the very slow reply, I have been pretty busy recently. I have made a few updates to the class:

* If `acceptors_sel` is not provided, then it is guessed by considering the atomic charges.
* If `hydrogens_sel` is not provided, then it is guessed by considering the atomic charges and masses.
* If `donors_sel` is not provided, then the topology is used to find donor-hydrogen pairs.
* Added an argument `update_selections`, which if False will not update atom groups at every step.
* Added a `count_by_ids` method, which is equivalent to the `count_by_type` method in the original class.
* Added a `count_by_type` method which returns a count of each type of hydrogen bond, where a unique type consists of a unique combination of donor resname, donor name, acceptor resname and acceptor name. This is therefore a coarser analysis of hydrogen bond types than the `count_by_ids` method.
* Added test cases for the new methods.
* Updated the documentation.

p-j-smith:
I'm just wondering whether there are any plans to update the version of numpy used by MDAnalysis, as the convenience functions I have written require numpy v.1.13 or later? If not, I can make some minor adjustments to fix it.

richardjgowers:
What’s the function that you need?
  

  
  

  
  
>

p-j-smith:
I'm using np.unique() with the argument axis=0, but the axis argument was only introduced in version 1.13.0

tylerjereddy:
Oh yeah, `np.unique` with the axis argument is probably pretty powerful for simulation / coordinate analysis since you can i.e., filter out duplicate coordinates without using pandas.

I think I've forgotten what our policy is on minimum supported NumPy version, but I suspect we do have to be somewhat conscious of the slower adoption of new software that sometimes happens in academia.

zemanj:
> I think I've forgotten what our policy is on minimum supported NumPy version

I'd also be interested in knowing on what basis such things are decided. Currently we have 1.10.2 (I think?) as a minimum requirement, and that misses quite a bunch of handy features such as `isin()` etc.

@p-j-smith We have `lib.util.unique_rows()` as a drop-in for 2d arrays, but you'd still have to make it return counts in order to work with the data structure you chose for storing hbond info etc. I'd prefer bumping the minimum NumPy version to 1.13 but I don't know if that's easily possible.

orbeckst:
"We" just decided that the minimum supported version of numpy will be *1.13.3.* (see #2272). PR should be coming soon. Does this help?

orbeckst:
@VOD555 please have a look at this PR #2237 for the current work on the new H-bond analysis.

We need to get going/reviewing on this one – sorry to @p-j-smith and @bieniekmateusz for the stall.

orbeckst:
@p-j-smith sorry for the long silence on this PR. Will you be able to pick it up again? I would like to move forward with it, especially with a view towards 1.0.

As @kain88-de said, the existing H-bond analysis is heavily used so we cannot just kill it. I would propose to put the existing H-bond module `MDAnalysis.analysis.hbonds` into legacy mode and see that we can get your new version up and running in parallel. If you also want to add correlations etc then we should have this in its own submodule, but just *not* named "hbonds". What would be a good name for the new submodule? "nonbonded", "interactions", "hydrogenbonds", ...?

p-j-smith:
@orbeckst no problem. I won't be able to do much for the next couple of weeks, but I'm definitely still keen to see this through.

I think that putting the current `MDAnalysis.analysis.hbonds` into legacy mode makes sense. @bieniekmateusz and I are planning to add hydrogen bond lifetimes to this new version, so it will indeed need to become a submodule. For the name of the submodule, I prefer "hydrogenbonds" but I'm open to other suggestions.

orbeckst:
@p-j-smith @bieniekmateusz I rebased against develop and force-pushed (sorry about that). You should still be able to pull – but better pull before making any edits.

I moved your module into a new `analysis.hydrogenbonds` so that it can be cleanly developed in parallel to the old `analysis.hbonds` (see commit message for 5c72ed5).

orbeckst:
@VOD555 – if you want to try the new hbond analysis then you should be able to use this branch. I ran the tests locally and they passed.

package/MDAnalysis/analysis/hbond_analysis.py
orbeckst: This is now `analysis.hydrogenbonds.hbond_analysis`

package/MDAnalysis/analysis/hbond_analysis.py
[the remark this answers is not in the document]
orbeckst: @richardjgowers can you raise an issue for your idea that `select_atoms` should be able to pass through ags, please?

orbeckst:
Some of the tests fail, perhaps because an older version of numpy is being installed, see,  e.g., https://travis-ci.com/MDAnalysis/mdanalysis/jobs/221063112 with error
```
>       indices, tmp_counts = np.unique(self.hbonds[:, 0], axis=0, return_counts=True)
E       TypeError: unique() got an unexpected keyword argument 'axis'
```

Under ` source ci-helpers/travis/setup_conda.sh`:
```
UnsatisfiableError: The following specifications were found to be in conflict:
  - numpy=1.16 -> python[version='>=2.7,<2.8.0a0'] -> *[track_features=debug]
  - python=3.5
Use "conda info <package>" to see the dependencies for each package.
Installing numpy=1.16 with conda was unsuccessful, removing from conda install and adding numpy==1.16.*  to PIP_DEPENDENCIES instead.
...
...
numpy:                1.11.3-py35_blas_openblashd3ea46f_205 conda-forge [blas_openblas]
```

VOD555:
@p-j-smith @bieniekmateusz @orbeckst It's a good idea to identify donors and acceptors using mass and charge information. But some of the topology files such as `pdb` and `gro` files don't have charge information. This will cause an error when using the 'new' `HydrogenBondAnalysis`. 

I think we still need some pre-guess donor and acceptor atoms similar to the ones in the 'old' `HydrogenBondAnalysis` for these topology without charge information.

VOD555:
@p-j-smith @orbeckst @bieniekmateusz Another problem is about the selection string. 

Suppose my `universe` (topology) contains a protein and water molecules, and I want to find out the h-bonds between atoms of the protein. If I give the `universe` to the function without selection strings, `HydrogenBondAnalysis` will take all water molecules into account. If i give the `universe` with donor and acceptor selection `protein`, it will take all atoms in the protein as potential acceptors and donors. To get what I want, I have to generate a new topology without water or give detailed selection string (such as 'name O or OC1 or OC2 ...'). This is not friendly to users.

May be we need to define a pre-process function to generate such detailed donor and acceptor selection strings.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -0,0 +1,384 @@
...
+
+    def __init__(self, universe, donors_sel=None, hydrogens_sel=None, acceptors_sel=None,
+                 d_h_cutoff=1.2, d_a_cutoff=3.0, d_h_a_angle_cutoff=150, update_selections=True):
+        """Set up atom selections and geometric criteria for finding hydrogen bonds in a Universe.
+
+        Parameters
+        ----------
+        universe : Universe
+            MDAnalysis Universe object
+        donors_sel : str
```
orbeckst: Note: when hydrogens_sel is provided, leave donors_sel None so that donors are found automatically.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -0,0 +1,384 @@
...
+            Selection string for the hydrogen bond donor atoms
+        hydrogens_sel :  str
```
orbeckst: If hydrogens are left empty and are guessed then donors should also be left to None so that donors are properly guessed.

p-j-smith:
@orbeckst Sorry for my silence over the past few weeks and thanks for rebasing against develop. I've pulled your changes and will have some time this week to make the changes or address the comments above.

Re https://github.com/MDAnalysis/mdanalysis/pull/2237#issuecomment-516683793, it does look like numpy=1.11 is being used for some reason, rather than numpy=1.13. Do you know why this might be or how I can force the use of v1.13?

p-j-smith:
@VOD555 I completely agree that the interface is not ideal, but I have not yet come up with a better way of simplifying the api whilst maintaining the flexibility or generality of the class. You mention that:

> I think we still need some pre-guess donor and acceptor atoms similar to the ones in the 'old' `HydrogenBondAnalysis` for these topology without charge information.

and that:

> Maybe we need to define a pre-process function to generate such detailed donor and acceptor selection strings.

The reasoning behind the `guess_acceptors` and `guess_hydrogens` methods was indeed to generate the donor and acceptor strings. I'm somewhat reluctant to include a pre-defined list of acceptor and donor atoms because it makes the class much less general. Currently, the pre-defined list includes only atoms from the protein force fields of CHARMM and Amber. This completely neglects anyone who works with lipids, carbohydrates or organic molecules, as well as LAMMPS users (as LAMMPS does not keep the CHARMM/Amber atom names). It even cannot accommodate users looking at ligand-protein interactions.

By using charge to identify potential donors and acceptors, it means that anyone using gromacs (`tpr`), namd (`psf`), lammps (`lammps data`) or amber (`prmtop`) can use the class to find hydrogen bonds for any system, so long as they create their universe using a file that contains the partial charges. By using, for example, a `psf` rather than a `pdb`, bonding information will also be available, which, as @orbeckst has previously pointed out, is the only guaranteed way of correctly identify donor-hydrogen pairs.

What do you think about making the `guess_acceptor` and `guess_hydrogen` methods class methods, which can be called with a selection string such as "protein" to generate the required selections? This would mean that not all water molecules would have to be included in the calculation, whilst still allowing the use of complex selection strings when creating a 'HydrogenBondAnalysis' object.


For example, something like this may be preferable to the current implementation:

```
# find all atoms excluding water
acceptors_sel = HydrogenBondAnalysis.guess_acceptors(u.select_atoms("not resname TIP3"), max_charge=-0.5))
hydrogens_sel = HydrogenBondAnalysis.guess_acceptors(u.select_atoms("not resname TIP3"), max_mass=1.1, min_charge=0.4)

# water selections
water_acceptors_sel = "resname TIP3 and name OH2"
water_hydrogens_sel = "resname TIP3 and name H1 H2"

# hydrogen bonds of all molecules (protein, lipid, carbohydrate, organic) as well as water within 10 A
    universe=u,
    donors_sel=None,
    hydrogens_sel=f"({hydrogens_sel}) or ({water_hydrogens_sel} and around 10 not resname TIP3})",
    acceptors_sel=f"({acceptors_sel}) or ({water_acceptors_sel} and around 10 not resname TIP3})",
    d_h_cutoff=1.2,
    d_a_cutoff=3.0,
    d_h_a_angle_cutoff=150,
    update_selections=True
)

```

This is still probably not ideal, but I feel like it offers a good compromise between usability and flexibility.

VOD555:
@p-j-smith Actually, I used this Hbond analysis class for my project in a similar way to your example, and it worked very well. I think making the `guess_acceptor` and `guess_hydrogen` methods class methods is a  good idea.

p-j-smith:
@VOD555 I'm glad it worked well for you. I tried changing guess_acceptors() and guess_donors() to be static methods but it actually seemed to make the usage a bit more clunky. I've added some examples to the documentation which will hopefully make it relatively easy for new users to use the class.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
p-j-smith: I've updated the docs to this effect

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
p-j-smith: I've added this to the docs

p-j-smith:
@orbeckst I think I'm nearly there now with this PR. I've added a `guess_donors` method that creates a donor selection string (so long as there is no bonding information in the topology).
Otherwise, in terms of the points you raised:

> * address comments as far as possible or state if you can't/don't want to address with a reason
I think I've addressed all the comments made on this PR

> * make use of numpy >= 1.13 functionality (unique something or other, see comments)
The helper functions (for counting the number of hydrogen bonds etc) make used of the `axis` argument in `np.unique`

> * add docs (make sure the docs build and show your docs – ask if you have questions how to do it)
I've added the docs and they build fine (I've uploaded a zipped html)
[hydrogenbonds.zip](https://github.com/MDAnalysis/mdanalysis/files/3581287/hydrogenbonds.zip)

> * update CHANGELOG
I've not yet done this, as I'm not sure which version to add it under, 0.20.2 or 1.0?

orbeckst:
Please rebase against the latest develop (or merge). In the CHANGELOG, change the `0.20.2` to `0.21.0` – this is a new feature so has to go into a new minor release.

I'll look through the other things as I managed to make time.

Sorry for the delay in replying... busy times.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -31,12 +31,18 @@
 This module provides methods to find and analyse hydrogen bonds in a Universe.
 
 The :class:`HydrogenBondAnalysis` class is a new version of the original :class:`HydrogenBondAnalysis`
```
orbeckst:
```reST
original :class:`MDAnalysis.analysis.hbonds.HydrogenBondAnalysis`
```

orbeckst:
ping @p-j-smith

package/MDAnalysis/analysis/hbond_analysis.py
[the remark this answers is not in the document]
orbeckst: raised #2369

KCLPhysics:
Hi @orbeckst sorry for being so slow with this, I should have some time tomorrow evening to finish this of, or if not then definitely by the end of this week

orbeckst:
@richardjgowers do you want to do a quick 0.20.2 release before this gets merged?

p-j-smith:
So I think I have made all the changes:

> * apparently, `guess_donors()` is not covered by the tests, see coverage report for [hydrogenbonds/hbond_analysis.py](https://codecov.io/gh/MDAnalysis/mdanalysis/pull/2237/diff?src=pr&el=tree#D1-306). Can you please add a test that exercises lines [306–326](https://codecov.io/gh/MDAnalysis/mdanalysis/pull/2237/diff?src=pr&el=tree#D1-306...326)

I've added a simple test for the `guess_donors()` method

> * Very minor comments on the docs

I've updated the docs.

> * rebase against develop

I've rebased against develop (hopefully I did it correctly!)

> * add entry to CHANGELOG under 0.21.0

I've updated the CHANGELOG

Also, I was wondering whether it would be appropriate to add a reference for users of the class to cite? It would be for the paper for which I wrote a hydrogen bond analysis script on which this class is based.

orbeckst:
@KCLPhysics please ping me when ready to review.

Also, @bieniekmateusz / @p-j-smith you might have to add a previous email address to your GitHub account so that some of the older commits are attributed to a GutHub user – unless you don't care.

Should I add @KCLPhysics to the contributors – you seem to be using this account since recently.

orbeckst:
> Also, I was wondering whether it would be appropriate to add a reference for users of the class to cite? It would be for the paper for which I wrote a hydrogen bond analysis script on which this class is based.

Absolutely!!!! 

* add the paper to references https://www.mdanalysis.org/docs/documentation_pages/references.html#citations-for-included-algorithms-and-modules
* use reST referencing to cite the paper
* (optional but HIGHLY recommended): add a duecredit citation so that it can be auto-cited with [duecredit](https://www.mdanalysis.org/docs/documentation_pages/references.html#citations-using-duecredit), see e.g., https://github.com/MDAnalysis/mdanalysis/blob/9372d093acebb78bc5b1a2b4dba815a4e3281b04/package/MDAnalysis/analysis/psa.py#L238-L244

p-j-smith:
> @KCLPhysics please ping me when ready to review.
> 
> Also, @bieniekmateusz / @p-j-smith you might have to add a previous email address to your GitHub account so that some of the older commits are attributed to a GutHub user – unless you don't care.
> 
> Should I add @KCLPhysics to the contributors – you seem to be using this account since recently.

Sorry for the confusion with the different accounts. I forgot that I was signed in as @KCLPhysics when I commented a couple of times. Thanks for the heads up about the previous commits, I'll add my other email address to my account.

p-j-smith:
Brilliant, thanks - I've added a reference to cite, and fixed a small error with the docs, so it looks like we're there!

orbeckst:
Don't merge until @richardjgowers has stated what he wants to do in terms of a 0.20.2 release or a 0.21.0 release. This new feature _must_ go into a minor release and cannot go into a patch level release.

orbeckst:
@p-j-smith please fix the docs error https://travis-ci.com/MDAnalysis/mdanalysis/jobs/248888347

orbeckst:
Yay! All green.

@p-j-smith do you have the time to clean up the history of this commit? We can't merge this right away (until I hear from @richardjgowers how we want to do the next release). The history of this PR would be much more readable if you could

1. rebase against latest develop
2. merge commits into largely independent units

By the latter I mean getting rid of little "I fixed this to make that work" commits. I use `git rebase -i develop` (interactive rebase). This is a completely new feature so in principle this can be reduced to the new code, new tests, changes to docs, changes in other code (eg deprecations).

richardjgowers:
When this is merged the next release becomes a 21.0 so put the change log
entries for this into 20.2 (which won’t ever exist) and I’ll make
everything say (ie sed) 21.0 when I mangle the version numbers on release.

orbeckst:
I merged the CHANGELOG and switched it to 0.21.0.

@p-j-smith if your code has "versionadded" and similar things then please change to 0.21.0. Then just squash my merge and your version changes into a single "version update" commit.

Sorry, these are little tedious things but the heavy lifting is done!

p-j-smith:
Brilliant, thanks @orbeckst I think we're there!

orbeckst:
Congratulations @p-j-smith and @bieniekmateusz – a major new analysis class. Well done.

@VOD555 – new H-bond analysis is now in develop, in case you want to add it to PMDA https://github.com/MDAnalysis/pmda/issues/95

PR #2254, opened 2019-04-18
Adds test coverage to Cython files

Fenil3510:
 - Added test coverage for cython, works fine locally. Minor changes from PR #2029. This looks kind of a hacky way to me, please review

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ os:
 
 env:
   global:
-    - secure: "f8EMSWeYC38elhpB4B/ddxlklEvQoycaxnt90Xw2tH/+ThdP1qteQ2vdgNFy1KL7Am/xnbrRhavI5K+ayfxJ93NoE2adaJ9f9aljXK+Oeu+buv5MVo2E2HhN9mX9opSSxiqGmnHIVYcdLP+1soIsDD78SGL7hB/u5nQ1aTzkbaM="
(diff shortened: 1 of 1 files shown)


richardjgowers:
@Fenil3510 we can't add the c/cpp files to ignore because we *do* include them in the master branch.  So releases don't require Cython.

package/setup.py
```diff
@@ -94,6 +94,7 @@
               "parallelization module".format(
                Cython.__version__, required_version))
         cython_found = False
+    cython_linetrace = True
```
richardjgowers: Ideally this isn't on by default as I'm guessing it has a performance cost.  We could probably look for one of the travis/CI environment variables to trigger this?

Fenil3510: Yes, this is not a default one and there's an environment variable to trigger this as you had already done in #2029, but I was forcing this to look at the results for now.

Fenil3510:
`.pyx` still not visible in `codecov` report. Have to work on it.

Fenil3510:
> @Fenil3510 we can't add the c/cpp files to ignore because we _do_ include them in the master branch. So releases don't require Cython.

The `.c` files were due to local build, will have to remove them anyway.

PR #2268, opened 2019-05-29
[WIP] Extending implementation of unwrap

### LINKED ISSUE #2264: Extend Implementation of Unwrap
The current implementation of Group.unwrap(PR [#2189](https://github.com/MDAnalysis/mdanalysis/pull/2189)) works for bonded structures.
This method should also work for molecules which are part of the same group, e.g., DNA molecules.

**Proposed Solution**
- Add keyword `reference_point`
- Unwrap molecules such that they are closest to the `reference_point`

**Questions to answered before implementation**
- Is there a more efficient method for unwrapping?
- Which test cases will be used?

NinadBhat:
Fixes #2264 

 - 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


package/MDAnalysis/core/groups.py
```diff
@@ -1641,12 +1644,45 @@ def unwrap(self, compound='fragments', reference='com', inplace=True):
...
+                ctrpos = unique_atoms.center_of_mass(pbc=False)
+            elif ref == 'cog':
+                ctrpos = unique_atoms.center_of_geometry(pbc=False)
+            else:
+                raise ValueError("Unrecognized reference '{}'. Please use one "
+                                 "of 'com', 'cog', or None.".format(centre))
+
+        difference = reference_point - ctrpos
+
+        shift = np.rint(np.divide(difference, unique_atoms.dimensions[:3]))
```
richardjgowers: @NinadBhat this only works for orthogonal boxes, for triclinic boxes the math is a little more complicated:

https://github.com/MDAnalysis/mdanalysis/blob/develop/package/MDAnalysis/lib/include/calc_distances.h#L34

package/MDAnalysis/core/groups.py
```diff
@@ -1641,12 +1644,45 @@ def unwrap(self, compound='fragments', reference='com', inplace=True):
...
         if not atoms.isunique:
             positions = positions[atoms._unique_restore_mask]
         return positions
 
+    def choose_periodic_image(self, reference_point, centre='com', inplace=True):
+        atoms = self.atoms
+
+        unique_atoms = atoms.unique
+
+        if centre is not None:
```
richardjgowers: looks like centre can't be `None`?

package/MDAnalysis/core/groups.py
```diff
@@ -1641,12 +1644,45 @@ def unwrap(self, compound='fragments', reference='com', inplace=True):
...
         if inplace:
             unique_atoms.positions = positions
         if not atoms.isunique:
             positions = positions[atoms._unique_restore_mask]
         return positions
 
+    def choose_periodic_image(self, reference_point, centre='com', inplace=True):
+        atoms = self.atoms
+
+        unique_atoms = atoms.unique
```
richardjgowers: `unique_atoms = self.atoms.unique`?

package/MDAnalysis/core/groups.py
```diff
@@ -1520,6 +1520,8 @@ def unwrap(self, compound='fragments', reference='com', inplace=True):
             unwrapped compounds will be shifted so that their individual
             reference point lies within the primary unit cell.
             If ``None``, no such shift is performed.
+        reference_point: coordinate, option
+            position where molecule should be unwrapped
```
jbarnoud: What is the behaviour if `reference_point` is `None` (default)?

package/MDAnalysis/core/groups.py
```diff
@@ -1624,7 +1628,6 @@ def unwrap(self, compound='fragments', reference='com', inplace=True):
                 c = unique_atoms[mask]
                 positions[mask] = mdamath.make_whole(c, inplace=False)
                 # Apply reference shift if required:
-                if reference is not None:
                     if ref == 'com':
```
jbarnoud: Could you check the indentation here. From the diff it looks like you removed the `if` statement but did not remove the indentation on the block.

NinadBhat:
@richardjgowers, @jbarnoud  I get the following error while importing minimizing_vector
`ImportError: cannot import name 'periodic_image'`
Any suggestions?

package/MDAnalysis/lib/distances.py
```diff
@@ -1523,3 +1526,16 @@ def apply_PBC(coords, box, backend="serial"):
         _run("triclinic_pbc", args=(coords, box), backend=backend)
 
     return coords
+
+def minimizing_vector(reference_point, ctrpos, box, backend="serial"):
```
jbarnoud: I am not sure the name is explicit enough as many things are vectors and minimizing may mean various things. `minimize_periodic_vector` perhaps?

NinadBhat: Will change it.

richardjgowers:
Looking good, try and include some tests. As something very easy to set up, take a molecule, shift its atoms a few box vectors in random directions, (eg atom 1 + 2* box in x direction, atom 2 - 3* box in y etc) then see if make_whole with reference point works. I think maybe I already wrote tests similar....
  

  
  

  
  
>

NinadBhat:
@jbarnoud , @richardjgowers  I have update unwrap such that it unwraps molecules to their closest periodic image to the reference_point. I will start updating the current tests and adding more tests.

package/MDAnalysis/lib/distances.py
```diff
@@ -1523,3 +1526,19 @@ def apply_PBC(coords, box, backend="serial"):
         _run("triclinic_pbc", args=(coords, box), backend=backend)
 
     return coords
+
+
+def minimize_periodic_vector(reference_point, ctrpos, box, backend="serial"):
```
richardjgowers: docs here, what does this do? when would I use it? etc

richardjgowers:
@NinadBhat I'm not sure you can change the default behaviour of the unwrap function... though it shouldn't really break anything for people if you're using a different image.  Can still add tests for the keyword you've added

testsuite/MDAnalysisTests/core/util.py
```diff
@@ -438,7 +440,29 @@ def unwrapped_coords(self, compound, reference):
...
+                    shift = np.rint((reference_point - loc_centre) /self._box_edge)
+                    relpos[base:base + 4, :] += shift
+
+                # type D
+                for base in range(23, 47, 8):
+                    loc_centre = np.mean(relpos[base:base + 8, :], axis=0)
+                    shift = np.rint((reference_point - loc_centre) / self._box_edge)
+                    relpos[base:base + 8, :] += shift
+
         if self._is_triclinic:
```
NinadBhat: @richardjgowers , @jbarnoud I am stuck at writing tests for triclinic systems. Does this **if block** convert the coordinates of a molecule from orthorhombic system to triclinic system?

jbarnoud: Indeed. From what I understand, the coordinates of each molecule is provided in an orthorhombic system and are shifted in the case of a non-orthorhombic system (orthorhombic is a special case of triclinic, so in all cases it is triclinic) so that they still match their description.

zemanj: @NinadBhat I designed the UnWrapUniverse so that it first initializes all coordinates in an orthorhombic box. Thereafter, if the box is actually supposed to be triclinic, it applies a position-dependent shift (the "skew" of the box) to the x- and y-components of all coordinates. That's what's being done here.

jbarnoud:
> @richardjgowers, @jbarnoud I get the following error while importing minimizing_vector
ImportError: cannot import name 'periodic_image'
Any suggestions?

I realize I did not answer that. Is this still a problem?

testsuite/MDAnalysisTests/core/util.py
zemanj: Think of it as taking an orthorhombic box and then shearing the top face in the x- and y-direction without altering the height of the box.

package/MDAnalysis/lib/distances.py
```diff
@@ -1523,3 +1526,45 @@ def apply_PBC(coords, box, backend="serial"):
...
+    """Returns the periodic image of ctrpos which is closest to the reference_point
+
+    Parameters
+    ----------
+
+    reference_point : numpy.ndarray
+        Coordinate array of shape ``(3,)`` or ``(n, 3)`` containing the
+        position of reference_point (dtype is arbitrary, will be converted to
+        ``numpy.float32`` internally)
+    ctrpos : numpy.ndarray
```
jbarnoud: I do not like the name `ctrpos`. It is not very explicit, I'd rather have something like `center_position`. Also, I do not fully understand what it is about from the [arameter description.

testsuite/MDAnalysisTests/core/test_unwrap.py
```diff
@@ -37,13 +37,21 @@ class TestUnwrap(object):
     which is specifically designed for wrapping and unwrapping tests.
     """
     precision = 5
+    reference_point_pos = (None,
+                           np.array([5, 5, 5], dtype=np.float32),
```
jbarnoud: I would avoid having x, y, and z set to the same value. If you have an error in your indices it may not make the test fail.

package/MDAnalysis/core/groups.py
jbarnoud: Also, how does this option behave relative to the `reference` one? If `reference` is `None` but I provide a `reference_point`, what shall I expect?

Maybe `reference` and `reference_point` should be merged so that `reference` can accept a position vector? Or maybe I do not understand how the two options work together.

zemanj:
I'm sorry that I jump in here a little late, but I'm extremely busy at the moment.
My main concern here is that we are about to include too much in the unwrap method, which makes it very hard to explain the different functionalities in the doc string (especially how the different options interact). This has a very high potential to leave users puzzled and frustrated.

IMHO moving molecules by individual box shifts so that they reside as closely to some reference point is definitely more than just unwrapping, and should therefore go into a separate method. In fact, it is an additional *independent* step that doesn't even necessarily require unwrapping.

Furthermore, even though the newly implemented functionality is definitely a step forward, yet it won't suffice in many situations. Think, for example, of a system comprising several micellar complexes where one wants each of the micelles to be unwrapped so that the shells stay as close as possible to their central compounds. Or maybe a DNA quadruplex where you want the individual strands being unwrapped with respect to their partners.

If we include all such functionality into one method, this will result in a complete mess and make it less flexible in the end. If we want to avoid a maintenance nightmare and spare the users a lot of confusion, we should really separate these things!

testsuite/MDAnalysisTests/core/test_unwrap.py
NinadBhat: Sure.

NinadBhat:
> > @richardjgowers, @jbarnoud I get the following error while importing minimizing_vector
> > ImportError: cannot import name 'periodic_image'
> > Any suggestions?
> 
> I realize I did not answer that. Is this still a problem?

@jbarnoud Found the solution to that.

zemanj:
BTW, don't we already have something very similar in ``MDAnalysis.transformations.center_in_box(ag, center='geometry', point=None, wrap=False)``?
Along these lines, the unwrap extension here could go into a function ``AtomGroup.arrange_around(self, target_position, compound='fragments', reference='com', inplace=True)``.

NinadBhat:
> IMHO moving molecules by individual box shifts so that they reside as closely to some reference point is definitely more than just unwrapping, and should therefore go into a separate method. In fact, it is an additional _independent_ step that doesn't even necessarily require unwrapping.
> 

Makes sense to me I will put reference_point in a different method which will be called by unwrap.

> Furthermore, even though the newly implemented functionality is definitely a step forward, yet it won't suffice in many situations. Think, for example, of a system comprising several micellar complexes where one wants each of the micelles to be unwrapped so that the shells stay as close as possible to their central compounds. Or maybe a DNA quadruplex where you want the individual strands being unwrapped with respect to their partners.
> 

@zemanj I planned to unwrap individual strand and then move them to periodic image closest to the reference point. If this does not do the needful maybe I will have to add pseudo bonds between the strands (As mentioned in [comment](https://github.com/MDAnalysis/mdanalysis/issues/2264#issuecomment-497620258) and then unwrap it. @richardjgowers  and @jbarnoud  Any thoughts on this?

zemanj:
@NinadBhat 
> Makes sense to me I will put reference_point in a different method which will be called by unwrap.

I think it's better to do this the other way round and make the the extra method *optionally* call ``unwrap()``.
Its signature would then look like this:
``AtomGroup.arrange_around(self, target_position, compound='fragments', compound_reference='com', unwrap=True, inplace=True)``

NinadBhat:
> Also, how does this option behave relative to the reference one? If reference is None but I provide a reference_point, what shall I expect?

> Maybe reference and reference_point should be merged so that reference can accept a position vector? Or maybe I do not understand how the two options work together.

@jbarnoud  could not reply to this comment above, so answer it here. 

**reference**: defines which centre should be moved i.e 'cog' or 'com'.
**reference_point**: defines where the centre will be moved

I am not sure what the behaviour should be when **reference** is not defined. Should be default to one of 'cog' or 'com'?

NinadBhat:
> @NinadBhat
> 
> > Makes sense to me I will put reference_point in a different method which will be called by unwrap.
> 
> I think it's better to do this the other way round and make the the extra method _optionally_ call `unwrap()`.
> Its signature would then look like this:
> `AtomGroup.arrange_around(self, target_position, compound='fragments', compound_reference='com', unwrap=True, inplace=True)`

@zemanj Thank for the reply. I wil do that.

zemanj:
@NinadBhat 
> I will have to add pseudo bonds between the strands

I know this sounds tempting, but tinkering with additional bonds is something I'd avoid at all cost given the way our topology system is currently implemented. We will run into a lot of problems handling the addition of bonds  because we have a couple of cached data structures that are invalidated by changing the topology (``Universe._fragdict``, for example), and there's currently no easy way of clearing/updating these structures.

zemanj:
> > Also, how does this option behave relative to the reference one? If reference is None but I provide a reference_point, what shall I expect?

That exactly hits the point I'm trying to make here:
If ``reference=None``, each compound will be unwrapped relative to the position of its first atom.
Having ``reference_point=<some position>`` combined with ``reference=None`` therefore doesn't really make sense and should result in an exception. => complicated!
> > Maybe reference and reference_point should be merged so that reference can accept a position vector? Or maybe I do not understand how the two options work together.

That certainly won't work. 

> 
> @jbarnoud could not reply to this comment above, so answer it here.
> 
> **reference**: defines which centre should be moved i.e 'cog' or 'com'.
> **reference_point**: defines where the centre will be moved
> 
> I am not sure what the behaviour should be when **reference** is not defined. Should be default to one of 'cog' or 'com'?

Implicitly changing the meaning of a kwarg depending on the value of some other kwarg is something you shouldn't even consider. Ever.

NinadBhat:
> I know this sounds tempting, but tinkering with additional bonds is something I'd avoid at all cost given the way our topology system is currently implemented. We will run into a lot of problems handling the addition of bonds because we have a couple of cached data structures that are invalidated by changing the topology (Universe._fragdict, for example), and there's currently no easy way of clearing/updating these structures.

@zemanj Thanks for the comments and reviews. Did you have any mechanism for unwrapping non bonded molecule, e.g., DNA quadruplex in mind?

richardjgowers:
> IMHO moving molecules by individual box shifts so that they reside as closely to some reference point is definitely more than just unwrapping, and should therefore go into a separate method. In fact, it is an additional _independent_ step that doesn't even necessarily require unwrapping

Yeah I think maybe this functionality is better split out.



> Furthermore, even though the newly implemented functionality is definitely a step forward, yet it won't suffice in many situations. Think, for example, of a system comprising several micellar complexes where one wants each of the micelles to be unwrapped so that the shells stay as close as possible to their central compounds. Or maybe a DNA quadruplex where you want the individual strands being unwrapped with respect to their partners.

I think this is sufficiently complex that it should become many function calls, so pseudocode:
```python
from the_user import identify_micelles, calculate_center

micelles = identify_micelles(u.atoms)

for m in micelles:
    center = calculate_center(m)  # whatever the definition of "core" is, not MDA's problem
    m.center_around(center)  # this *is* MDA's problem now it is tightly defined, this is what this PR should solve
```

richardjgowers:
@NinadBhat I think your proposal to unwrap individual items then gather them around a given reference point sounds sane.  I don't think MDAnalysis needs to be smart enough to automatically identify these structures, the user should provide this information (via AtomGroups)

Is there actually a performance / usability increase by having `unwrap` included in the `arrange_around` method?  Otherwise I think it's clearer to have the two as completely different steps?

richardjgowers:
> BTW, don't we already have something very similar in `MDAnalysis.transformations.center_in_box(ag, center='geometry', point=None, wrap=False)`?

These aren't meant to be used as functions directly, but as a function attached to a `Reader`.  Really they shouldn't contain any cool code, but instead just know which `AtomGroup` methods to call....  (ie squeeze any good code out of them into more fundamental objects, the transformation functions are just "recipes"?)

zemanj:
> @zemanj Thanks for the comments and reviews. Did you have any mechanism for unwrapping non bonded molecule, e.g., DNA quadruplex in mind?

Just like @richardjgowers just outlined, that is probably something involving several intermediate steps. Assuming we have 4 double strands in the system (i.e., 8 single strands with no artificial bonds connecting each of the 4 pairs), the procedure might look as follows:
1. Select all strands so that all of them are accessible in one AtomGroup `ag`.
2. Unwrap all 8 single strands with
```python
ag.unwrap(compound='fragments', inplace=True)
```
3. Define 4 groups containing one double strand per group ``ags=[ag1, ag2, ag3, ag4]``
4. Unwrap the centers of mass (or whatever reference) of the single strands in each group with respect to each other:
```python
for ag in ags:
    ag.arrange_closest(compound='fragments', reference='com', inplace=True)
```
5. Arrange AtomGroups such that their reference positions (`'com'` by default) lie as close as possible to some `target_position`:
```python
arrange_groups_around(ags, target_position, reference='com', inplace=True)
```

Obviously, this requires two additional functionalities:
1. A method `AtomGroup.arrange_closest()`. This method could also have an optional kwarg `unwrap=True` so one could save step 2.
This method is very similar to what you're currently implementing but places all compounds of the group as close as possible relative to each other instead of relative to a target position. This should be identical to selecting the center of mass of the first compound as the target position and applying your routine, so we could reuse it there. Maybe we could even have your method doing just that if no target position is given (``target_position=None``).
2. A method that takes an arbitrary number of AtomGroups and arranges them so that their reference positions (`'com'` by default) lie as close to the reference position. That is again very similar to what you are implementing, the only difference is that it acts on groups instead of compounds of groups.

Of course, the names of the methods are examples I just came up with, so their might be better ones.
And maybe you find some points where we could simplify the whole procedure.

zemanj:
Ok before we continue here, we should probably find a consensus about where this is going.
Here's my suggestion:

1. Implement a method that does pretty much what @NinadBhat was already about to be doing:
```python
AtomGroup.arrange_closest(target_position=None,
                          compound='fragments',
                          reference='com',
                          unwrap=True,
                          inplace=True)
```
  * If `unwrap` is `True`, the method must act on unwrapped coordinates obtained by internally calling `self.unwrap(compound=compound, reference=reference, inplace=inplace)` before doing anything else.
  * `reference` defines the type of reference coordinate per `compound` and must be either `'com'` or `'cog'` (most importantly, it must not be `None`).
  * If `target_position` is `None`, the `reference` of the first `compound` is taken as the target position.
  * If `target_position` is `None` and the group contains only one `compound`, that compound is already arranged with respect to itself and we're done here.
  * Finally, proceed as proposed by @NinadBhat and arrange all compounds such that their references lie as close as possible to whatever `target_position` dictates.

2. Implement a method that takes an arbitrary iterable of `AtomGroup`s and arranges them as closely as possible around some target point:
```python
arrange_closest(groups, target_position=None, reference='com', inplace=True)
```
  * This method is similar to `AtomGroup.arrange_closest()` but treats each group as a single compound. Thus, it doesn't have the `compound` and `unwrap` arguments.

@NinadBhat @richardjgowers @jbarnoud what's your opinion on this approach?

richardjgowers:
@zemanj if we leave `unwrap` out of that function, is it just the same as calling `unwrap` first? If so, isn't it easier just to not include unwrap in the signature (and have it called manually first)?

zemanj:
@richardjgowers Yes, exactly. Removing the `unwrap` keyword will simplify the code from a maintenance/testing perspective, so that's a good idea!

zemanj:
I think the second method I proposed should probably go into a separate PR since it exceeds the scope of the current one and requires additional discussion (what happens to groups from different universes, empty groups, etc.)

package/MDAnalysis/core/groups.py
zemanj: No, `atoms` is needed further below for `atoms._unique_restore_mask`

NinadBhat:
@richardjgowers, @jbarnoud, @zemanj 
Currently, there are some tests which are failing in triclinic system

I am getting an wierd error.


```
float box[9] = {10.,          0.,          0.,        0.0000001,  10.,          0, 0.0000005,   0, 10.};
double kf[3] = {4.616667, 2.983333, 4.833333};
minimum_image_triclinic(kf, box);
```
return kf = {4.616667, 2.983333, 4.833333};
The C file of above:- [gist](https://gist.github.com/NinadBhat/31a8ae6e878fd1c7acb46b87f0ba4bb9)

But the same thing through [here](https://github.com/MDAnalysis/mdanalysis/pull/2268/files#diff-b1ed937687dab7f27429e8e7ffbd9684R408) .

Returns 
dx = {3.616666, 1.983334, -5.166667}

NinadBhat:
On a broader case if I due the follow in the C file
```

    float box[9] = {10.,          0.,          0.,        0.0000001,  10.,          0, 0.0000005,   0, 10.};


    float centre[3] = {0.3833332,  2.016667,   0.16666667   };
    float ref[3] = {5.0, 5.0, 5.0};
    double x[3]= {6.999998,  0.5,       0.0          };

    float result[3] = {0, 0, 0};
    coordinate ref_t[1] = {  0.3833332,  2.016667,   0.16666667  };
    _translate_periodic_triclinic(ref, centre, result, box);
     for (int i=0; i<3; i++) {
    printf("%f\n", result[i]);
  }

I will get result = {0.383333, 2.016667, 0.166667}

But through cython it is giving me {1.3833337,  3.0166662, 10.166666}

```

NinadBhat:
Still, haven't figured out this. Any suggestions?

NinadBhat:
Got it. There was a mistake in the box configuration I was assuming.

NinadBhat:
@richardjgowers , @jbarnoud 
I have updated the tests for othrombic systems. I expected to [this](https://github.com/MDAnalysis/mdanalysis/blob/develop/testsuite/MDAnalysisTests/core/util.py#L442) to make the necessary changes for triclinic system. But this does not work when the reference_point is outside the unit cell (Kindly have a look at this [gist](https://gist.github.com/NinadBhat/9ac98140e3b9e9aeead0b752309889ef)). Should I hard code the results for triclinic system?

NinadBhat:
@richardjgowers @jbarnoud @orbeckst @micaela-matta Just a reminder to reply to the above question.

orbeckst:
1. Is your question about changing `UnWrapUniverse`?
1. In [your gist](https://gist.github.com/NinadBhat/9ac98140e3b9e9aeead0b752309889ef) you call `u.unwrapped_coords()` with the `reference_point=reference_point` kwarg
   ```python
   u = UnWrapUniverse(is_triclinic=is_triclinic)
   # select group appropriate for compound:
   group = u.atoms[6:9]
   # get the expected result:
   ref_unwrapped_pos = u.unwrapped_coords(compound, reference, reference_point=reference_point)[6:9]
   ```
   but as far as I can tell this kwarg is not available in `UnWrapUniverse.unwrapped_coords()` https://github.com/MDAnalysis/mdanalysis/blob/c820cc5143c504d06680b5e3fa42a3eb495c7be7/testsuite/MDAnalysisTests/core/util.py#L339
   [Your gist](https://gist.github.com/NinadBhat/9ac98140e3b9e9aeead0b752309889ef) seems to run with this kwarg so I am not sure if I am looking that the same code for `UnWrapUniverse.unwrapped_coords()`  that you're using. Why this difference, or am I missing something?

2. Have you drawn the unwrapped coordinates of the UnWrapUniverse and are you sure which ones are correct (i.e., the one that you computed with `unwrapped_pos = group.unwrap(compound=compound, reference=reference, reference_point=reference_point, inplace=False)` or the ones with `ref_unwrapped_pos = u.unwrapped_coords(..., reference_point=reference_point)`)?

NinadBhat:
>     1. Is your question about changing `UnWrapUniverse`?
Yes
> 
>     2. In [your gist](https://gist.github.com/NinadBhat/9ac98140e3b9e9aeead0b752309889ef) you call `u.unwrapped_coords()` with the `reference_point=reference_point` kwarg
>        ```python
>        u = UnWrapUniverse(is_triclinic=is_triclinic)
>        # select group appropriate for compound:
>        group = u.atoms[6:9]
>        # get the expected result:
>        ref_unwrapped_pos = u.unwrapped_coords(compound, reference, reference_point=reference_point)[6:9]
>        ```
>        
>        
>        but as far as I can tell this kwarg is not available in `UnWrapUniverse.unwrapped_coords()` https://github.com/MDAnalysis/mdanalysis/blob/c820cc5143c504d06680b5e3fa42a3eb495c7be7/testsuite/MDAnalysisTests/core/util.py#L339

I have updates UnWrapUniverse.unwrapped_coords() such that it has kwarg reference_point. When a reference_point is entered it will shift the fragments to the periodic image closet to the refernece_poinnt.

>        
>        [Your gist](https://gist.github.com/NinadBhat/9ac98140e3b9e9aeead0b752309889ef) seems to run with this kwarg so I am not sure if I am looking that the same code for `UnWrapUniverse.unwrapped_coords()`  that you're using. Why this difference, or am I missing something?
> 
My pull request includes the changes.

>     3. Have you drawn the unwrapped coordinates of the UnWrapUniverse and are you sure which ones are correct (i.e., the one that you computed with `unwrapped_pos = group.unwrap(compound=compound, reference=reference, reference_point=reference_point, inplace=False)` or the ones with `ref_unwrapped_pos = u.unwrapped_coords(..., reference_point=reference_point)`)?

Yes. For orthorhombic system both ref_unwrapped_pos and unwrapped_pos are same.  But for triclinic system, both are periodic images of the same molecule, but unwrapped_pos is closer to the reference_point compared to ref_unwrapped_pos. This is what I wanted to show from my gist.

orbeckst:
>  I have updates UnWrapUniverse.unwrapped_coords() such that it has kwarg reference_point. When a reference_point is entered it will shift the fragments to the periodic image closet to the refernece_poinnt.

You mean (with the proper link – please link to the actual code) in https://github.com/NinadBhat/mdanalysis/blob/4415fa82003328f17b46576dd8806e800d8c6856/testsuite/MDAnalysisTests/core/util.py#L339 
Perhaps the newly added code https://github.com/NinadBhat/mdanalysis/blob/4415fa82003328f17b46576dd8806e800d8c6856/testsuite/MDAnalysisTests/core/util.py#L443-L463 is not correct?

Can you make an actual plot of the positions?

NinadBhat:
@orbeckst Thanks for the suggestions. I realised one of the mistakes I was making was the reference_point should be shifted for the triclinic system. 

What I mean by the above is that `unwrapped_coords` calculated the position for orthorhombic cell and then shift it to a triclinic cell.  So, if I reference_point (`rf_trclinic`) is passed to `unwrapped_coords`, it should first calculate the orthorhombic position of the reference_point and the shift the atoms.

I have made the changes for the above, but there are still some tests that are failing. I am currently trying to figure out the changes for those.

NinadBhat:
@jbarnoud, @richardjgowers  This is ready for review.

NinadBhat:
@zemanj Yes I had added that functionality. In the last meeting, it was discussed that the method should be added in a separate PR as it could also have uses other than unwrap. I do not mind adding it here, @richardjgowers What do you think about this?

zemanj:
@jbarnoud I'm not quite sure why this PR is still open since IIRC its intended functionality should *not* be implemented directly in the `*Group.unwrap()` method but instead live in a separate `*Group` method.
@NinadBhat @richardjgowers Could you please comment on the status of this?

jbarnoud:
I am sure either. I read again the discussion but I was not completely sure what was the final verdict. Le 15 août 2019 17:03, Johannes Zeman <notifications@github.com> a écrit :@jbarnoud I'm not quite sure why this PR is still open since IIRC its intended functionality should not be implemented directly in the *Group.unwrap() method but instead live in a separate *Group method.
@NinadBhat @richardjgowers Could you please comment on the status of this?

NinadBhat:
@jbarnoud @zemanj I have created a different function called arrange_closest and planned to create a new PR for it. I have to add tests for that. Should I get that merged before moving ahead with this PR?

jbarnoud:
> 
> 
> @jbarnoud @zemanj I have created a different function called arrange_closest and planned to create a new PR for it. I have to add tests for that. Should I get that merged before moving ahead with this PR?

I think so yes.

PR #2275, opened 2019-06-17
 Adds Unwrap keyword to center

### LINKED ISSUE #2274: Add unwrap to center
Unwrap keywords needs to be added to center before it can be added to other methods.

**Additional context**
One questions that we need to discuss is whether we are removing keyword pbc?

NinadBhat:
Fixes #2274 

 -  Adds unwrap keyword to center

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


NinadBhat:
@jbarnoud , @richardjgowers Can I get a review on this?

package/MDAnalysis/core/groups.py
```diff
@@ -679,6 +679,11 @@ def center(self, weights, pbc=None, compound='group'):
             will be returned as an array of position vectors, i.e. a 2d array.
             Note that, in any case, *only* the positions of :class:`Atoms<Atom>`
             *belonging to the group* will be taken into account.
+        unwrap : bool or None, optional
```
zemanj:
```suggestion
        unwrap : bool, optional
```

package/MDAnalysis/core/groups.py
```diff
@@ -679,6 +679,11 @@ def center(self, weights, pbc=None, compound='group'):
...
+            If ``True`` and `compound` is ``'group'``, the atoms will be unwrapped
+            before calculations. If ``True`` and `compound` is
+            ``'segments'`` or ``'residues'``, all molecules will be unwrapped before
+             calculation to keep the compounds intact.
```
zemanj: Shorter:
```
If ``True``, compounds will be unwrapped before computing their centers.
```

package/MDAnalysis/core/groups.py
```diff
@@ -734,6 +739,8 @@ def center(self, weights, pbc=None, compound='group'):
 
         comp = compound.lower()
         if comp == 'group':
+            if unwrap:
+                coords = atoms.unwrap(compound=comp, reference=None, inplace=False)
             if pbc:
                 coords = atoms.pack_into_box(inplace=False)
```
zemanj: Having both `unwrap=True` and `pbc=True` doesn't make sense if `compound='group'`.
IMHO this should throw either an exception or a warning stating the precedence of operations.
I'm in favor of throwing an exception.

package/MDAnalysis/core/groups.py
```diff
@@ -769,11 +776,18 @@ def center(self, weights, pbc=None, compound='group'):
                              " one of 'group', 'residues', 'segments', "
                              "'molecules', or 'fragments'.".format(compound))
 
+
+
```
zemanj: remove superfluous newlines

testsuite/MDAnalysisTests/core/util.py
```diff
@@ -545,3 +546,67 @@ def wrapped_coords(self, compound, center):
...
+
+        pos = 0
+
+        if compound=="residues":
+            center_pos = np.zeros((15, 3), dtype=np.float32)
+        else:
+            center_pos = np.zeros((12, 3), dtype=np.float32)
+
+        for base in range(3):
+            loc_centre = relpos[base, :]
```
zemanj: It's not wrong, but I think we use American English throughout the code base. So use 'center' instead of 'centre'.

orbeckst: Yes, sorry...

NinadBhat:
@zemanj Thank you for the review. I have made the changes you suggested.

package/MDAnalysis/core/groups.py
```diff
@@ -734,6 +736,11 @@ def center(self, weights, pbc=None, compound='group'):
 
         comp = compound.lower()
         if comp == 'group':
+            if unwrap and pbc:
```
richardjgowers: This checking for if unwrap and wrap aren't both set is going to be very common, so it makes sense to check this in a decorator, eg how we use `@warn_if_not_unique` or `@_only_same_level`.

This decorator can also handle the deprecation of `pbc` into `unwrap` too....

```python
In [1]: def AorB(func): 
   ...:     def inner(*args, **kwargs): 
   ...:         if kwargs.get('A', False) and kwargs.get('B', False): 
   ...:             raise ValueError("both A and B set to true") 
   ...:         return func(*args, **kwargs) 
   ...:     return inner 
   ...:                                                                                                                                       

In [13]: @AorB 
    ...: def thingy(A=False, B=False): 
    ...:     return bool(A) or bool(B) 
    ...:                                                                                                                                      

In [14]: thingy()                                                                                                                             
Out[14]: False

In [15]: thingy(A=True)                                                                                                                       
Out[15]: True

In [16]: thingy(B=True)                                                                                                                       
Out[16]: True

In [17]: thingy(A=True, B=True)                                                                                                               
ValueError                                Traceback (most recent call last)
<ipython-input-17-8707bab20e96> in <module>
----> 1 thingy(A=True, B=True)

<ipython-input-1-baa4bbe6971a> in inner(*args, **kwargs)
      2     def inner(*args, **kwargs):
      3         if kwargs.get('A', False) and kwargs.get('B', False):
----> 4             raise ValueError("both A and B set to true")
      5         return func(*args, **kwargs)
      6     return inner

ValueError: both A and B set to true

```

NinadBhat:
@richardjgowers Thanks for the suggestion. I have made the changes.

package/MDAnalysis/core/groups.py
```diff
@@ -736,15 +748,13 @@ def center(self, weights, pbc=None, compound='group', unwrap=False):
...
-            if unwrap and pbc:
-                raise ValueError("'unwrap' and 'pbc' cannot be true at the same time "
-                                 "for compound='group")
-            if unwrap:
-                coords = atoms.unwrap(compound=comp, reference=None, inplace=False)
             if pbc:
                 coords = atoms.pack_into_box(inplace=False)
             else:
                 coords = atoms.positions
+            if unwrap:
```
richardjgowers: This logic isn't quite right.  For the case where `pbc=False, unwrap=True` then `coords` is generated twice.  Use an `elif` to fix this

package/MDAnalysis/core/groups.py
```diff
@@ -349,7 +349,15 @@ class _ImmutableBase(object):
     #  cache lookup if the class is reused (as in ag._derived_class(...)).
     __new__ = object.__new__
 
-
+def check_pbc_and_unwrap(function):
```
zemanj: Note that `pbc` and `unwrap` are only mutually exclusive if `compound='group'`.
There is nothing wrong with having both `pbc` and `unwrap` set to `True` if `compound` is e.g. `molecules`.

richardjgowers: How would this work?

zemanj: just as the docs say: "[...] Instead, the resulting position vectors will be moved to the primary unit cell after calculation."
In other words: 
1. `unwrap=True` => Unwrap components.
2. Compute center of each component
3. `pbc=True` => Wrap the computed centers into the primary unit cell.

richardjgowers: We're renaming `pbc=True` to `wrap=True`, having `method(wrap=True, unwrap=True)` just makes no sense to read.  Doing this would have to be two function calls...

```python
ag.unwrap()  # make molecules whole
ag.wrap(compound='group')  # move whole molecules into primary unit cell
```
```

NinadBhat:
@jbarnoud @richardjgowers This is ready for review.

richardjgowers:
@NinadBhat this looks like it's 99% done, I had a play and it seems to be giving the right results just needs:
- CHANGELOG entry
- docstring updates (`.. versionchanged:: `) for modified methods
- tests for the new decorator, make a dummy function and decorate it, then check that both kwargs set raises the right error

orbeckst:
Congrats @ninadbhat!

PR #2283, opened 2019-07-01
MAINT: bumpy minimum NumPy required to 1.13.3

### LINKED ISSUE #2272: Test failures due to wrong scipy version being installed
Currently, all `NUMPY_VERSION=1.10.4` tests fail on Travis.

The problem is that conda installs a too old scipy version:
```
scipy: 0.19.1-py35_blas_openblas_202 conda-forge [blas_openblas]
```
even though we require scipy>=1.0.
Later, when `$BUILD_CMD` is evaluated, the build notices the missing requirement
```
Collecting scipy>=1.0.0 (from MDAnalysis==0.19.3.dev0)
```
And a little further below:
```
ERROR: scipy 1.3.0 has requirement numpy>=1.13.3, but you'll have numpy 1.10.4 which is incompatible.
```
As expected, the tests then fail due to numpy/scipy incompatibilities.

tylerjereddy:
Fixes #2272

I adjusted the minimum required NumPy version to `1.13.3` for the relevant machinery I could find specifying `1.10.4` using `git grep` on our code base, but I did this quickly so we'll see what CI says.

I added a CHANGELOG entry too--one thought that occurred to me that is that people who put in review effort might not get acknowledged in that log, but that's a debate for another time maybe.

Also perhaps for another time, given current bandwidth, is establishing a clearer policy on major dependency version bumps, as noted in linked issue I think.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ env:
     - PYTHON_VERSION=3.7 CODECOV="true" SETUP_CMD="${PYTEST_FLAGS} --cov=MDAnalysis"
     - PYTHON_VERSION=3.6 CODECOV="true" SETUP_CMD="${PYTEST_FLAGS} --cov=MDAnalysis"
     - PYTHON_VERSION=2.7 CODECOV="true" SETUP_CMD="${PYTEST_FLAGS} --cov=MDAnalysis"
-    - NUMPY_VERSION=1.10.4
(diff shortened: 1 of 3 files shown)


tylerjereddy:
Can't remember if we have different build-time (i.e., numpy-distutils) vs. runtime NumPy version requirements, but maybe having them the same is less confusing anyway.

PR #2290, opened 2019-07-09
Adds unwrap keyword to Asphericity method

NinadBhat:
Fixes #2291 

Changes made in this Pull Request:
 - Adds unwrap keyword to Asphericity method

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [ ] CHANGELOG updated?
 - [x] Issue raised/referenced?

PR #2288, opened 2019-07-07
Adds unwrap keyword to moment of inertia

NinadBhat:
Fixes #2289 

Changes made in this Pull Request:
 - Adds unwrap keyword to moment of inertia

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?
 - [x] Issue raised/referenced?

PR #2294, opened 2019-07-09
Add unwrap keyword to radius of gyration

### LINKED ISSUE #1760: Radius of Gyration doesn't work with periodic boundaries.
### Expected behaviour

Calculating correct radius of gyration when protein is split on the periodic boundary and we set `pbc=True`

### Actual behaviour

Neither with the pbc option set or without it do I get the correct radius of gyration. This is because to correctly handle the split of a protein on periodic boundaries we would need to recombine proteins that are split (unwrap them)

[Example protein](https://gist.github.com/kain88-de/cca3fe830772931d76ddca24273dfd53) for which the correct value should be 2-4 nm.

``` python
import MDAnalysis as mda

u = mda.Universe('protein.gro')
u.atoms.radius_of_gyration(pbc=True)

```

NinadBhat:
Fixes #1760 

 - Adds unwrap keyword to radius of gyration

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/core/topologyattrs.py
+++ b/package/MDAnalysis/core/topologyattrs.py
@@ -953,15 +953,23 @@ class Masses(AtomAttr):
 
 
         .. versionchanged:: 0.8 Added *pbc* keyword
+        .. versionchanged:: 0.20.0 Added *unwrap* parameter
 
         """
         atomgroup = group.atoms
         pbc = kwargs.pop('pbc', flags['use_pbc'])
         masses = atomgroup.masses
+        unwrap = kwargs.pop('unwrap', False)
+        compound = kwargs.pop('compound', 'group')
 
-        com = atomgroup.center_of_mass(pbc=pbc)
+        com = atomgroup.center_of_mass(pbc=pbc, unwrap=unwrap, compound=compound)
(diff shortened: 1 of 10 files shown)


NinadBhat:
Will rebase after #2290 and #2288

package/MDAnalysis/core/topologyattrs.py
```diff
@@ -953,15 +953,23 @@ def radius_of_gyration(group, **kwargs):
 
 
         .. versionchanged:: 0.8 Added *pbc* keyword
+        .. versionchanged:: 0.20.0 Added *unwrap* parameter
 
         """
         atomgroup = group.atoms
         pbc = kwargs.pop('pbc', flags['use_pbc'])
         masses = atomgroup.masses
+        unwrap = kwargs.pop('unwrap', False)
```
richardjgowers: You need to update the docstring to include all keywords

NinadBhat:
Not sure why this PR is showing changes from #1991 even after rebase. Should I create a new PR?

NinadBhat:
Copy of this PR is at #2299

PR #2296, opened 2019-07-10
Adds unwrap keyword to shape_parameter

NinadBhat:
Fixes #2292 

Changes made in this Pull Request:
 - Adds unwrap keyword to shape_parameter

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [ ] CHANGELOG updated?
 - [x] Issue raised/referenced?

Issue #2293, opened 2019-07-09
Add unwrap keyoword to bsphere


NinadBhat:
**Is your feature request related to a problem? Please describe.**
"unwrap" keyword needs to be added to the bsphere method

PR #2297, opened 2019-07-10
Asv benchmarks

NinadBhat:
Fixes #

 - Add benchmarks for unwrap

PR Checklist
------------
 - [ ] Tests?
 - [x] Docs?
 - [ ] CHANGELOG updated?


NinadBhat:
@orbeckst, @richardjgowers, @jbarnoud, @micaela-matta Can you kindly look if I am adding the benchmarks correctly. I will keep on adding the benchmarks to this PR as `unwrap` is added to different methods.

benchmarks/benchmarks/ag_methods.py
```diff
@@ -24,7 +24,9 @@ class AtomGroupMethodsBench(object):
 
     def setup(self, num_atoms):
         self.u = MDAnalysis.Universe(GRO)
+        self.u_unwrap = mda.Universe(TRZ_psf, TRZ)
         self.ag = self.u.atoms[:num_atoms]
+        self.ag_unrap = self.u_unwrap.residues[0:3]
```
orbeckst: typo – "unwrap"

orbeckst:
Do you want this to be merged or do you want to keep adding to it?

My suggestion would be to add benchmarks in the PRs that implement the feature so that ASV does not choke on tests that rely on features that aren't merged yet.

Can you add a check list of the PRs that this PR depends on?

NinadBhat:
> Do you want this to be merged or do you want to keep adding to it?

I wanted to keep adding to it

> My suggestion would be to add benchmarks in the PRs that implement the feature so that ASV does not choke on tests that rely on features that aren't merged yet.
> Can you add a checklist of the PRs that this PR depends on?

At this point, this PR is not dependent on any other PRs. I plan to add other benchmarks to this PR once #2299, #2296, #2293 get merged. Then this will be ready to merge.

orbeckst:
> Am Jul 12, 2019 um 12:37 schrieb Ninad <notifications@github.com>:
> 
> I wanted to keep adding to it

Ok. Let me know when you need a final review.

benchmarks/benchmarks/ag_methods.py
```diff
@@ -202,7 +204,78 @@ def time_wrap(self, num_atoms):
...
+        """Benchmark center_of_mass calculation with
+        pbc active.
+        """
+        self.ag.center_of_mass(pbc=True)
+
+    def time_center_of_mass_unwrap(self, num_atoms):
+        """Benchmark center_of_mass calculation with
+        unwrap active.
+        """
+        self.ag_unwrap.center_of_mass(unwrap=True, compound='residues')
```
richardjgowers: These look cool, but some are only going to work from 0.20.0 onwards.  Is there a way to skip benchmarks if they're not valid? (@tylerjereddy )  Some sort of `@requires('0.20.0')` decorator?

tylerjereddy: It is certainly common for imports to be wrapped in try: except: pass to skip the benchmarks that don't have the imports.

Not sure how easy it would be to cook that up for this case though.

Perhaps more promising is that [if setup raises a `NotImplementedError`, the benchmark is marked as skipped](https://asv.readthedocs.io/en/stable/writing_benchmarks.html#setup-and-teardown-functions).

NinadBhat:
@orbeckst can you have a look at [line](https://github.com/MDAnalysis/mdanalysis/pull/2297/files#diff-06aae81f2586f827b49990403286ab40R267) and let me know if my implementation is right?

benchmarks/benchmarks/ag_methods.py
```diff
@@ -202,7 +202,94 @@ def time_wrap(self, num_atoms):
...
+    """
+
+    params = (10, 100, 1000, 10000)
+    param_names = ['num_atoms']
+
+    def setup(self, num_atoms):
+        self.u_unwrap = mda.Universe(TRZ_psf, TRZ)
+        self.ag_unwrap = self.u_unwrap.residues[0:3]
+
+        if MDAnalysis.__version__ < '0.20':
```
orbeckst: Let's not rely on simple ordering, follow https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python
```python
from packaging import version
if version.parse(MDAnalysis.__version__) < version.parse("0.20.0"):
   ...
```

(In our `setup.py` we use `LooseVersion()` but it looks as if this is pretty much deprecated. If the above does not work, maybe look at the other answer https://stackoverflow.com/a/21065570 using `from pkg_resources import parse_version`, which is guaranteed to assist)

PR #2299, opened 2019-07-11
Add unwrap to radius_of_gyration

### LINKED ISSUE #1760: Radius of Gyration doesn't work with periodic boundaries.

NinadBhat:
Fixes #1760 

 - Adds unwrap keyword to radius of gyration

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/core/topologyattrs.py
+++ b/package/MDAnalysis/core/topologyattrs.py
@@ -937,6 +937,7 @@ class Masses(AtomAttr):
         ('moment_of_inertia', moment_of_inertia))
 
     @warn_if_not_unique
+    @check_pbc_and_unwrap
     def radius_of_gyration(group, **kwargs):
         """Radius of gyration.
 
@@ -945,6 +946,10 @@ class Masses(AtomAttr):
         pbc : bool, optional
             If ``True``, move all atoms within the primary unit cell before
             calculation. [``False``]
+        unwrap : bool, optional
+            If ``True``, compounds will be unwrapped before computing their centers.
+        compound : {'group', 'segments', 'residues', 'molecules', 'fragments'}, optional
+            Which type of component to keep together during unwrapping.
 
         Note
         ----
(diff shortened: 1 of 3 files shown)


NinadBhat:
@richardjgowers this is ready to be merged.

NinadBhat:
@richardjgowers @orbeckst @jbarnoud @micaela-matta This needs to be reserved?

package/MDAnalysis/core/topologyattrs.py
```diff
@@ -953,15 +958,23 @@ def radius_of_gyration(group, **kwargs):
...
         """
         atomgroup = group.atoms
         pbc = kwargs.pop('pbc', flags['use_pbc'])
         masses = atomgroup.masses
+        unwrap = kwargs.pop('unwrap', False)
+        compound = kwargs.pop('compound', 'group')
+
+        com = atomgroup.center_of_mass(pbc=pbc, unwrap=unwrap, compound=compound)
+        if compound is not 'group':
+            com = (com * group.masses[:, None]).sum(axis=0) / group.masses.sum()
```
richardjgowers: You've added a compound parameter and not written tests for it, so this line never gets tested.  What is this line doing?

NinadBhat: it gets tested [here]( https://github.com/MDAnalysis/mdanalysis/pull/2299/files#diff-ab92738dd8fc8e62b290ddbdd5eb909dR912)

If the compound is 'residue', atomgroup.center_of_mass will return COM of each residue. This line takes the mean of each COM.

NinadBhat:
I rebased this branch with develop and now the pull request is showing changes from all other commits. I have created  #2305 in case I cant resolve this.

package/MDAnalysis/core/topologyattrs.py
NinadBhat: @richardjgowers is this sufficient or should I add some more tests?

PR #2316, opened 2019-08-05
Index selection

### LINKED ISSUE #1959: add index selection keyword
### Expected behaviour
The [selection language](https://www.mdanalysis.org/docs/documentation_pages/selections.html) should allow me to use the output from `AtomGroup.indices` for selections, which are the *0-based indices*.

### Actual behaviour
Only [`bynum`](https://www.mdanalysis.org/docs/documentation_pages/selections.html#index) is implemented as a selection keyword for indices. It uses *1-based* indices. 

See https://github.com/MDAnalysis/mdanalysis/issues/1958#issuecomment-400804285.

### Proposed solution
Implement a new selection keyword `index` that does the same as `bynum` but for 0-based indices.

develop

IAlibay:
Fixes #1959 

 - Adds new selection keyword `index`. This has the exact same behaviour as `bynum` but is 0-based rather than 1-based.

Notes:
 - I might have gone a little bit overboard in terms of duplicating the `bynum` tests for `index` in `test_atomselections`. If it's too much I can revert back to only including `test_index`.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/testsuite/MDAnalysisTests/core/test_atomselections.py
+++ b/testsuite/MDAnalysisTests/core/test_atomselections.py
@@ -179,24 +179,34 @@ class TestSelectionsCHARMM(object):
         assert_equal(len(sel), 2486)
 
     def test_around(self, universe):
-        sel = universe.select_atoms('around 4.0 bynum 1943')
-        assert_equal(len(sel), 32)
+        for atom in ['bynum 1943', 'index 1942']:
+            sel_string = 'around 4.0 ' + atom
+            sel = universe.select_atoms(sel_string)
+            assert_equal(len(sel), 32)
 
     def test_sphlayer(self, universe):
-        sel = universe.select_atoms('sphlayer 4.0 6.0 bynum 1281')
-        assert_equal(len(sel), 66)
+        for atom in ['bynum 1281', 'index 1280']:
+            sel_string = 'sphlayer 4.0 6.0 ' + atom
+            sel = universe.select_atoms(sel_string)
+            assert_equal(len(sel), 66)
 
     def test_sphzone(self, universe):
-        sel = universe.select_atoms('sphzone 6.0 bynum 1281')
-        assert_equal(len(sel), 86)
+        for atom in ['bynum 1281', 'index 1280']:
+            sel_string = 'sphzone 6.0 ' + atom
+            sel = universe.select_atoms(sel_string)
+            assert_equal(len(sel), 86)
 
     def test_cylayer(self, universe):
-        sel = universe.select_atoms('cylayer 4.0 6.0 10 -10 bynum 1281')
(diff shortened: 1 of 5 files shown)


testsuite/MDAnalysisTests/core/test_atomselections.py
```diff
@@ -392,25 +420,32 @@ def test_resid_single(self, universe):
...
+            sel_string = 'cylayer 10 20 20 -20 ' + atom
+            sel = atgp.select_atoms(sel_string)
+            assert_equal(len(sel), 1155)
 
     def test_cyzone(self, universe):
         """Cylinder zone selections with tricilinic periodicity (Issue 274)"""
         atgp = universe.select_atoms('name OW')
-        sel = atgp.select_atoms('cyzone 20 20 -20 bynum 3554')
-        assert_equal(len(sel), 1556)
+        for atom in ['bynum 3554', 'index 3553']:
```
richardjgowers: This sort of thing is best done with fixtures

testsuite/MDAnalysisTests/core/test_atomselections.py
```diff
@@ -392,25 +420,32 @@ def test_resid_single(self, universe):
 
     def test_same_coordinate(self, universe):
         """Test the 'same ... as' construct (Issue 217)"""
-        sel = universe.select_atoms("same x as bynum 1 or bynum 10")
-        assert_equal(len(sel), 12,
-                     "Found a wrong number of atoms with same x as ids 1 or 10")
-        target_ids = np.array([0, 8, 9, 224, 643, 3515,
-                               11210, 14121, 18430, 25418, 35811, 43618])
-        assert_equal(sel.indices, target_ids,
-                     "Found wrong atoms with same x as ids 1 or 10")
```
richardjgowers: Does the implicit OR work here? Ie “index 0 9”

IAlibay: It does, I've added those cases to the list of strings being tested, and also added index to `test_range_selections` in `TestImplicitOr`.

IAlibay:
Thanks for the comments @richardjgowers. Unfortunately I am not very versed in pytest fixtures, I've changed the `for` constructs for `pytest.mark.parametrize`. If you had something else in mind, I will gladly change this though.

richardjgowers:
@IAlibay yeah looks good.  The annoying thing with tests is that they will fail on the first thing wrong, so if you've got lots of things in one test method, you'll only know the first failure.  If you instead parametrize you get more detail about how/why it's failing which helps with debugging.

orbeckst:
Thank you both!

orbeckst:
Thank you both!

PR #2321, opened 2019-08-08
Unpin GSD from version 1.7.0

IAlibay:
Reverts #2320 

 - As per https://github.com/glotzerlab/gsd/issues/30 the py2.7 packages for GSD 1.8.0 have now been labelled as `broken` on conda-forge and shouldn't be picked up by travis-ci anymore.
- I have therefore unpinned the version from 1.7.0.
- <s>However, since the max version for py2.7 is now 1.7.0, I haven't changed setup.py, I can revert that too if the consensus is that it isn't necessary.</s>

PR Checklist
------------
 - [ ] Tests? - Not needed
 - [ ] Docs? - Not needed
 - [ ] CHANGELOG updated? - Not needed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,8 +27,7 @@ env:
     - MAIN_CMD="pytest ${PYTEST_LIST}"
     - SETUP_CMD="${PYTEST_FLAGS}"
     - BUILD_CMD="pip install -e package/ && (cd testsuite/ && python setup.py build)"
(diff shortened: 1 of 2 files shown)


orbeckst:
Can you try a clean `git revert` of the other commit (i.e. also revert `setup.py`) and see if the tests still pass? The less extra code the better.

Many thanks for following up with upstream!

IAlibay:
Thanks, it should now all be reverted in 5c0d166 @orbeckst

Issue #2282, opened 2019-06-24
Separate class docstrings and __init__ docstrings in API


lilyminium:
There are a few open issues on documentation, but it seems to mostly focus on tutorials rather than API reference structure -- do let me know if I've missed somewhere this would be better located. 

**Is your feature request related to a problem? Please describe.**
There's a lot of documentation in the code itself that isn't necessarily where it belongs, mostly in `Class.__init__` vs `Class` docstrings. You've noted before that you have different kinds of users, who can be mostly grouped into MD-applications people and developer people. Applications people will be looking for examples in the tutorials, while developers will be looking at the API to understand how the code itself works. Having examples and notes in `__init__()`, therefore, is unexpected for both and clutters up the code. On the flip side, the class docstring is not where I would first go to look for parameters to pass to `__init__`.

**Describe the solution you'd like**
1. Move long `Examples` etc. to tutorial `.rst` pages. Keep method docstrings limited to parameters that the methods take, and what the method returns. Keep class docstrings limited to description, short notes/examples, references, and attributes.

2. Have self-documenting code with clear names and explicit arguments instead of `kwargs` (although I see that PR #1190 is already addressing that).  Splitting up big methods into multiple smaller ones would also help interpretability. For example, `Universe.__init__` could be turned into several defined class methods: `Universe.from_files`, `Universe.from_streams`, etc. 

**Additional context**
Here's an example in the `RMSF.__init__` docstring:
```
r"""Parameters
        ----------
        atomgroup : AtomGroup
            Atoms for which RMSF is calculated
        start : int (optional)
            starting frame, default None becomes 0.
        stop : int (optional)
            Frame index to stop analysis. Default: None becomes
            n_frames. Iteration stops *before* this frame number,
            which means that the trajectory would be read until the end.
        step : int (optional)
            step between frames, default None becomes 1.
        verbose : bool (optional)
             Show detailed progress of the calculation if set to ``True``; the
             default is ``False``.

        Raises
        ------
        ValueError
             raised if negative values are calculated, which indicates that a
             numerical overflow or underflow occured


        Notes
        -----
        The root mean square fluctuation of an atom :math:`i` is computed as the
        time average

        .. math::

          \rho_i = \sqrt{\left\langle (\mathbf{x}_i - \langle\mathbf{x}_i\rangle)^2 \right\rangle}

        No mass weighting is performed.

        This method implements an algorithm for computing sums of squares while
[...]

PR #2322, opened 2019-08-08
Tidying Universe.__init__

### LINKED ISSUE #2527: Universe() should either work as Universe.empty() or raise an exception
**Expected behavior**

```python
import MDAnalysis as mda
u = mda.Universe()
print(u)
```

Prints something sensible.

**Actual behavior**

```
<ipython-input-4-f7a893c68ed4> in <module>
----> 1 print(u)

~/code/mdanalysis/package/MDAnalysis/core/universe.py in __repr__(self)
    775
    776         return "<Universe with {n_atoms} atoms>".format(
--> 777             n_atoms=len(self.atoms))
    778
    779     def __getstate__(self):

TypeError: object of type 'NoneType' has no len()
```

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`): 0.20.2-dev0
- Which version of Python (`python -V`)? Python 3.7.6
- Which operating system? macOS 10.14.5

lilyminium:
Related to #2282
Fixes #2527 

 - ~Added class methods to Universe~
 - Tidied `__init__` by having explicit kw/args and moving the topology/coordinate sorting to external functions
- ~An empty universe can no longer be created with `Universe()`. Use `Universe.empty()`~ `Universe()` creates a universe with 0 atoms
- Failing to read a topology for coordinates (because `len(coordinates)==0`, or `all_coordinates==True`) now raises a warning

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?


richardjgowers:
@lilyminium I agree that universe creation is a messy process, and there's lots of confusing options!

But I think that filenames (ie `mda.Universe('file.gro', 'traj.trr')`) always need to be the default option.  Streams work a lot like filenames I think... (I have never used this) but I think they ducktype to act like file object so these can share a "front door".

This then leaves passing the `Topology` object, which is probably not used often (I added it mostly as a dev way of building universes) so if anything, `Universe.from_Topology` should be the new class method (but probably not actually...)

I'm not sure what the no arguments use case is, especially now we have the `Universe.empty()` classmethod.  We can probably remove this.

I really like the cleaner structure of `Universe.__init__` - would it be possible to make this refactor instead something along the lines of:

```python
class Universe
    def __init__(self, args):
        if filenames_given:
            top = self._top_from_filenames(args)
        elif streams_given:
            top = self._top_from_streams(args)
        elif Topology_given:
            top = args
        # at this point we've unified all the mess into a proper Topology object
        self._generate_from_topology()

        # mess of coordinates can also be hidden away before this too?
        coordinates = figure_out_coordinates(args)
        self.load_new(coordinates)

```

So we hide the messiness of the different creation arguments, but we don't force a lot of different creation methods.

In terms of docs, I get that the `U.from_X` classmethods are nicer and self documenting.  We could add these as syntactic sugar, so they'd just route straight into `U.__init__`, but their docstrings could be more helpful than the init docstring, which is a little long.

package/MDAnalysis/core/universe.py
```diff
@@ -170,62 +229,65 @@ class Universe(object):
...
-        A CHARMM/XPLOR PSF topology file, PDB file or Gromacs GRO file; used to
-        define the list of atoms. If the file includes bond information,
-        partial charges, atom masses, ... then these data will be available to
-        MDAnalysis. A "structure" file (PSF, PDB or GRO, in the sense of a
-        topology) is always required. Alternatively, an existing
-        :class:`MDAnalysis.core.topology.Topology` instance may also be given.
-    topology_format
+    topology : `~MDAnalysis.core.topology.Topology`, str, stream, `np.ndarray`, None
+        The topology defines the list of atoms. This can be provided 
+        as filenames, streams of filenames, numpy arrays, or None for
```
richardjgowers: Mention you can pass a `mda.Topology` object here?  I think the most prominent option (ie first in lists of possibilities) has to be filename as string as this covers 90% of users, then the more involved options

package/MDAnalysis/core/universe.py
```diff
@@ -170,62 +229,65 @@ class Universe(object):
...
-        MDAnalysis. A "structure" file (PSF, PDB or GRO, in the sense of a
-        topology) is always required. Alternatively, an existing
-        :class:`MDAnalysis.core.topology.Topology` instance may also be given.
-    topology_format
+    topology : `~MDAnalysis.core.topology.Topology`, str, stream, `np.ndarray`, None
+        The topology defines the list of atoms. This can be provided 
+        as filenames, streams of filenames, numpy arrays, or None for 
+        an empty universe.
+    coordinates : str, stream, list of str, list of stream (optional)
+        Coordinates can be provided as streams of filenames either of
```
richardjgowers: Same here, reword so that the possibility of giving a path to the users' single coordinate file is the first in list of options

package/MDAnalysis/core/universe.py
```diff
@@ -170,62 +229,65 @@ class Universe(object):
...
-        the file extension [``None``] Can also pass a subclass of
+        the file extension. Can also pass a subclass of
         :class:`MDAnalysis.topology.base.TopologyReaderBase` to define a custom
         reader to be used on the topology file.
-    format
+    format: str, ``None``, default ``None``
         Provide the file format of the coordinate or trajectory file; ``None``
         guesses it from the file extension. Note that this keyword has no
         effect if a list of file names is supplied because the "chained" reader
         has to guess the file format for each individual list member.
```
richardjgowers: Tbh this is a minor bug, maybe can raise an issue that chainedreader should respect the format keyword.

package/MDAnalysis/core/universe.py
```diff
@@ -244,113 +306,67 @@ class Universe(object):
         master ConnectivityGroups for each connectivity type
 
     """
+    _topology = None
```
richardjgowers: probably best to set this in the top of `__init__`

package/MDAnalysis/core/universe.py
```diff
@@ -244,113 +306,67 @@ class Universe(object):
...
+        if not isinstance(topology, Topology) and not topology is None:
+            self.filename = _check_file_like(topology)
+            topology = _topology_from_file_like(self.filename, 
+                                                topology_format=topology_format,
+                                                **kwargs)
+
+        self._topology = topology
+
+        if topology is not None:
+            self._generate_from_topology()  # make real atoms, res, segments
```
richardjgowers: To keep the ongoing pattern, can remove `_generate_from_topology()` from the class and place it above with your other tidying functions.  Maybe it's cleaner to not have methods in the Universe namespace that can't be called after `__init__`

package/MDAnalysis/core/universe.py
```diff
@@ -244,113 +306,67 @@ class Universe(object):
...
+        
+        coordinates = _resolve_coordinates(self.filename, *coordinates,
+                                           format=format,
+                                           all_coordinates=all_coordinates)
+        
+        self.load_new(coordinates, format=format, in_memory=in_memory,
```
richardjgowers: This method however is safe to keep as it can be reused on an existing U

testsuite/MDAnalysisTests/core/test_universe.py
```diff
@@ -97,12 +97,12 @@ def test_load_trajectory_stringio(self):
         u = mda.Universe(StringIO(CHOL_GRO), StringIO(CHOL_GRO),  format='GRO', topology_format='GRO')
         assert_equal(len(u.atoms), 8, "Loading universe from StringIO failed somehow")
 
-    def test_make_universe_no_args(self):
```
richardjgowers: So this is now broken and replaced with `mda.Universe(None)`? If so CHANGELOG it

richardjgowers (changes requested):
Looking good, few comments but I like how much easier `Universe.__init__` is now

package/MDAnalysis/core/universe.py
lilyminium: We could -- but _generate_from_topology() modifies the instance, unlike the other methods. Other methods like `_gen_anchor_hash` and `_process_attr` also modify self and shouldn't really be called after `__init__` or by the user. I think it's somewhat common to have code that should only run in `__init__` wrapped in other functions to make code more readable. Moving `_generate_from_topology` outside would result in the below, which looks a bit weird.
```suggestion
            _generate_from_topology(self)  # make real atoms, res, segments
```

By the way, I think you could just move the contents of `_gen_anchor_hash` to `anchor_name`...

```python
    @property
    def anchor_name(self):
        return self._gen_anchor_hash()

    @anchor_name.setter
    def anchor_name(self, name):
        self.remove_anchor()  # clear any old anchor
        self._anchor_name = str(name) if not name is None else name
        self.make_anchor()  # add anchor again

    def _gen_anchor_hash(self):
        # hash used for anchoring.
        # Try and use anchor_name, else use (and store) uuid
        if self._anchor_name is not None:
            return self._anchor_name
        else:
            try:
                return self._anchor_uuid
            except AttributeError:
                # store this so we can later recall it if needed
                self._anchor_uuid = uuid.uuid4()
                return self._anchor_uuid
```

lilyminium: Also, correct me if I'm wrong, but I think calling it again after __init__ would have only have a real change on the universe if the topology has changed, which does seem useful in certain situations.

package/MDAnalysis/core/universe.py
lilyminium: What do you think of it now? I brought back some of the old wording

package/MDAnalysis/core/universe.py
richardjgowers: Yeah squashing in _gen_anchor_hash seems like a good idea.

I have no idea what calling `_generate_from_topology` would do post init, so probably not a good idea to do so.  It redefines what the `AtomGroup` class fundamentally is, so old AtomGroups would.... stop working?  I don't mind either way on `gen_from_top`.

package/MDAnalysis/core/universe.py
```diff
@@ -229,14 +278,14 @@ class Universe(object):
 
     Parameters
     ----------
-    topology : str, stream, `~MDAnalysis.core.topology.Topology`, `np.ndarray`, None
+    topology: str, stream, `~MDAnalysis.core.topology.Topology`, `np.ndarray`, None
```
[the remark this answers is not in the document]
lilyminium: Builds ok locally. I actually removed the spaces before because it wasn't rendering nicely (locally) when I tried to autosummarise from the UserGuide.

Screenshot 2019-09-23 at 3 33 36 AM

package/MDAnalysis/core/universe.py
```diff
@@ -243,114 +357,68 @@ class Universe(object):
...
-        # __getattr__.
-        self._instant_selectors = {}
-        # hold on to copy of kwargs; used by external libraries that
-        # reinitialize universes
-        self._kwargs = copy.deepcopy(kwargs)
-
-        # managed attribute holding Reader
-        self._trajectory = None
+    
+    def __init__(self, topology, *coordinates, all_coordinates=False,
```
richardjgowers: So Python 2.7 doesn't support this syntax, which is why we have the fugly `*args, **kwargs` syntax.  So if we want this merged before the Py3 switch, we can change this to args/kwargs (and lose readability for a while) else we can put this PR on ice for a while and merge it once we've dropped Py2.

richardjgowers (changes requested):
Needs either mushing into gross py2 syntax or putting on ice for a little

package/MDAnalysis/core/universe.py
lilyminium: The only visible changes are to the more articulate argument documentation, so it's probably more efficient to just wait until Python 2 finally goes away. Same with #2333

orbeckst:
In #2527 it came up that `Universe()` would be the "pythonic" way to create an empty universe (i.e., what most users would try first). In the discussion here the point

> An empty universe can no longer be created with `Universe()`. Use `Universe.empty()`

is included. What was the rationale for explicitly forbidding `Universe()`?

For context: #2527 states that at the moment `Universe()` produces something that is different from `Universe.empty()` which raises an exception for its own `repr()` so either `Universe()` should immediately raise an exception or do the same as `Universe.empty()`.

richardjgowers:
@orbeckst I think the signature&options and intent of `mda.Universe.empty()` is so different than `mda.Universe(file1, file2)` that it should have a separate entry point. I think 90+% of people want `mda.Universe(files)` so that should be what people see.  The power use of `.empty()` should be separate in a classmethod that has all pertinent documentation there.

I thought this PR had been merged already, oops.

orbeckst:
In this case Universe without args should raise an error — I assume that’s what’s happening in this PR (sorry, I’m lazily replying from mobile). 

The other issue that I linked above could then be closed.

lilyminium:
I changed it so that `Universe()` is the same as calling `Universe.empty(0, 0, 0)`. So in that respect it can close #2527. Now that I've done it, though, I'm not sure it's a good idea -- the universe winds up being pretty useless and you cannot add atoms as people may expect. There could be a warning?

```python
>>> import MDAnalysis as mda
>>> u = mda.Universe()
>>> print(u)
<Universe with 0 atoms>
>>> u.residues
<ResidueGroup with 0 residues>
>>> u._topology
<MDAnalysis.core.topology.Topology object at 0x105fe0590>
```

orbeckst:
> Now that I've done it, though, I'm not sure it's a good idea -- the universe winds up being pretty useless and you cannot add atoms as people may expect. There could be a warning?

Thank you for the explanation, that helps a lot.

So the Universe is more like a numpy array (unsurprisingly...) than a list. In this case I'll back-paddle on what I said before and change my opinion:

* `Universe()` should just fail — whatever the appropriate exception is when an argument is missing, I think `TypeError`. The error message can say that one should consider `Universe.empty(n_atoms, ...)`. (note that `np.array()` raises `TypeError`)
* `Universe.empty()` should require at least one argument. `np.array([])` does not complain about anything so I would also say, we don't have to warn about anything if the user sets `n_atoms=0`: it's the user's decision. The important distinction is not to allow empty arguments (as you would do with `list()`).

richardjgowers (approved):
Ok mushed the nice signatures into py2.7 friendly syntax, can't wait to change them back

PR #2326, opened 2019-08-15
Fixes velocity units and adds support for scale_factors in NCDFReader

### LINKED ISSUE #2323: NCDFReader velocities can sometimes be improperly scaled
**Overview**

So whilst I was working on a new NCRST reader, I noticed that the NCDFReader appears to not be properly handling the reading of velocities, leading to values that don't adhere to MDAnalysis velocity units standards (usually _Angstrom per AKMA time_ instead of _Angstrom per ps_).

***_A little bit of background:_***
 
As per the AMBER NetCDF convention (http://ambermd.org/netcdf/nctraj.xhtml) all trajectory data variables can have a `scale_factor` value associated with them, which is essentially a factor by which stored data should be multiplied by on read. From what I can tell, for most variables the default behaviour of AMBER MD engines is to forgo the use of `scale_factor` attributes, storing the values as they are in the NetCDF file. The one exception to this is velocities, where the values are stored in units of _Angstrom per AKMA time unit_ and given a `scale_factor` of 20.455 in order to obtain _Angstrom per ps_ values.

MDAnalysis' NCDFReader current approach to `scale_factor`, is to not handle them and throw a `NotImplementedError`. However, the current check for this is only done on the `coordinates` data variable rather than all of them (see lines 518-519), which means that velocities get read irrespective of whether or not a `scale_factor` attribute exists and are always treated as values of units _Angstrom per ps_.

***Proposed fix***

I am currently working on a temporary hot-fix for this. What I propose to do is to extend the `scale_factor` check to all NetCDF file variables, if it exists for the velocities data and is equal to 20.455 overwrite `self.units['velocity']` from _Angstrom/ps_ to _Angstrom/AKMA_. Otherwise, throw a `NotImplementedError`.

In the long run, I will see what I can do to write a clean NCDFReader implementation that fully implements handling of `scale_factor`.

**Expected behavior**

MDAnalysis universe data should always adhere to the units defined here: https://www.mdanalysis.org/mdanalysis/documentation_pages/units.html

**Actual behavior**

Velocities can by off by a factor of `scale_factor` from standard units. 

``` python

[1]: import MDAnalysis as mda

[2]: u = mda.Universe('../../ace_tip3p.parm7', 'mdcrd.nc')

# Values stored in universe (should be A/ps)
[3]: u.atoms[0].velocity
[3]: array([-0.477188  , -0.8108647 , -0.24172044], dtype=float32)

[4]: from scipy.io import netcdf

[5]: read_f = netcdf.netcdf_file('mdcrd.nc')

# This is how the data is stored in the file
[6]: read_f.variables['velocities'][0][0]
[6]: array([-0.477188  , -0.8108647 , -0.24172044], dtype=float32)

# This is the scale factor
[7]: read_f.variables['velocities'].scale_factor
[7]: 20.455

# This is the intended velocity units (post scaling)
[8]: read_f.variables['velocities'].units
[8]: b'angstrom/picosecond'

```

0.19.3-dev
3.7.3
- Which operating system?
Ubuntu 18.04 LTS

**Test files**

[MDA_VelNetCDF_Issue.zip](https://github.com/MDAnalysis/mdanalysis/files/3485835/MDA_VelNetCDF_Issue.zip)

IAlibay:
Fixes #2323 

 - Now checks for presence of `scale_factor` in all AMBER NetCDF convention variables
 - Adds support for `scale_factor` values, set in NCDFReader `__init__` and multiplied through in `_read_frame`.
 - Implements several new exception checks, mainly revolving around ensuring that units are properly adhered to.
 - Also included in this PR is part framework for the introduction of an NCRST file reader (singleframe version of the NCDFReader for .ncrst files). These can be seen in the `_NCDFGenerator` class of `test_netcdf.py`

Questions/Issues:

 - The NCDFWriter uses `degrees` units for `cell_angles` instead of the NetCDF convention `degree` (singular). I have kept support for the plural version, however, should we be thinking about changing the writer to using the singular version so that we properly adhere to the convention?
 - The NCDFWriter currently writes everything without `scale_factor` values. However, as far as I can tell all AMBER MD engines instead write out velocities with a `scale_factor` of 20.455 and the actual values in Angstrom/AKMA time units. Whilst this shouldn't be a problem for readers that properly implement the AMBER NetCDF convention, I have a suspicion that isn't actually the case for a lot of other tools/libraries. Would it be worth being more prudent by altering writing so that it matches the AMBER MD engines?

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


orbeckst:
@IAlibay thanks fore the detailed analysis:

> Questions/Issues:
>
>  *  The NCDFWriter uses degrees units for cell_angles instead of the NetCDF convention degree (singular). I have kept support for the plural version, however, should we be thinking about changing the writer to using the singular version so that we properly adhere to the convention?

Which convention do you refer to? AMBER or NetCDF? Can you link to what you're referring to?

In principle we should implement whatever the specs for the file format state. IIRC the AMBER NetCDF specs are actually pretty thorough (unlike the mess that "DCD" is). What do they say?

>   * The NCDFWriter currently writes everything without scale_factor values. However, as far as I can tell all AMBER MD engines instead write out velocities with a scale_factor of 20.455 and the actual values in Angstrom/AKMA time units. Whilst this shouldn't be a problem for readers that properly implement the AMBER NetCDF convention, I have a suspicion that isn't actually the case for a lot of other tools/libraries. Would it be worth being more prudent by altering writing so that it matches the AMBER MD engines?

Sounds sensible to me, especially once we properly support `scale_factor`. 

Such changes need to be properly advertised
- CHANGELOG in the Changes section (in addition to a fix for supporting scale_factor)
- ``.. versionchanged:: 0.20.0 + explanation similar to the above`` in the docs

package/MDAnalysis/coordinates/TRJ.py
```diff
@@ -422,7 +422,7 @@ class NCDFReader(base.ReaderBase):
     Current limitations:
 
     * only trajectories with time in ps and lengths in Angstroem are processed
-    * scale_factors are not supported (and not checked)
+    * scale_factors are supported but are not kept for writing
```
orbeckst: also support scale_factor for writing

package/MDAnalysis/coordinates/TRJ.py
```diff
@@ -448,6 +448,11 @@ class NCDFReader(base.ReaderBase):
        kwarg `delta` renamed to `dt`, for uniformity with other Readers.
     .. versionchanged:: 0.17.0
        Uses :mod:`scipy.io.netcdf` and supports the *mmap* kwarg.
+    .. versionchanged:: 0.20.0
```
orbeckst: 👍

package/MDAnalysis/coordinates/TRJ.py
```diff
@@ -458,74 +463,164 @@ class NCDFReader(base.ReaderBase):
...
+                          "http://ambermd.org/netcdf/nctraj.xhtml "
+                          "('AMBER' must be one of the token in attribute "
+                          "Conventions)".format(self.filename))
+                logger.fatal(errmsg)
+                raise TypeError(errmsg)
+        except AttributeError:
+            errmsg = "NCDF trajectory {0} is missing Conventions".format(
+                      self.filename)
+            logger.fatal(errmsg)
+            raise AttributeError(errmsg)
```
orbeckst: This should be either a `ValueError` (data contains wrong/missing information) or [MDAnalysis.NoDataError](https://www.mdanalysis.org/mdanalysis/documentation_pages/exceptions.html#MDAnalysis.exceptions.NoDataError) – opinions on which one? What do we use elsewhere?

package/MDAnalysis/coordinates/TRJ.py
```diff
@@ -458,74 +463,164 @@ class NCDFReader(base.ReaderBase):
...
+                wmsg = ("NCDF trajectory format is {0!s} but the reader "
+                        "implements format {1!s}".format(
+                         self.trjfile.ConventionVersion.decode('utf-8'),
+                         self.version))
+                warnings.warn(wmsg)
+                logger.warning(wmsg)
+        except AttributeError:
+            errmsg = "NCDF trajectory {0} is missing ConventionVersion".format(
+                      self.filename)
+            raise AttributeError(errmsg)
```
orbeckst: not AttributeError – that's a technicality, the user needs to know that the data are wrong

package/MDAnalysis/coordinates/TRJ.py
```diff
@@ -458,74 +463,164 @@ class NCDFReader(base.ReaderBase):
...
+            self.n_atoms = self.trjfile.dimensions['atom']
+            if n_atoms is not None and n_atoms != self.n_atoms:
+                errmsg = ("Supplied n_atoms ({0}) != natom from ncdf ({1}). "
+                          "Note: n_atoms can be None and t hen the ncdf value "
+                          "is used!".format(n_atoms, self.n_atoms))
+                raise ValueError(errmsg)
+        except KeyError:
+            errmsg = ("NCDF trajectory {0} does not contain atom "
+                      "information".format(self.filename))
+            raise KeyError(errmsg)
```
orbeckst: not "technical" KeyError – see above

package/MDAnalysis/coordinates/TRJ.py
```diff
@@ -458,74 +463,164 @@ class NCDFReader(base.ReaderBase):
...
+            # dimensions['frame'] == None (indicating a record dimension that
+            # can grow) whereas if read with netCDF4 I get
+            # len(dimensions['frame']) ==  10: in any case, we need to get
+            # the number of frames from somewhere such as the time variable:
+            if self.n_frames is None:
+                self.n_frames = self.trjfile.variables['time'].shape[0]
+        except KeyError:
+            errmsg = ("NCDF trajectory {0} does not contain frame "
+                      "information".format(self.filename))
+            raise KeyError(errmsg)
```
orbeckst: not "technical" KeyError – see above

package/MDAnalysis/coordinates/TRJ.py
```diff
@@ -458,74 +463,164 @@ class NCDFReader(base.ReaderBase):
              'length': 'Angstrom',
              'velocity': 'Angstrom/ps',
              'force': 'kcal/(mol*Angstrom)'}
+
     _Timestep = Timestep
 
     def __init__(self, filename, n_atoms=None, mmap=None, **kwargs):
+
+        def verify_units(eval_unit, expected_units):
```
orbeckst: Can you make this function a static method `_verify_units()` instead of hiding it?

You can then also explicitly test it.

package/MDAnalysis/coordinates/TRJ.py
```diff
@@ -458,74 +463,164 @@ class NCDFReader(base.ReaderBase):
...
+            verify_units(self.trjfile.variables['forces'].units,
+                         ('kilocalorie/mole/angstrom',))
 
         self.periodic = 'cell_lengths' in self.trjfile.variables
+        if self.periodic:
+            verify_units(self.trjfile.variables['cell_lengths'].units,
+                         ('angstrom',))
+            # Accept plural as this is what MDA writes as of 0.19.2
+            verify_units(self.trjfile.variables['cell_angles'].units,
+                         ('degree', 'degrees'))
```
orbeckst: How about we accept "degrees" until 1.0 but issue a deprecation warning right now.

Can we 
- change degrees to degree when we encounter it
- warn the user when we correct degrees
- issue a deprecation warning that from 1.0 onwards, files with cell_angles in degrees will raise an error
- only write degree

testsuite/MDAnalysisTests/coordinates/test_netcdf.py
```diff
@@ -195,6 +201,393 @@ def test_dt(self, u):
...
+        assert_almost_equal(self.coord_refs[expected],
+                            universe.atoms.positions[:3], self.prec)
+
+    @pytest.mark.parametrize('index,expected', ((0, 0), (8, 1)))
+    def test_forces(self, universe, index, expected):
+        """Here we multiply the forces by 4.184 to convert from
+        kcal to kj
+        """
+        universe.trajectory[index]
+        assert_almost_equal(self.frc_refs[expected] * 4.184,
```
orbeckst: Are the reference forces in Amber units kcal/mole/Å  and you do the manual conversion to see if MDA reads it correctly in MDA units?

testsuite/MDAnalysisTests/coordinates/test_netcdf.py
```diff
@@ -195,6 +201,393 @@ def test_dt(self, u):
...
+
+
+class TestNCDFReader3(object):
+    """NCDF trajectory with box, positions, forces and velocities
+
+    Added to address Issue #2323
+    """
+    prec = 3
+
+    coord_refs = np.array([
```
orbeckst: Add the units in comments.

testsuite/MDAnalysisTests/coordinates/test_netcdf.py
```diff
@@ -195,6 +201,393 @@ def test_dt(self, u):
...
+        [[15.249873, 12.578178, 15.191731],
+         [14.925511, 13.58888, 14.944009],
+         [15.285703, 14.3409605, 15.645962]],
+        [[14.799454, 15.214347, 14.714555],
+         [15.001984, 15.870884, 13.868363],
+         [16.03358, 16.183628, 14.02995]]
+    ], dtype=np.float32)
+
+    frc_refs = np.array([
```
orbeckst: Add the units in comments.

testsuite/MDAnalysisTests/coordinates/test_netcdf.py
```diff
@@ -195,6 +201,393 @@ def test_dt(self, u):
...
+        [[8.583388, 1.8023694, -15.0033455],
+         [-21.594835, 39.09166, 6.567963],
+         [4.363016, -12.135163, 4.4775457]],
+        [[-10.106646, -7.870829, -10.385734],
+         [7.23599, -12.366022, -9.106191],
+         [-4.637955, 11.597565, -6.463743]]
+    ], dtype=np.float32)
+
+    vel_refs = np.array([
```
orbeckst: Add the units in comments.

testsuite/MDAnalysisTests/coordinates/test_netcdf.py
```diff
@@ -195,6 +201,393 @@ def test_dt(self, u):
...
+        [[-0.5301689, -0.16311595, -0.31390688],
+         [0.00188578, 0.02513031, -0.2687525],
+         [0.84072256, 0.09402391, -0.7457009]],
+        [[-1.7773226, 1.2307, 0.50276583],
+         [-0.13532305, 0.1355039, -0.05567304],
+         [-0.6182481, 1.6396415, 0.46686798]]
+    ], dtype=np.float32)
+
+    box_refs = np.array([
```
orbeckst: Add the units in comments.

testsuite/MDAnalysisTests/coordinates/test_netcdf.py
```diff
@@ -195,6 +201,393 @@ def test_dt(self, u):
...
+        assert_almost_equal(ref, universe.trajectory.ts.dt, self.prec)
+
+    @pytest.mark.parametrize('index,expected', ((0, 0), (8, 1)))
+    def test_box(self, universe, index, expected):
+        universe.trajectory[index]
+        assert_almost_equal(self.box_refs[expected], universe.dimensions)
+
+
+class _NCDFGenerator(object):
+    """A class for generating abitrary ncdf files and exhaustively test
```
orbeckst: Interesting.

Perhaps [hypothesis](https://hypothesis.readthedocs.io/en/latest/) (which we are already using elsewhere) might also be of interest?

IAlibay:
Thanks for reviewing @orbeckst 

So with regards to changing the NCDFWriter behaviour, I was thinking it might make sense to address this in a separate PR, if you think it's appropriate? My rationale was mainly that it addresses a different feature change to what was mentioned in issue #2323.

> @IAlibay thanks fore the detailed analysis:
> 
> > Questions/Issues:
> > 
> > * The NCDFWriter uses degrees units for cell_angles instead of the NetCDF convention degree (singular). I have kept support for the plural version, however, should we be thinking about changing the writer to using the singular version so that we properly adhere to the convention?
> 
> Which convention do you refer to? AMBER or NetCDF? Can you link to what you're referring to?
> 
> In principle we should implement whatever the specs for the file format state. IIRC the AMBER NetCDF specs are actually pretty thorough (unlike the mess that "DCD" is). What do they say?
> 
> EDIT: see comments in review (read both until 1.0, auto-correct, issue deprecation warning for auto-correct degrees)

Just for reference, my "degrees" -> "degree" coment was based on http://ambermd.org/netcdf/nctraj.xhtml Section 6.2: `cell_angles`.

> 
> > * The NCDFWriter currently writes everything without scale_factor values. However, as far as I can tell all AMBER MD engines instead write out velocities with a scale_factor of 20.455 and the actual values in Angstrom/AKMA time units. Whilst this shouldn't be a problem for readers that properly implement the AMBER NetCDF convention, I have a suspicion that isn't actually the case for a lot of other tools/libraries. Would it be worth being more prudent by altering writing so that it matches the AMBER MD engines?
> 
> Sounds sensible to me, especially once we properly support `scale_factor`.
> 
> Such changes need to be properly advertised
> 
> * CHANGELOG in the Changes section (in addition to a fix for supporting scale_factor)
> * `.. versionchanged:: 0.20.0 + explanation similar to the above` in the docs

So with regards to the writer, do we want to implement a single writing format, i.e. no scale_factors for everything but `velocities`? Or would it be preferred to have either user defined scale_factors or somehow pick up whatever has been provided on read?

The latter case definitely offers more user freedom, albeit at the cost of a lot more complexity. Although, honestly can't think of a normal use case for it. In my experience AMBER style NetCDF files generally just adhere to whatever format the MD engines write in (note: I think those generated by Yank might be different, but I'm not sure).

package/MDAnalysis/coordinates/TRJ.py
IAlibay: So looking through all the other coordinate readers/writers, it looks like `NoDataError` is exclusively used by writers. `ValueError` is more commonly used for this, so as per your suggestion I'll swap all the `AttributeError` and `KeyError` instances to that.

orbeckst:
We can address the Writer in a separate PR – can you open an issue and add a comment to the code with a TODO to reference the issue? Just in case @richardjgowers creates 0.20.0 before the other PR is done...

> Or would it be preferred to have either user defined scale_factors or somehow pick up whatever has been provided on read?

That seems the best approach, especially if velocities gets the scale_factor. If scale_factor are supported for all dimensions then let's support them if we can.

package/MDAnalysis/coordinates/TRJ.py
IAlibay: As discussed, writing `scale_factor` will be handled in a future request. Extra docs have been written here and in the NCDFWriter to reflect this in 54da919.

package/MDAnalysis/coordinates/TRJ.py
IAlibay: Should be dealt with in 54da919.

package/MDAnalysis/coordinates/TRJ.py
IAlibay: As above in 54da919.

package/MDAnalysis/coordinates/TRJ.py
IAlibay: As above in 54da919.

package/MDAnalysis/coordinates/TRJ.py
IAlibay: As above in 54da919.

package/MDAnalysis/coordinates/TRJ.py
IAlibay: This should be handled in 54da919, let me know if you were thinking of something else/additional tests though.

package/MDAnalysis/coordinates/TRJ.py
IAlibay: Should be addressed in 54da919.

testsuite/MDAnalysisTests/coordinates/test_netcdf.py
IAlibay: Yep, I essentially extracted the values manually from the NetCDF file as is. Since this is your typical _pmemd_ generated trajectory, energies are in kcal instead of kJ. I have added some extra comments in 54da919 to explain this more.

testsuite/MDAnalysisTests/coordinates/test_netcdf.py
IAlibay: Should be addressed in 54da919.

testsuite/MDAnalysisTests/coordinates/test_netcdf.py
IAlibay: As above addressed in 54da919.

testsuite/MDAnalysisTests/coordinates/test_netcdf.py
IAlibay: As above, addressed in 54da919.

testsuite/MDAnalysisTests/coordinates/test_netcdf.py
IAlibay: As above, addressed in 54da919.

testsuite/MDAnalysisTests/coordinates/test_netcdf.py
IAlibay: So I had a look at hypothesis. I'll be honest, in part because I don't think I fully understand the API, I have no idea how best to employ hypothesis instead (or if it's applicable to this scenario). Any pointers here would be greatly appreciated.

orbeckst: @tylerjereddy used it – maybe he has some pointers. For this PR I am totally happy with your tests. I just thought that learning new tools that might make life easier is not a bad thing.

package/MDAnalysis/coordinates/TRJ.py
```diff
@@ -458,74 +470,168 @@ class NCDFReader(base.ReaderBase):
...
+            # singular form 'degree' agreed in the conventions. As discussed
+            # in PR #2326 from MDA 1.0 only and 'degree' will be accepted.
+            cell_angle_units = self.trjfile.variables['cell_angles'].units
+            if (cell_angle_units.decode('utf-8') == 'degrees'):
+                wmsg = ("NCDF trajectory {0} uses units of `degrees` for "
+                        "the `cell_angles` variable instead of `degree`. "
+                        "Support for non-AMBER convention units is now "
+                        "deprecated and will end in MDAnalysis version 1.0. "
+                        "Afterwards, reading this file will raise an error.")
+                warnings.warn(wmsg)
```
orbeckst: Needs to be a DeprecationWarning. We have fancy code for deprecation decorators somewhere (git grep...) but here it might just be enough to set the warning type appropriately.

To make the logger output clearer, lead wmsg with "DEPRECATED (1.0):"

package/MDAnalysis/coordinates/TRJ.py
```diff
@@ -693,7 +813,7 @@ class NCDFWriter(base.WriterBase):
         File "<string>", line 1, in <module>
         File "/scratch2/miniconda/envs/py35/lib/python3.5/site-packages/netCDF4/__init__.py", line 3, in <module>
           from ._netCDF4 import *
-      ImportError: /scratch2/miniconda/envs/py35/lib/python3.5/site-packages/netCDF4/_netCDF4.cpython-35m-x86_64-linux-gnu.so: undefined symbol: nc_inq_var_fletcher32
+      ImportError: /scratch2/minicthis will be changedonda/envs/py35/lib/python3.5/site-packages/netCDF4/_netCDF4.cpython-35m-x86_64-linux-gnu.so: undefined symbol: nc_inq_var_fletcher32
```
orbeckst: This looks odd... text inserted wrongly??

package/MDAnalysis/coordinates/TRJ.py
```diff
@@ -728,6 +848,10 @@ class NCDFWriter(base.WriterBase):
        Use fast :mod:`netCDF4` for writing but fall back to slow
        :mod:`scipy.io.netcdf` if :mod:`netCDF4` is not available.
 
+    .. TODO:
+       * Change the `cell_angles` units to `degree`, implement `scale_factor`
```
orbeckst: thanks!

package/MDAnalysis/coordinates/TRJ.py
IAlibay: My bad, seems like an erroneous middle mouse button click. Fixed in 
4b12592.

package/MDAnalysis/coordinates/TRJ.py
IAlibay: Added DeprecationWarning category in 
4b12592.

IAlibay:
@orbeckst Looks like Travis is failing in the MacOS py3.6 build.

I *think* I have it narrowed down to the recent release of pytest 5.1.0. Looks like the explicit kwargs handling was changed: https://github.com/pytest-dev/pytest/commit/683b2632b4e32ed935854e42bafd18b0178b4368

The other builds aren't pip-installing pytest, so they are still using the older 5.0.1 version of pytest.

I'll raise an issue on this.

orbeckst:
All looking good to me, let's wait for green CI. Feel free to ping me when you think it's ready for a squash-merge.

IAlibay:
@orbeckst Travis is failing on the py2.7 MacOS build for `test_libdcd.py`. It looks like it might be a non-deterministic timings issue (line 1549 in https://travis-ci.com/MDAnalysis/mdanalysis/jobs/226411993). 

I'm not sure if this is common? I could try to re-start Travis by closing/re-opening the PR if you think it might help?

orbeckst:
This does not look like something related to this PR and in principle I'd be happy to merge. However, I just restarted Job #1176.10 and I expect it come back as passing, in which case we can all be happy and merge.

orbeckst:
Thanks @IAlibay !

PR #2330, opened 2019-08-18
Addresses pytest 5.1.0 failures

### LINKED ISSUE #2329: Tests failing with pytest 5.1.0
**Expected behavior**

Issue with pytest 5.1.0 (released 15th August 2019)
`pytest test_util.py` in /lib should not fail.

**Actual behavior**

When using pytest 5.1.0, `test_deprecate` fails due to the `match_expr` keyword. This can be seen in https://travis-ci.com/MDAnalysis/mdanalysis/jobs/226110046

Conda currently provides pytest 5.0.1, so all the non-MacOS builds are not failing. But the py3.6 MacOS build installs pytest via pip, which picks up 5.1.0. I have tested 5.1.0 locally on a linux machine and the failures happen too.

**Possible fix**

It looks like the `match_expr` keyword was removed in: https://github.com/pytest-dev/pytest/commit/683b2632b4e32ed935854e42bafd18b0178b4368 (and https://github.com/pytest-dev/pytest/pull/5529)
If `match_expr` is replaced with `match` the tests will run normally.

I'm not 100% sure when match_expr first got deprecated though. It looks like it was ~ 2 years ago, so somewhere around pytest 3.X?

Iterating through older versions of pytest, it looks like the minimum version where the `match` keyword will work is 3.3.0. I should note that in pytest versions prior to that, changing the value of `match_expr` to something that should fail doesn't actually cause test failures either (tested 3.1, 3.2.0 and 3.2.5).

If it's ok I'll create a PR switching `match_expr` to `match` and bump up the minimum version of pytest to 3.3.0?

0.19.3-dev
Python 3.7.3
- Which operating system?
Ubuntu 18.04 LTS

IAlibay:
Fixes #2329 

Just creating this PR ahead of responses to #2329, if raising the pytest version is not desired at this point, I don't mind deleting the PR.

 - Raised the minimum pytest version to 3.3.0 from 3.1.2.
 - Changed the `match_expr` keyword to `match` in pytest.warns which was causing issues with pytest 5.1.0 (see #2329 ). 

PR Checklist
------------
 - [ ] Tests? - Fixes existing test
 - [ ] Docs? - No docs associated as far as I am aware
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -130,6 +130,8 @@ Deprecations
     method to be used, this is done automatically in run()
     
 Testsuite
+  * Raised minimum pytest version to 3.3.0, warns now uses the `match` 
+    argument instead of `match_expr` (Issue #2329)
   * Add tests for the new water bridge analysis (PR #2087)
 
 11/06/18 richardjgowers
--- a/testsuite/MDAnalysisTests/lib/test_util.py
+++ b/testsuite/MDAnalysisTests/lib/test_util.py
@@ -1861,7 +1861,8 @@ def test_deprecate(old_name, new_name, remove, message, release="2.7.1"):
                              new_name=new_name,
                              release=release, remove=remove,
                              message=message)
-    with pytest.warns(DeprecationWarning, match_expr="`.+` is deprecated"):
+    # match_expr changed to match (Issue 2329)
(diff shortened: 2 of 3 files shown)


richardjgowers:
merging to get the CI train rolling again, :steam_locomotive:  thanks @IAlibay

PR #2332, opened 2019-08-20
Adds arrange_closest

NinadBhat:
Fixes #

 - Adds arrange_closest method with translates atoms to the periodic image which is closest to the target position.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/core/groups.py
+++ b/package/MDAnalysis/core/groups.py
@@ -1674,6 +1674,66 @@ class GroupBase(_MutableBase):
             positions = positions[atoms._unique_restore_mask]
         return positions
 
+    def arrange_closest(self, target_position, reference='com', inplace=True):
+        """Move atoms of the group to the periodic image closest to the target position.
+
+        Parameters
+        ----------
+        target_position: numpy.ndarray
+            Co-ordinates of the target position where the atoms should move
+        reference : {'com', 'cog', None}, optional
+            If ``'com'`` (center of mass) or ``'cog'`` (center of geometry), the
+            unwrapped compounds will be shifted so that their individual
+            reference point lies within the primary unit cell.
+            If ``None``, no such shift is performed.
+        inplace : bool, optional
+            If ``True``, coordinates are modified in place.
+
+        Returns
+        -------
+        coords : numpy.ndarray
+            Unwrapped atom coordinate array of shape ``(n, 3)``.
+        """
+        atoms = self.atoms
+        unique_atoms = atoms.unique
+        positions = unique_atoms.positions
+        ref = reference.lower()
+        if ref  == 'com':
+            # Don't use hasattr(self, 'masses') because that's incredibly
+            # slow for ResidueGroups or SegmentGroups
+            if not hasattr(unique_atoms, 'masses'):
+                raise NoDataError("Cannot perform unwrap with "
+                                  "reference='com', this requires masses.")
+        elif ref != 'cog':
+            raise ValueError("Unrecognized reference '{}'. Please use one "
+                             "of 'com', 'cog'".format(reference))
+
+        if ref == 'com':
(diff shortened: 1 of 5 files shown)


NinadBhat:
@jbarnoud @zemanj Can you kindly review this?

NinadBhat:
Currently adding test for reference='com'. WIll push the changes later today.

package/MDAnalysis/core/groups.py
```diff
@@ -1674,6 +1674,66 @@ def unwrap(self, compound='fragments', reference='com', inplace=True):
...
+                raise ValueError("Cannot perform shift with "
+                                 "reference='com' because the total "
+                                 "mass of the group is zero.")
+            refpos = np.sum(positions * masses[:, None], axis=0)
+            refpos /= total_mass
+        else:  # ref == 'cog'
+            refpos = positions.mean(axis=0)
+
+        refpos = refpos.astype(np.float32, copy=False)
+        if target_position is None:
```
richardjgowers: How can `target_position` be None?

package/MDAnalysis/lib/include/calc_distances.h
```diff
@@ -356,6 +356,59 @@ static void _triclinic_pbc(coordinate* coords, int numcoords, float* box)
     }
 }
 
+static void _translate_periodic_ortho(float* reference, float* centre, float* result, float* box)
```
zemanj: would be nicer if this worked not only for single coordinates

package/MDAnalysis/lib/include/calc_distances.h
```diff
@@ -356,6 +356,59 @@ static void _triclinic_pbc(coordinate* coords, int numcoords, float* box)
...
+  inverse_box[2] = 1.0 / box[2];
+
+  dx[0] = reference[0] - centre[0];
+  dx[1] = reference[1] - centre[1];
+  dx[2] = reference[2] - centre[2];
+
+  minimum_image(dx, box, inverse_box);
+
+  for (int i=0; i<3; i++) {
+      *(result+i)= reference[i] - dx[i];
```
zemanj: With respect to the `arrange_closest()` method, aren't we doing double work here? I mean here we compute a distance vector which we use to translate a coordinate. Back in `arrange_closest()`, we take this coordinate to compute the distance vector again... or am I getting something wrong here?

package/MDAnalysis/core/groups.py
```diff
@@ -1674,6 +1674,64 @@ def unwrap(self, compound='fragments', reference='com', inplace=True):
             positions = positions[atoms._unique_restore_mask]
         return positions
 
+    def arrange_closest(self, target_position, reference='com', inplace=True):
```
zemanj: I see some use cases where this should work for multiple target positions, i.e., an array of target coordinates. (Think of the micelles example)
Furthermore, the method should have a `compound` parameter so that the shifts can be performed per compound.

package/MDAnalysis/core/groups.py
```diff
@@ -1674,6 +1674,64 @@ def unwrap(self, compound='fragments', reference='com', inplace=True):
...
+            unwrapped compounds will be shifted so that their individual
+            reference point lies within the primary unit cell.
+            If ``None``, no such shift is performed.
+        inplace : bool, optional
+            If ``True``, coordinates are modified in place.
+
+        Returns
+        -------
+        coords : numpy.ndarray
+            Unwrapped atom coordinate array of shape ``(n, 3)``.
```
zemanj: Looks like you got some unwrap remains in the docstring...

NinadBhat:
Sorry of the delay in response here. I will make the required changes over the next week.

PR #2339, opened 2019-09-04
Changes cell_angle units to degree (Issue #2327)

### LINKED ISSUE #2327: Improving the NCDFWriter behaviour (units + scale_factors)
A discussed in PR #2326, improving the NCDFReader has highlighted the fact that the NCDFWriter probably needs to be updated to ensure consistency with the AMBER NCDF convention (http://ambermd.org/netcdf/nctraj.xhtml) and the way AMBER MD engines write out certain variables by default.

Notably the following questions/issues have been raised:
 * The writer currently writes out `cell_angles` in units of `degrees` unlike the convention agreed `degree`.
 * The writer currently writes out `velocities` directly in whatever values MDAnalysis holds (usually units of Angstrom/ps). However, the standard behaviour of the AMBER MD engines appears to be to store the values in Angstrom/AKMA time units with a `scale_factor` of 20.455 in order to yield Angstrom/ps values when multiplied through. Assuming all readers properly implement the AMBER NetCDF convention and/or display scaled values to users, this should not be an issue. However in practice this is not actually the case (see this conversation for an example where things can easily become confusing: http://archive.ambermd.org/201908/0067.html).

As discussed with @orbeckst the current plan is to:
 * Change the units of `cell_angles` to `degree`, deprecate the reading of `degrees` and remove it from MDAnalysis 1.0 onwards. (Addressed by #2339 )
 * Default the writing style to that of trajectories written by the AMBER MD engines (or at least _pmemd_ as of Amber 18 + AmberTools 19).
 * Implement an option for the writer to pick up whatever `scale_factors` were used on read, and allow for user-defined values too.

**Additional context**
Work on a PR related to this will begin after #2326.

IAlibay:
Partly fixes #2327 

Apologies, this is very much a micro commit. Since I didn't manage to get this through before 0.20.0, I thought I might as well get it done asap so that the reader units deprecation makes more sense in 1.0.

A PR addressing the rest of #2327 is in progress, but considering how close we are to the start of term, it might get stalled for a few months.

 - Changes the units of the `cell_angle` variable in the NetCDF writer to `degree` instead of `degrees`.
 - Adds tests for units of files created by the NetCDF writer.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/testsuite/MDAnalysisTests/coordinates/test_netcdf.py
+++ b/testsuite/MDAnalysisTests/coordinates/test_netcdf.py
@@ -795,6 +795,30 @@ class _NCDFWriterTest(object):
                                       self.prec,
                                       err_msg="unitcells are not identical")
 
+    @pytest.mark.parametrize('var, expected', (
+        ('coordinates', 'angstrom'),
+        ('time', 'picosecond'),
+        ('cell_lengths', 'angstrom'),
+        ('cell_angles', 'degree'),
+        ('velocities', 'angstrom/picosecond'),
+    ))
+    def test_writer_units(self, universe, outfile, var, expected):
+        """Tests that the writer adheres to AMBER convention units
+        TODO: switch to input trajectory that also has force information.
+        """
+        trr = mda.Universe(GRO, TRR)
+
+        with mda.Writer(outfile, trr.trajectory.n_atoms, velocities=True,
+                        format='ncdf') as W:
+            for ts in trr.trajectory:
+                W.write_next_timestep(ts)
+
+        with netcdf.netcdf_file(outfile, mode='r') as ncdf:
+            unit = ncdf.variables[var].units.decode('utf-8')
+            assert_equal(unit, expected)
+
+        del trr
+
 
 class TestNCDFWriter(_NCDFWriterTest, RefVGV):
     pass
--- a/testsuite/MDAnalysisTests/coordinates/test_netcdf.py
+++ b/testsuite/MDAnalysisTests/coordinates/test_netcdf.py
@@ -37,7 +37,7 @@ from MDAnalysis.coordinates.TRJ import NCDFReader
 
 from MDAnalysisTests.datafiles import (PFncdf_Top, PFncdf_Trj,
                                        GRO, TRR, XYZ_mini,
-                                       PRM_NCBOX, TRJ_NCBOX)
+                                       PRM_NCBOX, TRJ_NCBOX, DLP_CONFIG)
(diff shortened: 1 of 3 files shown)


testsuite/MDAnalysisTests/coordinates/test_netcdf.py
```diff
@@ -795,6 +795,30 @@ def test_write_AtomGroup(self, universe, outfile, outtop):
...
+        with mda.Writer(outfile, trr.trajectory.n_atoms, velocities=True,
+                        format='ncdf') as W:
+            for ts in trr.trajectory:
+                W.write_next_timestep(ts)
+
+        with netcdf.netcdf_file(outfile, mode='r') as ncdf:
+            unit = ncdf.variables[var].units.decode('utf-8')
+            assert_equal(unit, expected)
+
+        del trr
```
richardjgowers: The `del` here isn't necessary because (usually) once variables fall out of scope they automatically get del'd.

testsuite/MDAnalysisTests/coordinates/test_netcdf.py
```diff
@@ -795,6 +795,30 @@ def test_write_AtomGroup(self, universe, outfile, outtop):
...
+        ('time', 'picosecond'),
+        ('cell_lengths', 'angstrom'),
+        ('cell_angles', 'degree'),
+        ('velocities', 'angstrom/picosecond'),
+    ))
+    def test_writer_units(self, universe, outfile, var, expected):
+        """Tests that the writer adheres to AMBER convention units
+        TODO: switch to input trajectory that also has force information.
+        """
+        trr = mda.Universe(GRO, TRR)
```
richardjgowers: Can we change this to a smaller universe? I think one of the TRZ files has force information (see `test_trz`)

testsuite/MDAnalysisTests/coordinates/test_netcdf.py
IAlibay: Fixed in 85e4508. I have also moved the whole thing out of the `_NCDFWriterTest` class since the test was being run twice.

testsuite/MDAnalysisTests/coordinates/test_netcdf.py
IAlibay: Unfortunately the files in `test_trz` didn't seem to have any force information, the `DLP_CONFIG` file is quite small (single frame, 216 atoms).

richardjgowers:
Cool, thanks @IAlibay , micro fixes are fine and easier to review

PR #2352, opened 2019-09-17
try changing all ints to np.intp

richardjgowers:
Fixes #

 - 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


richardjgowers:
@tylerjereddy I've tried to make MDA always generate the correct indexing int type, (`np.intp`) so that the `astype()` calls aren't ever needed (I imagine these will slow things down if they're copying things).  Hopefully this makes the azure tests have fewer failures than before....

tylerjereddy:
Looks like about an order of magnitude more failures than my current PR for 32-bit support. Happy to replace some of my changes if there are more efficient alternatives when ready, although focusing energy on resolving the remaining failures in that PR may be more helpful since there's only about 80 left or so.

Issue #2303, opened 2019-07-18
discontinue Python 2 support


orbeckst:
Python 2 reaches end of life on **1 January, 2020**, according to [PEP 373](https://www.python.org/dev/peps/pep-0373/) and https://github.com/python/devguide/pull/344 based on https://mail.python.org/pipermail/python-dev/2018-March/152348.html.

Many of our dependencies (notably numpy, see [Plan for dropping Python 2.7 support](https://docs.scipy.org/doc/numpy-1.14.0/neps/dropping-python2.7-proposal.html)) have ceased Python 2.7 support in new releases or will also drop Python 2.7 in 2020.

I know that science is rolling slowly and surely some scientific projects will continue with Python 2.7 beyond 2020. MDAnalysis has been supporting Python 2 and Python 3 now for a while. However, given how precious developer time is, I think **we also need to decide that we will stop caring for 2.7 after the official Python 2.7 drop date.**

We need to decide how to do this. I am opening this issue with the intent that it gets edited into an actionable list of items.

orbeckst:
@tylerjereddy pointed to https://github.com/numpy/numpy/pull/14086 : 
*This is being proposed by maintainers from Matplotlib, scikit-learn, IPython, Jupyter, yt, SciPy, NumPy, and scikit-image. [...] We propose only supporting versions of Python initially released in the preceding 42 months of a major or minor release of any of our projects.*

orbeckst:
Note that the proposed deprecation window of 42 months (2 minor Python releases) is much more aggressive than what we have been doing. We are used to projects that are carried out over years with the same versions of software.

(Nevertheless, dropping Python 2.7 has to happen...).

orbeckst:
Discussion thread on the mailing list at https://groups.google.com/d/msg/mdnalysis-devel/UIPEFxwh_7o/WRWA--DtDgAJ

orbeckst:
Current Plan

## Deprecating Python 2.7 support

- released 0.20.x (possibly do a 0.21.0 for features that didn't make it into 0.20.x in time)
- release 1.0.0 by the end of 2019;
  - 1.0 will be the last version of MDAnalysis that fully supports Python 2.7
  - 1.0 will be similar to 0.20 (or 0.21) (i.e., we wouldn't include major annoying API breaks, there are lot of clean-up and deprecations that we would still want to do, see Milestone 1.0)
[...]

Issue #597, opened 2015-12-16
Tests should make MDAnalysis raise exceptions


jbarnoud:
Looking at #592, and where the exclamation marks in formatted string may be an issue, I realized we rarely test the cases where exceptions are raised. These parts account for a significant part of the uncovered code. While it could seem harmless, it means we are not sure the expected exception is raised for a given error. Also, we are not sure the code to raise the exception is correct as it is the case in #592.

Tests should account for bad inputs and case MDAnalysis to raise its exceptions.

orbeckst:
Good idea. However, it's a pretty big task and although not very difficult, a bit daunting. It would be better to split it into more well defined tasks such as "tests for lib.transformations should check exception handling". This will (hopefully) make it more likely that someone picks up an issue (and it would also make it easier for new contributors to contribute).

New sub issues should reference this issue as a "parent" issue.

richardjgowers:
Yep, definitely something that needs doing when writing tests.

The guide on [writing tests](https://github.com/MDAnalysis/mdanalysis/wiki/UnitTests#writing-test-cases) probably needs to get spun out into its own page (separate from details on running them).  And then an example of how both good & bad usage should be tested.

orbeckst:
Anyone applying for GSoC and looking for a fixable issue should look at this one. Follow @fiona-naughton 's lead and identify one module or even just one class where we don't test an exception:
1. go to the [coverage report](https://coveralls.io/github/MDAnalysis/mdanalysis)
2. click on the build number and 
3. under *FILES*, click on the *All* tab 
4. pick a file name and click on it
5. search for red lines with `except` and identify file and class/function

Then find the test file under `testsuite/MDAnalysisTests` that tests the code that you identified. 
1. Raise an issue in the tracker for the specific exceptions that you want to test (and mention this issue #597)
2. Add a test or tests that raise the exception.
3. Submit a PR (referencing your new issue).

CalebMadsen:
Hey there,

I really appreciate you spelling things out for me. I went to the coverage report and just clicked the top build number (1700). I clicked through a few files and found this cone : https://coveralls.io/builds/5512450/source?filename=miniconda%2Fenvs%2Fpyenv%2Flib%2Fpython2.7%2Fsite-packages%2FMDAnalysis%2Fcore%2F__init__.py .
[...]

PR #2357, opened 2019-09-29
Improved Exception handling

joaomcteixeira:
I did not rise any issue previously to this  PR.

# Problem

Exceptions raised while handling previous caught exceptions were not raised from the caught exception nor from `None`, producing the output:

```
During handling of the above exception, another exception occurred:
```

Which in Python means *"failed to handle exception properly"* or that an *uncontrolled* exception occurred, which is not the case in MDAnalysis whatsoever.

# Solution

By capturing exception as `as e` and rising exceptions `from e` the correct output yields:

```
The above exception was the direct cause of the following exception:
```

Custom exceptions are raised `from None`, instead.

I did not entered into the detail of fine tuning `from e` and `from None` aside from using `from None` in custom exceptions because I considered such to be already changing MDAnalysis functionality/interface and I did not want to enter in such detail for this PR.

Thanks to @mariocj89 for his [talk at PyCon2019](https://www.youtube.com/watch?v=V2fGAv2R5j8) about correct exception handling.

# Tests

Tests run without failure (reds):

```
platform linux -- Python 3.7.4, pytest-5.1.3, py-1.8.0, pluggy-0.13.0
plugins: hypothesis-4.37.0
15402 passed, 84 skipped, 2 xfailed, 2 xpassed, 31871 warnings in 663.39s (0:11:03)
```

and pylint,

```
pylint 2.3.1
astroid 2.2.5
Python 3.7.3 (default, Mar 27 2019, 22:11:17) 
[GCC 7.3.0]

$ pylint --rcfile=package/.pylintrc package/MDAnalysis

------------------------------------
Your code has been rated at 10.00/10
```

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/analysis/align.py
+++ b/package/MDAnalysis/analysis/align.py
@@ -507,11 +507,12 @@ def alignto(mobile, reference, select="all", weights=None,
             # treat subselection as AtomGroup
             mobile_atoms = subselection.atoms
         except AttributeError:
-            error_ = TypeError(
-                "subselection must be a selection string, an"
-                " AtomGroup or Universe or None"
-                )
-            raise_from(error_, None)
+            raise_from(
+                TypeError(
+                    "subselection must be a selection string, an"
+                    " AtomGroup or Universe or None"
+                    ),
+                None)
 
     # _fit_to DOES subtract center of mass, will provide proper min_rmsd
     mobile_atoms, new_rmsd = _fit_to(mobile_coordinates, ref_coordinates,
--- a/package/MDAnalysis/analysis/align.py
+++ b/package/MDAnalysis/analysis/align.py
@@ -506,9 +506,9 @@ def alignto(mobile, reference, select="all", weights=None,
         try:
             # treat subselection as AtomGroup
             mobile_atoms = subselection.atoms
-        except AttributeError:
+        except AttributeError as e:
             raise TypeError("subselection must be a selection string, an"
-                            " AtomGroup or Universe or None")
+                            " AtomGroup or Universe or None") from e
 
     # _fit_to DOES subtract center of mass, will provide proper min_rmsd
     mobile_atoms, new_rmsd = _fit_to(mobile_coordinates, ref_coordinates,
@@ -1167,7 +1167,7 @@ def get_matching_atoms(ag1, ag2, tol_mass=0.1, strict=False):
                   "Try to improve your selections for mobile and reference.").format(
                       ag1.n_atoms, ag2.n_atoms)
         logger.error(errmsg)
-        raise SelectionError(errmsg)
+        raise SelectionError(errmsg) from None
 
     if np.any(mass_mismatches):
         # Test 2 failed.
--- a/package/MDAnalysis/analysis/align.py
+++ b/package/MDAnalysis/analysis/align.py
@@ -190,7 +190,7 @@ import warnings
 import logging
 
 from six.moves import range, zip, zip_longest
-from six import string_types
+from six import raise_from, string_types
 
 import numpy as np
 
@@ -506,9 +506,12 @@ def alignto(mobile, reference, select="all", weights=None,
         try:
             # treat subselection as AtomGroup
             mobile_atoms = subselection.atoms
-        except AttributeError as e:
-            raise TypeError("subselection must be a selection string, an"
-                            " AtomGroup or Universe or None") from e
+        except AttributeError:
+            error_ = TypeError(
+                "subselection must be a selection string, an"
+                " AtomGroup or Universe or None"
+                )
+            raise_from(error_, None)
 
     # _fit_to DOES subtract center of mass, will provide proper min_rmsd
     mobile_atoms, new_rmsd = _fit_to(mobile_coordinates, ref_coordinates,
@@ -1167,7 +1170,7 @@ def get_matching_atoms(ag1, ag2, tol_mass=0.1, strict=False):
                   "Try to improve your selections for mobile and reference.").format(
                       ag1.n_atoms, ag2.n_atoms)
         logger.error(errmsg)
-        raise SelectionError(errmsg) from None
+        raise_from(SelectionError(errmsg), None)
 
     if np.any(mass_mismatches):
         # Test 2 failed.
--- a/package/MDAnalysis/analysis/density.py
+++ b/package/MDAnalysis/analysis/density.py
@@ -161,7 +161,7 @@ can be used in downstream processing).
 
 from __future__ import print_function, division, absolute_import
 from six.moves import range, zip
-from six import string_types
+from six import raise_from, string_types
 
 import numpy as np
 import sys
@@ -372,12 +372,15 @@ class Density(Grid):
                 try:
                     units.conversion_factor[unit_type][value]
                     self.units[unit_type] = value
-                except KeyError as e:
-                    raise ValueError('Unit ' + str(value) + ' of type ' + str(unit_type) + ' is not recognized.') from e
-        except AttributeError as e:
+                except KeyError:
+                    raise_from(
+                        ValueError('Unit ' + str(value) + ' of type ' + str(unit_type) + ' is not recognized.'),
+                        None,
+                        )
+        except AttributeError:
             errmsg = '"unit" must be a dictionary with keys "length" and "density.'
             logger.fatal(errmsg)
-            raise ValueError(errmsg) from e
+            raise_from(ValueError(errmsg), None)
         # need at least length and density (can be None)
         if 'length' not in self.units:
             raise ValueError('"unit" must contain a unit for "length".')
@@ -485,8 +488,11 @@ class Density(Grid):
         try:
             self.grid *= units.get_conversion_factor('density',
                                                      self.units['density'], unit)
-        except KeyError as e:
-            raise ValueError("The name of the unit ({0!r} supplied) must be one of:\n{1!r}".format(unit, units.conversion_factor['density'].keys())) from e
+        except KeyError:
+            raise_from(
+                ValueError("The name of the unit ({0!r} supplied) must be one of:\n{1!r}".format(unit, units.conversion_factor['density'].keys())),
+                None,
+                )
         self.units['density'] = unit
 
     def __repr__(self):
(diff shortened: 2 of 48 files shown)


mariocj89:
This might not work here as the package needs Python2.7 support. You can use [six.raise_from](https://six.readthedocs.io/#six.raise_from) though.

joaomcteixeira:
Hi @mariocj89 , thanks very much for you comment. My fault not realizing the need to support Py2, I always fall into the trap of never considering Py2, may be that's because I actually never code with it.

I've just arrived to the project (as user and as, now, contributor) so I am not yet much deep into the details. Looking into it, I see `six` is used across the package already, though six.raise_from is never used.

I will wait for the developers to comment further before submitting any other changes.

Best and Thanks!

orbeckst:
@joaomcteixeira thank you for contributing!

Please use `six` and make the code Py 2 compatible. We will discontinue Py 2, possibly around the beginning of 2020 but until then it's important that the code runs in 2 and 3.

orbeckst:
As a first time contributor, please also add yourself to AUTHORS and add an entry to CHANGELOG.

package/MDAnalysis/analysis/align.py
```diff
@@ -506,9 +506,9 @@ def alignto(mobile, reference, select="all", weights=None,
         try:
             # treat subselection as AtomGroup
             mobile_atoms = subselection.atoms
-        except AttributeError:
+        except AttributeError as e:
```
orbeckst: We prefer variable names with multiple letters such as `err` as it is a lot easier to search for.

package/MDAnalysis/analysis/density.py
```diff
@@ -372,12 +372,12 @@ def _check_set_unit(self, u):
                 try:
                     units.conversion_factor[unit_type][value]
                     self.units[unit_type] = value
-                except KeyError:
-                    raise ValueError('Unit ' + str(value) + ' of type ' + str(unit_type) + ' is not recognized.')
-        except AttributeError:
+                except KeyError as e:
```
orbeckst: We prefer variable names with multiple letters such as `err` as it is a lot easier to search for. (`err` is also already being used elsewhere, as you saw.)

(This comment applies throughout.)

package/MDAnalysis/analysis/align.py
```diff
@@ -506,9 +506,9 @@ def alignto(mobile, reference, select="all", weights=None,
...
             raise TypeError("subselection must be a selection string, an"
-                            " AtomGroup or Universe or None")
+                            " AtomGroup or Universe or None") from e
```
orbeckst: Do you think the user needs to know why we ended up with the TypeError?

joaomcteixeira:
Hi @orbeckst 
Thanks very much for your reply,
Let me explain this issue further.

> Do you think the user needs to know why we ended up with the TypeError?

In the current implementation (master branch), the user will know about both `AttributeError` and `TypeError` in the *bad* way. Which is what happened to me.
Here's a simplification:
```python
try:
    # do something and a AttributeError raises!
    None.atoms()
except AttributeError:
    # We catch the exception
    # and we convert it to a TypeError
    raise TypeError('error to show the user')
    # and this is what the user sees
```
```
<ipython-input-1-0ef4d959c996> in <module>
      2     # do something and a AttributeError raises!
----> 3     None.atoms()
      4 except AttributeError:

AttributeError: 'NoneType' object has no attribute 'atoms'

During handling of the above exception, another exception occurred:

<ipython-input-1-0ef4d959c996> in <module>
      5     # We catch the exception
      6     # and we convert it to a TypeError
----> 7     raise TypeError('error to show the user')
      8     # and this is what the user sees

TypeError: error to show the user
```

The key problem here is the sentence:
```
During handling of the above exception, another exception occurred:
```

In Python, this strictly means that we were not able to handle the `AttributeError` properly and inside that `except` block another `Exception` occurred that we could not control and were not expecting. The user sees both `AttributeError` and `TypeError` where the latter raised uncontrollably. May be a general user won't notice the difference, but a user/developer will and get worried because the message is clear: *MDAnalysis crashed*, though the error message is user directed.

**Solution 1:**
Raise `TypeError` in a controlled manner, simply by raising it from the previous error:

```python
try:
    # do something and a AttributeError raises!
    None.atoms()
except AttributeError as err:
    # We catch the exception
    # and we convert it to a TypeError
    raise TypeError('error to show the user') from err
    # and this is what the user sees
```
```
<ipython-input-2-4cc0bdc2d9f8> in <module>
      2     # do something and a AttributeError raises!
----> 3     None.atoms()
      4 except AttributeError as err:

AttributeError: 'NoneType' object has no attribute 'atoms'

The above exception was the direct cause of the following exception:

<ipython-input-2-4cc0bdc2d9f8> in <module>
      5     # We catch the exception
      6     # and we convert it to a TypeError
----> 7     raise TypeError('error to show the user') from err
      8     # and this is what the user sees

TypeError: error to show the user
```
Hence we have:
```
The above exception was the direct cause of the following exception:
```

Whatever happened was expected to happen.

**Solution 2**
Raise `TypeError` in a controlled fashion hiding `AttributeError` to the user (notice the `from None`):

```
try:
    # do something and a AttributeError raises!
    None.atoms()
except AttributeError:
    # We catch the exception
    # and we convert it to a TypeError
    raise TypeError('error to show the user') from None
    # and this is what the user sees
```
which outputs only:
```
<ipython-input-3-ca5b07d51994> in <module>
      5     # We catch the exception
      6     # and we convert it to a TypeError
----> 7     raise TypeError('error to show the user') from None
      8     # and this is what the user sees

TypeError: error to show the user
```

**Comments**
I really think MDAnalysis should raise `from err` or `from None` and, decide based on the relevance of that traceback. This is a question that relates to API implementation already. I see from your comment that errors caught should not be presented and only the error raised should; this means that all errors, when raised within `except:` blocks, should be raised `from None`. Are these the intentions of the project?

I hope this explanation was helpful.
Looking forward your comments before applying any further changes.
Thanks for all the feedback

joaomcteixeira:
Hi,
have you considered any further on this matter? Should I actually make it py27 compatible now that it will be over after 2020? Have you decided whether to show the original exception to the user or not?

Look forward to your answer so I can complete the task,
best,

orbeckst:
@joaomcteixeira thank you for the explanation. You are right that the intention has been
>  that errors caught should not be presented and only the error raised should

(that's how it worked in Py2, see below)

## Make the change – or wait for MDA > 1.0

The Python 3 code should behave in the way the Python 2 code did, as you identified with `raise from None`.

I don't know if it easy to make this work in Python 2.7: `raise ... from None` gives `SyntaxError` in Py2.7. In which version of Python was this featured introduced?

If it is difficult/ugly to make the code work for Py 2 and Py 3 then I would wait for when we retire Python 2 support. The current plan is still somewhat in flux #2303 but we want to do a (more or less) "final" MDAnalysis 1.0 that still supports Python 2.7 with all the features that users know. We then start on 2.0, which will drop Py 2.

(The timeline says "MDA 1.0 by the end of 2019 but this might be a bit ambitious.)


## Py2 vs Py3 behavior
I had to check how Py2 behaved because I was surprised to read your explanation. MDAnalysis had started around Python 2.3... 

This is how it used to work in Python 2:
```python
Python 2.7.15 | packaged by conda-forge | (default, Jul  2 2019, 00:42:22)
Type "copyright", "credits" or "license" for more information.

IPython 5.8.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: foo = 1

In [2]: try:
   ...:     foo.bar
   ...: except AttributeError as err:
   ...:     raise TypeError("Don't use foo")
   ...:
<ipython-input-2-b01af9547409> in <module>()
      2     foo.bar
      3 except AttributeError as err:
----> 4     raise TypeError("Don't use foo")

TypeError: Don't use foo
```
but Python 3 change it:
```python
Python 3.6.7 | packaged by conda-forge | (default, Feb 28 2019, 02:16:08)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.5.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: foo = 1

In [2]: try:
   ...:     foo.bar
   ...: except AttributeError as err:
   ...:     raise TypeError("Don't use foo")
   ...:
<ipython-input-2-2127b0be432b> in <module>
      1 try:
----> 2     foo.bar
      3 except AttributeError as err:

AttributeError: 'int' object has no attribute 'bar'

During handling of the above exception, another exception occurred:

<ipython-input-2-2127b0be432b> in <module>
      2     foo.bar
      3 except AttributeError as err:
----> 4     raise TypeError("Don't use foo")
      5

TypeError: Don't use foo
```

joaomcteixeira:
Hi @orbeckst ,

Thanks for you comments. I have made the changes accordingly. As suggested [above](https://github.com/MDAnalysis/mdanalysis/pull/2357#issuecomment-536338056), I used `six.raise_from` for the Py27 and Py3 compatibility. I maintained the import statements that where already on the original code, that is, where `from six import ...` was already there I added the `raise_from`, while where `import six` was present I used `six.raise_from(....)` in the code.

The last commit passes all tests for py37 in my computer.

Travis has one broken build [1297.7](https://travis-ci.com/MDAnalysis/mdanalysis/jobs/252386721), says:

```
************* Module MDAnalysis.units

package/MDAnalysis/units.py:170: [W1618(no-absolute-import), ] import missing `from __future__ import absolute_import`
```

I think that is nothing introduced in this PR. Can you help me out, is there something I am missing?

Look forward your reply,
Thanks.

orbeckst:
Can you just add the `from __future_ ...` to units.py, please? I know that this has nothing to do with your PR. We continuously use updated versions of the CI tools including the linter and it seems a new warning was enabled. It's much easier to ok a PR when all CI comes back green.

Thanks for your massive effort!

joaomcteixeira:
Thanks @orbeckst ,
Tests already pass. I will later investigate the coverage issue. It might be because on some calls I assign a string var before raising the exception, so I will try to send the string directly to the `raise_from` in those cases. I will come back to this later and let you know the outcome.

orbeckst:
Can you please check why the tests are failing now and fix? (Please excuse my brevity, I’m travelling)

joaomcteixeira:
Hello @orbeckst 
The previous commit (8b0582a) failed because of a syntax error; I didn't run the test locally on that commit, sorry.

The latest commits clean the additional variables presented in the first commits of this PR that defined strings that were then passed to the `Exceptions` call. Now, errors are raised directly without strings being assigned beforehand. That is the only correction I can think of to clear the coverage reduction of this PR.

Yet,

Coverage reports **differ** from Appveyor and Travis. Actually, when AppVeyor finishes, coverage reports are higher than previously, while these get reduced during Travis, I do not know if these difference in coverage reports is an acknowledged and expected behavior within the project.

By analyzing the coverage reports I cannot understand where more can I operate to increase coverage percentage for the changes I propose apart from actually writing the tests of the cases where `Exceptions` are expected. Unfortunately, I currently lack the time to dedicate that much to the understanding of MDAnalysis so that I could write those tests properly in a short period of time. I have seen you have been discussion Exception coverage in #597.

Please let me know if there is something more I can help with.

orbeckst:
@joaomcteixeira thank you for doing all the extra work. I had a quick look and I don't understand what coverage thinks here either. What you did looks fine to me and I'll squash+merge in a minute.

orbeckst:
Congratulations @joaomcteixeira to your first merged commit! Thank you for all your work. I look forward to more contributions!

(I'll send you an invite to the MDAnalysis contributor group.)

joaomcteixeira:
Hello @orbeckst 
Thank you very much for your feedback and for this invitation! I have been using more and more MDAnalysis in my current research; so I am really keen to continue contributing to the project.
Best wishes,

PR #2359, opened 2019-10-01
Fixes Issue #2358

### LINKED ISSUE #2358: make_whole exits early on small box
I've got a system with a very small box (it's a crystal, so only a few bonds wide), `make_whole` currently doesn't work and returns the molecule unchanged.  This is because of an optimisation I made that assumes that any "wrong/overlong" bonds will be detectable relative to the boxsize, PR incoming...

In MDA 20.2

richardjgowers:
Fixes #2358

 - 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


package/MDAnalysis/lib/_cutil.pyx
```diff
@@ -185,6 +185,9 @@ def make_whole(atomgroup, reference_atom=None, inplace=True):
     .. versionchanged:: 0.20.0
         Inplace-modification of atom positions is now optional, and positions
         are returned as a numpy array.
+    .. versionchanged:: 0.20.2
```
orbeckst: should be 0.21.0

zemanj:
> Removing code is great.

In this case, I don't think so. I added this shortcut in PR #2189 (see [here](https://github.com/MDAnalysis/mdanalysis/pull/2189/files#diff-d1e4be9d5761ce14019f7d96bbac42eaL213)) because it made `make_whole()` substantially faster (see my [comment](https://github.com/MDAnalysis/mdanalysis/pull/2189#discussion_r252387721) for timings).
I'd only throw that out if there really is no better way of fixing the small box problem.

As I mentioned before, I don't understand why this shortcut causes `make_whole()` to return early if the box is small. I'd really appreciate it if @richardjgowers could provide a minimal example reproducing the issue!

orbeckst:
@zemanj my review was clearly only scratching the surface whereas you have more in-depth insight. You should turn your last comment into an official review — you have a clear request there and you also clearly don’t want the change to happen until your concerns are addressed. The best way to express this is is a blocking review. 

Thank you for having a look at the PR!

package/MDAnalysis/lib/_cutil.pyx
```diff
@@ -235,17 +237,6 @@ def make_whole(atomgroup, reference_atom=None, inplace=True):
...
-        is_unwrapped = True
-        for i in range(1, natoms):
-            for j in range(3):
-                if fabs(oldpos[i, j] - oldpos[0, j]) >= half_box[j]:
-                    is_unwrapped = False
-                    break
-            if not is_unwrapped:
-                break
-        if is_unwrapped:
-            return np.array(oldpos)
```
zemanj: I added this shortcut in PR #2189 (see [here](https://github.com/MDAnalysis/mdanalysis/pull/2189/files#diff-d1e4be9d5761ce14019f7d96bbac42eaL213)) because it made `make_whole()` substantially faster (see my [comment](https://github.com/MDAnalysis/mdanalysis/pull/2189#discussion_r252387721) for timings).
I'd only throw that out if there really is no better way of fixing the small box problem.

Issue #2369, opened 2019-10-22
use AtomGroup in place of selection strings


orbeckst:
**Is your feature request related to a problem? Please describe.**

Many analysis methods (`Contacts`, old and new H-bond analysis PR #2237, ...) require the user to provide _selection strings_ and then internally call `u.select_atoms(sel)`. Given the goal of MDAnalysis to enable object-oriented programming, it should be possible to provide _AtomGroups_ directly instead of selections.

**Describe the solution you'd like**

As discussed by @richardjgowers @p-j-smith @bieniekmateusz in https://github.com/MDAnalysis/mdanalysis/pull/2237#discussion_r272833451 
> This could be neatly solved with Universe.selectAtoms(AG) returning AG. This way all interfaces would automatically accept str and AG selections.

**Describe alternatives you've considered**

- We can leave everything as is, but this presents a bit of a hodge-podge of user interface choices (probably does not make @lilyminium happy, either, who would likely prefer to have a general way).
- Some cases are difficult where selection strings are actually necessary because the selection has to be recreated at each step (waterdynamics is an example, H-bond analysis is also a common case). This could be solved with updating (dynamics) AtomGroups but the user needs to be aware.

lilyminium:
#2375 implemented the following, where `type(ag) == AtomGroup`:

* `Universe.select_atoms(ag)` returns `Universe.atoms.intersection(ag)`, i.e. copy of `ag` with any duplicates and ordering lost.
* `AtomGroup.select_atoms(ag)` returns `AtomGroup.intersection(ag)`. 
* `Universe.select_atoms(ag, updating=True)` returns UpdatingAtomGroup, but the actual behaviour over frames is to return the same static atoms in the AtomGroup. Again, I believe ordering/duplicates are lost.

This was closed due to being more complicated than anticipated. The last point in particular is quite misleading, and users could lose ordered AtomGroups easily. It was also most likely that I would have to look at every analysis class and change their selection behaviour. This issue is [probably on hold until 2.0](https://github.com/MDAnalysis/mdanalysis/pull/2375#issuecomment-547187106).

orbeckst:
To add more notes from the discussion: In 2.0 we could also simply abandon selection strings for analysis user interfaces and rewrite them to just use AtomGroups. That would create a cleaner API anyway.

hmacdope:
I am also going to revive the spectre of this for 3.0 ? I think it would be good to have a more consistent interface.

[...]

PR #2375, opened 2019-10-25
select_atoms accepts atomgroups

lilyminium:
Implements #2369 

 - If `u.select_atoms` is given an Updating/AtomGroup, it returns that Updating/AtomGroup

@orbeckst commented [on the need for careful testing](https://github.com/MDAnalysis/mdanalysis/pull/2237#discussion_r272883354):

> I thought the same. Raise a separate issue. It looks like a simple change but this is such a fundamental method that it will require some careful testing. If it works consistently then it would be great. I think there was an old issue related to the "use AG where strings are allowed".

There is a simple test to check that it does return an AtomGroup, but otherwise I haven't got many ideas for others. Suggestions are welcome.

PR Checklist
------------
 - [ ] Tests? Kind of
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/testsuite/MDAnalysisTests/core/test_atomselections.py
+++ b/testsuite/MDAnalysisTests/core/test_atomselections.py
@@ -363,6 +363,15 @@ class TestSelectionsCHARMM(object):
             "resname LYS and name NZ and around 4 backbone")
         ag2 = ag.select_atoms("around 4 global backbone")
         assert_equal(ag2.indices, ag1.indices)
+    
+    @pytest.mark.parametrize('selstr',
+        ["resname LYS and name NZ",
+        "protein", "backbone", "resid 100",
+        "resid 100:105",
+        "around 4.0 bynum 1943"])
+    def test_accept_atomgroup(self, universe, selstr):
+        ag = universe.select_atoms(selstr)
+        assert ag is universe.select_atoms(ag)
 
 
 class TestSelectionsAMBER(object):
--- a/testsuite/MDAnalysisTests/core/test_atomselections.py
+++ b/testsuite/MDAnalysisTests/core/test_atomselections.py
@@ -364,21 +364,12 @@ class TestSelectionsCHARMM(object):
         ag2 = ag.select_atoms("around 4 global backbone")
         assert_equal(ag2.indices, ag1.indices)
     
-    @pytest.mark.parametrize('selstr',
-        ["resname LYS and name NZ",
-        "protein", "backbone", "resid 100",
-        "resid 100:105",
-        "around 4.0 bynum 1943"])
-    def test_accept_atomgroup(self, universe, selstr):
-        ag = universe.select_atoms(selstr)
+    def test_accept_AtomGroup(self, universe):
+        ag = universe.select_atoms('protein')
         assert ag is universe.select_atoms(ag)
     
-    @pytest.mark.parametrize('selstr',
-        ["around 4.0 bynum 1943",
-        "around 10.0 bynum 1943"
-        ])
-    def test_accept_updatingatomgroup(self, universe, selstr):
-        ag = universe.select_atoms(selstr, updating=True)
+    def test_accept_UpdatingAtomGroup(self, universe):
+        ag = universe.select_atoms('resid 100', updating=True)
         assert ag is universe.select_atoms(ag)
 
 
--- a/testsuite/MDAnalysisTests/core/test_atomselections.py
+++ b/testsuite/MDAnalysisTests/core/test_atomselections.py
@@ -364,13 +364,35 @@ class TestSelectionsCHARMM(object):
         ag2 = ag.select_atoms("around 4 global backbone")
         assert_equal(ag2.indices, ag1.indices)
     
-    def test_accept_AtomGroup(self, universe):
+    def test_returns_equal_AtomGroup_copy(self, universe):
         ag = universe.select_atoms('protein')
-        assert ag is universe.select_atoms(ag)
+        assert ag == universe.select_atoms(ag)
+        assert ag is not universe.select_atoms(ag)
     
-    def test_accept_UpdatingAtomGroup(self, universe):
-        ag = universe.select_atoms('resid 100', updating=True)
-        assert ag is universe.select_atoms(ag)
+    def test_returns_equal_UpdatingAtomGroup_copy(self, universe):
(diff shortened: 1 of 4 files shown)


testsuite/MDAnalysisTests/core/test_atomselections.py
```diff
@@ -372,6 +372,14 @@ def test_global(self, universe):
...
         ag = universe.select_atoms(selstr)
         assert ag is universe.select_atoms(ag)
+    
+    @pytest.mark.parametrize('selstr',
+        ["around 4.0 bynum 1943",
+        "around 10.0 bynum 1943"
+        ])
+    def test_accept_updatingatomgroup(self, universe, selstr):
+        ag = universe.select_atoms(selstr, updating=True)
+        assert ag is universe.select_atoms(ag)
```
richardjgowers: So the ‘is’ statement here checks that they are literally the same object, but I think there’s (possible) implementations where it returns a copy. So the test might be better with ==

testsuite/MDAnalysisTests/core/test_atomselections.py
```diff
@@ -363,6 +363,23 @@ def test_global(self, universe):
...
+        ["resname LYS and name NZ",
+        "protein", "backbone", "resid 100",
+        "resid 100:105",
+        "around 4.0 bynum 1943"])
+    def test_accept_atomgroup(self, universe, selstr):
+        ag = universe.select_atoms(selstr)
+        assert ag is universe.select_atoms(ag)
+    
+    @pytest.mark.parametrize('selstr',
+        ["around 4.0 bynum 1943",
```
richardjgowers: The around statements are gonna be slow, especially if this is a real universe. Technically any selection (eg on names) with updating will be updating.

testsuite/MDAnalysisTests/core/test_atomselections.py
```diff
@@ -363,6 +363,23 @@ def test_global(self, universe):
             "resname LYS and name NZ and around 4 backbone")
         ag2 = ag.select_atoms("around 4 global backbone")
         assert_equal(ag2.indices, ag1.indices)
+    
+    @pytest.mark.parametrize('selstr',
+        ["resname LYS and name NZ",
```
richardjgowers: Parametrise seems unnecessary

testsuite/MDAnalysisTests/core/test_atomselections.py
lilyminium: I'm not sure I understand what you mean -- can checking `isinstance` modify the UpdatingAtomGroup?

https://github.com/MDAnalysis/mdanalysis/blob/cc71ab887bf5ee4609adef48a5ba8936ee65e963/package/MDAnalysis/core/groups.py#L2841-L2842

testsuite/MDAnalysisTests/core/test_atomselections.py
lilyminium: Removed extra tests.

testsuite/MDAnalysisTests/core/test_atomselections.py
lilyminium: Changed the selection string to 'resid 100'

testsuite/MDAnalysisTests/core/test_atomselections.py
richardjgowers: The test is using “assert A is B” not “assert A == B”, these are slightly different

orbeckst: Perhaps have two tests: one that checks that you get the same object back (`is`) the other that it "is equal" (`==`). In this way we get noticed if our code does a copy somewhere where we don't expect it.

orbeckst:
The macOS job ran into the 50min time limit. I restarted it but we might be back in the bad, old days where we have to run our tests faster...

orbeckst:
In terms of tests I was thinking about MDAnalysis.analysis.rms.RMSD, ... such tests would be added to the tests of the analysis code to make sure it works with selection strings _and_ AGs.

orbeckst:
Re: CHANGELOG: has to go into a 0.21.0

lilyminium:
Do we *want* `select_atoms` to return copies of AtomGroups?

What should happen when:

```
>>> ag = u.select_atoms('around 2 protein')
>>> u.select_atoms(ag, updating=True)
```
Should this generate an error or should the static AtomGroup be turned into a selection string of indices?

lilyminium:
`u.select_atoms(ag)` now returns the intersection of `u.atoms` and `ag`. I've added an `apply` method to AtomGroup that should probably be renamed to something private like `_apply_selection`, but for now it's the same as `Selection.apply`. 

I modified rms.RMSD to accept AtomGroups and added a couple tests. I'll probably need to go through all the analysis classes individually.

richardjgowers:
Yeah the updating case is weird because it won’t work as maybe it looks like at first. Once the ‘around ...’ selection is done, the resulting AG has no memory of how/why it was made, so can’t then retroactively be made updating.

I’m not super sold on the new way to perform intersections but I’ll think on it.

This whole patch is getting messy and I’m wondering if it’s better that analysis classes learn to accept either strings or (U)AGs.

orbeckst:
I agree that switching from selection strings to only (U)AGs is the cleaner way to do this. Given that this might be more complicated than anticipated then let's put this on hold for 2.0. There's no good reason to break lots of user code just because we find this to be cleaner/neater... at least not in the short term.

I see 1.0 as the "stable/old" MDAnalysis release (minus crud like flags #782 ...).

For 2.0 we can apply our lessons learned.

orbeckst:
I can be persuaded to change my opinion, of course. 

But I learned that a good way to save time is to say "no" and stop work while you can. @lilyminium sorry to vote for killing the PR at this stage after you put the work in but this work was necessary to show potential problems. And I think that other things have higher priority.

Again, I am happy to be disagreed with.

lilyminium:
Definitely happy to focus on other things. Not much work put in 😄 Analyses like rms.RMSD already accept both Universes and AtomGroups anyway, so it did feel a little unnecessary there. Should this be closed or left open?

orbeckst:
I’d close the PR (but don’t delete it) and leave the issue open. Add a quick note to the issue summarizing the discussion for future generations/future us, please. Thanks!

PR #774, opened 2016-03-16
DATAWriter, velocity unit conversion, unit test

khuston:
Adds feature in issue #763 

### DATAWriter
Writes selection at current trajectory frame to file, including sections
Atoms, Masses, Velocities, Bonds, Angles, Dihedrals, and Impropers (if
these are defined). Atoms section is written in the "full" sub-style if
charges are available or "molecular" sub-style if they are not.
Molecule id in atoms section is set to to 0.

No other sections are written to the DATA file.
As of this writing, other sections are not parsed into the topology
by the `DATAReader`.

If the selection includes a partial fragment, then the outputted DATA
file will be invalid, because it will describe bonds between atoms
which do not exist in the Atoms section.

By default the writer assumes "conventional" or "real" LAMMPS units
where length is measured in Angstroms and velocity is measured in
Angstroms per femtosecond. If other units are desired, they must be
specified.

Raises ValueError if atom types are not convertable to integers or if
atoms of the same type don't all have the same mass.

An AttributeError will be raised if the atom group doesn't have masses.

### Tests
I added tests to make sure topology attributes of the written file
(after being read again) match the topology attributes read from
the original file. Checks types, bonds, positions, velocities, etc.

There was a problem with test data mini.data because it had atom type
3 which had undefined mass, so I changed the atom type to 1.

### DATAParser change
The DATAParser before was not converting velocity units, so I added
default units Angstroms/fs which are the "real" units in LAMMPS.

PR Checklist
------------
 - [X] Tests?
 - [x] Docs?
 - [ ] CHANGELOG updated?
 - [X] Issue raised/referenced?

Issue #763, opened 2016-03-11
LAMMPS.DATAWriter and bond types etc.


khuston:
I would like to write a `LAMMPS.DATAWriter`. Here is a use case:

```
u = mda.Universe('polymer.data', 'coil.dcd')
for ts in u.trajectory:
    ... analyze the trajectory, decide which frame I'd like to write out as a data file ...
    if save_this_frame:
        with mda.Writer('configuration_xx.data') as W:
            W.write(u.atoms)
```

There are a few concerns I can think of:
1. I would need to parse bond, angle, dihedral, and improper types with `DATAParser`.
2. Would these types be stored as `Bondtypes`, `Angletypes`, `Dihedraltypes`, `Impropertypes`?
3. Would the current `bond.type` etc. that returns an array of the constituent atom types be overridden? Or would these be some separate attribute of bonds, angles, dihedrals, and impropers?
4. LAMMPS data files can also have numerous coefficients sections. In principle someone might want to read that information and write it again to the DATA file later. Personally, I don't use those sections and would happily ignore them.
5. If someone tries to write a LAMMPS data file based on a topology which doesn't have `Bondtypes`, `Angletypes`, `Dihedraltypes`, `Impropertypes`, should an error result?

I've tried to catch up with the issue-363 discussion, but I've probably missed a lot, so I apologize if this is misguided.

richardjgowers:
Hey Kyle

This would be awesome.  

I think you're right that you'll need to override the current `.type` behaviour somehow.

The current 363 behaviour for atoms is something like sticking whatever we find in a topology file to atoms.  It would be nice to extend this philosophy to bonds too.. but I'll have to think about what this would mean in total.

In the short term, you should be able to read the bond types into a dictionary of `{(atom indices): bond type}`, and then write a Writer which accepts this.  I think once we've got a prototype Writer it'll be much clearer what we need to make it work seamlessly.

So tl;dr
1 - yes, but you can do a hack job in the short term
2 - ideally in the long run stuck to the bond objects
3 - same as above
4 - eventually it'd be nice to squeeze everything, but lets just get this working
5 - should error if the resulting file doesn't work (`NoDataError: No bond types specified, specify them by ...`)

Maybe hack together a Writer and we'll see what needs doing

khuston:
[...]

PR #2381, opened 2019-10-28
wrote ITPParser

lilyminium:
I'd like to read some ITP files.

 - Added parser for ITP files

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/CHANGELOG
```diff
@@ -27,6 +27,7 @@ Fixes
 Enhancements
   * Expanded selection wildcards to the start and middle of strings (Issue #2370)
   * Added type checking and conversion to Connection TopologyAttrs (Issue #2373)
+  * Added ITPParser
```
richardjgowers: Expand this Gromacs ITP topology or whatever

package/MDAnalysis/topology/ITPParser.py
```diff
@@ -0,0 +1,226 @@
...
+    - resids
+    - resnames
+    - segids
+    - moltypes
+    - bonds
+    - angles
+    - dihedrals
+    - impropers
+
+    .. _ITP: http://www.charmm.org/documentation/c35b1/struct.html
```
richardjgowers: Charmm 🤨

package/MDAnalysis/topology/ITPParser.py
```diff
@@ -0,0 +1,226 @@
...
+        impropers = []
+
+        segid = 'SYSTEM'
+        molnum = 1
+
+        section = None
+
+        # Open and check itp validity
+        with openany(self.filename) as itpfile:
+            header = re.compile(r'\[\s*(?P<section>\w+)\s*\]')
```
richardjgowers: Fwiw I think usually regex is slower than just checking for a left bracket and it looks like this is being called on all lines. I guess itps are probably gonna be small though

package/MDAnalysis/topology/ITPParser.py
lilyminium: Oops!

package/MDAnalysis/topology/ITPParser.py
lilyminium: Changed it anyway, and updated CHANGELOG

lilyminium:
I added bond/angle/dihedral/improper types. They are strings to match what appears to be the behaviour of #763 and #774.

testsuite/MDAnalysisTests/datafiles.py
```diff
@@ -473,5 +476,12 @@
 PDB_rama = resource_filename(__name__, 'data/19hc.pdb.gz')
 PDB_janin = resource_filename(__name__, 'data/1a28.pdb.gz')
 
+ITP_ala = resource_filename(__name__, 'data/gromacs_ala10.itp')
```
richardjgowers: this is a sneaky gotcha, but new data files need to get added to the testsuite setup.py so they get bundled up into the tarballs of the release. There’ll be a bunch of glob patterns but I don’t think the new itps will get caught

testsuite/MDAnalysisTests/datafiles.py
```diff
@@ -473,5 +476,12 @@
...
+XTC_ala = resource_filename(__name__, 'data/gromacs_ala10.xtc')
+
+ITP_atb = resource_filename(__name__, 'data/atb_hema.itp')
```
richardjgowers: Also these files are unused so either use them or lose them. You can also compress the files and MDA should just work still

package/MDAnalysis/topology/ITPParser.py
```diff
@@ -0,0 +1,237 @@
...
+   :members:
+   :inherited-members:
+
+"""
+from __future__ import absolute_import, division
+from six.moves import range
+
+import logging
+import functools
+import re
```
richardjgowers: Unused imports?

lilyminium:
@richardjgowers is MDAnalysis interested in topology writers? (I would also like to write the ITPs out). All the Writers seem to be associated with coordinates.

package/doc/sphinx/source/documentation_pages/topology_modules.rst
```diff
@@ -31,6 +31,7 @@ topology file format in the *topology_format* keyword argument to
    topology/GMSParser
    topology/GROParser
    topology/HoomdXMLParser
+   topology/ITPParser
```
richardjgowers: You'll need to add a .rst stub for this new module

richardjgowers:
@lilyminium yeah I think topology writers could make sense... Though they often contain a superset of information than a Universe (ie force field information)

orbeckst:
Topology writers are a thorny issue. It's really hard to get them right – do you believe that the written ITPs will be good enough to run in Gromacs?

There are ParmEd and intermol, which contain an awful lot of knowledge about force fields and topologies. I'd rather leave topology writing/hacking to them instead of having to deal with the problems.

orbeckst:
I'd rather look into the question how MDAnalysis can call ParmEd to write a topology.

lilyminium:
Yes... last I checked (last year) ParmEd refused to read GROMACS topology files with G96 type information (e.g. quartic bond potentials), which also happen to be the most common ones for native GROMOS force fields. (I haven't looked at InterMol yet). Converting quartic potentials to harmonics for another force field does sound like a terrible idea, but being able to load in a topology, select certain atoms, and write out the applicable bonds/angles/dihedrals/impropers to the same format is quite useful. Or even renumbering atoms, etc. 

Faithfully reproducing ITP files in MDAnalysis would be a pain because you can have multiple entries for the same bond, etc. in an ITP file that work additively. (That being said I've found the ITP file spec reasonably forgiving regarding whitespace and so on). 

Using ParmEd from MDAnalysis would be interesting.

PR #2391, opened 2019-11-05
Added correct parsing for huge GRO boxes

### LINKED ISSUE #2371: Huge box sizes in GRO-files
I currently facing the problem of HUGE box sizes in my GRO files. Due to the [current implementation](https://github.com/MDAnalysis/mdanalysis/blob/9372d093acebb78bc5b1a2b4dba815a4e3281b04/package/MDAnalysis/coordinates/GRO.py#L203) they can not be parsed at all. To fix this one can just replace

```python
unitcell = np.float32(line.split())
```

by 

```python
split = re.split("(\......)", line)
unitcell = np.float32([i+j for i, j in zip(split[::2], split[1::2])])
```
since the box dimensions always contain 5 decimals places.

### Expected behaviour

Consider the following GRO of a cubic box with a box length of 40000 nm.

```
huge cubic box
    1
    1SOL      H    1   0.000   0.000   0.000
40000.0000040000.0000040000.00000
```

This should give a dimension array of `[4.e+05 4.e+05 4.e+05 9.e+01 9.e+01 9.e+01]`

### Actual behaviour

ValueError during parsing
```python
ValueError: could not convert string to float: '40000.0000040000.0000040000.00000'
```

``` python
import MDAnalysis as mda

u = mda.Universe("cubic.gro")
```

0.20.2-dev0

PicoCentauri:
Fixes #2371 

GROReader and GROWriter now can handle boxes with box vectors >1000nm. Follows the 
changes made in the GROMACS Suite (see: https://gerrit.gromacs.org/c/gromacs/+/14117)

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/coordinates/GRO.py
```diff
@@ -200,7 +201,7 @@ def _read_first_frame(self):
             for pos, line in enumerate(grofile, start=1):
                 # 2 header lines, 1 box line at end
                 if pos == n_atoms:
-                    unitcell = np.float32(line.split())
+                    unitcell = np.float32(re.findall(r"(\d+\.\d{5})", line))
```
richardjgowers: I think this is probably a little strict, eg it won't parse "`1.0 1.0 1.0`" - There's probably people writing GRO files to less than 5dp precision.  I think something like: `(\d+\.?\d{0,5}\s*)` might work?

PicoCentauri: I'm not an expert in regular expression, sorry. Your suggestion works. Thanks

richardjgowers:
thanks again @PicoCentauri

PR #2404, opened 2019-11-16
ParmEd i/o

lilyminium:
https://github.com/MDAnalysis/mdanalysis/pull/2381#issuecomment-550104207
> I'd rather look into the question how MDAnalysis can call ParmEd to write a topology.

Library interoperability is convenient. There are a few issues:

- ParmEd doesn't try to guess elements if they aren't read. The current ParmEdParser does. This means, e.g., the PSF test file can't roundtrip. This could be an advantage for when you want a Parmed structure with the right elements

- ParmEd's guessing isn't great. It thinks that NA+ has an atomic number of 7, for example. 

- MD engines and ParmEd support multiple bonds etc., while MDAnalysis doesn't. Currently ParmEdParser turns these into a tuple of bond(etc.)types

**Example**
```python

In [2]: import parmed as pmd

In [3]: from MDAnalysis.tests.datafiles import *

In [4]: pgro = pmd.load_file(GRO)

In [5]: mgro = mda.Universe(pgro)

In [6]: ogro = mgro.atoms.write(file_format='PARMED')
```

 - Added ParmEdParser
 - Added ParmEdReader
 - Added ParmEdWriter
 - Added UreyBradleys + CMaps

PR Checklist
------------
 - [x] Tests? (some)
 - [x] Docs?
 - [x] CHANGELOG updated?

Still missing tests for:
- [x] Writing a subset of atoms
- [x] Writing an MDAnalysis Universe read from a non-Parmed source
- [x] Anything with Urey-bradley angles?


package/MDAnalysis/coordinates/ParmEd.py
```diff
@@ -0,0 +1,264 @@
...
+MDA2PMD = {
+    'tempfactor': 'bfactor',
+    'gbscreen': 'screen',
+    'altLoc': 'altloc',
+    'nbindex': 'nb_idx',
+    'solventradius': 'solvent_radius',
+    'id': 'number'
+}
+
+SYMB2Z = {'H': 1,                                                   'He': 2,
```
richardjgowers: this should probably go somewhere like `topology.tables` or something

.appveyor.yml
```diff
@@ -9,8 +9,8 @@ cache:
 
 environment:
     global:
-        CONDA_CHANNELS: conda-forge
```
richardjgowers: I'd rather move the parmed package to conda-forge

testsuite/MDAnalysisTests/coordinates/test_parmed.py
```diff
@@ -0,0 +1,153 @@
...
+            "ParmEdReader failed to get unitcell dimensions from ParmEd")
+    
+    def test_coordinates(self):
+        up = self.universe.atoms.positions
+        rp = self.ref.atoms.positions
+        assert_array_almost_equal(up, rp, decimal=3)
+    
+
+class TestParmEdReaderPDB(BaseTestParmEdReader):
+    __test__ = True
```
richardjgowers: shouldn't need `__test__`

.appveyor.yml
orbeckst: I agree.

Can both be used together, with conda-forge higher priority? IIRC, conda's package resolving process gets slower and slower with more channels so we really try to minimize time there.

package/MDAnalysis/coordinates/ParmEd.py
```diff
@@ -0,0 +1,264 @@
...
+
+class ParmEdWriter(base.WriterBase):
+    """Convert MDAnalysis AtomGroup or Universe to ParmEd Structure
+    """
+
+    format = 'PARMED'
+    units = {'time': None, 'length': 'Angstrom'}
+
+    def __init__(self, n_atoms=None, convert_units=None, **kwargs):
+        """Set up a ParmEdWriter with a precision of 3 decimal places.
```
orbeckst: Why "precision of 3 decimal places" ?

package/MDAnalysis/coordinates/ParmEd.py
```diff
@@ -0,0 +1,264 @@
...
+# N. Michaud-Agrawal, E. J. Denning, T. B. Woolf, and O. Beckstein.
+# MDAnalysis: A Toolkit for the Analysis of Molecular Dynamics Simulations.
+# J. Comput. Chem. 32 (2011), 2319--2327, doi:10.1002/jcc.21787
+#
+
+"""ParmEd structure reader --- :mod:`MDAnalysis.coordinates.ParmEd`
+================================================================
+
+Reads coordinates data from a ParmEd Structure.
+
```
orbeckst: Minimal usage example would be really good.

.appveyor.yml
orbeckst: Btw, omnia requires conda-forge https://github.com/ParmEd/ParmEd/blob/7e411fd03c7db6977e450c2461e065004adab471/devtools/ci/travis/install.sh#L25-L26

orbeckst: The full dependency list from https://github.com/ParmEd/ParmEd/blob/7e411fd03c7db6977e450c2461e065004adab471/devtools/ci/travis/install.sh#L25-L39
```bash
    if [ -z "$MINIMAL_PACKAGES" ]; then
        # Install all prerequisites
        conda create -y -n myenv python=$PYTHON_VERSION \
            numpy scipy pandas nose openmm coverage nose-timer \
            netCDF4
        conda update -y -n myenv --all
        conda install -y -n myenv rdkit==2018.09.1
        conda install -y -n myenv boost==1.69.0
        conda install -y -n myenv nglview
        conda install -y -n myenv ambertools=18 -c ambermd
        conda install -y -n myenv networkx
        conda install -y -n myenv lxml
```
Note that it contains ambertools from the ambermd channel, rdkit, and boost as some heavy dependencies.

The minimal list https://github.com/ParmEd/ParmEd/blob/7e411fd03c7db6977e450c2461e065004adab471/devtools/ci/travis/install.sh#L41-L43 is essentially numpy:
```bash
        conda create -y -n myenv python=$PYTHON_VERSION numpy nose coverage \
            nose-timer
```

I don't really want any MDA user have to install these massive dependencies. But I also don't think one should create a crippled ParmEd conda-forge package (which does not pull all its dependencies) as this is a bad user experience for anyone who installs ParmEd for its own sake.

Should we make the ParmEdReader/Writer fail gracefully if no parmed package is installed? This would, however, go against our [importing rules](https://github.com/MDAnalysis/mdanalysis/wiki/Style-Guide#importing-modules ) for [modules in the core](https://github.com/MDAnalysis/mdanalysis/wiki/Style-Guide#list-of-core-module-dependencies).

orbeckst: Regarding "guidelines" vs "rules": https://www.youtube.com/watch?v=WJVBvvS57j0 – should our rules be more something that we would call "guidelines"?

mattwthompson: Yes and (painfully) yes

`omnia` is a pretty well-maintained channel, although fairly specialized to the projects of the Pande, Chodera, Mobley, etc. labs. I'm not sure who is responsible for it but it also happens to be packaged on [`conda-forge`](https://anaconda.org/conda-forge/parmed). Compared to other dependencies, ParmEd is updated quite infrequently, so the risk of upstream complications should be fairly low (of course, there's also the downside of it possibly being slow to update to newer Python versions; not sure how that fits into your plans).

package/MDAnalysis/coordinates/ParmEd.py
```diff
@@ -0,0 +1,264 @@
...
+          'Pr': 59, 'Nd': 60, 'Pm': 61, 'Sm': 62, 'Eu': 63, 'Gd': 64, 'Tb': 65, 'Dy': 66, 
+          'Ho': 67, 'Er': 68, 'Tm': 69, 'Yb': 70, 'Lu': 71, 'Hf': 72, 'Ta': 73, 'W': 74, 
+          'Re': 75, 'Os': 76, 'Ir': 77, 'Pt': 78, 'Au': 79, 'Hg': 80, 'Tl': 81, 'Pb': 82, 
+          'Bi': 83, 'Po': 84, 'At': 85, 'Rn': 86, 'Fr': 87, 'Ra': 88, 'Ac': 89, 'Th': 90, 
+          'Pa': 91, 'U': 92, 'Np': 93, 'Pu': 94, 'Am': 95, 'Cm': 96, 'Bk': 97, 'Cf': 98, 
+          'Es': 99}
+
+
+class ParmEdWriter(base.WriterBase):
+    """Convert MDAnalysis AtomGroup or Universe to ParmEd Structure
```
orbeckst: That's a very interesting interoperability idea!

package/MDAnalysis/core/topologyattrs.py
```diff
@@ -1286,6 +1286,82 @@ class AltLocs(AtomAttr):
     def _gen_initial_values(na, nr, ns):
         return np.array(['' for _ in range(na)], dtype=object)
 
+class GBScreens(AtomAttr):
```
orbeckst: Being able to add all these attributes shows off the flexibility of the topology system!

package/MDAnalysis/topology/ParmEdParser.py
```diff
@@ -0,0 +1,317 @@
...
+===================
+
+Converts a ParmEd_ structure into a Topology.
+
+.. _ParmEd: https://parmed.github.io/ParmEd/html/index.html
+
+Example
+-------
+
+>>> import parmed as pmd
```
orbeckst: This is a writing example. Can you also add a reading example?

package/MDAnalysis/topology/ParmEdParser.py
```diff
@@ -0,0 +1,317 @@
...
+
+"""
+ParmEd topology parser
+===================
+
+Converts a ParmEd_ structure into a Topology.
+
+.. _ParmEd: https://parmed.github.io/ParmEd/html/index.html
+
+Example
```
orbeckst: I *think* such a heading will not render nicely with numpy docs/napoleon. We normally rewrite it to not be the word "Example" or "Examples", such as "Example use" or something more meaningful.

(I thought we had a note regarding this problem at https://github.com/MDAnalysis/mdanalysis/wiki/WritingDocumentation#numpy-style-gotchas but we don't so I am not 100% sure if this is still a problem.)

package/MDAnalysis/topology/ParmEdParser.py
```diff
@@ -0,0 +1,317 @@
...
+    Impropers,
+    CMaps
+)
+from ..core.topology import Topology
+
+logger = logging.getLogger("MDAnalysis.topology.ParmEdParser")
+
+# >>> import mendeleev as mv
+# >>> {i:mv.element(i).symbol for i in range(1, 100)}
+Z2SYMB = {1: 'H',                                                    2: 'He',
```
orbeckst: Can't you generate this table from `coordinates.ParmEd.SYMB2Z` ? 

Furthermore, perhaps just putting the periodic table into `lib.tables` might be a better place anyway?

testsuite/MDAnalysisTests/coordinates/test_parmed.py
```diff
@@ -0,0 +1,153 @@
...
+    GRO,
+    PDB_small,
+    PRM
+)
+
+class TestParmEdReaderGRO:
+    __test__ = True
+    ref_filename = GRO
+
+    universe = mda.Universe(pmd.load_file(GRO))
```
orbeckst: Add this as an example to the ParmEdReader and Parser!

testsuite/MDAnalysisTests/coordinates/test_parmed.py
```diff
@@ -0,0 +1,153 @@
...
+                         'number', 'altloc', 
+                         'atomic_number'):
+                ra = getattr(r, attr)
+                oa = getattr(o, attr)
+                assert ra == oa
+            
+            for attr in ('mass', 'charge', 'occupancy', 'rmin', 'epsilon'):
+                ra = getattr(r, attr)
+                oa = getattr(o, attr)
+                assert abs(ra-oa) < self.prec
```
orbeckst: Don't roll your own – just use `assert_almost_equal`

testsuite/MDAnalysisTests/coordinates/test_parmed.py
```diff
@@ -0,0 +1,153 @@
...
+        assert_array_almost_equal(
+            self.universe.trajectory.ts.dimensions, 
+            self.ref.trajectory.ts.dimensions,
+            self.prec,
+            "ParmEdReader failed to get unitcell dimensions from ParmEd")
+    
+    def test_coordinates(self):
+        up = self.universe.atoms.positions
+        rp = self.ref.atoms.positions
+        assert_array_almost_equal(up, rp, decimal=3)
```
orbeckst: We stopped using `assert_array_almost_equal` and replaced it with `assert_almost_equal` — although I forget the reason...

.appveyor.yml
hainm: @orbeckst parmed only requires `numpy`. And it's pretty easy to install via pip (or conda if exclude the requirements).

orbeckst: That's good to know.

Thanks for comments, @mattihappy and @hainm !

package/MDAnalysis/coordinates/ParmEd.py
lilyminium: I have no idea. Leftover from something I copied? No precision is ever considered, anyway. Removed.

package/MDAnalysis/topology/ParmEdParser.py
lilyminium: I think normally the heading used in MDAnalysis is just Example/s? It displays alright for [Universe](https://www.mdanalysis.org/mdanalysis/documentation_pages/core/universe.html#MDAnalysis.core.universe.Universe.add_TopologyAttr).

.appveyor.yml
richardjgowers: Looks like the conda-forge recipe only does linux for now, I raised an issue

orbeckst:
I _think_ converters are a good idea, especially on the writing side. Nevertheless, one could also argue that anything that takes in a data structure/class with a given API should be called a converter instead of a Reader. Once this is in, we could classify MMTF and MemoryReader as a converters as they all take a data structure as input as opposed to a filename.

package/MDAnalysis/core/groups.py
```diff
@@ -3032,6 +3032,82 @@ def improper(self):
...
+            An instance of the structure type from another library.
+        
+        Raises
+        ------
+        TypeError:
+            No converter was found for the required format
+        
+        """
+        try:
+            writer = _CONVERTERS[format]
```
richardjgowers: Rather than exposing (and pinning down) how we're doing the magic around `_CONVERTERS`, it's better to have a `get_converter_for` method like writers have, where later we can put more/different magic if things change.

lilyminium: Thanks, moved.

package/MDAnalysis/core/groups.py
```diff
@@ -3032,6 +3032,80 @@ def improper(self):
...
+
+        Example
+        -------
+
+        .. code-block:: python
+
+            import MDAnalysis as mda
+            from MDAnalysis.tests.datafiles import GRO
+            u = mda.Universe(GRO)
+            parmed_structure = u.atoms.convert_to('PARMED')
```
orbeckst: Add sentence stating that `parmed_structure` is now a [ParmEd.structure.Structure](https://parmed.github.io/ParmEd/html/structobj/parmed.structure.Structure.html) instance.

Link to the ParmEd doc.

Add an entry for https://parmed.github.io/ParmEd/html/ to the intersphinx links
https://github.com/MDAnalysis/mdanalysis/blob/972a8b1893b5e353b9d796b15d6fcdd022a93f00/package/doc/sphinx/source/conf.py#L329-L336

package/MDAnalysis/topology/ParmEdParser.py
```diff
@@ -0,0 +1,303 @@
...
+Example
+-------
+
+::
+
+    import parmed as pmd
+    import MDAnalysis as mda
+    from MDAnalysis.tests.datafiles import PSF
+    universe = mda.Universe(pmd.load_file(PSF))
+
```
orbeckst: See comments above – say something about the resulting object (ParmEd.structure.Structure).

package/MDAnalysis/topology/__init__.py
```diff
@@ -312,3 +312,4 @@
 from . import GSDParser
 from . import MinimalParser
 from . import ITPParser
+from . import ParmEdParser
```
orbeckst: newline

package/MDAnalysis/topology/tables.py
```diff
@@ -377,3 +377,20 @@ def kv2dict(s, convertor=str):
...
+          27: 'Co', 28: 'Ni', 29: 'Cu', 30: 'Zn', 31: 'Ga', 32: 'Ge', 33: 'As', 34: 'Se', 
+          35: 'Br', 36: 'Kr', 37: 'Rb', 38: 'Sr', 39: 'Y',  40: 'Zr', 41: 'Nb', 42: 'Mo', 
+          43: 'Tc', 44: 'Ru', 45: 'Rh', 46: 'Pd', 47: 'Ag', 48: 'Cd', 49: 'In', 50: 'Sn', 
+          51: 'Sb', 52: 'Te', 53: 'I',  54: 'Xe', 55: 'Cs', 56: 'Ba', 57: 'La', 58: 'Ce', 
+          59: 'Pr', 60: 'Nd', 61: 'Pm', 62: 'Sm', 63: 'Eu', 64: 'Gd', 65: 'Tb', 66: 'Dy', 
+          67: 'Ho', 68: 'Er', 69: 'Tm', 70: 'Yb', 71: 'Lu', 72: 'Hf', 73: 'Ta', 74: 'W', 
+          75: 'Re', 76: 'Os', 77: 'Ir', 78: 'Pt', 79: 'Au', 80: 'Hg', 81: 'Tl', 82: 'Pb', 
+          83: 'Bi', 84: 'Po', 85: 'At', 86: 'Rn', 87: 'Fr', 88: 'Ra', 89: 'Ac', 90: 'Th', 
+          91: 'Pa', 92: 'U',  93: 'Np', 94: 'Pu', 95: 'Am', 96: 'Cm', 97: 'Bk', 98: 'Cf', 
+          99: 'Es'}
```
orbeckst: Why stop at 99?  ;-)

package/MDAnalysis/topology/tables.py
```diff
@@ -377,3 +377,20 @@ def kv2dict(s, convertor=str):
...
+
+SYMB2Z = {v:k for k, v in Z2SYMB.items()}
```
orbeckst: newline.

richardjgowers:
@orbeckst yeah I'm not sure where the line on converters is, but maybe a nice definiton is that a converter is moving from one package's namespace/object to another.

.appveyor.yml
```diff
@@ -9,8 +9,8 @@ cache:
...
-        CONDA_DEPENDENCIES: pip setuptools wheel cython mock six biopython networkx joblib matplotlib scipy vs2015_runtime pytest mmtf-python GridDataFormats hypothesis pytest-cov codecov
+        CONDA_CHANNELS: conda-forge omnia
```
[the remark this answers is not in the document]
richardjgowers: parmed is in conda-forge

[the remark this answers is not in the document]
orbeckst: Then drop omnia, let us know what dependencies it pulls in, and then I am happy.

lilyminium:
Thanks @orbeckst. Added a couple usage examples and took out the `omnia` channel.

> Maybe we don't need to reorganize their code but their docs should be prominent.

I added a [brief page on converters](https://lilyminium.github.io/mdanalysis/documentation_pages/converters.html) -- is that what you meant?

orbeckst:
> I added a brief page on converters -- is that what you meant?

Yes, that's a good starting point!

orbeckst:
I approved it because you addressed my concerns – thank you.

Obviously, tests should be passing, too. I am not sure why many of the last tests failed but I restarted them. If they fail again, have a look...

PR #2406, opened 2019-11-18
Improved exception message for traj file without extension.

joaomcteixeira:
*This is a very small addition.*

**Improves Exception message when `format=None` or `filename` string has no extension.**

**Previous behavior:**
```python
with mda.Writer('output_traj', selection=selection.n_atoms) as W:
    for ts in u.trajectory:
        W.write(selection)
  File "/home/joao/GitHub/fork_mdanalysis/package/MDAnalysis/coordinates/core.py", line 134, in writer
    multiframe=kwargs.pop('multiframe', None))
  File "/home/joao/GitHub/fork_mdanalysis/package/MDAnalysis/core/_get_readers.py", line 199, in get_writer_for
    raise_from(TypeError(errmsg.format(format)), None)
  File "<string>", line 3, in raise_from
TypeError: No trajectory or frame writer for format ''
```

**PR behavior**:
```python
with mda.Writer('output_traj', selection=selection.n_atoms) as W:
    for ts in u.trajectory:
        W.write(selection)
  File "/home/joao/GitHub/fork_mdanalysis/package/MDAnalysis/coordinates/core.py", line 134, in writer
    multiframe=kwargs.pop('multiframe', None))
  File "/home/joao/GitHub/fork_mdanalysis/package/MDAnalysis/core/_get_readers.py", line 187, in get_writer_for
    ).format(filename))
MDAnalysis.exceptions.FileIOError: No extension assigned for trajectory output file 'output_traj'.
```

Enhances:
* raises custom Exception
* exception message is clearer

Tests pass on my computer.

### DIFF
--- a/package/MDAnalysis/core/_get_readers.py
+++ b/package/MDAnalysis/core/_get_readers.py
@@ -26,7 +26,6 @@ import copy
 import inspect
 import mmtf
 import numpy as np
-from MDAnalysis.exceptions import FileIOError
 from MDAnalysis.lib.util import isstream
 
 from .. import _READERS, _PARSERS, _MULTIFRAME_WRITERS, _SINGLEFRAME_WRITERS
@@ -163,10 +162,10 @@ def get_writer_for(filename, format=None, multiframe=None):
     """
     if filename is None:
         format = 'NULL'
-    elif format is None:
+    elif format in ('', None):
         try:
             root, ext = util.get_ext(filename)
-        except (TypeError, AttributeError):
+        except (TypeError, AttributeError, ValueError):
             # An AttributeError is raised if filename cannot
             # be manipulated as a string.
             # A TypeError is raised in py3.6
@@ -178,13 +177,6 @@ def get_writer_for(filename, format=None, multiframe=None):
         else:
             format = util.check_compressed_format(root, ext)
         
-    if format == '':
-        # Improves detail of error message when ext is not assigned in
-        # output file. Otherwise TypeError is raised when except the
-        # return statement
-        raise FileIOError((
-            "No extension assigned for trajectory output file '{}'."
-            ).format(filename))
     format = format.upper()
     if multiframe is None:
         # Multiframe takes priority, else use singleframe
--- a/package/MDAnalysis/core/_get_readers.py
+++ b/package/MDAnalysis/core/_get_readers.py
@@ -26,6 +26,7 @@ import copy
 import inspect
 import mmtf
 import numpy as np
+from MDAnalysis.exceptions import FileIOError
 from MDAnalysis.lib.util import isstream
 
 from .. import _READERS, _PARSERS, _MULTIFRAME_WRITERS, _SINGLEFRAME_WRITERS
@@ -176,6 +177,14 @@ def get_writer_for(filename, format=None, multiframe=None):
                 None)
         else:
             format = util.check_compressed_format(root, ext)
+        
+    if format == '':
+        # Improves detail of error message when ext is not assigned in
+        # output file. Otherwise TypeError is raised when except the
+        # return statement
+        raise FileIOError((
+            "No extension assigned for trajectory output file '{}'."
+            ).format(filename))
     format = format.upper()
     if multiframe is None:
         # Multiframe takes priority, else use singleframe
--- a/package/MDAnalysis/core/_get_readers.py
+++ b/package/MDAnalysis/core/_get_readers.py
@@ -162,10 +162,10 @@ def get_writer_for(filename, format=None, multiframe=None):
     """
     if filename is None:
         format = 'NULL'
-    elif format in ('', None):
+    elif format is None:
         try:
             root, ext = util.get_ext(filename)
-        except (TypeError, AttributeError, ValueError):
+        except (TypeError, AttributeError):
             # An AttributeError is raised if filename cannot
             # be manipulated as a string.
             # A TypeError is raised in py3.6
@@ -176,7 +176,14 @@ def get_writer_for(filename, format=None, multiframe=None):
                 None)
         else:
             format = util.check_compressed_format(root, ext)
-        
+    
+    if format == '':
+        raise ValueError((
+            'File format could not be guessed from {}, '
+            'resulting in empty string - '
+            'only None or valid formats are supported.'
+            ).format(filename))
+
     format = format.upper()
     if multiframe is None:
         # Multiframe takes priority, else use singleframe
--- a/package/MDAnalysis/exceptions.py
+++ b/package/MDAnalysis/exceptions.py
@@ -60,10 +60,6 @@ class FileFormatWarning(Warning):
     """Warning indicating possible problems with a file format."""
 
 
-class FileIOError(Exception):
-    """Exception indicating errors with I/O paths."""
-
-
 class StreamWarning(Warning):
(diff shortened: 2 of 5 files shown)


package/MDAnalysis/exceptions.py
```diff
@@ -60,6 +60,10 @@ class FileFormatWarning(Warning):
     """Warning indicating possible problems with a file format."""
 
 
+class FileIOError(Exception):
```
orbeckst: We don't want to introduce custom exceptions unless really necessary. (Originally we were quite generous with new warnings and exceptions but this tends to make things confusing for downstream code, which then has to import the custom exceptions to test against them etc.)

Unless there's a really good reason, please remove FileIOError.

package/MDAnalysis/core/_get_readers.py
```diff
@@ -26,6 +26,7 @@
 import inspect
 import mmtf
 import numpy as np
+from MDAnalysis.exceptions import FileIOError
```
orbeckst: remove (see below)

package/MDAnalysis/core/_get_readers.py
```diff
@@ -176,6 +177,14 @@ def get_writer_for(filename, format=None, multiframe=None):
                 None)
         else:
             format = util.check_compressed_format(root, ext)
+        
+    if format == '':
```
orbeckst: Under which circumstances can this happen? In particular, can [`util.check_compressed_format(root, ext)`](https://github.com/MDAnalysis/mdanalysis/blob/3b0bdcebd9f087d8b8497815616e53003de6ab2a/package/MDAnalysis/lib/util.py#L933) ever return `""`? — I think the answer is yes but I didn't try it and it's not really documented.

package/MDAnalysis/core/_get_readers.py
```diff
@@ -176,6 +177,14 @@ def get_writer_for(filename, format=None, multiframe=None):
...
+        # Improves detail of error message when ext is not assigned in
+        # output file. Otherwise TypeError is raised when except the
+        # return statement
+        raise FileIOError((
```
orbeckst: This should  also be `ValueError` as the format could not be guessed.

If the format was set by the user as empty string then that should be an error (also `ValueError`) as there is no way to determine a format (it enables auto-detection (`None`) and does not provide a valid format).

testsuite/MDAnalysisTests/lib/test_util.py
```diff
@@ -41,7 +41,7 @@
 from MDAnalysis.lib.util import (cached, static_variables, warn_if_not_unique,
                                  check_coords)
 from MDAnalysis.core.topologyattrs import Bonds
-from MDAnalysis.exceptions import NoDataError, DuplicateWarning
+from MDAnalysis.exceptions import NoDataError, DuplicateWarning, FileIOError
```
orbeckst: no new custom exception

testsuite/MDAnalysisTests/lib/test_util.py
```diff
@@ -1103,15 +1103,21 @@ def test_precedence(self):
     def test_missing_extension(self):
         # Make sure ``get_writer_for`` behave as expected if *filename*
         # has no extension
-        with pytest.raises(TypeError):
+        with pytest.raises(FileIOError):
```
orbeckst: no new custom exception

orbeckst: Also test giving `format=""`. maybe a separate test.

package/MDAnalysis/core/_get_readers.py
joaomcteixeira: This happens under the circumstances described in the PR message:

```python
with mda.Writer('output_traj', selection=selection.n_atoms) as W:
    for ts in u.trajectory:
        W.write(selection)
  File "/home/joao/GitHub/fork_mdanalysis/package/MDAnalysis/coordinates/core.py", line 134, in writer
    multiframe=kwargs.pop('multiframe', None))
  File "/home/joao/GitHub/fork_mdanalysis/package/MDAnalysis/core/_get_readers.py", line 199, in get_writer_for
    raise_from(TypeError(errmsg.format(format)), None)
  File "<string>", line 3, in raise_from
TypeError: No trajectory or frame writer for format ''
```

joaomcteixeira:
I understand your requirements for avoiding the custom Exceptions, I will correct that accordingly.

package/MDAnalysis/core/_get_readers.py
joaomcteixeira: Do you prefer `get_ext()` to raise an error if `ext==''` ? In this way we would not need to rewrite the error message and we could actually use the except block `except (TypeError, AttributeError):`.

testsuite/MDAnalysisTests/lib/test_util.py
```diff
@@ -931,6 +931,10 @@ def test_get_extention(self, extention):
 
         assert a == 'file'
         assert b == extention.lower()
+    
+    def test_get_extension_without_extension(self):
```
orbeckst: good that you added a test but `get_ext` should not raise a ValueError so the test is not necessary

package/MDAnalysis/core/_get_readers.py
```diff
@@ -162,10 +162,10 @@ def get_writer_for(filename, format=None, multiframe=None):
     """
     if filename is None:
         format = 'NULL'
-    elif format is None:
+    elif format in ('', None):
```
orbeckst: It's better to just raise an error for `format == ""` – I don't like making "" equivalent to `None` (which also changes the API). Just do
```python
if format == "":
   raise ValueError("format cannot be the empty string --- only None or valid formats are supported")
```

package/MDAnalysis/core/_get_readers.py
```diff
@@ -176,6 +176,7 @@ def get_writer_for(filename, format=None, multiframe=None):
                 None)
         else:
```
orbeckst: I can't think of a better way than adding another check after the `else:`
```python
if ext == "":
   raise ValueError('File format could not be guessed from "{0}"'.format(filename))
```

package/MDAnalysis/core/_get_readers.py
```diff
@@ -162,10 +162,10 @@ def get_writer_for(filename, format=None, multiframe=None):
     """
     if filename is None:
         format = 'NULL'
-    elif format is None:
+    elif format in ('', None):
         try:
             root, ext = util.get_ext(filename)
-        except (TypeError, AttributeError):
+        except (TypeError, AttributeError, ValueError):
```
orbeckst: Let's not have `get_ext()` raise ValueError so remove ValueError here.

joaomcteixeira:
Okay, I understand your review. I was not sure if the `ValueError` on `get_ext` would be a desired implementation but I gave it a try. I did that basically to avoid the raise error duplication but, as you said, it may, on the other hand, complicate the `get_ext` API. I will change according to your review during today and commit again.

Thanks to you! It is very nice to participate in the project now that I am using it quite intensively.

Best,

joaomcteixeira:
@orbeckst I have addressed your reviews. Let me know your opinion on this new commit.

orbeckst:
Thank you @joaomcteixeira !

Issue #588, opened 2015-12-15
What should be a bond in Gromacs topologies?


jbarnoud:
A few days ago, @mnmelo and I talked about what should be considered a bond in Gromacs topologies. Indeed, Gromacs offers various ways to define a bond, an angle, a dihedral, or an improper; some of these ways are not always used with topological bonds in mind. For instance, depending on the force field and the molecule, a constraint can be a bond or not. The same goes for virtual sites.

For now, the TPR parser has a list of interactions it considers as bonds. Constraints, for instance, are part of this list meaning that there can be to much bonds read for some molecules where constraints serve an other purpose.

With the new AtomGroup model (#363), it will be possible to attach any topological attributes to a universe. Would it make sense to have each interaction type be a separate attribute? Then lists of bonds, angles, dihedrals, and impropers could get composed based on these attributes, without information getting lost in the process.

In the long shot, this information can be used in the context of #507 to guess part of the topology based on the force field.

richardjgowers:
Hey

Yeah definitely something we can play with now. I've been thinking how I
could store information against the bonds, and have this attached to the
TopologyGroups too. We could then tag harmonic and constraints differently
from example.

I think ideally the extra info would be similar to Attributes again. I'm
pretty busy elsewhere in the code, so if you want to tinker feel free.

dotsdl:
@jbarnoud I'd have a look at [`Bonds` in ``core/topologyattrs.py``](https://github.com/MDAnalysis/mdanalysis/blob/issue-363/package/MDAnalysis/core/topologyattrs.py#L444) and perhaps make a subclass for constraints. I really like the idea of individual topology formats being able to find complete representation in the `Topology` object that's delivered by their parser. I think this could allow us to more easily *write* topologies in the future, too, since the information is well-contained.

orbeckst:
On a semantic level, I would reserve "bond" for a chemical bond and not include artificial constructs such as constraints. Bonds is also what would be used to define *fragments* (i.e. molecules).

jbarnoud:
Constraints can be used to describe chemical bonds. It is often used to avoid bonds vibrations and increase the step. The issue is that the same constraints can be used for other purposes. 
I am playing with the new topology system to store all the interactions as topology attributes and let the user choose what interactions to keep as bonds. 
[...]

PR #2408, opened 2019-11-20
Modifications to ITPParser

lilyminium:
 - Types parsed to Bonds/Angles/Dihedrals/Impropers match that of [TPRParser](https://lilyminium.github.io/UserGuide/formats/reference/tpr.html)

*Although*, more interactions are treated as bonds by TPRParser/ITPParser than should be, according to the manual. These directives are pretty unusual in an ITP file, though.

>Particles are considered bonded when they are connected by “chemical” bonds ([ bonds ] types 1 to 5, 7 or 8) or constraints ([ constraints ] type 1). Type 5 [ bonds ] can be used to create a connection between two atoms without creating an interac- tion. There is a harmonic interaction ([ bonds ] type 6) that does not connect the atoms by a chemical bond. There is also a second constraint type ([ constraints ] type 2) that fixes the distance, but does not connect the atoms by a chemical bond.

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?


richardjgowers:
Related (but ancient) #588

So back in the past @orbeckst said that constraints aren't bonds, but then later wanted his water constraints to show up as bonds :)

orbeckst:
In my feeble defense: the world is a complicated place. ;-)

More seriously: it’s not easy to make these decisions automagically.  Maybe there is something that allows us to distinguish the different use cases — @jbarnoud might have some insights. I haven’t looked into the gmx topologies in detail in a while.

jbarnoud:
> @jbarnoud might have some insights. I haven’t looked into the gmx topologies in detail in a while.

If I recall correctly, I used the logic quoted from the manual when I rewored the TPR parser.

lilyminium:
I think bond type 6 (HARMONIC) and constraints type 2 (CONSTRNC) are still included in creating bonds:

https://github.com/MDAnalysis/mdanalysis/blob/9834b09f4fb86ad12652cf0d562d7da90c4ee1f2/package/MDAnalysis/topology/tpr/utils.py#L625-L629

jbarnoud:
Hum... They should not. It seems I mistook my "this is what I should do" with a "this is what I have done" (https://github.com/MDAnalysis/mdanalysis/issues/588#issuecomment-199182320).

lilyminium:
`[ settles ]` is a bit hard to do here since it requires knowing which atoms are Hs to add bonds from the constraints. Below is tip5p from OPLS to illustrate the problem. I'm not sure how GROMACS figures out constraints, unless the 2nd and 3rd atoms are always hydrogen?

After looking at the generic water ITPs they all define bonds if the FLEXIBLE keyword isn't passed in anyway. This means unless the `#if`s are treated properly there'll be redundant interactions.

```
; original reference: [M. W. Mahoney and W. L. Jorgensen, J. Chem. Phys. 112 , 2000]
;
; Note that there are various issues with tip5p and the different forcefields.
; Discussion is here: http://redmine.gromacs.org/issues/1348

[ moleculetype ]
; molname       nrexcl
SOL             2

[ atoms ]
; id    at type res nr  residu name     at name         cg nr   charge
1       opls_118     1       SOL              OW             1       0
2       opls_119     1       SOL             HW1             1       0.241
3       opls_119     1       SOL             HW2             1       0.241
4       opls_120     1       SOL             LP1             1      -0.241
5       opls_120     1       SOL             LP2             1      -0.241

#ifndef FLEXIBLE

[ settles ]
; i     funct   doh     dhh
1       1       0.09572 0.15139

#else

[ bonds ]
; i     j       funct   length  force.c.
1       2       1       0.09572 502416.0 0.09572        502416.0
1       3       1       0.09572 502416.0 0.09572        502416.0

[ angles ]
; i     j       k       funct   angle   force.c.
2       1       3       1       104.52  628.02  104.52  628.02

#endif

[ virtual_sites3 ]
; The position of the virtual site is computed as follows:
;
; The distance from OW to L is 0.07 nm, the geometry is tetrahedral
; (109.47 deg)
; Therefore, a = b = 0.07 * cos (109.47/2) / | xOH1 + xOH2 |
;            c = 0.07 * sin (109.47/2) / | xOH1 X xOH2 |
;
; Using | xOH1 X xOH2 | = | xOH1 | | xOH2 | sin (H1-O-H2)
;       | xOH1 + xOH2 | = 2 | xOH1 | cos (H1-O-H2)
; Vsite pos x4 = x1 + a*x21 + b*x31 + c*(x21 X x31)

; Vsite from                    funct   a       b               c
4       1       2       3       4       -0.344908  -0.344908  -6.4437903493
5       1       2       3       4       -0.344908  -0.344908   6.4437903493

[ exclusions ]
1       2       3       4       5
2       1       3       4       5
3       1       2       4       5
4       1       2       3       5
5       1       2       3       4

```

jbarnoud:
> This means unless the #ifs are treated properly there'll be redundant interactions.

I do not know how the ITP parser deals with the preprocessor instructions, but it does not make sense to read an ITP file without the set of defined variables. 

An ITP file can describe completely different systems depending on these `#if` macros.

For `[settle]`, the TPR file assumes (or use to assume at least) that the triangle was `i`, `i+1`, and `i+2`. Since the TPR file does not change the order of the atoms, I guess it is safe to make the same assumption for the ITP.

jbarnoud:
> I think bond type 6 (HARMONIC) and constraints type 2 (CONSTRNC) are still included in creating bonds:

I have been working on the TPR parser these last few days and remember why I ended up including all the bonds as bond. By excluding some bonds or constraints that are defined in the TPR from the bond we read, we make it impossible to access these bonds. Since we cannot know why the bonds are there, and what the user wants to analyze, it is safer to read everything. One use case I encountered is the analysis of an elastic network. Such a network can be defined with bonds of type 6 which do not cause an exclusion. If we follow the logic we discussed above, they become impossible to access from MDAnalysis, making their analyses much harder.

The solution would be to store what type of bond each bond is and to expose a way to filter them. Until then, it is better to read everything and not assume the intent of the user.

package/MDAnalysis/topology/base.py
```diff
@@ -181,3 +181,10 @@ def get_borders(*arrays):
             # Should be the same for self consistency...
 
     return residx, new_others
+
+
+def squash_identical(values):
```
richardjgowers: docs pls

package/MDAnalysis/topology/ITPParser.py
```diff
@@ -215,16 +246,55 @@ def parse(self, **kwargs):
...
+            tattr = Attr(indices, types=types)
             top.add_TopologyAttr(tattr)
 
         return top
+    
+    def skip_if(self):
+        """
+        Skip lines in if condition, *inclusive* of the #endif or #else.
+        """
+        line = next(self.itpfile)
```
richardjgowers: this is a fun example of where the walrus operator is actually useful (if we could jump straight to 3.8)

package/MDAnalysis/topology/ITPParser.py
```diff
@@ -111,29 +148,45 @@ def parse(self, **kwargs):
...
-        with openany(self.filename) as itpfile:
-            for line in itpfile:
+        with openany(self.filename) as self.itpfile:
+            for line in self.itpfile:
                 line = line.split(';')[0].strip()  # ; is for comments
                 if not line:  # Skip line if empty
                     continue
-                if line.startswith('#'): #ignore include, ifdefs, etc
+
+                if line.startswith('#ifdef'):
```
richardjgowers: Maybe being pedantic, are nested `ifdef`s allowed in itp files?  If so I think maybe this approach won't work?

jbarnoud: I am almost sure they are allowed.

richardjgowers: In which case you need some sort of if stack thing like it looks like was planned.

jbarnoud:
I wrote this ITP parser not long ago. It may help with inspiration: https://github.com/jbarnoud/omm-top

package/MDAnalysis/topology/ITPParser.py
lilyminium: I think I did something similar to @jbarnoud's ``_skip_lines_until_endif()`` thing, although without the IOError catching. There's a test case for nested ifs.

lilyminium:
Thanks @jbarnoud! Today I learned about ``yield from`` :-)

Should ITPParser parse ``#include`` and ``#define``? I don't think many molecule files have them.

jbarnoud:
`yield from` is python 3 only, but it should soon not be a problem 😉 

`'#include` is crucial. A top file will use it to include the force field and the molecule description.

`#define` is less crucial, but should ultimately be there to be full-featured and avoid surprises to the user.

lilyminium:
> '#include is crucial. A top file will use it to include the force field and the molecule description.

Yes...  I can't think of a time I've seen it in an ITP file. I suppose it's not too difficult to put in and make this a generic top/itp parser. I'll look at this tomorrow. And add testing for flawed files (e.g. missing #endifs, etc). 

> `yield from` is python 3 only, but it should soon not be a problem 😉

I look forward to being able to use f-strings everywhere someday.

jbarnoud:
There is no difference between a top and itp file. The difference in file extension is only a matter of convention to show which file are meant to be top level and which ones are meant to be included.

jbarnoud:
I'll have a look over the week-end.

PR #2411, opened 2019-12-03
Bugfix #2410

### LINKED ISSUE #2410: Problem with step_to_frame in base
The following code in `auxiliary/base.py` seems to return the time difference between the first frame and `step`, 

```python
time_frame = time_frame_0 + frame_index*ts.dt
time_diff = abs(time_frame_0 - time_step)
return frame_index, time_diff
```
while it should return the time difference between `frame_index` and `step` (according to the documentation)

```python
time_frame = time_frame_0 + frame_index*ts.dt
time_diff = abs(time_frame - time_step)
return frame_index, time_diff
```

RMeli:
Fixes #2410

 - Changed `time_frame_0` with `time_frame`, according to documentation.

Tests to follow.

PR Checklist
------------
 - [x] Tests
 - [x] Issue raised/referenced


package/MDAnalysis/auxiliary/base.py
```diff
@@ -908,7 +908,7 @@ def _restart(self):
 
     def _reopen(self):
         """ Close and then reopen *auxfile*. """
-        if self.auxfile != None:
```
richardjgowers: this seems to reverse the logic?

RMeli: Wooops, I wanted to revert the change (not really associated to the PR) but I must have pressed `Ctrl-Z` only once. I will change it to `is not` if it's ok to leave this change within this PR.

orbeckst:
@fiona-naughton can you have a look at this PR, please?

orbeckst:
@RMeli we'll also need a test, please.

orbeckst:
@RMeli why did you mark the PR as WIP? (Sorry that it's been taking a while to get back to your PR.)

(The macOS failure on Travis was likely spurious – a timeout of a hypothesis tests for comments in DCD files, which has been flaky in the past. I restarted the CI.)

orbeckst:
ping @fiona-naughton

package/MDAnalysis/auxiliary/base.py
orbeckst: @richardjgowers – did @RMeli address all your issues?

RMeli:
@orbeckst Yes, I marked it as WIP since tests are still not written. I'll try to add them ASAP, but I might take a while since I'll have to understand what the code does exactly and hot to hit it.

fiona-naughton:
Apologies for the delay in getting to this. This fix looks good.

Currently this only gets used to check if an auxilliary step is within the cutoff time from a trajectory frame when using `iter_as_aux` on the trajectory reader. The current test for `iter_as_aux` with cutoff just used an offset auxilliary/cutoff=0 and checked all frames were skipped which is why this wasn't getting picked up, since they end up being skipped with the error anyway. This could be tested using a higher cutoff than the offset, or the 0 cutoff with no offset, and checking we do get all the frames (or could probably be tested more directly, too).

RMeli:
Thanks for the explanations @fiona-naughton ! I'll try to make a test ASAP.

testsuite/MDAnalysisTests/auxiliary/base.py
```diff
@@ -359,22 +359,31 @@ def test_load_from_description(self, reader):
...
+        for idx, expected_frame, expected_time_diff in zip(
+            [0, 1, 2], # idx
+            [0, 1, 2], # expected frame
+            [0, 0.2, 0.4] # expected time difference
+            ):
+
+            frame, time_diff = reader.step_to_frame(idx, ts, return_time_diff=True)
+            print(frame, time_diff)
+            assert frame == expected_frame
+            assert time_diff == pytest.approx(expected_time_diff)
```
richardjgowers: If memory serves pytest approx has caused grief in the past, try and use the np testing equivalent

testsuite/MDAnalysisTests/auxiliary/base.py
```diff
@@ -357,7 +357,33 @@ def test_load_from_description(self, reader):
...
+        assert reader.step_to_frame(-1, ts) is None
+        assert reader.step_to_frame(reader.n_steps, ts) is None
+        assert reader.step_to_frame(0, ts) == 0
+        
+        frame, time_diff = reader.step_to_frame(1, ts, return_time_diff=True)
+
+        assert frame == 1
+        assert time_diff == pytest.approx(0)
+
+        # Timestep is 0.2 longer than auxiliary data
```
richardjgowers: Try and split tests so each test does one thing. This ts is independent of the previous

RMeli:
I went for a direct test of `step_to_frame` by setting a `Timestep` with `dt=1.2` and checking the returned time difference.

testsuite/MDAnalysisTests/auxiliary/base.py
RMeli: Good point. I created `test_step_to_frame_out_of_bound`, `test_step_to_frame_no_time_diff` and `test_step_to_frame_time_diff`.

testsuite/MDAnalysisTests/auxiliary/base.py
RMeli: Good to know (I didn't notice it is mentioned in the User Guide).  I changed it to `np.testing.assert_almost_equal`.

orbeckst:
@fiona-naughton in the spirit of apprentice reviewing I suggest you review this PR. 

@richardjgowers and I will have to do a final ok and merge but it’s a lot easier (for me) if I can look at your review first and then write ‘LGTM as she said’.

testsuite/MDAnalysisTests/auxiliary/base.py
```diff
@@ -357,8 +357,32 @@ def test_load_from_description(self, reader):
...
+
+        # Timestep is 0.1 longer than auxiliary data
+        ts = mda.coordinates.base.Timestep(0, dt=1.1)
+
+        for idx in range(reader.n_steps):
+
+            frame, time_diff = reader.step_to_frame(idx, ts, return_time_diff=True)
+
+            assert frame == idx
+            np.testing.assert_almost_equal(time_diff, idx * 0.1)
```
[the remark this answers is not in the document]
RMeli: Good point, thanks. I hard-coded a loop over 5 frames and added a comment, so that this will work if the trajectory length is changed. Let me know if you prefer I only add a comment so that this test will likely fail if the trajectory length is increase.

fiona-naughton:
This looks good to me now!

orbeckst:
Thanks @RMeli and @fiona-naughton !

PR #2415, opened 2019-12-13
TST: add Python 3.8 Linux to CI

tylerjereddy:
Full test suite passes locally on Linux + Python 3.8. We'll see if the new CI matrix entry gets the deps all right with the Python version bump. NumPy should be `1.17.x` series and SciPy should be `1.3.x` series.


tylerjereddy:
3.8/Linux CI results look the same as 3.7, which is good I think

tylerjereddy:
Simple testing improvement with core dev approval--in it goes.

I'll see if I can check the Windows/3.8 situation soon as well.

Issue #601, opened 2015-12-28
AtomGroup methods from TopologyAttrs (transplants)


richardjgowers:
So currently (on the issue-363 branch) arbitrary attributes can be defined (usually by a parser) and these are then accessible at the AtomGroup (and Atom, ResidueGroup etc) level.  For example

``` python
class Colour(TopologyAttr):
    level = 'atom'
    plural = 'colours'
    # code here

atomgroup.colours
# [red, blue, blue, red etc]
```

The next step in this is to allow a `TopologyAttr` to also define additional methods which appear in the `Groups` and `Components` namespace.  The code for these would live in the TA, but get moved over into the AtomGroup on Universe init.  This will look something like:

``` python
class Colour(TopologyAttr):
    # fancy code
    def complimentary_colours(ag):
        return complimentary[ag.colours]

atomgroup.colours
atomgroup.complimentary_colours()
```

Reasons for this approach
- Prevents the base AtomGroup class from becoming bloated with every imaginable method
- All code related to a given Attribute lives in the same module
- Can include extra Attributes & methods without installing MDA from source

TODO:
- [x] get it working for `Masses`
- [ ] write tests for adding arbitrary attribute (`Happiness`)
- [x] write dev wiki page for how this works

dotsdl:
Once you have the test you mention, perhaps we'll close this. Up to you. It's enough for me if all the transplants we define actually work, but it is cool if we can show how easy it is to dynamically add things in.

PR #2418, opened 2019-12-20
Generate doc tables for transplanted methods, continued

### LINKED ISSUE #1845: docs for many AtomGroup methods missing

lilyminium:
Fixes #1845

Continues #2176 
 - Writes a sorted list of methods and properties without arguments like ``group`` or ``self``. 

I haven't addressed the duplicate citation issue. Links just go to the original `MDAnalysis.core.topologyattrs.Attrname.Method` docstring citation.

To be honest, IMO the best solution would be to just move the methods to the core classes and fail gracefully when the required topology attribute isn't there. This approach has many downsides, including:

- I sorted the properties/methods alphabetically, but all of them nevertheless appear before the *Group's *actual* properties/methods
- You must manually add `.. include:: XXX_methods_table.txt` and `.. include:: XXX_methods_docs.txt` to each docstring every time you add a new transplant target
- The Universe transplanted methods are now in a docstring that's largely about its `__init__` method and will show up in `help(Universe)`
- citations confusingly link to an identical entry on a different page
- People looking at the code now have a mismatch between code and documentation
- what happens when you have a method that requires >1 topology attribute?

PR Checklist
------------
 - [ ] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,6 @@ authors.py
 *.DS_Store
 # ignore files from tests
 .hypothesis/
+
+# Ignore generated documentation stubs
+package/doc/sphinx/source/documentation_pages/core/*.txt
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,7 +54,7 @@ matrix:
            BUILD_DOCS=true
            BUILD_CMD="cd ${TRAVIS_BUILD_DIR}/package && python setup.py build_ext --inplace"
            INSTALL_HOLE="false"
-           PIP_DEPENDENCIES="${PIP_DEPENDENCIES} sphinx==2.1 sphinx-sitemap sphinx_rtd_theme"
+           PIP_DEPENDENCIES="${PIP_DEPENDENCIES} sphinx==2.2.0 sphinx-sitemap sphinx_rtd_theme"
 
     - env: NAME="Lint"
            PYLINTRC="${TRAVIS_BUILD_DIR}/package/.pylintrc"
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,7 +54,7 @@ matrix:
            BUILD_DOCS=true
            BUILD_CMD="cd ${TRAVIS_BUILD_DIR}/package && python setup.py build_ext --inplace"
            INSTALL_HOLE="false"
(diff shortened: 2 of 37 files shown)


lilyminium:
Looks like the screenshot below or have a look at these [built docs](https://lilyminium.github.io/mdanalysis/documentation_pages/core/groups.html#MDAnalysis.core.groups.Atom). I felt a little blurb was needed to explain the sudden table.

Screenshot 2019-12-20 at 5 14 55 PM

richardjgowers:
I’m ok with moving methods into the base class. I think the whole transplanted methods thing went a little too far.

jbarnoud:
To add to @lilyminium argument, having the transplanted method also means that the methods are less discoverable, that they are clearly not IDE friendly, and that error messages are unhelpful when switching topology format.

jbarnoud:
> I’m ok with moving methods into the base class. I think the whole transplanted methods thing went a little too far.

I gave a go at the move in #2425. I am not done with it, but it feels already like a massive regression. The reasons @richardjgowers gave to justify the transplanted methods stand and become more obvious when undoing the change.

From #601:

> Reasons for this approach
> - Prevents the base AtomGroup class from becoming bloated with every imaginable method
> - All code related to a given Attribute lives in the same module
> - Can include extra Attributes & methods without installing MDA from source

orbeckst:
This PR has been superseded by PR #2427 and #3007 so I am closing it.

Issue #2039, opened 2018-08-10
Iterative Average structures


kain88-de:
**Is your feature request related to a problem? Please describe.**

In a recent [paper](https://pubs.acs.org/doi/abs/10.1021/acs.jpcb.7b11988) of mine, I describe an iterative algorithm to calculate an average structure from simulations.  Below is the description. 

> As a reference structure, I used the average structure in the simulations, which for B-DNA was
obtained by RMSD aligning the inner five base pairs and averaging the simulation structures iteratively to convergence. The first frame was used for the initial alignment.

I would expect that this algorithm was already previously described. If anyone knows an earlier paper please add it to this issue so we can properly cite it. I  currently do not have time to include this properly into MDAnalysis.

*Note*: I find this a good example to show how the changes of our analysis module in the last two years allow it to quickly write up a new analysis type that is easy to reuse and share.

**Describe the solution you'd like**

This is the complete implementation I used including tests. I'm posting it here should anyone be interested in including this into the align module of MDAnalysis. If anyone wants to just use this please consider to [cite my paper](https://pubs.acs.org/doi/abs/10.1021/acs.jpcb.7b11988) when you do. The code can be used under the same license that MDAnalysis uses.

```python
class AverageStructure(base.AnalysisBase):
    """Calculate an average structure after RMSD alignment to a reference."""

    def __init__(self, mobile, ref, weights=None, calc_msf=True, **kwargs):
        """init function

        Parameters
        ----------
        mobile : mda.AtomGroup
            atomgroup that is being fitted
        ref : mda.AtomGroup
            reference structure
        weights : str, array_like (optional)
            weights that can be used. If `None` use equal weights, if `'mass'`
            use masses of ref as weights or give an array of arbitrary weights.
        calc_msf : bool (optional)
            calc mean square fluctuation
        kwargs : dict
            options for AnalysisBase to adjust iteration behavior

        """
        super(AverageStructure, self).__init__(mobile.universe.trajectory,
                                               **kwargs)
        self.mobile = mobile
        self.ref_xyz = ref.positions.copy() - ref.centroid()
        self.weights = weights
        self.calc_msf = calc_msf

    def _prepare(self):
        self.structure = np.zeros(self.mobile.positions.shape)
        self.rmsd = 0
        if self.calc_msf:
            self.msf = np.zeros(self.mobile.n_atoms)

    def _single_frame(self):
        self.mobile.translate(-self.mobile.center(self.weights))
[...]

PR #2430, opened 2019-12-26
align.AverageStructure for out-of-memory average structures

lilyminium:
Relates to #2039 

 - added align.AverageStructure

PR Checklist
------------
 - [x] Tests?
 ~- [ ] Missing one to check the ``msf`` -- anyone have any existing tools to compute MSF?~ removed MSF
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/analysis/align.py
+++ b/package/MDAnalysis/analysis/align.py
@@ -599,7 +599,7 @@ class AlignTraj(AnalysisBase):
             Array of the rmsd values of the least rmsd between the mobile_atoms
             and reference_atoms after superposition and minimimization of rmsd
         filename : str
-            String reflecting the filename of the file where mobile_atoms
+            String reflecting the filename of the file where the aligned
             positions will be written to upon running RMSD alignment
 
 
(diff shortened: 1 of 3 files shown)


package/MDAnalysis/analysis/align.py
```diff
@@ -709,6 +709,200 @@ def save(self, rmsdfile):
         np.savetxt(rmsdfile, self.rmsd)
         logger.info("Wrote RMSD timeseries  to file %r", rmsdfile)
 
+class AverageStructure(AnalysisBase):
```
orbeckst: add duecredit for DOI 10.1021/acs.jpcb.7b11988

package/MDAnalysis/analysis/align.py
```diff
@@ -709,6 +709,200 @@ def save(self, rmsdfile):
...
+
+        import MDAnalysis as mda
+        from MDAnalysis.tests.datafiles import PSF, DCD
+        from MDAnalysis.analysis import align
+
+        u = mda.Universe(PSF, DCD)
+
+        # align to the third frame and average structure
+        av = align.AverageStructure(u, ref_frame=3).run()
+        averaged_universe = av.universe
```
orbeckst: Cite the paper DOI https://doi.org/10.1021/acs.jpcb.7b11988 ; we typically put all references into references.rst and link to there. This is less convenient for `help(class)` but better for the online docs. If you want to also add a References section and don't link to the local section.

orbeckst: As far as I can tell, `gmx trjconv` has been able to do an iterative fit for a long time: http://manual.gromacs.org/documentation/2020/onlinehelp/gmx-trjconv.html with `-fit progressive`:

> There are two options for fitting the trajectory to a reference either for essential dynamics analysis, etc. The first option is just plain fitting to a reference structure in the structure file. The second option is a progressive fit in which the first timeframe is fitted to the reference structure in the structure file to obtain and each subsequent timeframe is fitted to the previously fitted structure. This way a continuous trajectory is generated, which might not be the case when using the regular fit method, e.g. when your protein undergoes large conformational transitions.

If `AverageStructure` does the same as `-fit progressive` then it should be mentioned.

@kain88-de – did you ever look at `gmx trjconv -fit progressive`?

lilyminium: @orbeckst Right now the class doesn't actually implement @kain88-de 's iterative fit from the paper, although it is heavily based off his `AverageStructure` class. I would suggest the AlignTraj class is a better fit for implementing `-fit progressive`.

orbeckst: Ah – I misunderstood.

package/MDAnalysis/analysis/align.py
orbeckst: not needed

package/MDAnalysis/analysis/align.py
kain88-de: @orbeckst I did not look into the gmx progressive fit. From the description, it is not identical to my algorithm though. This PR makes the implementation of the iterative algorithm a lot easier though. I updated the issue accordingly.

Issue #2428, opened 2019-12-25
Native endianess in TPR files


ezavod:
This is probably an upstream problem, so this mostly aims at MDAnalysis' docs:

[Here](https://github.com/MDAnalysis/mdanalysis/blob/4adbd6c01673fcc9d13d4cd5c9089c1139257a57/package/MDAnalysis/topology/tpr/utils.py#L133-L136) it says that the new body is little-endian. I guess that's not correct.
In [tpr_test.zip](https://github.com/MDAnalysis/mdanalysis/files/4000298/tpr_test.zip) one could find `n3_le.tpr` and `n3_be.tpr` which are created on a little- and big-endian system respectively. Those could be inspected with `xxd n3_be.tpr`.
The header (until offset `0000:0063`), done with the original XDR serializer, is the same. But hereafter the new in memory serializer takes over and the files differ. The order of the bytes (i.e. padded 4 byte blocks) is reverse. GROMACS itself can't read big-endian files on a little-endian system and vice versa.
So it's native byte order instead of the documented little-endian.

There seem to be [efforts](http://redmine.gromacs.org/projects/gromacs/repository/revisions/ed91f1f7032b1411d33c254165d4cf6836ef318f/diff/src/gromacs/utility/inmemoryserializer.cpp) to enable endian swapping, but I don't know if this will apply to the TPR body and if the `standard` will be set to big-endian or little-endian.
Should one raise an issue in GROMACS regarding the TPR body?

PS: The comment goes on and explains the padding to four bytes. Curiously at some places they are not zero padded but one padded. E.g. Following the aforementioned header there are the box dimension. Those are in version 2019.4 `40 80 00 00` and in the new version (2020, rc1) `00 00 00 40  FF FF FF 80  0...`. I have yet to figure out where those Fs are coming from (and why this padding is even nescesary).

ping @jbarnoud

jbarnoud:
Oh! Thank you for this comment! Indeed, based on how the TPR body ends up little endian (packing each byte as a "char") it makes sense that it is native endian, I just did not realize that as I only worked on this on little-endian systems.

Is there a way of knowing what is the endianness of the TPR body? Ideally we want to be able to read TPR generated on any endianness, regardless of the native endianness. I guess I could look at where the padding is?

About the padding, I do not know why it sometime is `\x00` and sometimes `\xFF`.

[...]

Issue #2412, opened 2019-12-04
TPR parser is not compatible with Gromacs 2020


jbarnoud:
**Expected behavior**

The TPR parser reads topologies from TPR file produced by Gromacs 2020.

**Actual behavior**

The TPR parser produces an error:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/jon/dev/mdanalysis/package/MDAnalysis/core/universe.py", line 312, in __init__
    "Error: {2}".format(self.filename, parser, err)), None)
  File "<string>", line 3, in raise_from
ValueError: Failed to construct topology from file /home/jon/imdtest_gmx/imdexample/topol.tpr with parser <class 'MDAnalysis.topology.TPRParser.TPRParser'>.
Error: Your tpx version is 119, which this parser does not support, yet
```

**Code to reproduce the behavior**

Show us how to reproduce the failiure. If you can, use trajectory files from the test data.

``` python
>>> import MDAnalysis as mda
>>> mda.Universe('/home/jon/imdtest_gmx/imdexample/topol.tpr')
```

**Currently version of MDAnalysis**

- Which version are you using? (run `python -c "import MDAnalysis as mda; `'0.20.2-dev0'`
- Which version of Python (`python -V`)? `3.7.5`
- Which operating system? `Windows Subsystem for Linux, Ubuntu`

PR #2431, opened 2020-01-02
TPR for gromacs 2020 non-beta

jbarnoud:
 - Fix compatibility with TPR files produced by gromacs 2020
 - Remove support for TPR files produced by gromacs 2020-beta, raise an error instead

See #2412 and #2414 for the initial support for gromacs 2020-beta, see #2428 for the changes between 2020-beta and 2020.

PR Checklist
------------
 - [X] Tests?
 - [X] Docs?
 - [X] CHANGELOG updated?


package/MDAnalysis/topology/tpr/utils.py
```diff
@@ -157,57 +146,25 @@ def from_unpacker(cls, unpacker):
...
+        # on the contrary to the IO serializer.
+        return self._unpack_value(2, '>H')
+
+    def unpack_uchar(self):
+        # The InMemorySerializer implements uchar according to the XDR standard
+        # on the contrary to the IO serializer.
+        return self._unpack_value(1, '>B')
 
 
 def do_string(data):
```
ezavod: I guess that `do_string` has to be different for the 2020 files as well: The `InMemorySerializer` [packs the length as `uint64`](https://github.com/gromacs/gromacs/blob/6edce54dff1102ed59b8ab1c671b18c1702640bd/src/gromacs/utility/inmemoryserializer.cpp#L113-L117). The current implementation discards the first 32 bits (by reading an `int`) and then [tries to interpret the next 32 bits as `unsigned int`](https://github.com/python/cpython/blob/master/Lib/xdrlib.py#L215-L217).

jbarnoud: The IO serializer in gromacs used to do a few odd things that are now fixed or add in a different way...

PR #2441, opened 2020-01-09
TST: pin NumPy version for Travis MacOS

### LINKED ISSUE #2434: Travis systematically fails with python 3.6 on macos
**Expected behavior**

All the travis job pass when the code is valid.

**Actual behavior**

The travis job on macos for python 3.6 fails to install the dependencies. See https://travis-ci.com/MDAnalysis/mdanalysis/jobs/272135973

My understanding is that something requires numpy=0.18 to be installed (conda 4.8.0?) and that the failing job fails to install the requirements of biopython from pip.

tylerjereddy:
Fixes #2434, if I'm lucky

* try pinning the NumPy version used for
the Travis CI MacOS matrix entry, which
is currently failing on develop branch

* at the time of writing, the most
recent NumPy version available from
conda-forge in `1.17.3`, despite
the newer version on PyPI

The exact cause of the CI failure seems hard to dissect from the messy log. CI helpers is maybe to blame for that, but anyway..

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,6 +73,7 @@ matrix:
 
     - os: osx
       env: PYTHON_VERSION=3.6
+           NUMPY_VERSION=1.17.3
 
 # macOS Python 2.7 works in principle but times out on Travis CI:
 # Disabled so that the tests can pass.


tylerjereddy:
Ok, that seems to have done the trick.

PR #2447, opened 2020-01-15
TST: pin setuptools in Python 2.7 CI

### LINKED ISSUE #2444: Python 2.7 build fails at install on travis
**Expected behavior**

CI runs a build with python 2.7

**Actual behavior**

The travis run with python 2.7 fails at install. See <https://travis-ci.com/MDAnalysis/mdanalysis/jobs/275292493?utm_medium=notification&utm_source=email>.

This is dur to setuptools 45.0.0 dropping python 2 support. We need to pin setuptool < 45.0.0 for the python 2 job. See <https://setuptools.readthedocs.io/en/latest/history.html#v45-0-0>.

tylerjereddy:
Fixes #2444, I hope

* pin `setuptools` to a version < `45.0.0`
in Travis CI Python 2.7 matrix entry to
prevent pulling in newer/unsupported versions

Incidentally, there are solutions like [dependabot](https://dependabot.com/) for
pinning & gradually raising/adjusting CI deps versions with automatic PRs for review, but I
think we don't have the bandwidth to set this up right now (I know NumPy was trying this,
SciPy hasn't so far). Not sure if it handles the reverse case of suggesting a pin
when forward-bumps cause breaks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,6 +70,7 @@ matrix:
            CODECOV="true"
            NUMPY_VERSION=1.16
            SETUP_CMD="${PYTEST_FLAGS} --cov=MDAnalysis"
+           PIP_DEPENDENCIES="${PIP_DEPENDENCIES} setuptools<45.0.0"
 
     - os: osx
       env: PYTHON_VERSION=3.6


tylerjereddy:
Travis `2.7` matrix entry looks ok now. I've restarted the (hopefully spurious) Python 3.8 failure in Travis.

Issue #2421, opened 2019-12-22
XYZ writer ignores the `elements` attribute


jbarnoud:
Related to #2420 

**Expected behavior**

The first column of a XYZ file is, [by definition](http://openbabel.org/wiki/XYZ_%28format%29), the element symbol. If the `elements` topology attribute is present, it should be used to fill the first column.

**Actual behavior**

The atom name is always used and the `elements` field is ignored.

**Code to reproduce the behavior**

``` python
import io
import MDAnalysis as mda
from MDAnalysis.tests.datafiles import XYZ

u = mda.Universe(XYZ)
u.add_TopologyAttr('elements', ['C'] * len(u.atoms))

output_io = io.StringIO()
output = mda.lib.util.NamedStream(output_io, 'plop.xyz')
u.atoms.write(output, file_format='XYZ')
print(output_io.getvalue())
```

The first column is filled with `H`, if should be filled with `C`.

**Currently version of MDAnalysis**

- Which version are you using? 0.20.2-dev0 4adbd6c01
- Which version of Python (`python -V`)?  `'3.8.0 | packaged by conda-forge | (default, Nov 22 2019, 19:11:38) \n[GCC 7.3.0]'`
- Which operating system? Ubuntu 18.04

PR #2456, opened 2020-01-17
XYZ elements

### LINKED ISSUE #2420: XYZ parser guesses the elements
**Expected behavior**

The first column of a XYZ file is, [by definition](http://openbabel.org/wiki/XYZ_%28format%29), the element symbol (or the element number in the UniChem variant. The element being provided by the file, `AtomGroup.elements` should be filled and that should be used to guess the mass.

**Actual behavior**

The first column is treated as atom type and atom name, but not as element symbol. The elements are guessed in order to latter guess the masses. The `elements` attribute is not filled.

``` python
import MDAnalysis as mda
from MDAnalysis.tests.datafiles import XYZ

u = mda.Universe(XYZ)
u.atoms.elements
```

```
<ipython-input-22-1d38b5e1bdf4> in <module>
      3 
      4 u = mda.Universe(XYZ)
----> 5 u.atoms.elements

~/dev/mdanalysis/package/MDAnalysis/core/groups.py in __getattr__(self, attr)
   2291             except selection.SelectionError:
   2292                 pass
-> 2293         raise AttributeError("{cls} has no attribute {attr}".format(
   2294             cls=self.__class__.__name__, attr=attr))
   2295 

AttributeError: AtomGroup has no attribute elements
```

- Which version are you using? 0.20.2-dev0 4adbd6c01
- Which version of Python (`python -V`)?  `'3.8.0 | packaged by conda-forge | (default, Nov 22 2019, 19:11:38) \n[GCC 7.3.0]'`

RMeli:
Fixes #2420 and #2421

 - `XYZParser` adds elements attribute
 - `XYZWriter` tries to use the `atoms.elements` attribute (fallback to `atoms.names`)

PR Checklist
------------
 - [x] Tests
 - [x] CHANGELOG updated
 - [x] Issue raised/referenced


package/MDAnalysis/topology/XYZParser.py
```diff
@@ -92,7 +93,7 @@ def parse(self, **kwargs):
 
         # Guessing time
         atomtypes = guessers.guess_types(names)
-        masses = guessers.guess_masses(atomtypes)
+        masses = guessers.guess_masses(names)
```
RMeli: Here I assume that `names` contain element symbols (XYZ file format). As discusses in #2420 this might not always be the case. Should I make this more flexible?

richardjgowers: I think (hope) guess_masses is smart enough to skip a bad value and print a warning?

RMeli:
@richardjgowers  tests for the writer are still missing...

Any suggestions on how to best write them?

richardjgowers:
@RMeli something like...

```python
import io
import MDAnalysis as mda

u = mda.Universe.empty(n_atoms=5)

u.add_TopologyAttr('elements', values=['Te', 'S', 'Ti', 'N', 'Ga'])

ns = mda.lib.util.NamedStream(io.StringIO(), 'file.xyz')

with mda.Writer(ns) as w:
    w.write(u)

names = ''.join(l.split()[0] for l in ns.readlines())

assert names[:-1].lower() == 'testing'
```

RMeli:
@richardjgowers Thanks for the funny suggestion. I meant to ask how to best incorporate them in the existing suite. I put the tests in `TestXYZWriterNames` and the they try writing with both `elements` and `names` attributes.

orbeckst:
Given that @richardjgowers approved and I checked the codecov is nothing to worry about (the low diff coverage comes from changes in indentations that don't even show up in these files – weird) so I am merging,

PR #2457, opened 2020-01-18
Removes explicit use of Bio.Alphabet (deprecation pending)

### LINKED ISSUE #2448: Bio.Alphabet Deprecation
When running some MDAnalysis tests I run into the following `PendingDeprecationWarning`:
```
PendingDeprecationWarning: We intend to remove or replace Bio.Alphabet in 2020, ideally avoid using it explicitly in your code. Please get in touch if you will be adversely affected by this. https://github.com/biopython/biopython/issues/2046
```

- MDA: 0.20.2-dev0
- Python: 3.6.7
- OS: macOS 10.5.2

RMeli:
Fixes #2448

 - Remove `Bio.Alphabet` from `analysis/align.py` by setting the standard gap symbol `-`
 - Remove `BioAlphabet` from `core/topologyattrs.py` by using the default alphabet argument in Bio.Seq.Seq

Test still pass, but they might not cover all different user cases and therefore such changes might break things. Advice on possible user cases that might break would be greatly appreciated.

PR Checklist
------------
 - [x] Tests
 - [x] CHANGELOG updated
 - [x] Issue raised/referenced


RMeli:
Changelog updated. 

I am not sure I know what I'm doing either here, this is why I was asking for advice on what might break outside of the provided tests.

PR #2458, opened 2020-01-21
Add radius_cut_q as a method to contacts.Contacts

lilyminium:
 - adds `radius_cut_q` as a method to contacts.Contacts

It seems like this method was made for contacts.Contacts anyway.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/analysis/contacts.py
+++ b/package/MDAnalysis/analysis/contacts.py
@@ -438,8 +438,6 @@ class Contacts(AnalysisBase):
         # compute distance array for a frame
         d = distance_array(self.grA.positions, self.grB.positions)
 
-        y = np.empty(len(self.r0) + 1)
-        y[0] = self._ts.frame
         for i, (initial_contacts, r0) in enumerate(zip(self.initial_contacts,
                                                        self.r0), 1):
             # select only the contacts that were formed in the reference state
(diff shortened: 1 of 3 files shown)


package/MDAnalysis/analysis/contacts.py
```diff
@@ -397,10 +397,15 @@ def __init__(self, u, selection, refgroup, method="hard_cut", radius=4.5,
...
         super(Contacts, self).__init__(self.u.trajectory, **basekwargs)
 
+        self.fraction_kwargs = kwargs if kwargs is not None else {}
+
         if method == 'hard_cut':
             self.fraction_contacts = hard_cut_q
         elif method == 'soft_cut':
             self.fraction_contacts = soft_cut_q
+        elif method == 'radius_cut':
+            self.fraction_contacts = radius_cut_q
```
richardjgowers: I think here it's cleaner to use [function currying](https://en.wikipedia.org/wiki/Currying).

If you do `self.fraction_contacts = functools.partial(radius_cut_q, radius=radius)` it should work the same

lilyminium: Sure, done :-)

package/MDAnalysis/analysis/contacts.py
```diff
@@ -424,28 +429,25 @@ def __init__(self, u, selection, refgroup, method="hard_cut", radius=4.5,
...
+    def _prepare(self):
+        self.timeseries = np.empty((self.n_frames, len(self.r0)+1))
 
     def _single_frame(self):
+        self.timeseries[self._frame_index][0] = self._ts.frame
+
         # compute distance array for a frame
         d = distance_array(self.grA.positions, self.grB.positions)
 
         y = np.empty(len(self.r0) + 1)
```
orbeckst: I think this and the next line should go as `y` is not used anymore.

PR #2473, opened 2020-01-27
Fixes to ENCORE's ClusterCollection and Cluster output

### LINKED ISSUE #2464: encore ClusterCollection.get_ids error and unexpected representations
**Expected behavior**

1. `encore.clustering.ClusterCollection.ClusterCollection.get_ids()` works without an error.

2. `encore.clustering.ClusterCollection.ClusterCollection.__repr__` makes it look indexable, so I expect that I can index it.

3. `encore.clustering.ClusterCollection.Cluster.__repr__` looks like a numpy array, so I expect that I can index it.

**Actual behavior**

1. It raises an AttributeError because Clusters do not have an `idn` attribute. Clusters just call it `id`.

https://github.com/MDAnalysis/mdanalysis/blob/972a8b1893b5e353b9d796b15d6fcdd022a93f00/package/MDAnalysis/analysis/encore/clustering/ClusterCollection.py#L218

2. Not indexable. The clusters are in `encore.clustering.ClusterCollection.ClusterCollection.clusters`. Perhaps a `repr` that looks like `<ClusterCollection with 49 clusters>` would be more informative.

3. Not indexable. `repr` returns the Cluster's `elements`, which are a numpy array. Perhaps a `repr` that gives something like `<Cluster with 5 elements, centroid=1, id=1>` would be more informative.

[Notebook](https://lilyminium.github.io/UserGuide/examples/analysis/trajectory_similarity/clustering_ensemble_similarity.html)

mtiberti:
Fixes #2464 

 - fixes `ClusterCollection.get_ids`
- changes output of `__repr__` so that it's more compact and also handles non-initialised instances

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


package/MDAnalysis/analysis/encore/clustering/ClusterCollection.py
```diff
@@ -215,7 +219,7 @@ def get_ids(self):
         ids : list of int
         list of cluster ids
         """
-        return [v.idn for v in self.clusters]
```
richardjgowers: This change needs a test, typos like this would have been caught

mtiberti: Thanks - I've added a test for that and a couple of other explicit tests for similar methods of the class

richardjgowers:
cool thanks @mtiberti

PR #2474, opened 2020-01-27
Density docs + padding warning (Issue 2372)

### LINKED ISSUE #2372: density profiles from mda.analysis.density.density_from_universe does not conserve number of atoms
Hello 

Expected behavior:
I'm currently trying to calculate liquid and vapour densities from liquid slab MD simulations using density profiles. For this purpose I read a trajectory into an universe and add the center_in_box(center="mass", wrap=True) and a subsequent pack_into_box transformation, to get the liquid slab centered in the original box.

After this I call mda.analysis.density.density_from_universe() with a predefined box and center (see Issue #1928). The predefined box has padding in xy direction but not in z direction and the delta is chosen to exactly fit the box in z direction (as this is the elongated axis along which we go from vapour to liquid slab back to vapour). As a check, I calculate the sum over the returned three dimensional array and compare it to number of atoms in the trajectory. 

Actual behavior:
The actual behaviour is that the number of atoms is smaller than the number of atoms in the trajectory. This behaviour is independent of the padding in z direction (I tried to add padding), but i found out that its dependent on the delta keyword. Only for delta = 1 I get the right number of atoms. If I choose delta>1 I get lower number of atoms and for delta<1 I get higher number of atoms. This is consistent over multiple trajectories (simulations at different temperature but same number of atoms). 

``` python
    

    import MDAnalysis.analysis.density as DENS
    from MDAnalysis.transformations import center_in_box
    import MDAnalysis as mda
    
    box=None
    def put_in_box(atoms):
        # Put atoms in box during each call of the trajectory
        def wrapped(ts):
            ts.positions = atoms.pack_into_box(box=None, inplace=True)
            return ts
        return wrapped


    sys = mda.Universe(LAMMPS_DATAFILE, 
                       XTC/XYZ_TRAJECTORY,
                       atom_style='id type x y z')
    
    # If trajectory does not hold box infos, define yourself
    # This could usually also be done with the datafile...
    if box is not None:
        sys.universe.dimensions = box

    # PBC
    sys.atoms.pack_into_box(box=None, inplace=True) 

    # Center slap at COM + PBC
    workflow = [center_in_box(sys.atoms, center='mass', wrap=True), 
                put_in_box(sys.atoms)]       
    sys.trajectory.add_transformations(*workflow)

    # Get box dimensions and padding in xy directions. Define delta, so that
    # an integer number of bins fits in z direction
    xydim = sys.dimensions[0]
    zdim  = sys.dimensions[2]
    pad = 0.0
    zpad = 0.0
    #delta = (zdim+zpad)/(int(zdim))
    delta = 1.0 
    # Call Density object, disregard first 300 frames and define box manually.
    Dens = DENS.density_from_Universe(sys, atomselection="all", start=300, 
                                      xdim=xydim+pad, 
                                      ydim=xydim+pad, 
                                      zdim=zdim+zpad, 
                                      gridcenter=[xydim/2,xydim/2,zdim/2],
                                      delta=delta) 
    # Calculate Centers of density profile in z direction and initialise
    # density vector
    _, _, edges_z = Dens.edges
    dim_z = len(edges_z) - 1
    dens_z = np.zeros((dim_z,1))

    # Calculate Density profile and give warning if atoms are less
    # than expected
    for i in range(dim_z):
        dens_z[i] = np.sum(Dens.grid[:, :, i])
    print("Atoms: ", np.sum(dens_z))
    print("Atoms: ", np.sum(Dens.grid[:, :, :]))
....

```

Currently version of MDAnalysis:
- Which version are you using? 0.19.2
- Which version of Python: Python 3.73
- Which operating system: Linux, Ubuntu 18.04.3

IAlibay:
Fixes #2372

 - Clarifies in docstrings that density_from_Universe returns a `physical` density, and adds user information on how to recover the underlying histogram.
 - Adds a user warning for when the `padding` keyword is used with user defined grids.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/analysis/density.py
```diff
@@ -564,8 +564,10 @@ def density_from_Universe(universe, delta=1.0, atomselection='name OH2',
                           gridcenter=None, xdim=None, ydim=None, zdim=None):
     """Create a density grid from a :class:`MDAnalysis.Universe` object.
 
-    The trajectory is read, frame by frame, and the atoms selected with `atomselection` are
-    histogrammed on a grid with spacing `delta`.
+    The trajectory is read, frame by frame, and the atoms selected with
+    `atomselection` are histogrammed on a grid with spacing `delta`.
+    A physical density of units [Angstroem^{-3}] is returned (see
```
RMeli: Typo: Angstroem

package/MDAnalysis/analysis/density.py
```diff
@@ -689,6 +691,23 @@ def density_from_Universe(universe, delta=1.0, atomselection='name OH2',
       (It should be noted that the `padding` keyword is not used when a user
       defined grid is assigned).
 
+    As detailed above, the :class:`Density` object returned contains a
+    physical density in units of Angstroem^{-3}. If you are interested in
```
RMeli: (As above)

package/MDAnalysis/analysis/density.py
```diff
@@ -689,6 +691,23 @@ def density_from_Universe(universe, delta=1.0, atomselection='name OH2',
...
+
+      physical_density = density_from_Universe(universe, delta=1.0,
+                                               atomselection='name OW')
+
+      probability_density = physical_density / physical_density.grid.sum()
+
+    Similarly, if you would like to recover a grid containing a histogram of
+    atom counts, simply multiply by the volume::
+
+      atom_count_histogram = physical_density * (physical_density.delta[0] ** 3)
```
RMeli: Is `delta` the same in all directions?

IAlibay:
Ooo this brings about a good discussion point, the MDA docs use both **Angstrom** and **Angstroem**. The former is used in more cases and the assumption is that py2.7 isn't fully utf-8 compliant.

Any thoughts here?

package/MDAnalysis/analysis/density.py
IAlibay: Good shout :)

So the code calls `fixedwidth_bins`, which allegedly will always return a set of bins that obey the input `delta` value. Since a single `delta` currently can only be defined by users, delta is the same in all directions. However, maybe it might be safer to not make this assumption in the event that future versions of the density code allow for non-uniform bin widths?

IAlibay: @RMeli this should be fixed now

package/MDAnalysis/analysis/density.py
IAlibay: To make things easier in future efforts, I have now standardised the entire of density.py to Angstrom (matches what is declared in units.py).

package/MDAnalysis/analysis/density.py
IAlibay: To make things easier in future efforts, I have now standardised the entire of density.py to Angstrom (matches what is declared in units.py).

IAlibay:
> Ooo this brings about a good discussion point, the MDA docs use both **Angstrom** and **Angstroem**. The former is used in more cases and the assumption is that py2.7 isn't fully utf-8 compliant.
> 
> Any thoughts here?

To make things easier in future efforts, I have now standardised the entire of density.py to Angstrom (matches what is declared in units.py).

We might want to go back and standardise things across other parts of MDA at a later date?

Issue #2443, opened 2020-01-13
Release 1.0


richardjgowers:
I think we're due another release.  I think the only ongoing thing is @lilyminium 's work on docs, so ideally this release captures all changes necessary to the package for their doc work. (ie if this is an awful idea tell me and we can postpone).

Looking at #2303 (esp https://github.com/MDAnalysis/mdanalysis/issues/2303#issuecomment-529680541 ) it looks like the plan is that this release is called 1.0 and is the last python 2.7 release.  Work then continues ~~(supporting only py 3)~~ as 1.x until we finish the annoying API change stuff and we call the result 2.0 (and drop py 2)

following feedback, todo:
 - [x] flags @richardjgowers 
 - [x] ncdf @IAlibay 
 - [x] tpr jazz @jbarnoud 
 - [x] #2408  @lilyminium 
 - [x] #2382 @lilyminium 
 - [x] instant selectors removal @richardjgowers

IAlibay:
As previously discussed with @richardjgowers I'd like to at least fix the NCDF reader portion of #2323 and #2327 before release 1.0.

I'll hopefully get this done within the next couple of weekends if that's ok with your schedule?

richardjgowers:
![doeeet](https://i.giphy.com/media/bKnEnd65zqxfq/200.webp)

jbarnoud:
With gromacs 2020 released, I'd like to see #2431 merged before 1.0. Especially as, in the current state there will be unreadable errors as the TPR parser expects a TPR from gromacs 2020-beta that are very different.

lilyminium:
Sounds good! Will you be going through and looking at currently open PRs before release? There's some GSOC stuff and I would prefer to have #2408 in the release, as the current ITPParser is very flawed. #2382 would be nice too.

orbeckst:
In an ideal world we would have everything ready for 1.0. 

But the things mentioned above should be in 1.0 and we also wanted to remove the flags #782 . It would be easier to make this 0.21 and then follow up with 1.0. A release is overdue, and it might as well be 0.21 and then we follow up with 1.0.

Btw,  @richardjgowers – you looked at the wrong comment, the last one was https://github.com/MDAnalysis/mdanalysis/issues/2303#issuecomment-543937601 and the official roadmap is the (ambitious) https://www.mdanalysis.org/2019/11/06/roadmap/. 
[...]

PR #2482, opened 2020-01-30
Removing deprecated reader code for v1.0

IAlibay:
Partly addresses #2327 and #2443 

As far as I am aware, the NetCDF writer is the only part of the reader/writers that has code set for removal by v1.0.0, please let me know if this isn't the case.

 - Removes support for the `degrees` unit for `cell_angles` in the AMBER NetCDF reader.

Note: this does not completely fix the NetCDF reader/writer (`scale_factor` writer support is needed, and we really should switch the writing of velocities from unscaled to the same factor used by the AMBER MD engines use [see #2327]). However, assuming a non API complete v1.0.0, further fixes can probably be done by the next release.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -19,6 +19,8 @@ mm/dd/yy richardjgowers, kain88-de, lilyminium, p-j-smith, bdice, joaomcteixeira
   * 0.21.0
 
 Fixes
+  * Removes support for reading AMBER NetCDF files with `cell_angle` units set
+    to `degrees` instead of the convention agreed `degree` (Issue #2327). 
   * Removes the MassWeight option from gnm (PR #2479).
   * AtomGroup.guess_bonds now uses periodic boundary information when available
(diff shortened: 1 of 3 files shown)


IAlibay:
> LGTM if CI passes. The deprecation was issued in August 2019---do we have a timeline for `deprecation->removal` workflow established (i.e., two years? two releases minimum? etc.).

Thanks for the review!

With regards to the `deprecation->removal` workflow, I am currently going through all the things that need to be removed for v1.0.0, and there doesn't seem to be much consistency. For the most part a minimum of two versions have been released before removal (this particular case has the shortest `deprecation->removal` time, deprecated in 0.20 -> 0.20.1 -> removal 1.0). 

That being said, if the consensus is that this removal is too soon, I don't mind shelving it for a later date.

PR #2484, opened 2020-01-31
Changes encore.hes() details output

### LINKED ISSUE #2465: encore.hes returns an numpy array of a dict
**Expected behavior**

When I call `hes, details = encore.hes(ensemble, details=True` then `details` is readable.

**Actual behavior**

`details` is a `np.ndarray` of a `dict`. 

https://github.com/MDAnalysis/mdanalysis/blob/972a8b1893b5e353b9d796b15d6fcdd022a93f00/package/MDAnalysis/analysis/encore/similarity.py#L940-L945

This is not indexable either with numbers or keys. I don't know how to get information out from it. It's probably easier to just not convert it to an array.

[Notebook showing failure to index `details`](https://lilyminium.github.io/UserGuide/examples/analysis/trajectory_similarity/harmonic_ensemble_similarity.html#Calculating-harmonic-similarity)

mtiberti:
Fixes #2465 

- Removed arrayfication of encore.hes details 
 - I'm also changing the condition for which details are returned - they are now always returned without the need to use the `details` flag. This is more consistent with the behaviour of `ces` and `dres`. Previously, `None` was returned instead if the `details` flag was `False`.
- Changed docs accordingly

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


orbeckst:
@lilyminium would you mind having a quick look and comment if this PR fixes your original issue?

Please feel free to ping me when you need a final review for merge. Thanks.

lilyminium:
Looks good to me, thanks @mtiberti!

package/CHANGELOG
```diff
@@ -19,6 +19,7 @@ mm/dd/yy richardjgowers, kain88-de, lilyminium, p-j-smith, bdice, joaomcteixeira
   * 0.21.0
 
 Fixes
+  * encore.hes() returns non-array details (Issue #2465)
```
orbeckst: This is not a fix but a Change because for a user who wants to process the output something changes – please move to Change section.

package/CHANGELOG
```diff
@@ -48,6 +49,7 @@ Fixes
   * Added parmed to setup.py
 
 Enhancements
+  * encore.hes() always returns details, consistently with ces and dres (Issue #2465)
```
orbeckst: Are these really two different things that the PR does? To me it looks like a single entry under *Changes*.

package/MDAnalysis/analysis/encore/similarity.py
```diff
@@ -76,17 +76,16 @@
...
 
     >>> ens1 = Universe(PSF, DCD)
     >>> ens2 = Universe(PSF, DCD2)
-    >>> print encore.hes([ens1, ens2])
-    (array([[        0.        ,  38279683.95892926],
-            [ 38279683.95892926,         0.        ]]), None)
+    >>> HES, details = encore.hes([ens1, ens2])
+    >>> print HES
+    [[        0.        ,  38279683.95892926],
+    [ 38279683.95892926,         0.        ]]
```
orbeckst: add a `.. versionchanged:: 1.0.0` with the change description – when the API changes we need to document it

package/MDAnalysis/analysis/encore/similarity.py
```diff
@@ -721,7 +720,6 @@ def hes(ensembles,
         cov_estimator="shrinkage",
         weights='mass',
         align=False,
-        details=False,
```
orbeckst: removal of arguments also needs to be documented — add a `versionchanged`.

Also add a note to CHANGELOG – along the lines that hes() does not take `details` kwarg anymore and that the output is now a ...

orbeckst: Remove `details` from the doc string!

mtiberti: Thanks, definitely - not sure how I missed this one. 

Should I add just one `versionchanged` entry overall? Or do you mean `deprecated` for the removal of the option?

orbeckst: Versionchanged states that the option was removed.

Deprecated would mean that the option is still available but should not be used. As the option is gone, it has to be a versionchanged.

orbeckst:
Can you please also resolve the conflict on CHANGELOG so that it can be merged cleanly?

package/CHANGELOG
orbeckst: Also add to Changes that hes() does not take the details keyword argument anymore.

mtiberti:
sorry - I hadn't realised the build was failing because of wrong spacing in the docstring. It should be fixed now.

orbeckst:
@lilyminium , does this look good to you know? (I don't want to squash-merge without giving you a chance to look, given that @richardjgowers asked for your opinion.)

package/MDAnalysis/analysis/encore/similarity.py
```diff
@@ -76,17 +76,16 @@
...
 To calculate the Harmonic Ensemble Similarity (:func:`hes`)
 two ensemble objects are first created and then used for calculation: ::
 
     >>> ens1 = Universe(PSF, DCD)
     >>> ens2 = Universe(PSF, DCD2)
-    >>> print encore.hes([ens1, ens2])
-    (array([[        0.        ,  38279683.95892926],
-            [ 38279683.95892926,         0.        ]]), None)
+    >>> HES, details = encore.hes([ens1, ens2])
+    >>> print HES
```
lilyminium: Could you please just change this to `print(HES)` for python 3 users?

orbeckst:
Many thanks @lilyminium

mtiberti:
Thank you both!

Issue #2491, opened 2020-02-01
PendingDeprecationWarning instead of DeprecationWarning


RMeli:
I was looking at #2472 and found out that there are two sub-packages for hydrogen bond analysis: `MDAnalysis/analysis/hbond` and `MDAnalysis/analysis/hydrogenbonds`. In `MDAnalysis/analysis/hbond` there is this warning:

```
warnings.warn(
            "This module will be deprecated in version 1.0."
            "Please use MDAnalysis.analysis.hydrogenbonds.hbond_analysis instead.",
            category=DeprecationWarning
        )
```

I think a `PendingDeprecationWarning` is more appropriate in this case.

IAlibay:
This is code that is in progress of getting removed as part of #2443

RMeli:
Should it be removed? The message says that it will be deprecated in version `1.0`, not removed...

IAlibay:
That is a very good point, I hadn't read it properly it seems.

Maybe one of the coredevs can clarify here, I'm glad I hadn't pushed part 2 of my removals yet :sweat_smile:

RMeli:
I think if we are going for `0.21` it should be changed to `PendingDeprecationWarning`, while if we are going for `1.0` we should simply change the error message to something like this:
```
warnings.warn(
            "This module is deprecated."
            "Please use MDAnalysis.analysis.hydrogenbonds.hbond_analysis instead.",
            category=DeprecationWarning
        )
```

IAlibay:
Yes, I agree, I'll add the change assuming v1.0 as part of the removal changes.

orbeckst:
In this case PendingDeprecationWarning seems more appropriate than DeprecationWarning. Is this an existing warning in 2 and 3?

1.0 will contain both hbonds modules as we want to make it easy for users to keep the last 2-capable version of MDA for their projects. 

For 2.0 we would want to remove hbonds — if we can.

RMeli:
Yes, `PendingDeprecationWarning` is also [available in Python 2.7](https://docs.python.org/2/library/exceptions.html#exceptions.PendingDeprecationWarning). I can change it to `PendingDeprecationWarning` if there is a version 0.21; otherwise I'll live it as it is for 1.0.

I checked other deprecation warnings and this is the only one where "pending" is more appropriate.

orbeckst:
Thanks for checking. It seems that we’re heading for 1.0 so leave it.

PR #2486, opened 2020-01-31
Remove save() and changes filename in AlignTraj

IAlibay:
Completes #2099 
Addresses part of #1745 #2443 

Changes made in this Pull Request:
 - `save()` method has now been removed.
 - `filename` should now default to the current working directory.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?
 - [x] Issue raised/referenced?

PR #2494, opened 2020-02-02
Remove save part2

### LINKED ISSUE #2492: Deprecation of _reformat_hb in hbonds/hbond_analysis.py for v1.0
Currently going through and doing the required removals/deprecations for #1745 and I noticed that _reformat_hb was set for deprecation as part of #1339 

```
        .. deprecated:: 1.0
           This is a compatibility layer so that we can provide the same output
           in timeseries as before. However, for 1.0 we should make timeseries
           just return _timeseries, i.e., change the format of timeseries to
           the un-ambiguous representation provided in _timeseries.
```

It seems like some change is required here prior to v1.0.0, but it's unclear to me as to what.

IAlibay:
Partly addresses #1745 (completes `save()` removal with #2487 and #2486 )

Also addresses #2491, since the deprecation was unclear and it seems like there were no releases between the announced deprecation of `analysis/hbonds/hbond_analysis.py` and now, the deprecation was set as effective v1.0 and set for removal in v2.0 (this is just a placeholder date).

Fixes #2492 

 - Set deprecation date for `analysis/hbonds/hbond_analysis.py`.
 - Removed `save_table` from :class:`HydrogenBondAnalysis`.
 - Removed `save_results` from :class:`HydrogenBondAutoCorrel`.
 - Removed `filename`, `store` and `save_result` from :class:`PSAnalysis`.
 - Deprecation notice for :meth:`HydrogenBondAnalysis._reformat_hb` has been removed, entire class will be removed in 2.0 (#2492 )

Things left to do:
 - [x] Confirm the removal date for `MDAnalysis/analysis/hbonds/hbond_analysis`.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


orbeckst:
Set removal date for hbonds to 2.0 — that’s the plan.

orbeckst:
I didn’t check why CI fails. But ping me when you need a review.

IAlibay:
@orbeckst this should be good for review now

package/MDAnalysis/analysis/hbonds/hbond_analysis.py
```diff
@@ -30,7 +30,7 @@
 :Copyright: GNU Public License v3
 
 ..Warning:
-    This module will be deprecated in version 1.0.
+    This module will be deprecated in version 2.0.
```
orbeckst: Should be "will be removed in 2.0"

package/MDAnalysis/analysis/hbonds/hbond_analysis.py
```diff
@@ -567,7 +572,8 @@ def __init__(self, universe, selection1='protein', selection2='all', selection1_
         super(HydrogenBondAnalysis, self).__init__(universe.trajectory, **kwargs)
 
         warnings.warn(
-            "This class will be deprecated in version 1.0."
+            "This class is deprecated as of MDANalysis version 1.0 and will "
```
orbeckst: MDAnalysis

package/MDAnalysis/analysis/psa.py
```diff
@@ -1509,31 +1507,19 @@ def run(self, **kwargs):
...
-
-             .. deprecated:: 0.19.0
-                `store` will be removed together with :meth:`save_results` in 1.0.0.
-
-        filename : str
-             string, filename to save :attr:`PSAnalysis.D`
 
-             .. deprecated:: 0.19.0
-                `filename` will be removed together with :meth:`save_results` in 1.0.0.
+        .. deprecated:: 0.19.0
```
orbeckst: Remove this deprecation note – they are gone now, aren't they?

package/MDAnalysis/analysis/hbonds/hbond_analysis.py
IAlibay: Should be fixed in 4c7f557

package/MDAnalysis/analysis/hbonds/hbond_analysis.py
IAlibay: Fixed in 4c7f557

package/MDAnalysis/analysis/psa.py
IAlibay: Thanks for the clarification here, this got brought up in #2487 but we weren't sure what the appropriate approach was here. Fixed in 4c7f557 and I'll update #2487 accordingly.

orbeckst:
Yay, less code!

Issue #268, opened 2015-05-04
Add 'global' keyword to selections


mnmelo:
When performing atom selections from an AtomGroup instance all subexpressions are always evaluated within the set of the atoms of that instance. However, that isn't always intuitive:
```python
complex_preselected_solvent_group.selectAtoms("around 10 protein")
# This returns an empty AtomGroup because there is
# no protein in complex_preselected_solvent_group

# It can always be worked around...
universe.selectAtoms("group solv and around 10 protein", solv=complex_preselected_solvent_group)
```
A `global`keyword could be added to make the subsequent subexpressions be evaluated in the universe context, instead of the calling AtomGroup instance:
```python
complex_preselected_solvent_group.selectAtoms("around 10 global protein")
# This would work as expected
```

This construct would also eliminate the need for the `fullgroup` keyword, as we could now use `global group` instead, with equal legibility.
It also makes more sense now that we have the `same ... as` construct:
```python
# Suppose we have multiple protein copies in our system and want the active-site residues
# of those copies that are close to a POPC membrane
active_site_residues.selectAtoms("same fragment as global (protein and around 10 resname POPC)")
```

This is not a serious problem, since it can always be worked around by issuing the selection from the universe object, and using the `group` keyword to bring in the AtomGroup's atoms. But it makes for more readable code, and easier to write.

Opinions?

orbeckst:
I don't see anything wrong with this suggestions and it certainly makes for more readable selections, so yes, +1 ;-).

PR #2500, opened 2020-02-04
FullGroup selection keyword removal

IAlibay:
Completes #268

 - Removes the `fullgroup` keyword from selections (and related test).

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


IAlibay:
So the test removal wasn't necessary since it was just testing that an error would occur, so if we want I can always add it back in.

PR #2515, opened 2020-02-09
Issue 782 removeflags

### LINKED ISSUE #782: remove flag registry

richardjgowers:
Fixes #782

 - removed flags

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


package/MDAnalysis/core/topologyattrs.py
```diff
@@ -772,7 +771,7 @@ def get_segments(self, sg):
 
     @warn_if_not_unique
     @check_pbc_and_unwrap
-    def center_of_mass(group, pbc=None, compound='group', unwrap=False):
+    def center_of_mass(group, pbc=False, compound='group', unwrap=False):
         """Center of mass of (compounds of) the group.
 
         Computes the center of mass of :class:`Atoms<Atom>` in the group.
```
IAlibay: Would it be worth adding in the parameter docstring that pbc defaults to `False` now?

richardjgowers: It will have always defaulted to False, unless someone had used flags, which seems unlikely.

package/MDAnalysis/coordinates/XYZ.py
```diff
@@ -148,10 +147,8 @@ def __init__(self, filename, n_atoms=None, atoms=None, convert_units=None,
         """
```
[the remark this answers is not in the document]
richardjgowers: This is on the Writer, so this being true would convert inputs to the native units ... for xyz?  I'm not sure if XYZ has a native unit? So I'm not going to break anything here, but I'm not sure I want to document the fact that this option exists?

package/MDAnalysis/core/topologyattrs.py
IAlibay: That's fair 🙃

package/MDAnalysis/coordinates/XYZ.py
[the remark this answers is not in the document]
RMeli: XYZ units are _generally_ in angstroms and that's how most visualization packages interpret them (VMD for example).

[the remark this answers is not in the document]
IAlibay: So MDA does seem to assume `ps/Angstrom` on read, and the default writer also assume those units on the conversion. I guess the only case I could think of where this might be useful, is if for some reason you created a `Universe` with non-base units, and then wanted to use the writer to have a non-`ps/Angstrom` write. With the removal of flags, I'm not sure how likely that would be to happen though...

package/MDAnalysis/core/groups.py
```diff
@@ -848,13 +839,13 @@ def center_of_geometry(self, pbc=None, compound='group', unwrap=False):
...
         ----------
-        pbc : bool or None, optional
+        pbc : bool, optional
             If ``True`` and `compound` is ``'group'``, move all atoms to the
             primary unit cell before calculation. If ``True`` and `compound` is
             ``'segments'`` or ``'residues'``, the center of each compound will
             be calculated without moving any :class:`Atoms<Atom>` to keep the
             compounds intact. Instead, the resulting position vectors will be
-            moved to the primary unit cell after calculation.
+            moved to the primary unit cell after calculation. Default False.
```
orbeckst: Typeset False as 
```reST
``False``
```

orbeckst:
@IAlibay and @RMeli thanks for the reviews!

richardjgowers:
@orbeckst ok docs now all done, grepping doesn't show any more flag stuff in the repo

orbeckst:
Thanks go ahead and do what needs to be done. 

(I’m traveling and on mobile...)

PR #2516, opened 2020-02-09
Minimum mass for hydrogen guess in HydrogenBondAnalysis

### LINKED ISSUE #2472: Have a min_mass for HydrogenBondAnalysis.guess_hydrogens

``HydrogenBondAnalysis.guess_hydrogens`` only looks for a maximum mass and minimum charge for hydrogen-like atoms.

https://github.com/MDAnalysis/mdanalysis/blob/b5cb49fc06c05d4038f75ba98c66dc30f856d303/package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py#L268-L274

However, we know that sometimes the mass guesser will give masses of 0.0 if it can't figure out what a particle is (#2348). This seems like it would lead to a fair few false positives.

Have a `min_mass` keyword, or just don't choose atoms with a mass of 0.0

Making users validate the output of ``HydrogenBondAnalysis.guess_hydrogens`` themselves.

**Additional context**

RMeli:
Fixes #2472

 - Added a `min_mass` parameter (set to `0.9` by default) to `guess_hydrogens` in `HydrogenBondAnalysis`

When the mass can't be guessed it is usually set to `0.0`. This could lead to an erroneous identification of hydrogen atoms.
 
PR Checklist
------------
 - [x] Tests?
 - [x] CHANGELOG updated?


package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -263,14 +268,20 @@ def guess_hydrogens(self, selection='all', max_mass=1.1, min_charge=0.3):
...
+        .. versionchanged: 1.0.0
+            Added `min_mass` parameter to specify minimum mass (Issue #2472)
         """
 
+        assert min_mass < max_mass
+
         ag = self.u.select_atoms(selection)
         hydrogens_ag = ag[
-            np.logical_and(
+            np.logical_and.reduce((
```
richardjgowers: Maybe being pedantic here, but would it be better to apply to minimum mass filter after the max mass filter?  In terms of number of comparisons you want to apply the most selective filter (max_mass I imagine) first.  Can you check if it's more performant to slice a large atomgroup (to a "medium" size) by max_mass then apply charge, then apply min mass?

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -263,14 +268,20 @@ def guess_hydrogens(self, selection='all', max_mass=1.1, min_charge=0.3):
         Alternatively, this function may be used to quickly generate a :class:`str` of potential hydrogen atoms involved
         in hydrogen bonding. This str may then be modified before being used to set the attribute
         :attr:`hydrogens_sel`.
+
+        .. versionchanged: 1.0.0
+            Added `min_mass` parameter to specify minimum mass (Issue #2472)
         """
 
+        assert min_mass < max_mass
```
richardjgowers: I think we probably don't want assertionerrors in code, maybe valueerror is required here.

testsuite/MDAnalysisTests/analysis/test_hydrogenbonds_analysis.py
```diff
@@ -139,7 +139,56 @@ def test_guess_donors(self, h):
...
+        h = HydrogenBondAnalysis(universe, **self.kwargs)
+        return h
+
+    def test_guess_hydrogens(self, h):
+
+        ref_hydrogens = "(resname TIP3 and name H1) or (resname TIP3 and name H2)"
+        hydrogens = h.guess_hydrogens(selection='all')
+        assert hydrogens == ref_hydrogens
+
+    def test_guess_hydrogens_min_mass(self, h):
```
IAlibay: Considering these three tests are very similar in construction, would using pytest.mark.parametrize be worth it here?

testsuite/MDAnalysisTests/analysis/test_hydrogenbonds_analysis.py
```diff
@@ -139,7 +139,56 @@ def test_guess_donors(self, h):
...
+
+        hydrogens = h.guess_hydrogens(selection='all', max_mass=0.95)
+        assert hydrogens == ""
+
+    def test_guess_hydrogens_min_charge(self, h):
+
+        hydrogens = h.guess_hydrogens(selection='all', min_charge=1.0)
+        assert hydrogens == ""
+
+    def test_guess_hydrogens_min_max_mass(self, h):
```
IAlibay: Would be good to have a better user error message and use `match` to check it's being raised properly here.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
RMeli: Do you mean something like the following?
```python
hag = ag[ag.masses < max_mass]
hag = hag[hag.charges > min_charge]
hag = hag[hag.masses > min_mass]
```

On the provided files `waterPSF, waterDCD` this is actually slightly slower (`21.9 µs ± 326 ns` compared to `17.6 µs ± 770`) than the implemented solution, possibly because `np.logical_and.reduce` performs lazy evaluation? In contrast

```python
hydrogens_ag = ag[
            np.logical_and.reduce((
                ag.masses < max_mass,
                ag.charges > min_charge,
                ag.masses > min_mass
            ))
        ]
```
has a comparable speed (`17.7 µs ± 951 ns`). However, `len(ag)==15` which might not be very informative; can you suggest a larger file for this test?

testsuite/MDAnalysisTests/analysis/test_hydrogenbonds_analysis.py
RMeli: They are testing different parameters, not different parameters' values. How would you suggest to use `pytest.mark.parametrize` in this case?

testsuite/MDAnalysisTests/analysis/test_hydrogenbonds_analysis.py
RMeli: Yes, I agree. I changed the assertion to `ValueError` exception ([as suggested above](https://github.com/MDAnalysis/mdanalysis/pull/2516#discussion_r376797303)) and now the test check the error message.

testsuite/MDAnalysisTests/analysis/test_hydrogenbonds_analysis.py
IAlibay: So I might be misunderstanding what you are trying to do, so ignore me if that's the case, but I think something like this should work:

```python
pytest.mark.parametrize('min_mass, max_mass, min_charge', [
        (1.05, 1.1, 0.3),
        (0.9, 0.95, 0.3),
        (0.9, 1.1, 1.0)
    ])
def test_guess_hydrogens_criteria_noreturn(self, h, min_mass, max_mass, min_charge):
    hydrogens = h.guess_hydrogens(selection='all', min_mass=min_mass, max_mass=max_mass, min_charge=min_charge)
    assert hydrogens == ""
```

So this actually brings another point, shouldn't the code raise a `ValueError` if min_mass > max_mass is provided?

RMeli: I see. I didn't like this too much because it's not immediately clear which parameter  is changing with respect to the default, but on a second thought less repetitive code is better therefore I adopted this stile.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
richardjgowers: Huh you're right, numpy strikes again

PR #2518, opened 2020-02-09
Fix PDBWriter with StringIO for invalid coordinates

### LINKED ISSUE #2512: PDBWriter incompatible with StringIO if atom coordinates go out of range
**Expected behavior**

Assuming that PDBWriter is supposed to support a StringIO object as a valid destination, I expect that attempting to write a PDB with atom coordinates out of range will raise the "PDB files must have coordinate values between [...]" exception.

**Actual behavior**

The check attempts to `os.remove` the StringIO object and does not catch the resulting exception, so the coordinate exception never gets raised.

```
~/narupa-protocols/narupa-protocol/python-libraries/narupa-ase/examples/nglclient.py in mda_to_pdb_str(universe)
     54 def mda_to_pdb_str(universe: mda.Universe):
     55     with StringIO() as str_io, mda.coordinates.PDB.PDBWriter(str_io) as writer:
---> 56         writer.write(universe.atoms)
     57         pdb = str_io.getvalue()
     58     return pdb
~/anaconda3/envs/narupa-dev/lib/python3.7/site-packages/MDAnalysis/coordinates/PDB.py in write(self, obj)
    765         # write_all_timesteps() to dump everything in one go, or do the
    766         # traditional loop over frames
--> 767         self.write_next_timestep(self.ts, multiframe=self._multiframe)
    768         self._write_pdb_bonds()
    769         # END record is written when file is being close()d
~/anaconda3/envs/narupa-dev/lib/python3.7/site-packages/MDAnalysis/coordinates/PDB.py in write_next_timestep(self, ts, **kwargs)
    839                 raise NoDataError("PBDWriter: no coordinate data to write to "
    840                                   "trajectory file")
--> 841         self._check_pdb_coordinates()
    842         self._write_timestep(ts, **kwargs)
    843 
~/anaconda3/envs/narupa-dev/lib/python3.7/site-packages/MDAnalysis/coordinates/PDB.py in _check_pdb_coordinates(self)
    670             self.close()
    671             try:
--> 672                 os.remove(self.filename)
    673             except OSError as err:
    674                 if err.errno == errno.ENOENT:
TypeError: remove: path should be string, bytes or os.PathLike, not _io.StringIO
```

```
import io
import MDAnalysis as mda

u = mda.Universe(PDB)
u.atoms.translate([-9999, -9999, -9999])

with io.StringIO() as str_io, mda.coordinates.PDB.PDBWriter(str_io) as writer:
    #writer.filename = ""  # uncomment to avoid issue
    writer.write(u.atoms)
```

0.20.1
Python 3.7.3
- Which operating system?
Windows 10

RMeli:
Fixes #2512

 - Ignore `TypeError` when `PDBWriter` is trying to remove a `StringIO` object with invalid coordinates, so that the correct exception can be raised instead.

PR Checklist
------------
 - [x] Tests?
 - [x] CHANGELOG updated?


package/MDAnalysis/coordinates/PDB.py
```diff
@@ -678,6 +682,10 @@ def _check_pdb_coordinates(self):
             except OSError as err:
                 if err.errno == errno.ENOENT:
                     pass
+            except TypeError:
+                if isinstance(self.filename, StringIO):
+                    pass
```
jbarnoud: This is equivalent to

```python
except TypeError:
    pass
```

Shouldn't it be the following?

```python
except TypeError:
    if isinstance(self.filename, StringIO):
        pass
    else:
        raise
```

RMeli: Definitely! Thanks for spotting this problem!

RMeli: The same logic should apply to the previous exception as well, which is suffering from the same problem. 

However, this rise a question: should we ignore all the exceptions that can occur while trying `os.remove` so that the correct exception `ValueError` (which is the problem triggering `os.remove`) is thrown instead?

jbarnoud: We can at least wonder what would be the equivalent of an `os.remove` on a `StringIO`. Now, we have an odd discrepancy in behaviour between files and StringIO. Maybe it should be a different issue?

RMeli: I think the equivalent of `os.remove` for `StringIO` is just getting out of scope. However, it looks like that calling `close()` already frees the memory buffer:
> **StringIO.close()**
Free the memory buffer. Attempting to do further operations with a closed StringIO object will raise a ValueError.

jbarnoud: I guess, then, that the handling is appropriate.

PR #2519, opened 2020-02-09
Replace the distance search of finding hydrogens with a dictionary lookup

xiki-tempula:

In the original water bridge analysis implementation, the hydrogen associated with the hydrogen bond donor heavy atom were looked up through a distance search, which is repeated every single frame. This PR attempts to generate the dictionary of hydrogen bond donor heavy atom vs hydrogen pair in the first frame and do a dictionary lookup for the rest of the frames.

This PR should reduce the processing time for all the rest of the frames at the cost of increasing the time for the first frame. This PR should also reduce the computational overhead for turning on the option of `update_selection `.

Here are the benchmark results.

| Frames | new update_selection=False (sec) | old update_selection=False (sec) | new update_selection=True (sec) | old update_selection=True (sec) |
|--------|----------------------------------|----------------------------------|---------------------------------|---------------------------------|
| 1      | 12.0                             | 1.9                              | 10.0                            | 2.5                             |
| 10     | 13.7                             | 7.8                              | 14.0                            | 13.4                            |
| 100    | 37.9                             | 63.3                             | 59.7                            | 119.0                           |
| 1000   | 288.6                            | 649.6                            | 478.7                           | 1137.1                          |

Combined with #2480, the water bridge analysis is now 4X faster than the original implementation and I felt there is probably little room for performance improvement now.

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/analysis/hbonds/wbridge_analysis.py
+++ b/package/MDAnalysis/analysis/hbonds/wbridge_analysis.py
@@ -832,6 +832,17 @@ class WaterBridgeAnalysis(AnalysisBase):
                 if atom.name in self.donors:
                     self._residue_dict[residue.resname][atom.name].update(self._get_bonded_hydrogens(atom).names)
 
+    def _update_donor_h(self, atom_ix, h_donors, donors_h):
+        atom = self.u.atoms[atom_ix]
+        residue = atom.residue
+        hydrogen_names = self._residue_dict[residue.resname][atom.name]
+        if hydrogen_names:
+            hydrogens = residue.atoms.select_atoms('name {0}'.format(
+                ' '.join(self._residue_dict[residue.resname][atom.name]))).ix
+            for atom in hydrogens:
+                h_donors[atom] = atom_ix
+                donors_h[atom_ix].append(atom)
+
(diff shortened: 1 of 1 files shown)


package/MDAnalysis/analysis/hbonds/wbridge_analysis.py
```diff
@@ -820,6 +820,29 @@ def _log_parameters(self):
...
+        # The content is the hydrogen bond donor hydrogen atom names
+        atom_group = self.u.select_atoms(selection)
+        for residue in atom_group.residues:
+            if not residue.resname in self._residue_dict:
+                self._residue_dict[residue.resname] = defaultdict(set)
+            for atom in residue.atoms:
+                if atom.name in self.donors:
+                    self._residue_dict[residue.resname][atom.name].update(self._get_bonded_hydrogens(atom).names)
+
+    def _update_donor_h(self, atom_ix, h_donors, donors_h):
```
richardjgowers: this needs a docstring, it's not clear who/what/where atom_ix is

PR #1814, opened 2018-03-08
ENH: hole trajectory analysis: can use dynamic cpoint

orbeckst:
Changes made in this Pull Request:
 
The hole.HOLEtraj kwarg cpoint can now also a AtomGroup. In this case, the cpoint is recalculated for each frame as the COG of the ag, which helps with proteins that undergo conformational changes.

Basic tests were added.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?
 - [ ] Issue raised/referenced?

PR #2523, opened 2020-02-10
Created hole2 module

### LINKED ISSUE #2522: hole.HOLEtraj makes pore VMD files for every frame
["Producing a colored dot surface using mdanalysis hole over a trajectory"](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/mdnalysis-discussion/0a-IAOOCtQ8/OyZBD4BRAwAJ) seems like a reasonable request

Have hole.HOLEtraj make a vmd / .sph files for each frame of a trajectory.

n/a

**Additional context**
This is not hard, just could be slow. Also writing tests for it doesn't sound fun.

lilyminium:
Fixes #2522 

 - Wrote a new ``hole2`` module

Working example in [this notebook](https://lilyminium.github.io/UserGuide/examples/analysis/polymers_and_membranes/hole2.html#Visualising-the-VMD-surface).

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/analysis/hole.py
+++ b/package/MDAnalysis/analysis/hole.py
@@ -22,7 +22,7 @@
 #
 
 r"""Generation and Analysis of HOLE pore profiles (Deprecated) --- :mod:`MDAnalysis.analysis.hole`
-=================================================================================
+==================================================================================================
 
 :Author: Lukas Stelzl, Oliver Beckstein
 :Year: 2011-2012
--- a/package/MDAnalysis/analysis/hole.py
+++ b/package/MDAnalysis/analysis/hole.py
@@ -21,12 +21,17 @@
 # J. Comput. Chem. 32 (2011), 2319--2327, doi:10.1002/jcc.21787
 #
 
-r"""Generation and Analysis of HOLE pore profiles --- :mod:`MDAnalysis.analysis.hole`
+r"""Generation and Analysis of HOLE pore profiles (Deprecated) --- :mod:`MDAnalysis.analysis.hole`
 =================================================================================
 
 :Author: Lukas Stelzl, Oliver Beckstein
 :Year: 2011-2012
 :Copyright: GNU Public License v2
+
+.. warning:
+
+    This module is deprecated and will be removed in version 2.0.
+    Please use :mod:`MDAnalysis.analysis.hole2` instead.
 
 With the help of this module, the :program:`hole` program from the HOLE_ suite
 of tools [Smart1993]_ [Smart1996]_ can be run on frames in an MD trajectory or
--- a/package/MDAnalysis/analysis/hole.py
+++ b/package/MDAnalysis/analysis/hole.py
@@ -2,7 +2,7 @@
 # vim: tabstop=4 expandtab shiftwidth=4 softtabstop=4
 #
 # MDAnalysis --- https://www.mdanalysis.org
-# Copyright (c) 2006-2017 The MDAnalysis Development Team and contributors
+# Copyright (c) 2006-2020 The MDAnalysis Development Team and contributors
 # (see the file AUTHORS for the full list of names)
 #
 # Released under the GNU Public Licence, v2 or any higher version
@@ -399,7 +399,8 @@ class BaseHOLE(object):
         kw = {}
         frames = kwargs.pop('frames', None)
         if frames is None:
-            frames = np.sort(list(self.profiles.keys())[::kwargs.pop('step', 1)])
+            frames = np.sort(list(self.profiles.keys())
+                             [::kwargs.pop('step', 1)])
         else:
             frames = asiterable(frames)
         kw['frames'] = frames
@@ -807,15 +808,18 @@ class HOLE(BaseHOLE):
         self.dcd = kwargs.pop('dcd', None)
         if self.dcd:
             self.dcd = self.check_and_fix_long_filename(self.dcd)
-        self.dcd_step = kwargs.pop("step", 1) - 1  # HOLE docs description is confusing: step or skip??
+        # HOLE docs description is confusing: step or skip??
+        self.dcd_step = kwargs.pop("step", 1) - 1
         self.dcd_iniskip = 0
         self.cpoint = kwargs.pop("cpoint", None)
         self.cvect = kwargs.pop("cvect", None)
         self.sample = float(kwargs.pop("sample", 0.20))
         self.dotden = int(kwargs.pop("dotden", 15))
         self.endrad = float(kwargs.pop("endrad", 22.))
-        self.shorto = int(kwargs.pop("shorto", 0))  # look at using SHORTO 2 for minimum output
-        self.ignore_residues = kwargs.pop("ignore_residues", self.default_ignore_residues)
+        # look at using SHORTO 2 for minimum output
+        self.shorto = int(kwargs.pop("shorto", 0))
+        self.ignore_residues = kwargs.pop(
+            "ignore_residues", self.default_ignore_residues)
         self.radius = self.check_and_fix_long_filename(
             realpath(kwargs.pop('radius', None) or write_simplerad2()))
         self.raseed = kwargs.pop('raseed', None)
@@ -828,7 +832,8 @@ class HOLE(BaseHOLE):
         hole_exe_name = kwargs.pop('executable', 'hole')
         self.exe['hole'] = which(hole_exe_name)
         if self.exe['hole'] is None:
-            errmsg = "HOLE binary {hole_exe_name!r} not found.".format(**vars())
+            errmsg = "HOLE binary {hole_exe_name!r} not found.".format(
+                **vars())
             logger.fatal(errmsg)
             logger.fatal("%(hole_exe_name)r must be on the PATH or provided as keyword argument 'executable'.",
                          vars())
@@ -931,7 +936,8 @@ class HOLE(BaseHOLE):
         # try a relative path
         newname = os.path.relpath(filename)
         if len(newname) <= self.HOLE_MAX_LENGTH:
-            logger.debug("path check: Using relative path: %r --> %r", filename, newname)
+            logger.debug(
+                "path check: Using relative path: %r --> %r", filename, newname)
             return newname
 
         # shorten path by creating a symlink inside a safe temp dir
@@ -941,8 +947,10 @@ class HOLE(BaseHOLE):
         self.tempfiles.append(newname)
         self.tempdirs.append(dirname)
         if len(newname) > self.HOLE_MAX_LENGTH:
-            logger.fatal("path check: Failed to shorten filename %r --> %r", filename, newname)
-            raise RuntimeError("Failed to shorten filename %r --> %r", filename, newname)
+            logger.fatal(
+                "path check: Failed to shorten filename %r --> %r", filename, newname)
+            raise RuntimeError(
+                "Failed to shorten filename %r --> %r", filename, newname)
         os.symlink(filename, newname)
         logger.debug("path check: Using symlink: %r --> %r", filename, newname)
         return newname
@@ -955,9 +963,12 @@ class HOLE(BaseHOLE):
         # NOTE: If cvect and cpoint had been None in the constructor then they are
         #       ignored here. Arguably a bug... but then again, the keywords for run() are
         #       not even officially documented :-).
-        kwargs.setdefault("cvect_xyz", seq2str(kwargs.pop('cvect', self.cvect)))
-        kwargs.setdefault("cpoint_xyz", seq2str(kwargs.pop('cpoint', self.cpoint)))
-        kwargs.setdefault("ignore", seq2str(kwargs.pop('ignore_residues', self.ignore_residues)))
+        kwargs.setdefault("cvect_xyz", seq2str(
+            kwargs.pop('cvect', self.cvect)))
+        kwargs.setdefault("cpoint_xyz", seq2str(
+            kwargs.pop('cpoint', self.cpoint)))
+        kwargs.setdefault("ignore", seq2str(
+            kwargs.pop('ignore_residues', self.ignore_residues)))
         holeargs = vars(self).copy()
         holeargs.update(kwargs)
 
@@ -967,10 +978,13 @@ class HOLE(BaseHOLE):
                 f.write(inp)
             logger.debug("Wrote HOLE input file %r for inspection", inpname)
 
-        logger.info("Starting HOLE on %(filename)r (trajectory: %(dcd)r)", holeargs)
-        logger.debug("%s <%s >%s", self.exe['hole'], (inpname if inpname else "(input)"), outname)
+        logger.info(
+            "Starting HOLE on %(filename)r (trajectory: %(dcd)r)", holeargs)
+        logger.debug(
+            "%s <%s >%s", self.exe['hole'], (inpname if inpname else "(input)"), outname)
         with open(outname, "w") as output:
-            hole = subprocess.Popen([self.exe['hole']], stdin=subprocess.PIPE, stdout=output)
+            hole = subprocess.Popen(
+                [self.exe['hole']], stdin=subprocess.PIPE, stdout=output)
             stdout, stderr = hole.communicate(inp.encode('utf-8'))
         with open(outname, "r") as output:
             # HOLE is not very good at setting returncodes so check ourselves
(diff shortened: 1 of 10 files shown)


orbeckst:
The original idea had been that `HOLE` is like running hole on a single structure from the command line. But I guess it makes sense to just make it work in such a way that it doesn't matter if you feed a single frame or a trajectory. For backwards compatibility I would just create a new `HoleAnalysis` and deprecate the other two.

(There is also the stalled PR #1814 which I am just mentioning as another HOLE thing...)

orbeckst:
Should do this
> For backwards compatibility I would just create a new HoleAnalysis and deprecate the other two.

and deprecate in 1.0 for removal in 2.0.

package/MDAnalysis/analysis/hole.py
```diff
@@ -1276,29 +1306,17 @@ def _process_orderparameters(self, data):
             raise ValueError(errmsg)
         return q
 
-    def run(self, start=None, stop=None, step=None):
```
orbeckst: This looks as if it reverted an existing PR in which start/stop/step were made compliant.

package/MDAnalysis/analysis/hole2/hole.py
```diff
@@ -0,0 +1,1210 @@
...
+# simulations. In S. Benthall and S. Rostrup editors, Proceedings of the 15th
+# Python in Science Conference, pages 102-109, Austin, TX, 2016. SciPy.
+# doi: 10.25080/majora-629e541a-00e
+#
+# N. Michaud-Agrawal, E. J. Denning, T. B. Woolf, and O. Beckstein.
+# MDAnalysis: A Toolkit for the Analysis of Molecular Dynamics Simulations.
+# J. Comput. Chem. 32 (2011), 2319--2327, doi:10.1002/jcc.21787
+#
+
+"""HOLE Analysis --- :mod:`MDAnalysis.analysis.hole2.hole`
```
orbeckst: Good idea to make it a new module.

package/MDAnalysis/analysis/hole.py
```diff
@@ -399,7 +399,8 @@ def _process_plot_kwargs(self, kwargs):
         kw = {}
         frames = kwargs.pop('frames', None)
         if frames is None:
-            frames = np.sort(list(self.profiles.keys())[::kwargs.pop('step', 1)])
```
orbeckst: Add deprecations to hole.py

package/MDAnalysis/analysis/hole.py
richardjgowers: Probably wants a `git rebase origin/master`.  start/stop/step are now only in `run()` and never `__init__`, so `self.start/step/stop` can never exist

lilyminium: Yes, forgot to rebase.

package/MDAnalysis/analysis/hole2/__init__.py
```diff
@@ -0,0 +1,18 @@
+
```
orbeckst: add boring boilerplate file header (comments/licence...)

package/MDAnalysis/analysis/hole2/helper.py
```diff
@@ -0,0 +1,498 @@
+import logging
```
orbeckst: add boring boilerplate file header (comments/licence...)

package/MDAnalysis/analysis/hole2/templates.py
```diff
@@ -0,0 +1,97 @@
+
```
orbeckst: add boring boilerplate file header (comments/licence...)

package/MDAnalysis/analysis/hole2/templates.py
```diff
@@ -0,0 +1,97 @@
...
+BOND O??? 0.7
+BOND S??? 1.1
+BOND H??? 0.5
+BOND P??? 1.0
+BOND ???? 0.85
+"""
+
+hole_input = """
+! Input file for Oliver Smart's HOLE program
+! written by MDAnalysis.analysis.hole.HOLE
```
orbeckst: change?

package/MDAnalysis/analysis/hole2/templates.py
lilyminium: Updated `# Copyright (c) 2006-2017 The MDAnalysis Development Team and contributors
` to `# Copyright (c) 2006-2020 The MDAnalysis Development Team and contributors
` -- should this be done everywhere?

lilyminium:
``hole2`` (now) has:
 * ``hole``: a function that replaces hole.HOLE to work on PDB files.
    * it deletes HOLE files if `keep_files=False` (default: True) 
 * ``HoleAnalysis``: a class that replaces hole.HOLEtraj
    * It no longer takes order parameters in `__init__`, but returns profiles ordered over given order parameters with `over_order_parameters()`
    * It deletes HOLE files if used as a context manager
    * Unlike any other AnalysisBase subclass, takes more arguments in `run()` than `start, stop, step, verbose`. It takes what I think of as runtime arguments:
        * cpoint: coordinate inside the pore
        * cvect: search direction
        * sample: distance of sample points
        * end_radius: end radius of pore
        * output_level: output level of HOLE
        * random_seed: random seed
        * prefix: prefix for HOLE files
        * write_input_files: whether to write input files
    Given that no other classes do this, it can be moved to `__init__`. Opinions @richardjgowers @orbeckst?

richardjgowers:
disclaimer, I have no idea what HOLE does yet..

@lilyminium I'd move those things to `__init__` if they're parameters of how the analysis is done, whereas start/stop/step are more like sampling variables (so random seed should stay in run).  Maybe philosophically an analysis class with identical `__init__` args is measuring the same thing, the `run` kwargs just define how well it is sampled?

I'm confused about the name, is `hole2.py` using HOLE 2.0, or rather it's our second version of HOLE bindings? (Or Both).  Either way it's a confusing name, I'd rather just it have a different name (eg hbonds vs hbond_analysis) and nag users to use the newer one.

orbeckst:
Hole2 calculates radius of molecular pores in proteins. It is the de facto standard in ion channel structural biology. 

The program is hole2 so it would be appropriate to call the module hole2. 

Note: I’d love to have Python bindings, just for the package name pyhole ;-)

orbeckst:
@lilyminium I would put the parameters into init , following @richardjgowers rationale (and the original implementation).

orbeckst:
raseed should be in run(), though.

lilyminium:
Added new functions: `gather()`, `bin_radii()`, `histogram_radii()`, `plot_mean_profile()`. Are these similar to what you mean, @orbeckst? (See [example usage in this notebook](https://lilyminium.github.io/UserGuide/examples/analysis/polymers_and_membranes/hole2.html#Working-with-the-data))

Tests and documentation coming soon.

orbeckst:
Yes, these functions are very useful and the plot_mean_profile will allow users to quickly have a look at the data. Nice.

richardjgowers:
version 1.0 is 99% complete, is this close enough to completion to delay by a few days? If so I can dive into it and review

lilyminium:
@richardjgowers I was hoping it would be nearly done; the tests aren't failing on my local Python 2.7 or 3.7 environments so not sure what's going on there. When did you want to release?

package/MDAnalysis/analysis/hole2/__init__.py
richardjgowers: Copyright header on new files (gotta get that line count 🙃)

package/MDAnalysis/analysis/hole2/helper.py
```diff
@@ -0,0 +1,556 @@
+# -*- Mode: python; tab-width: 4; indent-tabs-mode:nil; coding:utf-8 -*-
```
richardjgowers: I think util is probably what we’d call this submodule.

package/MDAnalysis/analysis/hole2/helper.py
```diff
@@ -0,0 +1,556 @@
...
+                    records.append((rxncoord, radius, cenlineD))
+                    continue
+                # end of data
+                else:
+                    toggle_read = False
+                    names = ['rxn_coord', 'radius', 'cen_line_D']
+                    recarr = np.rec.fromrecords(records,
+                                                formats='f8, f8, f8',
+                                                names=names)
+                    recarrays[profile] = recarr
```
richardjgowers: Is it worth breaking the loop here? To avoid iterating further lines in the file. Or can there be multiple sections to read?

package/MDAnalysis/analysis/hole2/hole.py
```diff
@@ -0,0 +1,1451 @@
...
+import logging
+import itertools
+import warnings
+
+import numpy as np
+import matplotlib
+import matplotlib.pyplot as plt
+from collections import OrderedDict
+
+from MDAnalysis.exceptions import ApplicationError
```
richardjgowers: Try and not mix relative and absolute imports, I think relative is preferred

orbeckst:
> There’s a lot of bending over backwards to get data in to this program, which obviously isn’t going to be fast (ie writing and reading files is slow). How important is this analysis program? If it’s open source then I can look at building real bindings to it, maybe this is a better long term solution.

HOLE2 is FORTRAN with various limitations on file name lengths etc. Python bindings would be great (it is open source nowadays, link to repo in docs) ([`pyhole`](https://www.merriam-webster.com/dictionary/piehole)!) but it's not worthwhile holding up this PR, which is almost done and is a definitive improvement over what we currently have.

Performance in file I/O is pretty much a non-issue IIRC because HOLE is slow per frame. It would be a prime candidate for PMDA or similar approaches.

package/MDAnalysis/analysis/hole2/helper.py
```diff
@@ -0,0 +1,556 @@
...
+    """
+
+    if not os.path.exists(filename):
+        with open(filename, "w") as rad:
+            rad.write(SIMPLE2_RAD + "\n")
+        logger.debug("Created simple radii file {}".format(filename))
+    return filename
+
+
+def check_and_fix_long_filename(filename, tmpdir=os.path.curdir,
```
orbeckst: This function is not well tested, see coverage. (It wasn't well tested before, either.)

package/MDAnalysis/analysis/hole2/helper.py
```diff
@@ -0,0 +1,556 @@
...
+    if len(filename) <= max_length:
+        return filename
+
+    msg = ('HOLE will not read {} '
+           'because it has more than {} characters.')
+    logger.debug(msg.format(filename, max_length))
+
+    # try a relative path
+    new_name = os.path.relpath(filename)
+    if len(new_name) <= max_length:
```
orbeckst: The if block is not tested.

package/MDAnalysis/analysis/hole2/helper.py
```diff
@@ -0,0 +1,556 @@
...
+        msg = 'Using relative path: {} -> {}'
+        logger.debug(msg.format(filename, new_name))
+        return new_name
+
+    if make_symlink:
+        # shorten path by creating a symlink inside a safe temp dir
+        root, ext = os.path.splitext(filename)
+        dirname = tempfile.mkdtemp(dir=tmpdir)
+        newname = os.path.join(dirname, os.path.basename(filename))
+        if len(newname) > max_length:
```
orbeckst: The if block is not tested.

package/MDAnalysis/analysis/hole2/helper.py
```diff
@@ -0,0 +1,556 @@
...
+            fd, newname = tempfile.mkstemp(suffix='.pdb', dir=dirname)
+            os.close(fd)
+        os.symlink(filename, newname)
+        msg = 'Using symlink: {} -> {}'
+        logger.debug(msg.format(filename, newname))
+        return newname
+
+    msg = 'Failed to shorten filename {}'
```
orbeckst: `RuntimeError` is not tested.

package/MDAnalysis/analysis/hole2/hole.py
```diff
@@ -0,0 +1,1451 @@
...
+                    **kwargs)
+
+        ax.set_xlabel(r"Pore coordinate $\zeta$ ($\AA$)")
+        ax.set_ylabel(ylabel)
+        ax.set_zlabel(r"HOLE radius $R$ ($\AA$)")
+        plt.tight_layout()
+
+        return ax
+
+    def over_order_parameters(self, order_parameters, frames=None):
```
orbeckst: method has patchy test coverage

package/MDAnalysis/analysis/hole2/hole.py
```diff
@@ -0,0 +1,1451 @@
...
+        idx = np.argsort(order_parameters[frames])
+        sorted_frames = frames[idx]
+
+        profiles = OrderedDict()
+        for frame in sorted_frames:
+            profiles[order_parameters[frame]] = self.profiles[frame]
+
+        return profiles
+
+    def plot_order_parameters(self, order_parameters,
```
orbeckst: method needs a test (not tested currently)

package/MDAnalysis/analysis/hole2/hole.py
```diff
@@ -0,0 +1,1451 @@
...
+            Edges of each bin
+        """
+        agg = self.gather(frames=frames, flat=True)
+        coords = agg['rxn_coord']
+
+        if not util.iterable(bins):
+            if range is None:
+                range = (coords.min(), coords.max())
+            xmin, xmax = range
+            if xmin == xmax:
```
orbeckst: if block not tested

package/MDAnalysis/analysis/hole2/hole.py
```diff
@@ -0,0 +1,1451 @@
...
+                xmin -= 0.5
+                xmax += 0.5
+            bins = np.linspace(xmin, xmax, bins+1, endpoint=True)
+        else:
```
orbeckst: else block not tested

package/MDAnalysis/analysis/hole2/hole.py
```diff
@@ -0,0 +1,1451 @@
...
+        -------
+        array of floats
+            histogrammed, aggregate value of radii
+        array of (float, float)
+            Edges of each bin
+        """
+        binned, bins = self.bin_radii(frames=frames, bins=bins, range=range)
+        return np.array(list(map(aggregator, binned))), bins
+
+    def plot_mean_profile(self, bins=100, range=None,
```
orbeckst: method needs test (not tested at the moment)

package/MDAnalysis/analysis/hole2/hole.py
```diff
@@ -0,0 +1,1451 @@
...
+        Returns
+        -------
+        collections.OrderedDict
+            sorted dictionary of {order_parameter:profile}
+
+        """
+        if not self.profiles:
+            raise ValueError('No profiles available. Try calling run()')
+        if isinstance(order_parameters, six.string_types):
+            order_parameters = np.loadtxt(order_parameters)
```
orbeckst: loading from file note tested

package/MDAnalysis/analysis/hole2/helper.py
lilyminium: Changed name to ``utils`` to match ``encore.utils``

package/MDAnalysis/analysis/hole2/helper.py
lilyminium: This is for reading CHARMM DCD files using the hole function, although there is not current test for this that works on Travis (the actual HOLE program breaks on them, despite my local version working)

package/MDAnalysis/analysis/hole2/helper.py
lilyminium: Added a test class.

package/MDAnalysis/analysis/hole2/helper.py
lilyminium: I can't trigger the `RuntimeError` because the temporary file inside a temporary directory is short enough to return before then. In Python 2.7 the default names are 'tmp'+6 random characters -- can't find notes on Python 3 but they look similar. In theory the `RuntimeError` shouldn't happen, I guess.

lilyminium:
Added tests for plots and checking long filenames.

@richardjgowers [hole gets a few questions on the mailing list](https://groups.google.com/forum/#!searchin/mdnalysis-discussion/hole%7Csort:date). I started this PR because combining the pore surfaces for each frame into a single VMD script from the original surface files isn't immediately easy (and had to rewrite the class because otherwise it would keep overwriting intermediate files). So it's worth keeping that part, at least, the pore videos are neat.

orbeckst:
This is a veritable amount of tests. Kudos! (Some are falling for some versions, though, e.g. "AttributeError: 'Line3D' object has no attribute 'get_data_3d'" — for any tests that only fail on 2.7 I'd be ok with a skipif on the test, provided it's only the testing code that fails.)

orbeckst:
@lilyminium do you want to rewrite your history into a smaller number of commits? I don't think that squashing all of them is very clear (there's stuff in old and new hole, tests, etc) but I also don't want to merge 31 commits where some a little corrections.

Awesome work!!

lilyminium:
Thanks @orbeckst , I've squashed the history from 31 commits -> 7.

orbeckst:
Nice work!

Y0geshh:
Hi oliver is this module ready to use now?

lilyminium:
@Y0geshh Yes -- as it is now merged into the ``develop`` version of MDAnalysis, you can just clone the ``develop`` branch and build that.

PR #2505, opened 2020-02-05
Remove start/stop/step from AnalysisBase kwargs

IAlibay:
Addresses #1463

Changes made in this Pull Request:
 - Removed support for `start`, `stop`, and `step` keywords on object creation for :class:`AnalysisBase` and :class:`AnalysisFromFunction` 
 - Where appropriate docstring was changed. If the docstring indicated that `start`, `stop`, and `step` keywords could be passed to `__init__` a `versionchanged` tag was added.

To do:
 - [x] Currently the tests for `AnalysisFromFunction` don't seem to test out the `start`/`stop` arguments very well, it's probably worth adding an extra test there.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?
 - [x] Issue raised/referenced?

PR #2256, opened 2019-04-18
Consistent Autocorrelation and Intermittency - Across Different Analysis

bieniekmateusz:
Fixes #2247 

Changes made in this Pull Request:
 - Provides well defined autocorrelation which can be used for both, SurvivalProbability and HydrogenLifetimes (and more?)
 - Autocorrelation function and intermittency function are now independant with their own test cases
 - Removed the class HydrogenBondLifetimes and moved the autocorrelation functionality into the hbond_analysis class as a .autocorrelation  method. This method should replace also hbond_autocorrel.py file in the future. 
 - New test cases for the method .autocorrelation (HydrogenBondAnalysis) were added 
 - Added a test cases that, with our definition of intermittency, would fail in the case of the hbond_autocorrel.py.

The main issue with HydrogenBondLifetimes (besides the bug) is that it defined its own definition of a hydrogen bond, without allowing the user to change it. It should be clear that the hydrogen bonds first have to be computed by the user, and then the autocorrelation for the bonds can be found. This is achieved with the method .autocorrelation. 

PR Checklist
------------
 - [X] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?
 - [X] Issue raised/referenced?

Work done in duo @p-j-smith

PR #2525, opened 2020-02-10
Waterdynamics.SurvivalProbability start/stop/step/tau_max

### LINKED ISSUE #2524: Align behaviour of waterdynamics.SurvivalProbability with other analysis methods
**Description**

Relates to https://github.com/MDAnalysis/mdanalysis/pull/2505#pullrequestreview-354040434

Currently waterdynamics.SurvivalProbability:
 1. Accepts the deprecated `t0`, `tf`, and `dtmax` keywords in :meth:`SurvivalProbability.__init__`, which override the setting of `start`, `stop`, and `tau_max` in :meth:`SurvivalProbability.run`.
 2. Uses the `stop` (or aforementioned `tf`) keyword to set the final frame inclusively (rather than exclusively as done by most other analysis methods).

 1. Remove the deprecated keywords, making the behaviour similar to :class:`AnalysisBase` methods.
 2. Make `stop` exclusive rather than inclusive.

**Additional context**
This is would complement #2505 and is a goal towards #2443

IAlibay:
Fixes #2524

 - Removes the `t0`, `tf`, and `dtmax` keywords from the :class:`Waterdynamics.SurvivalProbability` constructor.
 - Makes the `stop` keyword for :meth:`WaterDynamics.SurvivalProbability.run` exclusive rather than inclusive.

PR Checklist
------------
 - [x] Tests? (using existing tests, just ammending stop to be 1 higher).
 - [x] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/analysis/waterdynamics.py
```diff
@@ -1316,55 +1305,63 @@ def _correct_intermittency(self, intermittency, selected_ids):
...
-            Note that `step` and `tau_max` work consistently with intermittency.
+            Jump every `step`-th frame. This is compatible but independant of
+            the taus used, and it is good to consider using the  `step` equal
+            to `tau_max` to remove the overlap. Note that `step` and `tau_max`
+            work consistently with intermittency. Default: None
+            (use every frame).
         tau_max : int, optional
-            Survival probability is calculated for the range 1 <= `tau` <= `tau_max`
+            Survival probability is calculated for the range
+            1 <= `tau` <= `tau_max`.
```
richardjgowers: If stop is exclusive does this docstring need updating?

IAlibay: Not sure here. `tau_max` should be independent of `stop`, i.e. previously the case was that `stop` was set in the code to be `stop+1`. The changes done here only change it so that `stop=stop`, `tau` and `tau_max`'s behaviour shouldn't be affected (if that makes sense).

IAlibay:
> Does the exclusive/inclusive change line this class up with the others?

@richardjgowers I think I have pretty much failed to understand what you meant from your review 😅 Any chance you could add some more info?

richardjgowers:
You made stop exclusive (or inclusive) when it was the other way around. Is
it now like other classes?

IAlibay:
> You made stop exclusive (or inclusive) when it was the other way around. Is it now like other classes?

@richardjgowers Yeah, so as per https://github.com/MDAnalysis/mdanalysis/pull/2505#pullrequestreview-354040434 the idea in making it exclusive was to align it with the AnalysisBase's exclusive `stop` behaviour:

https://github.com/MDAnalysis/mdanalysis/blob/977fb79c51b6a2fbdfa1c419a7bad0a97aa8d117/package/MDAnalysis/analysis/base.py#L179-L180

With regards to the other waterdynamics class, unfortunately I'm not sure. I'm having a bit of a hard time working out how `HydrogenBondLifetimes`'s `tf` keyword is working, but it looks like it's exclusive? 

For `WaterOrientationalRelaxation`, this line seems to indicate that `self.tf` (passed to `totalFrames`) is exclusive (similar thing for `MeanSquareDisplacement`):

https://github.com/MDAnalysis/mdanalysis/blob/977fb79c51b6a2fbdfa1c419a7bad0a97aa8d117/package/MDAnalysis/analysis/waterdynamics.py#L784

orbeckst:
@alejob could you please also review this PR on WaterDynamics?

There were a few questions https://github.com/MDAnalysis/mdanalysis/pull/2525#issuecomment-593115108 and maybe you as one of the authors can answer them.

Thanks!

alejob:
Ok, let me check this and I'll answer in a couple of days. I didn't realize all the current changes.

IAlibay:
> Ok, let me check this and I'll answer in a couple of days. I didn't realize all the current changes.

Just re-pinging @alejob as a reminder for review. I do realise things are a little bit (that's probably an understatement) chaotic for everyone, so no worries if you don't have time to review this.

alejob:
Hey there! 
All the module was written with inclusion in mind, regarding the tf parameter.

> With regards to the other waterdynamics classes, unfortunately I'm not sure. I'm having a bit of a hard time working out how `HydrogenBondLifetimes`'s `tf` keyword is working, but it looks like it's exclusive?

It is also inclusive, in the code it is define as
https://github.com/MDAnalysis/mdanalysis/blob/977fb79c51b6a2fbdfa1c419a7bad0a97aa8d117/package/MDAnalysis/analysis/waterdynamics.py#L515

but later I use this in this way
https://github.com/MDAnalysis/mdanalysis/blob/977fb79c51b6a2fbdfa1c419a7bad0a97aa8d117/package/MDAnalysis/analysis/waterdynamics.py#L573

tf is a value fill by self.tf.

> 
> For `WaterOrientationalRelaxation`, this line seems to indicate that `self.tf` (passed to `totalFrames`) is exclusive (similar thing for `MeanSquareDisplacement`):
> 
The -1 is because the list start in zero, so if there is not -1, there will be a "list index out of bound".
> https://github.com/MDAnalysis/mdanalysis/blob/977fb79c51b6a2fbdfa1c419a7bad0a97aa8d117/package/MDAnalysis/analysis/waterdynamics.py#L784

I hava a question about this PR, this is only for make it consistent with the AnalysisBase? I'm pretty sure I lost a discussion about this, but does this PR improve something in the usability of this class? and also, If I want to obtain the same results that this module delivers, with the new implementation I should add an unit to the tf value? ex: If before this PR I wanted to calculate something until the frame 1000, now should I write 1001?

IAlibay:
Thanks for having a look through @alejob 

It seems like I've completely failed to understand the behaviour of `tf` for most of these classes, I'll have a longer look at this.

> I hava a question about this PR, this is only for make it consistent with the AnalysisBase?

So there are two aims to this PR.
  1. Probably the most important thing re: v1.0 here is to remove the deprecated `t0`, `tf`, and `dtmax` keywords from the :`Waterdynamics.SurvivalProbability`.
  2. As pointed out by @orbeckst  (https://github.com/MDAnalysis/mdanalysis/pull/2505#pullrequestreview-354040434) there is also a descrepancy between how `Waterdynamics` treats `stop` relative to how `AnalysisBase` does. So this PR aims to at least fix this for `SurvivalProbability` (although the question was raised as to whether or not this should be done for the whole of `WaterDynamics`).

> I'm pretty sure I lost a discussion about this, but does this PR improve something in the usability of this class?

Whilst the current behaviour is documented in `SurvivalProbability`, one could see, especially considering my above failure to understand `tf` across most of the classes, how a user error could come up here. So the view here is that making things similar to `AnalysisBase` would avoid these kinds of errors / allow for a more consistent API.

> and also, If I want to obtain the same results that this module delivers, with the new implementation I should add an unit to the tf value? ex: If before this PR I wanted to calculate something until the frame 1000, now should I write 1001?

Indeed, assuming one had frames ranging from 0 to 1000, then if passed explicitly `stop` should be equal to 1001 (or higher) to cover the full range of frames. As far as I am aware, this is the current way in which `AnalysisBase` works.

orbeckst:
@IAlibay, I just merged PR #2256 which changed waterdynamics. Can you do resolve the conflicts and have a look again?

orbeckst:
This looks mostly done (bringing WaterAnalysis in-line with the other analysis classes) although the following seems needed:
- resolving conflicts (might not be trivial so @IAlibay should do it)
- @richardjgowers 's blocking review – I think @IAlibay answered his question (now all exclusive, in line with AnalysisBase use) but please check that you're satisfied
- @alejob can you please do a formal review of the changes with an "approve" or "changes requested"?
- I'll be happy to do a review after @alejob – but I don't know the code as well as the original author so I'll mostly follow  @alejob 's lead.

I added it to the 1.0 milestone so let's get this done.

IAlibay:
Sorry about the delay here @orbeckst, I've got a slot in my calendar specifically to look at this tomorrow :) 

A thought, do we want to extend this beyond survival probability to make stop exclusive across all of waterdynamics?

orbeckst:
> A thought, do we want to extend this beyond survival probability to make stop exclusive across all of waterdynamics?

That would be great – even if we don't have everything AnalysisBase-based yet, having a consistent interface is necessary. And with semantic versioning, backwards-incompatible changes in the API are very annoying because you need to bump the major version. Changing the meaning of the `stop` argument is such a major API change because it will lead to _wrong results_ between different versions. I don't think anyone wants to go to MDA 2.0 just in order to update the arguments in waterdynamics...

orbeckst:
> Sorry about the delay here

Oh, forgot: please do not apologize!!!!! You're doing an awful lot of work here, thank you, thank you! 

I see it as my job to occasionally check on PRs and help shepherd them to completion so that all the work that has already gone into them also counts. Sometimes one just has to clarify what needs doing.

orbeckst:
> A thought, do we want to extend this beyond survival probability to make stop exclusive across all of waterdynamics?

I would open a new PR, though, just to get this one done. It can then serve as a blueprint for the rest.

As we're preaching to our GSoC students: many small PRs are better than one big one ;-).

IAlibay:
The conflicts _should_ now be solved. Pinging @richardjgowers and @alejob for review.

orbeckst:
@alejob your review is important here. Do you have an idea when you might be able to do it? Thanks!

alejob:
Sorry for the late! Today at night I'll Look this.

package/MDAnalysis/analysis/waterdynamics.py
```diff
@@ -1172,84 +1172,80 @@ class SurvivalProbability(object):
...
-            self.stop = tf
-            warnings.warn("tf is deprecated, use run(stop=tf) instead", category=DeprecationWarning)
-
-        if dtmax is not None:
-            self.tau_max = dtmax
-            warnings.warn("dtmax is deprecated, use run(tau_max=dtmax) instead", category=DeprecationWarning)
-
-
-    def run(self, tau_max=20, start=0, stop=None, step=1, residues=False, intermittency=0, verbose=False):
+    def run(self, tau_max=20, start=None, stop=None, step=None, residues=False,
```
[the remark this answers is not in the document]
IAlibay: @alejob by this do you mean the removal of `t0`, `tf`, and `dtmax`? These have been slated for deprecation for a long time. As part of the upcoming 1.0.0 release, we are aiming to remove all historical deprecated code.

IAlibay:
> In the other side I don't see changes on the examples of documentation (or a comment), please change that, because it will be the only updated class and that could be confusing.

Thanks for the review @alejob. I have included documentation (beyond what was already in :meth:`SurvivalProbability.run`. Let me know if this is what you had in mind & if you think it needs changes.

orbeckst:
Ping @richardjgowers for final approval?

PR #2528, opened 2020-02-13
Standardise select keyword

### LINKED ISSUE #2461: select vs selection
**Expected behavior**

MDAnalysis analysis functions and classes have a standard keyword for selecting atoms from a Universe.

**Actual behavior**

It varies largely between 'select' and 'selection', meaning that I have to look up each function when I want to use it. It's usually a keyword argument (everything below has it as such), so I do have to give the right keyword.

**select**
- rms.RMSD, rms.RMSF
- align.AlignTraj, align.AverageStructure
- diffusionmap.DistanceMatrix
- pca.PCA

**selection**
- contacts.q1q2, contacts.Contacts
- gnm.GNMAnalysis, gnm.closeContactGNMAnalysis
- helanal.helanal_trajectory, helanal.helanal_main
- hole.HOLEtraj
- all of encore
- hydrogenbonds.HydrogenBondAnalysis

**atomselection**
- density.density_from_Universe, density.density_from_PDB, density.BfactorDensityCreator

**ref_select**
- psa.PSAnalysis (also has `path_select`)

- Which operating system?

RMeli:
Fixes #2461 

- [x] Should all the selections in `encore` change as well?
- [x] In `psa.PSAnalysis` there is `ref_select` and `path_select`. How this case should be standardized?

 - Change `selection` to `select`
 - Change `atomselection` to `select`

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


lilyminium:
Thanks for doing this <3

> In psa.PSAnalysis there is ref_select and path_select. How this case should be standardized?

If changed at all, I would change `ref_select` to `select` and keep `path_select` as is. `ref_select` is for trajectory alignment, while `path_select` is for calculating paths. If `path_select=None` then it uses `ref_select`, so the latter is the "main" selection. `ref_select` also behaves exactly as `select` does in rms.RMSD.

RMeli:
Thanks for the insights @lilyminium. I changed `ref_select` to `select` as you suggested.

PR #2533, opened 2020-02-15
added READER_HINTS

richardjgowers:
I think this is maybe a more elegant way to deal with optional dependencies.

Readers can define a `format_test` staticmethod, which when provided a "filename" (which is the thing provided to `Universe`, so can be eg numpy array) returns a bool of if they reckon they can read it.  `MemoryReader` is an example of a Reader that would implement this.  The `ParmedReader` is another example, which then means that only this Reader module needs to import Parmed, the rest of MDA is oblivious.

Then things like `get_reader_for` are simplified, as they just rely on information provided by Readers during the metaclass magic that happens. (ie `_READERS` and the new `_READER_HINTS`)

This would also be how the Chemfiles PR (and other converter type readers) would be resolved, you would load the system with chemfiles, then pass this to Universe.

TODO: identical treatment for parsers

Possible problem: race conditions in `format_test` (though is also currently a problem)

Thoughts? @MDAnalysis/coredevs

### DIFF
--- a/package/MDAnalysis/coordinates/MMTF.py
+++ b/package/MDAnalysis/coordinates/MMTF.py
@@ -62,7 +62,7 @@ class MMTFReader(base.SingleFrameReaderBase):
     format = 'MMTF'
 
     @staticmethod
-    def format_test(thing):
+    def _format_hint(thing):
         return isinstance(thing, mmtf.MMTFDecoder)
 
     @due.dcite(
--- a/package/MDAnalysis/coordinates/MMTF.py
+++ b/package/MDAnalysis/coordinates/MMTF.py
@@ -61,6 +61,10 @@ class MMTFReader(base.SingleFrameReaderBase):
     """Coordinate reader for the Macromolecular Transmission Format format (MMTF_)."""
     format = 'MMTF'
 
+    @staticmethod
+    def format_test(thing):
+        return isinstance(thing, mmtf.MMTFDecoder)
+
     @due.dcite(
         Doi('10.1371/journal.pcbi.1005575'),
         description="MMTF Reader",
--- a/package/MDAnalysis/coordinates/MMTF.py
+++ b/package/MDAnalysis/coordinates/MMTF.py
@@ -63,6 +63,10 @@ class MMTFReader(base.SingleFrameReaderBase):
 
     @staticmethod
     def _format_hint(thing):
+        """Can this Reader read *thing*?
+
+        .. versionadded:: 1.0.0
+        """
         return isinstance(thing, mmtf.MMTFDecoder)
 
     @due.dcite(
--- a/package/MDAnalysis/coordinates/ParmEd.py
+++ b/package/MDAnalysis/coordinates/ParmEd.py
@@ -87,6 +87,15 @@ class ParmEdReader(base.SingleFrameReaderBase):
     # Structure.coordinates always in Angstrom
     units = {'time': None, 'length': 'Angstrom'}
 
+    @staticmethod
+    def format_test(thing):
+        try:
+            import parmed as pmd
+        except ImportError:
+            return False
+        else:
+            return isinstance(thing, pmd.Structure)
+
     def _read_first_frame(self):
         self.n_atoms = len(self.filename.atoms)
 
--- a/package/MDAnalysis/coordinates/ParmEd.py
+++ b/package/MDAnalysis/coordinates/ParmEd.py
@@ -88,7 +88,7 @@ class ParmEdReader(base.SingleFrameReaderBase):
     units = {'time': None, 'length': 'Angstrom'}
 
     @staticmethod
-    def format_test(thing):
+    def _format_hint(thing):
         try:
             import parmed as pmd
         except ImportError:
--- a/package/MDAnalysis/coordinates/ParmEd.py
+++ b/package/MDAnalysis/coordinates/ParmEd.py
@@ -89,9 +89,14 @@ class ParmEdReader(base.SingleFrameReaderBase):
 
     @staticmethod
     def _format_hint(thing):
+        """Can this reader read *thing*?
+
+        .. versionadded:: 1.0.0
+        """
         try:
             import parmed as pmd
         except ImportError:
+            # if we can't import parmed, it's probably not parmed
             return False
         else:
             return isinstance(thing, pmd.Structure)
--- a/package/MDAnalysis/core/_get_readers.py
+++ b/package/MDAnalysis/core/_get_readers.py
@@ -24,9 +24,6 @@ from six import raise_from
 
 import copy
 import inspect
-import mmtf
-import parmed as pmd
-import numpy as np
 
 from .. import (_READERS, _READER_HINTS,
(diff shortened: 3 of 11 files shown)


package/MDAnalysis/core/_get_readers.py
```diff
@@ -83,20 +84,15 @@ def get_reader_for(filename, format=None):
...
-            # memoryreader slurps numpy arrays
-            format = 'MEMORY'
-        elif isinstance(filename, mmtf.MMTFDecoder):
-            # mmtf slurps mmtf object
-            format = 'MMTF'
-        elif isinstance(filename, pmd.Structure):
-            format = 'PARMED'
-        else:
+    if format is None:
+        for f, test in _READER_HINTS.items():
```
jbarnoud: `format_name` instead of `f`?

richardjgowers:
Re errors: the idea here is that users input their eg Parmed object, so
there’s no dependency within MDA (except the typecheck).

In the chemfiles case, rather than providing a file and asking MDA to load
with chemfiles, instead the order is load with chemfiles then ask MDA to
load from that object. So in a way the responsibility for the import has
been pushed to the user.

kain88-de:
Can you give some code examples for a user. It is unclear to me when the format_test functions are called.

package/MDAnalysis/core/_get_readers.py
```diff
@@ -27,9 +27,10 @@
 import mmtf
 import parmed as pmd
 import numpy as np
```
lilyminium: Don't need this import anymore

package/MDAnalysis/core/_get_readers.py
```diff
@@ -83,20 +84,15 @@ def get_reader_for(filename, format=None):
...
-            # mmtf slurps mmtf object
-            format = 'MMTF'
-        elif isinstance(filename, pmd.Structure):
-            format = 'PARMED'
-        else:
+    if format is None:
+        for f, test in _READER_HINTS.items():
+            if test(filename):
+                format = f
+                break;
```
lilyminium: Don't need the semicolon

lilyminium:
>  where does the nice error messages about missing dependencies fit in

Each class will still need its own error messages about dependencies, for the cases where:
* the user passes ``format='PARMED'`` and does not actually have parmed installed -- better to have some clear error message than to fail somewhere along the path of trying to convert the mystery object
* they want to write out to a particular library's format and do not have the required package/s

package/MDAnalysis/core/_get_readers.py
richardjgowers: haha I've been doing too little Python recently :D

richardjgowers:
WRT user code:
```python
import parmed as pmd

s = pmd.load_file('something.pdb')

u = mda.Universe(s)
```

Or more generically, this is now possible.  (Previously this would have required passing the `XReader` class as a `format=` keyword argument.)

```python
# XReader.py
class XReader(ReaderBase):
    @staticmethod
    def format_test(thing):
        try:
            import X
        except ImportError:
            return False
        return isinstance(thing, X.XClass)

# User script
import MDAnalysis as mda
import XReader  # monkey patches _READERS and _READER_HINTS
import X

x_object = X.load(something)

u = mda.Universe(x_object)
```

@kain88-de the `format_test` functions are really useful for a user, they're just used in `get_reader_for` for format detection, maybe we can hide them behind an underscore

@lilyminium the ParmedReader only reads a parmed object (ie it won't call `pmd.load_file`), so if the user passes `format="PARMED"` they must have the package to have the object to supply anyway?  With converting, you've already put the import inside the method and got a check around the import with a nice error message.

lilyminium:
Yep, I was thinking more along the lines of catching cases like these:

```python
>>> import numpy as np
>>> import MDAnalysis as mda
>>> arr = np.loadtxt('parmed_structure.dat')
>>> u = mda.Universe(arr, format='PARMED')
  File ".../MDAnalysis/core/universe.py", line 293, in __init__
    self._topology = p.parse(**kwargs)
  File ".../MDAnalysis/topology/ParmEdParser.py", line 178, in parse
    for atom in structure.atoms:
AttributeError: 'numpy.ndarray' object has no attribute 'atoms'
```

> With converting, you've already put the import inside the method and got a check around the import with a nice error message.

Yes, I guess future converters will have to keep that in mind too.

RMeli:
I like this idea @richardjgowers , but looking at the code snippet it seems to me that the user has to manually import `import XReader` in order to monkey patch `_READERS` and `_READER_HINTS`, even though `XReader` is  never explicitly used in the user script (linters will likely complain). What happens if the user forgets the `import XReader`?

richardjgowers:
@RMeli the example is for if `XReader.py` is a file sitting alongside some analysis script.  If XReader were in the MDAnalysis package it would get imported with everything else and work as normal

richardjgowers:
@MDAnalysis/coredevs ok this seems to simplify format guessing and makes adding new readers with new dependencies easier (ie chemfiles & cclib & rdkit) so can this get a review?

package/MDAnalysis/coordinates/MMTF.py
```diff
@@ -61,6 +61,10 @@ class MMTFReader(base.SingleFrameReaderBase):
     """Coordinate reader for the Macromolecular Transmission Format format (MMTF_)."""
     format = 'MMTF'
 
+    @staticmethod
+    def _format_hint(thing):
```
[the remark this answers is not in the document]
orbeckst: Also: versionadded: 1.0.0.

package/MDAnalysis/coordinates/ParmEd.py
```diff
@@ -81,14 +81,22 @@
...
     """Coordinate reader for ParmEd."""
     format = 'PARMED'
 
     # Structure.coordinates always in Angstrom
     units = {'time': None, 'length': 'Angstrom'}
 
+    @staticmethod
+    def _format_hint(thing):
+        try:
+            import parmed as pmd
```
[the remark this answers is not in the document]
richardjgowers: yeah, if you've gone down the `ImportError` branch then there's no chance of thing being a pmd structure

package/MDAnalysis/topology/MMTFParser.py
```diff
@@ -152,6 +152,10 @@ def apply(self, group):
 class MMTFParser(base.TopologyReaderBase):
     format = 'MMTF'
 
+    @staticmethod
+    def _format_hint(thing):
+        return isinstance(thing, mmtf.MMTFDecoder)
```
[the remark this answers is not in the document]
richardjgowers: Yes MMTF could always do both, the filename option is just syntactic sugar. File extension reading is unaffected by this, this is just a different formalism for how we deal with when objects are passed in to Universe

richardjgowers:
@orbeckst thanks for review, your thinking out loud is handy for knowing where to put comments :)

richardjgowers:
ok more docs, hopefully good to go

orbeckst:
> your thinking out loud is handy for knowing where to put comments

I think I found my role in the project: my job is to "simulate" and idiot reading the code!

PR #2534, opened 2020-02-15
Use tempdir when testing align.AlignTraj

### LINKED ISSUE #2532: rmsfit_align_test.dcd writtein outside of temporary directory
**Expected behavior**

When running the `analysis/test_align.py` I expect no file to be created outside of temporary directories.

**Actual behavior**

The file `rmsfit_align_test.dcd` is created in the `testsuite` directory (or any other directory where `pytest` is run).

``` bash
pytest analysis/test_align.py
```

- Which version are you using? 0.20.2-dev0
- Which version of Python (`python -V`)? 3.7.6
- Which operating system? macOS 10.15.3

RMeli:
Fixes #2532

 - Added `with tempdir.in_tempdir()` context manager when testing `align.AlignTraj`

PR Checklist
------------
 - [x] Tests?


testsuite/MDAnalysisTests/analysis/test_align.py
```diff
@@ -226,14 +223,16 @@ def test_AlignTraj_outfile_default_exists(self, universe, reference, tmpdir):
         fitted = mda.Universe(PSF, outfile)
 
         # ensure default file exists
-        with mda.Writer(str(tmpdir.join('rmsfit_align_test.dcd')),
```
richardjgowers: not calling `str()` is messing up the py2.7 tests

RMeli: Wooops, that was a leftover for the first attempt. I reverted it to the original syntax.

PR #2537, opened 2020-02-18
new DensityAnalysis

### LINKED ISSUE #2502: upgrade DensityAnalysis to use AnalysisBase
**Expected behavior**

The density analysis in `analysis.density` should implement the common analysis API, namely be based on `analysis.base.AnalysisBase`. 

**Actual behavior**

The density generation code in `analysis.density` consist of a rag-tag of idiosyncratic functions with no common interface.

current develop (> 0.20)

orbeckst:
Fixes #2502 

 - new DensityAnalysis (based on @nawtrey 's pmda.density.DensityAnalysis)
 - added tests (same as for the old density_from_universe() function) so old and new give the same answers
 - deprecated density_from_universe(), density_from_PDB(), Bfactor2RMSF(), BfactorDensityCreator (remove in 2.0)
- added superficial tests for density_from_PDB() and Bfactor2RMSF() (and fixed a bug...)
 - removed superfluous tests for the case when gridDataFormats package cannot be found: gridDataFormats is a dependency so we can always assume it is present

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


IAlibay:
So not directly related to issue #2502, but whilst things are being deprecated. As far as I am aware, `density_from_PDB` had historically been untested, would it be worth maybe considering removing it unless someone is willing to write tests for it?

package/MDAnalysis/analysis/density.py
```diff
@@ -180,11 +177,259 @@
...
+        self._bins = bins
+        #: density as a :class:`Density`
+        self.density = None
+
+    def _single_frame(self):
+        h, _ = np.histogramdd(self._atomgroup.positions,
+                              bins=self._bins, range=self._arange,
+                              normed=False)
+        # manually reduce (not yet part of AnalysisBase but exists in pmda.parallel.AnalysisBase)
+        self._reduce(self._grid, h)
```
IAlibay: Probably being silly here so please ignore if that's the case. Is there a plan to introduce _reduce to `AnalysisBase` in the future? Otherwise, it's not clear to me how the extra `if/else` isn't introducing an unnecessary couple of extra CPU cycles per iteration.

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -154,20 +151,126 @@ class Test_density_from_Universe(object):
...
+                "you may want to define a larger box")
+        with pytest.warns(UserWarning, match=regex):
+            D = density.density_from_Universe(
+                universe, select=self.selections['static'],
+                delta=self.delta, xdim=1.0, ydim=2.0, zdim=2.0, padding=0.0,
+                gridcenter=self.gridcenters['static_defined'])
+
+    def test_ValueError_userdefn_gridcenter_shape(self, universe):
+        # Test len(gridcenter) != 3
+        with pytest.raises(ValueError):
```
IAlibay: Not that we were doing it before, but is it worth adding a match argument here to check that we are picking up the right error?

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -154,20 +151,126 @@ class Test_density_from_Universe(object):
...
+            D = density.DensityAnalysis(
+                universe.select_atoms(self.selections['static']),
+                delta=self.delta, xdim=100.0, ydim=100.0, zdim=100.0, padding=1.0,
+                gridcenter=self.gridcenters['static_defined']).run(step=5)
+
+    def test_warn_userdefn_smallgrid(self, universe):
+        regex = ("Atom selection does not fit grid --- "
+                "you may want to define a larger box")
+        with pytest.warns(UserWarning, match=regex):
+            D = density.density_from_Universe(
```
IAlibay: Shouldn't this be a `DensityAnalysis` call?

package/MDAnalysis/analysis/density.py
```diff
@@ -180,11 +177,259 @@
...
+    artifacts and is suitably superimposed to provide a fixed reference frame)
+    [#testraj]_, first  create the :class:`DensityAnalysis` object by
+    supplying an AtomGroup, then use  the :meth:`run` method::
+
+        from MDAnalysis.analysis import density
+        u = Universe(TPR, XTC)
+        ow = u.select_atoms("name OW")
+        D = density.DensityAnalysis(ow, delta=1.0)
+        D.run()
+        D.convert_density('TIP4P')
```
lilyminium: This should be `D.density.convert_density('TIP4P')`.

package/MDAnalysis/analysis/density.py
```diff
@@ -180,11 +177,259 @@
...
+    *delta* = 1 Å. Initially the density is measured in inverse cubic
+    ångström. With the :meth:`Density.convert_density` method,
+    the units of measurement are changed. In the example we are now measuring
+    the density relative to the literature value of the TIP4P water model at
+    ambient conditions (see the values in :data:`MDAnalysis.units.water` for
+    details). In particular, the density is stored as a NumPy array in
+    :attr:`grid`, which can be processed in any manner. Results are available
+    through the :attr:`density` attribute, which has the :attr:`grid`
+    attribute that contains the histogrammed density data.
+    :attr:`DensityAnalysis.density` is a :class:`gridData.core.Grid` object.
```
lilyminium: Here (or earlier, since you `convert_density` is a Density function) could you explicitly say that it's a Density object, which is a subclass of Grid? Density has some extra methods defined and you can point users to those docs to work out what to do with their results.

package/MDAnalysis/analysis/density.py
```diff
@@ -180,11 +177,259 @@
...
+    Parameters
+    ----------
+    atomgroup : AtomGroup or UpdatingAtomGroup
+            Group of atoms (such as all the water oxygen atoms) being analyzed.
+            This can be an :class:`~MDAnalysis.core.groups.UpdatingAtomGroup` for
+            selections that change every time step.
+    delta : float (optional)
+            Bin size for the density grid in ångström (same in x,y,z).
+            Slice the trajectory as "trajectory[start:stop:step]"; default
+            is to read the whole trajectory.
```
lilyminium: These lines don't belong.

package/MDAnalysis/analysis/density.py
```diff
@@ -180,11 +177,259 @@
...
+    metadata : dict (optional)
+            `dict` of additional data to be saved with the object; the meta
+            data are passed through as they are.
+    padding : float (optional)
```
lilyminium: Should the default arguments of kwargs be noted in the docs?

package/MDAnalysis/analysis/density.py
lilyminium: Also, an `Attributes` section in the docs (even for the one thing) is more immediately discoverable.

orbeckst:
@IAlibay and @lilyminium many thanks for the excellent review, I'll work on it. 

I'll throw out `metadata` and some of the other historic crud – this is the best opportunity to clean up a bit.

package/MDAnalysis/analysis/density.py
orbeckst: If they are explicit in the call signature (i.e., we did not use `**kwargs`) then they are visible in the generated docs and with `help(...)`. I don't think we have an explicit policy but there's something to be said for being DRY.

Do you prefer them duplicated in the docs?

package/MDAnalysis/analysis/density.py
```diff
@@ -180,11 +177,259 @@
...
+
+        metadata = self._metadata if self._metadata is not None else {}
+        metadata['topology'] = metadata['psf'] = self._atomgroup.universe.filename
+        metadata['trajectory'] = metadata['dcd'] = self._trajectory.filename
+        metadata['n_frames'] = self.n_frames
+        metadata['totaltime'] = self._trajectory.totaltime
+        metadata['dt'] = self._trajectory.dt
+        metadata['time_unit'] = "ps"
+        parameters = self._parameters if self._parameters is not None else {}
+        parameters['isDensity'] = False  # must override
```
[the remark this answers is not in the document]
orbeckst: `isDensity` tells `Density` if it is a density (something per volume) or just counts. IIRC the idea had been that `parameters` could hold more things but obviously that never happened so it would be better to just have a `isDensity` kwarg.

`isDensity` is necessary so that one can initialize `Density` from either a histogram or a real density. In this case we know that we have a histogram so no matter what a user might have set in `parameters`, we must override it ... nevermind that there's only one thing.... so that's why I am removing it :-).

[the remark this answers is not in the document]
orbeckst: Changing `parameters` in `Density` is a good idea... 2.0

package/MDAnalysis/analysis/density.py
```diff
@@ -180,11 +177,259 @@
...
+        # average:
+        self._grid /= float(self.n_frames)
+
+        metadata = self._metadata if self._metadata is not None else {}
+        metadata['topology'] = metadata['psf'] = self._atomgroup.universe.filename
+        metadata['trajectory'] = metadata['dcd'] = self._trajectory.filename
+        metadata['n_frames'] = self.n_frames
+        metadata['totaltime'] = self._trajectory.totaltime
+        metadata['dt'] = self._trajectory.dt
+        metadata['time_unit'] = "ps"
```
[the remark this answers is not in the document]
orbeckst: removed the whole thing from `DensityAnalysis`.

package/MDAnalysis/analysis/density.py
```diff
@@ -180,11 +177,259 @@
...
+        # manually reduce (not yet part of AnalysisBase but exists in pmda.parallel.AnalysisBase)
+        self._reduce(self._grid, h)
+
+    def _conclude(self):
+        # average:
+        self._grid /= float(self.n_frames)
+
+        metadata = self._metadata if self._metadata is not None else {}
+        metadata['topology'] = metadata['psf'] = self._atomgroup.universe.filename
+        metadata['trajectory'] = metadata['dcd'] = self._trajectory.filename
```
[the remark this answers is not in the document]
orbeckst: removed (don't have to be there)

package/MDAnalysis/analysis/density.py
```diff
@@ -180,11 +177,259 @@
...
+
+    1. have been centered on the protein of interest;
+    2. have all molecules made whole that have been broken across periodic
+       boundaries [#pbc]_;
+    3. have the solvent molecules remapped so that they are closest to the
+       solute (this is important when using triclinic unit cells such as
+       a dodecahedron or a truncated octahedron) [#pbc]_;
+    4. have a fixed frame of reference; for instance, by superimposing a
+       protein on a reference structure so that one can study the solvent
+       density around it [#fit]_.
```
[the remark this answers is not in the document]
orbeckst: Should be done with on-the-fly transformations and not patched into every single analysis class... I think.

testsuite/MDAnalysisTests/analysis/test_density.py
orbeckst: yes, thanks

package/MDAnalysis/analysis/density.py
```diff
@@ -180,11 +177,259 @@
...
+    ----------
+    atomgroup : AtomGroup or UpdatingAtomGroup
+            Group of atoms (such as all the water oxygen atoms) being analyzed.
+            This can be an :class:`~MDAnalysis.core.groups.UpdatingAtomGroup` for
+            selections that change every time step.
+    delta : float (optional)
+            Bin size for the density grid in ångström (same in x,y,z).
+            Slice the trajectory as "trajectory[start:stop:step]"; default
+            is to read the whole trajectory.
+    metadata : dict (optional)
```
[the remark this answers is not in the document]
orbeckst: removed from DensityAnalysis, I leave it in Density for 1.0, can be rewritten for 2.0

testsuite/MDAnalysisTests/analysis/test_density.py
orbeckst: added

package/MDAnalysis/analysis/density.py
orbeckst: I think for 2.0 I'd like to rewrite AnalysisBase to match pmda so that we don't maintain two subtly different implementations. When we started with the serial version, we didn't really understand what was necessary to make the parallel one work. I opened #2542 .

orbeckst: I commented out the _reduce code and replaced it with the simple serial version.

package/MDAnalysis/analysis/density.py
```diff
@@ -180,11 +177,259 @@
...
+      # In this case, we update the atom selection as shown above.
+      D_water = DensityAnalysis(universe.select_atoms('name OW and around 5 resname LIG',
+                                                      updating=True),
+                                delta=1.0,
+                                gridcenter=ligand_COM,
+                                xdim=20, ydim=20, zdim=20)
+
+      (It should be noted that the `padding` keyword is not used when a user
+      defined grid is assigned).
+
```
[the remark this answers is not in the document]
orbeckst: added it back, thanks for the reminder

package/MDAnalysis/analysis/density.py
```diff
@@ -180,11 +177,259 @@
...
+    histogrammed on a 3D grid with spacing `delta`.
+
+    Parameters
+    ----------
+    atomgroup : AtomGroup or UpdatingAtomGroup
+            Group of atoms (such as all the water oxygen atoms) being analyzed.
+            This can be an :class:`~MDAnalysis.core.groups.UpdatingAtomGroup` for
+            selections that change every time step.
+    delta : float (optional)
+            Bin size for the density grid in ångström (same in x,y,z).
```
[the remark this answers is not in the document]
orbeckst: Seems consistently spelled as ångström.

orbeckst:
Cleaned up history. Please review. Thanks.

package/MDAnalysis/analysis/density.py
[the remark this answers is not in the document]
IAlibay: So my comment was relative to the other classes in `density.py`, e.g. :class:`Density` uses a mix of ångström and Angstrom. It doesn't really matter though.

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -323,17 +396,44 @@ def u():
...
+                         [(None, (21, 26, 29), 108.710149),
+                          (1.5,  (21, 26, 29), 160.050167)])
+def test_density_from_PDB(sigma, ref_shape, ref_gridsum):
+    # simple test that the code runs: use B-factors from PDB or fixed sigma
+    D = density.density_from_PDB(PDB_full, delta=2.0, sigma=sigma)
+
+    assert isinstance(D, density.Density)
+    assert_equal(D.grid.shape, ref_shape)
+    assert_almost_equal(D.grid.sum(), ref_gridsum, decimal=6)
+
```
IAlibay: Would it be possible to add deprecation warning tests for density_from_PDB and associated classes?

package/MDAnalysis/analysis/density.py
```diff
@@ -180,11 +192,287 @@
...
+    def _conclude(self):
+        # average:
+        self._grid /= float(self.n_frames)
+        density = Density(grid=self._grid, edges=self._edges,
+                          units={'length': "Angstrom"},
+                          parameters={'isDensity': False})
+        density.make_density()
+        self.density = density
+
+    # _reduce is not strictly necessary for the serial version but is necessary for
```
IAlibay: So this is more to try to avoid ending up with commented out code that is forgotten about, assuming that #2542 goes ahead, my understanding is that the `_reduce` method code will be added to :class:AnalysisBase. Is keeping this in the meantime necessary or could the above comment in `_single_frame` be sufficient?

package/MDAnalysis/analysis/density.py
lilyminium: I think so long as there is an explicit signature, they don't need to be there (and protects against changing the defaults in future versions) -- that being said I have been putting default values at the end of the explanatory part, [following examples in PEP 257](https://www.python.org/dev/peps/pep-0257/). I was just wondering if you had a specific policy on it.

orbeckst: We don't have an explicit policy in https://github.com/MDAnalysis/mdanalysis/wiki/WritingDocumentation. I vaguely remember discussion with @kain88-de where he convinced me that we should avoid `**kwargs` and use explicit kwargs because they immediately expose defaults, which reduces the need to list them explicitly.

package/MDAnalysis/analysis/density.py
orbeckst: This particularly `_reduce` method is different from the standard append one 
https://github.com/MDAnalysis/pmda/blob/26eb31e8269124c0d9e1382adcac8007c04681db/pmda/parallel.py#L446-L450 
so I would leave the commented out code.

orbeckst:
Docs should pass, too – took me a while to figure out that I cannot just `@deprecate` a class... using it on `__init__` works although it leaves a crappy message.

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -323,17 +397,64 @@ def u():
...
+    @pytest.mark.parametrize('sigma,ref_shape,ref_gridsum',
+                             [(None, (21, 26, 29), 108.710149),
+                              (1.5,  (21, 26, 29), 160.050167)])
+    def test_density_from_PDB(self, sigma, ref_shape, ref_gridsum):
+        # simple test that the code runs: use B-factors from PDB or fixed sigma
+        D = density.density_from_PDB(PDB_full, delta=2.0, sigma=sigma)
+
+        assert isinstance(D, density.Density)
+        assert_equal(D.grid.shape, ref_shape)
+        assert_almost_equal(D.grid.sum(), ref_gridsum, decimal=6)
```
IAlibay: Looks like the numpy 1.13.3 tests are failing here because it only matches to 5 d.p. instead of 6.

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -231,86 +334,57 @@ def test_density_from_Universe_userdefn_neqbox(self, universe, tmpdir):
...
-                import MDAnalysis.analysis.density
-            except ImportError:
-                pytest.fail(msg='''MDAnalysis.analysis.density should not raise
-                             an ImportError if gridData is available.''')
+    def test_has_DeprecationWarning(self, universe):
+        with pytest.warns(DeprecationWarning,
+                          match="will be removed in release 2.0.0"):
+            D = density.density_from_Universe(
+                universe, select=self.selections['static'],
+                delta=self.delta)
```
IAlibay: Probably missing something obvious here, but why was `test_presence_griddata` removed?

package/MDAnalysis/analysis/density.py
```diff
@@ -961,13 +1260,15 @@ def Bfactor2RMSF(B):
 
         \rm{RMSF} = \sqrt{\frac{3 B}{8\pi^2}}
 
-    .. rubric:: References
+    References
+    ----------
 
     .. [Willis1975]  BTM Willis and AW Pryor. *Thermal vibrations in crystallography*. Cambridge Univ. Press, 1975
```
IAlibay: Should this also have a `.. deprecated` docstring entry?

orbeckst:
I am not ignoring all your excellent reviews, I just don't have time to work on it at the moment! If anyone else wants to finish the PR, please do! Otherwise I'll get to it... when I get to it.

testsuite/MDAnalysisTests/analysis/test_density.py
orbeckst: mistake, probably

orbeckst: should be brought back

testsuite/MDAnalysisTests/analysis/test_density.py
orbeckst: need to decrease `decimal=5`

package/MDAnalysis/analysis/density.py
orbeckst: sure

orbeckst:
@IAlibay I am assigning you as the official chaperone for this PR.

Perhaps @wvandertoorn wants to finish this PR? That would be awesome.

wvandertoorn:
@IAlibay @orbeckst Definitely, I'm on it!

testsuite/MDAnalysisTests/analysis/test_density.py
wvandertoorn: I'm assuming that the whole ```TestGridImport``` class should be brought back, i.e. also```test_absence_griddata```, is that correct?

IAlibay: @wvandertoorn that's correct.

IAlibay:
@orbeckst just cleaned up the merge conflict to get travis et al. to run (I assume that's why appveyor/pr failed straight away...).

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -36,8 +36,8 @@
 import MDAnalysis as mda
 from MDAnalysis.analysis import density
 
-from MDAnalysisTests.datafiles import TPR, XTC, GRO
-from mock import Mock, patch
+from MDAnalysisTests.datafiles import TPR, XTC, GRO, PDB_full
+from unittest.mock import Mock, patch
```
IAlibay: Another one I failed to see in @wvandertoorn's PR. This import seems to be making Travis' python 2.7 tests fail. If you can add it in a new PR @wvandertoorn that would be great.

wvandertoorn:
@IAlibay I'm sorry to have missed it myself, wasn't familiar with sphynx before. I'll take care of the indentation issues right away, but the import failure isn't trivial too me, that might take a bit longer. Would you prefer the indentation and import fixes in separate PR's or in one?

IAlibay:
> @IAlibay I'm sorry to have missed it myself, wasn't familiar with sphynx before. I'll take care of the indentation issues right away, but the import failure isn't trivial too me, that might take a bit longer. Would you prefer the indentation and import fixes in separate PR's or in one?

@wvandertoorn if you look at the file difference, it looks like all you have to do is replace the unittest.mock import with: `from mock import Mock, patch`. My understanding here is that MDAnalysis uses the mock package: https://github.com/testing-cabal/mock which backports unittest.mock to pre-3.6 python.

wvandertoorn:
> @wvandertoorn if you look at the file difference, it looks like all you have to do is replace the unittest.mock import with: `from mock import Mock, patch`. My understanding here is that MDAnalysis uses the mock package: https://github.com/testing-cabal/mock which backports unittest.mock to pre-3.6 python.

Take it back, trivial indeed. Wasn't aware of testing-cabal's mock package, neat.

IAlibay:
> This looks great, I really like all the examples! Super tiny nitpicks.
> 
> I did think that `..versionchanged::` etc entries came in reverse-chronological order though, i.e. from most recent to least?

@lilyminium I'm not sure we've been super consistent in this unfortunately.

I've had a look through and for the most part there seems to be quite a few cases of `.. versionchanged` being in the order of oldest to newest. We probably need to agree on a convention and stick to it though.

Example cases:
topologyobjects, rms, lib/distances, PQRParser, MOL2Parser, analysis/__init__, core/groups, core/universe, topology/base, (kinda gave up looking after that).

@wvandertoorn would you like to raise a separate PR to address @lilyminium's comments?

wvandertoorn:
@IAlibay Yes, will do

orbeckst:
@IAlibay , if you could do the honors?

orbeckst:
p.s.: Merge so that @wvandertoorn 's commits survive.

IAlibay:
> p.s.: Merge so that @wvandertoorn 's commits survive.

@orbeckst Being wary of these new merge powers, it's probably best I ask for clarification here. What's the preferred option for MDAnalysis when trying to conserve commits on merge, merge rebase or a merge commit?

orbeckst:
If it's a simple fix from a single contributor then we typically *squash&merge*. We especially want to remove little micro-updates and reversals that make no sense when you read the commit history.

If it's a bigger thing with
- multiple non-trivial contributions 
- multiple self-contained commits 
then we typically just merge.

If committers are experienced with git we sometimes ask them to rewrite their history, especially if they care about individual commits.

If someone who isn't the original author of the commits rebased then we typically also merge because in the past we found that on a double-rebase, the original committer is lost. This might also be a case when history rewriting and commit-condensing is necessary.

IAlibay:
Thanks for the explanation @orbeckst that helped clear things up 😄 

Congrats on your first commits @wvandertoorn 🎉

orbeckst:
Thanks @wvandertoorn for finalizing this features – makes me very happy that this managed to get into 1.0, thanks to your push.

Issue #2489, opened 2020-01-31
ASV benchmarks: time_find_fragments    failure


orbeckst:
**Expected behavior**

ASV benchmarks for ag_methods.FragmentFinding.time_find_fragments all run.

**Actual behavior**

**Running ag_methods.FragmentFinding.time_find_fragments      failed**

```
               For parameters: '~/MDA/repositories/mdanalysis/benchmarks/env/491757a14a3783160c2e9816e3ecb44c/lib/python2.7/site-packages/MDAnalysisTests/data/adk_oplsaa.tpr', '~/MDA/repositories/mdanalysis/benchmarks/env/491757a14a3783160c2e9816e3ecb44c/lib/python2.7/site-packages/MDAnalysisTests/data/adk.psf', '~/MDA/repositories/mdanalysis/benchmarks/env/491757a14a3783160c2e9816e3ecb44c/lib/python2.7/site-packages/MDAnalysisTests/data/trz_psf.psf'
               Traceback (most recent call last):
                 File "~/MDA/miniconda3/envs/benchmark/lib/python3.6/site-packages/asv/benchmark.py", line 867, in <module>
                   commands[mode](args)
                 File "~/MDA/miniconda3/envs/benchmark/lib/python3.6/site-packages/asv/benchmark.py", line 838, in main_run
                   skip = benchmark.do_setup()
                 File "~/MDA/miniconda3/envs/benchmark/lib/python3.6/site-packages/asv/benchmark.py", line 442, in do_setup
                   result = Benchmark.do_setup(self)
                 File "~/MDA/miniconda3/envs/benchmark/lib/python3.6/site-packages/asv/benchmark.py", line 376, in do_setup
                   setup(*self._current_params)
               TypeError: setup() takes exactly 2 arguments (4 given)
```
and many others like this.

**Currently version of MDAnalysis**

Tested ones ... logfile available on request.

PR #2546, opened 2020-02-22
TST, BENCH: add asv check to Travis CI

### LINKED ISSUE #2545: TST/BENCH: add `asv check` to CI
Adding [`asv check`](https://asv.readthedocs.io/en/stable/commands.html#asv-check) to CI would help prevent issues like #2489. An example of usage in production CI: https://github.com/scipy/scipy/blob/master/.travis.yml#L264

Looks like the trick may be to specify usage of an existing environment to avoid paying the time cost of building a new one just for the syntax/asv lint check.

For example, I was able to reproduce Oli's original report without running expensive benchmarks:

```
· Discovering benchmarks
·· Creating virtualenv for virtualenv-py3.6-Cython-MDAnalysisTests-mock-nose-numpy-psutil-pytest-scipy-six.
·· Installing requirements for virtualenv-py3.6-Cython-MDAnalysisTests-mock-nose-numpy-psutil-pytest-scipy-six.........................................
·· Uninstalling from virtualenv-py3.6-Cython-MDAnalysisTests-mock-nose-numpy-psutil-pytest-scipy-six
·· Building 3e8ea8e0 <develop> for virtualenv-py3.6-Cython-MDAnalysisTests-mock-nose-numpy-psutil-pytest-scipy-six................
·· Installing 3e8ea8e0 <develop> into virtualenv-py3.6-Cython-MDAnalysisTests-mock-nose-numpy-psutil-pytest-scipy-six..
· Checking benchmarks
·· ag_methods.FragmentFinding.time_find_fragments: setup: wrong number of arguments (for <bound method FragmentFinding.setup of <benchmarks.ag_methods.FragmentFinding object at 0x7ff2a7413d68>> in ag_methods.py:336): expected 3, has 1
·· Benchmark suite check failed.

```

When I fixed that issue, `asv check` found a second one:
```
· Discovering benchmarks
· Checking benchmarks
·· analysis.rdf.SimpleRdfBench.time_interrdf: call: wrong number of arguments (for <bound method SimpleRdfBench.time_interrdf of <benchmarks.analysis.rdf.SimpleRdfBench object at 0x7fb9d2c86160>> in analysis/rdf.py:46): expected 3, has 2
·· Benchmark suite check failed.

```

After fixing the second issue:

```
· Discovering benchmarks
· Checking benchmarks
·· No problems found.

```

tylerjereddy:
* add `asv check` to Travis CI to help
catch issues in the benchmark suite without
running the full suite

Fixes #2545

- may require some iteration to get it working; the `time` command has been added as it should only take a few seconds once properly incorporated to use existing Python env for checking
- even if it does work, the CI should fail this check until the fix for #2489 is merged
- we could probably restrict this to a single matrix entry, though that would make it easier to accidentally remove it

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -121,6 +121,10 @@ script:
   - echo $MAIN_CMD $SETUP_CMD
   - if [[ $TRAVIS_OS_NAME == 'osx' ]]; then pip install pytest; fi
   - eval $MAIN_CMD $SETUP_CMD
+  # check benchmark suite for basic issues
+  - python -m pip install asv
+  - cd testsuite && python setup.py install
+  - cd ../benchmarks && time python -m asv check -E existing
 
 after_success:
   - |
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,6 +72,15 @@ matrix:
            SETUP_CMD="${PYTEST_FLAGS} --cov=MDAnalysis"
            PIP_DEPENDENCIES="${PIP_DEPENDENCIES} setuptools<45.0.0"
 
+    - env: NAME="asv check"
+           PYTHON_VERSION=2.7 
+           CODECOV="false"
+           MAIN_CMD=""
+           SETUP_CMD=""
+           NUMPY_VERSION=1.16
+           PIP_DEPENDENCIES="${PIP_DEPENDENCIES} setuptools<45.0.0"
+           ASV_CHECK="true"
+
     - os: osx
       env: PYTHON_VERSION=3.6
            NUMPY_VERSION=1.17.3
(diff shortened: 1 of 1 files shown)


tylerjereddy:
We may also need to update a few things in our `asv.conf.json` file, which is used by default & may have some settings that are too old now.

Let's see what CI says first though..

tylerjereddy:
Ok, the CI is correctly finding the failure reported in #2489 now. The actual check seems to take 1-2 seconds, which seems reasonable. The extra installs seem to be a few more seconds.

orbeckst:
Thank you, excellent idea.

Can you merge develop so that hopefully CI passes with the asv check?

.travis.yml
```diff
@@ -121,6 +121,10 @@ script:
   - echo $MAIN_CMD $SETUP_CMD
   - if [[ $TRAVIS_OS_NAME == 'osx' ]]; then pip install pytest; fi
   - eval $MAIN_CMD $SETUP_CMD
+  # check benchmark suite for basic issues
```
orbeckst: Would it be useful to make `asv check` a separate job and just run it on whatever we are actually running it on (namely Python 2.7...) instead of running it for every job?

I know that it's only a few seconds but the diagnostics would be cleaner if, say, only the asv check job fails while everything else still passes.

tylerjereddy:
I tried to revise accordingly & also rebase. We'll see what CI says..

tylerjereddy:
Ok, CI is all green now and for the asv check Travis job I see:

```
· Discovering benchmarks
· Checking benchmarks
·· No problems found.
```

There is some noisy stuff above it in the CI log, but we are using Python 2.7, etc. so probably not worth investigating.

PR #2548, opened 2020-02-23
Add Plotly plots to DiffusionMap

lilyminium:
Is MDAnalysis interested in supporting optional Plotly graphics in DiffusionMap? [(e.g. in this notebook, some scrolling required)](https://lilyminium.github.io/UserGuide/examples/analysis/reduced_dimensions/diffusion_map.html#Passing-your-own-distance-metric-to-DiffusionMap) It's convenient for quickly looking through the mapping at different time scales. 

 - Added 2D and 3D animated plots
 - Added frame checking/subsetting to `DiffusionMap.run()`. It's not clear that if the user `run()`s `DistanceMatrix` themselves, that they cannot alter the frames of the analysis in `DiffusionMap.run()`
 - Added NumPy arrays to `DiffusionMap`. `DiffusionMap` doesn't actually operate on a universe or trajectory, just the distance matrix thereof. If I have my own distance matrix, I would like to use `DiffusionMap` on it instead of writing my own analysis.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/analysis/diffusionmap.py
+++ b/package/MDAnalysis/analysis/diffusionmap.py
@@ -522,15 +522,15 @@ class DiffusionMap(object):
             embeds the data at integer ``time`` scales from 0 to ``n_frame``, 
             exclusive.  If ``None``, uses the number of frames in the analysis.
         colorscale: str, optional
-            Plotly colorscale for coloring. Default: 'Viridis'
+            Plotly colorscale for coloring.
         marker_size: float, optional
-            Marker size for plot. Default: 3
+            Marker size for plot.
         x: int, optional
-            Eigenvector to plot on x-axis. Default: 1
+            Eigenvector to plot on x-axis.
         y: int, optional
-            Eigenvector to plot on y-axis. Default: 2
+            Eigenvector to plot on y-axis.
         z: int, optional
-            Eigenvector to plot on z-axis. Default: 3
+            Eigenvector to plot on z-axis.
 
         Returns
         -------
--- a/package/MDAnalysis/analysis/diffusionmap.py
+++ b/package/MDAnalysis/analysis/diffusionmap.py
@@ -418,8 +418,11 @@ class DiffusionMap(object):
         diffusion_space : array (n_frames, n_eigenvectors)
             The diffusion map embedding as defined by [Ferguson1]_.
         """
-        return (self._eigenvectors[1:n_eigenvectors+1, ].T *
-                (self.eigenvalues[1:n_eigenvectors+1]**time))
+        try:
+            return (self._eigenvectors[1:n_eigenvectors+1, ].T *
+                    (self.eigenvalues[1:n_eigenvectors+1]**time))
+        except AttributeError:
+            raise ValueError('Call run() on DiffusionMap before using transform')
 
     def plot_animated_transform(self, n_frames=None, colorscale='Viridis', marker_size=3,
                                 x=1, y=2):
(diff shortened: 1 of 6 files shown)


package/MDAnalysis/analysis/diffusionmap.py
```diff
@@ -364,10 +365,191 @@ def transform(self, n_eigenvectors, time):
...
                 (self.eigenvalues[1:n_eigenvectors+1]**time))
+
+    def plot_animated_transform(self, n_time=25, colorscale='Viridis', marker_size=3,
+                                x=1, y=2):
+        """ Plots an interactive 2D plot of chosen eigenvectors at different timescales.
+
+        Parameters
+        ----------
+        n_time: int, optional
+            number of time values for ``transform()``. Default: 25
```
RMeli: I think the default parameters are automatically included in the documentation from the function definition. Should we avoid specifying them in the docstring as well, in order to avoid possible divergences between documentation and code?

package/MDAnalysis/analysis/diffusionmap.py
```diff
@@ -364,10 +365,191 @@ def transform(self, n_eigenvectors, time):
...
+
+
+        .. versionadded:: 1.0.0
+        """
+        try:
+            import plotly.graph_objects as go
+        except ImportError:
+            raise ImportError('Plotly is required for this animated graph')
+
+        if n_time is None:
```
RMeli: `n_frames` instead of `n_time` cold be a clearer name?

package/MDAnalysis/analysis/diffusionmap.py
```diff
@@ -364,10 +365,191 @@ def transform(self, n_eigenvectors, time):
...
+
+    def plot_animated_transform3D(self, n_time=25, colorscale='Viridis', marker_size=3,
+                                  x=1, y=2, z=3):
+        """
+        Plots an interactive 3D plot of chosen eigenvectors at different timescales.
+
+        Parameters
+        ----------
+        n_time: int, optional
+            number of time values for ``transform()``. Default: 25
```
RMeli: See comment above.

package/MDAnalysis/analysis/diffusionmap.py
```diff
@@ -364,10 +365,191 @@ def transform(self, n_eigenvectors, time):
...
+        """
+        try:
+            import plotly.graph_objects as go
+        except ImportError:
+            raise ImportError('Plotly is required for this animated graph')
+
+        eig = np.array([x, y, z]) - 1
+        n_eig = max(eig) + 1
+
+        if n_time is None:
```
RMeli: See comment above.

package/MDAnalysis/analysis/diffusionmap.py
lilyminium: Yes, I reckon so. I think I just do it because [everyone](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings) [else](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html) [seems](https://numpy.org/doc/1.18/reference/generated/numpy.trace.html#numpy.trace) [to](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.apply.html#pandas.DataFrame.apply), but that's not a very good reason -- maybe it's from when `**kwargs` was everywhere.

package/MDAnalysis/analysis/diffusionmap.py
lilyminium: I've changed it to `n_frames` for now because it *does* pass integers representing frames to ``transform()``, but I'm not sure that doesn't make it more confusing given that the keyword for ``transform`` is ``time``. What do you think of docstring/s?

lilyminium:
Added bits to deal with raw numpy arrays and start/stop/step mismatches in `run()`; updated the first comment to explain why.

hmacdope:
@lilyminium do we still want this ? Otherwise close.?

lilyminium:
Yea let’s close, at this point I think the last thing we want is more dependencies!On Oct 7, 2023, at 16:44, Hugo MacDermott-Opeskin ***@***.***> wrote:﻿
@lilyminium do we still want this ? Otherwise close.?

Issue #2470, opened 2020-01-27
mda.Merge for Universes without coordinates


lilyminium:
**Expected behavior**

I can make a new Universe with mda.Merge from my old Universe without positions.

**Actual behavior**

I get an AttributeError because mda.Merge does not wrap the `coords` in a try/except clause. MDAnalysis can make Universes without coordinates, so when the `args` don't have `positions`, the `coords[None, :, :]` argument to `Universe()` can be left out.

https://github.com/MDAnalysis/mdanalysis/blob/b5cb49fc06c05d4038f75ba98c66dc30f856d303/package/MDAnalysis/core/universe.py#L1534-L1537

**Code to reproduce the behavior**

Show us how to reproduce the failiure. If you can, use trajectory files from the test data.

``` python
>>> import MDAnalysis as mda
>>> from MDAnalysis.tests.datafiles import PSF
>>> u = mda.Universe(PSF)
>>> subset = mda.Merge(u.atoms[:10])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/lily/pydev/mdanalysis/package/MDAnalysis/core/universe.py", line 1535, in Merge
    coords = np.vstack([a.positions for a in args])
  File "/Users/lily/pydev/mdanalysis/package/MDAnalysis/core/universe.py", line 1535, in <listcomp>
    coords = np.vstack([a.positions for a in args])
  File "mdanalysis/package/MDAnalysis/core/groups.py", line 2294, in __getattr__
    cls=self.__class__.__name__, attr=attr))
AttributeError: AtomGroup has no attribute positions
```

**Currently version of MDAnalysis**

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) 0.20.2-dev
- Which version of Python (`python -V`)? 3.6
- Which operating system? MacOS

AnshulAngaria:
Hi! @lilyminium @orbeckst I want to work on this issue

orbeckst:
@AnshulAngaria sure, put up a PR and mention this issue in the "Fixes xx#" line.

PR #2549, opened 2020-02-24
test AnalysisFromFunction args order

### LINKED ISSUE #2503: analysis.AnalysisFromFunction processes arguments incorrectly
**Expected behavior**

If I don't pass a trajectory to AnalysisFromFunction, it looks for the first trajectory in my AtomGroup/s and then saves my given arguments. 

https://github.com/MDAnalysis/mdanalysis/blob/9d5520c611cb8ccfe7a35b1741b948e92475acb5/package/MDAnalysis/analysis/base.py#L226-L237

**Actual behavior**

When the trajectory is not passed, it moves the first argument to the end of `args` instead of the start. It then looks for the trajectory in the last AtomGroup.

``` python
>>> import MDAnalysis as mda
>>> from MDAnalysis.tests.datafiles import PSF, DCD
>>> from MDAnalysis.analysis.base import AnalysisBase, AnalysisFromFunction

>>> u = mda.Universe(PSF, DCD)
>>> protein = u.select_atoms('protein')

>>> def mass_xyz(atomgroup, masses):
...    return atomgroup.positions*masses

>>> xyz = AnalysisFromFunction(mass_xyz, protein, protein.masses)
>>> xyz.args
(array([14.007,  1.008,  1.008, ..., 12.011, 15.999, 15.999]),
 <AtomGroup with 3341 atoms>)
>>> xyz.run()
```

Error from `mass_xyz` receiving `(masses, atomgroup)` as the argument instead of `(atomgroup, masses)`
```
<ipython-input-4-50c897828315> in mass_xyz(atomgroup, masses)
      1 def mass_xyz(atomgroup, masses):
----> 2     return atomgroup.positions*masses
      3 
AttributeError: 'numpy.ndarray' object has no attribute 'positions'
```

**Possible solution**

```diff
        if (trajectory is not None) and (not isinstance(
                trajectory, coordinates.base.ProtoReader)):
---       args = args + (trajectory,)
+++       args = (trajectory,) + args
            trajectory = None

        if trajectory is None:
            for arg in args:
                if isinstance(arg, AtomGroup):
                    trajectory = arg.universe.trajectory
+++                 break
            # when we still didn't find anything
            if trajectory is None:
                for arg in six.itervalues(kwargs):
                    if isinstance(arg, AtomGroup):
                        trajectory = arg.universe.trajectory
```

**Tests**

One way to test a solution would be to verify that the breaking analysis in the example, actually works.

HTian1997:
Fixes #2503

 - Correct argument order inspired by @lilyminium 
 - More compact coding for AnalysisFromFunction function
 - Add new testing on AnalysisFromFunction. 

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/analysis/base.py
```diff
@@ -231,18 +231,16 @@ def __init__(self, function, trajectory=None, *args, **kwargs):
...
         if (trajectory is not None) and (not isinstance(
                 trajectory, coordinates.base.ProtoReader)):
-            args = args + (trajectory,)
+            args = (trajectory,) + args
             trajectory = None
 
         if trajectory is None:
-            for arg in args:
+            # all possible places to find trajectory
+            possible = [arg for arg in args] + [arg for arg in six.itervalues(kwargs)]
```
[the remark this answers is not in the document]
HTian1997: I think the test function should be written in analysis/test_base.py right?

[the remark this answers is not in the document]
lilyminium: Sorry, yes. And test `AnalysisFromFunction.args`. I was very distracted 🤦‍♀️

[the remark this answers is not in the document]
HTian1997: Don't fully understand "test `AnalysisFromFunction.args`". I have updated `test_base.py`. Please check whether that is what you mean.

HTian1997:
Sorry I was trying to solve issue [#2470](https://github.com/MDAnalysis/mdanalysis/issues/2470) but I didn't know that I should PR in a different branch. Is that working on multiple PRs at the same time should be in different branches? 😕

RMeli:
@HTian1997 Yes, if you are working on different PRs, they should be in different branches, otherwise all the changes will be updated in the same PR.

orbeckst:
Different PRs please. The smaller the PR, the easier it is to review.

HTian1997:
@orbeckst I have reverted the change. Please review.

testsuite/MDAnalysisTests/analysis/test_base.py
```diff
@@ -162,6 +162,22 @@ def test_AnalysisFromFunction(u, start, stop, step, nframes):
...
+    
+    protein = u.select_atoms('protein')
+    ans1 = base.AnalysisFromFunction(mass_xyz, u.trajectory, protein, protein.masses.reshape(-1, 1))
+    ans2 = base.AnalysisFromFunction(mass_xyz, protein, protein.masses.reshape(-1, 1))
+    
+    assert_equal(isinstance(ans1.args[0], mda.core.groups.AtomGroup), True)
+    assert_equal(isinstance(ans1.args[1], np.ndarray), True)
+    assert_equal(ans1.args[0], ans2.args[0])
+    assert_equal(ans1.args[1], ans2.args[1])
+    assert_equal(ans1.run().results, ans2.run().results)
```
lilyminium: Here you could have used ``assert isinstance(ans1.args[0], mda.core.groups.AtomGroup)`` instead.

In general I think the test should be a bit stronger, and test that `args[0] is protein` and same for the masses. It should also check that there is no mysterious `args[2]`. It would also be best to test that `ans2._trajectory is protein.universe.trajectory`, since that is what the code is actually looking for.

After that you probably don't need to actually `run()` the analysis.

testsuite/MDAnalysisTests/analysis/test_base.py
```diff
@@ -162,6 +162,22 @@ def test_AnalysisFromFunction(u, start, stop, step, nframes):
         assert_equal(results, ana.results)
 
 
+def test_AnalysisFromFunction_args_order(u):
+    # Issue 2503
+    def mass_xyz(atomgroup, masses):
+        return atomgroup.positions * masses
+    
+    protein = u.select_atoms('protein')
+    ans1 = base.AnalysisFromFunction(mass_xyz, u.trajectory, protein, protein.masses.reshape(-1, 1))
```
lilyminium: In my opinion, you don't need to have `ans1` here, since this test is for when the trajectory is not given. However, the tests don't actually have a case for when the trajectory is given, so it would be great if you made this a separate test and checked the argument order, and `ans1._trajectory` identity, again.

testsuite/MDAnalysisTests/analysis/test_base.py
```diff
@@ -162,6 +162,22 @@ def test_AnalysisFromFunction(u, start, stop, step, nframes):
...
+    ans2 = base.AnalysisFromFunction(mass_xyz, protein, protein.masses.reshape(-1, 1))
```
lilyminium: Please follow [PEP8](https://www.mdanalysis.org/UserGuide/contributing_code.html#code-formatting-in-python) line lengths when possible.

HTian1997:
@lilyminium I have added testing on 1) multiple traj inputs; 2) check `args[0] is protein` and `args[1] is masses`; 3) no mysterious `args[2]`; 4) `ans2._trajectory is protein.universe.trajectory`; 5) deleted unnecessary `ans1`. I forgot to pay attention to the line length. Will change that in the next commit.

testsuite/MDAnalysisTests/analysis/test_base.py
HTian1997: what do you mean by `made this a separate test and checked the argument order`, one function for testing case and another for testing argument order?

testsuite/MDAnalysisTests/analysis/test_base.py
```diff
@@ -162,6 +162,25 @@ def test_AnalysisFromFunction(u, start, stop, step, nframes):
         assert_equal(results, ana.results)
 
 
+def test_AnalysisFromFunction_args_order(u):
```
orbeckst: As a general rule, one test function for one assertion _especially_ if you are testing different things such as different arguments.

testsuite/MDAnalysisTests/analysis/test_base.py
```diff
@@ -162,6 +162,25 @@ def test_AnalysisFromFunction(u, start, stop, step, nframes):
...
+    # Issue 2503
+    def mass_xyz(atomgroup, masses):
+        return atomgroup.positions * masses
+    
+    protein = u.select_atoms('protein')
+    sub = u.atoms[:100]
+    #ans1 = base.AnalysisFromFunction(mass_xyz, u.trajectory, protein, protein.masses.reshape(-1, 1))
```
orbeckst: This should be one test function.

Each test should only create one `ans`.

testsuite/MDAnalysisTests/analysis/test_base.py
```diff
@@ -162,6 +162,25 @@ def test_AnalysisFromFunction(u, start, stop, step, nframes):
...
+    # lack traj but can find in protein
+    ans2 = base.AnalysisFromFunction(mass_xyz, protein, protein.masses.reshape(-1, 1))
```
orbeckst: This can be the center of the second test.

testsuite/MDAnalysisTests/analysis/test_base.py
```diff
@@ -162,6 +162,25 @@ def test_AnalysisFromFunction(u, start, stop, step, nframes):
...
+    # multiple traj input
+    ans3 = base.AnalysisFromFunction(mass_xyz, protein, sub, protein.masses.reshape(-1, 1))
```
orbeckst: Third test.

testsuite/MDAnalysisTests/analysis/test_base.py
```diff
@@ -162,6 +162,25 @@ def test_AnalysisFromFunction(u, start, stop, step, nframes):
...
+    
+    assert len(ans2.args) == 2
+    assert (ans2.args[0] == protein) and not (ans2.args[1] - protein.masses.reshape(-1, 1)).any()
+    assert ans2._trajectory == protein.universe.trajectory
+    assert ans3.args[0] == protein and ans3.args[1] == sub
+
```
orbeckst: Inspecting `ans` is good. (And  multiple `assert` for the _same_ `ans` are fine here.)

I would also add a short `ans.run(stop=5)` to make sure it really runs for a few steps. You should get something in `ans.result`. Just do a mean or a sum over whatever values you find there and check that value as an additional test.

testsuite/MDAnalysisTests/analysis/test_base.py
orbeckst: You might be able to write the three tests as a parametrized test with `@pytest.mark.paramterize` – however, if you haven't done this before, just write three separate tests to see if there's enough common structure and then we can help with condensing in a parametrized test.

testsuite/MDAnalysisTests/analysis/test_base.py
lilyminium: Also, just a note, `sub.universe.trajectory is protein.universe.trajectory` i.e. the [trajectories are the same object in memory](http://net-informations.com/python/iq/is.htm), so your test for multiple trajectory inputs would always succeed but for the wrong reasons. To test that you are always finding the *first* trajectory, you would need to create different universes and check `ans._trajectory is protein.universe.trajectory`.

testsuite/MDAnalysisTests/analysis/test_base.py
lilyminium: As @orbeckst elaborated on below, yes, checking the arguments when a trajectory is not given should be done in a separate test to checking the arguments when a trajectory is given.

HTian1997:
@orbeckst Separated into three functions. 0) self-defined function; 1) test args length and check `ans.run()` result; 2) check args content; 3) check whether it can catch first traj given multiple trajs.

testsuite/MDAnalysisTests/analysis/test_base.py
```diff
@@ -161,6 +161,31 @@ def test_AnalysisFromFunction(u, start, stop, step, nframes):
...
+    masses = protein.masses.reshape(-1, 1)
+    ans = base.AnalysisFromFunction(mass_xyz, protein, masses)
+    assert len(ans.args) == 2
+    assert int(sum(sum(sum(ans.run().results)))) == -317054
+
+def test_AnalysisFromFunction_args_content(u):
+    protein = u.select_atoms('protein')
+    masses = protein.masses.reshape(-1, 1)
+    ans = base.AnalysisFromFunction(mass_xyz, protein, masses)
+    assert (ans.args[0] == protein) and not (ans.args[1] - masses).any()
```
orbeckst: `(ans.args[1] - masses).any()` tries to compare floating point numbers to 0. That's fragile. Avoid doing this.

Furthermore, what is going to happen if the code is still wrong? Are you able to calculate `ans.args[1] - masses` or will this throw a TypeError or something like that? In this case, this is not a clean failure. Please explain a bit more what the reasoning is behind this test.

testsuite/MDAnalysisTests/analysis/test_base.py
```diff
@@ -161,6 +161,31 @@ def test_AnalysisFromFunction(u, start, stop, step, nframes):
...
+    assert  ans._trajectory == protein.universe.trajectory
+
+def test_AnalysisFromFunction_args_multiple_traj(u):
+    protein = u.select_atoms('protein')
+    masses = protein.masses.reshape(-1, 1)
+    another = mda.Universe(TPR, XTC).select_atoms("protein")
+    ans = base.AnalysisFromFunction(mass_xyz, another, protein, masses)
+    assert ans._trajectory == another.universe.trajectory
```
orbeckst: @lilyminium is this what it should be testing here?

lilyminium: Yes, although if you use [is instead of ==](http://net-informations.com/python/iq/is.htm) you could avoid importing new files. In general, using `is` would simplify these tests.

Also, `mass_xyz` wouldn't actually run here.

testsuite/MDAnalysisTests/analysis/test_base.py
HTian1997: if the code is wrong, that is, `ans.args[0] is masses` and `ans.args[1] is protein`, will return `AssertionError`. Do we need to deal with this error, like `pytest.raises` ?

testsuite/MDAnalysisTests/analysis/test_base.py
HTian1997: okay, I will change `==` to `is`. Even `mass_xyz` couldn't run here, we can still test the args order, is this a problem?

testsuite/MDAnalysisTests/analysis/test_base.py
orbeckst: `assert (ans.args[0] is protein) and (ans.args[1] is masses)` is better as you'll always get True/False answer and not an exception.

I'll call it a  "clean failure" when the `assert` raises `AssertionError` because of whatever you're asserting. Any other exception that gets raised will also make the test fail but in a way that we did not anticipate or explicitly test and it will show up differently when pytest runs. It's a sign of a bug, not a controlled test.

testsuite/MDAnalysisTests/analysis/test_base.py
HTian1997: @lilyminium I have revised `mass_xyz` which can run given two trajs.

HTian1997:
@orbeckst  I combined previous three functions into one as there is only one `ans` and only testing on `ans`.

HTian1997:
I really need to learn git command to amend instead of commit again and again on GitHub website 😭. So sorry about that.

RMeli:
@HTian1997 As long as your commit messages are sensible and informative, I think it's better to keep the whole history of your work. Rewriting/altering git history is a dangerous game, in my experience... They will be squashed together when merging, so they will not all end up in the MDAnalysis history anyway! ;)

HTian1997:
> @HTian1997 As long as your commit messages are sensible and informative, I think it's better to keep the whole history of your work. Rewriting/altering git history is a dangerous game, in my experience... They will be squashed together when merging, so they will not all end up in the MDAnalysis history anyway! ;)

HTian1997:
accidentally clicked close...

lilyminium:
Thanks @HTian1997! Please [update CHANGELOG](https://www.mdanalysis.org/UserGuide/contributing_code.html#documenting-your-code) as well with your username and the issue that you fixed :-)

package/CHANGELOG
```diff
@@ -134,6 +135,8 @@ Changes
 Deprecations
   * analysis.hbonds.HydrogenBondAnalysis is deprecated in 1.0 (remove in 2.0)
 
+Testsuite
+  * add test for args order of base.AnalysisFromFunction (PR #2594)
```
lilyminium: Tests don't typically get their own entry in the CHANGELOG. Otherwise, looks good to me!

orbeckst:
Thank you @HTian1997 ! We're waiting for CI to come back green and then we squash-merge.

HTian1997:
@lilyminium @orbeckst Thank you for the help!

orbeckst:
@HTian1997 congratulations to your first merged PR! Thank you. 🥇

Issue #1453, opened 2017-07-06
change timeseries "format" kw to "order"


orbeckst:
In various issues and discussions we sort-of decided that the `Reader.timeseries()` method should not use `format="afc"` as a kwarg but `order="afc"` (which make it consistent with the `MemoryReader`, avoids clashes with the builtin `format`, ... and avoids clashes with the `format` kwarg of `Universe` (yes, that one clashes with the builtin, too)).

orbeckst:
I think we should consider adding this to 0.17 so that we have some deprecation time.

orbeckst:
Note to myself: While playing around with the memory reader for the _Examples_ for [RMSF}(http://www.mdanalysis.org/mdanalysis/documentation_pages/analysis/rms.html#MDAnalysis.analysis.rms.RMSF) I found that not all orderings (keyword `order=...`) work for `Universe.load_new(array_like, order="...")` work. At least I got a cryptic error with `fac` (which is supposedly the default) while it worked with `fac`. Need to investigate more and possibly add tests for *all permutations*.

orbeckst:
Also note that the docs for the [MemoryReader](http://www.mdanalysis.org/mdanalysis/documentation_pages/coordinates/memory.html) need to be update:
- When a array_like is provided then it is automatically passed to the MemoryReader and the `format="MEMORY"` or `format=MemoryReader` is not needed anymore.
- Same for `Universe.load_new()`

In general we need to clarify how file types are recognized.

orbeckst:
I'll try to do this in my ample spare time ;-) – unless anyone else wants to: feel free to snatch this issue from me.

kain88-de:
@orbeckst do you have any code for your experiments of `order` that you did in august.

orbeckst:
No, sorry. But you can take the Example at https://www.mdanalysis.org/mdanalysis/documentation_pages/analysis/rms.html#MDAnalysis.analysis.rms.RMSF and just change `order` in the line
```python
# make a reference structure (need to reshape into a 1-frame "trajectory")
reference = mda.Merge(protein).load_new(
            reference_coordinates[:, None, :], order="afc")
```
and it should fail. 

(I am travelling and if you want to work on this please by all means do so and also assign yourself/un-assign me. Thanks!)

kain88-de:
Your example case works now with my changes.

orbeckst:
[...]

PR #2567, opened 2020-02-29
Remove format from MemoryReader.timeseries

IAlibay:
Towards v1.0 #2443 
Completes #1453 for MemoryReader 

 - Removes support for the deprecated `format` keyword in :meth:`MemoryReader.timseries`.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/coordinates/memory.py
```diff
@@ -510,26 +512,11 @@ def timeseries(self, asel=None, start=0, stop=-1, step=1, order='afc', format=No
...
 
-        Note
-        ----
-        The `format` parameter name is used to mimic the
-        :class:`MDAnalysis.coordinates.DCD.timeseries` interface. It is
-        identical to the `order` parameter for :class:`MemoryReader`. In a
-        future version, `format` will be renamed to `order`.
 
-
-        .. deprecated:: 0.17.0
```
orbeckst: IIRC, we leave in the old deprecation notices, sort of a history – or correct me...

IAlibay: I'm not sure we've been consistent with this, (see: https://github.com/MDAnalysis/mdanalysis/pull/2494#discussion_r374805918). Personally, I'm always in favour of maximising history in docstrings, so I am more than willing to add this one back if you would like to.

orbeckst: Thanks for showing that I have no idea what I said in the past. I am feeling very old  and senile now....

If we removed similar things recently then let's be consistent with the recent past. In any case, it's 1.0, who wants to look back? ;-)

PR #2571, opened 2020-02-29
Update groups.py

shfrz:
Fixes #

 - Improve missing AtomGroup property error messages

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


orbeckst:
Do we have an issue related to this patch? If yes, please update your issue report and add the issue number in the `Fixes #...` line. If not, please describe briefly what the problem is that you're solving. 

Generally speaking, when we identify a problem then we first raise an issue and then create a PR that addresses the issue. The reason is that PRs might stall or closed without a merge and then the issue report remains. For instance, sometimes the issue turns out to be much more complicated than initially thought and cannot be easily addressed but we still need to remember that it exists. Issues are permanent records, PRs are not.

package/MDAnalysis/core/groups.py
```diff
@@ -2222,8 +2222,8 @@ def __getattr__(self, attr):
             # eg:
             # if attr in _ATTR_ERRORS:
             # raise NDE(_ATTR_ERRORS[attr])
-            raise NoDataError("AtomGroup.{} not available; this requires Bonds"
-                              "".format(attr))
+            raise NoDataError("AtomGroup.{} not available; The attribute/method you are trying to access is not available because your topology doesn't contain information about {attr}"
+                              .format(attr))
```
fiona-naughton: You're using a named placeholder in your format string but only passing it a value rather than a key/value pair, so this will presumably raise a `KeyError`

shfrz: Hi @fiona-naughton. Oops! Thank you for pointing that out.

orbeckst:
@shfrz do you intend to continue working on this issue?

PR #2572, opened 2020-03-01
Fixed the run() method of analysis.hydrogenbonds.hbond_analysis method to use atom indices instead of ids

### LINKED ISSUE #2396: AttributeError when using MDAnalysis.analysis.hydrogenbonds.hbond_analysis with topology file without bonds information
**Expected behavior**

When given a universe without bonds attribute, should raise Exception 
``` python
  Cannot assign donor-hydrogen pairs via topology as no bonded information is present.                           
  Please either: load a topology file with bonded information; use the guess_bonds() 
  topology guesser; or set HydrogenBondAnalysis.donors_sel so that a distance cutoff                              
  can be used.
```

**Actual behavior**

And AttributeError was raised:
``` python
KeyError                                  Traceback (most recent call last)
~/anaconda3/envs/py36/lib/python3.6/site-packages/MDAnalysis/core/universe.py in __getattr__(self, key)
    509         try:
--> 510             segment = self._instant_selectors[key]
    511         except KeyError:

KeyError: 'bonds'

During handling of the above exception, another exception occurred:

<ipython-input-12-763dc8d4b193> in <module>
----> 1 gro.bonds

~/anaconda3/envs/py36/lib/python3.6/site-packages/MDAnalysis/core/universe.py in __getattr__(self, key)
    510             segment = self._instant_selectors[key]
    511         except KeyError:
--> 512             raise AttributeError('No attribute "{}".'.format(key))
    513         else:
    514             warnings.warn("Instant selector Universe.<segid> "

AttributeError: No attribute "bonds".
```

``` python
import MDAnalysis
from MDAnalysisTests.datafiles import GRO

kwargs = {
        'donors_sel': None,
        'hydrogens_sel': None,
        'acceptors_sel': None,
        'd_h_cutoff': 1.2,
        'd_a_cutoff': 3.0,
        'd_h_a_angle_cutoff': 120.0
        }

u = MDAnalysis.Universe(GRO)
h = HydrogenBondAnalysis(u, **kwargs)
h._get_dh_pairs()

```

  0.21.0
- Which version of Python?
  3.6.9

bieniekmateusz:
 - The run() method of `analysis.hydrogenbonds.hbond_analysis.HydrogenBondsAnalysis` currently stores the ids of atoms in a hydrogen bond, rather than the indices. The documentation states, and the helper methods assume, that it is atom indices that are stored.
- We have fixed this and added test cases that check the ids and indices of atoms in a hydrogen bond.
- Also Fixes #2396 

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

Co-authored-by: @p-j-smith  <paul.smith@kcl.ac.uk>


package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -408,8 +407,7 @@ def _get_dh_pairs(self):
 
         # If donors_sel is not provided, use topology to find d-h pairs
         if not self.donors_sel:
-
-            if len(self.u.bonds) == 0:
+            if not (hasattr(self.u, 'bonds') and len(self.u.bonds) != 0):
```
orbeckst: is it worthwhile testing for this case?

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -408,8 +407,7 @@ def _get_dh_pairs(self):
...
                 raise Exception('Cannot assign donor-hydrogen pairs via topology as no bonded information is present. '
```
orbeckst: Can you change this to something that's not a bare exception? I am not 100% sure what it should be, possibly our own [`NoDataError`](https://www.mdanalysis.org/mdanalysis/documentation_pages/exceptions.html#MDAnalysis.exceptions.NoDataError) or maybe @richardjgowers @jbarnoud @zemanj have an idea what we typically use when bond information is missing?

testsuite/MDAnalysisTests/analysis/test_hydrogenbonds_analysis.py
```diff
@@ -87,18 +87,111 @@ def test_count_by_type(self, h):
...
         assert int(counts[0, 2]) == ref_count
 
-    def test_count_by_ids(self, h):
+    def test_count_by_ids(self, h, universe):
 
         ref_counts = [1.0, 1.0, 0.5, 0.4, 0.2, 0.1]
         unique_hbonds = h.count_by_ids()
 
+        most_common_hbond_ids = [12, 14, 9]
+        assert np.allclose(unique_hbonds[0,:3], most_common_hbond_ids)
```
orbeckst: We prefer numpy's array comparison functions over bare asserts
```suggestion
        assert_equal(unique_hbonds[0,:3], most_common_hbond_ids)
```
See https://github.com/MDAnalysis/mdanalysis/wiki/Style-Guide%3A-Tests

testsuite/MDAnalysisTests/analysis/test_hydrogenbonds_analysis.py
```diff
@@ -87,18 +87,111 @@ def test_count_by_type(self, h):
...
+
         # count_by_ids() returns raw counts
         # convert to fraction of time that bond was observed
         counts = unique_hbonds[:, 3] / len(h.timesteps)
 
         assert_array_equal(counts, ref_counts)
```
orbeckst: While you're at it:
```suggestion
        assert_equal(counts, ref_counts)
```

testsuite/MDAnalysisTests/analysis/test_hydrogenbonds_analysis.py
```diff
@@ -87,18 +87,111 @@ def test_count_by_type(self, h):
...
+        'hydrogens_sel': 'name H1 H2',
+        'acceptors_sel': 'name O',
+        'd_h_cutoff': 1.2,
+        'd_a_cutoff': 3.0,
+        'd_h_a_angle_cutoff': 120.0
+    }
+
+    @staticmethod
+    @pytest.fixture(scope='class')
+    def universe():
```
orbeckst: This is cute!

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -408,7 +409,7 @@ def _get_dh_pairs(self):
         # If donors_sel is not provided, use topology to find d-h pairs
         if not self.donors_sel:
             if not (hasattr(self.u, 'bonds') and len(self.u.bonds) != 0):
```
zemanj: I also [posted](https://github.com/MDAnalysis/mdanalysis/issues/2396#issuecomment-596251787) this in the related issue for reference:

`hasattr(u, 'bonds')` is an *incredibly slow* test because it involves a lot of object instantiations under the hood. `hasattr(u._topology, 'bonds')` is **much** faster:
```python
import MDAnalysis as mda
from MDAnalysis.tests.datafiles import GRO, TPR

u = mda.Universe(GRO)  # universe *without* bonds
%timeit hasattr(u, 'bonds')
# 758 µs ± 4.33 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
%timeit hasattr(u._topology, 'bonds')
# 216 ns ± 2.88 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

u = mda.Universe(TPR)  # universe *with* bonds
%timeit hasattr(u, 'bonds')
# 88.8 ms ± 524 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)   <== AAARGH!
%timeit hasattr(u._topology, 'bonds')
# 86.4 ns ± 0.266 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```
You see that in the latter example, `hasattr(u, 'bonds')` is **a million times slower** than `hasattr(u._topology, 'bonds')`. This may be irrelevant for a setup routine that is run only once, but when being called in a loop, `hasattr(u, 'bonds')` is a nice way of very thoroughly killing performance. :grin:

richardjgowers:
The bonds lookup is because it actually calculates them too right? I think
that example is why I wanted to make Topology first class, so you could do
“‘bonds’ in u.topology”, else there’s not a clean/canonical way to check
for existence of attributes.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
bieniekmateusz: Hi @zemanj , 

Thanks for having a look at it. You've clearly done the work here. I can see that as you say, it is significantly slower. However, in a situation like this, I would advice to prioritise the simplicity of the code over performance. I believe you point this out as well. Thanks

p-j-smith: Thanks for pointing this out @zemanj - that's a major difference! If the atom selection is set to update, and the topology is to be used to find donor-hydrogen pairs, then this method (and thus the check for bonds) will be called for each frame of the analysis. So it's probably worth making this change.

orbeckst:
@zemanj are you happy with the changes? I'll leave it to you to shepherd the PR to the final merge.

testsuite/MDAnalysisTests/analysis/test_hydrogenbonds_analysis.py
```diff
@@ -87,15 +88,126 @@ def test_count_by_type(self, h):
...
+        kwargs = {
+            'donors_sel': None,
+            'hydrogens_sel': None,
+            'acceptors_sel': None,
+            'd_h_cutoff': 1.2,
+            'd_a_cutoff': 3.0,
+            'd_h_a_angle_cutoff': 120.0
+        }
+
+        with pytest.raises(NoDataError):
```
zemanj: Could you please test for the correct error message to appear? Sorry for nit-picking but I think your nice error message is the crucial part here. `with pytest.raises(NoDataError, match=<text or regex>):` should do the trick.

testsuite/MDAnalysisTests/analysis/test_hydrogenbonds_analysis.py
```diff
@@ -87,15 +88,126 @@ def test_count_by_type(self, h):
...
+        h = HydrogenBondAnalysis(universe, **self.kwargs)
+        h.run()
+
+        assert len(h.hbonds) == 2
+
+        frame_no, donor_index, hydrogen_index, acceptor_index, da_dst, dha_angle = h.hbonds[0]
+        assert donor_index == 0
+        assert hydrogen_index == 2
+        assert acceptor_index == 3
+        assert da_dst == 2.5
```
zemanj: Maybe use `assert_almost_equal` with a reasonably high precision here?

testsuite/MDAnalysisTests/analysis/test_hydrogenbonds_analysis.py
bieniekmateusz: Thanks @zemanj , I've added it as requested. However, I would advice against it. This would make the test rely on strings, which should not have such power. It is the Exception type here that is clearly defined. If you mean to say that we should inherit a more specific Exception type, then that would be more appropriate. Anyway, it is a minor thing! Thanks, Mat

testsuite/MDAnalysisTests/analysis/test_hydrogenbonds_analysis.py
bieniekmateusz: Updated the da_dst and dha_angle to `assert_almost_equal`. 

I also updated the indices to numpy `assert_equal` since they should be integers in the first place. Possibly a good task for a first time contributor? 

Let me know that's fine! Cheers

testsuite/MDAnalysisTests/analysis/test_hydrogenbonds_analysis.py
zemanj: I think that in general you're right, one usually shouldn't test for an error message. In this particular case, I think it's worth doing so. The reason is that we're soon gonna have `NoDataError`s for missing `AtomGroup` properties where appropriate, and I suspect that the same might eventually apply to `Universe` as well. Since your error message contains a lot of useful information, I want to avoid that it might be changed into a `NoDataError` with a generic error message in the future.

zemanj:
Thank you @bieniekmateusz @p-j-smith!

bieniekmateusz:
> @zemanj are you happy with the changes? I'll leave it to you to shepherd the PR to the final merge.

Thanks @zemanj, would you mind merging it when you get a chance? Thanks, Mat

PR #2573, opened 2020-03-01
improved unknown attribute errors in atomgroup and atom classes

### LINKED ISSUE #2565: Improve missing AtomGroup property error messages
**Expected behavior**
When I try to access an `AtomGroup` property that is generally valid but the corresponding data is missing from the underlying topology/trajectory, I expect a `NoDataError` with a message stating that the data I am requesting is not available from the topology/trajectory.

**Actual behavior**
When I load a topology that doesn't contain information about, e.g., charges, and try to access that `AtomGroup` property, an `AttributeError: AtomGroup has no attribute charges` is raised.

``` python
import MDAnalysis as mda
from MDAnalysis.tests.datafiles import GRO
u = mda.Universe(GRO)
u.atoms.charges
```

**Version of MDAnalysis**
Current 0.20.1 dev branch with Python 3.6 under Fedora 31
Should be version-independent though.

**Detailed problem description**
The `AttributeError` still makes sense if one tries to access an attribute that is generally unknown to `Atom` or `AtomGroup`. Thus, one needs to distinguish between attributes which *could* be there if the respective topology/trajectory info was there, and others that generally don't exist.
Furthermore, there are *derived* attributes such as `fragindices`, which are constructed from the topology's `bond` information. Thus, the error message for such derived attributes should state the missing information from which they are derived. The same is true for methods that require certain data, so that trying to access `AtomGroup.center_of_mass()` should raise a `NoDataError` due to missing information about masses.

Some examples: in the code example given above
* `u.atoms.boogiewoogie` should raise an `AttributeError`
* `u.atoms.charges` should raise a `NoDataError` due to missing charge information
* `u.atoms.fragindices` should raise a `NoDataError` due to missing bond information
* `u.atoms.unwrap()` should raise a `NoDataError` due to missing bond information
* `u.atoms.total_charge()` should raise a `NoDataError` due to missing charge information
* ...

**Proposed solution**
1. Create a dictionary with all *known* `AtomGroup` attributes as keys and the corresponding topology/trajectory information which is required for the attribute to exists.
2. Do the same for all known `Atom` attributes.
3. In the `__getattr__()` methods of `Atom` and `AtomGroup`, check whether the requested attribute is present in the respective dictionary's keys. If (and only if) that is the case, raise a `NoDataError` with an error message clarifying the problem. Otherwise, do nothing so that an `AttributeError` is automatically raised whenever a truly unknown attribute is accessed.
4. Write tests which cover all possible outcomes.

Regarding points 1 and 2, this involves parsing the `transplants` of `AtomGroup` and probably also those of `Atom`. Thanks to @richardjgowers we know how to automate the process so that the bookkeeping is done automatically, see his [comment](https://github.com/MDAnalysis/mdanalysis/issues/2565#issuecomment-594092648) below.

AnshulAngaria:
Fixes #2565

 Changed error message when an unknown attribute/method is accessed.

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/core/groups.py
+++ b/package/MDAnalysis/core/groups.py
@@ -2218,14 +2218,10 @@ class AtomGroup(GroupBase):
     def __getattr__(self, attr):
         # is this a known attribute failure?
         # TODO: Generalise this to cover many attributes
-        if attr in ('fragments', 'fragindices', 'n_fragments', 'unwrap'):
-            # eg:
-            # if attr in _ATTR_ERRORS:
-            # raise NDE(_ATTR_ERRORS[attr])
-            raise NoDataError("AtomGroup.{} not available; this requires Bonds"
-                              "".format(attr))
-        raise AttributeError("{cls} has no attribute {attr}".format(
-            cls=self.__class__.__name__, attr=attr))
+        raise NoDataError("The attribute/method you are trying to access is not available because"
+                            " your topology file doesn't contain information about {}."
+                            "".format(attr))
+
 
     def __reduce__(self):
         return (_unpickle, (self.universe.anchor_name, self.ix))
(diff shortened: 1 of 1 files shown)


orbeckst:
@zemanj , when you take the time to review (thank you!!) and you find that the PR is not ready to be merged, make it a "request changes" instead of a neutral comment as this sends a much clearer message.

Especially for GSoC, we want to communicate clearly with the students. They should ask for clarification if anything is unclear or if they don't know how to move forward.

zemanj:
@orbeckst that was on purpose, see [here](https://github.com/MDAnalysis/mdanalysis/issues/2565#issuecomment-593128763).

AnshulAngaria:
Thanks @orbeckst @zemanj for the response. I understand that someone else was already working on the issue. Right now I have already started working on #2470

orbeckst:
Ok, thanks for the context. Makes sense.

PR #2577, opened 2020-03-03
Weights rework for #2429

### LINKED ISSUE #2429: RMSD weights not applied to groupselections
**Expected behavior**

I expect that when `weights='mass'` is passed to `MDAnalysis.analysis.rms.RMSD`, it means that an array of masses is created from the atoms I have selected, for ``groupselections`` as well as ``select``.

**Actual behavior**

If any atom group that does not match ``select`` is selected for RMSD calculation in ``groupselections``, we get a ValueError because the size of the weights array does not match the selected atom group. Also, the masses selected are those for the reference atoms, not the mobile ones.

https://github.com/MDAnalysis/mdanalysis/blob/3e8ea8e0744f3163ddbf7b4f5f66874dd26796a7/package/MDAnalysis/analysis/rms.py#L557-L558

This is because `self.weights` is precomputed in `self._prepare()` as a single array. Instead, following @orbeckst's suggestion below, it should always be an iterable of arrays and raise an error when `groupselections` is provided, but only a single array of `weights` is provided.

eg.
```python
        if self.weights == 'mass':
            self.weights = [self.mobile_atoms.masses]
            for atoms in self._groupselections_atoms:
                self.weights.append(atoms['mobile'].masses)
        if self.weights is not None:
            if len(self.weights) != len(self.groupselections) + 1:
                raise ValueError('... message about weights!...')
```

```python
>>> import MDAnalysis as mda
>>> from MDAnalysis.tests.datafiles import PSF, DCD, CRD
>>> import MDAnalysis.analysis.rms as rms
>>> mda.__version__
'0.20.1'
>>> u = mda.Universe(PSF, DCD)  # closed AdK (PDB ID: 1AKE)
>>> R = rms.RMSD(u,  # universe to align
...              u,  # reference universe or atomgroup
...              select='backbone',  # group to superimpose and calculate RMSD
...              groupselections=["all"],  # groups for RMSD
...              weights='mass')  # weights for each atom
>>> R.run()
  File "MDAnalysis/analysis/base.py", line 197, in run
    self._single_frame()
  File "MDAnalysis/analysis/rms.py", line 638, in _single_frame
    center=False, superposition=False)
  File "MDAnalysis/analysis/rms.py", line 247, in rmsd
    raise ValueError('weights must have same length as a and b')
ValueError: weights must have same length as a and b
```

Note: ``weights='mass'`` does work when only ``select`` is passed and there are no ``groupselections``.

0.20.1
3.7.x
- Which operating system?
MacOS

I get that this becomes tricky when an array of weights is passed, but when ``RMSD`` is given "mass" as a string it should be fairly straightforward to apply it to the ``groupselections`` as well.

Also, I think it would be more flexible (if less convenient) to require that a ``weights`` array be the same shape as the entire universe, so that all selections can be weighted, rather than passing an array the same shape as ``select`` and failing for ``groupselections`` that necessarily will have different atoms.

yuxuanzhuang:
Fixes #2429

- Weights now accept *mass*, *None*, 1D float array or a list of them.
  - None and *mass* applies to all selections.
  - 1D float array applies only to select.
  - a list of them with length 1 + len(groupselection) applies to each
  selection correspondingly.

- Length mismatch between each groupselection and each weight is not yet checked.

Code fixed:

```python3
import MDAnalysis as mda
from MDAnalysis.tests.datafiles import PSF, DCD, CRD
import MDAnalysis.analysis.rms as rms
u = mda.Universe(PSF, DCD)  # closed AdK (PDB ID: 1AKE)
R = rms.RMSD(u,  # universe to align
            u,  # reference universe or atomgroup
            select='backbone',  # group to superimpose and calculate RMSD
            groupselections=["all"],  # groups for RMSD
            weights='mass')  # weights for each atom
R.run()
```

PR Checklist
------------
- [ ] Tests?
- [ ] Docs?
- [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/analysis/rms.py
+++ b/package/MDAnalysis/analysis/rms.py
@@ -575,20 +575,16 @@ class RMSD(AnalysisBase):
 
     def _prepare(self):
         self._n_atoms = self.mobile_atoms.n_atoms
-        # apply 'mass' weights for all selections
-        if self.weights == 'mass':
+        if self.weights == 'mass':         # apply 'mass' weights for all selections
            self.weights = ['mass'] * (len(self.groupselections) + 1)
-        # apply 'None' weights for all selections
-        elif self.weights is None:
-           self.weights = [None] * (len(self.groupselections) + 1)
-        # if weight is an 1D num array, apply it to select, and  apply 'None' weights to groupselections
+        elif self.weights is None:         # apply 'None' weights for all selections
+           self.weights = [None] * (len(self.groupselections) + 1)  
         elif (np.array(self.weights).ndim == 1) & (np.array(self.weights).dtype in (np.dtype('float64'),np.dtype('int64'))):
-           none_selection =[None] * len(self.groupselections)
-           none_selection.insert(0, self.weights)
+           none_selection =[None] * len(self.groupselections)       # if weight is an 1D num array, apply it to select, 
+           none_selection.insert(0, self.weights)                   # and apply 'None' weights to groupselections
            self.weights = none_selection
 
-        # add the array of weights to select
-        if not iterable(self.weights[0]) and self.weights[0] == 'mass':
+        if not iterable(self.weights[0]) and self.weights[0] == 'mass': # add the array of weights to weights_select
             self.weights_select = self.ref_atoms.masses
         elif self.weights[0] is None:
             self.weights_select = None
@@ -597,8 +593,7 @@ class RMSD(AnalysisBase):
         if self.weights_select is not None:
             self.weights_select = np.asarray(self.weights_select, dtype=np.float64) / np.mean(self.weights_select)
 
-       # add a list of arrays of weights for each groupselection
-        if self._groupselections_atoms:
+        if self._groupselections_atoms:                             # add the array of weights to select
             for igroup, atoms in enumerate(
                         self._groupselections_atoms, 3):
                 if not iterable(self.weights[igroup-2]) and self.weights[igroup-2] == 'mass':
@@ -687,7 +682,7 @@ class RMSD(AnalysisBase):
                         self._groupselections_atoms), 3):
                 self.rmsd[self._frame_index, igroup] = rmsd(
                     refpos, atoms['mobile'].positions,
-                    weights=self.weights_groupselection[igroup-3],
+                    weights=self.weights_groupselection[igroup-3],     #apply weights to each groupselection
                     center=False, superposition=False)
         else:
             # only calculate RMSD by setting the Rmatrix to None (no need
--- a/package/MDAnalysis/analysis/rms.py
+++ b/package/MDAnalysis/analysis/rms.py
@@ -544,11 +544,7 @@ class RMSD(AnalysisBase):
 
         # check weights type
 
-        if self.weights == 'mass':
-            pass
-        elif self.weights is None:
-            pass
-        elif (np.array(self.weights).ndim == 1) & (np.array(self.weights).dtype 
+        if (np.array(self.weights).ndim == 1) & (np.array(self.weights).dtype 
 						in (np.dtype('float64'),np.dtype('int64'))):
             self.weights = get_weights(self.mobile_atoms, self.weights)
         elif not iterable(self.weights):
@@ -559,11 +555,7 @@ class RMSD(AnalysisBase):
         else:
             weights_list = self.weights
             for weights in weights_list:
-                if weights == 'mass':
-                    pass
-                elif weights is None:
-                    pass
-                elif (np.array(weights).ndim == 1) & (np.array(weights).dtype 
+                if (np.array(weights).ndim == 1) & (np.array(weights).dtype 
 						in (np.dtype('float64'),np.dtype('int64'))):
                     # TODO:
                     # check each length of weights matches the length of each group selection?
--- a/package/MDAnalysis/analysis/rms.py
+++ b/package/MDAnalysis/analysis/rms.py
@@ -548,12 +548,14 @@ class RMSD(AnalysisBase):
             pass
         elif self.weights is None:
             pass
-        elif (np.array(self.weights).ndim == 1) & (np.array(self.weights).dtype in (np.dtype('float64'),np.dtype('int64'))):
+        elif (np.array(self.weights).ndim == 1) & (np.array(self.weights).dtype 
+						in (np.dtype('float64'),np.dtype('int64'))):
             self.weights = get_weights(self.mobile_atoms, self.weights)
         elif not iterable(self.weights):
             self.weights = get_weights(self.mobile_atoms, self.weights)
         elif len(self.weights) != len(self.groupselections) + 1:
-            raise ValueError("Length of array of weights is not equal to length of groupselections + 1")
+            raise ValueError("Length of array of weights is not equal to " 
+					"length of groupselections + 1")
         else:
             weights_list = self.weights
             for weights in weights_list:
@@ -561,7 +563,8 @@ class RMSD(AnalysisBase):
                     pass
                 elif weights is None:
                     pass
-                elif (np.array(weights).ndim == 1) & (np.array(weights).dtype in (np.dtype('float64'),np.dtype('int64'))):
+                elif (np.array(weights).ndim == 1) & (np.array(weights).dtype 
+						in (np.dtype('float64'),np.dtype('int64'))):
                     # TODO:
                     # check each length of weights matches the length of each group selection?
                     pass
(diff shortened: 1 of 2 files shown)


orbeckst:
Have a look at your test failures and see what you need to fix so the tests pass, e.g., https://travis-ci.com/MDAnalysis/mdanalysis/jobs/293359717

orbeckst:
(Generally speaking, we start reviewing in earnest once we see that tests pass.)

package/MDAnalysis/analysis/rms.py
```diff
@@ -535,29 +541,77 @@ def __init__(self, atomgroup, reference=None, select='all',
...
+            raise ValueError("Length of array of weights is not equal to length of groupselections + 1")
+        else:
+            weights_list = self.weights
+            for weights in weights_list:
+                if weights == 'mass':
+                    pass
+                elif weights is None:
+                    pass
+                elif (np.array(weights).ndim == 1) & (np.array(weights).dtype in (np.dtype('float64'),np.dtype('int64'))):
+                    # TODO:
```
orbeckst: complete this

package/MDAnalysis/analysis/rms.py
```diff
@@ -535,29 +541,77 @@ def __init__(self, atomgroup, reference=None, select='all',
...
+                    raise ValueError("Each groupselection can only be combined with "
+                                     "weight None or 'mass' or 1D float array")
+      
 
     def _prepare(self):
         self._n_atoms = self.mobile_atoms.n_atoms
+        # apply 'mass' weights for all selections
+        if self.weights == 'mass':
+           self.weights = ['mass'] * (len(self.groupselections) + 1)
+        # apply 'None' weights for all selections
```
orbeckst: put the comment inside its block instead of before the conditional (here and elsewhere) – that makes the code more readable

package/MDAnalysis/analysis/rms.py
```diff
@@ -535,29 +541,77 @@ def __init__(self, atomgroup, reference=None, select='all',
...
+        elif len(self.weights) != len(self.groupselections) + 1:
+            raise ValueError("Length of array of weights is not equal to length of groupselections + 1")
+        else:
+            weights_list = self.weights
+            for weights in weights_list:
+                if weights == 'mass':
+                    pass
+                elif weights is None:
+                    pass
+                elif (np.array(weights).ndim == 1) & (np.array(weights).dtype in (np.dtype('float64'),np.dtype('int64'))):
```
orbeckst: break long lines (see [Style Guide](https://github.com/MDAnalysis/mdanalysis/wiki/Style-Guide))

package/MDAnalysis/analysis/rms.py
```diff
@@ -535,29 +541,77 @@ def __init__(self, atomgroup, reference=None, select='all',
...
-        # be used with groupselection. (get_weights() returns the mass array
-        # for "mass")
-        if not iterable(self.weights) and self.weights == "mass":
+
+        # check weights type
+
+        if self.weights == 'mass':
             pass
-        else:
+        elif self.weights is None:
```
orbeckst: I would try to get rid of all the `pass` clauses as this will also simplify testing.

yuxuanzhuang:
Hi! As within the rework, this no long becomes an error. Should I modify the test code, or should I change the functionality of this implementation for now? 
masses as 1D float array

package/MDAnalysis/analysis/rms.py
orbeckst: Note: If `self.weights` is an array you will get a `FutureWarning` and eventually this will become an error. Should be changed to work with both scalars and arrays.

orbeckst:
If you change functionality that leads to code now doing things it couldn't do before then you change the tests accordingly.

package/MDAnalysis/analysis/rms.py
```diff
@@ -535,29 +540,90 @@ def __init__(self, atomgroup, reference=None, select='all',
...
-                iterable(self.weights) or self.weights not in ("mass", None))):
-            raise ValueError("groupselections can only be combined with "
-                             "weights=None or weights='mass', not a weight "
-                             "array.")
+        # check weights type
+        if (np.array(self.weights).ndim == 1) & (np.array(self.weights).dtype 
+						in (np.dtype('float64'),np.dtype('int64'))):
+            self.weights = get_weights(self.mobile_atoms, self.weights)
+        elif not iterable(self.weights):
+            self.weights = get_weights(self.mobile_atoms, self.weights)
```
fiona-naughton: Per above, the desired behaviour for `weights='mass'` is to use `'mass'` for both `select` and `groupselections`, which is what [this](https://github.com/MDAnalysis/mdanalysis/pull/2577/files#diff-d19ae35d57b874dea08d858a08da1041R580) in `_prepare()`  is for - but if we do have `weights='mass'` it'll be overwritten here when `get_weights` returns the atomic masses and won't be picked up below.

package/MDAnalysis/analysis/rms.py
```diff
@@ -535,29 +540,90 @@ def __init__(self, atomgroup, reference=None, select='all',
...
+					"length of groupselections + 1")
+        else:       # list of weights conditions
+            weights_list = self.weights
+            if (np.array(weights_list[0]).ndim == 1) & (np.array(weights_list[0]).dtype 
+                                                in (np.dtype('float64'),np.dtype('int64'))):
+                self.weights = get_weights(self.mobile_atoms, weights_list[0]) # check first one exclusively
+            for weights, atoms, selection in zip(weights_list[1:],
+                        self._groupselections_atoms,self.groupselections):
+                if str(weights) == 'mass':
+                    pass
```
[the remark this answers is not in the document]
yuxuanzhuang: Thank for all your comments! One thing I am not sure is that if I remove the None check pass, 
 `np.array(self.weights).ndim == 1` will fail. So I left this part for now. Hope the others get tidier.

orbeckst:
@fiona-naughton can you please continue reviewing? (And instead of "comment", select either "request changes" or "approve".) Once you're at "approve", please ping me. Of course, ping me or any other core dev if you need a second opinion.

Maybe @PicoCentauri could also help with some reviewing here, as it is a reasonably complicated PR?

package/MDAnalysis/analysis/rms.py
```diff
@@ -310,6 +309,20 @@ def process_selection(select):
     select['reference'] = asiterable(select['reference'])
     return select
 
+def weight_type_check(weights,atoms,selection):
```
PicoCentauri: Add a docstring or define this as a private method `weight_type_check`. I would prefer the latter.

package/MDAnalysis/analysis/rms.py
```diff
@@ -310,6 +309,20 @@ def process_selection(select):
...
+
+    if (np.array(weights).ndim == 1) & (np.array(weights).dtype 
+                            in (np.dtype('float64'),np.dtype('int64'))):
```
PicoCentauri: Do you really need the bitwise operator `&` here? More pythonic would be a pure `and`

package/MDAnalysis/analysis/rms.py
```diff
@@ -310,6 +309,20 @@ def process_selection(select):
...
+        if len(weights) != atoms.n_atoms:
+            raise ValueError("Length of provided weights {} do not match the number of atoms "
+                "in the selection {}: {}".format(weights, selection['mobile'], atoms.n_atoms))
+        else:
+            tmp_weights = get_weights(atoms, weights)
```
PicoCentauri: Remove the `tmp_weights` and just call the get_weights(atoms, weights) to check wether your weights are okay. 

Maybe there is also a way to combine your method with `get_weights` from utils.

package/MDAnalysis/analysis/rms.py
```diff
@@ -477,7 +494,9 @@ def __init__(self, atomgroup, reference=None, select='all',
         self.weights = weights
         self.tol_mass = tol_mass
         self.ref_frame = ref_frame
-
+        self.weights_ref = []              # save weights for reference
+        self.weights_select = []           # save weights for select
+        self.weights_groupselection = []   # save weights for each groupselection
```
PicoCentauri: Remove the comments here. Your property names are self-explanatory.

package/MDAnalysis/analysis/rms.py
```diff
@@ -381,11 +394,15 @@ def __init__(self, atomgroup, reference=None, select='all',
...
+        weights : {"mass", ``None``} or array_like or list of array_like (optional)
+             choose weights. With ``"mass"`` uses masses as weights for both `select`
+             and `groupselections`; with ``None`` weigh each atom equally for both 
+             `select` and `groupselections`. If a float array of the same length as
              `atomgroup` is provided, use each element of the `array_like` as a
-             weight for the corresponding atom in `atomgroup`.
+             weight for the corresponding atom in `select`, and assume ``None`` 
+             for `groupselections`. 
+             If weights is a list or array_like or 'mass' or None of length 
+             (1 + length(`groupselections`)), apply the weights correspondingly.
```
PicoCentauri: For me this is a bit confusing. Couldn't we split this into two weight attributes? One for `select` and one for `groupselections`. I'm not an expert for RMSDs but this reads like it is too much for a single attribute.

package/MDAnalysis/analysis/rms.py
```diff
@@ -535,29 +554,79 @@ def __init__(self, atomgroup, reference=None, select='all',
...
-        # cannot use arbitrary weight array (for superposition) with
-        # groupselections because arrays will not match
-        if (len(self.groupselections) > 0 and (
-                iterable(self.weights) or self.weights not in ("mass", None))):
-            raise ValueError("groupselections can only be combined with "
-                             "weights=None or weights='mass', not a weight "
-                             "array.")
+        # check weights type
+        #print(self.mobile_atoms)
+        if str(self.weights) == 'None':
```
PicoCentauri: Why are you converting this into a string? You can also use `if self.weights is None:`?

package/MDAnalysis/analysis/rms.py
yuxuanzhuang: I agree it becomes quite confusing/complicated (also for the code) after the change. But from a normal user's perspective, one normally uses `None`––which is the default for all selections; one might use `mass` in some rare cases––if splitting ´weight´ into two, you have to manually set both into `mass` (which is kind of prone to error?); I might never use custom weight...

package/MDAnalysis/analysis/rms.py
yuxuanzhuang: I thought it might have an impact when the comparison is elementwise in the future when I fix the FutureWarning for `self.weights == 'mass'`...ofc I was wrong.

yuxuanzhuang:
I accidentally messed up on my repository and have to start a new PR #2610 for more committing. Sorry for any inconvenience.

fiona-naughton:
Do you know what went wrong/did you try using `git reset`? Git can be frustrating when things go wrong, but it's convenient to keep all the comments etc in one place, so fixing this branch would be preferable to opening a new one.

package/MDAnalysis/analysis/rms.py
fiona-naughton: I agree with it might be better to split them. You could set it up to mimic the current behaviour - i.e. if weights aren't provided for `groupselections`, values are set based on the value provided for `select` rather than using a default.

orbeckst:
Btw, you could have recreated this branch (I am looking at the instructions that GitHub provides to me for manual merging, which you probably cannot see) although I don't know how you would re-link to the PR – perhaps pushing again to the original branch:
```
git checkout -b yuxuanzhuang-develop develop
curl -L https://github.com/MDAnalysis/mdanalysis/pull/2577.patch | git am -3   # git/GitHub MAGIC!!
# re-populate the branch
git push https://github.com/yuxuanzhuang/mdanalysis HEAD:develop
```

However, given that you restarted work on #2610 I am closing this PR.

PR #2580, opened 2020-03-04
Merge universes without coordinates Fix 2470

### LINKED ISSUE #2470: mda.Merge for Universes without coordinates

AnshulAngaria:
Fixes #2470

 - created and populated universes under try/except clause so that when args don't have positions the coords[None, :, :] argument to Universe() can be left out.

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/core/universe.py
```diff
@@ -1498,8 +1498,11 @@ def Merge(*args):
...
     # Create and populate a universe
-    coords = np.vstack([a.positions for a in args])
-    u = Universe(top, coords[None, :, :],
+    try:
+        coords = np.vstack([a.positions for a in args])
+        u = Universe(top, coords[None, :, :],
                  format=MDAnalysis.coordinates.memory.MemoryReader)
-
+    except:
+        u = Universe(top, format=MDAnalysis.coordinates.memory.MemoryReader)
```
richardjgowers: You don't need to specify `format=` here as there are no coordinates

AnshulAngaria: I have added tests and changelog entry and removed format=.
Thanks for the help.

package/CHANGELOG
```diff
@@ -20,6 +20,7 @@ mm/dd/yy richardjgowers, kain88-de, lilyminium, p-j-smith, bdice, joaomcteixeira
   * 0.21.0
 
 Fixes
+  * Create universes under try/except clause.(Issue #2470)(PR #2580)
```
orbeckst: Also add your GitHub to the authors list

package/MDAnalysis/core/universe.py
```diff
@@ -1498,8 +1498,11 @@ def Merge(*args):
...
 
     # Create and populate a universe
-    coords = np.vstack([a.positions for a in args])
-    u = Universe(top, coords[None, :, :],
+    try:
+        coords = np.vstack([a.positions for a in args])
+        u = Universe(top, coords[None, :, :],
                  format=MDAnalysis.coordinates.memory.MemoryReader)
-
+    except:
```
orbeckst: Explicitly list the exceptions on which the except clause triggers. Blanket exception catching will lead to very hard to find bugs.

package/CHANGELOG
richardjgowers: The entries in CHANGELOG are meant to indicate to users what has been fixed, ie here Merging Universes without coordinates.  This entry details what you did to fix it, but it's not useful information here

AnshulAngaria: I have updated CHANGELOG. Will take care of it next time.

richardjgowers:
cool thanks for the fix @AnshulAngaria !

AnshulAngaria:
@richardjgowers Thank you! Excited for doing more contributions!

PR #2584, opened 2020-03-06
[WIP] ITP topology Writer

xiki-tempula:
My project requires me to write an ITP writer. Thus, I'm thinking of creating a topology writer which will write the text-based itp file from the topology file read with ITPPraser. I haven't seen any open PR about this so I guess I will just go on and write it?

However, I'm not too sure about where the file should be. I cannot seem to find any topology writer in the folder topology.

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/coordinates/ITP.py
+++ b/package/MDAnalysis/coordinates/ITP.py
@@ -1,0 +1,30 @@
+# -*- Mode: python; tab-width: 4; indent-tabs-mode:nil; coding:utf-8 -*-
+# vim: tabstop=4 expandtab shiftwidth=4 softtabstop=4
+#
+# MDAnalysis --- https://www.mdanalysis.org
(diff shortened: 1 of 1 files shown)


IAlibay:
So an MD theological question if you'd like, does an ITP file count as a coordinates file?

xiki-tempula:
@IAlibay I'm very confused at the moment as I can only find TopologyReaderBase within the _Topologymeta.

richardjgowers:
We don't really do writing Topologies currently, mostly because they contain a lot more information (ie forcefields) than we currently handle.  To further @IAlibay 's theological concerns, I think this package deals with MD stuff post simulation, rather than pre.  I think maybe the best way to write a topology would be to convert to a parmed structure (via converters) and work from there.

xiki-tempula:
@richardjgowers Thank for the clarification. I will close this PR then.

PR #2610, opened 2020-03-10
Weights rework for #2429

### LINKED ISSUE #2429: RMSD weights not applied to groupselections

yuxuanzhuang:
Fixes #2429

- Weights now accept *mass*, *None*, 1D float array or a list of them.
  - None and *mass* applies to all selections.
  - 1D float array applies only to select.
  - a list of them with length 1 + len(groupselection) applies to each
  selection correspondingly.

- Length mismatch between each groupselection and each weight is not yet checked.

Code fixed:

```python3
import MDAnalysis as mda
from MDAnalysis.tests.datafiles import PSF, DCD, CRD
import MDAnalysis.analysis.rms as rms
u = mda.Universe(PSF, DCD)  # closed AdK (PDB ID: 1AKE)
R = rms.RMSD(u,  # universe to align
            u,  # reference universe or atomgroup
            select='backbone',  # group to superimpose and calculate RMSD
            groupselections=["all"],  # groups for RMSD
            weights='mass')  # weights for each atom
R.run()
```

PR Checklist
------------
- [x] Tests?
- [x] Docs?
- [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/analysis/rms.py
+++ b/package/MDAnalysis/analysis/rms.py
@@ -574,7 +574,7 @@ class RMSD(AnalysisBase):
                 self.weights_groupselections = [None] * len(self.groupselections)
 
         for igroup, (weights, atoms) in enumerate(zip(self.weights_groupselections,
-                                                      self._groupselections_atoms), 0):
+                                                      self._groupselections_atoms)):
             if str(weights) == 'mass':
                 self.weights_groupselections[igroup] = atoms['mobile'].masses
             if weights is not None:
--- a/package/MDAnalysis/analysis/rms.py
+++ b/package/MDAnalysis/analysis/rms.py
@@ -317,7 +317,7 @@ def weight_type_check(weights,atoms,selection):
             raise ValueError("Length of provided weights {} do not match the number of atoms "
                 "in the selection {}: {}".format(weights, selection['mobile'], atoms.n_atoms))
         else:
-            tmp_weights = get_weights(atoms, weights) 
+            get_weights(atoms, weights) 
     elif not iterable(weights) and str(weights) != 'mass' and weights is not None:
         raise ValueError("Each groupselection can only be combined with "
                                "weights: None, 'mass' or 1D float array")
--- a/package/MDAnalysis/analysis/rms.py
+++ b/package/MDAnalysis/analysis/rms.py
@@ -380,16 +380,22 @@ class RMSD(AnalysisBase):
 
             .. Note:: Experimental feature. Only limited error checking
                       implemented.
+
         weights : {"mass", ``None``} or array_like (optional)
-             Choose weights. With ``"mass"`` uses masses as weights for both `select`
-             and `groupselections`; with ``None`` weigh each atom equally for both
-             `select` and `groupselections`. If a float array of the same length as
-             `atomgroup` is provided, use each element of the `array_like` as a
-             weight for the corresponding atom in `select`, and assume ``None``
-             for `groupselections`.
-        weights_groupselections : False or list of {"mass", ``None``} or array_like (optional)
-            Default: ``False`` will apply imposed weights from ``weights`` option
-             Or a list with length of `groupselections`, apply the weights correspondingly.
+             1. "mass" will use masses as weights for both `select` and `groupselections`.
+
+             2. ``None`` will weigh each atom equally for both `select` and `groupselections`.
+
+             3. If 1D float array of the same length as `atomgroup` is provided,
+             use each element of the `array_like` as a weight for the
+             corresponding atom in `select`, and assumes ``None`` for `groupselections`.
+
+        weights_groupselections : False or list of {"mass", ``None`` or array_like} (optional)
+             1. ``False`` will apply imposed weights from ``weights`` option
+
+             2. a list of {"mass", ``None`` or array_like} with the length of `groupselections`,
+             apply the weights correspondingly.
+
         tol_mass : float (optional)
              Reject match if the atomic masses for matched atoms differ by more
              than `tol_mass`.
@@ -404,17 +410,15 @@ class RMSD(AnalysisBase):
         SelectionError
              If the selections from `atomgroup` and `reference` do not match.
         TypeError
-             If `weights` is not of the appropriate type; see also
-             :func:`MDAnalysis.lib.util.get_weights`
+             If `weights` or `weights_groupselections` is not of the appropriate type;
+             see also :func:`MDAnalysis.lib.util.get_weights`
         ValueError
              If `weights` are not compatible with `atomgroup` (not the same
              length) or if it is not a 1D array (see
              :func:`MDAnalysis.lib.util.get_weights`).
 
-             A :exc:`ValueError` is also raised if `weights` are not compatible
-             with `groupselections`: only equal weights (``weights=None``) or
-             mass-weighted (``weights="mass"``) are supported for additional
-             `groupselections`.
+             A :exc:`ValueError` is also raised if the length of `weights_groupselections` 
+             are not compatible with `groupselections`.
 
         Notes
         -----
@@ -481,8 +485,6 @@ class RMSD(AnalysisBase):
         self.weights = weights
         self.tol_mass = tol_mass
         self.ref_frame = ref_frame
-        self.weights_ref = []
-        self.weights_select = []
         self.weights_groupselections = weights_groupselections
         self.ref_atoms = self.reference.select_atoms(*select['reference'])
         self.mobile_atoms = self.atomgroup.select_atoms(*select['mobile'])
@@ -542,9 +544,7 @@ class RMSD(AnalysisBase):
                         len(atoms['reference']), len(atoms['mobile'])))
 
         # check weights type
-        if not iterable(self.weights) and self.weights == "mass":
-            pass
-        else:
+        if iterable(weights) or weights != "mass":
             get_weights(self.mobile_atoms, self.weights)
 
         if self.weights_groupselections:
@@ -555,9 +555,7 @@ class RMSD(AnalysisBase):
                                                  self._groupselections_atoms,
                                                  self.groupselections):
                 try:
-                    if not iterable(weights) and weights == "mass":
-                        pass
-                    else:
+                    if iterable(weights) or weights != "mass":
                         get_weights(atoms['mobile'], weights)
                 except Exception as e:
                     raise type(e)(str(e) + ' happens in selection %s' % selection['mobile'])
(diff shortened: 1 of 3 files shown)


orbeckst:
@fiona-naughton @PicoCentauri I assigned you to look after this PR. I am sorry that you will have to look back at your original comments in PR #2577 .

package/MDAnalysis/analysis/rms.py
```diff
@@ -381,11 +394,15 @@ def __init__(self, atomgroup, reference=None, select='all',
...
+        weights : {"mass", ``None``} or array_like or list of array_like (optional)
+             choose weights. With ``"mass"`` uses masses as weights for both `select`
+             and `groupselections`; with ``None`` weigh each atom equally for both 
+             `select` and `groupselections`. If a float array of the same length as
              `atomgroup` is provided, use each element of the `array_like` as a
-             weight for the corresponding atom in `atomgroup`.
+             weight for the corresponding atom in `select`, and assume ``None`` 
+             for `groupselections`. 
+             If weights is a list or array_like or 'mass' or None of length 
+             (1 + length(`groupselections`)), apply the weights correspondingly.
```
fiona-naughton: As discussed in #2577, I think it would be better to split `weights` into two attributes, say `weights` and `groupweights`. You can keep the default for the former as `None`, but have the latter default to say `False` so that can be caught and set to an appropriate value based on `weights` as with the current behaviour.

package/MDAnalysis/analysis/rms.py
```diff
@@ -535,29 +554,79 @@ def __init__(self, atomgroup, reference=None, select='all',
...
-            raise ValueError("groupselections can only be combined with "
-                             "weights=None or weights='mass', not a weight "
-                             "array.")
+        # check weights type
+
+        if self.weights is None:
+            pass
+        elif (type(self.weights) == str) or \
+            (np.array(self.weights).ndim == 1) and (np.array(self.weights).dtype 
+                                                in (np.dtype('float64'),np.dtype('int64'))):
```
fiona-naughton: You could combine these using `if not iterable(self.weights) or np.array( ...` to avoid needing the separate pass statement.

package/MDAnalysis/analysis/rms.py
```diff
@@ -335,6 +335,19 @@ class RMSD(AnalysisBase):
...
+        if (np.array(weights).ndim == 1) and (np.array(weights).dtype 
+                                in (np.dtype('float64'),np.dtype('int64'))):
+            if len(weights) != atoms.n_atoms:
+                raise ValueError("Length of provided weights {} do not match the number of atoms "
+                      "in the selection {}: {}".format(weights, selection['mobile'], atoms.n_atoms))
+            else:
+                get_weights(atoms, weights) 
+        elif not iterable(weights) and str(weights) != 'mass' and weights is not None:
+            raise ValueError("Each groupselection can only be combined with "
+                           "weights: None, 'mass' or 1D float array")
```
fiona-naughton: So having now looked at `get_weights`, it looks like that already does most of these checks/raises errors, so might be easier to use that rather than adding a new function. It would be useful to keep the more specific error message identifying which selection cause the error, since this wouldn't be clear from `get_weights` alone, but you could use say try/except to catch and expand on the errors when calling `get_weights`

package/MDAnalysis/analysis/rms.py
```diff
@@ -535,29 +554,79 @@ def __init__(self, atomgroup, reference=None, select='all',
...
+        # add the array of weights to weights_select
+        if str(self.weights[0]) == 'mass':
+            self.weights_select = self.mobile_atoms.masses
+            self.weights_ref = self.ref_atoms.masses
+        elif self.weights[0] is None:
+            self.weights_select = None
+            self.weights_ref = None
+        else:
+            self.weights_select = self.weights[0]
+            self.weights_ref = self.weights[0]
```
fiona-naughton: You could use `get_weights` with `self.weights[0]` and each `self.mobile_atoms` and `self.ref_atoms` to achieve this more simply

package/MDAnalysis/analysis/rms.py
```diff
@@ -535,29 +554,79 @@ def __init__(self, atomgroup, reference=None, select='all',
...
+                                                                        dtype=np.float64) / \
+                                       np.mean(self.weights_groupselection[igroup-3])                  
+                else:
+                    self.weights_groupselection.append(None)
+
+                if self.weights_groupselection[igroup-3] is not None:
+                    self.weights_groupselection[igroup-3] =   \
+                                       np.asarray(self.weights_groupselection[igroup-3], 
+                                                                        dtype=np.float64) / \
+                                       np.mean(self.weights_groupselection[igroup-3])
```
fiona-naughton: Weights set manually will end up being rescaled by the mean twice - the first rescaling is unnecessary

package/MDAnalysis/analysis/rms.py
yuxuanzhuang: Now the error raises as:
````
u = mda.Universe(PSF, DCD)
R = rms.RMSD(u, 
             u,              
             groupselections=['name CA and resid 1-5','name CA and resid 1'],
             weights_groupselections=[[1,1,1,1,1],[1,1]]) 

R.run()

weights (length 2) must be of same length as the atoms (1) happens in selection ['name CA and resid 1']
````

**Comment by yuxuanzhuang** [comment]
@fiona-naughton Thanks for all your comments! Hope the codes get tidier after weight splitting and `get_weights`:)

File: package/MDAnalysis/analysis/rms.py [hunk 6]
```diff
@@ -535,29 +541,52 @@ def __init__(self, atomgroup, reference=None, select='all',
...
-        if self.weights is not None:
-            self.weights = np.asarray(self.weights, dtype=np.float64) / np.mean(self.weights)
+        if not self.weights_groupselections:
+            if not iterable(self.weights):         # apply 'mass' or 'None' to groupselections
+                self.weights_groupselections = [self.weights] * len(self.groupselections)
+            else:
+                self.weights_groupselections = [None] * len(self.groupselections)
+
+        for igroup, (weights, atoms) in enumerate(zip(self.weights_groupselections,
+                                                      self._groupselections_atoms), 0):
```
PicoCentauri: 0 is the default. You don't need it here.

package/MDAnalysis/analysis/rms.py
```diff
@@ -535,29 +541,52 @@ def __init__(self, atomgroup, reference=None, select='all',
...
+            if weights == 'mass':
```
PicoCentauri: if weights is a numerical numpy array like masses this will raise 
```
`FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison`. 
```

We should probably catch this warning to avoid confusion.

package/MDAnalysis/analysis/rms.py
```diff
@@ -535,29 +541,52 @@ def __init__(self, atomgroup, reference=None, select='all',
...
+                raise ValueError("Length of weights_groupselections is not equal to "
+                                 "length of groupselections ")
+            for weights, atoms, selection in zip(self.weights_groupselections,
+                                                 self._groupselections_atoms,
+                                                 self.groupselections):
+                try:
+                    if not iterable(weights) and weights == "mass":
+                        pass
+                    else:
+                        get_weights(atoms['mobile'], weights)
```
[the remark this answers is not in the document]
yuxuanzhuang: assume it should be `iterable(weights) or weights != "mass"`.

yuxuanzhuang:
Thanks all for your help and comments!

orbeckst:
> (@orbeckst - don't know if you still wanted to be pinged per the original PR?)

No, thank you – you and @PicoCentauri are clearly do great here.

orbeckst:
There's an error on AppVeyor that seems to be related to this PR 
```
================================== FAILURES ===================================
___________________ TestRMSD.test_ref_mobile_mass_mismapped ___________________
self = <MDAnalysisTests.analysis.test_rms.TestRMSD object at 0x00000059CABD02E0>
universe = <Universe with 3341 atoms>
    def test_ref_mobile_mass_mismapped(self, universe):
        reference = MDAnalysis.Universe(PSF, DCD)
        universe.atoms.masses = np.zeros(len(universe.atoms.masses))
        with pytest.raises(ZeroDivisionError):
            RMSD = MDAnalysis.analysis.rms.RMSD(universe,
                                                reference=reference,
                                                select='all',
                                                weights='mass',
                                                tol_mass=100)
>           RMSD.run()
E           Failed: DID NOT RAISE <class 'ZeroDivisionError'>
analysis\test_rms.py:358: Failed
----------- coverage: platform win32, python 3.8.2-final-0 -----
```

Once this is addressed, it looks ready for a squash&merge.

yuxuanzhuang:
Not sure what the underlying problem is (I don't have a Windows machine at hand), but the test did pass in my Ubuntu env. I have changed to another way for this test. Hope it work this time. @orbeckst

PR #2611, opened 2020-03-10
Read XDR offsets from trajectory if file read-in fails

### LINKED ISSUE #1893: Failing to read XDR offsets is a fatal error
Sometimes loading xdr offsets fails. 

```python
  File "/home/user/miniconda2/envs/env/lib/python2.7/site-packages/MDAnalysis/coordinates/XDR.py", line 84, in read_numpy_offsets
    return {k: v for k, v in six.iteritems(np.load(filename))}
  File "/home/user/miniconda2/envs/env/lib/python2.7/site-packages/numpy/lib/npyio.py", line 429, in load
    "Failed to interpret file %s as a pickle" % repr(file))
IOError: Failed to interpret file '/host/user/dir/files/.file.xtc_offsets.npz' as a pickle
```

This shouldn't cause the entire file loading process to fail.  It should be possible wrap offset loading in a try/except to regenerate the offsets if needed

mlnance:
Fixes #1893 

 - Wrapped `read_numpy_offsets` in a try/except that allows for handling `IOError`
 - Added a check for if read-in failed and to `_read_offsets` instead
 - Provided user with warning if file read-in failed and defaulting to reading from trajectory
 - Added necessary tests to `test_xdr.py`
   - Add assertion to all tests that call `read_numpy_offsets` that return value is a `dict`
   - Test that an `IOError` returns `False` for `read_numpy_offsets`
   - Test that even with `IOError`, `_load_offsets` still returns expected value

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/AUTHORS
+++ b/package/AUTHORS
@@ -135,6 +135,7 @@ Chronological list of authors
   - Hugo MacDermott-Opeskin
   - Anshul Angaria
   - Shubham Sharma
+  - Morgan L. Nance
 
 
 External code
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -16,11 +16,13 @@ The rules for this file:
 mm/dd/yy richardjgowers, kain88-de, lilyminium, p-j-smith, bdice, joaomcteixeira,
          PicoCentauri, davidercruz, jbarnoud, RMeli, IAlibay, mtiberti, CCook96,
          Yuan-Yu, xiki-tempula, HTian1997, Iv-Hristov, hmacdope, AnshulAngaria,
-         ss62171, Luthaf
+         ss62171, Luthaf, mlnance
 
   * 0.21.0
 
 Fixes
+  * XDR offsets now read from trajectory if offsets file read-in fails on
+    IOError. Added relevant tests (Issue #1893, PR #2611)
   * Fixed test for the `save_paths` and `load` methods of :class:`PSAnalysis`
     (Issue #2593)
   * Fixes outdated :class:`TRJReader` documentation on random frame access
--- a/package/MDAnalysis/coordinates/XDR.py
+++ b/package/MDAnalysis/coordinates/XDR.py
@@ -113,6 +113,9 @@ class XDRBaseReader(base.ReaderBase):
     Reader. However, the  next time the trajectory is opened,  the offsets will
     have to be rebuilt again.
 
+    .. versionchanged:: 0.21.0
+       XDR offsets read from trajectory if offsets file read-in fails
+
     """
     def __init__(self, filename, convert_units=True, sub=None,
                  refresh_offsets=False, **kwargs):
--- a/package/MDAnalysis/coordinates/XDR.py
+++ b/package/MDAnalysis/coordinates/XDR.py
@@ -82,8 +82,11 @@ def read_numpy_offsets(filename):
         dictionary of offsets information
 
     """
-    return {k: v for k, v in six.iteritems(np.load(filename))}
-
+    try:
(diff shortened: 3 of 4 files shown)


mlnance:
@orbeckst thanks for the info on forking

This addition needs a test (and code review to ensure I added this feature properly, of course). Can anyone provide me with a relevant test file? Tagging @richardjgowers as you authored this issue originally

mlnance:
@richardjgowers thanks for the info on forking

This addition needs a test (and code review to ensure I added this feature properly, of course). Can anyone provide me with a relevant test file?

richardjgowers:
I think there’s two options;

You could create a bad file for a given trr file, where a bad file would be
something that np.load fails on, such as  txt file saying “hello”. Maybe a
clean way to do this is in a fixture that copies the trr to a tmp directory
then writes the silly bad file.

Or you could use mock to make np.load raise an IOError to spoof the failure
we’re hoping to observe. This is maybe a more “pure” way of doing it.

mlnance:
Sigh, accidentally closed...

mlnance:
@richardjgowers I believe I understand the mock procedure in theory (I've been googling and looking at how MDAnalysis uses mock to see if I can replicate it, I hadn't heard of `mock` until you mentioned it). My assumption is that I would pass a mocked path of some sort (off an offsets filename that does exist) to `_load_offsets` so that `read_numpy_offsets` can be called but will given `IOError`. I just have yet to figure out how to implement that. I'll keep researching, but if you have an any examples of where something like this has been done before in MDAnalysis testing, I'd appreciate it

fiona-naughton:
@mlnance the idea would be to create a mock of `np.load` that raises `IOError` rather than actually loading anything, and then run `_load_offsets`. Something like https://github.com/MDAnalysis/mdanalysis/blob/e416aa7d9adb9a9c14adb0774e65162da89a0d1c/testsuite/MDAnalysisTests/analysis/test_waterdynamics.py#L89-L95 with `np.load` instead of `universe.select_atoms` and setting `side_effect` to `IOError`.

mlnance:
@fiona-naughton thank you for the info, that was really helpful. The `mock` module is really cool, I'm glad I know about it now. I think I got the necessary tests going for what I have added. Tagging @richardjgowers

mlnance:
I'm not sure if this PR requires an update in documentation (I'm thinking not), but I know I do need to update the CHANGELOG

richardjgowers:
@mlnance this looks good.  Few things

- add a quick note to CHANGELOG (this is user facing, so briefly explain what you've fixed for them)
- add your github handle to CHANGELOG for 0.21
- add a quick note in the class with the change under a `.. versionchaged::` note
- add yourself to AUTHORS

mlnance:
> just needs documenting

@richardjgowers addressed, let me know if there is anything else

mlnance:
Thanks @richardjgowers @orbeckst @fiona-naughton for all the help/input/merging

I was wondering how I should interpret the change in the Codecov Report? Previously it said that my changes would change coverage by 0%, but after the branch-merge it is showing red for a decrease of 0.1%. However, only comments had been changed in both my last commit and the branch-merge. What exactly is the Codecov checking in this case?

orbeckst:
Codecov is not always super-reliable and fluctuates. We ignore these small changes.

orbeckst:
@richardjgowers will have a final look & squash-merge. Thank you @mlnance !

mlnance:
@orbeckst thanks again! this has been a fun process so far

richardjgowers:
Cool thanks for the fix @mlnance !

PR #2613, opened 2020-03-11
Added PCA subspace comparison methods and fixed n_components selection

### LINKED ISSUE #2623: PCA p_components and cumulated_variance incorrect when n_components is specified
## Expected behavior ##

p_components is a matrix that defines the first n principal components, where n is specified upon initialization. cumulated_variance is a vector that reflects the proportion of the total variance that is explained by the corresponding principal component and the ones that precede it, no matter how many PCs are kept.

## Actual behavior ##

When n_components is specified, p_components is a matrix with n rows. This produces an incorrect result because the components are represented as the column vectors. The values of the cumulated_variance vector change depending on n_components, and always converge to 1. 

This code uses my trajectory, but it illustrates the point. There are 238 atoms, and selecting 5 components. p_components is a 5x714 matrix where it should instead be a 714x5. The cumulated_variance converges to 1 even though the first 5 components do not actually explain 100% of the variation, but rather about 62%.

``` python
import MDAnalysis.analysis.pca as pca

p = pca.PCA(u, select='name CA', n_components=5).run()
p.p_components.shape
# (5, 714)
p.cumulated_variance
# array([0.6930588 , 0.80182406, 0.88727841, 0.94546782, 1.        ])
```

MDAnalysis 0.21.1
Python 3.7.5
Ubuntu 18.04

## Proposed solution ##

The _conclude function (line 272-281) of pca.py is currently:

``` python
    def _conclude(self):
        self.cov /= self.n_frames - 1
        e_vals, e_vects = np.linalg.eig(self.cov)
        sort_idx = np.argsort(e_vals)[::-1]
        self.variance = e_vals[sort_idx]
        self.variance = self.variance[:self.n_components]
        self.p_components = e_vects[:self.n_components, sort_idx]
        self.cumulated_variance = (np.cumsum(self.variance) /
                                   np.sum(self.variance))
        self._calculated = True
```

Fix:

``` python
    def _conclude(self):
        self.cov /= self.n_frames - 1
        e_vals, e_vects = np.linalg.eig(self.cov)
        sort_idx = np.argsort(e_vals)[::-1]
        self.variance = e_vals[sort_idx]
        self.cumulated_variance = (np.cumsum(self.variance) /
                                   np.sum(self.variance)) # calculated before variance slice
        self.variance = self.variance[:self.n_components]
        self.cumulated_variance = self.cumulated_variance[:self.n_components]
        self.p_components = e_vects[:, sort_idx[:self.n_components]] # all rows, slice of columns
        self._calculated = True
```

And the documentation at line 136 should read "p_components: array, (n_atoms * 3, n_components)" instead of "p_components: array, (n_components, n_atoms * 3)"

lilyminium:
Fixes #2623

This could be helpful for looking at subspaces.

 - method to calculate the root mean square inner product of subspaces
 - method to calculate the cumulative overlap of a vector in a subspace
 - fixed n_component selection of principal components as it was necessary for the subspace comparisons

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/analysis/pca.py
```diff
@@ -265,12 +269,25 @@ def _conclude(self):
...
+    @property
+    def n_components(self):
+        return self._n_components
+
+    @n_components.setter
+    def n_components(self, n):
+        if self._calculated:
+            if n is None:
+                n = len(self._variance)
+            self.variance = self._variance[:n]
```
orbeckst, line 286: Do we really want to permanently change `self.variance`?

In a recent discussion with @ianmkenney and @VOD555 we found it confusing that the "total" variance might only be  calculated from a subset of components. The docs were not very clear on this to start with. This can be confusing if you want to plot how much each PC contributes to the total observed variance but only selected, say,  `n=5`.

package/MDAnalysis/analysis/pca.py
```diff
@@ -327,10 +344,96 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+            dot[i] = np.dot(xyz, self._p_components[:, :dim])
 
         return dot
 
+    def rmsip(self, other, n_components=None):
+        """Compute the root mean square inner product between subspaces.
+
+        This is only symmetric if the number of components is the same for 
+        both instances. The RMSIP effectively measures how 
+        correlated the vectors of this instance are to those of ``other``.
```
orbeckst, line 356: Is there a reference available?

package/MDAnalysis/analysis/pca.py
```diff
@@ -327,10 +344,96 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+            number of components to compute for the inner products.
+            ``None`` computes all of them.
+
+        Returns
+        -------
+        float:
+            Root mean square inner product of the selected subspaces. 
+            0 indicates that they are mutually orthogonal, whereas 1 indicates 
+            that they are identical. 
+
```
orbeckst, line 372:
```suggestion
        See Also
        ---------
        rmsip
```

package/MDAnalysis/analysis/pca.py
orbeckst:
```suggestion

       Please cite [Amadei1999]_ and [Leo-Macias2004]_ if you use this function.
```

orbeckst: Adding `duecredit` citation would be great!

package/MDAnalysis/analysis/pca.py
```diff
@@ -365,3 +468,96 @@ def cosine_content(pca_space, i):
     cos = np.cos(np.pi * t * (i + 1) / T)
     return ((2.0 / T) * (scipy.integrate.simps(cos*pca_space[:, i])) ** 2 /
             scipy.integrate.simps(pca_space[:, i] ** 2))
+
+
+def rmsip(a, b, n_components=None):
```
orbeckst, line 473: Adding duecredit citation would be great!

package/MDAnalysis/analysis/pca.py
```diff
@@ -327,10 +344,96 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+        except AttributeError:
+            if isinstance(other, type(self)):
+                raise ValueError(
+                    'Call run() on the other PCA before using rmsip')
+            else:
+                raise ValueError('other must be another PCA class')
+
+        return rmsip(a.T, b.T, n_components=n_components)
+
+    def cumulative_overlap(self, other, i=0, n_components=None):
```
orbeckst, line 391: Adding duecredit citation would be great!

package/MDAnalysis/analysis/pca.py
```diff
@@ -365,3 +468,96 @@ def cosine_content(pca_space, i):
...
+        n_a = len(a)
+    if n_b is None:
+        n_b = len(b)
+
+    sip = np.matmul(a[:n_a], b[:n_b].T) ** 2
+    msip = sip.sum()/n_a
+    return msip**0.5
+
+
+def cumulative_overlap(a, b, i=0, n_components=None):
```
orbeckst, line 519: Adding duecredit citation would be great!

package/MDAnalysis/analysis/pca.py
lilyminium: No, hence the addition of the new attributes `_variance` and `_p_components`. The cumulative variance is also calculated from the total `_variance` so that now it does not add up to 1 if `n_components` is not `None`. TBH I don't really get the idea of throwing away information that was expensive to compute and is *already in memory*, so setting `n_components` is more of a convenient way to truncate or extend the visible variance, cumulated_variance, and p_components. It does not alter their values. We could suggest to users that they manually set ``self._variance = self.variance`` if they want to save space.

orbeckst: This should fix the issue that  @ianmkenney and @VOD555 saw – guys, if you still see a problem here, please say something now.

package/MDAnalysis/analysis/pca.py
```diff
@@ -265,12 +270,25 @@ def _conclude(self):
...
+        return self._n_components
+
+    @n_components.setter
+    def n_components(self, n):
+        if self._calculated:
+            if n is None:
+                n = len(self._variance)
+            self.variance = self._variance[:n]
+            self.cumulated_variance = (np.cumsum(self._variance) /
+                                       np.sum(self._variance))[:n]
```
orbeckst, line 289: Specifically, _this line_should fix the issue that @ianmkenney and @VOD555 saw.

package/MDAnalysis/analysis/pca.py
```diff
@@ -163,6 +167,7 @@ class PCA(AnalysisBase):
     Computation can be speed up by supplying a precalculated mean structure
 
     .. versionchanged:: 1.0.0
```
orbeckst, line 169: Also add a versionchanged entry for correct cumulative variance calculation for custom n_components.

orbeckst:
@lilyminium Please merge when the CI is happy. Either squash-merge or consolidate your history yourself if you think it makes more sense as separate commits.

lilyminium:
Thanks @orbeckst!

PR #2617, opened 2020-03-12
tqdm ProgressBar implementation

### LINKED ISSUE #928: replace custom ProgressMeter with tqdm
EDIT (@orbeckst): Based on the discussion in this issue we decided to replace our `lib.util.ProgressMeter` with [tqdm](https://github.com/tqdm/tqdm). Use this issue for discussion to refine this idea.

----
The progress bar could give an estimate of when the calculation is going to be done. This is reassuring for the end user, and allows to plan other things during long calculation (or to know when to come back from a coffee break).

This is easily achievable by keeping track of the time elapsed between two calls of the update method as we know how many frames remain to be analyzed.

It can be misleading for analyses that need to iterate several time through the trajectory, or that have a very long `_conclude` method. But the current progress bar is misleading already.

### LINKED ISSUE #2504: verbose not set in analysis.AnalysisBase.run
**Expected behavior**

From #1463, I think that `AnalysisBase.run(verbose=True)` is supposed to override the existing `self._verbose`. ~From #2206, it seems that `verbose` is supposed to be deprecated, but there are no deprecation warnings about it.~ (For what it's worth, I think the progress meter is cute.)

EDIT (@orbeckst): Analysis classes should be able to take the `verbose` kwarg for _both_ the `__init__()` constructor _and_ the `run()` method. The value give for `run()` should override the constructor value _while `run()` is executing_, as described in #1463 .

**Actual behavior**

Nothing happens.

https://github.com/MDAnalysis/mdanalysis/blob/9d5520c611cb8ccfe7a35b1741b948e92475acb5/package/MDAnalysis/analysis/base.py#L187

Nothing's done with `verbose`. `_setup_frames` defaults to False.

https://github.com/MDAnalysis/mdanalysis/blob/9d5520c611cb8ccfe7a35b1741b948e92475acb5/package/MDAnalysis/analysis/base.py#L149

cbouy:
Fixes #928 
Fixes #2504 

 - Added the `ProgressBar` class which inherits the `tqdm.auto.tqdm` object. For now, the `disable` keyword takes precedence over `verbose`, i.e. if both are set to True the progress bar won't show. Setting `disable=True` will disable the progress bar as expected. Setting `verbose=False` will disable the progress bar.
- Deprecate `ProgressMeter` through warning and text
- Replace `ProgressMeter` with `ProgressBar`

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


cbouy:
Now it should be able to automatically detect Jupyter (notebook or lab) versus the rest and use the appropriate `tqdm` class.
For JupyterLab, it requires a few things beforehand:
```
conda install -c conda-forge nodejs
jupyter labextension install @jupyter-widgets/jupyterlab-manager
```

.appveyor.yml
```diff
@@ -10,8 +10,8 @@ cache:
 environment:
     global:
         CONDA_CHANNELS: conda-forge
-        CONDA_DEPENDENCIES: pip setuptools wheel cython mock six biopython networkx joblib matplotlib scipy vs2015_runtime pytest mmtf-python GridDataFormats hypothesis pytest-cov codecov chemfiles
```
orbeckst: Why was chemfiles deleted?

.appveyor.yml
```diff
@@ -10,8 +10,8 @@ cache:
...
-        PIP_DEPENDENCIES: gsd==1.9.3 duecredit parmed
+        CONDA_DEPENDENCIES: pip setuptools wheel cython mock six biopython networkx joblib matplotlib scipy vs2015_runtime pytest mmtf-python GridDataFormats hypothesis pytest-cov codecov
+        PIP_DEPENDENCIES: gsd==1.9.3 duecredit parmed tqdm
```
orbeckst: Is there no tqdm conda package for Windows? (Generally we prefer the conda package.)

package/MDAnalysis/lib/log.py
```diff
@@ -394,3 +396,34 @@ def echo(self, step, **kwargs):
             return
         echo(self.format_handler(format, vars(self)),
              replace=self.dynamic, newline=newline)
+
+
+# detect if running from jupyter, ipython or python
+try:
+    shell = get_ipython().__class__.__name__
```
orbeckst: Where does `get_ipython()` come from? Does it only exist in jupyter/ipython?

package/MDAnalysis/lib/log.py
```diff
@@ -394,3 +396,34 @@ def echo(self, step, **kwargs):
...
+except: # non-interactive shell
```
orbeckst: Can you be more precise about the exception that is raised, e.g., `NameError`?

Some comment on how you do the detection would make it clearer.

package/MDAnalysis/lib/log.py
```diff
@@ -394,3 +396,34 @@ def echo(self, step, **kwargs):
...
+    from tqdm import tqdm as _tqdm_class
+else:
+    if shell == 'ZMQInteractiveShell': # Jupyter notebook and JupyterLab
+        from tqdm.notebook import tqdm_notebook as _tqdm_class
+    else: # Ipython
+        from tqdm import tqdm as _tqdm_class
+
+class ProgressBar(_tqdm_class):
+    r"""tqdm-like progress bar. Replaces the :class:`ProgressMeter` class to
```
orbeckst: This is placeholder text and eventually needs to be replace by something more meaningful.

If you leave out the doc string, will it inherit the one from `_tqdm_class`?

orbeckst:
@jbarnoud I am adding you to this issue as the original driving force behind #928

.appveyor.yml
cbouy: That's an error from my part, my bad

.appveyor.yml
cbouy: Yes there is, good to know. I'll change that

package/MDAnalysis/lib/log.py
cbouy: It comes from IPython and is automatically imported when running an ipython shell or jupyter notebook/lab

package/MDAnalysis/lib/log.py
cbouy: I noticed that tqdm actually has a class for automatic detection of this so I'll be changing the code to use their solution instead

package/MDAnalysis/lib/log.py
cbouy: Yes, no doc-string = use the doc-string from the parent class. Should I do that instead ?

package/MDAnalysis/analysis/base.py
```diff
@@ -177,13 +170,16 @@ def run(self, start=None, stop=None, step=None, verbose=None):
         self._setup_frames(self._trajectory, start, stop, step)
         logger.info("Starting preparation")
         self._prepare()
-        for i, ts in enumerate(
-                self._trajectory[self.start:self.stop:self.step]):
+        verbose = getattr(self, '_verbose', False)
+        n_frames = len(range(*self._trajectory.check_slice_indices(
```
richardjgowers: Rather than calculating the n_frames, I thought that a trajectory slice knew it's length?  Ie `len(u.trajectory[::2])` should work, so maybe tqdm can infer the length of the traj slice

cbouy: That would be true if there wasn't an `enumerate` around the traj slice, and we need it here ! But I just realised I can remove the `n_frames` for hbond_analysis.py, remove the `enumerate` in density.py, helanal.py and remove both `n_frames` and `enumerate` in pca.py. Thanks !

jbarnoud: You should still use len on the sliced trajectory to get n_frames. It would make the code more readable and avoid potential issues we had mixing slices and range.

cbouy: Makes sense ! I also just realised I can put the ProgressBar inside enumerate instead of the other way around, which makes it even simpler.

package/MDAnalysis/lib/log.py
orbeckst: yes

orbeckst:
The Python 2.7 version of the test is failing https://travis-ci.com/github/MDAnalysis/mdanalysis/jobs/298190141, perhaps something with comparing utf-8 strings.

cbouy:
@jbarnoud should I make a proper test module for the ProgressBar and ProgressMeter then ? I find it counter-intuitive to have the test for the ProgressMeter alongside the tests for the RMSD

jbarnoud:
I expect a test that creates a ProgresdMeter and makes sure the deprecation warning is issued. It should probably go with the tests of the log module.The point is to make sure that users that use the progress meter on its own actually see the warning, and that displaying it does not cause an error. We had issues like that in the past, where raising an exception was itself causing an error, because of format errors in the message. On 15 Mar 2020 17:44, Cédric Bouysset <notifications@github.com> wrote:
@jbarnoud should I make a proper test module for the ProgressBar and ProgressMeter then ? I find it counter-intuitive to have the test for the ProgressMeter alongside the tests for the RMSD

cbouy:
I don't understand the report from codecov, could someone explain what's missing ? I've added a new class, a test for this new class, I moved a test from test_rms.py to test_log.py and created a new test for the deprecation. There should be an increase in coverage or am I getting this wrong ?

orbeckst:
Codecov can be mysterious... 

Sorry for the long silence. Will have a look now.

package/MDAnalysis/analysis/hbonds/hbond_autocorrel.py
```diff
@@ -392,12 +392,10 @@ def run(self, force=False):
         # for normalising later
         counter = np.zeros_like(master_results, dtype=np.float32)
 
-        pm = ProgressMeter(self.nruns, interval=1,
-                           format="Performing run %(step)5d/%(numsteps)d"
-                                  "[%(percentage)5.1f%%]")
-
-        for i, (start, stop) in enumerate(zip(self._starts, self._stops)):
-            pm.echo(i)
+        n_frames = len(self.u.trajectory)
```
orbeckst: `n_frames` is the wrong total, should be `self.nruns`

package/MDAnalysis/analysis/hbonds/hbond_autocorrel.py
```diff
@@ -392,12 +392,10 @@ def run(self, force=False):
...
+        for i, (start, stop) in ProgressBar(enumerate(zip(self._starts,
+                                            self._stops)), total=n_frames,
```
orbeckst:
```suggestion
                                            self._stops)), total=self.nruns,
```

orbeckst:
> Codecov can be mysterious...

The above was a bit flippant but what I wanted to say is that it's not always clear that codecov is showing us what we expect to see and although we like to see everything green in the CI, we do overrule Codecov once we are happy.

jbarnoud:
@lilyminium noticed in https://github.com/MDAnalysis/mdanalysis/pull/2631#issuecomment-601649236 that the API changes. It probably should not.

lilyminium:
https://github.com/MDAnalysis/mdanalysis/pull/2631#issuecomment-601649236 was phrased poorly -- as far as I can tell this does not change the API, except that `AnalysisBase._pm` no longer exists. #2631 did alter the arguments given to `_setup_frames`. This PR also fixes #2504.

package/MDAnalysis/analysis/hbonds/hbond_autocorrel.py
cbouy: Oh yes ! my bad

package/CHANGELOG
```diff
@@ -17,7 +17,7 @@ mm/dd/yy richardjgowers, kain88-de, lilyminium, p-j-smith, bdice, joaomcteixeira
          PicoCentauri, davidercruz, jbarnoud, RMeli, IAlibay, mtiberti, CCook96,
          Yuan-Yu, xiki-tempula, HTian1997, Iv-Hristov, hmacdope, AnshulAngaria,
          ss62171, Luthaf, yuxuanzhuang, abhishandy, mlnance, shfrz, orbeckst,
-         wvandertoorn
+         wvandertoorn, cbouy
```
orbeckst: Could you please update the CHANGELOG to also say that #2504 is fixed?

orbeckst:
@lilyminium @jbarnoud  thanks for pointing out that this PR fixes #2504, too https://github.com/MDAnalysis/mdanalysis/pull/2631#issuecomment-601649236.

package/CHANGELOG
lilyminium: This wasn’t in the scope of the original PR, but if you do, could you add a test that the verbose keyword does actually toggle the ProgressBar?

cbouy: Done !

cbouy:
Thank you @orbeckst @jbarnoud @richardjgowers for your help ! And thanks @lilyminium for mentioning the other issue !

orbeckst:
One of the macOS jobs failed right at the start. I restarted https://travis-ci.com/github/MDAnalysis/mdanalysis/jobs/300680256 – hopefully this will all look good in an hour.

orbeckst:
@cbouy very nice contribution!

PR #2609, opened 2020-03-10
Secondary structure determination

lilyminium:
Fixes #2608 

Changes made in this Pull Request:
 - added `secondary_structure` module
 - contains wrapper classes for:
   - mdtraj.compute_dssp
   - DSSP (mkdssp)
   - STRIDE (stride)

Let’s see how these installation scripts go...

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?
 - [x] Issue raised/referenced?

[Demo notebook here](https://lilyminium.github.io/UserGuide/examples/analysis/structure/dssp.html).

PR #2622, opened 2020-03-13
🧬 helanal refactor 🧬

### LINKED ISSUE #2452: update helanal to AnalysisBase
**Expected behavior**

When I run ``helanal.helanal_trajectory``, it produces readable output.

**Actual behavior**

``helanal.helanal_trajectory`` writes 8 files (there is no way to turn this off):

- bending_matrix.dat
- fit_tilt.xvg
- local_bend.xvg
- local_tilt.xvg
- origin.pdb
- screw.xvg
- summary.txt
- unit_twist.xvg

None of the `.xvg` files have delimiters, resulting in something like this (unit_twist.xvg):

>```
>0173.50089152.36047157.11974168.0601134.1332682.78972102.89294480.54708134.09038143.31886166.2364105.43222124.6999295.75029122.1871694.43125101.5204499.2620497.0216893.1698991.6502786.83874136.04012134.16911119.11755132.66452166.35683171.27223122.5070494.534966102.8902595.2906597.6656998.8994297.9797.82383102.0121695.7262297.1709147.01093149.21326110.0322688.56357102.8416599.71714105.5752799.8775199.056145101.6334296.66122494.9870894.4209499.65817135.18112141.92354114.745056120.963615134.35696130.1924107.3315199.60976101.5804996.4396102.3210396.43453102.259056104.7718392.5648796.93780596.2955106.55468119.2636123.093636104.8041102.03329133.9388796.147766111.26341105.42476127.773766158.32767162.8811374.15202146.2733362.067436170.5584664.37851110.36818101.4832895.5074697.5645998.09011101.22569103.59741100.17954100.5833197.70348126.24198155.93248169.31538163.89807153.8199364.76133159.19398172.53307168.54634154.64725157.2409160.78711117.229416130.50992101.69997498.69098102.3421398.5284197.9638897.9031384.2355692.51822101.50628112.175995141.17561179.15326115.91146162.714297.18241148.72191177.6465675.098206162.01685178.84647153.61559143.4729582.728778.1872974.47471169.2817782.82487176.68405159.40784127.06898576.81265108.926895116.31709143.0244387.8953476.859406138.5434143.4651297.01566138.7108172.56514115.02348133.1791475.7148175.10802104.24445131.14348144.5514105.2866192.4977393.9222296.55736101.0048695.58964100.96390594.60849102.6927897.4528593.37078100.4637390.25625692.76536109.01559498.313156104.478905100.8604699.90523101.04384695.73096101.6445100.1566798.6450398.35263101.63966104.76728139.75473134.5733364.02203109.86525159.32852177.16191148.17155163.89886151.0398174.66324114.77936119.296104104.45007126.39088499.2082396.272104.2805799.6220799.19309105.1044101.61230598.236244101.4117290.6837682.89361
>```

These aren't fixed-length either. Throwing in a newline after each value gives values with a mix of 4 or 5 decimal places.

> ```
> 0
> 173.50089
> 152.36047
> 157.11974
> 168.0601
> 134.13326
> 82.78972
> 102.892944
> 80.54708
> 134.09038
> ...
> ```
> 

Am I missing some obvious way to interpret these files? If not, IMO these should either be written in an easy way for `np.loadtxt` to bring them back again, or left as attributes in a class.

The documentation also could use a lot of clarification and updating.

``` python
import MDAnalysis as mda
from MDAnalysis.align import helanal
u = mda.Universe(PDB)
helanal.helanal_trajectory(u, 'name CA')

```

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) '0.20.2-dev0'

lilyminium:
Fixes #2452 

 - Refactored HELANAL to use AnalysisBase

~Well, it produces the same numbers as the old ``helanal``.~ Changed the algorithm for calculating screw angles. A few things left (which the old module didn't implement):

* fit a sphere and classify the helix as linear or curved
> The local helix origins are reoriented
in *X-Y* plane and the reoriented points are used to fit a circle as well as a
line, by least squares method. Based on the relative goodness of line and
circle fit to local helix origins, the helix is *classified as being linear or
curved*. 
* classify helix as kinked or not
> A helix is classified as being *kinked*, if at least one local bending
angle in the middle of the helix is greater than 20 degrees.
* Finish #2609 
* Add tests for ref_axis != [0, 0, 1]

[Demo notebook here](https://lilyminium.github.io/UserGuide/examples/analysis/structure/helanal.html). It also has pictures of the attributes, I would appreciate feedback on if they make it clear what HELANAL actually calculates.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


lilyminium:
@WG150 @fiona-naughton I would love your thoughts on this as it is now and anything you would add!

package/MDAnalysis/analysis/helix_analysis.py
```diff
@@ -0,0 +1,538 @@
...
+    a: :class:`numpy.ndarray` of shape (N, M)
+
+    Returns
+    -------
+    :class:`numpy.ndarray` of shape (N,)
+    """
+    return pdot(a, a)**0.5
+
+
+def bound(arr, min_value, max_value):
```
richardjgowers: np.clip does this (probably faster?)

package/MDAnalysis/analysis/helix_analysis.py
```diff
@@ -0,0 +1,538 @@
...
+    #     v           origin
+    #   CA_i+3
+    #
+    # V: vectors
+    # bi: bisectors
+    # θ: local_twists
+    # origin: origins
+    # local_axes: in the plane of the screen. Orthogonal to the bisectors
+
+    vectors = positions[1:] - positions[:-1]  # (n_res-1, 3)
```
richardjgowers: maybe a silly question, what's the anticipated size of `positions` here?

package/MDAnalysis/analysis/helix_analysis.py
lilyminium: Thanks, numpy is the best.

package/MDAnalysis/analysis/helix_analysis.py
lilyminium: I think helices are usually < 50 residues long so ``(50, 3)`` for ideal use, but there are no error-raising checks on atom groups so you could just put your entire trajectory into it.

package/MDAnalysis/analysis/helix_analysis.py
```diff
@@ -0,0 +1,515 @@
...
+
+import warnings
+import numpy as np
+
+import MDAnalysis as mda
+from ..lib import util, mdamath
+from .base import AnalysisBase
+
+
+def pdot(a, b):
```
richardjgowers: There's some functions in `lib._cutil.pyx` that might do this sort of thing, so maybe either move those there or use them?

package/MDAnalysis/analysis/helix_analysis.py
```diff
@@ -0,0 +1,515 @@
...
+        self.global_axes = [self._zeros_per_frame(3) for n in n_res]
+        self.all_bends = [self._zeros_per_frame(n-3, n-3) for n in n_res]
+
+    def _single_frame(self):
+        _f = self._frame_index
+        for i, ag in enumerate(self.atomgroups):
+            results = helix_analysis(ag.positions)
+            for key, value in results.items():
+                attr = getattr(self, key)
+                attr[i][_f] = value
```
richardjgowers: haha this melted my brain for a minute but it's very clever

package/MDAnalysis/lib/util.py
```diff
@@ -1685,6 +1686,29 @@ def blocks_of(a, n, m):
...
+    -------
+    list of :class:`numpy.ndarray`
+
+    Examples
+    >>> arr = np.array([ 2,  3,  4,  7,  8,  9, 10, 11, 15, 16])
+    >>> group_consecutive_integers(arr)
+    [array([2, 3, 4]), array([ 7,  8,  9, 10, 11]), array([15, 16])]
+    """
+    mask = np.ones(len(arr)+1, dtype=bool)
+    mask[1:-1] = (arr[1:]-arr[:-1]) > 1
```
richardjgowers: So a weird numpy trick that might be better here....

```python
np.r_[True, (arr[1:] - arr[:-1]) > 1, True]
```

From what I can understand, `np.r_` is a weird hack for concatenating arrays

richardjgowers:
This looks like good code, but someone who actually knows the science ought to review it

package/MDAnalysis/analysis/helix_analysis.py
lilyminium: Thanks @richardjgowers -- this is because I'm lazy and kept changing what the properties were called to try to make them clearer. However if it impairs readability I'm happy to make it a longform

```
self.local_twists[i][self._frame_index] = results['local_twists']
...
```

package/MDAnalysis/analysis/helix_analysis.py
lilyminium: Is it worth cythonising them?

lilyminium: Is it worth cythonising them?

lilyminium:
Hm, ok, something to fix here. (x-axis is time)

Screenshot 2020-03-17 at 7 09 43 AM

richardjgowers:
In general your vector and angle calculations aren’t respecting PBC. Try
unwrapping and see if this fixes it?

lilyminium:
These are already preprocessed and unwrapped -- I think the 30 degrees is swinging around to 150 because I'm only calculating the angle from one direction. But yes I should put in PBC

lilyminium:
I'm going to have to come up with better tests than just comparing to old helanal's output...
image

Is this what `local_screw` is actually meant to be? I couldn't find this in the paper.
image

package/MDAnalysis/analysis/helix_analysis.py
```diff
@@ -0,0 +1,515 @@
...
+
+
+def local_screw(best_fit, ref_axis, rotation_vectors):
+    """
+    Get local screw angles.
+
+    Parameters
+    ----------
+    best_fit: :class:`numpy.ndarray` of shape (3,)
+        Vector of best fit. Tilts are calculated perpendicular to
```
fiona-naughton: Should be 'Screw angles are calculated...' rather than 'Tilts are ...'

package/MDAnalysis/analysis/helix_analysis.py
```diff
@@ -0,0 +1,515 @@
...
+    # replace low angles
+    low = screw_angles < quarter_pi
+    screw_angles[low] = np.where(alt_angles[low] < half_pi,
+                                 half_pi - alt_angles[low],
+                                 alt_angles[low] - half_pi)
+    # replace high angles
+    high = screw_angles > 3*quarter_pi
+    screw_angles[high] = np.where(alt_angles[high] < half_pi,
+                                  half_pi + alt_angles[high],
+                                  3*half_pi - alt_angles[high])
```
fiona-naughton: I may be missing something, but why is this necessary? As far as I can tell, this won't actually result in any changes to any of the angles.

package/MDAnalysis/analysis/helix_analysis.py
```diff
@@ -0,0 +1,515 @@
...
+
+    ortho = np.cross(ref_1, rotation_vectors)
+    cos_theta = np.matmul(best_fit, ortho.T)/(pnorm(ortho)*ref_norms[0])
```
fiona-naughton: Should be using the norm of `best_fit` rather than `ref_norms[0]`?

package/MDAnalysis/analysis/helix_analysis.py
```diff
@@ -0,0 +1,515 @@
...
+    #         CA_i+2 <----- CA_i+1
+    #         /    \       /   ^
+    #        /    r \     /     \
+    #     V /        \ θ /       \
+    #      /          \ /       CA_i
+    #     v           origin
+    #   CA_i+3
+    #
+    # V: vectors
+    # bi: bisectors
```
fiona-naughton: This might be a bit nit-picky - given the way they're defined below the 'biscectors' are only true biscectors if the CA_i -> CA_i+1 etc vectors have the same length in this plane. This isn't a bad assumption and as far as I can tell the radius calculation later on assumes this anyway, and I don't know what a better name would be; but might be worth clarifying here/in the notebook that they're not directly calculated as biscectors (assuming I'm doing my geometry right of course - do ignore me if I'm wrong!)

fiona-naughton:
Re: tests/testing data - those plots do look a little suspicious, though I'm not immediately sure what exactly would be the issue. What input were you using? Doesn't look like the right number of residues for the one the tests use.

Also, might be worth adding a test for when a different `ref_axis` is selected? There was a bug in the old helanal I never got around to properly reporting where it ignored `ref_axis` for some calculations and just used the z-axis instead - as far as I can tell this should be fine with the new code, though!

package/MDAnalysis/analysis/helix_analysis.py
```diff
@@ -0,0 +1,515 @@
...
+    :class:`numpy.ndarray` of shape (3,)
+        Vector of best fit.
+    """
+    centered = coordinates - coordinates.mean(axis=0)
+    Mt_M = np.matmul(centered.T, centered)
+    u, s, vh = np.linalg.linalg.svd(Mt_M)
+    vector = vh[0]
+
+    # does vector face first residue?
+    angle = mdamath.angle(coordinates[0], vector)
```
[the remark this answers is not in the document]
lilyminium: > I think this should be centered[0] rather than coordinates[0]

D'oh. Yes, you're right.

> Related thought: is there a reason to chose the direction of the helix vector to point towards the 1st residue? I don't remember if it's in any of the references, but at least to me it makes more sense intuitively for the helix axis to point towards the last residue (which would be the same direction as the local helix axes).

Each origin point would have its own local helix axis associated with it, so I don't understand if this matters.

I do agree with your intuition, it feels upside-down. However since it only affects whether the global tilt is obtuse or acute, I left it like that -- also because I peeked at how [these guys](https://github.com/davecao/biHelix/blob/00f6e3905b0171e604be2edc4c7a61096893b3ea/src/helanal.f90#L367-L374) did it and they also point at the first local origin point (that comment is inaccurate).

package/MDAnalysis/analysis/helix_analysis.py
lilyminium: Aha, yes, I have no idea what this is meant to achieve, partially because I don't know what a screw angle is meant to describe -- but the original author went to a lot of trouble so I kept it.

> As far as I can tell, this won't actually result in any changes to any of the angles.

I think it does, unless the `local_axis` is exactly in the plane created by `ref_1` and `ref_2` (which depend on `ref_axis` and `global_axes` only) -- this picture is how I understand the angles:
Screenshot 2020-03-20 at 8 55 50 PM

You can see anyway that the screw angles in the comment below are bouncing between 0.5pi-ish, -0.5pi-ish and -pi so my comment `# the effect of the below is to move angles into the (pi/4, 3pi/4 domain)` is clearly wrong.

package/MDAnalysis/analysis/helix_analysis.py
lilyminium: No, you're right -- they're not normalised here because otherwise the cylinder radius winds up being dependent only on the angle between the inaccurately named ``bisectors``, instead of including the length as well, which made sense to me. It isn't a variable that users can see without looking at the code anyway -- I could rename it to something like ``dvectors`` or ``in_between``, or even just add a comment clarifying that they're not real bisectors?

lilyminium:
Thanks for the review @fiona-naughton! You brought up many of the things I wondered about myself. The inputs are helices from my own trajectories (this refactor was borne out of necessity) but I can put some graphs together from the test data.

> Also, might be worth adding a test for when a different ref_axis is selected

Good idea, I think in general the module needs more tests than the old one had.

package/MDAnalysis/analysis/helix_analysis.py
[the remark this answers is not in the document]
fiona-naughton: Yeah, that's fair - no harm in keeping it as is! Either way, it might be worth clarifying in the docs that it's end-to-start in case anyone uses the `global_axes` output and the orientation matters

package/MDAnalysis/analysis/helix_analysis.py
fiona-naughton: Ah yes, I'd forgotten it wouldn't be exactly in the plane, sorry!

In that case, my assumption is that the screw angle is intended to approximate the cylindrical azimuth angle of each atom about the helix axis, with `ref_1` as 0, but it'll be slightly offset due to the rotation vectors not being quite in plane. Having this code in place would mean we're using the larger of the angles from `ref_1` or the perpendicular `ref_2` to calculate the approx azimuth, which should reduce the relative error we get due to that offset.

package/MDAnalysis/analysis/helix_analysis.py
fiona-naughton: Yeah, fair - I think bisectors probably does actually work better as a descriptive name to get across what they're being used for, so that's fine.

package/MDAnalysis/analysis/helix_analysis.py
lilyminium: But we could just calculate the cylindrical azimuth as an angle to the plane of `ref_1` and `global_axes`? Is the approximation a feature?

package/MDAnalysis/lib/util.py
tylerjereddy: Since there's already `np.ediff1d` I like using that:

```python
np.array_split(arr, np.where(np.ediff1d(arr) != 1)[0])
```

tylerjereddy: I bet there is even cleaner with some careful thought..

package/MDAnalysis/analysis/helix_analysis.py
lilyminium: I updated the function to what I think it's trying to achieve: the cylindrical azimuth angle to the plane of the global_axis and ref_axis.

package/MDAnalysis/lib/mdamath.py
```diff
@@ -92,7 +94,39 @@ def normal(vec1, vec2):
...
+    Parameters
+    ----------
+    a: :class:`numpy.ndarray` of shape (N, M)
+    b: :class:`numpy.ndarray` of shape (N, M)
+
+    Returns
+    -------
+    :class:`numpy.ndarray` of shape (N,)
+    """
+    return np.einsum('ij,ij->i', a, b)
```
orbeckst, line 115: +1 for einsum!

package/MDAnalysis/lib/mdamath.py
```diff
@@ -92,7 +94,39 @@ def normal(vec1, vec2):
...
+
+    Parameters
+    ----------
+    a: :class:`numpy.ndarray` of shape (N, M)
+
+    Returns
+    -------
+    :class:`numpy.ndarray` of shape (N,)
+    """
+    return pdot(a, a)**0.5
```
orbeckst, line 129: I was just curious if `np.sqrt` is any different
```
In [5]: x = np.random.random((5000,))

In [6]: %timeit x**0.5
11.5 µs ± 60.6 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [7]: %timeit np.sqrt(x)
11.5 µs ± 67.7 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```
but it gives the same time on my laptop.

package/MDAnalysis/lib/util.py
```diff
@@ -1689,6 +1690,25 @@ def blocks_of(a, n, m):
...
+    Returns
+    -------
+    list of :class:`numpy.ndarray`
+
+    Examples
+    >>> arr = np.array([ 2,  3,  4,  7,  8,  9, 10, 11, 15, 16])
+    >>> group_consecutive_integers(arr)
+    [array([2, 3, 4]), array([ 7,  8,  9, 10, 11]), array([15, 16])]
+    """
+    return np.split(arr, np.where(np.ediff1d(arr)-1)[0]+1)
```
orbeckst, line 1710: pep 8 spaces...

lilyminium: PEP8 is ambiguous on spaces around operators; I typically just go with what makes the most sense/readability to me. 

Screenshot 2020-06-06 at 5 56 52 AM

Is the below easier to read?

```python
return np.split(arr, np.where(np.ediff1d(arr) - 1)[0] + 1)
```

orbeckst: It's mildly confusing either way; maybe a middle ground:
```python
return np.split(arr, np.where(np.ediff1d(arr)-1)[0] + 1)
```

package/MDAnalysis/lib/mdamath.py
lilyminium: The performance differs for bigger numbers.

Screenshot 2020-06-06 at 6 40 52 AM

Had to quietly take that out from my "Intro to numpy and why it's faster than everything else" tutorial once.

orbeckst: I am easily swayed by data – thanks!

package/MDAnalysis/analysis/helix_analysis.py
```diff
@@ -339,75 +345,48 @@ class HELANAL(AnalysisBase):
         'local_screw_angles': (-2,),
     }
 
-    @classmethod
-    def from_SecondaryStructure(cls, secondary_structure, select='name CA',
```
lilyminium: The SecondaryStructure PR is nowhere near done, this can get added back in that one.

package/MDAnalysis/analysis/helix_analysis.py
```diff
@@ -339,75 +345,48 @@ class HELANAL(AnalysisBase):
...
-            ids, counts = np.unique(ag.resindices, return_counts=True)
-            if np.any(counts > 1):
-                dup = ', '.join(map(str, ids[counts > 1]))
-                warnings.warn('Your selection {} includes multiple atoms '
-                              'for residues with these resindices: {}.'
-                              'HELANAL is designed to work on one carbon-alpha '
-                              'per residue.'.format(s, dup))
+        atomgroups = [universe.select_atoms(s) for s in selections]
+        consecutive = []
+        # check that residues are consecutive and long enough sequence
```
lilyminium, line 356: Added checks and warnings that residue sequences are a) consecutive b) at least 9 residues/atoms long

package/MDAnalysis/lib/util.py
```diff
@@ -1704,10 +1705,10 @@ def group_consecutive_integers(arr):
 
     Examples
     >>> arr = np.array([ 2,  3,  4,  7,  8,  9, 10, 11, 15, 16])
-    >>> group_consecutive_integers(arr)
+    >>> group_same_or_consecutive_integers(arr)
     [array([2, 3, 4]), array([ 7,  8,  9, 10, 11]), array([15, 16])]
     """
-    return np.split(arr, np.where(np.ediff1d(arr)-1)[0]+1)
+    return np.split(arr, np.where(np.ediff1d(arr)-1 > 0)[0] + 1)
```
lilyminium, line 1711: Changed to accept same integers as well, in case the user really wants multiple atoms from the same residue.

lilyminium:
Update: added some more checks and tests. Putting the original to-dos on hold:
- fit a sphere and classify the helix as linear or curved
- classify helix as kinked or not

because the original helanal didn't have them anyway and I don't have much time to work on this in the near future.

package/MDAnalysis/analysis/helix_analysis.py
```diff
@@ -0,0 +1,474 @@
+# -*- Mode: python; tab-width: 4; indent-tabs-mode:nil; coding:utf-8 -*-
```
richardjgowers, line 1: rst stub for this?

package/CHANGELOG
```diff
@@ -97,6 +97,7 @@ Fixes
   * Contact Analysis class respects PBC (Issue #2368)
 
 Enhancements
+  * Refactored helanal in helix_analysis (Issue #2452, PR #2622)
```
richardjgowers, line 100: this seems a little understated, analysis.helanal is deprecated and analysis.helix_analysis is the new thing right?

package/MDAnalysis/analysis/helix_analysis.py
lilyminium: I've got one [here](https://github.com/MDAnalysis/mdanalysis/pull/2622/files#diff-0f3a2ec710c389155f8e7a2261d553b4); is it not showing up?

package/CHANGELOG
lilyminium: Yeah, I think I meant "into", is that clearer?

```suggestion
  * Refactored helanal into helix_analysis (Issue #2452, PR #2622)
```

package/MDAnalysis/analysis/helix_analysis.py
```diff
@@ -0,0 +1,474 @@
...
+    vector = vh[0]
+
+    # does vector face first local helix origin?
+    angle = mdamath.angle(centered[0], vector)
+    if angle > np.pi/2:
+        vector *= -1
+    return vector
+
+
+def local_screw_angles(global_axis, ref_axis, helix_directions):
```
fiona-naughton, line 106: The screw angles here are calculated in the range 0 to pi, but I think they should cover the full -pi to pi range?

package/MDAnalysis/analysis/helix_analysis.py
```diff
@@ -0,0 +1,474 @@
...
+    # guesstimate radius = half bisector magnitude
+    radii = np.where(cos_theta != 1, radii, (adjacent_mag**0.5)/2)
+    # height of local cylinder
+    heights = np.abs(mdamath.pdot(vectors[1:-1], local_axes))  # (n_res-3,)
+
+    local_helix_directions = (bisectors.T/bimags).T  # (n_res-2, 3)
+
+    # get origins by subtracting radius from atom i+1
+    origins = positions[1:-1][:]  # (n_res-2, 3)
+    origins[:-1] -= (radii*local_helix_directions[:-1].T).T
```
fiona-naughton, line 252: I think this needs to use `copy()` rather than `[:]` to stop the input `positions` array from being altered here

package/MDAnalysis/analysis/helix_analysis.py
```diff
@@ -0,0 +1,474 @@
...
+                dims, n_positions=n) for n in n_res]
+            setattr(self, key, empty)
+
+        self.global_axis = [self._zeros_per_frame((3,)) for n in n_res]
+        self.all_bends = [self._zeros_per_frame((n-3, n-3)) for n in n_res]
+
+    def _single_frame(self):
+        _f = self._frame_index
+        for i, ag in enumerate(self.atomgroups):
+            results = helix_analysis(ag.positions)
```
fiona-naughton, line 413: `self.ref_axis` should be passed on to `helix_analysis` here

package/MDAnalysis/analysis/helix_analysis.py
lilyminium: I guess I thought of it just as the angle to the plane of `ref_axis` and `global_axis`, which is usually [0, pi). I'm really awful at 3D visualisation so I could be missing an obvious answer to this, but which side of the plane is the "origin" i.e. which part of the helix should be negative vs positive pi?

richardjgowers:
@fiona-naughton I think this is taken care of, can you take a quick peek so this makes the release (or not)

package/MDAnalysis/analysis/helix_analysis.py
fiona-naughton: So the way I understand it is that the screw angle is measuring the location of each CA around the circle if you were looking down the helix axis - so you could use it to, say, select all the residues on a particular 'face' of the helix by defining an angle range. So positive/negative angles tell us whether it's clockwise or anticlockwise relative to the reference axis.

On the calculation of the screw angles - I don't seem to be getting the numbers I expect, with offsets in multiples of 90/flipped signs. I'm still gathering my thoughts and making sure what I'm expecting is actually right, but for what it's worth the following does seem to work so far (adding back ortho, which I believe is the projection of `ref_axis` on the helix cross-section plane, as the second reference; and changing how the angle is adjusted for quadrants).

    orth = np.cross(-plane, global_axis)  ### 

    # angles to plane normal
    refs = np.array([plane, orth]) ###
    norms = np.outer(mdamath.pnorm(refs), mdamath.pnorm(helix_directions))
    cos = np.matmul(refs, helix_directions.T)/norms
    to_plane, to_orth = np.arccos(np.clip(cos, -1, 1))  # (2, n_vec)

    # angles to plane in (-pi, pi]
    to_orth[to_plane > np.pi/2] *= -1

    return np.rad2deg(to_orth)

lilyminium: Thanks @fiona-naughton for looking at it, I remain a bit confused as to what this should be. I added some tests that involve circles as a kind of expected spec: a circle in the xy plane with the x-axis as global_axis and y-axis as ref_axis should have 0s as the angles, and a circle in the xz plane should be `array([   0,   12,   24,   36,   48,   60,   72,   84,   96,  108,  120,
        132,  144,  156,  168,  180, -168, -156, -144, -132, -120, -108,
        -96,  -84,  -72,  -60,  -48,  -36,  -24,  -12])`. I interpreted this function as calculating the angle to the plane of ref_axis and global_axis (i.e. the xy plane)

fiona-naughton: Ah, hadn't looked properly at the new tests, sorry. So having looked through everything again and having another thing, I'm going to gather my thoughts here, just so we can make sure we're on the same page - might get a bit long, so please bear with me!

If I'm following correctly, basically what I think the current code is doing is: **1.** Define an axis (`plane` in the code, but I'll call it `perp` here) perpendicular to both `ref` and `global`. **2.** Measure angles of position vectors to `perp` and `global`.  **3.** Work out quadrants from these angles to give an angle to `global` in the range -pi to pi. 
We get an angle *to* `global `- I guess this would represent how high/low (along the helix axis) an atom is relative to the center of it's helix window? However, using both angles/quadrants like this assumes that the position vectors are ~in the `global`-`perp` plane, which is the case in your second example. I think the first still works because it's a trivial case, elsewhere it'd get less accurate.


I think what it should do is: **1.** Define an axis (`perp`) perpendicular to both `ref` and `global`. **2.** Define a second axis (`orth` perpendicular to both `perp` and `global`). **3.** Measure angles of position vectors to `perp` and `orth`. **4.** Work out quadrants from these angles to give an angle to `orth` in the range -pi to pi.
Since `perp` and `orth` are both perpendicular to `global`, this gives an angle *about* `global` (with 0 in some way based on the direction of `ref`), so represents which side of the helix the atom is on. As above this assumes that the position vectors are in the `perp`-`orth` plane - i.e. perpendicular to `global`. But in general helanal already assumes this is ~the case, so here I think it's (more) ok.


(Related: I also used a circle in xy to test the code, but defined `global` along the z-axis so the position vectors were perpendicular to global as they would be for a 'proper' helix. I don't think it makes sense to test with `global` in the plane of the circle). 

I hope that all makes sense...!

**tl;dr** code currently calculated angle *to* helix axis, but the method for this makes an assumption; tests pass because they're cases where assumption holds. An angle *around* helix axis should require less assumption + more intuitive?

lilyminium: Ok, I've just drawn this terribly, just pretend ortho is perpendicular to `perp` and `GLOBAL`:

IMG_3282 JPG


> Work out quadrants from these angles to give an angle to `orth` in the range -pi to pi.

So it sounds like you think the screw angles should be the blue psi angle to the `ortho` vector, whereas I've been interpreting it as the green theta angle to the plane of REF and GLOBAL. Does that look about right?

> Ah, hadn't looked properly at the new tests, sorry. So having looked through everything again and having another thing, I'm going to gather my thoughts here, just so we can make sure we're on the same page - might get a bit long, so please bear with me!
> 
> If I'm following correctly, basically what I think the current code is doing is: **1.** Define an axis (`plane` in the code, but I'll call it `perp` here) perpendicular to both `ref` and `global`. **2.** Measure angles of position vectors to `perp` and `global`. **3.** Work out quadrants from these angles to give an angle to `global` in the range -pi to pi.
> We get an angle _to_ `global `- I guess this would represent how high/low (along the helix axis) an atom is relative to the center of it's helix window? However, using both angles/quadrants like this assumes that the position vectors are ~in the `global`-`perp` plane, which is the case in your second example. I think the first still works because it's a trivial case, elsewhere it'd get less accurate.

What I *think* I'm doing is figuring out the angle to `perp`, the normal of the ref-global plane, and also figuring out the angle to global as way to get quadrants. The final result after some arithmetic I *think* is the angle to the ref-global plane. I envisioned vectors rotating around the `global` axis being calculated to a static plane, I guess. 

> so we can make sure we're on the same page

I'm happy to be on either page. I don't follow that getting angles to the cross section of the axis (as defined by either `perp` or `ortho`) is the right thing here though, does that capture rotation around `global`?

> (Related: I also used a circle in xy to test the code, but defined global along the z-axis so the position vectors were perpendicular to global as they would be for a 'proper' helix. I don't think it makes sense to test with global in the plane of the circle).

I think I made the circle in the xz-plane, the global axis the x-axis, and the ref-axis the y-axis to test angles to the xy plane. The position vectors in a real helix would rarely be perpendicular because the helix goes up/down as it winds around.

> I hope that all makes sense...!

I think we're making progress? I won't change anything just yet until we settle just what the screw angles represent, though

lilyminium: Looking at it again I think the best angle would be from the red vector to the plane of `perp` and `global` (not on the picture)?

lilyminium: Ok, I did what was in the last comment and added a lot more comments to the tests to try to make it clear what I'm going for

fiona-naughton: > I think I made the circle in the xz-plane, the global axis the x-axis, and the ref-axis the y-axis to test angles to the xy plane. The position vectors in a real helix would rarely be perpendicular because the helix goes up/down as it winds around.

Using positions here was probably misleading - I mean the `local_helix_directions`, which are what is passed into the `local_screw_angles` calculation. They're calculated as the normalised `bisectors`, which should be roughly perpendicular to the helix axis (e.g. if I feed the test helix along the z axis `np.array([[1.,0,0], [0,1,1], [-1,0,2], [0,-1,3], [1,0,4], [0,1,5], [-1,0,6],[0,-1,7]])` into `helix_analysis`, the `local_helix_directions` are `[[0.  1.  0.],  [-1.  0.  0.], [ 0. -1.  0.], [ 1.  0.  0.], [ 0.  1.  0.], [-1.  0.  0.]]` - ie. in all in the xy plane. 

When trying with an actual helix it's not as perfect, but still close - e.g.

    u = mda.Universe(PDB_small)
    sel = u.select_atoms('name CA and resnum 161-170')
    pos = sel.atoms.positions
    data = hel.helix_analysis(pos)
    #  get angle between global and each helix direction
    print([np.rad2deg(np.arccos(np.dot(i, c['global_axis']))) for i in c['local_helix_directions']])

Gives `[89.38454, 90.11533, 85.983376, 93.158035, 89.046326, 89.97367, 89.75803, 91.40081]`. 

That being said - I suppose it's not unreasonable to test the extreme cases to check the code even if it might not reflect a 'realistic' situation. It might actually come up for quite bendy/kinked helicies, but other parts of helanal currently aren't great for those situations, either.

> So it sounds like you think the screw angles should be the blue psi angle to the ortho vector, whereas I've been interpreting it as the green theta angle to the plane of REF and GLOBAL. Does that look about right?

Yeah, that diagram looks right for what I think we were each considering, though I had been mistakenly assuming you meant the green angle to represent the angle to the `global` axis so my comments about inaccuracy don't matter - my bad there. 

Here's my own diagram - I'm sorry it got a bit busy, especially after adding the new angle-to-perp/global in purple, labeled alpha (though I accidentally drew it going from orth rather than xyz), but hopefully you can still sort of see what's going on.

IMG_20200610_222617

So I could probably have been clearer that what I think the 'real' screw angle should be is the pink angle, which is basically a projection of the angle between ref and the helix direction vector in the plane perpendicular to global. But I've been talking about the blue angle as an approximation to this, since this was what I interpreted the old code as doing - but we should be able to directly calculate it rather than using the approximate. 

It seems both of what we're doing approximate to the same thing when we assume the helix directions are perpendicular to global (or well, with a quadrant offset, but since where we set 0 is a bit arbitrary this isn't huge) - and that's what I'm seeing testing the current code, so that's something (I was getting the occasional weird thing - like angles above 360 - previously). 

So I guess where I am at the moment is that the code currently calculates approximately the same thing that the original helanal code did/what I was trying to argue, just that it may not be the most 'intuitive' method given what I think the screw angles is trying to represent (though that might just be for me personally); and that I probably should have been advocating for the pink angle all along rather than the blue angle.

lilyminium: Thanks so much for drawing it out and clearing things up @fiona-naughton! In my last commit we were approximating to the same thing when you were talking about the blue angle to ortho, but you're right, it makes way more sense to just directly calculate the pink psi angle of the projection onto the ortho-perp plane.

I've just pushed a commit that (I think):

- projects the input vectors onto the global axis
- subtracts the global_axis projection from the input vectors, giving the projection onto the plane
- calculates angles from the projection to `ortho`
- uses the angles to `perp` to assign quadrants

and tests that check that changing the direction of the input vectors, along the global_axis axis, do not result in different angles.

Thank you for your patience in sorting this out with me <3

fiona-naughton: That makes sense, and the it's looking good to me! I might do one last quick lookover but hopefully this is everything :)

(And thank you, too! Now, here's hoping we don't have to think about 3D vector geometry again any time soon... )

lilyminium: Thank you @fiona-naughton for all the work you've put into this, I really would have had no idea what to do here without you! I'm definitely looking forward to giving my lack of 3D spatial visualisation skills a break.

fiona-naughton:
Sorry for another delay - looks good to me! (though I see one of the travis jobs has failed, but I don't know enough to know what's gone wrong there...)

lilyminium:
Thank you @fiona-naughton!!! Thank you for getting through this 😁 😁 😁 Travis was failing on some double RDF docs, I've fixed that now.

orbeckst:
@fiona-naughton you can squash-merge this. Just rewrite the commit message using the information from the PR, include the "Fixes" line, and clean up the commit message lines that come from the individual commits. (E.g., leave out any "fixed pep8", "now finally got X working" and all other "PR internal stuff".) The guiding principle is simply that this single commit message should have all the relevant information for the whole thing.

orbeckst:
@lilyminium can you please 
- make another PR in which you remove old helanal, targeted to milestone 2.0.
- make a PR against master to add the deprecation notice to old helanal & CHANGELOG (and target to milestone 1.0.x)

Thank you!

Issue #2553, opened 2020-02-26
Parsers should try and provide an Elements attribute


xiki-tempula:
Related to #2521, I'm interested in having a method of finding the element of an atom, which is traditionally defined using `type` attribute. However, as we are now able to read more topology files, the type is used to store the `atom type` of a certain atom instead of the element of the atom.

```
>>> import MDAnalysis as mda
>>> from MDAnalysis.tests.datafiles import *
>>> u = mda.Universe(gro)
>>> u.atoms.types
array(['N', 'N', 'C', ..., 'O', 'H'], dtype=object)
```

```
>>> import MDAnalysis as mda
>>> from MDAnalysis.tests.datafiles import *
>>> u = mda.Universe(TPR)
>>> u.atoms.types
array(['opls_287', 'opls_290', 'opls_290', ..., 'opls_407', 'opls_407',
       'opls_407'], dtype=object)
```
I'm thinking if it is necessary to have a new attribute 'element' which is dedicated to storing the element of the atom. Or it is sufficient to guess element from the mass.

richardjgowers:
Elements already exists it’s just not commonly used. Trying to enforce more
Elements usage in parsers would be a nice addition. Then once it’s more
common (mandatory?) we can use it downstream more.

xiki-tempula:
@richardjgowers That's nice. I haven't seen anyone mentioning it in #2521 and I didn't get any elements back when I load gro file so I thought it didn't exist.

IAlibay:
Just for information here, `TOPParser` is an example of a reader that feeds elements based on the topology's atomic number record. I'm surprised that the TPR parser doesn't do the same since (at least based on a quick look at a `gmx dump` output) this information is available. This might be worth looking into if anyone is comfortable with the TPR file format.

xiki-tempula:
Reading binary is certainly more difficult compared with reading a txt file and a binary which changes its format dramatically every few months is even more annoying.

lilyminium:
@IAlibay it gets complicated for united-atom stuff where a particle could be a methyl group, although that' definitely not exclusive to TPR. You'd have to guess based on the (iirc) mass of around 15.035 since atom type names vary, and decide to represent it as a C or H or CH3 (which then veers into atom types again). Definitely agree that elements should be the default topology attribute in formats like XYZ.

IAlibay:
[...]

PR #2624, opened 2020-03-14
Added elements attributes to the PDBParser.py

AmeyaHarmalkar:
Fixes #

-Added the elements attributes to the PDBParser.py
-Non-physical atoms types will be denoted by an X
-Can make it more specific to discriminate between CG and non-physical by adding a check_CG_atom function in guesser.py 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/topology/PDBParser.py
+++ b/package/MDAnalysis/topology/PDBParser.py
@@ -72,6 +72,7 @@ from ..core.topologyattrs import (
     Bonds,
     ChainIDs,
     Atomtypes,
+    Elements,
     ICodes,
     Masses,
     Occupancies,
@@ -159,6 +160,7 @@ class PDBParser(TopologyReaderBase):
      - resids
      - resnames
      - segids
+     - elements
 
     Guesses the following Attributes:
      - masses
@@ -210,6 +212,7 @@ class PDBParser(TopologyReaderBase):
         resnames = []
 
         segids = []
+        elements = []
 
         self._wrapped_serials = False  # did serials go over 100k?
         last_wrapped_serial = 100000  # if serials wrap, start from here
@@ -241,6 +244,8 @@ class PDBParser(TopologyReaderBase):
                 altlocs.append(line[16:17].strip())
                 resnames.append(line[17:21].strip())
                 chainids.append(line[21:22].strip())
+                #Saving the elements type in a list
+                elements.append(line[76:78].strip())
 
                 # Resids are optional
                 try:
@@ -302,6 +307,22 @@ class PDBParser(TopologyReaderBase):
 
         masses = guess_masses(atomtypes)
         attrs.append(Masses(masses, guessed=True))
+
+        # Guessed attributes
+        # Need to pull elements from Atom names
+        # Similar to the check for atomtypes function
+        if not any(elements):
+            # Can further add a check_cg_atom(elements) function in guessers.py to check for CG atom.
+            elements = guess_atom_element(elements)
+            if elements == '':
+                warnings.warn("Element record found to be non-physical.")
+                #Nomenclature that X will denote any non-physical element.
+                elements = 'X'
+            attrs.append(Elements(elements, guessed=True))
+        else:
+            attrs.append(Elements(guessers.guess_types(elements),
+                                  guessed=True))
+
 
         # Residue level stuff from here
         resids = np.array(resids, dtype=np.int32)


AmeyaHarmalkar:
@richardjgowers I added the elements attribute to the PDBParser.py. I am wondering if you could review it for me?

package/MDAnalysis/topology/PDBParser.py
```diff
@@ -241,6 +244,8 @@ def _parseatoms(self):
                 altlocs.append(line[16:17].strip())
                 resnames.append(line[17:21].strip())
                 chainids.append(line[21:22].strip())
+                #Saving the elements type in a list
+                elements.append(line[76:78].strip())
```
lilyminium: This column isn't guaranteed in PDB files -- MDAnalysis itself doesn't write it out (fixed in #2609 because it was breaking DSSP). You need to check the line length to avoid IndexErrors.

package/MDAnalysis/topology/PDBParser.py
```diff
@@ -303,6 +308,22 @@ def _parseatoms(self):
         masses = guess_masses(atomtypes)
         attrs.append(Masses(masses, guessed=True))
 
+        # Guessed attributes
+        # Need to pull elements from Atom names
+        # Similar to the check for atomtypes function
+        if not any(elements):
+            # Can further add a check_cg_atom(elements) function in guessers.py to check for CG atom.
+            elements = guess_atom_element(elements)
```
lilyminium: `guess_atom_element` guesses from names -- this will return a list of empty strings. It also checks one value at a time. You could use `guess_types` here instead, which just passes it down to `guess_atom_element` anyway.

package/MDAnalysis/topology/PDBParser.py
```diff
@@ -303,6 +308,22 @@ def _parseatoms(self):
...
+            if elements == '':
```
lilyminium: Isn't `elements` a list?

package/MDAnalysis/topology/PDBParser.py
```diff
@@ -303,6 +308,22 @@ def _parseatoms(self):
...
+                warnings.warn("Element record found to be non-physical.")
+                #Nomenclature that X will denote any non-physical element.
+                elements = 'X'
```
lilyminium: From #2553 I think the consensus is that `element` should be a False-y value if it's non-physical, not a special name.

package/MDAnalysis/topology/PDBParser.py
```diff
@@ -303,6 +308,22 @@ def _parseatoms(self):
...
+            attrs.append(Elements(elements, guessed=True))
+        else:
+            attrs.append(Elements(guessers.guess_types(elements),
+                                  guessed=True))
```
lilyminium: Why are you guessing the elements again?

lilyminium:
@AmeyaHarmalkar I just noticed you're making changes on your ``develop`` branch. This makes it difficult to keep current with the main MDAnalysis ``develop`` branch, which gets code merged into it all the time. It also means that any further branches you checkout will contain these commits. I would create a new PR and delete this one, since there isn't much discussion yet.

1. Make a new branch from ``develop``:
```
git checkout -b my-new-branch
git push origin my-new-branch
```
2. Roll back changes on ``develop``:
```
git checkout develop
git reset --hard e416aa7
```
3. Open a new PR from my-new-branch and delete this one. Link to this PR number and note that it replaces this one.

The user guide [describes the typical workflow for contributing to MDAnalysis](https://www.mdanalysis.org/UserGuide/contributing_code.html#contributing-to-the-main-codebase).

package/MDAnalysis/topology/PDBParser.py
richardjgowers: Python lets you slice a string beyond it's end ie `'cat'[100:110] == ''`, so these will just be blank strings if the column isn't there

package/MDAnalysis/topology/PDBParser.py
richardjgowers: Yeah I think you'll want something like `any(not e for e in elements)` to detect missing elements

package/MDAnalysis/topology/PDBParser.py
IAlibay: Just adding to @lilyminium's comment. As per #2222, it would be good to throw a user warning when guessing here. That way users are aware of what MDA is doing being the scenes.

package/MDAnalysis/topology/PDBParser.py
AmeyaHarmalkar: I did it as a double check. I was think of providing a None/False in the else loop

package/MDAnalysis/topology/PDBParser.py
AmeyaHarmalkar: Sounds good. I thought of it as 3 conditionals. 1. In case it is an existing and valid element. 2. The element exists, but is not valid. .i.e. non-physical maybe because of the limitations of the tables or being a CG type? 3. It does not exist. 
Should I just have 2 instead?

package/MDAnalysis/topology/PDBParser.py
AmeyaHarmalkar: So, my understanding was that guess_atom_element will guess from a string and I am feeding it the elements string as an input. But I realize that checking the entire string would be the needed option. I can change it to guess_types. 
Also, I added a warning in the following if conditional. Should I shift that a loop above?

package/MDAnalysis/topology/PDBParser.py
AmeyaHarmalkar: Yes, I missed that. So, an alternative can be:
`if not any(elements):
     do (element parsing)
else if not any(atomnames):
     # Guess elements types from Atom names
     do (same task of parsing to get elements)
else:
     return False for element type`

Does this sound reasonable?

AmeyaHarmalkar:
Thanks @lilyminium and @richardjgowers for the review. I will implement your suggestions right away and create a different branch as well.

AmeyaHarmalkar:
@lilyminium I am closing this Pull request and submitting a different one like you suggested. Sorry for the inconvenience!

PR #2442, opened 2020-01-09
[WIP] PDB parser and writer use the "elements" topology attribute

jbarnoud:
Fixes #2422 #2423

Changes made in this Pull Request:
 - The PDB parser populates the `elements` topology attribute if the appropriate column is filled in the input. The content of `elements` is the same as `atomtypes`.
- The PDB reader uses the `elements` topology attribute if available.

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?
 - [X] Issue raised/referenced?

PR #2627, opened 2020-03-14
Providing Elements attribute to parsers  (#2553)

### LINKED ISSUE #2553: Parsers should try and provide an Elements attribute

AmeyaHarmalkar:
Fixes #2553

 - Added the elements attributes to the PDBParser.py
 - Added test cases with 
     1. PDB that is correct and provides the elements attributes correctly (pdb added to testsuites)
     2. A PDB file with missing element types that prints approriate warnings. (pdb added to testsuites)
 - Added the test check for guessed and expected attributes.
 - Added necessary warning statements for the user. 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


AmeyaHarmalkar:
@lilyminium @richardjgowers (sorry for so many pings, just bringing your attention to the alternative Pull Request for issue #2553)

IAlibay:
@AmeyaHarmalkar this PR seems to include changes for both the PDBParser and TPRParser.
Since you are aiming to fix the PDBParser here, your branch should ideally only have changes associated with that.

AmeyaHarmalkar:
@IAlibay I committed to this branch by mistake(I switched the branch and used a different PR for that). However, I used git reset --hard on this branch to reset it to the parent. Is there any other way I should have done it?

IAlibay:
@AmeyaHarmalkar that seems to have done it 👍

AmeyaHarmalkar:
I actually forced it on the head just now. Made a rookie mistake. Sorry about that!

package/MDAnalysis/topology/guessers.py
```diff
@@ -147,13 +147,14 @@ def guess_atom_element(atomname):
...
                 return name
-            if name[:-1] in tables.elements:
+            elif name[:-1] in tables.elements:
                 return name[:-1]
-            if name[1:] in tables.elements:
+            elif name[1:] in tables.elements:
                 return name[1:]
-            if len(name) <= 2:
-                return name[0]
-            name = name[:-1]  # probably element is on left not right
```
lilyminium: It's a bit silly but you need this for names like "AO5x" which is probably maybe an O (and would return 'AOx' here). But yeah good choice returning an empty string instead of an unphysical element.

package/MDAnalysis/topology/guessers.py
```diff
@@ -147,13 +147,14 @@ def guess_atom_element(atomname):
...
+            elif len(name) <= 2:
+                name = ''
+                return name
+            name = "00"  # probably element is on left not right
```
lilyminium: I don't understand why this line is here?

package/MDAnalysis/topology/guessers.py
```diff
@@ -147,13 +147,14 @@ def guess_atom_element(atomname):
...
 
         # if it's numbers
         return no_symbols
```
lilyminium: I don't remember why this is here, but from https://github.com/MDAnalysis/mdanalysis/issues/2553#issuecomment-591883179 I think this function should either return a valid element symbol, or an empty string.

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -244,3 +248,26 @@ def test_PDB_metals():
...
+
+
+def test_PDB_elements():
+    from MDAnalysis.topology import tables
+
+    u = mda.Universe(PDB_elements, format='PDB')
+    element_list = ['N', 'C', 'C', 'O', 'C', 'C', 'O', 'N', 'H', 'H', 'H',
+                    'H','H','H', 'H', 'H', 'CU', 'FE', 'C', 'MG']
+    assert len(u.atoms.elements) == len(element_list)
+    assert all(i == j for i,j in zip(element_list, u.atoms.elements))
```
lilyminium: You can use ``assert_equal`` for this kind of thing. The user guide has [some pointers](https://www.mdanalysis.org/UserGuide/testing.html#general-conventions) on writing tests.

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -244,3 +248,26 @@ def test_PDB_metals():
...
     assert u.atoms[2].mass == pytest.approx(tables.masses["CA"])
     assert u.atoms[3].mass == pytest.approx(tables.masses["MG"])
+
+
+def test_PDB_elements():
+    from MDAnalysis.topology import tables
+
+    u = mda.Universe(PDB_elements, format='PDB')
+    element_list = ['N', 'C', 'C', 'O', 'C', 'C', 'O', 'N', 'H', 'H', 'H',
+                    'H','H','H', 'H', 'H', 'CU', 'FE', 'C', 'MG']
```
lilyminium: Should the elements all be upper case like that? Genuine question.

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -244,3 +248,26 @@ def test_PDB_metals():
     assert u.atoms[1].mass == pytest.approx(tables.masses["FE"])
     assert u.atoms[2].mass == pytest.approx(tables.masses["CA"])
     assert u.atoms[3].mass == pytest.approx(tables.masses["MG"])
+
+
+def test_PDB_elements():
+    from MDAnalysis.topology import tables
+
+    u = mda.Universe(PDB_elements, format='PDB')
```
lilyminium: Do we really not have a PDB file on hand with elements? (I haven't looked.)

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -244,3 +248,26 @@ def test_PDB_metals():
...
+    assert all(i == j for i,j in zip(element_list, u.atoms.elements))
+
+
+
+def test_PDB_elements_warnings():
+    from MDAnalysis.topology import tables
+
+    u = mda.Universe(PDB_wrong_ele, format='PDB')
+    with pytest.warns(UserWarning) as record:
+        warnings.warn("Element record found to be non-physical. Guessing element from atom name: CGA", UserWarning)
```
lilyminium: That's not quite how pytest.warns works. You probably want this:
```python
with pytest.warns(UserWarning) as record:
    u = mda.Universe(PDB_wrong_ele)
```

In this case you yourself raise the warning you are looking for.

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -244,3 +248,26 @@ def test_PDB_metals():
...
+    u = mda.Universe(PDB_elements, format='PDB')
+    element_list = ['N', 'C', 'C', 'O', 'C', 'C', 'O', 'N', 'H', 'H', 'H',
+                    'H','H','H', 'H', 'H', 'CU', 'FE', 'C', 'MG']
+    assert len(u.atoms.elements) == len(element_list)
+    assert all(i == j for i,j in zip(element_list, u.atoms.elements))
+
+
+
+def test_PDB_elements_warnings():
+    from MDAnalysis.topology import tables
```
lilyminium: Why are you importing tables?

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -244,3 +248,26 @@ def test_PDB_metals():
...
+
+    u = mda.Universe(PDB_wrong_ele, format='PDB')
```
lilyminium: Again, do we really not have a PDB file with missing elements? Surely one of your new files is redundant. Generally we try not to add more files to the test suite if an existing one suffices, to keep it as light as possible.

richardjgowers: using `StringIO` is a way of having a throwaway PDB file inlined in the code, I think there's examples in either PDB or GRO...

testsuite/MDAnalysisTests/topology/test_pdb.py
AmeyaHarmalkar: Yeah, there was an initial suggestion in the code to compare everything in Upper case. I was unsure whether it was for some alternative functionality. So, I preferred to avoid the case change.

testsuite/MDAnalysisTests/topology/test_pdb.py
AmeyaHarmalkar: We do have one with elements (metals.pdb for eg but it contains all hetatoms), but I took @richardjgowers opinion and created an alternative one, which contains an amino acid and some arbitrary metals. I can redirect it to some other pdb if you think that's a better alternative?

testsuite/MDAnalysisTests/topology/test_pdb.py
AmeyaHarmalkar: Oh, my bad. Will correct this

testsuite/MDAnalysisTests/topology/test_pdb.py
AmeyaHarmalkar: That actually sounds like a better option, I will use that instead

package/MDAnalysis/topology/guessers.py
AmeyaHarmalkar: That makes sense. I was actually wondering why the loop was return that way. What would you suggest I do then? Returning empty string instead of unphysical element would be one of the cases, and another would be return O in cases such as AO5x?

package/MDAnalysis/topology/guessers.py
AmeyaHarmalkar: I kept '00' to indicate a missing element. From the discussion, the third case was that if we detect a missing element. Should I return a None?

testsuite/MDAnalysisTests/topology/test_pdb.py
AmeyaHarmalkar: The tables have the location for the PDB file with missing elements. But I would use the 'String IO' option instead as suggested in the next comment. That should resolve this

package/MDAnalysis/topology/PDBParser.py
```diff
@@ -303,6 +308,23 @@ def _parseatoms(self):
         masses = guess_masses(atomtypes)
         attrs.append(Masses(masses, guessed=True))
 
+        # Guessed attributes
+        # Need to pull elements from Atom names
+        # Similar to the check for atomtypes function
+        if any(elements):
```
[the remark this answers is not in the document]
AmeyaHarmalkar: The logic was if any of the element identifier exists, then get element information, else guess based on atom names, and report individual missing elements. However, if the entire row is missing, then the element identifier does not exist, and hence, we would be guessing completely based on the atom names. The subset of missing element identifiers can vary, hence I wanted to prompt a warning only in the cases when it was missing, hence I did a if any instead of a if not all. Because only specific elements are being guessed instead of the entire atom set in the PDB. In your opinion should I change this logic?

orbeckst:
@AmeyaHarmalkar – is there an existing issue that this PR and PR #2628  are addressing? If so, please reference it. 

If there isn't an issue yet then it would be great if you raised one in the issue tracker and then reference your new issue. In this way we have a much better record of problems and the problem definition is neatly separated from the solution.

jbarnoud:
I don't have the time to work on this but I just wanted to point to #2442 where I started solving the issue. The discussion may be of interest.

AmeyaHarmalkar:
@lilyminium I made the changes you suggested. It would be really helpful if you could have a look and let me know if I need to make any further changes. 
Thanks!

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -244,3 +247,102 @@ def test_PDB_metals():
...
+TER    1617
+"""
+
+def test_PDB_elements():
+    """
+    The elements attribute is equipped to obtain the element symbols,
+    if the PDB file contains the element symbol
+    """
+    u = mda.Universe(StringIO(PDB_elements), format='PDB')
+    element_list = np.array(['N', 'C', 'C', 'O', 'C', 'C', 'O', 'N', 'H', 'H', 'H', 'H', 'H', 'H', 'H', 'H', 'CU', 'FE', 'MG', 'S', 'O', 'C', 'C', 'S', 'O', 'C', 'C'], dtype=object)
```
lilyminium: Please follow PEP8 and split this list across multiple lines. Also, it is my strong opinion is that the element symbols should follow the capitalisation of the periodic table. I think the suggestion to compare elements in upper case was to catch stuff like "h" and match it to "H".

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -244,3 +247,102 @@ def test_PDB_metals():
...
+    assert np.all(u.atoms.elements == element_list)
```
lilyminium: Please use ``assert_equal`` for this. The [user guide](https://www.mdanalysis.org/UserGuide/testing.html#general-conventions) has more tips on writing tests.

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -244,3 +247,102 @@ def test_PDB_metals():
...
+    """
+    The test checks whether it asserts the appropriate warning on
+    encountering missing elements column.
+    """
+
+    with pytest.warns(UserWarning) as record:
+        u = mda.Universe(StringIO(PDB_missing_ele), format='PDB')
+
+    assert len(record) == 1
+    assert record[0].message.args[0] == "Element information missing. Guessing elements from atom names : ['N', 'CA', 'C', 'O', 'CU', 'FE', 'Mg']"
```
lilyminium: Again, please break the message over multiple lines to follow PEP8. You can also test to check that the guessed elements are what you expect here.

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -244,3 +247,102 @@ def test_PDB_metals():
...
+def test_wrong_elements_warnings():
+    """
+    The test checks whether it fills in the wrong elements appropriately
+    with the atom names where element symbols are missing.
+    """
+    with pytest.warns(UserWarning) as record:
+        u = mda.Universe(StringIO(PDB_wrong_ele), format='PDB')
+    
+    assert len(record) == 2
+    assert record[1].message.args[0] == "Element record found to be non-physical. Guessing element from atom name: ['X', 'FE']"
```
lilyminium: As above, please follow PEP8. What's the first warning that you're expecting here?

> "Element record found to be non-physical.

I don't see any non-physical elements in `PDB_wrong_ele`, just missing ones. You may need to reconsider how you guess the elements.

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -244,3 +247,102 @@ def test_PDB_metals():
...
+
+    # Checking if missing atoms are filled with an empty string.
+
+    element_list = np.array(['N', 'C', 'C', 'O', '', 'CU', 'FE', 'MG'], dtype=object)
+    assert np.all(u.atoms.elements == element_list)
```
lilyminium: As above, please use `assert_equal`.

package/MDAnalysis/topology/PDBParser.py
```diff
@@ -303,6 +308,30 @@ def _parseatoms(self):
...
+            attrs.append(Elements(elements, guessed=True))
+        else:
+            elements = guess_types(elements)
+            uni_elements, indices = np.unique(elements, return_inverse=True)
+
+            if any(np.isin(uni_elements, '')):
+                indexlist = [ j for j in range(len(indices)) if indices[j] == 0 ]
+                # This would be 0 always as guess_types would return a set
+                # of alphabets and empty string, and empty string will be
+                # sorted to be a index 0.
```
lilyminium: This is a bit convoluted. You're right that the empty string will always be sorted to be first, so you could just check `if uni_elements[0] == ''`. However what I had in mind was more checking the unique atom names. So you can get a masking array of ``missing = elements==''``, use that to choose the names you want to look at with ``names_to_guess_from = names[missing]``, and apply `np.unique` to *that* array. (note: you may need to convert `elements` and `names` into numpy arrays first.)

IAlibay:
> Thanks for these changes @AmeyaHarmalkar -- right now I'm not sure there's a consensus on what to do for intermittently missing or invalid elements.

With the idea of getting this done on time for the GSOC deadline, maybe this PR should just aim to add the elements if they exist in the file, and if they don't warn the user and not populate elements. If needed, further changes could then be tackled in a separate PR (most likely based on #2630). Thoughts on this @lilyminium?

lilyminium:
@IAlibay sounds good. Just to be clear, you're suggesting that Elements only be added if all of the symbols are present and valid, and no guessing happen based on names?

Alternatively, @AmeyaHarmalkar, you could focus on #2628 since the TPR format is more straightforward.

AmeyaHarmalkar:
Sounds great! @lilyminium Should I finish this PR with the suggestions and changes you requested, merge it and then work on TPR next?

zemanj:
>@IAlibay sounds good. Just to be clear, you're suggesting that Elements only be added if all of the symbols are present and valid, and no guessing happen based on names?

Very conservative, won't break anything - seems legit!

> I think @jbarnoud and @richardjgowers had different opinions on that in #2442 if they want to chime in. And @zemanj has opinions on guessing in general -- what do you think about trying to validate read information?

I never though about that before. I think that (optional?) data validation (comparing with guessed data?) would be a handy feature! For this PR, though, I'd simply go with what @IAlibay said:

* If all atoms in the PDB file have elements defined AND all these elements are known elements -> use them
* Otherwise: discard all element info, throw warning, continue

IAlibay:
> @IAlibay sounds good. Just to be clear, you're suggesting that Elements only be added if all of the symbols are present and valid, and no guessing happen based on names?

Yeah exactly. It's possible that I misread your comment https://github.com/MDAnalysis/mdanalysis/pull/2627#pullrequestreview-378912385, but I thought this might be the fastest solution here.

I'm not super fond of my own suggestion, since it technically amounts to asking @AmeyaHarmalkar to throw away code they wrote. However, considering there's 9 days left on the GSOC submission deadline, and at least to me adding elements on pdbv3 complient files is already a good PR, I thought it might be reasonable to give @AmeyaHarmalkar an option to get this done without having to wait for consensus.

> Alternatively, @AmeyaHarmalkar, you could focus on #2628 since the TPR format is more straightforward.

So I think we might face the same issue about intermittently missing or invalid elements in this one too. I was inspecting a few tprs yesterday and realised that in my protein-ligand complexes, my ligands all had -1 atomic numbers whilst my proteins were fine (some parmed conversion thing it seems) 😅

lilyminium:
@AmeyaHarmalkar yes I think it would be a good idea to finish off this PR following @IAlibay's and @zemanj's suggestions for element reading for now, just to get it in before the GSOC deadline :-)

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -233,6 +235,7 @@ def test_PDB_hex():
     assert u.atoms[3].id == 100002
     assert u.atoms[4].id == 100003
 
+@pytest.mark.filterwarnings("ignore::UserWarning")
```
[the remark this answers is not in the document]
AmeyaHarmalkar: Sorry, I failed to reply to this. The no_symbols is essential in case there is an element which not has been detected (.i.e. is essentially missing). It basically returns an empty string in case the element provided is non-physical or missing, which was the consensus. Does it sound reasonable?

testsuite/MDAnalysisTests/topology/test_pdb.py
AmeyaHarmalkar: ATOM 5 was named X, that was a non-physical element which is returned as a string.

testsuite/MDAnalysisTests/topology/test_pdb.py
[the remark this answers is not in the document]
lilyminium: I'm not sure you replied to the right comment here. If you're talking about `no_symbols` in `guess_atom_element`, `no_symbols` is just the name without any *, +, or - symbols in it. So `no_symbols` for 'AO*5' would be 'AO5', rather than the empty string.

package/MDAnalysis/topology/PDBParser.py
```diff
@@ -303,6 +308,28 @@ def _parseatoms(self):
         masses = guess_masses(atomtypes)
         attrs.append(Masses(masses, guessed=True))
 
+        # Guessed attributes
+        # Need to pull elements from Atom names
+        # Similar to the check for atomtypes function
+        if not any(elements):
+            warnings.warn("Element information absent.")
+            attrs.append(Elements(elements, guessed=True))
```
lilyminium: > guessed=True

But you are not guessing these. In addition, I interpreted @IAlibay's suggestion to just not add the `Elements` if they are not all present and valid.

package/MDAnalysis/topology/PDBParser.py
```diff
@@ -303,6 +308,28 @@ def _parseatoms(self):
...
+            missing = np.isin(elements,'')
+            uni_elements, indices = np.unique(elements, return_inverse=True)
+            # indexlist contains the index of those elements
+            # whose atom record is missing from the column or is
+            # non-physical.
+            # This can be further used in guessing from atomnames
+            # if needed.
+            indexlist = [ j for j in range(len(indices)) if missing[j] == True]
+            warnings.warn("Element record found to be either non-physical or missing for some elements.")
+            attrs.append(Elements(elements, guessed=True))
```
lilyminium: Same as above, I would simply check that all the read elements are valid, and only add them if so.

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -244,3 +247,108 @@ def test_PDB_metals():
...
+def test_PDB_elements():
+    """
+    The elements attribute is equipped to obtain the element symbols,
+    if the PDB file contains the element symbol
+    """
+    u = mda.Universe(StringIO(PDB_elements), format='PDB')
+    element_list = np.array([
+        'N', 'C', 'C', 'O', 'C', 'C', 'O', 'N', 'H',
+        'H', 'H', 'H', 'H', 'H', 'H', 'H', 'CU', 'FE',
+        'MG', 'S', 'O', 'C', 'C', 'S', 'O', 'C', 'C'
```
lilyminium: Again, I think the elements should follow the capitalisation of the periodic table: 'Cu', 'Fe', etc.

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -244,3 +247,108 @@ def test_PDB_metals():
...
+        u = mda.Universe(StringIO(PDB_wrong_ele), format='PDB')
+
+    assert len(record) == 2
+    assert record[1].message.args[0] == "Element record found to "\
+        "be either non-physical or missing for some elements."
+
+    # Checking if missing atoms are filled with an empty string.
+
+    element_list = np.array(['N', 'C', 'C', 'O', '', 'CU', '', 'MG'
+        ], dtype=object)
```
lilyminium: As above, I think we are simply not adding `Elements` if even a single one is missing.

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -72,7 +74,7 @@ class PDBBase(ParserBase):
     expected_attrs = ['ids', 'names', 'record_types', 'resids',
                       'resnames', 'altLocs', 'icodes', 'occupancies',
                       'bonds', 'tempfactors', 'chainIDs']
-    guessed_attrs = ['types', 'masses']
+    guessed_attrs = ['types', 'masses','elements']
```
lilyminium: You may have to remove `elements` from `guessed_attrs` if they are not all present in the files for the tests that subclass `PDBBase`.

lilyminium:
@IAlibay would you be able help out with this PR, as the GSOC deadline is coming up quite soon? @AmeyaHarmalkar is really close to being done and I think you know this issue very well. I won't have as much time to spend on MDAnalysis in the next few days due to dealing with departmental shutdown. Of course I'm happy to keep up with this if you have too much on your plate already (*global* pandemic).

IAlibay:
@AmeyaHarmalkar if you can complete the changes requested by @lilyminium we can try to get this done before the GSOC submission deadline.

AmeyaHarmalkar:
@IAlibay @lilyminium sorry for not getting back on this sooner. With the universities shutting down, it has been difficult, but I am on it now, and would try to get this done in a few hours.
Just to be clear:
1. We are geting elements only if the elements column is completely filled. If any of the elements is missing then we are not getting it? I was unclear on this because I thought we want to get whatever data is available even if a single one is present, but from @lilyminium's comment above I find it is otherwise. Please correct me if I am wrong. I though that we would not want to miss on any available information (even if we are not guessing).
2. About the atom names, they are in that order as the keys in the dictionary I believe. And I would need to redo the dictionary keys which I would do now.

Please let me know if these are the right interpretations that I got?

IAlibay:
> @IAlibay @lilyminium sorry for not getting back on this sooner. With the universities shutting down, it has been difficult.

Completely understandable and no worries, the most important thing is that people stay safe. Unfortunately GSOC hasn't decided to extend the submission deadline, but I'm sure we can get this done before the deadline.

> 1. We are getting elements only if the elements column is completely filled. If any of the elements is missing then we are not getting it? I was unclear on this because I thought we want to get whatever data is available even if a single one is present, but from @lilyminium's comment above I find it is otherwise. Please correct me if I am wrong. I though that we would not want to miss on any available information (even if we are not guessing).

So the concensus here was that in order to accelerate this PR (ensuring that this gets done before the GSOC deadline), it would be best to just fill the elements if all of them are available. It's a conservative approach, but at least we can temporarily avoid some of the discussion on what to do when we only have partial information available. My recommendation would then be to open up a second PR to deal specifically with partial information.

> 2. About the atom names, they are in that order as the keys in the dictionary I believe. And I would need to redo the dictionary keys which I would do now.

Sorry, I'm coming into this PR a little bit oblivious of some of the finer details. Could you elaborate as to which dictionary are you referring to?

AmeyaHarmalkar:
> Sorry, I'm coming into this PR a little bit oblivious of some of the finer details. Could you elaborate as to which dictionary are you referring to?

The guess_atom_elements() in guesser.py has a try except which uses the atomelements dictionary in `MDAnalysis.topology.tables`. Alternatively, elements uses a list and I can make that list a dictionary so as to compare in upper and print the expected information just like in the periodic table.

AmeyaHarmalkar:
@IAlibay just letting you know that I implemented the requested changes. Do let me know if I need to change anything. Thanks!

package/MDAnalysis/topology/tables.py
```diff
@@ -89,75 +89,110 @@ def kv2dict(s, convertor=str):
 # atomname   element
 # --------- ------------------
 
+# Aluminium
```
IAlibay: I'm not sure here, there have been attempts at updating this table in the past and there have been plenty of arguments for/against. Possibly @orbeckst is the best person to say if this is a good idea / won't break anything else.

One possible alternative here is to add an additional dictionary entry to MDAnalysis.topology._elements.py which just reverses `NUMBER_TO_ELEMENTS`. This would then provide a complete match to the entire periodic table rather than only accepting a subset of it (if anything Cobalt is a frequent ion in the PDB and as far as I can see it isn't listed here).

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -244,3 +247,110 @@ def test_PDB_metals():
...
+    with pytest.warns(UserWarning) as record:
+        u = mda.Universe(StringIO(PDB_wrong_ele), format='PDB')
+
+    assert len(record) == 2
+    assert record[1].message.args[0] == "Element information absent "\
+        "or inadequate."
+
+    # Checking if missing atoms are filled with an empty string.
+
+    element_list = np.array(['', '', '', '', '', '', '', ''
```
IAlibay: As per @lilyminium's comment, here we would be checking that elements isn't a attribute of `u.atoms`.

package/MDAnalysis/topology/tables.py
AmeyaHarmalkar: Sure, that sounds like a better alternative. I was a bit concerned about making changes in this table.Should I go ahead and 1. Change this dictionary back to original, and create an additional dictionary entry in MDAnalysis.topology._elements.py to reverse it like you said?

package/MDAnalysis/topology/PDBParser.py
```diff
@@ -303,6 +308,21 @@ def _parseatoms(self):
...
+        # Need to pull elements from Atom names
+        # Need to update dependent on the treatment of non-physical
+        # elements (such as the CG atoms) or missing elements as
+        # per consensus in the community
+        if all(elements):
+            elements = guess_types(elements)
+            attrs.append(Elements(elements))
+        else:
+            warnings.warn("Element information absent or inadequate.")
+            miss_elements = [ '' for i in range(len(elements))]
```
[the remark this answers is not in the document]
AmeyaHarmalkar: Okay, so just give a warning and exit the loop? This would lead the `u.atoms.element ` call to fail, so I should change my test case for that accordingly then

package/MDAnalysis/topology/tables.py
IAlibay: Might be best to see what any of the others have to say about this before you make any further changes to this table. There's a good chance there's a very simple solution here that I'm just not seeing.

testsuite/MDAnalysisTests/topology/test_pdb.py
AmeyaHarmalkar: Yes right. Sorry about that.

package/MDAnalysis/topology/tables.py
AmeyaHarmalkar: Oh okay. Definitely. I would just update the remaining changes in the attributes for now then...

package/MDAnalysis/topology/PDBParser.py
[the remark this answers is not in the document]
IAlibay: Yeah, at least to me, that would be the easiest option. You probably would want to document this behaviour in the class docstring though.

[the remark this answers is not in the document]
AmeyaHarmalkar: Sure. Also, the code fragment for the alternative of treating the non-physical elements I wrote is available in some of the earlier commits. It might be useful to save time while tackling the issue in the near future. Should I comment the code and keep it in the file, or should I just remove it (like in this commit)? Just asking to ensure that whenever this issue pops up, it would be resolved quickly

package/MDAnalysis/topology/tables.py
lilyminium: Just popping in to point out the `SYMB2Z` and `Z2SYMB` dictionaries in `tables.py`, which feature more modern element names.

Also, yeah, this is a really good example of how hard it is to guess anything with names. `elements` is so short because of MD systems where an atom called '1he2' is more probably hydrogen than helium, and so on. This is probably better discussed in #2630.

Anyway you're just validating element symbols so `SYMB2Z` should do.

package/MDAnalysis/topology/PDBParser.py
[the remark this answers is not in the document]
lilyminium: You might get some trouble with the tests with the expected attributes, as IIRC the base testing class actually checks to see that every `expected_attr` is available, and that attributes not in `expected_attrs` or `guessed_attrs` are not there (you'll probably want to check this). So you might have to manually modify the attribute list for the files where you are/n't expecting elements.

package/MDAnalysis/topology/tables.py
IAlibay: I knew I was probably missing something obvious 😅 Thanks @lilyminium, although, if I'm getting this right, we have a whole extra file that just duplicates `Z2SYMB` (except that `NUMBER_TO_ELEMENT` contains pre-2016 IUPAC names)?

lilyminium: I suppose so! Although more like `Z2SYMB` duplicates `NUMBER_TO_ELEMENT` due to me not realising the latter existed until you mentioned it just now. `grep` says it's only used for `TOPParser`.

AmeyaHarmalkar: Just to ensure that I am on the same page before I add it to the guess_atom_element option, I think the discussion above primarily implies to do something as follows:

```
if name in tables.atomelements:
            return Z2SYMB[SYMB2Z[tables.atomelements[name].lower().capitalize()]]
elif name in tables.elements:
                return Z2SYMB[SYMB2Z[name.lower().capitalize()]]

```

package/MDAnalysis/topology/PDBParser.py
```diff
@@ -303,6 +313,24 @@ def _parseatoms(self):
         masses = guess_masses(atomtypes)
         attrs.append(Masses(masses, guessed=True))
 
+
+        # Need to pull elements from Atom names
+        substitute_elements = {v.upper():v for k,v in Z2SYMB.items()}
+        if all(elements):
+            elements = guess_types(elements)
```
IAlibay: So as far as I can tell, this is still going to cause the same issue of rejecting most valid elements.
The problem here is that passing the elements through to `guess_types` essentially will always cause MDA to make a guess of some form or another (in this case based on FF atom names).

Rather than relying on `guess_types` and attempting to fix its behaviour for this use case (which might have an impact on user code downstream), I would propose here just purely string matching with `Z2SYMB`.

Something like this might work:

```python
periodic_table = set(SYMB2Z) # set over list might not actually be faster here
elements = [ elem.capitalize() for elem in element ]

if all(elem in periodic_table for elem in elements):
    attrs.append(Elements(elements)
else:
    warnings.warn("some warning message here")
```

package/MDAnalysis/topology/guessers.py
```diff
@@ -36,6 +36,7 @@
 
 from ..lib import distances
 from . import tables
+from .tables import Z2SYMB
```
IAlibay: As per my above comment, it would be simpler to just leave `guessers.py` as it was, I have a feeling that otherwise we'll unecessarily affect downstream user code.

IAlibay: That being said the changes made here are probably quite useful, could you open a separate issue/PR set?

package/MDAnalysis/topology/PDBParser.py
AmeyaHarmalkar: @IAlibay I thought of doing that earlier, but there was another problem in that case. For example, say the elements column has some elements named as Ox, CA+2, S1, POT, N+, etc. In that case, checking based on just string matching would result in wrong elements being added. Hence, I decided to use guess_types first and then do a string matching.
What do you think?

package/MDAnalysis/topology/guessers.py
AmeyaHarmalkar: My bad. I completely forgot to remove that line. 
Also, do you want me to do a changes in a new branch, commit it and then make a PR there?

package/MDAnalysis/topology/PDBParser.py
IAlibay: So personally I'd consider all those cases to be "non-normal" elements. At least within the scope of this PR, we'd want to just drop them here (maybe we can somehow get a list of all unique "non-normal" elements and feed it to the warning in the `else` case.

In the long term, my view is that guessing is always best applied only to the cases where a "non-normal" element is found, rather than by default guessing for every entry.

Note: here POT and CA+2 would fail because you are reading a two character portion of `line`.

package/MDAnalysis/topology/guessers.py
IAlibay: So I'm not 100% sure if there were any changes to `guessers.py` that was beyond just purely the scope of adding elements to the PDB parser. If so in the first instance I would open a new issue detailing what changes are being proposed and why. Then in a separate branch introduce those changes and open a PR.

package/MDAnalysis/topology/PDBParser.py
AmeyaHarmalkar: Gotcha. I will rewrite this part according to what you said. 
Also, I can add the warning message as requested.
Should I increase the read limits from [76:78] to [76:80]?

package/MDAnalysis/topology/guessers.py
AmeyaHarmalkar: Sure, that makes sense. I also don't quite understand why TravisCI builds might be failing? Is there something I can do to improve on that?

IAlibay: If you look at the Travis output: https://travis-ci.com/github/MDAnalysis/mdanalysis/jobs/303605080, it looks like it's failing because of the changes that were made to `guessers.py`. Since we'll probably be reverting those changes here, these tests should stop failing. However, it would be worth noting these failures if you plan to introducing any changes to `guessers.py` in the future.

AmeyaHarmalkar:
@IAlibay does this look good? I close the other PR. Just wanted you to look through this one once if possible. Thanks!

package/MDAnalysis/topology/PDBParser.py
```diff
@@ -159,6 +166,7 @@ class PDBParser(TopologyReaderBase):
      - resids
      - resnames
      - segids
+     - elements
```
IAlibay: In addition to this, could you also add a `.. versionchanged:: 1.0.0` section in the docstring? Saying that valid elements are now read.

package/MDAnalysis/topology/PDBParser.py
```diff
@@ -35,6 +35,10 @@
 :mod:`~MDAnalysis.topology.ExtendedPDBParser`) that can handle residue
 numbers up to 99,999.
 
+TODO:
+    Add attributes to guess elements for non-physical or missing elements
+
+
 .. Note::
 
    The parser processes atoms and their names. Masses are guessed and set to 0
```
IAlibay: Maybe worth adding here that elements are processed if all are valid.

testsuite/MDAnalysisTests/topology/test_pdb.py
IAlibay: As per @lilyminium's comment, what expected warnings are being filtered here?

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -244,3 +247,104 @@ def test_PDB_metals():
     assert u.atoms[1].mass == pytest.approx(tables.masses["FE"])
     assert u.atoms[2].mass == pytest.approx(tables.masses["CA"])
     assert u.atoms[3].mass == pytest.approx(tables.masses["MG"])
+
+
+# An example of ideal PDB file used to test elements attribute.
```
IAlibay: Rather than putting this as a comment, could you put it in the REMARK section of the PDB? (see `PDB_hex` above).

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -244,3 +247,104 @@ def test_PDB_metals():
...
+HETATM 1612  C2  DMS A 101      18.750  40.066  17.029  1.00 20.50           C
+HETATM 1613  S   DMS A 102      22.157  39.211  12.217  1.00 27.26           S1
+HETATM 1614  O   DMS A 102      20.622  38.811  11.702  1.00 25.51           O
+HETATM 1615  C1  DMS A 102      22.715  40.764  11.522  1.00 26.00           C
+HETATM 1616  C2  DMS A 102      22.343  39.515  13.971  1.00 25.46           C
+TER    1617
+"""
+
+def test_PDB_elements():
+    """
```
IAlibay: The indentation of the docstring here seems off, see the following for an example of how I'd expect it to be indented:

https://github.com/MDAnalysis/mdanalysis/blob/5f513831301492ffcf054c7b57ee6092bf153c41/package/MDAnalysis/topology/TOPParser.py#L316-L323

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -244,3 +247,104 @@ def test_PDB_metals():
...
+        The elements attribute is equipped to obtain the element symbols,
+        if the PDB file contains the element symbol
+        """
+    u = mda.Universe(StringIO(PDB_elements), format='PDB')
+    element_list = np.array([
+                             'N', 'C', 'C', 'O', 'C', 'C', 'O', 'N', 'H',
+                             'H', 'H', 'H', 'H', 'H', 'H', 'H', 'Cu', 'Fe',
+                             'Mg', 'Ca', 'S', 'O', 'C', 'C', 'S','O', 'C',
```
IAlibay: Missing space between 'S','O'

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -244,3 +247,104 @@ def test_PDB_metals():
...
+"""
+
+def test_PDB_elements():
+    """
+        The elements attribute is equipped to obtain the element symbols,
+        if the PDB file contains the element symbol
+        """
+    u = mda.Universe(StringIO(PDB_elements), format='PDB')
+    element_list = np.array([
+                             'N', 'C', 'C', 'O', 'C', 'C', 'O', 'N', 'H',
```
IAlibay: So more of a personal preference so ignore if you'd like, but here I would probably just do the following here, just because it saves you a line of code 🙃 

```
np.array(['N', 'C', 'C', 'O', 'C', 'C', 'O', 'N', 'H',
               'H', 'H', 'H', 'H', 'H', 'H', 'H', 'Cu', 'Fe',
               .... ], dtype=object)
```

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -244,3 +247,104 @@ def test_PDB_metals():
...
+ATOM      5  X   ASN A   1      -9.437   3.396  -2.889  1.00  0.00          XX
+TER       6
+HETATM    7 CU    CU A   2      03.000  00.000  00.000  1.00 00.00          CU
+HETATM    8 FE    FE A   3      00.000  03.000  00.000  1.00 00.00          Fe
+HETATM    9 Mg    Mg A   4      03.000  03.000  03.000  1.00 00.00          MG
+TER       10
+"""
+
+def test_missing_elements_warnings():
+    """
```
IAlibay: See above comment on docstring.

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -244,3 +247,104 @@ def test_PDB_metals():
...
+    with pytest.warns(UserWarning) as record:
+        u = mda.Universe(StringIO(PDB_missing_ele), format='PDB')
+    
+    assert len(record) == 1
+    assert record[0].message.args[0] == "Element information absent "\
+        "or inadequate."
+
+
+def test_wrong_elements_warnings():
+    """
```
IAlibay: See above comment on docstring.

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -244,3 +247,104 @@ def test_PDB_metals():
...
+ATOM      3  C   ASN A   1      -7.117   2.964  -1.897  1.00  0.00
+ATOM      4  O   ASN A   1      -6.634   1.849  -1.758  1.00  0.00
+TER       5
+HETATM    6 CU    CU A   2      03.000  00.000  00.000  1.00 00.00
+HETATM    7 FE    FE A   3      00.000  03.000  00.000  1.00 00.00
+HETATM    8 Mg    Mg A   4      03.000  03.000  03.000  1.00 00.00
+TER       9
+"""
+
+PDB_wrong_ele = """\
```
IAlibay: Again more of a personal preference, but I'd consider maybe moving this until after `test_missing_elements_warnings`, that way you have a more consistent PDB_file -> function -> PDB_file -> function type of pattern.

package/MDAnalysis/topology/PDBParser.py
```diff
@@ -303,6 +313,23 @@ def _parseatoms(self):
         masses = guess_masses(atomtypes)
         attrs.append(Masses(masses, guessed=True))
 
+
+        # Getting element information from element column.
+        periodic_table = [k for k,v in SYMB2Z.items()]
```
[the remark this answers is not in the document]
AmeyaHarmalkar: That's quite helpful! I was actually concerned that list might not be the efficient way to parse a huge PDB. I will change this right away.
Thanks for this!

package/MDAnalysis/topology/PDBParser.py
```diff
@@ -303,6 +313,23 @@ def _parseatoms(self):
...
+        if all(elements):
+            
+            elements = [i.capitalize() for i in elements]
+            
+            if all( element in periodic_table for element in elements):
+                attrs.append(Elements(elements))
+            else:
+                warnings.warn("Element information invalid.")
```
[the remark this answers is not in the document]
AmeyaHarmalkar: Makes sense. Editing that

testsuite/MDAnalysisTests/topology/test_pdb.py
AmeyaHarmalkar: The PDB_metals popped up the UserWarnings for the elements attributes. So, these warnings are filtered here. I think tis happened because a few of the test PDBs would not have the elements column. But, I didn't want those cases to affect the test case for metals(which happens here), hence I filtered the UserWarnings. Is there any other way to go around this problem?

testsuite/MDAnalysisTests/topology/test_pdb.py
AmeyaHarmalkar: That actually makes perfect sense. Consistency is better!

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -244,3 +247,104 @@ def test_PDB_metals():
...
+HETATM 1613  S   DMS A 102      22.157  39.211  12.217  1.00 27.26           S1
+HETATM 1614  O   DMS A 102      20.622  38.811  11.702  1.00 25.51           O
+HETATM 1615  C1  DMS A 102      22.715  40.764  11.522  1.00 26.00           C
+HETATM 1616  C2  DMS A 102      22.343  39.515  13.971  1.00 25.46           C
+TER    1617
+"""
+
+def test_PDB_elements():
+    """The elements attribute is equipped to obtain the element symbols,
+        if the PDB file contains a valid element symbol
```
IAlibay: the extra indentation isn't necessary here I think, so:

```
    """The elements attribute...
    if the PDB file
```

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -244,3 +247,104 @@ def test_PDB_metals():
...
+TER       5
+HETATM    6 CU    CU A   2      03.000  00.000  00.000  1.00 00.00
+HETATM    7 FE    FE A   3      00.000  03.000  00.000  1.00 00.00
+HETATM    8 Mg    Mg A   4      03.000  03.000  03.000  1.00 00.00
+TER       9
+"""
+
+def test_missing_elements_warnings():
+    """The test checks whether it asserts the appropriate warning on
+        encountering missing elements.
```
IAlibay: As above.

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -244,3 +247,104 @@ def test_PDB_metals():
...
+TER       6
+HETATM    7 CU    CU A   2      03.000  00.000  00.000  1.00 00.00          CU
+HETATM    8 FE    FE A   3      00.000  03.000  00.000  1.00 00.00          Fe
+HETATM    9 Mg    Mg A   4      03.000  03.000  03.000  1.00 00.00          MG
+TER       10
+"""
+
+def test_wrong_elements_warnings():
+    """The test checks whether it fills in the wrong elements appropriately
+        with the atom names where element symbols are missing.
```
IAlibay: As above.

testsuite/MDAnalysisTests/topology/test_pdb.py
IAlibay: Because `@pytest.mark.filterwarnings("error")` is set below it (i.e. make all warnings into errors), I'm not sure if blanket disabling UserWarnings adheres to the intended test behaviour.

Instead here I would suggest using `@pytest.mark.filterwarnings("ignore: attributes will not be populated.")`. Thankfully there's overlap between the two warnings 🙃

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -244,3 +247,104 @@ def test_PDB_metals():
...
+HETATM 1610  O   DMS A 101      19.279  37.904  18.777  1.00 23.69           Ox
+HETATM 1611  C1  DMS A 101      21.344  39.260  17.532  1.00 24.07           C
+HETATM 1612  C2  DMS A 101      18.750  40.066  17.029  1.00 20.50           C
+HETATM 1613  S   DMS A 102      22.157  39.211  12.217  1.00 27.26           S1
+HETATM 1614  O   DMS A 102      20.622  38.811  11.702  1.00 25.51           O
+HETATM 1615  C1  DMS A 102      22.715  40.764  11.522  1.00 26.00           C
+HETATM 1616  C2  DMS A 102      22.343  39.515  13.971  1.00 25.46           C
+TER    1617
+"""
+
```
IAlibay: PEP8: I believe you need two empty lines here.

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -244,3 +247,104 @@ def test_PDB_metals():
...
+ATOM      2  CA  ASN A   1      -8.608   3.135  -1.618  1.00  0.00
+ATOM      3  C   ASN A   1      -7.117   2.964  -1.897  1.00  0.00
+ATOM      4  O   ASN A   1      -6.634   1.849  -1.758  1.00  0.00
+TER       5
+HETATM    6 CU    CU A   2      03.000  00.000  00.000  1.00 00.00
+HETATM    7 FE    FE A   3      00.000  03.000  00.000  1.00 00.00
+HETATM    8 Mg    Mg A   4      03.000  03.000  03.000  1.00 00.00
+TER       9
+"""
+
```
IAlibay: As above about PEP8.

testsuite/MDAnalysisTests/topology/test_pdb.py
IAlibay: I think this docstring needs changing since this is no longer what this tests does?

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -244,3 +247,104 @@ def test_PDB_metals():
...
+def test_missing_elements_warnings():
+    """The test checks whether it asserts the appropriate warning on
```
IAlibay: Since `assert` means something in python, I would instead suggest replacing that word with "gives".

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -244,3 +247,104 @@ def test_PDB_metals():
...
+HETATM 1612  C2  DMS A 101      18.750  40.066  17.029  1.00 20.50           C
+HETATM 1613  S   DMS A 102      22.157  39.211  12.217  1.00 27.26           S1
+HETATM 1614  O   DMS A 102      20.622  38.811  11.702  1.00 25.51           O
+HETATM 1615  C1  DMS A 102      22.715  40.764  11.522  1.00 26.00           C
+HETATM 1616  C2  DMS A 102      22.343  39.515  13.971  1.00 25.46           C
+TER    1617
+"""
+
+def test_PDB_elements():
+    """The elements attribute is equipped to obtain the element symbols,
```
IAlibay: Might be worth changing the docstring to be more descriptive of what the test does. So "This test checks that the elements attribute is assigned properly given a PDB file containing a valid elements record".

testsuite/MDAnalysisTests/topology/test_pdb.py
AmeyaHarmalkar: I didn't really get this one. Do you mean something like this?

```
HETATM 1616  C2  DMS A 102      22.343  39.515  13.971  1.00 25.46           C
TER    1617
"""


def test_PDB_elements():
```

AmeyaHarmalkar: Also, for @pytest.mark.filterwarnings, if I give it a string, there are several Internal errors and pytest exits with just 3/47 passed tests. It also doesn't point out the error which is weird but basically the module doesn't exist is a common Internal Error which has been popping

IAlibay: Yeah so according to PEP8 there should be two empty lines between class or function definitions. According to flake8 this also extends to global variables like these PDB variables.

Note: you'll also want two lines above line 237.

Related note: You'll notice that this isn't consistent throughout `test_pdb`. As far as I understand this is because a lot of this code got written before PEP8 became the accepted standard for MDA. So our current approach is to ensure all future test is PEP8 compliant and fix PEP8 issues when we encounter them 😄. In case you don't already, if you use an IDE of some kind for your code (e.g. VScode, atom), I'd recommend you have flake8 installed so you can pick up these issues.

IAlibay: > Also, for @pytest.mark.filterwarnings, if I give it a string, there are several Internal errors and pytest exits with just 3/47 passed tests. It also doesn't point out the error which is weird but basically the module doesn't exist is a common Internal Error which has been popping.
> Update: I don't think I need to have the UserWarning filter option. Since, we have made it more conservative, I think it is not popping any user valid for Invalid elements as all the PDBs used are valid.

I'm not sure I understand what's going on, do you want to push the code and we can have a look at it?

AmeyaHarmalkar: Oh gotcha. My understanding was that there must be 2 blank lines between function definitions. I didn't know that we would need it for the string declaration and the function as well. 
I created a MDAnalysis workspace in XCode, I will check if there is a possible alternative to flake8 or maybe flake8 to install in there. This is helpful. 
Just pushing through the changes in a moment

AmeyaHarmalkar:
@IAlibay pushed the changes. I made the changes in the AUTHORS and CHANGELOG, but there is some issue which I am not sure of.

IAlibay:
> @IAlibay pushed the changes. I made the changes in the AUTHORS and CHANGELOG, but there is some issue which I am not sure of.

By issues, do you mean the merge conflicts?

package/CHANGELOG
```diff
@@ -16,7 +16,8 @@ The rules for this file:
 mm/dd/yy richardjgowers, kain88-de, lilyminium, p-j-smith, bdice, joaomcteixeira,
          PicoCentauri, davidercruz, jbarnoud, RMeli, IAlibay, mtiberti, CCook96,
          Yuan-Yu, xiki-tempula, HTian1997, Iv-Hristov, hmacdope, AnshulAngaria,
-         ss62171, Luthaf
+         ss62171, Luthaf, yuxuanzhuang, abhishandy, mlnance, shfrz, orbeckst,
+         wvandertoorn, cbouy, AmeyaHarmalkar
```
IAlibay: You'll also need to add an entry under "enhancements".

package/MDAnalysis/topology/PDBParser.py
```diff
@@ -303,6 +317,24 @@ def _parseatoms(self):
         masses = guess_masses(atomtypes)
         attrs.append(Masses(masses, guessed=True))
 
+
+        # Getting element information from element column.
+        periodic_table = set(SYMB2Z)
+        if all(elements):
+
```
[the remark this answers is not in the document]
IAlibay: Not sure from the thumbs up if you wanted to fix this one but didn't get around to in your latest commit?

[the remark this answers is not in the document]
lilyminium: If you're just looking at the keys in `SYMB2Z`, there's no need to convert it to a set. `if x in dict` returns true if `x` is a key in `dict`.

lilyminium:
Looks good, thanks for the fix @AmeyaHarmalkar and for taking over @IAlibay! Travis won't run while there are merge conflicts, I think. It's just with AUTHORS and CHANGELOG so you could just resolve on GitHub if you had to

package/MDAnalysis/topology/PDBParser.py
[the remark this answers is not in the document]
IAlibay: So just for fun, I thought it'd be interesting to see if there was actually a speed difference between directly accessing the `dict` keys or doing the set conversion. Interestingly, it looks like there is a ~ 8-10% cost.

That being said on 5 million elements that's a ~ 0.02 seconds (+/- ~ 0.001) difference, so at this point it's probably all just academic 🙃 Definitely @lilyminium's approach is a much better/cleaner call here.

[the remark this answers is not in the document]
AmeyaHarmalkar: So, should I switch it to @lilyminium's approach or keep it a set?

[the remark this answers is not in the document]
lilyminium: Interesting, @IAlibay, I'm getting the opposite direction on that speed difference 😁. It's negligible either way. *However*, it occured to me that you could use `set` elsewhere....

**functions and input**
Screenshot 2020-03-29 at 10 26 13 AM

**timings**
Screenshot 2020-03-29 at 10 26 55 AM

On 5 million elements, 1 s --> 50 ms. Even on the 10k max elements a PDB should have, a 10x speed up looks pretty good, even though the absolute difference is 1 millisecond. So @AmeyaHarmalkar although I would personally just use SYMB2Z just to save a line of code, it's up to you -- but converting elements to a set seems like a good idea.

AmeyaHarmalkar:
I don't have right access, so I am unable to resolve the conflicts in AUTHORS and CHANGELOG. Let me know if the changes look good so far, and it would be really helpful if someone could help me with resolving the conflicts. Thanks @lilyminium and @IAlibay!

lilyminium:
Ok @AmeyaHarmalkar I fixed the merge conflicts on this end, now if you want to make changes you'll need to `git pull origin test-develop` to pull the changes into your local repo.

AmeyaHarmalkar:
Thanks @lilyminium! 
The codecov/project check failed. Everything else seems alright. Any further suggestions to work on the code @IAlibay?

package/MDAnalysis/topology/PDBParser.py
```diff
@@ -303,6 +315,18 @@ def _parseatoms(self):
         masses = guess_masses(atomtypes)
         attrs.append(Masses(masses, guessed=True))
 
+        # Getting element information from element column.
+        if all(elements):
+            elements = [i.capitalize() for i in elements]
```
IAlibay: As per @lilyminium's comment: https://github.com/MDAnalysis/mdanalysis/pull/2627#discussion_r399722309

Doing a set comparison for elements here would be faster. So:

```python
if all(elements):
    elem_set = set(i.capitalize() for i in i in set(elements))
    if all(element in SYMB2Z for element in elem_set):
        attrs.append(Elements(np.array([i.capitalize() for i in elements], dtype=object))
    else:
        .....
```

Note: I noticed that every other attrs being appended are actually being converted to numpy arrays, so we probably should be consistent here (unless there's a reason not to?).

package/MDAnalysis/topology/PDBParser.py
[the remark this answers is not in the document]
IAlibay: TIL set creation is way more expensive than I realised 🙃

package/CHANGELOG
```diff
@@ -76,10 +76,11 @@ Fixes
   * Added missing selection module to leaflet.py (Issue #2612)
 
 Enhancements
+  * Added elements attribute to PDBParser (Issue #2553)
```
IAlibay: You should also add 2647 here.

AmeyaHarmalkar: Sure!
@IAlibay, I git pulled on my local repo and I can't import the mda universe due a module error. It has issues with importing the tqdm package (it was one of the newly merged changes I believe). Can you help me in figuring out what am I doing wrong? Thanks and sorry for the trouble

IAlibay:
@AmeyaHarmalkar I just got a notification with something along the lines of "I git pulled on my local repo and I can't import the mda universe due a module error. It has issues with importing the tqdm package (it was one of the newly merged changes I believe)."

For some reason the comment isn't showing up here, did you fix the issue?

AmeyaHarmalkar:
Yeah @IAlibay it was a rookie mistake. I figured it out. No worries!

package/CHANGELOG
```diff
@@ -76,10 +76,11 @@ Fixes
   * Added missing selection module to leaflet.py (Issue #2612)
 
 Enhancements
+* Added elements attribute to PDBParser (Issue #2553, #2647)
```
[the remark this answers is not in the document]
AmeyaHarmalkar: It's actually very helpful @IAlibay. Makes me aware of my mistakes and definitely helps me write a better code

package/MDAnalysis/topology/PDBParser.py
```diff
@@ -303,6 +315,19 @@ def _parseatoms(self):
         masses = guess_masses(atomtypes)
         attrs.append(Masses(masses, guessed=True))
 
+        # Getting element information from element column.
+        if all(elements):
+            element_set = set(i.capitalize() for i in set(elements))
+            if all( element in SYMB2Z for element in element_set):
+                attrs.append(Elements(np.array([i.capitalize() for i in elements],
```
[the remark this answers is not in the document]
zemanj: I usually solve such situations with intermediate variables, e.g.
```python
            if all(element in SYMB2Z for element in element_set):
                element_list = [i.capitalize() for i in elements]
                attrs.append(Elements(np.array(element_list, dtype=object)))
```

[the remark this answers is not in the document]
AmeyaHarmalkar: I could make that change. I thought probably we are saving one line of code, but can definitely make the change if needed.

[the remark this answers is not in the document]
IAlibay: @AmeyaHarmalkar probably in this case, adding one line for a bit more readability is worth it. If you feel up to making the change please do.

package/MDAnalysis/topology/PDBParser.py
```diff
@@ -303,6 +315,20 @@ def _parseatoms(self):
...
+            if all(element in SYMB2Z for element in element_set):
+                attrs.append(Elements(np.array(
+                             [i.capitalize() for i in elements],
+                             dtype=object)))
+            else:
+                warnings.warn("Invalid elements found in the PDB file, "
+                              "elements attributes will not be populated.")
+        else:
+            warnings.warn("Element information is absent or missing for a few "
+                          "residues. Elements attributes will not be populated.")
```
IAlibay: @AmeyaHarmalkar if you are fixing @zemanj's comment: https://github.com/MDAnalysis/mdanalysis/pull/2627#discussion_r399829470 then I probably would also change `a few residues` here to `some atoms` (since elements are atom-wise).

AmeyaHarmalkar: Sure thing. Updating and pushing through

IAlibay:
Thanks so much for all your work @AmeyaHarmalkar and congrats on your first merged MDA PR 🎉

AmeyaHarmalkar:
Thank you @IAlibay and @lilyminium for all your help! This was a great learning experience!

Issue #2206, opened 2019-02-17
Parameters not saved in Analysis classes


PicoCentauri:
In all analysis classes the parameters `start`, `stop`, `step` and `verbose` are not saved if they are changed after the object was run once.

After looking into the [code](https://github.com/MDAnalysis/mdanalysis/blob/develop/package/MDAnalysis/analysis/base.py#L137-L139) I found that for the `start`, `stop` and `step` parameters this is an issue due to the fix of the old deprecated syntax. For the `verbose` one just have to save this in the `_verbose` attribute of the object. All these things are easy to fix and I could provide a patch.

**Code to reproduce the behavior**

``` python
from MDAnalysis.analysis.lineardensity import LinearDensity
from MDAnalysis.tests.datafiles import PSF_TRICLINIC, DCD_TRICLINIC

u = mda.Universe(PSF_TRICLINIC, DCD_TRICLINIC)
ldens = LinearDensity(u.atoms, verbose=True)

ldens.run(start=0, stop=10)
print(ldens.start, ldens.stop)

ldens.run(start=0, stop=5, verbose=False)
print(ldens.start, ldens.stop)
```

**Currently version of MDAnalysis**

Python 3.6, mda 0.19.3, MacOS

richardjgowers:
Yeah we used to save them on `__init__` and use this during `run()`.  We could save the last used values in `run()` I guess.

PicoCentauri:
Yes. I would just add simple test in `_setup_frames`. See PR #2207.

orbeckst:
I don't think we should save them. That's a left over when we set them in init. When providing parameters to run I would expect that they only apply to _this_ invocation.

orbeckst:
I don't think we should save them. That's a left over when we set them in init. When providing parameters to run I would expect that they only apply to _this_ invocation.

PicoCentauri:
One the one hand I could imagine that sometimes its useful to access this information i.e. at what frame the analysis started. On the other the user already put the information in so maybe its superfluous.

zemanj:
@orbeckst 
>I don't think we should save them. That's a left over when we set them in init. When providing parameters to run I would expect that they only apply to _this_ invocation.

I agree. Since the `start`/`stop`/`step` parameters are not passed to the initializer, they shouldn't be reused. The `run()` method should *always* work as documented, including its kwarg defaults. Permanently storing and reusing the parameters would invalidate the documented signature.

@PicoCentauri 
[...]

PR #460, opened 2015-09-30
Feature polymers & analysis base

richardjgowers:
So I'm trying to polish up a few of the tools I've made lately, starting with persistence length.

I made an `_AnalysisBase` class to run things.  Hopefully this can be used by lots of tools so that initialising and using analysis tools uses a similar set of keywords.  For the init, I've got the following kwargs to control where in the trajectory is used:
 - start, first frame to use
 - stop, last frame to use
 - skip, spacing between frames

Then this is started using `run`.  Does this look remotely useful to other people writing stuff?  If so, it could move into `analysis.base`.

I wasn't sure what the best way to make results available is.  Maybe a dictionary called `results` that has all possible things that people might be interested in?

Thoughts/opinions/random heckling?

PR #2631, opened 2020-03-16
Setting Verbose 

### LINKED ISSUE #2504: verbose not set in analysis.AnalysisBase.run

Purva-Chaudhari:
Fixes #2504 

 - Deprecation warning for start/stop/ step
- verbose set

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


Purva-Chaudhari:
@lilyminium over-ridding needs to be done only when verbose is set to True in run (as i observed in some of the test cases) Hence i set it this way. But still if this doesn,t work i have another approach in mind . Do let me know,
Thanking you

package/MDAnalysis/analysis/base.py
```diff
@@ -126,9 +126,14 @@ def _setup_frames(self, trajectory, start=None, stop=None, step=None):
         """
         self._trajectory = trajectory
         start, stop, step = trajectory.check_slice_indices(start, stop, step)
-        self.start = start
-        self.stop = stop
-        self.step = step
+        if start is not None:
```
IAlibay: Quick question here, is there a case where `check_slice_indices` returns `None` for either of `start`, `stop` or `step`? As far as I was aware it already returned ints?

package/MDAnalysis/analysis/base.py
```diff
@@ -126,9 +126,14 @@ def _setup_frames(self, trajectory, start=None, stop=None, step=None):
...
+            self._start = start
+        elif stop is not None:
+            self._stop = stop
+        elif step is not None:
+            self._step = step
+        else:
+            warnings.warn("Deprecation Warning", DeprecationWarning)
```
IAlibay: Aside from the above comment, I'm not sure I understand how this relates to setting verbose. Apologies if I'm just missing something simple, any chance you could provide extra information on why you're doing this?

package/MDAnalysis/analysis/base.py
```diff
@@ -173,7 +178,7 @@ def run(self, start=None, stop=None, step=None, verbose=None):
         logger.info("Choosing frames to analyze")
         # if verbose unchanged, use class default
         verbose = getattr(self, '_verbose', False) if verbose is None else verbose
-
+        if verbose == True : self.verbose = verbose
```
IAlibay: Based on line 142, shouldn't this be `self._verbose`? 
Also, I'm probably getting the logic wrong here, but wouldn't it be best to always set `self._verbose` here? What if you have a case where you pass `False` to `run()` but `True` to `__init__`, in this construct `self._verbose` would be set to True even if verbose is currently `False` right?

package/MDAnalysis/analysis/base.py
Purva-Chaudhari: I am sorry got confused with issue #2206 requirements asking to Adjust AnalysisBase.__init__() so that start/stop/step are stored in the respective private variables. Also make sure that the DeprecationWarning is raised even if any of the deprecated kwargs is None (currently missing).

Will remove that part for now and work on it after solving this one

package/MDAnalysis/analysis/base.py
Purva-Chaudhari: In case of parameter overriding for the run class can it be directly:
 verbose = getattr(self, '_verbose', False) if verbose is None else verbose
 -> self._verbose = verbose after the line

Purva-Chaudhari: Please do correct me if i am going wrong in understanding the requirement. (Just re-read the comments regarding deprecation.)

Purva-Chaudhari:
@IAlibay Could you please have a look and suggest. I think there is no need to introduce a new subclass / method for overriding if its only within for the run class. Do correct me if I am wrong

IAlibay:
> @IAlibay Could you please have a look and suggest. I think there is no need to introduce a new subclass / method for overriding if its only within for the run class. Do correct me if I am wrong .
> 
> If this works fine I would simultaneously start with issue #2206 Since they are all related

@Purva-Chaudhari as far as I can tell this *should* work, you'll probably want to introduce a test though. Probably something similar to:

https://github.com/MDAnalysis/mdanalysis/blob/7cb41846a8eebe1d2f46764ee5cb3a708fbeb20d/testsuite/MDAnalysisTests/analysis/test_base.py#L93-L95

But testing the influence of verbose being set when calling `run`.

IAlibay:
I do wonder if it might be cheaper / safer to just explicitly pass `verbose` to `_setup_frames` (in the same manner as `start`, `stop`, and `step`). Especially if the long term idea is to make all those values only temporarily overriden by `run` (#2206).

Maybe one of the @MDAnalysis/coredevs might have some thoughts on this?

orbeckst:
I like @IAlibay 's idea to pass `verbose` through to https://github.com/MDAnalysis/mdanalysis/blob/7cb41846a8eebe1d2f46764ee5cb3a708fbeb20d/package/MDAnalysis/analysis/base.py#L111
instead of setting/unsetting `AnalysisBase._verbose`. This looks like a clean solution that makes sure that `run(..., verbose=VALUE)` is only temporary.

Purva-Chaudhari:
@IAlibay so should I pass verbose explicitly to  _setup_frames , in that case of temporary overriding, the line : verbose = getattr(self, '_verbose', False) in _setup_frames can be removed and i guess then we dont have to set self._verbose in run/_setup_frames.

Should I make this change and then introduce test?

IAlibay:
> @IAlibay so should I pass verbose explicitly to _setup_frames . In that case of temporary overriding, the line : verbose = getattr(self, '_verbose', False) in _setup_frames can be removed and i guess then we don,t have to set self._verbose in run/_setup_frames.
> 
> Should I make this change and then introduce test?

@Purva-Chaudhari Looks like the idea is popular, please do go ahead 👍

Purva-Chaudhari:
@IAlibay Could you please suggest me further

package/MDAnalysis/analysis/base.py
```diff
@@ -108,7 +108,7 @@ def __init__(self, trajectory, verbose=False, **kwargs):
         self._trajectory = trajectory
         self._verbose = verbose
 
-    def _setup_frames(self, trajectory, start=None, stop=None, step=None):
+    def _setup_frames(self, trajectory, start=None, stop=None, step=None, verbose = None):
         """
         Pass a Reader object and define the desired iteration pattern
         through the trajectory
```
IAlibay: The docstring needs updating to reflect that `verbose` is now an explicit argument to `_setup_frames`. Please also include a `.. versionchanged` entry detailing that the `verbose` argument was added in 1.0.0.

package/MDAnalysis/analysis/base.py
```diff
@@ -134,7 +134,7 @@ def _setup_frames(self, trajectory, start=None, stop=None, step=None):
         if interval == 0:
             interval = 1
 
-        verbose = getattr(self, '_verbose', False)
+
```
IAlibay: Probably don't need two empty lines here.

package/MDAnalysis/analysis/base.py
```diff
@@ -134,7 +134,7 @@ def _setup_frames(self, trajectory, start=None, stop=None, step=None):
         if interval == 0:
```
IAlibay: Do also add a versionchanged to `run` to document this new behaviour (i.e. temporary overriding of `__init__` set `verbose`).

IAlibay:
@Purva-Chaudhari also don't forget to update CHANGELOG and AUTHORS!

Purva-Chaudhari:
@IAlibay , @orbeckst  for the case of testing of temporary overriding should I return verbose too from run method and in  testsuite/MDAnalysisTests/analysis/test_base.py, i would then introduce test_verbose_run() as
 a, run_verbose = FrameAnalysis(u.trajectory).run(verbose = True)
    assert run_verbose

jbarnoud:
The verbosity is not a result, it should not be returned. Checking stdout or stderr would be better.

orbeckst:
@joaomcteixeira could you please also keep an eye on this PR and help out @IAlibay who is already doing many others. Thanks!

lilyminium:
Sorry for not seeing this earlier @Purva-Chaudhari. I think you should be able to check the verbosity of the ProgressMeter at ``FrameAnalysis._pm.verbose`` and compare that to the verbosity of the analysis class at ``FrameAnalysis._verbose``?

Purva-Chaudhari:
> Sorry for not seeing this earlier @Purva-Chaudhari. I think you should be able to check the verbosity of the ProgressMeter at `FrameAnalysis._pm.verbose` and compare that to the verbosity of the analysis class at `FrameAnalysis._verbose`?

@lilyminium The temporary overriding of verbose is done by passing it to `_setup_frames` as suggested by @IAlibay , `self._verbose` not set. So I guess I need to compare `FrameAnalysis._pm.verbose` with the verbose in run, right?.   `FrameAnalysis._verbose` would is not overridden. 
Please do correct me if its wrong
So do I have to check for assert_equal (both the verbose)

lilyminium:
@Purva-Chaudhari yes, you would be checking that the ``verbose`` you pass into ``run`` is the same as the ProgressMeter but not necessarily the same as the FrameAnalysis value.

IAlibay:
@Purva-Chaudhari can you solve your merge conflicts, that way we can see if Travis runs green.

package/MDAnalysis/analysis/base.py
```diff
@@ -123,6 +123,9 @@ def _setup_frames(self, trajectory, start=None, stop=None, step=None):
             stop frame of analysis
         step : int, optional
             number of frames to skip between each analysed frame
```
IAlibay: Please do add verbose to the parameter list please.

package/MDAnalysis/analysis/base.py
```diff
@@ -123,6 +123,9 @@ def _setup_frames(self, trajectory, start=None, stop=None, step=None):
...
+            
+        .. versionchanged:: 1.0.0
+        added verbose argument
```
IAlibay: Identation issue, should be:
```
.. versionchanged:: 1.0.0
   Verbose argument is now passed explicitly from :meth:`AnalysisBase.run`.
```

package/MDAnalysis/analysis/base.py
```diff
@@ -169,12 +171,15 @@ def run(self, start=None, stop=None, step=None, verbose=None):
             number of frames to skip between each analysed frame
         verbose : bool, optional
             Turn on verbosity
+            
+        .. versionchanged:: 1.0.0
+        verbose explicitly passed to _setup_frames as argument for temporary
```
IAlibay: Same as above, please do use :meth: and :class: docstring decorators too if you can, that way users can navigate to the right place when looking at the sphinx documentation.

testsuite/MDAnalysisTests/analysis/test_base.py
```diff
@@ -95,6 +95,12 @@ def test_verbose(u):
     assert a._verbose
 
 
+def test_verbose_run(u):
+    a = FrameAnalysis(u.trajectory, verbose=False).run(verbose = True)
```
IAlibay: PEP8 correction: no spaces around `=` please.

testsuite/MDAnalysisTests/analysis/test_base.py
```diff
@@ -95,6 +95,12 @@ def test_verbose(u):
     assert a._verbose
 
 
+def test_verbose_run(u):
```
IAlibay: Here it would be good to test multiple cases using pytest.mark.parametrize, i.e. the cases of:

init, run
True, True
False, True
True, False
False, False

Let us know if you need any help to work out how to use parametrize. See `test_AnalysisFromFunction` for an example of it.

Purva-Chaudhari: @pytest.mark.parametrize('verbose, verbose_r',[(True,True), (True,False), (False,True), (False,False)])
def test_verbose_run(u,verbose, verbose_r)
   a = FrameAnalysis(u.trajectory, verbose=verbose).run(verbose = verbose_r)
(should I include the four conditions )  
 if (a._verbose == True and a._pm.verbose == True): 
      assert a._verbose
      assert a._pm.verbose
   elif (a._verbose == True and a._pm.verbose == False): 
      assert a._verbose
      assert not a._pm.verbose
   elif (a._verbose == False and a._pm.verbose == True): 
      assert not a._verbose
      assert a._pm.verbose
   elif (a._verbose == False and a._pm.verbose == False): 
      assert not a._verbose
      assert not a._pm.verbose
or simply 
      assert a._verbose
      assert a._pm.verbose

jbarnoud:
There is a pull request about replacing the progress meter to use tqmd. It is worth checking that the testing strategy remains valid.

lilyminium:
Unfortunately, I think #2617 actually makes this PR redundant since it never sets ``verbose`` or passes it to ``_setup_frames``.

jbarnoud:
That other PR should probably not change the API. We should check that. On 20 Mar 2020 11:18, Lily Wang <notifications@github.com> wrote:
Unfortunately, I think #2617 actually makes this PR redundant since it never sets verbose or passes it to _setup_frames.

testsuite/MDAnalysisTests/analysis/test_base.py
IAlibay: You don't need the if/else construct, just assert that the a._verbose and a._pm.verbose values match the expected inputs verbose and verbose_r.

Obviously, this is all dependent on https://github.com/MDAnalysis/mdanalysis/pull/2631#issuecomment-601640537

Purva-Chaudhari:
What should I do further then?

package/MDAnalysis/analysis/base.py
```diff
@@ -109,7 +109,7 @@ def __init__(self, trajectory, verbose=False, **kwargs):
         self._trajectory = trajectory
         self._verbose = verbose
 
-    def _setup_frames(self, trajectory, start=None, stop=None, step=None):
+    def _setup_frames(self, trajectory, start=None, stop=None, step=None, verbose = None):
```
joaomcteixeira: use `verbose=None` without spaces

joaomcteixeira:
Hi @Purva-Chaudhari ,

Just a small comment on to properly develop on a fork. I see you have made the changes in your forked `develop` branch. This is not completely wrong, but it is best to branch out first to a new branch that addresses only the PR/issue in question. And then, Pull Request that branch to the main repository base. If you continue developing new features in the fork base branch, as per habit, it may get really problematic if you have to deal with complex merge conflicts in future situations.

package/MDAnalysis/analysis/base.py
```diff
@@ -135,7 +141,6 @@ def _setup_frames(self, trajectory, start=None, stop=None, step=None):
         if interval == 0:
             interval = 1
 
-        verbose = getattr(self, '_verbose', False)
         self._pm = ProgressMeter(self.n_frames if self.n_frames else 1,
```
joaomcteixeira: Python tip: 

```python
a = None or  1
# or
b = 0 or 77
```
:wink:

package/MDAnalysis/analysis/base.py
```diff
@@ -170,12 +175,15 @@ def run(self, start=None, stop=None, step=None, verbose=None):
...
         verbose : bool, optional
             Turn on verbosity
+            
+        .. versionchanged:: 1.0.0
+            Verbose argument is now passed explicitly to
+            :meth:`AnalysisBase._setup_frames`.
         """
         logger.info("Choosing frames to analyze")
         # if verbose unchanged, use class default
         verbose = getattr(self, '_verbose', False) if verbose is None else verbose
```
joaomcteixeira: To my understanding, there is no need to use `getattr` here. `getattr` and `setattr` are used on dynamic operations, not on static ones.

testsuite/MDAnalysisTests/analysis/test_base.py
```diff
@@ -95,6 +95,12 @@ def test_verbose(u):
...
+    a = FrameAnalysis(u.trajectory, verbose=False).run(verbose = True)
+    assert not a._verbose
```
joaomcteixeira: `assert a._verbose is False`

package/MDAnalysis/analysis/base.py
```diff
@@ -170,12 +175,15 @@ def run(self, start=None, stop=None, step=None, verbose=None):
...
-
-        self._setup_frames(self._trajectory, start, stop, step)
+        self._setup_frames(self._trajectory, start, stop, step, verbose)
```
joaomcteixeira: I don't think this is correct. Why should a method of a class receive an argument that is an attribute of that same class when the method being called is not a `@staticmethod` ? (may be this is not from this PR and I am missing something)

testsuite/MDAnalysisTests/analysis/test_base.py
```diff
@@ -95,6 +95,12 @@ def test_verbose(u):
...
+    assert a._pm.verbose
```
joaomcteixeira: same here `is True`

package/MDAnalysis/analysis/base.py
Purva-Chaudhari: > Hi @Purva-Chaudhari ,
> 
> Just a small comment on to properly develop on a fork. I see you have made the changes in your forked `develop` branch. This is not completely wrong, but it is best to branch out first to a new branch that addresses only the PR/issue in question. And then, Pull Request that branch to the main repository base. If you continue developing new features in the fork base branch, as per habit, it may get really problematic if you have to deal with complex merge conflicts in future situations.

Sorry was sort of new to open source, will ensure this henceforth

joaomcteixeira: No problem. I understood that. You did a very nice first attempt. :wink: that's why I commented on it.

package/MDAnalysis/analysis/base.py
```diff
@@ -124,6 +124,12 @@ def _setup_frames(self, trajectory, start=None, stop=None, step=None):
...
             number of frames to skip between each analysed frame
+        verbose : bool, optional
+            Turn on verbosity
+            
+        .. versionchanged:: 1.0.0
+            Verbose argument is now passed explicitly from
+            :meth:`AnalysisBase.run`.
         """
         self._trajectory = trajectory
         start, stop, step = trajectory.check_slice_indices(start, stop, step)
```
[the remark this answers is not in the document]
IAlibay: Since this is historical AnalysisBase stuff, @richardjgowers might be best placed to discuss if there were any specific reasons for this (I think it was added in #460).

This behaviour is probably going to need to change in light of #2206, so I'm not sure if it's not best to just tackle all this in a separate PR?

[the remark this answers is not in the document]
jbarnoud: These are valid comments, though these changes are out of scope of this pull request.

package/MDAnalysis/analysis/base.py
jbarnoud: Here `self._verbose` may not be defined, so the `getattr` is necessary. Another possibility would be to wrap the call in a `try`, but it would be more cluttered.

package/MDAnalysis/analysis/base.py
jbarnoud: `start`, `stop`, `step`, and `verbose` are historically argument of the class, though we are deprecating them from the initializer in favour of the same arguments in the `run` method.

testsuite/MDAnalysisTests/analysis/test_base.py
jbarnoud: `not a._verbose` is correct. There is no reason to compare the identity with the `False` singleton; in theory, everything that is considered false by python would be a valid value here.

testsuite/MDAnalysisTests/analysis/test_base.py
jbarnoud: Same as above.

package/MDAnalysis/analysis/base.py
Purva-Chaudhari: > To my understanding, there is no need to use `getattr` here. `getattr` and `setattr` are used on dynamic operations, not on static ones.

should I make it verbose = False if none else verbose ?

package/MDAnalysis/analysis/base.py
joaomcteixeira: Sorry, I was referring to the `self._trajectory`. Because it also gets assigned again on `_setup_frames`. But this is for other PR/issue, indeed.

testsuite/MDAnalysisTests/analysis/test_base.py
joaomcteixeira: Correct. Thanks for the feedback!

package/MDAnalysis/analysis/base.py
joaomcteixeira: isn't `self._verbose` defined in the `__init__`? [link](https://github.com/MDAnalysis/mdanalysis/blob/452655df8cbfd19e4f4dd52e848c8d238481e349/package/MDAnalysis/analysis/base.py#L110)

jbarnoud: It is indeed. My bad.

orbeckst:
@Purva-Chaudhari , PR #2617 (which will be merged soon) makes issue #2504 go away by design so I am sorry to say that this PR that you have been working on will be closed without merging. I understand that you and various mentors have put a lot of effort into it. This would have been your first contribution and in this way you would have qualified for submitting a GSoC proposal.

From what I can see from this PR, we would have eventually merged it as it was close to acceptance. So I am going to say that **we will accept your GSoC proposal submission on the basis of your work seen in this PR #2631** unless you get another PR merged by 3/31/2020 (and we obviously encourage you to do so).

— @orbeckst for @MDAnalysis/gsoc-mentors

Purva-Chaudhari:
@orbeckst Thank you for your acceptance regarding proposal, wanted to ask if I could work on an issue not labelled as gsoc starter. I see many of the gsoc labelled issues already taken. Any suggestions regarding any small issue to get solved please let me know,
Thanking you

PR #2634, opened 2020-03-17
Updated mdamath

### LINKED ISSUE #2632: Non-continuous angles from mdamath.angle
## Expected behavior ##

`mdamath.angle` calculates angles that are continuous in the domain [-1.0, 1.0].

## Actual behavior ##
```python
def angle(a, b):
    """Returns the angle between two vectors in radians

    .. versionchanged:: 0.11.0
       Moved into lib.mdamath
    """
    x = np.dot(a, b) / (norm(a) * norm(b))
    # catch roundoffs that lead to nan otherwise
    if x > 1.0:
        return 0.0
    elif x < -1.0:
        return -np.pi
    return np.arccos(x)
```

`np.arccos` returns values within [0, pi]. This means that `x==-1.0` returns pi, but `x==-1.00001` due to a rounding off error returns -pi. 

``` python
>>> import numpy as np
>>> from MDAnalysis.lib import mdamath
>>> a = np.array([0.1, 0, 0.2])
>>> mdamath.angle(a, -a)
3.141592653589793
>>> mdamath.angle(a**0.5, -(a**0.5))
-3.141592653589793

```

- Which operating system?

## Suggested fix ##

Use `np.clip` instead and leave it to `np.arccos` to return. Probably that code above would suffice as a test.

shfrz:
Fixes #2632

 - Used `np.clip` to fix the rounding off error 
- Added `test_angle_lower_clip` in `test_util.py`

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -22,6 +22,8 @@ mm/dd/yy richardjgowers, kain88-de, lilyminium, p-j-smith, bdice, joaomcteixeira
   * 0.21.0
 
 Fixes
+  * mdamath.angles now returns np.pi instead of -np.pi in cases of
+    lower bound roundoff errors. (Issue #2632, PR #2634)
   * Proper error message for AlignTraj on trajectory without mass (Issue #2469)
   * Updated tests to have explicit fixtures (Issue #2618)
   * XDR offsets now read from trajectory if offsets file read-in fails on
@@ -115,8 +117,6 @@ Enhancements
   * PersistenceLength.plot() now create new axes if current axes not provided (Issue #2590)
 
 Changes
-  * Use `np.clip()` to counter the rounding off error in `mdamath.angles` and 
-    added `test_angle_lower_clip` in `test_util.py` (Issue #2632, PR #2634)
   * Deprecated :class:`ProgressMeter` and replaced it with :class:`ProgressBar` using
     the tqdm package (Issue #928, PR #2617). Also fixes issue #2504.
   * Removed `details` from `ClusteringMethod`s (Issue #2575, PR #2620)
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -115,6 +115,8 @@ Enhancements
   * PersistenceLength.plot() now create new axes if current axes not provided (Issue #2590)
 
 Changes
+  * Use `np.clip()` to counter the rounding off error in `mdamath.angles` and 
+    added `test_angle_lower_clip` in `test_util.py` (Issue #2632, PR #2634)
   * Deprecated :class:`ProgressMeter` and replaced it with :class:`ProgressBar` using
     the tqdm package (Issue #928, PR #2617). Also fixes issue #2504.
   * Removed `details` from `ClusteringMethod`s (Issue #2575, PR #2620)
--- a/package/MDAnalysis/lib/mdamath.py
+++ b/package/MDAnalysis/lib/mdamath.py
@@ -101,10 +101,10 @@ def angle(a, b):
     """
     x = np.dot(a, b) / (norm(a) * norm(b))
     # catch roundoffs that lead to nan otherwise
-    x = np.clip(x,-1.0,1.0)
     if x > 1.0:
         return 0.0
     elif x < -1.0:
+        x = np.clip(x,-1.0,-1.0)      
         return -np.pi
     return np.arccos(x)
 
--- a/package/MDAnalysis/lib/mdamath.py
+++ b/package/MDAnalysis/lib/mdamath.py
@@ -98,6 +98,8 @@ def angle(a, b):
 
     .. versionchanged:: 0.11.0
        Moved into lib.mdamath
+    .. versionchanged:: 1.0.0
+       Added np.clip() to fix the rounding-off error
     """
     x = np.dot(a, b) / (norm(a) * norm(b))
     # catch roundoffs that lead to nan otherwise
--- a/package/MDAnalysis/lib/mdamath.py
+++ b/package/MDAnalysis/lib/mdamath.py
@@ -99,7 +99,8 @@ def angle(a, b):
     .. versionchanged:: 0.11.0
        Moved into lib.mdamath
     .. versionchanged:: 1.0.0
-       Added np.clip() to fix the rounding-off error
+       Changed rounding-off code to use `np.clip()`. Values lower than
+       -1.0 now return `np.pi` instead of `-np.pi`
     """
     x = np.dot(a, b) / (norm(a) * norm(b))
     # catch roundoffs that lead to nan otherwise
--- a/package/MDAnalysis/lib/mdamath.py
+++ b/package/MDAnalysis/lib/mdamath.py
@@ -101,11 +101,7 @@ def angle(a, b):
     """
     x = np.dot(a, b) / (norm(a) * norm(b))
     # catch roundoffs that lead to nan otherwise
-    if x > 1.0:
-        return 0.0
-    elif x < -1.0:
-        x = np.clip(x,-1.0,-1.0)      
-        return -np.pi
+    x = np.clip(x,-1.0,1.0)
     return np.arccos(x)
 
 
--- a/package/MDAnalysis/lib/mdamath.py
+++ b/package/MDAnalysis/lib/mdamath.py
@@ -101,7 +101,7 @@ def angle(a, b):
     """
     x = np.dot(a, b) / (norm(a) * norm(b))
     # catch roundoffs that lead to nan otherwise
-    x = np.clip(x,-1.0,1.0)
+    x = np.clip(x, -1.0, 1.0)
     return np.arccos(x)
 
 
--- a/package/MDAnalysis/lib/mdamath.py
+++ b/package/MDAnalysis/lib/mdamath.py
@@ -101,6 +101,7 @@ def angle(a, b):
     """
     x = np.dot(a, b) / (norm(a) * norm(b))
     # catch roundoffs that lead to nan otherwise
+    x = np.clip(x,-1.0,1.0)
     if x > 1.0:
         return 0.0
     elif x < -1.0:
--- a/testsuite/MDAnalysisTests/lib/test_util.py
+++ b/testsuite/MDAnalysisTests/lib/test_util.py
@@ -221,9 +221,8 @@ class TestGeometryFunctions(object):
         assert_equal(mdamath.normal(vec1, vec2), value)
         # add more non-trivial tests
 
-    @pytest.fixture()
-    def test_angle(self, a, b):
-      a = np.array([0.1, 0, 0.2] ** 0.5)
+    def test_angle_lower_clip(self, a, b, x):
+      a = np.array([0.1, 0, 0.2]) ** 0.5
       b = -a
       assert x < -1.0
       assert mdamath.angle(a, b) == np.pi
--- a/testsuite/MDAnalysisTests/lib/test_util.py
+++ b/testsuite/MDAnalysisTests/lib/test_util.py
@@ -222,9 +222,11 @@ class TestGeometryFunctions(object):
         # add more non-trivial tests
 
     @pytest.fixture()
-    def test_angle(self, a):
+    def test_angle(self, a, b):
       a = np.array([0.1, 0, 0.2])
-      assert mdamath.angle(a, -a) == mdamath.angle(a**0.5, -(a**0.5))
+      b = -a
+      assert x < -1.0
+      assert mdamath.angle(a, b) == np.pi
             
     def test_stp(self):
         assert mdamath.stp(self.e1, self.e2, self.e3) == 1.0
--- a/testsuite/MDAnalysisTests/lib/test_util.py
+++ b/testsuite/MDAnalysisTests/lib/test_util.py
@@ -221,6 +221,10 @@ class TestGeometryFunctions(object):
         assert_equal(mdamath.normal(vec1, vec2), value)
         # add more non-trivial tests
 
+    @pytest.mark.parametrize('a', np.array([0.1, 0, 0.2]))
+    def test_angle(self, a):
+    	assert_equal(mdamath.angle(a, -a), mdamath.angle(a**0.5, -(a**0.5)))
+            
     def test_stp(self):
         assert mdamath.stp(self.e1, self.e2, self.e3) == 1.0
         # add more non-trivial tests
--- a/testsuite/MDAnalysisTests/lib/test_util.py
+++ b/testsuite/MDAnalysisTests/lib/test_util.py
@@ -221,8 +221,8 @@ class TestGeometryFunctions(object):
         assert_equal(mdamath.normal(vec1, vec2), value)
         # add more non-trivial tests
 
-    @pytest.mark.parametrize('a', np.array([0.1, 0, 0.2]))
     def test_angle(self, a):
+      a = np.array([0.1, 0, 0.2])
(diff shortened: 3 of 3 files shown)


package/MDAnalysis/lib/mdamath.py
```diff
@@ -101,6 +101,7 @@ def angle(a, b):
     """
     x = np.dot(a, b) / (norm(a) * norm(b))
     # catch roundoffs that lead to nan otherwise
+    x = np.clip(x,-1.0,1.0)
```
richardjgowers: if we're clipping here then the if conditions below can never be triggered?

shfrz: Thanks @richardjgowers 
A silly mistake indeed! /

package/MDAnalysis/lib/mdamath.py
```diff
@@ -104,6 +104,7 @@ def angle(a, b):
     if x > 1.0:
         return 0.0
     elif x < -1.0:
+        x = np.clip(x,-1.0,-1.0)
```
IAlibay: @shfrz if I think @richardjgowers' point here was that the `if`/`else` construct should just go away if we use `np.clip`. The way you have now changed the code is such that this `np.clip` call will have no effect on solving issue #2632.

shfrz: @IAlibay , if I am getting it right  `x==-1.0001` would become `-1.0` which works fine?

Thanks!

IAlibay: @shfrz Unless I'm missing something, by the time you've entered this part of the `if`/`else` construct, either way you'll return `-np.pi`?

shfrz: @IAlibay , so the right approach would be to remove the `if`/`else` block and set `x = np.clip(x,-1.0,1.0)?

IAlibay: Yes, at least if we are going for a clip option, I believe that was @richardjgowers' original suggestion.

shfrz:
@lilyminium, please have a look :)

orbeckst:
@fiona-naughton would you mind also keeping an eye on this PR and helping out @lilyminium who already has quite a few on her plate.  Thanks!

shfrz:
@lilyminium, I don't see any direct tests for mdamath in MDAnalysisTests?

IAlibay:
> @lilyminium, I don't see any direct tests for mdamath in MDAnalysisTests?

@shfrz please see under:

https://github.com/MDAnalysis/mdanalysis/blob/4102481643dd59cf40db382a190f8b35ece9b831/testsuite/MDAnalysisTests/lib/test_util.py#L168

testsuite/MDAnalysisTests/lib/test_util.py
```diff
@@ -221,6 +221,10 @@ def test_normal(self, vec1, vec2, value):
         assert_equal(mdamath.normal(vec1, vec2), value)
         # add more non-trivial tests
 
+    @pytest.mark.parametrize('a', np.array([0.1, 0, 0.2]))
+    def test_angle(self, a):
+    	assert_equal(mdamath.angle(a, -a), mdamath.angle(a**0.5, -(a**0.5)))
```
lilyminium: Why are you using `parametrize` here? It turns `a` into the scalars `0.1`, `0`, and `0.2` sequentially. Does that test what you want to test? Does comparing the angle between `(a, -a)` and `(a**0.5, -(a**0.5))` test your fix, i.e. that when `np.dot(a, b) / (norm(a) * norm(b)) < -1.0`, `angle()` still returns `pi`?

Note: I specifically used the vector `[0.1, 0, 0.2] ** 0.5` in the issue because it [takes advantage of floating point imprecision](https://medium.com/better-programming/why-is-0-1-0-2-not-equal-to-0-3-in-most-programming-languages-99432310d476).

shfrz: @lilyminium , that article was helpful!
Thanks :)

shfrz:
@lilyminium, could you please elaborate a bit, it seems a little unclear to me

lilyminium:
In #2632 the problem was that when:

```python
x = np.dot(a, b) / (norm(a) * norm(b)) < -1.0
```

the function `mdamath.angle` would return `-np.pi`, when in fact, it should return `np.pi`.

To test your fix, you should check that this scenario no longer occurs. The best way to do that is to use a vector that you *know* will result in `x < -1.0`, check that with `assert x < -1.0`, and then check that `mdamath.angle` nonetheless returns pi.

shfrz:
@lilyminium , in #2632 "Probably that code above would suffice as a test." What was that for? 
If we get `mdamath.angle(a, -a) == mdamath.angle(a**0.5, -(a**0.5))` probably that would suffice for the test ?

lilyminium:
> On 24 Mar 2020, at 1:19 AM, shfrz <notifications@github.com> wrote:
> 
> 
> @lilyminium <https://github.com/lilyminium> , in #2632 <https://github.com/MDAnalysis/mdanalysis/issues/2632> "Probably that code above would suffice as a test." What was that for?
> 

Well firstly, the code never directly compares  mdamath.angle(a, -a) == mdamath.angle(a**0.5, -(a**0.5)). More importantly, sure, that was poor phrasing. I often make mistakes, am wrong, or people will disagree with me. For this to be your contribution to MDAnalysis, you must take ownership of your code and be able to explain why you believe your test works. 

> If we get mdamath.angle(a, -a) == mdamath.angle(a**0.5, -(a**0.5)) probably that would suffice for the test ?
> 
> 

Well, your earlier test passed, despite comparing a when a is 0.1, 0, and 0.2 separately. This is clearly does not address the original issue. Does x ever < -0.1 when a is 0.1 or 0? Moreover, future readers may not be aware that the vector [0.1, 0, 0.2]**0.5 takes advantage of floating point imprecision. They may not be aware that x < -0.1 when you use that vector and its negative. An assert statement makes it clear what scenario you are testing.

I also want to say that the vector [0.1, 0, 0.2] is irrelevant for this example, which may not have been clear if you do not have a background in linear algebra. I could have easily used [1, 0, 0]. I used  [0.1, 0, 0.2] as an easy way to contrast its behaviour to its square root. The statement mdamath.angle(a, -a) == mdamath.angle(a**0.5, -(a**0.5)) is trivial and meaningless because it all reduces down to the unit vector anyway.

shfrz:
@lilyminium , thanks for your invaluable feedback and suggestions! I appreciate your time.

So the problems with my earlier code was that it didn't specifically take into consideration the cases where x < -0.1, it took  unit vectors sequentially and and the future readers wouldn't know about x being less than -0.1...?

Thanks again!

lilyminium:
> So the problem with my earlier code was that it didn't specifically take into consideration the cases where x < -0.1

The 0.1 was a typo, but yes. If your fix is for the scenario where x < -1.0, your test must specifically test for the scenario where x < -1.0. 

>  it took unit vectors sequentially

[``parametrize`` takes a list of arguments and feeds it into tests as parameters.](https://docs.pytest.org/en/latest/parametrize.html) When you used it, it split your vector into scalar values: a=0.1, then a=0, then a=0.2. None of these are [unit vectors](https://www.mathsisfun.com/algebra/vector-unit.html), nor [0.1, 0, 0.2] itself. None of this is inherently a problem, except that it doesn't test what you want it to test, as elaborated on below.

> the future readers wouldn't know about x being less than -0.1...?

Software unit tests [can be considered a kind of specification](https://python-3-patterns-idioms-test.readthedocs.io/en/latest/UnitTesting.html#write-tests-first). Often users can look at tests to figure out what code is *supposed* to be doing. If they looked at a test that compared `mdamath.angle(a, -a) == mdamath.angle(a**0.5, -(a**0.5))`, that tells them that the angle between opposite vectors is supposed to be the same as the angle between the same opposite vectors, but square-rooted. As mentioned before, this is a tautology. However, if they looked at a test that asserted `x < -1.0` and then `mdamath.angle(a, -a) == np.pi`, *that* tells them that in the rare case of roundoff errors, the angle returned is still pi.

lilyminium:
@fiona-naughton can you take over this issue please -- I won't be able to spend time on MDAnalysis for the next few days, global pandemic, etc. @shfrz already has an issue merged so no rush.

testsuite/MDAnalysisTests/lib/test_util.py
```diff
@@ -221,6 +221,13 @@ def test_normal(self, vec1, vec2, value):
         assert_equal(mdamath.normal(vec1, vec2), value)
         # add more non-trivial tests
 
+    @pytest.fixture()
```
fiona-naughton: The point of `@pytest.fixture()` is to set up a function that is run before any test function we pass it to. This is unnecessary here (and since you're not passing it anywhere this test is currently not run, which is why the issues below weren't picked up).

testsuite/MDAnalysisTests/lib/test_util.py
```diff
@@ -221,6 +221,13 @@ def test_normal(self, vec1, vec2, value):
...
+    def test_angle(self, a, b):
```
fiona-naughton: I'd add to the test name to reflect that you're testing a specific edgecase, just to help make things clearer for anyone reading in the future - say something like `test_angle_lower_clip()`

testsuite/MDAnalysisTests/lib/test_util.py
```diff
@@ -221,6 +221,13 @@ def test_normal(self, vec1, vec2, value):
...
+      a = np.array([0.1, 0, 0.2] ** 0.5)
```
fiona-naughton: `**` won't work on lists; you can apply it to the array instead

testsuite/MDAnalysisTests/lib/test_util.py
```diff
@@ -221,6 +221,13 @@ def test_normal(self, vec1, vec2, value):
...
+      b = -a
+      assert x < -1.0
```
fiona-naughton: You need to actually define `x` first

testsuite/MDAnalysisTests/lib/test_util.py
shfrz: @fiona-naughton , thanks for the review!

shfrz: after removing `@pytest.fixture()`, and running tests locally, `pytest` prompts `fixture a not found`, how to deal with that?
Should we use `@pytest.mark.parametrize()` instead?

IAlibay: @shfrz I think the problem here is that you are passing `a` and `b` as arguments to `test_angle` when they are just variables defined within the function. So pytest is probably getting confused expecting `a` and `b` to be fixtures. Try just `def test_angle(self)` and see if that works.

shfrz: @IAlibay, I tried your suggestion and it worked fine for that part but it seems even after clipping, `mdamath.angle(a**0.5, -(a**0.5))`, returns -np.pi 

Capturehbh

testsuite/MDAnalysisTests/lib/test_util.py
```diff
@@ -221,6 +221,13 @@ def test_normal(self, vec1, vec2, value):
         assert_equal(mdamath.normal(vec1, vec2), value)
         # add more non-trivial tests
 
+    def test_angle_lower_clip(self):
+    	a = np.array([0.1, 0, 0.2])
```
IAlibay: Travis is picking up some tabs instead of spaces on some of these lines. Could you make sure that you are solely using tabs?

package/MDAnalysis/lib/mdamath.py
```diff
@@ -101,10 +101,7 @@ def angle(a, b):
     """
     x = np.dot(a, b) / (norm(a) * norm(b))
     # catch roundoffs that lead to nan otherwise
-    if x > 1.0:
-        return 0.0
-    elif x < -1.0:
-        return -np.pi
+    x = np.clip(x,-1.0,1.0)
```
IAlibay: PEP8, please add spaces between your function arguments (i.e. `np.clip(x, -1.0, 1.0)`).

testsuite/MDAnalysisTests/lib/test_util.py
orbeckst: I hope you mean "spaces" and not "tabs" – please never, ever use tabs in Python code!

IAlibay: Sorry yes, that is a very very bad typo. Spaces not tabs!

shfrz:
@IAlibay , we were almost done. What needs to be done now?

IAlibay:
> @IAlibay , we were almost done. What needs to be done now?

@shfrz I'm not sure what happened to your git commits, but somehow we went from ~ dozen line changes & 2 files changed, to +438/-1508 lines & 34 files changes. A lot of the changes are also reverting recent commits on the develop branch (including changes to tdqm). I think it went wrong around 8341dbf could you make sure that your branch is even with develop (aside from the changes to mdamath & associated tests)?

lilyminium:
@shfrz you'll want to reset back to just after @IAlibay's review and go from there.

```bash
git reset --hard d883a3a
git push -f origin issue-2632
```

I'm not sure what PR's you're merging from MDAnalysis/develop. If you're trying to update your branch to `develop`, [rebasing will create a cleaner commit history](https://www.mdanalysis.org/UserGuide/contributing_code.html#rebasing-your-code).

shfrz:
@IAlibay , are we ready for merger :)

shfrz:
Thanks @fiona-naughton  :+1:

package/MDAnalysis/lib/mdamath.py
```diff
@@ -101,10 +101,7 @@ def angle(a, b):
     """
```
IAlibay: This needs a `.. verionschanged:: 1.0.0` describing what has changed.

package/MDAnalysis/lib/mdamath.py
```diff
@@ -98,13 +98,12 @@ def angle(a, b):
 
     .. versionchanged:: 0.11.0
        Moved into lib.mdamath
+    .. versionchanged:: 1.0.0
+       Added np.clip() to fix the rounding-off error
```
IAlibay: Rather than added, maybe this should be "Changed rounding-off code to use `np.clip()`. Values lower than -1.0 now return `np.pi` instead of `-np.pi`".

The latter part is possibly important in case anyone was using the `-np.pi` return value downstream and suddenly finds that things no longer work as intended.

package/CHANGELOG
```diff
@@ -115,6 +115,8 @@ Enhancements
   * PersistenceLength.plot() now create new axes if current axes not provided (Issue #2590)
 
 Changes
+  * Use `np.clip()` to counter the rounding off error in `mdamath.angles` and
```
IAlibay: Just noticed this, 
a) The main change here is that you now enforce a return of `np.pi` rather than `-np.pi` when x is less than -1.0. That's probably what you want to detail here, i.e. "`mdamath.angles` now returns `np.pi` instead of `-np.pi` in cases of lower bound roundoff errors."

b) this probably should be under "Fixes" rather than "Changes".

c) as far as I remember we don't usually detail new tests in the changelog, soI'd probably just end the entry at `mdamath.angles` (followed by the Issue/PR number) as your entry is sufficiently decriptive 😃

shfrz:
Thank you @lilyminium , @fiona-naughton and @IAlibay 👍

shfrz:
Merge? :)

PR #2636, opened 2020-03-20
Make NoDataError subclass of AttributeError and add informative error messages

### LINKED ISSUE #2565: Improve missing AtomGroup property error messages

### LINKED ISSUE #2635: Testing for fragment attribute
## Expected behavior ##
Checking for the `fragment` attribute 

```python
from MDAnalysisTests.core.util import UnWrapUniverse
u = UnWrapUniverse(have_molnums=False, have_bonds=False)

print(hasattr(u,"fragments"))
print(hasattr(u.atoms, "fragments"))
```

should give the following consistent output

```python
False
False
```

## Actual behavior ##

However, while the first works fine and returns `False` the second raises a `NoDataError`.

```python
/usr/local/lib/python3.7/site-packages/MDAnalysis/core/groups.py in __getattr__(self, attr)
   2222             # raise NDE(_ATTR_ERRORS[attr])
   2223             raise NoDataError("AtomGroup.{} not available; this requires Bonds"
-> 2224                               "".format(attr))
   2225         raise AttributeError("{cls} has no attribute {attr}".format(
   2226             cls=self.__class__.__name__, attr=attr))

NoDataError: AtomGroup.fragments not available; this requires Bonds
```

- Which version are you using? 0.20.2-dev0
- Which version of Python: 3.7

lilyminium:
Fixes #2635
Fixes #2565 
Mitigates https://github.com/MDAnalysis/mdanalysis/issues/2649#issuecomment-605176500

 - added more informative error messages for topology attributes
 - made NoDataError subclass of AttributeError

```python
>>> import MDAnalysis as mda
>>> from MDAnalysis.tests.datafiles import GRO
>>> u = mda.Universe(GRO)
>>>
>>> u.atoms.altlocs
  File "mdanalysis/package/MDAnalysis/core/groups.py", line 2285, in __getattr__
    return super(AtomGroup, self).__getattr__(attr)
  File "mdanalysis/package/MDAnalysis/core/groups.py", line 560, in __getattr__
    return super(GroupBase, self).__getattr__(attr)
  File "mdanalysis/package/MDAnalysis/core/groups.py", line 385, in __getattr__
    raise AttributeError(err)
AttributeError: AtomGroup has no attribute altlocs. Did you mean altLocs?
>>>
>>> u.atoms[0].resnames
  File "mdanalysis/package/MDAnalysis/core/groups.py", line 3701, in __getattr__
    return super(Atom, self).__getattr__(attr)
  File "mdanalysis/package/MDAnalysis/core/groups.py", line 3576, in __getattr__
    singular=cls.singular))
AttributeError: Atom has no attribute resnames. Do you mean resname?
>>>
>>> u.atoms.unwrap()
  File "mdanalysis/package/MDAnalysis/core/groups.py", line 1643, in unwrap
    "".format(self.__class__.__name__))
MDAnalysis.exceptions.NoDataError: AtomGroup.unwrap() not available; this requires Bonds
>>>
>>> u.residues.bonds
  File "mdanalysis/package/MDAnalysis/core/groups.py", line 558, in __getattr__
    raise NoDataError(err.format(singular=cls.singular))
MDAnalysis.exceptions.NoDataError: This Universe does not contain bonds information
>>>
>>> u.atoms[0].total_charge()
  File "mdanalysis/package/MDAnalysis/core/groups.py", line 3705, in __getattr__
    return super(Atom, self).__getattr__(attr)
  File "mdanalysis/package/MDAnalysis/core/groups.py", line 3585, in __getattr__
    return super(ComponentBase, self).__getattr__(attr)
  File "mdanalysis/package/MDAnalysis/core/groups.py", line 373, in __getattr__
    selfcls=selfcls))
AttributeError: total_charge() is a method of AtomGroup, not Atom
```

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -23,6 +23,8 @@ mm/dd/yy richardjgowers, kain88-de, lilyminium, p-j-smith, bdice, joaomcteixeira
   * 0.21.0
 
 Fixes
+  * Added more informative error messages about topology attributes
+    (Issue #2565)
   * Made NoDataError a subclass of ValueError *and* AttributeError
     (Issue #2635)
   * Fixed select_atoms("around 0.0 ...") selections and capped_distance 
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -23,9 +23,10 @@ mm/dd/yy richardjgowers, kain88-de, lilyminium, p-j-smith, bdice, joaomcteixeira
   * 0.21.0
 
 Fixes
+  * Made NoDataError a subclass of ValueError *and* AttributeError
+    (Issue #2635)
   * Fixed select_atoms("around 0.0 ...") selections and capped_distance 
-    causing a segfault
-    (Issue #2656 PR #2665)
+    causing a segfault (Issue #2656 PR #2665)
   * `PDBWriter` writes unitary `CRYST1` record (cubic box with sides of 1 Å)
     when `u.dimensions` is `None` or `np.zeros(6)` (Issue #2679, PR #2685)
   * Ensures principal_axes() returns axes with the right hand convention (Issue #2637)
--- a/package/MDAnalysis/__init__.py
+++ b/package/MDAnalysis/__init__.py
@@ -177,7 +177,10 @@ _PARSER_HINTS = {}
 _SELECTION_WRITERS = {}
 _CONVERTERS = {}
 # Registry of TopologyAttributes
-_TOPOLOGY_ATTRS = {}
+_TOPOLOGY_ATTRS = {}  # {attrname: cls}
+_TOPOLOGY_TRANSPLANTS = {}   # {name: [attrname, method, transplant class]}
+_TOPOLOGY_ATTRNAMES = {}  # {lower case name w/o _ : name}
+
 
 # Storing anchor universes for unpickling groups
 import weakref
--- a/package/MDAnalysis/coordinates/base.py
+++ b/package/MDAnalysis/coordinates/base.py
@@ -446,6 +446,14 @@ class Timestep(object):
             return self._pos[atoms]
         else:
             raise TypeError
+    
+    def __getattr__(self, attr):
+        # special-case timestep info
+        if attr in ('velocities', 'forces', 'positions'):
+            raise NoDataError('This Timestep has no ' + attr)
+        err = "{selfcls} object has no attribute '{attr}'"
+        raise AttributeError(err.format(selfcls=type(self).__name__,
+                                        attr=attr))
 
     def __len__(self):
         return self.n_atoms
--- a/package/MDAnalysis/core/groups.py
+++ b/package/MDAnalysis/core/groups.py
@@ -350,6 +350,7 @@ class _MutableBase(object):
 
     def __getattr__(self, attr):
         selfcls = type(self).__name__
+
         if attr in _TOPOLOGY_TRANSPLANTS:
             topattr, meth, clstype = _TOPOLOGY_TRANSPLANTS[attr]
             if isinstance(meth, property):
@@ -363,8 +364,12 @@ class _MutableBase(object):
             if not isinstance(self, clstype):
                 mname = 'property' if isinstance(meth, property) else 'method'
                 err = '{attr} is a {method} of {clstype}, not {selfcls}'
-                raise AttributeError(err.format(attr=attr, method=attrtype,
-                                                clstype=clstype.__name__,
+                clsname = clstype.__name__
+                if clsname == 'GroupBase':
+                    clsname = selfcls + 'Group'
+                raise AttributeError(err.format(attr=attrname,
+                                                method=attrtype,
+                                                clstype=clsname,
                                                 selfcls=selfcls))
             # missing required topologyattr
             else:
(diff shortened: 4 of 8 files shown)


richardjgowers:
Ok I'm confused by this, what did this break exactly? It seems like a good idea.

lilyminium:
@richardjgowers Making NoDataError a subclass of AttributeError (instead of just making a new error, as in this iteration of the PR) broke because of how `__getattribute__` handles `AttributeError`s. 

A few property getters e.g. ``Atom.velocity`` raise `NoDataError`s if there is missing information. When `NoDataError` is just a `ValueError` it behaves normally and gives the user the error message. When `NoDataError` is also an `AttributeError`, `__getattribute__` sends the original key to `__getattr__`, completely ignoring the original traceback. So originally:

```python
>>> u.atoms.velocities
  File "mdanalysis/package/MDAnalysis/core/groups.py", line 2454, in velocities
    raise_from(NoDataError("Timestep does not contain velocities"), None)
  File "<string>", line 3, in raise_from
MDAnalysis.exceptions.NoDataError: Timestep does not contain velocities
```
When NoDataError is subclass of AttributeError:

```python
>>> u.atoms.velocities
  File "mdanalysis/package/MDAnalysis/core/groups.py", line 2226, in __getattr__
    cls=self.__class__.__name__, attr=attr))
AttributeError: AtomGroup has no attribute velocities
```

This is because currently `__getattr__` only checks a very limited number of known attributes to return a NoDataError. Since a lot of the code base relies on catching NoDataErrors rather than AttributeErrors, it broke a lot of downstream stuff like writing out files. So, it could be a workable solution if #2565 is fixed.

zemanj:
Yeah I also think that making `NoDataError` a subclass of `AttributeError` in conjunction with what's to be done in #2565 is the way to go. FWIW, that should also unify Py27/3x behavior (IIRC `hasattr()` swallows *any* exception in `__getattr__()`  in Py27).

lilyminium:
> (IIRC hasattr() swallows any exception in __getattr__() in Py27).

Yep that's right.

Alrighty, I'll put this on hold and add in #2565 stuff if no one fixes that by the end of the GSOC application period?

zemanj:
> Alrighty, I'll put this on hold and add in #2565 stuff if no one fixes that by the end of the GSOC application period?

Awesome! :+1:

lilyminium:
Sorry @IAlibay, misclick while I was trying to add the Work In Progress label...

lilyminium:
@richardjgowers I think so; I did add some extra stuff beyond the original issue (e.g. suggesting when methods are AtomGroup vs Atom or checking for capitalisation and plurals), so just checking that that's both welcome and doesn't overly affect performance?

richardjgowers:
From what I can tell it’s only performance heavy when things fail, and that’s to give a better error message.

PR #2638, opened 2020-03-21
Fixed mdamath

### LINKED ISSUE #2632: Non-continuous angles from mdamath.angle

Oscuro-Phoenix:
Fixes #2632

 - 
I have fixed mdamath using np.clip( ). 

Testing added.

Thanks.

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


IAlibay:
@Oscuro-Phoenix unless I'm missing something, a PR fixing this issue already exists (#2634). Unless this PR stalls, it will be given priority instead of this one.

IAlibay:
Closing this PR as #2634 has now been merged.

PR #2645, opened 2020-03-25
Added warning for 1 A^3 CRYST1 Record 

### LINKED ISSUE #2599: 1 A^3 boxes from EM structures in the PDB
## Context ##

This is related to this discussion: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/mdnalysis-discussion/IKYv18MqlFs

From a cursory look at a couple of systems (6NT5, 6CVM, 3JAC), it looks like PDB files for EM structures have a tendency to have the following CRYST1 records:

`CRYST1    1.000    1.000    1.000  90.00  90.00  90.00 P 1           1`

As dicussed in the dicussion thread, this is causing issues with periodic selections as it is folding the system into a cube of 1 A.

## Proposed solution ##

My solution here would be to catch cases where the CRYST1 record is defining a 1 A cube, warn the user and make the system dimensionless.

Very slightly altered code from Lucas Siemons' post.

``` python
import MDAnalysis as mda
import numpy as np

u = mda.Universe('6cvm.pdb')

cur_sele = u.select_atoms('around 0.1 (resid 4 and name CA and segid A)')

ref_ca = u.select_atoms('resid 4 and name CA and segid A')

print('sele len: ', len(cur_sele))
for i in cur_sele:
    dist = np.linalg.norm(i.position-ref_ca[0].position)
    print('Distance: %f '%(dist), i)
```

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) 0.20.2-dev0
- Which version of Python (`python -V`)? 3.7.6
- Which operating system? Ubuntu 18.04.3 LTS

Oscuro-Phoenix:
Fixes #2599

 - Added a warning for 1 A^3 CRYST1 Record 

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


Oscuro-Phoenix:
Hi!,
Is this what we need? Do let me know if any changes are needed. 
Thanks.

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -398,6 +398,8 @@ def _read_frame(self, frame):
                     warnings.warn("Failed to read CRYST1 record, "
                                   "possibly invalid PDB file, got:\n{}"
                                   "".format(line))
+                if (self.ts._unitcell[:3] == np.array([1,1,1],dtype=float)).all():
```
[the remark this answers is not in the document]
Oscuro-Phoenix: ls this somewhat closer to what you're asking?
               `try:
                    cell_check = [line[6:15], line[15:24],
                                  line[24:33], line[33:40],
                                  line[40:47], line[47:54]]
                    if (np.array(cell_check,dtype=float) == 
                        np.array([1,1,1,90,90,90],dtype=float)).all():
                                      "possibly an EM structure file, setting box-size to 0",stacklevel=2)
                        self.ts._unitcell=np.array([0,0,0,90,90,90],dtype=float)    
                except ValueError:
                    warnings.warn("Failed to read CRYST1 record, "
                                  "possibly invalid PDB file, got:\n{}"
                                  "".format(line))
                else:
                    try:
                        if not (self.ts._unitcell==np.array([0,0,0,90,90,90])).all():
                            self.ts._unitcell[:] = [line[6:15], line[15:24],
                                                    line[24:33], line[33:40],
                                                    line[40:47], line[47:54]]
                    except ValueError:
                        warnings.warn("Failed to read CRYST1 record, "
                                      "possibly invalid PDB file, got:\n{}"
                                      "".format(line))`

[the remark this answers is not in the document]
Oscuro-Phoenix: > Is this what we'd like?

                try:
                    cell_check = [line[6:15], line[15:24],
                                  line[24:33], line[33:40],
                                  line[40:47], line[47:54]]
                    if (np.array(cell_check,dtype=float) == 
                        np.array([1,1,1,90,90,90],dtype=float)).all():
                                      "possibly an EM structure file, setting box-size to 0",stacklevel=2)
                        self.ts._unitcell=np.array([0,0,0,90,90,90],dtype=float)    
                except ValueError:
                    warnings.warn("Failed to read CRYST1 record, "
                                  "possibly invalid PDB file, got:\n{}"
                                  "".format(line))
                else:
                    try:
                        if not (self.ts._unitcell==np.array([0,0,0,90,90,90])).all():
                            self.ts._unitcell[:] = [line[6:15], line[15:24],
                                                    line[24:33], line[33:40],
                                                    line[40:47], line[47:54]]
                    except ValueError:
                        warnings.warn("Failed to read CRYST1 record, "
                                      "possibly invalid PDB file, got:\n{}"
                                      "".format(line))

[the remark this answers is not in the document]
IAlibay: Just a general point to start with. It is usually easier for us to review if you can push larger code changes to you branch rather than place it in the comments. That way we can get a better feeling for how the code changes fit within the context of the existing code. It also makes it easier for you to read our comments afterwards 😄 

Whilst you have most of the right core elements, I don't think the structure works.

#### `try` block:

What you want to do here is try to extract the lines into a temporary variable (here `cell_dims` might be a more understandable variable name than `cell_check`). There's no point testing/warning for cryo-em default values here because you are already doing this in the `else` block (you don't want to use up extra CPU cycles if you can avoid it).

#### `except` block:

This seems fine.

#### `else` block:

The `else` block is where I'm having the hardest time following what you are doing.
Since you will have already extracted the cell dimensions in the `try` block, there is no point trying to extract them again. Additionally, you have already tested if a `ValueError` will be raised when reading the appropriate parts of `line`. So you don't need the addtiional `try`/`except` here.

In this part of the code I would expect you to be testing the contents of the temporary variable hosting the cell dimensions against the default cryo-em CRYST1 values (i.e. [1,1,1,90,90,90]). If it matches (i.e. is a likely cryo-em structure), then raise a warning. If not then feed the cell dimensions to `self.ts._unitcell`.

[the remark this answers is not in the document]
Oscuro-Phoenix: Thank you so much for taking this. I will keep your points in mind when I revert later.

What I wanted to do is basically have the `cell_dims `variable store the box values which are checked for `CRYST1` values in the first try block. If they turn out to be the same as the specific set of values [1,1,1,90,90,90] then I set the `self.ts._unitcell` as having dimensions of [0,0,0,90,90,90].  And in else block, if the `self.ts._unitcell `is not set as having box-size of 0 then only I go ahead and set it as the cell_dims (again, I should have directly set it as =`cell_dims`). I see that the last try/except was extraneous. I'm really sorry for that.

IAlibay:
@Oscuro-Phoenix not had a look at the code yet, but adding & parsing a 38k line (cryo-em files are indeed quite large 😅 ) file to test a CRYST1 record might a bit overkill. I'd instead recommend looking at using the much smaller [5A7U](http://www.rcsb.org/structure/5A7U).

testsuite/MDAnalysisTests/datafiles.py
```diff
@@ -509,5 +510,7 @@
 ITP_no_endif = resource_filename(__name__, 'data/no_endif_spc.itp')
 
 NAMDBIN = resource_filename(__name__, 'data/adk_open.coor')
+
+CRYST1_BOX = resource_filename(__name__, 'data/6cvm.pdb')
```
IAlibay: So it might be best to add this with the rest of the `PDB` named filed. Also it might be pedantic but I would instead consider using something like `PDB_CRYOEM_BOX` instead of `CRYST1_BOX`. The former will help future developers understand why the file was included (and how it can be re-used).

testsuite/MDAnalysisTests/coordinates/test_pdb.py
```diff
@@ -961,3 +961,10 @@ def test_partially_missing_cryst():
     assert len(u.atoms) == 3
     assert len(u.trajectory) == 2
     assert_array_almost_equal(u.dimensions, 0.0)
+
+def test_CRYST1_Warning():
```
IAlibay: To keep with the style of the other functions, I would suggest renaming this to a fully lowercase function name. Also, there are two warnings that could happen with the CRYST1 record, so I would suggest adding the word `em` in there somewhere.

testsuite/MDAnalysisTests/coordinates/test_pdb.py
```diff
@@ -961,3 +961,10 @@ def test_partially_missing_cryst():
...
+    #issue 2599  
+    with pytest.warns(UserWarning) as record:
+        u = mda.Universe(CRYST1_BOX)
+        cur_sele = u.select_atoms('around 0.1 (resid 4 and name CA and segid A)')
```
IAlibay: You _shouldn't_ need to call `select_atoms` to get the warning here (it should hopefully just be thrown on universe building (worth checking that's the case here). So the `select_atoms` shouldn't be necessary here. However, we definitely should be checking that selecting 0.1 A around one of the alpha carbons works now, so do please add this as a second test.

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -391,14 +391,22 @@ def _read_frame(self, frame):
             elif line[:6] == 'CRYST1':
                 # does an implicit str -> float conversion
                 try:
-                    self.ts._unitcell[:] = [line[6:15], line[15:24],
-                                            line[24:33], line[33:40],
-                                            line[40:47], line[47:54]]
+                    cell_dims = [line[6:15], line[15:24],
+                                  line[24:33], line[33:40],
+                                  line[40:47], line[47:54]]
+                    if (np.array(cell_dims,dtype=float) ==
```
IAlibay: PEP8 you'll need spaces in between the `,` in this construct.

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -391,14 +391,22 @@ def _read_frame(self, frame):
             elif line[:6] == 'CRYST1':
                 # does an implicit str -> float conversion
                 try:
-                    self.ts._unitcell[:] = [line[6:15], line[15:24],
-                                            line[24:33], line[33:40],
-                                            line[40:47], line[47:54]]
+                    cell_dims = [line[6:15], line[15:24],
```
IAlibay: As per my previous comment, you only want the `cell_sim` extraction in the `try` block. The reason for not adding the rest is that you want to limit the amount of operations you do here so that your `try` is specifically encompassing the operation you are expecting to fail.

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -391,14 +391,22 @@ def _read_frame(self, frame):
...
+                                  line[24:33], line[33:40],
+                                  line[40:47], line[47:54]]
+                    if (np.array(cell_dims,dtype=float) == 
+                        np.array([1,1,1,90,90,90],dtype=float)).all():
+                        warnings.warn("1 A^3 CRYST1 record," 
+                                      "possibly an EM structure file, setting box-size to 0",stacklevel=2)
+                        self.ts._unitcell=np.array([0,0,0,90,90,90],dtype=float)
```
IAlibay: Apologies, I think my previous explanation wasn't great. Here I would be expecting that `self.ts._unitcell` is not assigned at all (in the same way that you would if you triggered the `except` block).

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -391,14 +391,22 @@ def _read_frame(self, frame):
...
+                        warnings.warn("1 A^3 CRYST1 record," 
+                                      "possibly an EM structure file, setting box-size to 0",stacklevel=2)
+                        self.ts._unitcell=np.array([0,0,0,90,90,90],dtype=float)    
                 except ValueError:
                     warnings.warn("Failed to read CRYST1 record, "
                                   "possibly invalid PDB file, got:\n{}"
                                   "".format(line))
-
+                else:
+                    if not (self.ts._unitcell==np.array([0,0,0,90,90,90])).all():
```
IAlibay: See above, I would be expecting most of what is currently in `try` (except assigning `cell_dims`) to be here.

Oscuro-Phoenix: Do let me know if it has been done the right way in my latest commit.

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -391,14 +391,21 @@ def _read_frame(self, frame):
...
-                                            line[24:33], line[33:40],
-                                            line[40:47], line[47:54]]
+                    cell_dims = [line[6:15], line[15:24],
+                                  line[24:33], line[33:40],
+                                  line[40:47], line[47:54]]
+                    if (np.array(cell_dims, dtype=float) == 
+                        np.array([1, 1, 1, 90, 90, 90], dtype=float)).all():
+                        warnings.warn("1 A^3 CRYST1 record," 
+                                      " possibly an EM structure file, setting box-size to 0",stacklevel=2)
+                        cell_dims = np.array([0, 0, 0, 90, 90, 90],dtype=float)
```
IAlibay: If we match the CRYST1 record of a cryo-em file, we don't want to set the box at all. Setting `cell_dims` (and then eventually `self.ts._unitcell[:]` to [0, 0, 0, 90, 90, 90] technically means that some of the box was read in, which is not true here. Cryo-em CRYST1 records are just placeholder values so they don't have physicality.

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -391,14 +391,21 @@ def _read_frame(self, frame):
...
                 except ValueError:
                     warnings.warn("Failed to read CRYST1 record, "
                                   "possibly invalid PDB file, got:\n{}"
                                   "".format(line))
-
+                else:
+                    if (np.array(cell_dims, dtype=float) == 
+                        np.array([1, 1, 1, 90, 90, 90], dtype=float)).all():
+                        warnings.warn("1 A^3 CRYST1 record," 
+                                      " possibly an EM structure file",stacklevel=2)
```
IAlibay: I'm not sure that setting stacklevel to 2 here is appropriate. If I remember how this works properly, this should make the warning trace back to `_read_next_timestep` instead. It probably is best to leave the warning pointing at this exact part of the code so that it's easier to understand why this warning is occuring (see above warning).

testsuite/MDAnalysisTests/coordinates/test_pdb.py
```diff
@@ -38,7 +38,7 @@
                                        XPDB_small, PSF, DCD, CONECT, CRD,
                                        INC_PDB, PDB_xlserial, ALIGN, ENT,
                                        PDB_cm, PDB_cm_gz, PDB_cm_bz2,
-                                       PDB_mc, PDB_mc_gz, PDB_mc_bz2)
+                                       PDB_mc, PDB_mc_gz, PDB_mc_bz2, PDB_CRYOEM_BOX)
```
IAlibay: PEP8: the addition of `PDB_CRYOEM_BOX` to this line is making the number of characters exceed 79. I would add it on a newline.

testsuite/MDAnalysisTests/coordinates/test_pdb.py
```diff
@@ -961,3 +961,11 @@ def test_partially_missing_cryst():
     assert len(u.atoms) == 3
     assert len(u.trajectory) == 2
     assert_array_almost_equal(u.dimensions, 0.0)
+
+def test_cryst_em_warning():
+    #issue 2599  
+    with pytest.warns(UserWarning) as record:
+        u = mda.Universe(PDB_CRYOEM_BOX)
+        cur_sele = u.select_atoms('around 0.1 (resid 4 and name CA and segid A)')
```
IAlibay: As per my previous comment, I don't think you need to have a `select_atoms` call here to trigger the warning.

Instead I would create a second test that does an atom selection on `PDB_CRYOEM_BOX` that has this check.

testsuite/MDAnalysisTests/coordinates/test_pdb.py
```diff
@@ -961,3 +961,11 @@ def test_partially_missing_cryst():
...
     assert len(u.trajectory) == 2
     assert_array_almost_equal(u.dimensions, 0.0)
+
+def test_cryst_em_warning():
+    #issue 2599  
+    with pytest.warns(UserWarning) as record:
+        u = mda.Universe(PDB_CRYOEM_BOX)
+        cur_sele = u.select_atoms('around 0.1 (resid 4 and name CA and segid A)') 
+    assert record[0].message.args[0] == "1 A^3 CRYST1 record, possibly an EM structure file"
+    assert len(cur_sele) == 0
```
IAlibay: As mentioned above, because this test specifically tests for a warning, I would suggest putting this in a separate test rather than trying to test multiple things at the same time (it'll make for a cleaner error message if the test ever fails).

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -391,14 +391,21 @@ def _read_frame(self, frame):
             elif line[:6] == 'CRYST1':
                 # does an implicit str -> float conversion
                 try:
-                    self.ts._unitcell[:] = [line[6:15], line[15:24],
-                                            line[24:33], line[33:40],
-                                            line[40:47], line[47:54]]
+                    cell_dims = [line[6:15], line[15:24],
```
IAlibay: My bad here for not realising this sooner, but the way this construct works now, you won't be triggering the `ValueError` if a non-numerical CRYST1 record is read in, i.e. if `line[6:15] == "A"` this would be perfectly normal assignement for `cell_dims`.

This however is an issue as we would expect an error to be thrown (hence the test failures).

To fix this I would suggest doing:

```python
cell_dims = np.array([line[6:15], line[15:24], line[24:33], line[33:40],
                     line[40:47], line[47:54]], dtype=np.float32)
```

With the associated changes for llines 402-404, this _should_ work.

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -391,14 +391,25 @@ def _read_frame(self, frame):
...
                 except ValueError:
                     warnings.warn("Failed to read CRYST1 record, "
                                   "possibly invalid PDB file, got:\n{}"
                                   "".format(line))
-
+                else:
+                    if (np.array(cell_dims, dtype=np.float32) == 
+                        np.array([1, 1, 1, 90, 90, 90], 
+                        dtype=np.float32)).all():
+                        warnings.warn("1 A^3 CRYST1 record,"
```
IAlibay: That probably should say "A 1^3 cell was found in the CRYST1 record"
Also "its" -> "it's" (i.e. it is), but here I would go with "this is usually a placeholder value in cryo-em structures" (the user of "for" is a bit ambiguous).

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -391,14 +391,25 @@ def _read_frame(self, frame):
...
+                                  line[24:33], line[33:40],
+                                  line[40:47], line[47:54]], 
+                                  dtype=np.float32)  
                 except ValueError:
                     warnings.warn("Failed to read CRYST1 record, "
                                   "possibly invalid PDB file, got:\n{}"
                                   "".format(line))
-
+                else:
+                    if (np.array(cell_dims, dtype=np.float32) ==
```
IAlibay: You don't need to make a `np.array` from cell_dims, it already is one (see your recent change above).

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -391,14 +391,25 @@ def _read_frame(self, frame):
             elif line[:6] == 'CRYST1':
                 # does an implicit str -> float conversion
                 try:
-                    self.ts._unitcell[:] = [line[6:15], line[15:24],
-                                            line[24:33], line[33:40],
-                                            line[40:47], line[47:54]]
+                    cell_dims = np.array([line[6:15], line[15:24],
+                                  line[24:33], line[33:40],
```
IAlibay: I'm not sure if it's a PEP8 or just a flake8 thing, but my editor is giving me some warning about this indentation. Can you shift the start of these lines so that the first character lines up with the first "[" from the line above?

testsuite/MDAnalysisTests/coordinates/test_pdb.py
```diff
@@ -961,3 +962,18 @@ def test_partially_missing_cryst():
     assert len(u.atoms) == 3
     assert len(u.trajectory) == 2
     assert_array_almost_equal(u.dimensions, 0.0)
+
+def test_cryst_em_warning():
+    #issue 2599  
+    with pytest.warns(UserWarning) as record:
+        u = mda.Universe(PDB_CRYOEM_BOX)
+    assert record[0].message.args[0] == "1 A^3 CRYST1 record," \
+                                      " its usually a placeholder for" \
```
IAlibay: Getting a flake8 warning here, can you shift these lines so that the " of each line is aligned (assuming it doesn't break 80 characters per line).

testsuite/MDAnalysisTests/coordinates/test_pdb.py
```diff
@@ -961,3 +962,18 @@ def test_partially_missing_cryst():
...
+                                      " cryo-em structures. Unit cell" \
+                                      " dimensions will not be set."
+
```
IAlibay: PEP8: you need two empty lines between these functions.

testsuite/MDAnalysisTests/coordinates/test_pdb.py
```diff
@@ -961,3 +962,18 @@ def test_partially_missing_cryst():
...
+    with pytest.warns(UserWarning) as record:
+        u = mda.Universe(PDB_CRYOEM_BOX)
+    assert record[0].message.args[0] == "1 A^3 CRYST1 record," \
+                                      " its usually a placeholder for" \
+                                      " cryo-em structures. Unit cell" \
+                                      " dimensions will not be set."
+                                    
+def test_cryst_em_select():
+    #issue 2599
+    u=mda.Universe(PDB_CRYOEM_BOX)
```
IAlibay: PEP8, can you add spaces around `=` please? (one the lines below too).

testsuite/MDAnalysisTests/coordinates/test_pdb.py
```diff
@@ -961,3 +962,18 @@ def test_partially_missing_cryst():
...
+    cur_sele=u.select_atoms('around 0.1 (resid 4 and name CA and segid A)')
+    assert len(cur_sele) == 0
```
IAlibay: More of a personal change here. Rather than `len` here, I would use `cur_sele.n_atoms` (technically both work, but using `n_atoms` is easier to read code-wise).

package/MDAnalysis/coordinates/PDB.py
Oscuro-Phoenix: Very true. I'm sorry for that grammar!

IAlibay: No need to be happens to all of us, as you can see from my own comment having a few typos 🙃

testsuite/MDAnalysisTests/coordinates/test_pdb.py
```diff
@@ -961,3 +962,20 @@ def test_partially_missing_cryst():
...
+    assert record[0].message.args[0] == "1 A^3 CRYST1 record," \
+                                        " this is usually a placeholder in" \
+                                        " cryo-em structures. Unit cell" \
+                                        " dimensions will not be set."
+                                    
+
+def test_cryst_em_select():
+    #issue 2599
+    u = mda.Universe(PDB_CRYOEM_BOX)
+    cur_sele=u.select_atoms('around 0.1 (resid 4 and name CA and segid A)')
```
IAlibay: PEP8: As per my previous review, you'll want spaces around the `=` here.

Oscuro-Phoenix: Is there anything more that I must do ? Thanks.

IAlibay: So sorry for the delay here, the PR just got lost in a sea of GSOC related things. I'll try to re-review by the end of the day.

package/CHANGELOG
```diff
@@ -111,6 +111,7 @@ Enhancements
   * Improve the distance search in water bridge analysis with capped_distance (PR #2480)
   * Added weights_groupselections option in RMSD for mapping custom weights 
     to groupselections (PR #2610, Issue #2429)
+  * Added user warning for CRYST1 cryo-em structures (PR #2645, Issue #2599)
```
IAlibay, line 114: Two things here:
1) Since this does more than just add a warning I'd be more explicit here, something along the lines of "Checks for cryo-em 1 A^3 default CRYST1 record, disabling setting of box dimensions if found (Issue #2599)".
2) This probably should go under "fixes". Also as per the CHANGELOG rules, entries should be added "newest first", i.e. at the top of the list.

package/CHANGELOG
```diff
@@ -22,6 +22,8 @@ mm/dd/yy richardjgowers, kain88-de, lilyminium, p-j-smith, bdice, joaomcteixeira
   * 0.21.0
 
 Fixes
+  * Checks for cryo-em 1 A^3 default CRYST1 record,
```
IAlibay, line 25: You also need to add yourself to the list above (line 20).

Oscuro-Phoenix: Done. 😅 I guess I'm having issues rebasing right now. Could you take care of the merge conflicts? Thanks.

IAlibay: Sure now worries 👍

Oscuro-Phoenix: Should I close this?

Oscuro-Phoenix: *this PR

IAlibay: No need to do anything, I'll be merging the PR once Travis CI completes. (I'm just checking that everything is all done & we didn't break anything since it last ran).

IAlibay:
Sorry about the close-open, just re-starting Travis in the hope that this MacOS build actually runs through...

Oscuro-Phoenix:
Looks like there's some issue with Travis.

IAlibay:
I'm going to ping @richardjgowers here, thoughts on the Travis failures? Different builds keep timing out. I guess cumulatively the whole thing came back green eventually. Just merge or try to fix Travis somehow?

richardjgowers:
@Oscuro-Phoenix Thanks for the fix!

@IAlibay yeah travis is just a helper tool, you can merge stuff if you think you know better :)

PR #2646, opened 2020-03-25
Add pbc kwarg to Contacts

### LINKED ISSUE #2368: Contact analysis does not respect PBC
**Expected behavior**

* Contact analysis should be able to calculate distances with PBC taken into account.
* There should be a kwarg that allows me to select if I want to use PBC or not.
* At a *minimum* the behavior should be documented as not PBC save.

**Actual behavior**

Code review of `contacts.Contacts`, specifically https://github.com/MDAnalysis/mdanalysis/blob/9372d093acebb78bc5b1a2b4dba815a4e3281b04/package/MDAnalysis/analysis/contacts.py#L430-L432 (and other parts of the code) shows that [`distance_array()`](https://www.mdanalysis.org/docs/documentation_pages/lib/distances.html?highlight=distance_array#MDAnalysis.lib.distances.distance_array) is never called with the `box=ts.dimensions` argument, therefore will never take PBC into account.

(If the class is used on trajectories where intra-molecule contacts are analyzed for a molecule that is whole then no problems arise. However, for molecules broken across PBC or for inter-molecule contacts such as multiple peptides in solution, some contacts will be missed and the contact count is *too low*.)

Based on code review.

current develop (0.21-dev)

ss62171:
Fixes #2368 

 - Added a kwargs boolean option 'is_box' whether to take PBC into account or not.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/analysis/contacts.py
+++ b/package/MDAnalysis/analysis/contacts.py
@@ -428,26 +428,27 @@ class Contacts(AnalysisBase):
         self.grA = u.select_atoms(select[0])
         self.grB = u.select_atoms(select[1])
         self.pbc = pbc
-        self.box_dim = None
         
         # contacts formed in reference
         self.r0 = []
         self.initial_contacts = []
 
+        #get dimension of box if pbc set to True
+        if self.pbc:
+            self._get_box = lambda ts: ts.dimensions
+        else:
+            self._get_box = lambda ts: None
+
         if isinstance(refgroup[0], AtomGroup):
             refA, refB = refgroup
-            if(self.pbc):
-                self.box_dim = refA.universe.dimensions
-
-            self.r0.append(distance_array(refA.positions, refB.positions, box=self.box_dim))
+            self.r0.append(distance_array(refA.positions, refB.positions,
+                                            box=self._get_box(refA.universe)))
             self.initial_contacts.append(contact_matrix(self.r0[-1], radius))
 
         else:
             for refA, refB in refgroup:
-                if(self.pbc):
-                    self.box_dim = refA.universe.dimensions
-
-                self.r0.append(distance_array(refA.positions, refB.positions, box=self.box_dim))
+                self.r0.append(distance_array(refA.positions, refB.positions,
+                                                box=self._get_box(refA.universe)))
                 self.initial_contacts.append(contact_matrix(self.r0[-1], radius))
 
     def _prepare(self):
@@ -455,9 +456,10 @@ class Contacts(AnalysisBase):
 
     def _single_frame(self):
         self.timeseries[self._frame_index][0] = self._ts.frame
-    
+        
         # compute distance array for a frame
-        d = distance_array(self.grA.positions, self.grB.positions, box=self.box_dim)
+        d = distance_array(self.grA.positions, self.grB.positions,
+                            box=self._get_box(self._ts))
         
         for i, (initial_contacts, r0) in enumerate(zip(self.initial_contacts,
                                                        self.r0), 1):
--- a/package/MDAnalysis/analysis/contacts.py
+++ b/package/MDAnalysis/analysis/contacts.py
@@ -392,6 +392,9 @@ class Contacts(AnalysisBase):
         method : string | callable (optional)
             Can either be one of ``['hard_cut' , 'soft_cut', 'radius_cut']`` or a callable
             with call signature ``func(r, r0, **kwargs)`` (the "Contacts API").
+        pbc : bool (optional)
+            Uses periodic boundary conditions to calculate distances if set to ``True``; the
+             default is ``False``.
         kwargs : dict, optional
             dictionary of additional kwargs passed to `method`. Check
             respective functions for reasonable values.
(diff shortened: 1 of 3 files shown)


testsuite/MDAnalysisTests/analysis/test_contacts.py
```diff
@@ -302,6 +304,29 @@ def test_non_callable_method(self, universe):
         with pytest.raises(ValueError):
             self._run_Contacts(universe, method=2, stop=2)
 
+    def test_distance_box(self):
+        u = mda.Universe(TPR, XTC)
+        sel1 = "(resname ARG LYS) and (name NH* NZ)"
+        sel2 = "(resname ASP GLU) and (name OE* OD*)"
+        acidic = u.select_atoms(sel2)
+        basic = u.select_atoms(sel1)
```
fiona-naughton: You could focus the test on a particular pair we know is split across the periodic boundary, to make it easier to directly check the distances measured are what we expect given PBC.

testsuite/MDAnalysisTests/analysis/test_contacts.py
```diff
@@ -302,6 +304,29 @@ def test_non_callable_method(self, universe):
...
+                              kwargs={'is_box': True})
+
+        r = contacts.Contacts(u,
+                              select=(sel1, sel2),
+                              refgroup=(acidic, basic),
+                              method="hard_cut")
+
+        q.r0 = np.array(q.r0).squeeze()
+        r.r0 = np.array(r.r0).squeeze()
+        assert_array_equal(q.r0,r.r0)
```
fiona-naughton: We expect the distance to be different with or without consideration of the periodic boundaries, so this test is currently failing; maybe you could compare directly with what we expect the values to be instead?

ss62171:
@fiona-naughton @orbeckst can you please review these changes?

lilyminium:
Just FYI, `distance_array` will take a `box=None` argument.

ss62171:
> Just FYI, `distance_array` will take a `box=None` argument.

@lilyminium I didn't get this. Can you please clarify :)

package/MDAnalysis/analysis/contacts.py
```diff
@@ -461,7 +479,8 @@ def _new_selections(u_orig, selections, frame):
     """create stand alone AGs from selections at frame"""
     u = MDAnalysis.Universe(u_orig.filename, u_orig.trajectory.filename)
     u.trajectory[frame]
-    return [u.select_atoms(s) for s in selections]
+    temp_u = u.copy()
```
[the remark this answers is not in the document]
ss62171: can I fix this in another pull request?

lilyminium:
> you could set a value box based on pbc (ie. None or dimensions, as appropriate) and run distance_array with that, rather than needing an if/else every time.

Yep, thanks for clarifying @fiona-naughton. However the box dimensions will change for every frame in an NPT ensemble so it might be better to set a box getting function to use in `_single_frame` instead e.g.

```python
if pbc:
    self._get_box = lambda ts: ts.dimensions
else:
    self._get_box = lambda ts: None
```

 and call it with `box=self._get_box(self._ts)`. This is less readable than if/else-ing though.

ss62171:
@lilyminium thanks for clarifying and providing better way to implement this :)

ss62171:
@fiona-naughton any other changes needed?

ss62171:
@lilyminium can you please review this.

testsuite/MDAnalysisTests/analysis/test_contacts.py
```diff
@@ -302,6 +303,26 @@ def test_non_callable_method(self, universe):
...
+        sel_basic = "(resname ARG LYS)"
+        sel_acidic = "(resname ASP GLU)"
+        acidic = u.select_atoms(sel_acidic)
+        basic = u.select_atoms(sel_basic)
+        
+        r = contacts.Contacts(u, select=(sel_acidic, sel_basic),
+                        refgroup=(acidic, basic), radius=6.0, pbc=pbc)
+        r.run()
+        if pbc:
+            expected = [1., 0.43138152, 0.3989021, 0.43824337, 0.41948765,
```
lilyminium, line 318: Looks good @ss62171 -- personally I would change the test so that the expected values are also part of `parametrize`, e.g.

```python
@pytest.mark.parametrize("pbc,expected", [
    
    (True, [1., 0.43138152, 0.3989021, 0.43824337, 0.41948765,
            0.42223239, 0.41354071, 0.43641354, 0.41216834, 0.38334858]),
    (False, [1., 0.42327791, 0.39192399, 0.40950119, 0.40902613,
             0.42470309, 0.41140143, 0.42897862, 0.41472684, 0.38574822])
    
    ])
```

A new push would also restart the Appveyor tests, which as far as I can tell have failed due to running out of time. I can't seem to restart these jobs individually without a commit -- @MDAnalysis/coredevs have any insight?

package/MDAnalysis/analysis/contacts.py
```diff
@@ -373,6 +373,8 @@ class Contacts(AnalysisBase):
     .. versionchanged:: 1.0.0
        ``save()`` method has been removed. Use ``np.savetxt()`` on
        :attr:`Contacts.timeseries` instead.
+    .. versionchanged:: 2.0.0
```
lilyminium, line 376: Thanks @ss62171, just one final thing -- this should be for version 1.0.0 which hasn't been released yet.

lilyminium:
Thanks @ss62171, looks good to me! Are you happy with this @fiona-naughton?

ss62171:
@lilyminium any updates on this?

lilyminium:
Sorry for the delay @ss62171, I've been busy. I've requested another review from @fiona-naughton and if she doesn't do so within 24 hours I'll presume that she is too and merge the PR.

lilyminium:
Thanks for your patience @ss62171, this will make Contacts analysis much nicer to use.

PR #2652, opened 2020-03-29
added test for alignto and fasta2select

### LINKED ISSUE #2598: Increasing coverage for align.py

The coverage for `analysis/align.py` is currently at ~ 87%. There are some easy enough parts of the code that are missing coverage that could be tested. See: https://codecov.io/gh/MDAnalysis/mdanalysis/src/develop/package/MDAnalysis/analysis/align.py

More tests, example cases that should be easy to cover:
  - `alignto`'s `subselection` (lines: 507-512)
  - `fasta2select`'s `alnfilename` and `treefilename` == `None` (lines: 1040-1045)
  - `get_matching_atoms`'s `strict` diagnostics (lines: 1296-1320)

## Additional context ##
As discussed with @richardjgowers this is being raised as a feature request to point GSOC candidates towards code areas where increased coverage would be useful.

ss62171:
Fixes #2598 

 - added test to increase coverage

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/testsuite/MDAnalysisTests/analysis/test_align.py
+++ b/testsuite/MDAnalysisTests/analysis/test_align.py
@@ -151,6 +151,19 @@ class TestGetMatchingAtoms(object):
         with pytest.raises(SelectionError):
             align.alignto(u, ref, select='all', match_atoms=False)
 
+    def test_subselection_alignto(self, universe, reference):
+        sub_string = 'resname ALA and name CA'
+        rmsd = align.alignto(universe, reference, subselection=sub_string)
+        assert_almost_equal(rmsd[1], 5.27e-7, decimal=9)
+
+        sub_ag = universe.select_atoms('resname ALA and name CA')
+        rmsd = align.alignto(universe, reference, subselection=sub_ag)
+        assert_almost_equal(rmsd[1], 5.27e-7, decimal=9)
+
+        with pytest.raises(TypeError):
+            random_no = 9821
+            align.alignto(universe, reference, subselection=random_no)
+
 
 class TestAlign(object):
     @staticmethod
@@ -447,6 +460,14 @@ class TestAlignmentProcessing(object):
         assert len(
             sel['mobile']) == 30623, "selection string has unexpected length"
 
+    def test_fasta2select_file(self, tmpdir):
+        sel = align.fasta2select(self.seq, is_aligned=False,
+                                 alnfilename=None, treefilename=None)
+        assert len(
+            sel['reference']) == 23080, "selection string has unexpected length"
+        assert len(
+            sel['mobile']) == 23090, "selection string has unexpected length"
+
     @pytest.mark.skipif(executable_not_found("clustalw2"),
                         reason="Test skipped because clustalw2 executable not found")
     def test_fasta2select_ClustalW(self, tmpdir):
@@ -462,7 +483,6 @@ class TestAlignmentProcessing(object):
             sel['reference']) == 23080, "selection string has unexpected length"
         assert len(
             sel['mobile']) == 23090, "selection string has unexpected length"
-
 
 def test_sequence_alignment():
     u = mda.Universe(PSF)
--- a/testsuite/MDAnalysisTests/analysis/test_align.py
+++ b/testsuite/MDAnalysisTests/analysis/test_align.py
@@ -21,6 +21,7 @@
 # J. Comput. Chem. 32 (2011), 2319--2327, doi:10.1002/jcc.21787
 #
 from __future__ import absolute_import, division, print_function
+from contextlib import nullcontext as does_not_raise
 
 import MDAnalysis as mda
 import MDAnalysis.analysis.align as align
@@ -151,15 +152,14 @@ class TestGetMatchingAtoms(object):
         with pytest.raises(SelectionError):
             align.alignto(u, ref, select='all', match_atoms=False)
 
-    @pytest.mark.parametrize('sub_string',('resname ALA and name CA', 1234,
-    mda.Universe(PSF, DCD).select_atoms('resname ALA and name CA')))
-    def test_subselection_alignto(self, universe, reference, sub_string):
-        
-        if isinstance(sub_string, int):
-            with pytest.raises(TypeError):
-                align.alignto(universe, reference, subselection=sub_string)
+    @pytest.mark.parametrize('sub_string, expectation', [
+        ('resname ALA and name CA', does_not_raise()),
+        (mda.Universe(PSF, DCD).select_atoms('resname ALA and name CA'), does_not_raise()),
+        (1234, pytest.raises(TypeError)),
+    ])
+    def test_subselection_alignto(self, universe, reference, sub_string, expectation):
 
-        else:
+        with expectation:
             rmsd = align.alignto(universe, reference, subselection=sub_string)
             assert_almost_equal(rmsd[1], 0.0, decimal=9)
 
--- a/testsuite/MDAnalysisTests/analysis/test_align.py
+++ b/testsuite/MDAnalysisTests/analysis/test_align.py
@@ -154,11 +154,11 @@ class TestGetMatchingAtoms(object):
     def test_subselection_alignto(self, universe, reference):
         sub_string = 'resname ALA and name CA'
         rmsd = align.alignto(universe, reference, subselection=sub_string)
-        assert_almost_equal(rmsd[1], 5.27e-7, decimal=9)
+        assert_almost_equal(rmsd[1], 0.0, decimal=9)
 
         sub_ag = universe.select_atoms('resname ALA and name CA')
         rmsd = align.alignto(universe, reference, subselection=sub_ag)
-        assert_almost_equal(rmsd[1], 5.27e-7, decimal=9)
+        assert_almost_equal(rmsd[1], 0.0, decimal=9)
 
         with pytest.raises(TypeError):
             random_no = 9821
--- a/testsuite/MDAnalysisTests/analysis/test_align.py
+++ b/testsuite/MDAnalysisTests/analysis/test_align.py
@@ -21,7 +21,7 @@
 # J. Comput. Chem. 32 (2011), 2319--2327, doi:10.1002/jcc.21787
 #
 from __future__ import absolute_import, division, print_function
-from contextlib import nullcontext as does_not_raise
+from contextlib import ExitStack as does_not_raise
 
 import MDAnalysis as mda
 import MDAnalysis.analysis.align as align
--- a/testsuite/MDAnalysisTests/analysis/test_align.py
+++ b/testsuite/MDAnalysisTests/analysis/test_align.py
@@ -21,7 +21,7 @@
 # J. Comput. Chem. 32 (2011), 2319--2327, doi:10.1002/jcc.21787
 #
 from __future__ import absolute_import, division, print_function
-from contextlib import ExitStack as does_not_raise
+from contextlib import contextmanager
 
 import MDAnalysis as mda
 import MDAnalysis.analysis.align as align
@@ -39,6 +39,10 @@ from numpy.testing import (
     assert_array_almost_equal,
 )
 
+#Function for Parametrizing conditional raising
+@contextmanager
+def does_not_raise():
+    yield
 
 class TestRotationMatrix(object):
     a = np.array([[0.1, 0.2, 0.3], [1.1, 1.1, 1.1]])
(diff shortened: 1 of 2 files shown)


ss62171:
couldn't figure out to add test for `get_matching_atoms`'s `strict` diagnostics (lines: 1296-1320).
Any idea how to do this?

ss62171:
@IAlibay can you review this?

IAlibay:
Apologies @ss62171, things have been a bit hectic this week. I'll try to get to reviewing this over the weekend unless someone else wants to take this on.

testsuite/MDAnalysisTests/analysis/test_align.py
```diff
@@ -151,6 +151,19 @@ def test_toggle_atom_nomatch_mismatch_atoms(self, universe, reference):
         with pytest.raises(SelectionError):
             align.alignto(u, ref, select='all', match_atoms=False)
 
+    def test_subselection_alignto(self, universe, reference):
+        sub_string = 'resname ALA and name CA'
+        rmsd = align.alignto(universe, reference, subselection=sub_string)
+        assert_almost_equal(rmsd[1], 5.27e-7, decimal=9)
```
tylerjereddy, line 157: I see this one is failing in CI. A bit strange at first glance since it passes for me locally, whether I run the test class or just reproduce the code outside of the testing framework.

Thinking a little more, since the two universes are identical, I think this is basically applying the identity rotation matrix to the `subselection`, so maybe there's an instability there in some cases?

Even when the tests passes locally, the `old_rmsd` is 0.0 while the `new_rmsd` is the `5.27..e-7`. I can't decide if that makes sense or not--nothing should change unless the comparison is only using the subselection after, but then what is the subselection compared with?

tylerjereddy: There's a line of Cython code in `MDAnalysis/lib/qcprot.pyx`

`rms = sqrt(fabs(2.0 * (E0 - mxEigenV)/N))`

Where the square root of `2.78755634425465e-13` is taken to produce the `5.27.. e-7` number. That's inside `FastCalcRMSDAndRotation()` function. Maybe there's a precision guard in there that needs a bit of thought.

RMeli: @tylerjereddy I recently implemented the QCP method and noticed this problem as well. For identical structures we should have `E0 = mxEigenV`, therefore I think that because of small numerical errors in `mxEigenV`,  they end up being slightly different. I was actually getting a negative number there, which is how I spotted the problem.

tylerjereddy: Here's a diff I used locally to fix the issue and get this PR to pass tests on the expectation there is `0.0` rmsd here:

```diff
diff --git a/package/MDAnalysis/analysis/align.py b/package/MDAnalysis/analysis/align.py
index 04101b7d7..b694abd48 100644
--- a/package/MDAnalysis/analysis/align.py
+++ b/package/MDAnalysis/analysis/align.py
@@ -271,6 +271,9 @@ def rotation_matrix(a, b, weights=None):
     if a.shape != b.shape:
         raise ValueError("'a' and 'b' must have same shape")

+    if np.allclose(a, b) and weights is None:
+        return np.eye(3, dtype=np.float64), 0.0
+
     N = b.shape[0]

     if weights is not None:
diff --git a/testsuite/MDAnalysisTests/analysis/test_align.py b/testsuite/MDAnalysisTests/analysis/test_align.py
index a0a202863..0fbe8dc0a 100644
--- a/testsuite/MDAnalysisTests/analysis/test_align.py
+++ b/testsuite/MDAnalysisTests/analysis/test_align.py
@@ -154,11 +154,11 @@ class TestGetMatchingAtoms(object):
     def test_subselection_alignto(self, universe, reference):
         sub_string = 'resname ALA and name CA'
         rmsd = align.alignto(universe, reference, subselection=sub_string)
-        assert_almost_equal(rmsd[1], 5.27e-7, decimal=9)
+        assert_almost_equal(rmsd[1], 0.0, decimal=9)

         sub_ag = universe.select_atoms('resname ALA and name CA')
         rmsd = align.alignto(universe, reference, subselection=sub_ag)
-        assert_almost_equal(rmsd[1], 5.27e-7, decimal=9)
+        assert_almost_equal(rmsd[1], 0.0, decimal=9)

         with pytest.raises(TypeError):
             random_no = 9821
```

If other core devs agree with that (or similar) it may dislodge this PR. I also note that MDA seems to be a bit particular about identical structures when weights are specified (that doesn't always give `0.0` rmsd); for now I've said just "no weights and identical structures means 0.0 rmsd."

ss62171: @tylerjereddy so how should I proceed from here?

ss62171: like I got your point that since both `universe` and `reference` is identical then using `subselection` their `new_rmsd` value must be `0.0` but it is not. Should I change the mentioned changes in `rotation_matrix` function in my PR?

tylerjereddy: I'd suggest making those changes yes.

tylerjereddy:
> couldn't figure out to add test for `get_matching_atoms`'s strict diagnostics (lines: 1296-1320).
Any idea how to do this?

Looks like you'll want to generate two atomgroups with the same number of residues. However, some of the residues should have different numbers of atoms. For example, the fourth residue in group 2 might have 2 more atoms than the fourth residue in group 1. On top of that, you'll want to use the `strict=True` keyword invocation.

ss62171:
```python
if ag1.n_atoms != ag2.n_atoms:
        if not match_atoms:
            errmsg = ("Mobile and reference atom selections do not "
                      "contain the same number of atoms and atom "
                      "matching is turned off. To match atoms based "
                      "on residue and mass, try match_atoms=True")
            logger.error(errmsg)
            raise SelectionError(errmsg)
        if ag1.n_residues != ag2.n_residues:
            errmsg = ("Reference and trajectory atom selections do not contain "
                    "the same number of atoms: \n"
                    "atoms:    N_ref={0}, N_traj={1}\n"
                    "and also not the same number of residues:\n"
                    "residues: N_ref={2}, N_traj={3}").format(
                        ag1.n_atoms, ag2.n_atoms,
                        ag1.n_residues, ag2.n_residues)
            logger.error(errmsg)
            raise SelectionError(errmsg)
        else:
            msg = ("Reference and trajectory atom selections do not contain "
                   "the same number of atoms: \n"
                   "atoms:    N_ref={0}, N_traj={1}").format(
                       ag1.n_atoms, ag2.n_atoms)
            if strict:
                logger.error(msg)
                raise SelectionError(msg)
```
If I have different number of atoms and same number of residues having `strict` as `True` then this will raise `SelectionError` as last line of this code. I don't think this works. Any other approach?

testsuite/MDAnalysisTests/analysis/test_align.py
```diff
@@ -460,6 +472,14 @@ def test_fasta2select_aligned(self):
         assert len(
             sel['mobile']) == 30623, "selection string has unexpected length"
 
+    def test_fasta2select_file(self, tmpdir):
```
tylerjereddy, line 475: It looks like a few CI matrix entries are failing this new test because the `clustalw2` program is not available. To deal with that, you can add `@pytest.mark.skipif...` following one of the examples in this file that skips when `clustalw2` is absent.

testsuite/MDAnalysisTests/analysis/test_align.py
```diff
@@ -151,6 +151,19 @@ def test_toggle_atom_nomatch_mismatch_atoms(self, universe, reference):
         with pytest.raises(SelectionError):
             align.alignto(u, ref, select='all', match_atoms=False)
 
+    def test_subselection_alignto(self, universe, reference):
+        sub_string = 'resname ALA and name CA'
+        rmsd = align.alignto(universe, reference, subselection=sub_string)
```
tylerjereddy, line 156: The core devs may prefer that you [parametrize](https://docs.pytest.org/en/latest/parametrize.html) this test as well to reduce duplication and have a separate test issued for each case in the test suite. Probably not critical, but I think it is fairly prevalent in MDAnalysis.

I think you can even do that for the `TypeError` case with an extra mark.

tylerjereddy:
> If I have different number of atoms and same number of residues having strict as True then this will raise SelectionError as last line of this code. I don't think this works. Any other approach?

I stand by my original suggestion--the total number of atoms can be the same without each residue having the same number of atoms.

For example, atomgroup 1 may have the same residues as atomgroup 2, but in a different order.

testsuite/MDAnalysisTests/analysis/test_align.py
```diff
@@ -151,6 +151,18 @@ def test_toggle_atom_nomatch_mismatch_atoms(self, universe, reference):
         with pytest.raises(SelectionError):
             align.alignto(u, ref, select='all', match_atoms=False)
 
+    @pytest.mark.parametrize('sub_string',('resname ALA and name CA', 1234,
+    mda.Universe(PSF, DCD).select_atoms('resname ALA and name CA')))
+    def test_subselection_alignto(self, universe, reference, sub_string):
+        
+        if isinstance(sub_string, int):
+            with pytest.raises(TypeError):
```
tylerjereddy, line 159: for conditional parametrized raising see here: http://doc.pytest.org/en/latest/example/parametrize.html#parametrizing-conditional-raising

ss62171:
> > If I have different number of atoms and same number of residues having strict as True then this will raise SelectionError as last line of this code. I don't think this works. Any other approach?
> 
> I stand by my original suggestion--the total number of atoms can be the same without each residue having the same number of atoms.
> 
> For example, atomgroup 1 may have the same residues as atomgroup 2, but in a different order.

`universe.select_atoms("resid 13") + universe.select_atoms("resid 14")[1:]`
`universe.select_atoms("resid 14") + universe.select_atoms("resid 13")`
Will this work?

testsuite/MDAnalysisTests/analysis/test_align.py
```diff
@@ -21,6 +21,7 @@
 # J. Comput. Chem. 32 (2011), 2319--2327, doi:10.1002/jcc.21787
 #
 from __future__ import absolute_import, division, print_function
+from contextlib import ExitStack as does_not_raise
```
tylerjereddy, line 24: `ExitStack` requires Python `3.3`, but we *still* support Python `2.7` unfortunately. So, that is causing some CI issues.

ss62171:
> I think that would violate the requirement for the total number of atoms to remain the same, since there's a slice operation on one of the residues.

`universe.select_atoms("resid 13") + universe.select_atoms("resid 14")
universe.select_atoms("resid 14") + universe.select_atoms("resid 13")`
Then should this work?

tylerjereddy:
> Then should this work?
Since these contains both residue in different orders and contain different number of atoms but total atoms remains same.

You should be able to find out by trying locally. You can run the coverage analysis locally with `pytest-cov` and the right `pytest` invocation; `--cov=...` flag I think.

ss62171:
> > Then should this work?
> > Since these contains both residue in different orders and contain different number of atoms but total atoms remains same.
> 
> You should be able to find out by trying locally. You can run the coverage analysis locally with `pytest-cov` and the right `pytest` invocation; `--cov=...` flag I think.

I did but couldn't invoke those lines. I'm also confused by the method you described, like why we want same no. of atoms in both `atom group` since first condition is 
`if ag1.n_atoms != ag2.n_atoms:`

tylerjereddy:
Ah, ok, I think you confused me above because you were trying *not* to get funneled into a conditional underneath that.

In that case, you'll indeed need to "edit"/slice one of the residues to achieve having both a different number of atoms and the same number of residues, assuming you start construction from a single atomgroup.

There are probably a few creative ways to achieve the desired situation of "different number of atoms with same number of residues."

tylerjereddy:
You could probably just make two different N-residue selections, until you find one with a different total n-atoms, and then use that in the test case?

tylerjereddy:
That might be a fair bit simpler actually.

ss62171:
Since I've to invoke lines `1296-1320` I need to set `strict` to `True`. Even when doing what you described above and setting `strict` to `True` this will raise `Selection error` at line `1262`. If I'm wrong please correct me. I'm bit confused here :)

tylerjereddy:
Ok, it seems it is an unreachable code path--so not clear to me why @IAlibay suggested it should be covered. Probably they didn't notice it couldn't be reached.

I'd suggest just avoiding that block of coverage then--the code is clearly a bit of a mess, but that's a separate issue to just bumping the coverage a bit.

ss62171:
@tylerjereddy so what should I do from here now to get this PR merged?

tylerjereddy:
I restarted the erroring Travis jobs--let me take another look later.

tylerjereddy:
merging, thanks @ss62171

Issue #2345, opened 2019-09-07
`FastNS` yields wrong results on simple test case


Linux-cpp-lisp:
**Expected behavior**

`FastNS` misses neighbors in a perfect 111 FCC slab (an ideal, oversimplified case).

**Actual behavior**

Instead of identifying 9 or 12 neighbors for every atom, `FastNS` identifies 9 or 12 for _most_ atoms, with some strange exceptions:

![MDAnalysisBug1](https://user-images.githubusercontent.com/1473644/64477951-1d01ba80-d167-11e9-9a75-5d3a26777ef0.png)

Or see the structure, with atoms colored by coordination number:

![image](https://user-images.githubusercontent.com/1473644/64477973-53d7d080-d167-11e9-84e1-d2ee5dc12351.png)

**Code to reproduce the behavior**

(Requires `ase`.)

The Ag-Ag distance in this system is 2.89Å, so a 3.2Å cutoff is quite sufficient.

``` python
from ase.build import fcc111
from MDAnalysis.lib.mdamath import triclinic_box 
from MDAnalysis.lib.nsgrid import FastNS 
slab = fcc111('Ag', size = (5, 5, 4), vacuum = 10.0)
tcbox = triclinic_box(*slab.cell) 
fns = FastNS(cutoff = 3.2, coords = slab.positions, box = tcbox, pbc = True)
ns = fns.self_search()                                                                                              
cns = [len(n) for n in ns.get_indices()]                                                                            
print(np.unique(cns, return_counts = True))                                                                         
# => (array([ 7,  8,  9, 11, 12, 13]), array([ 1,  3, 46,  3, 45,  2]))
```

The results are identical for cutoffs of 3.0Å and 2.9Å. At a cutoff of 2.5Å, as expected, all atoms have a CN of 0. 

The results are correct, for example, using `pymatgen`'s (very, very, very slow) [`CutOffDictNN`](https://pymatgen.org/pymatgen.analysis.local_env.html#pymatgen.analysis.local_env.CutOffDictNN):

```python
from pymatgen.io.ase import AseAtomsAdaptor 
from pymatgen.analysis.local_env import CutOffDictNN 
cdnn  = CutOffDictNN(cut_off_dict={('Ag', 'Ag') : 3.0})
slabstruct = AseAtomsAdaptor.get_structure(slab)
cs_matcutoff = [len(e) for e in cdnn.get_all_nn_info(slabstruct)]
[...]

Issue #2229, opened 2019-03-30
Neighbor search with nsgrid.FastNS yields different pairs for swapped coords and search_coords


ezavod:
**Expected behavior**

Let's say one has _A_ and _B_ particles. It should make no difference if one constructs `FastNS` with coordinates of _A_ and searches with coordinates of _B_ or vice versa. The pairs of _A-B_ that are found within a certain cutoff should be invariant against exchanging _A around B_ or _B around A_.

**Actual behavior**

In some cases one gets different pairs depending on the order.

**Code to reproduce the behavior**

This is `simple.gro` (notice the extra `5` in the last decimal place of the first box vector):
```
Written by MDAnalysis
   2
    1RA       A    1   0.000   0.000   0.000
    2RB       B    2   5.500   0.000   0.000
   5.745585  5.00000   5.00000
```
This script searches pairs of _A-B_ within a cutoff of 3. With PBC _A_ and _B_ are well within this distance.
``` python
import MDAnalysis
from MDAnalysis.lib.nsgrid import FastNS

u = MDAnalysis.Universe("simple.gro")

max_cutoff = 3
configuration = u.select_atoms("name A").positions
reference = u.select_atoms("name B").positions
box = u.trajectory[0].dimensions

gridsearch = FastNS(max_cutoff, configuration, box=box)
results = gridsearch.search(reference)
pairs = results.get_pairs()
print("Constructed with A; searched with B:", pairs)

gridsearch = FastNS(max_cutoff, reference, box=box)
results = gridsearch.search(configuration)
pairs = results.get_pairs()
print("Constructed with B; searched with A:", pairs)
```
Output:
```
Constructed with A; searched with B: [[0 0]]
Constructed with B; searched with A: []
```
If one now **removes the last decimal place** of the first box vector every thing works as expected. One finds the pair no matter which order. So the bug only happens under special circumstances.

**Current version of MDAnalysis**

- MDAnalysis version 0.19.2
- Python version 3.7.2
- openSUSE Tumbleweed 64bit

I also played around with the current develop branch (94288667302c6d54321f0237e39804f69781867d) but this problem is still persistent. However one has to adjust the box size slightly, probably because of the extra margin in #2136.

**Impact**
[...]

PR #2657, opened 2020-04-08
BUG: fix segfault with 0.0 around sels

### LINKED ISSUE #2656: select_atoms("around 0.0 protein") produces segmentation fault.
## Expected behavior ##
The result should be an empty AtomGroup.

## Actual behavior ##
A segmentation fault results and the program crashes:
```
~$ python aroundfail.py
/home/(myusername)/anaconda3/envs/py37mdanalysis/lib/python3.7/site-packages/MDAnalysisTests/__init__.py:125: MatplotlibDeprecationWarning: The 'warn' parameter of use() is deprecated since Matplotlib 3.1 and will be removed in 3.3.  If any parameter follows 'warn', they should be pass as keyword, not positionally.
  matplotlib.use('agg', warn=False)
Segmentation fault (core dumped)
```
gdb backtrace:

```
~$ gdb -ex r --args python aroundfail.py
GNU gdb (Ubuntu 8.1-0ubuntu3.2) 8.1.0.20180409-git
Copyright (C) 2018 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from python...done.
Starting program: /home/jouni/anaconda3/envs/py37mdanalysis/bin/python aroundfail.py
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
/home/jouni/anaconda3/envs/py37mdanalysis/lib/python3.7/site-packages/MDAnalysisTests/__init__.py:125: MatplotlibDeprecationWarning: The 'warn' parameter of use() is deprecated since Matplotlib 3.1 and will be removed in 3.3.  If any parameter follows 'warn', they should be pass as keyword, not positionally.
  matplotlib.use('agg', warn=False)

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff2c225fb in __pyx_f_10MDAnalysis_3lib_6nsgrid_7_NSGrid_fill_grid(__pyx_obj_10MDAnalysis_3lib_6nsgrid__NSGrid*, __Pyx_memviewslice) ()
   from /home/jouni/anaconda3/envs/py37mdanalysis/lib/python3.7/site-packages/MDAnalysis/lib/nsgrid.cpython-37m-x86_64-linux-gnu.so
(gdb) backtrace
#0  0x00007ffff2c225fb in __pyx_f_10MDAnalysis_3lib_6nsgrid_7_NSGrid_fill_grid(__pyx_obj_10MDAnalysis_3lib_6nsgrid__NSGrid*, __Pyx_memviewslice) ()
   from /home/jouni/anaconda3/envs/py37mdanalysis/lib/python3.7/site-packages/MDAnalysis/lib/nsgrid.cpython-37m-x86_64-linux-gnu.so
#1  0x00007ffff2c2bd33 in __pyx_pw_10MDAnalysis_3lib_6nsgrid_6FastNS_1__init__(_object*, _object*, _object*) () from /home/jouni/anaconda3/envs/py37mdanalysis/lib/python3.7/site-packages/MDAnalysis/lib/nsgrid.cpython-37m-x86_64-linux-gnu.so
#2  0x000000000816c208 in type_call (kwds=0x7fffeed75f50, args=0x7ffff05d70f0, type=<optimized out>) at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Objects/typeobject.c:971
#3  _PyObject_FastCallKeywords () at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Objects/call.c:199
#4  0x00000000081d25c6 in call_function (kwnames=0x7ffff2edfc80, oparg=<optimized out>, pp_stack=<synthetic pointer>) at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Python/ceval.c:4619
#5  _PyEval_EvalFrameDefault () at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Python/ceval.c:3139
#6  0x00000000081158f9 in _PyEval_EvalCodeWithName () at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Python/ceval.c:3930
#7  0x0000000008116c60 in _PyFunction_FastCallDict () at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Objects/call.c:376
#8  0x00000000081ced6a in do_call_core (kwdict=0x7fffeed75eb0, callargs=0x7fffeefda190, func=0x7ffff2cba710) at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Python/ceval.c:4645
#9  _PyEval_EvalFrameDefault () at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Python/ceval.c:3191
#10 0x00000000081160b8 in _PyEval_EvalCodeWithName () at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Python/ceval.c:3930
#11 0x00000000081649e7 in _PyFunction_FastCallKeywords () at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Objects/call.c:433
#12 0x00000000081ce2e7 in call_function (kwnames=0x7ffff3296aa0, oparg=<optimized out>, pp_stack=<synthetic pointer>) at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Python/ceval.c:4616
#13 _PyEval_EvalFrameDefault () at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Python/ceval.c:3139
#14 0x00000000081158f9 in _PyEval_EvalCodeWithName () at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Python/ceval.c:3930
#15 0x00000000081649e7 in _PyFunction_FastCallKeywords () at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Objects/call.c:433
#16 0x00000000081ce2e7 in call_function (kwnames=0x7ffff2b6e320, oparg=<optimized out>, pp_stack=<synthetic pointer>) at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Python/ceval.c:4616
#17 _PyEval_EvalFrameDefault () at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Python/ceval.c:3139
#18 0x000000000816475b in function_code_fastcall (globals=<optimized out>, nargs=2, args=<optimized out>, co=<optimized out>) at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Objects/call.c:283
#19 _PyFunction_FastCallKeywords () at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Objects/call.c:408
#20 0x00000000081cd4a0 in call_function (kwnames=0x0, oparg=<optimized out>, pp_stack=<synthetic pointer>) at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Python/ceval.c:4616
#21 _PyEval_EvalFrameDefault () at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Python/ceval.c:3110
#22 0x00000000081160b8 in _PyEval_EvalCodeWithName () at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Python/ceval.c:3930
#23 0x0000000008116a35 in _PyFunction_FastCallDict () at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Objects/call.c:376
#24 0x0000000008134e03 in _PyObject_Call_Prepend () at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Objects/call.c:908
#25 0x000000000812775e in PyObject_Call () at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Objects/call.c:245
#26 0x00000000081ced6a in do_call_core (kwdict=0x7fffef1add70, callargs=0x7ffff02f5310, func=0x7ffffe2e51e0) at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Python/ceval.c:4645
#27 _PyEval_EvalFrameDefault () at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Python/ceval.c:3191
#28 0x00000000081158f9 in _PyEval_EvalCodeWithName () at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Python/ceval.c:3930
#29 0x00000000081649e7 in _PyFunction_FastCallKeywords () at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Objects/call.c:433
#30 0x00000000081d1ab9 in call_function (kwnames=0x0, oparg=<optimized out>, pp_stack=<synthetic pointer>) at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Python/ceval.c:4616
#31 _PyEval_EvalFrameDefault () at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Python/ceval.c:3093
#32 0x00000000081158f9 in _PyEval_EvalCodeWithName () at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Python/ceval.c:3930
#33 0x0000000008116824 in PyEval_EvalCodeEx () at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Python/ceval.c:3959
#34 0x000000000811684c in PyEval_EvalCode (co=<optimized out>, globals=<optimized out>, locals=<optimized out>) at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Python/ceval.c:524
#35 0x000000000822c684 in run_mod () at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Python/pythonrun.c:1035
#36 0x0000000008236ab1 in PyRun_FileExFlags () at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Python/pythonrun.c:988
#37 0x0000000008236ca3 in PyRun_SimpleFileExFlags () at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Python/pythonrun.c:429
#38 0x0000000008237d95 in pymain_run_file (p_cf=0x7ffffffeda60, filename=0x8371740 L"aroundfail.py", fp=0x83b9ff0) at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Modules/main.c:428
#39 pymain_run_filename (cf=0x7ffffffeda60, pymain=0x7ffffffedb70) at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Modules/main.c:1607
#40 pymain_run_python (pymain=0x7ffffffedb70) at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Modules/main.c:2868
#41 pymain_main () at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Modules/main.c:3029
#42 0x0000000008237ebc in _Py_UnixMain () at /home/conda/feedstock_root/build_artifacts/python_1585001848288/work/Modules/main.c:3064
#43 0x00007ffffee01b97 in __libc_start_main (main=0x80f64a0 <main>, argc=2, argv=0x7ffffffedcc8, init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7ffffffedcb8) at ../csu/libc-start.c:310
#44 0x00000000081dcbd0 in _start () at ../sysdeps/x86_64/elf/start.S:103

```

``` python
import MDAnalysis as mda

u = mda.Universe(PSF, DCD)
failing = u.select_atoms('around 0.0 resid 1')
```

0.20.1 
The problem is reproducable in multiple versions of MDAnalysis. This is what I have currently.

Python 3.7.6
Does not seem to be specific to Python version either.

- Which operating system?
 Ubuntu 18.04.4 LTS (Windows Subsystem for Linux)
Found the problem originally in another Linux, so it does not seem to be OS-specific.

tylerjereddy:
Fixes #2656 

* atomgroup selections with `around 0.0` were
segfaulting; fix this by returning an empty
slice of the query atomgroup

* add a regression test for the fix

### DIFF
--- a/package/MDAnalysis/core/selection.py
+++ b/package/MDAnalysis/core/selection.py
@@ -260,6 +260,10 @@ class AroundSelection(DistanceSelection):
         sel = self.sel.apply(group)
         # All atoms in group that aren't in sel
         sys = group[~np.in1d(group.indices, sel.indices)]
+        # all "around" selections with a zero
+        # distance will produce an empty atomgroup
+        if self.cutoff == 0.0:
+            return sys[0:0]
 
         if not sys or not sel:
             return sys[[]]
--- a/package/MDAnalysis/core/selection.py
+++ b/package/MDAnalysis/core/selection.py
@@ -268,6 +268,7 @@ class AroundSelection(DistanceSelection):
             # check for entries with the same position
             result = cdist(sel.positions, sys.positions)
             indices = np.where(result == 0)[1]
+            warnings.warn("No support for 'around 0.0' with wrapping")
             if result.size > 0:
                 return sys[np.asarray(indices, dtype=np.int64)].unique
             else:
--- a/package/MDAnalysis/core/selection.py
+++ b/package/MDAnalysis/core/selection.py
@@ -49,6 +49,9 @@ import functools
 import warnings
 
 import numpy as np
+import scipy
+import scipy.spatial
+from scipy.spatial.distance import cdist
 
 
 from ..lib.util import unique_int_1d
@@ -260,10 +263,15 @@ class AroundSelection(DistanceSelection):
         sel = self.sel.apply(group)
         # All atoms in group that aren't in sel
         sys = group[~np.in1d(group.indices, sel.indices)]
-        # all "around" selections with a zero
-        # distance will produce an empty atomgroup
+
         if self.cutoff == 0.0:
-            return sys[0:0]
+            # check for entries with the same position
+            result = cdist(sel.positions, sys.positions)
+            indices = np.where(result == 0)[1]
+            if result.size > 0:
+                return sys[np.asarray(indices, dtype=np.int64)].unique
+            else:
+                return sys[0:0]
 
         if not sys or not sel:
             return sys[[]]
--- a/package/MDAnalysis/core/selection.py
+++ b/package/MDAnalysis/core/selection.py
@@ -49,13 +49,10 @@ import functools
 import warnings
 
 import numpy as np
-import scipy
-import scipy.spatial
-from scipy.spatial.distance import cdist
 
 
 from ..lib.util import unique_int_1d
-from ..lib import distances
+from ..lib import distances, pkdtree
 from ..exceptions import SelectionError, NoDataError
 
 
@@ -264,23 +261,21 @@ class AroundSelection(DistanceSelection):
         # All atoms in group that aren't in sel
         sys = group[~np.in1d(group.indices, sel.indices)]
 
-        if self.cutoff == 0.0:
-            # check for entries with the same position
-            result = cdist(sel.positions, sys.positions)
-            indices = np.where(result == 0)[1]
-            warnings.warn("No support for 'around 0.0' with wrapping")
-            if result.size > 0:
-                return sys[np.asarray(indices, dtype=np.int64)].unique
-            else:
-                return sys[0:0]
-
         if not sys or not sel:
             return sys[[]]
 
         box = self.validate_dimensions(group.dimensions)
-        pairs = distances.capped_distance(sel.positions, sys.positions,
-                                          self.cutoff, box=box,
-                                          return_distances=False)
+        pkdt = pkdtree.PeriodicKDTree(box=box)
+
+        if box is not None:
+            pkdt.set_coords(coords=sys.positions,
+                            cutoff=self.cutoff)
+        else:
+            pkdt.set_coords(coords=sys.positions,
+                            cutoff=None)
+
+        pairs = pkdt.search_tree(centers=sel.positions,
+                                 radius=self.cutoff)
         if pairs.size > 0:
             indices = np.sort(pairs[:, 1])
 
--- a/testsuite/MDAnalysisTests/core/test_atomselections.py
+++ b/testsuite/MDAnalysisTests/core/test_atomselections.py
@@ -178,13 +178,14 @@ class TestSelectionsCHARMM(object):
         sel = universe.select_atoms('not backbone')
         assert_equal(len(sel), 2486)
 
-    @pytest.mark.parametrize('selstr', [
-        'around 4.0 bynum 1943', 
-        'around 4.0 index 1942'
+    @pytest.mark.parametrize('selstr, size', [
+        ('around 4.0 bynum 1943', 32),
+        ('around 4.0 index 1942', 32),
+        ('around 0.0 resid 1', 0),  # gh-2656
     ])
-    def test_around(self, universe, selstr):
+    def test_around(self, universe, selstr, size):
         sel = universe.select_atoms(selstr)
-        assert_equal(len(sel), 32)
+        assert_equal(len(sel), size)
 
     @pytest.mark.parametrize('selstr', [
         'sphlayer 4.0 6.0 bynum 1281',
--- a/testsuite/MDAnalysisTests/core/test_atomselections.py
+++ b/testsuite/MDAnalysisTests/core/test_atomselections.py
@@ -188,14 +188,10 @@ class TestSelectionsCHARMM(object):
     @pytest.mark.parametrize('selstr, size', [
         ('around 4.0 bynum 1943', 32),
         ('around 4.0 index 1942', 32),
-        # ('around 0.0 resid 1', 0),  # gh-2656
+        ('around 0.0 resid 1', 0),  # gh-2656
     ])
     def test_around(self, universe, selstr, size):
-        if 'around 0.0' in selstr:
-            with pytest.warns(UserWarning):
-                sel = universe.select_atoms(selstr)
-        else:
-            sel = universe.select_atoms(selstr)
+        sel = universe.select_atoms(selstr)
         assert_equal(len(sel), size)
 
     @pytest.mark.xfail(reason="passes in develop, fails in gh-2657, len(res)==7")
@@ -210,8 +206,6 @@ class TestSelectionsCHARMM(object):
             'around 0.0 resid 3')
         assert len(ag) == 1
 
-    @pytest.mark.xfail(reason="nsgrid broken, see gh-2345, finds 10 atoms")
-    # https://github.com/MDAnalysis/mdanalysis/issues/2345#issuecomment-529888097
     def test_around_triclinic_nsgrid(self, u_pbc_triclinic,
                                      cutoff=1.5e-4):
         u_small = u_pbc_triclinic.atoms[:50]
@@ -249,8 +243,7 @@ class TestSelectionsCHARMM(object):
         for i in range(n_dup):
             new_ag[i + 1].position = reference_pos
 
-        with pytest.warns(UserWarning):
-            sel = new_ag.select_atoms(selstr)
+        sel = new_ag.select_atoms(selstr)
         assert_equal(len(sel), size)
 
     @pytest.mark.xfail(reason="see gh-2657")
@@ -258,8 +251,7 @@ class TestSelectionsCHARMM(object):
         # around 0.0 with wrapping
         u = mda.Universe(PDB)
         u.dimensions = [1e-2, 1e-2, 1e-2, 90, 90, 90]
-        with pytest.warns(UserWarning):
-            ag = u.select_atoms('around 0.0 resid 1')
+        ag = u.select_atoms('around 0.0 resid 1')
         assert len(ag) == len(u.atoms) - len(u.residues[0].atoms)
 
     @pytest.mark.parametrize('selstr', [
--- a/testsuite/MDAnalysisTests/core/test_atomselections.py
+++ b/testsuite/MDAnalysisTests/core/test_atomselections.py
@@ -187,6 +187,30 @@ class TestSelectionsCHARMM(object):
         sel = universe.select_atoms(selstr)
         assert_equal(len(sel), size)
 
+    @pytest.mark.parametrize('selstr, n_dup, size', [
+        # no duplicate atom positions means no matches
+        ('around 0.0 bynum 1', 0, 0),
+        # 1 duplicate atom position means 1 match
+        ('around 0.0 bynum 1', 1, 1),
+        # 5 duplicate atom positions means 5 matches
+        ('around 0.0 bynum 1', 5, 5),
+        # the first residue will encompass the duplications
+        # so no non-self entities are selected
+        ('around 0.0 resid 1', 5, 0),
+    ])
+    def test_around_degenerate(self, universe, selstr, n_dup, size):
+        # for cases with duplicate (spatially identical)
+        # atoms separated by 0.0 distance
+        new_ag = universe.atoms
+        reference_pos = new_ag[0].position
+        # produce atoms with unique ids but
+        # the same position
+        for i in range(n_dup):
+            new_ag[i + 1].position = reference_pos
+
+        sel = new_ag.select_atoms(selstr)
+        assert_equal(len(sel), size)
+
     @pytest.mark.parametrize('selstr', [
         'sphlayer 4.0 6.0 bynum 1281',
         'sphlayer 4.0 6.0 index 1280'
--- a/testsuite/MDAnalysisTests/core/test_atomselections.py
+++ b/testsuite/MDAnalysisTests/core/test_atomselections.py
@@ -64,6 +64,12 @@ class TestSelectionsCHARMM(object):
     def universe_copy(self, universe):
         return MDAnalysis.Universe(PSF, DCD)
 
+    @pytest.fixture()
+    def u_pbc_triclinic(self):
+        u = mda.Universe(PDB)
+        u.dimensions = [1e-3, 1e-3, 1e-3, 60, 60, 60]
+        return u
+
     def test_segid(self, universe):
         sel = universe.select_atoms('segid 4AKE')
         assert_equal(sel.n_atoms, 3341, "failed to select segment 4AKE")
@@ -142,7 +148,8 @@ class TestSelectionsCHARMM(object):
         assert_equal(sel.n_residues, 16,
                      "Failed to find all 'resname LEU' residues.")
         assert_equal(sorted(sel.indices),
-                     sorted(universe.select_atoms('segid 4AKE and resname LEU').indices),
+                     sorted(universe.select_atoms(
+                         'segid 4AKE and resname LEU').indices),
                      "selected 'resname LEU' atoms are not the same as auto-generated s4AKE.LEU")
 
     def test_name(self, universe):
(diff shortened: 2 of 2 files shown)


package/MDAnalysis/core/selection.py
```diff
@@ -260,6 +260,10 @@ def apply(self, group):
         sel = self.sel.apply(group)
         # All atoms in group that aren't in sel
         sys = group[~np.in1d(group.indices, sel.indices)]
+        # all "around" selections with a zero
+        # distance will produce an empty atomgroup
+        if self.cutoff == 0.0:
+            return sys[0:0]
```
lilyminium, line 266: Just being pedantic here, but what if I have two atoms superposed? Is there an easy solution for this case?

tylerjereddy: I'll see what I can do. It isn't a case in our testsuite though, so we're not explicitly supporting it yet.

tylerjereddy: Ok, I've revised with code + unit test for the scenario. It can be a little mind bending because of self-exclusion rules.

package/MDAnalysis/core/selection.py
```diff
@@ -261,6 +264,15 @@ def apply(self, group):
         # All atoms in group that aren't in sel
         sys = group[~np.in1d(group.indices, sel.indices)]
 
+        if self.cutoff == 0.0:
```
richardjgowers, line 267: It's pedantic, but this misses the case of atoms in different images at the same position.  I'd sooner actually fix what is causing the segfault than try and special-case it.

tylerjereddy: Can you provide an example unit test for that case? If these things are genuinely important there probably should be unit tests already in place for them.

lilyminium: Would something like this work?
```python
def test_around_superposed_atoms(self):
    u = mda.Universe(PDB)
    u.dimensions = [1e-2, 1e-2, 1e-2, 90, 90, 90]
    ag = u.select_atoms('around 0.0 resid 1')
    assert len(ag) == len(u.atoms)-len(u.residues[0].atoms)
```

from below:
```python
>>> import MDAnalysis as mda
>>> u = mda.Universe(PDB)
>>> u.select_atoms('around 0.1 resid 1')
<AtomGroup with 0 atoms>
>>> u.dimensions = [1e-2, 1e-2, 1e-2, 90, 90, 90]
>>> u.select_atoms('around 0.1 resid 1')
<AtomGroup with 47662 atoms>
>>> u.atoms.wrap()
array([[6.9999695e-03, 3.8146973e-06, 5.0010681e-03],
       [7.9994202e-03, 1.9989014e-03, 2.0008087e-03],
       [9.9945068e-04, 7.9994202e-03, 9.0007782e-03],
       ...,
       [2.0065308e-03, 1.9989014e-03, 7.9994202e-03],
       [3.9978027e-03, 4.0016174e-03, 4.0006638e-03],
       [9.9945068e-04, 8.9988708e-03, 3.0000210e-03]], dtype=float32)
>>> pos = [tuple(x) for x in u.atoms.positions]
>>> pos[1] in pos[len(u.residues[0].atoms):]
True
>>> u.select_atoms('around 0.1 resid 1')
<AtomGroup with 47662 atoms>
>>> u.select_atoms('around 0.0 resid 1')
Segmentation fault: 11
```

Re: actual cause of segmentation fault, I haven't looked v. closely at this but it looks like cutoff=0 triggers an infinite loop and multiple divisions-by-0 in `_NSGrid`.

https://github.com/MDAnalysis/mdanalysis/blob/e0e3462e51714f0115cf009a506a0bd352d7329a/package/MDAnalysis/lib/nsgrid.pyx#L583-L588

tylerjereddy: Thanks, that will be useful.

zemanj:
NSGrid is currently broken and needs a rewrite anyway, see #2345

lilyminium:
I was talking it over with @nbgl and put in a few prints, cutoff=0.0 sets off a chain of incorrect values that winds up dying at 
https://github.com/MDAnalysis/mdanalysis/blob/e01176f5849426c9ee24a0b4a9aa8b4d6a811b29/package/MDAnalysis/lib/nsgrid.pyx#L674

due to indexing a zero-sized array with a very large negative number.

https://github.com/MDAnalysis/mdanalysis/blob/e01176f5849426c9ee24a0b4a9aa8b4d6a811b29/package/MDAnalysis/lib/nsgrid.pyx#L601

Although as @zemanj points out in #2345 NSGrid should just be disabled until fixed.

richardjgowers:
Yeah I think @lilyminium has nailed it here, the real problem is that capped_distance doesn't work with r=0, so putting an if statement in the selection just stops one way of hitting the segfault.

tylerjereddy:
Right, but collecting/adding tests that should already be in place seems sensible to me. They can be `xfail` marked if needed for now. 

I still think I can make the case that it is best to prevent the segfault to the extent reasonable until a "complete rewrite" of that low-level stuff, so I'm not sure I understand the resistance to adding the conditional. Indeed, if the rewrite takes a while a few more guards might be in order I'd think.

tylerjereddy:
Might even mean adding a warning for r=0 or something similar.

tylerjereddy:
> so putting an if statement in the selection just stops one way of hitting the segfault.

Can you provide a unit test for a case that goes around that `if` statement I've added. That would be helpful too. I agree maybe switching it to a warning or something for now might be good. I don't immediately see the other routes to segfault though?

tylerjereddy:
I've revised to add the test case from @lilyminium, which does not segfault on this feature branch. I've marked it with an `xfail` and revised to emit a warning on `around 0.0` sels to note that periodic handling is not in place for that yet.

This still allows for some correct behavior for the single/compacted unit cell scenarios--I don't see another route to segfault through the `around` keyword?

Feel free to edit (push to) the PR to just add the tests though, if that's what the consensus prefers--and just mark them with xfail until fixed at lower level.

lilyminium:
The general gist of #2345 and #2229 seems to be that `_nsgrid_capped` doesn't deal well with triclinic boxes, as well as dying on zero-cutoffs here. Disabling it in favour of bruteforce or (I assume) pkdtree would temporarily solve all three issues until the class gets rewritten, e.g. the below should pass:

```python
    @pytest.fixture()
    def u_pbc_triclinic(self):
        u = mda.Universe(PDB)
        u.dimensions = [1e-3, 1e-3, 1e-3, 60, 60, 60]
        return u

    @pytest.mark.xfail(reason="passes in develop, fails in gh-2657, len(res)==7")
    def test_around_superposed_small_res(self, u_pbc_triclinic):
        ag = u_pbc_triclinic.select_atoms('around 0.0 resid 10')
        assert len(ag) == 1

    @pytest.mark.xfail(reason="segfaults in develop, fails in gh-2657, len(res)==19")
    def test_around_superposed_large_res(self, u_pbc_triclinic):
        ag = u_pbc_triclinic.select_atoms('around 0.0 resid 3')
        assert len(ag) == 1

    @pytest.mark.xfail(reason="nsgrid broken, see gh-2345, finds 14486 atoms")
    # https://github.com/MDAnalysis/mdanalysis/issues/2345#issuecomment-529888097
    def test_around_triclinic_nsgrid(self, u_pbc_triclinic,
                                     cutoff=1.5e-4):
        r1 = u_pbc_triclinic.residues[0].atoms
        notr1 = u_pbc_triclinic.residues[1:].atoms
        dist_arr = distance_array(r1.positions, notr1.positions,
                                  box=u_pbc_triclinic.dimensions)
        n_in_cutoff = np.count_nonzero(np.any(dist_arr <= cutoff, axis=0))
        ag = u_pbc_triclinic.select_atoms('around {} resid 1'.format(cutoff))
        assert len(ag) == n_in_cutoff == 28838

    def test_around_triclinic_bruteforce(self, u_pbc_triclinic,
                                         cutoff=1.500001e-4):
        r1 = u_pbc_triclinic.residues[0].atoms
        notr1 = u_pbc_triclinic.residues[1:].atoms
        dist_arr = distance_array(r1.positions, notr1.positions,
                                  box=u_pbc_triclinic.dimensions)
        n_in_cutoff = np.count_nonzero(np.any(dist_arr <= cutoff, axis=0))
        ag = u_pbc_triclinic.select_atoms('around {} resid 1'.format(cutoff))
        assert len(ag) == n_in_cutoff == 28838
```

I added them to the PR as you suggested, but a smaller test system would be better because they take forever.

tylerjereddy:
I've revised to use pkdtree instead of the capped distance calculation for `around` selections, and reactivated some of the xfailed tests added in this PR.

I can't reactivate all the tricinlic tests you've added though--are we confident that pkdtree is feature complete for periodic handling? I can never really tell which parts of MDA are truly feature-complete for periodic boundaries. Maybe kdtree is "decent but not perfect?"

The segfault is gone without use of my conditional anyway.

lilyminium:
> I can't reactivate all the tricinlic tests you've added though

`test_around_superposed_atoms` might just be wrong (sorry) -- the box looks a bit 'big' so I may have modified the universe before looking at atom images. `test_around_triclinic_nsgrid` and `test_around_triclinic_bruteforce` should make it redundant, but if you want to keep an orthorhombic box, lengths of 1e-4 might be better at shifting everything to the same coordinates.

> are we confident that pkdtree is feature complete for periodic handling? 

If `test_around_superposed_small_res` isn't passing with pkdtree then maybe not -- any insight @zemanj? It might be that pkdtree also doesn't like cutoff=0.

`capped_distance` uses the `bruteforce_capped` method when:
 * one group has <10 atoms (=> difference in segfaulting between `test_around_superposed_small_res` and `test_around_superposed_large_res`), or 
 * the cutoff is < 0.3 the triclinic box size (=> the difference between `test_around_triclinic_nsgrid` and `test_around_triclinic_bruteforce`). 

The brute force method just calculates a distance array and picks the ones under the cutoff. Outside those conditions, `capped_distance` uses `nsgrid_capped`. Personally I would switch those cases to bruteforce or pkdtree handling more upstream in `lib.distances.capped_distance`, because there are other geometric selections and analyses that use that function too.

tylerjereddy:
> there are other geometric selections and analyses that use that function too.

I guess I'm just trying to keep the PR somewhat focused. If other tests aren't failing, we probably need more tests for other code paths if there are other selection issues, so this PR might get a bit too big.

tylerjereddy:
In case it helps identify if tests or `pkdtree` or something else is to blame, here are the failures that result when I remove the `xfail` markers from the tests in this feature branch:

```
========================================================================================================================= FAILURES =========================================================================================================================
__________________________________________________________________________________________________ TestSelectionsCHARMM.test_around_superposed_small_res ___________________________________________________________________________________________________
[gw2] linux -- Python 3.7.4 /mnt/c/Users/treddy/venv_py37/bin/python

self = <MDAnalysisTests.core.test_atomselections.TestSelectionsCHARMM object at 0x7fb5ff2ba110>, u_pbc_triclinic = <Universe with 47681 atoms>

    def test_around_superposed_small_res(self, u_pbc_triclinic):
        ag = u_pbc_triclinic.atoms[6550:6600].select_atoms(
            'around 0.0 resid 10')
>       assert len(ag) == 1
E       assert 0 == 1
E        +  where 0 = len(<AtomGroup with 0 atoms>)

testsuite/MDAnalysisTests/core/test_atomselections.py:200: AssertionError
__________________________________________________________________________________________________ TestSelectionsCHARMM.test_around_superposed_large_res ___________________________________________________________________________________________________
[gw2] linux -- Python 3.7.4 /mnt/c/Users/treddy/venv_py37/bin/python

self = <MDAnalysisTests.core.test_atomselections.TestSelectionsCHARMM object at 0x7fb5fe0952d0>, u_pbc_triclinic = <Universe with 47681 atoms>

    def test_around_superposed_large_res(self, u_pbc_triclinic):
        ag = u_pbc_triclinic.atoms[46800:46900].select_atoms(
            'around 0.0 resid 3')
>       assert len(ag) == 1
E       assert 0 == 1
E        +  where 0 = len(<AtomGroup with 0 atoms>)

testsuite/MDAnalysisTests/core/test_atomselections.py:205: AssertionError
____________________________________________________________________________________________________ TestSelectionsCHARMM.test_around_superposed_atoms _____________________________________________________________________________________________________
[gw2] linux -- Python 3.7.4 /mnt/c/Users/treddy/venv_py37/bin/python

self = <MDAnalysisTests.core.test_atomselections.TestSelectionsCHARMM object at 0x7fb5fc9aebd0>

    def test_around_superposed_atoms(self):
        # around 0.0 with wrapping
        u = mda.Universe(PDB)
        u.dimensions = [1e-2, 1e-2, 1e-2, 90, 90, 90]
        ag = u.select_atoms('around 0.0 resid 1')
>       assert len(ag) == len(u.atoms) - len(u.residues[0].atoms)
E       assert 47 == (47681 - 19)
E        +  where 47 = len(<AtomGroup with 47 atoms>)
E        +  and   47681 = len(<AtomGroup with 47681 atoms>)
E        +    where <AtomGroup with 47681 atoms> = <Universe with 47681 atoms>.atoms
E        +  and   19 = len(<AtomGroup with 19 atoms>)
E        +    where <AtomGroup with 19 atoms> = <Residue MET, 1>.atoms

testsuite/MDAnalysisTests/core/test_atomselections.py:252: AssertionError
```

tylerjereddy:
I'll see if I can take a closer look at those tests soon-ish; I may be a little slow

tylerjereddy:
Looks like Lily and Richard prefer the competing PR from Richard so I'll close this, thanks.

PR #2660, opened 2020-04-11
print message if C extension not installed

### LINKED ISSUE #2550: Catch when C extensions haven't been compiled on import
There's a common error where users import MDAnalysis and the C extensions haven't been compiled & created.  This (currently) ends up manifesting as `no module called _cutil` as this is the first C extension submodule imported when MDAnalysis is imported.

I think putting a try/except around this first C-extension import and then issuing a better error message (possibly with some debug info) would be a nice feature

ss62171:
Fixes #2550 

 - Issued an error message if C extensions are not installed properly.
- Added test for the same

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/lib/util.py
+++ b/package/MDAnalysis/lib/util.py
@@ -207,7 +207,11 @@ from numpy.testing import assert_equal
 import inspect
 
 from ..exceptions import StreamWarning, DuplicateWarning
-from ._cutil import unique_int_1d
+try:
+    from ._cutil import unique_int_1d
+except ImportError:
+    print("MDAnalysis not installed properly.")
+    raise
 
 # Python 3.0, 3.1 do not have the builtin callable()
 try:
--- a/package/MDAnalysis/lib/util.py
+++ b/package/MDAnalysis/lib/util.py
@@ -207,8 +207,11 @@ from numpy.testing import assert_equal
 import inspect
 
 from ..exceptions import StreamWarning, DuplicateWarning
-from ._cutil import unique_int_1d
-
+try:
+    from ._cutil import unique_int_1d
+except ImportError:
+    print('MDAnalysis not installed properly')
+    sys.exit(-1)
 
 # Python 3.0, 3.1 do not have the builtin callable()
 try:
--- a/package/MDAnalysis/lib/util.py
+++ b/package/MDAnalysis/lib/util.py
@@ -211,7 +211,7 @@ try:
     from ._cutil import unique_int_1d
 except ImportError:
     print("MDAnalysis not installed properly.")
-    raise
+    raise ImportError
 
 # Python 3.0, 3.1 do not have the builtin callable()
 try:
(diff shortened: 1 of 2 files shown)


ss62171:
@orbeckst can you review this please?

orbeckst:
Sorry, I'm fairly busy with other things at the moment so I won't be able to review immediately. Thank you for your patience. Maybe someone else can step in.

orbeckst:
Generally speaking: make sure that the CI does not give any errors. Once all is green, people are more likely to review.

Click on the travis link and find out why your tests are failing.

ss62171:
> this doesn't seem to have any changes to the code to catch the failed import? Maybe you forgot to push something?

My mistake :)

package/MDAnalysis/lib/util.py
```diff
@@ -207,8 +207,11 @@
 import inspect
 
 from ..exceptions import StreamWarning, DuplicateWarning
-from ._cutil import unique_int_1d
-
+try:
+    from ._cutil import unique_int_1d
+except ImportError:
+    print("MDAnalysis not installed properly.")
+    raise
```
tylerjereddy, line 214: I'm pretty sure you'd want to raise an `ImportError` rather than a generic `raise`? Otherwise, capturing it might broadly capture all other Exception classes as well.

testsuite/MDAnalysisTests/lib/test_util.py
```diff
@@ -48,6 +50,25 @@
     Make_Whole, TPR, GRO, fullerene, two_water_gro,
 )
 
+def test_absence_cutil():
+    with patch.dict('sys.modules', {'MDAnalysis.lib._cutil':None}):
+        if sys.version_info[:2] <= (3, 3):
```
tylerjereddy, line 55: Doing robust version comparisons in Python is a little tricky. PEP 440 compliance is the best way, but I'm not sure if we want to add another dependency for that just yet (like `packaging`: https://pypi.org/project/packaging/).

Since we already depend on NumPy, using `LooseVersion` from numpy-distutils might be an option. SciPy still uses that in some places. See examples on StackOverflow: https://stackoverflow.com/a/11887885/2942522

ss62171: > Doing robust version comparisons in Python is a little tricky. PEP 440 compliance is the best way, but I'm not sure if we want to add another dependency for that just yet (like `packaging`: https://pypi.org/project/packaging/).
> 
> Since we already depend on NumPy, using `LooseVersion` from numpy-distutils might be an option. SciPy still uses that in some places. See examples on StackOverflow: https://stackoverflow.com/a/11887885/2942522

Can't we use `sys.hexversion`  as described [here](https://stackoverflow.com/questions/1093322/how-do-i-check-what-version-of-python-is-running-my-script/3018075#3018075). I think it's fairly easy to check using this.

tylerjereddy: You could, though I find those hex version numbers hard to read.

tylerjereddy:
Thanks @ss62171

PR #2662, opened 2020-04-14
Added frames and times to AnalysisBase

### LINKED ISSUE #2661: Add .frames, .times to AnalysisBase

I find myself needing to re-generate the frame indices / times of the trajectory slice used in AnalysisBase for almost every analysis to make a meaningful graph. Conversely it's extra work to slice `RMSD.rmsd` and it's the only class that does that to its results (add the frame index and time to the first 2 columns).

Add `.frames` and `.times` arrays to AnalysisBase, populated in `._run()`

## Additional context ##

Would allow for:

```python
for line in rmsd.rmsd.T:
	plt.plot(rmsd.times, line)
```

vs

```python
# after I look up again which column is the frame and which is time
# the contents of .rmsd is not in the class docstring but the rms module instead
# frames should be ints but times and rmsds are floats
for line in rmsd.rmsd.T[2:]:
	plt.plot(rmsd.rmsd[:, 1], line)
```

lilyminium:
Fixes #2661

 - Added frames and times to AnalysisBase

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -78,6 +78,7 @@ Fixes
   * Added missing selection module to leaflet.py (Issue #2612)
 
 Enhancements
+  * Added .frames and .times arrays to AnalysisBase (Issue #2661)
   * Added elements attribute to PDBParser (Issue #2553, #2647)
   * Added new density.DensityAnalysis (Issue #2502)
   * Added ability to use Chemfiles as a trajectory reader backend (PR #1862)
--- a/package/MDAnalysis/analysis/base.py
+++ b/package/MDAnalysis/analysis/base.py
@@ -89,6 +89,13 @@ class AnalysisBase(object):
        na = NewAnalysis(u.select_atoms('name CA'), 35).run(start=10, stop=20)
        print(na.result)
 
+    Attributes
+    ----------
+    times: np.ndarray
+        array of Timestep times. Only exists after calling run()
+    frames: np.ndarray
+        array of Timestep frame indices. Only exists after calling run()
+
     """
 
     def __init__(self, trajectory, verbose=False, **kwargs):
@@ -124,6 +131,11 @@ class AnalysisBase(object):
             stop frame of analysis
         step : int, optional
             number of frames to skip between each analysed frame
+
+
+        .. versionchanged:: 1.0.0
+            Added .frames and .times arrays as attributes
+
         """
         self._trajectory = trajectory
         start, stop, step = trajectory.check_slice_indices(start, stop, step)
@@ -143,14 +155,14 @@ class AnalysisBase(object):
 
     def _prepare(self):
         """Set things up before the analysis loop begins"""
-        pass # pylint: disable=unnecessary-pass
+        pass  # pylint: disable=unnecessary-pass
 
     def _conclude(self):
         """Finalise the results you've gathered.
 
         Called at the end of the run() method to finish everything up.
         """
-        pass # pylint: disable=unnecessary-pass
+        pass  # pylint: disable=unnecessary-pass
 
     def run(self, start=None, stop=None, step=None, verbose=None):
         """Perform the calculation
@@ -168,7 +180,8 @@ class AnalysisBase(object):
         """
         logger.info("Choosing frames to analyze")
         # if verbose unchanged, use class default
-        verbose = getattr(self, '_verbose', False) if verbose is None else verbose
+        verbose = getattr(self, '_verbose',
+                          False) if verbose is None else verbose
 
         self._setup_frames(self._trajectory, start, stop, step)
         logger.info("Starting preparation")
@@ -176,7 +189,7 @@ class AnalysisBase(object):
         for i, ts in enumerate(ProgressBar(
                 self._trajectory[self.start:self.stop:self.step],
                 verbose=verbose)):
-            self._frame_index =  i
+            self._frame_index = i
             self._ts = ts
             self.frames[i] = ts.frame
             self.times[i] = ts.time
--- a/package/MDAnalysis/analysis/base.py
+++ b/package/MDAnalysis/analysis/base.py
@@ -131,6 +131,8 @@ class AnalysisBase(object):
         self.stop = stop
         self.step = step
         self.n_frames = len(range(start, stop, step))
+        self.frames = np.zeros(self.n_frames, dtype=int)
+        self.times = np.zeros(self.n_frames)
 
     def _single_frame(self):
         """Calculate data from a single frame of trajectory
@@ -174,8 +176,10 @@ class AnalysisBase(object):
         for i, ts in enumerate(ProgressBar(
                 self._trajectory[self.start:self.stop:self.step],
                 verbose=verbose)):
-            self._frame_index = i
+            self._frame_index =  i
             self._ts = ts
+            self.frames[i] = ts.frame
+            self.times[i] = ts.time
             # logger.info("--> Doing frame {} of {}".format(i+1, self.n_frames))
             self._single_frame()
         logger.info("Finishing up")
--- a/package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
+++ b/package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
@@ -442,9 +442,8 @@ class HydrogenBondAnalysis(base.AnalysisBase):
     def _prepare(self):
 
         self.hbonds = [[], [], [], [], [], []]
-        self.frames = np.arange(self.start, self.stop, self.step)
-        self.timesteps = (self.frames * self.u.trajectory.dt) + self.u.trajectory[0].time
-
+        self.timesteps = self.times
+        
         # Set atom selections if they have not been provided
         if not self.acceptors_sel:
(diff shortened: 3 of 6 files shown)


package/MDAnalysis/analysis/base.py
```diff
@@ -174,8 +176,10 @@ def run(self, start=None, stop=None, step=None, verbose=None):
         for i, ts in enumerate(ProgressBar(
                 self._trajectory[self.start:self.stop:self.step],
                 verbose=verbose)):
-            self._frame_index = i
+            self._frame_index =  i
```
IAlibay, line 179: Typo: extra space?

testsuite/MDAnalysisTests/analysis/test_base.py
```diff
@@ -65,29 +65,41 @@ def _single_frame(self):
 def u():
     return mda.Universe(PSF, DCD)
```
IAlibay, line 67: PEP8 - two empty lines between functions not in the same class?

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -442,9 +442,8 @@ def _get_dh_pairs(self):
     def _prepare(self):
 
         self.hbonds = [[], [], [], [], [], []]
-        self.frames = np.arange(self.start, self.stop, self.step)
-        self.timesteps = (self.frames * self.u.trajectory.dt) + self.u.trajectory[0].time
-
+        self.timesteps = self.times
```
IAlibay, line 445: Nice catch 👍 

Rather than creating a case where we have two attributes that do the same thing, I would suggest here just deperecating `self.timesteps` in favour for `self.times`?

package/MDAnalysis/analysis/base.py
```diff
@@ -131,6 +131,8 @@ def _setup_frames(self, trajectory, start=None, stop=None, step=None):
         self.stop = stop
         self.step = step
         self.n_frames = len(range(start, stop, step))
+        self.frames = np.zeros(self.n_frames, dtype=int)
```
IAlibay, line 134: You'll probably want to add a `versionchanged` detailing these somewhere.

testsuite/MDAnalysisTests/analysis/test_base.py
```diff
@@ -65,29 +65,41 @@ def _single_frame(self):
...
-def test_default(u):
+def test_default(u, n_frames):
     an = FrameAnalysis(u.trajectory).run()
-    assert an.n_frames == len(u.trajectory)
-    assert_equal(an.frames, list(range(len(u.trajectory))))
+    assert an.n_frames == n_frames
+    assert_equal(an.found_frames, np.arange(n_frames))
+    assert_equal(an.frames, np.arange(n_frames))
+    times = (an.frames+1)/u.trajectory.dt  # DCD starts at 1
+    assert_almost_equal(an.times, times, decimal=4)
```
IAlibay, line 79: Probably too pedantic, please ignore if that's the case. I'm thinking it might be good to also test that times returns as expected when using start & stop too. Also in order to make test failure easier to understand, maybe it'd be better to move times-checking to a separate function and use parametrize?

lilyminium: > I'm thinking it might be good to also test that times returns as expected when using start & stop too. 

Good point, I've done that, but

> Also in order to make test failure easier to understand, maybe it'd be better to move times-checking to a separate function and use parametrize?

I'm not sure what you had in mind here. I did add an error message for help in understanding.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -439,11 +440,16 @@ def _get_dh_pairs(self):
 
         return donors, hydrogens
 
-    def _prepare(self):
+    @property
+    def timesteps(self):
```
IAlibay, line 444: Would it be worth adding a test for this?

IAlibay: Also you'll probably also want a `versionchanged` in the docstring somewhere?

testsuite/MDAnalysisTests/analysis/test_base.py
IAlibay: I guess it's more of a personal preference of mine, I like to put separate out tests so as to avoid a series of asserts that test different things, and then you can just have a more descriptive function signature, but honestly it doesn't matter. These changes are just as, if not more, descriptive.

IAlibay:
@lilyminium I may have just realised that `hydrogenbonds.hbond_analysis` module was added in 0.21 (which is essentially 1.0.0)😅 , the `versionchanged` might not make sense in this case. Neither is a same-version deprecation.

Thoughts on recommended path here? I'd assume this code hasn't made its way downstream yet, so maybe just making the changes without deprecation would be the simplest way?

lilyminium:
Oops. I know some people have been [using HydrogenBonds in paper results](https://groups.google.com/forum/#!searchin/mdnalysis-discussion/hydrogenbondanalysis%7Csort:date/mdnalysis-discussion/rztJG2Yicrg/c_czDylyCAAJ). I don't think this counts as a significant change but pinging @richardjgowers and @bieniekmateusz for their opinions. The alternative is calling the array `timesteps` (which, ehh, they're not really) or keeping both attributes around for that class only.

lilyminium:
Ok, I've taken that as a yes and also removed all other `.. versionchanged::` notes. Let me know if I've been over-eager.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -222,6 +223,7 @@ def __init__(self, universe, donors_sel=None, hydrogens_sel=None, acceptors_sel=
 
         It is highly recommended that a universe topology with bonding information is used, as this is the only way
         that guarantees the correct identification of donor-hydrogen pairs.
+
```
IAlibay, line 226: Extra blank line probably isn't needed?

package/CHANGELOG
```diff
@@ -78,6 +78,7 @@ Fixes
   * Added missing selection module to leaflet.py (Issue #2612)
 
 Enhancements
+  * Added .frames and .times arrays to AnalysisBase (Issue #2661)
```
IAlibay, line 81: Thoughts on also introducing a CHANGELOG entry under `changes` for hbond_analysis?

package/MDAnalysis/analysis/base.py
```diff
@@ -124,13 +131,20 @@ def _setup_frames(self, trajectory, start=None, stop=None, step=None):
...
+            Added .frames and .times arrays as attributes
+
         """
         self._trajectory = trajectory
         start, stop, step = trajectory.check_slice_indices(start, stop, step)
         self.start = start
         self.stop = stop
         self.step = step
         self.n_frames = len(range(start, stop, step))
+        self.frames = np.zeros(self.n_frames, dtype=int)
```
p-j-smith, line 146: Would there be any drawback to instead doing:
```
self.frames = np.arange(start, stop, step, dtype=int)
self.times = np.arange(start, stop, step) * trajectory.dt
```
which would remove the need to add these values in `run()`, and also mean that these attributes exist before `run()` is called?

lilyminium: I didn't do that for `.times` because I wasn't sure how the `times` were obtained, e.g. I think xtc files must encode them somewhere as actual times instead of time steps because `gmx check` can tell you when you don't have regular time steps. `.frames` is a good idea, but I might see if #2664 is welcome first because I [already stuck it in there as `.frame_indices`. ](https://github.com/MDAnalysis/mdanalysis/pull/2664/files#diff-6d44da489d8c88a50d896dd8df8fd393R133)

package/CHANGELOG
lilyminium: I would think a CHANGELOG entry would be about the same thing as a `.. versionchanged::` note wrt updating a new class, though.

package/MDAnalysis/analysis/base.py
orbeckst: In general you do not want to compute times yourself. Believe what the reader gives you.

orbeckst: I am not in favor because the Reader is responsible for interpreting trajectory data, not the upstream code. Believe what the reader gives you.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -440,10 +431,7 @@ def _get_dh_pairs(self):
         return donors, hydrogens
 
     def _prepare(self):
-
         self.hbonds = [[], [], [], [], [], []]
-        self.frames = np.arange(self.start, self.stop, self.step)
-        self.timesteps = (self.frames * self.u.trajectory.dt) + self.u.trajectory[0].time
```
orbeckst, line 446: Yay for removing these lines!

Issue #2578, opened 2020-03-04
Can only run encore convergence on DCDReader/MemoryReader trajectories


lilyminium:
**Expected behavior**

I can run `encore.[dr/c]es_convergence` on any MDAnalysis universe trajectory.

**Actual behavior**

It fails because `prepare_ensembles_for_convergence_increasing_window` fails to transfer trajectories into memory before calling `timeseries` on them.

**Code to reproduce the behavior**

Show us how to reproduce the failiure. If you can, use trajectory files from the test data.

``` python
>>> import MDAnalysis as mda
>>> from MDAnalysis.tests.datafiles import PSF, DCD,  GRO, PDB, TPR, XTC, TRR,  PRMncdf, NCDF
>>> from MDAnalysis.analysis import encore

>>> u = mda.Universe(TPR, XTC)
>>> encore.dres_convergence(u, 10)

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-23-24fe8767487d> in <module>
----> 1 encore.dres_convergence(u, 10)

~/pydev/mdanalysis/package/MDAnalysis/analysis/encore/similarity.py in dres_convergence(original_ensemble, window_size, select, dimensionality_reduction_method, nsamples, ncores)
   1672 
   1673     ensembles = prepare_ensembles_for_convergence_increasing_window(
-> 1674         original_ensemble, window_size, select=select)
   1675 
   1676     coordinates, dimred_details = \

~/pydev/mdanalysis/package/MDAnalysis/analysis/encore/similarity.py in prepare_ensembles_for_convergence_increasing_window(ensemble, window_size, select)
    693     """
    694 
--> 695     ens_size = ensemble.trajectory.timeseries(ensemble.select_atoms(select),
    696                                               order='fac').shape[0]
    697 

AttributeError: 'XTCReader' object has no attribute 'timeseries'
```

**Currently version of MDAnalysis**

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) 0.20.2-dev
- Which version of Python (`python -V`)?
- Which operating system?

richardjgowers:
Related work:

https://github.com/MDAnalysis/mdanalysis/pull/1400#issuecomment-308715045

https://github.com/MDAnalysis/mdanalysis/pull/2271

[...]

PR #2664, opened 2020-04-20
Ensemble container for Universes

### LINKED ISSUE #2663: Ensemble of Universes

I think a lot of people simulate in replicate and need to account for that, calculating PCA for a shared subspace or clustering over an ensemble, etc. It's always a bit tedious tracking which frame is from where. Also, analyses like the encore library operate on ensembles. Encore deals with this by merging them into a new Universe and pulling all the frames into memory, sometimes multiple times. This gets prohibitively large quite quickly, especially when you only care about a small selection of atoms (e.g. alpha-carbons) in your full solvated universe.

Have a Universe container that holds Universes and has convenience methods for iterating over selected frames and selecting atoms from each Universe, allowing for different atom indices from each.

## Additional context ##

@orbeckst once [noted that analysis functions can't use `FrameIterators`](https://github.com/MDAnalysis/UserGuide/pull/40#discussion_r376516624) so I overengineered a container that could be used in place of a `Universe.trajectory` in AnalysisBase classes.

lilyminium:
Fixes MDAnalysis/mdanalysis#2663

 - added an Ensemble class to hold Universes

I wrote most of this a while ago to make `encore` analysis more tractable, but thought it might be useful in general. The `Ensemble` class is designed to:

* hold and track the frames of separate Universes
* enable easy atom selection with the `select_atoms` method only, such that `Ensemble.atoms.positions` contains the positions of the selected atoms at the current time frame, when iterating over combined trajectory.
  * This is alternative to creating a new Universe with `mda.Merge` and having to load the entire trajectory into memory to select a subset of coordinates, as the current ENCORE analysis does
* easily get the relative frame (e.g. 2nd frame of 3rd universe) from an absolute frame (e.g. frame 100 of the combined trajectory), which is helpful for e.g. identifying which actual frame is the centroid of a cluster
* mimic a trajectory to the extent that it can be used in AnalysisBase. It mostly behaves like a FrameIterator, I think -- you can slice it and pass boolean masks to select particular frames. Indexing frames is relative to the Ensemble rather than absolute trajectories.

```python
>>> import MDAnalysis as mda
>>> from MDAnalysis.tests.datafiles import PSF, DCD, PDB, TPR, XTC
>>> u1 = mda.Universe(PSF, DCD)
>>> len(u1.atoms)
3341
>>> u2 = mda.Universe(PDB)
>>> len(u2.atoms)
47681
>>> u3 = mda.Universe(TPR, XTC)
>>> len(u3.atoms)
47681
>>> ens = mda.Ensemble([u1, u2, u3], select='name CA')
>>> len(ens.atoms)
214
>>> len(ens.trajectory)  # 98 + 1 + 10 frames
109
>>> u3.trajectory.frame
0
>>> ens[-1]
< Timestep 9 with unit cell dimensions [80.08523 80.08523 80.08523 60.      60.      90.     ] >
>>> u3.trajectory.frame  # changes persist
9
>>> ens2 = ens.select_atoms('resid 1:50')  # select subset
>>> len(ens2.atoms)
50
>>> len(ens2.trajectory)
109
>>> ens3 = ens2[[0, 3, 2, -19]]  # select whatever frames
>>> len(ens3.trajectory)
4
```

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


jbarnoud:
It is an interesting concept, especially in term of semantics. It could lead to very exciting things about error estimates in analyses when the tools know what replicas are.

I find the interface confusing, though. Indexing an ensemble, for instance, can get you either a universe, a frame, or another ensemble. This is a surprising behaviour as most collections will return a single element, or a new collection of that element. I find this even more confusing because the `Ensemble` object tries to mimic a `Universe`, a `Trajectory`, and an `AtomGroup`, so it cannot fulfil the all the expectations.

What the ideal interface should look like is not obvious to me. I tried suggesting several things and was not fully convinced by any of them. The best I could come up is a combination of the chain reader and on-the-fly transformations to normalise the selections. I guess seeing how the object helps you with encore would help figuring things out.

lilyminium:
Thank you @jbarnoud for your thoughts!

> Indexing an ensemble, for instance, can get you either a universe, a frame, or another ensemble. 

I thought of the frame/ensemble duality similarly to the Atom/Group duality: AtomGroup[:1] returns an AtomGroup of length 1, AtomGroup[[0]] returns an AtomGroup, but AtomGroup[0] returns an Atom. So indexing should only give the item (frame), but slicing should yield the container (in this case, an ensemble).

Getting universe by name is more of a convenience thing. It's not possible to index an array with a string, so I didn't think it would be too strange to add it in. However, it could be more explicit: perhaps `Ensemble.universe_by_name["universe_name"]`?

> the Ensemble object tries to mimic a Universe, a Trajectory, and an AtomGroup, so it cannot fulfil the all the expectations.

Yes -- it's functionally closest to a Trajectory, but to be useful in existing analysis classes it needed atom selection functionality. And most (all?) analysis classes pass `u.trajectory` instead of asking for the trajectory directly. It doesn't pretend to be a Universe/AtomGroup to users, just the analysis API, if that makes sense -- I envisioned people interacting with it just in an indexing/slicing way.

>  The best I could come up is a combination of the chain reader and on-the-fly transformations to normalise the selections. I guess seeing how the object helps you with encore would help figuring things out.

Mostly I wanted to run encore on a collection of Universes. Currently `encore` transfers them all to memory, presumably because it merges them all into the one Universe (this also breaks if any of the Universes have different topologies, because it merges into a single Universe *before* applying the atom selection). Limited memory meant I could not use `encore` as is. * If I therefore use external tools (e.g. cpptraj) and want to load in my pairwise RMSD matrices or PCA projections or cluster indices, the Ensemble class can tell `encore` which frames belong to one trajectory, and then it can proceed. 

It also makes a task like visualising the centroid of a cluster essentially a single step -- instead of working out which trajectory it came from, you could `nglview.show_mdanalysis(ensemble[cluster.centroid_index])`. In addition, if we create a cluster class that contains indices, you could run further analysis on that selection of frames by creating ``cluster_ens = ensemble[cluster_1.indices]``. Currently `Analysis` classes can't take FrameIterators or arbitrary frames, so this could be helpful, especially for aggregate methods such as `AverageStructure`.

*It would be nice to revamp `encore` to not require Universes at all, but simply be a collection of lightweight functions that operate on a matrix or dataframe of values. I do already have `ces` and `dres` functions that operate on a dataframe of labelled subspace projections / cluster indices and treat outliers as distinct clusters.

lilyminium:
@jbarnoud I refactored encore into `reencore`. I hope some of the changes I made to the code here demonstrate how I wanted to make use of the `Ensemble` class. For example, being able to pass an arbitrary array of frame indices to analyse (which does not necessarily need to use an Ensemble, as I made changes to Universe also).

The main idea is to be able to plug and play an Ensemble into e.g. PCA or diffusionmap.DistanceMatrix without having to make multiple new Universes, which can be heavy on the memory and can lead to issues such as missing `.timeseries` methods or `.filename` attributes. (see: MDAnalysis/mdaencore#30, MDAnalysis/mdanalysis#2948, MDAnalysis/mdanalysis#2578, MDAnalysis/mdaencore#32). Moreover, the way `encore` is currently structured, it first merges the Universes and then applies `select_atoms` -- so I cannot, for example, analyse an ensemble of trajectories of proteins if one of the trajectories has water, due to mismatching numbers of atoms.

> If I therefore use external tools (e.g. cpptraj) and want to load in my pairwise RMSD matrices or PCA projections or cluster indices, the Ensemble class can tell encore which frames belong to one trajectory, and then it can proceed.

The above still holds.

Apologies for mixing the encore refactor into this, I can split into two different PRs. It just relied on Ensemble features.

A comparison of timings and examples of usage can be found here: https://mdanalysisuserguide--132.org.readthedocs.build/en/132/examples/analysis/trajectory_similarity/encore.html. The `ces` and `dres` functions especially are considerably faster without merging into all the new Universes.

Edit: the failing test is `test_written_remarks_property`, it seems to be unrelated?

orbeckst:
@mnmelo as you were interested in overloading the ChainReader for ensemble analysis, have a look at this PR for an alternative idea.

lilyminium:
This is currently on hold because it's a new feature that no one asked for and does not solve any current problems, and as @jbarnoud pointed out it suffers from a lack of clarity as to what an ensemble should do. I can put it back on the to-do list if people think it should be in the core library (as opposed to an encore mdakit, for example).

mnmelo:
I can have a look, but can already say that my idea leveraged what the ChainReader already does, ensemble-wise, rather than creating a new collection object.

mnmelo:
So, from what I understand, this idea, when compared to simply using a ChainReader, allows the user to
1. combine trajectories/Universes under the same collection, even if they come from incompatible topologies, and treat them as a single trajectory;
2. explicitly break the reader abstraction and access individual trajectories/Universes when needed.

Also added is the capability to perform analysis runs over arbitrary sequences of frames, which could be considered as an addition on its own (on this aspect, I'd probably overload the `start` argument to take lists, rather than introducing a `frames` one, but I think this could be its own PR and discussed there).

Did I miss any other major features?

The downsides I see to this approach are:
1. overall extensibility for use by other analyses (for replica considerations, for instance) probably requires their adaptation;
2. Since the previous point might stymie adoption, Ensemble could end up becoming little-used API bloat.

If you still think this is a relevant addition, perhaps you could exemplify how this could benefit, say, an RMSD analysis? That way it'd be clearer how this would be advantageous over an overloading of ChainReader (I can already see that one of the advantages is use of incompatible Universes/topologies, but again, that likely requires analysis adaptation).

lilyminium:
I suppose it would save you a little bit of work for RMSD from a single reference structure for different replicas?

```python
import MDAnalysis as mda
from MDAnalysis.tests.datafiles import PSF, DCD, DCD2
from MDAnalysis.analysis.rms import RMSD

u1 = mda.Universe(PSF, DCD)
u2 = mda.Universe(PSF, DCD2)
u3 = mda.Universe(PSF, DCD)

# ensemble way
ens = mda.Ensemble([u1, u2, u3])
rmsd = RMSD(ens, u3, select="name CA").run(step=2)
## Create sub-ensemble just for convenience, could have also passed that into the RMSD analysis
ens2 = ens[::2]
## gives a list of arrays, one array per universe
rmsd_arrays_per_replica = list(ens2.split_array(rmsd.results.rmsd))

# chainreader way
u = mda.Universe(PSF, [DCD, DCD2, DCD])
rmsd = RMSD(u, u3, select="name CA").run(step=2)
rmsd_arrays_per_replica = []
current_replica = []
current_filename = u.trajectory.filename
for frame, rmsd_value in zip(rmsd.results.frames, rmsd.results.rmsd):
  u.trajectory[frame]
  if u.trajectory.filename != current_filename:
    rmsd_arrays_per_replica.append(current_replica)
    current_replica = []
  current_replica.append(rmsd_value)
```

Under the hood, the ensemble `split_array` function just matches the universe indices:
```python
    def split_array(self, arr):
        """Convenience function to split arrays of data by Universe
        into an iterable of data arrays"""
        for i in range(self.n_universes):
            ix = np.where(self._universe_frames[:, 0] == i)[0]
            yield arr[ix[0]:ix[-1]+1]
```

What I frequently like doing is pairwise RMSD comparisons between many different trajectories. The `split_array` function above is pretty convenient for chunking a huge square matrix into many smaller square trajectory-to-trajectory matrices.

However, all of these examples are contrived, because in reality it is unusual to work with replicas with numbers of atoms and different numbers of frames. I suppose I got a little tired of writing the same one-liner over and over to figure out which actual trajectory and which actual frame correlated with a particular cluster index in the combined cluster analysis.

> (on this aspect, I'd probably overload the start argument to take lists, rather than introducing a frames one, but I think this could be its own PR and discussed there).

This is a little off-topic but I think that would be very confusing to users, and I would ask what `start=[0, 1, 2, 3], stop=10, step=2` would give me.

mnmelo:
Thanks for the use-case @lilyminium. I don't think your example is contrived at all. I sometimes also do analyses in a similar spirit (also with incompatible numbers of atoms).

---
_Begin Devil's advocacy_

I think, however, that it's not necessarily easy to generalize the ensemble/replica case. What if you'd wanted each trajectory RMSD'ed to its own reference universe? (say, `u1` with `u1_ref`, etc.) You wouldn't be able to use a single call for the ensemble.

There's also a simpler alternative to ensembles and `ChainReaders`:
```python
u1 = mda.Universe(PSF, DCD)
u2 = mda.Universe(PSF, DCD2)
u3 = mda.Universe(PSF, DCD)

rmsd_arrays_per_replica = []
for u in [u1, u2, u3]:
    rmsd = RMSD(u, u3, select="name CA").run(step=2)
    rmsd_arrays_per_replica.append(rmsd.results.rmsd)
```

What I'm trying to highlight is that there are many possibilities of ensemble analyses, each perhaps too specialized to permit/warrant a common API. I can't tell if there's a specific type more common than others for which it'd be worth investing in API specialization.

_End Devil's advocacy_

---

That said, I can see myself using an `Ensemble`, mostly for statistics, as @jbarnoud pointed out, and perhaps we'd be able to adapt `AnalysisBase` in a way that doesn't involve rewriting every analysis.

> This is a little off-topic but I think that would be very confusing to users, and I would ask what start=[0, 1, 2, 3], stop=10, step=2 would give me.

As to overloading `start`, I meant it in a way analogous to numpy's indexing, where you either specify `[start:stop:step]` or `[iterable]` (no `stop`/`step` allowed in this case) or `[matching_bool_array]` (also no `stop`/`step` allowed in this case).

lilyminium:
@mnmelo thank you for the devil's advocacy, it is easy to overcomplicate things :p In reply to that, I should mention the original use case I had for `ensemble`, which has gotten a bit lot now: it was for "analyses that take an ensemble of trajectory data to be one dataset, but then I want to split up the analysis to look at individual trajectories". For example (and it was indeed originally created for encore), using principal component analysis to construct a shared subspace between multiple trajectories, but then inspecting each trajectory's projection individually to calculate the dimension reduction ensemble similarity.

> perhaps we'd be able to adapt AnalysisBase in a way that doesn't involve rewriting every analysis.

I would hope there wouldn't be much AnalysisBase adaptation at all -- what's in this PR is what I thought was necessary for the existing analyses to work. Primarily setting each frame individually in `fun`, instead of iterating over the trajectory.

> As to overloading start, I meant it in a way analogous to numpy's indexing,

At that stage it might be easier to just adapt `run` to accept a trajectory `FrameIterator`?

mnmelo:
Ok, I see that Analysis adaptation is not very extensive, but analysis creators still have to bear in mind quirks such as using `select_atoms(...).atoms`.

Also, if this goes ahead, the time to implement this is now, so we can advertise this new API requirement without breaking people's analyses.

IAlibay:
> Also, if this goes ahead, the time to implement this is now, so we can advertise this new API requirement without breaking people's analyses.

Please don't scare me like this @mnmelo :P Can we make this is 2.1.0+ please?

mnmelo:
Ok, but if we know we're gonna change the API soon, at least make some obvious warnings from 2.0.0.

lilyminium:
@mnmelo I'd think small changes that we make to our own classes (like `u.select_atoms().atoms`) would be irrelevant to users, and if they have written their own analysis, it is still usable with a normal Universe -- we could just advertise that if you want to use Ensemble these are the things you should be aware of. It's a new feature rather than changing an old one so it'll be backwards compatible.

mnmelo:
That's an ok compromise, I guess: analyses don't have to be Ensemble compatible. If it catches on, we then get to be stricter in the API definition.

hmacdope:
Closing as stale, feel free to re-open if you want to continue.

PR #2668, opened 2020-04-22
Bond-Angle-Torsions coordinate module

daveminh:

I've written a module that will convert coordinates between a Cartesian and a Bond-Angle-Torsion coordinate system.

PR Checklist
------------
 - [X] Tests?
 - [X] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,422 @@
...
+      List to sort
+    reverse : bool
+      Atoms will be in descending order of mass
+
+    Returns
+    -------
+    ag_n : list of Atoms
+      Sorted list
+    """
+    return sorted(sorted(atoms, key = lambda atom:atom.name), \
```
richardjgowers, line 153: With this double sort, is the idea to sort by mass then by name as a tiebreaker?  If so I think better is to do `lambda x: (x.mass, x.name)`.  Tuples will get sorted by their first item, then second etc

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,422 @@
...
+      Atoms that are allowed to be part of the torsion angle
+
+    Returns
+    -------
+    new_torsion : AtomGroup
+      an AtomGroup that defines the torsion angle
+    """
+    for a1 in selected_atoms:
+        # Loop over new atoms connected to the selected atom
+        for a0 in _sort_atoms_by_mass([a for a in a1.bonded_atoms \
```
richardjgowers, line 177: if you feed the input to `_sort_atoms` as a generator, it will be a little more efficient.  ie you can do `_sort_atoms_by_mass(a for a in a1.bonded_atoms etc....)`

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,422 @@
...
+          if (a in self._ag) and (a!=initial_atom)][-1]
+        self._root = mda.AtomGroup([initial_atom, second_atom, third_atom])
+
+        # Construct a list of torsion angles
+        self._torsions = []
+        selected_atoms = mda.AtomGroup(self._root)
+        while len(selected_atoms) < self._ag.n_atoms:
+            try:
+                torsion = _find_torsion(selected_atoms, self._ag)
+            except:
```
richardjgowers, line 250: catching bare exceptions isn't a good idea, make the function throw a known type of error (`ValueError`) then catch that one specifically

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,422 @@
...
+                raise ValueError('AtomGroup is more than one molecule')
+
+            self._torsions.append(torsion)
+            selected_atoms += torsion[0]
+
+        # Get indices of the root and torsion atoms
+        # in a Cartesian positions array that matches the AtomGroup
+        ag_top_inds = list(self._ag.indices)
```
richardjgowers, line 258: `AtomGroup.indices` is going to be a numpy array, so it seems a little unwise to convert this to a list for the sake of the `.index()` method.  Instead you can use `np.where` and not have to make a copy of all the indices

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,422 @@
...
+    def _prepare(self):
+        self.bat = []
+
+    def _single_frame(self):
+        # Calculate coordinates based on the root atoms
+        # The rotation axis is a normalized vector pointing from atom 0 to 1
+        # It is described in two degrees of freedom
+        # by the polar angle and azimuth
+        (p0, p1, p2) = self._root.positions
+        v01 = p1 - p0
```
richardjgowers, line 283: So this assumes that p0, p1, p2 are all in the same periodic image right?  If the p1-p2 bond was wrapped over a box boundary what would happen?

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,422 @@
...
+                             self._ag2.positions,
+                             self._ag3.positions,
+                             box=self._ag1.dimensions)
+        torsions = calc_dihedrals(self._ag1.positions,
+                                  self._ag2.positions,
+                                  self._ag3.positions,
+                                  self._ag4.positions,
+                                  box=self._ag1.dimensions)
+        # When appropriate, calculate improper torsions
+        torsions = np.array([\
```
richardjgowers, line 322: I think there's probably a faster way to do this with numpy masking rather than a list comprehension.  Things like this are going to get slow for big systems.

testsuite/MDAnalysisTests/analysis/test_bat.py
```diff
@@ -0,0 +1,56 @@
...
+        R = BAT(selected_residues)
+        R.run()
+
+        assert_equal(len(R.bat), 98,
+                     err_msg="error: list is not length of trajectory")
+
+        bat = R.bat[0]
+        XYZ = R.Cartesian(bat)
+
+        assert_almost_equal(XYZ, selected_residues.positions, 5,
```
richardjgowers, line 54: Split this test into two, it's checking two different things

testsuite/MDAnalysisTests/analysis/test_bat.py
```diff
@@ -0,0 +1,56 @@
...
+from numpy.testing import assert_equal, assert_almost_equal
+import matplotlib
+import pytest
+
+import MDAnalysis as mda
+from MDAnalysisTests.datafiles import PSF, DCD
+from MDAnalysis.analysis.bat import BAT
+
+
+class TestBAT(object):
```
richardjgowers, line 35: I think some more tests here would be useful.  If you can do a system you know by hand, then hard code the internal coordinates that should be produced as the test requirement (ie a regression test) then it will be easier to maintain.  Just because round tripping works doesn't mean that the coordinates on the far end are going to be what they should be.

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,422 @@
...
+# MDAnalysis: A Python package for the rapid analysis of molecular dynamics
+# simulations. In S. Benthall and S. Rostrup editors, Proceedings of the 15th
+# Python in Science Conference, pages 102-109, Austin, TX, 2016. SciPy.
+# doi: 10.25080/majora-629e541a-00e
+#
+# N. Michaud-Agrawal, E. J. Denning, T. B. Woolf, and O. Beckstein.
+# MDAnalysis: A Toolkit for the Analysis of Molecular Dynamics Simulations.
+# J. Comput. Chem. 32 (2011), 2319--2327, doi:10.1002/jcc.21787
+#
+r"""Bond-Angle-Torsion coordinates analysis --- :mod:`MDAnalysis.analysis.bat`
```
richardjgowers, line 23: You'll need to add a `rst` file stub in the `docs/` for sphinx to build docs for this.

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,422 @@
...
+   Chemistry B* 107(4): 1048–55. doi:`10.1021/jp027149c
+   <https://doi.org/10.1021/jp027149c>`_
+
+.. [Hikiri2016] Hikiri, Simon, Takashi Yoshidome, and Mitsunori Ikeguchi (2016).
+   "Computational Methods for Configurational Entropy Using Internal and
+   Cartesian Coordinates." *Journal of Chemical Theory and Computation*
+   12(12): 5990–6000. doi:`10.1021/acs.jctc.6b00563
+   <https://doi.org/10.1021/acs.jctc.6b00563>`_
+
+.. [Minh2020] Minh, David D L (2020). "Alchemical Grid Dock (AlGDock): Binding
```
richardjgowers, line 119: if you want you can check out how we use duecredit for citations

daveminh:
> @daveminh this looks like a very cool contribution! Would it make sense for this to also be available as a [on the fly transformation](https://www.mdanalysis.org/docs/documentation_pages/trajectory_transformations.html#module-MDAnalysis.transformations)?

Thanks for the useful feedback! I will work on addressing each point. This is my first contribution to the project (and actually any community-run open-source project), so I am unfamiliar with some of the practices.

daveminh:
> @daveminh this looks like a very cool contribution! Would it make sense for this to also be available as a [on the fly transformation](https://www.mdanalysis.org/docs/documentation_pages/trajectory_transformations.html#module-MDAnalysis.transformations)?

I'm open to the idea, but I don't think it makes sense as an on-the-fly transformation because it doesn't change the Cartesian coordinates of the system.

package/MDAnalysis/analysis/bat.py
daveminh: Yes, that is the idea. I want the ordering to be reproducible and was unaware of this trick of sorting by tuples. I am switching the atom name to atom index.

package/MDAnalysis/analysis/bat.py
daveminh: I was also unaware of this trick.

package/MDAnalysis/analysis/bat.py
daveminh: I'm changing _find_torsion to return None if no torsions are found and for this loop to check if torsion is None.

package/MDAnalysis/analysis/bat.py
daveminh: This is another trick I was unaware of. The np.where help recommends the use of np.nonzero, so I'm using that instead.

package/MDAnalysis/analysis/bat.py
daveminh: Yes, that is an assumption. I will use MDAnalysis.lib._cutil.make_whole to unwrap the positions of the root atoms.

package/MDAnalysis/analysis/bat.py
daveminh: I've rewritten these lines to be based on numpy arrays instead of a list comprehension

package/MDAnalysis/analysis/bat.py
daveminh: Looks straightforward.

package/MDAnalysis/analysis/bat.py
daveminh: Hmm, that's an interesting feature. How would I add citations to the duecredit report for those who use the BAT module (or any other module I write)?

(mdanalysis-dev) Minh-IIT-MBP2018:[~/clusters/CCB/FEPiC]: more test_credit.py
import MDAnalysis as mda
u = mda.Universe(PSF, DCD)

# selection of atomgroups
selected_residues = u.select_atoms("resid 5-10")

from MDAnalysis.analysis.bat import BAT

self = BAT(selected_residues)

# Calculates BAT coordinates for a trajectory
self.run()
(mdanalysis-dev) Minh-IIT-MBP2018:[~/clusters/CCB/FEPiC]: python -m duecredit test_credit.py
/Users/dminh/Documents/GitHub/mdanalysis/testsuite/MDAnalysisTests/__init__.py:126: MatplotlibDeprecationWarning: The 'warn' parameter of use() is deprecated since Matplotlib 3.1 and will be removed in 3.3.  If any parameter follows 'warn', they should be pass as keyword, not positionally.
  matplotlib.use('agg', warn=False)

DueCredit Report:
- Molecular simulation analysis library / MDAnalysis (v 0.20.2-dev0) [1, 2]
- Scientific tools library / numpy (v 1.18.1) [3]

2 packages cited
0 modules cited
0 functions cited

References
----------

[1] Gowers, R. et al., 2016. MDAnalysis: A Python Package for the Rapid Analysis of Molecular Dynamics Simulations. In Proceedings of the 15th Python in Science Conference.  SciPy.
[2] Michaud-Agrawal, N. et al., 2011. MDAnalysis: A toolkit for the analysis of molecular dynamics simulations. Journal of Computational Chemistry, 32(10), pp.2319–2327.
[3] Van Der Walt, S., Colbert, S.C. & Varoquaux, G., 2011. The NumPy array: a structure for efficient numerical computation. Computing in Science & Engineering, 13(2), pp.22–30.

testsuite/MDAnalysisTests/analysis/test_bat.py
daveminh: Done

testsuite/MDAnalysisTests/analysis/test_bat.py
daveminh: Added test for ethanol. Checked the internal coordinates by VMD and visual inspection.

daveminh:
The latest commit should resolve all issues except for inclusion in duecredit

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,439 @@
...
+the primary torsion. The others are defined as improper torsions, differences
+between the raw torsion angle and the primary torsion. This definition reduces
+the correlation between the torsion angles.
+
+Each molecule also has six external coordinates that define its translation and
+rotation in space. The three Cartesian coordinates of the first atom are the
+molecule's translational degrees of freedom. Rotational degreres of freedom are
+specified by the axis-angle convention. The rotation axis is a normalized vector
+pointing from the first to second atom. It is described by the polar angle, phi,
+and azimuthal angle, theta. omega is a third angle that describes the rotation
```
orbeckst, line 56: You can use math
```suggestion
pointing from the first to second atom. It is described by the polar angle, :math:`phi`,
and azimuthal angle, :math:`theta`. :math:`omega` is a third angle that describes the rotation
```

You can also include full equations such as 
```reST
.. math::

    \cos\alpha = \frac{\matbf{a}\cdot\mathbf{b}}{ab}
```

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,439 @@
...
+
+   # Reconstruct Cartesian coordinates from BAT coordinates
+   bat = R.bat[0]
+   XYZ = R.Cartesian(bat)
+
+   # The difference between the original and reconstructed coordinates
+   # should be zero.
+   print(np.sum(np.abs(XYZ - selected_residues.positions)>1E-6))
+
+After R.run(), the coordinates can be accessed with :attr:`R.bat`.
```
orbeckst, line 102: Can you please add a section where you explain the data structures? This can be either here or as part of the class. But the user needs to know what they get in `R.bat`. It's confusing that you apparently need to take the first element `R.bat[0]` but then for XYZ you get the full array (presumably shape `(N, 3)`). In any case,  be very clear about what the data structures look like and even more importantly, what the entries in the data structures _mean_.

By all means, add the appropriate equations to explain. You already have references to papers, which is great, but you also want to make sure that when people read the papers they know exactly what they get here.

package/MDAnalysis/analysis/bat.py
orbeckst: For the whole module:
```python
from ..due import due, Doi

due.cite(Doi("10.1002/jcc.2603"),
         description="Bond-angle-torsions coordinates",
         path="MDAnalysis.analysis.bat",
         cite_module=True)
del Doi
```

For a method or function
```python
class BAT():
    @due.dcite(Doi("10.1002/jcc.2603"),
            description="Bond-angle-torsions coordinates analysis",
            path="MDAnalysis.analysis.bat.BAT")
     def __init__(....):
        ....
```
(at least I think that's how you decorate the class; see the [duecredit README](https://github.com/duecredit/duecredit/blob/master/README.md) for details)

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,503 @@
...
+                                  box=self._ag1.dimensions)
+        # When appropriate, calculate improper torsions
+        shift = torsions[self._primary_torsion_indices]
+        shift[self._unique_primary_torsion_indices] = 0.
+        torsions -= shift
+        # Wrap torsions to between -np.pi and np.pi
+        torsions = ((torsions + np.pi) % (2 * np.pi)) - np.pi
+
+        self.bat.append(np.concatenate((root_based, bonds, angles, torsions)))
+
```
orbeckst, line 385: How about adding a `_conclude` method to cast `bat` to an array? You're doing it already for `save` and `load` and it seems a much more sensible thing, especially in order to slice it along time:
```suggestion
    def _conclude(self):
        self.bat = np.asarray(self.bat, dtype=np.float64)
```

On that note, I would make the collection list `self._bat` to make sure that users understand that they don't need it (and also safer for parallel code) and then have
```python

    def _conclude(self):
        self.bat = np.asarray(self._bat, dtype=np.float64)
        self._bat = []
```

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,503 @@
...
+    def load_bat(self, FN):
```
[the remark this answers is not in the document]
orbeckst: If you cast `bat` to array as suggested above then you don't need the cast here.

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,503 @@
...
+        """Loads the bat trajectory from a netcdf file
+        """
+        nc_F = Dataset(FN,'r')
+        self.bat = list(np.array(nc_F.variables['bat']))
+        nc_F.close()
```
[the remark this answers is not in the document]
orbeckst: If you cast `bat` to array as suggested above then you don't need the cast here.
```python
        with Dataset(filename,'r') as nc_F:
            self.bat = np.asarray(nc_F.variables['bat'])
```
(`asarray()` will only do the cast if necessary... I think ncdf already gives you an array, right?)

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,503 @@
...
+        self._ag1 = mda.AtomGroup([ag[0] for ag in self._torsions])
+        self._ag2 = mda.AtomGroup([ag[1] for ag in self._torsions])
+        self._ag3 = mda.AtomGroup([ag[2] for ag in self._torsions])
+        self._ag4 = mda.AtomGroup([ag[3] for ag in self._torsions])
+
+        if bat_nc is not None:
+            self.load_bat(bat_nc)
+
+    def _prepare(self):
+        self.bat = []
```
orbeckst, line 328: See comments below why I suggest this
```suggestion
        self._bat = []
```

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,503 @@
...
+   bat = R.bat[0]
+   XYZ = R.Cartesian(bat)
+
+   # The difference between the original and reconstructed coordinates
+   # should be zero.
+   print(np.sum(np.abs(XYZ - selected_residues.positions)>1E-6))
+
+After R.run(), the coordinates can be accessed with :attr:`R.bat`.
+
+:attr:`R.bat` is a list of numpy arrays. Each list element corresponds
```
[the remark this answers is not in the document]
orbeckst: Make it *one numpy array* instead of a list of arrays as this makes it a lot easier to work with it. See more comments below (especially for `_conclude()` and `load`/`save`)

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,503 @@
...
+to a frame in the trajectory. Each numpy array has the shape (3N,).
```
orbeckst, line 106: Rewrite for `R.bat` as a `(n_frames, 3N)` array.

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,503 @@
...
+
+"""
+from __future__ import absolute_import, division
+
+import numpy as np
+from netCDF4 import Dataset
+import warnings
+
+import MDAnalysis as mda
+from MDAnalysis.analysis.base import AnalysisBase
```
orbeckst, line 146:
```suggestion
from .base import AnalysisBase
```

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,503 @@
...
+        """Loads the bat trajectory from a netcdf file
+        """
+        nc_F = Dataset(FN,'r')
+        self.bat = list(np.array(nc_F.variables['bat']))
+        nc_F.close()
+
+    def save_bat(self, FN):
+        """Saves the bat trajectory to a netcdf file
+        """
+        bat = np.array(self.bat)
```
orbeckst, line 396: will be superfluous when `self.bat` is an array

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,503 @@
...
+   # of the first frame
+   bat = R.bat[0]
+   XYZ = R.Cartesian(bat)
+
+   # The difference between the original and reconstructed coordinates
+   # should be zero.
+   print(np.sum(np.abs(XYZ - selected_residues.positions)>1E-6))
+
+After R.run(), the coordinates can be accessed with :attr:`R.bat`.
+
```
orbeckst, line 104: The description of the `bat` data structure can come under the actual attribute docs or as a separate subsection in the module docs. 

1. You can  add it under the attribute documentation (see my comment below on _Document the attributes that contain data_). Then whenever you link to ``attr:bat``, a user will immediately know what's in there. 

2. If you decide to make it a subsection in the module docs, add a sub header, put a
    ```reST
    .. _bat-datastructure-description:
    
    BAT data structure description
    ------------------------------
    
    The :attr:`BAT.bat` attribute contains the bond-angle-torsion coordinates for all frames
    in the trajectory. It is structures as NumPy :class:`numpy.ndarray` with shape `(n_frames, 3N)`
    where ...
    ```

    Then in the attribute text add a ``:ref:...`` link to the section into the attribute docs like so
    ```reST

    .. autoclass:: BAT 
        :members: 
        :inherited-members: 
  
        .. attribute:: bat 
      
         Bond-angle-torsions internal coordinates for all trajectory frames as a numpy array
         with shape (`n_frames, 3*N)` for the `N` atoms in the :class:`AtomGroup` `ag`.
         See :ref:`bat-datastructure-description` for details of the data structure.
    ```

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,503 @@
...
+    """Calculate BAT coordinates for the specified AtomGroup.
+
+    BAT coordinates will be computed for the group of atoms and all frames
+    in the trajectory belonging to `ag'.`
+
+    """
+    @due.dcite(Doi("10.1002/jcc.26036"),
+        description="Bond-Angle-Torsions Coordinate Transformation",
+        path="MDAnalysis.analysis.bat.BAT")
+    def __init__(self, ag, initial_atom=None, bat_nc=None, **kwargs):
```
orbeckst, line 236: Use `filename` instead of `bat_nc` to be more consistent across the library.

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,503 @@
...
+            :meth:`AtomGroup.guess_bonds <MDAnalysis.core.groups.AtomGroup.guess_bonds>`.
+            ag must only include one molecule.
+            If a trajectory is associated with the atoms then the computation
+            iterates over the trajectory.
+        initial_atom : Atom
+            The atom whose Cartesian coordinates define the translation
+            of the molecule. If not specified, the heaviest terminal atom
+            will be selected.
+        bat_nc : str
+            File name of a netCDF4 file containing a saved bat attribute.
```
orbeckst, line 252: `filename`; you need to explain somewhere that you can load the data from a netcdf file instead of using `run()`.

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,503 @@
...
+   R.run()
+
+   # Reconstruct Cartesian coordinates from BAT coordinates
+   # of the first frame
+   bat = R.bat[0]
+   XYZ = R.Cartesian(bat)
+
+   # The difference between the original and reconstructed coordinates
+   # should be zero.
+   print(np.sum(np.abs(XYZ - selected_residues.positions)>1E-6))
```
orbeckst, line 101: Also show `save`/`load`.

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,503 @@
...
+        bat = np.array(self.bat)
+
+        nc_F = Dataset(FN,'w')
+        frames = nc_F.createDimension("frames", bat.shape[0])
+        dims = nc_F.createDimension("dims", bat.shape[1])
+        bat_in_F = nc_F.createVariable("bat", "f4", ("frames","dims"))
+        bat_in_F[:] = bat
+        nc_F.close()
+
+    def Cartesian(self, bat):
```
orbeckst, line 405: I find it odd that you pass `bat` in here. I think this should just be
```python
def Cartesian(self):
   bat = self.bat
   origin = bat[:3]
   ...
   return XYZ
```
and then use `self.bat` (the object knows about its data).

You are using `self._primary_torsion_indices` and `self._unique_primary_torsion_indices` so I don't see how you could make it a stand-alone function.

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,503 @@
...
+            vu = np.cross(vp, v21)
+
+            XYZ[a0] = p1 + \
+              r01*(vu*sn_ang*cs_tor + vp*sn_ang*sn_tor - v21*cs_ang)
+        return XYZ
+
+    def getAtomGroup(self):
+        """Returns the atomgroup
+        """
+        return self._ag
```
orbeckst, line 503: We generally don't have setters/getters. Instead we use properties. If you want the ag as a "read-only" attribute:
```suggestion
    @property
    def atoms(self):
        """The atomgroup for which BAT are computed (read-only property)"""
        return self._ag
```

package/MDAnalysis/analysis/bat.py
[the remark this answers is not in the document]
daveminh: I actually went with the list because I thought it would be more consistent with the rest of MDAnalysis. Doesn't AnalysisBase._single_frame() method usually append to a list? If I make it an numpy array, should I do that in the AnalysisBase._conclude()?

package/MDAnalysis/analysis/bat.py
daveminh: This function is really for interfacing with other modules. For example, you may want to change one dihedral angle and get a new set of Cartesian coordinates.

package/MDAnalysis/analysis/bat.py
daveminh: Should I keep the names load_bat and save_bat?

package/MDAnalysis/analysis/bat.py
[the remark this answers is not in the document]
orbeckst: We typically change it to an array in `_conclude()`. However, if you can create the empty array in `_prepare()` and just fill it as you go along then that would be even better. When `run()` is run, it will populate `self.start`, `self.stop`, `self.step` and `self.n_frames` that are all available when it calls `_prepare()`. Thus you can allocate an array in 
```python
def _prepare(self):
    self.bat = np.zeros((self.n_frames, 3*self._ag.n_atoms), dtype=np.float64)
```

daveminh:
I still need to build the documents.

package/MDAnalysis/analysis/bat.py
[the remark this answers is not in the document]
daveminh: I've addressed this but will build the documents before requesting another review.

package/MDAnalysis/analysis/bat.py
orbeckst: You mean you take `R.bat`, change something, and then want to get new Cartesian coordinates back without changing the original bat?

In this case I'd do the following
- make `bat=None` optional and default to `self.bat`
- add a sanity check that `bat` can be used (at least dimensions)
- write more docs, outlining what the method is good for

I assume you can't make it a function, independent of the `BAT` class?

package/MDAnalysis/analysis/bat.py
orbeckst: No, call them `load()` and `save()`.

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,503 @@
...
+            raise ValueError('Additional torsions not found.')
+    return torsions
+
+
+class BAT(AnalysisBase):
+    """Calculate BAT coordinates for the specified AtomGroup.
+
+    BAT coordinates will be computed for the group of atoms and all frames
+    in the trajectory belonging to `ag'.`
+
```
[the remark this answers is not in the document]
daveminh: I'm moving most of the description from below the example to here.

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,503 @@
...
+            (p0, p1, p2) = make_whole(self._root, inplace=False)
+        v01 = p1 - p0
+        v21 = p1 - p2
+        # Internal coordinates
+        r01 = np.sqrt(np.sum(v01 *
+                             v01))  # Distance between first two root atoms
+        r12 = np.sqrt(np.sum(v21 *
+                             v21))  # Distance between second two root atoms
+        a012 = np.arccos(max(-1.,min(1.,np.sum(v01*v21)/\
+          np.sqrt(np.sum(v01*v01)*np.sum(v21*v21))))) # Angle between root atoms
```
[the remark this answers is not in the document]
daveminh: This was from yapf, but I manually aligned them.

package/MDAnalysis/analysis/bat.py
daveminh: Okay, I took option 1 based on your other comment on "Document the attributes that contain data"

package/MDAnalysis/analysis/bat.py
daveminh: Switched to an numpy array

package/MDAnalysis/analysis/bat.py
daveminh: Now I'm initiating it as an array, not a list.

package/MDAnalysis/analysis/bat.py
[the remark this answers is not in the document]
daveminh: I'm switching to numpy binaries because the API is simpler and there are fewer compatibility issues. The code was not passing continuous integration tests with netCDF4.

package/MDAnalysis/analysis/bat.py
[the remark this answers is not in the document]
daveminh: I've added the suggested sanity checks on load()

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,578 @@
...
+    """
+    @due.dcite(Doi("10.1002/jcc.26036"),
+               description="Bond-Angle-Torsions Coordinate Transformation",
+               path="MDAnalysis.analysis.bat.BAT")
+    def __init__(self, ag, initial_atom=None, filename=None, **kwargs):
+        r"""Parameters
+        ----------
+        ag : AtomGroup or Universe
+            Group of atoms for which the BAT coordinates are calculated.
+            ag must have a bonds attribute. If not available,
```
orbeckst, line 280:
```suggestion
            `ag` must have a bonds attribute. If not available,
```

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,578 @@
...
+                                  self._ag4.positions,
+                                  box=self._ag1.dimensions)
+        # When appropriate, calculate improper torsions
+        shift = torsions[self._primary_torsion_indices]
+        shift[self._unique_primary_torsion_indices] = 0.
+        torsions -= shift
+        # Wrap torsions to between -np.pi and np.pi
+        torsions = ((torsions + np.pi) % (2 * np.pi)) - np.pi
+
+        self.bat[self._frame_index,:] = \
```
orbeckst, line 427:
```suggestion
        self.bat[self._frame_index, :] = \
```

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,578 @@
...
+        # Check position of initial atom
+        for i, ts in enumerate(self._trajectory[self.start:self.stop:self.step]):
+            self._frame_index = i
+            self._ts = ts
+            self.frames[i] = ts.frame
+            self.times[i] = ts.time
+            if (self.bat[i,:3] != self._root[0].position).any():
+                raise ValueError('Position of initial atom in file ' + \
+                    'inconsistent with current trajectory.')
+        return self
```
orbeckst, line 468: This is an expensive test: re-analyzing the whole trajectory... I think what you're doing here is the gold standard. If you think it's still reasonable safe to just check
- root atom in frame `start`
- length of trajectory = length of bat
then perhaps the cheaper test is ok? But that's your decision here. Think about what your and the users' use cases are going to be.

orbeckst: If you keep the long test, add it to the documentation to tell the user why loading will take a long time. In this case I would also wrap the trajectory in `lib.log.ProgressBar` and add the `verbose` kwarg to mimic what `run()` provides.

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,578 @@
...
+    def save(self, filename):
+        """Saves the bat trajectory in a file in numpy binary format
+
+        See Also
+        --------
+        load: Loads the bat trajectory from a file in numpy binary format
+        """
+        np.save(filename, self.bat)
+
+    def Cartesian(self, bat):
```
orbeckst, line 479: I now understand the use case better. I find it confusing that `bat` here (3N,) and `self.bat` (n_frames, 3N) are different data structures. To make this a bit clearer I would rename the argument here to something like `bat_frame` or `bat_timestep`.

Or alternatively, rename `self.bat` to `self.bat_trajectory`, which immediately signifies what it is and people will happily index it along axis 0 with the idea that that is frame number.

package/MDAnalysis/analysis/bat.py
```diff
@@ -0,0 +1,578 @@
...
+   print(np.sum(np.abs(XYZ - selected_residues.positions)>1E-6))
+
+Change a single torsion angle by :math:`\pi`::
+
+   bat = R.bat[0,:]
+   bat[bat.shape[0]-12] += np.pi
+   XYZ = R.Cartesian(bat)
+
+   # A good number of Cartesian coordinates should have been modified
+   np.sum((XYZ - selected_residues.positions)>1E-5)
```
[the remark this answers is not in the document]
daveminh: Sure

package/MDAnalysis/analysis/bat.py
daveminh: I'll switch to the short sanity check.

package/MDAnalysis/analysis/bat.py
daveminh: I'm renaming the argument to bat_frame

orbeckst:
Looks good to me, @daveminh . I assume when you build the docs locally they look ok? If you want to reorganize your commit history in a few (<5) commits that stand on their own (e.g., feature, tests, CHANGELOG+AUTHORS) then please rewrite your history. Otherwise we will squash everything into a single commit and do a merge.

@richardjgowers any final comments?

orbeckst:
Regarding OTF transformations (https://github.com/MDAnalysis/mdanalysis/pull/2668#pullrequestreview-398866049) one could start thinking about something where you use BAT, modify internal coordinates on the fly, and then pass that as the actual coordinates. Similarly, one could provide BAT as an AUXReader to have the internal coordinates as a parallel trajectory. But that's all beyond the current PR.

Another future improvement might be to select the correct BAT for a given atom – this is currently really complicated because there isn't a 1-to-1 correspondence between the BAT 3N array and the atoms. Not sure what this "topology-like" data structure would be but it could mirror our standard topology.

daveminh:
I see no need to reorganize the history, so please go ahead and squash everything into one commit and do a merge.

Oliver and Richard, thanks for looking things over carefully several times! I've never had code reviewed so thoroughly before. I've also never tried many of the software engineering practices. It was tedious the first time but I'm sure with practice it'll be easier. In time, I hope to be able to develop code that's more broadly useful in the scientific community and better guide my group members to do the same.

Presently I'm not personally planning to pursue more work with BAT itself. Rather I am using it as a stepping stone to a new(ish) class of end-point free energy calculations. I'll probably put in a new pull request around the time I'm ready to submit a paper.

daveminh:
I'm just checking in on this.

orbeckst:
ping @richardjgowers

orbeckst:
@daveminh can you please fix the conflict in CHANGELOG?

I can't do it for you because I don't have permission to modify your branch. There's a way to enable this and then I could have done it myself.

Please ping me when the PR is mergeable and the CI is green. Thanks.

package/CHANGELOG
```diff
@@ -12,8 +12,7 @@ The rules for this file:
   * accompany each entry with github issue/PR number (Issue #xyz)
   * release numbers follow "Semantic Versioning" http://semver.org
 
-------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
daveminh: I'd confused that line as part of the conflict.

orbeckst:
Thank you and congratulations for your first contribution to MDAnalysis, @daveminh !  I also sent you an invitation as a MDAnalysis/contributor.

BAT will be in 2.0.0.  🎉

If there are issues related to the module then we will almost certainly ping you and possibly ask you for comments or code reviews. It really helps when original code authors respond so that we can quickly address issues so please look out for @-mentions from MDAnalysis.

PR #2683, opened 2020-05-16
Fix reading new ts after reading last one (#2674)

### LINKED ISSUE #2674: Aux does not rewind after getting last frame
## Expected behavior ##

Getting timesteps after retrieving the last frame should give the correct auxiliary information.

## Actual behavior ##

I ran into an issue with a trajectory augumented with an `XVGReader` where after obtaining the last `Timestep` with
```python
trajectory[last_frame_idx]
```
the auxiliary information of all the next timesteps obtained through `__getitem__` would be `nan`. I narrowed down the issue to [these lines](https://github.com/MDAnalysis/mdanalysis/blob/6d6b86202a16bff69fad3d8017aca19eed76578c/package/MDAnalysis/auxiliary/base.py#L421-L425):
```python
frame_for_step = self.step_to_frame(self.step, ts)
frame_for_next_step = self.step_to_frame(self.step+1, ts)
if (frame_for_step is not None and frame_for_next_step is not None
        and not (frame_for_step < ts.frame <= frame_for_next_step)):
    self.move_to_ts(ts)
```
when `self.step` is the last one, `frame_for_next_step` is `None` so that `move_to_ts(ts)` is never called.

Patching my local installation with this
```python
frame_for_step = self.step_to_frame(self.step, ts)
frame_for_next_step = self.step_to_frame(self.step+1, ts)
if frame_for_step is not None:
    if (frame_for_next_step is None) and not (frame_for_step < ts.frame):
        # self.step is the last auxiliary step in memory.
        self.move_to_ts(ts)
    elif ((frame_for_next_step is not None) and
          not (frame_for_step < ts.frame <= frame_for_next_step)):
        self.move_to_ts(ts)
```
solved the problem for me.

My code is quite complicated and I don't have time to create a minimal case right now, but I can work on it or even open a PR with the fix if necessary.

MDAnalysis version 0.19.2 installed from the `conda-forge` channel.
Python 3.6.7.
macOS 10.13.6.

andrrizzi:
Fixes #2674.

 - Implemented the bugfix proposed in #2674.
- Added unit test for the bug. I verified that the test failed before the fix.

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -22,6 +22,8 @@ mm/dd/yy richardjgowers, kain88-de, lilyminium, p-j-smith, bdice, joaomcteixeira
   * 0.21.0
 
 Fixes
+  * Fixed retrieval of auxiliary information after getting the last timestep
+    of the trajectory (Issue #2674).
   * Use internal residue indices instead of resids for similarity and connectivity
     selections (Issues #2669 and #2672, PR #2673)
   * Checks for cryo-em 1 A^3 default CRYST1 record, 
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -17,13 +17,13 @@ mm/dd/yy richardjgowers, kain88-de, lilyminium, p-j-smith, bdice, joaomcteixeira
          PicoCentauri, davidercruz, jbarnoud, RMeli, IAlibay, mtiberti, CCook96,
          Yuan-Yu, xiki-tempula, HTian1997, Iv-Hristov, hmacdope, AnshulAngaria,
          ss62171, Luthaf, yuxuanzhuang, abhishandy, mlnance, shfrz, orbeckst,
-         wvandertoorn, cbouy, AmeyaHarmalkar, Oscuro-Phoenix
+         wvandertoorn, cbouy, AmeyaHarmalkar, Oscuro-Phoenix, andrrizzi
 
   * 0.21.0
 
 Fixes
   * Fixed retrieval of auxiliary information after getting the last timestep
-    of the trajectory (Issue #2674).
+    of the trajectory (Issue #2674, PR #2683).
   * n_components correctly selects PCA components (Issue #2623)
   * Use internal residue indices instead of resids for similarity and connectivity
     selections (Issues #2669 and #2672, PR #2673)
--- a/testsuite/MDAnalysisTests/auxiliary/base.py
+++ b/testsuite/MDAnalysisTests/auxiliary/base.py
@@ -352,6 +352,20 @@ class BaseAuxReaderTest(object):
                          "representative value does not match when iterating through all trajectory timesteps")
         u.trajectory.close()
 
+    def test_auxiliary_read_ts_rewind(self, ref):
+        # add to trajectory
+        u = mda.Universe(COORDINATES_TOPOLOGY, COORDINATES_XTC)
+        u.trajectory.add_auxiliary('test', ref.testdata)
+        # AuxiliaryBase.read_ts() should retrieve the correct step after
+        # reading the last one. Issue #2674 describes a case in which the
+        # object gets stuck on the last frame.
+        aux_info_0 = u.trajectory[0].aux.test
+        u.trajectory[-1].aux.test
+        aux_info_0_rewind = u.trajectory[0].aux.test
+        assert_equal(aux_info_0, aux_info_0_rewind,
+                     "aux info was retrieved incorrectly after reading the last step")
+        u.trajectory.close()
+
     def test_get_description(self, ref, reader):
         description = reader.get_description()
         for attr in ref.description:
--- a/testsuite/MDAnalysisTests/auxiliary/base.py
+++ b/testsuite/MDAnalysisTests/auxiliary/base.py
@@ -339,7 +339,7 @@ class BaseAuxReaderTest(object):
 
     def test_iterate_as_auxiliary_from_trajectory(self, ref, ref_universe):
         # check representative values of aux for each frame are as expected
-        # trahectory here has same dt, offset, so there's a direct correspondence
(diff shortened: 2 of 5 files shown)


testsuite/MDAnalysisTests/auxiliary/base.py
```diff
@@ -352,6 +352,20 @@ def test_iterate_as_auxiliary_from_trajectory(self, ref):
...
+        u = mda.Universe(COORDINATES_TOPOLOGY, COORDINATES_XTC)
+        u.trajectory.add_auxiliary('test', ref.testdata)
+        # AuxiliaryBase.read_ts() should retrieve the correct step after
+        # reading the last one. Issue #2674 describes a case in which the
+        # object gets stuck on the last frame.
+        aux_info_0 = u.trajectory[0].aux.test
+        u.trajectory[-1].aux.test
+        aux_info_0_rewind = u.trajectory[0].aux.test
+        assert_equal(aux_info_0, aux_info_0_rewind,
+                     "aux info was retrieved incorrectly after reading the last step")
```
IAlibay, line 366: PEP8 > 79 characters [sorry, this is probably overly picky, especially since there are PEP8 violations above and below]

testsuite/MDAnalysisTests/auxiliary/base.py
```diff
@@ -352,6 +352,20 @@ def test_iterate_as_auxiliary_from_trajectory(self, ref):
...
+        u.trajectory.close()
```
orbeckst, line 367: This seems unnecessary – once you leave the function, `u` gets garbage collected and that should close everything.

testsuite/MDAnalysisTests/auxiliary/base.py
```diff
@@ -352,6 +352,20 @@ def test_iterate_as_auxiliary_from_trajectory(self, ref):
                          "representative value does not match when iterating through all trajectory timesteps")
         u.trajectory.close()
 
+    def test_auxiliary_read_ts_rewind(self, ref):
+        # add to trajectory
+        u = mda.Universe(COORDINATES_TOPOLOGY, COORDINATES_XTC)
+        u.trajectory.add_auxiliary('test', ref.testdata)
```
orbeckst, line 358: This could become a fixture (universe with aux file)

testsuite/MDAnalysisTests/auxiliary/base.py
andrrizzi: No problem. I've fixed it in a couple of previous tests as well.

testsuite/MDAnalysisTests/auxiliary/base.py
andrrizzi: Removed here and in the similar tests above.

testsuite/MDAnalysisTests/auxiliary/base.py
andrrizzi: Done.

andrrizzi:
Thanks for the feedback! I think I've addressed all your comments. Let me know if there is anything else I should fix.

orbeckst:
Thanks. I’ll leave it to @IAlibay to finalize.

package/CHANGELOG
```diff
@@ -22,6 +22,8 @@ mm/dd/yy richardjgowers, kain88-de, lilyminium, p-j-smith, bdice, joaomcteixeira
   * 0.21.0
 
 Fixes
+  * Fixed retrieval of auxiliary information after getting the last timestep
```
IAlibay, line 25: We also have a list of contributors for the upcoming version (line 20), could you add yourself there too?

testsuite/MDAnalysisTests/auxiliary/base.py
```diff
@@ -328,29 +328,34 @@ def test_get_auxreader_for(self, ref, reader):
...
-        u.trajectory.close()
+                         "representative value does not match when "
+                         "iterating through all trajectory timesteps")
+
+    def test_auxiliary_read_ts_rewind(self, ref_universe):
+        # AuxiliaryBase.read_ts() should retrieve the correct step after
+        # reading the last one. Issue #2674 describes a case in which the
+        # object gets stuck on the last frame.
+        aux_info_0 = ref_universe.trajectory[0].aux.test
+        ref_universe.trajectory[-1].aux.test
```
fiona-naughton, line 354: I think just `ref_universe.trajectory[-1]` should work here?

testsuite/MDAnalysisTests/auxiliary/base.py
```diff
@@ -328,29 +328,34 @@ def test_get_auxreader_for(self, ref, reader):
...
+                         "representative value does not match when "
+                         "iterating through all trajectory timesteps")
 
-    def test_iterate_as_auxiliary_from_trajectory(self, ref):
-        # add to trajectory
-        u = mda.Universe(COORDINATES_TOPOLOGY, COORDINATES_XTC)
-        u.trajectory.add_auxiliary('test', ref.testdata)
+    def test_iterate_as_auxiliary_from_trajectory(self, ref, ref_universe):
         # check representative values of aux for each frame are as expected
         # trahectory here has same dt, offset, so there's a direct correspondence
```
fiona-naughton, line 342: While you're here, 'trajectory' has been misspelt in this comment

andrrizzi:
Thanks for taking a look! They should be fixed now.

testsuite/MDAnalysisTests/auxiliary/base.py
andrrizzi: You're right! Just to be sure I've checked, and the test indeed fails with just `ref_universe.trajectory[-1]` without the fix.

IAlibay:
Thanks for getting this done @andrrizzi  I'll squash-merge once Travis returns green.

PR #2685, opened 2020-05-19
Fix #2679

### LINKED ISSUE #2679: Can't write PDB if dimensions is None
## Expected behavior ##

I am using `mda.Universe` objects to handle multiformat IO in one project. A common workflow involves ingesting PDB IDs and writing PDBs to disk.

## Actual behavior ##

I have found at least one PDB ID that makes this simple process fail with this error:

```
  File "mda_bug.py", line 7, in <module>
    m.atoms.write("2bbm.pdb")
  File "/home/jaime/.conda/envs/structuralalignment/lib/python3.8/site-packages/MDAnalysis/core/groups.py", line 3123, in write
    w.write(self.atoms)
  File "/home/jaime/.conda/envs/structuralalignment/lib/python3.8/site-packages/MDAnalysis/coordinates/PDB.py", line 763, in write
    self._write_pdb_header()
  File "/home/jaime/.conda/envs/structuralalignment/lib/python3.8/site-packages/MDAnalysis/coordinates/PDB.py", line 641, in _write_pdb_header
    self.CRYST1(self.convert_dimensions_to_unitcell(u.trajectory.ts))
  File "/home/jaime/.conda/envs/structuralalignment/lib/python3.8/site-packages/MDAnalysis/coordinates/base.py", line 2150, in convert_dimensions_to_unitcell
    lengths, angles = ts.dimensions[:3], ts.dimensions[3:]
TypeError: 'NoneType' object is not subscriptable
```

``` python
from MDAnalysis import fetch_mmtf
m = fetch_mmtf("2BBM")
m.atoms.write("2bbm.pdb")

```

0.20.1

3.8.2

- Which operating system?
Ubuntu on WSL

RMeli:
Fixes #2679

 - `PDBWriter` checks `u.dimensions` ans skips `CRYST1` record if `u.dimensions == None`

The PDB standard states that

> If the structure was not determined by crystallographic means, CRYST1 simply provides the unitary values, with an appropriate REMARK. 

Should I add unitary values (see suggestion in #2679) of skip the CRYST1 record all together (current solution)?

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


testsuite/MDAnalysisTests/coordinates/test_pdb.py
```diff
@@ -291,6 +296,26 @@ def test_write_single_frame_AtomGroup(self, universe2, outfile):
                                                "agree with original coordinates from frame %d" %
                                                u.trajectory.frame)
 
+    def test_write_nodimes(self, universe4, outfile):
```
IAlibay, line 299: More of a thought, please do ignore; would `nodims` be more descriptive than `nodimes` here?

IAlibay: Probably more worth it, maybe consider adding some slight docstring to point out the issue this is testing for, as per the above test?

testsuite/MDAnalysisTests/coordinates/test_pdb.py
```diff
@@ -291,6 +296,26 @@ def test_write_single_frame_AtomGroup(self, universe2, outfile):
...
+        u = universe4
+
```
IAlibay, line 301: Again very pedantic, but are the empty lines on 301 and 303 necessary?

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -651,7 +656,9 @@ def _write_pdb_header(self):
             self.REMARK(*remarks)
         except AttributeError:
             pass
-        self.CRYST1(self.convert_dimensions_to_unitcell(u.trajectory.ts))
+
+        if u.dimensions is not None:
```
IAlibay, line 660: Thought: Whether or not we follow the PDB standard here, do we want to maybe throw a user warning here if dimensions is `None`?

testsuite/MDAnalysisTests/coordinates/test_pdb.py
RMeli: This is actually a typo. Adding a reference to the issue in the docstring is a great idea!

package/MDAnalysis/coordinates/PDB.py
RMeli: I'm not sure about this, some warnings are more annoying than useful. My view here is that if there is no `CRYST1` record in the output it means that the information needed to write it was absent.

However I would be 100% for a warning if we want to set the dimensions automatically (to follow the PDB standard).

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -631,6 +631,11 @@ def _write_pdb_title(self):
                        "".format(self.start, self.remarks))
 
     def _write_pdb_header(self):
+        """
+        .. versionchanged: 1.0.0
```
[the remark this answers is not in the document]
RMeli: Sure, but what would you like to see here? Just re-writing the function name is probably not that helpful (especially for a "private" function).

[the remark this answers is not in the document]
IAlibay: I was thinking something along the lines of `_check_pdb_coordinates`, possibly with some explanation of what's happening with the REMARKS stuff. It's not exactly necessary though.

orbeckst:
> Re: following standard, if we do so I'd definitely suggest we throw a user Warning if we enforce specific dimensions. As per #2599 and #2645, our approach on reading systems that don't have dimensions (and therefore have [1,1,1,90,90,90] boxes defined) is just to have no dimensions. So at least in that case we'd be recovering the original PDB file.

In PR #2645 we forgot to add documentation to the PDBReader to say what we're doing with these boxes. Can you please add it? 

> That being said, I'm worried about gas phase MD simulations where no box is defined and we end up enforcing a [1,1,1,90,90,90] box on PDB writing. That might lead to a lot of downstream issues...

I am in favor of being standard-conform when we can. This is _a lot easier_ than guessing what people want. We do this for reading but for writing we can simply say that we obey the standard. If anyone feels strongly about not including certain elements of a PDB file they can either propose a kwarg to modify the behavior (messy, handle in different PR!) or (better!) derive their custom writer class and use it for their special purposes (didn't @richardjgowers want to write a blog post or User Guide article for that... we do have an advanced tutorial https://github.com/MDAnalysis/WorkshopHackathon2018/blob/master/06_AdvancedTutorials/Writing_custom_readers.ipynb for readers).

Therefore, I propose for this PR: When _writing_ a PDB file with `dimensions == None`
* write a 1 Å<sup>3</sup> dummy CRYST1
* write a REMARK record
* warn the user
* document the behavior in the PDBWriter docs

RMeli:
Thank your for the comments @orbeckst. Do we really need the warning if this is standard-conforming and well documented in the docs?

richardjgowers:
I’d say yes because we’re adding extra info, the 1A boxes caught us out
until lately. And nobody actually follows/knows the standard.

IAlibay:
Re Warning: my vote here is to have it. We throw a warning when we read in a cryo-em default box that we'll throw away the dimensions, if we don't tell users we're adding it back it'll be confusing.

orbeckst:
I agree, add the warning.

RMeli:
There are different `REMARK`s record depending on _why_ the unit cell is meaningless (215 for solution NMR, 217 for solid-state NMR, 247 for EM, ...). Should I just add the remark under number 285, which just includes information about the unit cell? I'd go for the following (or something similar):

```
REMARK 285 UNITARY VALUES FOR THE UNIT CELL AUTOMATICALLY SET
REMARK 285 BY MDANALYSIS PDBWRITER BECAUSE UNIT CELL INFORMATION
REMARK 285 WAS MISSING.
REMARK 285 PROTEIN DATA BANK CONVENTIONS REQUIRE THAT
REMARK 285 CRYST1 AND SCALE RECORDS BE INCLUDED, BUT THE  VALUES ON
REMARK 285 THESE RECORDS ARE MEANINGLESS.
```

This makes me realize that I should add the `SCALEn` record as well?

testsuite/MDAnalysisTests/coordinates/test_pdb.py
```diff
@@ -291,6 +296,40 @@ def test_write_single_frame_AtomGroup(self, universe2, outfile):
...
+                u.atoms.positions, uout.atoms.positions,
+                self.prec,
+                err_msg="Written coordinates do not "
+                        "agree with original coordinates from frame %d" %
+                        u.trajectory.frame
+            )
+
+        expected_msg = "Unit cell dimensions not found. CRYST1 record set to unitary values."
+
+        assert record[0].message.args[0] == expected_msg
```
RMeli, line 331: Use [pytest match keyword](https://docs.pytest.org/en/latest/warnings.html#warns) instead.

IAlibay:
@RMeli  The Travis py2.7 duecredit failures are rather odd, is this linked with #2697 ?

RMeli:
@IAlibay Weird, I'll look into it. 

#2697 should only "fix" the fact that since I installed `duecredit`, `.duecredit.p` files are created when I run the tests.

RMeli:
Actually, I didn't realize (despite your comment) that one `.duecredit.p` file was actually included in this PR by mistake. I removed it now, let's see if this solves the failing build.

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -631,6 +643,23 @@ def _write_pdb_title(self):
...
+        """
+        Write PDB header.
+
+        The HEADER record is set to :code: `trajectory.header`.
+        The TITLE record explicitly mentions MDAnalysis and contains
+        information about trajectory frame(s).
+        The COMPND record is set to :code:`trajectory.compound`.
+        The REMARKS records are set to :code:`u.trajectory.remarks`
+        The CRYST1 record specifies the unit cell. This record is set to
+        unitary values if unit cell dimensions are nor set.
```
orbeckst, line 655:
```suggestion
        unitary values if unit cell dimensions are not set.
```

package/CHANGELOG
```diff
@@ -22,6 +22,8 @@ mm/dd/yy richardjgowers, kain88-de, lilyminium, p-j-smith, bdice, joaomcteixeira
   * 0.21.0
 
 Fixes
+  * `PDBWriter` skips `CRYST1` record when `u.dimensions` is
```
orbeckst, line 25: It doesn't skip but sets it to unitary values

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -465,6 +476,8 @@ class PDBWriter(base.WriterBase):
     The maximum frame number that can be stored in a PDB file is 9999 and it
     will wrap around (see :meth:`MODEL` for further details).
 
+    If unit cell dimensions are not set, unitary values are used and an
```
orbeckst, line 479: Add explicit explanation what unitary values means: cubic box with sides 1 Å.

Link to the PDB specs that we are following. Say that we are following PDB standard here. It will make it sooo much easier when we start answering questions...

testsuite/MDAnalysisTests/coordinates/test_pdb.py
```diff
@@ -190,6 +191,10 @@ def universe2(self):
     def universe3(self):
         return mda.Universe(PDB)
 
+    @pytest.fixture
+    def universe4(self):
+        return mda.fetch_mmtf("2BBM")
```
orbeckst, line 196: Bundle the file as bzip2-compressed file or even better, just a minimal example.

We should not make our test suite depend on internet connectivity. Also, the file can be changed externally and that will compromise our tests.

RMeli: Very good points, thanks!

RMeli: If I download the `2BBM` PDB file and use that for testing, the test fails because the unit cell is set to `[0.0, 0.0, 0.0, 0.0, 0.0, 0.0]`. @IAlibay Shouldn't `u.dimensions` be `None` after #2645? If I missed something, I can add a check for `[0.0, 0.0, 0.0, 0.0, 0.0, 0.0]` in my fix.

testsuite/MDAnalysisTests/coordinates/test_pdb.py
```diff
@@ -291,6 +299,41 @@ def test_write_single_frame_AtomGroup(self, universe2, outfile):
...
+        Test fix for Issue #2679.
+        """
+
+        u = universe4
+
+        print(u.dimensions)
+
+        expected_msg = "Unit cell dimensions not found. CRYST1 record set to unitary values."
+
+        with pytest.warns(UserWarning, match=expected_msg):
```
RMeli, line 317: This now fails because `u.dimensions` is `[0.0, 0.0, 0.0, 0.0, 0.0, 0.0]`.

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -651,7 +700,24 @@ def _write_pdb_header(self):
             self.REMARK(*remarks)
         except AttributeError:
             pass
-        self.CRYST1(self.convert_dimensions_to_unitcell(u.trajectory.ts))
+
+        if u.dimensions is not None:
```
RMeli, line 704: Check for `[0.0, 0.0, 0.0, 0.0, 0.0, 0.0]`?

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -395,22 +421,20 @@ def _read_frame(self, frame):
...
-                                         dtype=np.float32)  
+                                         line[40:47], line[47:54]],
+                                         dtype=np.float32)
                 except ValueError:
                     warnings.warn("Failed to read CRYST1 record, "
                                   "possibly invalid PDB file, got:\n{}"
                                   "".format(line))
                 else:
-                    if (cell_dims == np.array([1, 1, 1, 90, 90, 90], 
-                        dtype=np.float32)).all():
```
RMeli, line 406: Possibly dangerous `==` comparison with floating point numbers?

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -395,22 +421,20 @@ def _read_frame(self, frame):
...
-                        warnings.warn("1 A^3 CRYST1 record," 
-                                      " this is usually a placeholder in"
-                                      " cryo-em structures. Unit cell"
-                                      " dimensions will not be set.")
+                    if np.allclose(cell_dims, np.array([1.0, 1.0, 1.0, 90.0, 90.0, 90.0])):
+                        warnings.warn("1 A^3 CRYST1 record,"
+                                      " this is usually a placeholder."
```
RMeli, line 433: Removed explicit reference to cryo-EM. Let me know it this is not ok.

testsuite/MDAnalysisTests/coordinates/test_pdb.py
IAlibay: #2645 only attempted to handle cryo-em default cases, if there are more default cases than [1,1,1,90,90,90] then those will need handling.

package/MDAnalysis/coordinates/PDB.py
IAlibay: Question here (and also relates to the above removal of the reference to cryo-em), are the default only [0,0,0,0,0,0] and [1,1,1,90,90,90] or are there more? If it's the latter case, then we might have to start handing things by remark number.

RMeli: Good point. Maybe this should be addressed in a different place? Here I could check for both `None` and the default `[0.0, 0.0, 0.0, 0.0, 0.0, 0.0]`

testsuite/MDAnalysisTests/coordinates/test_pdb.py
RMeli: It dimensions is `None` with `fetch_mmtf("2BBM")` but `[0.0, 0.0, 0.0, 0.0, 0.0, 0.0]` if I load `2bbm.pdb.gz` manually.

package/MDAnalysis/coordinates/PDB.py
IAlibay: Yeah, I'd vote for that. Either way, issues should probably be opened for further changes to the behaviour of PDBReader.

testsuite/MDAnalysisTests/coordinates/test_pdb.py
IAlibay: Yeah, so the MMTFReader specifically sets `_unitcell` to `None`, whilst we otherwise just default to whatever is set to ReaderBase in the PDBReader (as mentioned below, this would probably need alignment in behaviour).

package/MDAnalysis/coordinates/PDB.py
RMeli: Opened #2698 for meaningless/unspecified cell dimensions consistency.

testsuite/MDAnalysisTests/coordinates/test_pdb.py
RMeli: Solved by adding `[0.0, 0.0, 0.0, 0.0, 0.0, 0.0]` to the checks. See #2698.

package/MDAnalysis/coordinates/PDB.py
RMeli: This results in a small change to test `test_cryst_em_warning()`.

RMeli:
`test_atomname_alignment` and `test_atomtype_alignment` are modified because of the additional `REMARKS` introduced with this PR.

package/MDAnalysis/coordinates/PDB.py
orbeckst: In this particular case it's fine, I think, because it's really only this exact unitcell that we are using as the magic marker.

package/CHANGELOG
```diff
@@ -22,6 +22,8 @@ mm/dd/yy richardjgowers, kain88-de, lilyminium, p-j-smith, bdice, joaomcteixeira
   * 0.21.0
 
 Fixes
+  * `PDBWriter` writeS unitary `CRYST1` record (cubic box with sides of 1Å)
```
orbeckst, line 25:
```suggestion
  * `PDBWriter` writes unitary `CRYST1` record (cubic box with sides of 1 Å)
```

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -217,6 +216,17 @@ class PDBReader(base.ReaderBase):
     79 - 80        LString(2)    charge       Charge  on the atom.
     =============  ============  ===========  =============================================
 
+    Notes
+    -----
+    If a system does not have unit cell parameters (such as in EM structures),
```
orbeckst, line 221: EM = electron microscopy

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -217,6 +216,17 @@ class PDBReader(base.ReaderBase):
...
+    the PDB file format requires the CRYST1 field to be provided with unitary
+    values (cubic box with sides of 1Å) and an appropriate REMARK. If unitary
```
orbeckst, line 223: space between value and unit: 1 Å

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -231,7 +241,8 @@ class PDBReader(base.ReaderBase):
     .. versionchanged:: 0.20.0
        Strip trajectory header of trailing spaces and newlines
     .. versionchanged:: 1.0.0
-       User warning for CRYST1 cryo-em structures
+       Raise user warning for CRYST1 record with unitary valuse
+       (cubic box with sides of 1Å) and do not set cell dimensions.
```
orbeckst, line 245: 1 Å

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -364,6 +375,21 @@ def _read_next_timestep(self, ts=None):
         return self._read_frame(frame)
 
     def _read_frame(self, frame):
+        """
+        Read frame from PDB file.
+
+        Notes
+        -----
+        When the CRYST1_ record has unitary values (cubic box with sides of 1Å),
```
orbeckst, line 383: 1 Å

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -364,6 +375,21 @@ def _read_next_timestep(self, ts=None):
...
+        cell dimensions are considered fictitious. An user warning is raised
+        and cell dimensions are not set.
```
orbeckst, line 385: Can we be more precise here or are we waiting for a solution for #2698  ?

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -364,6 +375,21 @@ def _read_next_timestep(self, ts=None):
...
+
+        .. _CRYST1: http://www.wwpdb.org/documentation/file-format-content/format33/sect8.html#CRYST1
+
+        .. versionchanged:: 1.0.0
+           Raise user warning for CRYST1 record with unitary valuse
+           (cubic box with sides of 1Å) and do not set cell dimensions.
```
orbeckst, line 391: 1 Å

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -217,6 +216,17 @@ class PDBReader(base.ReaderBase):
     79 - 80        LString(2)    charge       Charge  on the atom.
     =============  ============  ===========  =============================================
 
+    Notes
+    -----
+    If a system does not have unit cell parameters (such as in EM structures),
+    the PDB file format requires the CRYST1 field to be provided with unitary
```
orbeckst, line 222: Link CRYST1

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -217,6 +216,17 @@ class PDBReader(base.ReaderBase):
...
+    values (cubic box with sides of 1Å) and an appropriate REMARK. If unitary
+    values are found within the CRYST1 field, :code:`PDBReader` will not set
+    unit cell dimensions and it
+    will warn the user.
+
+    .. _CRYST1: http://www.wwpdb.org/documentation/file-format-content/format33/sect8.html#CRYST1
```
orbeckst, line 228: Does sphinx complain about a duplicated link between here and further down? If so, keep only 1.

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -364,6 +375,21 @@ def _read_next_timestep(self, ts=None):
...
+        """
+        Read frame from PDB file.
+
+        Notes
+        -----
+        When the CRYST1_ record has unitary values (cubic box with sides of 1Å),
+        cell dimensions are considered fictitious. An user warning is raised
+        and cell dimensions are not set.
+
+        .. _CRYST1: http://www.wwpdb.org/documentation/file-format-content/format33/sect8.html#CRYST1
```
orbeckst, line 387: possibly duplicated reST link

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -395,22 +421,20 @@ def _read_frame(self, frame):
...
-                    if (cell_dims == np.array([1, 1, 1, 90, 90, 90], 
-                        dtype=np.float32)).all():
-                        warnings.warn("1 A^3 CRYST1 record," 
-                                      " this is usually a placeholder in"
-                                      " cryo-em structures. Unit cell"
-                                      " dimensions will not be set.")
+                    if np.allclose(cell_dims, np.array([1.0, 1.0, 1.0, 90.0, 90.0, 90.0])):
+                        warnings.warn("1 A^3 CRYST1 record,"
+                                      " this is usually a placeholder."
+                                      " Unit cell dimensions will not be set.")
```
orbeckst, line 434: Is it true that "unset" simply means that we default to the reader initialization and that will have to be harmonized by fixing  #2698 ?

package/MDAnalysis/coordinates/PDB.py
orbeckst: This is the place where docs are really needed, basically the ones on CRYST1 that you had above.

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -631,7 +656,31 @@ def _write_pdb_title(self):
...
+        """
+        Write PDB header.
+
+        The HEADER_ record is set to :code: `trajectory.header`.
+        The TITLE record explicitly mentions MDAnalysis and contains
+        information about trajectory frame(s).
+        The COMPND record is set to :code:`trajectory.compound`.
+        The REMARKS records are set to :code:`u.trajectory.remarks`
+        The CRYST1_ record specifies the unit cell. This record is set to
+        unitary values (cubic box with sides of 1Å) if unit cell dimensions
```
orbeckst, line 668: 1 Å

orbeckst: These docs are good but never visible to users in the online docs. Put anything (such as the CRYST1 stuff) into the class docs.

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -631,7 +656,31 @@ def _write_pdb_title(self):
...
+        .. _HEADER: http://www.wwpdb.org/documentation/file-format-content/format33/sect2.html#HEADER
+        .. _TITLE: http://www.wwpdb.org/documentation/file-format-content/format33/sect2.html#TITLE
+        .. _COMPND: http://www.wwpdb.org/documentation/file-format-content/format33/sect2.html#COMPND
+        .. _REMARKS: http://www.wwpdb.org/documentation/file-format-content/format33/remarks.html
+        .. _CRYST1: http://www.wwpdb.org/documentation/file-format-content/format33/sect8.html#CRYST1
+
+        .. versionchanged: 1.0.0
+           Fix writing of PDB file without unit cell dimensions (Issue #2679).
+           If cell dimensions are not found, unitary values (cubic box with
+           sides of 1Å) are used (PDB standard).
```
orbeckst, line 680: 1 Å

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -651,7 +700,25 @@ def _write_pdb_header(self):
             self.REMARK(*remarks)
         except AttributeError:
             pass
-        self.CRYST1(self.convert_dimensions_to_unitcell(u.trajectory.ts))
+
+        if u.dimensions is None or np.allclose(u.dimensions, np.zeros(6)):
```
orbeckst, line 704: Maybe add a comment for ourselves that we internally consider both None and zeros equivalent, and that #2698 will address this.

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -651,7 +700,25 @@ def _write_pdb_header(self):
...
+            # Unitary unit cell by default (PDB standard)
```
orbeckst, line 705: Add link in comment to where the standard is described (CRYST1 link?)

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -651,7 +700,25 @@ def _write_pdb_header(self):
...
+            self.CRYST1(np.array([1.0, 1.0, 1.0, 90.0, 90.0, 90.0]))
+
+            # Add CRYST1 REMARK (285)
+            self.REMARK("285 UNITARY VALUES FOR THE UNIT CELL AUTOMATICALLY SET")
+            self.REMARK("285 BY MDANALYSIS PDBWRITER BECAUSE UNIT CELL INFORMATION")
+            self.REMARK("285 WAS MISSING.")
+            self.REMARK("285 PROTEIN DATA BANK CONVENTIONS REQUIRE THAT")
+            self.REMARK("285 CRYST1 AND SCALE RECORDS BE INCLUDED, BUT THE VALUES ON")
```
orbeckst, line 713: Just state that CRYST1 records be included and don't mention SCALE. Then add a _comment_ to the code saying that we decided against including SCALE records anywhere as we are only implementing a subset of the PDB standard. (We do deal with CRYST1 so we adhere to any standard with regards to that record.)

testsuite/MDAnalysisTests/coordinates/test_pdb.py
```diff
@@ -190,6 +191,23 @@ def universe2(self):
...
+            [PDB_NOCRYST, np.zeros(6)],
+            [MMTF_NOCRYST, None]
+        ])
+    def universe_and_expected_dims(self, request):
+        """
+        File with meaningless CRYST1 record and expected dimensions.
+
+        Notes
+        -----
+        This will need to be made consistent, see Issue #2698
```
orbeckst, line 204: Great note

package/MDAnalysis/coordinates/PDB.py
RMeli: I added the following since for the PDB reader cell dimensions are given by `_init_unitcell` in `Timestep`

> cell dimensions are set to :code:`np.array([0.0, 0.0, 0.0, 0.0, 0.0, 0.0])` (see Issue #2698)

package/MDAnalysis/coordinates/PDB.py
RMeli: Should I revert it back?

package/MDAnalysis/coordinates/PDB.py
RMeli: Yes. In this case they are actually set to `np.array([0.0, 0.0, 0.0, 0.0, 0.0, 0.0])` so I changed the code to reflect that:

```
                        # FIXME: Dimensions set to zeros.
                        # FIXME: This might change with Issue #2698
                                      " this is usually a placeholder."
                                      " Unit cell dimensions will be set"
                                      " to zeros.")
```

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -217,6 +216,17 @@ class PDBReader(base.ReaderBase):
     79 - 80        LString(2)    charge       Charge  on the atom.
     =============  ============  ===========  =============================================
 
+    Notes
+    -----
+    If a system does not have unit cell parameters (such as in electron
+    microscopy structures), the PDB file format requires the CRYST1_ field to
+    be provided with unitary values (cubic box with sides of 1 Å) and an
+    appropriate REMARK. If unitary values are found within the CRYST1_ field,
+    :code:`PDBReader` will not set unit cell dimensions and it will warn the
```
IAlibay, line 225: I know this is going to get changed with #2698, but is it worth temporarily detailing that this means dimensions will return [0,0,0,0,0,0]? (i.e. "will not set unit cell dimensions (leaving it at the default [0,0,0,0,0,0[ value)...".

My thought here is mainly "if we have a big gap between this PR and the next" (e.g. a release), then more details is better.

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -465,6 +491,9 @@ class PDBWriter(base.WriterBase):
     The maximum frame number that can be stored in a PDB file is 9999 and it
     will wrap around (see :meth:`MODEL` for further details).
 
+    The CRYST1_ record specifies the unit cell. This record is set to
+    unitary values (cubic box with sides of 1 Å) if unit cell dimensions
+    are not set.
```
IAlibay, line 496: As per my comment above, it might be good to be a bit more explicit to users here (i.e. "not set" in practical terms means either `None` or [0,0,0,0,0,0]).

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -651,7 +699,30 @@ def _write_pdb_header(self):
...
+
+        # FIXME: Values for meaningless cell dimensions are not consistent.
+        # FIXME: See Issue #2698. Here we check for both None and zeros
+        if u.dimensions is None or np.allclose(u.dimensions, np.zeros(6)):
+            # Unitary unit cell by default. See PDB standard:
+            # http://www.wwpdb.org/documentation/file-format-content/format33/sect8.html#CRYST1
+            self.CRYST1(np.array([1.0, 1.0, 1.0, 90.0, 90.0, 90.0]))
+
+            # Add CRYST1 REMARK (285)
+            # The SCALE record is not included
```
IAlibay, line 711: Thought here, do we/should we be accounting for the SCALE record in the PDBReader?

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -651,7 +699,30 @@ def _write_pdb_header(self):
...
+            self.REMARK("285 UNITARY VALUES FOR THE UNIT CELL AUTOMATICALLY SET")
+            self.REMARK("285 BY MDANALYSIS PDBWRITER BECAUSE UNIT CELL INFORMATION")
+            self.REMARK("285 WAS MISSING.")
+            self.REMARK("285 PROTEIN DATA BANK CONVENTIONS REQUIRE THAT")
+            self.REMARK("285 CRYST1 RECORD IS INCLUDED, BUT THE VALUES ON")
+            self.REMARK("285 THIS RECORD ARE MEANINGLESS.")
+
+            warnings.warn("Unit cell dimensions not found. "
+                          "CRYST1 record set to unitary values."
+                          )
```
IAlibay, line 722: Super nit-picky, but is keeping a whole line for a `)` really necessary 😆?

testsuite/MDAnalysisTests/datafiles.py
```diff
@@ -187,7 +187,9 @@
     "GMX_TOP_BAD", # file with an #include that doesn't exist
     "ITP_no_endif", # file missing an #endif
     "PDB_CRYOEM_BOX", #Issue 2599
-    "PDB_CHECK_RIGHTHAND_PA" # for testing right handedness of principal_axes
+    "PDB_CHECK_RIGHTHAND_PA", # for testing right handedness of principal_axes
+    "PDB_NOCRYST", # File with meaningless CRYST1 record (Issue #2679, PR #2685)
```
IAlibay, line 191: A thought here, probably not worth too much attention. In order to reduce the amount of data we have to store, would it make sense to use `PDB_CRYOEM_BOX` instead of `PDB_NOCRYST`? I realise that 2BBM was what was reported in the original issue and that it's NMR whilst the other Cryo-EM, but we are looking for the same CRYST1 record right? Alternatively a tiny NMR structure could be used like 6QYR.

RMeli:
* [ ] Fix failing `TestHoleAnalysis`

```bash
______________________ TestHoleAnalysis.test_output_level ______________________

self = <MDAnalysisTests.analysis.test_hole2.TestHoleAnalysis object at 0x7fdbbd33bef0>

tmpdir = local('/tmp/pytest-of-travis/pytest-0/test_output_level1')

universe = <Universe with 264 atoms>

    def test_output_level(self, tmpdir, universe):

        with tmpdir.as_cwd():

            with pytest.warns(UserWarning) as rec:

                h = hole2.HoleAnalysis(universe,

                                       output_level=100)

                h.run(start=self.start,

                      stop=self.stop, random_seed=self.random_seed)

>           assert len(rec) == 3

E           assert 5 == 3

E            +  where 5 = len(WarningsChecker(record=True))
```

package/MDAnalysis/coordinates/PDB.py
RMeli: Sure. I think I already changed it elsewhere but obviously missed a few spots.

testsuite/MDAnalysisTests/datafiles.py
RMeli: Yes, @orbeckst mentioned something on those lines before (and I forgot):

> Bundle the file as bzip2-compressed file or even better, just a minimal example.

I'll switch the `PDB_NOCRYST` with `PDB_CRYOEM_BOX` and change the MMTF file to something smaller (`6QYR`).

lilyminium:
🤦‍♀️ Sorry @RMeli, that's just a terribly written test (that I wrote) that has caused problems before. Not checking how many warnings you get and just looking for ```assert 'needs to be < 3' in rec[0].message.args[0]``` or ```assert any('needs to be < 3' in r.message.args[0] for r in rec)``` might be more robust. I can make this a separate issue/fix if you like.

orbeckst:
This PR lgtm.

@IAlibay can I leave it to you? You still have some discussions going.

IAlibay:
> @IAlibay can I leave it to you? You still have some discussions going.

Can do @orbeckst. 

As far as I see it, there's only a few things left to do:
 - [ ] Fix the failing test - @lilyminium since it's a small-ish fix, IMHO it's better to get it done here rather than merging on a failing Travis, but it's up to you both.
 - [ ] Check the SCALE record stuff, if necessary raise a new issue that looks to directly tackle SCALE records on read/write.
 - [ ] Be a little bit more explicit in the class docs (i.e. define what `unset` means).
 - [ ] Use minimal Cryo-EM/NMR example files (if necessary).

RMeli:
@lilyminium I just copy-pasted the test here as a reminder to fix it. I think fixing it here is ok since it's a very small thing and it's related to this PR. I was thinking  of using pytest's `match`, I just need to figure out how to use it with multiple warnings (if at all possible).

package/MDAnalysis/coordinates/PDB.py
RMeli: Looks like we should add the `SCALE` record as well?

> For NMR, fiber diffraction, and EM entries, SCALE is given as an identity matrix with no translation.

RMeli:
I think I addressed all the issues apart from the `SCALE` record. The `PDBReader` is not reading the `SCALE` record (or any other record of the _Crystallographic and Coordinate Transformation Section_):

```
    The following *PDB records* are parsed (see `PDB coordinate section`_ for
    details):

     - *CRYST1* for unitcell A,B,C, alpha,beta,gamma
     - *ATOM* or *HETATM* for serial,name,resName,chainID,resSeq,x,y,z,occupancy,tempFactor
     - *HEADER* (:attr:`header`), *TITLE* (:attr:`title`), *COMPND*
       (:attr:`compound`), *REMARK* (:attr:`remarks`)
     - all other lines are ignored
```

If we want `PDBReader` to be able to read other records I think we can open another issue to discuss, bit it's probably outside the scope of this PR.

testsuite/MDAnalysisTests/analysis/test_hole2.py
```diff
@@ -307,10 +308,13 @@ def test_output_level(self, tmpdir, universe):
                                        output_level=100)
                 h.run(start=self.start,
                       stop=self.stop, random_seed=self.random_seed)
-            assert len(rec) == 3
-            assert 'needs to be < 3' in rec[0].message.args[0]
-            assert 'has no dt information' in rec[1].message.args[0]
-            assert 'has no dt information' in rec[2].message.args[0]
+
+            assert len(rec) == 5
+
```
IAlibay, line 313: Any reason for the insertion of an extra empty line here?

package/MDAnalysis/coordinates/PDB.py
IAlibay: With further thought, this wouldn't be a bad thing to do here, As per your comment https://github.com/MDAnalysis/mdanalysis/pull/2685#issuecomment-635984874 it would be good to raise an issue about this for reading / non-default writing if possible.

testsuite/MDAnalysisTests/coordinates/test_pdb.py
```diff
@@ -291,6 +309,44 @@ def test_write_single_frame_AtomGroup(self, universe2, outfile):
...
+            assert np.allclose(u.dimensions, expected_dims)
+
+        expected_msg = "Unit cell dimensions not found. CRYST1 record set to unitary values."
+
+        with pytest.warns(UserWarning, match=expected_msg):
+
+            u.atoms.write(outfile)
+
+        uout = mda.Universe(outfile)
+
```
IAlibay, line 335: Ideally at this point the reader should also recognise this as a default box, could you check either that the dimensions returned are [0, 0, 0, 0, 0, 0] or that you get raise the "1 A^3 CRYST1 record" warning?

testsuite/MDAnalysisTests/analysis/test_hole2.py
RMeli: Personal preference, but I reverted it back.

RMeli:
After a private discussion @IAlibay and I convened that adding the SCALE record and going towards a fully standard-compliant `PDBWriter` is outside the scope of this PR, which solves a very specific issue.

See  #2711 for further discussion.

package/CHANGELOG
```diff
@@ -23,6 +23,8 @@ mm/dd/yy richardjgowers, kain88-de, lilyminium, p-j-smith, bdice, joaomcteixeira
   * 0.21.0
 
 Fixes
+  * `PDBWriter` writes unitary `CRYST1` record (cubic box with sides of 1 Å)
+    when `u.dimensions` is `None` (Issue #2679, PR #2685)
```
IAlibay, line 27: Sorry, just saw this last minute as I was writing the squash merge entry.
This isn't fully true at the moment, could you change this to "when `u.dimensions` is `None` or `np.zero(6)`" (or something to that effect)?

PR #2346, opened 2019-09-08
WIP, TST: add 32-bit support

tylerjereddy:
Fixes #2342

Currently WIP, with my fork feature branch `fix_32bit` reducing total failures to 137 from around 1000 or so. I will first aim to add 32-bit testing to CI.

I'm currently stuck trying to get the `testsuite/MDAnalysisTests/core/test_wrap.py` tests passing--some of the C++ stuff in `_cutil.pyx` is not friendly for Python maintainers...

Anyway, will aim to get the large number of 32 bit test failures reproduced in CI here for demo purposes, then will try pushing in the WIP improvements to drop the fail count down...

PR #2689, opened 2020-05-21
bugfix for issue #2687

### LINKED ISSUE #2687: AtomGroup.residues
## Expected behavior ##
COMMAND - list(u.atoms[100:130].residues)
[<Residue LEU, 6>, <Residue GLY, 7>, <Residue ALA, 8>]

## Actual behavior ##
  File "exmple.py", line 11, in <module>
    print(list(u.atoms.n_residues))
  File "c:\users\user\opens\mdanalysis\package\MDAnalysis\core\groups.py", line 2301, in n_residues
    return len(self.residues)
  File "c:\users\user\opens\mdanalysis\package\MDAnalysis\core\groups.py", line 2261, in residues
    rg = self.universe.residues[unique_int_1d(self.resindices)]
  File "MDAnalysis\lib\_cutil.pyx", line 52, in MDAnalysis.lib._cutil.unique_int_1d
ValueError: Buffer dtype mismatch, expected 'int64_t' but got 'long

The reason for the error is that the unique_int_1d requires a numpy array of int64 datatype while only a simple array is being passed in the function unique_int_1d() so before passing the value it needs to be typed as numpy int64 array

``` python
import MDAnalysis as mda

u = mda.Universe(PSF, DCD)
print(list(u.atoms[1:50].residues))


....

```

MDAnalysis version: 0.20.2-dev0
Python version:  Python 3.8.2
Operating System: Windows 10(Home)
- Which operating system?

darmis007:
Fixes #2687

 - The 'self.resindices' was not being passed as a numpy array of int64 type in the argument of AtomGroup 's residues' function which is relayed to _cutil.pyx whose unique_int_1d function requires a numpy array of int64 datatype . The commit intends to solve the particular issue.

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/core/groups.py
+++ b/package/MDAnalysis/core/groups.py
@@ -2258,7 +2258,7 @@ class AtomGroup(GroupBase):
         """A sorted :class:`ResidueGroup` of the unique
         :class:`Residues<Residue>` present in the :class:`AtomGroup`.
         """
-        rg = self.universe.residues[unique_int_1d(self.resindices)]
+        rg = self.universe.residues[unique_int_1d(np.array((self.resindices),dtype='int64'))]
         rg._cache['isunique'] = True
         rg._cache['unique'] = rg
         return rg


package/MDAnalysis/core/groups.py
```diff
@@ -2258,7 +2258,7 @@ def residues(self):
         """A sorted :class:`ResidueGroup` of the unique
         :class:`Residues<Residue>` present in the :class:`AtomGroup`.
         """
-        rg = self.universe.residues[unique_int_1d(self.resindices)]
+        rg = self.universe.residues[unique_int_1d(np.array((self.resindices),dtype='int64'))]
```
orbeckst, line 2261: Use `ndarray.astype()` to save a conversion when it is already correctly int64, and using the dtype object directly:
```suggestion
        rg = self.universe.residues[unique_int_1d(self.resindices.astype(np.int64)]
```

orbeckst:
By the way, it's not clear yet that the code above will really fix the issue.

1. Did you try it out for yourself and does it make the problem reported in issue #2687 go away?
2. We need a test that _fails on our Windows CI system_ ("Appveyor") without the fix. You'd need to add the test, then disable the fix (put back the original code), show that Appveyor fails, and then add the fix and then Appveyor should pass all tests, including the new one.

darmis007:
sir the code above gave the required result without any error on my windows system
sir i will write the required test and get back to you

IAlibay:
@darmis007 this PR closure is rather sudden/unexpected, could you let us know what happened?

tylerjereddy:
The exact change Oli is suggesting is part of https://github.com/MDAnalysis/mdanalysis/pull/2346

This is looking a lot like a 32-bit Python interpreter on Windows to me.

PR #2693, opened 2020-05-22
Testforissue#2687

### LINKED ISSUE #2687: AtomGroup.residues

darmis007:
Fixes #2687 

 - I have added required tests regarding issue #2687 and my suggested pull request #2689 

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?


testsuite/MDAnalysisTests/lib/test_cutil.py
```diff
@@ -45,6 +52,10 @@ def test_unique_int_1d(values):
     assert type(res) == type(ref)
     assert res.dtype == ref.dtype
 
+def test_array_unique_int_1d():
+    universe=u()
+    assert unique_int_1d(universe.atoms[100:130].resindices) == [5,6,7]
```
orbeckst, line 57: For array comparisons, use the numpy functions, like
```suggestion
    assert_equal(unique_int_1d(universe.atoms[100:130].resindices),  [5,6,7])
```
(Make sure that `assert_equal` is imported from `numpy.testing`)

testsuite/MDAnalysisTests/core/test_groups.py
```diff
@@ -885,6 +885,10 @@ def test_segmentgroup_repr(self, u):
     def test_segmentgroup_str(self, u):
         sg = u.segments[:10]
         assert str(sg) == '<SegmentGroup [<Segment 4AKE>]>'
+    
+    def test_atomgroupresidue_dtype(self,u):
```
orbeckst, line 889: PEP8: space after comma

testsuite/MDAnalysisTests/lib/test_cutil.py
```diff
@@ -37,6 +40,10 @@
     [1, 2, 2, 4, 4, 6, ],  # duplicates, monotonic
     [1, 2, 2, 6, 4, 4, ],  # duplicates, non-monotonic
 ))
+
+def u():
+    return mda.Universe(PSF, DCD)
+
```
orbeckst, line 46: remove, when testing `lib` functions just use normal arrays because all these functions work without universes and AtomGroups

testsuite/MDAnalysisTests/lib/test_cutil.py
```diff
@@ -45,6 +52,10 @@ def test_unique_int_1d(values):
     assert type(res) == type(ref)
     assert res.dtype == ref.dtype
 
+def test_array_unique_int_1d():
+    universe=u()
```
orbeckst, line 56: Instead of a universe, create numpy arrays with the appropriate dtypes so that you can test behavior _without_ needing a Universe.

testsuite/MDAnalysisTests/lib/test_cutil.py
orbeckst: `universe.atoms[100:130].resindices` should be replaced with something like `np.array([6, 5, 5, 7], dtype=np.int32)` or something like that – check what you get on your system  as the dtype of `universe.atoms[100:130].resindices.dtype`.

testsuite/MDAnalysisTests/core/test_groups.py
```diff
@@ -885,6 +885,10 @@ def test_segmentgroup_repr(self, u):
...
+        rg = list(u.atoms[100:130].residues)
+        assert rg == '[<Residue LEU, 6>, <Residue GLY, 7>, <Residue ALA, 8>]'
```
orbeckst, line 891: This will fail. `rg` is a list of residue objects and not strings.

Try
```suggestion
assert len(rg) == 3
```
or something like that.

testsuite/MDAnalysisTests/core/test_groups.py
```diff
@@ -885,6 +885,10 @@ def test_segmentgroup_repr(self, u):
     def test_segmentgroup_str(self, u):
         sg = u.segments[:10]
         assert str(sg) == '<SegmentGroup [<Segment 4AKE>]>'
+    
+    def test_atomgroupresidue_dtype(self,u):
+        rg = list(u.atoms[100:130].residues)
```
orbeckst, line 890: If we know that this can fail then we want to properly fail the test:
```suggestion
       try:
           rg = list(u.atoms[100:130].residues)
       except ValueError:
           raise AssertionError("Issue #2687 failure")
```

darmis-blip:
Thank you so much sir for your valuable suggestions i will get back with the required changes

darmis007:
Thank you so much sir for the valuable suggestions i will get back to with the required changes

testsuite/MDAnalysisTests/lib/test_cutil.py
```diff
@@ -46,6 +48,14 @@ def test_unique_int_1d(values):
     assert res.dtype == ref.dtype
 
 
+def test_array_unique_int_1d():
+    try:
+        res_indices = np.array([6, 5, 5, 7], dtype=np.int32)
+        assert_equal(unique_int_1d(res_indices),  [5, 6, 7])
```
lilyminium, line 54: I know @orbeckst suggested this, but a dtype of `np.int_` would be more representative of the `resindices` array across Windows and Unix systems, as it is created for most topology formats with the dtype `np.int`, i.e. `np.int_`. [`np.int_` is either an int32 or int64 depending on the platform](https://numpy.org/devdocs/user/basics.types.html), which is likely why this has not been an issue earlier.

https://github.com/MDAnalysis/mdanalysis/blob/b341090aaa0bace9b946316ebf1e5bf965b026da/package/MDAnalysis/topology/base.py#L203

As it is this test is guaranteed to fail from a type mismatch.

darmis007: Yes i will surely make the change and update the PR

darmis007:
The following tests failed in AppVeyor
FAILED core\test_groups.py::TestReprs::test_atomgroupresidue_dtype - Assertio...
FAILED lib\test_cutil.py::test_array_unique_int_1d - ValueError:

testsuite/MDAnalysisTests/lib/test_cutil.py
orbeckst: I wasn't sure what the right choice for the `dtype` should have been, so by all means, follow @lilyminium 's suggestion.

On a more general note: should we avoid the platform-dependent types in MDA and be explicitly using the numpy-guaranteed types instead?

testsuite/MDAnalysisTests/core/test_groups.py
```diff
@@ -886,17 +890,26 @@ def test_segmentgroup_str(self, u):
         sg = u.segments[:10]
         assert str(sg) == '<SegmentGroup [<Segment 4AKE>]>'
 
+    def test_atomgroupresidue_dtype(self, u):
+        try:
+            rg = list(u.atoms[100:130].residues)
+            assert len(rg) == 3
```
tylerjereddy, line 896: normally one would let `pytest`/Python propagate the exception and just reference the issue in a comment `# see gh-2687`

testsuite/MDAnalysisTests/lib/test_cutil.py
```diff
@@ -46,6 +48,14 @@ def test_unique_int_1d(values):
     assert res.dtype == ref.dtype
 
 
+def test_array_unique_int_1d():
+    try:
+        res_indices = np.array([6, 5, 5, 7], dtype=np.int_)
+        assert_equal(unique_int_1d(res_indices),  [5, 6, 7])
+    except ValueError:
+        raise AssertionError("Issue #2687 failure")
```
tylerjereddy, line 56: I would just let `pytest`/Python handle the exception and use a `# see gh-2687` comment at the start of the test function

testsuite/MDAnalysisTests/lib/test_cutil.py
tylerjereddy: > should we avoid the platform-dependent types in MDA

Probably, to the extent reasonable

testsuite/MDAnalysisTests/core/test_groups.py
tylerjereddy: > we want to properly fail the test

Hmm, maybe we have different views on this--I don't really mind, but I rarely see custom handling instead of letting `pytest` report the exception? Maybe this is becoming more common in MDA?

testsuite/MDAnalysisTests/lib/test_cutil.py
orbeckst: The reason why I suggested this approach is because if we let the code raise ValueError then the test will fail differently that if it raised AssertionError. Because we know about one specific failure I suggested to make this known failure clear.

testsuite/MDAnalysisTests/core/test_groups.py
orbeckst: Only `rg = list(u.atoms[100:130].residues)` should be inside try/exept, everything else should be after it!

tylerjereddy:
People have differing views on this, but there are a bunch of unrelated whitespace cleanups in this PR too.

testsuite/MDAnalysisTests/core/test_groups.py
orbeckst: The reason why I suggested this approach is because if we let the code raise ValueError then the test will fail differently that if it raised AssertionError. Because we know about one specific failure I suggested to make this known failure clear.

orbeckst:
I only had a quick look but most of the white space cleanups appear reasonable/PEP8 compliant. There are some extra empty lines which are not necessary, I think.

What's your opinion @tylerjereddy ?

tylerjereddy:
I think unrelated cleanups make PRs harder to review, but I'm often in the minority on that, since i.e., SciPy often only allows such cleanups if you happen to see them while editing for other stuff..

testsuite/MDAnalysisTests/core/test_groups.py
orbeckst: I can see an advantage of just letting the exception trigger the failure because in this way you get more information and we don't accidentally mask some other failure we didn't think about.

You also have plenty of experience with numpy/scipy so I am happy to defer to your opinion here.

@darmis007 – let's ignore my suggestion and go back to simply
```python
# If this fails with ValueError see gh-2687
rg = list(u.atoms[100:130].residues)
assert ...
```
(no special exception handling)

testsuite/MDAnalysisTests/core/test_groups.py
orbeckst: Please see https://github.com/MDAnalysis/mdanalysis/pull/2693#discussion_r429423973 – I now concur with @tylerjereddy to _not_ use the try/except.

testsuite/MDAnalysisTests/lib/test_cutil.py
tylerjereddy: I'm a little unclear on the suggestion to use `np.int_` here. It seems like we only have a problem when `np.int32` is used because of the platform, and it seems like we'd want to test for that regardless of what Appveyor happens to use for it.

So, if Appveyor does someday not use `np.int32` for `np.int_` anymore, we'd effectively stop testing for the regression in our CI.

Perhaps just parametrize over the two and then both people are right :)

tylerjereddy:
> but most of the white space cleanups appear reasonable/PEP8 compliant

Yeah, let's agree to keep those then--to have a clear message for review here.

testsuite/MDAnalysisTests/lib/test_cutil.py
lilyminium: > So, if Appveyor does someday not use np.int32 for np.int_ anymore, we'd effectively stop testing for the regression in our CI.

I don't really follow -- aren't the platform differences (and the fact that apparently no one has tried this in Windows until now) the source of the bug? I would rather keep this test and stick an `xfail` over it to track changes in the Appveyor types. Passing an int32 array to a function that wants an int64 array should always fail, right, even on Unix, and probably not provide useful information.

Anyway, oops, looking at `change_squash()` was too upstream; the arrays get converted to `np.intp` in TransTable.

https://github.com/MDAnalysis/mdanalysis/blob/b341090aaa0bace9b946316ebf1e5bf965b026da/package/MDAnalysis/core/topology.py#L219

tylerjereddy: We want to simulate the presence of *any* problematic platform that might substitute `np.int32` in `np.int_`, so I think always using `np.int32` makes sense to catch the regression consistently.

The currently-proposed fix that I see is just going to coerce to `np.int64` without a platform-specific guard, which seems right, since that can be made a no-op with `astype` when the input is already `int64`.

If you really want to decorate the test on POSIX machines, you could skip it when not windows, but probably not worth it--it should pass on all platforms really.

lilyminium:
Style decisions aside, it is interesting to note that the only test that failed was `test_array_unique_int_1d()` when an `np.int_` array is passed to `unique_int_1d`. We see [the same error](https://ci.appveyor.com/project/orbeckst/mdanalysis/builds/33054671/job/6snbi67sgvog5jwr#L1212) as in #2687:

```
>   ???
E   ValueError: Buffer dtype mismatch, expected 'int64_t' but got 'long'
MDAnalysis\lib\_cutil.pyx:52: ValueError
```

So, just calling `AtomGroup.residues` works. This makes sense because unless I missed [further type conversion after creating](https://github.com/MDAnalysis/mdanalysis/pull/2693#discussion_r429456968) the `TransTable` for the topology, `resindices` are actually usually `np.intp`. (`intp` is the type supposed to be used for indexing, as well). It depends on ``ssize_t``, so should be 64 bit on a 64 bit system. This may also explain why @IAlibay is not seeing the bug. However, it raises the question of why @darmis007 is getting different behaviour.

Solution-wise, it seems like a good idea to keep the array `np.intp`. Perhaps changing the type of `unique_int_1d` to `intptr_t` would work better?

tylerjereddy:
Right, the single failure was noted above.

> However, it raises the question of why @darmis007 is getting different behaviour.

Was a 32-bit Python interpreter/setup decisively ruled out there? The 64-bit operating system may not necessarily rule out a 32-bit Python setup and I believe 32-bit Python may indeed still be quite common on Windows.

When I looked into 32-bit support, a huge number of coercions had to be added to avoid low-level changes to the C code, so adding the coercion doesn't seem surprising to me.

tylerjereddy:
see also: https://github.com/MDAnalysis/mdanalysis/pull/2346

lilyminium:
Thanks @tylerjereddy, I see none of these are new issues; you definitely know more here than I do. According to [the python wiki](https://wiki.python.org/moin/BeginnersGuide/Download) the 32-bit version is the default one downloaded by the big yellow button, so, Windows support should probably take that into account...

testsuite/MDAnalysisTests/lib/test_cutil.py
lilyminium: But using `np.int32` here will make the test fail on all platforms -- the test passes the array directly into `unique_int_1d`, without trying to convert it to int64. Unless I'm misunderstanding and that's the point.

tylerjereddy: Right, I think the test should probably be removed since `unique_int_1d` is not public?

It is on a private import path: `MDAnalysis.lib._cutil.unique_int_1d`

Or do you want users to be able to use it directly? Then the test is "ok" and we'd need a lower-level patch there, but I suspect that's intended to be private and the test should be removed.

If you agree, that would mean we're back to the stage that CI is not able to reproduce the failure for the first test, which is closer to the reported issue. For that I suspect we need a 32-bit entry in CI, but I don't know if we want to go down that rabbit hole based on the other PR/discussions there. I think it may be a good idea, but it is a maintenance burden to consider.

I'll see if I can reproduce locally now with a 32-bit win interpreter.

tylerjereddy: That was a very good point though, I didn't actually look close enough to notice that the proposed patch wouldn't have an effect for the second test.

tylerjereddy:
I've added the 32-bit label here, since that is pretty clear now. There's probably some discussion to be had about level of 32-bit support before we proceed too far, but this was a detailed/useful discussion so far I think!

darmis007:
All builds of the current commit are passing except one in Travis CI on checking the logs no fail tests are to be seen,the recent commit passed without any build problem in this one the try and except blocks are removed just

richardjgowers:
I'm a bit late to the party here, but I looked through the code underneath and found that we're pretty good at declaring the integer type for groups as `np.intp` [eg here and throughout](https://github.com/MDAnalysis/mdanalysis/blob/develop/package/MDAnalysis/core/groups.py#L474).  This is the integer large enough to hold a pointer on a given platform, ie 64 or in this case 32 bits. [so much fun](https://numpy.org/devdocs/user/basics.types.html#array-types-and-conversions-between-types)

I think the problem happens when we later assume that this is `np.int64` [here](https://github.com/MDAnalysis/mdanalysis/blob/develop/package/MDAnalysis/lib/_cutil.pyx#L52) [and here](https://github.com/MDAnalysis/mdanalysis/blob/develop/package/MDAnalysis/lib/_cutil.pyx#L74).  @darmis007 if you change the 2 lines I've linked above to use `intptr_t` does this also fix your problem?

tylerjereddy:
This patch seems to work for me on win 32-bit locally; `intptr_t` was not a valid identifier, at least not without some more header includes. There are other issues in the 32-bit full test suite though & we need it in CI first, which should probably be Azure instead of Appveyor for turnaround time/parallel jobs.

```diff
--- a/package/MDAnalysis/lib/_cutil.pyx
+++ b/package/MDAnalysis/lib/_cutil.pyx
@@ -49,7 +49,7 @@ ctypedef cmap[int, intset] intmap

 @cython.boundscheck(False) # turn off bounds-checking for entire function
 @cython.wraparound(False)  # turn off negative index wrapping for entire function
-def unique_int_1d(np.int64_t[:] values):
+def unique_int_1d(long[:] values):
     """Find the unique elements of a 1D array of integers.

     This function is optimal on sorted arrays.
@@ -71,7 +71,7 @@ def unique_int_1d(np.int64_t[:] values):
     cdef int i = 0
     cdef int j = 0
     cdef int n_values = values.shape[0]
-    cdef np.int64_t[:] result = np.empty(n_values, dtype=np.int64)
+    cdef long[:] result = np.empty(n_values, dtype=np.int)
```

tylerjereddy:
This is probably closer to what Richard wants, using the actual NumPy/Cython types. This also passes locally on 32-bit Win for the problem case

```diff
--- a/package/MDAnalysis/lib/_cutil.pyx
+++ b/package/MDAnalysis/lib/_cutil.pyx
@@ -49,7 +49,7 @@ ctypedef cmap[int, intset] intmap

 @cython.boundscheck(False) # turn off bounds-checking for entire function
 @cython.wraparound(False)  # turn off negative index wrapping for entire function
-def unique_int_1d(np.int64_t[:] values):
+def unique_int_1d(np.intp_t[:] values):
     """Find the unique elements of a 1D array of integers.

     This function is optimal on sorted arrays.
@@ -71,7 +71,7 @@ def unique_int_1d(np.int64_t[:] values):
     cdef int i = 0
     cdef int j = 0
     cdef int n_values = values.shape[0]
-    cdef np.int64_t[:] result = np.empty(n_values, dtype=np.int64)
+    cdef np.intp_t[:] result = np.empty(n_values, dtype=np.intp)
```

richardjgowers:
Yeah sorry, I thought it was called something  different. This should fix
the problem at the root. I think there’s another function further down (for
fragments?) that also assumes int64 which likely has the same bug.

tylerjereddy:
Superseded by #2696, so closing--we've added broader support for 32-bit Windows now. Sorry for the confusion here.

PR #2695, opened 2020-05-24
WIP, CI: add 32-bit Win to CI

tylerjereddy:
* Add 32-bit Windows testing via Azure CI (not Appveyor because of the lack of parallel jobs there)

* It is up to the team if we want to just place fixes in this PR before merging, switch to "allowed failure" somehow, or just merge in when ready and have "red CI" for a bit until the issues are fixed/skipped/xfailed, etc.

* this is a pared-down version of the  32-bit Windows testing I set up for NumPy/SciPy--probably could pare it down even more, but seems to catch the test failures we'd expect

Related to:
https://github.com/MDAnalysis/mdanalysis/pull/2693
https://github.com/MDAnalysis/mdanalysis/issues/2687
https://github.com/MDAnalysis/mdanalysis/pull/2689
https://github.com/MDAnalysis/mdanalysis/pull/2346

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,49 @@
+trigger:
+  # start a new build for every push
+  batch: False
+  branches:
+    include:
+      - develop
+  paths:
+    include:
+      - '*'
+
+pr:
+  branches:
+    include:
+    - '*'  # must quote since "*" is a YAML reserved character; we want a string
+
+
+jobs:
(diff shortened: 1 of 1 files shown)


tylerjereddy:
Ah, maybe I can't open the first Azure addition from my fork for some chicken-and-egg reason. I see  #2346 didn't originate from a fork so let me try that instead..

PR #2696, opened 2020-05-24
CI: add 32-bit Win to CI

tylerjereddy:
Supersedes #2695

Add 32-bit Windows testing via Azure CI (not Appveyor because of the lack of parallel jobs there)

It is up to the team if we want to just place fixes in this PR before merging, switch to "allowed failure" somehow, or just merge in when ready and have "red CI" for a bit until the issues are fixed/skipped/xfailed, etc.

this is a pared-down version of the 32-bit Windows testing I set up for NumPy/SciPy--probably could pare it down even more, but seems to catch the test failures we'd expect

Related to:
#2693
#2687
#2689
#2346

### DIFF
--- a/package/MDAnalysis/core/selection.py
+++ b/package/MDAnalysis/core/selection.py
@@ -226,7 +226,7 @@ class ByResSelection(UnarySelection):
 
     def apply(self, group):
         res = self.sel.apply(group)
-        unique_res = unique_int_1d(res.resindices.astype(np.int64))
+        unique_res = unique_int_1d(res.resindices.astype(np.intp))
         mask = np.in1d(group.resindices, unique_res)
 
         return group[mask].unique
--- a/package/MDAnalysis/core/selection.py
+++ b/package/MDAnalysis/core/selection.py
@@ -226,7 +226,7 @@ class ByResSelection(UnarySelection):
 
     def apply(self, group):
         res = self.sel.apply(group)
-        unique_res = unique_int_1d(res.resindices.astype(np.intp))
+        unique_res = unique_int_1d(res.resindices)
         mask = np.in1d(group.resindices, unique_res)
 
         return group[mask].unique
--- a/testsuite/MDAnalysisTests/analysis/test_density.py
+++ b/testsuite/MDAnalysisTests/analysis/test_density.py
@@ -398,6 +398,8 @@ class TestNotWithin(object):
     def u():
         return mda.Universe(GRO)
 
+    @pytest.mark.skipif(sys.maxsize <= 2**32,
+                        reason="non-kdtree density too large for 32-bit")
     def test_within(self, u):
         vers1 = density.notwithin_coordinates_factory(u, 'resname SOL',
                                                       'protein', 2,
@@ -410,6 +412,8 @@ class TestNotWithin(object):
 
         assert_equal(vers1, vers2)
 
+    @pytest.mark.skipif(sys.maxsize <= 2**32,
+                        reason="non-kdtree density too large for 32-bit")
     def test_not_within(self, u):
         vers1 = density.notwithin_coordinates_factory(u, 'resname SOL',
                                                      'protein', 2,
--- a/testsuite/MDAnalysisTests/analysis/test_hole2.py
+++ b/testsuite/MDAnalysisTests/analysis/test_hole2.py
@@ -73,6 +73,8 @@ class TestCheckAndFixLongFilename(object):
             fixed = check_and_fix_long_filename(abspath)
             assert fixed == self.filename
     
+    @pytest.mark.skipif(os.name == 'nt' and sys.maxsize <= 2**32,
+                        reason="FileNotFoundError on Win 32-bit")
     def test_symlink_dir(self, tmpdir):
         dirname = 'really_'*20 +'long_name'
         short_name = self.filename[-20:]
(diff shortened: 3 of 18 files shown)


tylerjereddy:
Ok, it is running now. I tried to delete the spurious/duplicate `CI (1)` entry for Azure, so that should go away in the future.

tylerjereddy:
@lilyminium @richardjgowers I pushed in a commit that allows 32-bit Windows and 64-bit Linux test suites to pass locally.

We'll see if the CI agrees, but you'll probably want to review carefully anyway. In short, mostly `np.int64` -> `np.intp` changes, and I think I avoided any coercion to `int64`.

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -398,6 +398,8 @@ class TestNotWithin(object):
     def u():
         return mda.Universe(GRO)
 
+    @pytest.mark.skipif(sys.maxsize <= 2**32,
+                        reason="non-kdtree density too large for 32-bit")
```
tylerjereddy, line 402: memory allocation issue specifically, I think

testsuite/MDAnalysisTests/topology/test_tprparser.py
```diff
@@ -55,7 +55,7 @@ def test_moltypes(self, top):
     def test_molnums(self, top):
         molnums = top.molnums.values
         assert_equal(molnums, self.ref_molnums)
-        assert molnums.dtype == np.int64
+        assert molnums.dtype == np.intp
```
tylerjereddy, line 58: While nobody likes changing unit tests to make them pass, I think the proposed approach (from Richard and maybe Lily?) for portability is that integer indexing will now be platform-sized rather than coerced  to `int64`, so some tests will have platform-specific typing.

tylerjereddy:
There's only 1 32-bit test failure left in Azure CI--it is an unreliable failure according to `hypothesis`. I'll let the team decide what to do with that one.

richardjgowers:
Ok this is cool.  So the one failure is something string related that hypothesis found.  Can we tag this from pytest is a knownfail with win32 platforms then we can include this in our test matrix?

tylerjereddy:
> Can we tag this from pytest is a knownfail with win32 platforms then we can include this in our test matrix?

@richardjgowers Ok, I added a third commit to `xfail` that specific test on 32-bit windows. I tried to reproduce locally, but my own 32-bit Windows setup is happy to pass that test. I suspect we can just investigate this later--maybe I have a few extra add-ons installed that bare-bones Azure doesn't.

testsuite/MDAnalysisTests/analysis/test_hole2.py
```diff
@@ -73,6 +73,8 @@ def test_relative(self):
             fixed = check_and_fix_long_filename(abspath)
             assert fixed == self.filename
     
+    @pytest.mark.skipif(os.name == 'nt' and sys.maxsize <= 2**32,
+                        reason="FileNotFoundError on Win 32-bit")
     def test_symlink_dir(self, tmpdir):
```
lilyminium, line 78: Was this also because of not having symlinking privileges, or is it some weirdness with Windows paths and `os.path.join`?

(Just out of interest; I'm not sure hole2 even supports Windows.)

package/MDAnalysis/core/selection.py
```diff
@@ -226,7 +226,7 @@ class ByResSelection(UnarySelection):
 
     def apply(self, group):
         res = self.sel.apply(group)
-        unique_res = unique_int_1d(res.resindices.astype(np.int64))
+        unique_res = unique_int_1d(res.resindices.astype(np.intp))
```
lilyminium, line 229: Is this `astype` necessary? I think it required coercion earlier as `resindices` are `intp` typically.

lilyminium:
Thanks @tylerjereddy! This looks great. I don't know much about win32 but only 5 xfails seems really good; I suppose the non-kdtree densities could have warnings, same with the symlinking (I guess symlinking privileges would be a Windows-wide issue?).

package/MDAnalysis/core/selection.py
tylerjereddy: I've pushed in a commit to remove it--32-bit Windows suite passed locally without it. I didn't have time to run the suite for 64-bit linux as well, but fingers crossed.

orbeckst:
Great work, @tylerjereddy !

orbeckst:
Are we now in a position to officially support 32-bit Windows? 

Should there be something in CHANGELOG?

tylerjereddy:
> Should there be something in CHANGELOG?

We could add something--I don't know if I'd go as far as "official support," but pretty close.

tylerjereddy:
FWIW, I'd suggest not using a single plain text changelog for entries since it can be a source of merge conflicts. A simple wiki page in between releases can work well enough. Some projects use fancy tools to add separate notes that get merged later or whatever.

Anyway, I'll try to push something in shortly.

tylerjereddy:
Ok, I've pushed in a changelog entry--I put the recent change at top, but not my name at top--mostly to avoid reformatting the author list, but also technically I've made other changes for the release already without recording the name so that's my excuse.

tylerjereddy:
Alright, CI is green, 2 core dev approvals and I added a short release note, so merging

orbeckst:
> FWIW, I'd suggest not using a single plain text changelog for entries since it can be a source of merge conflicts. 

It is true that most merge conflicts come from our custom to add new changes at the top of each category. Maybe it would be better to add at the bottom.

> A simple wiki page in between releases can work well enough. 

However, I don't like using a Wiki because then we need to check something separately when reviewing PRs, much better to have everything in one place, especially with new contributors.

> Some projects use fancy tools to add separate notes that get merged later or whatever.

Yes, I've seen these. Is it worth the extra work?

Issue #2651, opened 2020-03-29
Removing guess by default in TOPParser


IAlibay:
## Is your feature request related to a problem? ##

So a mix between a bug and feature request.
In light of #2630, especially https://github.com/MDAnalysis/mdanalysis/issues/2630#issuecomment-599784226 I went back and looked at the element parser for the TOPParser. I noticed that not only is it by default guessing elements (using `guess_types` if the `ATOMIC_NUMBER` record isn't provided), but it also reports bad elements for two of our test files (`NA` is reported instead of `N` for PRM and PRM7).

## Describe the solution you'd like ##

I would like to remove the guess by default functionality, and make it so that elements are only assigned if a full `ATOMIC_NUMBER` record is present (possibly assigning "empty" records when some of the elements are partially missing [i.e. if the ATOMIC_NUMBER contains negative values].

Guessing can then be done a later user-defined point if necessary.

## Additional context ##
I am complicit in this "bug" so I'm willing to fix this, but we probably need a consensus on what an "empty" element record would look like (this might be agreed upon and I've missed the discussion, pinging @lilyminium who would probably know better here).

## Update

So the way I see it there are three things that probably need to be done here:
  - [ ] Remove guess by default for elements in the TOPParser (this should also partly fix #2449 [at least users will be told that it wasn't recognised as P]
  - [ ] Add use of the atomnum attribute to the TOPParser

IAlibay:
To add context here, I suddenly realised that AMBER does offer a united atom force field (ff03ua) [as far as I am aware no CG though?]. It looks like the default here is to set the ATOMIC_NUMBER to the element of the heavy atom (i.e. C in CH3). See the following parm7 snippet fo alanine dipeptide:
```
%FLAG ATOM_NAME
%FORMAT(20a4)
CH3 C   O   N   H   CA  HA  CB  C   O   N   H   CH3
%FLAG CHARGE
%FORMAT(5E16.8)
  1.04557735E+00  8.25646946E+00 -9.85343539E+00 -7.37589504E+00  5.36238555E+00
 -1.57422450E-01  1.91674907E+00 -2.08827558E-02  1.03907928E+01 -1.01157272E+01
 -8.77327211E+00  5.21844761E+00  4.10619486E+00
%FLAG ATOMIC_NUMBER
%FORMAT(10I8)
       6       6       8       7       1       6       1       6       6       8
       7       1       6
%FLAG MASS
```
[...]

PR #2700, opened 2020-05-27
Remove `NUMBER_TO_ELEMENTS`

### LINKED ISSUE #2699: Remove _elements.py (NUMBER_TO_ELEMENT)
## Context
We currently have two dictionaries that do the exact same thing, `NUMBER_TO_ELEMENTS` in MDAnalysis.topology._elements and `Z2SYMB` in MDAnalysis.topology.tables. These are quasi-identical except the fact that Z2SYMB is a little bit more IUPAC compliant for some of the newer elements (which makes sense since `Z2SYMB` was added after `NUMBER_TO_ELEMENTS`).

At the moment, `NUMBER_TO_ELEMENTS` is used solely for the TOPParser, and could be swapped out for `Z2SYMB` at nearly no cost.

Remove `NUMBER_TO_ELEMENTS`, this should help us lose a decent chunk of code for free.

Keeping it if we really want it.

IAlibay:
Fixes #2699

 - Removes MDAnalysis.topology._elements.py
 - Switches `NUMBER_TO_ELEMENTS` to `Z2SYMB` in TOPParser
 - Adds a more thorough elements test and starts cleaning up tests in anticipation of #2651

## Additional context:

Unfortunately it seems I tackled this the wrong way around. The elements testing for TOPParser is a little bit all over the place at the moment, but I plan to more thoroughly change things with #2651, if anything, the `test_elements` function for PRM12 is going to get moved to TOPBase.

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?


package/CHANGELOG
```diff
@@ -133,6 +133,7 @@ Enhancements
     the capability to allow intermittent behaviour (PR #2256)
 
 Changes
+  * Removes duplicate `NUMBER_TO_ELEMENTS` table (Issue #2699)
```
orbeckst, line 136: add, "use Z2SYMB instead" or something like that for anyone who might have used it...

IAlibay: Fixed in 4019011.

Issue #2468, opened 2020-01-24
rdkit interoperability


richardjgowers:
This is mostly an idea for a GSOC project.  There's a bunch of cool stuff in rdkit which isn't even close to being in MDA (and vice versa) so rather than reinvent wheels, it would be cool to do:
```python
import MDAnalysis as mda
from MDAnalysis import convert_to

u = mda.Universe('bleh')

rdkit_moi = convert_to('rdkit', u)
```
and
```python
from rdkit import Chem
from MDAnalysis import convert_to

rdkit_mol = Chem.MolFromSmiles('CCOC')

u = convert_to('mda', rdkit_mol)
```

Which would expand upon the converters idea that @lilyminium has got rolling with parmed.

The data structures are going to be very different, and rdkit is quite picky about what it lets you load, but it would be cool to get something going.

RMeli:
This would be very useful indeed! The following blog post might be relevant for further discussion: [Why the RDKit isn't available on PyPi](https://rdkit.blogspot.com/2019/11/why-rdkit-isnt-available-on-pypi.html).

richardjgowers:
Yeah rdkit via conda is the only way to stay sane tbh.  So this project might end up in a side package/non default submodule which is optionally installed and does some monkey patching to `_CONVERTERS`.

IAlibay:
At least on our end, rdkit integration would be very much appreciated. It would definitely make tools that depend on both packages like [lintools](https://github.com/bigginlab/lintools) (which we still have aims to revive properly) a lot easier.

orbeckst:
I like the idea of expanding on converters – working towards API interoperability instead of files. It's the future.

We might have to [review our policy on package dependencies in the core](https://github.com/MDAnalysis/mdanalysis/wiki/Style-Guide#list-of-core-module-dependencies). Maybe it's ok for the majority of users to get a well-define failure if they try to do something with a specialized reader/converter, especially if they are told what to do if they want to install the package. I am thinking along mocking missing optional packages in such a way that only users who want to use exactly the functionality get a failure. 

Perhaps the policy can be changed as to "packages that are used in a single convertor or reader can be optional". Or we make converters a submodule with its own policy (but it would be convenient to also be able to include readers/parsers with exotic dependencies).

orbeckst:
Do you know anyone from the RDKit community who might want to co-mentor for GSoC? That would be extremely valuable.

RMeli:
[...]

Issue #2667, opened 2020-04-22
Documentation doesn't build with sphinx >= 2.0


daveminh:
## Expected behavior ##

Documentation would build and I could see it in doc/html/index.html, as described here: https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard
## Actual behavior ##

Attempting to autodetect OpenMP support... Did not detect OpenMP support.
No openmp compatible compiler found default to serial build.
Will attempt to use Cython.
/miniconda2/envs/mdanalysis-dev/lib/python3.6/site-packages/setuptools/dist.py:454: UserWarning: Normalizing '0.20.2-dev0' to '0.20.2.dev0'
  warnings.warn(tmpl.format(**locals()))
running build_sphinx
Running Sphinx v3.0.2
loading intersphinx inventory from https://docs.python.org/objects.inv...
loading intersphinx inventory from https://docs.scipy.org/doc/numpy/objects.inv...
loading intersphinx inventory from https://docs.scipy.org/doc/scipy/reference/objects.inv...
loading intersphinx inventory from https://matplotlib.org/objects.inv...
loading intersphinx inventory from https://networkx.github.io/documentation/stable/objects.inv...
loading intersphinx inventory from https://www.mdanalysis.org/GridDataFormats/objects.inv...
loading intersphinx inventory from https://gsd.readthedocs.io/en/stable/objects.inv...
loading intersphinx inventory from https://parmed.github.io/ParmEd/html/objects.inv...
intersphinx inventory has moved: https://docs.python.org/objects.inv -> https://docs.python.org/3/objects.inv
building [mo]: all of 0 po files
building [html]: all source files
updating environment: [new config] 146 added, 0 changed, 0 removed
reading sources... [  9%] documentation_pages/analysis/encore/dimensionality_red/Users/dminh/Documents/GitHub/mdanalysis/package/MDAnalysis/analysis/hbonds/hbond_analysis.py:345: DeprecationWarning: This module is deprecated as of MDAnalysis version 1.0.It will be removed in MDAnalysis version 2.0Please use MDAnalysis.analysis.hydrogenbonds.hbond_analysis instead.
  category=DeprecationWarning
reading sources... [100%] index
sphinx-sitemap warning: No pages generated for sitemap.xml

Warning, treated as error:
[...]

PR #2707, opened 2020-06-01
RDKitReader and RDKitParser

cbouy:
Part of the fixes for #2468 

 - added the RDKitParser which creates a `core.topology.Topology` object from an `rdkit.Chem.rdchem.Mol` object.
- added the RDKitReader (based on the MemoryReader) to read coordinates from RDKit conformers
- added the `Aromaticities` topology attributes, and the `aromatic` selection token (for now, only usable when the Universe was created from an RDKit molecule)
- added the `from_smiles` classmethod to the Universe (can add hydrogens and generate multiple conformers)

This ~~is~~ was a minimal version complemented iteratively by adding new attributes, tests and the doc. Code heavily inspired from the ParmEdParser.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


tylerjereddy:
Unrelated, but I wonder why Azure pipelines is not showing up in CI.

testsuite/MDAnalysisTests/topology/test_rdkit.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def test_bond_orders(self, top, filename):
+        # not exactly what's written in the mol2 file but the 2 double bonds in 
+        # the thiophene moiety should be written as aromatic, and they should 
+        # be perceived as such by RDKit
+        expected = [1.5, 1, 1, 2, 1, 2, 1, 1, 1.5, 1, 1, 1.5, 2, 2, 1, 1, 1, 1,
+        1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1.5, 1.5, 1.5, 1, 1.5, 1, 1.5, 1, 1.5, 1,
+        1, 1, 1.5, 1, 1, 1, 1, 1.5, 1, 1, 2, 1, 1]
+        expected = [float(i) for i in expected]
+        assert sorted(top.bonds.order) == sorted(expected)
```
richardjgowers, line 74: Sorting here just checks that the correct count of different order bonds are parsed, rather than the order being correctly assigned to different bonds.

testsuite/MDAnalysisTests/topology/test_rdkit.py
```diff
@@ -0,0 +1,90 @@
...
+    guessed_attrs = ['types']
+    
+    expected_n_atoms = 137
+    expected_n_residues = 13
+    expected_n_segments = 1
+    expected_n_bonds = 137
+
+    @pytest.fixture
+    def filename(self):
+        return Chem.MolFromPDBFile(self.ref_filename, removeHs=False)
```
richardjgowers, line 90: I think a `MolFromSmiles` test would be good too, smiles parsing is something very important that this will give us.

testsuite/MDAnalysisTests/topology/test_rdkit.py
cbouy: Yes I thought rdkit was changing the order in which bonds are iterated during sanitization, but it was actually another problem. Should be fixed now

testsuite/MDAnalysisTests/topology/test_rdkit.py
cbouy: Done!

cbouy:
Can I add an SDF file in the test data to check that it's read correctly as well ?

IAlibay:
> Can I add an SDF file in the test data to check that it's read correctly as well ?

Please do! The more tests the better :)

richardjgowers:
Yup a small one but with corner cases preferably

testsuite/MDAnalysisTests/topology/test_rdkit.py
```diff
@@ -0,0 +1,108 @@
...
+#
+from __future__ import absolute_import
+
+import pytest
+
+import MDAnalysis as mda
+from MDAnalysisTests.topology.base import ParserBase
+from MDAnalysisTests.datafiles import mol2_molecule, PDB_helix
+
+Chem = pytest.importorskip('rdkit.Chem')
```
IAlibay, line 31: Probably missing something here with my end-of-the-day-brain, but you'll probably want to tell appveyor and travis to install rdkit (see the similar parmed entries in `.travis.yml` and `.appveyor.yml`

cbouy:
Looks like there are a few more formats that RDKit supports but not MDA: 
- SDF
- MOL
- MAE (from Maestro)
- TPL (from Catalyst)
- HELM (from Pistoia Alliance)
- Fasta
- sequence of amino-acids

But I guess just SDF is enough. I'll try to find a file with charges, stereochemistry, custom data and whatnot...

testsuite/MDAnalysisTests/topology/test_rdkit.py
cbouy: Completely forgot about that, my bad

IAlibay:
> Looks like there are a few more formats that RDKit supports but not MDA:
> 
> * SDF
> * MOL
> * MAE (from Maestro)
> * TPL (from Catalyst)
> * HELM (from Pistoia Alliance)
> * Fasta
> * sequence of amino-acids
> 
> But I guess just SDF is enough. I'll try to find a file with charges, stereochemistry, custom data and whatnot...

Note here, to save on space, we could probably store things in gzip format and feed rdkit with file-like objects.

cbouy:
Is the `Charges` attribute supposed to be used for formal or partial charges ?

IAlibay:
> Is the `Charges` attribute supposed to be used for formal or partial charges ?

I believe it's the partial charges of each atom in electron charge units.

cbouy:
and there's no attribute for formal charges ?

IAlibay:
I don't think so, quickly scanning through TopologyAttrs doesn't bring anything. @fiona-naughton or @richardjgowers might know better though.

richardjgowers:
I think technically pdb is meant to be formal, but we just call it charge.
Is there a use case where people would have both partial and formal?

cbouy:
I wasn't speaking specifically about PDB, RDKit will assign formal charges to any atom no matter the file format.
But for MOL2 for instance, RDKit will give the usual `atom.GetFormalCharge()` and also return partial charges read from the file with `atom.GetProp('_TriposPartialCharge')`.
RDKit can also compute Gasteiger charges with `AllChem.ComputeGasteigerCharges(mol)` and they become accessible with `atom.GetProp('_GasteigerCharge')`

Anyway I'll stick to partial charges for now, thanks!

package/MDAnalysis/topology/RDKitParser.py
```diff
@@ -0,0 +1,249 @@
...
+                # atom type (MOL2 only)
+                try:
+                    atomtypes.append(atom.GetProp('_TriposAtomType'))
+                except KeyError:
+                    pass
+                # gasteiger charge (computed):
+                # if the user took the time to compute them, make it a priority
+                # over charges read from a MOL2 file
+                try:
+                    charges.append(atom.GetProp('_GasteigerCharge'))
```
IAlibay, line 145: We'd probably want to warn users here, if they happen to have both types of charges available, that we are defaulting to Gasteiger over the user-read ones.

package/MDAnalysis/topology/RDKitParser.py
```diff
@@ -0,0 +1,249 @@
...
+                except KeyError:
+                    # partial charge (MOL2 only)
+                    try:
+                        charges.append(atom.GetProp('_TriposPartialCharge'))
```
IAlibay, line 149: An additional thought here, does RDKIT guarantee the charge units? I couldn't see anything specific in the docs.

cbouy: For Gasteiger, they are in `e`, but for Tripos there's no way of knowing where they come from so no guarantee

IAlibay: Looks like our own parse assumes `e` so that should be fine.

cbouy:
I guess I'm done for the topology... Should I create new topology attributes for aromaticity and formal charges ? What else is missing ?

package/MDAnalysis/topology/RDKitParser.py
```diff
@@ -0,0 +1,289 @@
...
+        atomtypes = []
+        segids = []
+        altlocs = []
+        chainids = []
+        occupancies = []
+        tempfactors = []
+
+        # check if multiple charges present
+        atom = mol.GetAtomWithIdx(0)
+        if atom.HasProp('_GasteigerCharge') and (
```
IAlibay, line 149: Might be easier to read if we can move this closer to where we feed `Charges`?

package/MDAnalysis/topology/RDKitParser.py
```diff
@@ -0,0 +1,289 @@
...
+        atom.HasProp('_TriposPartialCharge')
+        ):
+            warnings.warn(
```
IAlibay, line 152: We'll need a test to pick this warning up.

package/MDAnalysis/topology/RDKitParser.py
```diff
@@ -0,0 +1,289 @@
...
+                # atom type (MOL2 only)
+                try:
+                    atomtypes.append(atom.GetProp('_TriposAtomType'))
+                except KeyError:
+                    pass
+                # gasteiger charge (computed):
+                # if the user took the time to compute them, make it a priority
+                # over charges read from a MOL2 file
+                try:
+                    charges.append(atom.GetDoubleProp('_GasteigerCharge'))
```
IAlibay, line 185: We'll want tests here to check that we are picking up the gasteiger and tripos charges as expected.

package/MDAnalysis/topology/RDKitParser.py
```diff
@@ -0,0 +1,289 @@
...
+                resnames.append(mi.GetResidueName())
+                segids.append(mi.GetSegmentNumber())
+                altlocs.append(mi.GetAltLoc().strip())
+                chainids.append(mi.GetChainId().strip())
+                occupancies.append(mi.GetOccupancy())
+                tempfactors.append(mi.GetTempFactor())
+            else:
+                # atom name (MOL2 only)
+                try:
+                    names.append(atom.GetProp('_TriposAtomName'))
```
IAlibay, line 173: Probably will want to add tests for these two to `TestRDKitParserMOL2`

package/MDAnalysis/topology/RDKitParser.py
cbouy: I put it outside of the loop because I didn't want to check if both charges are present for every atom. Should I put it after the loop on atoms then ?

IAlibay: No I think it's fine, my bad here, I somehow hadn't taken the looping into account.

cbouy:
> How about adding a Universe.from_smiles classmethod? Should be trivial with what you have here already.

Sounds good to me!

Also, I just realized for my bond order tests that I'm comparing what's in the Topology to what's in the file. Shouldn't I be comparing to what's in the RDKit molecule instead ? i.e. if RDKit suddenly reads an aromatic bond as a triple bond, should that make our tests fail ?

cbouy:
For your PDBParser I noticed that if SegIds are not present you assign them the values of ChainIds. Should I also do that here ?

With the PDB_chainidrepeat testfile I'm currently getting 5 residues and 1 segment with RDKitParser instead of 6 residues 6 segments with PDBParser.

I guess both parsers should be consistent right ?

.appveyor.yml
```diff
@@ -9,8 +9,8 @@ cache:
 
 environment:
     global:
-        CONDA_CHANNELS: conda-forge
-        CONDA_DEPENDENCIES: pip setuptools wheel cython mock six biopython networkx joblib matplotlib scipy vs2015_runtime pytest mmtf-python GridDataFormats hypothesis pytest-cov codecov chemfiles tqdm
+        CONDA_CHANNELS: conda-forge rdkit
```
richardjgowers, line 12: can use the conda-forge rdkit packaging right?

package/MDAnalysis/topology/RDKitParser.py
```diff
@@ -0,0 +1,282 @@
+# -*- Mode: python; tab-width: 4; indent-tabs-mode:nil; coding:utf-8 -*-
```
richardjgowers, line 1: new modules will need a .rst stub

package/MDAnalysis/core/universe.py
```diff
@@ -1289,6 +1289,41 @@ def _fragdict(self):
                 fragdict[a.ix] = fraginfo(i, f)
 
         return fragdict
+    
+    @classmethod
+    def from_smiles(cls, smiles, sanitize=True, addHs=True, **kwargs):
+        """Create a Universe from a SMILES string with rdkit
+
```
richardjgowers, line 1296: This is very cool.  Add an example into the docstring.  Also what happens if I pass RDKit a bad SMILES string?

package/MDAnalysis/core/universe.py
```diff
@@ -1289,6 +1289,41 @@ def _fragdict(self):
...
+        ----------
+        smiles : str
+            SMILES string
+
+        sanitize : bool (optional, default True)
+            Toggle the sanitization of the molecule
+
+        addHs : bool (optional, default True)
+            Add all necessary hydrogens to the molecule
+
```
richardjgowers, line 1307: How hard would it be to also generate coordinates?

testsuite/MDAnalysisTests/core/test_universe.py
```diff
@@ -216,6 +216,12 @@ def test_universe_topology_class_with_coords(self):
         assert_equal(u.trajectory.n_frames, u2.trajectory.n_frames)
         assert u2._topology is u._topology
 
+    def test_universe_from_smiles(self):
+        smi = "CN1C=NC2=C1C(=O)N(C(=O)N2C)C"
```
IAlibay, line 220: The minimal dependency tests are failing here, you'll probably want an importorskip.

.appveyor.yml
cbouy: I completely forgot it's now available through conda-forge!

package/MDAnalysis/core/universe.py
cbouy: pretty easy

package/MDAnalysis/core/universe.py
cbouy: RDKit is gonna complain in a not so graceful manner about mol being `None` if we reach `Chem.AddHs(mol)`
I'll add a check on that.

Also, should I use rdkit's camelcase names for the `from_smiles` parameters ? i.e. `AddHs` or `add_hydrogens` ?

richardjgowers: Might be best to follow RDKit's lead in argument names

package/MDAnalysis/core/universe.py
richardjgowers: `generate_coordinates=True` would be a fun kwarg then

richardjgowers:
@cbouy the `Universe.from_smiles()` is very cool, it should probably have a section in the [user guide](https://www.mdanalysis.org/UserGuide/universe.html#creating-a-universe) somewhere near "Constructing from scratch". Can you open a PR on [that repo](https://github.com/MDAnalysis/UserGuide) adding some documentation?

package/MDAnalysis/topology/RDKitParser.py
cbouy: I'm not too familiar with sphinx yet, are these 2 files the only thing I need to create/modify ?
package/doc/sphinx/source/documentation_pages/topology/RDKitParser.rst
package/doc/sphinx/source/documentation_pages/topology_modules.rst

cbouy:
3 questions left:
> Also, I just realized for my bond order tests that I'm comparing what's in the Topology to what's in the file. Shouldn't I be comparing to what's in the RDKit molecule instead ? i.e. if RDKit suddenly reads an aromatic bond as a double bond, should that make our tests fail ?

> For your PDBParser I noticed that if SegIds are not present you assign them the values of ChainIds. Should I also do that here ? It directly influences how many residues are found in some cases: with the PDB_chainidrepeat testfile I'm currently getting 5 residues and 1 segment with RDKitParser instead of 6 residues 6 segments with PDBParser.
I guess both parsers should be consistent right ?

> New topology attribute for aromaticity ?

richardjgowers:
@cbouy 
1 - Yes we're going for matching RDKit rather than the file, I guess we're assuming that this part of RDKit is mature enough that it won't change.

2 - The PDB issue will need looking into, there are probably big differences with how this format is parsed in the two packages.

3 - Yes, a boolean dtype attribute for aromaticity would be good.  Then including this into the `select_atoms` language would be cool.  Finally, and tangentially related, adding a `topology.guessers.guess_aromaticity` function to generate this (which would dump the molecule into RDKit, try and perceive then return this) would be handy.

cbouy:
I started adding the Reader, several questions:

- How do I handle rdkit mols without conformers ? Right now I'm creating an array of np.NaN with the appropriate dimension (1 frame, right num_atoms, 3 coordinates) but I guess there's something more relevant to do.
- Should the `filename` of the resulting Universe be `None` or the original RDKit molecule ? This goes for both the Parser and the Reader.

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -0,0 +1,92 @@
...
+"""
+from __future__ import absolute_import
+
+import numpy as np
+
+from . import memory
+
+
+class RDKitReader(memory.MemoryReader):
+    """Coordinate reader for RDKit."""
```
IAlibay, line 64: You'll probably want some docstring here & on __init__ that documents inputs, etc.. right now if you call `??RDKITReader`, I think you'll get the docstring for _Readermeta.

package/MDAnalysis/core/selection.py
```diff
@@ -573,6 +573,20 @@ class AltlocSelection(StringSelection):
     field = 'altLocs'
 
 
+class AromaticSelection(Selection):
+    """Contains all aromatic atoms"""
```
IAlibay, line 577: I realise it's self-evident, but you might want to expand on the docstring here a little bit, at the very least to say which attribute you are selecting on. That way it's more obvious to users that are reading the documentation.

package/MDAnalysis/core/topologyattrs.py
```diff
@@ -1665,6 +1665,18 @@ def _gen_initial_values(na, nr, ns):
...
+class Aromaticities(AtomAttr):
+    """Aromaticity (RDKit)"""
+    attrname = "aromaticities"
+    singular = "aromaticity"
+    per_object = "atom"
+    dtype = bool
+
+    @staticmethod
+    def _gen_initial_values(na, nr, ns):
+        return np.array([False for _ in range(na)], dtype=bool)
```
IAlibay, line 1677: I _think_ here you can just do `np.zeros(na, dtype=bool)`.

package/MDAnalysis/core/universe.py
```diff
@@ -1289,6 +1289,69 @@ def _fragdict(self):
...
+            raise SyntaxError('Error while parsing SMILES {0}'.format(smiles))
+        if addHs:
+            mol = Chem.AddHs(mol)
+        if generate_coordinates:
+            if not addHs:
+                raise ValueError("Generating coordinates requires adding "
+                "hydrogens with `addHs=True`")
+            assert (type(numConfs) is int) and (numConfs > 0), ("numConfs must"
+            " be a non-zero positive integer instead of {0}".format(numConfs))
+            AllChem.EmbedMultipleConfs(mol, numConfs)
```
IAlibay, line 1352: EmbedMultipleConfs has a few useful options that users might want to set manually, is there any way we could capture these from **kargs?

testsuite/MDAnalysisTests/coordinates/test_rdkit.py
```diff
@@ -0,0 +1,54 @@
...
+
+    @pytest.fixture
+    def universe(self, mol):
+        return mda.Universe(mol)
+
+    def test_coordinates(self, mol, universe):
+        expected = np.array([
+            conf.GetPositions() for conf in mol.GetConformers()], 
+            dtype=np.float32)
+        assert universe.trajectory.n_frames == 1
```
IAlibay, line 49: you'll want to add more test cases here, specifically looking for multiple frames & making sure you get the right coordinates on "trajectory" iteration (probably worth looking at base.MultiframeReaderTest)

testsuite/MDAnalysisTests/coordinates/test_rdkit.py
```diff
@@ -0,0 +1,54 @@
...
+    def mol(self):
+        return Chem.MolFromMol2File(mol2_molecule, removeHs=False)
+
+    @pytest.fixture
+    def universe(self, mol):
+        return mda.Universe(mol)
+
+    def test_coordinates(self, mol, universe):
+        expected = np.array([
+            conf.GetPositions() for conf in mol.GetConformers()],
```
IAlibay, line 47: Would it also be worth testing that we get the same coordinates as our own internal MOL2 reader?

testsuite/MDAnalysisTests/core/test_atomselections.py
```diff
@@ -520,6 +520,20 @@ def test_molnum(self, universe, selection_string, reference):
         assert_equal(sel.ids, np.array(reference, dtype=np.int32))
 
 
+class TestSelectionRDKit(object):
+    def setup_class(self):
+        pytest.importorskip("rdkit.Chem")
+        
+    def test_aromatic(self):
```
IAlibay, line 527: rather than two asserts here, would it be better to just use parametrize? that way we can add more combinations if necessary?

IAlibay: pedantic, but since it's core stuff, it might be worth adding a few more cases with `and`, `around`, etc..

testsuite/MDAnalysisTests/topology/test_rdkit.py
```diff
@@ -0,0 +1,188 @@
...
+# MDAnalysis: A Python package for the rapid analysis of molecular dynamics
+# simulations. In S. Benthall and S. Rostrup editors, Proceedings of the 15th
+# Python in Science Conference, pages 102-109, Austin, TX, 2016. SciPy.
+# doi: 10.25080/majora-629e541a-00e
+#
+# N. Michaud-Agrawal, E. J. Denning, T. B. Woolf, and O. Beckstein.
+# MDAnalysis: A Toolkit for the Analysis of Molecular Dynamics Simulations.
+# J. Comput. Chem. 32 (2011), 2319--2327, doi:10.1002/jcc.21787
+#
+from __future__ import absolute_import
```
IAlibay, line 23: I _think_ we can drop this (and __future__ imports in other files) now that we're past 1.0.

testsuite/MDAnalysisTests/topology/test_rdkit.py
```diff
@@ -0,0 +1,188 @@
...
+    def test_tripos_charges(self, top, filename):
+        expected = np.array([
+            a.GetDoubleProp('_TriposPartialCharge') for a in filename.GetAtoms()
+            ], dtype=np.float32)
+        assert_array_equal(expected, top.charges.values)
+
+    def test_aromaticity(self, top, filename):
+        expected = np.array([atom.GetIsAromatic() for atom in filename.GetAtoms()])
+        assert_array_equal(expected, top.aromaticities.values)
+
```
IAlibay, line 137: PEP8 two blank lines needed

package/MDAnalysis/core/selection.py
```diff
@@ -573,6 +573,20 @@ class AltlocSelection(StringSelection):
     field = 'altLocs'
 
 
+class AromaticSelection(Selection):
```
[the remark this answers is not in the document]
richardjgowers: Probably, but it's fine here.  This PR got a bit big as it was in a "holding pattern" while 1.0 was done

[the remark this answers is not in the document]
richardjgowers: This needs tests though.  Just simple `select_atoms('type C and aromatic')` and `select_atoms('type C and not aromatic')` would be enough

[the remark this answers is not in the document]
cbouy: tests are in testsuite/MDAnalysisTests/core/test_atomselections.py

testsuite/MDAnalysisTests/core/test_atomselections.py
richardjgowers: Missed these. Yeah I’d split this into two tests and combine with at least one other selection.

package/MDAnalysis/core/universe.py
cbouy: Since there are so many arguments that can be passed to this function, I went for a separate dict of kwargs. i.e. if you want to fix the random seed:
`mda.Universe.from_smiles("CCO", rdkit_kwargs=dict(randomSeed=42))`

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -0,0 +1,105 @@
...
+        ----------
+
+        filename : rdkit.Chem.rdchem.Mol
+            RDKit molecule
+        """
+        n_atoms = filename.GetNumAtoms()
+        coordinates = np.array([
+            conf.GetPositions() for conf in filename.GetConformers()], 
+            dtype=np.float32)
+        if coordinates.size == 0:
```
richardjgowers, line 102: Does this happen when the Mol has no coordinates? Worth raising a warning at least.

cbouy: Yes, and since the memory reader expects a numpy array I thought it would be best to have it at least with the right shape

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -0,0 +1,107 @@
...
+
+.. autoclass:: RDKitReader
+   :members:
+
+.. autoclass:: RDKitConverter
+   :members:
+
+
+"""
+from __future__ import absolute_import
```
IAlibay, line 56: We no longer need to support py2, so it's ok to remove the __future__ imports. Same for RDKITParser and the rdkit test files (the other files are probably covered by one of the other py2 removal PRs that got opened).

package/MDAnalysis/core/universe.py
```diff
@@ -1285,6 +1285,81 @@ def _fragdict(self):
...
+        >>> u = mda.Universe.from_smiles('CCO', rdkit_kwargs=dict(
+                                         params=AllChem.ETKDGv3()))
+        >>> u.trajectory
+        <RDKitReader with 1 frames of 9 atoms>
+        """
+        try:
+            from rdkit import Chem
+            from rdkit.Chem import AllChem
+        except ImportError as e:
+            six.raise_from(ImportError(
```
IAlibay, line 1343: Similarly to my import comment, we are dropping six, so you should just do "raise .. from e" here, although you probably don't need to do a raise from here.

package/MDAnalysis/core/universe.py
```diff
@@ -1285,6 +1285,81 @@ def _fragdict(self):
...
+        >>> u
+        <Universe with 9 atoms>
+        >>> u.trajectory
+        <RDKitReader with 10 frames of 9 atoms>
+
+        To use a different conformer generation algorithm, like ETKDGv3 ::
+        >>> u = mda.Universe.from_smiles('CCO', rdkit_kwargs=dict(
+                                         params=AllChem.ETKDGv3()))
+        >>> u.trajectory
+        <RDKitReader with 1 frames of 9 atoms>
```
IAlibay, line 1337: Could you put a versionadded here?

package/MDAnalysis/core/universe.py
```diff
@@ -1285,6 +1285,81 @@ def _fragdict(self):
...
+            raise SyntaxError('Error while parsing SMILES {0}'.format(smiles))
+        if addHs:
+            mol = Chem.AddHs(mol)
+        if generate_coordinates:
+            if not addHs:
+                raise ValueError("Generating coordinates requires adding "
+                "hydrogens with `addHs=True`")
+            
+            numConfs = rdkit_kwargs.pop("numConfs", numConfs)
+            assert (type(numConfs) is int) and (numConfs > 0), ("numConfs must"
```
IAlibay, line 1358: I remember @orbeckst telling me in my very first MDA PR that asserts should be kept in the realm of tests, any change we could use either a try/except or an if/else->except?

testsuite/MDAnalysisTests/coordinates/test_rdkit.py
```diff
@@ -0,0 +1,82 @@
...
+        return mda.Universe(mol)
+
+    def test_coordinates(self, mol, universe):
+        expected = np.array([
+            conf.GetPositions() for conf in mol.GetConformers()], 
+            dtype=np.float32)
+        assert universe.trajectory.n_frames == 1
+        assert_equal(expected, universe.trajectory.coordinate_array)
+
+    def test_multi_coordinates(self):
```
IAlibay, line 52: The overlap between this test and the above one is quite large, any chance we could squash them together and do a parametrize instead?

testsuite/MDAnalysisTests/core/test_atomselections.py
```diff
@@ -516,6 +516,32 @@ def test_molnum(self, universe, selection_string, reference):
...
+        pytest.importorskip("rdkit.Chem")
+
+    @pytest.fixture
+    def u(self):
+        smi = "Cc1cNcc1"
+        u = MDAnalysis.Universe.from_smiles(smi, addHs=False,
+                                            generate_coordinates=False)
+        return u
+        
+    def test_aromatic(self, u):
```
IAlibay, line 530: I think the following 4 tests can be squashed into one just pytest.mark.parametrize

package/MDAnalysis/core/universe.py
```diff
@@ -1285,6 +1285,81 @@ def _fragdict(self):
...
+            " be a non-zero positive integer instead of {0}".format(numConfs))
+            AllChem.EmbedMultipleConfs(mol, numConfs, **rdkit_kwargs)
```
IAlibay, line 1360: 👍 We probably should test this too (apologies if I happened to have missed the test for rdkit_kwargs).

package/MDAnalysis/coordinates/RDKit.py
cbouy: Do we also drop `six` ?

IAlibay: yup :)

package/MDAnalysis/core/universe.py
cbouy: It's implicitly tested in test_universe.py/test_coordinates_SMILES where I pass rdkit a randomSeed parameter through the rdkit_kwargs, but I will add one np

testsuite/MDAnalysisTests/coordinates/test_rdkit.py
```diff
@@ -0,0 +1,78 @@
...
+    mol = Chem.MolFromSmiles("CCO")
+    mol = Chem.AddHs(mol)
+    cids = AllChem.EmbedMultipleConfs(mol, numConfs=3)
+    return mol
+
+class TestRDKitReader(object):
+    @pytest.mark.parametrize("rdmol, n_frames", [
+        (mol2_mol(), 1),
+        (smiles_mol(), 3),
+    ])
```
cbouy, line 49: Not sure if this is a very pythonic way of testing things, but since parametrize won't take fixtures as args I decided to do it this way

package/MDAnalysis/coordinates/RDKit.py
IAlibay: Do we have a feeling for what the standard behaviour across the MDA readers is? I believe the netcdf reader will just throw an error if it can't find a reasonable dimension (my thought here is that we would want to have a similar behaviour to how other things work).

testsuite/MDAnalysisTests/coordinates/test_rdkit.py
richardjgowers: Yeah pytest can't parametrize over fixtures, it's annoying.  Not sure there's much better you can do here.

package/MDAnalysis/topology/RDKitParser.py
```diff
@@ -0,0 +1,286 @@
...
+                tempfactors.append(mi.GetTempFactor())
+            else:
+                # atom name (MOL2 only)
+                try:
+                    names.append(atom.GetProp('_TriposAtomName'))
+                except KeyError:
+                    pass
+                # atom type (MOL2 only)
+                try:
+                    atomtypes.append(atom.GetProp('_TriposAtomType'))
```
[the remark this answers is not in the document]
cbouy: I don't think "guaranteed" and "MOL2" can be used in the same sentence unfortunately, especially when it comes to atom types. Nevertheless, I made a quick test with the `Chem.MolFromMol2Block`: neither the atom name nor atom type can be blank, otherwise the resulting object is just `None`. So I guess if reading from a MOL2 file, they both have to be present otherwise RDKit will fail.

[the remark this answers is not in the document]
IAlibay: > I don't think "guaranteed" and "MOL2" can be used in the same sentence unfortunately,

😆 That's a very good point.

> So I guess if reading from a MOL2 file, they both have to be present otherwise RDKit will fail.

If you think it's worth it, it might make sense to put them within the same try/except block then , although maybe with a comment stating that we are making that assumption. My main concern here was whether or not there would be a large performance penalty to triggering the except twice if you were passing a non-MOL2 file, maybe worth benchmarking?

cbouy:
> Have you had a chance to build the docs and have a look at them? If you haven't, I'd recommend doing so, there's always a tiny thing that just doesn't build you expect it to.

I have to admit I skipped building the doc locally until now. I think I installed everything now but when I run `make html` in `/package/doc/sphinx` I get the following:
```
Warning, treated as error:
/mnt/e/Projets/mdanalysis/package/MDAnalysis/analysis/encore/similarity.py:docstring of MDAnalysis.analysis.encore.similarity.ces:1:duplicate object description of MDAnalysis.analysis.encore.similarity.ces, other instance in documentation_pages/analysis/encore/similarity, use :noindex: for one of them
Makefile:42: recipe for target 'html' failed
make: *** [html] Error 2
```
Also, is `make html` the right command to use ?

IAlibay:
> > Have you had a chance to build the docs and have a look at them? If you haven't, I'd recommend doing so, there's always a tiny thing that just doesn't build you expect it to.
> 
> I have to admit I skipped building the doc locally until now. I think I installed everything now but when I run `make html` in `/package/doc/sphinx` I get the following:
> 
> ```
> Warning, treated as error:
> /mnt/e/Projets/mdanalysis/package/MDAnalysis/analysis/encore/similarity.py:docstring of MDAnalysis.analysis.encore.similarity.ces:1:duplicate object description of MDAnalysis.analysis.encore.similarity.ces, other instance in documentation_pages/analysis/encore/similarity, use :noindex: for one of them
> Makefile:42: recipe for target 'html' failed
> make: *** [html] Error 2
> ```
> 
> Also, is `make html` the right command to use ?

I'm not sure where you got to in the process, @lilyminium made a very good guide on building the documentation: https://userguide.mdanalysis.org/contributing_code.html#working-with-the-code-documentation

You should just need to do `python setup.py build_sphinx -E` in `package`.

orbeckst:
Sphinx must be < 3. See #2667 .

For 

     python setup.py build_sphinx

You can also temporarily disable the warnings = error by changing https://github.com/MDAnalysis/mdanalysis/blob/788f294e53ca19fdfba8cc7e2f86d0de7be82bbf/package/setup.cfg#L18

This will always build all docs and for historical reasons they will be built in *package/doc/html/html/index.html*.

If you want faster turn-around while working on docs then use `make html`; edit the `Makefile` https://github.com/MDAnalysis/mdanalysis/blob/develop/package/doc/sphinx/Makefile#L6
and temporarily remove `-W` (warn -> errors) then run
```
cd package/doc/sphinx
make html
```
These docs will show up at *package/doc/html/index.html* (sigh... it's ugly).

IAlibay:
> > You should just need to do python setup.py build_sphinx -E in package
> 
> Got it, thanks! Still the same error though

Interesting, what version of sphinx are you running? I know that 2+ have issues with our docs (I think there's an issue open about it). With 1.8.5, I get the following:

```
Warning, treated as error:
/home/bioc1523/github/mdanalysis/package/MDAnalysis/core/universe.py:docstring of MDAnalysis.core.universe.Universe.from_smiles:34:Unexpected indentation.
```

Have you had any luck with compiling the docs from the standard develop branch?

(I have just noticed that your PR is coming from your fork's develop branch rather than a separate branch, you probably want to avoid doing that in the future).

package/MDAnalysis/topology/RDKitParser.py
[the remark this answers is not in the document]
cbouy: Tried with the SMILES string of caffeine (24 atoms, without generating coordinates), 10 runs, 100 loops each:
* same block: 8.51 ms ± 363 µs per loop
* 2 try/except blocks: 11.3 ms ± 1.07 ms per loop

With PDB_helix file (137 atoms):
* same block: 7.73 ms ± 255 µs per loop
* 2 try/except blocks: 7.31 ms ± 472 µs per loop

With MOL2 file (49 atoms):
* same block: 13.9 ms ± 458 µs per loop
* 2 try/except blocks: 14.7 ms ± 648 µs per loop

I don't see a large performance drop there so I guess we can keep it at 2 ?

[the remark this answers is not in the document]
IAlibay: > I don't see a large performance drop there so I guess we can keep it at 2 ?

I think that's fair unless @fiona-naughton or @richardjgowers have a different opinion here.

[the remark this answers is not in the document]
richardjgowers: This looks fine, we can do a performance pass if it ever becomes notably slow. I’m imagining any chem perception RDKit does is magnitudes slower than try/excepts on our end.

package/MDAnalysis/core/universe.py
```diff
@@ -1285,6 +1285,84 @@ def _fragdict(self):
...
+            Parameters passed on Universe creation
+
+        Returns
+        -------
+        :class:`~MDAnalysis.core.Universe`
+
+        Examples
+        --------
+        To create a Universe with 10 conformers of ethanol ::
+        >>> u = mda.Universe.from_smiles('CCO', numConfs=10)
```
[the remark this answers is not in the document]
cbouy: I actually just needed blank lines before the first `>>>` otherwise it wasn't detected as code, hence the indentation warning

cbouy:
Downgrading from 3.0.4 to 1.8.5 fixed the issues I mentioned earlier, thanks! And thanks for the tips @orbeckst !

@IAlibay I'll create a new branch next time my bad

cbouy:
Is this still for 1.0.0 ? I see that it's been timestamped to last week on the changelog. Or should I mark it for 1.0.1 in the docs and the changelog ?

richardjgowers:
This will be 2.0

package/MDAnalysis/core/selection.py
```diff
@@ -568,6 +568,21 @@ class AltlocSelection(StringSelection):
...
+    Aromaticity is available in the `aromaticities` attribute and is made 
+    available through RDKit"""
+    token = 'aromatic'
+    field = 'aromaticities'
+
+    def __init__(self, parser, tokens):
+        pass
+
+    def apply(self, group):
+        return group[group.aromaticities].unique
```
richardjgowers, line 583: I've changed this so that it will fail if there aren't aromaticities available, rather than silently putting everything as false

package/MDAnalysis/topology/RDKitParser.py
```diff
@@ -0,0 +1,285 @@
...
+            ):
+                attrs.append(Attr(np.array(vals, dtype=dtype)))
+        else:
+            attrs.append(Resids(np.array([1])))
+            attrs.append(Resnums(np.array([1])))
+            residx = None
+            n_residues = 1
+
+        # Segment
+        if any(segids) and not any(val is None for val in segids):
```
IAlibay, line 270: @cbouy Was just checking codecov to double check that there were enough tests and it looks like this branch of the code isn't tested at all (might be looking at an old commit). Could you double check that this is the case & add a test if so? https://codecov.io/gh/MDAnalysis/mdanalysis/src/dc482401df0d5c5839b2b37343b49a778987c2b1/package/MDAnalysis/topology/RDKitParser.py

richardjgowers:
I'll merge this so it's done, we can look at coverage + corner cases in future issues.

@cbouy first step done!

Issue #2362, opened 2019-10-10
Consolidate TopologyAttrs with consistent meanings


lilyminium:
TopologyAttrs are a bit of a mess right now, and the documentation is not up-to-date. This is a list of some inconsistencies I found.

## atomiccharge

The GAMESS parser uses the TopologyAttr `atomiccharge` for the atomic number. 

```python
>>> from MDAnalysis.tests.datafiles import *
>>> gms = mda.Universe(GMS_ASYMOPT)
>>> gms.atoms
<AtomGroup with 6 atoms>
>>> g = gms.atoms[0]
>>> g.atomiccharge
8
>>> g.name
'O'
>>> g.type
'O'
>>> g.element
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/lily/anaconda3/envs/mdanalysis/lib/python3.7/site-packages/MDAnalysis/core/groups.py", line 3566, in __getattr__
    cls=self.__class__.__name__, attr=attr))
AttributeError: Atom has no attribute element
```

This is *technically* correct, but I think most people would expect the `atomiccharge` to refer to a partial atomic charge. A more informational name might be `atomicnumber`.

## atomnum

The Desmond DMS parser uses the TopologyAttr `atomnum`. The [Desmond User Guide](http://www.deshawresearch.com/Desmond_Users_Guide-0.7.pdf) (section 17.1.1) asserts that 'anum' refers to the atomic number of a particle, but this seems implausible:

```python
>>> from MDAnalysis.tests.datafiles import DMS
>>> dms = mda.Universe(DMS)
>>> dms.atoms.names
array(['N', 'HT1', 'HT2', ..., 'C', 'OT1', 'OT2'], dtype=object)
>>> dms.atoms.atomnums
array([0, 0, 0, ..., 0, 0, 0], dtype=int32)
```

## element vs type

A non-exhaustive search indicates that only the prmtop TOPParser implements the `element` TopologyAttr. 

```python
>>> prm = mda.Universe(PRM)
>>> prm.atoms[:6].names
array(['N', 'H1', 'H2', 'H3', 'CA', 'HA'], dtype=object)
>>> prm.atoms[:6].types
array(['N3', 'H', 'H', 'H', 'CT', 'HP'], dtype=object)
>>> prm.atoms[:6].elements
array(['N', 'H', 'H', 'H', 'C', 'H'], dtype=object)
```

[...]

PR #2714, opened 2020-06-06
Removes element guessing for TOPParser

### LINKED ISSUE #2449: EP handling in parm7 files
**Expected behavior**

When loading a parm7 topology, extra points are not assumed to be elements.

**Actual behavior**

Loading in an ff19SB parm7 topology (or actually anything with the SPC/E water model) will yield a user warning that atoms of types EP have been assigned to P (since mass couldn't be recognised).

```
u = mda.Universe(<some parm7 topology>)
```

- Which version are you using? 0.20.1
- Which operating system? Ubuntu 18.04.3 LTS

**Actual comment**

Sorry not the most descriptive bug report, this is in some ways closer to feature request as I am opening this issue in order to get ideas on how this could be tackled.

So I don't think this is a new issue, but if possible I would like to fix this if possible as getting a bunch of phosphates is making things a tad bit more complicated.

Does anyone know if MDA already has a way to deal with extra points (I unfortunately couldn't find anything after an initial quick look through)?

### LINKED ISSUE #2651: Removing guess by default in TOPParser

IAlibay:
Fixes #2449, Partly fixes #2651 

 - Removes the element guessing default behaviour of the TOPParser

Other changes:
 - Cleans up the test_top.py to remove duplicate code
 - Adds new test datafile PRM19SBOPC which contains EPW atoms and CMAP entries (the latter may be useful in the future).

Note: Considering we are likely to go to 1.0.0 soon, I haven't updated the CHANGELOG and left the version as X.0.0. This PR is not necessary to go through before 1.0.0 is finalised, so I'll update this once it is reviewed or 1.0.0 happens (whichever happens first).

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/topology/TOPParser.py
```diff
@@ -277,27 +285,18 @@ def next_getter():
...
                               attrs.pop('diha'), attrs.pop('dihh'))
 
-        # Guess elements if not in topology or if any dummy atoms are found
-        if not 'elements' in attrs:
-            msg = ("ATOMIC_NUMBER record not found, guessing atom elements "
-                   "based on their atom types")
+        # Warn user if elements not in topology
+        if 'elements' not in attrs:
+            msg = ("ATOMIC_NUMBER record not found, elements attribute will "
+                   "not be populated")
```
lilyminium, line 291: Could you add something about `guess_elements` (or whatever the method's called?)

package/MDAnalysis/topology/TOPParser.py
```diff
@@ -60,6 +60,7 @@
 
 TODO:
   Add support for Chamber-style topologies
+  Add support for atomnum topology attributes
```
lilyminium, line 63: Is this atomic number style attributes? Not really relevant here but we might want to avoid calling them `atomnum` to a) avoid confusion with `resnum` and b) avoid confusion with whatever Desmond calls `atomnum`

package/MDAnalysis/topology/TOPParser.py
```diff
@@ -69,8 +70,12 @@
...
 
    Chamber-style Amber topologies (i.e. topologies generated via parmed
-   conversion of a CHARMM topology to an AMBER one) are not currently supported.
-   Support will likely be added in future MDAnalysis releases.
+   conversion of a CHARMM topology to an AMBER one) are not currently
+   supported. Support will likely be added in future MDAnalysis releases.
+
+   As of version X.0.0, elements are no longer guessed if ATOMIC_NUMBER records
+   are missing. In those scenarios, if elements are necessary, users will have
+   to invoke the element guessers after parsing the topology file.
```
lilyminium, line 78: Could you give an example or link to the relevant method?

package/MDAnalysis/topology/TOPParser.py
IAlibay: I think I might have misread https://github.com/MDAnalysis/mdanalysis/issues/2362#issue-505245572, I was under the impression that `atomnum` was our topology attribute for atomic numbers. As you mentioned in https://github.com/MDAnalysis/mdanalysis/issues/2651#issuecomment-605711880 it probably would be an idea to have an atomic number attribute. I can probably just change this to "Add support for storing atomic numbers", and we can decide what we should call this topology attribute at a later date?

lilyminium: Sounds good! We have an `atomnum`, but [it's only used in the Desmond parser and is all 0s](https://github.com/MDAnalysis/mdanalysis/issues/2362); not sure if that's a bug or real information. 

```python
>>> from MDAnalysis.tests.datafiles import DMS
>>> dms = mda.Universe(DMS)
>>> dms.atoms.names
array(['N', 'HT1', 'HT2', ..., 'C', 'OT1', 'OT2'], dtype=object)
>>> dms.atoms.atomnums
array([0, 0, 0, ..., 0, 0, 0], dtype=int32)
```

package/MDAnalysis/topology/TOPParser.py
IAlibay: Warnings now direct users to MDAnalysis.topology.guessers in 95cb49e

package/MDAnalysis/topology/TOPParser.py
IAlibay: I've added a link to MDAnalysis.topology.guessers, and then added detailed example uses of element guessing there. If this is reasonable, I'll try to add a more detailed tutorial to the user guide in the long term?

package/MDAnalysis/topology/TOPParser.py
richardjgowers: Hm, the desmond thing is a sqlite database if I remember right, so not sure how we're not getting that right.  I'd sooner coerce atomic numbers into elements for some uniformity rather than literalism in what the file provided.

IAlibay: >  I'd sooner coerce atomic numbers into elements for some uniformity rather than literalism in what the file provided.

So my thought there was that as @lilyminium mentioned https://github.com/MDAnalysis/mdanalysis/issues/2651#issuecomment-605711880, whilst we have to trust that the information provided in topology files is accurate, there might be some use cases where (i.e. dummy atoms) where one wants to have a copy of the atomic number record to do some kind of further processing. Additionally, I think @cbouy mentioned that it might be good to have this available for RDKIT parsing. So it might be a good idea to have both attributes available.

That being said, there's an argument that having both is redundant and maybe it would be easier to assume atomic number -> element conversion is a "guess" and drop populating elements in the TOPParser (I think at the moment, it's the only one that does this kind of automatic conversion)?

My suggested way forward here would be that we pin this PR until after v1.0 is out, then we go ahead with this PR (my view is that if anything not guessing from atom names here is going to reduce user errors), and then we have a longer discussion on atomic numbers vs elements as a separate issue aimed at v2.0? This way if we happen to change our mind, it won't be reflected too much on users.

package/MDAnalysis/topology/TOPParser.py
```diff
@@ -426,14 +427,18 @@ def parse_elements(self, num_per_record, numlines):
...
         Note
         ----
-        If the record contains atomic numbers <= 0, these will be treated as
-        dummy elements and an attempt will be made to guess the element based
-        on atom type. See issue #2306 for more details.
+        If the record contains unknown atomic numbers (e.g. <= 0), these will
+        be treated as unknown elements and assigned an empty string value. See
+        issues #2306 and #2651 for more details.
+
+        .. versionchanged X.0.0
```
lilyminium, line 434:
```suggestion
        .. versionchanged:: 2.0.0
```

testsuite/MDAnalysisTests/topology/test_top.py
```diff
@@ -36,13 +36,10 @@
     PRMErr1,
     PRMErr2,
     PRMErr3,
-    PRM_UreyBradley
+    PRM_UreyBradley,
+    PRM19SBOPC
```
lilyminium, line 40:
```suggestion
    PRM19SBOPC,
```

Just helps the next time we need to add a new file

testsuite/MDAnalysisTests/topology/test_top.py
```diff
@@ -460,3 +481,19 @@ def test_value_errors(self, parm, errmatch):
     def test_flag_index_error(self):
         with pytest.raises(IndexError, match="%FLAG section not found"):
             u = mda.Universe(PRMErr3)
+
+    @pytest.mark.parametrize("parm, errmsgs", (
+        [PRM, [ATOMIC_NUMBER_MSG, COORDINATE_READER_MSG]],
+        [PRM7, [ATOMIC_NUMBER_MSG, COORDINATE_READER_MSG]],
```
lilyminium, line 487: Nice, much more sustainable way to check warnings than we've been doing before

package/MDAnalysis/topology/TOPParser.py
IAlibay: Oops my bad, thanks for for the review :)

orbeckst:
@IAlibay – should this PR  be backported for 1.0.1 #2768 ?

I am currently assembling PR #2798 with the backport fixes for 1.0.1. If you want to add it to the backport, please do so. Any _fix_ may go in.

orbeckst:
Also, feel free to add anything for the backport to the [1.0.x milestone](https://github.com/MDAnalysis/mdanalysis/milestone/17) for keeping track.

IAlibay:
@orbeckst it makes things safer for users, so I'm not against backporting it. It might break a few workflows if they relied on automatically getting elements on reading prmtop files though.

I can make the PR and we can judge if it's too disruptive?

orbeckst:
I don't think we need a PR. The question is if it is a fix that is needed for correctness (then it has a place in 1.0.1) or we consider it a better choice (and we're making a design decision – then it's more a change than a fix). The latter is perfectly suited for 2.0.0.

IAlibay:
It's a bit of a tough call, technically MDA's own example files were leading to incorrect elements being guessed, however the decision to use the guesser to generate elements from atom types was intentional and we were warning users that this was happening. Since we're not exactly fixing the guessers but preventing users from automatically making what could be a bad decision, I'm going to says it's not "unintended behaviour" (if that makes sense) and therefore doesn't need to be backported to 1.0.1.

orbeckst:
Ok, thanks!

PR #2726, opened 2020-06-09
fix for gcc 9 (post 7 really)

### LINKED ISSUE #2722: lib.test_distances.test_coords_in_central_image_tric fails on some gcc version
## Expected behavior ##

All tests run.

## Actual behavior ##

Both the serial and openmp versions of lib.test_distances.test_coords_in_central_image_tric fail on "assert np.all(res >= 0.0)".

As pointed out by @lilyminium it looks like a precision issue. On generating `res`, res[0,0] and res[2, 0] have negative values of order - 1 e-7. This propagates down when calling `transform_RtoS` and causes the test to fail.

pytest test_distances.py

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) 0.20.2-dev0
- Which version of Python (`python -V`)? Many (3.6+)
- Failing gcc version: 9.3, 8.4 (tested by @lilyminium)
- Working versions of gcc: 7.4
- Working versions of clang: 11.0 (tested by @RMeli)
- Which operating system? Many

richardjgowers:
fixes failing triclinic image test

Fixes #2722

 - toned down optimisations to get tests passing

Longer term we'll need to take a closer look at handling of precision, but this fixes this failure.

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/setup.py
+++ b/package/setup.py
@@ -261,7 +261,8 @@ def extensions(config):
     use_cython = config.get('use_cython', default=not is_release)
     use_openmp = config.get('use_openmp', default=True)
 
-    extra_compile_args = ['-std=c99', '-ffast-math', '-O3', '-funroll-loops']
+    extra_compile_args = ['-std=c99', '-ffast-math', '-O3', '-funroll-loops',
+                          '-fsigned-zeros']  # see #2722
     define_macros = []
     if config.get('debug_cflags', default=False):
         extra_compile_args.extend(['-Wall', '-pedantic'])


richardjgowers:
this is @IAlibay 's fix which I've ruthlessly stolen.  There's no tests as the CI doesn't currently build for multiple compilers, but I'll raise an issue for that.  We've confirmed locally that this fix works.

Issue #2541, opened 2020-02-19
remove Python 2 constructs


orbeckst:
Release 2.0 does not support Python 2 anymore (see the [Roadmap](https://www.mdanalysis.org/2019/11/06/roadmap/)) so we should get rid of compatibility code (six and friends).

RMeli:
Which versions of Python 3 will be supported? (Asking mainly for f-strings, PEP 498)

orbeckst:
I don’t think we discussed this in any detail. Start a discussion on the dev list?

richardjgowers:
I think we usually followed whatever numpy was also doing, which I think was what continuous integration and conda-forge supported.  I think we usually have to be quite slow because we need to support that one cluster that still happens to run 2.7 in 2020.

IAlibay:
I don't know when 1.19 comes out, but `numpy` seems to have dropped support for 3.5 on its development branch: https://github.com/numpy/numpy/blob/master/INSTALL.rst.txt#prerequisites

My understanding is that major releases of `numpy` are scheduled ~ 6 months, which means 1.19 should come in late June. So that might align itself quite well with a late Q2 goal for MDA 2.0.

tylerjereddy:
SciPy will also be 3.6+ only for the next release (1.5.x). NumPy & SciPy have started to remove large amounts of Python 2 compatibility constructs in their master branches now that their Python 2.7 LTS branches have had what should be their final upates/need for backports.

orbeckst:
For 2.0 I don't have a problem starting with Python >= 3.6.

IAlibay:
Going to link https://github.com/MDAnalysis/mdanalysis/issues/2739#issuecomment-642362469 back to here.

My current approach has been to remove six / __future__ things every time I deal with a file. We probably want a means of tracking progress though.

IAlibay:
As mentioned in https://github.com/MDAnalysis/mdanalysis/pull/2750#issuecomment-643514970 it might be that we want to chip away at this in stages. Having discussed this with @lilyminium, it might be an idea to break it down per "section" where we cover what is both in package & testsuite:
  - [ ] analysis
  - [ ] auxiliary
  - [ ] core
  - [ ] lib
  - [ ] selections
  - [ ] tests
  - [ ] topology
  - [ ] transformations
  - [ ] visualisation
  - [ ] misc (benchmarks, maintainer, setup scripts, datafiles, etc...)

PR #2729, opened 2020-06-10
Start 1.0.1

richardjgowers:
Start of the next release

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,13 +65,6 @@ matrix:
            CONDA_DEPENDENCIES=""
            INSTALL_HOLE="false"
 
-    - env: NAME="python 2.7"
-           PYTHON_VERSION=2.7
-           CODECOV="true"
-           NUMPY_VERSION=1.16
-           SETUP_CMD="${PYTEST_FLAGS} --cov=MDAnalysis"
-           PIP_DEPENDENCIES="${PIP_DEPENDENCIES} setuptools<45.0.0"
-
     - env: NAME="asv check"
            PYTHON_VERSION=2.7
            CODECOV="false"
@@ -84,14 +77,6 @@ matrix:
     - os: osx
       env: PYTHON_VERSION=3.6
            NUMPY_VERSION=1.17.3
-
-# macOS Python 2.7 works in principle but times out on Travis CI:
-# Disabled so that the tests can pass.
-#    - os: osx
-#      env: PYTHON_VERSION=2.7
-#           NUMPY_VERSION=1.16
-#           CODECOV="true"
-#           SETUP_CMD="${PYTEST_FLAGS} --cov=MDAnalysis"
 
     - env: NAME='minimal'
            PIP_DEPENDENCIES=""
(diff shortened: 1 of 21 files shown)


.travis.yml
```diff
@@ -65,13 +65,6 @@ matrix:
            CONDA_DEPENDENCIES=""
            INSTALL_HOLE="false"
 
-    - env: NAME="python 2.7"
```
IAlibay, line 68: Can't remember the exact issue but I think we said we would drop 3.5 too.

IAlibay: I think it was this one: https://github.com/MDAnalysis/mdanalysis/issues/2541#issuecomment-590611719

lilyminium: I *thought* dropping py2.7 and py3.5 would be a 2.0 thing, and that 1.xx releases would be just bug fixes, based on that same comment?

IAlibay: Pinging @orbeckst here

richardjgowers: It looks like everyone else is moving on, I'd like to just ditch python 2.  If we find anything nasty that we fix I could do a backport/LTS thing, but I'd rather just force everyone to move to py 3, anaconda makes this possible enough.

RMeli: I was about to tag that comment! ;) I think this discussion is suited for the dev list, but if major packages are already dropping support for 3.5 (see https://github.com/MDAnalysis/mdanalysis/issues/2541#issuecomment-589831731) probably we should drop it too.

RMeli: > I thought dropping py2.7 and py3.5 would be a 2.0 thing, and that 1.xx releases would be just bug fixes, based on that same comment?

My view here (probably to be discussed on the dev list) is that it would be nice to increase the number of releases to include small bug fixes. The reason is that many issues are solved relatively quickly, but make it on master after a very long time (and therefore most users don't benefit from them).

I think we should start dropping Python 2 and 3.5 support with 1.0.x releases (as done here) and tidy up the code as much as possible on the way to 2.0 (possibly with type hints, and other cool stuff).

package/MDAnalysis/version.py
```diff
@@ -67,4 +67,4 @@
 # e.g. with lib.log
 
 #: Release of MDAnalysis as a string, using `semantic versioning`_.
-__version__ = "0.20.2-dev0"  # NOTE: keep in sync with RELEASE in setup.py
+__version__ = "1.0.0-dev0"  # NOTE: keep in sync with RELEASE in setup.py
```
RMeli, line 70: Tests are failing because this tag is different from the one in `MDAnalysisTests/__init__.py`.

```
E       AssertionError: MDAnalysis release 1.0.1-dev0 must be installed to have meaningful tests, not 1.0.0-dev0
E       assert '1.0.0-dev0' == '1.0.1-dev0'
E         - 1.0.1-dev0
E         ?     ^
E         + 1.0.0-dev0
E         ?     ^

```

.travis.yml
orbeckst: In https://www.mdanalysis.org/2019/11/06/roadmap/ we said _1.x will be the last version of MDAnalysis that fully supports Python 2.7_

This means that 1.x supports and tests 2.7 so we cannot remove Python 2.7 tests. Drop it for 2.0.

I'll start a discussion on the dev list instead of an issue thread...

PR #2733, opened 2020-06-10
Added space to separate field in output files.

rcrehuet:
The output files had all the the floating point numbers joined in a single string. 
I added a whitespace separator. I may have added some unnecessary whitespaces, but I don't think that creates any problem. In other words, I did not check that all added whitespeces where needed, some may go at the end of the line, where they are superflous but not problematic.

### DIFF
--- a/package/MDAnalysis/analysis/helanal.py
+++ b/package/MDAnalysis/analysis/helanal.py
@@ -376,28 +376,28 @@ def helanal_trajectory(universe, select="name CA",
 
         #print out rotations across the helix to a file
         with open(twist_filename, "a") as twist_output:
-            print(frame, end='', file=twist_output)
+            print(frame, end=' ', file=twist_output)
             for loc_twist in twist:
-                print(loc_twist, end='', file=twist_output)
+                print(loc_twist, end=' ', file=twist_output)
             print("", file=twist_output)
 
         with open(bend_filename, "a") as bend_output:
-            print(frame, end='', file=bend_output)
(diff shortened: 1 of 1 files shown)


lilyminium:
I can't remember the details anymore but #2622 does alter the algorithm, e.g. it safeguards helix height so it can no longer be negative, and the screw angles are now different. It is currently stalled as @fiona-naughton and I have been working off different interpretations of what a screw angle is -- [outside opinions are *very welcome*](https://github.com/MDAnalysis/mdanalysis/pull/2622#discussion_r437239683) -- but otherwise nearly there.

fiona-naughton:
Progress with #2622 has unstalled for now; but this does make sense if we're keeping the old helanal around. 

There's also an issue where `ref_axis` is never actually passed to the function that calculates the helix tilt angles (`vector_of_best_fit`) - [the z-axis is always used by defalt](https://github.com/MDAnalysis/mdanalysis/blob/795a5b71066ae93742cc93c4ad59cef13c9b8e1f/package/MDAnalysis/analysis/helanal.py#L810). Wouldn't take much to fix, so that could be done here too?

orbeckst:
We can back-port it as a fix in #2798 .

All I'd need here are additions to CHANGELOG and AUTHORS.

orbeckst:
I am including the patch 8f729d8150d086968f660947ed941d8a2360ec4d as a backport for 1.0.1 in PR #2798 and I added @rcrehuet to the AUTHORS file.

Thank you!

Issue #2739, opened 2020-06-10
Deprecated and slated for 2.0 removal


IAlibay:
We have quite a lot of things that are slated for removal in 2.0:

- [x] old density code #2740 
- [x] old hole code #2778 
- [ ] old hbonds code #2746 
- [x] deprecated writers Timestep code #2754 
- [x] progressmeter #2743 

With #2729 I guess we started 2.0 a little bit sooner than anticipated, so I thought I'd check that we are good at doing these removals before we go ahead and start cutting out code?

orbeckst:
Let‘s get started. We‘ll have to figure out a way to do patches for 1.0 but it seems we can just do PRs against master at the moment and treat master as 1.x branch. 

We‘ll figure out a way to backport fixes from develop; I assume that if we do squash merges we get fairly clean diffs that we can cherry-pick or merge into master/1.x.

orbeckst:
Also: remove six!

IAlibay:
@richardjgowers do you mind if I just move hbond_autocorrel to hydrogenbonds for this?

richardjgowers:
Yeah moving it is fine

PR #2740, opened 2020-06-11
Removes deprecated density code

IAlibay:
Towards #2739 

 - Removes density_from_Universe, density_from_PDB, Bfactor2RMSF, BfactorDensityCreator, and notwithin_coordinates_factory
 - Moves _set_user_grid to be a staticmethod of DensityAnalysis
 - Unused tests removed

Todo:
  - Do we need a versionchanged somewhere in the docs?

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


IAlibay:
In the interest of not having too many PRs dependent on other PRs, I've reverted `from __future__` changes. I can remove them once we deal with the linter.

I've also added a versionchanged to make it clear to anyone looking at the old density code is gone.

package/MDAnalysis/analysis/density.py
```diff
@@ -35,11 +35,17 @@
 The module provides classes and functions to generate and represent
 volumetric data, in particular densities.
 
+
+.. versionchanged:: 2.0.0
```
orbeckst, line 39: This is good, thanks.

package/MDAnalysis/analysis/density.py
```diff
@@ -170,10 +153,7 @@
 .. _`gmx trjconv`: http://manual.gromacs.org/programs/gmx-trjconv.html
 
 """
-
-from __future__ import print_function, division, absolute_import
-from six.moves import range, zip
-from six import raise_from, string_types
+from __future__ import absolute_import
```
orbeckst, line 156: Thank you for the explanation why this is still there. I didn’t understand it on the other PR that I looked at. The linter PR is merged so this can go (when you merge develop or rebase).

package/MDAnalysis/analysis/density.py
```diff
@@ -455,6 +436,63 @@ def _conclude(self):
         density.make_density()
         self.density = density
 
+    @staticmethod
```
orbeckst, line 439: Can you make sure that this private method is actually showing up in the docs? This seems useful to have documented so that one knows what happens when the grid is set.

package/MDAnalysis/analysis/density.py
```diff
@@ -785,709 +822,3 @@ def __repr__(self):
...
-            gridcenter is ``None``
-
-    Returns
-    -------
-    :class:`Density`
-            A :class:`Density` instance with the histogrammed data together
-            with associated metadata.
-
-
-    Notes
```
orbeckst, line 920: There’s quite a bit of docs. Are these docs (at least in spirit) available in the module/DensityAnalysis?

(Sorry, a bit difficult to see this from the diff.)

package/MDAnalysis/analysis/density.py
IAlibay: I'm going to open a separate "remove py2 stuff from analysis methods" PR coming along at some point this morning, I'll remove it with that if that works for you?

IAlibay: See #2753

package/MDAnalysis/analysis/density.py
IAlibay: Fixed in c2c94b7

orbeckst:
> On Jun 13, 2020, at 12:48 AM, Irfan Alibay <notifications@github.com> wrote:
> 
> I'm going to open a separate "remove py2 stuff from analysis methods" PR coming along at some point this morning, I'll remove it with that if that works for you?

Absolutely!

package/MDAnalysis/analysis/density.py
IAlibay: Looks to me like you moved everything that was necessary over when you created `DensityAnalysis` :)
The only thing I could see that was "missing" was a `Return` entry, so I added that (emphasizing on the fact that this is a physical density so users are aware).

orbeckst: thanks for checking & updating, much appreciated!

orbeckst:
Thanks for the work!

PR #2743, opened 2020-06-11
Remove ProgressMeter

IAlibay:
Towards #2739 

 - Removes deprecated ProgressMeter

Questions:
 - Do we need a versionchanged in lib.log?

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/lib/log.py
```diff
@@ -80,12 +80,11 @@
 .. autogenerated, see Online Docs
 
 """
-from __future__ import print_function, division, absolute_import
+from __future__ import absolute_import
```
orbeckst, line 84: Remove?

orbeckst:
And yay, -282 lines of code!

Btw, if there’s widespread support for not adding versionchanged for dropped content then say so. I am just thinking that the technical docs can be technical and have useful information. Anything we put in the docs now will hopefully allow users to come to their own conclusions as opposed to us explaining.

IAlibay:
> And yay, -282 lines of code!
> 
> Btw, if there’s widespread support for not adding versionchanged for dropped content then say so. I am just thinking that the technical docs can be technical and have useful information. Anything we put in the docs now will hopefully allow users to come to their own conclusions as opposed to us explaining.

I think my main thing was that I wasn't too sure where the versionchanged should go, I've just placed one under "13.2.6.2. Other functions and classes for logging purposes" hopefully it looks reasonable.

package/MDAnalysis/lib/log.py
IAlibay: Sure, I'll remove all the py2 stuff from lib at the same time.

PR #2749, opened 2020-06-11
Add scripts for doc versioning

### LINKED ISSUE #2741: Versioned docs

With 2.0 coming soon (hopefully) we really need versioned documentation.

Based on [related discussion on the user guide](https://github.com/MDAnalysis/UserGuide/issues/67) I'm going to try building this into the Travis pipeline instead of hosting on RTD or similar services.

The alternative is hosting on ReadTheDocs, but as @orbeckst points out that comes with potential added problems if RTD breaks and also a small RTD ad in the sidebar.

## Additional context ##

Currently our docs are at mdanalysis.org/docs. We could:

- keep the status quo and have mdanalysis.org/docs/$version urls, where mdanalysis.org/docs redirects to mdanalysis.org/docs/$latest_release and mdanalysis.org/mdanalysis redirects to mdanalysis.org/docs/$dev_version
- switch it up to docs.mdanalysis.org/$version , where both mdanalysis.org/docs and docs.mdanalysis.org/ redirect to docs.mdanalysis.org/$latest_release

I really like the conceptual separation of the second one. All it requires is changing the URL in settings from mdanalysis.org/mdanalysis to docs.mdanalysis.org/, and adding a CNAME DNS record. Any opinions?

lilyminium:
Fixes #2741 ... maybe

 - added msmb_theme because they have a cool version sidebar that replicates the RTD one
 - added doc moving scripts to maintainer/
 - get travis to call them in place of the old deploy scripts
 - write stubs for index.html, dev.html and latest.html to redirect them to the relevant versions
 - ~add RTD config~

~This last one is so ~people can choose to hook up ReadTheDocs to their own forks and see the docs built there~ we can add an RTD check that builds the docs for you, which also makes it easier to review that kind of thing. This is currently in progress.~ It's quite hard to build MDAnalysis and have it accessible to `conf.py`, which hardcodes in looking three levels up, except RTD I'm pretty sure builds it in the same level as the root directory.

Examples:
- https://lilyminium.github.io/mdanalysis/1.0.0/
- https://lilyminium.github.io/mdanalysis/2.0.0-dev0/

I've kept the status quo for now (everything branches off mdanalysis.org/mdanalysis) but final URLs are still up for discussion at #2741 

PR Checklist
------------
 - [ ] Tests?
 - [x] Docs? yes!
 - [ ] CHANGELOG updated?

TO DO
-----
Figure out what the following need to be and make them like so:
- [ ] sitemap.xml
- [ ] searchindex.js
- [ ] genindex.html
- [ ] objects.inv
- [ ] py-modindex.html


lilyminium:
I guess one thing I haven't thought through is how to trigger this for old versions, i.e. 1.0.0 . The one I have at the address up there is just `develop` with `version.py` edited to say differently

package/MDAnalysis/version.py
```diff
@@ -68,3 +68,4 @@
 
 #: Release of MDAnalysis as a string, using `semantic versioning`_.
 __version__ = "2.0.0-dev0"  # NOTE: keep in sync with RELEASE in setup.py
+release = False
```
richardjgowers, line 71: Ideally we can have this as something like `release = not 'dev0' in __version__`?

lilyminium: Oops, yeah, that's what the script does anyway so this line is redundant now.

lilyminium:
Ok it's weirdly hard to build MDAnalysis in the right place to be accessible to docs on ReadTheDocs; I give up for now.

I added in a function to get the absolute paths of files to setup.py, when trying to get RTD to work; I can remove this if unwanted. It doesn't really do any harm and might help the odd person wanting to go `python package/setup.py install`.

orbeckst:
Very cool.

If we do a backport to 1.0.x then it should create the proper 1.0.x docs for correct linking, no? (If I am in your "1.0.0" docs then the version link to 2.0.0-dev is broken https://lilyminium.github.io/mdanalysis/1.0.0/lilyminium.github.io/mdanalysis/2.0.0-dev0).

(RTD seems a bit limited in how you can play with it, I remember being pretty frustrated as soon as stuff didn't work out of the box... when it works it's great, when not, it's pain.)

lilyminium:
Oops. I hope that does it. (lilyminium.github.io/mdanalysis now redirects to mdadocs.minium.com.au, just temporarily)

I also added index.html, dev.html and latest.html stubs so that mdadocs.minium.com.au/ redirects to mdadocs.minium.com.au/1.0.0 , mdadocs.minium.com.au/latest redirects to mdadocs.minium.com.au/1.0.0, mdadocs.minium.com.au/dev redirects to mdadocs.minium.com.au/2.0.0-dev0 . This is written by the `update_json_versions.py` script every time it's updated.

MDTraj uses Javascript to accomplish the same redirect, but this seemed easier to me both in case people don't have JS turned on and also so we don't have to wait for the JS to figure out the dev URL.

```html
<html><head>
<!-- redirect to /development after 3 seconds, in case something fails or
    the user has javascript disabled -->
<meta http-equiv="refresh" content="3;URL='/development'"/>

<script type="text/javascript">
// download the versions.json file and redirect to the URL associated
// with the entry that has latest=true
var xhr = new XMLHttpRequest();
xhr.open('GET', encodeURI('versions.json'));
xhr.onload = function() {
    if (xhr.status === 200) {
        var versions = JSON.parse(xhr.responseText);
        for (var key in versions) {
            if (versions[key]['latest']) {
                var latest = versions[key];
                window.location.replace(latest['url']);
            }
        }
    }
}
xhr.send();
</script>
</head></html>
```

.travis.yml
```diff
@@ -131,6 +132,12 @@ after_success:
       codecov; \
     fi
   # can't use test here since this leads to travis fails even though the build passes
-  - if [[ ${TRAVIS_PULL_REQUEST} == "false" ]] && [[ ${BUILD_DOCS} == "true" ]] && [[ ${TRAVIS_BRANCH} == ${GH_DOC_BRANCH} ]]; then
-            bash ${TRAVIS_BUILD_DIR}/maintainer/deploy_docs.sh;
+  - |
+    if [[ ${TRAVIS_PULL_REQUEST} == "false" ]] && [[ ${BUILD_DOCS} == "true" ]] && [[ ${TRAVIS_BRANCH} == ${GH_DOC_BRANCH} ]]; then
+      python -c 'import os,sys,fcntl; flags = fcntl.fcntl(sys.stdout, fcntl.F_GETFL); fcntl.fcntl(sys.stdout, fcntl.F_SETFL, flags&~os.O_NONBLOCK);'
```
orbeckst, line 137: Just out of curiosity: why is this needed?

maintainer/update_versions_json.py
```diff
@@ -0,0 +1,93 @@
+# Adapted from mdtraj/devtools/travis-ci/update_versions_json.py by Robert T McGibbon
```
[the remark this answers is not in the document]
lilyminium: Yes it is (well the entire repo is), and 80% of the script is different, but it was adapted from his script originally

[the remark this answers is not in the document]
orbeckst: If you add "used under LGPL" then there's not doubt.

maintainer/update_versions_json.py
```diff
@@ -0,0 +1,93 @@
...
+import json
+import os
+
+try:
+    from urllib.request import Request, urlopen
+except ImportError:
+    from urllib2 import Request, urlopen
+
+# ========= WRITE JSON =========
+URL = "https://mdanalysis.github.io/mdanalysis/"
```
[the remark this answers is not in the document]
lilyminium: We could. I left it as such in case the URL would change in the future. mdanalysis.github.io/mdanalysis will redirect to docs.mdanalysis.org/ now, and will redirect to whatever is set as the URL in the GitHub settings in the future, so this seemed more extensible. I'm not sure if it actually slows down web traffic by redirecting; I can definitely change it

.travis.yml
lilyminium: It was due to this error, and the solution is from there: https://github.com/travis-ci/travis-ci/issues/4704. The build kept erroring from an unavailable resource; I think what was happening was that I had multiple branches going at once (1.0.0 and 2.0.0-dev0), causing the other branch to error instead of waiting its turn, but I didn't precisely narrow down the cause per se.

orbeckst: Thanks for the explanation. Perhaps add a comment with a link to https://github.com/travis-ci/travis-ci/issues/4704 ?

Kudos to you for finding the fix!

maintainer/update_versions_json.py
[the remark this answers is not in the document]
orbeckst: I assume @rmcgibbo will be ok with this attribution? (Thank you!)

maintainer/deploy_docs_via_travis.sh
```diff
@@ -0,0 +1,73 @@
...
+git remote add upstream "https://${GH_TOKEN}@${GH_REPOSITORY}"
+git fetch --depth 50 upstream ${GH_DOC_BRANCH} gh-pages
+git reset upstream/gh-pages
+
+touch .
+touch .nojekyll
+
+echo $VERSION
+echo "ls *"
+ls *
```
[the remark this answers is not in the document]
lilyminium: Only if it breaks -- removed.

lilyminium:
>  By the way, are the sitemap.xml files created correctly?

No, at present they're caught up in each $version directory. They're built by Sphinx before moving all the docs into each subdirectory, so I think what we would need to do is create a new one by concatenating them all into one big `sitemap.xml` at the root directory.

lilyminium:
So, each subdirectory should function as its own little website for the most part. Each of the following uses relative links, so they *should* be fine in the subdirectories instead of the root directory without further editing:

* search.html
* py-modindex.html
* searchindex.js
* genindex.html

I guess people intersphinxing with `objects.inv` will have to link to the right URL? Unfortunately, `mdadocs.minium.com.au/latest` redirects to the relevant release, but `mdadocs.minium.com.au/latest/objects.inv` will error. We could copy the relevant `objects.inv` into `latest/`, but I think intersphinx uses the given URL in `conf.py` as the base URL, so the same problem remains. Two suggestions:

1. copy the entire set of docs into a latest/ and a dev/ directory. This is redundant but easy and robust. Or,
2. add a CloudFlare pagerule converting \*.mdanalysis.org/latest/\* to $1.mdanalysis.org/$version/$2. We would need another for `/dev/`. This is doable if we remove the current rule for docs.mdanalysis.org (currently off) and the 2018 workshop. It would require manual updating for each release.

The new root sitemap.xml just links to the other sitemaps for reasons explained in the script (the file must be a max of 50 MB or 50k URLs) but it can easily be one very big sitemap.xml. By the way, I think the original sitemaps used incorrect URLs with shortened versions:

**docs.mdanalysis.org/sitemap.xml**
```xml
<?xml version='1.0' encoding='utf-8'?>
<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9"><url><loc>https://www.mdanalysis.org/mdanalysis/2.0/documentation_pages/analysis/align.html</loc>
</url>
<url>
    <loc>https://www.mdanalysis.org/mdanalysis/2.0/documentation_pages/analysis/base.html</loc>
</url>
<url>
    <loc>https://www.mdanalysis.org/mdanalysis/2.0/documentation_pages/analysis/contacts.html</loc>
</url>
```

package/doc/sphinx/source/conf.py
```diff
@@ -46,7 +47,7 @@
 #       are served from https://www.mdanalysis.org/docs/ and the sitemap.xml
 #       is manually fixed when deploying the release docs with the
 #       maintainer/deploy_master_docs.sh script
```
orbeckst, line 49: update comment

package/doc/sphinx/source/conf.py
```diff
@@ -46,7 +47,7 @@
...
-site_url = "https://www.mdanalysis.org/mdanalysis/"
+site_url = "https://docs.mdanalysis.org/"
```
orbeckst, line 50: Add comment regarding our DNS set up (requires a CNAME etc). The previous "https://www.mdanalysis.org/mdanalysis/" would work automatically with the standard GitHub setup where any gh-pages repo would appear under our top level. With the new set up we need correct DNS.

package/doc/sphinx/source/conf.py
```diff
@@ -156,8 +158,15 @@
     'titles_only': False,
 }
 
+html_context = {
+    'versions_json_url': 'https://mdanalysis.github.io/mdanalysis/versions.json'
```
orbeckst, line 162: Does this have to be the true URL or could it be our CNAME, for consistency?

lilyminium: It can be any valid address, so I've changed it to docs.mdanalysis.org :-)

lilyminium:
Well, this is interesting. I think until this PR is merged, every other merge will reset the docs URL back to mdanalysis.org/mdanalysis by wiping the CNAME record from `gh-pages` -- specifically due to this line that removes all entries not built in the docs by Travis:

https://github.com/MDAnalysis/mdanalysis/blob/46e505c0805e6a6f3f0933c62731be95edd1c59f/maintainer/deploy_docs.sh#L49

lilyminium:
https://docs.mdanalysis.org/2.0.0-dev0/

* ✅ versions.json looks fine
* ✅ sitemap_index.xml looks fine
* ✅ URLs in 2.0.0-dev0/sitemap.xml look sensible
* ✅ CNAME not overwritten
* ✅ [docs.mdanalysis.org/](docs.mdanalysis.org/) , [docs.mdanalysis.org/dev](docs.mdanalysis.org/dev) and [docs.mdanalysis.org/latest](docs.mdanalysis.org/latest) all redirecting to 2.0.0-dev0 (if no proper release is found, it just goes to the most recent URL)
* 😕 forgot that we need to remove existing files like objects.inv and sitemap.xml from `gh_pages`, but that doesn't necessarily mess anything up

Next: revert back to old 1.0.0 docs and do those...

PR #2750, opened 2020-06-12
Pylint checks for Python 3 only

### LINKED ISSUE #2744: Pylint config for python 3+
## Expected behavior ##

Now that we are using python 3+ removing __future__ imports shouldn't fail (unless I'm missing something obvious).

## Actual behavior ##

Remove a future import in a PR

## Possible fix ##

Looks like we have to remove some of the future imports from .pylintrc?

lilyminium:
Fixes #2744

 - get the linter check to error on Python 3 things only

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,7 @@ matrix:
 
     - env: NAME="Lint"
            PYLINTRC="${TRAVIS_BUILD_DIR}/package/.pylintrc"
-           MAIN_CMD="pylint package/MDAnalysis && pylint testsuite/MDAnalysisTests"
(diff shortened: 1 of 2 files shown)


lilyminium:
We can either merge all `__future__` cleanups into this PR or merge this into develop first and have separate ones for clean-up. I haven't checked if the old linter looks for `six` etc.

orbeckst:
I know that I previously said to remove everything in one go together with the check. But looking at the PR, it seems more conducive to moving things along if we merged it right away and then people can chip away at ripping out Py 2 when they feel they need some light distraction that still passes as doing something productive.

PR #2754, opened 2020-06-14
Remove ts from writers

IAlibay:
Towards #2739 

 - Removed deprecated use of timestep from writers.
 - Removed Writer.write_next_timestep
 - Removes support for passing `atoms` to XYZWriter (code paths here no longer worked since we removed support for Timestep).
 - Partial py2 removals for coordinates (rest to follow in a separate PR).
 - Mostly aligned error checking for using Timestep on writers.

To do:
  - [x] Fix broken tests (chainreader, netcdf, lammps, trz, xdr, xyz currently failing).
  - [x] Add checks for raising TypeError on calling with timestep?

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/coordinates/null.py
```diff
@@ -57,4 +57,9 @@ def __init__(self, filename, **kwargs):
         pass
 
     def _write_next_frame(self, obj):
+        try:
```
IAlibay, line 60: I went ahead and added this check to the null writer. My understanding is that the null writer should behave like all other writers, so it should complain if we pass it a Timestep. Thoughts anyone?

richardjgowers:
Sounds sane

package/MDAnalysis/coordinates/GRO.py
```diff
@@ -357,31 +353,24 @@ def write(self, obj):
            *resName* and *atomName* are truncated to a maximum of 5 characters
         .. versionchanged:: 0.16.0
            `frame` kwarg has been removed
-        .. deprecated:: 1.0.0
```
richardjgowers, line 360: I think we’re meant to keep the old tag and just keep appending new stuff

package/MDAnalysis/coordinates/DCD.py
```diff
@@ -402,23 +399,22 @@ def _write_next_frame(self, ag):
...
            Renamed from `write_next_timestep` to `_write_next_frame`.
+        .. versionchanged:: 2.0.0
+           Deprecated support for Timestep argument has now been removed.
+           Use AtomGroup or Universe as an input instead.
         """
-        if isinstance(ag, base.Timestep):
-            ts = ag
-        else:
+        try:
+            ts = ag.ts
```
richardjgowers, line 410: Constructing an AtomGroup timestep ain’t free, so maybe use ag.positions to grab the positions etc

package/MDAnalysis/coordinates/MOL2.py
```diff
@@ -307,7 +307,12 @@ def encode_block(self, obj):
         obj : AtomGroup or Universe
         """
         # Issue 2717
-        obj = obj.atoms
+        try:
+            obj = obj.atoms
```
richardjgowers, line 311: I’d hope stuff is type checked by this point

package/MDAnalysis/coordinates/NAMDBIN.py
```diff
@@ -114,31 +112,30 @@ def __init__(self, filename, n_atoms=None, **kwargs):
...
         .. versionchanged:: 1.0.0
            Renamed from `write` to `_write_next_frame`.
-        .. deprecated:: 1.0.0
-           Passing a Timestep is deprecated for removal in version 2.0
+        .. versionchanged:: 2.0.0
+           Deprecated support for Timestep argument has now been removed.
+           Use AtomGroup or Universe as an input instead.
         """
-        # TODO 2.0: Remove Timestep logic
-        if isinstance(obj, base.Timestep):
```
richardjgowers, line 133: Same here, put a type check in base.Write

package/MDAnalysis/coordinates/GRO.py
IAlibay: This came up before in: https://github.com/MDAnalysis/mdanalysis/pull/2494#discussion_r374805918 and I think the idea at the time was to remove deprecation notices since they get replaced by the versionchanged.

package/MDAnalysis/coordinates/MOL2.py
IAlibay: Unfortunately this is the failure point if you pass a Timestep object to the MOL2Writer. I could move it up to _write_next_frame if you think it's more appropriate?

package/MDAnalysis/coordinates/DCD.py
IAlibay: So from a quick look, I think DCD, TRJ, TRR, TRZ, XTC, XYZ, and chemfiles all do something similar. In the interest of not making this PR too hard to follow, would you be ok if I opened an issue for this and dealt with it in a separate PR?

package/MDAnalysis/coordinates/GRO.py
orbeckst: I am glad that you have a good memory @IAlibay ! 

I added a section under [Writing Documentation: Documenting changes](https://github.com/MDAnalysis/mdanalysis/wiki/WritingDocumentation#documenting-changes) to capture what I believe is our current consensus. It's up for discussion, of course.

@lilyminium I know that the User Guide copied most of the wiki style guide so I created PR https://github.com/MDAnalysis/UserGuide/pull/76 with these changes.

PR #2755, opened 2020-06-14
Faster name selections

### LINKED ISSUE #2751: simple selection by name is slow due to fnmatch
## Expected behavior ##

Simple atom selections such as `u.select_atoms("protein")` or `u.select_atoms("name CA")` or "resname LYS" are not slowed down by advanced selection capabilities.

## Actual behavior ##

The benchmarks  https://www.mdanalysis.org/benchmarks/#selections.SimpleSelectionBench.time_simple_selections?p-selection_string='name%20CA'&six=%5Bnone%5D&ram=16GB clearly show that performance got worse by 53.58x (from 997.988μs to 53.474ms) when PR #2551 , which introduced *fnmatch*, was merged. Specifically, the commit https://github.com/MDAnalysis/mdanalysis/commit/eb18a338 is the point in the benchmark when performance gets worse.

## Note
This slowdown might have contributed to #2671  (tests timing out) and the need for PR #2706.

richardjgowers:
Fixes #2751

 - AtomNames now use a lookup table in their implementation

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?


richardjgowers:
This is a different way of storing atom names in the `Topology` object, where instead of an array of `['N', 'C', etc]` you instead have two arrays, one a set of known atom names, i.e. `['C', 'N', ...]` then the atom names are stored as indices into this array.

This is then beneficial for quickly matching strings (ie names), as you match against a **set** of names to yield integers, then quickly match integers. There's an assumption that the number of unique atom names is very small compared to the number of atoms in the system, ie atom names are repeated often.

I did some timing with this sequence of actions:
```python
import MDAnalysis as mda
from MDAnalysisTests.datafiles import GRO

u = mda.Universe(GRO)

names = u.atoms.names

ag = u.select_atoms('name O*')

ag.names = 'X' 
```

* Universe creation is a bit slower (600 -> 625 ms) for me.
* Attribute access is identical.
* `select_atoms` goes from 69 ms to 0.69 ms
* setting atom names gets slower from 0.6ms to 1.1ms.

I think I can probably make Universe creation faster by rewriting how the Parsers are building the AtomName attribute (ie build the lookup array as I go).  I think atoms are selected more often than they are set, so I'm not too concerned about the performance drop in attribute setting.

Then in a very hand wavy way, I think this will also help serialisation efforts as this is more compact to store/send.

richardjgowers:
So if @MDAnalysis/coredevs could try this on their typical workloads and see if it works for them.

richardjgowers:
@IAlibay as a way of slowing you down from making more PRs, can you review this?  Is it understandable/maintainable and does it show performance improvements on your typical systems?

richardjgowers:
Ok I've made all string topology attrs use this approach.  I think I can improve selections like `protein` to leverage the new stuff too.

IAlibay:
So the code adds a numpy array and a smaller (unless everything is unique) list + dictionary set per attribute. At what point do we need to consider the extra memory cost of Universe creation?

richardjgowers:
The memory cost is likely almost always smaller. Yes there are more python
objects, but each resname is stored as an integer (2 bytes for uint16)
rather than a Python string, likely 3+? I’ll double check these assumptions
though :)

IAlibay:
Update on my million atom benchmark for this new patch.

  - Universe creation has further increased to 12.4s from 10.9s (develop).
  - Selecting waters has decreased to 54.2 ms from 1.98s (develop)
  - water renaming hasn't substantially changed (slight reduction but within error).

It's probably worth noting that I suspect that the TOPParser is particularly inefficient, so it's likely that the difference in universe creation is going to be a lot different if say a GRO file was read.

package/MDAnalysis/core/topologyattrs.py
```diff
@@ -487,10 +487,61 @@ class Atomnames(AtomAttr):
...
+            try:
+                self.nmidx[i] = self.namedict[val]
+            except KeyError:
+                nextidx = len(self.namedict)
+                self.namedict[val] = nextidx
+                name_lookup.append(val)
+
+                self.nmidx[i] = nextidx
+
+        self.name_lookup = np.array(name_lookup, dtype=object)
```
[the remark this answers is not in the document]
richardjgowers: Yeah there's also possibly a way to exploit the guaranteed ordering of the keys that I'm not exploiting here.

richardjgowers:
@IAlibay thanks for benchmarking.  I'm not shocked that this is a little slower to construct, we're constructing an index of possible types, so there's a Python loop over all atom values, for each attribute.  I'll see if I can take the edge off that... Killing a 2s selection time is nice though :)

package/MDAnalysis/core/selection.py
```diff
@@ -518,11 +518,19 @@ def __init__(self, parser, tokens):
...
     def apply(self, group):
-        mask = np.zeros(len(group), dtype=np.bool)
-        for val in self.values:
-            values = getattr(group, self.field)
-            mask |= [fnmatch.fnmatch(x, val) for x in values]
-        return group[mask].unique
+        # rather than work on group.names, cheat and look at the lookup table
+        nmattr = getattr(group.universe._topology, self.field)
+
+        matches = []  # list of passing indices
```
jbarnoud, line 524: The code in this function is duplicated several times. Would there be a way of factoring it out?

package/MDAnalysis/core/topologyattrs.py
```diff
@@ -477,8 +477,65 @@ def _gen_initial_values(na, nr, ns):
...
+        self.values = self.name_lookup[self.nmidx]
+
+    @staticmethod
+    def _gen_initial_values(na, nr, ns):
+        return np.array(['' for _ in range(na)], dtype=object)
+
+    @_check_length
+    def set_atoms(self, ag, values):
```
jbarnoud, line 509: This logic as well is replicated several times.

package/MDAnalysis/core/selection.py
richardjgowers: The difference is here it's `group.ix` as we're getting an atom attribute, later it's `group.resindices`/`group.segindices` as it's resnames/segids.  We could have this as `getattr(group, self.attrlevel)`, where attrlevel = `ix | resindices | segindices`.  This is arguably a little less readable but more compact in terms of repetition.  Not fussed either way, what's easier to read/maintain?

package/MDAnalysis/core/topologyattrs.py
richardjgowers: Same as above

package/MDAnalysis/core/selection.py
jbarnoud: My concern is that if we get to fix or change anything here, the odds are that one of the copies will be left out leading to very confusing discrepancies. I would go for the `getattr` solution here. Though I am not doing the work, so I won't block the PR either way.

richardjgowers: Yeah that's a good point, I'll refactor so it breaks/fixes as one unit :)

package/MDAnalysis/core/selection.py
```diff
@@ -518,11 +518,19 @@ def __init__(self, parser, tokens):
...
+        # iterate through set of known atom names, check which pass
+        for nm, ix in nmattr.namedict.items():
+            if any(fnmatch.fnmatch(nm, val) for val in self.values):
```
IAlibay, line 527: So just out of curiosity I looked into how much `fnmatch` itself was a culprit in the slowdown we see on `develop`.

It looks like we can get a decent speedup by just switching over to regex by doing something along the lines of `fnmatch.translate(val)` combined with `reobj = re.compile` and `reobj .match()`.

Modifying the `develop` code to switch from pure fnmatch to regex gives a ~ 5x speed up, so definitely the approach used here is worth it. However I wonder if we can somehow leverage the improved performance of `re.match()` somewhere here?

I've tried:

```python
     def apply(self, group):
         # rather than work on group.names, cheat and look at the lookup table
         nmattr = getattr(group.universe._topology, self.field)
         vals = [ re.compile(fnmatch.translate(i)) for i in self.values ]

         matches = []  # list of passing indices
         # iterate through set of known atom names, check which pass
         for nm, ix in nmattr.namedict.items():
             if any(bool(val.match(nm)) for val in vals):
                 matches.append(ix)
```

But the performance difference is at best within error. It might be that we could do a better job by making `grab_not_keyword` return regex objects instead?

richardjgowers: iirc `re` already caches things so the speedup from explicitly `re.compile` before use is minimal.

Something I haven't looked at yet is here we're matching against all possible names in the Universe, but really we only need to match all names in the selection scope.  So you could first create a set of `nmidx[group.ix]` and only run fnmatch/re against that subset.

IAlibay: So for context this is what I did on develop to get the speed up (+/- indicating the pre and post regex code):

```python
     @return_empty_on_apply
     def apply(self, group):
         mask = np.zeros(len(group), dtype=np.bool)
+        values = getattr(group, self.field)
         for val in self.values:
-            values = getattr(group, self.field)
-            mask |= [fnmatch.fnmatch(x, val) for x in values]
+            reobj = re.compile(fnmatch.translate(val))
+            mask |= [bool(reobj.match(x)) for x in values]
         return group[mask].unique
```

Pre-regex my million atom benchmark got:
alpha carbon selection: 1.32 s
water oxygen selection: 2.72 s

Post-regex, my million atom benchmark got:
alpha carbon selection: 269 ms
water oxygen selection: 712 ms

It's worth noting that the cost of creating the regex object is probably high, so it's possible that on smaller systems the gains aren't actually real.

orbeckst:
Is the only thing left here the refactoring of the duplicated code? It would be nice to get this regression fixed sooner than later (also for PR #2798 ).

orbeckst:
CHANGELOG, too.

package/MDAnalysis/core/topologyattrs.py
richardjgowers: This isn't possible to fix as the method names are also different (set_atoms / set_residues / set_segments) so I'm leaving this as duplicated

package/MDAnalysis/core/selection.py
```diff
@@ -828,14 +845,20 @@ class ProteinSelection(Selection):
         'CLEU', 'CILE', 'CVAL', 'CASF', 'CASN', 'CGLN', 'CARG', 'CHID', 'CHIE',
         'CHIP', 'CTRP', 'CPHE', 'CTYR', 'CGLU', 'CASP', 'CLYS', 'CPRO', 'CCYS',
         'CCYX', 'CMET', 'CME', 'ASF',
-    ])
+    }
 
     def __init__(self, parser, tokens):
         pass
 
     def apply(self, group):
```
richardjgowers, line 853: `select_atoms('protein')` went from 50ms to 0.5ms on GRO

IAlibay:
Something is messing up the azure pipelines stuff here but it's not immediately obvious to me as to what it is.

package/CHANGELOG
```diff
@@ -39,6 +39,8 @@ Enhancements
   * Added computation of Mean Squared Displacements (#2438, PR #2619)
   * Improved performances when parsing TPR files (PR #2804)
   * Added converter between Cartesian and Bond-Angle-Torsion coordinates (PR #2668)
+  * Improved performance of select_atoms on strings (e.g. name, type, resname) and
```
orbeckst, line 42: I’d also put that under Fixes because it fixes the performance regression. (Having other faster things is a nice side-effect).

orbeckst:
AppVeyor fails on *TestICodes.test_set_residues_singular* – is this related to this PR?

package/MDAnalysis/core/selection.py
```diff
@@ -530,11 +530,23 @@ def __init__(self, parser, tokens):
...
-            values = getattr(group, self.field)
-            mask |= [fnmatch.fnmatch(x, val) for x in values]
-        return group[mask].unique
+        # rather than work on group.names, cheat and look at the lookup table
+        nmattr = getattr(group.universe._topology, self.field)
+
+        matches = []  # list of passing indices
+        # iterate through set of known atom names, check which pass
+        for nm, ix in nmattr.namedict.items():
+            if any(fnmatch.fnmatch(nm, val) for val in self.values):
```
orbeckst, line 539: I think we need to be using [fnmatchcase()](https://docs.python.org/3/library/fnmatch.html#fnmatch.fnmatchcase) so that it doesn't depend on your OS if you match "GLU" with "Glu" on Windows (where [os.path.normcase()](https://docs.python.org/3/library/os.path.html#os.path.normcase) is applied!)  but not on Linux.

richardjgowers: good catch, thanks

richardjgowers:
@IAlibay should be good to go hopefully...

package/MDAnalysis/core/topologyattrs.py
```diff
@@ -473,8 +473,65 @@ def _gen_initial_values(na, nr, ns):
...
+        newnames = []
+
+        # two possibilities, either single value given, or one per Atom
+        if isinstance(values, str):
```
richardjgowers, line 509: I think checking against `str` here is the only 2/3 incompatible thing.  This will need to be `six.string_types` I think..

richardjgowers:
@MDAnalysis/coredevs does someone fancy giving this a once over?

package/MDAnalysis/core/topologyattrs.py
orbeckst: Ok, thanks for the heads-up for PR #2798

richardjgowers:
@MDAnalysis/coredevs please

package/MDAnalysis/core/selection.py
```diff
@@ -907,29 +977,63 @@ class BaseSelection(NucleicSelection):
...
 
 class NucleicSugarSelection(NucleicSelection):
     """Contains all atoms with name C1', C2', C3', C4', O2', O4', O3'.
+
+    .. versionchanged:: 2.0.0
+       sug_atoms changed to set (from numpy array)
+       performance improved by ~100x on larger systems
     """
     token = 'nucleicsugar'
     sug_atoms = np.array(["C1'", "C2'", "C3'", "C4'", "O4'"])
```
jbarnoud, line 1018: Shouldn't this be a set?

package/MDAnalysis/core/topologyattrs.py
```diff
@@ -473,8 +473,65 @@ def _gen_initial_values(na, nr, ns):
...
 
+class _AtomStringAttr(AtomAttr):
+    def __init__(self, vals, guessed=False):
+        self._guessed = guessed
+      
+        self.namedict = dict()  # maps str to nmidx
+        name_lookup = []  # maps idx to str
+        # eg namedict['O'] = 5 & name_lookup[5] = 'O'
+
+        self.nmidx = np.zeros_like(vals, dtype=int)  # the lookup for each atom
```
jbarnoud, line 484: I would not mind a more explicit name than `nmidx`. `name_index` perhaps? But there may be something even better I do not think of.

orbeckst:
ping @IAlibay for approval

ping @richardjgowers to address @jbarnoud 's comments

package/MDAnalysis/core/selection.py
orbeckst: That's what the versionchanged says....

package/MDAnalysis/core/topologyattrs.py
orbeckst: Does`nmidx` need to be public or could it just be hidden behind an underscore?

orbeckst:
@IAlibay @jbarnoud do you see any blocking issues here?

I am keen to see if it looks better in our benchmarks.

jbarnoud:
I am not in a position to review in depth, thought from what I read on my phone it looks good to me. I am uneasy with the amount of code duplication, but it can be a separate pull request.

package/MDAnalysis/core/selection.py
```diff
@@ -802,10 +819,14 @@ class ProteinSelection(Selection):
     See Also
     --------
     :func:`MDAnalysis.lib.util.convert_aa_code`
+
+    .. versionchanged:: 2.0.0
```
IAlibay, line 823: I think you'll need two empty lines here otherwise the versionchanged won't render properly.

package/MDAnalysis/core/selection.py
```diff
@@ -875,29 +911,63 @@ class BackboneSelection(ProteinSelection):
 
     This excludes OT* on C-termini
     (which are included by, eg VMD's backbone selection).
+
```
IAlibay, line 914: As above, two empty lines (unless it somehow renders fine as-is).

package/MDAnalysis/core/selection.py
```diff
@@ -875,29 +911,63 @@ class BackboneSelection(ProteinSelection):
...
+        return group.unique
 
 
 class NucleicBackboneSelection(NucleicSelection):
     """Contains all atoms with name "P", "C5'", C3'", "O3'", "O5'".
 
     These atoms are only recognized if they are in a residue matched
     by the :class:`NucleicSelection`.
+
```
IAlibay, line 946: As above.

package/MDAnalysis/core/selection.py
```diff
@@ -907,29 +977,63 @@ class BaseSelection(NucleicSelection):
 
      'N9', 'N7', 'C8', 'C5', 'C4', 'N3', 'C2', 'N1', 'C6',
      'O6','N2','N6', 'O2','N4','O4','C5M'
+
```
IAlibay, line 980: As above.

package/MDAnalysis/core/selection.py
```diff
@@ -907,29 +977,63 @@ class BaseSelection(NucleicSelection):
...
+                           if nm in self.nucl_res]
+        nmidx = resnames.nmidx[group.resindices]
+        group = group[np.in1d(nmidx, resname_matches)]
+
+        return group.unique
 
 
 class NucleicSugarSelection(NucleicSelection):
     """Contains all atoms with name C1', C2', C3', C4', O2', O4', O3'.
+
```
IAlibay, line 1012: As above.

package/MDAnalysis/core/topologyattrs.py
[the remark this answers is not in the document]
IAlibay: I realise in practice this isn't going to happen over many iterations, but `append` can be quite slow, especially if you're just doing a numpy conversion at the end anyways. If there's any way to avoid it (and make the code look cleaner at the same time), I would encourage it (same thing for the similar appends below).

package/MDAnalysis/core/topologyattrs.py
```diff
@@ -1910,8 +2008,65 @@ def set_segments(self, sg, values):
...
+
+        # two possibilities, either single value given, or one per Atom
+        if isinstance(values, str):
+            try:
+                newidx = self.namedict[values]
+            except KeyError:
+                newidx = len(self.namedict)
+                self.namedict[values] = newidx
+                newnames.append(values)
+        else:
```
IAlibay, line 2051: This part of the code (i.e. the `else` part of the `if` construct) isn't being tested (at least according to codecov). Could we add at least a quick test for this? Since it's core, it's probably a bad idea to leave stuff untested.

package/MDAnalysis/core/topologyattrs.py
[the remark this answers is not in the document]
orbeckst: I don't think that with this code you know in advance what size `name_lookup` will be. And finding out in advance will be relatively costly.

I benchmarked  a typical system
```python
u = mda.Universe(data.GRO_MEMPROT)
%timeit build_name_index_numpy(u.atoms.names)
```
with `len(set(u.atoms.names)) == 219` different atom names. 

Just finding out how many different names there are
```python
%timeit len(set(u.atoms.names))
```
costs sizable 2.3 ms (on my ageing laptop)
```
2.29 ms ± 39.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

How does this compare to what we have?

## Current code

The current code (@richardjgowers 's version)
```python
def build_name_index_append(vals):
        namedict = dict()  # maps str to nmidx
        name_lookup = []  # maps idx to str
        # eg namedict['O'] = 5 & name_lookup[5] = 'O'

        nmidx = np.zeros_like(vals, dtype=int)  # the lookup for each atom
        # eg Atom 5 is 'C', so nmidx[5] = 7, where name_lookup[7] = 'C'

        for i, val in enumerate(vals):
            try:
                nmidx[i] = namedict[val]
            except KeyError:
                nextidx = len(namedict)
                namedict[val] = nextidx
                name_lookup.append(val)

                nmidx[i] = nextidx

        name_lookup = np.array(name_lookup, dtype=object)
        values = name_lookup[nmidx]
        
        return namedict, nmidx, name_lookup, values
```
is already pretty fast
```python
%timeit build_name_index_append(u.atoms.names)
```
gives the speed to beat:
```
9.91 ms ± 664 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

Note how `len(set(vals))` would already be 1/5 of the total run time?

## Replace append with pre-allocated array

Following @IAlibay suggestion and replacing append with a pre-allocated array
```python
def build_name_index_insert(vals):
        namedict = dict()  # maps str to nmidx
        name_lookup = np.empty(len(set(vals)), dtype=object)  # maps idx to str
        # eg namedict['O'] = 5 & name_lookup[5] = 'O'

        nmidx = np.zeros_like(vals, dtype=int)  # the lookup for each atom
        # eg Atom 5 is 'C', so nmidx[5] = 7, where name_lookup[7] = 'C'

        nextidx = 0
        for i, val in enumerate(vals):
            try:
                nmidx[i] = namedict[val]
            except KeyError:
                namedict[val] = nextidx
                name_lookup[nextidx] = val
                nmidx[i] = nextidx
                nextidx += 1

        values = name_lookup[nmidx]
        
        return namedict, nmidx, name_lookup, values
```
gives
```python
%timeit build_name_index_insert(u.atoms.names)
```
```
11.1 ms ± 712 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
which is a tad slower than the version with append.

So I would say unless we expect O(10^3) different string values, the append version is almost certainly ok.

## NumPy 

I tried out a "numpy" version
```python
def build_name_index_numpy(vals):
    name_lookup = np.unique(vals)
    namedict = {name: idx for idx, name in enumerate(name_lookup)}
    nmidx = np.zeros_like(vals, dtype=int)
    
    for name in name_lookup:
        nmidx[vals == name] = namedict[name]
        
    values = name_lookup[nmidx]
    
    return namedict, nmidx, name_lookup, values
```
but that gave an abysmal
```
240 ms ± 34.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
which is 24 times **slower** than the try/except/append version. You can shave off 60 ms by replacing `np.unique` with `name_lookup = np.array(list(set(vals)))` but the >200 repeated assignments `nmidx[vals == name] = namedict[name]` kill performance:
```python
name = "H1"
vals = u.atoms.names
%timeit nmidx[vals == name] = namedict[name]
```
```
752 µs ± 24.8 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
_Maybe_ if there was a way to do all assignments in one go, one would have a hope of doing better.

But even just building the dict itself
```python
%timeit namedict = {name: idx for idx, name in enumerate(name_lookup)}
```
```
19.5 µs ± 776 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```
is slower than the existing code so I think it's going to be tough to do much better – maybe someone with better numpy skills than I can see it.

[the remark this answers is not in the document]
IAlibay: So my suggestion was actually to do this:

```python
def build_name_index_append(vals):
        namedict = dict()  # maps str to nmidx
        # eg namedict['O'] = 5 & name_lookup[5] = 'O'

        nmidx = np.zeros_like(vals, dtype=int)  # the lookup for each atom
        # eg Atom 5 is 'C', so nmidx[5] = 7, where name_lookup[7] = 'C'

        for i, val in enumerate(vals):
            try:
                nmidx[i] = namedict[val]
            except KeyError:
                nextidx = len(namedict)
                namedict[val] = nextidx
                nmidx[i] = nextidx

        name_lookup = np.array([i for i in namedict.keys()], dtype=object)
        values = name_lookup[nmidx]

        return namedict, nmidx, name_lookup, values
```

The idea here was that if you end up having to do lots of irregular append operations in the original construct, you could avoid some of the memory access overheads (plus it looks a bit nicer imho).

However, in practice that idea doesn't hold up, in part since the KeyError is triggered so infrequently, you end up with code that has within error performance (13.5 ms vs 13.6 ms in the above GRO_MEMPROT system).

Even if all 43480 elements in the name array are unique, you only save ~ 2 ms [44 ms -> 42 ms], although this is mainly an academic point as passing an array of unique names is out of scope of this function (i.e. triggering the KeyError probably becomes the biggest cost).

orbeckst:
@richardjgowers could you please have a look at @IAlibay 's comments? Then this should be _finally_ good to go!

package/MDAnalysis/core/topologyattrs.py
richardjgowers: hopefully covered by new test on setting multiple segids

richardjgowers:
re: avoiding the append loop.  Not sure it's possible as we don't know the length beforehand.  Then we definitely want it as a numpy array later as we fancy index it to death.

there's some code duplication, this is because the different versions work on atoms/residues/segments or have different name arrays (nucl_res / sug_atoms).  You could make it all more generic and accessing a variable variable but I think it gets less readable as you try and play code golf with it.

IAlibay (approved):
codecov seems reasonable, lgtm!

package/MDAnalysis/core/topologyattrs.py
orbeckst: included in backport 8a29e83cf33f59f0d0507b26acd8f4d71dbb0dae

PR #2767, opened 2020-06-17
decorate Selection.apply to ignore empty atomgroups

### LINKED ISSUE #2765: Cannot select atoms with StringSelection from empty AtomGroup
## Expected behavior ##

`empty_atomgroup.select_atoms('name anything')` returns an empty AtomGroup.

## Actual behavior ##

A TypeError is raised.

```python

In [2]: from MDAnalysis.tests.datafiles import PSF

In [3]: mda.__version__
Out[3]: '2.0.0-dev0'

In [4]: u = mda.Universe(PSF)

In [5]: ag = u.select_atoms('resname blahblah')

In [6]: len(ag)
Out[6]: 0

In [7]: ag.select_atoms('name O')
<ipython-input-7-c43a12bdaf29> in <module>
----> 1 ag.select_atoms('name O')

~/pydev/mdanalysis/package/MDAnalysis/core/groups.py in select_atoms(self, sel, *othersel, **selgroups)
   2861             # Apply the first selection and sum to it
   2862             atomgrp = sum([sel.apply(self) for sel in selections[1:]],
-> 2863                           selections[0].apply(self))
   2864         return atomgrp
   2865

~/pydev/mdanalysis/package/MDAnalysis/core/selection.py in apply(self, group)
    517         for val in self.values:
    518             values = getattr(group, self.field)
--> 519             mask |= [fnmatch.fnmatch(x, val) for x in values]
    520         return group[mask].unique
    521

TypeError: ufunc 'bitwise_or' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) 2.0.0-dev0

lilyminium:
Fixes #2765 

 - if an empty AtomGroup is passed to Selection.apply, just return it

I decorated `Selection.apply` in `_Selectionmeta.__init__`, which I baselessly feel is a bit iffy. Any opinions on if I should move it somewhere else?

Edit: alternately it could be expensive to have so many ifs, I could just do the `StringSelection` and sphzone/layer ones individually

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/core/selection.py
+++ b/package/MDAnalysis/core/selection.py
@@ -156,6 +156,16 @@ class OrOperation(LogicOperation):
 
         return group.universe.atoms[idx]
 
+def return_empty_on_apply(func):
+    """
+    Decorator to return empty AtomGroups from the apply() function
+    without evaluating it
+    """
+    def apply(self, group):
+        if len(group) == 0:
+            return group
+        return func(self, group)
+    return apply
 
 class _Selectionmeta(type):
     def __init__(cls, name, bases, classdict):
@@ -163,6 +173,11 @@ class _Selectionmeta(type):
         try:
             _SELECTIONDICT[classdict['token']] = cls
         except KeyError:
+            pass
+        
+        try:
+            cls.apply = return_empty_on_apply(cls.apply)
+        except AttributeError:
             pass
 
 
--- a/package/MDAnalysis/core/selection.py
+++ b/package/MDAnalysis/core/selection.py
@@ -176,13 +176,6 @@ class _Selectionmeta(type):
         except KeyError:
             pass
 
-        # GlobalSelections can return something even if empty AG
-        if not cls.__name__ == 'GlobalSelection':
-            try:
-                cls.apply = return_empty_on_apply(cls.apply)
-            except AttributeError:
-                pass
-
 
 class Selection(object, metaclass=_Selectionmeta):
     pass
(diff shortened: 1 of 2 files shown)


package/MDAnalysis/core/selection.py
```diff
@@ -156,6 +156,16 @@ def apply(self, group):
 
         return group.universe.atoms[idx]
 
+def return_empty_on_apply(func):
```
richardjgowers, line 159: Iirc this kills docstrings so there's a `functools.decorate` or something to preserve docstrings after wrapping

richardjgowers:
I like the fix but it's a little hacky to put the decorator application in the `_SelectionMeta` - really the Selection going through shouldn't be modified, so maybe it's better to explicitly include the `@emptyAGguard` decorator onto each `apply` method

lilyminium:
I added the decorator to the apply methods the geometric and string selection classes instead.

orbeckst:
This PR did not include a CHANGELOG comment. 

Further discussion in https://github.com/MDAnalysis/mdanalysis/issues/2765#issuecomment-654519327

lilyminium:
Yes, sorry, we forgot; the CHANGELOG entry was updated in the next merge to develop (by @IAlibay iirc)

PR #2770, opened 2020-06-18
Make dependencies files

### LINKED ISSUE #2769: Have dependencies files

It's easier to update one `requirements.txt` and `environment.yml` file than to update all of `.travis.yml`, `.appveyor.yml`, and `azure-pipelines.yml` (potentially), and the user guide, every time we add a new package.

Have a `requirements.txt` and `environment.yml` file that we point both CI and people to.

Status quo

lilyminium:
Fixes #2769 

 - Added full and minimal `requirements.txt` and `environment.yml` files to maintainer
 - Moved CI to use them

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


maintainer/envs/minimal_requirements.txt
```diff
@@ -0,0 +1,18 @@
+biopython
+codecov
+cython
+duecredit
```
tylerjereddy, line 4: I'm pretty sure you're changing the Azure run here--for example, I was not using `duecredit` on purpose to avoid the annoying spam at the end of the test run.

I don't have specific views on the `conda` installs, but for `pip` I'm a bit more concerned here.

More broadly, I'm also a little concerned that the abstractions may not really be worth it unless they are for standardized things like [PEP 518](https://www.python.org/dev/peps/pep-0518/), which is indeed precisely for "Specifying Minimum Build System Requirements for Python Projects." For that, `pyproject.toml` should be used.

tylerjereddy: There's also "bare minimum to build," "bare minimum to test," and what if I want to use a slightly different dependency pin for a while on 32-bit Windows because of some issue? The abstraction to a file becomes a burden then if another CI service/run is using it.

The abstraction vs. convenience to just edit in the CI yml itself is something that should probably at least be discussed.

Not saying we shouldn't do this, but I do have my reservations about some aspects.

azure-pipelines.yml
```diff
@@ -36,15 +36,7 @@ jobs:
...
-  - script: >-
-      python -m pip install
-      cython
-      matplotlib
-      numpy
-      pytest
-      pytest-xdist
-      scikit-learn
-      scipy
+  - script: python -m pip install -r maintainer\envs\minimal_requirements.txt
```
tylerjereddy, line 39: This is also only the replacement of a single use case. If there were at least 3 places where the replacement was performed for `pip`, and where it did not change the meaning whatsoever, I might be in favor, but both the meaning (the set of deps installed) and the duplication requirement are not met for `pip`  at the moment.

tylerjereddy: If you want to tell users to use that file, I think it is fine, but I'm currently at least -0.5 on using it in CI, at least until our CI grows further to use this more commonly, etc.

maintainer/envs/environment.yml
```diff
@@ -0,0 +1,38 @@
+
+name: mdanalysis
+channels:
+  - biobuilds
```
richardjgowers, line 4: what's biobuilds for?

lilyminium: clustalw, I think

azure-pipelines.yml
lilyminium: Yes, I agree that `minimal_requirements.txt` is probably unnecessary. I wanted one of the CI suites to use `requirements.txt` and `environment.yml` to make sure they stayed up to date, and then for Travis it seemed easier to have a `minimal_environment.yml` that users could use than to write out packages again, so the `minimal_requirements.txt` was analogous. I'll remove it.

maintainer/envs/environment.yml
orbeckst: yes

lilyminium:
Closed as we have changed CI methods and added an `environment.yml` for ReadTheDocs builds in #3060 and #3071.

Issue #1901, opened 2018-05-16
Alias tempfactor & bfactors


richardjgowers:
These are currently separate attributes, but are in fact the same thing.  Because both names are widely used maybe it makes more sense to alias them, so a single `TopologyAttr` gives you both access points...

orbeckst:
There's also an issue that MMTFReader used bfactors and PDBReader/Writer uses tempfactors. This leads to the annoying situation that writing a PDB from MMTF looses the tempfactors:
```python
import MDAnalysis as mda
u = mda.fetch_mmtf('1AKE')
u.select_atoms("segid A").write("1AKE_A.pdb")
```
gives
```
~/anaconda3/lib/python3.5/site-packages/MDAnalysis/coordinates/PDB.py:906: UserWarning: Found no information for attr: 'tempfactors' Using default value of '0.0'
  "".format(attrname, default))
```

orbeckst:
Workaround for the MMTF/PDB case:
```python
import MDAnalysis as mda
u = mda.fetch_mmtf('1AKE')
u.add_TopologyAttr('tempfactors', u.atoms.bfactors)
```

IAlibay:
@lilyminium and I are keen to see this get fixed for 2.0, if anything it breaks colour-by-bfactor visualisation in NGLView.

Do we need to warn users in 1.0.2? (I don't think so, but maybe I'm missing something).

lilyminium:
To repeat what I said on Discord, IMO we should in 2.0 but maybe not 1.0.2. My arguments is that users *may* have assigned different information to bfactors and tempfactors in existing code. The most likely scenario to select atoms with custom tags (e.g. "bfactor 2" vs "tempfactor 2")

orbeckst:
Note that 1.1.0 promised 1eede18 :

> In version 2.0, MDAnalysis will stop treatiing  tempfactors and bfactors as separate attributes. Instead, they will be aliases of the same attribute.

If we decide to _not_ honor this promise we would technically have to do 1.2.0 with a deprecation note. (We can decide that we're breaking semver a little bit here — but at least we should be clear about what we're doing.)

IAlibay:
> Note that 1.1.0 promised [1eede18](https://github.com/MDAnalysis/mdanalysis/commit/1eede18101cedf2be9ba686cc41429719d8bb3da) :
> 
[...]

PR #2775, opened 2020-06-19
Simple RDKitConverter

cbouy:
Part of the fixes for #2468 

 - Added a converter that works for simple cases

Inspired from the ParmEdConverter

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/coordinates/RDKit.py
+++ b/package/MDAnalysis/coordinates/RDKit.py
@@ -59,7 +59,25 @@ import warnings
 import numpy as np
 
 from . import memory
+from . import base
 
+try:
+    from rdkit import Chem
+except ImportError:
+    pass
+else:
+    RDBONDTYPE = {
+        'AROMATIC': Chem.BondType.AROMATIC,
+        'SINGLE': Chem.BondType.SINGLE,
+        'DOUBLE': Chem.BondType.DOUBLE,
+        'TRIPLE': Chem.BondType.TRIPLE,
+    }
+    RDBONDORDER = {
+        1: Chem.BondType.SINGLE,
+        1.5: Chem.BondType.AROMATIC,
+        2: Chem.BondType.DOUBLE,
+        3: Chem.BondType.TRIPLE,
+    }
 
 class RDKitReader(memory.MemoryReader):
     """Coordinate reader for RDKit.
@@ -102,3 +120,62 @@ class RDKitReader(memory.MemoryReader):
             coordinates = np.empty((1,n_atoms,3), dtype=np.float32)
             coordinates[:] = np.nan
         super(RDKitReader, self).__init__(coordinates, order='fac', **kwargs)
+
+
+class RDKitConverter(base.ConverterBase):
+    """Convert MDAnalysis AtomGroup or Universe to `RDKit <https://www.rdkit.org/docs/source/rdkit.Chem.rdchem.html#rdkit.Chem.rdchem.Mol>`_ :class:`rdkit.Chem.rdchem.Mol`.
+
+    Example
+    -------
+
+    .. code-block:: python
+
+        import MDAnalysis as mda
+        from MDAnalysis.tests.datafiles import PDB_full
+        u = mda.Universe(PDB_full)
+        mol = u.select_atoms('resname DMS').convert_to('RDKIT')
+    
+
+    .. versionadded:: 2.X.X
+    """
+
+    lib = 'RDKIT'
+    units = {'time': None, 'length': 'Angstrom'}
+
+    def convert(self, obj):
+        """Write selection at current trajectory frame to :class:`~rdkit.Chem.rdchem.Mol`.
+
+        Parameters
+        -----------
+        obj : AtomGroup or Universe or :class:`Timestep`
+        """
+        try:
+            from rdkit import Chem
+        except ImportError:
+            raise ImportError('RDKit is required for RDKitConverter but '
+                              'is not installed. Try installing it with \n'
+                              'conda install -c conda-forge rdkit')
+        try:
+            # make sure to use atoms (Issue 46)
+            ag_or_ts = obj.atoms
+        except AttributeError as e:
+            if isinstance(obj, base.Timestep):
+                ag_or_ts = obj.copy()
+            else:
+                raise TypeError("No Timestep found in obj argument") from e
+
+        mol = Chem.RWMol()
+        atom_mapper = {}
+        for atom in ag_or_ts:
+            rdatom = Chem.Atom(atom.element)
+            index = mol.AddAtom(rdatom)
+            atom_mapper[atom.ix] = index
+
+        for bond in ag_or_ts.bonds:
+            bond_indices = [atom_mapper[i] for i in bond.indices]
+            bond_type = RDBONDTYPE.get(bond.type.upper(), RDBONDORDER.get(
+                bond.order, Chem.BondType.SINGLE))
+            mol.AddBond(*bond_indices, bond_type)
+
+        Chem.SanitizeMol(mol)
+        return mol


IAlibay:
@cbouy since this seemd like WIP, I've been holding off on reviewing, do ping us when you'd like a review.

cbouy:
@IAlibay @richardjgowers @fiona-naughton 

It's starting to look good!
image

It's still missing quite some tests but I guess you can start having a look now

IAlibay:
> It's starting to look good!

Oooo that's some really cool stuff! Do make sure to include this (or a similar image) in your next blog :)

RMeli:
This looks really cool! =)

Would it be easier to call the conversion method `rdkit()` or `to_rdkit()`? The reason I'm asking is that one could use code autocompletion to quickly see the conversion methods available, while `convert_to()` requires to look at the documentation (which might get quite hairy if multiple conversions are supported?). It's also less typing. ;)

This is just a quick thought I had looking at your code snippet, please feel free to ignore!

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -101,4 +135,113 @@ def __init__(self, filename, **kwargs):
...
+            rdatom.SetMonomerInfo(mi)
+            # TODO other properties (charges)
+            index = mol.AddAtom(rdatom)
+            # map index in universe to index in mol
+            atom_mapper[atom.ix] = index
+
+        try:
+            bonds = ag.bonds
+        except NoDataError:
+            ag.guess_bonds()
```
IAlibay, line 232: To be discussed, bond guessing isn't one of those I think are really inaccurate, so I'm not as opposed to guessing by default, but do we think it might be a good idea to throw a warning here?

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -101,4 +135,113 @@ def __init__(self, filename, **kwargs):
...
+                element = atom.element
+            except NoDataError:
+                # guess atom element
+                # capitalize: transform CL to Cl and so on
+                element = guess_atom_element(atom.name).capitalize()
+            rdatom = Chem.Atom(element)
+            # add properties
+            mi = Chem.AtomPDBResidueInfo()
+            for attr, rdattr in RDATTRIBUTES.items():
+                try: # get value in MDA atom
```
IAlibay, line 196: Here until 222 is doing quite a lot of things that probably would be worth documenting thoroughly (even if docstring in a private function, at least future developers would know exactly what is going on). How would we feel about moving this code into a private function that get's called within the loop? If anything it might make things more readable (although that could just be me and my tiny laptop screen).

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -101,4 +135,113 @@ def __init__(self, filename, **kwargs):
...
+                    value = getattr(atom, attr)
+                except AttributeError:
+                    pass
+                else:
+                    if isinstance(value, np.generic):
+                        # convert numpy types to python standard types
+                        value = value.item()
+                    if attr == "segid":
```
IAlibay, line 204: Could we use segindices here instead?

richardjgowers:
The cis/trans is rotating, I’m guessing this is because we’re not labelling
the bond? Iirc rdkit labels stereo around bonds as a group of atoms and a
stereo type right?  This will probably be too tricky to quickly solve but
is worth raising warnings and an issue on the tracker about

Looks awesome nonetheless!

cbouy:
> Would it be easier to call the conversion method rdkit() or to_rdkit()? The reason I'm asking is that one could use code autocompletion to quickly see the conversion methods available, while convert_to() requires to look at the documentation (which might get quite hairy if multiple conversions are supported?). It's also less typing. ;)

I decided to follow what was already there for the parmed converter, but yes I completely agree that a `to_rdkit` method would be very nice, although I guess it should be in another PR so that we can add the `to_parmed` as well. 
I prefer the pandas style `to_rdkit` so that you can quickly see which converters are available by autocompleting after typing `to_`.

cbouy:
> The cis/trans is rotating, I’m guessing this is because we’re not labelling the bond?

Yes, although atom and bond stereochemistry can be inferred from the coordinates with `Chem.AssignStereochemistryFrom3D` and `Chem.DetectBondStereoChemistry`. But I'm not reading the coordinates yet, there's still a lot to do on the topology conversion first.

RMeli:
> I decided to follow what was already there for the `parmed` converter, but yes I completely agree that a `to_rdkit` method would be very nice, although I guess it should be in another PR so that we can add the `to_parmed` as well.

I wasn't aware of the syntax of the `parmed` converter. +1 for a separate PR changing both and I also prefer the `pandas` style for the reason you mentioned.

IAlibay:
> I wasn't aware of the syntax for the `parmed` converter. +1 for a separate PR changing both and I also prefer the `pandas` style for the reason you mentioned.

pinging @lilyminium who will know better here, but I don't think there's a specific convention yet. Unless we have multiple converters already using the same naming style, I can't see a reason why we can't try to agree on a good naming scheme now.

@cbouy it might be a good idea to bring this up an issue on this. As we have more and more "converters" it would be good to have a unified API.

lilyminium:
I like the current model of the metaclass registering the format. We could do a similar thing to TopologyAttrs though, and get the metaclass to add methods to AtomGroups. e.g. 

```python
def __init__(cls, name, bases, classdict):
    method_name = f'to_{cls.lib.lower()}'
    setattr(AtomGroup, method_name, cls.convert)
    setattr(Universe, method_name, cls.convert)
```

The to_lib() method would show up in dir() etc and I think a fix to docstrings in sphinx is in the works.

This would require ParmEdWriter to change too, though, so it's probably best as a separate issue @cbouy

cbouy:
Currently I'm automatically guessing the `element` attribute as well as bonds if they are not present, which is not necessarily a good idea, even if we are warning the user.
However making the converter fail on such cases would require them to add the missing attributes themselves, but they might not be too familiar with the topology API (although it's very well documented).
Would it be acceptable to add `guess_elements=False` and `guess_bonds=False` to the `convert` method, and pass these through `**kwargs` in the `convert_to` method of the AtomGroup ? So users have a simple way of doing it ?

IAlibay:
> Currently I'm automatically guessing the `element` attribute as well as bonds if they are not present, which is not necessarily a good idea, even if we are warning the user.
> However making the converter fail on such cases would require them to add the missing attributes themselves, but they might not be too familiar with the topology API (although it's very well documented).
> Would it be acceptable to add `guess_elements=False` and `guess_bonds=False` to the `convert` method, and pass these through `**kwargs` in the `convert_to` method of the AtomGroup ? So users have a simple way of doing it ?

So my worry here is that there are many ways in which elements can be guessed and hopefully in the future we'll want to provide guessers from things like mass, etc... instead of just atom names. Accounting for all of this should be outside the scope of the method in my opinion. Indeed, not being aware of the API, and especially the assumptions our guessers make, might lead to a lot more unnoticed user errors.

For elements, I think it might be a better call to instead give an error message that points users to the right part of the documentation / user guide (I recently wrote an addition to the guessers docstring to show how one could guess elements [and the dangers of doing so], I could probably port that to the user guide).

Bonds are a different issue in my opinion, guessing should be rather consistent, but we should warn users that we are doing a guess.

@fiona-naughton, @richardjgowers might have different opinions here.

I'm also going to ping @jbarnoud and @lilyminium who have been very involved in the element guessing discussions.

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -95,10 +132,164 @@ def __init__(self, filename, **kwargs):
...
+
+        try:
+            elements = ag.elements
+        except NoDataError:
+            raise AttributeError(
+                "The `elements` attribute is required for the RDKitConverter "
+                "but is not present in this AtomGroup. Please refer to the "
+                "documentation to guess elements from other attributes. "
+                "If `types` are present in the AtomGroup, a good starting "
+                "point would be:\n"
```
IAlibay, line 193: Rather than having code in an error message, I'd recommend just pointing users to the guessers, maybe a help(MDAnalysis.topology.guessers)?

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -95,10 +134,163 @@ def __init__(self, filename, **kwargs):
...
+    value : object, np.int or np.float
+        Attribute value as found in the AtomGroup
+    mi : rdkit.Chem.rdchem.AtomPDBResidueInfo
+        MonomerInfo object that will store the relevant atom attributes
+    """
+    if isinstance(value, np.generic):
+        # convert numpy types to python standard types
+        value = value.item()
+    if attr == "names":
+        # RDKit needs the name to be properly formated for a
```
richardjgowers, line 286: @jbarnoud didn't you already write code to this effect in the PDBWriter? Or was that for parsing still?+

testsuite/MDAnalysisTests/coordinates/test_rdkit.py
```diff
@@ -24,24 +24,44 @@
...
+        return Chem.MolFromMol2File(mol2_molecule, removeHs=False)
+
+    def smiles_mol():
+        mol = Chem.MolFromSmiles("CN1C=NC2=C1C(=O)N(C(=O)N2C)C")
+        mol = Chem.AddHs(mol)
+        cids = AllChem.EmbedMultipleConfs(mol, numConfs=3)
+        return mol
+
+
+requires_rdkit = pytest.mark.skipif(rdkit_installed == False,
```
IAlibay, line 60: We recently added "import_not_available" to the test utils, it could probably replace this (see test_msd for some of the usage).

testsuite/MDAnalysisTests/coordinates/test_rdkit.py
```diff
@@ -65,14 +85,119 @@ def test_no_coordinates(self):
...
+    def test_other_attributes(self, mol2, idx):
+        mol = mol2.atoms.convert_to("RDKIT")
+        rdprops = mol.GetAtomWithIdx(idx).GetPropsAsDict()
+        for prop in ["charge", "segid", "type"]:
+            rdprop = rdprops["_MDAnalysis_%s" % prop]
+            mdaprop = getattr(mol2.atoms[idx], prop)
+            assert rdprop == mdaprop
+
+
+@pytest.mark.skipif(rdkit_installed == True, reason="test minimal dependency")
```
IAlibay, line 197: One other possibility is to move this up and use block_import to allow to fake a minimal dependency.

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -95,10 +134,164 @@ def __init__(self, filename, **kwargs):
...
+                "on atoms coordinates")
+            ag.guess_bonds()
+            bonds = ag.bonds
+
+        for bond in bonds:
+            try:
+                bond_indices = [atom_mapper[i] for i in bond.indices]
+            except KeyError:
+                # one of the atoms of the bond is not part of the atomgroup
+                continue
```
cbouy, line 253: Not sure if this is an issue or not: 
The KeyError is raised when one of the atom in bond.indices wasn't seen when converting MDA atoms to RDKit, and happens when selecting a fragment of a molecule (`u.select_atoms("resid 2").convert_to("RDKIT")` for example). 
Which means that this selection (resid2) will contain bonds between atoms of resid2, but also the bond between resid1 and the Nterm of resid2, and the bond between the Cterm of resid2 and resid3, although resid1 and resid3 are not part of the selection. 
Is this intended ? Or should the selection only return bonds where both atoms of the bond belong to the selection ?

cbouy:
@IAlibay @richardjgowers @fiona-naughton should I keep this PR "short" or do I add the inferring of charges and bond orders here as well ? I've already created a draft in #2799 but I can move it here if it's preferable. 
Also I'm still not reading charges, should I do that first ? 
Finally, do you see something missing apart from updating the changelog ?

package/MDAnalysis/coordinates/RDKit.py
cbouy: moved this to #2821

cbouy:
So I investigated this `OverflowError` which happens with `numpy==1.13.3` a bit. Locally, this requirement on numpy installed rdkit `2019.03.2` instead of the more recent `2020.03.2`. 
The problem occurs when using the `rdatom.SetIntProp` method with a large int (anything bigger than 2,147,483,647, which is the max value for an int32). 
This happens frequently when using the `Chem.MolFromSequence` which I added in the tests recently. This method will randomly assign a large int to the segment number in v2019 of rdkit but not in v2020 where it just assigns 0. When converting the MolFromSequence to a Universe, and then back to an rdkit molecule, since the `rdatom.SetIntProp` method only accepts int32 it will sometimes throw the error. 
Note that during the Mol-to-Universe-to-Mol conversion the SegmentNumber in the resulting mol is the equivalent of the segindex in the universe (and not the segid as one would expect). This is done because rdkit needs SegmentNumber to be an int32.
I've switched the test to use one of the existing testfiles instead of MolFromSequence for now.

Should I add something in the converter to make sure that people don't use too large segids ? Or maybe a warning in case of OverflowError that this is likely due to large segids (I don't see it happening for other numeric properties like bfactors or charges) ?

package/CHANGELOG
```diff
@@ -36,6 +36,8 @@ Enhancements
   * Added computation of Mean Squared Displacements (#2438, PR #2619)
   * Improved performances when parsing TPR files (PR #2804)
   * Added converter between Cartesian and Bond-Angle-Torsion coordinates (PR #2668)
+  * Added a simple version of the RDKitConverter that handles uncharged 
+    molecules with bond orders/bond types assigned (#2468, PR #2775)
```
IAlibay, line 40: I think the convention we've taken is "Issue #<number>", it's pedantic but probably worth adding for consistency.

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -95,10 +135,163 @@ def __init__(self, filename, **kwargs):
...
+        try:
+            elements = ag.elements
+        except NoDataError:
+            raise AttributeError(
+                "The `elements` attribute is required for the RDKitConverter "
+                "but is not present in this AtomGroup. Please refer to the "
+                "documentation to guess elements from other attributes or "
+                "type `help(mda.topology.guessers)`") from None
+
+        # attributes accepted in PDBResidueInfo object
```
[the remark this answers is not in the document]
IAlibay: (I'm thinking something like the table we have in the TOPParser docs, but that indicates what is read to MDA and from MDA?)

cbouy:
@IAlibay @richardjgowers @fiona-naughton
* Should I continue with the bond order guessing on this PR or on a separate one ? Same question for coordinates
* Should I check for too large segment indices (see my previous comment on numpy 1.13.3) ?

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -95,10 +137,165 @@ def __init__(self, filename, **kwargs):
...
+                    rdatom.SetProp("_MDAnalysis_%s" % attr, value)
+            rdatom.SetIntProp("_MDAnalysis_index", int(atom.ix))
+            # add atom
+            index = mol.AddAtom(rdatom)
+            # map index in universe to index in mol
+            atom_mapper[atom.ix] = index
+
+        try:
+            bonds = ag.bonds
+            if (len(bonds) == 0) and (ag.n_atoms > 1):
```
IAlibay, line 243: I think you could simplify this with a call to `ag.bonds.values()` that should raise an IndexError if you have no bonds in your system.

testsuite/MDAnalysisTests/coordinates/test_rdkit.py
```diff
@@ -65,14 +92,146 @@ def test_no_coordinates(self):
...
+
+    def test_identical_topology(self):
+        rdmol = smiles_mol()
+        u = mda.Universe(rdmol)
+        umol = u.atoms.convert_to("RDKIT")
+        assert rdmol.HasSubstructMatch(umol) and umol.HasSubstructMatch(rdmol)
+
+    def test_raise_requires_elements(self):
+        u = mda.Universe(mol2_molecule)
+        with pytest.raises(AttributeError) as e:
```
IAlibay, line 181: mostly a comment for next time, you can use the `match` parameter of pytest.raises here to save yourself a little bit of hassle.

testsuite/MDAnalysisTests/coordinates/test_rdkit.py
```diff
@@ -65,14 +92,146 @@ def test_no_coordinates(self):
...
+            u.atoms.convert_to("RDKIT")
+            assert ("`elements` attribute is required for the RDKitConverter"
+                    in str(e.value))
+
+    def test_warn_guess_bonds(self, pdb):
+        pdb.delete_bonds(pdb.bonds)
+        ag = pdb.select_atoms("resnum 101 and segid A")
+        pdb.delete_bonds(ag.bonds)
+        with warnings.catch_warnings(record=True) as w:
```
IAlibay, line 190: As above, I think you can also save yourself a little bit of code complexity by just using pytest.warns here.

IAlibay:
> * Should I continue with the bond order guessing on this PR or on a separate one ? Same question for coordinates

I think it depends on what it means to not have bond order guessing at this point. I.e. is there any case where this leads to the right answer? If not, then we should probably have it in this PR (@richardjgowers or @fiona-naughton may have a different opinion here).

> * Should I check for too large segment indices (see my previous comment on numpy 1.13.3) ?

Do we have any idea of what is the minimum version of numpy required for rdkit 2020? 1.13.3 is now three years old and we no longer support py2.7 & 3.5. It might be worth considering bumping up the minimum version.

cbouy:
Without charges and bond order guessing, you will get a valid molecule if the input universe comes from an uncharged molecule (not very common for a biomolecule) and have either only single bonds (even less common), or the universe has a usable bond types or bond orders attribute (which I think only happens with MOL2 files ?)
For an uncharged molecule without bond orders you get something like this, which might be usable:
image

As a side note, does `u.atoms.bonds.types()` actually returns the bond types ("single"...etc.) in some cases ? I've mostly seen it as a tuple for the atomtypes, but the [doc](https://www.mdanalysis.org/docs/documentation_pages/core/topologyobjects.html#MDAnalysis.core.topologyobjects.Bond) says it's for the "type of the bond" so I'm a bit confused about this one

jbarnoud:
> 1.13.3 is now three years old and we no longer support py2.7 & 3.5. It might be worth considering bumping up the minimum version.

@orbeckst started a discussion about that on the dev mailing list: https://groups.google.com/forum/#!topic/mdnalysis-devel/vCDDPH_2iTo

If the proposal goes through, we would bump the minimum version to 1.15 or even 1.16 assuming it waits for the end of the month.

IAlibay:
> Without charges and bond order guessing, you will get a valid molecule if the input universe comes from an uncharged molecule (not very common for a biomolecule) and have either only single bonds (even less common), or the universe has a usable bond types or bond orders attribute (which I think only happens with MOL2 files ?)
> For an uncharged molecule without bond orders you get something like this, which might be usable:

So at least to me it's hard to justify merging something that doesn't offer reasonable behaviour in most cases. If the extra code necessary to get to "works in most cases" is essentially just what is your draft PR, I say let's just add it here. 

It probably doesn't need to cover every single input file type, but we can get to a point where we can put in checks for the file types we expect will fail, then I'd say that's reasonable to merge.

> 
> As a side note, does `u.atoms.bonds.types()` actually returns the bond types ("single"...etc.) in some cases ? I've mostly seen it as a tuple for the atomtypes, but the [doc](https://www.mdanalysis.org/docs/documentation_pages/core/topologyobjects.html#MDAnalysis.core.topologyobjects.Bond) says it's for the "type of the bond" so I'm a bit confused about this one

I _think_ that u.bond.types() actually returns "force field bond types" not so much "bond order" (see: https://www.mdanalysis.org/docs/documentation_pages/topology/TPRParser.html).

You can see this behaviour with TPR files if you do:

```python
import MDAnalysis as mda

u = mda.Universe(TPR, XTC)

u.atoms.bonds.types()
```

> For numpy, I'm not sure if it's RDKit 2020 that directly requires it or if it's one of the dependencies, I'll have to check

Thanks for looking into it.

cbouy:
`numpy==1.15.0` is compatible with rdkit 2020.03.2

IAlibay:
> `numpy==1.15.0` is compatible with rdkit 2020.03.2

Awesome, I've pinged the list discussion @jbarnoud mentioned, let's see what others think.

cbouy:
> I think that u.bond.types() actually returns "force field bond types" not so much "bond order"

Ah okay, I should probably not try using it then

PS: you've made a typo in your message on the mailing list, it's 1.15 not 1.5 😅

IAlibay:
> PS: you've made a typo in your message on the mailing list, it's 1.15 not 1.5 😅

😅 oops I think I've been making a lot of typos today... Based on Oliver's latest [message](https://groups.google.com/d/msg/mdnalysis-devel/vCDDPH_2iTo/3E6hD_26BgAJ) do you want to put up an issue to raise the numpy version to at least 1.15?

package/MDAnalysis/coordinates/RDKit.py
cbouy: I ended up not using bonds.values() as it needs coordinates to work, but people might just want to convert their topology to rdkit without needing the trajectory.

IAlibay: That's reasonable.

I'm trying to work out if this does the intended behaviour though. My understanding is that calling `AtomGroup.bonds` should raise a NoDataError if no bond information is present. If bond information is present but the number of bonds is 0, that _should_ be a valid input, at least according to whatever the user provided right?

Is there a particular case where the bonds attribute get populated but we still require guessing?

cbouy: The only case I could find where the bonds attribute was present but empty is a PDB file without a CONECT record, I don't know if that's intended or not (the PDB_helix test file for example).
The corresponding AtomGroup probably still requires guessing bonds though, as I don't really see a scenario where people need to convert their simulation of monoatomic ions or noble gases to RDKit.
For now I can just try if ag.bonds is present if you prefer, PDB files are a mess anyway !

IAlibay: That's probably not normal, we should have a consistent API, if bonds aren't present then the bonds attribute doesn't get assigned :/ Can you raise this as an issue too? It might be that there's a rationale for it, but I can't seem to understand what it could be.

cbouy: raised #2832

cbouy:
A few examples that work:
* From a PRM7, charges are correctly assigned
image
* From a PSF, when selecting a single residue, charges are correctly assigned and radical are shown where the peptide was cut
image

Some with problems (not sure if we deal with them in this PR or later):
* GRO file that shows 2 problems:  
  * the arginine is not correctly charged,
  * residues that were cut by the selection are charged instead of showing radicals, this happens when we have to guess bonds on a selection
image
* same for this PSF file where bonds have to be guessed for the selection, we get a -C(=O) with a minus charge and a -NH with a minus charge
image

Some other groups that aren't working correctly (first image is the input, second image is after guessing):
* charges instead of double bonds
image
* charges and double bonds on the wrong atoms
image

IAlibay:
Thanks for the thorough look into this @cbouy 

Is there a way we can categorise when it works and when it doesnt? From the looks of things, if bonds & charges are already present it'll work, is this a correct intepretation?

cbouy:
Formal charges are never present in mdanalysis if I understood correctly. 
If bond types are already present then yes it should work without any problem. 
Same if the bonds attribute is present (except if there's an arginine and a few other cases).
If bonds were not present and had to be guessed, it will only work if the bonds were guessed on the whole universe. 
If bonds were guessed only for a selection of atoms, the corresponding Nterm and Cterm will be charged.

If the input doesn't have all hydrogen atoms, it will do some ugly stuff (even if the correct bond types are given, but I can probably check if the bond order attribute is present to fix that):
```python
u = mda.Universe.from_smiles("CN1C=NC2=C1C(=O)N(C(=O)N2C)C", addHs=False, generate_coordinates=False)
u.atoms.convert_to("RDKIT")
```
image

cbouy:
Some of these examples (arginine, sulfates, Nterm and Cterm, and maybe the keto-enolate tautomer) should be fixable by using SMARTS patterns.

cbouy:
It's working 🎉
image

I still need to add a few tests and update the changelog and some docs

IAlibay:
> It's working 🎉

Congrats :)

Does this fix all the issues with when there's no bonds or just part of them? I forgot to reply on Friday, but worst case scenario we just tell users that they need to guess bonds for the entire Universe.

Are some of the above cases still failing? (sorry I'm probably asking very obvious questions it's just a tad bit hard to keep track of all the things you did :), if there's still "stuff to do" it might be a good call to have a checklist somewhere we can look at to keep track of things).

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -95,10 +132,353 @@ def __init__(self, filename, **kwargs):
...
+            rdatom.SetMonomerInfo(mi)
+            # other properties
+            for attr in other_attrs.keys():
+                value = other_attrs[attr][i]
+                attr = _TOPOLOGY_ATTRS[attr].singular
+                _set_atom_property(rdatom, attr, value)
+            _set_atom_property(rdatom, "index", int(atom.ix))
+            # add atom
+            index = mol.AddAtom(rdatom)
+            # map index in universe to index in mol
```
IAlibay, line 271: might be good to move this comment from here to above line 251. That way it's more obvious as to what the dictionary is when you construct it?

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -95,10 +132,353 @@ def __init__(self, filename, **kwargs):
...
+        # NUE is negative, it means we can only add a positive charge to
+        # the atom
+        if (len(nue) == 1) and (nue[0] < 0):
+            atom.SetFormalCharge(-nue[0])
+            mol.UpdatePropertyCache(strict=False)
+            continue
+        else:
+            neighbors = atom.GetNeighbors()
+            # check if one of the neighbors has a common NUE
+            for i, na in enumerate(neighbors, start=1):
```
IAlibay, line 395: It mostly makes sense to me, but I'm having trouble wrapping my head around whether or not this could lead to order specific behavoiur i.e. raising the bond order for one neighbour pair means that another won't be raised. I can't really think of a case where this would be a problem though.

Do you have any thoughts here?

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -95,10 +132,353 @@ def __init__(self, filename, **kwargs):
...
+    +-----------------------+-------------------------------------------+
+    | resnames              | atom.GetMonomerInfo().GetResidueName()    |
+    +-----------------------+-------------------------------------------+
+    | resids                | atom.GetMonomerInfo().GetResidueNumber()  |
+    +-----------------------+-------------------------------------------+
+    | segindices            | atom.GetMonomerInfo().GetSegmentNumber()  |
+    +-----------------------+-------------------------------------------+
+    | tempfactors           | atom.GetMonomerInfo().GetTempFactor()     |
+    +-----------------------+-------------------------------------------+
+    | bfactors              | atom.GetDoubleProp("_MDAnalysis_bfactor") |
```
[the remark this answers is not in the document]
cbouy: I completely agree, although I'm not sure which one should be added as a `TempFactor` in RDKit if both attributes are present in the AtomGroup...

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -95,10 +132,353 @@ def __init__(self, filename, **kwargs):
...
+    MDanalysis attributes are stored in each RDKit atom of the resulting 
+    molecule in two different ways: 
+
+    * in an `AtomPDBResidueInfo` object available through the 
+      ``atom.GetMonomerInfo()`` method if it's an attribute that is typically 
+      found in a PDB file,
+    * directly as an atom property available through the 
+      ``atom.GetPropsAsDict()`` method for the others.
+
+    Supported attributes:
```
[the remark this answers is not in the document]
cbouy: Sure, can do!

package/MDAnalysis/coordinates/RDKit.py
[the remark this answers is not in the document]
IAlibay: I'm tempted to say that @orbeckst's comment on what to do with the PDBWriter wouldn't be a bad idea here: https://github.com/MDAnalysis/mdanalysis/issues/1901#issuecomment-389578569

A sort of exclusive OR; if you have both error out letting users know that two sets of entries were found. It's probably overly prudent, but at the same time I'm not sure there are that many cases that users would end up with both sets of topology attributes.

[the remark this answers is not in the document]
cbouy: sounds good to me

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -95,10 +132,353 @@ def __init__(self, filename, **kwargs):
...
+            if hasattr(ag, attr):
+                other_attrs[attr] = getattr(ag, attr)
+
+        mol = Chem.RWMol()
+        atom_mapper = {}
+
+        for i, (atom, element) in enumerate(zip(ag, elements)):
+            # create atom
+            rdatom = Chem.Atom(element.capitalize())
+            # disable adding H to the molecule
```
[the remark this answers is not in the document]
cbouy: Like that ? left: NoImplicit=False, right: NoImplicit=True (the default). I'm still not sure if it's usable but it definitely looks less worse 😅
image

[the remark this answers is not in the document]
IAlibay: Yeah it is a tad bit better, although going from too few to too many hydrogens doesn't really fix the issue.

Any thoughts here @fiona-naughton @richardjgowers ?

package/MDAnalysis/coordinates/RDKit.py
cbouy: It is indeed order-specific, for example for a carboxylate group `R-C(-O)-O` the first oxygen read will receive a double bond and the other one will be charged. Of course IRL both oxygens share the charge and the bond order but that's not something RDKit was made to handle. In other words, I don't think there's anything I can do about it...

package/MDAnalysis/coordinates/RDKit.py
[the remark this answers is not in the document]
cbouy: It fixes the case where bond orders are present but hydrogens where removed though, so it could still be useful in a MOL2 file without hydrogens

cbouy:
> Does this fix all the issues with when there's no bonds or just part of them? 

Yup, this is fixed now with the `_standardize_patterns` function, the Nterm and Cterm of a selection in a protein won't be charged anymore. I'm not sure about nucleic acids though, I don't have a proper file to try that on, is there one with hydrogens in the test files ?

> Are some of the above cases still failing?

In the cases I previously mentioned, no, but there are probably some niche functional groups that I didn't include in  `_standardize_patterns`. Here's what's working so far, if you can think of something that's not here that would be helpful:
image

> if there's still "stuff to do" it might be a good call to have a checklist somewhere we can look at to keep track of things).

I've updated the first post

cbouy:
Is there a reason why the tests are not running ?

tylerjereddy:
> Is there a reason why the tests are not running ?

There was/is a major GitHub stability incident today: https://www.githubstatus.com/

I saw quite a few complaints that some repos were not even opening in a web browser.

IAlibay:
> > Does this fix all the issues with when there's no bonds or just part of them?
> 
> Yup, this is fixed now with the `_standardize_patterns` function, the Nterm and Cterm of a selection in a protein won't be charged anymore. I'm not sure about nucleic acids though, I don't have a proper file to try that on, is there one with hydrogens in the test files ?

Great work :) Sorry I probably won't get a chance to review until tomorrow, but in the mean time the types of systems that might be worth trying out are multi-ring systems like purines, beta-lactams (penicillin), and cases where you have multiple rings bonded to each other, such as 2e and f from: https://www.biorxiv.org/content/10.1101/286542v1.full.pdf

These are the types of systems where normal force field typing usually fails, so if we do better than those that's probably good enough.

Regarding hydrogens, do you think it's worth putting a check somewhere in the code that throws a warning to users if no H atoms are found in the structure? I.e. it's not validated and in most cases it can fail. We can also make adding implicit hydrogens an option and document that it _might_ help in cases of a MOL2 file (as you mentioned above).

cbouy:
> Sorry I probably won't get a chance to review until tomorrow

No worries! 

> the types of systems that might be worth trying out are multi-ring systems like purines, beta-lactams (penicillin), and cases where you have multiple rings bonded to each other

I'll check on those kind of systems and add them to our test cases.

> Regarding hydrogens, do you think it's worth putting a check somewhere in the code that throws a warning to users if no H atoms are found in the structure? I.e. it's not validated and in most cases it can fail. We can also make adding implicit hydrogens an option and document that it might help in cases of a MOL2 file (as you mentioned above).

Yes, a warning would definitely help. 
For the implicit hydrogens, I guess that means adding kwargs to the `ag.convert_to` method ? Should that be in a separate PR ?

cbouy:
Okay so I tried the first molecule:

IAlibay:
> Yes, a warning would definitely help.
> For the implicit hydrogens, I guess that means adding kwargs to the `ag.convert_to` method ? Should that be in a separate PR ?

If there's a warning (and/or docs regarding the limitations of a hydrogen-less world if you think adding the warning at this point is redundant), then I have no issues with the implicit stuff being addressed in a separate PR.

cbouy:
Okay so I tried the first molecule you mentioned:
image
And I remembered our discussion on the inferring of bond orders being dependent on the order of the atoms, so I tried converting this molecule to SMILES with a random order several times, and this is what happened:
image

So I guess I'll have to change a few things. I think I'll have to go back to the first version that I had in the draft PR, which was requesting that both atoms have the same number of unpaired electrons, instead of the current version where I take the smallest positive one.

Anyway, no need for the code review tomorrow then 😆

cbouy:
The possible fix that I mentioned yesterday actually didn't change anything, so for now I've added a quick fix for the above case:
Whenever the inferring function produces a molecule with the following pattern: an anion, followed by n times a single and double bond, then a single bond and another anion (`[*-] (-*=*)n -[*-]`) it can safely be converted to the form with n+1 alternating double and single bonds and no charges (`[*] (=*-*)n =[*]`). 
image
For now I'm doing the corresponding SMARTS patterns manually up to n=3, but ideally I should find a more general solution.

Here it is working on the previously failing test case:
image
I added a few more test cases like this one.

I also need to correct these weird crossed double bonds

IAlibay:
@cbouy just to let you know, I've not forgotten the review for this, I'm still trying to wrap my head around `_standardize_patterns`. I think part of the issue here is that I've not really used reactions properly in the past so I'm trying to get myself up to date on that one ASAP.

cbouy:
No worries, I'm still working on the general solution to the problem with conjugated bonds, hopefully I'll be able to publish it today!

For reactions the SMARTS syntax is relatively easy when you're familiar with SMILES, and the [documentation](https://www.daylight.com/dayhtml/doc/theory/theory.smarts.html) is pretty clear. 
What was harder for me to understand is how the RDKit implementation works, especially when the reactant reacts with itself. [This RDKit tutorial](https://www.rdkit.org/docs/GettingStartedInPython.html#chemical-reactions) might help a little.

cbouy:
So I ended up in a dead end yesterday with the first "general solution" I tried that worked with reactions. I might try something similar but just checking manually for alternating single and double bonds instead.

cbouy:
It's finally consistent no matter the order in which atoms are read 😄
image

cbouy:
While playing around with nglview I found this molecule in their datafiles which has a nitrogen at one edge of the conjugated system. It made the inferring fail hence my latest commit:
image
and before the fix:
image

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -59,7 +59,25 @@
...
+
+try:
+    from rdkit import Chem
+except ImportError:
+    pass
+else:
+    RDBONDTYPE = {
+        'AROMATIC': Chem.BondType.AROMATIC,
+        'SINGLE': Chem.BondType.SINGLE,
+        'DOUBLE': Chem.BondType.DOUBLE,
```
good-luck123: Duplicate of #

IAlibay: Any chance you could elaborate here @good-luck123? It's not immediately clear to me as to what you mean.

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -95,10 +133,538 @@ def __init__(self, filename, **kwargs):
...
+            _set_atom_property(rdatom, "_MDAnalysis_index", int(atom.ix))
+            # add atom
+            index = mol.AddAtom(rdatom)
+            atom_mapper[atom.ix] = index
+
+        try:
+            if (len(ag.bonds) == 0) and (ag.n_atoms > 1):
+                # force guessing bonds
+                raise NoDataError
+        except NoDataError:
```
tylerjereddy, line 311: What is the motivation for raising `NoDataError` and then capturing it right away to issue a warning and guess the bonds vs. just directly issuing the warning and guessing the bonds when the conditional is satisfied?

package/MDAnalysis/coordinates/RDKit.py
tylerjereddy: Also, less useful comment than Richard's, but `formated`->`formatted` I think

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -95,10 +133,538 @@ def __init__(self, filename, **kwargs):
...
+    elif isinstance(value, (int, np.int)):
+        atom.SetIntProp(attr, int(value))
+    else:
+        atom.SetProp(attr, value)
+
+
+def _infer_bo_and_charges(mol, terminal_atom_indices=[]):
+    """Infer bond orders and formal charges from a molecule.
+
+    Since most MD topology files don't explicitely retain informations on bond
```
tylerjereddy, line 389: typo: "explicitly" I think; and probably no "s" for "information"

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -95,10 +133,538 @@ def __init__(self, filename, **kwargs):
...
+    orders or charges, it has to be guessed from the topology. This is done by
+    looping other each atom and comparing its expected valence to the current
+    valence to get the Number of Unpaired Electrons (NUE).
+    If an atom has a negative NUE, it needs a positive formal charge (-NUE).
+    If two neighbouring atoms have UEs, the bond between them most
+    likely has to be increased by the value of the smallest NUE.
+    If after this process, an atom still has UEs, it's either a radical
+    (because one its bonds was cut when creating the AtomGroup) or it needs a
```
tylerjereddy, line 397: "one of"

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -95,10 +133,538 @@ def __init__(self, filename, **kwargs):
...
+            # exit the n_matches loop if there's no product. Example
+            # where this is needed: SO^{4}_{2-} will match the sulfone
+            # pattern 6 times but the reaction is only needed once
+            break
+    return reactant
+
+
+def _rebuild_conjugated_bonds(mol, max_iter=200):
+    """Rebuild conjugated bonds without negatively charged atoms at the
+    begining and end of the conjugated system
```
tylerjereddy, line 551: typo " beginning" I think

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -95,10 +133,538 @@ def __init__(self, filename, **kwargs):
...
+    behaviour by using an iterative procedure.
+    The problematic molecules always follow the same pattern: 
+    `anion(-*=*)n-anion` instead of `*=(*-*=)n*`, where `n` is the number of
+    successive single and double bonds. The goal of the iterative procedure is
+    to make `n` as small as possible by consecutively transforming
+    `anion-*=*` into `*=*-anion` until it reaches the smallest pattern with
+    `n=1`. This last pattern is then transformed from `anion-*=*-anion` to
+    `*=*-*=*`.
+    Since `anion-*=*` is the same as `*=*-anion` in terms of SMARTS, we can
+    control that we don't tranform the same triplet of atoms back and forth by
```
tylerjereddy, line 566: typo "transform"

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -95,10 +133,538 @@ def __init__(self, filename, **kwargs):
...
+            mol.UpdatePropertyCache(strict=False)
+            # start new iteration
+            continue
+
+        # no more changes to apply
+        return
+
+    # reached max_iter
+    warnings.warn("The standardization could not be completed within a "
+                  "reasonable ammount of iterations")
```
tylerjereddy, line 648: typo "amount"

testsuite/MDAnalysisTests/coordinates/test_rdkit.py
```diff
@@ -20,28 +20,104 @@
...
+        RDATTRIBUTES,
+        _add_mda_attr_to_rdkit,
+        _infer_bo_and_charges,
+        _standardize_patterns,
+        _rebuild_conjugated_bonds,
+        _set_atom_property,
+        _reassign_props_after_reaction,
+    )
+except ImportError:
+    def mol2_mol():
```
tylerjereddy, line 57: Just curious--the machinery that skips the tests when rdkit is absent is not sufficient to alleviate the need for "mocking" these functions?

Is this because of the trickery with using the functions in fixtures, so there's an issue where the skipping hasn't happened yet at test collection time?

That probably means there's a bit too much magic going on here, but I can see why I guess.

[Indirect parametrization](https://docs.pytest.org/en/latest/example/parametrize.html#indirect-parametrization) to allow fixture usage in the parametrization may be something to aim for eventually, even if this is "ok" here for now, to cut down on this top-level "mocking."

tylerjereddy:
FWIW, many projects try to avoid heavy usage of fixtures because of exactly these kinds of confusion/magic, though I think MDAnalysis is pretty heavily committed to them now, and I believe some of the core devs are quite keen on them.

package/MDAnalysis/coordinates/RDKit.py
cbouy: I do 2 things here: I implicitly check that ag.bonds exist, if it doesn't it will raise the NoDataError. And sometimes ag.bonds exist but is empty (PDB file without a CONECT record) so I need the if statement as well, and to avoid duplicate code I raise the same NoDataError. All this was prior to #2855 so I can safely remove the if statement now

testsuite/MDAnalysisTests/coordinates/test_rdkit.py
cbouy: Yup exactly, but I'm using indirect param now thanks to your comment 😃 I somehow missed this when looking at the pytest docs.
I think it's much better now

cbouy:
Regarding the coordinates and stereochemistry, should I keep going on this PR or do I open another one ? Not sure if this still qualifies as a "small PR" 😅

package/MDAnalysis/coordinates/RDKit.py
tylerjereddy: I see, yeah that probably deserved a comment at least, since it is a bit confusing as written. The repetition might have even been 'ok' to avoid the confusion (i.e., clarity over concision).

Sounds like you can simplify this now IIRC.

cbouy:
Coordinates are here 🙌
image

and stereochemistry as well:
image

IAlibay:
> Regarding the coordinates and stereochemistry, should I keep going on this PR or do I open another one ? Not sure if this still qualifies as a "small PR" 😅

Sorry, I didn't get to reply, but it's good enough that you've gone ahead with it. Unfortunately I'm a bit swamped right now, maybe one of the other @MDAnalysis/gsoc-mentors might want to jump in with some reviews.

cbouy:
> Edit; because I didn't say it, this is some really cool stuff! I can't wait to use it :)

Thanks 😄 

I saw that there was a decorator in `lib.util` to cache things but I couldn't find a way to cache only one AtomGroup and avoid clogging up the memory so I ended up implementing one. This way when we iterate over a trajectory we only have to build the RDKit topology once and save it in the cache, and the converter returns a copy of the cached mol with the corresponding coordinates. If a new universe or atomgroup is converted, the cache is cleared.

cbouy:
retinal_traj

Reading trajectories with RDKit: ✅

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -246,9 +249,10 @@ def convert(self, obj, **kwargs):
             raise TypeError("No `atoms` attribute in object of type {}, "
                             "please use a valid AtomGroup or Universe".format(
                                 type(obj))) from None
-        
+
         # create the topology
-        key = id(ag)
+        key = "<%x>" % id(ag) + ",".join(
+            ["%s=%s" % (str(k), str(v)) for k, v in kwargs.items()])
```
orbeckst, line 255: Maybe use f-strings? We seem to be getting rid of `%` interpolation.

orbeckst:
@IAlibay @fiona-naughton @richardjgowers what needs to happen for this to get merged?

cbouy:
I added a `cache=True` arg so that people might skip the "copy cached topology" part if necessary by setting it to False

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -95,10 +134,591 @@ def __init__(self, filename, **kwargs):
...
+        return mol
+
+    def atomgroup_to_mol(self, ag, NoImplicit=True):
+        """Converts an AtomGroup to an RDKit molecule.
+
+        Parameters
+        -----------
+        ag : AtomGroup
+
+        NoImplicit : bool
```
IAlibay, line 298: As above, we'll want to stick to PEP8 convention for variable naming.

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -95,10 +134,591 @@ def __init__(self, filename, **kwargs):
...
+
+        terminal_atom_indices = []
+        for bond in ag.bonds:
+            try:
+                bond_indices = [atom_mapper[i] for i in bond.indices]
+            except KeyError:
+                # one of the atoms of the bond is not part of the atomgroup.
+                # can happen for terminal atoms.
+                # save the bond atom that is in the atomgroup for later
+                terminal_atom_indices.extend([atom_mapper[i]
```
IAlibay, line 379: Note, my understanding here is that this heavily relies on atomgroup_intersection acting the way it currently does. If not already doing so, we explicitly test terminal atom indices somehow to make sure that any future changes (I think it was mentioned somewhere), don't unintentionally prevent this from working properly.

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -95,10 +134,591 @@ def __init__(self, filename, **kwargs):
...
+    if isinstance(value, (float, np.float)):
+        atom.SetDoubleProp(attr, float(value))
+    elif isinstance(value, (int, np.int)):
+        atom.SetIntProp(attr, int(value))
+    else:
+        atom.SetProp(attr, value)
+
+
+def _infer_bo_and_charges(mol, terminal_atom_indices=[]):
+    """Infer bond orders and formal charges from a molecule.
```
IAlibay, line 441: I've not tried generating the docs locally, so might be wrong, but I think by default because these are private functions they won't be added to the docs, however I do think that they should be available for users to see. It should be able to force sphinx to build the docs for these if it's not happening already.

(note mainly referring to this and the standardize patterns below)

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -95,10 +134,591 @@ def __init__(self, filename, **kwargs):
...
+                                              if i in atom_mapper.keys()])
+                # skip adding this bond
+                continue
+            bond_type = RDBONDORDER.get(bond.order, Chem.BondType.SINGLE)
+            mol.AddBond(*bond_indices, bond_type)
+
+        mol.UpdatePropertyCache(strict=False)
+
+        # infer bond orders and formal charges from the connectivity
+        _infer_bo_and_charges(mol, terminal_atom_indices)
```
IAlibay, line 390: shouldn't these be guarded by an `if not NoImplicit` statement?

package/MDAnalysis/coordinates/RDKit.py
IAlibay: Can we put this down in the docstring as a Note? If anything it might be useful to know for some user at some point.

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -95,10 +134,591 @@ def __init__(self, filename, **kwargs):
...
+
+
+def _standardize_patterns(mol):
+    """Standardizes functional groups
+
+    Uses :func:`_rebuild_conjugated_bonds` to standardize conjugated systems,
+    and SMARTS reactions for other functional groups.
+    Due to the way reactions work, we first have to split the molecule by
+    fragments. Then, for each fragment, we apply the standardization reactions.
+    Finally, the fragments are recombined.
```
IAlibay, line 526: I think I've finally wrapped my head around this. Here we will need to document a list of all the different reactions we are enforcing to happen. Users will need to be aware of it, in case it happens that they have a system with a feature that is unintentionally converted by one of these reactions.

testsuite/MDAnalysisTests/coordinates/test_rdkit.py
```diff
@@ -20,59 +20,480 @@
...
+        elements = mda.topology.guessers.guess_types(u.atoms.names)
+        u.add_TopologyAttr('elements', elements)
+        return u.select_atoms("resid 2-12")
+
+    @pytest.mark.parametrize("smi", ["[H]", "C", "O", "[He]"])
+    def test_single_atom_mol(self, smi):
+        u = mda.Universe.from_smiles(smi, addHs=False,
+                                     generate_coordinates=False)
+        mol = u.atoms.convert_to("RDKIT")
+        assert mol.GetNumAtoms() == 1
```
IAlibay, line 175: This might be a simple enough case for testing that the right elements were assigned to the mol?

testsuite/MDAnalysisTests/coordinates/test_rdkit.py
```diff
@@ -20,59 +20,480 @@
...
+            rd_value = getattr(mi, "Get%s" % rd_attr)()
+            if hasattr(sel, mda_attr):
+                mda_value = getattr(sel, mda_attr)[mda_index]
+                if mda_attr == "names":
+                    rd_value = rd_value.strip()
+                assert rd_value == mda_value
+
+    @pytest.mark.parametrize("rdmol", ["mol2_mol", "smiles_mol"],
+                             indirect=True)
+    def test_identical_topology(self, rdmol):
```
IAlibay, line 210: Would it be worth testing the full roundtrip too? i.e. MDA->RDKit->MDA

testsuite/MDAnalysisTests/coordinates/test_rdkit.py
```diff
@@ -20,59 +20,480 @@
...
+        u = mda.Universe(rdmol)
+        umol = u.atoms.convert_to("RDKIT")
+        assert rdmol.HasSubstructMatch(umol) and umol.HasSubstructMatch(rdmol)
+
+    def test_raise_requires_elements(self):
```
IAlibay, line 215: I reaslise these are set in code order, but I do wonder if it might be more readable by grouping all the error/warning testing together at the end of the class? (it's pedantic and doesn't change anything, so please don't feel like you have to do this)

testsuite/MDAnalysisTests/coordinates/test_rdkit.py
```diff
@@ -20,59 +20,480 @@
...
+        with pytest.raises(ValueError, match="Bad Conformer Id"):
+            mol.GetConformer()
+        # only 1 molecule should be cached
+        u = mda.Universe.from_smiles("C")
+        mol = u.atoms.convert_to("RDKIT")
+        assert len(cache) == 1
+        assert cache != previous_cache
+        # cache should depend on passed arguments
+        previous_cache = copy.deepcopy(cache)
+        mol = u.atoms.convert_to("RDKIT", NoImplicit=False)
```
IAlibay, line 339: This seems to be failing in the tests, do double check it works.

package/MDAnalysis/coordinates/RDKit.py
orbeckst: One can tell sphinx to also show undocumented members, or, include it explicitly in the list of members. See as example how `_reopen()` was documented https://github.com/MDAnalysis/mdanalysis/blob/5c0eb207d2141bb4166112fe3af4fe61d0fad2fc/package/MDAnalysis/coordinates/H5MD.py#L181

testsuite/MDAnalysisTests/coordinates/test_rdkit.py
orbeckst: Good idea -- a general acid test for converters!

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -95,10 +134,591 @@ def __init__(self, filename, **kwargs):
...
+        -----------
+        obj : AtomGroup or Universe
+
+        cache : bool
+            Use a cached copy of the molecule's topology when available. To be
+            used, the cached molecule and the new one have to be made from the
+            same AtomGroup object (same id) and with the same arguments passed
+            to the converter (with the exception of this `cache` argument)
+
+        NoImplicit : bool
```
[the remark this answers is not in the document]
cbouy: I chose to call this parameter `NoImplicit` for the same reason I named `AddHs` the parameter to add hydrogen atoms when using the `Universe.from_smiles` method: those are named after functions/methods of RDKit so it's easier for RDKit users to remember the argument name and what it does  
In #2707 @richardjgowers suggested that following RDKit's lead in argument names would be best so I followed the same principle here  
With that being said, I agree that mixing snake_case and PascalCase can be very confusing especially for newcomers  
So I don't really know what's best... should I create a poll on twitter ?

package/MDAnalysis/coordinates/RDKit.py
cbouy: With the API change discussed in #2821 the try except won't be needed anymore.
Regarding the terminal_atom_indices, I was using it to avoid adding charges to these atoms, but standardize_patterns should be able to cover this now so I can simplify it a bit

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -95,10 +134,591 @@ def __init__(self, filename, **kwargs):
...
+    .. versionadded:: 2.0.0
+
+    .. _RDKit: https://www.rdkit.org/docs/source/rdkit.Chem.rdchem.html#rdkit.Chem.rdchem.Mol
+    """
+
+    lib = 'RDKIT'
+    units = {'time': None, 'length': 'Angstrom'}
+    _cache = dict()
+
+    def convert(self, obj, cache=True, **kwargs):
```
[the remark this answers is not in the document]
cbouy: Will add it explicitly! And I can remove the kwargs then

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -95,10 +134,591 @@ def __init__(self, filename, **kwargs):
...
+                # create the topology
+                self._cache[key] = mol = self.atomgroup_to_mol(ag, **kwargs)
+            # continue on copy of the cached molecule
+            mol = copy.deepcopy(mol)
+        else:
+            self._cache.clear()
+            mol = self.atomgroup_to_mol(ag, **kwargs)
+
+        # add a conformer for the current Timestep
+        if hasattr(ag, "positions") and not np.isnan(ag.positions).any():
```
fiona-naughton, line 277: It might we worth throwing a warning here in case someone doesn't realise they have NaNs in their data and is expecting to have coordinates assigned

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -95,10 +134,591 @@ def __init__(self, filename, **kwargs):
...
+                other_attrs[attr] = getattr(ag, attr)
+
+        mol = Chem.RWMol()
+        # map index in universe to index in mol
+        atom_mapper = {}
+
+        for i, (atom, element) in enumerate(zip(ag, elements)):
+            # create atom
+            rdatom = Chem.Atom(element.capitalize())
+            # disable adding H to the molecule
```
fiona-naughton, line 346: This comment is outdated now `NoImplicit` can be False

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -95,10 +134,591 @@ def __init__(self, filename, **kwargs):
...
+    else:
+        atom.SetProp(attr, value)
+
+
+def _infer_bo_and_charges(mol, terminal_atom_indices=[]):
+    """Infer bond orders and formal charges from a molecule.
+
+    Since most MD topology files don't explicitly retain information on bond
+    orders or charges, it has to be guessed from the topology. This is done by
+    looping other each atom and comparing its expected valence to the current
```
fiona-naughton, line 445: other -> over

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -95,10 +134,591 @@ def __init__(self, filename, **kwargs):
...
+
+    terminal_atom_indices : list
+        List of terminal atoms indices, i.e. atoms at the edges of a molecule
+    """
+
+    for atom in mol.GetAtoms():
+        # get NUE for each possible valence
+        expected_vs = PERIODIC_TABLE.GetValenceList(atom.GetAtomicNum())
+        current_v = atom.GetTotalValence() - atom.GetFormalCharge()
+        nue = [v - current_v for v in expected_vs]
```
fiona-naughton, line 470: This same basic calculation is repeated several times here to get/update NUEs, so it might be worth making it its own function?

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -95,10 +134,591 @@ def __init__(self, filename, **kwargs):
...
+                if common_nue != 0:
+                    # increase bond order
+                    bond = mol.GetBondBetweenAtoms(
+                        atom.GetIdx(), na.GetIdx())
+                    order = common_nue + 1
+                    bond.SetBondType(RDBONDORDER[order])
+                    mol.UpdatePropertyCache(strict=False)
+                    if i < len(neighbors):
+                        # recalculate nue for atom
+                        current_v = atom.GetTotalValence()
```
fiona-naughton, line 504: I might be missing something, but is there a reason the formal charge gets used in the other NUE calculations, but not here?

testsuite/MDAnalysisTests/coordinates/test_rdkit.py
cbouy: You mean something like this ?
```python
u = mda.Universe(...)
mol = u.atoms.convert_to("RDKIT")
umol = mda.Universe(mol)
assert u == umol
```
I'm not sure how Universe comparison works

orbeckst: Yes, that would be the idea way – but I also don't think that we have `Universe.__eq__` defined...  so comparing topology things (list of atoms, bonds) (`assert_equal` for exact array comparisons) and positions (`assert_almost_equal` for floating point numbers) should do the job.

(Or raise an issue to make `Universe.__eq__` work, which will likely lead to having to make `Topology.__eq__` and `Reader.__eq__` work... at least `Timestep.__eq__` exists already.)

package/MDAnalysis/coordinates/RDKit.py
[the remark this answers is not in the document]
IAlibay: Best just pinging @richardjgowers here, I'd like to aim to get this merged by early next week at the latest, and going via a Twitter poll is probably going to be too time consuming.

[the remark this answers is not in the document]
cbouy: No problem, it would also affect Universe.from_smiles so it can be in a separate PR later

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -84,21 +126,708 @@ def _format_hint(thing):
...
+    and SMARTS reactions for other functional groups.
+    Due to the way reactions work, we first have to split the molecule by
+    fragments. Then, for each fragment, we apply the standardization reactions.
+    Finally, the fragments are recombined.
+
+    Notes
+    -----
+    The following functional groups are transformed:
+
+    +---------------+-------------------------------------------------------------------------+
```
IAlibay, line 590: Awesome, this is exactly what we needed here. However, I think that usually we have Parameters & Returns before Notes. Also could you add a line that mentions that these operations are done in order (I doubt there's much overlap, but this is really in case someone has something crazy that just gets overridden by all this).

package/MDAnalysis/coordinates/RDKit.py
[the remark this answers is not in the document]
orbeckst: I'd be ok with keeping RDKit names if they are passed through or do exactly the same thing. Keeping them like RDKit increases interoperability. 

For anything else I would follow our own conventions.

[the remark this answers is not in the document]
orbeckst: See also: [Guidelines vs Actual Rules](https://www.youtube.com/watch?v=jl0hMfqNQ-g)

IAlibay:
Just restarted the azure-pipelines CI, should be fixed now.
@cbouy looks like docs building is failing: "/home/travis/build/MDAnalysis/mdanalysis/package/MDAnalysis/coordinates/RDKit.py:docstring of MDAnalysis.coordinates.RDKit:36:Duplicate explicit target name: "rdkit""

Could you look into this & build things locally to see if there's any further issues?

cbouy:
@IAlibay should be good now 🤞

package/MDAnalysis/coordinates/RDKit.py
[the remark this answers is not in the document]
cbouy: I just noticed that I'm using `addHs` in Universe.from_smiles instead of `AddHs`, so I'll probably open a PR next week anyway

fiona-naughton:
Looks good to me!

Issue #2764, opened 2020-06-16
Setting attributes should have dtype checks


richardjgowers:
Our Topology system is essentially statically typed.  I recently had a test fail when you tried setting an atom name to an integer:  https://github.com/MDAnalysis/mdanalysis/pull/2755/commits/97e1782942a8262acfab2990460a129a15ed52bd

This would have previously worked as `dtype=object` (what we use for numpy arrays of python strings) will allow a Python integer into the array.  

We should have a `check_dtype` decorator around `TopologyAttr.set_atoms` (and residue/segment) to check that the supplied values are the correct type.

darmis007:
@richardjgowers i would like to work on this issue

b-thebest:
I would like to work on this issue

darmis007:
@richardjgowers  sir, can you please review my Pull Request #2788  regarding this issue

orbeckst:
@richardjgowers is this issue still relevant? Is there anything where you'd want to describe more clearly what you'd want to see so that the issue becomes suitable for the GSOC-starter label?

olivia632:
is this issue still open to work on

orbeckst:
I don't see an open PR referencing this issue so please feel free to open a PR. Make sure that you fill in the PR comment template properly and reference *this* issue.

fortune-max:
Hello @orbeckst. I'm Fortune from the Outreachy program. I just opened a pull request for this issue.

lilyminium:
@richardjgowers I'm not sure you can blindly enforce dtypes without also giving a preferred actual type; the dtype of string arrays is set to `object`, which an integer also is. Other people have also wanted to set TopologyAttrs of arbitrary type as well, so we can't assume every object is a string. The fix for strings might need to edit `_StringInternerMixin._set_X`  to just warn or error if a non-string is given.

The TypeError raised in the initial message isn't just because the dtype is wrong, it's because the value setter assumes the given value is either a string or a list. Then no further type checking is done. Setting multiple names works fine:

```
>>> u.atoms[:3].names = [1, 2, 3]
>>> u.atoms[:3].names
array([1, 2, 3], dtype=object)
```

Tagging @fortune-max here as well as you've opened a PR for the issue :)

richardjgowers:
Ok I see, its possible there are object topologyattrs that aren’t string,
so hacking the internet mixin seems good yeah.

fortune-max:
[...]

PR #2776, opened 2020-06-20
#FixForIssue2764

darmis007:
Fixes #

 - 
Fixed issue #2764 by adding the necessary decorator

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/core/topologyattrs.py
+++ b/package/MDAnalysis/core/topologyattrs.py
@@ -487,9 +487,58 @@ class Atomnames(AtomAttr):
     dtype = object
     transplants = defaultdict(list)
 
+    def __init__(self, vals, guessed=False):
+        self.namedict = dict()  # maps str to nmidx
+        name_lookup = []  # maps idx to str
+        # eg namedict['O'] = 5 & name_lookup[5] = 'O'
+
+        self.nmidx = np.zeros_like(vals, dtype=int)  # the lookup for each atom
+        # eg Atom 5 is 'C', so nmidx[5] = 7, where name_lookup[7] = 'C'
+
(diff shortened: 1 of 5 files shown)


package/MDAnalysis/core/topologyattrs.py
```diff
@@ -59,6 +59,34 @@
 from .. import _TOPOLOGY_ATTRS, _TOPOLOGY_TRANSPLANTS, _TOPOLOGY_ATTRNAMES
 
 
+def _check_dtype(func):
+    def _attr_dtype(values):
+        # check dtype of values array
+        dtypes = ['int8','int16','int32','int64']
```
richardjgowers, line 65: This is a correct check for the attrs which are strictly ints (eg atom ids, resids).  Some attrs are strings (which we store as numpy arrays of `object` dtype.  I think with this approach we'll need a decorator for each case

package/MDAnalysis/core/topologyattrs.py
```diff
@@ -59,6 +59,34 @@
...
+        try:
+            if values.dtype in dtypes:
+                return "yes"
```
richardjgowers, line 68: rather than return "yes"/"no" you can use the boolean True/False

package/MDAnalysis/core/topologyattrs.py
darmis007: So should i increase the extend the datatypes in the dtypes array

darmis007:
I am making a new Pull request with build from develop branch instead from the above mentioned branch . Hence i m closing this branch

Issue #2773, opened 2020-06-18
set status to mature


orbeckst:
Our classifiers should indicate that we're out of beta (after > 12 years?).

https://github.com/MDAnalysis/mdanalysis/blob/788f294e53ca19fdfba8cc7e2f86d0de7be82bbf/package/setup.py#L534

Should be changed for 
- 1.x
- 2.x

IAlibay:
Somehow the autoclose keywords on the two PRs didn't work, closing now.

orbeckst:
Apparently, it only works for merges into develop.

PR #2746, opened 2020-06-11
Remove deprecated hbonds.hbond_analysis

RMeli:
Related to #2739, also address #2745

Changes made in this Pull Request:
 - Removed deprecated `hbonds.hbond_analysis`
 - Moved `hbonds.hbond_autocorrel` and `hbonds.wbridge_analysis` to `hydrogenbonds`
   - Raise `DeprecationWarning` if `hbonds` is imported instead of `hydrogenbonds`

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?
 - [x] Issue raised/referenced?

PR #2778, opened 2020-06-20
Remove deprecated hole, and completes py2 removal

### LINKED ISSUE #2541: remove Python 2 constructs

IAlibay:
Fixes #2541, Towards #2773 and #2739 , Changelog for #2765

 - Removed original deprecated hole code in favour of new hole2.
 - Removed six, and future from remaining files (note, as #2746 may take some time to complete, I removed the six & future calls here, even if it will just get deleted in the end).
 - Setup scripts no longer take sub-3.6 python versions.
 - Development status set of `6 - Mature` for 2.x.x.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/CHANGELOG
```diff
@@ -18,6 +18,8 @@ The rules for this file:
   * 2.0.0
 
 Fixes
+  * Selections on emtpy AtomGroups now return an empty AtomGroup instead of an
```
IAlibay, line 21: Added this one in as requested by @lilyminium

lilyminium: Thank you! Could you add in my username above as well?

IAlibay: Oops my bad, done

package/MDAnalysis/analysis/hole2/__init__.py
```diff
@@ -26,14 +26,14 @@
 
 due.cite(Doi("10.1016/S0006-3495(93)81293-1"),
          description="HOLE program",
-         path="MDAnalysis.analysis.hole",
+         path="MDAnalysis.analysis.hole2",
```
IAlibay, line 29: Not super sure on these (nor on the addition of hol2 to MDAnalysis.analysis.__init__ would be worth specifically double checking here).

benchmarks/asv.conf.json
```diff
@@ -31,7 +31,6 @@
         "Cython": [],
         "numpy": [],
         "scipy": [],
-	    "six": [],
```
orbeckst, line 34: You can’t remove six from asv because it is needed for testing the older versions.

orbeckst: More precisely, we need to check what should be there and what we are actually using. I think i only have NumPy and cython in the actual cfg file, see the benchmark repo/wiki.  However, just leave six here until this gets sorted out in a separate asv cleanup PR. Better to leave the history in place because ASV needs to build historical versions of the code.

package/MDAnalysis/analysis/__init__.py
```diff
@@ -59,7 +59,7 @@
 :mod:`~MDAnalysis.analysis.helanal`
     Analysis of helices with the HELANAL_ algorithm.
 
-:mod:`~MDAnalysis.analysis.hole`
+:mod:`~MDAnalysis.analysis.hole2.hole`
```
orbeckst, line 62: Would be nicer to have a page for the hole2 package instead of going to its module. But can be separate PR if it requires work.

package/MDAnalysis/analysis/hole2/__init__.py
orbeckst: What are you unsure about?

package/MDAnalysis/analysis/__init__.py
IAlibay: So it turns out this `__init__` text isn't actually used anywhere for docs generation (as far as I can tell), the hole2 package links to: https://www.mdanalysis.org/docs/documentation_pages/analysis/hole2.html#hole-analysis-mdanalysis-analysis-hole2

I can just get rid of the `.hole` portion if you think it's more appropriate.

package/MDAnalysis/analysis/hole2/__init__.py
IAlibay: Honestly it's my first time dealing with duecredit, I wasn't sure if this should be triggered on importing `MDAnalysis.analysis.hole2` or `MDAnalysis.analysis.hole2.hole`.

benchmarks/asv.conf.json
IAlibay: Makes sense, done in e5a3fe1

richardjgowers:
@IAlibay sort out the conflicts and we can merge this probably

package/MDAnalysis/analysis/__init__.py
orbeckst: I think hole2 is more appropriate. 

It looks as if the doc string requires an update anyway... probably getting rid of most of it and just saying something simple for anyone using `help(MDA.analysis)`. But can be different issue.

IAlibay: I changed it in `fa48e06`, but on testing on a page for which docs actually do get built for, it seems like it won't link to anything. Either way if the whole thing is going to get re-written, and considering that these docs don't get built anyways, then it might be good enough for now?

GeraZerbetto:
Hello everyone! I am a MDA. Hole user but I am quite interested in the development of this tool. I have been using this module to run analysis on quite long trajectories, resulting in long computing times. I don't know if this is the place of doing such a question (I am very new at Github, so I don't know the manners here): is there a way to make the code run parallel frames using python multiprocessing, so a single trajectory could be run on multiple threads? I am very new to all of this, sorry if this is not the place for this!
Regards!

IAlibay:
Hi @GeraZerbetto,

Welcome to the MDA (or at least the github side of things if you are already a use the code 😄 )!

Ideally could you ask this question on either the [developer](https://groups.google.com/group/mdnalysis-devel) or [user](https://groups.google.com/group/mdnalysis-discussion) discussion email lists?

Roughly, the way we try to organise things are in the following manner:

1) code usage discussion: [user mailing list](https://groups.google.com/group/mdnalysis-discussion)
2) code development discussions: [developer mailing list](https://groups.google.com/group/mdnalysis-devel)
3) issues & feature requests: [github issues](https://github.com/MDAnalysis/mdanalysis/issues)
4) active code contributions: [github pull requests](https://github.com/MDAnalysis/mdanalysis/pulls)

GeraZerbetto:
Hi @IAlibay 
Thanks a lot!

PR #2783, opened 2020-06-22
Makes tidynamics a non required install

### LINKED ISSUE #2782: Inconsistent install requirements for tidynamics
## Expected behavior ##

Optional packages should not be installed by default.

## Actual behavior ##

The code for MSD seems to indicate that tidynamics is an optional import, however, unless other optional imports (e.g. parmed and RDKIT) it's in the `install_requires` lists for both setup.py scripts. Should these entries be there?

Current develop branch code.

IAlibay:
Fixes #2782

 - Removes tidynamics from install_requires lists
 - Refactors MSD tests to allow for partial testing when tidynamics is not available.
 - Make MSD FFT tests skip if no tidynamics

To do:
 - <s>There are two reasonably easy tests for MSD that I could add here whilst I'm at it (ImportError and ValueError). For the ImportError, I think the only way to get this done cleanly is to put the current tests in a class and put the skip as a decorator for that class only, thoughts?</s>
 - <s>PEP8 fixes</s>

PR Checklist
------------
 - [x] Tests?
 - [x] Docs? N/A
 - [x] CHANGELOG updated? Nothing changed for users here, is this necessary?

### DIFF
--- a/testsuite/MDAnalysisTests/analysis/test_msd.py
+++ b/testsuite/MDAnalysisTests/analysis/test_msd.py
@@ -91,16 +91,6 @@ def step_traj_arr(NSTEP):  # constant velocity
     return traj
 
 
-def random_walk_3d(NSTEP):
-    np.random.seed(1)
-    steps = -1 + 2*np.random.randint(0, 2, size=(NSTEP, 3))
-    traj = np.cumsum(steps, axis=0)
-    traj_reshape = traj.reshape([NSTEP, 1, 3])
-    u = mda.Universe.empty(1)
-    u.load_new(traj_reshape)
-    return u, traj
-
-
 def characteristic_poly(n, d):  # polynomial that describes unit step traj MSD
     x = np.arange(0, n)
     y = d*x*x
--- a/testsuite/MDAnalysisTests/analysis/test_msd.py
+++ b/testsuite/MDAnalysisTests/analysis/test_msd.py
@@ -103,6 +103,12 @@ class TestMSDSimple(object):
         # test some basic size and shape things
         assert_equal(msd.n_particles, 10)
 
+    @pytest.mark.parametrize('msdtype', ['foo', 'bar', 'yx', 'zyx'])
+    def test_msdtype_error(self, u, SELECTION, msdtype):
+        errmsg = f"invalid msd_type: {msdtype}"
+        with pytest.raises(ValueError, match=errmsg):
+            m = MSD(u, SELECTION, msd_type=msdtype)
+
     @pytest.mark.parametrize("dim, dim_factor", [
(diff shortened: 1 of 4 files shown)


testsuite/MDAnalysisTests/analysis/test_msd.py
```diff
@@ -41,14 +41,29 @@ def SELECTION():
...
-    return nstep
+def u():
+    return mda.Universe(PSF, DCD)
+
+
+@block_import('tidynamics')
+def test_notidynamics(u, SELECTION):
+    # Have to avoid importing until here as otherwise tidynamics will already
+    # be imported
+    from MDAnalysis.analysis.msd import EinsteinMSD as MSD
```
IAlibay, line 52: I'm not super happy about how messy this looks, but it's the cleanest way I could think of testing this without having the global import override the presence of tidynamics..

testsuite/MDAnalysisTests/analysis/test_msd.py
```diff
@@ -83,133 +90,138 @@ def step_traj(NSTEP):  # constant velocity
...
+    def test_random_walk_u_simple(self, random_walk_u):
+        # regress against random_walk test data
+        msd_rw = MSD(random_walk_u, 'all', msd_type='xyz', fft=False)
+        msd_rw.run()
+        norm = np.linalg.norm(msd_rw.timeseries)
+        val = 3932.39927487146
+        assert_almost_equal(norm, val, decimal=5)
+
+
+def import_check(module):
```
IAlibay, line 142: Unfortunately had to to do this to avoid doing pytest.importorskip everywhere (as far as I can tell you can just use pytest.importorskip at the class level, it just defaults to the whole module).

If we think it might be useful somewhere else, I can probably move this to MDAnalysisTests.utils though.

richardjgowers: I think maybe the way to do this is `pytest.mark.skipif()` then apply that to each class? This isn't hideous though

IAlibay: Sorry, probably going to need clarification here, I'm not sure if you mean what's on line 157 or something different?

orbeckst: Perhaps write the equivalent of `pytest.importorskip` that actually works at the class level, just like what you have, and put the new decorator to `utils.skipif_noimport` (or somesuch)... but what you have here is already pretty clear, especially if you put `import_check()` into utils.

IAlibay: For simplicity I've just moved the function over to utils as `import_not_available`, it made more sense considering we already have a similar `executable_not_found` there.

orbeckst:
Good choice!

Issue #2247, opened 2019-04-14
HydrogenBondLifetimes (Autocorrelation) has an increasing value


bieniekmateusz:
The autocorrelation has to be decreasing. 

Adjusted test case from the test_waterdynamics.py file:
```python
def test_HydrogenBondLifetimes_growing_continuous(universe):
    # The autocorrelation cannot grow
    hbl = waterdynamics.HydrogenBondLifetimes(
        universe, SELECTION1, SELECTION1, t0=0, tf=9, dtmax=5)
    hbl.run()
    print("results", hbl.timeseries)
    # Index 0 in frameX[0] refers to the continuous, 1 to intermittent
    assert all([frameX[0] > frameXplus1[0] for frameX, frameXplus1 in zip(hbl.timeseries, hbl.timeseries[1:])])
```
The current autocorrelation code in HydrogenBondLifetimes returns lifetimes that increase over time. We  are planning to address this in the next Pull Request with @p-j-smith where we add the test case and use our standalone autocorrelation function. 

- Which version are you using? developing branch
- Python 2.7
- Linux Arch

orbeckst:
Are you saying that the calculated ACF must be monotonically decreasing? Not even a little bit of noise allowed?

bieniekmateusz:
In the actual experiments I imagine the noise could create this kind of data. However, in the case of continuous ACF this cannot happen. This is because at each point t0, we take our bonds and check if they survive at every step up until a given tau, say tau=t4. If they leave at t3, we do not allow it to return, and so t4 is thrown away (we say it has left). So, by definition, the continuous ACF will be monotonically decreasing. 

Now, some implementations of the intermittent ACF could lead to the noise you suggest. In our case, it is not possible: we say that if the molecule leaves only at t3 and returns at t4, with the intermittency=1, then we treat this bond as if it was there at every step from t0 to t4, including at t3. So again, it is not possible for the ACF to increase.

Paul & Mat

richardjgowers:
Yeah this is slightly worrying. I’ve not looked in too much detail at this yet, is there anything this class does that the hbond autocorrel doesn’t do? And how hard would it be to merge the two?
  

  
  

  
  
>

bieniekmateusz:
@richardjgowers The HydrogenBondLifetimes does not do anything extra. We're hoping to finish soon submit the next PR. We're going to add a function to the Richard's HydrogenBond finder to get the autocorrelation directly. 

Mat

PR #2791, opened 2020-06-25
Hydrogen Bond Lifetime in waterdynamics.hydrogenbonds

### LINKED ISSUE #2547: Turn waterdynamics.HydrogenBondLifetimes into function of HydrogenBondAnalysis
Given that `HydrogenBondLifetimes` just runs (the now deprecated) hbonds.HydrogenBondAnalysis under the hood, it seems it could be easily turned into a post-run() function of HydrogenBondAnalysis.

Turn it into a function of HydrogenBondAnalysis.

At least use the new class instead of the old one.

**Additional context**
This is the `run()` function:

```python
def run(self, **kwargs):
        """Analyze trajectory and produce timeseries"""
        h_list = MDAnalysis.analysis.hbonds.HydrogenBondAnalysis(self.universe,
                                                                 self.selection1,
                                                                 self.selection2,
                                                                 distance=3.5,
                                                                 angle=120.0)
        h_list.run(**kwargs)
        self.timeseries = self._getGraphics(h_list.timeseries, self.t0,
                                            self.tf, self.dtmax)
```

bieniekmateusz:
Fixes #2547

 - The standalone autocorrelation function is now used to calculate the hydrogen bond lifetime in the new implementation of the hydrogen bond analysis class. 
 - We deprecate the waterdynamics.HydrogenBondLifetime because of #2247.
 - This function can also replace the hbonds.hbond_autocorrel module. However, we haven't deprecated this module here as that is being addressed by PR #2746 
 - Basic unit tests added to check the integration of the separate components

Co-authored-by: @p-j-smith  <paul.smith@kcl.ac.uk> 

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/analysis/waterdynamics.py
```diff
@@ -513,6 +513,12 @@ class HydrogenBondLifetimes(object):
 
     def __init__(self, universe, selection1, selection2, t0, tf, dtmax,
                  nproc=1):
+        warnings.warn(
+            "This class is depracated. "
```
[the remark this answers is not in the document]
orbeckst: The decorator also adds a note to the reST string which you will need to do manually otherwise. Something like
```reST
.. deprecated:: 2.0.0
   This analysis class will be removed in release 3.0. 
    Please use :meth:`MDAnalysis.analysis.hydrogenbonds.HydrogenBondAnalysis.autocorrelation` instead
```
or similar (making sure that the reST link works).

package/MDAnalysis/lib/correlations.py
```diff
@@ -223,36 +223,36 @@ def correct_intermittency(list_of_sets, intermittency):
 
     list_of_sets = deepcopy(list_of_sets)
 
-    for i, ids in enumerate(list_of_sets):
+    for i, elements in enumerate(list_of_sets):
```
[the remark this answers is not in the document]
bieniekmateusz: I added a comment explaining that an element might refer to different things, such as:
- a pair of atoms in the case of hydrogen bond lifetime, 
- an atom ID when computing water survival probability

Another name that could be applicable could be "state" or "event", as we are checking for survival probability of these. Please let me know if you have a preference. 

Thanks!

hmacdope:
Hey all, looking great!

Not sure how applicable to this use case, but just thought I would let you know that tidynamics, which is now an optional dependency, provides an ACF computation option using the FFT based fast correlation algorithm.

Perhaps some of the computation can be outsourced? :)

bieniekmateusz:
@hmacdope Thanks for pointing this out. I'll look into it!

bieniekmateusz:
After further thought, we are going to extend this pull request to make it easier to obtain hydrogen bond lifetimes. Please don't merge yet. Thanks

p-j-smith:
In adding the hydrogen bond lifetime functionality we realised it was not possible to find, say, protein-water hydrogen bonds without also finding protein-protein and water-water-hydrogen bonds. This would therefore make the HBL function of limited use if one wanted to know the lifetime of protein-water hydrogen bonds. We have therefore added a `between` argument to `HydrogenBondsAnalysis ` to make it possible to specify pairs of atoms groups between which hydrogen bonds will be formed. For example `between=["protein", "resname SOL"]` will find only protein-water hydrogen bonds.

This is now in keeping with both the VMD hbonds plugin and `MDAnalysis.analysis.hbobds.hbond_analysis.HydrogenBondsAnalysis`, although we thought a `between` keyword, rather than `sel1` and `sel2`, would be preferable. This is because it allows for more flexibility, such as passing two pairs of selections, e.g `between=[["protein", "protein"], ["protein", "resname SOL"]]`  will find all protein-protein and protein-water hydrogen bonds but not water-water hydrogen bonds.

We have added test cases for this new argument and the interface is backwards compatible - the default is `between=None`, in which case all hydrogen bonds are found.

We also plan to add an `of` keyword to the HBL lifetime function, e.g `hbonds.lifetime(of=["resname ASP", "resname SOL"])` would calculate the hydrogen bond lifetime of hydrogen bonds between aspartic acid and water. One could therefore use the `between` keyword to find all protein-water hydrogen bonds, and subsequently use `hbonds.lifetime(of)` to calculate the HBL of particular specific hydrogen bonds. We don't want to bloat this PR too much and so we'll leave this `of` argument for another time.

And thanks @hmacdope for pointing us toward tidynamics. The autocorrelation function we're using to calculate the HBL is also used by `MDAnalysis.analysis.waterdynamics` to calculate the survival probability of water around proteins, and so it would probably be best to make any changes to the autocorrelation function in a separate PR. But it's definitely something to look into!

package/CHANGELOG
```diff
@@ -32,6 +32,7 @@ Enhancements
     token (Issue #2468, PR #2707)
   * Added the `from_smiles` classmethod to the Universe (Issue #2468, PR #2707)
   * Added computation of Mean Squared Displacements (#2438, PR #2619)
+  * Added Hydrogen Bond Lifetime via existing autocorrelation features (PR #2791)
```
orbeckst, line 35: Add another line for `between`

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -206,6 +209,14 @@ def __init__(self, universe, donors_sel=None, hydrogens_sel=None, acceptors_sel=
         acceptors_sel : str
             Selection string for the hydrogen bond acceptor atoms. Leave as `None` to guess which atoms to use in the
             analysis using :attr:`guess_acceptors`
+        between : List (optional)
+            Specify two selection strings for non-updating atom groups between which hydrogen bonds will be calculated.
+            For example, if the donor and acceptor selections include both protein and water, it is possible to
+            find only protein-water hydrogen bonds - and not protein-protein or water-water - by specifying
+            :attr`between=["protein", "SOL"]`. If a two-dimensional list is passed, hydrogen bonds between each pair
+            will be found. For example, :attr`between=[["protein", "SOL"], ["protein", "protein"]]` will calculate
+            all protein-water and protein-protein hydrogen bonds but not water-water hydrogen bonds. If None, hydrogen
```
orbeckst, line 218: Do you mean `None` (Python value) or string “None”? Use markup.

Is `None` the default?

orbeckst: Add a versionadded 2.0.0 for between.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -206,6 +209,14 @@ def __init__(self, universe, donors_sel=None, hydrogens_sel=None, acceptors_sel=
         acceptors_sel : str
             Selection string for the hydrogen bond acceptor atoms. Leave as `None` to guess which atoms to use in the
             analysis using :attr:`guess_acceptors`
+        between : List (optional)
```
orbeckst, line 212: Add an example to the docs above that shows how to use “between”, essentially, take the text here and add code.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -228,6 +239,29 @@ def __init__(self, universe, donors_sel=None, hydrogens_sel=None, acceptors_sel=
         self.donors_sel = donors_sel
         self.hydrogens_sel = hydrogens_sel
         self.acceptors_sel = acceptors_sel
+
+        # If hydrogen bonding groups are selected, then generate corresponding atom groups
+        if between is not None:
+
+            if type(between) is not list or len(between)==0:
```
orbeckst, line 246: Use `isinstance()` instead of the `type() == ...` checks (also below).

Or do you really have to type check? Can’t you just work with anything that will allow you to do the iteration, e.g., a numpy array? Handle with try/except and duck typing is probably more pythonic.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -496,6 +567,67 @@ def _conclude(self):
...
+        Unique hydrogen bonds are identified using hydrogen-acceptor pairs. This means an acceptor
+        switching to a different hydrogen atom - with the same donor - from one frame to the next
+        is considered a different hydrogen bond.
+
+        Please see :func:`MDAnalysis.lib.correlations.autocorrelation` and
+        :func:`MDAnalysis.lib.correlations.intermittency` functions for more details.
+
+
+        Parameters
+        ----------
```
orbeckst, line 587: THere is no `of` parameter. Remove doc?

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -496,6 +567,67 @@ def _conclude(self):
...
+            autcorrelation value for each value of `tau`
+        """
+
+        if not hasattr(self, 'hbonds'):
+            logging.error("Autocorrelation analysis of hydrogen bonds cannot be done before the hydrogen bonds are found")
+            logging.error("Autocorrelation: Please use the .run() before calling this function")
+            raise NoDataError("No .hbonds: use the .run() first")
+
+        if self.step != 1:
+            logging.warning("Autocorrelation: ideally autocorrelation function would be carried out on consecutive frames. ")
```
orbeckst, line 616: State that the hbonds were computed with a step>1 and that you recommend that they be recomputed with step=1.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -496,6 +567,67 @@ def _conclude(self):
...
+            logging.warning("Autocorrelation: if you would like to allow bonds to break and reform, please use 'intermittency'")
+
+        # Extract the hydrogen bonds IDs only in the format [set(superset(x1,x2), superset(x3,x4)), ..]
+        found_hydrogen_bonds = [set() for _ in self.frames]
```
orbeckst, line 620: Does `[set()] * len(self.frames)` replicate the same empty set or does it create what you need here?

package/MDAnalysis/analysis/waterdynamics.py
```diff
@@ -513,6 +513,12 @@ class HydrogenBondLifetimes(object):
 
     def __init__(self, universe, selection1, selection2, t0, tf, dtmax,
                  nproc=1):
+        warnings.warn(
+            "This class is deprecated. "
+            "Please use MDAnalysis.analysis.hydrogenbonds.HydrogenBondAnalysis.autocorellation() instead.",
```
orbeckst, line 518: Isn’t the method called `HhydrogenBondAnalysis.lifetime()`?

package/MDAnalysis/analysis/waterdynamics.py
```diff
@@ -513,6 +513,12 @@ class HydrogenBondLifetimes(object):
 
     def __init__(self, universe, selection1, selection2, t0, tf, dtmax,
                  nproc=1):
+        warnings.warn(
+            "This class is deprecated. "
```
orbeckst, line 517: Need to add
- When it will be removed (2.0.0)
- .. deprecated reST to the text

(Have a look at other code how we normally deprecate.)

package/MDAnalysis/lib/correlations.py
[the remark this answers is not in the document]
orbeckst: Comment is good, thanks.

testsuite/MDAnalysisTests/analysis/test_hydrogenbonds_analysis.py
```diff
@@ -207,6 +209,140 @@ def test_count_by_time(self, universe):
         assert_array_almost_equal(h.times, ref_times)
         assert_array_equal(counts, ref_counts)
 
+    def test_hydrogen_bond_lifetime(self, universe):
+        hbonds = HydrogenBondAnalysis(universe, **self.kwargs)
```
orbeckst, line 213: Can this be made a class-level fixture so that it only needs to run once?

orbeckst: Or does this create problems because the instance is modified in different ways?

Then it still makes sense as a normal method/function fixture to just reduce code repetition.

bieniekmateusz: Added a fixture to replace most of the cases.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
bieniekmateusz: I've attempted to address this and will read more about duck typing. Thanks

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
bieniekmateusz: Just tried it. I am afraid the reference points to the same empty set.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -496,6 +631,95 @@ def _conclude(self):
...
+
+        Returns
+        -------
+        tau_timeseries : np.array
+            tau from 1 to `tau_max`
+        timeseries : np.array
+            autcorrelation value for each value of `tau`
+        """
+
+        if not hasattr(self, 'hbonds'):
```
orbeckst, line 680: Wouldn't it be safe to always have `.hbonds` but set it to `None` (in init) or any empty data structure? If find testing `if not self.hbonds` clearer.

package/MDAnalysis/analysis/waterdynamics.py
```diff
@@ -508,11 +508,22 @@ class HydrogenBondLifetimes(object):
     .. versionchanged:: 1.0.0
        The ``nproc`` keyword was removed as it linked to a portion of code that
        may have failed in some cases.
+    .. deprecated:: 2.0.0
+       Instead, please use
```
orbeckst, line 512: The deprecation will have to be in a 1.0.x
```suggestion
    .. deprecated:: 1.0.1
       ``waterdynamics.HydrogenBondLifetimes`` is deprecated and will be removed in 2.0.0.
       Instead, please use (available in 2.0.0)
```

package/MDAnalysis/analysis/waterdynamics.py
```diff
@@ -508,11 +508,22 @@ class HydrogenBondLifetimes(object):
...
+
     """
 
 
     def __init__(self, universe, selection1, selection2, t0, tf, dtmax,
                  nproc=1):
+        warnings.warn(
+            "This class is deprecated. "
+            "Instrad, please use"
+            "MDAnalysis.analysis.hydrogenbonds.HydrogenBondAnalysis.lifetime",
```
orbeckst, line 523:
```suggestion
            "This class will be removed in 2.0.0; please prepare to use "
            "MDAnalysis.analysis.hydrogenbonds.HydrogenBondAnalysis.lifetime (in 2.0.0)",
```

testsuite/MDAnalysisTests/analysis/test_hydrogenbonds_analysis.py
```diff
@@ -166,6 +169,17 @@ def universe():
 
         return u
 
+    @staticmethod
+    @pytest.fixture(scope='class')
+    def hydrogen_bonds(universe):
+        h = HydrogenBondAnalysis(
+            universe,
+            **TestHydrogenBondAnalysisMock.kwargs
```
orbeckst, line 177: Mock?

bieniekmateusz:
Please let us know if you spot anything else! Thanks!

orbeckst:
Thank you @bieniekmateusz and @p-j-smith – sorry it took me a while to get back to the PR.

orbeckst:
@bieniekmateusz @p-j-smith – You know what? We should have removed `wateranalysis.HydrogenBondLifetimes` in this PR because _this_ PR is really for 2.0.0 when the class will have been removed. Sorry, I didn't quite realize it during review. I backported the deprecation to #2798 (for 1.0.1). Can you please submit another PR where you remove the deprecated functionality and add a CHANGELOG entry under Changes saying so? Thanks!

PR #2798, opened 2020-06-26
Backport 1.0.1

orbeckst:
Fixes #2768 

Changes made in this Pull Request:
- includes backports of all fixes tagged with [Milestone 1.0.x](https://github.com/MDAnalysis/mdanalysis/milestone/17) (currently 1.0.1)
- cherry-picked commits

Add to this branch and _merge_ into Master before releasing 1.0.1.

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?
 - [x] Issue raised/referenced?

PR #2804, opened 2020-06-29
Make the TPR parser a bit faster

jbarnoud:
A function in the TPR parser calls list.pop thousands of times, which is
slow. This commit avoids that exansive call.

Taking the TPR from
https://github.com/bioexcel/covid_modelling_simulation_data/tree/master/spike_protein/full_spike/trimer
the parsing time on my computer goes from 25.6s to 9.39s. On a more
pathological TPR file, it goes from 3 minutes to about 6s.

 - 

PR Checklist
------------
 ~- [ ] Tests?~
 ~- [ ] Docs?~
 - [x] CHANGELOG updated?


orbeckst:
Could be backported to 1.0.1 via PR #2798.

PR #2805, opened 2020-06-30
make Atom target class of ResidueAttr by default

### LINKED ISSUE #2803: Insertion code attribute not available for single atom
## Expected behavior ##

Accessing `u.atoms[0].icode` should return the icode value if the icodes are present

## Actual behavior ##

Even if the icodes attribute exists, the above will return a NoDataError. `u.atoms.icodes[0]` will work though

``` python
import MDAnalysis as mda
import numpy as np

u = mda.Universe.from_smiles("CCO")
icodes = np.array(["X"], dtype=object)
u.add_TopologyAttr("icodes", icodes)

u.atoms.icodes
# array(['X', 'X', 'X', 'X', 'X', 'X', 'X', 'X', 'X'], dtype=object)

u.atoms[0].icode
# NoDataError: This Universe does not contain icode information
```
Doing the same with the `resnames` attribute will not reproduce this error:

```python

resnames = np.array(["UNK"], dtype=object)
u.add_TopologyAttr("resnames", resnames)

u.atoms.resnames
# array(['UNK', 'UNK', 'UNK', 'UNK', 'UNK', 'UNK', 'UNK', 'UNK', 'UNK'], dtype=object)

u.atoms[0].resname
# 'UNK'

```

2.0.0-dev0
Python 3.6.10
- Which operating system?
Ubuntu 18.04.4 LTS through a VM (WSL2)

lilyminium:
Fixes #2803 

 - made Atom a target class of ResidueAttr by default
 - made ICode therefore an Atom attribute
 - ~left Molnums alone but I feel that this should also be a SegmentGroup attribute, like Moltypes? Or even a SegmentAttr as the code seems to start a new segment for each gmx molecule (pinging @jbarnoud)~
 - made Molnums have default target classes, thereby adding it as a SegmentGroup attribute
   - Molnums cannot be a SegmentAttr [per @jbarnoud's comment](https://github.com/MDAnalysis/mdanalysis/pull/2805#issuecomment-651590532)

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -23,6 +23,7 @@ Fixes
   * TOPParser no longer guesses elements when missing atomic number records
     (Issues #2449, #2651)
   * Testsuite does not any more matplotlib.use('agg') (#2191)
+  * ResidueAttrs now have Atom as a target class by default (#2803)
 
 Enhancements
   * Added the RDKitParser which creates a `core.topology.Topology` object from 
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -18,12 +18,13 @@ The rules for this file:
   * 2.0.0
 
 Fixes
+  * ResidueAttrs now have Atom as a target class by default; ICodes and 
+    Molnums now have default target_classes (#2803, PR #2805)
   * Selections on emtpy AtomGroups now return an empty AtomGroup instead of an
     error (Issue #2765)
   * TOPParser no longer guesses elements when missing atomic number records
     (Issues #2449, #2651)
   * Testsuite does not any more matplotlib.use('agg') (#2191)
-  * ResidueAttrs now have Atom as a target class by default (#2803)
 
 Enhancements
   * Added the RDKitParser which creates a `core.topology.Topology` object from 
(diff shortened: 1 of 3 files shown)


jbarnoud:
`molnum` cannot be at the segment level as they may be more than one `molnum` per segment. A segment is defined for a TPR file as a succession of molecules with the same moltype. For instance, in a system with a protein, a single component lipid bilayer, and water, there will be a segment for the protein, one segment for the lipid and one segment for the water; though there will be many molecules in the lipid and the water segments.

lilyminium:
I see, thanks @jbarnoud! In that case I think `SegmentGroup.molnums` should be alright then, in the same way as `SegmentGroup.resnames` gives a list of all resnames in the segmentgroup?

jbarnoud:
I am surprised it does not already do that. I barely ever use segments so I may have missed it.

lilyminium:
To me it makes sense that all `ResidueAttr`s should be accessible via AtomGroup, ResidueGroup, SegmentGroup, Atom, and Residue, so I've made these the default `target_classes` and changed ICodes and Molnums to use these defaults (all other ResidueAttrs already manually specified these targets)

PR #2810, opened 2020-07-01
Remove docs only building for develop

lilyminium:
Fixes #

 - It may be convenient to build docs for more than just `develop`, e.g. 1.0.1 and `master`
 - This removes the branch check
 - could be backported to 1.0.1 for docs there

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -136,9 +136,10 @@ after_success:
   # turn off blocking as it causes large writes to stdout to fail 
   # (see https://github.com/travis-ci/travis-ci/issues/4704)
   - |
-    if [[ ${TRAVIS_PULL_REQUEST} == "false" ]] && [[ ${BUILD_DOCS} == "true" ]] && [[ ${TRAVIS_BRANCH} == ${GH_DOC_BRANCH} ]]; then
+    if [[ ${TRAVIS_PULL_REQUEST} == "false" ]] && [[ ${BUILD_DOCS} == "true" ]] ; then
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -136,9 +136,10 @@ after_success:
   # turn off blocking as it causes large writes to stdout to fail 
   # (see https://github.com/travis-ci/travis-ci/issues/4704)
   - |
-    if [[ ${TRAVIS_PULL_REQUEST} == "false" ]] && [[ ${BUILD_DOCS} == "true" ]] && [[ ${TRAVIS_BRANCH} == ${GH_DOC_BRANCH} ]]; then
```
orbeckst, line 139: Will this trigger when I merge into _any_ branch, e.g., when merging into a feature branch?

If so, then that does not seem to be desirable as we would quickly have all kinds of weird doc branches (docs for "fix-little-thing-on-other-issue-branch"....?)

lilyminium: I think it should only run for `master` and `develop` at present due to this section:

https://github.com/MDAnalysis/mdanalysis/blob/343e10a257690e4312673f2a0fa4703b31b2e223/.travis.yml#L4-L8

But that is quite far away from the doc building so I can add a comment.

lilyminium: Although we might also want to keep doc building separate from all the branches we want to test in the future. 🤔 In that case checking against an array of branches is easier to keep distinct

PR #2812, opened 2020-07-01
Fix InterRDF_s when given density=True

### LINKED ISSUE #2811: Wrong rdf values from InterRDF_s when given 'density=True'
## Expected behavior ##

If we set `density=True`, the results from `rdf.InterRDF_s` should be the same as `rdf.InterRDF` when given a single-atom to single-atom pair.

See also https://github.com/MDAnalysis/pmda/issues/120

## Actual behavior ##

The results from `rdf.InterRDF_s` will be 1/N of `rdf.InterRDF`.
Where N = nA * nB, and nA, nB are the number of atoms in atomgroup A and atomgroup B.

``` python
import MDAnalysis as mda
from MDAnalysisTests.datafiles import GRO_MEMPROT, XTC_MEMPROT
from pmda.rdf import InterRDF_s
from MDAnalysis.analysis import rdf

u = mda.Universe(GRO_MEMPROT, XTC_MEMPROT)

s1 = u.select_atoms('name ZND and resid 289')
s2 = u.select_atoms('(name OD1 or name OD2) and resid 51 and sphzone 5.0
                         (resid 289)')
s3 = u.select_atoms('name ZND and (resid 291 or resid 292)')
s4 = u.select_atoms('(name OD1 or name OD2) and sphzone 5.0 (resid 291)')
ags = [[s1, s2], [s3, s4]]

rdfs = InterRDF_s(u, ags, density=True)
rdfs.run()


ref_s1 = u.select_atoms('name ZND and resid 289')
ref_s2 = u.select_atoms('name OD1 and resid 51 and sphzone 5.0 (resid 289)')
ref_rdf = rdf.InterRDF(ref_s1, ref_s2, density=True)
ref_rdf.run()

from numpy.testing import assert_almost_equal

assert_almost_equal(rdf_ref.rdf, rdfs[0][0][0])

```

MDAnalysis1.0.0
- Which version of Python?
python 3.6
- Which operating system?
Ubuntu 18.04

VOD555:
Fixes #2811 

 - Fix number of atom pairs in InterRDF_s.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -14,7 +14,7 @@ The rules for this file:
 
 ------------------------------------------------------------------------------
 ??/??/?? tylerjereddy, richardjgowers, IAlibay, hmacdope, orbeckst, cbouy,
-         lilyminium, daveminh, jbarnoud, yuxuanzhuang
+         lilyminium, daveminh, jbarnoud, yuxuanzhuang, VOD555
 
   * 2.0.0
 
@@ -24,11 +24,12 @@ Fixes
   * TOPParser no longer guesses elements when missing atomic number records
     (Issues #2449, #2651)
   * Testsuite does not any more matplotlib.use('agg') (#2191)
-  * In ChainReader, read_frame does not trigger change of iterating position. 
+  * In ChainReader, read_frame does not trigger change of iterating position.
     (Issue #2723, PR #2815)
   * TRZReader now checks `n_atoms` on reading. (Issue #2817, PR #2820)
   * TRZWriter now writes `n_atoms` to the file. (Issue #2817, PR #2820)
-
+  * fixed InterRDF_s function now gives correct results if density=True.
+    (Issue #2811, PR #2812)
 
 Enhancements
   * Added the RDKitParser which creates a `core.topology.Topology` object from
@@ -44,7 +45,7 @@ Enhancements
 Changes
   * deprecated NumPy type aliases have been replaced with their actual types
     (see upstream NumPy PR 14882), and our pytest configuration will raise an
-    error for any violation when testing with development NumPy builds 
+    error for any violation when testing with development NumPy builds
     (`1.20.0`)
   * Changes development status from Beta to Mature (Issue #2773)
   * Removes deprecated MDAnalysis.analysis.hole, please use
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -28,8 +28,10 @@ Fixes
     (Issue #2723, PR #2815)
   * TRZReader now checks `n_atoms` on reading. (Issue #2817, PR #2820)
   * TRZWriter now writes `n_atoms` to the file. (Issue #2817, PR #2820)
-  * fixed InterRDF_s function now gives correct results if density=True.
-    (Issue #2811, PR #2812)
+  * rdf.InterRDF_s density keyword documented and now gives correct results for
+    density=True; the keyword was available since 0.19.0 but with incorrect
+    semantics and not documented and did not produce correct results (Issue
+    #2811, PR #2812)
 
 Enhancements
   * Added the RDKitParser which creates a `core.topology.Topology` object from
--- a/package/MDAnalysis/analysis/rdf.py
+++ b/package/MDAnalysis/analysis/rdf.py
@@ -46,6 +46,8 @@ The radial cumulative distribution function is
 
 and the average number of :math:`b` particles within radius :math:`r`
 
+.. _equation-countab:
+
 .. math::
 
    N_{ab}(r) = \rho G_{ab}(r)
@@ -55,6 +57,13 @@ compute, for instance, coordination numbers such as the number of neighbors in
 the first solvation shell :math:`N(r_1)` where :math:`r_1` is the position of
 the first minimum in :math:`g(r)`.
 
+In :class:`InterRDF_s`, we provide an option `density`. When `density` is
+`True`, it will return the RDF :math:`g_{ab}(r)`; when `density` is False`,
+it will return the density of particle :math:`b` in a shell at distance
+:math:`r` around a :math:`a` particle, which is
+
+.. math::
+   n_{ab}(r) = \rho g_{ab}(r)
 
 .. _`pair distribution functions`:
    https://en.wikipedia.org/wiki/Pair_distribution_function
@@ -68,7 +77,7 @@ Average radial distribution function
 :class:`InterRDF` is a tool to calculate average radial distribution functions
 between two groups of atoms. Suppose we have two AtomGroups ``A`` and
 ``B``. ``A`` contains atom ``A1``, ``A2``, and ``B`` contains ``B1``,
-``B2``. Give ``A`` and ``B`` to class:`InterRDF`, the output will be the
+``B2``. Given ``A`` and ``B`` to :class:`InterRDF`, the output will be the
 average of RDFs between ``A1`` and ``B1``, ``A1`` and ``B2``, ``A2`` and
 ``B1``, ``A2`` and ``B2``. A typical application is to calculate the RDF of
 solvent with itself or with another solute.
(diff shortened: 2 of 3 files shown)


package/MDAnalysis/analysis/rdf.py
```diff
@@ -342,14 +342,11 @@ def _conclude(self):
...
         for i, (ag1, ag2) in enumerate(self.ags):
             # Number of each selection
-            nA = len(ag1)
-            nB = len(ag2)
-            N = nA * nB
             indices.append([ag1.indices, ag2.indices])
 
             # Average number density
             box_vol = self.volume / self.n_frames
-            density = N / box_vol
```
orbeckst, line 352: Can you quickly explain again how the normalization needs to be calculated? What's the equation?

orbeckst: By equation I mean the whole equation, something like the g_ab(r) at the top https://www.mdanalysis.org/docs/documentation_pages/analysis/rdf.html

orbeckst:
Btw, there's an error in the reST for `Suppose we have two AtomGroups A and B. A contains atom A1, A2, and B contains B1, B2. Give A and B to class:InterRDF, t` (the "class" is missing a leading colon), see https://www.mdanalysis.org/docs/documentation_pages/analysis/rdf.html#average-radial-distribution-function 

Could you please fix it while you're at it? Thanks.

package/MDAnalysis/analysis/rdf.py
VOD555: Where should I put the equation, at the top below the g_ab(r) or in the texts for InterRDF_s?

orbeckst:
Add a section or paragraph at the top where you explain InterRDF_s with the equation.

orbeckst:
To be used for squashed commit message
```
* Fixes #2811
* Fix normalization for InterRDF_s(..., density=True)
* add test to compare InterRDF and InterRDF_s
* add description for density option
* update CHANGELOG
```

orbeckst:
@VOD555, I'll quickly add the doc fix ups while I am merging...

package/CHANGELOG
```diff
@@ -24,11 +25,14 @@ Fixes
...
-  * In ChainReader, read_frame does not trigger change of iterating position. 
+  * In ChainReader, read_frame does not trigger change of iterating position.
     (Issue #2723, PR #2815)
   * TRZReader now checks `n_atoms` on reading. (Issue #2817, PR #2820)
   * TRZWriter now writes `n_atoms` to the file. (Issue #2817, PR #2820)
-
+  * rdf.InterRDF_s density keyword documented and now gives correct results for
+    density=True; the keyword was available since 0.19.0 but with incorrect
+    semantics and not documented and did not produce correct results (Issue
+    #2811, PR #2812)
```
orbeckst, line 35: I changed the meaning of the `density` keyword so that it makes more sense. Given that we never documented it, this should be ok and I will classify it as a fix so that it can go into 1.0.1 PR #2798

package/MDAnalysis/analysis/rdf.py
```diff
@@ -55,6 +59,15 @@
 the first solvation shell :math:`N(r_1)` where :math:`r_1` is the position of
 the first minimum in :math:`g(r)`.
 
+In :class:`InterRDF_s`, we provide an option `density`. When `density` is
+``False``, it will return the RDF :math:`g_{ab}(r)`; when `density` is
+``True``, it will return the density of particle :math:`b` in a shell at
+distance :math:`r` around a :math:`a` particle, which is
```
orbeckst, line 65: Note that `density=False` now means "calculate g(r)" and `density=True` means "calculate density".

package/MDAnalysis/analysis/rdf.py
```diff
@@ -77,28 +90,86 @@
...
+
+   .. attribute:: rdf
+
+      :class:`list` of the site-specific :ref:`radial distribution
+      functions<equation-gab>` or :ref:`density
+      functions<equation-nab>` for the :attr:`bins`. The list contains
+      ``len(ags)`` entries. Each entry for the ``i``-th pair ``[A, B]
+      = ags[i]`` in `ags` is a :class:`numpy.ndarray` with shape
+      ``(len(A), len(B))``, i.e., a stack of RDFs. For example,
+      `rdf[i][0, 2]` is the RDF between atom ``A[0]`` and ``B[2]``.
```
orbeckst, line 153: @VOD555 please check that this description is correct

package/MDAnalysis/analysis/rdf.py
```diff
@@ -77,28 +90,86 @@
...
+AtomGroup, ``[[A, B], [C, D], ...]``. Given the same ``A`` and ``B`` to
+:class:`InterRDF_s`, the output will be a list of individual RDFs between
+``A1`` and ``B1``, ``A1`` and ``B2``, ``A2`` and ``B1``, ``A2`` and ``B2`` (and
+similarly for ``C`` and ``D``). These site-specific radial distribution
+functions are typically calculated if one is interested in the solvation shells
+of a ligand in a binding site or the solvation of specific residues in a
+protein. A common use case is to choose ``A`` and ``C`` to be AtomGroups that
+only contain a single atom and ``W`` all solvent molecules: ``InterRDF_s(u,
+[[A, W], [B, W]])`` will then produce the RDF of solvent around atom ``A[0]``
+and around atom ``B[0]``.
```
orbeckst, line 128: @VOD555 I think that this sentence is not really correct: if we choose `W` to be all solvent molecules then we will get one RDF for _each solvent molecule_ – right? So the user still need to average over `rdf[i][0, :]`. The memory requirements might be very high for including hundreds of thousands of solvent molecules so I am not sure if we should recommend it here.

package/MDAnalysis/analysis/rdf.py
```diff
@@ -342,31 +434,35 @@ def _conclude(self):
...
         self.indices = indices
 
     def get_cdf(self):
-        """Calculate the cumulative distribution functions (CDF) for all sites.
-        Note that this is the actual count within a given radius, i.e.,
-        :math:`N(r)`.
+        r"""Calculate the cumulative counts for all sites.
+
+        This is the :ref:`cumulative count<equation-countab>` within a given
+        radius, i.e., :math:`N_{ab}(r)`.
```
orbeckst, line 455: Why are we calculating the counts and calling it CDF? This is confusing... 

I made it clear in the text but still: why did we do this instead of having separate methods that just do the thing that they are named after?

testsuite/MDAnalysisTests/analysis/test_rdf_s.py
```diff
@@ -90,15 +93,24 @@ def test_double_run(rdf):
...
 def test_cdf(rdf):
     rdf.get_cdf()
     assert rdf.cdf[0][0][0][-1] == rdf.count[0][0][0].sum()/rdf.n_frames
 
 
 @pytest.mark.parametrize("density, value", [
-    (True, 13275.775440503656),
-    (False, 0.021915460340071267)])
-
+    (None, 26551.55088100731),    # default, like False (no kwarg, see below)
```
orbeckst, line 103: This will check that the default is still "calculate g(r)".

package/MDAnalysis/analysis/rdf.py
VOD555: So we still need another PR to change the default `density`, right?

package/MDAnalysis/analysis/rdf.py
VOD555: It's correct.

package/MDAnalysis/analysis/rdf.py
VOD555: Yes, it will return one RDF for each solvent molecule. Maybe we can have another option to control whether to average over `rdf` for each pairs when doing the calculation, but for this version, I think we should not recommend it here.

package/MDAnalysis/analysis/rdf.py
VOD555: Yes, we should have seperate functions to give CDF and cummulative N.

package/MDAnalysis/analysis/rdf.py
orbeckst: No, this PR switched the behavior.

package/MDAnalysis/analysis/rdf.py
orbeckst: Ok, I remove the sentence.

package/MDAnalysis/analysis/rdf.py
orbeckst: Ok, new issue.

package/MDAnalysis/analysis/rdf.py
orbeckst: thanks

package/MDAnalysis/analysis/rdf.py
orbeckst: See comment above https://github.com/MDAnalysis/mdanalysis/pull/2812#discussion_r453102797

package/MDAnalysis/analysis/rdf.py
orbeckst: @VOD555 can you raise issues for MDAnalysis and PMDA, please?

orbeckst:
I rewrote the history so that it can be nicely merged.

PR #2815, opened 2020-07-02
Chainreader Refactoring

### LINKED ISSUE #2814: ChainReader read_frame does not update iteration
## Expected behavior ##

ChainReader functions as other readers.

```python
u = mda.Universe(GRO, [XTC, XTC])
u.trajectory[5]
u.trajectory.next()
< Timestep 6 with unit cell dimensions [79.93229 79.93229 79.93228 60.      60.      90.     ]>
```

## Actual behavior ##

`__next__` function and `_read_frame` are disconnected.

```python
#  XTCReader
u = mda.Universe(GRO, XTC)
u.trajectory[5]
u.trajectory.next()
< Timestep 6 with unit cell dimensions [79.93229 79.93229 79.93228 60.      60.      90.     ]>

# ChainReader
u = mda.Universe(GRO, [XTC, XTC])
u.trajectory[5]
u.trajectory.next()
< Timestep 1 with unit cell dimensions [79.93229 79.93229 79.93228 60.      60.      90.     ]>
```

- Which operating system? Ubuntu 20.08

yuxuanzhuang:
Fixes #2814

 - new `_read_frame` and `__iter__` are connected.
 - remove `_chained_iterator`.
 - add `__next__` to `ChainReader`.
 - add an extra attribute `current_frame` to internally monitor absolute current frame.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/coordinates/chain.py
```diff
@@ -561,17 +557,12 @@ def _read_frame(self, frame):
...
 
     def _chained_iterator(self):
         """Iterator that presents itself as a chained trajectory."""
         self._rewind()  # must rewind all readers
-        for i in range(self.n_frames):
-            j, f = self._get_local_frame(i)
-            self.__activate_reader(j)
-            self.ts = self.active_reader[f]
-            self.ts.frame = i
-            yield self.ts
```
orbeckst, line 574: Then delete all `_chained_reader()` – at least in the current form it does not make much sense.

package/MDAnalysis/coordinates/chain.py
```diff
@@ -561,17 +557,12 @@ def _read_frame(self, frame):
         # update Timestep
         self.ts = self.active_reader.ts
         self.ts.frame = frame  # continuous frames, 0-based
+        self._current_frame = frame
```
orbeckst, line 560: Possibly make it a double-underscore attribute `__current_frame`? How bad would it be if a user changed it?

package/MDAnalysis/coordinates/chain.py
```diff
@@ -44,7 +44,6 @@
 import warnings
 
 import os.path
-import itertools
```
orbeckst, line 47: Hah – I didn't realize that we weren't using it here anymore... the original code had a `itertools.chain([... readers ...])` at its heart.

package/MDAnalysis/coordinates/chain.py
```diff
@@ -656,3 +643,17 @@ def _apply_transformations(self, ts):
...
+            self.__activate_reader(j)
+            self.ts = self.active_reader[f]
+            self.ts.frame = self._current_frame + 1
+            self._current_frame += 1
+            return self.ts
+        else:
+            raise StopIteration()
+
+    def next(self):
+        return self.__next__()
```
kain88-de, line 659: The `next` method is not needed anymore. See [pep-3114](https://www.python.org/dev/peps/pep-3114/). I know we are not very good in following this pattern.

testsuite/MDAnalysisTests/coordinates/test_chainreader.py
```diff
@@ -96,6 +96,11 @@ def test_frame_numbering(self, universe):
         universe.trajectory[98]  # index is 0-based and frames are 0-based
         assert_equal(universe.trajectory.frame, 98, "wrong frame number")
 
+    def test_next_after_frame_numbering(self, universe):
+        universe.trajectory[98]  # index is 0-based and frames are 0-based
+        universe.trajectory.next()
+        assert_equal(universe.trajectory.frame, 99, "wrong frame number")
```
kain88-de, line 102: this failed previously? Wow.

package/MDAnalysis/coordinates/chain.py
yuxuanzhuang: The next frame ts will start from the newly set value, but the current ts will not change. Just double-underscored it to prevent from confusion.

package/MDAnalysis/coordinates/chain.py
orbeckst: If we backport this fix to 1.0.1 #2798 , will we need to add `next()` back for Python 2.7?

orbeckst:
Not really sure why the CI is not running on this PR....

package/MDAnalysis/coordinates/chain.py
yuxuanzhuang: `next()` still functions here from `ProtoReader` definition.

kain88-de: Than we should leave the next function and remove it once from all readers.

orbeckst:
Thanks a lot @yuxuanzhuang !

Issue #2450, opened 2020-01-16
Code formatting, style guide enforcement and static cheks


RMeli:
I recently started using code formatting (with [black](https://black.readthedocs.io/en/stable/)), style guide enforcement (with [flake8](http://flake8.pycqa.org/en/latest/index.html#)) and static code checks (with [mypy](http://mypy-lang.org/)). I tried to apply them to MDAnalysis for fun and came across to many warnings.

**Describe the solution you'd like**

I think it would be great to format MDA with [black](https://black.readthedocs.io/en/stable/) (or other tools) and then use  [flake8](http://flake8.pycqa.org/en/latest/index.html#) and [mypy](http://mypy-lang.org/) (or other tools) to spot and fix possible problems.

**Describe alternatives you've considered**

There are other tools that can be considered: [pylint](https://pylint.readthedocs.io/en/latest/), [isort](https://pypi.org/project/isort/), [ypaf](https://github.com/google/yapf), ...

**Additional context**

Many errors come from `MDAnalysis/migration/`.

You can have a look at [this commit](https://github.com/RMeli/mdanalysis/commit/2369d53615d15509f07db2fabb085adf6edf57cf) to see the result of automatic code formatting with [black](https://black.readthedocs.io/en/stable/).

richardjgowers:
We had (or have?) pylint on, but it sometimes throws annoying errors which stop a PR merging.

i think with something like black, we'd have to have every contributor follow the style, which would mean either a) instructions for everyone on how to apply it easily or b) some sort of CI system which fixes up the formatting on merging.  Both seem like they'd add ongoing effort for not much real gain?

RMeli:
@richardjgowers I just wanted to spark the discussion but I agree the ratio benefits/effort might be small. However, #2409 and #2411 could be easily spotted by static checks. I also have found some indentation errors here and there that will make the code fail when executed.

Automatic code formatting (with black or other) is useful only if used with pylint or flake8, so that it removes most of the useless formatting errors. (It might reduce commit diffs if everyone uses it, but this is a very marginal benefit). However, it's really easy to install (`pip install black`) and apply (`black .`).

jbarnoud:
I am all in favour of linters and static analyzers. I am considering adding type hints when we would have dropped python 2. I am less keen on formating tools, though: the way I format my code is often deliberate and, sometimes, going against the convention helps with readability or other things.

RMeli:
I think adding type hints will be a great! 

[...]

PR #2816, opened 2020-07-03
Add Lipid Enrichment Index analysis

lilyminium:
Fixes #

 - add analysis class to calculate a lipid enrichment index

This uses a different way to find membrane leaflets than the LeafletFinder. I haven't benchmarked it but it removes a tiny bit of thinking required (unless you need to change the delta, I guess).

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [ ] CHANGELOG updated?


package/MDAnalysis/analysis/leaflet.py
```diff
@@ -322,3 +322,255 @@ def optimize_cutoff(universe, select, dmin=10.0, dmax=20.0, step=0.5,
...
+                                       for h in self.leaflet_headgroups]
+
+        # set up results
+        self.residue_counts = [np.zeros((len(r), self.n_frames)) for r in lres]
+        self.total_counts = np.zeros((self.n_leaflets, self.n_frames))
+        self.leaflet_ids = ids = [getattr(r, self.attrname) for r in lres]
+        self.leaflets = [dict.fromkeys(np.unique(r)) for r in ids]
+        self.leaflets_summary = [dict.fromkeys(np.unique(r)) for r in ids]
+
+    def _single_frame(self):
```
orbeckst, line 507: You don't rebuild leaflet groups at every step. For really long (or cg) sims you might have lipid flip-flops and you start assigning lipids wrongly.  At least document.

One could also have a setting that leaflets are recomputed every n steps.

package/doc/sphinx/source/documentation_pages/references.rst
```diff
@@ -180,6 +180,11 @@ If you use the Mean Squared Displacement analysis code in
 
 .. _`10.21105/joss.00877`: https://doi.org/10.21105/joss.00877
 
+.. [Corradi2018] Corradi, V. et al. Lipid–Protein Interactions Are Unique Fingerprints for Membrane Proteins. ACS Cent Sci 4, 709–717 (2018). doi:`10.1021/acscentsci.8b00143`_
```
orbeckst, line 183: Please include all authors... bytes aren't that expensive nowadays, we can afford all authors.

testsuite/MDAnalysisTests/analysis/test_leaflet.py
```diff
@@ -173,3 +176,185 @@ def test_write_selection(universe, lipid_heads, tmpdir):
 def test_component_index_is_not_none(universe, lipid_heads):
     lfls_ag = LeafletFinder(universe, lipid_heads, cutoff=15.0, pbc=True)
     assert_almost_equal(len(lfls_ag.groups(component_index=0)), 180, decimal=4)
+
+
+class BaseTestLipidEnrichment:
+    pytest.importorskip('sklearn', minversion='0.23.1')
```
orbeckst, line 182: This works _inside_ classes? Does it really skip just the class and not the whole module?

testsuite/MDAnalysisTests/analysis/test_leaflet.py
```diff
@@ -173,3 +176,185 @@ def test_write_selection(universe, lipid_heads, tmpdir):
...
+
+    n_u_POPE = 113
+    n_l_POPE = 108
+    n_u_POPG = 28
+    n_l_POPG = 27
+    n_upper = 113+28
+    n_lower = 108+27
+    n_lipids = 276
+
+    def test_leaflet_finder(self, lipen):
```
orbeckst, line 242: Could be a separate test for a separate alternative LeafletFinder.

package/MDAnalysis/analysis/leaflet.py
```diff
@@ -322,3 +322,255 @@ def optimize_cutoff(universe, select, dmin=10.0, dmax=20.0, step=0.5,
...
+                leaflets[i].append(res)
+            self.leaflet_residues = lres = [np.array(x).sum() 
+                                            for x in leaflets]
+            self.leaflet_residues.sort(key=lambda x: x.center_of_geometry()[-1],
+                                       reverse=True)
+            self.leaflet_headgroups = [(self.headgroups & r.atoms) 
+                                       for r in lres]
+        else:
+            self.leaflet_residues = [self.residues]
+            self.leaflet_headgroups = [self.headgroups]
```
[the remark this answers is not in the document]
orbeckst: If you can make the API similar to the existing LeafletFinder then one could also just add a `method=...` kwarg to LeafletFinder and either use the classic one or the spectral_clustering.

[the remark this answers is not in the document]
lilyminium: > In any case, why hide it? Also, then you can test it independently.

I guess it felt very short to be its own class, but I can move it into LeafletFinder. 

> What's the difference between the build-in spectral-clustering leafletfinder and the old one? What's the advantage?

LeafletFinder constructs a networkx graph from a matrix of contacts, with a default cutoff of `cutoff=15`. The groups it returns are every connected component in the graph, sorted by size. It means you should seriously consider your `cutoff` parameter in case you accidentally make little lipid islands, or connect all the lipids; this directly affects whether the `n` biggest groups are true representations of your leaflets.

The spectral clustering method just clusters a similarity matrix into `n` leaflets, where you choose `n`. There's a `delta` parameter that indirectly affects the clustering because it's used to transform the distance matrix into the similarity matrix, but I haven't found the need to change it for the simulations we've used it on. I like this method because I have to do less thinking and I'm guaranteed to get `n` leaflets. If, e.g., cholesterol is floating around in the middle of flip-flopping it's guaranteed to get assigned to one of the leaflets. 

> Is the algorithm your own or from the paper? Would be worthwhile making it citable.

The leaflet-finding is just spectral clustering; the DEI formula is from the paper. I am thinking of putting in a soft potential to weight the distances of the lipids from the protein, though, as I am concerned that the hard cutoff + skipping frames neglects the transient movements of low-population lipids in a way that has an unrepresentatively large effect on the final index.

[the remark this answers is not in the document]
lilyminium: I should mention that I did try some different `delta` values under 1 and it didn't seem to have any effect.

package/MDAnalysis/analysis/leaflet.py
lilyminium: > One could also have a setting that leaflets are recomputed every n steps.

Yeah, I did consider that. I didn't want to cluster on every frame by default. I was thinking instead of computing distance to the center of geometry of the different lipids, which would be less computationally expensive. I would have to change up how I structure the results containers, though, as currently they're numpy arrays where each element represents a lipid :/

package/MDAnalysis/analysis/leaflet.py
```diff
@@ -322,3 +322,255 @@ def optimize_cutoff(universe, select, dmin=10.0, dmax=20.0, step=0.5,
...
+        whether to compute distances only using the headgroup.
+    attrname: str
+        The topology attribute used to group lipid species.
+    """
+
+    def __init__(self, universe, select_protein: str='protein', 
+                 select_residues: str='all', select_headgroup: str='name PO4',
+                 n_leaflets: int=2, count_by_attr:str ='resnames', 
+                 enrichment_cutoff: float=6, delta: float=0.5,
+                 compute_headgroup_only: bool=True, **kwargs):
```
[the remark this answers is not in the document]
orbeckst: I had forgotten about #2450 where there was agreement to start adding type hints once Python 2 was dropped.

package/MDAnalysis/analysis/leaflet.py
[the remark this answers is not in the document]
orbeckst: The similarity matrix clustering of distances is so general that it could just live in the distances module. It can be easily applied to contacts, salt bridges, h-bonds... Then you just call it here with the headgroup COM positions as input.

There is probably a way to speed it up with capped distances because your similarity matrix does not really care about values beyond a few `delta`. 

@ianmkenney and @VOD555 might be interested, too.

package/MDAnalysis/analysis/leaflet.py
orbeckst: For right now I would just note it in the documentation as a potential limitation. If there’s sufficient need then whoever has the greatest need can work on improving it.

lilyminium: Yes, on the other hand, that's probably me by Tuesday 😄

jbarnoud: I got to play with the original code from the reference paper, we often observed flip-flop, especially just near to the proteins. It is especially true when looking at small lipids or, worst, sterols. Accounting for flip-flop is the difficult part in this analysis.

package/MDAnalysis/analysis/leaflet.py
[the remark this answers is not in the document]
jbarnoud: If it works well, it would be beneficial to expose the spectral leaflet finder. At the very least, it would allow to be consistent in the definition of the leaflets between analyses.

jbarnoud:
@mnmelo was involved in the [Corradi *et al.*](https://pubs.acs.org/doi/10.1021/acscentsci.8b00143) that is the reference paper for this pull request. Maybe he would have some comments.

testsuite/MDAnalysisTests/analysis/test_leaflet.py
lilyminium: > This works *inside* classes? Does it really skip just the class and not the whole module?

No, but it would have been nice if it did. Changed to `skipif`

orbeckst:
Please ping when you need me to look at the PR again.

lilyminium:
This has grown a bit; I think I'll split it out into a couple different pull requests (LeafletFinder and LipidCounter are quite separate now).

jbarnoud:
For the record, the original script is available there: http://cgmartini.nl/index.php/tools2/proteins-and-bilayers

lilyminium:
Closing this PR to MDAnalysis as it is now implemented in the external toolkit [lipyds](https://github.com/lilyminium/lipyds).

Issue #2818, opened 2020-07-03
Running AlignTraj on GRO/XTC causes Travis segfault


IAlibay:
## Expected behavior ##

AlignTraj and Transform should not segfault when running Travis.

## Actual behavior ##

Running AlignTraj / Transform on the MDA GRO/XTC testfiles causes Travis to fail on python 3.6-3.8 w/ Linux. Interestingly we cannot reproduce that behaviour locally. See: https://github.com/MDAnalysis/UserGuide/pull/83 (transform), and https://travis-ci.com/github/bigginlab/MDRestraintsGenerator/jobs/357265043 (aligntraj)

Note: I'm not 100% convinced this isn't user error somehow (i.e. Travis setup related), so I'm going to open a test PR here to see if I can reproduce it.

## Code to reproduce the behavior ##

Will work fine for the PSF/DCD pair, but fails on GRO/XTC.

Transform:
```python
import MDAnalysis as mda
from MDAnalysis.tests.datafiles import TPR, XTC
from MDAnalysis import transformations as trans

def test_transform():
    u = mda.Universe(TPR, XTC)
    protein = u.select_atoms('protein')
    align_transform = trans.fit_rot_trans(protein, protein, weights=protein.masses)
    u.trajectory.add_transformations(align_transform)
    assert 1 == 1
```

Aligntraj:
``` python
import MDAnalysis as mda
from MDAnalysis.tests.datafiles import GRO, XTC

@pytest.mark.parametrize('top,traj', [(PSF, DCD), (GRO, XTC)])
def test_aligntraj(tmpdir, top, traj):
    """AlignTraj is failing, so let's test it here"""
    copy_u = mda.Universe(top, traj)
    prealigner = align.AlignTraj(copy_u, copy_u, select="protein and name CA",
                                                 in_memory=True)
    prealigner.run()

    assert 1 == 1
```

## Current version of MDAnalysis ##

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) 1.0.0 and 0.20.1 tested
- Which version of Python (`python -V`)? 3.6 - 3.8
- Which operating system? Ubuntu 18.04 & 16.04 LTS (Travis-CI)

lilyminium:
[...]

PR #2819, opened 2020-07-03
[test] aligntraj + transform on XTC file

IAlibay:
Test for #2818 

Apologies for the excess noise here, I'm trying to see if the downstream issue I'm seeing can be replicated here.

 - Adds an aligntraj test using GRO/XTC

Update:

Interestingly Travis here does not seem to fail in the way seen downstream. However, I can reproduce a similar segfault to what the userguide also gets, so I'm going to add the case for where the user guide fails and see if that also "works" here.

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


IAlibay:
Closing as no longer useful, it appears the answer is probably just to ensure we use astropy's CI helpers: https://github.com/MDAnalysis/mdanalysis/issues/2818#issuecomment-653561810. PR for user guide has been updated accordingly.

PR #2820, opened 2020-07-03
TRZReader check for n_atoms

### LINKED ISSUE #2817: TRZReader doesn't complain when wrong topology is provided
## Expected behavior ##

TRZReader raises a error (or warning) when a wrong topology (or n_atoms) is provided to `TRZReader`.

## Actual behavior ##

It passes without complaining during initialization.

```python
u = mda.Universe(TRZ, n_atoms=8180)  # should be 8184
#  pass
u = mda.Universe(PSF, TRZ) #  should be TRZ_psf
#  pass
```
If you check for `dt` or `read_frame`, it will run into error. Maybe add such checks during initialization?
```python
>>> u.trajectory.dt
~/mdanalysis/package/MDAnalysis/coordinates/TRZ.py in _seek(self, nframes)
    387             seeksize = int(seeksize)
    388 
--> 389             self.trzfile.seek(seeksize, 1)
    390 
    391     def _reopen(self):

OSError: [Errno 22] Invalid argument

>>> u.trajectory[0]
~/mdanalysis/package/MDAnalysis/coordinates/base.py in _apply_limits(self, frame)
   1565             frame += len(self)
   1566         if frame < 0 or frame >= len(self):
-> 1567             raise IndexError("Index {} exceeds length of trajectory ({})."
   1568                              "".format(frame, len(self)))
   1569         return frame

IndexError: Index 0 exceeds length of trajectory (0).
```

Iteration, on the other hand, works--with wrong result.
```python
>>> u.trajectory.__next__()
< Timestep 1023497743 with unit cell dimensions [5.6476583e-04 1.2095231e-03 1.3170627e-02 1.0533933e+01 1.7983182e+02
 1.6946487e+02] >
```

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) 2.0.0.dev
- Which operating system? Ubuntu 20.04

yuxuanzhuang:
Fixes #2817 

 - check for `dt` to raise `ValueError` if `n_atoms` is incompatible with the trajectory.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/CHANGELOG
```diff
@@ -25,6 +25,7 @@ Fixes
   * TOPParser no longer guesses elements when missing atomic number records
     (Issues #2449, #2651)
   * Testsuite does not any more matplotlib.use('agg') (#2191)
+  * TRZReader now checks `n_atoms` provided right during initilization.
```
orbeckst, line 28:
```suggestion
  * TRZReader now checks `n_atoms` during initialization (PR  #2820)
```

testsuite/MDAnalysisTests/coordinates/test_trz.py
```diff
@@ -129,6 +129,10 @@ def test_get_writer_2(self, universe, tmpdir):
             assert_equal(isinstance(W, mda.coordinates.TRZ.TRZWriter), True)
             assert_equal(W.n_atoms, 100)
 
+    def test_get_wrong_n_atoms(self):
+        with pytest.raises(ValueError, match=r"Supplied n_atoms *"):
+            mda.Universe(TRZ, n_atoms = 8080)
```
orbeckst, line 134:
```suggestion
            mda.Universe(TRZ, n_atoms=8080)
```

package/MDAnalysis/coordinates/TRZ.py
```diff
@@ -221,6 +226,14 @@ def __init__(self, trzfilename, n_atoms=None, **kwargs):
...
         self._read_next_timestep()
 
+        try:
+            self._get_dt()
+        except OSError:
+            raise ValueError("Supplied n_atoms {} is incompatible "
+                             "with provided trajectory file. "
+                             "Maybe `topology` is wrong?".format(self.n_atoms))
+
+
```
orbeckst, line 236: It is not clear why `_get_dt()` will fail with the wrong `n_atom`. What's the stack trace and where does the ValueError come from? (I assume this relies on being able to do `self.next()`, which actually fills the `ts`.... but then why does `self._read_next_timestep()` on the preceding line work? 🤷 )

This might work and fix an error but I think we should rather check sooner and in a more transparent manner if at all possible.

yuxuanzhuang: `_read_next_timestep` (used by `__next__`) only fills the `ts` by reading a chunk from the TRZ file using `np.fromfile` with the provided `frame_contents` (defined by `n_atoms`) no matter it is right or wrong.

`read_frame`, and other related functions (e.g. `_get_dt` will go next and read frame back), on the other hand, will use `self._seek`. With an abnormal `trajectory.next().frame`, This will always reach an error.

yuxuanzhuang:
Given my current knowledge on `TRZ` files (which is none), I cannot really come up with an ideal checking method. Maybe check if `self.next()` return a nonphysical frame? Or what may happen if `TRZ` only has one frame. 

Besides, I checked there are still other magic `n_atoms` numbers that will pass the "dt" test, may leading to false positive cases. 

Other people with more knowledge are welcome to step in.

IAlibay:
> Given my current knowledge on `TRZ` files (which is none), I cannot really come up with an ideal checking method. Maybe check if `self.next()` return a nonphysical frame? Or what may happen if `TRZ` only has one frame.
> 
> Besides, I checked there are still other magic `n_atoms` numbers that will pass the "dt" test, may leading to false positive cases.
> 
> Other people with more knowledge are welcome to step in.

I'm also not super well read on the TRZ format, but just looking at the code, is there any chance we could evaluate the return value of data['natoms']?

yuxuanzhuang:
Looks good? Also spot that `TRZWriter` does not write `n_atoms` to the file. Maybe a bug or intentionally?

package/MDAnalysis/coordinates/TRZ.py
```diff
@@ -141,7 +140,8 @@ class TRZReader(base.ReaderBase):
        Extra data (Temperature, Energies, Pressures, etc) now read
        into ts.data dictionary.
        Now passes a weakref of self to ts (ts._reader).
-
+    .. versionchanged:: 2.0.0
```
IAlibay, line 143: I'm pretty sure this won't work, you usually need 2 lines between the text and the versionchanged

testsuite/MDAnalysisTests/coordinates/test_trz.py
```diff
@@ -129,6 +129,10 @@ def test_get_writer_2(self, universe, tmpdir):
             assert_equal(isinstance(W, mda.coordinates.TRZ.TRZWriter), True)
             assert_equal(W.n_atoms, 100)
 
+    def test_get_wrong_n_atoms(self):
+        with pytest.raises(ValueError, match=r"Supplied n_atoms *"):
```
IAlibay, line 133: note,  I dont think you need to have the "*" in match calls, it essentially does a `string in string` type of comparison.

package/MDAnalysis/coordinates/TRZ.py
yuxuanzhuang: It's right after a (quite long) `.. versionchanged`:) The doc lgtm.

yuxuanzhuang:
I just want to discuss the necessity of raising an error, i.e. the TRZ file is still valid if the natoms written in the file is wrong (e.g. 0), which is case for the previous TRZWriter. Maybe a warning will suffice?

package/MDAnalysis/coordinates/TRZ.py
```diff
@@ -141,7 +140,8 @@ class TRZReader(base.ReaderBase):
...
+       Now checks for the right `n_atoms` during initilization.
```
orbeckst, line 144:
```suggestion
       Now checks for the correct `n_atoms` during initialization.
```

package/MDAnalysis/coordinates/TRZ.py
```diff
@@ -159,6 +159,11 @@ def __init__(self, trzfilename, n_atoms=None, **kwargs):
             number of atoms in trajectory, must be taken from topology file!
         convert_units : bool (optional)
             converts units to MDAnalysis defaults
+
+        Raises
+        ------
+        ValueError
+           If wrong `n_atoms`/topology file is provided.
```
orbeckst, line 166:
```suggestion
           If  `n_atoms` or the number of atoms in the topology file do not match the 
           number of atoms in the trajectory.
```

orbeckst:
> I just want to discuss the necessity of raising an error, i.e. the TRZ file is still valid if the natoms written in the file is wrong (e.g. 0), which is case for the previous TRZWriter. Maybe a warning will suffice? Or we still need to find something else for validation.

Let's raise an exception.

Did you fix the writer? If so, add a note to Fixes in the CHANGELOG, please.

package/MDAnalysis/coordinates/TRZ.py
IAlibay: Sorry my bad, I somehow managed to miss the above versionchanged.

package/CHANGELOG
```diff
@@ -28,6 +26,9 @@ Fixes
   * Testsuite does not any more matplotlib.use('agg') (#2191)
   * In ChainReader, read_frame does not trigger change of iterating position. 
     (Issue #2723, PR #2815)
+  * TRZReader now checks `n_atoms` during initilization. (PR #2820)
```
IAlibay, line 29: I might be missing something obvious, but the check is happening in `_read_next_timestep` right? So technically it's not limited to initialization but every step. It might be more accurate here to just say "checks `n_atoms` on reading"?

IAlibay: Could you also add the issue number here?

package/MDAnalysis/coordinates/TRZ.py
```diff
@@ -588,6 +600,7 @@ def _write_next_frame(self, obj):
         out['p1a'], out['p1b'] = 20, 20
         out['nframe'] = ts.frame + 1  # TRZ wants 1 based
         out['ntrj'] = data['step']
+        out['natoms'] = self.n_atoms
```
IAlibay, line 603: Could you also add a test for this too please?

yuxuanzhuang: It's implicitly checked during `test_write_trajectory`. That's where I spot the bug in the first place, do we need to do it more explicitly?

IAlibay: was it raising an error previously? If not we should explicitly check that the right n_atoms value is present in the file now.

yuxuanzhuang: It will copy and write to a TRZ file, then read the written file for comparison. So it will raise an error during reading if the written `natoms` is wrong,

IAlibay: Ok yes that works, thanks for walking me through it :)

Issue #2789, opened 2020-06-24
Aggregation analysis


aligh92:
## Is your feature request related to a problem? ##

No! Actually, it was my problem, and I scripted code for that using your package, and then I thought I should appreciate you by contributing my code to your package! 
I was stuck on measuring the aggregate size based on the distance between the molecule's centre of mass. Also, I needed to measure the aggregate properties such as gyration radius, shape and density. As far as I know, there is no analysis for that in your package. GROMACS has "Clustsize" feature to measure the aggregation of the molecules, but it is not able to solve my problem

## Describe the solution you'd like ##

The GROMACS feature, "Clustsize", can measure the number of aggregate with different size based on the distance between the closest atom or a specific atom. However, it is not able to consider the distance of the centre of mass. Also, it can not measure the cluster characterization such as gyration radius or shape throughout the simulation time. Therefore, I decided to write my code. Currently, my code is ready, and I proved that using GROMACS feature. My code can consider different types of distance, as a user request, and the optional distance cut-off. In "Clustsize" the monomer will be considered while my code is flexible in considering or eliminating them. I considered measuring the average gyration radius along the simulation time as a representative property for clusters. User can take that as an example and measure other properties, or I can add them upon your request. 

## Describe alternatives you've considered ##

There might be other software to solve my problem, but as far as I know, this feature is missing in your package and it would be pleasant for the MDAnalysis users to have this opportunity.

## Additional context ##

The code that I am suggesting has been proved by GROMCAS feature, "Clustsize". The results are going to be published as a part of a technical article.

orbeckst:
Hi @aligh92 , the best way to get some more feedback is to create a pull request and show the code. If you have a preprint of your article then also provide a link as well as any other papers you're relying on.

lilyminium:
Side note, radius of gyration currently exists as an example of writing your own analysis using our analysis building blocks, which is to be fair not a good reason to keep it from the main library.

https://userguide.mdanalysis.org/examples/analysis/custom_trajectory_analysis.html

orbeckst:
Actually, there's [AtomGroup.radius_of_gyration()](https://www.mdanalysis.org/docs/documentation_pages/core/topologyattrs.html#MDAnalysis.core.topologyattrs.Masses.radius_of_gyration) (see our [basic example](https://www.mdanalysis.org/pages/basic_example/) which you can wrap with `AnalysisFromFunction`.

[...]

PR #2822, opened 2020-07-03
analyze aggregate size

aligh92:
developed code for issue #2789 i.e to analyze the aggregate size
 - 

PR Checklist
------------
 - [ ] Tests?
 - [x] Docs?
 - [ ] CHANGELOG updated?
 - [x] New code proposal


orbeckst:
Thank you for the PR @aligh92 !

At the moment this still looks like  a notebook or a command line script. It will take considerable effort to turn it into an *MDAnalysis.analysis* module that follows the common layout and style of the other modules. If you want to proceed in this direction you'll need to learn how to write analysis classes with [AnalysisBase](https://www.mdanalysis.org/docs/documentation_pages/analysis/base.html) and how to write tests. The User Guide has a whole section on [Contributing to MDAnalysis](https://userguide.mdanalysis.org/contributing.html) which is geared towards developers.

Alternatively, we could add the script to the [CookBook](https://github.com/MDAnalysis/MDAnalysisCookbook) repository – we would create a _contributed_ directory where you could add it together with documentation.

Finally, we could also create a separate repository under MDAnalysis for you to work in. That's something that would need to be discussed by the development team @MDAnalysis/coredevs .

Tell us what you think.

lilyminium:
The user guide also has a specific tutorial on how to use the AnalysisBase and related tools; it even uses radius of gyration as the example :-)

https://userguide.mdanalysis.org/examples/analysis/custom_trajectory_analysis.html

package/MDAnalysis/analysis/Aggregation_size.py
```diff
@@ -0,0 +1,364 @@
...
+                    globals()["d" + str(j+1)][ii][j_time] = dists[i][j]
+                    ii += 1
+            j_time += 1
+    else:
+        warnings.warn("enter valid Type")
+    GYR = []
+    ij = 0
+    for ts in u.trajectory:
+        u.trajectory[ij]
+        ij += 1
```
jbarnoud, line 183: Here, you are doing a lot of redundant gymnastic. In MDAnalysis, when you iterate over a trajectory, you move the cursor so that the trajectory points at the frame you requested. When you index the trajectory with `u.trajectory[ij]`, you do the same thing. So here, you move to frame 0, `ts` corresponds to frame 0 so `ts.frame` is 0; then you move again to frame 0 by doing `u.trajectory[ij]`; because we sometimes but not always reuse the time step object, `ts` may or may not be the same as previously, but because you moved to the same frame `ts.frame` is 0 anyway, same as `ij`, same as `tss`.

One way of avoiding the redundant operations would be to only do:

```python
for ts in u.trajectory:
    tss = ts.frame
   ...
```

Her you save yourself the hassle and the time it takes to jump back to the frame you are already at.

Tough, I do not think you are using the fact that the trajectory actually advances in this loop, and you only want to run your loop once per frame. Moving around the trajectory is an slow operation, so you should avoid it when you can. Here I would suggest something lighter such as:

```python
for tss in range(len(u.trajectory)):
    ...
```

Note that `tss` is not a good variable name and you should probably use something more explicit like `frame_index`.

jbarnoud:
@orbeckst The cookbook has not been updated for the last 5 years, I doubt anything there still works. I would strongly advocate in favour of archiving the repo and avoiding accumulating untested code at the periphery of our sight as it will for sure end up the same as the cookbook.

aligh92:
@orbeckst  Thank you so much for the quick reply. The code needs changes to become compatible with _MDAnalysis.analysis_ module. In this stage, I would rather learn writing analysis class and test and make the code eligible for _MDAnalysis.analysis_ module other than having code in CookBook repository. Because in the end, I would like to be beneficial and I want my code to be helpful for other users :-).
@lilyminium I really appreciate your comment and guide. I will definitely use the link that you provided. 
@jbarnoud Thank you so much for your encouragement. At this point, I wanted to make sure if the code is beneficial for the package and your comment assure me about that which is fantastic! I really appreciate your comments, and I would be more than happy to get as much as possible feedback. I will try to point out all your comments and go through all learning stuff to make code ready to be a part of _MDAnalysis.analysis_ module. In meanwhile, I appreciate more feedback from you or any other person. 
I will be in touch with you guys. Thanks again for the useful feedback and guides

package/MDAnalysis/analysis/Aggregation_size.py
```diff
@@ -0,0 +1,364 @@
...
+    from MDAnalysis.lib import distances
+    u = mda.Universe(gro_file, xtc_file)
+    molnum = number_of_molecules
+    Limit = cut_off
+    i = 0
+    for it in reversed(range(1, molnum)):
+        i += 1
+        globals()["d" + str(i)] = np.zeros(
+            it * len(u.trajectory)).reshape(it, len(u.trajectory))
+    if Type == "Closest":
```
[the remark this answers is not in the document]
aligh92: I am now using the _single_frame function in the AnalysisBase class. So, I need to run through the frames more than once, I am wondering if I need to consider _single_frame for all loops which are going through frames? if yes how can I accommodate that?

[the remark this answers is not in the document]
jbarnoud: If I recall correctly, your initial logic was to first compute the distance matrix for all the frames, then cluster the the particles into aggregates for all the frames. With `_single_frame` I would recommand to change you logic a little bit. Instead of looping twice over the frames, I would, for each frame, calculate the distance matrix and do the clustering. This is possible because you do not need the information from the other frames when you compute the aggregate for one given frame.

By doing all the frames independently from each other, you avoid looping multiple times over the frames, but you also avoid storing large amount of data that you do not need. By building all the distance matrices at once, you have to store a large amount of data (n*n*f, with n the number of atoms, and f the number of frames) which limit the size of the trajectories you can work with. By doing the frames separately, you only have one distance matrix in memory at a time.

aligh92:
To: @jbarnoud 
cc: @lilyminium @orbeckst, 

Hello, 
I addressed your comments. I am wondering if you can advise me what is missing or what is the next step. 
Best regards, 
Ali

package/MDAnalysis/analysis/Aggregation_size.py
```diff
@@ -0,0 +1,370 @@
...
+        for i in range(1, self._molnum+1):
+            ii = 0
+            for j in range(i+1, self._molnum+1):
+                residue1 = u.select_atoms(f"resid {i}").positions
+                residue2 = u.select_atoms(f"resid {j}").positions
+                dists = distances.distance_array(
+                    residue1, residue2, box=u.universe.dimensions,
+                    backend='OpenMP')
+                d[i][ii] = np.amin(dists)
+                ii += 1
```
orbeckst, line 112: I would have the method return `d` – doesn't come with a performance penalty and makes it easier to understand, even though you're changing `d` in place (why?)

package/MDAnalysis/analysis/Aggregation_size.py
```diff
@@ -0,0 +1,370 @@
...
+        """
+        self._cut_off = self._cut_off or 3.5
+        for i in range(1, self._molnum+1):
+            ii = 0
+            for j in range(i+1, self._molnum+1):
+                residue1 = u.select_atoms(f"resid {i}").positions
+                residue2 = u.select_atoms(f"resid {j}").positions
+                dists = distances.distance_array(
+                    residue1, residue2, box=u.universe.dimensions,
+                    backend='OpenMP')
```
orbeckst, line 110: Don't hard-code the backend, it might fail when not available. If you need this control, make `backend` a kwarg for init and then use `self.backend` here.

package/MDAnalysis/analysis/Aggregation_size.py
```diff
@@ -0,0 +1,370 @@
...
+    def specific_atoms(self, d):
+        """
+        This function calculates the distance between the selected atom
+        of two molecules.
+        """
+        if (not(atom_1 or atom_2)):
+            raise ValueError("enter valid atom_1 and atom_2")
+        self._cut_off = self._cut_off or 8.5
+        residue1 = u.select_atoms(f"name {atom_1}").positions
+        residue2 = u.select_atoms(f"name {atom_2}").positions
```
orbeckst, line 141: Operate with your `self.atomgroup`

package/MDAnalysis/analysis/Aggregation_size.py
```diff
@@ -0,0 +1,370 @@
...
+            cm_coordintes, cm_coordintes, box=u.universe.dimensions)
+        for j in range(self._molnum):
+            ii = 0
+            for i in range(j+1, self._molnum):
+                d[j+1][ii] = dists[i][j]
+                ii += 1
+
+    def specific_atoms(self, d):
+        """
+        This function calculates the distance between the selected atom
```
orbeckst, line 134: docs

package/MDAnalysis/analysis/Aggregation_size.py
```diff
@@ -0,0 +1,370 @@
...
+        self.d = [0] * self._molnum
+        i = 0
+        for it in reversed(range(1, self._molnum+1)):
+            self.d[i] = np.zeros(it * 1).reshape(it, 1)
+            i += 1
+        self.size_num = [0] * self.n_frames
+
+    def aggregation(self, k, tss, counter, m, cut_off, connected):
+        """
+        This is a subfunction that enables the code to find all
```
orbeckst, line 164: Have a look at LeafletFinder in `leaflet.py` where networkx is used to do something similar. Maybe you can rewrite with networkx, which might be faster.

package/MDAnalysis/analysis/Aggregation_size.py
```diff
@@ -0,0 +1,370 @@
...
+        """
+        aggregate = [0] * self._molnum
+        if self._type == "Closest":
+            self.closest_atom(self.d)
+        elif self._type == "COM":
+            self.center_of_mass(self.d)
+        elif self._type == "Specific":
+            self.specific_atoms(self.d)
+        else:
+            raise ValueError("enter valid Type")
```
orbeckst, line 206: This check should be in init – when you call `_single_frame()` everything should be ok.

package/MDAnalysis/analysis/Aggregation_size.py
```diff
@@ -0,0 +1,370 @@
...
+            # this "if" will restrict double calculation of the same molecule
+            # to increase efficincy of code
+            if counter not in P_Level:
+                # define the ones that are attached
+                # to a molecule in first level
+                bool_sector = self.d[counter] < self._cut_off
+                Coordinates = self.d[counter][bool_sector]
+                index1 = np.where(bool_sector)
+                # list the connected molecules to a molecule in first level
+                listOfCoordinates = list(zip(index1[0], index1[1]))
```
orbeckst, line 226: When I see `list` and `tolist()` I wonder if this code can be written with arrays instead. Big lists are typically slow and the wrong data structure for coordinates.

package/MDAnalysis/analysis/Aggregation_size.py
```diff
@@ -0,0 +1,370 @@
...
+                    for i in set(mylist):
+                        if i in (list(P_Level)):
+                            mylist.remove(i)
+                    # substitute new "conncected" with previous one
+                    connected = np.array(mylist)
+                aggregate[counter-1] = (
+                    list(dict.fromkeys(aggregate[counter-1])))
+                Group = np.append(Group, counter)
+            else:
+                continue
```
orbeckst, line 272: Not very readable... move up

package/MDAnalysis/analysis/Aggregation_size.py
```diff
@@ -0,0 +1,370 @@
...
+        # len(Group) will be the number of clusters
+        Group = []
+        # this for will go through all molecules to
+        # check their connection with others
+        for counter in range(1, self._molnum):
+            # aggergate parameter include the molecules in each cluster
+            aggregate[counter-1] = [counter]
+            # this "if" will restrict double calculation of the same molecule
+            # to increase efficincy of code
+            if counter not in P_Level:
```
orbeckst, line 219:
```suggestion
            if counter not in P_Level:
                continue
```
Make the loop more readable with early continue.

Then de-dent the rest of the code.

package/MDAnalysis/analysis/Aggregation_size.py
```diff
@@ -0,0 +1,370 @@
...
+            index1 = np.where(bool_sector)
+            listOfCoordinates = list(zip(index1[0], index1[1]))
+            connect = ([j[0] for j in listOfCoordinates]
+                       + connected[m] + 1)
+            Connected_NLevel = np.array(connect)
+        else:
+            Connected_NLevel = []
+        return Connected_NLevel
+
+    def _single_frame(self):
```
orbeckst, line 185: This function is complicated and long. See if you can break it up into logical units and use methods for different parts to bring out the logic.

I also see lists and various casts with `map` and wonder if you are using the right data structures.

package/MDAnalysis/analysis/Aggregation_size.py
```diff
@@ -0,0 +1,370 @@
...
+#import os
+#os.chdir(r"path")
+#gro_file = "mdrun.gro"
+#xtc_file = "mdrun.xtc"
+#u = mda.Universe(gro_file, xtc_file)
+#Mol = u.select_atoms('resname AsphC')
+#rog_base = Aggregation_size(Mol, number_of_molecules = 50, Type = "COM", 
+#no_monomer = True, Gyr_calc=True).run()
+#print(rog_base.results)
+#rog_base.df
```
orbeckst, line 370: remove – you'll have to write tests and something like this can go in the tests

package/MDAnalysis/analysis/Aggregation_size.py
```diff
@@ -0,0 +1,370 @@
...
+        "aggregate" subfunction (to be efficient)
+        4- Check and merge two aggregates as one aggregate if the distance
+        between any element of them is less than cut_off
+        5- Eliminate monomers (optional)
+        6- measure average gyration radius (optional)
+        7- measure the number of aggregates with different size
+        """
+        aggregate = [0] * self._molnum
+        if self._type == "Closest":
+            self.closest_atom(self.d)
```
lilyminium, line 200: Relevant to @orbeckst's comment above, I would have an umbrella function `d = self._get_distances()` which you assign in `__init__`:

```python
def __init__(self, method="closest", ...):
    if method == "closest":
        self._get_distances = self._get_distances_by_closest_atoms
    ... # etc
```

Renaming your other functions with action words makes it clearer what they do, and returning the `d` distance array means you don't overwrite it every frame.

If you want to save your distances, you can prepare a placeholder array in `_prepare`:

```
self._distances = np.zeros((self.n_frames, self.n_residues, self.n_residues))
```

or if you would like to flatten it to save space:

```
n_triangle = self.n_residues * (self.n_residues-1) / 2
self._distances = np.zeros((self.n_frames, n_triangle))
```

And in `single_frame`, you could update `self._distances[self._frame_index] = d`.

package/MDAnalysis/analysis/Aggregation_size.py
```diff
@@ -0,0 +1,370 @@
...
+        for counter in range(1, self._molnum):
+            # aggergate parameter include the molecules in each cluster
+            aggregate[counter-1] = [counter]
+            # this "if" will restrict double calculation of the same molecule
+            # to increase efficincy of code
+            if counter not in P_Level:
+                # define the ones that are attached
+                # to a molecule in first level
+                bool_sector = self.d[counter] < self._cut_off
+                Coordinates = self.d[counter][bool_sector]
```
lilyminium, line 223: This variable name is a bit misleading, as `self.d` contains distances.

package/MDAnalysis/analysis/Aggregation_size.py
lilyminium: Also, `listofCoordinates` is again confusing, as it actually contains the indices of the residues with a sufficiently low cutoff. Although I think this will error as there should not be a second `index[1]` element in a 1D array.

package/MDAnalysis/analysis/Aggregation_size.py
lilyminium: I think @orbeckst meant ``if counter in P_level: continue`` 😄

package/MDAnalysis/analysis/Aggregation_size.py
```diff
@@ -0,0 +1,370 @@
...
+            # z_average
+            z_average.append(np.sum(np.array(num)*(np.array(size)**3)) /
+                             np.sum(np.array(num)*(np.array(size)**2)))
+        self.results[:, 1] = np.array(number_average)
+        self.results[:, 2] = np.array(weight_average)
+        self.results[:, 3] = np.array(z_average)
+        columns = ['Frame', 'number_average',
+                   'weight_average', 'z_average',
+                   'Radius of Gyration']
+        self.df = pd.DataFrame(self.results, columns=columns)
```
lilyminium, line 360: Instead of a data frame, consider saving your results as different attributes. We don't have [pandas as a core dependency](https://userguide.mdanalysis.org/1.0.0/module_imports.html#module-dependencies-in-the-code) so it's not guaranteed to be installed in the user's Python environment.

package/MDAnalysis/analysis/Aggregation_size.py
```diff
@@ -0,0 +1,370 @@
...
+            # to increase efficincy of code
+            if counter not in P_Level:
+                # define the ones that are attached
+                # to a molecule in first level
+                bool_sector = self.d[counter] < self._cut_off
+                Coordinates = self.d[counter][bool_sector]
+                index1 = np.where(bool_sector)
+                # list the connected molecules to a molecule in first level
+                listOfCoordinates = list(zip(index1[0], index1[1]))
+                # match array number with molecule number: 2-3[0] => 2-3 [3]
```
lilyminium, line 227: I don't understand this comment; what does "2-3" mean? Explicit comments are always better than short, but possibly confusing, ones. It's also often easier to use the variable names in the code, e.g. `p_level-d_level[counter]-> p_level-d_level[connected]` (just an example, as I do not know what you're referring to here)

package/MDAnalysis/analysis/Aggregation_size.py
jbarnoud: I think @orbeckst means:

```suggestion
            if counter in P_Level:
                continue
```

The idea is to abort the iteration early so the bulk can be one indentation level to the left.  The further the code is indented, the more difficult it becomes to read.

jbarnoud:
As @orbeckst and @lilyminium pointed out, you made good progress there! However, it is not ready yet and there will be several more rounds of review.

I let you address the open comments before I make another pass through the code. Here are the steps I think you should follow:

* **Write a first series of tests.** These should make sure that you can run your code with all your "types" and that the cut-offs are respected. This should catch errors such as undefined variables or unused cut-offs for instance.
* **Clarify your intent.** Some of your variable names are not descriptive or ambiguous as @lilyminium pointed out. Since you are at it, you can make sure your variable follow the naming conventions. After this step, your code will be easier to understand, both for you and for us. It makes a big difference with a little effort.
* **Match the user's expectations.** An analysis should fit as seemingly as possible in the normal MDAnalysis workflow. Users expect to work with atom groups for instance; it should not be necessary to do anything special to a trajectory before it is analysed.
* **Write more tests.** At the previous step, your user interface will have changed and this will have broken your tests. You will need to fix them. This time you may have more cases to test. Also, at this point, you should test the algorithm is working correctly, including in edge cases (what if everything is one big aggregate, what if this aggregate touches itself through the periodic box, what if there are only monomers...)
* **Simplify the code.** Are there things you do "by hand" that already have solutions? Using networkx to get the connected component is an example. The simpler the code is, the simpler it is to maintain in the long run. At this stage, it is crucial your tests cover well your code. If you break something at this stage, you want your tests to tell you.
* **Polish.** There are always small things to fix or improve here and there. For instance, are all the invalid inputs raising warnings or exceptions?

I know it looks daunting. But you are doing good so far, and after this process, you will have code you can be proud of.

aligh92:
Thanks to all of you for valuable feedback. I also appreciate your encouragement. I will write a test and address all your comments and get back to you soon. 
I really appreciate your help.

package/MDAnalysis/analysis/Aggregation_size.py
aligh92: self.d parameter looks like [array((n)*1), array((n-1) * 1), array((n-2) * 1), ... , array((1)*1)]. Each array consists of the distance between a molecule and the rest of the molecules. So, in the main function, provided in _single_frame, the code will go through one array (let's say the first array) and find the connected molecules to the first molecule. Then in the aggregation function, code will check if the molecules which were connected to the molecule 1 are connected to the other molecules or not. This loop will continue until there will not be any new connection. 
I read Networkx tutorial and looked to the "leaflet.py" code. Unfortunately, I could not find a similar function. Please advice me and let me know if I have to rewrite the code with Netwrokx.

package/MDAnalysis/analysis/Aggregation_size.py
aligh92: I changed that to return. 
It needs to be replaced because the distance between molecules will change along the simulation time and we do not want to store them since they only need to be used in each frame.

package/MDAnalysis/analysis/Aggregation_size.py
```diff
@@ -0,0 +1,370 @@
...
+        """
+        This function calculates the distance between the COM
+        of two molecules.
+        """
+        self._cut_off = self._cut_off or 8.5
+        cm = {}
+        for x in range((self._molnum)):
+            cm[x+1] = u.select_atoms(f"resid {x+1}").center_of_mass()
+        cm_coordintes = np.array(list(cm.values()))
+        dists = distances.distance_array(
```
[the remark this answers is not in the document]
aligh92: Thank you so much for the recommendation. I tried that, it is not possible for my case since d parameter is like bellow
 [array((n)*1), array((n-1) * 1), array((n-2) * 1), ... , array((1)*1)]

package/MDAnalysis/analysis/Aggregation_size.py
aligh92: results will be reported in self.results with 5 columns. Columns: 1- Frame number 2- number average 3- weight average 4- z average 5- Average Gyration radius. So I can delete the lines which are related to panda

package/MDAnalysis/analysis/Aggregation_size.py
aligh92: This comment we assume that we are studying the connected molecule with molecule 2, and we assume molecule 2 is connected to molecule 3. In this case, the first element of "self.d[1]", which has 0 index, will show up in "i[0] for i in listOfcoordinate])". But we would like to save it as 3 for further calculation. So, we are doing the next line code to adjust the number of the index (e.g. 0) with the number of the connected molecule (e.g. 3) to the main molecule (e.g. 2), if that makes sense. I will try to make it more clear in the code as well.

package/MDAnalysis/analysis/Aggregation_size.py
aligh92: Thank you, I used the return as @orbeckst mentioned. I do not need to save distance of all frame that is why I am replacing them every frame.  I also changed the name of parameter to make them more readable and understandable. you will see that in the new version of code.

package/MDAnalysis/analysis/Aggregation_size.py
aligh92: Sorry, I guess the bad name for parameter cause some misunderstanding. Actually, I do not have any molecule coordinate after measuring distances. I have changed the names and make them more rational. So here we only have distances between molecules. I agree with you that the arrays are faster and more suitable for coordination and distance calculation, that is why I used them at the beginning of the code. I agree that the code will be faster if everything will be with arrays but the point is some times I used dictionaries and sets which are more compatible with list. So if you do not mind, I will keep the code as it is in terms of this comment and make test. If after the test section you still think code is not efficient I will try my best to rewrite the lists with arrays.

package/MDAnalysis/analysis/Aggregation_size.py
```diff
@@ -0,0 +1,370 @@
...
+                ii += 1
+
+    def center_of_mass(self, d):
+        """
+        This function calculates the distance between the COM
+        of two molecules.
+        """
+        self._cut_off = self._cut_off or 8.5
+        cm = {}
+        for x in range((self._molnum)):
```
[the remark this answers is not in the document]
aligh92: Actually, when I was writing my code for my scientific purposes, I compared the unwrap function in MDAnalysis and GROMACS and I found out that they are a bit different. So since I was not sure about the algorithm in MDAnalysis, I decided to unwrap my box before entering that to the code. I mentioned that for users in the docs as you may see in P.S. section: number 2.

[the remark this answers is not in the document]
jbarnoud: What differences did you get between gromacs and MDAnalysis unwrapings? Ideally, the two should be identical if used properly. If it is not the case, it may be a bug and we need to fix it instead of avoiding it.

package/MDAnalysis/analysis/Aggregation_size.py
jbarnoud: What you describe is one way of getting the connected components of a graph. In short, what the leaflet code does is that it makes a graph where each "molecule" is a node, and edges are formed when the distance between these molecules are lesser than the cut-off. Then it uses networkx to find the connected component, being the parts of the graph where each node can be accessed from any other node by following the edges.

What you do in your code is basically the same thing. Each of your "types" are different ways of building the distance matrix to build the graph.

aligh92:
Hi, I actually did most of your revision and made a test file. For the test file, I needed to use my own output. So I needed to change the insert file name in the datafiles.py. However, when I implemented the pull request there was an error and I tried to resolve that quickly and it seems it did not work out. 
Could you please guide me?

jbarnoud:
From a very quick look on my phone, it looks that you removed lines from datafiles.py. Without these lines, the other tests do not find their reference.

aligh92:
Thank you for responding. So, I pull first this time, I added a few lines that were necessary based on my code and push them again. I did not delete lines. It seems there are still some problems. I would appreciate if you can look into that again, please.

testsuite/MDAnalysisTests/datafiles.py
```diff
@@ -405,6 +407,10 @@
 PRMErr2 = resource_filename(__name__, 'data/Amber/ace_mbondi3.error2.parm7')
 PRMErr3 = resource_filename(__name__, 'data/Amber/ace_mbondi3.error3.parm7')
 
+<<<<<<< HEAD
```
jbarnoud, line 410: It looks like you had a git merge conflict that you did not fix. These lines are not recognised by python and cause issues.

aligh92: Thanks, I fixed and resubmit that

aligh92:
@jbarnoud I am so sorry to bother you. but I do not know why it did not work out again :-(

testsuite/MDAnalysisTests/datafiles.py
```diff
@@ -179,6 +179,8 @@
     "JaninArray", "LYSJaninArray", # time series of chi1/chi2 angles
     "PDB_rama", "PDB_janin", # for testing failures of Ramachandran and Janin classes
     "PDB_metal", # PDB with metal atoms
+    "Asph_gro"
+    "Asph_xtc"
```
jbarnoud, line 183: The error in the test is

```
from MDAnalysisTests.datafiles import *
E   AttributeError: module 'MDAnalysisTests.datafiles' has no attribute 'Asph_groAsph_xtcBATArray'
```

It happens when a module does `from datafiles import *`. In such cases, python looks for the `__all__` variable in the module. If it exists, python imports everything that is listed in there. The `__all__` variable contains an element `"Asph_groAsph_xtcBATArray" where you would want 3 distinct elements. This is because there is no coma between the 3 elements, so python takes them as a single string.

Try the following code:

```python
my_variable = ("a" "b" "c")  # see how there is no coma!
print(my_variable)  # prints abc
```

The solution to that problem is to add the missing comas.

Since you are at it, you could also add a comment to tell what the files are about.

aligh92: Thank you sooooo much. you are very helpful.

aligh92:
@jbarnoud  I added the commas. Still, I get errors. Please give me advice

jbarnoud:
The error seems unrelated to your code. Maybe it has something to do with #2894 by @edisj?

orbeckst:
Yes, sorry, the test failures are due to issue #2890 . Please ignore these specific failures. We're working on a fix in PR #2894 and once that one is done, you can merge the fixed develop branch into your branch and it should all work again.

aligh92:
Thank you for informing me. So I will wait until the problem gets fixed.

orbeckst:
You don’t have to wait with your work. You can look at the output and look for any errors related to your work but ignore errors related to “H5MD”.

I would also recommend you run the tests locally for yourself. The User Guide has a section with notes on how to do this.

aligh92:
@orbeckst thank you so much. I actually ran the test locally before submission and it worked. I got convergence more than 90 % if I understood that correctly.

orbeckst:
Have a look at the failed tests: click on the "Details" link for Travis CI and then click on a failing job. Read the output. For instance, [Job 3140.1](https://travis-ci.com/github/MDAnalysis/mdanalysis/jobs/371240983) shows you
```
    command: /home/travis/miniconda/envs/test/bin/python3.6 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/home/travis/build/MDAnalysis/mdanalysis/package/setup.py'"'"'; __file__='"'"'/home/travis/build/MDAnalysis/mdanalysis/package/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-1rnxla0v

         cwd: /home/travis/build/MDAnalysis/mdanalysis/package/

    Complete output (6 lines):

      File "<string>", line 1, in <module>

      File "/home/travis/build/MDAnalysis/mdanalysis/package/setup.py", line 76

        <<<<<<< HEAD

         ^

    SyntaxError: invalid syntax

    ----------------------------------------

ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.

The command "eval $BUILD_CMD" failed and exited with 1 during .
```
There's a problem in file `setup.py`, something that looks like a merge commit that was not properly resolved. What did you do in the last "merge" commit c113a3f, specifically, which commands did you run?

aligh92:
I pulled everything first and then I pushed and merge. I guess I messed up somehow. Give me some time. I will redo it. So sorry for inconvenience.

orbeckst:
Don’t worry, no need to apologize. Thank you for putting in the effort!

aligh92:
Hello @orbeckst, 

Thanks for motivating me and helping me during this work. I fixed those little conflicting and here is the pull request. 
I went through detail of Travis CI as you mentioned before and tried to solve issues that I understood but it seems there is some more issues. I wonder if you can help me with them. 

Best regards, 
Ali

orbeckst:
There are segmentation faults due to some of the C code in nsgrid.
- https://travis-ci.com/github/MDAnalysis/mdanalysis/jobs/377042661#L1988
- https://ci.appveyor.com/project/orbeckst/mdanalysis/builds/34827567/job/wk4aqxkf35qfabyt

I haven't seen these specific problems come up elsewhere. It is possible that you added your changes (since commit 68e9450 ) to outdated code. I think it would be clearer if you were basing your code on the latest development branch. You can merge the current develop into your branch and then do a diff against develop and it should only show _your_ code.

aligh92:
Dear @orbeckst 

I think the main problem get fixed now. However, there is still a small problem. I am trying to figure out this. It would be appreciated if you can check that too as well. 

Thank you so much for all help

viral28:
The forked repository was behind the original develop branch previously, that is why travis gave an error. So we updated the forked repository and than merged it, however, there were too many changes to follow (in the other files) so we replaced all the files from this branch with the ones from the develop branch and then implemented the changes. This went further in the test process but it failed. I notice that the merge history also shows changes on files we did not change. I was wondering what the best way to do this is?

aligh92:
Dear @orbeckst , @lilyminium  and @jbarnoud 

I tried a lot to fix the issue with this pull request. I fixed many problems but it seems there is still an issue that I can not find that. I am wondering if you can help me with that. I really appreciate your help in advance.

aligh92:
Looks like commit c3b108a went very well but there is a small issue with that I would appreciate any help

lilyminium:
```
    def test_all_exports():

        import MDAnalysisTests.datafiles

        missing = [name for name in dir(MDAnalysisTests.datafiles)

                   if

                   not name.startswith('_') and name not in MDAnalysisTests.datafiles.__all__ and name != 'absolute_import']

>       assert_equal(missing, [], err_msg="Variables need to be added to __all__.")

E       AssertionError: 

E       Items are not equal: Variables need to be added to __all__.

E        ACTUAL: 2

E        DESIRED: 0
```

Hi @aligh92, what this error message means is that although you have added your files to `datafiles.py` as `Asph_gro` and `Asph_xtc`, you have not added the names `'Asph_gro'` and `'Asph_xtc'` to the all `__all__` list at the top of the file, so the files cannot be exported.

aligh92:
@lilyminium Thank you so much for looking into that. It is weird, I did that on my computer, I do not know why it did not transfer here!!! Anyway, I made the change again, let's see how it works.

aligh92:
Looks like there is some error here:
https://dev.azure.com/mdanalysis/mdanalysis/_build/results?buildId=950&view=logs&j=6337f8de-d220-572c-a53d-20c279e44ba9&t=38e5c318-25f5-5156-9fa7-6cd72ffb271c&l=627

aligh92:
@lilyminium Thank you again for previous help. Currently, I have a problem with these two I guess. Any help is appreciated.
1-https://dev.azure.com/mdanalysis/mdanalysis/_build/results?buildId=951&view=logs&j=1cdae27b-58e8-5c64-01e2-519b6cdf4b73&t=c13e15da-7c65-5b3c-5c17-f011eed8e1e8&l=443
2-https://dev.azure.com/mdanalysis/mdanalysis/_build/results?buildId=951&view=logs&j=b3e715e3-855a-545e-22f2-ec20d10eac17&t=18ebbd08-144c-517e-f71d-909fbbe1f815&l=384

lilyminium:
The error is odd. I think it's failing to read the xtc offsets?

```
    @pytest.fixture()
    def atomgroup():
>       return mda.Universe(Asph_gro, Asph_xtc).select_atoms('resname AsphC')

... [more error log, until it dies in np.load] ...
                # Try a pickle
                if not allow_pickle:
>                   raise ValueError("Cannot load file containing pickled data "
                                     "when allow_pickle=False")
```

@jbarnoud may be the one to ask here. The rest of the tests pass so it's likely not because of something with your file.

aligh92:
@lilyminium Thank you so much for looking into the errors. I really appreciate your help and time. 

Dear @jbarnoud,
Could you please look into this error and check if there is something wrong, please? 
Looking forward to hearing from you,

aligh92:
Dear @orbeckst 

I am wondering if you can help me with this issue. I could run the test locally but it seems there is some small issue here. 

Any help is appreciated. 
Thank you in advance, 
Best regards, 
Ali

jbarnoud:
From a quick look at the error message, it looks like there is an issue reading the offset for the test XTC file, indeed. As it only complains about the file introduced in this PR, I doubt there is an issue with the general system producing and reading the file (though it remains a possibility, of course). However, looking at the files introduced here, I do not see what could cause the problem if you can run the tests locally. Could you do the following test locally:

* if the offset file exists for Asph_xtc (hidden file next to the trajectory with a similar name to the trajectory and a `.npz` extension), delete it
* create a universe for the trajectory (the offset should be recreated)
* create another universe for the trajectory (the offset should be loaded)

If this works, then I do not know what the problem is. If it fails somewhere, then there is probably an issue with the trajectory. My hypothesis (and it is a not very substantiated one) is that the tests pass locally for you because you have an offset file already while the CI would not have that file.

aligh92:
Dear @jbarnoud, 

Thank you so much for your reply. I am looking to that based on your comments and I will get back to you as soon as possible. 

Best regards

aligh92:
Dear @jbarnoud, 
I deleted the .npz file and ran the pytest again. It went through successfully and the .npz file was build by itself again. I am not sure if this is what you meant. Please see the attached screenshot which is the output of local pytest. 


image

jbarnoud:
These are OK, they are warnings that have nothing to do with your PR. I will restart the failing tests to see if it is not just a fluke. On 5 Oct 2020 03:11, aligh92 <notifications@github.com> wrote:
Dear @jbarnoud,
I deleted the .npz file and ran the pytest again. It went through successfully and the .npz file was build by itself again. I am not sure if this is what you meant. Please see the attached screenshot which is the output of local pytest.

jbarnoud:
@aligh92 I did not manage to just re-run the pipeline to see if it works by just insisting. So I added a commit to your branch fixing an indentation error. The pipelines are running again now, we will see if it works any better.

aligh92:
Dear @jbarnoud , 

Looks like we still have a problem. The main problem that I saw before and it is still there is related to the attached photo

image

orbeckst:
Aggregate size analysis might be a good fit for a [MDAKit](https://www.mdanalysis.org/2021/08/31/CZI-EOSS4/#2-launching-the-mdakits-ecosystem) (independent analysis package linked to MDAnalysis). 

@aligh92 would you be interested in pursuing that route?

aligh92:
Hi Oliver,

I am so sorry for the late reply. I would be very pleased to have my code
published and functional for future people's use. Could you please let me
know how and what can I do to fit that to the MDAKit?

Looking forward to hearing from you,
Best regards,
Ali

orbeckst:
@aligh92 , good to hear that you're interested in doing an MDAKit. For right now I added the MDAKit label to the PR and issue.

We are working on guidelines. We'll have a document for discussion soon and then we'd really like to hear from developers like you what they think. We'll also be very happy if you'd want to try it out and give feedback on the process.

aligh92:
Thank you so much Oliver,

I just should mention that I am not very good with GitHub and I need
probably a bit of assistance there if I need to do any action. That was
probably one of the main reasons that I could not publish my code in the
first place. Because I still can not understand why I could not add my code
to the MDAnalysis package itself.

Best regards,
Ali

hmacdope:
Closing as stale, feel free to re-open if you want to continue.

PR #2824, opened 2020-07-04
Removes profraw file and adds ignore

### LINKED ISSUE #2823: extra default.profraw file
Just noticed that we have a `default.profraw` file that seems to have been added in #2668 : https://github.com/MDAnalysis/mdanalysis/blob/develop/package/MDAnalysis/analysis/default.profraw

As far as I'm aware this since some kind of clang binary file and doesn't seem necessary so it probably needs to be removed?

IAlibay:
Fixes #2823 

 - git rm'ed the default.profraw file
 - added a gitignore entry for *.profraw files

PR Checklist
------------
 - Tests? N/A
 - Docs? N/A
 - CHANGELOG updated? N/A


richardjgowers:
kind of annoying this slipped in as the ~1mb chunk will be in the repo history, but not sure it's viable to erase from history and break all the hashes..

RMeli:
For the records: https://github.com/conda-forge/python-feedstock/issues/339

PR #2830, opened 2020-07-08
Create .pep8speaks.yml

kain88-de:
I found that the 79 char limit of pep8 is pretty restricting. Often extending it to 100 allows for more natural code. It works great for me to still view two files side by side on my laptop. Jake VanderPlas had a good entry about line length statistics in python on his [blog](https://jakevdp.github.io/blog/2017/11/09/exploring-line-lengths-in-python-packages/)

Fixes #

 - 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/.pep8speaks.yml
+++ b/.pep8speaks.yml
@@ -1,0 +1,5 @@
+scanner:
+  diff_only: True
+
+pycodestyle:
+  max-line-length: 100
--- a/.pep8speaks.yml
+++ b/.pep8speaks.yml
@@ -2,4 +2,4 @@ scanner:
   diff_only: True
 
 pycodestyle:
-  max-line-length: 100
+  max-line-length: 79


IAlibay:
My 2 cents here is that I think overall 79 chars has resulted in reasonable code in most cases. Considering pep8speaks' comments are non-blocking, maybe this should be more of an unwritten policy instead?

My point here being that having a pep8 linter run and tell people to look at lines that could easily be split could be more useful than applying a blanket exception on all cases (sorry I'm probably making a terrible explanation of my point here, but hopefully it makes some sense?)

lilyminium:
I agree with @IAlibay; in most cases, the 79 character line length works. In some other cases they don't; for example, pep8speaks has pointed out 6 lines that are too long in #2816. I have reasons for not editing them down, but it was good to have them pointed out in the first place to make me think about it.

What we could also do is make the pep8speaks footer link to a relevant guideline of PEP8: [A Foolish Consistency is the Hobgoblin of Little Minds](https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds) and add a note on how you should prioritise readability over following the rules 100%.

kain88-de:
> or example, pep8speaks has pointed out 6 lines that are too long in #2816. I have reasons for not editing them down, but it was good to have them pointed out in the first place to make me think about it.

That is my reason for a 100 line length. With 100 characters I found that the need for this vanishes. I found that not needing to think about formatting is really nice when working.

orbeckst:
I think we need to keep in mind who PEP8speaks is mostly for. Most of the time we need PEP8 for new contributors. I'd rather have them adhere to 79 most of the time. 

When @kain88-de or @lilyminium or @jbarnoud or @IAlibay do not adhere to 79 characters I don't mention it because I _know_ that they have a good reason. Just keeping PEP8speaks non-blocking is good enough for me in this case.

orbeckst:
More importantly in my opinion is what PEP8speaks does _not_ catch at the moment and I'd be curious if we could make it. For example in https://github.com/MDAnalysis/mdanalysis/pull/2822#issuecomment-659701523 PEP8speaks did not highlight numerous variable and class naming issues
- https://github.com/MDAnalysis/mdanalysis/pull/2822#discussion_r456154353 
   ```python
   class Aggregation_size(AnalysisBase):
   ```
   (maybe that's not a PEP8 but still...)
- https://github.com/MDAnalysis/mdanalysis/pull/2822#discussion_r456155035 referring to 
   ```python
       def __init__(self, atomgroup, number_of_molecules, Type="Closest",
                 no_monomer=True, Gyr_calc=False, cut_off=False,
                 atom_1=None, atom_2=None, **kwargs):
    ```
    where `Gyr_calc` should be `gyr_calc` (and `Type`... well, should be `method` ;-).)
which I thought it should have picked up.
- others: see PR #2822

I think PR #2822 is a good example to show us what would help us to get a new user/developer to integrate their own code as effortlessly (for them and us) as possible.

orbeckst:
I have no idea if we wanted to move farther with this old PR but given the broad opinion in favor of 79 chars I edited the config file accordingly. Do we still need/want it?

IAlibay:
@orbeckst I genuinely don't know this one 🙈 how does the config file now differ from the default behaviour?

orbeckst:
No idea... I am just asking — if not needed then we should close the PR with a wontfix.

IAlibay:
I'm tempted to vote for a wontfix here.

That being said, I'm happy to default to @lilyminium's opinion who probably knows far more about pep8speak than I do.

lilyminium:
I'm pretty sure the 79-character limit is default, the bot has complained to me many times about it....

IAlibay:
For the sake of saving time, I'm going to label this one as `wontfix` and close. I would still be happy to discuss customising pep8speak further, but at the same time there are 54 other PRs we should also deal with.

PR #2831, opened 2020-07-09
Set numpy and rdkit min version

### LINKED ISSUE #2827: Raise the minimum numpy version

In PR #2775 @IAlibay suggested that it might be a good idea to set the minimum version for RDKit to 2020. 
This version is however not compatible with `numpy==1.13.3`.  
Since python < 3.6 is no longer supported and there's already a discussion to drop 3.6 on the [mailing list](https://groups.google.com/forum/#!topic/mdnalysis-devel/vCDDPH_2iTo), it might be a good time to raise the minimal version of numpy to at least `1.15.0` (compatible with rdkit 2020.03.1).

cbouy:
Fixes #2827 

 - raise numpy minimum version from 1.13.3 to 1.16.0
 - set the minimum RDKit version to 2020.03.1 for tests

I'm not sure if there's a way to enforce RDKit>=2020.03.1 for users on install since it's a conda dependency

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -10,7 +10,7 @@ cache:
 environment:
     global:
         CONDA_CHANNELS: conda-forge
-        CONDA_DEPENDENCIES: pip setuptools wheel cython biopython networkx joblib matplotlib scipy vs2015_runtime pytest mmtf-python GridDataFormats hypothesis pytest-cov codecov chemfiles tqdm tidynamics>=1.0.0 rdkit
+        CONDA_DEPENDENCIES: pip setuptools wheel cython biopython networkx joblib matplotlib scipy vs2015_runtime pytest mmtf-python GridDataFormats hypothesis pytest-cov codecov chemfiles tqdm tidynamics>=1.0.0 rdkit>=2020.03.1
         PIP_DEPENDENCIES: gsd==1.9.3 duecredit parmed
         DEBUG: "False"
         MINGW_64: C:\mingw-w64\x86_64-6.3.0-posix-seh-rt_v5-rev1\mingw64\bin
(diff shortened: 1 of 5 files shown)


IAlibay:
> I'm not sure if there's a way to enforce RDKit>=2020.03.1 for users on install since it's a conda dependency

Could we set that in extras_require the same way we do with tidynamics?

I'll be honest I'm not 100% sure how extras_requires behaves if you have something but it's not the right version number, someone else might know better.

cbouy:
With rdkit 2019:
```
$ pip install ".[analysis]"
...
ERROR: Could not find a version that satisfies the requirement rdkit>=2020.03.1 (from MDAnalysis==2.0.0.dev0) (from versions: none)
ERROR: No matching distribution found for rdkit>=2020.03.1 (from MDAnalysis==2.0.0.dev0)
```

lilyminium:
[rdkit is not on pypi](https://rdkit.blogspot.com/2019/11/why-rdkit-isnt-available-on-pypi.html). It must be in the conda package or pre-installed.

If it's a required feature that you need from 2020.03.1, could you raise a relevant error message on failure?

orbeckst:
If we know that rdkit cannot be installed from PyPi then it doesn't make sense to add it to setup.py as a requirement. Add a comment and link to the "why not on PyPi" but then have your module deal gracefully with the absence of rdkit. 

Make sure that the "minimal" tests on travis (without rdkit) pass.

orbeckst:
If you make this PR only about numpy then we merge it in no time.

cbouy:
Setting the RDKit version as extra in the setup.py is not really super important tbh, the cases where some of the parser and converter features fail are pretty rare but the errors aren't straightforward to debug, so we'll just have to keep that in mind if issues related to rdkit pop up in the future. 
I'll just keep this as a requirement for the CI then.

PR #2836, opened 2020-07-10
MAINT: master -> main where possible

tylerjereddy:
* replace `master` with `main` in locations where
it does not appear to cause burden for the development
team

* none of the cases replaced here appear to be related
to `master` branch on GitHub (indeed, we might leave
that transition to the GitHub team)

* this is similar in spirit to GitHub's indication that
they will do the same for branch renaming:
https://twitter.com/natfriedman/status/1271253144442253312

* this is not fully comprehensive, but catches the low-hanging
cases I think; I saw a few others, but they were more functional
like parts of test files I didn't want to mess with, at least
initially

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/core/universe.py
+++ b/package/MDAnalysis/core/universe.py
@@ -307,9 +307,9 @@ class Universe(object):
         current system dimensions (simulation unit cell, if set in the
         trajectory)
     atoms, residues, segments
-        master Groups for each topology level
+        main Groups for each topology level
     bonds, angles, dihedrals
-        master ConnectivityGroups for each connectivity type
+        main ConnectivityGroups for each connectivity type
 
     .. versionchanged:: 1.0.0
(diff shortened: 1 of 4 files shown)


package/MDAnalysis/core/universe.py
```diff
@@ -307,9 +307,9 @@ class Universe(object):
         current system dimensions (simulation unit cell, if set in the
         trajectory)
     atoms, residues, segments
-        master Groups for each topology level
+        main Groups for each topology level
     bonds, angles, dihedrals
-        master ConnectivityGroups for each connectivity type
+        main ConnectivityGroups for each connectivity type
```
lilyminium, line 312:
```suggestion
        principal ConnectivityGroups for each connectivity type
```

tylerjereddy:
Squash merged; codecov patch diff complaint seems spurious

PR #2844, opened 2020-07-13
Refactor LeafletFinder and add new methods

### LINKED ISSUE #2843: Other ways to find leaflets

Looking at the current LeafletFinder algorithm, it seems that the cutoff is very important in building the graph, as every node is connected providing its distance is lower than the cutoff. The cutoff must therefore be small enough to not include lipids in the other leaflet, but large enough to account for potentially bulky tail groups and lipids in the middle of flip-flopping. 

> 1. build a graph of all phosphate distances < cutoff
> 2. identify the largest connected subgraphs
> 3. analyse first and second largest graph, which correspond to the leaflets

If you are identifying the leaflets on every frame of a trajectory where the membrane is changing thickness and lipids are flip-flopping, a single cutoff may not cut it. An ``optimize_cutoff`` function is provided to help out, but it is self-admittedly slow.

* use whichever method I like
* have more methods pre-defined, e.g. spectral clustering, which removes a need to think much about the cutoff
* but have them flexible and accessible outside LeafletFinder

lilyminium:
Fixes #2843

Split off from #2816  

 - Refactor LeafletFinder to accept other methods
 - Move the graph grouping method to distances
 - Add a method to use spectral clustering and distances to centers-of-geometry
 - Allow LeafletFinder to use any method
 - Changed default cutoff to 20 (spectral clustering requires a slightly larger one)
 - Sort leaflets by z-coordinate in LeafletFinder
 - Remove most of the iteration behaviour

None of these tests run on vesicles, which might be a worthwhile addition.

Unfortunately clustering is quite slow compared to the graph method, and it will get slower with a larger cutoff. The lipid orientation one is much worse, but works slightly better for a small selection of lipids in a large membrane. Edit: but only slightly; running this on a reasonable selection of lipids (~50%) of a planar bilayer as it bends points to the spectral clustering with large cutoff (e.g. cutoff=80) being the most fool-proof and robust. In all likelihood users will want to experiment to find what works best

Screen Shot 2020-07-13 at 6 37 01 pm

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


lilyminium:
Sorry @orbeckst, I should look at some stability issues first before this is ready.

orbeckst:
Ok, ping me when I should look at it.

lilyminium:
Closing this PR to MDAnalysis as it is now implemented in the external toolkit [lipyds](https://github.com/lilyminium/lipyds).

PR #2849, opened 2020-07-14
Fixes AttributeError in HydrogenBondAnalysis 

### LINKED ISSUE #2818: Running AlignTraj on GRO/XTC causes Travis segfault

### LINKED ISSUE #2848: HydrogenBondAnalysis  AttributeError when empty `hydrogens` AtomGroup
## Expected behavior ##

When I run HydrogenBondAnalysis with an updating selection for the donor, hydrogen and acceptor atoms, if no hydrogen atoms are present at a given frame there should be no errors raised.

## Actual behavior ##

If the topology is used to find donor-hydrogen pairs (but not if the the D-H pairs are identified via a distance cutoff), an attribute error is thrown if the `hydrogen` AtomGroup is empty. 
This comes from creating the `donor` AtomGroup by: `donors = sum(h.bonded_atoms[0] for h in hydrogens)`.

```
from MDAnalysisTests.datafiles import waterPSF, waterDCD
import MDAnalysis as mda

kwargs = {
    'donors_sel': None,
    'hydrogens_sel': 'name H',  # no atoms in this universe have name H
    'acceptors_sel': 'name OH2',
    'd_h_cutoff': 1.2,
    'd_a_cutoff': 3.0,
    'd_h_a_angle_cutoff': 120.0
}

universe = mda.Universe(waterPSF, waterDCD)
h = HydrogenBondAnalysis(universe, **kwargs)
h.run()
```

```
E       AttributeError: 'int' object has no attribute 'positions'

package\MDAnalysis\analysis\hydrogenbonds\hbond_analysis.py:586: AttributeError
```

- Which version are you using? 1.00
- Which version of Python 3.8
- Which operating system? Windows and Linux

We (@bieniekmateusz and I) will make a PR to fix this issue shortly!

p-j-smith:
Fixes #2848 

 - An attribute error is thrown when finding D-H pairs via the topology if `hydrogens` is an empty AtomGroup. This has been fixed so that an empty `donors` group is created, rather than assigning `donors = 0`.
 - Added a test case that raises the attribute without the fix.

PR Checklist
------------
 - [X] Tests?
 - [X] Docs?
 - [X] CHANGELOG updated?

Co-authored-by: bieniekmateusz <bieniekmat@gmail.com>

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -34,8 +34,9 @@ Fixes
     density=True; the keyword was available since 0.19.0 but with incorrect
     semantics and not documented and did not produce correct results (Issue
     #2811, PR #2812)
-  * Attribute error was thrown when finding D-H pairs via the topology if
-    `hydrogens` was an empty AtomGroup (Issue #2848)
+  * In hydrogenbonds.hbond_analysis.HydrogenbondAnalysis an AttributeError 
+    was thrown when finding D-H pairs via the topology if `hydrogens` was an 
+    empty AtomGroup (Issue #2848)
 
 Enhancements
   * Added the RDKitParser which creates a `core.topology.Topology` object from
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -34,6 +34,8 @@ Fixes
     density=True; the keyword was available since 0.19.0 but with incorrect
     semantics and not documented and did not produce correct results (Issue
     #2811, PR #2812)
+  * Attribute error was thrown when finding D-H pairs via the topology if
+    `hydrogens` was an empty AtomGroup (Issue #2848)
 
 Enhancements
   * Added the RDKitParser which creates a `core.topology.Topology` object from
--- a/package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
+++ b/package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
@@ -219,7 +219,7 @@ from .. import base
 from MDAnalysis.lib.distances import capped_distance, calc_angles
 from MDAnalysis.lib.correlations import autocorrelation, correct_intermittency
 from MDAnalysis.exceptions import NoDataError
-from MDAnalysis import AtomGroup
+from MDAnalysis.core.groups import AtomGroup
 
 from ...due import due, Doi
 
--- a/package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
+++ b/package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
@@ -219,6 +219,7 @@ from .. import base
 from MDAnalysis.lib.distances import capped_distance, calc_angles
 from MDAnalysis.lib.correlations import autocorrelation, correct_intermittency
 from MDAnalysis.exceptions import NoDataError
+from MDAnalysis import AtomGroup
 
 from ...due import due, Doi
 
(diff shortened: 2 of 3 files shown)


package/CHANGELOG
```diff
@@ -34,6 +34,8 @@ Fixes
     density=True; the keyword was available since 0.19.0 but with incorrect
     semantics and not documented and did not produce correct results (Issue
     #2811, PR #2812)
+  * Attribute error was thrown when finding D-H pairs via the topology if
+    `hydrogens` was an empty AtomGroup (Issue #2848)
```
orbeckst, line 38:
```suggestion
  * In hydrogenbonds.hbond_analysis.HydrogenbondAnalysis an AttributeError 
    was thrown when finding D-H pairs via the topology if `hydrogens` was an 
    empty AtomGroup (Issue #2848)
```

orbeckst: Just a bit more context so that users would have an idea _where_ this error would be encountered.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -219,6 +219,7 @@
 from MDAnalysis.lib.distances import capped_distance, calc_angles
 from MDAnalysis.lib.correlations import autocorrelation, correct_intermittency
 from MDAnalysis.exceptions import NoDataError
+from MDAnalysis import AtomGroup
```
orbeckst, line 222: I think it would be better to import `AtomGroup` from core.groups (I think...) – or do we normally import the top level alias?

p-j-smith: > Can the same problem also happen for acceptors or is it restricted to the case of finding H via topology?

The `acceptors` AtomGroup is always created using a selection string (even when the acceptors are guessed with `guess_acceptors`), so this shouldn't be an issue here.

p-j-smith: woops, I mean to post the above reply at the bottom of the page.

I wasn't sure where to import `AtomGroup`  from, so thanks for pointing it out!

orbeckst:
No sure why Travis CI currently states "cancelled" when the most recent check https://travis-ci.com/github/MDAnalysis/mdanalysis/builds/175684432 passed. I'll merge.

PR #2787, opened 2020-06-23
add h5md format

edisj:
Fixes #762 

Changes made in this Pull Request:
 - add h5md reader 
 - ~~add h5md writer~~ (this will be handled in separate pull request)
 - add h5md test file 

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?
 - [x] Issue raised/referenced?

PR #2850, opened 2020-07-14
CI: migrate to Azure for Win

### LINKED ISSUE #2846: replace Windows tests on AppVeyor with Azure
I noticed that we are often waiting for AppVeyor to complete whereas Azure is fast. 

Could we replace AppVeyor's Windows testing with Windows testing under Azure?

tylerjereddy:
Fixes #2846

* the majority of Windows CI testing previously provided
by Appveyor is now avaiable in the Azure CI Windows test
matrix; 64-bit Python 3.6 and 3.8 have been added to Azure
CI Windows testing

* however, Azure CI has been kept PyPI/`pip`-only, to avoid
mixing `conda` and `pip`, which has meant the exclusion of the
`RDKit` dependency

* as a result, it may be advisable to keep the Appveyor testing
in our CI for the moment to test Windows + `RDKit` until
they provide a wheel; possibly, we could consider slimming
down the tests executed in Appveyor CI to only exercise `RDKit`
and other `conda`-only code paths

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -22,12 +22,16 @@ jobs:
   variables:
     MPLBACKEND: agg    
   strategy:
-    maxParallel: 4
     matrix:
         Python37-32bit-full:
           PYTHON_VERSION: '3.7'
           PYTHON_ARCH: 'x86'
-          BITS: 32
+        Python36-64bit-full:
+          PYTHON_VERSION: '3.6'
+          PYTHON_ARCH: 'x64'
+        Python38-64bit-full:
+          PYTHON_VERSION: '3.8'
+          PYTHON_ARCH: 'x64'
   steps:
   - task: UsePythonVersion@0
     inputs:
@@ -39,6 +43,7 @@ jobs:
   - script: >-
       python -m pip install
       cython
+      hypothesis
       matplotlib
       numpy
       pytest
@@ -47,6 +52,22 @@ jobs:
       scikit-learn
       scipy
     displayName: 'Install dependencies'
+  # TODO: recent rdkit is not on PyPI
+  - script: >-
+      python -m pip install
+      biopython
+      chemfiles
+      duecredit
+      gsd==1.9.3
+      joblib
+      GridDataFormats 
+      mmtf-python
+      networkx
+      parmed
+      tidynamics>=1.0.0
+      tqdm
+    displayName: 'Install additional dependencies for 64-bit tests'
+    condition: and(succeeded(), eq(variables['PYTHON_ARCH'], 'x64'))
   - powershell: |
       cd package
       python setup.py install
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -29,6 +29,9 @@ jobs:
         Python36-64bit-full:
           PYTHON_VERSION: '3.6'
           PYTHON_ARCH: 'x64'
+        Python37-64bit-full:
+          PYTHON_VERSION: '3.7'
+          PYTHON_ARCH: 'x64'
         Python38-64bit-full:
           PYTHON_VERSION: '3.8'
           PYTHON_ARCH: 'x64'
@@ -51,6 +54,7 @@ jobs:
       pytest-xdist
       scikit-learn
       scipy
+      tqdm
     displayName: 'Install dependencies'
   # TODO: recent rdkit is not on PyPI
   - script: >-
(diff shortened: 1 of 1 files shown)


orbeckst:
Is the rationale for pip-only on Azure speed?

orbeckst:
@cbouy @IAlibay @richardjgowers is there a chance that somebody will build RDKit wheels for Windows? Alternatively, how does one install and build it in the absence of conda?

cbouy:
For wheels I'm not sure, there seem to be some progress in [this PR](https://github.com/rdkit/rdkit/pull/2690) but nothing working yet apparently
Otherwise you'd have to [build from source](https://www.rdkit.org/docs/Install.html#windows)

tylerjereddy:
> Is the rationale for pip-only on Azure speed?

That's one motivation (why I didn't do pure conda instead), but a bigger one is just control and simplicity. Bad things can happen and they can be hard to track down when you mix the two: https://www.anaconda.com/blog/using-pip-in-a-conda-environment

We've also got this astropy harness doing a bunch of conda stuff behind the scenes on appveyor too, I think. It is a lot of extra complexity for the build environment when we're looking at 1-2 packages that are not pip ready.

Up to you though--feel free to push in whatever changes you want here, as long as you're aware of the drawbacks. Azure has built-in conda support if you want to do pure conda.

No idea if the astropy ci helpers stuff works with Azure, but I would have hoped we don't really need that harness anymore.

IAlibay:
> No idea if the astropy ci helpers stuff works with Azure, but I would have hoped we don't really need that harness anymore.

Just following up here. Unfortunately we still haven't gotten to the bottom of the segfaults that can happen when astropy's CI helpers don't get used: https://github.com/MDAnalysis/UserGuide/pull/92

However, I _think_ it might be limited to Linux only (or at least in my own project it wouldn't segfault on MacOS). With a bit of luck, it might be that we can get away with it on Windows.

azure-pipelines.yml
```diff
@@ -47,6 +52,22 @@ jobs:
...
+      chemfiles
+      duecredit
+      gsd==1.9.3
+      joblib
+      GridDataFormats 
+      mmtf-python
+      networkx
+      parmed
+      tidynamics>=1.0.0
+      tqdm
```
orbeckst, line 68: tqdm is a core dependency nowadays... if we are listing it anywhere in optional deps then we are doing something wrong.

azure-pipelines.yml
```diff
@@ -47,6 +52,22 @@ jobs:
       scikit-learn
       scipy
     displayName: 'Install dependencies'
+  # TODO: recent rdkit is not on PyPI
+  - script: >-
+      python -m pip install
```
orbeckst, line 57: When H5MD in PR #2787 is added, we will also need to add h5py, but that's already working in PR #2787 – apparently, hdf5 libraries are available.

azure-pipelines.yml
```diff
@@ -22,12 +22,16 @@ jobs:
   variables:
     MPLBACKEND: agg    
   strategy:
-    maxParallel: 4
     matrix:
         Python37-32bit-full:
           PYTHON_VERSION: '3.7'
           PYTHON_ARCH: 'x86'
-          BITS: 32
```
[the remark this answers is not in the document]
tylerjereddy: Yeah, it is a little subtle, the value gets fed into the `UsePythonVersion@0` Azure task further down in the control flow, in the `architecture` setting--that's where the decision is made presumably.

azure-pipelines.yml
```diff
@@ -22,12 +22,16 @@ jobs:
   variables:
     MPLBACKEND: agg    
   strategy:
-    maxParallel: 4
     matrix:
         Python37-32bit-full:
```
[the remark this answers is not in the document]
tylerjereddy: ok, I pushed in that extra matrix entry

azure-pipelines.yml
tylerjereddy: I've moved it up to the list of deps used for 32-bit and 64-bit then.

Perhaps you'll want to error out more swiftly/decisively if `tqdm` is missing, or maybe it just gets auto-installed through our `setup.py`/related checks instead of erroring. Anyway, that's probably for a separate discussion if it matters.

azure-pipelines.yml
tylerjereddy: Ah, that's good news, if we're down to just 1 `conda`-only binary that would be good.

orbeckst:
Thanks @tylerjereddy !

PR #2852, opened 2020-07-15
RDKitParser with table of attributes and icodes

cbouy:
 - Adds a table to the doc of the RDKitParser that lists the RDKit attributes and their equivalence in MDAnalysis
 - Adds the `icodes` attribute

I somehow forgot the "insertion codes" attribute when I created the parser...
Do I need to update the changelog for this ? Since the parser is not technically there yet

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [ ] CHANGELOG updated?


richardjgowers:
@cbouy yeah should be fine without a CHANGELOG as there's no user visible change.

PR #2855, opened 2020-07-18
Remove assignment of bond attribute when PDB file contains no CONECT records

### LINKED ISSUE #2832: Empty u.atoms.bonds for PDB files without a CONECT record
## Expected behavior ##

For API consistency, when creating a Universe from a PDB file that doesn't have a CONECT record, accessing `u.atoms.bonds` should raise a `NoDataError` as it's the case for other file formats.

## Actual behavior ##

`u.atoms.bonds` will return a TopologyGroup with 0 bonds.

``` python
>>> import MDAnalysis as mda
>>> from MDAnalysis.tests.datafiles import PDB_helix

>>> u.atoms.bonds
<TopologyGroup containing 0 bonds>

```

2.0.0-dev0
Python 3.6.10
- Which operating system?
Ubuntu 18.04.4 LTS through a VM (WSL2)

IAlibay:
Fixes #2832

 - If no bonds are parsed, bond attribute is no longer set in PDB files.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/topology/PDBParser.py
+++ b/package/MDAnalysis/topology/PDBParser.py
@@ -203,7 +203,7 @@ class PDBParser(TopologyReaderBase):
                           "bonds will not be parsed")
         else:
             # Issue 2832: don't append Bonds if there are no bonds
-            if len(bonds) > 0:
+            if bonds:
                 top.add_TopologyAttr(bonds)
 
         return top
--- a/package/MDAnalysis/topology/PDBParser.py
+++ b/package/MDAnalysis/topology/PDBParser.py
@@ -167,6 +167,7 @@ class PDBParser(TopologyReaderBase):
      - resnames
      - segids
      - elements
+     - bonds
 
     Guesses the following Attributes:
      - masses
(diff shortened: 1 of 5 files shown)


package/MDAnalysis/topology/PDBParser.py
```diff
@@ -198,7 +202,9 @@ def parse(self, **kwargs):
             warnings.warn("Invalid atom serials were present, "
                           "bonds will not be parsed")
         else:
-            top.add_TopologyAttr(bonds)
+            # Issue 2832: don't append Bonds if there are no bonds
+            if len(bonds) > 0:
```
richardjgowers, line 206: Can just do "if bonds". Usually better to check the bool-ness rather than qualifying length

jbarnoud: Do `TopologyAttribute` evaluate to `False` when empty?

jbarnoud:
Seems that there is a test broken about merging universes.

package/MDAnalysis/topology/PDBParser.py
IAlibay: I didn't think it'd work, but apparently they do :) (or at least Bonds does).

jbarnoud:
It appears that the print in the test about merge was the only thing covering the repr of `TopologyGroup`. Since it was covering it by accident, and was not really testing it, I will ignore the regression in coverage.

PR #2858, opened 2020-07-19
Add elements attribute to TPR parser

jbarnoud:
Elements are read from the TPR. TPR files contain the atomic number for
each atoms, when an atom does not match an element (some virtual sites,
coarse-grained particles), the atomic number is set to -1.

The TPR parser adds the 'elements' attribute only if at least 1 atom has
its atomic number corresponding to an element. In that case, atoms that
do not match an element have their element set to an empty string.

See #2553
Closes #2628

PR Checklist
------------
 - [X] Tests?
 - [X] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/topology/tpr/utils.py
```diff
@@ -403,11 +406,15 @@ def do_mtop(data, fver):
     top.add_TopologyAttr(Impropers([improper for improper in impropers
                                     if improper]))
 
+    if any(elements):
```
IAlibay, line 409: Is it worth throwing a warning here if the element list is imcomplete? We do something like that with the TOPParser: https://github.com/MDAnalysis/mdanalysis/blob/784ccc717dba2ceff7fa5233bc3cc9d4640c2464/package/MDAnalysis/topology/TOPParser.py#L291-L293

jbarnoud:
> Could a test also be added that checks that the right elements have been assigned?

I was afraid somebody would ask... Done now.

package/MDAnalysis/topology/tpr/utils.py
jbarnoud: I am not sure about that one. In normal operation, there should not be a warning. A virtual site in your water model or a coarse-grained models are normal operation. Also, the whole point of using a TPR is that there is nothing to guess.

IAlibay: So in all honesty when I wrote that comment I somehow had convinced myself that there were cases where you would have negative atomnumber values when doing antechamber -> parmed -> gmx conversions. But looking back at a few gmx dump outputs I was (happily) mistaken.

If the negative numbers only commonly occurs in CG or EP situations then I'm happy going without warning.

package/CHANGELOG
```diff
@@ -57,6 +57,7 @@ Enhancements
   * Added Hydrogen Bond Lifetime via existing autocorrelation features (PR #2791)
   * Added Hydrogen Bond Lifetime keyword "between" (PR #2791)
   * Dead code removed from the TPR parser and increased test coverage (PR #2840)
+  * TPR parser exposes the elements topology attribute (PR #2858)
```
IAlibay, line 60: Issue number could be useful here too.

package/MDAnalysis/topology/TPRParser.py
```diff
@@ -110,10 +110,10 @@
 The TPR reader is a pure-python implementation of a basic TPR
 parser. Currently the following sections of the topology are parsed:
 
-* Atoms: number, name, type, resname, resid, segid, mass, charge,
+* Atoms: number, name, type, resname, resid, segid, mass, charge, element
   [residue, segment, radius, bfactor, resnum, moltype]
 * Bonds
-* Angels
```
lilyminium, line 116: 😇

testsuite/MDAnalysisTests/topology/test_tprparser.py
```diff
@@ -258,3 +258,32 @@ def test_fail_for_unsupported_files(tpr_path, expected_exception):
     with pytest.raises(expected_exception):
         parser.parse()
 
+
+@pytest.mark.parametrize('tpr_path', BONDED_TPRS)
```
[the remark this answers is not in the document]
lilyminium: I think `BONDED_TPRS` would be an iterable of TPR files here.

package/MDAnalysis/topology/tpr/obj.py
```diff
@@ -112,6 +114,21 @@ def __init__(self, id, name, type, resid, resname, mass, charge):
...
         self.mass = mass
         self.charge = charge
+        self.atomic_number = atomic_number
+
+    @property
+    def element_symbol(self):
+        """
+        The symbol of the atom element.
+
+        Returns
```
lilyminium, line 124: It's odd to think of a property "returning" something, but on the other hand very few people will see this

package/MDAnalysis/topology/tpr/obj.py
```diff
@@ -112,6 +114,21 @@ def __init__(self, id, name, type, resid, resname, mass, charge):
...
+        -------
+        str:
+            The symbol corresponding to the atomic number. If the atomic number
+            is not recognized, which appends if a particle is not really an
```
lilyminium, line 128:
```suggestion
            is not recognized, which happens if a particle is not really an
```

testsuite/MDAnalysisTests/topology/test_tprparser.py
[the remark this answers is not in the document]
IAlibay: Makes sense, my bad.

jbarnoud:
@lilyminium @IAlibay Thank you for the review. I think I addressed all your comments.

orbeckst:
@IAlibay @lilyminium thank you for reviewing. I just put you down as the official assignees so one of you has the honor of merging.

PR #2723, opened 2020-06-09
Serialize FileIO and TextIOWrapper and Universe

yuxuanzhuang:
Fixes #2878
 
- Composition over inheritance for pickling `Universe` in #2704

Changes made in this Pull Request:
 - add new picklable BufferedReader, FileIO, and TextIOWrapper classes.
 - add new classes and `pickle_open` function to `picklable_file_io.py`
 - implement `__getstate__`, `__setstate__` to `Universe` and `BaseReader`
 - fix DCD, XDR pickle issue

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?
 - [x] Issue raised/referenced?

PR #2859, opened 2020-07-20
Refactor Transformations from closure into class

### LINKED ISSUE #2860: Refactor all Transformations into class
To make pickling transformation possible. Currently, due to the nature of python closure, `Transformations` cannot be pickled.

Refactor those into class object, with `__call__` function.

yuxuanzhuang:
Fixes #2860

 - refactor transformations from closure into class

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/transformations/fit.py
```diff
@@ -82,38 +81,42 @@ def fit_translation(ag, reference, plane=None, weights=None):
...
-        axes = {'yz' : 0, 'xz' : 1, 'xy' : 2}
+    def __init__(self, ag, reference, plane=None, weights=None):
+        self.ag = ag
+        self.reference = reference
+        self.plane = plane
+        self.weights = weights
+
+
+    def __call__(self, ts):
+        if self.plane is not None:
```
richardjgowers, line 92: to be equivalent, `__call__` should only have the contents of `wrapped`, all other logic should go into `__init__`

yuxuanzhuang: Makes sense!

yuxuanzhuang:
So for most parts I just moved code around, the only change is `PositionAverager`, where it now gets the same result when calling the same timestep. (to suit we call u.trajectory[frame] during pickling.

```python
trans = PositionAverager(3, check_reset=True)
u = mda.Universe(TPR, XTC)
u.trajectory.add_transformations(trans)
u.trajectory[3]
print(u.atoms.positions[0])
u.trajectory[3]
print(u.atoms.positions[0])
u.trajectory[3]
print(u.atoms.positions[0])
```
- Before
[27.95  20.195 13.645]
[55.9  40.39 27.29]
[55.9  40.39 27.29]
/home/scottzhuang/mdanalysis/package/MDAnalysis/transformations/positionaveraging.py:177: Warning: Cannot average position for non sequentialiterations. Averager will be reset.
  warnings.warn('Cannot average position for non sequential'
- Now
[53.96     41.975002 29.420002]
[53.96     41.975002 29.420002]
[53.96     41.975002 29.420002]

orbeckst:
There's a segfault in a pickle transformation test https://travis-ci.com/github/MDAnalysis/mdanalysis/jobs/365340906 – do you have an idea what's happening there?

orbeckst:
There are also test failures related to pickle transformations on WIndows, eg https://dev.azure.com/mdanalysis/mdanalysis/_build/results?buildId=658&view=logs&j=6337f8de-d220-572c-a53d-20c279e44ba9&t=38e5c318-25f5-5156-9fa7-6cd72ffb271c&l=99

yuxuanzhuang:
I am not sure why it's happening, but the reason is that I define a universe (PDB, XTC) at the top level of the file (line 44). I guess some specific version of python/numpy/? is not quite happy with that. 
```python
=================================== ERRORS ====================================
_ ERROR collecting MDAnalysisTests/parallelism/test_pickle_transformation.py __
MDAnalysisTests\parallelism\test_pickle_transformation.py:44: in <module>
    u = mda.Universe(TPR, XTC)
C:\hostedtoolcache\windows\Python\3.7.8\x64\lib\site-packages\mdanalysis-2.0.0.dev0-py3.7-win-amd64.egg\MDAnalysis\core\universe.py:386: in __init__
    in_memory_step=in_memory_step, **kwargs)
C:\hostedtoolcache\windows\Python\3.7.8\x64\lib\site-packages\mdanalysis-2.0.0.dev0-py3.7-win-amd64.egg\MDAnalysis\core\universe.py:578: in load_new
    self.trajectory = reader(filename, format=format, **kwargs)
C:\hostedtoolcache\windows\Python\3.7.8\x64\lib\site-packages\mdanalysis-2.0.0.dev0-py3.7-win-amd64.egg\MDAnalysis\coordinates\XDR.py:149: in __init__
    self._load_offsets()
C:\hostedtoolcache\windows\Python\3.7.8\x64\lib\site-packages\mdanalysis-2.0.0.dev0-py3.7-win-amd64.egg\MDAnalysis\coordinates\XDR.py:194: in _load_offsets
    data = read_numpy_offsets(fname)
C:\hostedtoolcache\windows\Python\3.7.8\x64\lib\site-packages\mdanalysis-2.0.0.dev0-py3.7-win-amd64.egg\MDAnalysis\coordinates\XDR.py:84: in read_numpy_offsets
    return {k: v for k, v in np.load(filename).items()}
C:\hostedtoolcache\windows\Python\3.7.8\x64\lib\site-packages\numpy\lib\npyio.py:444: in load
    raise ValueError("Cannot load file containing pickled data "
E   ValueError: Cannot load file containing pickled data when allow_pickle=False
```
After I changed it to DCD, it's fine now.

orbeckst:
Uh, oh, ... I think @IAlibay and @lilyminium have been chasing an elusive segfault that seems to be connected to numpy and xtc. Just pinging so that they can comment.

package/MDAnalysis/transformations/__init__.py
```diff
@@ -53,7 +53,30 @@ def wrapped(ts):
...
+            self.needed_var = args[0]
+
+        def __call__(self, ts):
+            #  apply changes to the Timestep object
+            return ts
+
+
+Note it does not mean that the old closure/wrapped function implementation will fail.
+They can still be used if one does not need serialization, which is a prerequisite for parallel
+analysis.
```
orbeckst, line 75: If a new reader reads this, will they know what "old" means? This should be written from the perspective of the present time. You might have to explain that you can also write functions (closures) as in MDA 1.x but that these cannot be serialized and thus will not work with all forms of parallel analysis.

package/MDAnalysis/transformations/__init__.py
```diff
@@ -53,7 +53,30 @@ def wrapped(ts):
 
 See `MDAnalysis.transformations.translate` for a simple example.
 
+To meet the need of serialization of universe, transformations are converted
```
orbeckst, line 56: The conversion is not really relevant to a user of the library – they deal with what's there. You can add a comment to the code talking about the thinking and motivation or you can write a specific developer section. But as I said below, the text needs to make sense to someone who comes to the code without the historical knowledge. Try to put yourself into the shoes of someone who sees this the first time.

package/MDAnalysis/transformations/__init__.py
```diff
@@ -26,40 +26,17 @@
...
-
-To meet the need of serialization of universe, transformations are converted
-into classes. They retain the similar API and functionality by implementing the
-aforementioned `wrapped(ts)` as a special method `__call__`. For example:
+A typical transformation class looks like this:
 
 .. code-blocks:: python
 
-    class transfomrations(object):
+    class transfomration(object):
```
orbeckst, line 39:
```suggestion
   class transformation(object):
```

package/MDAnalysis/transformations/__init__.py
```diff
@@ -69,14 +46,50 @@ def __call__(self, ts):
...
 
-Note it does not mean that the old closure/wrapped function implementation will fail.
-They can still be used if one does not need serialization, which is a prerequisite for parallel
-analysis.
+Note, to meet the need of serialization of universe, only transformation class
+are used after MDAnlaysis 2.0.0. One can still write functions (closures) as in
+MDA 1.x, but that these cannot be serialized and thus will not work with all
+forms of parallel analysis. For detailed descriptions about how to write a
+closure-style transformation, read the code in MDA 1.x as a reference
+or read MDAnalysis UserGuide.
```
orbeckst, line 87:
```suggestion

.. Note::
   Although functions (closures) work as transformations, they are not used in
   in MDAnalysis from release 2.0.0 onwards because they cannot be reliably
   serialized and thus a :class:`Universe` with such transformations cannot be 
   used with common parallelization schemes (e.g., ones based on 
   :mod:`multiprocessing`). 
   For detailed descriptions about how to write a closure-style transformation, 
   please refer to MDAnalysis 1.x documentation.
```

package/MDAnalysis/transformations/__init__.py
orbeckst: still a typo, if you take the suggestion then it is fixed

orbeckst:
ping @richardjgowers – this could go in right after PR #2723 if you're happy.

orbeckst:
@yuxuanzhuang @richardjgowers let's try to get the CI to pass again (when #2890 is fixed) before finally merging this one, just to make 100% sure. 

@richardjgowers in the mean time, anything else that needs to be done here or can you approve?

package/MDAnalysis/transformations/__init__.py
```diff
@@ -26,34 +26,72 @@
...
+modify the trajectory.
+Coordinate transformations, such as PBC corrections and molecule fitting
+are often required for some analyses and visualization, and the functions in
+this module allow transformations to be applied on-the-fly.
+
+A typical transformation class looks like this:
+
+.. code-blocks:: python
+
+    class transformation(object):
```
richardjgowers, line 39: It's kind of odd to have class names lowercase, normally this is uppercase for python, but I get that it's to preserve backwards compatibility

package/MDAnalysis/transformations/__init__.py
```diff
@@ -26,34 +26,72 @@
...
+        def __init__(self, *args, **kwargs):
+            #  do some things
+            #  save needed args as attributes.
+            self.needed_var = args[0]
```
richardjgowers, line 43: This is a little abstract, a trivial example would be better

package/MDAnalysis/transformations/__init__.py
```diff
@@ -26,34 +26,72 @@
...
+
+        def __call__(self, ts):
+            #  apply changes to the Timestep object
```
richardjgowers, line 46: Changes needn't be applied to the timestep oddly enough. You could E.g. modify an AtomGroup and return ts

package/MDAnalysis/transformations/__init__.py
```diff
@@ -26,34 +26,72 @@
...
+            return ts
+
+See `MDAnalysis.transformations.translate` for a simple example.
```
richardjgowers, line 49: The problem is they can't see the code from the docs, so instead have the above example be translate

package/MDAnalysis/transformations/__init__.py
```diff
@@ -26,34 +26,72 @@
...
-contain a wrapped function that takes a `Timestep` object as argument.
-So, a transformation can be roughly defined as follows:
+.. code-block:: python
+
+    u = mda.Universe(GRO, XTC)
+    ts = u.trajectory[0]
+    trans = transformation(args)
+    ts = trans(ts)
+
+    #  or add as a workflow
```
richardjgowers, line 64: I think we only want to promote the workflow usage, else it's just a function they can (forget to) apply

richardjgowers (dismissed):
Code looks good, docs need tweaking

package/MDAnalysis/transformations/__init__.py
yuxuanzhuang: Yeah I was hesitating about that. But I guess it's fine to use lowercase for callable classes according to PEP 8:
>Class Names
Class names should normally use the CapWords convention.
The naming convention for functions may be used instead in cases where the interface is documented and used primarily as a callable.

And for backwards compatibility of course.

package/MDAnalysis/transformations/__init__.py
yuxuanzhuang: I think this part (inside `__init__`) of the doc is not present anywhere in the doc webpages. A more explicit description of how transformation is done is written separately. By which I guess makes it fine to be a little more abstract?

package/MDAnalysis/transformations/__init__.py
orbeckst: Especially with the 
> The naming convention for functions may be used instead in cases where the interface is documented and used primarily as a callable.
I am totally ok with lower case.

orbeckst:
ping @richardjgowers for re-review

yuxuanzhuang:
Met a problem when I was wrapping things up. After we change the behaviour of `AtomGroup` pickling, one problem shows up:

```python
fit_trans = trans.fit_rot_trans(u.atoms, u.atoms)
u.trajectory.add_transformations(fit_trans)
u_p = pickle.loads(pickle.dumps(u))
<ipython-input-8-e66d2dd7c314> in <module>
----> 1 u_p = pickle.loads(pickle.dumps(u))

~/mdanalysis/package/MDAnalysis/core/groups.py in _unpickle(u, ix)
    113 def _unpickle(u, ix):
    114     print(u.__dict__)
--> 115     return u.atoms[ix]
    116 
    117 

AttributeError: 'Universe' object has no attribute 'atoms'
```
Not totally sure the exact reason. Seems that in the new process, transformations (in which have `AtomGroup`) are unpickled when `Universe` is still empty. Changing back to `__reduce__` can fix this pickle ordering(?) problem.

package/MDAnalysis/core/universe.py
```diff
@@ -687,15 +687,21 @@ def __repr__(self):
...
 
-    def __getstate__(self):
-        # Universe's two "legs" of topology and traj both serialise themselves
-        return self._topology, self._trajectory
+    @classmethod
+    def _unpickle_U(cls, top, traj):
+        """Special method used by __reduce__ to deserialise a Universe"""
+        #  top is a Topology obj at this point, but Universe can handle that.
+        u = cls(top)
+        u.trajectory = traj
```
orbeckst, line 695: Are you missing `return u`?

package/MDAnalysis/core/universe.py
```diff
@@ -687,15 +687,21 @@ def __repr__(self):
...
 
-    def __setstate__(self, args):
-        self._topology = args[0]
-        _generate_from_topology(self)
+        return u
```
orbeckst, line 697: oops, there it is... confusing diff

package/MDAnalysis/transformations/__init__.py
```diff
@@ -26,34 +26,77 @@
...
+            #  do some things
+            #  save needed args as attributes.
+            self.needed_var = args[0]
+
+        def __call__(self, ts):
+            #  apply changes to the Timestep,
+            #  or modify an AtomGroup and return Timestep
+
+            return ts
+
```
orbeckst, line 50: You showed the abstract class. Now show a concrete example to address @richardjgowers comment. For instance
```suggestion
As a concrete example we will write a transformation that rotates a group of atoms around 
the z-axis through the center of geometry by a fixed increment for every time step. We will use
:meth:`MDAnalysis.core.groups.AtomGroup.rotateby` and simply increment the rotation angle
every time the transformation is called::

  class spin_atoms(object):
      def __init__(self, atoms, dphi):
         """Rotate atoms by dphi degrees for every time step (around the z axis)"""
         self.atoms = atoms
         self.dphi = dphi
         self.axis = np.array([0, 0, 1])
         self.phi = 0
         
     def __call__(self, ts):
         self.atoms.rotateby(self.phi, self.axis)
         return ts

This transformation can be used as ::
  
   u = mda.Universe(PSF, DCD)
   u.trajectory.add_transformations(spin_atoms(u.select_atoms("protein"), 1.0))

```

I currently can't get nglview to work in my notebook so you'll need to check that this actually works... or come up with another example.

orbeckst:
And yeah, I don't know why we need to use `__reduce__`... are there downsides to doing so instead of `__get/setstate__`?

orbeckst:
Does it still work with dask?

yuxuanzhuang:
> And yeah, I don't know why we need to use `__reduce__`... are there downsides to doing so instead of `__get/setstate__`?

No, I think it is just low-level func for `__get/setstate__`. And it still works will dask.

On the other hand, for performance, it seems that the `Universe` is pickled twice in the presence of transformation.
```python
>>> u = mda.Universe(PSF, DCD)
>>> fit_trans = trans.fit_rot_trans(u.atoms, u.atoms)
>>> u.trajectory.add_transformations(fit_trans)

>>> dmp = pickle.dumps(u)
run `__reduce__` (Universe)
run `__reduce__` (AtomGroup)
run `__reduce__` (Universe)
>>> u_p = pickle.loads(dmp)
run `_unpickle_U`
run `unpickle AtomGroup`
run `_unpickle_U`
```

I have to admit I am a bit confused here...when pickling intertwined stuff, e.g when we pickle them apart, it's not an issue:
```python
>>> u = mda.Universe(PSF, DCD)
>>> fit_trans = trans.fit_rot_trans(u.atoms, u.atoms)

>>> dmp = pickle.dumps([u, fit_trans])  #  Note they are not linked together here.
run `__reduce__` (AtomGroup)
run `__reduce__` (Universe)
>>> u_p = pickle.loads(dmp)
run `_unpickle_U` 
run `unpickle AtomGroup`
```

package/MDAnalysis/transformations/__init__.py
yuxuanzhuang: Thanks! it works. And I think it makes sense to just rotate by a fixed angle. (for visualization perhaps)

orbeckst: I changed the snippet.

orbeckst:
Would be good to understand the path the code takes when it is pickling Universe twice. Even if it were not a performance issue, it is a potential bug somewhere down the line. Btw, is the Universe stored twice in the pickle?

yuxuanzhuang:
I think it's a normal behavior of python handling circular reference pickling.
```python
class Universe(object):
    def __init__(self, top):
        self.top = top
        
    def add_trajectory(self,traj):
        self.traj = traj
        
    def __reduce__(self):
        print('reduce')
        return self._unpickle_u, (self.top, self.traj)
    
    @classmethod
    def _unpickle_u(cls, top, traj):
        print('unpickle')
        u = cls(top)
        u.traj = traj
        return u

class Topology(object):
    def __init__(self):
        pass
    
class Trajectory(object):
    def __init__(self):
        pass
        
    def add_transformation(self, transformation):
        self.transformation = transformation
        
class Transformation(object):
    def __init__(self, atomgroup):
        self.atomgroup = atomgroup
        
class Atomgroup(object):
    def __init__(self, universe):
        self.universe = universe

top = Topology()
traj = Trajectory()
u = Universe(top)
u.add_trajectory(traj)
ag = Atomgroup(u)
trans = Transformation(ag)
traj.add_transformation(trans)

>>> import pickle
>>> u_dmp = pickle.dumps(u)
>>> u_p = pickle.loads(u_dmp)
reduce
reduce
unpickle
unpickle

File: package/MDAnalysis/transformations/__init__.py [hunk 14]
```diff
@@ -48,7 +50,32 @@ def __call__(self, ts):
...
+            self.atoms.rotateby(self.phi, self.axis)
+            self.phi += self.dphi
+            return ts
+
+This transformation can be used as ::
+
+   u = mda.Universe(PSF, DCD)
+   u.trajectory.add_transformations(spin_atoms(u.select_atoms("protein"), 1.0))
+
+Alos see :mod:`MDAnalysis.transformations.translate` for a simple example.
```
orbeckst, line 78:
```suggestion
Also see :mod:`MDAnalysis.transformations.translate` for a simple example.
```

orbeckst (approved):
Thanks for looking at the pickling order. If we only store the data once then I think that's not a problem if it gets pickled twice.

I found a minor typo in the docs (see suggestion). Just correct it. I am already approving the PR.

orbeckst:
@richardjgowers , were your comments addressed?

PR #2861, opened 2020-07-20
Numpy 1.19

IAlibay:
Investigates #2818 

 - Bump up numpy version from 1.18.5 (stable) to 1.19
 - try using both cores for travis

Just a summary:

  - Using -n2 with pytest-xdist did not yield significant changes in time: https://travis-ci.com/github/MDAnalysis/mdanalysis/jobs/362839273, it's not immediately clear to me what's going on here, further tests to come.

  - Switching to numpy 1.19 leads to instabilities for some parts of the code  (particularly trajectory aligning for XTC files).

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


IAlibay:
Looks like the segfaults are not consistent, but essentially on Travis' linux image you either end up with a segfault failure when running the first iteration of `test_rms.py::test_rmsf_xtc` (a test I added that usually fails downstream), or `test_chainreader.py::TestChainReaderFormats::test_set_all_format_tuples`

orbeckst:
Running the tests in 13:51 minutes is encouraging – maybe pytest-xdist `-n 2` is working after all.

orbeckst:
Perhaps only run the segfaulting tests (`pytest -k test_rmsf_xtc` or similar) and then run the whole thing in `gdb` to see where exactly the segfault happens? Or if we get a core file, run gdb on the core? (Not sure how to properly do any of these things but it would be good to know if the segfault originates inside the XDR code.)

Is the segfault related to repeated opening and closing of an XTC file? If so, there might be an issue with the `__del__` method, which has always been iffy (for all readers) and perhaps a subsequent attempt to double-free something.

Finally, is the travis image public and could one run it locally to debug inside the docker image?

IAlibay:
closed as stale - might still be a bug here, to be investigated some other time.

Issue #1387, opened 2017-06-06
Multiple segments reduce to one segid


tclick:
Using the MDAnalysis 0.16.1, a multisegmented system with overlapping residue numbers is reduced to a single segid by usage of topology.base.squash_by. In MDAnalysis 0.10.0, this did not occur because the topology handling system was different. Users who work with complex systems will want to have their systems separated by segments as well as residues and atoms.

As noted above, the root cause is in the squash_by function, which uses numpy.unique; if a list of resids is provided, then multisegmented systems with overlapping residues will reduce to the first instance within the first segment.

richardjgowers:
Hi Timothy

You're right around squash_by, but we've (I've) been trying to use
change_squash more recently which allows for better splitting of things
across multiple segments etc.

What format is it that you work with? I'll check how we're handling it.

tclick:
I've tried both PSF and PDB.

richardjgowers:
@tclick sorry for the delay, can you provide a toy example of a psf file that chokes?  I'll be able to push a fix in 0.16.2 that way. Thanks

tclick:
Sorry about the delay, but here is a PSF and CRD file. Also, I noticed that the segments are also alphabetized, which may not always correspond to the original setup (as in my case of PDZ3 and CRPT). 

[structure.zip](https://github.com/MDAnalysis/mdanalysis/files/1080584/structure.zip)

orbeckst:
@richardjgowers have you had a chance to look at this issue in more detail? It sounds like a regression that we should fix for 0.17.0.

tclick:
I was using the PSF topology. One way to counter this (as I have learned), might be to use

return atom_idx, unique_resids, [attr[np.argsort(sort_mask)] for attr in attributes]

The following topology files contain “squash_by” when doing a grep:

DMSParser.py:from .base import TopologyReaderBase, squash_by
DMSParser.py:        atom_residx, res_resids, (res_resnames, res_segids) = squash_by(
DMSParser.py:        res_segidx, seg_segids = squash_by(
LAMMPSParser.py:from .base import TopologyReaderBase, squash_by
LAMMPSParser.py:        residx, resids = squash_by(resids)[:2]
MOL2Parser.py:from .base import TopologyReaderBase, squash_by
MOL2Parser.py:        residx, resids, (resnames,) = squash_by(
PQRParser.py:from .base import TopologyReaderBase, squash_by, change_squash
[...]

PR #2872, opened 2020-07-26
Implemented extraction of chains from DMS formats

ianmkenney:
Used PDBParser as a reference

Addresses #1387 

 - Repeated residues are not squashed
 - Multiple segids are created when multiple chains are present

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


ianmkenney:
@richardjgowers I'm creating an altered ADK dms (`adk_closed_domains.dms`) from the already existing `adk_closed.dms`. Domain definitions are from:

Seyler SL, Kumar A, Thorpe MF, Beckstein O (2015) Path Similarity Analysis: A Method for Quantifying Macromolecular Pathways. PLoS Comput Biol 11(10): e1004568. https://doi.org/10.1371/journal.pcbi.1004568

Setting the domains as follows from sqlite3:
```sql
UPDATE particle SET segid = "CORE";

UPDATE particle 
SET segid = "NMP" 
WHERE resid BETWEEN 30 AND 59;

UPDATE particle 
SET segid = "LID" 
WHERE resid BETWEEN 122 AND 159;
```

A check on the selections of those domains should be sufficient for tests, right?

ianmkenney:
actually, I'm just going to edit `adk_closed.dms` directly. It should work better with the class structure.

ianmkenney:
@richardjgowers should be ready for a review

package/MDAnalysis/topology/DMSParser.py
```diff
@@ -176,21 +176,43 @@ def dict_factory(cursor, row):
...
 
-        # Segments
-        res_segidx, seg_segids = squash_by(
-            res_segids)[:2]
-        topattrs.append(Segids(seg_segids))
+        if any(res_segids) and not any(val is None for val in res_segids):
+            res_segidx, (res_segids,) = change_squash((res_segids,),
+                                                      (res_segids,))
+
+            uniq_seg = np.unique(res_segids)
```
orbeckst, line 199: Are we normally using our own `lib.util.unique_int_1d` https://github.com/MDAnalysis/mdanalysis/blob/2550d061a3134e0002fc8ec4eb31eb3c63bcfa5b/package/MDAnalysis/lib/_cutil.pyx#L52 instead of `np.unique` for performance reasons, @richardjgowers @jbarnoud @tylerjereddy ?

Or will it not matter here because `res_segids` is small?

ianmkenney: res_segids should be pretty small most of the time. The only time I could see it becoming a problem is if for whatever reason the DMS was structures so that similar lipids/solvent molecules were not grouped in the database. It would have to be extremely disorganized to cause performance issues I would think.

orbeckst: Ok, then I think we just leave it.

orbeckst:
Thanks @ianmkenney !

package/MDAnalysis/topology/DMSParser.py
tylerjereddy: Yeah performance, I wanted to get rid of our own version because it required maintenance to make 32-bit compatible but I think @lilyminium pointed out that it was a bit faster I guess. The NumPy one has quite a few additional features so that's probably part of the difference.

PR #2883, opened 2020-07-31
Add SMARTS selection

cbouy:
Part of #2468 
Depends on #2775 

 - Adds atom selection based on SMARTS queries: `u.select_atoms("[C;R;!a]", smarts=True)`

An example with Proline:
image

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

@richardjgowers @IAlibay @fiona-naughton


cbouy:
Since some SMARTS have parenthesis in them I'd need to find a way to bypass the "split original selection string on spaces and parenthesis" thing, but should be doable

cbouy:
Done!

package/MDAnalysis/core/selection.py
```diff
@@ -600,6 +600,38 @@ def apply(self, group):
...
+        except ImportError:
+            raise ImportError("RDKit is required for SMARTS-based atom "
+                              "selection but it's not installed. Try "
+                              "installing it with \n"
+                              "conda install -c conda-forge rdkit")
+        pattern = Chem.MolFromSmarts(self.pattern)
+        if not pattern:
+            raise ValueError("{!r} is not a valid SMARTS query".format(
+                self.pattern))
+        mol = group.convert_to("RDKIT")
```
richardjgowers, line 626: Might be nice to try and cache this in Topology.

package/MDAnalysis/core/selection.py
```diff
@@ -600,6 +600,38 @@ def apply(self, group):
...
+        matches = mol.GetSubstructMatches(pattern, useChirality=True)
+        indices = [idx for match in matches for idx in match]
+        indices = [mol.GetAtomWithIdx(i).GetIntProp("_MDAnalysis_index")
```
richardjgowers, line 629: Can we not predict what idx RDKit will assign? Ie do the atoms sometimes get reordered?

richardjgowers:
This is pretty awesome btw :)

package/MDAnalysis/core/selection.py
```diff
@@ -600,6 +600,38 @@ def apply(self, group):
         return group[group.aromaticities].unique
 
 
+class SmartsSelection(Selection):
+    """Select atoms based on SMARTS queries"""
+    token = 'smarts'
+
+    def __init__(self, parser, tokens):
+        pattern = []
```
richardjgowers, line 608: Is it simpler to enforce no white space allowed in the pattern?

package/MDAnalysis/core/selection.py
cbouy: The mol (topology, not coordinates) is already cached in the converter so that we don't need to rebuild it for every frame of a trajectory:
https://github.com/cbouy/mdanalysis/blob/rdkit-converter/package/MDAnalysis/coordinates/RDKit.py#L253-L264

package/MDAnalysis/core/selection.py
cbouy: They get reordered when using reactions to "standardize" some functional groups like nitro, sulfone...etc into the correct form. I could also force reordering atoms at the end of the conversion but that requires building a new molecule again I think.

package/MDAnalysis/core/selection.py
cbouy: White spaces are added automatically by the parser around parentheses:
https://github.com/cbouy/mdanalysis/blob/rdkit-converter/package/MDAnalysis/core/selection.py#L1182

jbarnoud:
That is a super cool feature!

I did not read the code so maybe it is obvious and super well documented, but I have 2 questions:
* what happens if I do not have RDKit installed?
* what happens if the SMART string matches in several places? What is these places overlap?

cbouy:
> what happens if I do not have RDKit installed?

It raises an ImportError with a message that you need RDKit to use the SMARTS selection

> what happens if the SMART string matches in several places? What if these places overlap?

You get a flat and "uniqued" AtomGroup with all atoms matching the query.

For example, with the molecule `C(C=O)(C=C)C`:
image

RDKit returns a tuple of matches for the query, i.e. if you query `C-C` it will return `((0, 1), (0, 3), (0, 5))`
`u.select_atoms("smarts C-C").indices` will return `array([0, 1, 3, 5])`

There's a limit of 1000 matches (the default in RDKit).

Maybe it could be useful to add a kwarg to `select_atoms` so that it yields each match as a separate AtomGroup (i.e. [0, 1], then [0, 3]...etc.) ?

jbarnoud:
> Maybe it could be useful to add a kwarg to select_atoms so that it yields each match as a separate AtomGroup (i.e. [0, 1], then [0, 3]...etc.) ?

For API consistency's sake, it would be better to have an other method to do this. I think it would be very useful, but it is clearly for later.

cbouy:
I agree that it could get confusing, a separate method is probably better

```python
for ag in u.smarts_matches("C-C"):
    print(ag.indices)
```

package/MDAnalysis/core/groups.py
```diff
@@ -2689,6 +2689,10 @@ def select_atoms(self, sel, *othersel, **selgroups):
             record_type *record_type*
                 for selecting either ATOM or HETATM from PDB-like files.
                 e.g. ``select_atoms('name CA and not record_type HETATM')``
+            smarts *SMARTS-query*
```
richardjgowers, line 2692: So with atom selections, there's two sets of documentation.  Somewhere there's a `.rst` file with essentially these docs duplicated but expanded a bit.  Can you find these and document `smarts` there? (Unless I'm wrong @lilyminium ?)

lilyminium: Yeah there's `selections.rst` in `sphinx/documentation_pages`; redundantly, there's also a separate copy in the user guide :/ we should pick which one we like and point to it instead of having 3 versions. I think the user guide *might* be the most up-to-date one at this point?

fiona-naughton (changes requested):
Just one comment from me, otherwise looking good!

package/MDAnalysis/core/selection.py
```diff
@@ -600,6 +600,63 @@ def apply(self, group):
...
+            # keep track of the number of open and closed parentheses
+            if tokens[0] in ["(", ")"]:
+                counter[tokens[0]] += 1
+            # if the next char is a closing ")" but there's no corresponding
+            # open "(" then we've reached then end of the smarts query and
+            # the ")" is part of a grouping parenthesis
+            if tokens[1] == ")" and counter["("] < (counter[")"] + 1):
+                val = tokens.popleft()
+                pattern.append(val)
+                break
```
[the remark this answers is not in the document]
cbouy: you're right my bad!

IAlibay (approved):
Assuming CI returns green lgtm!

pinging @fiona-naughton and @richardjgowers for re-review, we can probably get this merged today/tomorrow.

richardjgowers (approved):

PR #1781, opened 2018-02-22
Add reindex option to write atom id to gro file

xiki-tempula:
Fixes #1777 
Can I have some advice on which doc should I change? 
Should I only change https://www.mdanalysis.org/docs/documentation_pages/coordinates/GRO.html
or should I mention it on other writer pages? 
Suggestions are welcomed.

Changes made in this Pull Request:
 - 
```python
from MDAnalysis import Universe
from six import StringIO
gro = '''test
1
    2CL      CL20850   0.000   0.000   0.000
7.29748 7.66094 9.82962'''
u = Universe(StringIO(gro), format='gro')
u.atoms[0].id = 3
u.atoms.write('temp.gro', reindex=False)
```
Gives
```
test
1
    2CL      CL    3   0.000   0.000   0.000
7.29748 7.66094 9.82962
```

While
```python
from MDAnalysis import Universe
from six import StringIO
gro = '''test
1
    2CL      CL20850   0.000   0.000   0.000
7.29748 7.66094 9.82962'''
u = Universe(StringIO(gro), format='gro')
u.atoms[0].id = 3
u.atoms.write('temp.gro', reindex=True)
```
Gives
```
test
1
    2CL      CL    1   0.000   0.000   0.000
7.29748 7.66094 9.82962
```

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?
 - [ ] Issue raised/referenced?

PR #2886, opened 2020-08-06
Add a reindex argument to the PDBWritter

### LINKED ISSUE #1072: Add option to keep original Atom IDs when writing PDB
### Expected behaviour

When I use `ag.write('foo.pdb')` I expect the atom ids to be the same as in the AtomGroup

### Actual behaviour

The atom ids's always start from 0.

``` python
import MDAnalysis as mda

u = mda.Universe(PDB)
bb = u.select_atoms('backbone')
bb.write('foo.pdb')
u2 = mda.Universe('foo.pdb')
assert np.all(bb.ids == u2.atoms.ids)
```

develop branch

jbarnoud:
Fixes #1072

 - Add a `reindex` argument to the `PDBWriter`. The argument is `True` by default and the atom serials are generated starting from 1. If the argument is set to `False`, then `atoms.ids` is used for the serial. This is reminiscent of #1781.

PR Checklist
------------
 - [x] Tests?
 - [X] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/coordinates/PDB.py
+++ b/package/MDAnalysis/coordinates/PDB.py
@@ -827,7 +827,6 @@ class PDBWriter(base.WriterBase):
 
         con = collections.defaultdict(list)
         for a1, a2 in bonds:
-            print(a1, a2)
             if not (a1 in mapping and a2 in mapping):
                 continue
             con[a2].append(a1)
--- a/package/MDAnalysis/coordinates/PDB.py
+++ b/package/MDAnalysis/coordinates/PDB.py
@@ -534,7 +534,8 @@ class PDBWriter(base.WriterBase):
        An indexing issue meant it previously used the first charater (Issue #2224)
 
     .. versionchanged:: 2.0.0
-        Add the "reindex" argument.
+        Add the `redindex` argument. Setting this keyword to ``True``
+        (the default) preserves the behavior in earlier versions of MDAnalysis.
 
     """
     fmt = {
--- a/package/MDAnalysis/coordinates/PDB.py
+++ b/package/MDAnalysis/coordinates/PDB.py
@@ -806,7 +806,10 @@ class PDBWriter(base.WriterBase):
         bondset = set(itertools.chain(*(a.bonds for a in self.obj.atoms)))
         if self._reindex:
             index_attribute = 'index'
-            mapping = {index: i for i, index in enumerate(self.obj.atoms.indices, start=1)}
+            mapping = {
+                index: i
+                for i, index in enumerate(self.obj.atoms.indices, start=1)
+            }
             atoms = np.sort(self.obj.atoms.indices)
         else:
             index_attribute = 'id'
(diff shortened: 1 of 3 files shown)


package/MDAnalysis/coordinates/PDB.py
```diff
@@ -533,6 +533,9 @@ class PDBWriter(base.WriterBase):
        ChainID now comes from the last character of segid, as stated in the documentation.
        An indexing issue meant it previously used the first charater (Issue #2224)
 
+    .. versionchanged:: 2.0.0
+        Add the "reindex" argument.
```
orbeckst, line 537: State which value of `reindex` preserves old (pre 2.0.0) behavior.

orbeckst: I think we put kwargs as bare reST roles i.e.
```reST
Add the `redindex` argument. Setting this keyword to ``True`` (the default) preserves
the behavior in earlier versions of MDAnalysis.
```

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -624,6 +627,9 @@ def __init__(self, filename, bonds="conect", n_atoms=None, start=0, step=1,
            ``False``: write a single frame to the file; ``True``: create a
            multi frame PDB file in which frames are written as MODEL_ ... ENDMDL_
            records. If ``None``, then the class default is chosen.    [``None``]
+        reindex: bool (optional)
+            If ``True`` (default), the atom serial is set to be consecutive
```
orbeckst, line 631: 1. Which value preserves compatibility with 1.x?
2. Have we ever been able to rountrip a PDB when the serials were not consecutively numbered starting from 1? Is this something we would want to be able to do? Is this something a user would generally expect?

orbeckst (commented):
Only some half-backed comments, sorry, not much time.

Does the PDB standard prescribe continuous serials starting from 1? (IIRC we never really found out but maybe someone knows.) 

What is the user's expectation: save PDB with exactly the given serials or save with something looking like a "normal" PDB?

If users want renumbered serials, should we have them renumber a serial attribute in the topology instead and then use that for writing instead of just doing something on the fly?

jbarnoud:
What we called the atom id is actually called atom serial, which let me assume it should be consecutive and starting at 1. In the discussion in the issue, @arose commented in that direction. So the default seems to be what is expected. However, it is unconvenient for some usecase, especially when patching systems together; hence the new keyword.

orbeckst:
Thanks for the explanation. Then this makes sense, with the default the most likely PDB-standard implementation and `reindex=False` for the special cases.

orbeckst (changes requested):
I am happy with the UI. Perhaps someone else can have a look at the implementation.

The test coverage is still a bit low, so that's my only complaint.

jbarnoud:
@orbeckst 

> The test coverage is still a bit low, so that's my only complaint.

With 100% patch coverage, it should be good now.

orbeckst (approved):

orbeckst:
Thanks, could you please address the merge conflict in CHANGELOG? Otherwise all good from my side.

PR #321, opened 2015-06-21
Lightweight serialization of AtomGroups by reutilization of built Universes

mnmelo:
AtomGroups can now be pickled/unpickled (closes #293)
A weakref system was implemented so that Universes can have a __del__
method without memleaking (closes #297)
ChainReaders no longer leak (and also no longer have a __del__  method)
(closes #312)
Tests now specifically look for uncollectable objects at the end. Breaks
parallelization, though.

These issues/enhancements all got interdependent at one point, hence the common commit.
I found a way to include memory leak checks at `teardown_package`, in `MDAnalysisTests.__init__.py`, but the inclusion of this function apparently keeps nose from parallelizing (n processes/threads are still started, but they run in only one core).
Ignored the 59 known (Issue #319) test errors related to DL_Poly.

Issue #297, opened 2015-06-04
__del__ destructor for Universes?


mnmelo:
The title says it all. @richardjgowers mentioned it'd be interesting to implement, but currently that leads to mem leaks as the garbage collector can't clean up the Universe.

See the discussion on #293.

If Universe has a __del__ I end up with two uncollectables: the Universe itself and the trajectory (which also has a __del__). I tried to find any sort of circular reference between the two, but found nothing.
Is it worth pursuing this?

richardjgowers:
So adding the del destructor does create a memory leak.. :)

mnmelo:
So, I think I was able to finally track this down. I always thought it was a problem of having the `Reader` also having a `__del__`, because that's apparently the easiest way to have uncollectable circular references. But the snippet below shows I was wrong:
```python
import gc
class Universe():
     def __init__(self):
         self.trajectory = Reader()
     def __del__(self):
         print "Universe is dying"

class Reader():
    def __init__(self):
        self.somevar = 3
    def __del__(self):
        print "Trajectory is dying"

u = Universe()
u = None
gc.collect() # This is empty, meaning all got properly collected
```

Now let's add some more realism to our toy Universe:
```python
import gc
class Universe():
     def __init__(self):
         self.trajectory = Reader()
         self.atom = Atom(self) # Having this or the following lines breaks garbage collection
         self.universe = self
     def __del__(self):
         print "Universe is dying"

class Reader():
    def __init__(self):
        self.somevar = 3
    def __del__(self):
        print "Trajectory is dying"

class Atom():
    def __init__(self, parent):
        self.universe = parent

u = Universe()
u = None
gc.collect()
gc.garbage # [<__main__.Universe instance at 0x1004a91b8>, <__main__.Reader instance at 0x1004a9200>]
```

As far as I know there's no way to clean this automatically. A user would not only have to set `u.universe = None` but also do so for each `Atom.universe`, `AtomGroup.universe`, `Residue.universe` etc. listed under the `Universe` object.
Hopeless?

richardjgowers:
https://docs.python.org/2/library/weakref.html

If we made all the references to Universe (from Atom etc) a `weakref.ref, it`might fix everything?

Can you try?
``` python
class Atom():
   self.universe = weakref.ref(parent)
```

mnmelo:
Ah, was just writing to propose that!
[...]

PR #2893, opened 2020-08-09
Serialization of AtomGroup

yuxuanzhuang:
Fixes #

 - `AtomGroup` now is pickled without finding its anchored `Universe`

PR Checklist
------------
 - [ ] Tests?
 - [x] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/core/groups.py
+++ b/package/MDAnalysis/core/groups.py
@@ -110,16 +110,7 @@ from . import topologyobjects
 from ._get_readers import get_writer_for, get_converter_for
 
 
-def _unpickle(uhash, ix):
-    try:
-        u = _ANCHOR_UNIVERSES[uhash]
-    except KeyError:
-        # doesn't provide as nice an error message as before as only hash of universe is stored
-        # maybe if we pickled the filename too we could do better...
-        errmsg = (f"Couldn't find a suitable Universe to unpickle AtomGroup "
-                  f"onto with Universe hash '{uhash}'. Availble hashes: "
-                  f"{', '.join([str(k) for k in _ANCHOR_UNIVERSES.keys()])}")
-        raise RuntimeError(errmsg) from None
+def _unpickle(u, ix):
     return u.atoms[ix]
 
 
@@ -2252,10 +2243,16 @@ class AtomGroup(GroupBase):
     :class:`AtomGroup` instances are always bound to a
     :class:`MDAnalysis.core.universe.Universe`. They cannot exist in isolation.
 
+    During serialization, :class:`AtomGroup` will be pickled with its bound
+    :class:`MDAnalysis.core.universe.Universe`. If multiple :class:`AtomGroup`
+    are bound to the same :class:`MDAnalysis.core.universe.Universe`, they
+    will bound to the same one after serialization.
+
 
     See Also
     --------
     :class:`MDAnalysis.core.universe.Universe`
+
 
     .. deprecated:: 0.16.2
        *Instant selectors* of :class:`AtomGroup` will be removed in the 1.0
@@ -2263,10 +2260,13 @@ class AtomGroup(GroupBase):
     .. versionchanged:: 1.0.0
        Removed instant selectors, use select_atoms('name ...') to select
        atoms by name.
+    .. versionchanged:: 2.0.0
+       :class:`AtomGroup` can always be pickled with or without its universe,
+       instead of failing when not finding its anchored universe.
     """
 
     def __reduce__(self):
-        return (_unpickle, (self.universe.anchor_name, self.ix))
+        return (_unpickle, (self.universe, self.ix))
 
     def __getattr__(self, attr):
         # special-case timestep info


yuxuanzhuang:
I may miss the point in #321 about lightweight serialization. Don't we always have to serialize (or recreate) a `Universe` in the new process? In that regard, there's no literal difference.

And due to the nature of `pickle`, one `Universe` instance will only be pickled once during serialization. So in the new process, all `AtomGroup` will still be bound to the same `Universe` (not the old one though, but I don't think it is a problem).

Besides, another advantage to ditch anchoring is: we no longer need to worry about the order of `Universe` and `AtomGroup` during serialization:
```python
>>> def print_id(ag, u):
>>>     print('id u:',id(u))
>>>     print('id ag._u:',id(ag._u))
>>>     print('')


>>> ag = u.atoms[2:5]
>>> print('before pickle:')
>>> print('id u:',id(u))
>>> print('id ag_u:',id(ag._u))
>>> print('')
>>> print('after pickle:')


>>> p = multiprocessing.Pool(2)
>>> i = 0
>>> res = np.array([p.apply(print_id, args=(ag, u))  #  this won't work in the old version with AtomGroup before Universe.
>>>             for ts in u.trajectory[:3]])
>>> p.close()

before pickle:
id u: 139955715000496
id ag_u: 139955715000496

after pickle:
id u: 139956140148912
id ag._u: 139956140148912

id u: 139956140148912
id ag._u: 139956140148912

id u: 139955676441232
id ag._u: 139955676441232

old version:

before pickle:
id u: 139990197591440
id ag_u: 139990197591440

after pickle:
id u: 139990190787024
id ag._u: 139990197591440

id u: 139990190788224
id ag._u: 139990197591440

id u: 139990202526352
id ag._u: 139990190787024
```

yuxuanzhuang:
I am not sure how it helps to prevent duplicated universes, could you elaborate a bit more?

Also I am not certain whether it's fine to ditch the anchor completely--all the tests still pass without that. 

With that said, in #297, it also mentioned that it helps the mem leaks issue. (is that tested anywhere?)

orbeckst:
We used to be able to test for memleaks when we were using nose for testing but didn't manage to port it to pytest. I don't remember why. So we are not really testing for memleaks – this would be a useful thing to look at. 

I don't think that the ANCHOR_UNIVERSES were related to #297, though, that was more about weakrefs inside Universe. The ANCHOR_UNIVERSES came from PR #321. 

> I am not sure how it helps to prevent duplicated universes, could you elaborate a bit more?

You're right, it doesn't. I had thought that serializing a universe and unpacking it in the following way
```python
import pickle
import MDAnalysis as mda; from MDAnalysis.tests import datafiles as data
u = mda.Universe(data.PSF, data.DCD)
u_copy = pickle.loads(pickle.dumps(u))
```
would make `u` and `u_copy` magically the same
```python
u is u_copy    # nope
```
but that's `False`. 

When doing it with a group `g = u.atoms` then the new group `g_copy = loads(dumps(g))` correctly attaches to the old universe: `(u is g_copy.universe) is True` but that also works with your new approach.

orbeckst:
@yuxuanzhuang , with this PR, what does the following code do?
```python
import pickle
import MDAnalysis as mda; from MDAnalysis.tests import datafiles as data
u = mda.Universe(data.PSF, data.DCD)
g = u.atoms
h = u.atoms[:10]
g_copy = pickle.loads(pickle.dumps(g))
h_copy = pickle.loads(pickle.dumps(h))

print("g_copy.universe:", u is g_copy.universe)
print("h_copy:", u is h_copy.universe)
```
I.e., do groups attach to their own universe if it is already present?

I vaguely remember that you had already checked that previously but just wanted to confirm.

orbeckst:
Assuming all looking good then try removing any remnants of ANCHOR_UNIVERSE.

yuxuanzhuang:
No it is not after ditching anchoring.
```python
import pickle
import MDAnalysis as mda; from MDAnalysis.tests import datafiles as data
u = mda.Universe(data.PSF, data.DCD)
g = u.atoms
h = u.atoms[:10]
g_copy = pickle.loads(pickle.dumps(g))
h_copy = pickle.loads(pickle.dumps(h))

print("g_copy.universe is u:", u is g_copy.universe)
print("h_copy.universe is u:", u is h_copy.universe)
print("g_copy.universe is h_copy.universe", g_copy.universe == h_copy.universe)

g_copy.universe is u: False
h_copy.universe is u: False
g_copy.universe is h_copy.universe False

g_copy, h_copy, u_copy = pickle.loads(pickle.dumps([g, h, u]))

print("g_copy.universe is u_copy:", u_copy is g_copy.universe)
print("h_copy.universe is u_copy:", u_copy is h_copy.universe)
print("g_copy.universe is h_copy.universe", g_copy.universe == h_copy.universe)niverse)
print("h_copy:", u is h_copy.universe)

g_copy.universe is u_copy: True
h_copy.universe is u_copy: True
g_copy.universe is h_copy.universe True
```
They have to be pickled together. Is there any use cases for the former condition?

yuxuanzhuang:
No performance difference is observed when using pmda:
```python
u = mda.Universe('serialize_mda/benchmarking/YiiP_system.pdb','serialize_mda/benchmarking/YiiP_system_90ns_center.xtc')
ref = mda.Universe('serialize_mda/benchmarking/YiiP_system.pdb')
rmsd_ana = pmda.rms.RMSD(u.atoms,ref.atoms)
rmsd_ana.run(n_jobs=12, n_blocks=12)
print(rmsd_ana.timing._total)
```
Before: 26.77
Afrer: 26.20

orbeckst:
Thanks for the illuminating examples. 

I've been thinking more about the question *if unpickling an AtomGroup should attach to an existing Universe or create a new one*.
> Is there any use cases for the former condition?

I don't really envisage a use where I would keep serializing separate AtomGroups but users come up with all kinds of creative solutions so I can see people getting confused if they find that two AtomGroups are now diverging in their history (and don't update when iterating over the trajectory, for instance). 

On the other hand, I would say from a Python point of view, serialize-deserialize should behave like a deepcopy and with reattaching to existing universes we break this expectation. Furthermore, carrying global state around in ANCHOR_UNIVERSES is awkward.

With all of this I am now _strongly in favor of removing all ANCHOR_UNIVERSES stuff_ and be very explicit about what happens when we serialize:
- You always get an independent Universe.
- If you have multiple related objects then you _must_ serialize them in one go (e.g. as a tuple). 

If the docs don't say this already explicitly then that should be added.

@MDAnalysis/coredevs any thoughts?

PR #2894, opened 2020-08-10
Make H5MD serialize with pickle

### LINKED ISSUE #2890: H5MD fails to serialize with pickle
## Expected behavior ##
All readers can be pickled and pass the tests for MultiframeReaderTest

## Actual behavior ##

`TestH5MDReader.test_pickle_reader` fails, see https://travis-ci.com/github/MDAnalysis/mdanalysis/jobs/369892996

(Note that this test failure occured becaused the H5MD PR #2787 was merged before the serialization PR and the latter did not have the H5MD PR #2723  merged into it.) 

Run TestH5MDReader.test_pickle_reader

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) current develop at  563995c 
- Which version of Python (`python -V`)? any
- Which operating system? all on CI

edisj:
Fixes #2890 

 - Added `H5PYPicklable` class to `H5MDReader`

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/coordinates/H5MD.py
+++ b/package/MDAnalysis/coordinates/H5MD.py
@@ -803,18 +803,13 @@ class H5PYPicklable(h5py.File):
     """
     def __getstate__(self):
         try:
-            driver = self._driver
-            comm = self._comm
+            driver = self.driver
         except AttributeError:
             driver = None
-            comm = None
 
-        return {
-                    'name': self.filename,
-                    'mode': self.mode,
-                    'driver': driver,
-                    'comm': comm
-                   }
+        return {'name': self.filename,
+                'mode': self.mode,
+                'driver': driver}
 
     def __setstate__(self, state):
         if state['driver'] == 'mpio':
--- a/package/MDAnalysis/coordinates/H5MD.py
+++ b/package/MDAnalysis/coordinates/H5MD.py
@@ -859,7 +859,7 @@ def comm_pickle(obj):
     if obj == MPI.COMM_WORLD:
         return comm_unpickle, ('COMM_WORLD',)
     if isinstance(obj, MPI.Comm):
-        return comm_unpickle, ('COMM_WORLD')
+        return comm_unpickle, ('COMM_WORLD',)
     raise TypeError("cannot pickle object")
 
 
--- a/package/MDAnalysis/coordinates/H5MD.py
+++ b/package/MDAnalysis/coordinates/H5MD.py
@@ -819,22 +819,6 @@ class H5PYPicklable(h5py.File):
     .. versionadded:: 2.0.0
     """
 
-    if HAS_MPI:
-        def comm_unpickle(self, name):
-            return getattr(MPI, name)
-
-        def comm_pickle(self, obj):
-            if obj == MPI.COMM_NULL:
-                return self.comm_unpickle('COMM_NULL',)
-            if obj == MPI.COMM_SELF:
-                return self.comm_unpickle('COMM_SELF',)
-            if obj == MPI.COMM_WORLD:
-                return self.comm_unpickle('COMM_WORLD',)
-            raise TypeError("cannot pickle MPI.Comm object")
-
-        copyreg.pickle(MPI.Intracomm, comm_pickle, comm_unpickle)
-        copyreg.pickle(MPI.Comm, comm_pickle, comm_unpickle)
-
     def __getstate__(self):
         try:
             driver = self.driver
--- a/package/MDAnalysis/coordinates/H5MD.py
+++ b/package/MDAnalysis/coordinates/H5MD.py
@@ -832,6 +832,8 @@ class H5PYPicklable(h5py.File):
                 return self.comm_unpickle('COMM_WORLD',)
             raise TypeError("cannot pickle MPI.Comm object")
 
+        copyreg.pickle(MPI.Intracomm, comm_pickle, comm_unpickle)
+        copyreg.pickle(MPI.Comm, comm_pickle, comm_unpickle)
 
     def __getstate__(self):
         try:
@@ -846,9 +848,6 @@ class H5PYPicklable(h5py.File):
                     comm = self.id.get_access_plist().get_fapl_mpio()[0]
                 except AttributeError:
                     comm = None
-
-                copyreg.pickle(MPI.Intracomm, self.comm_pickle, self.comm_unpickle)
-                copyreg.pickle(MPI.Comm, self.comm_pickle, self.comm_unpickle)
 
             return {'name': self.filename,
                     'mode': self.mode,
--- a/package/MDAnalysis/coordinates/H5MD.py
+++ b/package/MDAnalysis/coordinates/H5MD.py
@@ -806,7 +806,8 @@ class H5PYPicklable(h5py.File):
             driver = self._driver
             comm = self._comm
         except AttributeError:
-            comm and not (driver == 'mpio')
+            driver = None
+            comm = None
 
         return {
                     'name': self.filename,
--- a/package/MDAnalysis/coordinates/H5MD.py
+++ b/package/MDAnalysis/coordinates/H5MD.py
@@ -805,7 +805,11 @@ class H5PYPicklable(h5py.File):
 
     def __getstate__(self):
         driver = self.driver
-        if driver == 'mpio':
+        # Current issues: Need a way to retrieve MPI communicator object
+        # from self and pickle MPI.Comm object. Parallel driver is excluded
+        # from test because h5py call for an MPI configuration when driver is
+        # 'mpio', so this will need to be patched in the test function.
+        if driver == 'mpio':  # pragma: no cover
             raise TypeError("Parallel pickling of `h5py.File` with"
                             " 'mpio' driver is currently not supported.")
 
--- a/package/MDAnalysis/coordinates/H5MD.py
+++ b/package/MDAnalysis/coordinates/H5MD.py
@@ -186,7 +186,6 @@ Classes
 """
 
 import numpy as np
-import copyreg
 import MDAnalysis as mda
 from . import base, core
 from ..exceptions import NoDataError
@@ -205,22 +204,6 @@ except ImportError:
 
 else:
     HAS_H5PY = True
-
-try:
-    from mpi4py import MPI
-except ImportError:
-    HAS_MPI = False
-
-    # Allow building documentation even if mpi4py is not installed
-    import types
-
-    class MockMPI:
-        pass
-    mpi4py = types.ModuleType("mpi4py")
-    MPI = MockMPI
-
-else:
-    HAS_MPI = True
 
 
 class Timestep(base.Timestep):
@@ -780,9 +763,8 @@ class H5PYPicklable(h5py.File):
 
     When the file is pickled, filename, mode, driver, and comm of
     :class:`h5py.File` in the file are saved. On unpickling, the file
-    is opened by filename, mode, driver, and comm (driver and comm are
-    optional arguments). This means that for a successful unpickle, the
-    original file still has to be accessible with its filename.
+    is opened by filename, mode, driver. This means that for a successful
+    unpickle, the original file still has to be accessible with its filename.
 
     Parameters
     ----------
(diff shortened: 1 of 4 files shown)


edisj:
Hi @orbeckst, hopefully this fixes the pickling issues. I followed the code @yuxuanzhuang gave in #2890 which does allow the tests to pass. I modified it a bit to work with `comm` arguments as well and added documentation. 

It seems to work with some testing:
image
as in the `h5py.File` object seems to have all the data in it that it should, but I'm not entirely confident with how pickling works. Please leave any comments/criticisms for changes

yuxuanzhuang:
The Travis error can be solved by adding `mock`, like in `chemfiles`:
```python
try:
    import h5py
except ImportError:
    HAS_H5PY = False

    # Allow building documentation even if h5py is not installed
    import imp

    class MockH5pyFile:
        pass
    h5py = imp.new_module("h5py")
    h5py.File = MockH5pyFile

else:
    HAS_H5PY = True
```

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -727,3 +728,93 @@ def has_forces(self):
...
+
+
+    .. versionadded:: 2.0.0
+    """
+    def __getstate__(self):
+        try:
+            driver = self._driver
+            comm = self._comm
+        except AttributeError:  # is this error necessary?
+            driver = None
```
yuxuanzhuang, line 796: Will that be a problem if `self._driver` exists but `self._comm` not?

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -727,3 +728,93 @@ def has_forces(self):
...
+    def __getstate__(self):
+        state = self.__dict__.copy()
+        del state['_particle_group']
+        return state
+
+    def __setstate__(self, state):
+        self.__dict__ = state
+        self._particle_group = self._file['particles'][
+                                list(self._file['particles'])[0]]
+
```
yuxuanzhuang, line 741:
```suggestion
        self[self.ts.frame]

```
This is not needed in this specific case, but just to fit to be broader API.

orbeckst:
@yuxuanzhuang thank you for reviewing.

@edisj please ping me when all tests are passing and when you have addressed all of @yuxuanzhuang 's comments. I can then do a final review + approval.

package/MDAnalysis/coordinates/H5MD.py
edisj: You're right I think it would have been a problem. `self._driver` should have been `self.driver` as well, since it's pulling the driver from the file object, right? The pickled file wasn't returning the right driver with `self._driver`, but does with `self.driver`

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -727,3 +740,89 @@ def has_forces(self):
...
+    .. versionadded:: 2.0.0
+    """
+    def __getstate__(self):
+        try:
+            driver = self.driver
+        except AttributeError:
+            driver = None
+
+        return {'name': self.filename,
```
orbeckst, line 810: I am confused: If the `comm` is not included in `__getstate__` then how can it be used in `__setstate__`?

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -727,3 +740,89 @@ def has_forces(self):
...
+                'mode': self.mode,
+                'driver': driver}
+
+    def __setstate__(self, state):
+        if state['driver'] == 'mpio':
+            self.__init__(name=state['name'],
+                          mode=state['mode'],
+                          driver=state['driver'],
+                          comm=state['comm'])
```
orbeckst, line 819: Where will `state['comm']` come from?

edisj: I thought it was taking the `state` dict as argument from https://github.com/MDAnalysis/mdanalysis/blob/c7087b8cb1473b98ca943057d1697b1a3f91a46d/package/MDAnalysis/coordinates/H5MD.py#L744-L747

package/MDAnalysis/coordinates/H5MD.py
orbeckst: I see that you had it included in b29e675765c71ee0e015b3690e6b38a902a86188 but then removed again.

Did you test that a `MPI.COMM_WORLD` could be pickled on its own, i.e.,
```python
from mpi4py import MPI
import pickle

comm = MPI.COMM_WORLD
comm_p = pickle.loads(pickle.dumps(comm))

assert comm == comm_p
```
or something similar?

package/MDAnalysis/coordinates/H5MD.py
orbeckst: Maybe I don't understand how pickling works but I assumed that the goal is to make each object picklable by itself, i.e., ,`H5PYPicklableFile` **by itself** can be pickled. That's the first thing to check (and add a test for!!). Once this works, `H5MDReader` can just treat it as one item in its own class dict, as it would a numpy array or a basic Python object. It would be bad program design if `H5PYPicklableFile` needed the help of `H5MDReader` because then you could never use it independently from `H5MDReader`.

edisj: I see, that makes sense, thanks for the clarification. I'm doing some testing with `comm` on the parallel build I have on spudda

package/MDAnalysis/coordinates/H5MD.py
edisj: That does not seem to work
image

yuxuanzhuang: I don’t have a working mpi4py in my desktop. But it seems that it cannot be pickled natively.
This might help meanwhile https://bitbucket.org/mpi4py/mpi4py/issues/104/pickling-of-mpi-comm

orbeckst:
@edisj please see #2897 and switch imp to using importlib.

edisj:
I'm making a lot of commits at the moment because I'm pulling them to my repository on workstation for testing with parallel h5py

orbeckst:
Don't worry, we can squash it all when merging.

edisj:
@orbeckst The latest commit successfully pickles `h5py.File` objects with `driver='mpio` and `comm=MPI.COMM_WORLD`, although I'm still unsure if the correct communicator is being unpickled. I started a conversation in the h5py group and will continue to poke around online in the meantime.

image

yuxuanzhuang:
I am trying to build a parallel h5py system. You could add a few tests (specifically for H5PYPicklable) inside ``testsuite/MDAnalysisTests/utils/test_pickleio.py`` to increase codecov. (add skipif marks for them)

yuxuanzhuang:
I am having trouble to install all that stuffs to make h5py work. But I assume if it is hard to find the "right" `comm`, you might  overwrite the `__init__` function to save the input `comm` with something like
```python
def __init__(self, ...comm):
   self._comm = comm
   super().__init__(...)
```

orbeckst:
The test failures are pretty annoying for all other PRs. Can we make H5PYPicklable work for _serial_ right now? It would be ok to raise a `TypeError` (or whatever else is normally raised) if a communicator is present. Add a test that checks that `comm != None` raises the error.

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -727,3 +757,112 @@ def has_forces(self):
...
+    :class:`MDAnalysis.lib.picklable_file_io.BZ2Picklable`
+
+
+    .. versionadded:: 2.0.0
+    """
+
+    def __getstate__(self):
+        driver = self.driver
+        if driver == 'mpio':
+            comm = self.id.get_access_plist().get_fapl_mpio()[0]
```
orbeckst, line 825: If `MPI.COMM_WORLD` cannot be natively pickled then I don't think that this will work.

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -727,3 +757,112 @@ def has_forces(self):
...
+    :class:`MDAnalysis.lib.picklable_file_io.GzipPicklable`
+    :class:`MDAnalysis.lib.picklable_file_io.BZ2Picklable`
+
+
+    .. versionadded:: 2.0.0
+    """
+
+    def __getstate__(self):
+        driver = self.driver
+        if driver == 'mpio':
```
orbeckst, line 824: In order to move the PR along, make `__getstate__` fail (with `TypeError`) when you have driver mpio and a comm object.

Add a test that specifically checks for the exception being raised. Then open a new issue to make pickling of parallel h5py files work.

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -727,3 +757,112 @@ def has_forces(self):
...
+    if obj == MPI.COMM_WORLD:
+        return comm_unpickle, ('COMM_WORLD',)
+    if isinstance(obj, MPI.Comm):
+        return comm_unpickle, ('COMM_WORLD',)
+    raise TypeError("cannot pickle object")
+
+
+if HAS_MPI:
+    copyreg.pickle(MPI.Intracomm, comm_pickle, comm_unpickle)
+    copyreg.pickle(MPI.Comm, comm_pickle, comm_unpickle)
```
orbeckst, line 868: Remove for the serial-pickle PR

package/MDAnalysis/coordinates/H5MD.py
edisj: The test fails with `ModuleNotFoundError` because h5py tries to import mpi4py when `driver='mpio'`. Is there any way around this?

orbeckst: Which test are we talking about?

edisj: something like this:
```
@pytest.mark.skipif(not HAS_H5PY, reason="h5py not installed")
def test_H5MD_parallel_pickle():
    from unittest.mock import MagicMock
    MPI_COMM_WORLD = MagicMock()
    h5md_io = H5PYPicklable(H5MD_xvf, 'r',
                            driver="mpio",
                            comm=MPI_COMM_WORLD)
    h5md_io_pickled = pickle.loads(pickle.dumps(h5md_io))
```

orbeckst:
See https://travis-ci.com/github/MDAnalysis/mdanalysis/jobs/370755668#L2255-L2257
```
    def test_read_charges(self, top):
>       assert_almost_equal(top.charges.values, 0.0)
E       NameError: name 'assert_almost_equal' is not defined
testsuite/MDAnalysisTests/topology/test_hoomdxml.py:82: NameError
```

package/MDAnalysis/coordinates/H5MD.py
orbeckst: Try mocking the mpi4py module, too.  The idea is to provide a thing (the mock) that just pretends to be there but doesn't really do anything.

orbeckst: @richardjgowers you have been dubbed in other PR "The King of Mocks" – any suggestions?

yuxuanzhuang: I assume it cannot be fixed as it's checked inside h5py code. (https://github.com/h5py/h5py/blob/8cc83410995996330c85867d0e56cbe922b709f8/h5py/_hl/files.py#L43)

Maybe just excluding these lines from coverage for now?

orbeckst: There might be a way to trick `h5py` into believing that it was built with mpi support (one could patch the call to `h5.get_config()` https://github.com/h5py/h5py/blob/8cc83410995996330c85867d0e56cbe922b709f8/h5py/_hl/files.py#L24 or just the `mpi` variable) but that's not worth it for right now so I concur with @yuxuanzhuang to just skip the test. Add a comment summarizing the current issues and move forward.

edisj: @orbeckst  just pushed commit with the lines excluded from coverage

orbeckst:
@yuxuanzhuang could you please review this PR within the next day? Sorry for asking for a quick turnaround but @edisj 's project is coming to an end and we would like to have it at a good stopping point, just in case he has other more urgent things to do. 

You should be able to leave a approve/request changes review.  I will wait with merging until I hear from you.

orbeckst:
Thank you @edisj and @yuxuanzhuang !

PR #2900, opened 2020-08-11
Wrap RDKit drawing code for AtomGroups

cbouy:
Part of #2468
Depends on #2775

 - In notebooks, by importing `from MDAnalysis.visualization.RDKit import RDKitDrawer`, the default representation of the AtomGroup is changed for small molecules: it displays an image of the AtomGroup using RDKit's drawing code (and the default representation for large AtomGroups)
image
- The default drawer displays PNG images for atomgroups below 200 atoms, removes non-polar hydrogens, kekulizes and flattens the structure, although all of these are configurable
image
- You can reuse the code that is used to generate the PNG/SVG by calling `drawer.atomgroup_to_image(ag, ...)`. There is also some code to automatically display or save gifs from an atomgroup:
image
- And all of this can be cancelled at any point
image

PR Checklist
------------
 - [ ] Tests?
 - [x] Docs?
 - [ ] CHANGELOG updated?

Tagging @MDAnalysis/coredevs 

I'm not sure if I've put the code in the right place but "visualization" made sense to me


cbouy:
I added a metaclass and a base class for "formatters" (i.e. classes that can modify the representation of other objects in interactive shells/notebooks), in case we want to add another visualization package like nglview to display atomgroups directly

package/MDAnalysis/visualization/RDKit.py
```diff
@@ -0,0 +1,212 @@
...
+    format = "RDKIT"
+
+    def __init__(self, size=(450, 250), max_atoms=200, removeHs=True,
+                 kekulize=True, drawOptions=None, useSVG=False):
+        """Set default parameters for the drawer
+
+        Parameters
+        ----------
+        size : tuple
+            default size for images
```
tylerjereddy, line 66: units are pixels or?

package/MDAnalysis/visualization/base.py
```diff
@@ -0,0 +1,97 @@
...
+    .. versionadded:: 2.0.0
+    """
+
+    def add_repr(self, obj, mime, func):
+        """Add a new formatter to an object
+
+        Parameters
+        ----------
+        obj : object
+            An MDAnalysis object
```
tylerjereddy, line 64: That's a really broad input type! There's no restriction here that it should be related to i.e., `Universe` or `atomgroup` as far as visualization-related things goes?

tylerjereddy (commented):
Not sure about testing images themselves. I don't think we do much/any of that for i.e., the `streamlines` modules at the moment. I think `matplotlib` does have a testing harness/exapmles for that kind of testing though.

IAlibay:
I'm moving this to the 2.1.0 milestone

ldomic:
Maybe this helps with the SVG testing? https://github.com/ldomic/mdanalysis/blob/2af2d99447cc06d96fd3695b400b9f3bcbb0f4d5/testsuite/MDAnalysisTests/visualization/test_lintools.py

package/MDAnalysis/visualization/RDKit.py
```diff
@@ -0,0 +1,212 @@
...
+    format = "RDKIT"
+
+    def __init__(self, size=(450, 250), max_atoms=200, removeHs=True,
+                 kekulize=True, drawOptions=None, useSVG=False):
+        """Set default parameters for the drawer
+
+        Parameters
+        ----------
+        size : tuple
+            default width and height of images, in pixels
```
ldomic, line 66: Would it be possible to set up your own width and height for an image? If so, then `default` is not necessary - the fact that it is set up as an attribute of the function suggests the default :)

package/MDAnalysis/visualization/RDKit.py
```diff
@@ -0,0 +1,212 @@
...
+
+class RDKitDrawer(FormatterBase):
+    """Wrapper class to RDKit's drawing code
+
+    .. versionadded:: 2.0.0
+    """
+    format = "RDKIT"
+
+    def __init__(self, size=(450, 250), max_atoms=200, removeHs=True,
+                 kekulize=True, drawOptions=None, useSVG=False):
```
ldomic, line 60: I am of course biased, but if SVG were to be `True` by default (or even as the only option), it would become much easier to test and then we could leave it to the users to convert the image in the format of their choosing - PNG, JPEG etc - or if something like this is needed, to have a separate function that converts SVG to PNG or JPEG - it seems like there are some libraries that could do this easily. That way the testing burden is reduced, as you know that `RDKitDrawer` produces expected content and only the conversion remains untested,

package/MDAnalysis/visualization/RDKit.py
```diff
@@ -0,0 +1,212 @@
...
+        """Set default parameters for the drawer
+
+        Parameters
+        ----------
+        size : tuple
+            default width and height of images, in pixels
+        max_atoms : int
+            AtomGroups with more atoms that this limit won't be displayed as
```
ldomic, line 68: More of a question - I reckon this would be enough to show a regular lipid molecule without the hydrogens? With increased interest in lipids in MD, it could be useful to consider their drawings as well.

package/MDAnalysis/visualization/RDKit.py
```diff
@@ -0,0 +1,212 @@
...
+        """Prepare the AtomGroup and resulting mol for the drawing code
+
+        Parameters
+        ----------
+        ag : MDAnalysis.core.groups.AtomGroup
+            The AtomGroup to prepare
+        keep_3D : bool
+            Keep or remove 3D coordinates to generate the image
+        """
+        mol = ag.convert_to("RDKIT")
```
ldomic, line 116: One of the tricks I found while trying to make this happen a few years ago was that RDKit drew the molecules best when they were loaded as SMILES - I suppose they might already be, but if not it is worth a shot!

package/MDAnalysis/visualization/RDKit.py
```diff
@@ -0,0 +1,212 @@
...
+            img, *imgs = [Image.open(BytesIO(png)) for png in pngs]
+            # write to file, or display if output is None
+            buffer = partial(open, output, "wb") if output else BytesIO
+            with buffer() as fp:
+                img.save(fp=fp, format='GIF', append_images=imgs,
+                         save_all=True, duration=frame_duration, loop=0)
+                if isinstance(fp, BytesIO):
+                    b64 = b64encode(fp.getvalue()).decode("ascii")
+                    display(HTML(f"<img src='data:image/gif;base64,{b64}' />"))
+
```
ldomic, line 209:
```suggestion
else:
     raise ValueError("Creation of GIF requires a trajectory as part of the Universe")
```
Or something along those lines

package/MDAnalysis/visualization/base.py
```diff
@@ -0,0 +1,99 @@
...
+    warnings.warn("You should be in an interactive python shell (IPython or "
+                  "Notebook) to use this properly")
+
+from .. import _FORMATTERS
+
+
+class _Formattermeta(type):
+    """Automatic Formatter registration metaclass
+
+    .. versionadded:: 2.0.0
```
ldomic, line 39: I guess these need to move up to `2.1.0` (provisionally?)

package/MDAnalysis/visualization/base.py
```diff
@@ -0,0 +1,99 @@
...
+# J. Comput. Chem. 32 (2011), 2319--2327, doi:10.1002/jcc.21787
+#
+
+import warnings
+
+try:
+    shell = get_ipython()
+except NameError:
+    shell = None
+    warnings.warn("You should be in an interactive python shell (IPython or "
```
ldomic, line 30: Could I not create visualizations outside of IPython ecosystem? If e.g. I would like to create a pipeline such as LINTools used to be, I might struggle with this requirement, as it requires chaining functions in this PR and new ones. I do wonder if this requirement could make the visualization packages harder to use..

testsuite/MDAnalysisTests/visualization/test_rdkit.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def test_ag_to_img_png(self, u, drawer):
+        png = drawer.atomgroup_to_image(u.atoms)
+        img = Image.open(BytesIO(png))
+        assert img.size == drawer.size
+        assert len(png) == 6349
+
+    def test_ag_to_img_svg(self, u, drawer):
+        svg = drawer.atomgroup_to_image(u.atoms, useSVG=True)
+        assert len(svg) == 1275
```
ldomic, line 83: I think this could be extended to check the contents of the file as well as the length :)

ldomic (commented):
Lots of little nitpicky comments and questions about how this works! Appreciate that I am coming late to the party and would like to say that this is looking really good @cbouy

PR #2908, opened 2020-08-13
backport fix for #2878 (pickle DCD and XDR readers at correct frame)

orbeckst:
Fixes #2878 (backport)

Changes made in this Pull Request:
 - change `__setstate__` for DCDReader and XTC/TRRReader according to https://github.com/MDAnalysis/mdanalysis/issues/2878#issuecomment-665973591
 - add tests (which xfail for any readers that cannot be pickled yet)

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?
 - [x] Issue raised/referenced?

PR #2911, opened 2020-08-15
DCD XDR pickle tests

### LINKED ISSUE #2878: Fail to pickle last frame of DCD, XDR files
## Expected behavior ##

The trajectory can be pickled no matter which frame it is current in.

## Actual behavior ##

XDR and DCD readers cannot pickle the last frame.
Raising the EOFError

```python
u = mda.Universe(GRO_MEMPROT, XTC_MEMPROT)
u.trajectory[4] #  last frame
pickle.loads(pickle.dumps(u.trajectory))
EOFError: Trying to seek over max number of frames
```

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) 2.0.0 dev
- Which operating system? Ubuntu 20

yuxuanzhuang:
Fixes #2878 

 - add inclusive tests for pickle low-level dcd and xdr files
 - fix the bug that was introduced in #2723  
 - may further backport in #2908 
PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/lib/formats/libmdaxdr.pyx
+++ b/package/MDAnalysis/lib/formats/libmdaxdr.pyx
@@ -306,8 +306,7 @@ cdef class _XDRFile:
 
         # where was I
         current_frame = state[1]
-        self.seek(current_frame - 1)
-        self.current_frame = current_frame
+        self.seek(current_frame)
 
     def seek(self, frame):
         """Seek to Frame.
--- a/package/MDAnalysis/lib/formats/libmdaxdr.pyx
+++ b/package/MDAnalysis/lib/formats/libmdaxdr.pyx
@@ -306,14 +306,21 @@ cdef class _XDRFile:
 
         # where was I
         current_frame = state[1]
-        if current_frame == self.offsets.size:
-            #  cannot seek to self.offsets.size (like in DCD file)
-            #  because here seeking depends on the offsets list size.
-            #  Instead, we seek to one frame ahead and read the next frame. 
+        if current_frame < self.offsets.size:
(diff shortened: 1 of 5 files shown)


package/MDAnalysis/lib/formats/libmdaxdr.pyx
```diff
@@ -306,8 +306,14 @@ cdef class _XDRFile:
 
         # where was I
         current_frame = state[1]
-        self.seek(current_frame - 1)
-        self.current_frame = current_frame
+        if current_frame == self.offsets.size:
```
orbeckst, line 309: Switch the decision in the `if` statement to make the more common situation come first
```python
if current_frame < self.offsets.size:
     ...
elif current_frame == self.offsets.size:
    ...
else:
    #pragma: no cov
    raise RuntimeError("Invalid frame number {} > {} -- this should not happen.".format(current_frame, self.offsets.size)
```
and add defensive code.

orbeckst: Add comment under which circumstances we have to do funny things – namely when we need to serialize at the end of the trajectory.

yuxuanzhuang:
I makes `DCDFile` to behave the same as `XDRFile`, i.e. not messing up with `seek`, but checking during `__setstate__`

yuxuanzhuang:
The `pragma: no cover` is not honered in cython files (https://github.com/cython/cython/issues/3680)
Any reason we have to check for that here (aside from for the completeness), since it will be eventually checked in `seek`.

Issue #2768, opened 2020-06-17
Release 1.0.1


orbeckst:
A few things are accumulating for a patch-level release for 1.0, related to

- installation
- performance regressions

The point is to make 1.0.x a stable platform for anyone who needs legacy MDAnalysis for the foreseeable feature. 1.0.x should not be worse than 0.20.1.

orbeckst:
One potentially controversial statement in _Challenge to scientists: does your ten-year-old code still run?_ Nature 2020 doi:[10.1038/d41586-020-02462-7](https://doi.org/10.1038/d41586-020-02462-7) 
> Newer languages’ rapidly evolving APIs and reliance on third-party libraries make them vulnerable to breaking. In that sense, the sunsetting of Python 2.7 at the start of this year represents an opportunity for scientists, Rougier and Hinsen note. Python 2.7 puts “at our disposal an advanced programming language that is guaranteed not to evolve anymore”, Rougier writes (Rougier, N. P. ReScience C https://doi.org/10.5281/zenodo.3886628).

provides some additional motivation for making sure that we have a stable Python 2.7-capable release.

orbeckst:
The backports from PR #2798 are now included.

The remaining major issues is to either disable capped-distance with nsgrid #2930 or fix it.

jbarnoud:
The fix for #2984 could be backported.

IAlibay:
So after some off-github discussions with @lilyminium and @richardjgowers, we were wondering how everyone would feel about pushing for a rapid 1.0.1 release where we disable nsgrid completely and then have aim for a 1.0.2 release with @richardjgowers' fix?

If the mailing list is any indication, quite a lot of people are facing issues with things that are now fixed in `master`, on top of that we've got this nsgrid bug that users probably aren't aware of. Personally I think it would be better for everyone to have some type of patch fix out there earlier than later?

pinging @MDAnalysis/coredevs here.

lilyminium:
+1 for releasing an interim 1.0.1 as strictly better than keeping 1.0.0 until the fix is in (e.g. all the pytest failures)

orbeckst:
Yes, I agree – disable nsgrid #2930 for 1.0.1.

I'd also include a backport fix for #2984 and then do a 1.0.1. We can really do a patch release in rapid succession. Doubt that anyone minds getting code fixed.

lilyminium:
And #2427!

orbeckst:
I'll backport PR #2427.

[...]

PR #2913, opened 2020-08-18
deprecate hydrogen bond analysis in water bridge analysis

xiki-tempula:
Fixes #

deprecate hydrogen bond analysis in water bridge analysis.
The hydrogen bond analysis is not used in the original water bridge analysis and I have cleared some reference to the hydrogen bond analysis in the doc sections.
Move water bridge analysis from hbonds to hydorgenbonds

Related to #2739 and  #2746

 - 

PR Checklist
------------
 - [ ] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


IAlibay:
Going to ping @RMeli here, I believe this move was started in PR #2746, but I don't know why it stalled.
(also @xiki-tempula can you add a reference to #2739 in the PR text + eventually changelog? that way we can remember to update the progress on it).

xiki-tempula:
@IAlibay I have chatted with @RMeli before initiating this PR. This PR is still WIP and I will do this after I solved all the PEP8 issue.

RMeli:
@IAlibay, PR #2746 stalled because of https://github.com/MDAnalysis/mdanalysis/issues/2745#issuecomment-644425372 and for lack of time. @xiki-tempula did ask me if he could supersede my PR as part of his current work and I have no problems with that.

package/MDAnalysis/analysis/hbonds/wbridge_analysis.py
```diff
@@ -33,1527 +33,17 @@
...
-                    tmp_acceptors.append(acceptor[pairs[idx,1]])
-
-        angles = np.rad2deg(
-            calc_angles(
-                self.u.atoms[tmp_donors].positions,
-                self.u.atoms[tmp_hydrogens].positions,
-                self.u.atoms[tmp_acceptors].positions,
-                box=self.box
-            )
+warnings.warn(
```
IAlibay, line 44: So in the past we've been a little bit more cautious with stubs, specifically setting simplewarning through something like:

```python
with warnings.catch_warnings():
    warnings.simplefilter('always', DeprecationWarning)
    warnings.warn(("This module has been moved to "
                   "MDAnalysis.analysis.hydrogenbonds.wbridge_analysis "
                   "It will be removed in release 3.0"),
                  category=DeprecationWarning)

from ..hydrogenbonds.wbridge_analysis import WaterBridgeAnalysis
```

Whilst I'm not sure the order of the import matters so much, the use of filtering is probably useful in cases where people tend to just blanket remove all warnings. Pinging @orbeckst here who probably has more context on why this was the decision taken for stubs circa ~ v0.21.

package/MDAnalysis/analysis/hydrogenbonds/__init__.py
```diff
@@ -26,3 +26,4 @@
 ]
```
IAlibay, line 26: Also add to `__all__`?
```suggestion
    'WaterBridgeAnalysis'
]
```

testsuite/MDAnalysisTests/analysis/test_wbridge.py
```diff
@@ -7,7 +7,8 @@
 
 import MDAnalysis
 import MDAnalysis.analysis.hbonds
-from MDAnalysis.analysis.hbonds.wbridge_analysis import WaterBridgeAnalysis
+from MDAnalysis.analysis.hydrogenbonds.wbridge_analysis import (
+    WaterBridgeAnalysis, )
 
 def test_import_from_hbonds():
```
IAlibay, line 13: Also needs a test for the new warning please.

package/MDAnalysis/analysis/hydrogenbonds/wbridge_analysis.py
```diff
@@ -0,0 +1,1799 @@
+# -*- Mode: python; tab-width: 4; indent-tabs-mode:nil; coding:utf-8 -*-
```
IAlibay, line 1: From a cursory look, it seems fine as just the moved file + PEP8 changes. Unfortunately the relatively unhelpful git diff make this hard to check thoroughly (PEP8 probably could have just been dealt with in a separate PR but too late now).

Is there any particular changes here that you'd like us to look at?

package/MDAnalysis/analysis/hydrogenbonds/wbridge_analysis.py
```diff
@@ -0,0 +1,1799 @@
...
+    """Perform a water bridge analysis
+
+    The analysis of the trajectory is performed with the
+    :meth:`WaterBridgeAnalysis.run` method. The result is stored in
+    :attr:`WaterBridgeAnalysis.timeseries`. See
+    :meth:`~WaterBridgeAnalysis.run` for the format.
+
+
+    .. versionadded:: 0.17.0
+
```
IAlibay, line 698: Unnecessary white space?

package/MDAnalysis/analysis/hbonds/wbridge_analysis.py
```diff
@@ -33,1527 +33,17 @@
...
+            "This module has been moved to MDAnalysis.analysis.hydrogenbonds"
+            "It will be removed in MDAnalysis version 2.0. Please use "
```
IAlibay, line 46: Next target version is 2.0, so removals have to be done afterwards unless we want to make this a v1.1 target instead. Thoughts @orbeckst ?

IAlibay (changes requested):
Couple of comments + need to work out what the target version will be for full removal.

package/MDAnalysis/analysis/hydrogenbonds/wbridge_analysis.py
xiki-tempula: As is described in the PR description. I have 

- Fixed the PEP8 issue
- Remove reference to the old HBA in the doc
- Added the code for deprecation related stuff.

None of the code inside the wba is modified.

IAlibay (changes requested):

testsuite/MDAnalysisTests/analysis/test_wbridge.py
```diff
@@ -17,6 +17,26 @@ def test_import_from_hbonds():
                              "importing WaterBridgeAnalysis from "
                              "MDAnalysis.analysis.hbonds failed.'")
 
+
+def test_import_warning():
+    with pytest.warns(DeprecationWarning) as record:
+        from MDAnalysis.analysis.hbonds.wbridge_analysis import (
+            WaterBridgeAnalysis, )
+    assert len(record) == 2
+    assert record[0].message.args[0] == 'This module is deprecated as of '\
```
IAlibay, line 24: This is going to go away next, so I probably wouldn't test for it.

testsuite/MDAnalysisTests/analysis/test_wbridge.py
```diff
@@ -17,6 +17,26 @@ def test_import_from_hbonds():
                              "importing WaterBridgeAnalysis from "
                              "MDAnalysis.analysis.hbonds failed.'")
 
+
+def test_import_warning():
+    with pytest.warns(DeprecationWarning) as record:
```
IAlibay, line 20: Given my other comment, rather than explicitly testing the record, it's probably easier/cleaner to just check using the `match` keyword here:
```suggestion
    wmsg = 'Please use MDAnalysis.analysis.hydrogenbonds.wbridge_analysis'

    with pytest.warns(DeprecationWarning, match=wmsg):
```

IAlibay (changes requested):

testsuite/MDAnalysisTests/analysis/test_wbridge.py
```diff
@@ -17,6 +15,29 @@ def test_import_from_hbonds():
...
+                                        'hbond_analysis instead.'
+    assert record[1].message.args[0] == 'This module has been moved to '\
+                                        'MDAnalysis.analysis.hydrogenbondsIt '\
+                                        'will be removed in MDAnalysis '\
+                                        'version 2.0. Please use MDAnalysis.'\
+                                        'analysis.hydrogenbonds.'\
+                                        'wbridge_analysis instead.'
+
+
+from MDAnalysis.analysis.hydrogenbonds.wbridge_analysis import (
```
IAlibay, line 38: Strange, does it not work if this import is at the top of the file?

xiki-tempula: @IAlibay  The warning is only issued once when it is first imported. Importing it again won't give the same warning.

IAlibay: I see, well as long as it works, it's a lot better than whatever I had to do for some of the other tests with import issues 😅

xiki-tempula: @IAlibay but it creates PEP8 problem as the module-level import is not at the top of the file.

IAlibay:
Thanks for all the work here @xiki-tempula,  just waiting on a secondary opinion on some of the comments I left above. Particularly, I'm not 100% sure if we should not be targeting 1.x instead of 2.0.

testsuite/MDAnalysisTests/analysis/test_wbridge.py
IAlibay: Unfortunately I can't pull your PR (since it's on your develop branch), so I can't tell if there is some weird way to go around this, but if this is the only way I think we can live with breaking PEP8 a little bit (it's more guidelines rather than strict rules to live by).

package/MDAnalysis/analysis/hbonds/wbridge_analysis.py
lilyminium: Are we doing more versions of 1? I would deprecate in 1.xx and remove in 2.0, as was done for #2791 and #2622.

orbeckst: If we have to do a 1.1 for whatever reasons then we keep things as unchanging as possible. But the goal was to keep 1.x boring and as similar to the old stuff as possible, so any removals target 2.0.

I _hope_ that we are only doing bug fixes for 1.x so that this stays 1.0.x

package/MDAnalysis/analysis/hbonds/wbridge_analysis.py
orbeckst: Sorry for the late (and unhelpful) reply: I don't remember and the more cautious approach by @IAlibay looks good to me.

Removal would be 2.0 (together with old h-bonds).

package/CHANGELOG
```diff
@@ -122,6 +122,8 @@ Changes
     `__call__` (Issue #2860, PR #2859)
   * deprecated ``analysis.helanal`` module has been removed in favour of
     ``analysis.helix_analysis`` (PR #2929)
+  * Move water bridge analysis from hbonds to hydrogenbonds and deprecate
+  hydrogen bond analysis in water bridge analysis (Issue #2739 PR #2913)
```
orbeckst, line 126: indentation
```suggestion
    hydrogen bond analysis in water bridge analysis (Issue #2739 PR #2913)
```

testsuite/MDAnalysisTests/analysis/test_wbridge.py
```diff
@@ -1,13 +1,15 @@
 from io import StringIO
 from collections import defaultdict
+from importlib import reload
```
orbeckst, line 3: Why do we need `reload`? This is very unusual and I'd not rather have it unless absolutely necessary. 

It is not explicitly used so should be removed.

orbeckst (changes requested):
This PR should really be against master to be part of release 1.0.1 #2768 . The deprecations should become part of 1.0.1 and the removal is scheduled for 2.0.

You then need a second PR against develop that does _not_ contain warnings etc and only contains the moved module.

I am not sure how you want to organize this in the best way – maybe make a second PR that targets master and then remove the warnings/deprecations from this one (and change the CHANGELOG accordingly).

testsuite/MDAnalysisTests/analysis/test_wbridge.py
xiki-tempula: The reload is for the checking of warnings.
The warning will only appear upon the import as I use reload to regenerate the warning so that it could be checked.

xiki-tempula:
@orbeckst Thank you for the advice. I have changed the changelog and remove the water_bridge from the hbond module. I will open another PR for master.

xiki-tempula:
@orbeckst Thank you for the advice. I was wondering if there is anything that I could do to get this PR merged. I'm working to port @p-j-smith 's MDAnalysis.analysis.hydrogenbonds.hbond_analysis.HydrogenBondAnalysis.lifetime() to water bridge analysis, so it would be nice if this PR could be merged.

orbeckst:
Sorry , I am snowed under with other things at the moment. Could you please ping me at the end of the week again? Sorry again for the long silence.

xiki-tempula:
@orbeckst Thanks for all the help. I wonder if you mind give a look, please? I will make another PR to the masters once this one is merged.

orbeckst:
Thanks for the reminder. I put it in my calendar for Wednesday — sorry, there’s no way I can get to it any earlier (and I had not enough spare time last week).

orbeckst (changes requested):
This PR moves wbridge_analysis to hydrogenbonds. Please open a PR against master that just includes the deprecations, stating that in 2.0, it will be in hydrogenbonds. When the deprecation PR is approved, I can also approve the move here.

orbeckst:
@xiki-tempula what is your plan for wbridge analysis? Do you want to rewrite it to base it on the new hydrogenbond analysis? It would be really good to have HydrogenBondAnalysis and WaterBridgeAnalysis behave in the same way as far as selections and performance go. 

If there's a way to avoid duplication of code then I would be very much in favor of, e.g., creating a base class for both HydrogenBondAnalysis and WaterBridgeAnalysis.

xiki-tempula:
@orbeckst The PR against the master is done in #3111.

The issue with HydrogenBondAnalysis is a quite tricky one. Since the WaterBridgeAnalysis is written based on the old HydrogenBondAnalysis, it has all the functionality of the old HydrogenBondAnalysis. 

Later on, @p-j-smith introduced capped-distance to HydrogenBondAnalysis, which brought a major speed improvement. So I have adopted the capped-distance to distance calculations. Currently, WaterBridgeAnalysis and HydrogenBondAnalysis offer the same performance in terms of computing hydrogen bonds.

The selection problem is a bit tricky. As is discussed in #2177, the WaterBridgeAnalysis adopted the same selection scheme as the old HydrogenBondAnalysis. I'm currently working on a RDKIT based selection scheme.

It is difficult to write WaterBridgeAnalysis based on HydrogenBondAnalysis as WaterBridgeAnalysis needs a more complex data structure but it is easy to write HydrogenBondAnalysis based on WaterBridgeAnalysis. The only functionality that is present in HydrogenBondAnalysis but not WaterBridgeAnalysis is the lifetime analysis, which I plan to add when this PR is merged.

p-j-smith:
Hi, I don't know the full extent of the functionality of the WaterBridgeAnalysis class, but I have used HydrogenBondAnalysis to find hydrogen bonds and then used NetworkX to look for water-bridging in the following way:

1. Find hydrogen bonds. Use an selection that include e.g. protein and water within 10 Angstrom of the protein, such as in the fourth example seen here: https://docs.mdanalysis.org/2.0.0-dev0/documentation_pages/analysis/hydrogenbonds.html#example-use-of-hydrogenbondanalysis Using the 10 Angstrom cutoff (as opposed to all water) makes both the hydrogen bond analysis and the path-finding faster.

2. Iterate over frames. At each frame construct an adjacency matrix, A, of size (num_donors, num_accepors), where Aij=1 if there is a hydrogen bond between donor_i and acceptor_j. Use NetworkX to construct a graph from this adjacency matrix. Iterate over donor-acceptor pairs of the protein only, and use the NetworkX `all_shorted_paths` method (https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.shortest_paths.generic.all_shortest_paths.html#networkx.algorithms.shortest_paths.generic.all_shortest_paths) to find the shortest water-mediated paths between each donor and acceptor, if one exists. If the path is greater than a cutoff, say 5 water molecules, don't count this as path. I'm not sure, but I think there is also an all-to-all method in NetworkX that could remove the need to iterate over donor-acceptor pairs.

Another PhD student in my group had the idea of using NetworkX for this. We've only used it to look at the statistics of the path-length distribution, but NetworkX also returns the list of atoms in the through-water path from donor-to acceptor. I'm not sure whether this is something you would be interested in, but offloading the path-finding to NetworkX might give a performance boost whilst also keeping the same selections etc. as HydrogenBondAnalysis. If this is something you're interested in @xiki-tempula I'd be happy to work with you on it.

xiki-tempula:
@p-j-smith Thanks for the insight. The WaterBridgeAnalysis basically does what you have said with various tweaks for speed improvement and robustness.

The first step is to find the hydrogen bonds within a defined space. Instead of `protein and water within 10 Angstrom of the protein`, I make three selections, one selection is the start selection (selection 1) another selection is the end selection (selection 2), while the third water selection is a convex which is both within a certain distance from selection 1 and selection 2. Since the hydrogen bond from selection 1 to selection 2 is implicitly detected in this stage, it is easy to go from WaterBridgeAnalysis to HydrogenBondAnalysis but not the other way around.

The second step is finding a path from selection 1 to selection 2. I have written an optimized path-finder for this type of job. Though I have not tested the NetworkX implementation, I think it would be interesting to see how would NetworkX perform.
I don't think NetworkX would be significantly slower than my implementation but most of the time is spent on distance search, so it is not likely to matter a lot.

I think one strength of the WaterBridgeAnalysis is the flexible analysis functionality. The `path-length distribution` could be easily computed by providing the custom analysis function to the count_by_type function.

IAlibay:
> The selection problem is a bit tricky. As is discussed in #2177, the WaterBridgeAnalysis adopted the same selection scheme as the old HydrogenBondAnalysis. I'm currently working on a RDKIT based selection scheme.

Just jumping in here and focusing very specifically on the last this specifically. I don't know what your plans are re: RDKIT-based selections, but if it's possible to make it optional that would be great. WaterBridgeAnalysis is a great core analysis feature of MDA, since we probably can't make RDKIT a core dependency, having the ability to use it without RDKIT is probably quite important.

p-j-smith:
> I make three selections, one selection is the start selection (selection 1) another selection is the end selection (selection 2), while the third water selection is a convex which is both within a certain distance from selection 1 and selection 2.

Sorry - it's been a while since I worked on this. You're completely right - I remember now that in post-processing the hydrogen bonds to find water-bridging I also had to use three selections - for the source, sink and path. So what I suggested would probably complicate rather than simplify things.

orbeckst (approved):

orbeckst:
With PR #3111 to be merged into 1.0.2-develop, this is good to go from my end.

package/CHANGELOG
```diff
@@ -156,6 +156,7 @@ Changes
     `__call__` (Issue #2860, PR #2859)
   * deprecated ``analysis.helanal`` module has been removed in favour of
     ``analysis.helix_analysis`` (PR #2929)
+  * Move water bridge analysis from hbonds to hydrogenbonds (Issue #2739 PR #2913)
```
IAlibay, line 159: Technically new changes go to the top of the list, but I'm willing to let that go.

IAlibay (approved):
Sorry, I've not been keeping up, from a cursory look it seems fine. I'll approve since @orbeckst did / so I'm not blocking merging :)

orbeckst:
I want to rebase all of this into a single commit (there's too much clutter in the history) but *Rebase and merge* will not work automatically.

@xiki-tempula please condense the history of this PR in not more than ~3 commits (or just a single one) and then force-push so that we have a clean history. I leave this to you because you know best what the end result ought to look like.

I will block until the history is cleaned up.

orbeckst:
Actually, forget my previous comment: *Squash and merge* works just fine. Will do this *now*. Sorry for the clutter!

package/CHANGELOG
orbeckst: I appreciate your attention to detail.

I merged the whole thing nevertheless, as this has been sitting for way too long.

Issue #2863, opened 2020-07-23
Interoperability with OpenMM


jbarnoud:
## Is your feature request related to a problem? ##
On the contrary to most MD engines, OpenMM gives access to most of its internals through a python API. In a typical OpenMM workflow, many operations are performed using that API including loading and manipulating topologies, and writing the trajectory to file. An interoperability layer between OpenMM and MDAnalysis would make it easier to use MDAnalysis to prepare OpenMM simulation, and benefit from the file formats supported by the library when saving a running OpenMM trajectory to disk.

### Topology parser for `simtk.openmm.app.Topology`

OpenMM stores topologies using the `simtk.openmm.app.Topology` object. That object provides name, element, and residue information for each atom, so as the topologically relevant connectivity between the atoms. The information stored in that object would allow building a `MDAnalysis.core.topology.Topology` object. This opens the door to creating a Universe.

### Universe from `simtk.openmm.app.PDBFile`, `simtk.openmm.app.PDBXFile`, and `simtk.openmm.app.Modeller` instances

These objects are the ones used by OpenMM to read and write PDB and mmCIF files, and to build or modify topologies (with operations such as adding water, or building a membrane). They contain both a topology and coordinates which is everything we need to create a Universe.

With the ability to create such Universes, MDAnalysis gains access to the file reading and topology modelling capabilities of OpenMM.

### Converters to `simtk.openmm.app.PDBFile`, `simtk.openmm.app.PDBXFile`, and `simtk.openmm.app.Modeller`

These objects do contain enough information to build a Universe, but a Universe should contain everything necessary to build them as well. This opens MDAnalysis compatibility and topology modelling capabilities to users preparing OpenMM simulations.

### MDAnalysis reporter for OpenMM

OpenMM uses a mechanism called reporters to stop the MD loop and let python do something such as writing a frame to disk. OpenMM provides a collection of reporters. However, a reporter built on top of MDAnalysis brings to capabilities: it allows OpenMM to write the running simulation is any format that MDAnalysis can write, and it allows to run analyses on the fly as the simulation runs.

### Universe from `simtk.openmm.System` or `simtk.openmm.app.Simulation`

[...]

Issue #2367, opened 2019-10-15
PDBx/mmCIF Reader/Topology Reader


JoaoRodrigues:
Hi all,

I've been using OpenMM to run simulations for the past couple of years. Unlike other tools, OpenMM does not write a 'topology' file (e.g. `.gro`, `.prmtop`) but instead creates a Topology object on the fly when loading structures. Because simulation systems tend to be quite large, I've take to use PDBx/mmCIF files as my default file format for writing structures that I use as topologies.

I'd like to start using MDAnalysis but right now this involves jumping through a bunch of hoops to get my topologies in a format that is parseable. It'd be much easier if I could just load a PDBx/mmCIF file as a topology, specially since it's now the default file format for structures in the PDB.

To this end, I've started working on writing a simple `PDBxParser` class. I wouldn't mind extending it to a PDBxReader/Writer class but that is not necessarily a priority (specially the writer). Would this be an interesting feature to add in your opinion? See the code [here](https://github.com/JoaoRodrigues/mdanalysis/blob/mmcif_topology_reader/package/MDAnalysis/topology/PDBxParser.py), I modeled the class after `PDBParser`.

Thanks for the great work with the library so far!

orbeckst:
Hi @JoaoRodrigues , new formats are a great addition. Can you just create a PR and then we can comment directly on the code, run tests, etc?

JoaoRodrigues:
Sounds good. Will do. Thanks!

orbeckst:
Pure python implementation: https://github.com/Electrostatics/mmcif_pdbx

orbeckst:
Also, [chemfiles reads mmCIF](http://chemfiles.org/chemfiles/latest/formats.html) and can be used inside MDAnalysis, see [Reading trajectories with chemfiles](https://docs.mdanalysis.org/2.0.0-dev0/documentation_pages/coordinates/chemfiles.html).

orbeckst:
Note on trying to use chemfiles:

With the 2.0.0b,
```python
u = mda.Universe("1ake.cif.gz", topology_format="chemfiles")
```

fails with 

    'CHEMFILES' isn't a valid topology format, nor a coordinate format

I can't just read it with
```pythin
u = mda.Universe("1ake.cif.gz", format="chemfiles")
```

either as it gives the same ValueError.

Just trying something stupid with the topology as a PDB also fails 
```python
[...]

Issue #2887, opened 2020-08-07
Attributes that cannot be serialized inside Universe


yuxuanzhuang:
To achieve flawless parallelism in analysis, everything inside Universe has to be serialized. In this PR list all the attributes that cannot be serialized (or have not been thoroughly tested) at the moment:
- `Transformation`
- `Auxiliary`
- `Converter`

Related Issues and fixes:
- Transformation #2860

orbeckst:
As discussed, we will need a way to fail serialization if any of the above are detected. Is `RuntimeError` the best choice for the exception when Transformations and Auxiliaries are present (because they can be added at run time by the user) and `TypeError` for Converters (because that is set when creating the Universe)?? Or should it be consistent, say, `TypeError` for everything because objects are present that cannot be serialized?

darmis007:
Hello @orbeckst sir and @yuxuanzhuang i would like to take up this issue under your guidance

orbeckst:
Hi @darmis007 , thank you for offering to help here. @yuxuanzhuang is working on these problems as part of his ongoing GSoC project (that's what the tag "GSoC" indicates) so unless @yuxuanzhuang indicates that he needs specific help on any of these issues I think your time will be better spent on other issues (e.g., completing your PR #2788).

It admit that it wasn't clear that @yuxuanzhuang is actively working on this issue so I assigned him.

PR #2917, opened 2020-08-25
Converter for OpenMM

ahy3nz:
Partially addresses https://github.com/MDAnalysis/mdanalysis/issues/2863 and https://github.com/MDAnalysis/mdanalysis/issues/2367

 - Add topology converters for openmm Simulation, PDBFile, Modeller, PDBxFile, and Topology objects
 - Add coordinate converters for openmm Simulation, PDBFile, Modeller, and PDBxFile objects
 - Tests included

Openmm [has a lot of classes that wrap fileformats](http://docs.openmm.org/7.0.0/api-python/app.html), but they all have properties for positions (numpy arrays) and topologies (openmm topology). I'm not sure what the "best" way to handle this is, but I went with trying to create one mdanalysis reader class for each openmm object (they end up re-using a lot of functions). Alternatively, could you create a universe from a numpy array of positions + the openmm topology? (if you can, I don't know how) This could be a uniform pattern for converting a lot of openmm objects into mdanalysis universe objects while reducing the number of reader classes.

When testing the openmm Topology reader, I had to specify the `topology_format` flag -- do the `_READERS` only infer for coordinate readers?

There's been nothing done with trying to infer angles/dihedrals/etc from the forces within an openmm system. There's also been nothing done with trying to create openmm objects from mdanalysis objects

I tried to adjust some yml files that install from conda. I don't think openmm is available on pip

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


orbeckst:
Hi @ahy3nz thanks for giving this one a try. Please be a little patient with the core devs... we're at the end of GSoC and most of the time for reviewing PRs currently goes into getting GSoC PRs wrapped up. You can also email the developer list with questions, sometimes that gets a quick reply, especially if one can answer in a sentence or two. But rest assured, you're not being ignore, we're just stretched thin!

orbeckst:
@ahy3nz if you need some additional eyes or have questions, feel free to ping; @IAlibay @lilyminium (and possibly others) were also interested in OpenMM integration.

.appveyor.yml
```diff
@@ -9,8 +9,8 @@ cache:
 
 environment:
     global:
-        CONDA_CHANNELS: conda-forge
-        CONDA_DEPENDENCIES: pip setuptools wheel cython biopython networkx joblib matplotlib scipy vs2015_runtime pytest mmtf-python GridDataFormats hypothesis pytest-cov codecov chemfiles tqdm tidynamics>=1.0.0 rdkit>=2020.03.1 h5py
+        CONDA_CHANNELS: conda-forge omnia
```
IAlibay, line 12: Just a thought for the future, we should raise an issue here to remind us to keep track of when openmm goes full conda-forge (I know it's in beta right now).

testsuite/MDAnalysisTests/datafiles.py
```diff
@@ -198,6 +198,7 @@
     "MMTF_NOCRYST", # File with meaningless CRYST1 record (Issue #2679, PR #2685)
     "FHIAIMS", # to test FHIAIMS coordinate files
     "SDF_molecule"  # MDL SDFile for rdkit
```
lilyminium, line 200:
```suggestion
    "SDF_molecule",  # MDL SDFile for rdkit
```

This missing comma is just breaking all the tests :-)

lilyminium (changes requested):

testsuite/MDAnalysisTests/topology/test_openmm.py
```diff
@@ -0,0 +1,137 @@
...
+# Python in Science Conference, pages 102-109, Austin, TX, 2016. SciPy.
+# doi: 10.25080/majora-629e541a-00e
+#
+# N. Michaud-Agrawal, E. J. Denning, T. B. Woolf, and O. Beckstein.
+# MDAnalysis: A Toolkit for the Analysis of Molecular Dynamics Simulations.
+# J. Comput. Chem. 32 (2011), 2319--2327, doi:10.1002/jcc.21787
+#
+import pytest
+import numpy as np
+from simtk.openmm import app
```
IAlibay, line 25: The tests are erroring out here, this needs an importorskip probably, although note we'll probably want to be able to test any warning that gets thrown if openmm isn't available, e.g.: https://github.com/MDAnalysis/mdanalysis/blob/b2b7bcbdf4e2fe4e221d5e67842c90cf317375e5/testsuite/MDAnalysisTests/coordinates/test_rdkit.py#L56-L63

ahy3nz:
Most of the tests are passing, just 1 in travis and 1 in appveyor aren't. I can't really debug why those two failed -- just a build error? Would a restart of those tests be sufficient?

orbeckst:
I tried to restart but wasn't allowed to because the PR was in non-mergeable state. Resolving the merge conflict in CHANGELOG kicked off a new round of CI anyway...

jbarnoud:
I think that the problem with CI is that OpenMM is not yet packaged for python 3.8 so python 3.8 jobs cannot pass the install stage.

I suggest that the OpenMM tests are skipped if OpenMM cannot be found and that the 3.8 jobs do not install OpenMM (as they cannot anyway).

testsuite/MDAnalysisTests/coordinates/test_openmm.py
```diff
@@ -0,0 +1,198 @@
...
+from MDAnalysisTests.coordinates.reference import RefAdKSmall
+
+from MDAnalysisTests.coordinates.base import _SingleFrameReader
+
+import MDAnalysis as mda
+
+mm = pytest.importorskip("simtk.openmm")
+unit = pytest.importorskip("simtk.unit")
+app = pytest.importorskip("simtk.openmm.app")
+pmd = pytest.importorskip("parmed")
```
jbarnoud, line 37: Why do you need parmed?

jbarnoud (commented):
I had only a very superficial look at the PR as I should be doing something else. Overall it looks quite good, but I would like to see more docs. Especially, I would like to see exemples of how to use the code as an end-user. I would also like to see tests with a minimal topology and simulation: not openmm objects created from a file, but ones created from scratch with as little information in. I would like to be sure you do not assume a field to be there while it may not be.

jbarnoud:
> When testing the openmm Topology reader, I had to specify the topology_format flag -- do the _READERS only infer for coordinate readers?

Readers are for coordinates, parsers are for topologies.

> There's been nothing done with trying to infer angles/dihedrals/etc from the forces within an openmm system. There's also been nothing done with trying to create openmm objects from mdanalysis objects

While we could infer some of the topology from the potentials, I doubt it is a good thing to do at this point. OpenMM flexibility when it comes to potentials means we can only get a small part right, I'd rather have nothing by default rather than unreliable data.

package/MDAnalysis/coordinates/OpenMM.py
```diff
@@ -0,0 +1,165 @@
...
+
+        state = self.filename.context.getState(-1, getVelocities=True, 
+                getForces=True, getEnergy=True)
+
+        n_atoms = self.filename.context.getSystem().getNumParticles()
+
+        ts = self._Timestep(n_atoms, **self._ts_kwargs)
+        ts.frame = 0
+        ts.data["time"] = state.getTime()._value
+        ts.data["potential_energy"] = state.getPotentialEnergy()
```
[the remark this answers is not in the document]
ahy3nz: Yes they'll be `simtk.unit.Quantity` -- I've added some functionality to convert the simtk energy into an internal float

ahy3nz:
Thanks for the comments. I've added some example usage and clarifying text on OpenMM objects within the Application layer. Changelog, unnecessary imports, and formatting should be addressed. 

OpenMM is indeed not available on python38, so I'm guessing that will crash the tests when trying to create any python38 environment, regardless of the pytest importorskip. Is there a clean way to modify the CI recipes to avoid installing openmm in any of the 38 environments?

package/MDAnalysis/topology/OpenMMParser.py
```diff
@@ -0,0 +1,186 @@
...
+        -------
+        top : MDAnalysis.core.topology.Topology
+
+        """
+
+        atom_resindex = [a.residue.index for a in omm_topology.atoms()]
+        residue_segindex = [r.chain.index for r in omm_topology.residues()]
+        atomids = [a.id for a in omm_topology.atoms()]
+        atomnames = [a.name for a in omm_topology.atoms()]
+        #atomtypes = guess_types(atomnames)
```
jbarnoud, line 115: Looks like the atom type is actually mandatory.

package/MDAnalysis/topology/OpenMMParser.py
```diff
@@ -0,0 +1,186 @@
...
+        bond_types = [b.type for b in omm_topology.bonds()]
+
+        n_atoms = len(atomids)
+        n_residues = len(resids)
+        n_segments = len(segids)
+
+        attrs = [
+            Atomids(np.array(atomids, dtype=np.int32)),
+            Atomnames(np.array(atomnames, dtype=object)),
+            #Atomtypes(np.array(atomtypes, dtype=object)),
```
jbarnoud, line 134: Same as above.

jbarnoud (commented):
Regarding OpenMM and python 3.8, this will require to do some changes in <https://github.com/MDAnalysis/mdanalysis/blob/develop/.travis.yml>. I think the simplest solution would be to add a global variable `CONDA_OPENMM=""`, to have `CONDA_DEPENDENCIES` use in the same way as it uses `CONDA_MIN_DEPENDENCIES`, and to set `CONDA_OPENMM` to the expected value (either `"openmm"` or `""`) when defining the test matrix on line <https://github.com/MDAnalysis/mdanalysis/blob/a9543ba29083449944b34d0b77429cab12f2409a/.travis.yml#L45>.

ahy3nz:
At this point, travis and appveyor configs have been modified to test openmm on the 36 environments, but not 38 environments. 

I'm not sure what's the issue with azure -- the tests are failing due to H5MD? From looking at the azure pipelines yml, I can't immediately tell if there's a place for me to add some lines for conda installing openmm (going off precedent, it looks like conda + rdkit isn't involved in azure either)

For code cov reports, which jobs should I be modifying in the build matrices?

tylerjereddy:
> I'm not sure what's the issue with azure 

It seems like the same failures are observed in `develop` so probably not related. I'll open an issue for that.

IAlibay:
Sorry I've not really been keeping up with this PR. Thought I'd just mention that the lastest OpenMM is now distributed via conda-forge and supports py3.8 which might help remove some of the complexity here.

ahy3nz:
Thanks, I think most of the code and tests are in place, but I'll have to go back and update for GHA as well

ahy3nz:
@IAlibay Let me know if there's anything else I should do. The code coverage dropped a little, but for a pretty insignificant unit conversion function

IAlibay:
Thanks @ahy3nz apologies the review has taken so long here. Unfortunately I'm a bit snowed under by teaching, but hopefully one of us can have a review at some point this one.

ahy3nz:
@IAlibay @lilyminium @orbeckst @jbarnoud Please let me know if I need any additional changes

jbarnoud:
@ahy3nz Sorry I did not look at this earlier: I did not have much time to spend on MDAnalysis lately. I'll try to have a look very soon.

ahy3nz:
@lilyminium @IAlibay @orbeckst @jbarnoud Please let me know what else needs to be done. Otherwise, if this is currently an undesired feature/approach, I can go ahead and close this PR

IAlibay:
@ahy3nz this is definitely still a very much desired feature.

I'm so sorry about the delay here, the main issue we're facing is that we have a crazy large queue of fixes to do before we can even start concentrating on v2.0 stuff + lots of GSoC stuff to deal with :(

If one of the @MDAnalysis/coredevs has capacity to take over the review here that would be great!

I'll put it as a milestone for 2.0, so hopefully that'll get someone to finish up the review ASAP.

.disabled-travis.yml
```diff
@@ -30,8 +30,9 @@ env:
     - SETUP_CMD="${PYTEST_FLAGS}"
     - BUILD_CMD="pip install -e package/ && (cd testsuite/ && python setup.py build)"
     - CONDA_MIN_DEPENDENCIES="mmtf-python biopython networkx cython matplotlib scipy griddataformats hypothesis gsd codecov"
-    - CONDA_DEPENDENCIES="${CONDA_MIN_DEPENDENCIES} seaborn>=0.7.0 clustalw=2.1 netcdf4 scikit-learn joblib>=0.12 chemfiles tqdm>=4.43.0 tidynamics>=1.0.0 rdkit>=2020.03.1 h5py==2.10.0"
-    - CONDA_CHANNELS='biobuilds conda-forge'
+    - CONDA_OPENMM=""
+    - CONDA_DEPENDENCIES="${CONDA_MIN_DEPENDENCIES} ${CONDA_OPENMM} seaborn>=0.7.0 clustalw=2.1 netcdf4 scikit-learn joblib>=0.12 chemfiles tqdm>=4.43.0 tidynamics>=1.0.0 rdkit>=2020.03.1 h5py==2.10.0"
+    - CONDA_CHANNELS='biobuilds conda-forge omnia'
```
lilyminium, line 35: I think openmm is on conda forge now

package/MDAnalysis/coordinates/OpenMM.py
```diff
@@ -0,0 +1,199 @@
...
+                state.getKineticEnergy()._value)
+        ts.triclinic_dimensions = state.getPeriodicBoxVectors(
+                asNumpy=True)._value
+        ts.positions = state.getPositions(asNumpy=True)._value
+        ts.velocities = state.getVelocities(asNumpy=True)._value
+        ts.forces = state.getForces(asNumpy=True)._value
+
+        return ts
+
+    def convert_energies_from_native(self, energy):
```
lilyminium, line 132: For this and the other conversion functions -- doesn't `getPotentialEnergy()` and the like return the quantity with units already attached (like simtk.physical.Quantity and simtk.physical.Unit)? Couldn't you just convert from that to the MDA unit instead?

package/MDAnalysis/coordinates/OpenMM.py
```diff
@@ -0,0 +1,199 @@
...
+        Parameters
+        ----------
+        energy: array_like
+          energies to transform
+
+        """
+        f = units.get_conversion_factor(
+            'energy', self.units['energy'], 'kJ/mol')
+        if f == 1.:
+            return energy
```
lilyminium, line 144: Could this be removed?

package/MDAnalysis/coordinates/OpenMM.py
```diff
@@ -0,0 +1,199 @@
...
+        self.ts = self._mda_timestep_from_omm_app()
+
+        if self.convert_units:
+            self.convert_pos_from_native(self.ts._pos)
+            self.ts.triclinic_dimensions = self.convert_pos_from_native(
+                self.ts.triclinic_dimensions, inplace=False
+            )
+
+    def _mda_timestep_from_omm_app(self):
+        """ Construct Timestep object from Openmm Application object """
```
lilyminium, line 184:
```suggestion
        """ Construct Timestep object from OpenMM Application object """
```

testsuite/MDAnalysisTests/coordinates/test_openmm.py
```diff
@@ -0,0 +1,241 @@
...
+    __test__ = True
+
+    def setUp(self):
+        self.universe = mda.Universe(app.PDBFile(RefAdKSmall.filename))
+        self.ref = mda.Universe(RefAdKSmall.filename)
+        self.prec = 3
+
+    def test_dimensions(self):
+        assert_almost_equal(
+            # Angles seem to parse differently when openmm reads the pdb file
```
lilyminium, line 93: How do they differ?

testsuite/MDAnalysisTests/coordinates/test_openmm.py
```diff
@@ -0,0 +1,241 @@
...
+
+class TestOpenMMModellerReader(_SingleFrameReader):
+    __test__ = True
+
+    def setUp(self):
+        pdb_obj = app.PDBFile(RefAdKSmall.filename)
+        modeller = app.Modeller(pdb_obj.topology, pdb_obj.positions)
+        self.universe = mda.Universe(modeller)
+        self.ref = mda.Universe(RefAdKSmall.filename)
+        self.prec = 3
```
lilyminium, line 115: You only use this once, it is easier to understand if you just use 3 in the test below.

testsuite/MDAnalysisTests/coordinates/test_openmm.py
```diff
@@ -0,0 +1,241 @@
...
+            "OpenMMSimulationReader failed to get unitcell dimensions " +
+            "from OpenMMSimulation",
+        )
+
+    def test_coordinates(self):
+        up = self.universe.atoms.positions
+        rp = self.ref.atoms.positions
+        assert_almost_equal(up, rp, decimal=3)
+
+    def test_pickle_singleframe_reader(self):
```
lilyminium, line 166: Was this going to be a test?

lilyminium (changes requested):
Thanks @ahy3nz and so sorry for the *long* delay. I'm not super familiar with OpenMM so I have some questions about units and the PDB angle parsing.

lilyminium:
Also if you could fix the merge conflicts and PEP8 as noted by the bot (https://github.com/MDAnalysis/mdanalysis/pull/2917#issuecomment-679459947), that'd be great.

package/MDAnalysis/coordinates/OpenMM.py
ahy3nz: Yes it's true the getters return simtk.Quantity objects, but I wasn't sure what the correct syntax would be to map from the simtk Unit to floats in a correct MDAnalysis manner. If preferred, we can do the unit conversion solely with the simtk unit API

testsuite/MDAnalysisTests/coordinates/test_openmm.py
ahy3nz: It seems like self.prec gets called somewhere in the base tests like in test_distance. I can hardcode the precision in the specific openmm tests, but it still seems like setting self.prec seems important for other tests?

testsuite/MDAnalysisTests/coordinates/test_openmm.py
ahy3nz: Since the underlying openmm objects are c-objects translated with swig (or something along those lines), there were issues trying to pickle these openmm python objects. To be honest, I don't know what's the best way to handle the pickle test, so I just ignored it with `pass`. And if I don't overwrite this test, I think it defaults to the basetest, which will try to pickle the reader, leading to `TypeError: can't pickle SwigPyObject objets`

testsuite/MDAnalysisTests/coordinates/test_openmm.py
ahy3nz: In the tests for a non-orthogonal box, the parsed angles are 120, 120, 90 while the ref trajectory angles are 60, 60, 90 (box lengths are fine though). These were parsed directly from the openmm Topology's `getPeriodicBoxVectors()` method. I don't _think_ openmm and mdanalysis define their triclinic unit cells differently, so the difference might be how openmm reads the pdb files (with openmm's pdb file reader) and parses box vectors vs how mdanalysis reads pdb files and parses box vectors, which leads to the downstream consequences here? For what it's worth, the orthogonal boxes convert angles perfectly fine

testsuite/MDAnalysisTests/coordinates/test_openmm.py
lilyminium: Ah, if it ain't broke, let's not fix it.

testsuite/MDAnalysisTests/coordinates/test_openmm.py
lilyminium: @orbeckst @yuxuanzhuang should we note this as an exception then?

package/MDAnalysis/coordinates/OpenMM.py
lilyminium: Thanks, I think that's a bit more foolproof than `_value`.

testsuite/MDAnalysisTests/coordinates/test_openmm.py
yuxuanzhuang: Generally speaking, we have not started to do a thorough test for the serialization of converters (#2887), so it largely depends on the upstream support. 

From my very quick test on the picklability of the `openmm.Simulation` object, it seems that most of the data structures can be serialized except `('context', <simtk.openmm.openmm.Context; proxy of <Swig Object of type 'OpenMM::Context *'`

It seems that the openmm people only test the picklability of the `state` obtained from the `Context` https://github.com/openmm/openmm/blob/be19e0222ddf66f612016a3c1f687161a53c2396/wrappers/python/tests/TestPickle.py#L65. As I have no experience in openmm, I cannot give a better solution other than noting this as an exception (and commenting on to this PR #2887).

orbeckst: Quick drive-by comment (sorry if off-topic): Please do not have tests just `pass` — that implies that we have functionality that's not present. Instead decorate it with an xfail (unless there's a better solution).

testsuite/MDAnalysisTests/coordinates/test_openmm.py
lilyminium: I see, it comes down to OpenMM reducing its triclinic vectors (https://github.com/openmm/openmm/blob/f3720f62e94eea15a369c1131ff3911b7d4d0933/wrappers/python/openmm/app/internal/unitcell.py#L76-L81) which in this case inverts the sign.

@richardjgowers will this affect calculations at all?

If not, @ahy3nz could you alter the test to check for equal angles or 180-angle?

testsuite/MDAnalysisTests/coordinates/test_openmm.py
lilyminium: Thank you @yuxuanzhuang for looking into it. @ahy3nz if you could change the test to xfail with this reasoning and maybe a link to #2887 or @yuxuanzhuang's comment, that would be great!

package/MDAnalysis/topology/OpenMMParser.py
```diff
@@ -0,0 +1,189 @@
...
+        -------
+        top : MDAnalysis.core.topology.Topology
+
+        """
+
+        atom_resindex = [a.residue.index for a in omm_topology.atoms()]
+        residue_segindex = [r.chain.index for r in omm_topology.residues()]
+        atomids = [a.id for a in omm_topology.atoms()]
+        atomnames = [a.name for a in omm_topology.atoms()]
+        atomtypes = guess_types(atomnames)
```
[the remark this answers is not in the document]
lilyminium: I agree with @jbarnoud here -- do you need to have types @ahy3nz?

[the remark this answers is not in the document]
jbarnoud: At some point, types were mandatory. IIRC @ahy3nz tried to not have types and it blew up. Did we change anything in that respect? In any case, they should not be guessed.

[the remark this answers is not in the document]
IAlibay: I think they still are (for now), that being said, we did agree at some point that if types weren't available but that elements were we should be using those. I'd recommend just doing that instead.

testsuite/MDAnalysisTests/coordinates/test_openmm.py
IAlibay: Even if it didn't affect calculations, is there any way we can convert this? Whilst we often fail to achieve this, it would be nice to have consistency when we return things on the MDA side of things.

jbarnoud:
I am looking at this PR now. I tested it on my use case and it worked neatly! I still need to understand why codecov reports numbers so low for the test coverage while I do see quite a lot of tests. I will also check how the doc looks and fix for PEP8 where relevant.

jbarnoud:
The low coverage was because of the restrictions github recently added to the CI. I allowed the workflows to run and the patch coverage is 98.43%.

The docstrings do not appear in the documentation and the linter still complains about some PEP8 errors. Instead of asking @ahy3nz to grind through a bunch of tiny trivial fixes, I opened a PR against their branch: https://github.com/ahy3nz/mdanalysis/pull/1

With these small fixes, the PR looks good to go for me.

jbarnoud:
Note that it would be very cool to have this merged in for version 2.0. The feature freeze for that version is on Friday so it would be great if you could merge my changes by then. Sorry to rush you after I made you wait for basically ever, but the feature really is great!

IAlibay:
@ahy3nz this is your first PR for MDAnalysis right? Please don't forget to add yourself to AUTHORS and update the CHANGELOG too! 🎉

jbarnoud (approved):
Looks good to me :rocket:

IAlibay (approved):
I'm not super acquainted with the inner workings of OpenMM, but lgtm

jbarnoud:
Congratulations on your first contribution @ahy3nz! It as a substantial one and I am very glad it is in. I am very sorry it took so long.

PR #2928, opened 2020-09-02
Deprecate helanal

lilyminium:
 - Deprecate helanal from #2622 

Sorry for the delay in adding the helanal deprecation notice.

### DIFF
--- a/package/MDAnalysis/analysis/helanal.py
+++ b/package/MDAnalysis/analysis/helanal.py
@@ -123,13 +123,14 @@ Functions
 .. autofunction:: helanal_main
 
 """
+from __future__ import print_function, division, absolute_import
 import os
 
 import numpy as np
 
 import MDAnalysis
-from MDAnalysis.lib.log import ProgressBar
-from MDAnalysis.lib import mdamath
+from ..lib.log import ProgressBar
+from ..lib import mdamath, util
 
 import warnings
 import logging
@@ -168,6 +169,9 @@ def mean_abs_dev(a, mean_a=None):
         mean_a = np.mean(a)
     return np.mean(np.fabs(a - mean_a))
 
+
+@util.deprecate(release="1.0.1", remove="2.0",
+                message="In 2.0 use the upcoming helix_analysis module")
 def helanal_trajectory(universe, select="name CA",
                        begin=None, finish=None,
                        matrix_filename="bending_matrix.dat",
@@ -258,12 +262,6 @@ def helanal_trajectory(universe, select="name CA",
        helanal_trajectory is deprecated and will be removed in 2.0.0.
        Please use helix_analysis.HELANAL instead (available in 2.0.0).
     """
-
-    warnings.warn("This function is deprecated as of MDAnalysis version 1.0. "
-                  "It will be removed in MDAnalysis version 2.0. Please use "
-                  "MDAnalysis.analysis.helix_analysis.HELANAL instead "
-                  "(available in 2.0.0)",
-                  category=DeprecationWarning)
 
     if ref_axis is None:
         ref_axis = np.array([0., 0., 1.])
@@ -546,6 +544,8 @@ def stats(some_list):
     return [list_mean, list_sd, list_abdev]
 
 
+@util.deprecate(release="1.0.1", remove="2.0",
+                message="In 2.0 use the upcoming helix_analysis module")
 def helanal_main(pdbfile, select="name CA", ref_axis=None):
     """Simple HELANAL_ run on a single frame PDB/GRO.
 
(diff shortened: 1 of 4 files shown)


package/MDAnalysis/analysis/helanal.py
```diff
@@ -116,9 +123,6 @@
 .. autofunction:: helanal_main
 
 """
-from __future__ import print_function, division, absolute_import
```
IAlibay, line 119: If this is targetting 1.0.1 then these need to come back in right?

package/MDAnalysis/analysis/helanal.py
```diff
@@ -131,6 +135,12 @@
 import logging
 logger = logging.getLogger("MDAnalysis.analysis.helanal")
 
+warnings.warn("This module is deprecated as of MDAnalysis version 1.0. "
```
IAlibay, line 138: We'll probably want some test for this? (not sure how strict we are being with 1.0.1 tests, but if it's getting released we should try to get coverage down?)

IAlibay: Actually the tests are failing here, so we it'll need fixing anyways.

package/MDAnalysis/analysis/helanal.py
```diff
@@ -243,7 +253,18 @@ def helanal_trajectory(universe, select="name CA",
 
     .. versionchanged:: 1.0.0
        Changed `selection` keyword to `select`
+
+    .. deprecated:: 1.0.1
+       helanal_trajectory is deprecated and will be removed in 2.0.0.
+       Please use helix_analysis.HELANAL instead (available in 2.0.0).
     """
+
+    warnings.warn("This function is deprecated as of MDAnalysis version 1.0. "
```
IAlibay, line 262: I think we have this nice decorator for deprecations under lib.utils, you should be able to just do 

`@deprecate(release="1.0.1", remove="2.0", message="In 2.0 use the upcoming helix_analysis module.")`

for all these functions

package/MDAnalysis/analysis/helanal.py
```diff
@@ -581,8 +602,18 @@ def helanal_main(pdbfile, select="name CA", ref_axis=None):
     .. versionchanged:: 1.0.0
        Changed `selection` keyword to `select`
 
+    .. deprecated:: 1.0.1
+       helanal_trajectory is deprecated and will be removed in 2.0.0.
+       Please use helix_analysis.HELANAL instead (available in 2.0.0).
+
     """
 
+    warnings.warn("This function is deprecated as of MDAnalysis version 1.0. "
```
IAlibay, line 611: As above.

IAlibay (changes requested):
Couple of changes, probably would be worth switching over to the deprecate decorator. Also we'll need to test warnings (or at least fix the test for warnings).

package/MDAnalysis/analysis/helanal.py
orbeckst: yes

orbeckst:
Just fyi, you can also make a PR against master, that's fine, too.

testsuite/MDAnalysisTests/analysis/test_helanal.py
```diff
@@ -198,14 +198,13 @@ def test_warnings(tmpdir):
                 u, select="name CA", begin=-1, finish=99999
             )
 
-            assert len(record) == 2
+            assert len(record) == 3
```
IAlibay, line 201: I guess that works as a really minimal test 👍

package/MDAnalysis/analysis/helanal.py
```diff
@@ -117,20 +124,24 @@
...
 import logging
 logger = logging.getLogger("MDAnalysis.analysis.helanal")
 
+warnings.warn("This module is deprecated as of MDAnalysis version 1.0. "
+              "It will be removed in MDAnalysis version 2.0."
+              "Please use (available in 2.0.0) "
+              "MDAnalysis.analysis.helix_analysis instead.",
+              category=DeprecationWarning)
+
 def center(coordinates):
```
IAlibay, line 145: These are all public facing functions, do we need to add deprecation warnings here too or do we just assume the warning on module importing will be sufficient?

IAlibay (approved):
lgtm, just a couple of comments that might be worth considering.

@orbeckst since this is going into your 1.0.1 PR I'll let you do the merge?

package/MDAnalysis/analysis/helanal.py
orbeckst: I think the module import warning is sufficient.

testsuite/MDAnalysisTests/analysis/test_helanal.py
orbeckst: Curious that DeprecationWarning counts as a UserWarning... but ok, if it works...

orbeckst:
Thank you @lilyminium and thanks @IAlibay for the review!

PR #2937, opened 2020-09-07
Capped distance fix

### LINKED ISSUE #2919: Wrong results from `capped_distance()` with "nsgrid" search and atom in the center of the box
## Expected behavior ##

`capped_distance()` function should always give the correct result regardless of the employed search method and atomic positions.
In the example below the expected result for cutoff=3.2 is:
* cutoff=3.2
bruteforce: 2497 pairs
pkdtree: 2497 pairs
nsgrid: 2497 pairs

## Actual behavior ##

When one of the atoms of the reference group is positioned exactly in the center of the box, `capped_distance()` sometimes returns erroneous results for "nsgrid" search method.

* cutoff=2.8
bruteforce: 1115 pairs
pkdtree: 1115 pairs
nsgrid: 1115 pairs
* cutoff=3.2
bruteforce: 2497 pairs
pkdtree: 2497 pairs
nsgrid: 2510 pairs

``` python
import MDAnalysis as mda
from MDAnalysis.transformations.translate import center_in_box
from MDAnalysis.tests.datafiles import PDB_xvf

u = mda.Universe(PDB_xvf)
ag = u.select_atoms('index 0')
u.trajectory.ts = center_in_box(ag)(u.trajectory.ts)

box = u.dimensions
reference = u.select_atoms('protein')
configuration = u.select_atoms('not protein')

for cutoff in [2.8, 3.2]:
    print(f"* cutoff={cutoff}")
    for method in ['bruteforce', 'pkdtree', 'nsgrid']:
        pairs, distances = capped_distance(
            reference.positions,
            configuration.positions,
            max_cutoff=cutoff,
            box=box,
            method=method,
            return_distances=True,
            )
        print(f"{method}: {len(pairs)} pairs")

```

- MDAnalysis 1.0.0 (conda-forge)
- Python 3.7.7
- Ubuntu 18.04.5 LTS

richardjgowers:
Fixes #2919 #2345 #2229

 - 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


richardjgowers:
This is mostly a rewrite of nsgrid, incorporating a lot of ideas that @zemanj raised, notably
- use our existing distance calculation code from `calc_distances.h` rather than rewriting it
- use integer shifts in cells (the subdivisions of system volume) to search neighbouring cells, (rather than an imaginary probe coordinate)

This is still WIP as I haven't looked at the (semi-) periodic case.  I do seem to get the results from issue #2919 correct though.

If I can get someone (@MDAnalysis/coredevs) else to read this and increase the bus factor that would be nice.  It might be easier to just read the new version and check that it seems correct rather than trying to understand the delta.

orbeckst:
This WIP currently segfaults https://travis-ci.com/github/MDAnalysis/mdanalysis/jobs/381842818#L1985
```
estsuite/MDAnalysisTests/lib/test_neighborsearch.py ..                  [ 51%]

testsuite/MDAnalysisTests/lib/test_nsgrid.py FFatal Python error: Segmentation fault

Thread 0x00007f5fa1c39700 (most recent call first):

  File "/home/travis/miniconda/envs/test/lib/python3.6/threading.py", line 299 in wait

  File "/home/travis/miniconda/envs/test/lib/python3.6/threading.py", line 551 in wait

  File "/home/travis/miniconda/envs/test/lib/python3.6/site-packages/tqdm/_monitor.py", line 69 in run

  File "/home/travis/miniconda/envs/test/lib/python3.6/threading.py", line 916 in _bootstrap_inner

  File "/home/travis/miniconda/envs/test/lib/python3.6/threading.py", line 884 in _bootstrap

Current thread 0x00007f5fa7219700 (most recent call first):

  File "/home/travis/build/MDAnalysis/mdanalysis/testsuite/MDAnalysisTests/lib/test_nsgrid.py", line 45 in run_grid_search

  File "/home/travis/build/MDAnalysis/mdanalysis/testsuite/MDAnalysisTests/lib/test_nsgrid.py", line 65 in test_nsgrid_badcutoff
```

orbeckst:
@richardjgowers any progress here?

orbeckst:
@zemanj could you have a look here, given that you have been instrumental in finding the issues?

zemanj:
I already skimmed through, but reviewing is a bit hard on a cell phone (I'm on vacation). Anyway, I'll try to do a proper review ASAP.

package/MDAnalysis/lib/nsgrid.pyx
```diff
@@ -407,314 +179,172 @@ cdef class NSResults(object):
...
-            Maximum total number of cells
-        force : boolean
-            Optimizes cutoff if set to ``False`` [False]
+            Minimum desired cutoff radius
+        box : numpy ndarray shape=(6,)
+            Box info, [lx, ly, lz, alpha, beta, gamma]
         """
+        cdef int i, j
+        cdef double relative_cutoff_margin
+        cdef double original_cutoff
```
tylerjereddy, line 228: Just checking the Cython analyzer locally I think these can be typed too?

```cython
        cdef int[::1] head_id
        cdef int[::1] next_id
```

package/MDAnalysis/lib/nsgrid.pyx
```diff
@@ -840,82 +457,58 @@ cdef class FastNS(object):
...
-        cdef ns_int cellindex, cellindex_probe
-        cdef ns_int xi, yi, zi
-
+        from MDAnalysis.lib import distances
+
+        cdef int i, j, size_search
+        cdef int cx, cy, cz
+        cdef int cellid
+        cdef int xi, yi, zi
+        cdef int cellcoord[3], searchcoord[3]
```
tylerjereddy, line 466: Cython seems to want these declaration on separate lines:

`warning: package/MDAnalysis/lib/nsgrid.pyx:469:26: Non-trivial type declarators in shared declaration (e.g. mix of pointers and values). Each pointer declaration should be on its own line.
`

tylerjereddy (commented):
Just read through the Cython analyzer output/html report for a quick initial look.

package/MDAnalysis/lib/nsgrid.pyx
```diff
@@ -407,314 +179,172 @@ cdef class NSResults(object):
...
+        
+        cz = <int> (coord[2] / self.cellsize[2])
+        cy = <int> ((coord[1] - cz * self.triclinic_dimensions[7]) / self.cellsize[1])
+        cx = <int> ((coord[0] - cy * self.triclinic_dimensions[3] - cz * self.triclinic_dimensions[6]) / self.cellsize[0])
+        
+        xyz[0] = cx
+        xyz[1] = cy
+        xyz[2] = cz
+
+    cdef int cellxyz2cellid(self, int cx, int cy, int cz) nogil:
```
zemanj, line 315: With PBC the branching in this function should better be replaced by modulo operations, i.e.,
```python
cx %= self.ncells[0]
cy %= self.ncells[1]
cz %= self.ncells[2]
```
For non-periodic cells, there either need to be bound checks within this function or before it is called.
Also, we must be aware of the fact that the number of cells in any dimension may be less than 3, which currently leads to double-counting cells! However, this problem could probably be better dealt with in a different place.

package/MDAnalysis/lib/nsgrid.pyx
```diff
@@ -407,314 +179,172 @@ cdef class NSResults(object):
...
+        if cz < 0:
+            cz = self.ncells[2] - 1
+        elif cz == self.ncells[2]:
+            cz = 0
+
+        return cx + cy * self.cell_offsets[1] + cz * self.cell_offsets[2]
 
 
 cdef class FastNS(object):
     """Grid based search between two group of atoms
```
zemanj, line 336: I think we're dealing with sets of coordinates, not with atoms.

package/MDAnalysis/lib/nsgrid.pyx
```diff
@@ -407,314 +179,172 @@ cdef class NSResults(object):
...
 
     cdef _NSGrid grid
-    cdef ns_int max_gridsize
     cdef bint periodic
 
-    def __init__(self, cutoff, coords, box, max_gridsize=5000, pbc=True):
+    def __init__(self, cutoff, coords, box, pbc=True):
         """
         Initialize the grid and sort the coordinates in respective
         cells by shifting the coordinates in a brick shaped box.
```
zemanj, line 350: This is no longer true, coordinates are not necessarily shifted, and the box stays as it is. Also the following two lines are now obsolete.
Also, the lines about a pseudobox don't really make sense to me here... Isn't the pseudobox automatically created if `box is None`?

package/MDAnalysis/lib/nsgrid.pyx
```diff
@@ -786,27 +412,18 @@ cdef class FastNS(object):
             raise ValueError("Any of the box dimensions cannot be 0")
```
zemanj, line 412: I think that `np.allclose()` is wrong. Shouldn't it be rather something like `if np.any(box[:3] == 0.0)` (or does `np.anyclose()` exist)?

package/MDAnalysis/lib/nsgrid.pyx
```diff
@@ -786,27 +412,18 @@ cdef class FastNS(object):
...
 
-        self.box = _PBCBox(box, self.periodic)
-
-        if cutoff < 0:
-            raise ValueError("Cutoff must be positive!")
-        if cutoff * cutoff > self.box.c_pbcbox.max_cutoff2:
-            raise ValueError("Cutoff greater than maximum cutoff ({:.3f}) given the PBC")
-
-        self.coords_bbox = self.box.fast_put_atoms_in_bbox(self.coords)
+        self.coords_bbox = distances.apply_PBC(coords, box)
```
zemanj, line 416: What if `pbc=False` and the box is too small? If a box is supplied, we should check if it is safe to use.

package/MDAnalysis/lib/nsgrid.pyx
zemanj: I second that

package/MDAnalysis/lib/nsgrid.pyx
```diff
@@ -840,82 +457,58 @@ cdef class FastNS(object):
...
             raise ValueError("search_coords must have a shape of (n, 3), got "
                              "{}.".format(search_coords.shape))
 
-        # Generate another grid to search
-        searchcoords = search_coords.astype(np.float32, order='C', copy=False)
-        searchcoords_bbox = self.box.fast_put_atoms_in_bbox(searchcoords)
-        searchgrid = _NSGrid(searchcoords_bbox.shape[0], self.grid.used_cutoff, self.box, self.max_gridsize, force=True)
-
-        size_search = searchcoords.shape[0]
+        searchcoords_bbox =  distances.apply_PBC(search_coords, self.grid.dimensions)
```
zemanj, line 478: This is problematic for non-PBC calculations if coordinates lie outside the pseudobox.

package/MDAnalysis/lib/nsgrid.pyx
```diff
@@ -840,82 +457,58 @@ cdef class FastNS(object):
...
+                        cx = cellcoord[0] - 1 + xi
+                        cy = cellcoord[1] - 1 + yi
+                        cz = cellcoord[2] - 1 + zi
+                        cellid = self.grid.cellxyz2cellid(cx, cy, cz)
+
+                        if not self.periodic and cellid == -1:  # out of bounds
+                            continue
+
+                        # for loop over atoms in searchcoord
+                        j = self.grid.head_id[cellid]
```
zemanj, line 498: This is where some of the tests (the triclinic ones) currently segfault. The reason is that `coord2cellxyz()` is called on line 484, which may yield an invalid `cellcoord` for triclinic boxes. Since `cellxyz2cellid()` (called on line 492) only checks if the coordinates are *equal to* (and not equal or greater than) `ncells`, the returned `cellid` can be out of bounds, causing a segfault.

package/MDAnalysis/lib/nsgrid.pyx
```diff
@@ -840,82 +457,58 @@ cdef class FastNS(object):
...
+                        while (j > 0):
+                            if self.grid.triclinic:
+                                d2 = calc_dist_triclinic(&searchcoords_bbox[i][0], &self.coords_bbox[j][0],
+                                                         &self.grid.triclinic_dimensions[0])
+                            else:
+                                d2 = calc_dist(&searchcoords_bbox[i][0], &self.coords_bbox[j][0],
+                                               &self.grid.dimensions[0], &self.grid.inverse_dimensions[0])
```
zemanj, line 506: The non-periodic case is not handled!

package/MDAnalysis/lib/nsgrid.pyx
```diff
@@ -840,82 +457,58 @@ cdef class FastNS(object):
...
+        size_search = search_coords.shape[0]
+        results = NSResults()
+
+        for i in range(size_search):
+            # which cell is atom *i* in
+            self.grid.coord2cellxyz(&searchcoords_bbox[i][0], &cellcoord[0])
+            # loop over all 27 neighbouring cells
+            for xi in range(3):
+                for yi in range(3):
+                    for zi in range(3):
```
zemanj, line 488: This currently won't work if the number of cells in any dimension is less than 3. Wouldn't it be nicer to have a function that computes all neighbor cell indices, taking this and pbc/non-pbc into account? Then one could just loop over an array of cell indices instead of having three nested loops. This might also be useful for the `self_search()`, as one could just omit half of the neighbor cells instead of the "routing".

package/MDAnalysis/lib/nsgrid.pyx
```diff
@@ -934,70 +527,69 @@ cdef class FastNS(object):
...
-        cdef bint check
+        cdef double cutoff2 = self.cutoff * self.cutoff
 
         size_search = self.coords.shape[0]
+        results = NSResults()
 
-        results = NSResults(self.cutoff, self.coords, self.coords)
+        # loop over 13 neighbouring pairs
+        cdef int[:, ::1] route
+        route = np.array([[1, 0, 0], [1, 1, 0], [0, 1, 0], [-1, 1, 0],
```
zemanj, line 545: As mentioned above for `search()`: It might be nicer if (relevant) neighbor cells are computed in an extra function.

zemanj (changes requested):
There are some issues with determining cell indices (causing the test segfaults), handling non-periodic boxes, and small grids with less than 3 cells in one dimension. Apart from that, this re-implementation looks really promising as it doesn't suffer from the problems the old code had! :+1:

zemanj (changes requested):
One more thing: We also need tests that reproduce the failures of issues #2229 and #2345, and maybe also #2919 with the *old* version of the code to make sure that the re-implementation really fixes all bugs.

orbeckst:
@richardjgowers is the idea to have [distopia](https://github.com/MDAnalysis/distopia) replace capped_distance and fixes in this PR or are you still working towards getting this PR ready for 1.0.1?

richardjgowers:
@orbeckst not really, for the short term distopia will just be a drop in replacement for lower down functions (`distance_array`).  This is still WIP and I'm chipping away at it.

package/MDAnalysis/lib/nsgrid.pyx
```diff
@@ -407,314 +179,172 @@ cdef class NSResults(object):
...
-    cdef ns_int coord2cellid(self, rvec coord) nogil:
+        for i in range(6):
+            self.dimensions[i] = box[i]
+        for i in range(3):
+            self.inverse_dimensions[i] = 1.0 / box[i]
+        self.triclinic_dimensions = triclinic_vectors(box).reshape((9,))
+        self.triclinic = (self.triclinic_dimensions[3] > 0 or
+                          self.triclinic_dimensions[6] > 0 or
+                          self.triclinic_dimensions[7] > 0)
+        # TODO: Something smart about defining the maximum cutoff for a given boxsize
```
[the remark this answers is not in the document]
richardjgowers: A bit late replying to this but...

Why half the boxsize?  It's ok if the sphere half fits in the subdivisions of the box.  I.e. we're looking for a sphere (with radius our search cutoff) that fits inside a 3x3 grid when the origin is in the central cubelet.

Isn't the spacing between faces just the diagonal of the triclinic_dimensions?

package/MDAnalysis/lib/nsgrid.pyx
richardjgowers: It's probably more correct as modulo math but I know that I'm calling this with `-1 <= cx <= ncells`.

WRT the minimum number of cells, this now done in `__init__` where there's either >3 or 1 cells in each dimension

The bounds check is done by returning `END (-1)` to give a NULL result

package/MDAnalysis/lib/nsgrid.pyx
richardjgowers: The cellxyz2cellid will do boundchecking, so it works for cell dimensions less than 3 now.

package/MDAnalysis/lib/distances.py
```diff
@@ -1096,9 +1096,9 @@ def _nsgrid_capped_self(reference, max_cutoff, min_cutoff=None, box=None,
             gridsearch = FastNS(max_cutoff, reference, box=box)
             results = gridsearch.self_search()
 
-        pairs = results.get_pairs()[::2, :]
+        pairs = results.get_pairs()
```
richardjgowers, line 1099: This looks odd, but `self_search()` used to add each pair twice, then we'd return every other pair

richardjgowers:
@MDAnalysis/coredevs Ok I think this is now fixed and working - I've also added the test cases from the three issues linked at the start.  It's pretty much a rewrite of the whole module so I wouldn't try and make sense of the diff, just look at the new version.  I'd also recommend a squash and merge, there's lots of embarrassingly bad commits in there.  If you've got tiny comments it might be quicker just to edit the file in place.

package/MDAnalysis/lib/nsgrid.pyx
```diff
@@ -69,249 +70,52 @@ not reflect in the results.
 .. [#] a pair correspond to two particles that are considered as neighbors .
 
 .. versionadded:: 0.19.0
+.. versionchanged:: 2.0.0
```
orbeckst, line 73: Can't this go in 1.0.2 ?

package/MDAnalysis/lib/nsgrid.pyx
```diff
@@ -69,249 +70,52 @@ not reflect in the results.
...
+    if a < b:
+        return a
+    else:
+        return b
 
-        return bbox_coords
+cdef inline float degsin(float deg):
+    # sin in degrees
+    if (deg > 90):
+        deg = 180 - deg
```
orbeckst, line 115: Just out of curiosity: why do you need to transform the angle? In general sin(π -x) = sin(x). I assume 0 ≤ x < π (*is this true?*) so you'd always have sin(x) ≥ 0. This only makes a difference if you can have x outside the range but then I don't quite understand what the ultimate goal of this function is.

package/MDAnalysis/lib/nsgrid.pyx
```diff
@@ -321,39 +125,13 @@ cdef class NSResults(object):
...
-
-        self.cutoff = cutoff
-        self.coords = coords
-        self.searchcoords = searchcoords
+    cdef vector[int] pairs
+    cdef vector[double] distances2
 
-        self.npairs = 0
+    def __init__(self):
+        pass
```
orbeckst, line 132: Do you need to have init when it doesn't do anything?

testsuite/MDAnalysisTests/lib/test_nsgrid.py
```diff
@@ -46,24 +50,25 @@ def run_grid_search(u, ref_id, cutoff=3):
 
     return searcher.search(searchcoords)
 
-def test_pbc_box():
+@pytest.mark.parametrize('box', [
+    np.zeros(3),  # Bad shape
+    np.zeros((3, 3)),  # Collapsed box
+    np.array([[0, 0, 0], [0, 1, 0], [0, 0, 1]]),  # 2D box
+    np.array([[1, 0, 0], [0, 1, 0], [0, 0, 1]]),  # Box provided as array of integers
+    np.array([[1, 0, 0], [0, 1, 0], [0, 0, 1]]),  # Box provided as array of double
```
orbeckst, line 58: looks like integers and not floats

package/MDAnalysis/lib/nsgrid.pyx
orbeckst: > I know that I'm calling this with -1 <= cx <= ncells.

Would it be worthwhile to add `assert` here for anything you know to be true?

orbeckst (commented):
I only had a quick look and I can't substantially comment on the algorithm. I like the tests for the issues #2919 #2229 #2345. 

Assuming someone with better insight into the algorithmic parts (@zemanj ?) approves it, I am ok, too. 

Could this fix be backported for a 1.0.2 so that we have a fully functioning 1.x again?

package/MDAnalysis/lib/nsgrid.pyx
richardjgowers: Yeah no reason why not, I think this is now @IAlibay 's call

package/MDAnalysis/lib/nsgrid.pyx
richardjgowers: Just me being bad at trig :)

PR #2882, opened 2020-07-30
Converters API

cbouy:
Fixes #2790 

Changes made in this Pull Request:
 - Makes `u.atoms.convert_to("PACKAGE")` case insensitive
 - Pass kwargs to converters: `u.atoms.convert_to("rdkit", NoImplicit=False)`
 - Adds `u.atoms.convert_to.package()` methods automatically from the metaclass magic, with TAB completion support, and docstring from the converter

The same logic can be applied to writers if that's needed

The tests will fail for now as I have written them with #2775 in mind.

Also, I'm not sure if I've put the code in the right place and if the names I've come up with are relevant.

I'm not sure how this will work out with Sphinx though, since all the `convert_to.package()` methods are made automatically through `setattr`. Similar concern for `convert_to` in AtomGroup, is there a way to tell Sphinx to show a particular docstring instead of considering it like an attribute, and if yes, which docstring should it show ?

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?
 - [x] Issue raised/referenced?

PR #2942, opened 2020-09-13
Improving the RDKitConverter caching system

cbouy:
The current "homemade" caching system in the RDKit converter only allows to store the most recent conversion.
This new version uses the `functools.lru_cache` which allows users to select how many molecules should be cached, and improves readability/maintainability IMO

Also, the new caching system retrieves the converted items from the hash of all the arguments passed to the decorated `atomgroup_to_mol` function, instead of the id of the atomgroup and the arguments, which makes more sense. I didn't know what a hash was until recently so please forgive me for the rookie mistake :D 
Now if you successively run `u.atoms.convert_to("RDKIT")` it will benefit from the caching system.

I needed to convert two different atomgroups (protein and ligand) while iterating over a trajectory and the previous system would just rebuild the whole topology (which takes quite some time for a protein) for each molecule at every frame hence why I think this is necessary. Now it works like a breeze.

 - Replaces the RDKitConverter cache system with `functools.lru_cache`
 - Adds the `set_converter_cache_size(maxsize)` function to modify how many items are retained in the cache
 - Moved `atomgroup_to_mol` outside of the RDKitConverter class (it's not really needed there anyway), otherwise I need to define hash and eq dunders for the caching to work

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/coordinates/RDKit.py
+++ b/package/MDAnalysis/coordinates/RDKit.py
@@ -67,6 +67,7 @@ Classes
 import warnings
 import re
 import copy
+from functools import lru_cache
 
 import numpy as np
 
@@ -244,19 +245,13 @@ class RDKitConverter(base.ConverterBase):
     Since one of the main use case of the converter is converting trajectories
     and not just a topology, creating a new molecule from scratch for every
     frame would be too slow so the converter uses a caching system. The cache
-    only remembers the id of the last AtomGroup that was converted, as well
-    as the arguments that were passed to the converter. This means that using
-    ``u.select_atoms("protein").convert_to("RDKIT")`` will not benefit from the
-    cache since the selection is deleted from memory as soon as the conversion
-    is finished. Instead, users should do this in two steps by first saving the
-    selection in a variable and then converting the saved AtomGroup. It also
-    means that ``ag.convert_to("RDKIT")`` followed by
-    ``ag.convert_to("RDKIT", NoImplicit=False)`` will not use the cache.
-    Finally if you're modifying the AtomGroup in place between two conversions,
-    the id of the AtomGroup won't change and thus the converter will use the
-    cached molecule. For this reason, you can pass a ``cache=False`` argument
-    to the converter to bypass the caching system.
-    Note that the cached molecule doesn't contain the coordinates of the atoms.
+    only remembers the 2 most recent AtomGroups that were converted, as well
+    as the arguments that were passed to the converter. The number of objects
+    cached can be changed with the function :func:`set_converter_cache_size`.
+    However, ``ag.convert_to("RDKIT")`` followed by ``ag.convert_to("RDKIT", NoImplicit=False)``
+    will not use the cache since the arguments given are different.
+    You can pass a ``cache=False`` argument to the converter to bypass the
+    caching system.
 
 
     .. versionadded:: 2.0.0
@@ -265,7 +260,6 @@ class RDKitConverter(base.ConverterBase):
 
     lib = 'RDKIT'
     units = {'time': None, 'length': 'Angstrom'}
-    _cache = dict()
 
     def convert(self, obj, cache=True, NoImplicit=True, max_iter=200):
         """Write selection at current trajectory frame to
@@ -278,16 +272,14 @@ class RDKitConverter(base.ConverterBase):
         cache : bool
             Use a cached copy of the molecule's topology when available. To be
             used, the cached molecule and the new one have to be made from the
-            same AtomGroup object (same id) and with the same arguments passed
-            to the converter (with the exception of this `cache` argument)
+            same AtomGroup selection and with the same arguments passed
+            to the converter
         NoImplicit : bool
             Prevent adding hydrogens to the molecule
         max_iter : int
             Maximum number of iterations to standardize conjugated systems.
             See :func:`_rebuild_conjugated_bonds`
         """
-        # parameters passed to atomgroup_to_mol and used by the cache
-        kwargs = dict(NoImplicit=NoImplicit, max_iter=max_iter)
 
         try:
             from rdkit import Chem
@@ -303,22 +295,13 @@ class RDKitConverter(base.ConverterBase):
                             "please use a valid AtomGroup or Universe".format(
                                 type(obj))) from None
 
+        # parameters passed to atomgroup_to_mol
+        kwargs = dict(NoImplicit=NoImplicit, max_iter=max_iter)
         if cache:
-            # key used to search the cache
-            key = f"<{id(ag):#x}>" + ",".join(f"{key}={value}"
-                                            for key, value in kwargs.items())
-            try:
-                mol = self._cache[key]
-            except KeyError:
-                # only keep the current molecule in cache
-                self._cache.clear()
-                # create the topology
-                self._cache[key] = mol = self.atomgroup_to_mol(ag, **kwargs)
-            # continue on copy of the cached molecule
+            mol = atomgroup_to_mol(ag, **kwargs)
             mol = copy.deepcopy(mol)
         else:
-            self._cache.clear()
-            mol = self.atomgroup_to_mol(ag, **kwargs)
+            mol = atomgroup_to_mol.__wrapped__(ag, **kwargs)
 
         # add a conformer for the current Timestep
         if hasattr(ag, "positions"):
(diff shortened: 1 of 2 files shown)


cbouy:
ping @IAlibay @richardjgowers

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -339,108 +322,124 @@ def convert(self, obj, cache=True, NoImplicit=True, max_iter=200):
...
+        See :func:`_rebuild_conjugated_bonds`
+    """
+    try:
+        elements = ag.elements
+    except NoDataError:
+        raise AttributeError(
+            "The `elements` attribute is required for the RDKitConverter "
+            "but is not present in this AtomGroup. Please refer to the "
+            "documentation to guess elements from other attributes or "
+            "type `help(mda.topology.guessers)`") from None
```
orbeckst, line 347:
```suggestion
            "type `help(MDAnalysis.topology.guessers)`") from None
```

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -339,108 +322,124 @@ def convert(self, obj, cache=True, NoImplicit=True, max_iter=200):
...
+
+    if "H" not in ag.elements:
```
orbeckst, line 349: Maybe this should raise an exception and have an explicit opt-in kwarg (like "force")? Can you think of a case where one would want this? ... How often do we have structures without hydrogens?

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -339,108 +322,124 @@ def convert(self, obj, cache=True, NoImplicit=True, max_iter=200):
...
+    """Set the maximum cache size of the RDKit converter
+
+    Parameters
+    ----------
+    maxsize : int or None
+        If int, the cache will only keep the ``maxsize`` most recent
+        conversions in memory. Using ``maxsize=None`` will remove all limits
+        to the cache size, i.e. everything is cached.
+    """
+    global atomgroup_to_mol
```
orbeckst, line 441: justified use of `global` ;-)

orbeckst: This is probably not thread-safe – not a big deal, though, and I don't have a better idea.

(Although, we don't really encourage use of threads for parallelization; multiprocessing should do just fine.)

orbeckst (commented):
I like the use of the lru cache from the stdlib. Peripheral comments inline.

testsuite/MDAnalysisTests/coordinates/test_rdkit.py
```diff
@@ -331,34 +331,44 @@ def test_nan_coords(self):
             mol.GetConformer()
 
     def test_cache(self):
```
IAlibay, line 333: This test is getting rather convoluted, whilst generally we don't need docstring for tests, I wonder if a quick summary here might help future readers?

i.e. what is being tested, a quick line or two about how you're going about it.

IAlibay (changes requested):
Overall lgtm! Just a few comments, mainly to do with tests & docs.

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -339,108 +322,124 @@ def convert(self, obj, cache=True, NoImplicit=True, max_iter=200):
...
-
-        if NoImplicit:
-            # infer bond orders and formal charges from the connectivity
-            _infer_bo_and_charges(mol)
-            mol = _standardize_patterns(mol, max_iter)
-
-        # sanitize
-        Chem.SanitizeMol(mol)
+            bond_indices = [atom_mapper[i] for i in bond.indices]
+        except KeyError:
```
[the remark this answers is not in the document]
cbouy: How can I test that though ? Its used to skip bonds that are between an atom in the atomgroup and an atom that is not.
We can discuss it in my upcoming PR on the converter anyway

package/MDAnalysis/coordinates/RDKit.py
cbouy: @orbeckst PDB files often don't have them all explicit. There are also the rather rare cases of small molecules with no hydrogen like CO2 and so on...
I'm not too sure if it's desirable to return a molecule and just raise a warning even if the conversion will output something that is most likely 💩

orbeckst: > I'm not too sure if it's desirable to return a molecule and just raise a warning even if the conversion will output something that is most likely 💩

No, I expect that nobody reads warnings.

If something will likely be wrong then better fail.

That should be the guiding principle. If there are special cases then maybe provide a kwarg for power users to override the safety check. It will be a bit annoying when you work with inorganic molecules (or CO2) but presumably these will be a minority and the exception message could state clearly what to do if the user thinks that this should work.

package/MDAnalysis/coordinates/RDKit.py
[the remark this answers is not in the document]
IAlibay: Maybe you could create an atom group made of a single atom that is bonded to other things in the universe?

Whilst pedantic, the issue is that there are no tests currently covering this (or the `AttributeError` in convert) and we are so close to 100% coverage.

testsuite/MDAnalysisTests/coordinates/test_rdkit.py
```diff
@@ -239,6 +239,11 @@ def test_warn_no_hydrogen(self):
             match="No hydrogen atom could be found in the topology"
         ):
             u.atoms.convert_to("RDKIT")
+        rdkit_converter = mda._CONVERTERS["RDKIT"]().convert
```
IAlibay, line 242: Whilst I appreciate concise code, I personally prefer that tests like these be split up for readability. If that's not feasible, then can you at least add a comment to detail that you are now testing a different condition?

testsuite/MDAnalysisTests/coordinates/test_rdkit.py
```diff
@@ -355,17 +371,17 @@ def test_cache(self):
         assert cache.misses == 3
         mol = ucc.atoms.convert_to("RDKIT")
         assert cached_func.cache_info().hits == 1
```
IAlibay, line 373: Is it worth also reloading `uc` and checking that the cache hits doesn't increase? (ignore if overly pedantic)

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -339,108 +322,125 @@ def convert(self, obj, cache=True, NoImplicit=True, max_iter=200):
...
+        elements = ag.elements
+    except NoDataError:
+        raise AttributeError(
+            "The `elements` attribute is required for the RDKitConverter "
+            "but is not present in this AtomGroup. Please refer to the "
+            "documentation to guess elements from other attributes or "
+            "type `help(MDAnalysis.topology.guessers)`") from None
+
+    if NoImplicit is True and "H" not in ag.elements:
+        warnings.warn(
```
IAlibay, line 351: @orbeckst is probably right here that we should raise an error instead.

I am a little bit concerned that say an alkyne might be an issue if the only way is to enforce implicit hydrogens. Thoughts? As @orbeckst suggested, should there be a `force` parameter?

IAlibay (changes requested):
Thanks @cbouy, couple of near-final comments with very minimal changes (one of which probably can be just ignored). The main discussion point remains this implicit hydrogens thing.

cbouy:
I updated my first post with the new changes.
The Azure fails seem to be related to H5MD, I'll let you investigate.
Also coverage is not 100% even with the new tests 😭

cbouy:
All tests are passing 💃 anything else ?

IAlibay:
> All tests are passing 💃 anything else ?

Apologies for taking so long here, I'll re-review over the weekend but I _think_ we should be good.

IAlibay:
@cboy if you want to update this against the current develop, it'll finally be on my list for the next thing I review.

cbouy:
Okay I think I finally managed to run a proper `git rebase upstream/develop`  this time 💁‍♂️

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -303,22 +299,13 @@ def convert(self, obj, cache=True, NoImplicit=True, max_iter=200):
...
-                self._cache.clear()
-                # create the topology
-                self._cache[key] = mol = self.atomgroup_to_mol(ag, **kwargs)
-            # continue on copy of the cached molecule
+            mol = atomgroup_to_mol(ag, **kwargs)
             mol = copy.deepcopy(mol)
         else:
-            self._cache.clear()
-            mol = self.atomgroup_to_mol(ag, **kwargs)
+            mol = atomgroup_to_mol.__wrapped__(ag, **kwargs)
```
IAlibay, line 308: Just thinking about what users might be wondering here. Is there a specific user case where users might not want to use the cache?

testsuite/MDAnalysisTests/coordinates/test_rdkit.py
```diff
@@ -167,11 +167,21 @@ def peptide(self):
         u.add_TopologyAttr('elements', elements)
         return u.select_atoms("resid 2-12")
 
+    @pytest.fixture
+    def uo2(self):
+        return mda.Universe.from_smiles("O=O")
+
+    def test_no_atoms_attr(self):
+        rdkit_converter = mda._CONVERTERS["RDKIT"]().convert
```
IAlibay, line 175: This pops up a lot in these tests, would it be worth making this a fixture?

testsuite/MDAnalysisTests/coordinates/test_rdkit.py
```diff
@@ -167,11 +167,21 @@ def peptide(self):
...
+        with pytest.raises(TypeError, match=""):
+            rdkit_converter("foo")
+
     @pytest.mark.parametrize("smi", ["[H]", "C", "O", "[He]"])
     def test_single_atom_mol(self, smi):
         u = mda.Universe.from_smiles(smi, addHs=False,
                                      generate_coordinates=False)
-        mol = u.atoms.convert_to("RDKIT")
```
IAlibay, line 174: Sorry, I think I'm just being silly and forgetting a very obvious thing. Could you remind me why these are all being switched away from `convert_to`?

testsuite/MDAnalysisTests/coordinates/test_rdkit.py
```diff
@@ -331,34 +360,58 @@ def test_nan_coords(self):
             mol.GetConformer()
 
     def test_cache(self):
-        u = mda.Universe.from_smiles("CCO", numConfs=5)
-        ag = u.atoms
-        cache = mda.coordinates.RDKit.RDKitConverter._cache
+        """5 tests are performed here:
```
IAlibay, line 363: 👍🏽

IAlibay (approved):
Sorry about the massive delay here @cbouy, overall lgtm!

Just a couple of small questions.

@orbeckst do you want to re-review or are we good to merge?

testsuite/MDAnalysisTests/coordinates/test_rdkit.py
cbouy: `convert_to` doesn't pass arguments to the underlying converter, it was in a PR at some point though (#2882 )

IAlibay: Doesn't this behaviour contradict the docstring? I.e. ":func:`set_converter_cache_size`. However, ``ag.convert_to("RDKIT")``
    followed by ``ag.convert_to("RDKIT", NoImplicit=False)`` will not use the"

Or was the argument that we would merge #2882 before this PR?

cbouy: The point is the converter modules weren't really documented to be instantiated like `c = mda.coordinates.RDKit.RDKitConverter(); c.convert(...)` but usually go through the `convert_to` AtomGroup method.
So yeah I assumed 2882 would be merged before v2.0 comes out

IAlibay: alright, let's see if we can revive #2882 then

orbeckst:
@IAlibay sorry, don't have time today to review — I'll leave it to you.

IAlibay (changes requested):
I don't know why I completely forgot 🙀 We need a changelog (and can you also add in an entry for #2926?).

package/MDAnalysis/coordinates/RDKit.py
cbouy: With the previous implementation being buggy, yes it made sense :') but now I can't really think of a good reason, at least for users

cbouy:
I'm adding the missing changelog now. For the changelog of this PR though, do I mention the fixes/changes I made, or just the enhancements (`set_converter_cache_size(maxsize)` and the `force` parameter) ? The RDKit converter isn't released yet so it's a bit weird fixing something that isn't officially out...

orbeckst:
If there was an issue for the fixes then I’d still add a CHANGELOG entry even though it’s a bit weird. However,for people living in the edge (using develop) it’s still helpful.

PR #2950, opened 2020-09-19
Limit numpy thread usage for `Transformation` classes

### LINKED ISSUE #2996: PositionAverager Transformation ends up with wrong results with parallel analysis
## Expected behavior ##

The results are the same as serial analysis

## Actual behavior ##

Due to the splitting approach, a new `PositionAverager` will be created for each block; no previous memory (`self.coord_array`) is saved.

``` python
import MDAnalysis as mda
from MDAnalysisData.adk_equilibrium import fetch_adk_equilibrium
adk = fetch_adk_equilibrium()

from MDAnalysis.transformations.positionaveraging import PositionAverager

from MDAnalysis.analysis.rms import RMSD as serial_RMSD
from pmda.rms import RMSD as parallel_RMSD

u = mda.Universe(adk.topology, adk.trajectory)

pos_avg_trans = PositionAverager(1000)
u.trajectory.add_transformations(pos_avg_trans)

rmsd = serial_RMSD(u.atoms, u.atoms).run()
rmsd = parallel_RMSD(u.atoms, u.atoms).run(n_blocks=8)

# plot rmsd
```
- Serial RMSD
serial RMSD
- Parallel RMSD
parallel RMSD

- Which operating system? Ubuntu 20

yuxuanzhuang:
Fixes #2996  and https://github.com/MDAnalysis/pmda/issues/144

 - add a `TransformationBase` class for handling thread limiting.
- `TransformationBase` has a `parallelizable` to check if it can be used in the parallel analysis (split-apply-combine approach)

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


orbeckst (changes requested):
This is an interesting addition. I might be commenting prematurely but here are initial comments
- I assume that under normal circumstances, multithreading is beneficial, i.e., whenever someone just runs MDA on a multicore machine without thinking about parallelization. Under these conditions we would not want to limit the threads, would we? Is there a sensible way by which we can make the thread limiting optional?
- tests
- docs
- changes

Did you  check the performance impact of limiting threads?

.travis.yml
```diff
@@ -30,7 +30,7 @@ env:
     - SETUP_CMD="${PYTEST_FLAGS}"
     - BUILD_CMD="pip install -e package/ && (cd testsuite/ && python setup.py build)"
     - CONDA_MIN_DEPENDENCIES="mmtf-python biopython networkx cython matplotlib scipy griddataformats hypothesis gsd codecov"
-    - CONDA_DEPENDENCIES="${CONDA_MIN_DEPENDENCIES} seaborn>=0.7.0 clustalw=2.1 netcdf4 scikit-learn joblib>=0.12 chemfiles tqdm>=4.43.0 tidynamics>=1.0.0 rdkit>=2020.03.1 h5py"
+    - CONDA_DEPENDENCIES="${CONDA_MIN_DEPENDENCIES} seaborn>=0.7.0 clustalw=2.1 netcdf4 scikit-learn joblib>=0.12 chemfiles tqdm>=4.43.0 tidynamics>=1.0.0 rdkit>=2020.03.1 h5py threadpoolctl"
```
IAlibay, line 33: threadpoolctl isn't a core dependency, so as it is, it will fail when running on minimal dependencies.

IAlibay (changes requested):
echoing @orbeckst's comment.

I realise this is linked to https://github.com/MDAnalysis/pmda/issues/144, but it might be good to have an issue on MDAnalysis itself. In part because it would be to go through all the use cases & see where benchmarks vary between users (I wouldn't be surprised if you end up getting a lot of variance on whether or not enabling multithreading speeds/slows things down for a given use case).

yuxuanzhuang:
I am afraid that with the decorator (the current implementation), we cannot change the thread limit at runtime. (e.g. passing a `self.n_thread` arg). However, we can do it inside `__init__` by `self.__call__ = threadpool_limits_decorator(self.n_thread)(self.__call__)`, which I think is not that clear/pretty as with the decorator.

As for the impact on performance, I can reproduce the results (https://github.com/MDAnalysis/pmda/issues/144) with another 6-core CPU. Sorry for the rough data, but you can see it's mainly the hyperthreading that is limiting the performance.
image
Note this is purely the serial code (no PMDA involved)

I will create an MDA issue when I have time.

orbeckst:
The decorator is pretty but if it does not provide enough flexibility, just use the context manager:
```python
class RotateTransformation:
   def __init__(self, rotmat, max_threads=None):
       self.rotmat = rotmat
       self.n_thread = max_threads  # possibly needs some logic here to "do the best thing"
  
   def __call__(self, ts):
        with threadpool_limits(self.n_thread):
              ts.positions[:] = np.dot(ts.positions, self.rotmat)
        return ts
```
That still looks reasonable to me.

Or is there a deeper reason for preferring the function decorator?

The bigger question is what the default for `max_threads` ought to be and if all transformations need to have it. I think the common use case is serial on multicore so it should probably default to whatever the thread-parallel code wants to do (even though with hyper-threading this might be sub-optimal). We should then give guidance on how to tune the settings.

yuxuanzhuang:
Sorry for the radio silence...was pretty busy last week. I think we can port this function into PMDA if we decide serial code just goes into its default setting. 

And in PMDA,  say if the user is requiring all the cores, then we limit the `self._single_frame` to one thread so that everything inside, `Transformation` included, will be run with one thread. This is also the reason I prefer to use a decorator.

orbeckst:
It looks like a worthwhile tuning knob to include in the MDAnalysis transformations. Then PMDA (and anyone else) can just use the optional `max_threads` as needed.

https://github.com/joblib/threadpoolctl/blob/32037cf43a61909282b5d07e6b21d9621fa03e25/threadpoolctl.py#L154 says that `threadpool_limits(limits=None)` does nothing so we can just use `None` as a default for `max_threads=None`.

Including `threadpoolctl` as dependency should be easy (conda-forge or pip).

orbeckst:
> I think we can port this function into PMDA if we decide serial code just goes into its default setting.

Do you mean then we don't have threadctl in MDA? Is this because then the user would have to limit the threads when they create the transformations, i.e., they need to know that they will use PMDA? 

> And in PMDA, say if the user is requiring all the cores, then we limit the self._single_frame to one thread so that everything inside, Transformation included, will be run with one thread. 

I can see that it's easier to just apply the limits to everything in PMDA.

I don't think that it will be easy to figure out if there's any oversubscription of the CPU, at least not when someone is using distributed. For multiprocessing you could try to find out how many cores are available but this starts getting complicated, I feel.

> This is also the reason I prefer to use a decorator.

Can you explain more? You mean so that you can decorate `single_frame()`?? Maybe an example would help me to understand.

orbeckst:
@yuxuanzhuang in your performance comparison https://github.com/MDAnalysis/mdanalysis/pull/2950#issuecomment-699856608, what is the default when you don't limit any threads? That number is missing from the comparison. _If_ the default runs 12 threads (or enough to always reduce performance) then I better understand your point that you want to limit to 1 thread by default.

Did you check with `threadpool_info()` what the default settings were, i.e., how many threads `np.dot` uses by default?

yuxuanzhuang:
It turns out there's an issue with this **decoration** approach that doesn't seem to be easily solvable---the thread limit is applied globally, instead of only the decorated function. Here is an example. https://gist.github.com/yuxuanzhuang/05b0da16a51f567e54f7f3f22591e316

yuxuanzhuang:
> @yuxuanzhuang in your performance comparison [#2950 (comment)](https://github.com/MDAnalysis/mdanalysis/pull/2950#issuecomment-699856608), what is the default when you don't limit any threads? That number is missing from the comparison. _If_ the default runs 12 threads (or enough to always reduce performance) then I better understand your point that you want to limit to 1 thread by default.
> 
> Did you check with `threadpool_info()` what the default settings were, i.e., how many threads `np.dot` uses by default?

The default is 12 threads (or whatever that computer has including hyper threads). I think for `numpy`, the threshold of asking for all the threads is quite low. So `Transformation` is always performing pretty bad in default.

IAlibay:
@yuxuanzhuang apologies, I think I'm missing part of the conversation here 😅 From your test, it seems that the context manager method might work here right? Would that be suitable or is there a barrier to using it?

yuxuanzhuang:
> @yuxuanzhuang apologies, I think I'm missing part of the conversation here 😅 From your test, it seems that the context manager method might work here right? Would that be suitable or is there a barrier to using it?

You are right, the context manager method should work. I was thinking that with that you have to add the code snippet everywhere needed.

IAlibay:
> You are right, the context manager method should work. I was thinking that with that you have to add the code snippet everywhere needed.

I'll admit I'm not super versed on the transformation code, but it seems like pretty much everything is a class with both an `__init__` and a `__call__`. Is there any way we could create a base class for all these so that we only have to implement one `__call__` that has the context manager, and then calls some other function (let's say `_run` or something)?

jbarnoud:
The transformations are any callable that takes and returns a timestep. In practice, callables are not always pickable, hence the use of classes. Since all the transformations we will ship will be such classes, they may as well pack some features and wrap the transformation logic with the core count one.

orbeckst:
Given that OpenMP (and therefore the BLAS implementations in numpy) default to the maximum number of threads and therefore hurt MDAnalysis performance I would say we 
- add threadpoolctl as a dependency
- make our transformation classes accept a `max_threads` kwarg in the constructor
- use the `threadpoolctl.threadpool_limits` context manager to limit our transformations 
- change the transformations base class  to have a `_transform()` method (similar to `_single_frame()` in AnalysisBase) that is called inside the context manager in the `__call__()` method

By the way, `np.dot()` shows up pretty frequently in the code (encore, diffusionmap, gnm, helix_analysis, rms, pca, atoms.transform). Is it always the case that the perform could be improved by limiting the thread number?

yuxuanzhuang:
It turns out the speed limiting step is not `np.dot` (which is actually **faster** with multi-threads).
Due to the previously mentioned globally-applied thread limiting, the performance is actually enhanced by other parts of the code. (Sorry for the confusion, should do the line profiler earlier).
What confused me now is that such acceleration only happens with `Transformation` present. (https://gist.github.com/yuxuanzhuang/c14f1889cd49c42fd661a860bf7e92d7) I will look into that later, but if you have any insight, please comment!

orbeckst:
So if I read your gist correctly then the profiling shows that `np.dot()` is not a problem and even with 12 threads runs marginally faster:

| class         	| function                         	| n=1    	| n=12   	|
|---------------	|----------------------------------	|--------	|--------	|
| _single_frame 	| CalcRMSDRotationalMatrix()       	| 295.9  	| 515.7  	|
|               	| mobile_atoms.center()            	| 2860.4 	| 5498.6 	|
| fit_rot_trans 	| align.rotation_matrix()          	| 1367.7 	| 1790.8 	|
|               	| mobile.atoms.center()            	| 2843.4 	| 5525.8 	|
|               	| np.dot(ts.positions, rotation.T) 	| 1030.1 	| 944.9  	|

However, the big problem is the **center of mass calculation** in `atoms.center()`. This is actually slower than the rotation matrix calculation and it gets much worse for higher thread numbers.

Is this run on 12 real cores or are you using hyperthreading?

orbeckst:
Judging from @yuxuanzhuang 's benchmarks, oversubscribing cores for OpenMP hurts performance in unexpected ways, even for code where there's no obvious parallelization going on (`AtomGroup.center()` takes twice the amount of time in his example!).

`np.dot()` shows some modest improvements in performance, as long as OpenMP does not oversubscribe cores.

orbeckst:
It is also odd that oversubscription appears to affect the performance of the actual read step in the XDRReader, see https://gist.github.com/yuxuanzhuang/4918cd1b5d8d62de79eab9df40de4bb7#gistcomment-3488210.

yuxuanzhuang:
Related gist:
- Thread effect on MDAnalysis RMSD analysis: https://gist.github.com/yuxuanzhuang/17ed6def63b08248db59f2c44e3e0419
- Thread effect on MDAnalysis position visiting: https://gist.github.com/yuxuanzhuang/4918cd1b5d8d62de79eab9df40de4bb7
- Thread effect on numpy (code after `np.dot`): https://gist.github.com/yuxuanzhuang/82e1e7b57d0cda80ac964d1cd138f618

The tests were down with

- 6-core CPU
- numpy 1.19.1
- python 3.8
```
threadpool_info()
[{'filepath': '/home/scottzhuang/anaconda3/envs/gsoc/lib/python3.8/site-packages/numpy.libs/libopenblasp-r0-ae94cfde.3.9.dev.so',
  'prefix': 'libopenblas',
  'user_api': 'blas',
  'internal_api': 'openblas',
  'version': '0.3.9.dev',
  'num_threads': 12,
  'threading_layer': 'pthreads'},
 {'filepath': '/home/scottzhuang/anaconda3/envs/gsoc/lib/libgomp.so.1.0.0',
  'prefix': 'libgomp',
  'user_api': 'openmp',
  'internal_api': 'openmp',
  'version': None,
  'num_threads': 12},
 {'filepath': '/home/scottzhuang/anaconda3/envs/gsoc/lib/libmkl_rt.so',
  'prefix': 'libmkl_rt',
  'user_api': 'blas',
  'internal_api': 'mkl',
  'version': '2020.0.1',
  'num_threads': 6,
  'threading_layer': 'intel'},
 {'filepath': '/home/scottzhuang/anaconda3/envs/gsoc/lib/libiomp5.so',
  'prefix': 'libiomp',
  'user_api': 'openmp',
  'internal_api': 'openmp',
  'version': None,
  'num_threads': 12}]
```

richardjgowers:
This fix looks good to me given the problem

yuxuanzhuang:
Still WIP
Update:
- finally, the new `threadpool_limits_decorator` can be used locally as a decorator.
   - tests TODO
- create a `TransformationBase` for `Transformations`.

orbeckst (commented):
Overall it looks neat and tidy, at least from the MDA-native side of things. 

My question/comment is on the documentation side of things:
- The `max_threads` kwarg needs to be documented everywhere 
- The docs on how to create transformations need to be updated (now that we would have `_transform()` instead of `__call__()`). The API needs to be properly described.
- Do we want to prescribe that _all_ users need to follow the TransformBase API? Or will we have the user-facing examples still show a normal class with `__call__()` (or even closures)?

Going to TransformBase sacrifices some of the original elegance of the approach (although that already suffered when we moved to classes instead of closures) but it shows the advantage of using classes by having all kinds of goodies built in... 

It would be good to hear from different people what they think, e.g., @jbarnoud @lilyminium .

package/MDAnalysis/transformations/base.py
```diff
@@ -0,0 +1,42 @@
...
+=================================================================
+
+.. autoclass:: TransformationBase
+
+"""
+from threadpoolctl import threadpool_limits
+
+
+class TransformationBase(object):
+    def __init__(self, max_threads):
```
lilyminium, line 34: Could you give this a default argument, e.g. ``max_threads=None``, or even make it a default class attribute? My concern is that users who don't understand or care about parallelisation will try to make custom transformations and override ``__init__`` or forget to pass it an argument, with predictable errors upon initialisation or ``__call__``ing.

lilyminium:
> Do we want to prescribe that all users need to follow the TransformBase API? Or will we have the user-facing examples still show a normal class with `__call__()` (or even closures)?

IMO simpler is better, and it would be clearer to have a structured and foolproof "follow this method" approach front-and-center with the TransformationBase API. That being said I think we already have some documentation (e.g. current user guide, blog post) with the classes and closures so we could keep them on as additional extended examples.

package/MDAnalysis/transformations/base.py
jbarnoud: I agree with @lilyminium. Most users will not care about the number of threads used so there should be a default value that 1) let users focus on what they care about and 2) keep the existing code working.

IAlibay (changes requested):
I agree with @lilyminium. Personally I really like the idea of the TransformationBase API.

Couple of comments, also would it be worth adding a couple tests that do things with and without the default thread limit? This would be to check that we get the same answer, not that we are actually using more/less threads (looking at the threadpoolctl tests, the latter seems like a pain).

Also CHANGELOG will need updating.

IAlibay:
Also the appveyor config file will probably need to have threadpoolctl added to it (see current test failures).

orbeckst:
Summarizing comments so far:
- formally document a **Transformations API** (or **On-the-fly Transformations API** to be more precise?), based on a `TransformationsBase` class and following the AnalysisBase API.
- build in threadpoool limitations
- default to 1 thread (without user intervention)
- enable users to increase the threadlimit (`max_threads`) as a kwarg for the transformation

(Please correct me if I didn't get everything.)

jbarnoud:
I am overall happy with @orbeckst summary. I just have a question about the default of `max_threads`: shall it be ` or whatever is the default if we do not touch it?

The benchmark has been done on one machine with one architecture and one bios/os setting, we do not know if it still holds true on something else, nor if it will still hold true in the future (in which case, would we notice the change?)

orbeckst:
Regarding the default: Given that in my tests with `threadpoolctl_info()` numpy/libopenblas (Linux, macOS) and numpy/libomp (macOS) always default to 2 x cores threads (i.e., with hyperthreads), I would set `max_threads=1`.

If one adds an ASV benchmark with different threads (up to 4; the benchmarking machine only has 4 real cores [Intel(R) Xeon(R) CPU E3-1226 v3 @ 3.30GHz](https://ark.intel.com/content/www/us/en/ark/products/80917/intel-xeon-processor-e3-1226-v3-8m-cache-3-30-ghz.html) and no hyperthreading) then we can see if it makes a difference.

lilyminium:
> Regarding the default: Given that in my tests with `threadpoolctl_info()` numpy/libopenblas (Linux, macOS) and numpy/libomp (macOS) always default to 2 x cores threads (i.e., with hyperthreads), I would set `max_threads=1`.

The default in `threadpoolctl` is `None`.

> - If None, this function does not do anything.

https://github.com/joblib/threadpoolctl/blob/32037cf43a61909282b5d07e6b21d9621fa03e25/threadpoolctl.py#L167

orbeckst:
`None` is not a good default for our purpose — WIth `None`, no thread limitation is enforced and  the default number of threads are used, which (in my cases) is N(hardware threads (cores)) + N(hyper threads).

jbarnoud:
Does `threadpoolctl` respects `OMP_NUM_THREADS`? If so, I'd say that `None` is the correct default as it is the one that respects the user's configuration. The OS default may be wrong but overwriting this behaviour breaks the user expectation.

yuxuanzhuang:
> Regarding the default: Given that in my tests with `threadpoolctl_info()` numpy/libopenblas (Linux, macOS) and numpy/libomp (macOS) always default to 2 x cores threads (i.e., with hyperthreads), I would set `max_threads=1`.
> 
> If one adds an ASV benchmark with different threads (up to 4; the benchmarking machine only has 4 real cores [Intel(R) Xeon(R) CPU E3-1226 v3 @ 3.30GHz](https://ark.intel.com/content/www/us/en/ark/products/80917/intel-xeon-processor-e3-1226-v3-8m-cache-3-30-ghz.html) and no hyperthreading) then we can see if it makes a difference.

I have done the benchmark both for a 6-core Intel chip and a 32-core AMD server CPU, confirming hyperthreading makes performance worse. I will look into the ASV benchmark when I have time. FYI, I think the multi-thread version of `np.dot` is only raised up with more than ~20k atoms.

yuxuanzhuang:
> Does `threadpoolctl` respects `OMP_NUM_THREADS`? If so, I'd say that `None` is the correct default as it is the one that respects the user's configuration. The OS default may be wrong but overwriting this behaviour breaks the user expectation.

`threadpoolctl` will overrule the `OMP_NUM_THREADS`...ofc only in its controlled region. We can set the limit to just `blas` by `with threadpool_limits(limits=1, user_api='blas'):` if we want to always respect the user's OMP expectation.

package/MDAnalysis/lib/util.py
```diff
@@ -2353,3 +2355,24 @@ def check_box(box):
     if np.all(box[3:] == 90.):
         return 'ortho', box[:3]
     return 'tri_vecs', triclinic_vectors(box)
+
+
+class threadpool_limits_decorator(threadpool_limits, ContextDecorator):
```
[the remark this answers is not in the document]
yuxuanzhuang: No, not really. It turns out to be not that useful in this case.

jbarnoud:
If I put my sysadmin hat on, my preference would be: respect the environment configuration by default (even if it is not optimal), locally overright the value when relevant (for instance because the parallellisation is done differently). I will not die on that hill, though: I did not even realize we were using much parallel code.

orbeckst:
I understand your view to do not interfere with default settings. Unfortunately, in this case MDAnalysis gets blamed for what looks to me like a problematic handling of defaults somewhere upstream. It seems really unlikely that anyone will complain if we restrict threads in transformations. Users can still set `max_threads=None` to be subjected to `OMP_NUM_THREADS`. We will document the behavior and the power users that require it will hopefully pick it up. The other 98% of users might get a 2x speed-up on a function that really needs to be faster to be useable.

However, I will concede that we should make the default *in the TransformationBase* base class `max_threads=None` so that any user-defined function works exactly as the author intended.

So, in conclusion, I am in favor of defaulting to `max_threads=1` for functions where we know that it will improve performance (namely the RMSD alignments ones) but leaving the default `None` in the base class, together with ample documentation and comments.

yuxuanzhuang:
TODO:
- Update the other part of the `Transformation` document
- any other tests need to be done?
- benchmark

testsuite/MDAnalysisTests/transformations/test_base.py
```diff
@@ -0,0 +1,92 @@
...
+from MDAnalysis.transformations.base import TransformationBase
+
+
+class DefaultTransformation(TransformationBase):
+    """Default values for max_threads and parallelizable"""
+    def __init__(self):
+        super().__init__()
+
+    def _transform(self, ts):
+        self.runtime_info = threadpool_info
```
orbeckst, line 39: shouldn't this be a call to `threadpool_info()`?

package/MDAnalysis/transformations/base.py
```diff
@@ -31,12 +31,84 @@
...
+    Transformation classes. This class will
+    1) set up a context manager framework on limiting the threads
+       per call, which may be the multi-thread OpenBlas backend of NumPy.
+       This backend may kill the performance when subscribing hyperthread
+       or oversubscribing the threads when used together with other parallel
+       engines e.g. Dask. (PR #2950)
+       Define `max_threads=1` when that is the case.
+    2) set up a boolean attribute `parallelizable` for checking if the
+       transformation can be applied in a **split-apply-combine** parallelism.
+       For example, the `MDAnalysis.transformations.PositionAverager`
```
orbeckst, line 46: make it a link with `:class:` role

package/MDAnalysis/transformations/base.py
```diff
@@ -31,12 +31,84 @@
...
+       is history-dependent and can not be used in parallel analysis natively.
+       (Issue #2996)
+
+    To define a new Transformation, `TransformationBase` has to be subclassed.
```
orbeckst, line 50: add reST role

orbeckst (changes requested):
This looks all very good, I only have doc comments

- see inline
- have a section at the top of transformations that explains the issues with parallelization and threads. Say that we use threadpoolcontrol and link to it. State that users can set the external environment variable OMP_MAX_THREADS (or whatever it was...) to globally limit thread usage. Show an example for sh-like shells `export OMP_MAX_THREADS=2` or similar. Users are advised to _benchmark_ code because interaction between different libraries can lead to sub-optimal performance with defaults.

yuxuanzhuang:
Thanks! I will wait for the h5py fix merge to see if all CI pass.

IAlibay:
> Thanks! I will wait for the h5py fix merge to see if all CI pass.

@yuxuanzhuang should be fixed now, do you want to try updating your branch and see how it fares?

IAlibay:
@orbeckst @yuxuanzhuang what's the status on this?

It looks like it was near completion, so I'm going to put this as a 2.0 milestone.

yuxuanzhuang:
Just merged to the developer branch. Looks fine from my side.

.appveyor.yml
```diff
@@ -0,0 +1,73 @@
+# config based on examples from SciPy & NumPy repositories, which
```
IAlibay, line 1: This needs to go (does gh actions need to be updated somehow?)

.travis.yml
```diff
@@ -45,7 +45,7 @@ matrix:
       if: type = cron
       before_install:
         - python -m pip install cython numpy scipy
-        - python -m pip install --no-build-isolation hypothesis matplotlib pytest pytest-cov pytest-xdist tqdm
+        - python -m pip install --no-build-isolation hypothesis matplotlib pytest pytest-cov pytest-xdist tqdm threadpoolctl
```
IAlibay, line 48: I've lost track of why this was being added. Do we need to also add this to ppc64le?

package/MDAnalysis/transformations/base.py
```diff
@@ -0,0 +1,118 @@
...
+#
+
+"""\
+Transformations Base Class --- :mod:`MDAnalysis.transformations.base`
+=====================================================================
+
+.. autoclass:: TransformationBase
+
+"""
+from threadpoolctl import threadpool_limits
```
IAlibay, line 30: Does this need to be in setup.py?

package/MDAnalysis/transformations/wrap.py
```diff
@@ -36,8 +36,10 @@
 
 from ..lib._cutil import make_whole
 
+from .base import TransformationBase
```
IAlibay, line 39: Does the new dimensions transformation need to be updated too?

IAlibay (commented):
Mostly catching up here. Just added some questions - more on the deployment side of things.

.travis.yml
orbeckst: threadpoolctl is essential for the PR; we might have to promote it to "required package" because the transformation code will fail without it (I think)

package/MDAnalysis/transformations/base.py
orbeckst: Yes, so that `pip install` installs it. Good catch!

.appveyor.yml
orbeckst, line 1: remove file (not sure why it is still here)

package/MDAnalysis/transformations/base.py
```diff
@@ -0,0 +1,118 @@
+# -*- Mode: python; tab-width: 4; indent-tabs-mode:nil; coding:utf-8 -*-
```
orbeckst, line 1: make sure that this page shows up in docs

package/MDAnalysis/transformations/base.py
```diff
@@ -0,0 +1,118 @@
...
+    """Base class for defining on-the-fly transformations
+
+    The class is designed as a template for creating on-the-fly
+    Transformation classes. This class will
+
+    1) set up a context manager framework on limiting the threads
+    per call, which may be the multi-thread OpenBlas backend of NumPy.
+    This backend may kill the performance when subscribing hyperthread
+    or oversubscribing the threads when used together with other parallel
+    engines e.g. Dask. (PR #2950)
```
orbeckst, line 43: make PR #2950 a link

package/MDAnalysis/transformations/base.py
```diff
@@ -0,0 +1,118 @@
...
+
+    Define `max_threads=1` when that is the case.
```
orbeckst, line 45: Need double back-ticks in reST

package/MDAnalysis/transformations/base.py
```diff
@@ -0,0 +1,118 @@
...
+
+    2) set up a boolean attribute `parallelizable` for checking if the
+    transformation can be applied in a **split-apply-combine** parallelism.
+    For example, the :class:`MDAnalysis.transformations.PositionAverager`
+    is history-dependent and can not be used in parallel analysis natively.
+    (Issue #2996)
```
orbeckst, line 51: make issue a link

package/MDAnalysis/transformations/base.py
```diff
@@ -0,0 +1,118 @@
...
+
+    To define a new Transformation, :class:`TransformationBase`
+    has to be subclassed.
+    `max_threads` will be set to `None` in default, i.e. does not do anything.
```
orbeckst, line 55:
```reST
``None``
```
double backticks

package/MDAnalysis/transformations/base.py
```diff
@@ -0,0 +1,118 @@
...
+    `parallelizable` will be set to `True` in default. You may need to double
```
orbeckst, line 56: ``True``

package/MDAnalysis/transformations/base.py
```diff
@@ -0,0 +1,118 @@
...
+    check if it can be used in parallel analysis; if not, override the value
+    to `False`. Note this attribute is not checked anywhere in MDAnalysis yet.
```
orbeckst, line 58: ``False``

package/MDAnalysis/transformations/base.py
orbeckst:
```reST
... does not do anything and any settings in the environment such as the environment variable :envvar:`OMP_NUM_THREADS`  (see the `OpenMP specification for OMP_NUM_THREADS<https://www.openmp.org/spec-html/5.0/openmpse50.html>`_) are used.
```

package/MDAnalysis/transformations/base.py
```diff
@@ -0,0 +1,118 @@
...
+       check if it can be used in parallel analysis.
+    """
+
+    def __init__(self, **kwargs):
+        """
+        Parameters
+        ----------
+        max_threads: int, optional
+           The maximum thread number can be used.
+           Default is `None`, which means the default or the external setting.
```
orbeckst, line 95: ``None``

package/MDAnalysis/transformations/base.py
```diff
@@ -0,0 +1,118 @@
...
+        parallelizable: bool, optional
+           A check for if this can be used in split-apply-combine parallel
+           analysis approach.
+           Default is `True`.
```
orbeckst, line 99: double backticks ``True``

package/doc/sphinx/source/documentation_pages/trajectory_transformations.rst
```diff
@@ -113,17 +113,22 @@ the following two methods can be used to create such transformation:
 Creating complex transformation classes
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ 
 
-It is implemented by defining :func:`__call__` for the transformation class
-and can be applied directly to a :class:`Timestep`.
+It is implemented by inheriting :class:`MDAnalysis.transformations.base.TransformationBase`,
```
orbeckst, line 116: inheriting from

package/doc/sphinx/source/documentation_pages/trajectory_transformations.rst
```diff
@@ -185,7 +190,12 @@ any of these transformations, the module must first be imported:
 
    import MDAnalysis.transformations
 
-A workflow can then be added to a trajectory as described above.
+A workflow can then be added to a trajectory as described above. Notably,
+the parameter `max_threads` can be defined when creating a transformation
+instance to limit the maximum threads.
+(See :class:`MDAnalysis.transformations.base.TransformationBase` for more details) 
+Whether a specific transformation can be used along with parallel analysis
+can be accessed by checking its `parallelizable` attribute.
```
orbeckst, line 198:
```suggestion
can be assessed by checking its :attr:`~MDAnalysis.transformations.base.TransformationBase.parallelizable` attribute.
```

package/doc/sphinx/source/documentation_pages/trajectory_transformations.rst
```diff
@@ -221,6 +231,31 @@ Giving a workflow as a keyword argument when defining the universe:
...
+Building blocks for Transformation Classes
+------------------------------------------
+Transformations normally ultilize the power of NumPy to get better performance
+on array operations. However, when it comes to parallelism, NumPy will sometimes
+oversubscribe the threads, either by hyper threading (when it uses OpenBlas backend),
+or by working with other parallel engines (e.g. Dask).
+
+In MDAnalysis, we use `threadpoolctl <https://github.com/joblib/threadpoolctl>`_
+inside :class:`TransformationBase` to control the maximum threads for
+transformations. (See :class:`MDAnalysis.transformations.base.TransformationBase` for more details)
```
orbeckst, line 245:
```suggestion
inside :class:`~MDAnalysis.transformations.base.TransformationBase` to control the maximum threads for
transformations.
```
(Just use the generated link)

orbeckst (changes requested):
Looks very good, almost there.
- Make sure that threadpoolctl is always installed (`setup.py` and elsewhere... @IAlibay might know about some other files where it needs to be).
- Ensure that the new transformations base page is included in the docs.
- Possibly raise an issue in the User Guide to update the Transformations docs there.

IAlibay:
> * Make sure that threadpoolctl is always installed (`setup.py` and elsewhere... @IAlibay might know about some other files where it needs to be).

As far as I am aware, we'll need it in the gh actions workflow, the travis cron job, azure pipelines, and setup.py.

yuxuanzhuang:
> > * Make sure that threadpoolctl is always installed (`setup.py` and elsewhere... @IAlibay might know about some other files where it needs to be).
> 
> As far as I am aware, we'll need it in the gh actions workflow, the travis cron job, azure pipelines, and setup.py.

Thanks! Let me know if I have missed anywhere.

package/MDAnalysis/transformations/base.py
yuxuanzhuang: Yes, under `8.5. Building blocks for Transformation Classes`.

package/MDAnalysis/transformations/wrap.py
yuxuanzhuang: Updated:)

IAlibay:
> > > * Make sure that threadpoolctl is always installed (`setup.py` and elsewhere... @IAlibay might know about some other files where it needs to be).
> > 
> > 
> > As far as I am aware, we'll need it in the gh actions workflow, the travis cron job, azure pipelines, and setup.py.
> 
> Thanks! Let me know if I have missed anywhere.

@lilyminium do we use the maintainer/conda/environment.yml for anything / should we update it too?

.travis.yml
IAlibay, line 48: @yuxuanzhuang if you can add threadpoolctl to either line 27 or 28 that'd be great too (the ppc64le build has it's own install call).

package/MDAnalysis/transformations/fit.py
```diff
@@ -174,8 +183,20 @@ class fit_rot_trans(object):
...
+    .. versionchanged:: 2.0.0
+       The transformation was changed from a function/closure to a class
+       with ``__call__``.
+    .. versionchanged:: 2.0.0
+       The transformation was changed to inherit from the base class for
+       limiting threads and checking if it can be used in parallel analysis.
     """
-    def __init__(self, ag, reference, plane=None, weights=None):
+    def __init__(self, ag, reference, plane=None, weights=None,
+                 max_threads=1, parallelizable=True):
```
IAlibay, line 196: In some of the transformations `max_threads` is set to 1, whilst in the others it is set to `None`. My understanding is that the ones that are set to 1 are the cases where you've benchmarked the transformation and it looks like setting to 1 is faster. Would it make sense in these cases to add a `Note` to the docstring documenting this? That way next time folks come along and see the differences they can easily see why that design decision was made.

IAlibay (changes requested):
Couple of small typos, a test and a question re: documenting thread limitations.

.travis.yml
orbeckst: @IAlibay  should we also add to CHANGELOG under Changes that threadpoolctl is now required? There's also a wiki page where we list required packages (I think in the developer guide... not sure if this has been superceeded by anything in the User Guide).

package/MDAnalysis/transformations/base.py
```diff
@@ -0,0 +1,124 @@
...
+        """
+        with threadpool_limits(self.max_threads):
+            return self._transform(ts)
+
+    def _transform(self):
+        """Transform the given `Timestep`
+
+        It deals with the transformation of a single `Timestep`.
+        """
+        raise NotImplementedError("Only implemented in child classes")
```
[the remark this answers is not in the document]
orbeckst: For NotImplementedError (and assertions) I would also be ok with a pragma nocover. Explicit test will be better, of course.

orbeckst (approved):
You addressed all my comments but @IAlibay raised important points so please address these.

.travis.yml
IAlibay: > @IAlibay should we also add to CHANGELOG under Changes that threadpoolctl is now required? 
👍🏽 

> There's also a wiki page where we list required packages (I think in the developer guide... not sure if this has been superceeded by anything in the User Guide).

I think we now link to the user guide (see here: https://github.com/MDAnalysis/mdanalysis/wiki/Guide-for-Developers).
We should definitely update the userguide install instructions (we probably need to have a blanket issue for this, since I think there are a few things that need ammending for 2.0? Tagging @lilyminium here.)

lilyminium: Made https://github.com/MDAnalysis/UserGuide/issues/139 and https://github.com/MDAnalysis/UserGuide/issues/138.

orbeckst:
There's one RDKit test failing https://github.com/MDAnalysis/mdanalysis/pull/2950/checks?check_run_id=2297717464
```
_____________ TestRDKitConverter.test_mol_from_selection[ILE-38-1] _____________
[gw0] linux -- Python 3.8.8 /usr/share/miniconda/envs/test/bin/python

self = <MDAnalysisTests.coordinates.test_rdkit.TestRDKitConverter object at 0x7f20c185ec70>
peptide = <AtomGroup with 155 atoms>, resname = 'ILE', n_atoms = 38
n_fragments = 1

    @pytest.mark.parametrize("resname, n_atoms, n_fragments", [
        ("PRO", 14, 1),
        ("ILE", 38, 1),
        ("ALA", 20, 2),
        ("GLY", 21, 3),
    ])
    def test_mol_from_selection(self, peptide, resname, n_atoms, n_fragments):
        mol = peptide.select_atoms("resname %s" % resname).convert_to("RDKIT")
>       assert n_atoms == mol.GetNumAtoms()
E       assert 38 == 1
E        +  where 1 = <bound method GetNumAtoms of <rdkit.Chem.rdchem.Mol object at 0x7f20b9a6a280>>()
E        +    where <bound method GetNumAtoms of <rdkit.Chem.rdchem.Mol object at 0x7f20b9a6a280>> = <rdkit.Chem.rdchem.Mol object at 0x7f20b9a6a280>.GetNumAtoms

/home/runner/work/mdanalysis/mdanalysis/testsuite/MDAnalysisTests/coordinates/test_rdkit.py:186: AssertionError
```

I assume that this one has nothing to do with this PR?

I am listed as the Assignee. However, @IAlibay when you're happy with the PR I will not object to you merging. Or ping me when a final look-over is needed & I'll happily do the merge.

lilyminium:
@orbeckst it sounds like it's related to #2958 but I have not personally seen this failure before.

IAlibay:
I'll restart CI, we really need to make @cbouy's PRs our next priority for 2.0.

IAlibay (approved):
lgtm, I'll let you have a final look/merge @orbeckst

IAlibay:
In the interest of progressing towards 2.0, I'll go ahead with the squash-merge.

orbeckst:
Thanks.

yuxuanzhuang:
Thanks for the review!

PR #2953, opened 2020-09-24
ParmEdConverter: fix NameError when catching NoDataError

cbouy:
Converting to parmed when the AtomGroup doesn't have a resname leads to a NameError: the current code tries to catch the NoDataError but because NoDataError is not imported from exceptions.py it triggers a `NameError: name 'NoDataError' is not defined`

 - import NoDataError from exceptions

Is the changelog update needed here ?

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/coordinates/ParmEd.py
+++ b/package/MDAnalysis/coordinates/ParmEd.py
@@ -77,6 +77,7 @@ import functools
 from . import base
 from ..topology.tables import SYMB2Z
 from ..core.universe import Universe
+from ..exceptions import NoDataError
 
 
 class ParmEdReader(base.SingleFrameReaderBase):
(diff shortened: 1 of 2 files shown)


lilyminium:
Oops, thanks for finding this. Could you please add a test as well so the tests will catch it in the future as well?

orbeckst:
You don't need to update the CHANGELOG as this just fixes something that hadn't been released yet. (We don't absolutely strictly adhere to what I just said because our release cycles are so stretched out that we need to track bugs in develop and often fix issues that never show up in a released version. But in the case here a CHANGELOG entry seems overkill.)

orbeckst (changes requested):
Can you please add a test?

Not for testing that your fix worked but rather because it indicates that we didn't test a possible (and perhaps not uncommon) failure mode earlier.

cbouy:
Sorry for the delay and the messed up commits, should be fixed now

orbeckst (approved):

orbeckst:
Thank you!

PR #2956, opened 2020-09-27
TST, CI: add ARM64 Graviton 2 to CI

tylerjereddy:
* start testing MDAnalysis on ARM64 using
new [AWS Graviton 2](https://blog.travis-ci.com/AWS-Graviton-2-support-comes-to-Travis-CI) architecture available
in Travis CI

* testing on my fork shows only two failures
in the MDAnalysis test suite using a fairly
minimal set of dependencies; we can probably
either fix those or skip them and open an issue

* the test failures are:
1. `test_written_remarks_property`
2. `TestEncoreClustering.test_clustering_three_ensembles_two_identical`

* run time for this CI entry is about
23 minutes, which is solid for ARM; we save a lot
of time using experimental upstream wheels in some
cases and excluding (source builds of) dependencies
where possible until the ARM64 binary wheel ecosystem
matures a bit more

### DIFF
--- a/testsuite/MDAnalysisTests/formats/test_libdcd.py
+++ b/testsuite/MDAnalysisTests/formats/test_libdcd.py
@@ -20,6 +20,7 @@ import os
 import sys
 import string
 import struct
+import platform
 
 import hypothesis.strategies as strategies
 from hypothesis import example, given
@@ -351,8 +352,10 @@ def write_dcd(in_name, out_name, remarks='testing', header=None):
             f_out.write(xyz=frame.xyz, box=frame.unitcell)
 
 
-@pytest.mark.xfail(os.name == 'nt' and sys.maxsize <= 2**32,
-                   reason="occasional fail on 32-bit windows")
+@pytest.mark.xfail((os.name == 'nt'
+                    and sys.maxsize <= 2**32) or
+                    platform.machine() == 'aarch64',
+                   reason="occasional fail on 32-bit windows and ARM")
 @given(remarks=strategies.text(
     alphabet=string.printable, min_size=0,
     max_size=239))  # handle the printable ASCII strings
--- a/.travis.yml
+++ b/.travis.yml
@@ -96,6 +96,31 @@ matrix:
            INSTALL_HOLE="false"
            CODECOV="true" SETUP_CMD="${PYTEST_FLAGS} --cov=MDAnalysis"
 
+    - os: linux
+      language: python
+      arch: arm64-graviton2
+      python:
+        - "3.7"
+      dist: focal
+      virt: vm
+      group: edge
+      before_install:
+        - python -m pip install cython numpy
+        # special test SciPy wheel for ARM64:
+        - python -m pip install https://anaconda.org/multibuild-wheels-staging/scipy/1.6.0.dev0+a240c17/download/scipy-1.6.0.dev0+a240c17-cp37-cp37m-manylinux2014_aarch64.whl
+        - python -m pip install --no-build-isolation hypothesis matplotlib pytest pytest-cov pytest-xdist tqdm
+      install:
+        - cd package
+        - python setup.py install
+        - cd ../testsuite
+        - python setup.py install
+        - cd ..
+      script:
+        - cd testsuite
+        - python -m pytest ./MDAnalysisTests --disable-pytest-warnings -n 8 -rsx --cov=MDAnalysis
+      after_success:
+        - echo "Override this stage for ARM64"
+
   allow_failures:
     - env: NUMPY_VERSION=dev EVENT_TYPE="cron"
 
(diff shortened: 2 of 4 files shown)


tylerjereddy:
Looks like subtle behavior differences in `MDAnalysis/analysis/encore/clustering/src/ap.c` on the platforms of the only test that really matters here (the other is already known to be a issue on other plaforms so we can probably `xfail` the remarks one on arm64 with the `hypothesis` case problems there).

Looking at the first few lines of a `diff` command on the debug output from that C file when running the test in the absence of scikit-learn on both platforms:

```
-6.07   -5.59   -5.17   -4.75   -4.33   -3.97   -3.56   -3.24 | -6.07   -5.59   -5.17   -4.75   -4.33   -3.97   -3.56   -3.24
tmp -0.20                                                     | tmp -0.21
tmp -0.20                                                     | tmp -0.21
-0.54   -0.46   -0.96   -1.42   -1.90   -2.33   -2.75   -3.14 | -0.54   -0.46   -0.96   -1.42   -1.90   -2.33   -2.75   -3.14
0.46    -0.46   -0.96   -1.42   -1.90   -2.33   -2.75   -3.14 | 0.46    -0.46   -0.96   -1.42   -1.90   -2.33   -2.75   -3.14
tmp -4.44                                                     | tmp -4.45
tmp -4.44                                                     | tmp -4.45
```

tylerjereddy:
A bunch of `float` -> `double` changes in that source file allows the test to pass on an ARM64 gcc compile farm now; I'll see if I can reduce the size of the diff/changes.

tylerjereddy:
Perhaps also useful to know if we *really* need `float` for "performance." Using `double` is often preferable anyway. Perhaps I could add a platform-conditional remap in the worst case.

testsuite/MDAnalysisTests/formats/test_libdcd.py
```diff
@@ -351,8 +352,10 @@ def write_dcd(in_name, out_name, remarks='testing', header=None):
             f_out.write(xyz=frame.xyz, box=frame.unitcell)
 
 
-@pytest.mark.xfail(os.name == 'nt' and sys.maxsize <= 2**32,
-                   reason="occasional fail on 32-bit windows")
+@pytest.mark.xfail((os.name == 'nt'
+                    and sys.maxsize <= 2**32) or
+                    platform.machine() == 'aarch64',
+                   reason="occasional fail on 32-bit windows and ARM")
```
tylerjereddy, line 358: this doesn't actually fail on the `gcc 115` gcc compile farm ARM64 machine, but the test is clearly already flaky on some other platforms--maybe depends on character encoding/support stuff?

tylerjereddy (commented):
We'll see if this passes ARM64 CI now..

tylerjereddy:
hmm, I still need to iterate more it seems; the partial switch to `double` was harmful for another encore test

tylerjereddy:
If I use `gcc 4.x` on ARM64 the failure is different than the 9.x series version in use in CI (and on the compile farm machine):

```
E       AssertionError: Unexpected result: <ClusterCollection with 42 clusters>
E       assert 42 == 40
E        +  where 42 = len(<ClusterCollection with 42 clusters>)
```

The "bad" value changes to 36 with `gcc 9.x`.

tylerjereddy:
Selectively reducing compiler optimization level to `-O1` for ARM64 machines allows the encore tests to pass in my hands. We'll see if the CI agrees.

tylerjereddy:
@lilyminium @richardjgowers Can you live with the ARM64 shims here? CI is green now; the new job still runs in ~23 minutes.

I think we'll be able to bump up the optimization level a bit--`-O1` was my second try, but `-O2` might be possible. We can always make that improvement later though.

IAlibay:
It's interesting that the tests still run in a reasonably short amount of time even at O1. Even though they output incorrect values, are they substantially faster at 02/03?

It would be good to eventually work out if we can isolate the opt flag that's causing this. If anything it would help us understand how/if the code can be rewritten in a safer manner.

tylerjereddy:
> are they substantially faster at 02/03?

The test suite as a whole in CI is about the same runtime at `-O1` on ARM64. You can see by comparing the time I reported for the ARM64 job when I first opened the PR at `-O3` vs. current run time after realizing that aggressive optimization was causing problems.

tylerjereddy:
Some performance loss in the testsuite may also be made up by reduced build time with optimizations turned down. In any case, I think correctness and portability trump the hunger for performance here, especially since I've isolated it to a single architecture.

Maybe open an issue to perfect things on ARM64 eventually, but that will be a separate task I think--the build/iteration times on ARM64 can still be slow on some of the openly available machines, and we're really only using a small subset of dependencies until that ecosystem matures anyway.

richardjgowers (approved):
Yeah this looks like a good step towards having Arm compatibility, thanks @tylerjereddy !

IAlibay (approved):
Thanks for all the work here @tylerjereddy Is the only thing that needs changing before we claim "minimal ARM support" the use of the scipy development wheels?

tylerjereddy:
Yeah, I released SciPy `1.5.3` over the weekend with "official" Linux ARM wheels so I'll try to clean that up in a follow-up

PR #2959, opened 2020-09-29
Removes element guessing from ParmedParser (Issue #2933)

### LINKED ISSUE #2933: Remove element guessing by default in ParmedParser
As previously discussed #2630, #2651, etc... we probably don't want to be guessing by default if we can avoid it, particularly for elements where guess_types/elements (synonymous) is particularly error prone.

The ParmEdParser currently guesses by default elements if the atomic number doesn't match anything in `Z2SYMB`.
https://github.com/MDAnalysis/mdanalysis/blob/b2b7bcbdf4e2fe4e221d5e67842c90cf317375e5/package/MDAnalysis/topology/ParmEdParser.py#L219-L223

Change the behaviour of the ParmeEdParser. If the element can't be found, then just assign the element as an empty string (as we do for the TOPParser). 

Alternatively we could either: a) not assign elements if there's a bad element, or b) keep things as they are.

IAlibay:
Fixes #2933

 - If elements are not present in the parmed structure, the element is assigned to empty string `''`.
 - Tests for ParmedParser element reading.
 - Some PEP8/Flake8 changes.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/testsuite/MDAnalysisTests/topology/test_parmed.py
+++ b/testsuite/MDAnalysisTests/topology/test_parmed.py
@@ -21,8 +21,10 @@
 # J. Comput. Chem. 32 (2011), 2319--2327, doi:10.1002/jcc.21787
 #
 import pytest
-import MDAnalysis as mda
+import numpy as np
+from numpy.testing import assert_equal
 
+import MDAnalysis as mda
 from MDAnalysisTests.topology.base import ParserBase
 from MDAnalysisTests.datafiles import (
     PSF_NAMD_GBIS,
@@ -30,6 +32,7 @@ from MDAnalysisTests.datafiles import (
 )
 
 pmd = pytest.importorskip('parmed')
+
 
 class BaseTestParmedParser(ParserBase):
     parser = mda.topology.ParmEdParser.ParmEdParser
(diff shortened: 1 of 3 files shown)


testsuite/MDAnalysisTests/topology/test_parmed.py
```diff
@@ -168,16 +177,22 @@ def test_dihedrals_identity(self, top, value):
...
     PRM
     """
-    
+
     ref_filename = PRM
-    
+
     expected_n_atoms = 252
     expected_n_residues = 14
+    elems_ranges = ((0, 8), (30, 38))
```
IAlibay, line 191: So strangely enough Parmed here actually guesses elements from masses (hence why PRM directly loaded in MDA has no elements, but this yields an element list). But since it's outside our hands... I guess it's fine?

lilyminium: I agree, IMO the goal should be to faithfully represent the ParmEd object, however it chooses to interpret the original file.

lilyminium (approved):
LGTM! Thanks for doing this and cleaning up the code 😅

PR #2968, opened 2020-10-04
Have copies of docs at stable/ and dev/

### LINKED ISSUE #2966: Expose "stable" docs endpoint
I refer to MDAnalysis in the [documentation for freud](https://freud.readthedocs.io/en/stable/topics/datainputs.html#mdanalysis-readers) through an [intersphinx mapping](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html). I saw that the docs have moved to docs.mdanalysis.org. It would be nice to have a "stable" version of the MDAnalysis docs that could be referenced.

Add a "stable" docs version pointing to the latest stable release. Something like https://docs.mdanalysis.org/stable/. This is also how some other Python projects label their docs, e.g.
- https://numpy.org/doc/stable/
- https://scikit-learn.org/stable/

I could pin my library to link to the 1.0.0 docs but that would become out of date as new MDAnalysis versions are released.

lilyminium:
Fixes #2966 (kind of)*

 - Copies the latest release to https://docs.mdanalysis.org/stable/
 - Copies latest dev docs to https://docs.mdanalysis.org/dev
 - Allows docs to be intersphinxed at above URLs

It will hopefully look like the way docs behave at https://minium.com.au/mdanalysis/ . In short:
* https://minium.com.au/mdanalysis/ redirects to https://minium.com.au/mdanalysis/stable
* https://minium.com.au/mdanalysis/documentation_pages/topology/GSDParser redirects to https://minium.com.au/mdanalysis/stable/documentation_pages/topology/GSDParser.html
* https://minium.com.au/mdanalysis/1.0.0/ is the exact same docs but at a different URL
* https://minium.com.au/mdanalysis/dev/ is the same docs as https://minium.com.au/mdanalysis/2.0.0-dev0/ at a different URL
* https://minium.com.au/mdanalysis/latest redirects to https://minium.com.au/mdanalysis/1.0.0 because I set up the "latest" endpoint earlier and am not sure if it will break anything if we remove it

This doubling up of docs is based off `numpy` behaviour, i.e. https://numpy.org/doc/stable/ appears to be the same docs as https://numpy.org/doc/1.19/ . If it's too redundant, we could remove the `1.0.0` and `2.0.0-dev0` entries and just keep `stable` and `dev`.

\* I manually made the "stable" docs for 1.0.0 because `master` has moved onto 1.0.1-dev, but the "stable" docs in https://minium.com.au/mdanalysis/ were made by changing the version to "1.0.0" (hence the different date for when they were last built). So https://docs.mdanalysis.org/stable/index.html already exists despite not merging this yet.

Edit: this should go into master as well because it changes the behaviour of the "dev/" URL.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,9 @@ os:
 
 env:
   global:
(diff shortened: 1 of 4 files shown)


orbeckst (approved):
I agree with your reasoning and decisions – that's the main thing for which I add the 👍 .

From skim-reading the code it was not clear to me if `latest` in the code refers to "dev" (which is how numpy labels their docs https://numpy.org/doc/). If you could just add a comment (deploy script and update Python script) to add semantics then that would be helpful.

PR #2981, opened 2020-10-13
Use a fixture for npz files in test_bat

### LINKED ISSUE #2979: test_bat.py::test_bat_incorrect_dims refers to a file that may not exist
## Expected behavior ##

Running the tests in parallel do not yield failed tests.

## Actual behavior ##

When running `pytest  --disable-pytest-warnings -n 6`

```
TestBAT.test_bat_incorrect_dims ____________________________________________
[gw2] linux -- Python 3.9.0 /home/jon/miniconda3/envs/py39/bin/python3.9

self = <MDAnalysisTests.analysis.test_bat.TestBAT object at 0x7f0a65f00940>

    def test_bat_incorrect_dims(self):
        u = mda.Universe(PSF, DCD)
        selected_residues = u.select_atoms("resid 1-3")
>           R = BAT(selected_residues, filename='test_bat_IO.npy')

analysis/test_bat.py:114:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../package/MDAnalysis/analysis/bat.py:364: in __init__
    self.load(filename)
../../package/MDAnalysis/analysis/bat.py:450: in load
    self.bat = np.load(filename)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

file = 'test_bat_IO.npy', mmap_mode = None, allow_pickle = False, fix_imports = True, encoding = 'ASCII'

    @set_module('numpy')
    def load(file, mmap_mode=None, allow_pickle=False, fix_imports=True,
             encoding='ASCII'):
        """
        Load arrays or pickled objects from ``.npy``, ``.npz`` or pickled files.

        .. warning:: Loading files that contain object arrays uses the ``pickle``
                     module, which is not secure against erroneous or maliciously
                     constructed data. Consider passing ``allow_pickle=False`` to
                     load data that is known not to contain object arrays for the
                     safer handling of untrusted sources.

        Parameters
        ----------
        file : file-like object, string, or pathlib.Path
            The file to read. File-like objects must support the
            ``seek()`` and ``read()`` methods. Pickled files require that the
            file-like object support the ``readline()`` method as well.
        mmap_mode : {None, 'r+', 'r', 'w+', 'c'}, optional
            If not None, then memory-map the file, using the given mode (see
            `numpy.memmap` for a detailed description of the modes).  A
            memory-mapped array is kept on disk. However, it can be accessed
            and sliced like any ndarray.  Memory mapping is especially useful
            for accessing small fragments of large files without reading the
            entire file into memory.
        allow_pickle : bool, optional
            Allow loading pickled object arrays stored in npy files. Reasons for
            disallowing pickles include security, as loading pickled data can
            execute arbitrary code. If pickles are disallowed, loading object
            arrays will fail. Default: False

            .. versionchanged:: 1.16.3
                Made default False in response to CVE-2019-6446.

        fix_imports : bool, optional
            Only useful when loading Python 2 generated pickled files on Python 3,
            which includes npy/npz files containing object arrays. If `fix_imports`
            is True, pickle will try to map the old Python 2 names to the new names
            used in Python 3.
        encoding : str, optional
            What encoding to use when reading Python 2 strings. Only useful when
            loading Python 2 generated pickled files in Python 3, which includes
            npy/npz files containing object arrays. Values other than 'latin1',
            'ASCII', and 'bytes' are not allowed, as they can corrupt numerical
            data. Default: 'ASCII'

        Returns
        -------
        result : array, tuple, dict, etc.
            Data stored in the file. For ``.npz`` files, the returned instance
            of NpzFile class must be closed to avoid leaking file descriptors.

        Raises
        ------
        IOError
            If the input file does not exist or cannot be read.
        ValueError
            The file contains an object array, but allow_pickle=False given.

        See Also
        --------
        save, savez, savez_compressed, loadtxt
        memmap : Create a memory-map to an array stored in a file on disk.
        lib.format.open_memmap : Create or load a memory-mapped ``.npy`` file.

        Notes
        -----
        - If the file contains pickle data, then whatever object is stored
          in the pickle is returned.
        - If the file is a ``.npy`` file, then a single array is returned.
        - If the file is a ``.npz`` file, then a dictionary-like object is
          returned, containing ``{filename: array}`` key-value pairs, one for
          each file in the archive.
        - If the file is a ``.npz`` file, the returned value supports the
          context manager protocol in a similar fashion to the open function::

            with load('foo.npz') as data:
                a = data['a']

          The underlying file descriptor is closed when exiting the 'with'
          block.

        Examples
        --------
        Store data to disk, and load it again:

        >>> np.save('/tmp/123', np.array([[1, 2, 3], [4, 5, 6]]))
        >>> np.load('/tmp/123.npy')
        array([[1, 2, 3],
               [4, 5, 6]])

        Store compressed data to disk, and load it again:

        >>> a=np.array([[1, 2, 3], [4, 5, 6]])
        >>> b=np.array([1, 2])
        >>> np.savez('/tmp/123.npz', a=a, b=b)
        >>> data = np.load('/tmp/123.npz')
        >>> data['a']
        array([[1, 2, 3],
               [4, 5, 6]])
        >>> data['b']
        array([1, 2])
        >>> data.close()

        Mem-map the stored array, and then access the second row
        directly from disk:

        >>> X = np.load('/tmp/123.npy', mmap_mode='r')
        >>> X[1, :]
        memmap([4, 5, 6])

        """
        if encoding not in ('ASCII', 'latin1', 'bytes'):
            # The 'encoding' value for pickle also affects what encoding
            # the serialized binary data of NumPy arrays is loaded
            # in. Pickle does not pass on the encoding information to
            # NumPy. The unpickling code in numpy.core.multiarray is
            # written to assume that unicode data appearing where binary
            # should be is in 'latin1'. 'bytes' is also safe, as is 'ASCII'.
            #
            # Other encoding values can corrupt binary data, and we
            # purposefully disallow them. For the same reason, the errors=
            # argument is not exposed, as values other than 'strict'
            # result can similarly silently corrupt numerical data.
            raise ValueError("encoding must be 'ASCII', 'latin1', or 'bytes'")

        pickle_kwargs = dict(encoding=encoding, fix_imports=fix_imports)

        with contextlib.ExitStack() as stack:
            if hasattr(file, 'read'):
                fid = file
                own_fid = False
            else:
>               fid = stack.enter_context(open(os_fspath(file), "rb"))
E               FileNotFoundError: [Errno 2] No such file or directory: 'test_bat_IO.npy'

/home/jon/miniconda3/envs/py39/lib/python3.9/site-packages/numpy/lib/npyio.py:416: FileNotFoundError
```

This error seems to be due to the test assuming "test_bat_IO.npy" has been created by another test. This may not have happened when the tests run in parallel.

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) `2.0.0-dev0` (a9543ba29083449944b34d0b77429cab12f2409a)
- Which version of Python (`python -V`)?  `Python 3.9.0`
- Which operating system? Ubuntu 20.04 over WSL

jbarnoud:
Fixes #2979

The file test_bat_IO.npy was created in one test and used in
another one. This leads to the second test possibly failing
when the tests are run in parallel. This commit moves to create
the file into a fixture that the two tests depend.

The new fixture also uses the tmpdir fixture to avoid the file
created during the tests to remain in the test directory.

PR Checklist
------------
 - [X] Tests?
 - ~[ ] Docs?~
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -58,6 +58,8 @@ Fixes
     Universe (PR #2893)
   * ensure that unistd.h is included on macOS when compiling ENCORE's spe.c
     (Issue #2934)
+  * Fix tests for analysis.bat that could fail when run in parallel and that
+    would create a test artifact (Issue #2979, PR #2981)
 
 Enhancements
   * Refactored analysis.helanal into analysis.helix_analysis
--- a/testsuite/MDAnalysisTests/analysis/test_bat.py
+++ b/testsuite/MDAnalysisTests/analysis/test_bat.py
@@ -31,8 +31,6 @@ from MDAnalysisTests.datafiles import (PSF, DCD, mol2_comments_header, XYZ_mini,
                                        BATArray)
 from MDAnalysis.analysis.bat import BAT
 
-import os
-
 
 class TestBAT(object):
     @pytest.fixture()
--- a/testsuite/MDAnalysisTests/analysis/test_bat.py
+++ b/testsuite/MDAnalysisTests/analysis/test_bat.py
@@ -31,6 +31,8 @@ from MDAnalysisTests.datafiles import (PSF, DCD, mol2_comments_header, XYZ_mini,
                                        BATArray)
 from MDAnalysis.analysis.bat import BAT
 
+import os
+
 
 class TestBAT(object):
     @pytest.fixture()
@@ -44,6 +46,14 @@ class TestBAT(object):
         R = BAT(selected_residues)
         R.run()
         return R.bat
+
+    @pytest.fixture
+    def bat_npz(self, tmpdir, selected_residues):
+        filename = str(tmpdir / 'test_bat_IO.npy')
+        R = BAT(selected_residues)
+        R.run()
+        R.save(filename)
+        return filename
 
     def test_bat_root_selection(self, selected_residues):
         R = BAT(selected_residues)
@@ -79,11 +89,8 @@ class TestBAT(object):
             err_msg="error: Reconstructed Cartesian coordinates " + \
                     "don't match original")
 
-    def test_bat_IO(self, selected_residues, bat):
-        R = BAT(selected_residues)
-        R.run()
-        R.save('test_bat_IO.npy')
-        R2 = BAT(selected_residues, filename='test_bat_IO.npy')
+    def test_bat_IO(self, bat_npz, selected_residues, bat):
+        R2 = BAT(selected_residues, filename=bat_npz)
         test_bat = R2.bat
         assert_almost_equal(
             bat,
@@ -107,8 +114,8 @@ class TestBAT(object):
         with pytest.raises(ValueError):
             R = BAT(selected_residues)
 
-    def test_bat_incorrect_dims(self):
+    def test_bat_incorrect_dims(self, bat_npz):
         u = mda.Universe(PSF, DCD)
         selected_residues = u.select_atoms("resid 1-3")
         with pytest.raises(ValueError):
-            R = BAT(selected_residues, filename='test_bat_IO.npy')
+            R = BAT(selected_residues, filename=bat_npz)


jbarnoud:
There is one build failing in an odd way that seems absolutelly unrelated to the PR:

```
_________________ ERROR collecting MDAnalysisTests/test_api.py _________________

../../../miniconda/envs/test/lib/python3.6/site-packages/_pytest/python.py:571: in _importtestmodule

    mod = import_path(self.fspath, mode=importmode)

../../../miniconda/envs/test/lib/python3.6/site-packages/_pytest/pathlib.py:517: in import_path

    importlib.import_module(module_name)

../../../miniconda/envs/test/lib/python3.6/importlib/__init__.py:126: in import_module

<frozen importlib._bootstrap>:994: in _gcd_import

    ???

<frozen importlib._bootstrap>:971: in _find_and_load

    ???

<frozen importlib._bootstrap>:955: in _find_and_load_unlocked

    ???

<frozen importlib._bootstrap>:665: in _load_unlocked

    ???

../../../miniconda/envs/test/lib/python3.6/site-packages/_pytest/assertion/rewrite.py:171: in exec_module

    exec(co, module.__dict__)

testsuite/MDAnalysisTests/test_api.py:27: in <module>

    import MDAnalysis as mda

package/MDAnalysis/__init__.py:207: in <module>

    from .coordinates.core import writer as Writer

package/MDAnalysis/coordinates/__init__.py:735: in <module>

    from . import H5MD

../../../miniconda/envs/test/lib/python3.6/site-packages/duecredit/injections/injector.py:285: in __import

    mod = self._orig_import(name, *args, **kwargs)

package/MDAnalysis/coordinates/H5MD.py:193: in <module>

    import h5py

../../../miniconda/envs/test/lib/python3.6/site-packages/duecredit/injections/injector.py:285: in __import

    mod = self._orig_import(name, *args, **kwargs)

../../../miniconda/envs/test/lib/python3.6/site-packages/h5py/__init__.py:62: in <module>

    from .tests import run_tests

../../../miniconda/envs/test/lib/python3.6/site-packages/duecredit/injections/injector.py:285: in __import

    mod = self._orig_import(name, *args, **kwargs)

../../../miniconda/envs/test/lib/python3.6/site-packages/h5py/tests/__init__.py:15: in <module>

    from . import old, hl

../../../miniconda/envs/test/lib/python3.6/site-packages/duecredit/injections/injector.py:285: in __import

    mod = self._orig_import(name, *args, **kwargs)

../../../miniconda/envs/test/lib/python3.6/site-packages/h5py/tests/old/__init__.py:4: in <module>

    from . import ( test_attrs,

../../../miniconda/envs/test/lib/python3.6/site-packages/duecredit/injections/injector.py:285: in __import

    mod = self._orig_import(name, *args, **kwargs)

<frozen importlib._bootstrap>:971: in _find_and_load

    ???

<frozen importlib._bootstrap>:955: in _find_and_load_unlocked

    ???

<frozen importlib._bootstrap>:665: in _load_unlocked

    ???

../../../miniconda/envs/test/lib/python3.6/site-packages/_pytest/assertion/rewrite.py:162: in exec_module

    source_stat, co = _rewrite_test(fn, self.config)

../../../miniconda/envs/test/lib/python3.6/site-packages/_pytest/assertion/rewrite.py:357: in _rewrite_test

    tree = ast.parse(source, filename=fn_)

../../../miniconda/envs/test/lib/python3.6/ast.py:35: in parse

    return compile(source, filename, mode, PyCF_ONLY_AST)

E     File "/home/travis/miniconda/envs/test/lib/python3.6/site-packages/h5py/tests/old/test_attrs_data.py", line 251

E       s = b"Hello\x00\Hello"

E          ^

E   SyntaxError: invalid escape sequence \H
```

package/CHANGELOG
```diff
@@ -58,6 +58,8 @@ Fixes
     Universe (PR #2893)
   * ensure that unistd.h is included on macOS when compiling ENCORE's spe.c
     (Issue #2934)
+  * Fix tests for analysis.bat that could fail when run in parallel and that
```
IAlibay, line 61: I guess technically it should be newest first, but I've seen plenty of cases where we've not followed this so it's a bit late to be pedantic now 🤷‍♂️

IAlibay (approved):
😅 I was initially confused as to how the original code worked without leaving files behind, but I guess it does....

Thanks @jbarnoud, lgtm!

IAlibay:
> There is one build failing in an odd way that seems absolutelly unrelated to the PR:

That's a weird one, I don't have a clue what's happening here :( I've restarted the job just to see if it's a consistent failure.

IAlibay:
Test still fails.

What's particularly interesting here is that unlike the py3.7+ tests, the h5py version which ends up being installed is 2.7 rather than 2.10.

Likely related, for some strange reason, the dry run does pick up 2.10, but then in the actual install it picks up 2.7:
https://travis-ci.com/github/MDAnalysis/mdanalysis/jobs/398748734#L1137

package/CHANGELOG
jbarnoud: Oops... I have been doing it wrong for a while 😅

jbarnoud:
Also, it looks like previous builds from 3 days ago were picking 2.10 as expected.

jbarnoud:
Looks like the travis error was a fluke, probably with the conda-forge repository.

IAlibay:
> Looks like the travis error was a fluke, probably with the conda-forge repository.

How strange, I can't see anything upstream that would be the cause of this, at least it works now 🤷‍♂️

Issue #2376, opened 2019-10-25
why is (cached) fragments slow?


orbeckst:
**Is your feature request related to a problem? Please describe.**
The "make whole" (unwrap) on the fly transformation PR #2038 needs the fragments at every time step. ~Apparently, the fragments are recalculated every time~ (EDIT: They are not – see discussion below) 

```python
>>> u = mda.Universe(data.TPR, data.XTC)
>>> %timeit u.atoms.fragments
556 ms ± 71.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Recalculating fragments takes 0.5 seconds per step.

**Describe the solution you'd like**
~There is no need to recalculate fragments as they are based on bonds. Fragments can be created when the topology is created.~

~Or only recalculate if the bonds change, or just tell users that if they change bonds then they also need to recalculate fragments.~

EDIT: Recalculating fragments is not the problem but building the actual data structure appears to be slow.

**Describe alternatives you've considered**
Recalculating fragments every time the managed attribute is accessed guards against stale data, which is good. However, user expectation is that attributes behave as instantaneous value stores (and the fact that we use managed attributes should not concern users). Furthermore, I contend that the case of accessing fragments for different time steps is more common than the case of changing the topology. 

**Additional context**

Doing the unwrap is slow no matter what, ca 4s/iteration
```python
In [81]: u = mda.Universe(TPR, XTC)
In [82]: u.trajectory.add_transformations(MDAnalysis.transformations.wrap.unwrap(u.atoms))
In [83]: with mda.Writer("unwrap.xtc", n_atoms=u.atoms.n_atoms) as W:
    ...:     for ts in tqdm.tqdm(u.trajectory):
    ...:         W.write(u.atoms)
    ...:
100%|██████████████████████████████████████████████████████████| 10/10 [00:39<00:00,  3.90s/it]
```
but at least shaving off 0.5 s is a start.

richardjgowers:
https://github.com/MDAnalysis/mdanalysis/blob/develop/package/MDAnalysis/core/universe.py#L998

I think they are cached at a universe level... so maybe something else
weird is happening here. Maybe it’s the construction of the fragments for
this AG from the master list?

orbeckst:
As they are cached then something else appears to be slow. Needs some line profiling, I think.

orbeckst:
[...]

PR #2982, opened 2020-10-13
Implements compact wrapping and nojump, and updates associated transforms.

mnmelo:
Implemented function `distances.apply_compact_PBC`. With it I could easily adapt `AtomGroup.wrap()` to also do compact wrapping, depending on a keyword. Backward compatibility is maintained.

Also optimized the way `compound_indices` are handled in `AtomGroup.wrap()`, with major speedup:
For an `AtomGroup` of 146k atoms:

### Before
```python
%timeit ag.wrap(compound='fragments', inplace=False)
1.08 s ± 1.02 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### After
```python
%timeit ag.wrap(compound='fragments', inplace=False)
72.7 ms ± 1.38 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

In `transformations`:
- Made `transformations.wrap` and `transformations.unwrap` transparently expose AtomGroup.wrap() and AtomGroup.unwrap(), to take advantage of the new functionality.

- Added `transformations.nojump` (should this be a separate PR?)

Also added the `center` keyword to `distances.apply_PBC` (with care to minimally impact performance), which allows the PBC wrapping to an arbitrarily centered box. This helps with vector treatment in `transformations.nojump`, and can be useful in other cases (say, to wrap atoms to a box centered on the origin or on a dynamic point). Should `AtomGroup.wrap()` expose this capability too?

Specific tests still to be written, but would like feedback in the meantime. Existing tests pass, except for the seemingly unrelated `testsuite/MDAnalysisTests/analysis/test_pca.py:test_given_mean` and `testsuite/MDAnalysisTests/analysis/test_encore.py:TestEncoreClustering.test_clustering_three_ensembles_two_identical` (which also fail for the unmodified `develop` branch).

Feedback much appreciated!
 
PR Checklist
------------
 - [ ] Tests?
 - [x] Docs?
 - [ ] CHANGELOG updated?
(There's no issue for these enhancements. Should I open one?)


orbeckst:
I didn't have the time to review, sorry. But: >10 times speedup for wrapping/unwrapping — that will finally make these transformations  usable 🚀 . Also, getting *compact* and *nojump* will be super-neat. 

I would break up the PR – one for the performance, one or more PRs for new features. The smaller the better, really. 

If the performance one can be separate from everything else then we can easily backport it.

mnmelo:
Perfect, I'll split it up, then. Just to be clear, I only optimized the wrapping. Was actually right now looking at the unwrapping and think we can look at some 5x speedup for a 100k atom system. Anyway, we can discuss better in the respective PRs.

Also, the atom-based wrapping is already quite fast (milliseconds for 100k atoms). What I optimized was the fragment based wrapping.

orbeckst:
The fragment-based wrapping is the crucial one for making sane trajectories.... so any improvements there are _very_ welcome!

mnmelo:
Thanks for the positive outlook, @orbeckst. To make this work I propose some refactoring to the compound and cache code in #3000 and #3005, and further unwrap-specific optimizations in #2376. I really need some approval of those ideas moving forward, because they involve some work and touch some central aspects that others might want to chip in.

orbeckst:
nojump would be really good for MSD (@hmacdope recently told someone on the mailing list that they had to use GROMACS for that part) and compact has always been one my favorites in trjconv. Having both would boost transformations.

I notice that all the prerequisite PRs have been merged so in principle we "just" need (1) merging of develop into this PR and (2) a few 👀 on the code. It would be a shame to let it rot.

hmacdope:
@mnmelo there is now an nojump method available, and this PR is quite stale. Is it okay for me to close here?

mnmelo:
Sure!

orbeckst:
We have the nojump but IIRC a major contribution of this PR to speed up unwrapping ("make molecules whole across PBC") which is still veeeeeeeeeeeeeeeeeery slow. 

It's become very clear to me that I don't have the bandwidth to move this PR forward (apologies) but I am also not sure that it makes sense to close it if there's code in there that we should really be working on. @hmacdope have you had a deeper look?

tylerjereddy:
@PardhavMaradani brought it to my attention that this PR does a good job with dodecahedral and octahedral GMX containers, which are pretty important for i.e., my virus work. See https://github.com/BradyAJohnston/MolecularNodes/issues/981#issuecomment-3369134809 and the impressive `MolecularNodes`-related box visualization therein that works with this branch.

So if this does get closed, please open an issue to track the capability to handle those container types and point to this as an attempt at it.

PR #2989, opened 2020-10-16
compile Cython extensions with Cython >= 3.0

### LINKED ISSUE #2988: compiling warns of using deprecated NumPy API
## Expected behavior ##

Compiling the MDA C extensions does not raise warnings about deprecated APIs.

## Actual behavior ##

Compiling the cython extension warns

*"Using deprecated NumPy API, disable it with " "#define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION"*.

This is due to Cython <3.0 using the deprecated NumPp API. This will change with upcoming Cython 3.0, see https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html#numpy-c-api at which point we can simply 
```python
define_macros.append("NPY_NO_DEPRECATED_API", "NPY_1_7_API_VERSION"))
```
in `setup.py`.

(However, with Cython < 3.0, compilation fails as it relies on the old API.)

```bash
python setup.py buil
```

- Cython: 0.29.21

orbeckst:
Fixes #2988

- Do not any longer use deprecated NumPy API.
- requires Cython >= 3.0 (tested locally with 3.0a6)
- fixed direct ndarray.base access (using https://github.com/cython/cython/issues/3690#issuecomment-645665793 )

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


orbeckst:
Note that this will fail until we install Cython 3.0.

richardjgowers:
Is there something in the travis yml to get cython 3?

orbeckst:
We could `pip install 'Cython==3.0.0a6'` (worked for me) or something like that (and remove cython from the conda/mamba installation).

IAlibay:
@orbeckst since cython 3.0 is still in alpha, I'm going to place this as a v2.1.0 milestone, we can shift as necessary.

IAlibay:
> @orbeckst since cython 3.0 is still in alpha, I'm going to place this as a v2.1.0 milestone, we can shift as necessary.

Moving this to 2.2.0 instead as cython 3 is still in alpha. We'll also need to consider if it should be a breaking change that gets introduced in MDA 3.0.

IAlibay:
Moving to 2.3.0 since cython 3.0 is still in alpha

orbeckst:
cython 3.0 is staying in alpha for quite a while...

IAlibay:
Cython 3.0 is still in alpha, re-targetting 2.4.0.

IAlibay:
Cython 3.0 is still in alpha, bumping to 2.5.0.

orbeckst:
PR #4129 solved the array problem. The only other thing that _this_ PR does is to get rid of the use of the deprecated NumPy API. However, this _requires_ Cython 3.x so cannot be done in a backward-compatible manner. 

I don't think it's worthwhile keeping the PR open for that little patch in setup.py so I am closing it.

PR #2991, opened 2020-10-16
Update indexing in Charges and Masses get_residues

### LINKED ISSUE #2990: `Residue.mass` and `Residue.charge` use deprecated numpy indexing
Running the tests throws a warning about MDAnalysis using a behaviour of numpy that is subject to change:

```
MDAnalysisTests/core/test_accumulate.py::TestTotals::test_total_charge_duplicates[residues]
  /home/jon/dev/mdanalysis/package/MDAnalysis/core/topologyattrs.py:1540: FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.
    charges = self.values[resatoms].sum()

MDAnalysisTests/core/test_accumulate.py::TestTotals::test_total_mass_duplicates[residues]
  /home/jon/dev/mdanalysis/package/MDAnalysis/core/topologyattrs.py:1119: FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.
    masses = self.values[resatoms].sum()
```

The warnings are thrown for `core/test_accumulate.py::TestTotals::test_total_charge` (and its mass equivalent): https://github.com/MDAnalysis/mdanalysis/blob/2ee4e9da5aa3a2c1b21fc3d1897bd70e0ab2064d/testsuite/MDAnalysisTests/core/test_accumulate.py#L121

The problematic call is a call to `Residue.charge` or `Residue.mass` (note the singular group name). This calls `topologyattrs.Charges.get_residues` which makes a problematic array indexing on line https://github.com/MDAnalysis/mdanalysis/blob/2ee4e9da5aa3a2c1b21fc3d1897bd70e0ab2064d/package/MDAnalysis/core/topologyattrs.py#L1540 or `topologyattrs.Masses.get_residues` on line https://github.com/MDAnalysis/mdanalysis/blob/2ee4e9da5aa3a2c1b21fc3d1897bd70e0ab2064d/package/MDAnalysis/core/topologyattrs.py#L1119

Here, an array is indexed with a list of arrays, hence the warning.

The `get_segments` counterpart are fixed and convert the list of arrays into a tuple of arrays, not triggering the warning.

``` python
import MDAnalysis as mda
from MDAnalysis.tests.datafiles import XTC, TRR

u = mda.Universe(TRR, XTC)
u.residues[0].mass  # or u.residues[0].charge
```

or

```
pytest core/test_accumulate.py::TestTotals
```

jbarnoud:
Fixes #2990

Avoid indexing a numpy array with a list of arrays, use a tuple of array
instead. Multidimensional indexing of arrays with a non-tuple argument
is deprecated.

PR Checklist
------------
 - [X] Tests?
 - ~[ ] Docs?~
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -20,6 +20,8 @@ The rules for this file:
   * 2.0.0
 
 Fixes
+  * Avoid using deprecated array indexing in topology attributes
+    (Issue #2990, PR #2991)
   * ParmedParser no longer guesses elements if they are not recognised, instead
     empty strings are assigned (Issue #2933)
   * Instead of using ATOM for both ATOM and HETATM, HETATM record type
--- a/package/MDAnalysis/core/topologyattrs.py
+++ b/package/MDAnalysis/core/topologyattrs.py
@@ -1116,7 +1116,7 @@ class Masses(AtomAttr):
 
         if isinstance(rg._ix, numbers.Integral):
             # for a single residue
-            masses = self.values[resatoms].sum()
+            masses = self.values[tuple(resatoms)].sum()
         else:
             # for a residuegroup
             masses = np.empty(len(rg))
@@ -1537,7 +1537,7 @@ class Charges(AtomAttr):
         resatoms = self.top.tt.residues2atoms_2d(rg.ix)
 
         if isinstance(rg._ix, numbers.Integral):
-            charges = self.values[resatoms].sum()
+            charges = self.values[tuple(resatoms)].sum()
         else:
             charges = np.empty(len(rg))
             for i, row in enumerate(resatoms):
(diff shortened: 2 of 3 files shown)


orbeckst (approved):
This seems non-controversial and easy to review: LGTM — and good idea to add the tests that now fail on the warning to avoid regression. We should follow the same pattern for future warning deprecation fixes (if we're not already doing it).

PR #2992, opened 2020-10-17
Fixes AtomGroup.center when using compounds + unwrapping

### LINKED ISSUE #2984: AtomGroup.center fails for compounds+wrapping
Digging into the optimization of `AtomGroup.unwrap` (from the discussion in #2982) I am trying to replicate the vectorization done in `AtomGroup.center`. I noticed that in there there's a sorting of compound indices that isn't being done if the user requests compound unwrapping. This breaks center calculation if for some reason a compound's atoms are not initially contiguous.

This is a corner case, but still possible. The fix is quite simple, and if it's ok with you I'd include it in the upcoming PR dedicated to the wrap/unwrap enhancement because that will also be dealing with similar code patterns, in the same source file (`core/groups.py`).

With a simple 5-water system:
```python
import MDAnalysis as mda
from MDAnalysis.tests.datafiles import waterPSF, waterDCD

u = mda.Universe(waterPSF, waterDCD)

testgrp = u.atoms
testndx = np.arange(len(testgrp))
np.random.shuffle(testndx)
testgrp_unordered = testgrp[testndx]
testgrp_unordered.fragindices
>> array([3, 0, 1, 1, 2, 2, 2, 4, 3, 4, 1, 0, 0, 4, 3])
```

The fragment COGs of the test group:
```python
testgrp.center_of_geometry(compound='fragments')
>> array([[ 16.86431313,  -1.89109834,  23.81008784],
          [ 18.23819478,   0.2692619 , -24.17409579],
          [ 18.62940788,  -3.88507994, -24.22752635],
          [ 20.89391009,  -0.74630801, -24.41894786],
          [ 21.33577983,  -3.48827068, -23.58818181]])
```

Same answer if atoms are unwrapped but consecutive (molecules are already whole in this test traj):
```python
testgrp.center_of_geometry(compound='fragments', unwrap=True)
>> array([[ 16.86431313,  -1.89109834,  23.81008784],
          [ 18.23819478,   0.2692619 , -24.17409579],
          [ 18.62940788,  -3.88507994, -24.22752635],
          [ 20.89391009,  -0.74630801, -24.41894786],
          [ 21.33577983,  -3.48827068, -23.58818181]])
```

Also ok if randomized but not unwrapped:
```python
testgrp_unordered.center_of_geometry(compound='fragments')
>> array([[ 16.86431313,  -1.89109834,  23.81008784],
          [ 18.23819478,   0.2692619 , -24.17409579],
          [ 18.62940788,  -3.88507994, -24.22752635],
          [ 20.89391009,  -0.74630801, -24.41894786],
          [ 21.33577983,  -3.48827068, -23.58818181]])
```

**Oops!**
```python
testgrp_unordered.center_of_geometry(compound='fragments', unwrap=True)
>> array([[ 18.76917013,  -0.9612124 ,  -8.50591214],
          [ 18.56716092,  -2.47748999, -23.95220439],
          [ 20.31393941,  -2.90605104, -23.84455808],
          [ 18.98535792,  -1.3438748 ,  -8.07818158],
          [ 19.32597733,  -2.05286684,  -8.21780777]])
```

`2.0.0-dev0` (at commit 64a7c051f1a51ff2a89b9da9257a21cea0ef00fa), Python 3.6.9, Ubuntu 18.04

EDIT: output clarification

mnmelo:
Fixes #2984

In `AtomGroup.center`, coordinates are now properly sorted after unwrapping, in tandem with `compound_indices` and `weights`. Additionally, sorting is now only even done if the `compound_indices` are not already monotonically increasing. Monotonicity is a rather cheap check for the most frequent use case, compared to the expensive sorting that's only really needed for the rarer cases of unordered/noncontiguous compounds. 

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/testsuite/MDAnalysisTests/core/test_atomgroup.py
+++ b/testsuite/MDAnalysisTests/core/test_atomgroup.py
@@ -992,6 +992,12 @@ class TestUnwrapFlag(object):
         return group
 
     @pytest.fixture()
+    def unordered_ag(self, ag):
+        ndx = np.arange(len(ag))
+        np.random.shuffle(ndx)
+        return ag[ndx]
+
+    @pytest.fixture()
     def ref_noUnwrap_residues(self):
         return {
             'COG': np.array([[21.356, 28.52, 36.762],
@@ -1013,12 +1019,12 @@ class TestUnwrapFlag(object):
             'COG': np.array([[21.356, 41.685, 40.501],
                              [44.577, 43.312, 79.039],
                              [ 2.204, 27.722, 54.023]], dtype=np.float32),
-            'COM': np.array([[20.815, 42.013, 39.802],
-                             [44.918, 43.282, 79.325],
-                             [2.045, 28.243, 54.127]], dtype=np.float32),
-            'MOI': np.array([[16747.486, -1330.489,  2938.243],
-                             [-1330.489, 19315.253,  3306.212],
-                             [ 2938.243,  3306.212,  8990.481]]),
+            'COM': np.array([[21.286, 41.664, 40.465],
+                             [44.528, 43.426, 78.671],
+                             [ 2.111, 27.871, 53.767]], dtype=np.float32),
+            'MOI': np.array([[16687.941, -1330.617, 2925.883],
+                             [-1330.617, 19256.178, 3354.832],
+                             [ 2925.883,  3354.832, 8989.946]]),
             'Asph': 0.2969491080,
         }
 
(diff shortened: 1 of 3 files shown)


testsuite/MDAnalysisTests/core/test_atomgroup.py
```diff
@@ -1058,6 +1064,12 @@ def test_UnWrapFlag_residues(self, ag, ref_Unwrap_residues):
         assert_almost_equal(ag.moment_of_inertia(unwrap=True, compound='residues'), ref_Unwrap_residues['MOI'], self.prec)
         assert_almost_equal(ag.asphericity(unwrap=True, compound='residues'), ref_Unwrap_residues['Asph'], self.prec)
 
+    def test_UnWrapFlag_residues_unordered(self, unordered_ag, ref_Unwrap_residues):
```
jbarnoud, line 1067: This could be parametrized so each AG method is tested separately.

testsuite/MDAnalysisTests/core/test_atomgroup.py
```diff
@@ -1058,6 +1064,12 @@ def test_UnWrapFlag_residues(self, ag, ref_Unwrap_residues):
...
+        assert_almost_equal(unordered_ag.center_of_geometry(unwrap=True, compound='residues'), ref_Unwrap_residues['COG'], self.prec)
```
jbarnoud, line 1068: These lines could be split.

jbarnoud (changes requested):
There are a couple of changes of indentation for split lines that are make the linter complain. Other than that there only is a small change on one test that I think would improve the PR.

testsuite/MDAnalysisTests/core/test_atomgroup.py
mnmelo: I rReally just kept them long to conform to how all other such tests are. I can split (shall I also split the analogous ones in neighboring tests?)

testsuite/MDAnalysisTests/core/test_atomgroup.py
mnmelo: You suggest also parameterizing the method? Like this?
```python
@pytest.mark.parametrize('method_name', ('center_of_geometry', 'center_of_mass',
                                         'moment_of_inertia', 'asphericity'))
def test_UnWrapFlag_residues_unordered(self, unordered_ag, ref_Unwrap_residues, method_name):
    method = getattr(unordered_ag, method_name)
    # requires that the keys of ref_Unwrap_residues are changed to match the method names
    assert_almost_equal(method(unwrap=True, compound='residues'),
                        ref_Unwrap_residues[method_name], self.prec)
```
I'm all for it, but several more tests are begging for the same cleanup. It seems a targeted refactoring effort would make more sense. If you think that at least new tests should be parameterized as above, I'm cool with having it thus.

mnmelo:
In the end, most of the non-parameterization or lint in the tests is to conform to the existing standard of neighboring tests. I'll gladly introduce these tests as exceptions and examples to follow from now on, but I think a refactor should be its own PR.

jbarnoud (approved):
Refactoring the tests is out of scope for this PR and I agree there is some value in keeping internal consistency. I openned an issue about the cleaning of the tests in that file: #2995.

PR #2998, opened 2020-10-19
Smoothen analysis class docstrings

PicoCentauri:
While @joaomcteixeira and I are working on the MDA command line interface, I found several (possible) improvements to the analysis class docstrings.

* Added more modules to the `__all__` list
* Changed `: ` -> ` : ` to follow numpydoc docstring guide
* removed default values from docstring
* removed trailing spaces and empty lines

PR Checklist
------------
 - [ ] Tests?
 - [x] Docs?
 - [ ] CHANGELOG updated?


package/MDAnalysis/analysis/encore/clustering/affinityprop.c
```diff
@@ -0,0 +1,9163 @@
+/* Generated by Cython 0.29.21 */
```
jbarnoud, line 1: This file is not supposed to be checked into git.

jbarnoud (commented):
These are very simple changes, but they are very welcome as well. There is one cython-generated C file that got added in the repo and that should be removed, though.

About the default, we have not been consistent in how we deal with default values in the docstrings. They are places where we write them explicitly between square brackets, places where we do not write them explicitly, and places where we write them using a different format. I think it is worth having a discussion about this (ping @MDAnalysis/coredevs) before we remove them.

I see pros and cons about having them:

* pro: as a user, I like knowing the default value for an argument and sometimes it is unpleasant to have to parse the signature to get it (many arguments,  very long docstring so the signature is far from the description of the parameters, obfuscated defaults, arguments in **kwargs...)
* con: it can easily not match the value in the code

package/MDAnalysis/analysis/encore/clustering/affinityprop.c
PicoCentauri: I removed the file.

PicoCentauri:
Thanks for the comment @jbarnoud. The generated `.c` file is gone.

Let's wait for the reply of the core devs. I removed them because for me it seems that there are more definitions only having the default values in the function or class definitions. The [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html#sections) docu are also not very clear how to handle them

> Optional keyword parameters have default values, which are displayed as part of the function signature. They can also be detailed in the description [...]

IAlibay:
> About the default, we have not been consistent in how we deal with default values in the docstrings. They are places where we write them explicitly between square brackets, places where we do not write them explicitly, and places where we write them using a different format. I think it is worth having a discussion about this (ping @MDAnalysis/coredevs) before we remove them.
> 
> I see pros and cons about having them:
> 
> * pro: as a user, I like knowing the default value for an argument and sometimes it is unpleasant to have to parse the signature to get it (many arguments,  very long docstring so the signature is far from the description of the parameters, obfuscated defaults, arguments in **kwargs...)
> * con: it can easily not match the value in the code

I have to say, I'm equally torn between the pro/cons here. I believe there were discussions about this between @orbeckst and @lilyminium in the past (with either the HOLE2 or new density code I think?). 

At the end of the day, as long as we can have uniformity in the docs, it's probably the best outcome. If removing the defaults is the fastest/safest means of ensuring that (which looks like it might be the case), then I'm happy to vote for it.

package/MDAnalysis/analysis/hole2/hole.py
```diff
@@ -1248,13 +1241,13 @@ def bin_radii(self, frames=None, bins=100, range=None):
...
+        frames : int or iterable of ints, optional
             Profiles to include by frame. If ``None``, includes
-            all frames. Default: ``None``
-
-        bins: int or iterable of edges, optional
-            If bins is an int, it defines the number of equal-width bins in the given range. If bins is a sequence, it defines a monotonically increasing array of bin edges, including the rightmost edge, allowing for non-uniform bin widths. Default: 100
-
+            all frames.
+        bins : int or iterable of edges, optional
+            If bins is an int, it defines the number of equal-width bins in the given range.
```
IAlibay, line 1248: Just a reminder to deal with the PEP8 issues (see pep8speaks)

orbeckst:
If we strip the defaults from the text description then we need to look carefully at any cases where we use `**kwargs` and if necessary make the kwargs explicit (which is better practice anyway).

When the text explains what the default does (e.g., "The default `None` selects all atoms.") then this explanation needs to remain. It's really important for docs to convey the _meaning_ and _intentions_ as this is often not even clear from looking at the code itself.

Overall, I see a slight advantage in clarity and maintainability if we remove "defaults in square brackets" if that's all there is -- this is information that can be kept [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). I just would not cut everything if real information is being lost.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -280,9 +280,9 @@ def __init__(self, universe,
             universe topology does not contain bonding information
         d_a_cutoff : float (optional)
             Distance cutoff for hydrogen bonds. This cutoff refers to the D-A distance. [3.0]
-        d_h_a_angle_cutoff: float (optional)
+        d_h_a_angle_cutoff : float (optional)
```
orbeckst, line 283: forgot to remove defaults -- the numbers in square brackets!

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -280,9 +280,9 @@ def __init__(self, universe,
             universe topology does not contain bonding information
         d_a_cutoff : float (optional)
             Distance cutoff for hydrogen bonds. This cutoff refers to the D-A distance. [3.0]
```
orbeckst, line 282: remov default

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -280,9 +280,9 @@ def __init__(self, universe,
...
-        d_h_a_angle_cutoff: float (optional)
+        d_h_a_angle_cutoff : float (optional)
             D-H-A angle cutoff for hydrogen bonds, in degrees. [150]
-        update_selections: bool (optional)
+        update_selections : bool (optional)
```
orbeckst, line 285: remove default [True]

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
orbeckst: There are more instances of missed defaults in the docs here.

package/MDAnalysis/analysis/pca.py
```diff
@@ -294,19 +294,19 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
 
         Parameters
         ----------
-        atomgroup: MDAnalysis atomgroup/ Universe
+        atomgroup : MDAnalysis atomgroup/ Universe
             The atomgroup or universe containing atoms to be PCA transformed.
-        n_components: int, optional
-            The number of components to be projected onto, Default none: maps
+        n_components : int, optional
+            The number of components to be projected onto, Default None: maps
```
orbeckst, line 300: Rewrite as 
```suggestion
            The number of components to be projected onto. The default value ``None`` maps
```

package/MDAnalysis/analysis/pca.py
```diff
@@ -294,19 +294,19 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
             onto all components.
-        start: int, optional
+        start : int, optional
```
orbeckst, line 302: rewrite, see above for example

package/MDAnalysis/analysis/pca.py
```diff
@@ -294,19 +294,19 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
             The frame to start on for the PCA transform. Default: None becomes
             0, the first frame index.
-        stop: int, optional
+        stop : int, optional
```
orbeckst, line 305: rewrite, see above for example. Also replace n_frames with explicit "the total number of frames in the trajectory".

package/MDAnalysis/analysis/pca.py
```diff
@@ -294,19 +294,19 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
             Frame index to stop PCA transform. Default: None becomes n_frames.
             Iteration stops *before* this frame number, which means that the
             trajectory would be read until the end.
-        step: int, optional
+        step : int, optional
```
orbeckst, line 309: rewrite, if set to ``None`` (the default) then every frame is analyzed (i.e., same as ``step=1``) -- makes things clearer

orbeckst (changes requested):
Generally I'm in favor and thank you for cleaning up!

Found a few missed ones  + suggestions for making things clearer

lilyminium:
I agree with @orbeckst, I am in favour of not repeating default values in the docs.

> pro: as a user, I like knowing the default value for an argument and sometimes it is unpleasant to have to parse the signature to get it (many arguments, very long docstring so the signature is far from the description of the parameters, obfuscated defaults, arguments in **kwargs...)
    con: it can easily not match the value in the code

IMO this con outweighs the pros. 

> if necessary make the kwargs explicit (which is better practice anyway).

And also that in most cases it is much better to just make the kwargs explicit.

package/MDAnalysis/analysis/hole2/hole.py
PicoCentauri: Thanks. I fixed the lengths

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
PicoCentauri: Done

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
PicoCentauri: Done

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
PicoCentauri: Done

package/MDAnalysis/analysis/pca.py
PicoCentauri: Done

package/MDAnalysis/analysis/pca.py
PicoCentauri: Done

package/MDAnalysis/analysis/pca.py
PicoCentauri: Done

package/MDAnalysis/analysis/pca.py
PicoCentauri: Why didn't we set the default to 1?

orbeckst: Because using `None` as step = 1 is the behavior for slicing and range so we keep the user interface "pythonic".

orbeckst (approved):
There's one open suggestion regarding `step` but either way, this looks good. Thanks @PicoCentauri .

package/MDAnalysis/analysis/pca.py
orbeckst: @PicoCentauri do you want to improve this doc string, too, or are we squash-merging as is?

orbeckst:
I am going to squash-merge the PR. The failures on Azure are widely varying timings for some of the hypothesis-based tests.

orbeckst:
Thank you @PicoCentauri and @joaomcteixeira !

Issue #2423, opened 2019-12-22
PDB writer ignores the `elements` topology attribute


jbarnoud:
Related to #2422 and #2421 

**Expected behavior**

If the `elements` topology attribute is present in an AtomGroup, it is used to write the element field of the PDB file instead of the element being guessed.

**Actual behavior**

The `elements` topology attribute is ignored by the PDB writer.

**Code to reproduce the behavior**

``` python
import io
import MDAnalysis as mda
from MDAnalysis.tests.datafiles import PDB

u = mda.Universe(PDB)
u.add_TopologyAttr('elements', ['Zn'] * len(u.atoms))

output_io = io.StringIO()
output = mda.lib.util.NamedStream(output_io, 'plop.pdb')
u.atoms.write(output, file_format='PDB')
print(output_io.getvalue())
```

Columns 77 and 78 contains element symbols guessed from atom names instead of `Zn` as specified in the `elements` attribute.

**Currently version of MDAnalysis**

- Which version are you using? 0.20.2-dev0 4adbd6c01
- Which version of Python (`python -V`)?  `'3.8.0 | packaged by conda-forge | (default, Nov 22 2019, 19:11:38) \n[GCC 7.3.0]'`
- Which operating system? Ubuntu 18.04

Issue #2918, opened 2020-08-25
On the guessing of atom types (mostly XYZParser)


IAlibay:
Partially related to #2553
Also one of @lilyminium's other issues but I can't seem to find it.

## Is your feature request related to a problem? ##

For a lot of topology parsers, our default stance for atom types is to use `guess_types` when atom types are not available, even when elements are present, for example:
https://github.com/MDAnalysis/mdanalysis/blob/45e56e8314c278e3eb98ed7a6029b74e7435e8be/package/MDAnalysis/topology/XYZParser.py#L94-L105

What doesn't make sense to me here is that as far as I understand it, `guess_types` is essentially an inaccurate way of guessing elements, and well if elements are present, shouldn't we just default to elements normally?

## Describe the solution you'd like ##

I think we've already agreed that guessing elements shouldn't be a thing by default, I don't think we came to an agreement with what to do with atom types. But assuming we want to always have atom types around, I would suggest that at the very least in cases where elements are present (e.g. XYZParser), atom types are not guessed if elements are present.

## Describe alternatives you've considered ##

We just keep things as-is.

IAlibay:
Other affected parsers: FHIAIMSParser, PDBParser, RDKitParser

orbeckst:
Your approach of using elements if they are present to determine elements sounds eminently sensible.

I don't remember why we should ensure that there are always types around. With a lot of hindsight (and the "new" topology system) it seems better to only have the attributes that are present. 

Maybe then have a bunch of guessers that users can use if they want to, along the lines of "guess_missing_types()", "guess_missing_elements()" etc... 

In the dim and distant past the philosophy might have been to magically make everything™ work but growing up (and being faced with so many different formats and use cases) one stops believing in magic and just starts putting responsibility on the users themselves as they (should) know their data best.

IAlibay:
> I don't remember why we should ensure that there are always types around. With a lot of hindsight (and the "new" topology system) it seems better to only have the attributes that are present.

Agreed, in the case of atom types though, we might need to be clear on what they actually are. I've always seen them as "force field" atom types (defaulting to elements in some cases), however it looks like the current definition is: "atom name, atom element, or force field atom type" (https://userguide.mdanalysis.org/1.0.0/topology_system.html#topology-attributes). If we were to make atom types non-necessary for universes it would make sense to make atom names a requirement instead?

[...]

PR #3001, opened 2020-10-20
Adds PDB element writing and fixes reading

### LINKED ISSUE #2422: PDB parser ignores the element column
Related to #2420 

**Expected behavior**

The penultimate field of the `ATOM` and `HETATM` records is the element symbol: <http://www.wwpdb.org/documentation/file-format-content/format33/sect9.html#ATOM>. The PDB reader should read that field when present and expose it through the `elements` topology attribute.

**Actual behavior**

The element field of a PDB file is ignored. This means the element must be guessed even when provided.

``` python
import MDAnalysis as mda
from MDAnalysis.tests.datafiles import PDB_full

u = mda.Universe(PDB_full)
u.atoms.elements
```

```
<ipython-input-61-b55fd0543db8> in <module>
      4 
      5 u = mda.Universe(PDB_full)
----> 6 u.atoms.elements

~/dev/mdanalysis/package/MDAnalysis/core/groups.py in __getattr__(self, attr)
   2291             except selection.SelectionError:
   2292                 pass
-> 2293         raise AttributeError("{cls} has no attribute {attr}".format(
   2294             cls=self.__class__.__name__, attr=attr))
   2295 

AttributeError: AtomGroup has no attribute elements
```

- Which version are you using? 0.20.2-dev0 4adbd6c01
- Which version of Python (`python -V`)?  `'3.8.0 | packaged by conda-forge | (default, Nov 22 2019, 19:11:38) \n[GCC 7.3.0]'`

IAlibay:
Completes/supersedes #2442 Fixes #2422 #2423 

 - PDB parser now allows for partial element parsing, setting an empty record if the element is not recognised.
 - PDB writer now uses the elements attribute instead of guessing.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


IAlibay:
There was quite a lot of discussion on #2422 about what to do when there's empty/unknown elements records. 

The approach I've taken here reflects the discussions in #2422, and other places where we've taken the conservative stance that unknown elements are not assigned an element record on reading, and therefore would not have an elements record written down either.

I'm sure many will have strong opinions here, including but not limited to; @lilyminium, @RMeli, and @richardjgowers.

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -1055,7 +1059,7 @@ def get_attr(attrname, default):
             vals['occupancy'] = occupancies[i]
             vals['tempFactor'] = tempfactors[i]
             vals['segID'] = segids[i][:4]
-            vals['element'] = guess_atom_element(atomnames[i].strip())[:2]
+            vals['element'] = elements[i][:2]
```
IAlibay, line 1062: Should this have an `upper()` call? I've seen PDB files with both mixed case or just all upper case. We enforce UpperLower within MDA, but it would be best to write out in the manner that's consistent, thoughts?

RMeli:
IMHO, not assigning unknown elements is very sensible. I'd rather have the code fail cleanly because the elements attribute is missing instead of getting erroneous results because of an incorrect guess.

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -1033,6 +1033,9 @@ def _write_timestep(self, ts, multiframe=False):
            When only :attr:`record_types` attribute is present, instead of
            using ATOM_ for both ATOM_ and HETATM_, HETATM_ record
            types are properly written out (Issue #1753).
+           Writing now only uses the contents of the elements attribute
+           instead of guessing by default. If the elements are missing,
+           empty records are written out (Issues #2423).
```
lilyminium, line 1038:
```suggestion
           empty records are written out (Issue #2423).
```

package/MDAnalysis/coordinates/PDB.py
lilyminium: Yeah I like `.capitalize()`. That being said, given that we already validate and capitalize elements on input, if the user reeeeeeally wants bespoke capitalisation then maybe we should let them.

package/MDAnalysis/topology/PDBParser.py
```diff
@@ -35,15 +35,13 @@
...
-TODO:
-    Add attributes to guess elements for non-physical or missing elements
 
 
 .. Note::
 
    The parser processes atoms and their names. Masses are guessed and set to 0
    if unknown. Partial charges are not set. Elements are parsed if they are
-   valid.
+   valid, if partially missing or incorrect empty records are assigned.
```
lilyminium, line 44:
```suggestion
   valid. If partially missing or incorrect, empty records are assigned.
```

lilyminium (changes requested):
Looks good, a couple nitpicks. So the Atomtypes attribute remains unchecked or guessed elements, and the Elements attribute is the same information run through a sensibility filter?

package/MDAnalysis/coordinates/PDB.py
IAlibay: So we normalize all the inputs to be `capitalize()`, which is where this gets complicated, I prefer then converting back to `all caps/upper()` elements because I think we see these a bit more often in PDBs, but if users had a specific preference then we'd be overriding that :/

IAlibay:
> Looks good, a couple nitpicks. So the Atomtypes attribute remains unchecked or guessed elements, and the Elements attribute is the same information run through a sensibility filter?

So this is an interesting one that links back to #2918, personally I wouldn't ever guess on read, but we have this default need for atomtypes, and what an atom type isn't properly defined. Given that atom types can be "atom name, atom element, or force field atom type" maybe the answer here is to assign atom types to the raw input atom name?

My only worry here is how much would break if we did this (I have no idea for how widely used atomtype is downstream).

package/MDAnalysis/coordinates/PDB.py
lilyminium: It seems that [all-caps used to be the spec](https://lists.sdsc.edu/pipermail/pdb-l/2004-January/001397.html), and here is [an example of capitalized symbols causing issues](https://gitlab.com/ase/ase/-/issues/78) in NGLView -- I vote for upper case unless/until someone specifically raises an issue about it.

lilyminium:
I would also be concerned that changing `atomtype` to name would break something, especially as elements weren't in our PDB format until recently iirc. Let's just keep as is until we figure out what to do with the Atomtype attribute as a whole?

testsuite/MDAnalysisTests/coordinates/test_pdb.py
```diff
@@ -45,6 +45,15 @@
 IGNORE_NO_INFORMATION_WARNING = 'ignore:Found no information for attr:UserWarning'
 
 
+@pytest.fixture
+def dummy_universe_without_elements():
```
jbarnoud, line 49: Could you look at minimizing the number of warnings that gets issued by your new tests? I used 2 different approaches for that in #2886: filling all the required attributes in the fixture or filtering the warnings in the tests.

IAlibay: 3d7857d should have done the trick, only warnings left are parmed's ABCs from collections import.

package/MDAnalysis/coordinates/PDB.py
IAlibay: That's convincing enough for me, I've changed it to upper in bd220dc :)

lilyminium (approved):
LGTM, thank you @IAlibay and sorry for the delay! 😶

PR #3004, opened 2020-10-23
Temporarily disable nsgrid (Issue #2930)

### LINKED ISSUE #2930: disable nsgrid neighbor search
## Expected behavior ##

neighbor search with capped_distance returns reliable and consistent results

## Actual behavior ##

It doesn't.

* #2919 
* #2229 
* #2345
* #2670

See https://github.com/MDAnalysis/mdanalysis/issues/2229#issuecomment-529910570 for detailed analysis of problems.

## Course of action
Disable nsgrid-based neighbor search until we have a stable solution. Use PBC-pkdtree instead.

## Impact
* slower distance-based searches and analysis (e.g. RDF)
* correct results in all cases

See linked issues.

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) 1.0.x, 2.0.0-dev
- Which version of Python (`python -V`)? any
- Which operating system? any

IAlibay:
Fixes #2930 

 - Temporarily replaces nsgrid with pkdtree
 - Adds note on the use of nsgrid

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/lib/nsgrid.pyx
```diff
@@ -70,6 +70,15 @@ not reflect in the results.
 
 .. versionadded:: 0.19.0
 
+Important Note
```
IAlibay, line 73: I have only left a note, would a warning be useful too? (not really sure how you do warnings in a cython file).

package/CHANGELOG
```diff
@@ -20,6 +20,7 @@ The rules for this file:
   * 1.0.1
 
 Fixes
+  * Disable NSGrid in favour fo PKDTree (Issue #2930)
```
orbeckst, line 23: I think we need to state that the nsgrid implementation could lead to wrong results (similar to what you have in you versionchanged messages).

package/MDAnalysis/lib/nsgrid.pyx
orbeckst: Do you mean `import warnings`? I _think_ that should work.

Can you also add a warning in the markup
```rest
.. warning::

   The current nsgrid code can return incorrect values. Do not use for production use.
   See issue `#2930<https://github.com/MDAnalysis/mdanalysis/issues/2930>`_ for further context.
```
(I hope that's the right [syntax for inline reST links](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#external-links)).

orbeckst (approved):
Thanks @IAlibay ! Looking good to me; I don't know if there are any more places where it might show up, though.

I assume the actual tests for the code are still in place?

I have minor comments/suggestions for docs – please see inline.

IAlibay:
> I don't know if there are any more places where it might show up, though.

So I can't seem to find any more cases of FastNS being imported (just doing a recursive grep for files in `package`

Outside of `lib.__init__`, the only other instance of `nsgrid` is in `setup.py`'s `extensions` (from what I gather it's just the cythonize stuff).

> I assume the actual tests for the code are still in place?

So the tests that have been affected here are those that invoke capped_distance or self_capped_distance, so it lowers coverage of nsgrid a little bit, but test_nsgrid is still there.

package/CHANGELOG
orbeckst: Also:
```python
"fo"[::-1]
```

IAlibay: I would claim that this is caused by writing docs at 2 am, but I'm pretty sure I make just as many typos during the day 🙄 

Should be fixed in ffc0a72 (along with more typos 😅 )

orbeckst:
@IAlibay please merge at your leisure (unless anyone else finds something, of course)

IAlibay:
> @IAlibay please merge at your leisure (unless anyone else finds something, of course)

Will do, just adding those warnings and this should be good :)

orbeckst:
Apart from my own stupidity (see #3011 ...) this is the last real thing to be completed for 1.0.1 to ship.

IAlibay:
Ok, changed the docs to have proper `.. warning` entries, and added a userwarning on FastNS.

pinging @MDAnalysis/coredevs can someone give this a quick re-review so we can merge with the aim of prepping the 1.0.1 release tomorrow (@richardjgowers nicely suggested walking those interested to learn through it).

orbeckst:
The tests fail because the fastns warning is not always raised. Perhaps we have to ensure that it is _always_ raised and never muted?

IAlibay:
> The tests fail because the fastns warning is not always raised. Perhaps we have to ensure that it is _always_ raised and never muted?

I honestly have no idea why this isn't working. It works fine if I build locally on my end 😕 The fact that azure builds and the other CIs don't probably mean there's an option we are/aren't turning on somewhere that disables this warning but I can't seem to work out what.

orbeckst:
If we make it something that's not a `UserWarning` such as `RuntimeWarning`???

Alternatively, can you enforce raising warnings in your test?  Some of the text in https://docs.python.org/3.8/library/warnings.html#testing-warnings and above says that any of the fancy temporarily changing warning behavior is not well defined in multi-threaded applications. I think parallel pytest uses different processes so that should be ok. There's also a note

> One thing to be aware of is that if a warning has already been raised because of a once/default rule, then no matter what filters are set the warning will not be seen again unless the warnings registry related to the warning has been cleared.

which suggests that we might have to globally set our warnings filter at the beginning of the tests to ensure that we always see our warnings.

IAlibay:
Now this is quite suspicious, `d4499f5` should have caused a ton of CI tests to fail (as per [the azure tests](https://dev.azure.com/mdanalysis/mdanalysis/_build/results?buildId=1126&view=logs&j=1cdae27b-58e8-5c64-01e2-519b6cdf4b73&t=c13e15da-7c65-5b3c-5c17-f011eed8e1e8)( because that `RuntimeError` should have been encountered on class construction (unless I typo-ed somewhere).

The fact that only the nsgrid test for the `RuntimeWarning` [failed](https://travis-ci.com/github/MDAnalysis/mdanalysis/jobs/415184272) is very puzzling.

Is it possible that somehow CI isn't recompiling the cython files? Pinging @richardjgowers and @tylerjereddy who might know better about the peculiarity of our Travis setup & cython compiling.

orbeckst:
Following https://docs.python.org/3/library/warnings.html#overriding-the-default-filter, perhaps we should add an "always" at the start of our tests in [`testsuite/MDAnalysisTests/__init__.py`](https://github.com/MDAnalysis/mdanalysis/blob/develop/testsuite/MDAnalysisTests/__init__.py)
```python
import sys

if not sys.warnoptions:
    import os, warnings
    warnings.simplefilter("always") # Change the filter in this process
    os.environ["PYTHONWARNINGS"] = "always" # Also affect subprocesses
```

(I would have assumed that `pytest` is already doing something like this but who knows?)

This should ensure that warnings are never counted as "have already occurred" and hopefully get around the problem
>  One thing to be aware of is that if a warning has already been raised because of a once/default rule, then no matter what filters are set the warning will not be seen again unless the warnings registry related to the warning has been cleared.

One thing to check would be if `pytest.warns()` can be made to reset this registry (with [warnings.resetwarnings()](https://docs.python.org/3/library/warnings.html#warnings.resetwarnings) to ensure that a warning is always shown. Or use [warnings.resetwarnings()](https://docs.python.org/3/library/warnings.html#warnings.resetwarnings) in your own test to check???

IAlibay:
@orbeckst the problem here isn't that the warning isn't being emitted, it's that the new cython code isn't being compiled to c to begin with.

Looking at differences between azure (which works) and the other two CI methods, I suspect the problem is how we call setup.py. @richardjgowers' suggestion here was to try just check the new c files in, however I'd prefer fix CI to begin with (if possible).

orbeckst:
If we are not rebuilding the C files then _what_ are we are actually testing?? 

How did you find out that the files are not being rebuild?

orbeckst:
The Travis output only contains ` Running setup.py develop for MDAnalysis` for the step that we're interested in. Do you want to try and replace
https://github.com/MDAnalysis/mdanalysis/blob/2ef90f2e88d62d1109cae94b2fc10216be813853/.travis.yml#L31
with
```
BUILD_CMD="(cd package && python setup.py install) && (cd testsuite/ && python setup.py build)"
```
to have the same across our CI providers? (I have no idea why we used a developer install there anyway...)

IAlibay:
Apologies, I've not had time to really spend on this. I'm going to open a separate test PR targeting `develop` just to see if I can isolate the cause of all this...

IAlibay:
Ok, so we know that it works on #3024, so it's `develop` specific, now to dig into why..

IAlibay:
So I've spent a fair amount on this and I'm still baffled that what's in `master` that causes this behaviour. Given that we really should have pushed 1.0.1 a month ago, I'm going to ping @MDAnalysis/coredevs here. If anyone can come up with a quick solution this would be great.

The only option I can think of here is to push the new `c` files generated by cython, but I'm not sure if that's such a great idea long term (especially if we keep working on `master` towards future 1.0.x releases).

orbeckst:
Thank you for all the hours you put into trying to figure it out!!!

Did you confirm locally that it works as expected? If so, in order to get 1.0.1 out I suggest you make the corresponding text XFAIL or XPASS (?)  in 1.0.1, add a comment describing the problem and reference the discussion here. Let's move forward.

IAlibay:
So finally yielding on fixing CI and following @richardjgowers' wisdom does the trick here.

It seems like we'll have to push re-generated `.c`/`.cpp` files on `master` for now.

Maybe the option is to create a separate `develop-1.x` branch for future releases (if needed), where we can remove all these files like we do on `develop`?

I've reverted all the CI yml changes for now since they had no impact.

@orbeckst does this work for you? If so we can merge & look to ship 1.0.1 soon.

orbeckst:
Sorry, behind on things....

Let's move forward with the way you suggest. Using master for 1.x is dirty anyway, better be explicit. 🚀

lilyminium:
Might the failure to recreate .c/pp files change now with GH actions?

IAlibay:
Ok, this should be good to merge?

orbeckst:
The sooner the better! 🚢

lilyminium (approved):
LGTM, thank you @IAlibay for pushing this through! 💫

PR #3005, opened 2020-10-23
Optimized and refactored common compound code.

### LINKED ISSUE #3000: Use of compounds involves excessive sorting and fancy-indexing of AtomGroups and compound indices.
Hi,
I realized that there is potential for optimization of most `AtomGroup` methods that implement compounds. Currently, compounds are implemented by cleverly splitting into compound groups of the same size, to allow vectorization at least on a per-size basis. This was a great idea! (`git blame` tells me kudos go to @kain88-de and @richardjgowers).

Current improvable aspects:
- By default, `compound_indices` are always being sorted, to be robust to cases where compounds in an `AtomGroup` are not contiguous. This is slow and unneeded for the most common use case of contiguous compounds. A contiguity check is cheap (this was already implemented for `AtomGroup.center` only, in the resolution of #2984 and #2992);
- Most compound-using methods currently do their own splitting. This is inefficient code- and readability-wise, and should be concentrated in a dedicated method;
- Splitting masks for a given compound type need to be calculated only once per `AtomGroup` and could be cached. This would positively impact a number of methods, especially when iterating over frames (with due care to invalidate such caches if the user changes the `AtomGroup`'s compound indices). Even within a single calculation this can already be impactful: when using `AtomGroup.center` with `unwrap=True` indices are split once to do the unwrapping and then again to do the center calculation. 

I'd like to work on such refactoring, including some benchmarking for performance, but I need input so I don't set off on an effort that will either run against current code philosophy or is too large for anyone to review properly.

Currently needed pointers:
- Should this be a project?
- What is the recommended way of implementing caching?
- This will enlarge the memory footprint of an `AtomGroup` due to caching. I expect about 1x-to-2x the size of `._ix` per cached compound-type splitting. Is this acceptable?
- What test cases should be used for performance benchmarking? This will of course be more visible for large `AtomGroups` with many compounds, but I wouldn't want to harm the small-case performance. I thought about systems with a combination of the following characteristics: number of atoms `(100, 10k, 100k)`, number of compounds `(1, 10, 100)`, heterogeneity of compound sizes `(all_same_size, all_different)`, contiguity `(ordered, unordered)`;
- What should be the scope of the enhancements? For instance, I stumbled upon this while trying to speed up `AtomGroup.unwrap` (#2982), and there are `unwrap`-specific aspects that can be tweaked to take advantage of a centralized splitting solution. Should those be left out for an `unwrap`-specific effort?
- Following from the previous point: what other aspects do you think could be optimized along with centralized splitting?
- Who wants to join in on the fun? 😀

mnmelo:
Fixes #3000

 - Implements the ideas in #3000;
 - Compound code in `core/group.py` refactored (more DRY now);
 - Caching functionally implemented, but cache invalidation still to do.

Performance benchmarks, in the discussion of #3000, are quite promising, especially with caching.

## Problem
Cached compound masks work great and can give large speedups, but invalidating them is tough: if I have an `AtomGroup` with cached masks for segments and the user moves a residue between segments those cached masks should no longer be valid, but how to notify an AtomGroup of something occurring at the `ResidueGroup`/`SegmentGroup` level? 

Pointers appreciated.

PR Checklist
------------
 - [ ] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/benchmarks/benchmarks/ag_methods.py
+++ b/benchmarks/benchmarks/ag_methods.py
@@ -324,6 +324,62 @@ class AtomGroupAttrsBench(object):
         self.ag[:2].bond
 
 
+class CompoundSplitting(object):
+    """Test how fast can we split compounds into masks and apply them
+    
+    The benchmark used in Issue #3000. Parameterizes multiple compound number
+    and size combinations.
+    """
+    
+    params = [(100, 10000, 1000000),  # n_atoms
+              (1, 10, 100),           # n_compounds
+              (True, False),          # homogeneous
+              (True, False)]          # contiguous
+
+    def setup(self, n_atoms, n_compounds, homogeneous, contiguous):
+        rg = np.random.Generator(np.random.MT19937(3000))
+
+        # some parameter screening for nonsensical combinations.
+        if n_compounds > n_atoms:
+            raise NotImplementedError
+
+        if n_compounds == 1 and not (homogeneous and contiguous):
+            raise NotImplementedError
+            
+        if n_compounds == n_atoms:
+            if not (homogeneous and contiguous):
+                raise NotImplementedError
+            compound_indices = np.arange(n_compounds)
+        elif homogeneous:
+            ats_per_compound, remainder = divmod(n_atoms, n_compounds)
+            if remainder:
+                raise NotImplementedError
+            compound_indices = np.tile(np.arange(n_compounds),
+                                       (ats_per_compound, 1)).T.ravel()
+        else:
+            compound_indices = np.sort(np.floor(rg.random(n_atoms)
+                                                * n_compounds).astype(np.int))
+                
+        unique_indices = np.unique(compound_indices)
+        if len(unique_indices) != n_compounds:
+            raise RuntimeError
+        
+        if not contiguous:
+            rg.shuffle(compound_indices)
+        
+        self.u = MDAnalysis.Universe.empty(n_atoms,
+                                           n_residues=n_compounds,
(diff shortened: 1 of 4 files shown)


mnmelo:
In #2376 I propose a solution to the cache invalidation problem, which is to cache at the `Universe` level.

richardjgowers (commented):
I like the DRY changes in this PR a lot.  I can't see where the caching is, has this been removed for now? If so I'm happy to merge these changes.

mnmelo:
I have dropped the caching for this PR, so it's more contained and easier to review. (Also, the caching invalidation mechanism is what is still up for discussion).

Gonna clean up these lingering bugs and update soon (should've tested locally 😬).

mnmelo:
Ok, this is ready for review. I guess codecov hit a roadbump or something because only `groups.py` was touched and diff coverage is almost 100% there.

mnmelo:
To get my feet wet in `asv` I also implemented the performance benchmarks I had posted in #3000. I really don't know what the status is on `asv` usage, so guidance is appreciated.

mnmelo:
Ping @richardjgowers. I think this is ready to merge. There seem to be some inconsistencies in Codecov computation triggering a fail there.
Also, this PR removes a couple hundred lines of repeated code, and the coverage may indeed go down just because of that, even if all remaining lines are as covered as before.

richardjgowers (approved):

richardjgowers:
@mnmelo thanks!

PR #3012, opened 2020-10-25
Cleanup and better parametrization of test_atomgroup.py

### LINKED ISSUE #2995: `testsuite/MDAnalysisTests/core/test_atomgroup.py` would benefit from some cleaning
`testsuite/MDAnalysisTests/core/test_atomgroup.py` has at least long lines and some tests that could be parametrized. See discussion in #2992.

mnmelo:
Fixes #2995

 - PEP8 cleanup of `test_atomgroup.py`
 - More `pytest` parametrization for more DRY tests (see discussion in #2992)

For more efficient parametrization I modified some of the tests to run on a parametrized fixture (in `TestUnwrapFlag`), which is something not used elsewhere in the same test file (but somewhat used throughout the testsuite). Please check.

Also, to allow efficient parametrization I modified a couple of fixtures that were just generating reference dicts to simply be class attributes (in `TestUnwrapFlag` and `TestPBCFlag`). I think these make more sense as attributes anyway, unless we fear some test may inadvertently modify the reference?

PR Checklist
------------
 - [x] Tests? (Duh...)
 - [ ] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/testsuite/MDAnalysisTests/core/test_atomgroup.py
+++ b/testsuite/MDAnalysisTests/core/test_atomgroup.py
@@ -993,7 +993,7 @@ class TestUnwrapFlag(object):
         group = universe.residues[0:3]
         group.wrap(inplace=True)
         if request.param:
-            rg = np.random.Generator(np.random.MT19937(121989))
+            rg = np.random.RandomState(121989)
             ndx = np.arange(len(group))
             rg.shuffle(ndx)
             group = group[ndx]
@@ -1632,7 +1632,7 @@ class TestAtomGroup(object):
                             err_msg="failed to update atoms 12:42 position "
                             "to new position")
 
-        rg = np.random.Generator(np.random.MT19937(121989))
+        rg = np.random.RandomState(121989)
         # create wrong size array
         badarr = rg.random((pos.shape[0] - 1, pos.shape[1] - 1))
         with pytest.raises(ValueError):
--- a/testsuite/MDAnalysisTests/core/test_atomgroup.py
+++ b/testsuite/MDAnalysisTests/core/test_atomgroup.py
@@ -1088,6 +1088,8 @@ class TestPBCFlag(object):
         universe = mda.Universe(TRZ_psf, TRZ)
         return universe.residues[0:3]
 
+    @pytest.mark.parametrize('pbc, ref', ((True, ref_PBC),
+                                          (False, ref_noPBC)))
     @pytest.mark.parametrize('method_name', ('center_of_geometry',
                                              'center_of_mass',
                                              'radius_of_gyration',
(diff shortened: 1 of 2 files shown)


mnmelo:
I'm checking that coverage change. Especially the 0.18% drop in `core/groups.py`.

testsuite/MDAnalysisTests/core/test_atomgroup.py
```diff
@@ -1680,13 +1592,15 @@ def test_dihedral_equals_improper(self, universe):
 
     def test_bond(self, universe):
         universe.trajectory.rewind()  # just to make sure...
```
richardjgowers, line 1682: While we're tidying, get rid of these rewinds

testsuite/MDAnalysisTests/core/test_atomgroup.py
```diff
@@ -984,189 +938,197 @@ class TestUnwrapFlag(object):
...
+            # We test unwrap=False as the default behavior
+            result = method()
+        assert_almost_equal(result, ref[method_name], self.prec)
 
 
 class TestPBCFlag(object):
 
     prec = 3
 
-    @pytest.fixture()
```
richardjgowers, line 1099: I'd probably leave these as fixtures speaking stylistically. - It's a better way to express that they're constant across tests. But we're hardly consistent with this (see entire coordinates module tests...)

testsuite/MDAnalysisTests/core/test_atomgroup.py
```diff
@@ -984,189 +938,197 @@ class TestUnwrapFlag(object):
...
+                                    [0.50622389, 0.18364489, 0.84262206]])
+    }
 
     @pytest.fixture()
     def ag(self):
         universe = mda.Universe(TRZ_psf, TRZ)
         return universe.residues[0:3]
 
-    def test_default(self, ag, ref_noPBC):
+    @pytest.mark.parametrize('method_name', ('center_of_geometry',
```
richardjgowers, line 1091: This is nicer!

richardjgowers (commented):
Looks like some good tidy up. The coverage diff always seems to have some noise in it. Best way to stay sane might be running before/after locally

testsuite/MDAnalysisTests/core/test_atomgroup.py
mnmelo: (Good you brought this up, because I had wrongly set up the reference parametrization in the tests using this. Fixed that now and actually was able to simplify the entire class to a single parametrized test.)
The issue is that if I want to parametrize the reference to be used in a test, it cannot be a method (I mean, it can, but then the test must call the method to obtain the reference, which is quite ugly IMO). As regular attributes this just works cleanly.

I think an alternative, if we really want to keep the `@pytest.fixture` _and_ the parametrization simplification, is to parametrize the fixture to access the `request` argument and figure out whether it's being used in a test with `pbc=True` or `pbc=False` and generate the appropriate reference dict accordingly. I'm unsure about the neatness-to-overkill ratio in such an approach, so I'll follow what you suggest is best.

testsuite/MDAnalysisTests/core/test_atomgroup.py
mnmelo: Done.

mnmelo:
Stabilized random number usage by using seeded RNGs.
One of the two RNG occurrences (in `TestAtomGroup.test_positions`) was used in a contrived function-inside-a-test pattern that I also simplified, without any ill effect (but maybe there is a reason for the complexity?)

mnmelo:
I did introduce a coverage drop because the seeded random shuffling to generate an unordered group turned out to be unlucky and preserved the order (facepalm). This was less improbable than it may seem because the group in question (a `ResidueGroup`) only has 3-elements. I had assumed the fixture was working with the underlying `AtomGroup` with tens of atoms. With only 3 elements the chance of hitting an ordered shuffling is 1 in 6, which I'd say is a borderline case of Murphy's Law.

mnmelo:
All is passing and I'm done cleaning the tests. Let me know if I should change anything else.

jbarnoud (approved):
This looks good to me. It is a massive omprovement compared to what we had before! Because it touches the tests of the AtomGroup and it is a quite essential/sensitive part of the codebase, I'll wait until tomorrow to merge so others can chip in if needed.

mnmelo:
Yea, I'd feel a bit more comfortable if more eyes could take a look. My additional checks were:
- Number of tests
This PR: `===================== 384 passed, 14223 warnings in 29.42s =====================`
develop: `===================== 341 passed, 14223 warnings in 27.99s =====================`
- Coverage diff of `test_atomgroup.py` only
(This PR is left, develop is right)
```
             Hits Hits
package/MDAnalysis/core/groups.py
  Line  836    1    0
package/MDAnalysis/core/selection.py
  Line  150    0    1
  Line  151    0    1
  Line  155    0    1
  Line  157    0    1
  Line 1311    0    1
  Line 1312    0    1
  Line 1368    0    1
  Line 1369    0    1
  Line 1370    0    1
```
The systematic parametrization ends up hitting an extra line in `core/groups.py`. As to `core/selection.py`, a syntax simplification I made
`select_atoms("resid 1:50 and not resname LYS and (name CA or name CB)")` -> `select_atoms("resid 1:50 and not resname LYS and name CA CB")`
ends up missing some token parsing lines, which I guess is ok because they're well covered elsewhere anyway and it's not the point of these tests (in the overall coverage, `core/selection.py` is just as covered as before).

PR #3013, opened 2020-10-27
Fixed import of several `__all__` lists

PicoCentauri:
Changes made in this Pull Request

When running `from MDAnalysis.analysis import * ` an `AttributeError` was raised since the `rdf_s` is missing.
If `sklearn` is not installed the `__all__` lists in `.analysis.encore` and `.analysis.encore.dimensionality_reduction` raised an error.

# PR Checklist

- [x] Tests?
- [ ] Docs?
- [x] CHANGELOG updated?

### DIFF
--- a/testsuite/MDAnalysisTests/test_api.py
+++ b/testsuite/MDAnalysisTests/test_api.py
@@ -24,7 +24,10 @@
 Test the user facing API is as we expect...
 """
 
+import importlib
+
 import MDAnalysis as mda
+import pytest
 
 
 def test_Universe():
@@ -49,3 +52,33 @@ def test_ResidueGroup():
 
 def test_SegmentGroup():
     assert mda.SegmentGroup is mda.core.groups.SegmentGroup
+
+
+@pytest.mark.parametrize('module',
+                         ('',
+                          '.analysis',
+                          '.analysis.data.filenames',
+                          '.analysis.distances',
+                          '.analysis.encore',
+                          '.analysis.encore.dimensionality_reduction',
+                          '.analysis.hbonds',
+                          '.analysis.hydrogenbonds',
+                          '.coordinates',
+                          '.core',
+                          '.lib._augment',
+                          '.lib._cutil',
+                          '.lib.formats',
+                          '.lib.pkdtree',
+                          '.topology',
+                          '.topology.tpr',
+                          '.visualization'
+                          )
+                         )
+def test_import_star(module):
+    try:
+        # This is really bad and ugly but I haven't found a way to run this
+        # using `importlib.import_module...`
+        exec("from MDAnalysis{} import *".format(module))
+    except (ImportError, AttributeError) as error:
+        # We want the test to fail, not to error
+        raise AssertionError(error)
--- a/testsuite/MDAnalysisTests/test_api.py
+++ b/testsuite/MDAnalysisTests/test_api.py
@@ -24,8 +24,15 @@
 Test the user facing API is as we expect...
 """
 
-import MDAnalysis as mda
+import importlib
+import os
+from pathlib import PurePath
+
 import pytest
+from numpy.testing import assert_equal
+
+import MDAnalysis as mda
+mda_dirname = os.path.dirname(mda.__file__)
 
 
 def test_Universe():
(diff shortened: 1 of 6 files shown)


jbarnoud:
Good catch. It would be worth having a test to avoid regressions, something very simple that just does `from MDAnalysis.analysis import *` and fails if there is an exception:

```python
def test_import_star():
    try:
        from MDAnalysis.analysis import *
    except ImportError as error:
        # We want the test to fail, not to error
        raise AssertError(str(error))
````

File: testsuite/MDAnalysisTests/test_api.py [hunk 1]
```diff
@@ -49,3 +52,33 @@ def test_ResidueGroup():
...
+                          '.lib._augment',
+                          '.lib._cutil',
+                          '.lib.formats',
+                          '.lib.pkdtree',
+                          '.topology',
+                          '.topology.tpr',
+                          '.visualization'
+                          )
+                         )
+def test_import_star(module):
```
PicoCentauri, line 77: If you know a better place for this test let me know.

testsuite/MDAnalysisTests/test_api.py
```diff
@@ -24,7 +24,10 @@
 Test the user facing API is as we expect...
 """
 
+import importlib
```
jbarnoud, line 27: Seems that you are not using this anymore.

testsuite/MDAnalysisTests/test_api.py
```diff
@@ -49,3 +52,33 @@ def test_ResidueGroup():
...
+                          '.topology',
+                          '.topology.tpr',
+                          '.visualization'
+                          )
+                         )
+def test_import_star(module):
+    try:
+        # This is really bad and ugly but I haven't found a way to run this
+        # using `importlib.import_module...`
+        exec("from MDAnalysis{} import *".format(module))
```
PicoCentauri, line 81: As written in the comment above this is bad practice but after testing a lot I haven't found a way to run something like `from module import *` using the `importlib` package... I'm open for ideas

PicoCentauri (commented):
Thanks @jbarnoud for the idea with a test case. I added a generic one checking all lists I found two more bugs.

PicoCentauri:
Hey @jbarnoud, sorry for the long delay. I created the testcase and found several other issues regarding the _all_ lists. I fixed them and hope the PR is fine.

Thanks for a comment. 🙂

testsuite/MDAnalysisTests/test_api.py
jbarnoud: Another approach is the one we use for datafiles: https://github.com/MDAnalysis/mdanalysis/blob/76d9ca1bd0b48cf692b6b3d720fa3a85dc13ef98/testsuite/MDAnalysisTests/utils/test_datafiles.py#L34 it checks that all the names in `__all__` are also in `dir`. This would avoid the `exec` that may break the coverage report.

testsuite/MDAnalysisTests/test_api.py
```diff
@@ -49,3 +50,33 @@ def test_ResidueGroup():
 
 def test_SegmentGroup():
     assert mda.SegmentGroup is mda.core.groups.SegmentGroup
+
+
+@pytest.mark.parametrize('module',
```
jbarnoud, line 55: My issue with such a list is that new modules are likely to not be added. What about walking the tree and find the modules? It might be done with importlib, but the simplest would be to check the directory tree for `__init__.py` files.

jbarnoud (commented):
I have some suggestions, but overall it is better than what we had before. Thank you for looking at it.

testsuite/MDAnalysisTests/test_api.py
PicoCentauri: Walking the complete package works fine and is a good idea,.

testsuite/MDAnalysisTests/test_api.py
```diff
@@ -49,3 +55,20 @@ def test_ResidueGroup():
...
 def test_SegmentGroup():
     assert mda.SegmentGroup is mda.core.groups.SegmentGroup
+
+def test_all_imports():
+    os.chdir(os.path.dirname(mda.__file__))
+
+    for root, dirs, files in os.walk("."):
+        if "__init__.py" in files:
+            submodule = ".".join(PurePath(root).parts)
+            submodule = "."*(len(submodule) > 0) + submodule
```
jbarnoud, line 65: You could extract the code from the beginning of this function up to here in a separate function that generate the list (or a generator? would that work with pytest?) to pass to `pytest.mark.parametrize` so that each module is tested separately. This way you can fix all the broken ones rather than fix one, rerun tests, fix the next broken one... This is just me becoming greedy here, though.

jbarnoud (commented):
Looks all good. I have one last optional comment in the review. This last thing is to make pep8speaks happy: https://github.com/MDAnalysis/mdanalysis/pull/3013#issuecomment-717965459

jbarnoud:
The tests do not pass, though. I did not follow very closely lately, is there a systematic failure going on at the moment?

lilyminium:
It looks like it mostly has to do with numpy deprecating `np.int` and `np.float`?

testsuite/MDAnalysisTests/test_api.py
PicoCentauri: Yes, good idea to split this into individual tests. This is also easy to implement. No greed needed :)

PicoCentauri:
I'm also a bit puzzled by the failing tests. I will take a look later.

IAlibay:
> I'm also a bit puzzled by the failing tests. I will take a look later.

Should be related to #3113, hopefully #3114 will fix it.

testsuite/MDAnalysisTests/test_api.py
```diff
@@ -49,3 +57,24 @@ def test_ResidueGroup():
 
 def test_SegmentGroup():
     assert mda.SegmentGroup is mda.core.groups.SegmentGroup
+
```
IAlibay, line 60: Please don't forget to check the pep8speak comments!

(from a quick look, needs an extra empty line here and at line 69, and line 79 is too long).

IAlibay:
@PicoCentauri if you wanna try merging in the latest develop, I think that _should_ fix the tests.

PicoCentauri:
Thanks @IAlibay for the fix. It worked for most of the tests. However, one ist still failing.

jbarnoud (approved):

Issue #3019, opened 2020-11-07
CI: h5py-related failures in develop branch Azure CI


tylerjereddy:
The Azure CI Windows jobs are failing on current `develop` branch/merges. The Azure logs tend to disappear after 10 days or so, so I'll summarize the types of failures:

- Python 3.7 32-bit: fails to `pip` install `h5py` version `3.0.0`
- Python 3.6, 3.7, 3.8 64-bit: 5 test suite failures in `TestH5MDReader`:

<details>

```
2020-10-30T20:10:03.6709718Z ================================== FAILURES ===================================
2020-10-30T20:10:03.6710719Z __________________________ TestH5MDReader.test_iter ___________________________
2020-10-30T20:10:03.6711428Z [gw1] win32 -- Python 3.8.6 C:\hostedtoolcache\windows\Python\3.8.6\x64\python.exe
2020-10-30T20:10:03.6711823Z 
2020-10-30T20:10:03.6712316Z self = <MDAnalysisTests.coordinates.test_h5md.TestH5MDReader object at 0x000001D67EDE0E50>
2020-10-30T20:10:03.6712867Z ref = <MDAnalysisTests.coordinates.test_h5md.H5MDReference object at 0x000001D67EDE0A90>
2020-10-30T20:10:03.6713508Z reader = <H5MDReader D:\a\1\s\testsuite\MDAnalysisTests\data\coordinates\test.h5md with 5 frames of 5 atoms>
2020-10-30T20:10:03.6713923Z 
2020-10-30T20:10:03.6714344Z     def test_iter(self, ref, reader):
2020-10-30T20:10:03.6714819Z >       for i, ts in enumerate(reader):
2020-10-30T20:10:03.6715149Z 
2020-10-30T20:10:03.6715554Z MDAnalysisTests\coordinates\base.py:292: 
2020-10-30T20:10:03.6716033Z _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
2020-10-30T20:10:03.6716783Z C:\hostedtoolcache\windows\Python\3.8.6\x64\lib\site-packages\mdanalysis-2.0.0.dev0-py3.8-win-amd64.egg\MDAnalysis\coordinates\base.py:1474: in __next__
2020-10-30T20:10:03.6717443Z     return self.next()
[...]

PR #3020, opened 2020-11-07
Write CONECT record only once in multi-MODEL PDB files

### LINKED ISSUE #3018: MDAnalysis.Writer exporting PDBs in bad format with bonds=True
## Expected behavior ##

Save each frame trajectory as a PDB with the contact information separate by the `ENDMDL` after each PDB.

## Actual behavior ##

The `ENDMDL` directive is wrote BEFORE the `CONECT` session, breaking the structure of the PDB.

Load a trajectory and export as with the `bonds=True , multiframe=True`. I am using:
complxx = u.select_atoms('segid PROT','segid LIG')
with MDAnalysis.Writer("output.pdb", bonds=True, multiframe=True) as W:
    for ts in u.trajectory:
        W.write(complxx)

- Which version of Python (`python -V`)? Python 3.8.6
- Which operating system? Ubuntu 20.04

RMeli:
Fixes #3018, as suggested by @IAlibay :

> If this is the correct interpretation, we should be applying CONECT in the same way we apply END, i.e. automatically on trajectory closing.

This PR allows to correctly write multi-frame PDB files with multiple calls to `write`. The behavior of `write_all_timesteps` was already correct and remains unchanged.

 - Write `CONECT` record just before the `END` record when calling `close()` 

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


IAlibay:
@RMeli I think the errors are h5py related (i.e. #3019) could you merge in the latest develop and see if that fixes things?

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -658,9 +658,11 @@ def __init__(self, filename, bonds="conect", n_atoms=None, start=0, step=1,
         self.first_frame_done = False
 
     def close(self):
-        """Close PDB file and write END record"""
+        """Close PDB file and write CONECT and END record"""
         if hasattr(self, 'pdbfile') and self.pdbfile is not None:
             if not self.has_END:
+                if hasattr(self, "obj"):  # Avoid problems with zero atoms
```
IAlibay, line 664: Is the no `self.obj` case covered in tests / if not how easy would it be to add a test?

IAlibay (changes requested):
Overall lgtm, just a couple of comments, mainly related to docs.

Let's see if CI returns green when you include the h5py changes.

package/MDAnalysis/coordinates/PDB.py
RMeli: Without such safeguard, `test_write_pdb_zero_atoms` was failing.

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -924,6 +930,9 @@ def write_all_timesteps(self, obj):
 
         .. versionchanged:: 0.11.0
            Frames now 0-based instead of 1-based
+
+        .. versionchanged:: 2.0.0
+           CONECT_ record moved to :fun:`close`
```
RMeli, line 935:
```suggestion
           CONECT_ record moved to :meth:`close`
```

package/MDAnalysis/coordinates/PDB.py
IAlibay: As a thought here, since we're already doing a `self.obj` check in `_write_pdb_bonds()` would it be cleaner / safer (for future changes) to instead have this check there?

IAlibay (changes requested):
couple of extra checks & questions, should be good after that.

testsuite/MDAnalysisTests/coordinates/test_pdb.py
```diff
@@ -810,6 +810,41 @@ def test_write_all_timesteps(self, multiverse, outfile):
...
+
+        u2 = mda.Universe(outfile)
+        assert_equal(len(u2.atoms), desired_group,
+                     err_msg="MultiPDBWriter trajectory written for an "
+                             f"AtomGroup contains {len(u2.atoms)} atoms, "
+                             f"it should contain {desired_group}")
+
+        assert_equal(len(u2.trajectory), desired_frames,
+                     err_msg=f"MultiPDBWriter trajectory written for an "
+                             "AtomGroup contains {len(u.trajectory)} "
```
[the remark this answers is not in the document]
RMeli: Nope, this is wrong. Well spotted!

package/MDAnalysis/coordinates/PDB.py
RMeli: Probably yes. I'll have a look and change it as you suggest.

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -658,9 +658,16 @@ def __init__(self, filename, bonds="conect", n_atoms=None, start=0, step=1,
         self.first_frame_done = False
 
     def close(self):
-        """Close PDB file and write END record"""
+        """
+        Close PDB file and write CONECT and END record
+
+
+        .. versionchanged 2.0.0
```
IAlibay, line 665:
```suggestion
        .. versionchanged:: 2.0.0
```

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -922,8 +930,12 @@ def write_all_timesteps(self, obj):
 
         will be writing frames 12, 14, 16, ...
 
+
         .. versionchanged:: 0.11.0
            Frames now 0-based instead of 1-based
+
+        .. versionchanged:: 2.0.0
+           CONECT_ record moved to :meth:`close`
```
IAlibay, line 938: I wonder if we should be a little bit more explicit, maybe in the module nodes, i.e. tell users specifically that CONECT records won't be written until they close the object (would users be likely to use the files before the objects have finished closing? thoughts?)

IAlibay (changes requested):
A docstring change and a question re: docs.

package/MDAnalysis/coordinates/PDB.py
RMeli: >would users be likely to use the files before the objects have finished closing?

Hopefully not, but I can be more explicit in the docstring for sure.

IAlibay: If possible? Maybe just a quick extra line after the bit about writing CONECT records in `notes`? I'll leave it to your discretion.

IAlibay (approved):
lgtm, thanks @RMeli !

PR #3021, opened 2020-11-07
MAINT, CI: pin Azure h5py

tylerjereddy:
* pin the version of `h5py` used in Azure Windows CI testing
because of gh-3019; CI failures were observed to start/coincide
with a recent release of `h5py`

Fixes #3019 (I hope)

PR #3022, opened 2020-11-08
MAINT: Travis/Appveyor pin h5py

tylerjereddy:
* pin the version of `h5py` used in Travis and Appveyor CI runs
because CI failures related to recent versions of this dependency
were observed in gh-3021 (where the same pinning was applied to Azure
CI)

* original issue was described in gh-3019

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -10,7 +10,7 @@ cache:
 environment:
     global:
         CONDA_CHANNELS: conda-forge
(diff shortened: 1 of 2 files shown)


tylerjereddy:
Ok, that gets the rest of the CI green when combined with gh-3021

IAlibay (approved):
Thanks @tylerjereddy !

PR #3025, opened 2020-11-09
Fix for h5py 3.0

### LINKED ISSUE #3023: CI; unpin h5py
As detailed in #3019 the 3.0 release of h5py was causing widespread test failures.

Currently CI is pinned to 2.10, however we should:

1. Investigate what caused the issue (is this an issue with MDA or h5py?)
2. Attempt to unpin

and / or

3. Define the pin in setup? (maybe under `extras_require`?)

yuxuanzhuang:
Fixes #3023

 - fix h5py format failure after version 3.0
- unpin h5py version

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


IAlibay:
Should we just be aiming at >3.0 for everything? (unless there's a reason not to?)

yuxuanzhuang:
> Should we just be aiming at >3.0 for everything? (unless there's a reason not to?)

Ohh...I failed to unpin h5py in the azure...thought pip in Windows still only installs h5py<3.0.

Now it looks good.

tylerjereddy:
> Now it looks good.

I'm not so sure--as far as I can tell something else also changed in newer `h5py` related to its requisite dependencies and the 32-bit Windows job seems to be missing that thing now, for some reason... That's part of why I pinned there.

yuxuanzhuang:
I was too optimistic. 

It seems that PyPI has not yet provided a 3.1.0 wheel for 32 bit Windows (https://pypi.org/project/h5py/3.1.0/#files), which should include the HDF5 package. Besides, I don't see any discussion from their side about ditching 32-bit Windows completely, so maybe it's just a matter of time to include it again?

Alternatively, maybe we can ask azure to build HDF5 from scratch?

> > Now it looks good.
> 
> I'm not so sure--as far as I can tell something else also changed in newer `h5py` related to its requisite dependencies and the 32-bit Windows job seems to be missing that thing now, for some reason... That's part of why I pinned there.

IAlibay:
> It seems that PyPI has not yet provided a 3.1.0 wheel for 32 bit Windows (https://pypi.org/project/h5py/3.1.0/#files), which should include the HDF5 package. Besides, I don't see any discussion from their side about ditching 32-bit Windows completely, so maybe it's just a matter of time to include it again?
> 
> Alternatively, maybe we can ask azure to build HDF5 from scratch?

The safest thing here would probably be to raise it upstream and see what they say. I had a very quick look (please double check) through their test matrix and I can't seem to see an obvious case on azure/appveyor where the environment runs 32bit, so there's a chance they might have just ditched it for 3.0.

tylerjereddy:
> so there's a chance they might have just ditched it for 3.0.

Hopefully not; as far as I know 32-bit Python interpreters are still more common than 64-bit for Windows. The same reasoning is used to justify all the 32-bit Windows wheels for NumPy/SciPy upstream.

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -600,7 +600,7 @@ def _read_frame(self, frame):
             else:
                 raise NoDataError("Provide at least a position, velocity"
                                   " or force group in the h5md file.")
-        except ValueError:
+        except IndexError:
```
orbeckst, line 603: Let's make this 2.0 and 3.0 compatible:
```suggestion
        except (ValueError, IndexError):
```

orbeckst (changes requested):
It's easy to keep the code compatible with hdf 2.x and 3.x so I would do that. Then the CI can install whatever it can find.

orbeckst:
I'll resolve the merge conflicts to develop...

orbeckst:
I wanted mdanalysi-CI (Azure) to rerun; I hit the "rerun" link but not sure if anything is happening. If it still says that it failed 21d ago then I'll need some advice on how to kickstart (or even better: how to make the 32bit tests pass).

IAlibay:
Was this ever fixed upstream? I see that the issue is still open: https://github.com/h5py/h5py/issues/1753

Re restarting, I've had the same issue as @orbeckst in the past :( The only solution I've ever had to it was closing/re-opening. I'm sure there's a better way to do it though.

tylerjereddy:
@orbeckst @IAlibay Maybe that trick will suffice for now? I believe  you can specify a specific matrix entry name too.

IAlibay:
Unless I'm missing something, it does seem to me that h5py isn't, at least for the short term, planning to add wheels for 32bit windows. 

We probably should consider this when we consider the future of 32bit support, but in the short term, could we just pin the version on 32bit CI? As long as we support pre and post 3.0, I guess we can make the argument that 32 bit windows users would just end up installing pre-3.0 h5py if they wanted to meet that dependency?

.github/workflows/gh-ci.yaml
```diff
@@ -0,0 +1,330 @@
...
+      - develop
+      - master
+
+defaults:
+  run:
+    shell: bash -l {0}
+
+env:
+  MDA_CONDA_MIN_DEPS: "pip pytest==6.1.2 mmtf-python biopython networkx cython matplotlib scipy griddataformats hypothesis gsd codecov"
+  MDA_CONDA_EXTRA_DEPS: "seaborn>=0.7.0 clustalw=2.1 netcdf4 scikit-learn joblib>=0.12 chemfiles tqdm>=4.43.0 tidynamics>=1.0.0 rdkit>=2020.03.1 h5py==2.10.0"
```
orbeckst, line 18: Why is h5py pinned to 2.10.0 here? I thought we would allow 2 or 3?

maintainer/conda/environment.yml
```diff
@@ -0,0 +1,34 @@
...
+channels:
+  - defaults
+  - conda-forge
+dependencies:
+  - chemfiles
+  - codecov
+  - cython
+  - griddataformats
+  - gsd
+  - h5py==2.10.0
```
orbeckst, line 11: unpin?

azure-pipelines.yml
```diff
@@ -49,16 +40,15 @@ jobs:
...
       matplotlib
       numpy
-      pytest
+      pytest==6.1.2
       pytest-cov
       pytest-xdist
       scikit-learn
       scipy
-      h5py==2.10.0
+      h5py
```
orbeckst, line 48: Are we  using azure _only_ for 32-bit Windows builds now? If so we could pin h5py==2.10.0 here (+ comment) although I don't understand why the package resolving algorithms are not able to get the right version by itself. Is this a pip or conda issue? (Naively, I thought that if h5py is not available for an arch then it will just use 2.10.0.)

orbeckst (commented):
Questions...

orbeckst:
I agree with @IAlibay .

There are things I don't understand so I added a comment/questions review.

azure-pipelines.yml
tylerjereddy: > Are we using azure only for 32-bit Windows builds now?

I don't think so--it looks like PRs to `develop` have more matrix entries than that for azure. It looks like this PR is deleting some Azure matrix entries--can't tell if that's because of the changes to github workflows CI here though.

I think the `pip` resolver is looking for the latest available version (it is definitely pip, not conda here), and it will build from source if it has to (and it fails to do that here). One option might be to use `--only-binary` flag for `pip install` for `h5py`; that should exclude usage of source packages and fall back to newest available wheel: https://pip.pypa.io/en/stable/reference/pip_install/#install-only-binary (assuming the older h5py has the 32-bit binary and wasn't just building that from source..)

I'm probably on the other side for 32-bit support--if deps don't support it, shed the deps/related feature or don't take it on in the first place or at least make it fully optional, especially for Windows where 32-bit is still important.

orbeckst:
I lost track of what needs to be done here to get this fix in. 

Can we make it "just work" if we require h5py > 2 ? 

(I would have liked supporting 2.x because it's easy from the code side but if it breaks everything in CI/dependency hell then I am willing to cut losses and ditch 2.x.)

ping @tylerjereddy @IAlibay @yuxuanzhuang @edisj

IAlibay:
> Can we make it "just work" if we require h5py > 2 ?

Do you mean > 3?

The issue here is that 3.0+ seem to have dropped 32bit support.

The easiest answer here would be to include the code paths for both <3 and >3, and have the 32bit CI pinned at <3?

orbeckst:
I did mean ">3" because I assumed then we only had to worry about one thing.

If we can just pin the 32 bit to < 3 then that's good enough for me. If we collect coverage from everyone then we'll cover all the code paths, right?

yuxuanzhuang:
Sorry, I may have lost track of how GH action and azure work now, but I don't see the azure check running?

yuxuanzhuang:
The method mentioned by @tylerjereddy works well here.
> One option might be to use --only-binary flag for pip install for h5py; that should exclude usage of source packages and fall back to newest available wheel

The current problem is after I unpin the h5py version in gh-ci, Those checks don't seem to be running now.

tylerjereddy:
> Those checks don't seem to be running now.

Just a drive-by comment, but which checks specifically are not running? 18 successful CI entries seems to be the current norm for PRs to `develop` here. Note: I didn't review the latest changes, just asking this question for now.

yuxuanzhuang:
> > Those checks don't seem to be running now.
> 
> Just a drive-by comment, but which checks specifically are not running? 18 successful CI entries seems to be the current norm for PRs to `develop` here. Note: I didn't review the latest changes, just asking this question for now.

Oh, It took quite some time to finally show up. Now it looks good.

IAlibay:
> > > Those checks don't seem to be running now.
> > 
> > 
> > Just a drive-by comment, but which checks specifically are not running? 18 successful CI entries seems to be the current norm for PRs to `develop` here. Note: I didn't review the latest changes, just asking this question for now.
> 
> Oh, It took quite some time to finally show up. Now it looks good.

There were issues with the gh action backend yesterday iirc, which is probably why the delay happened.

IAlibay:
@orbeckst is this ready for squash merge? (just trying to push along everything)

orbeckst:
Looks awesome from my end: 🚀 !

orbeckst:
Thanks @yuxuanzhuang !!

And thank you @IAlibay to move things forward!

PR #3027, opened 2020-11-11
MAINT, CI: Azure add Python 3.9 Win

tylerjereddy:
* the upstream ecosystem has had a bit of time
to release binaries for Python `3.9`; let's see
what happens if we turn on Python `3.9` Windows testing
in Azure CI


tylerjereddy:
Looks like there are no 3.9 wheels for `h5py` and `scikit-learn` on Windows, with the former causing an actual failure to build in our dependency install step. So, may need to wait a bit more/check with upstream plans. I know that the increased Python release cadence is causing a fair bit of maintainer burden to get things out.

tylerjereddy:
rebase/force push 1 month later to see if the ecosystem is a bit more developed on Python 3.9 yet

orbeckst:
It's a month later ;-)

tylerjereddy:
There is a binary wheel on PyPI for h5py version 3.1.0 and this Windows/64-bit combination. I think we have another MR related to unpinning h5py so I can't judge if just bumping the version for that is safe here or if it is optional so we could just not use it for now in that matrix entry?

IAlibay:
@tylerjereddy I believe there's an Error type change or something that @yuxuanzhuang had to implement to for >=3.1 h5py (I think one of the failures swapped from a ValueError to an IndexError).

Assuming we can merge #3025 following @orbeckst's latest comments this probably _should_ work here?

IAlibay:
@tylerjereddy I've taken the liberty to update this again the current develop, now that #3025 is merged it looks like this all works fine? (waiting on the updated CI to run)

tylerjereddy:
Thanks, I've merged--makes sense to test with 3.9 for Windows I think

Issue #386, opened 2015-08-07
automatically deploy development docs


orbeckst:
Following [Automatically Update Github Pages with Travis Example](https://github.com/steveklabnik/automatically_update_github_pages_with_travis_example) seems straighforward enough in order to have travis-ci build the sphinx docs and then push them to the gh-pages branch. (See also the discussion at the bottom of #350.) 

## Potential solutions
* [Automatically Update Github Pages with Travis Example](https://github.com/steveklabnik/automatically_update_github_pages_with_travis_example)
* [Auto-deploying built products to gh-pages with Travis](https://gist.github.com/domenic/ec8b0fc8ab45f39403dd)
*  [awestruct gh deployment](http://awestruct.org/auto-deploy-to-github-pages/) notes

## Progress so far
* [x] generated OAUTH token ([limit scope](https://developer.github.com/v3/oauth/#scopes) to `public_repo` access; I used the [tokens web interface](https://github.com/settings/tokens) but can also be done via `curl` as described in the [awestruct gh deployment](http://awestruct.org/auto-deploy-to-github-pages/) notes). DONE
* [x] encrypt for travis with the `travis` client:

    ```
    gem2.0 install travis
    travis encrypt GH_TOKEN=`cat ./github_OAUTH_token`
    ```

    yields the entry for the `.travis.yml` file:

   ```
     secure: "HIj3p+p2PV8DBVg/KGUx6n83KwB0ASE5FwOn0SMB9zxnzAqe8sapwdBQdMdq0sXB7xT1spJqRxuxOMVEVn35BNLu7bxMLfa4287C8YXcomnvmv9xruxAsjsIewnNQ80vtPVbQddBPxa4jKbqgPby5QhhAP8KANAqYe44pIV70fY="
   ```
   (This is encrypted with the travis public key. For more details, see [Travis CI: Encryption keys](http://docs.travis-ci.com/user/encryption-keys/).)
* [x] add lines to `.travis.yml` along the lines of
  ```yaml
  script:
    - ./testsuite/MDAnalysisTests/mda_nosetests -v --with-coverage --cover-package MDAnalysis --processes=2 --process-timeout=120 --with-memleak
    - coveralls
    - test $TRAVIS_PULL_REQUEST == "false" && test $TRAVIS_BRANCH == "develop" && cd package/doc/sphinx && make clean html

  after_success:
[...]

PR #3040, opened 2020-11-23
Adds basic GH Actions CI workflow

### LINKED ISSUE #3036: On the future of CI post Travis
Just putting this up as there's been a lot of discussion internally, off-github, and on the Twitterverse :)

## Current status

As of today Travis CI has switched over to using a credit based system for OSS CI. Unfortunately the amount of credits given are far too low to keep MDAnalysis going. As of writing this, we have already used up all our allocated credits (that's ~ 5 commits and 2 merges...). 

Thankfully, mostly due to the work of @tylerjereddy, half our CI infrastructure is still working (mostly Windows testing, but we have all the bits in place to add cover for other OSes).

## Solution

The current proposed solution (based on some off-github discusions) is that we should both move part of our CI matrix to existing still-free solutions (i.e. appveyor and azure), and also offload some of that to Github Actions.

## Timeframe

I'm currently in the process of testing out github actions on IAlibay/mdanalysis and am hoping to open a PR at some point tomorrow (how early will probably depend on how much more coffee I decide to drink tonight 🙃).

If someone wants to take the lead on reviewing the current matrix and see how things should be spread out to actions/azure/appveyor (maybe some minimal dependency pip based installs?) that would be awesome :)

IAlibay:
Fixes #3036 

## Changes made in this Pull Request:
- Adds basic GH Actions CI workflow

## To do:
- <s>Fix codecov</s>
- <s>Docs (separate workflow?)<s/> build tests is done, deployment will need to be done in a separate PR.
- <s>Numpy versions (maybe we have to ditch env files?)</s> -> numpy-dev to be done some other time
- <s>Full osx matrix on cron?</s> -> open separate issue for this

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


IAlibay:
Strangely enough, the `coverage.misc.CoverageException` error seen here originally also occured locally on my own own fork, but now is no longer reproducible. It's unclear as to why this is happening, so any insights here would be greatly appreciated.

lilyminium:
Is it https://github.com/nedbat/coveragepy/issues/703#issuecomment-575628293 ? I notice that on both your fork and here it's the 3.6 test failing, are we using 3.6.0?

.github/workflows/gh-ci.yaml
```diff
@@ -0,0 +1,191 @@
...
+
+    - name: install_asv
+      if: contains(matrix.name, 'asv_check')
+      run: |
+        pip install asv "setuptools<45.0.0"
+
+    - name: install_hole
+      if : matrix.install_hole
+      run: |
+        # cert issue, anyone with a mac can circumvent this?
```
lilyminium, line 101: fascinatingly, macOS stores its certificates in keychain access, which curl, being not being mac-specific, does not know. Adding this to the install_hole.sh exports the keychain certificates and tells curl where to find it:

```diff
OSNAME="$1"
case "$OSNAME" in
    Linux|linux)
	OSNAME=Linux
	OFORMAT=Linux
	DOWNLOAD_URL="${DOWNLOAD_URL_LINUX}"
	TARFILE=${TARFILE_LINUX}
+++    CURLARGS=""  # i think this dies if a variable is unset
	;;
    OSX|osx|Darwin|darwin)
	OSNAME=Darwin
	OFORMAT="Mach-O"
	DOWNLOAD_URL="${DOWNLOAD_URL_DARWIN}"
	TARFILE=${TARFILE_DARWIN}
+++    security export -p -t certs -k `security list-keychains -d system|cut -d '"' -f 2` -o certs.crt
+++    CURLARGS="--cacert $(PWD)/certs.crt"
	;;
    *)
	die "OS '${OSNAME}' not supported." 10;;
esac

PREFIX="$2"
test -n "$PREFIX" || die "missing second argument PREFIX (installation directory)" 10

#------------------------------------------------------------
# start installation
#------------------------------------------------------------

mkdir -p "$PREFIX" && cd "$PREFIX" || die "Failed to create and cd to $PREFIX" 1
if ! test -f ${TARFILE}; then
    echo "Downloading ${TARFILE} from ${DOWNLOAD_URL}..."
    # fixing curl on travis/anaconda, see http://stackoverflow.com/a/31060428/334357
    export CURL_CA_BUNDLE=/etc/ssl/certs/ca-certificates.crt
---    curl -L ${DOWNLOAD_URL} -o ${TARFILE} $CURLARGS || \
+++    curl -L ${DOWNLOAD_URL} -o ${TARFILE} $CURLARGS || \
    die "Failed to download ${TARFILE} from ${DOWNLOAD_URL}" 1
fi
```

IAlibay: Should we be doing this at the install_hole level or GH Actions, i.e. how likely are the Apple overlords going to decide to change things between osx versions?

IAlibay:
> Is it [nedbat/coveragepy#703 (comment)](https://github.com/nedbat/coveragepy/issues/703#issuecomment-575628293) ? I notice that on both your fork and here it's the 3.6 test on MacOS failing, are we using 3.6.0?

So looking at the conda list trace (in setup check), the version that's picked up is 3.6.11. Pinning to 3.6.2 seemed to have fixed it (although it is a flaky issue, so it's possible it only passed this one time, like what's happening with my own fork), but I don't think it's reasonable for us not to use the latest version of 3.6 :(

I'm baffled as to why we weren't facing this issue with the ci helpers :/

IAlibay:
Talking about coverage versions, this from astropy's ci-helpers might be worth keeping in mind/testing:

https://github.com/astropy/ci-helpers/blob/7393396f8ed26ec529e75bf902d16af7f6f1e463/travis/setup_dependencies_common.sh#L915-L923

maintainer/certs.crt
```diff
@@ -0,0 +1,40 @@
+-----BEGIN CERTIFICATE-----
```
IAlibay, line 1: Should this file be here?

IAlibay:
I'm going to try replicating the astropy ci helpers, and unpinning the py3.6 version (it _should_ pick up 3.6.11 by default), it works on my fork, but who knows if it'll work here (it's all very flaky).

IAlibay:
That seems to have worked (i.e. reproducible across both my own fork and here).

We should try to add the fancy codecov action next (don't think it adds anything, but why not).

One thing to note here is that the conda build time is really long, especially for MacOS where it exceeded 13 mins.
Given that we can't have fine grained control of numpy with environment files anyways, I think the answer here is this to a commandline `conda` and `pip` call. This would also allow us to use mamba which should speed things up.

.github/workflows/gh-ci.yaml
```diff
@@ -0,0 +1,189 @@
...
+    - name: install_asv
+      if: contains(matrix.name, 'asv_check')
+      run: |
+        pip install asv "setuptools<45.0.0"
+
+    - name: install_hole
+      if : matrix.install_hole
+      run: |
+        bash ./maintainer/install_hole.sh ${OS_NAME} "${HOME}"
+        HOLE_BINDIR=${HOME}/hole2/exe
```
IAlibay, line 102: This (and the export below) won't work for HOLE. This can be seen from the lack of coverage for hole2. The issue here, I think, is that env variables and exports aren't shared between job steps by default.

We'll have to do something similar to what we do with setup_linux.

IAlibay:
So it looks to me like the HOLE2 code isn't compatible with Catalina (it looks like the whole 32 bit incompatibility thing?).

Can someone with a Mac device and Catalina confirm this? pinging @RMeli and @lilyminium 

Also pinging @orbeckst who might know more about the future of hole2 and binary releases.

RMeli:
Yes, I can confirm this locally.

IAlibay:
@RMeli confirms that HOLE2 will not work on Catalina.
It is not possible with GH Actions to use anything older than Catalina, I suggest we just drop test support for it unless somehow new binaries are made available.

@lilyminium is there anything that needs to be done for a long term deprecation of MacOS support for the hole2 code?

IAlibay:
So issues #3043  and #3042 aside, the tests now have the same coverage (17225 passed, 67 skipped, 4 xfailed, 2 xpassed) as a similar local install.

What's still left to do is:

- [ ] Check that pip is behaving properly
- [ ] re-enable `-n 2` on pytest (I disabled it so we could have an idea of what was getting skipped).
- [ ] delete unused environment yaml files

I suggest that we add the numpy-dev jobs back at a later date (will probably need @tylerjereddy's expertise here), and the docs build in a different PR (@lilyminium from what I gather it's not meant to be activated on PR right? If it doesn't matter we can add it here).

orbeckst:
Thank you @IAlibay , you're a HERO ⚔️ 🛡️ .

I don't know anything about HOLE2 development. I will try to build it on Catalina and see if I can get it to work. Maybe that motivates me to make a conda package...

orbeckst:
HOLE2 builds just fine on macOS 10.15.7 Catalina *with my patch for using the `FC` compiler variable correctly*. I used macports with gcc 10 (`sudo port install gcc10`) but probably any version of gcc/gfortran will work. 

Instructions are in https://github.com/MDAnalysis/hole2/blob/develop/INSTALL.md – basically set the `FC` Makefile variable (could also be done in environment but I like to keep my env clean):
```bash
git clone https://github.com/MDAnalysis/hole2.git
cd hole2/src
source ../source.apache     # if you did NOT change the source code.
make FC=gfortran-mp-10      # set FORTRAN compiler in FC
make PREFIX=$HOME/hole2 FC=gfortran-mp-10 install
source ../source.unset       # clean up environment
```
gives
```
$ /Users/oliver/hole2/bin/hole
 *** Program HOLE ***

 If results of this run are reported or published, please
 cite the publication listed below.
 For the HOLE program:
   Smart, O.S., Goodfellow, J.M., Wallace, B.A. (1993).
   The pore dimensions of gramicidin A.
   Biophys. J., 65, 2455-2460.

 Originally written by Oliver Smart at Birkbeck College from April 1992.
 Copyright 1993-2005 Oliver Smart
 Copyright 2014-2015 SmartSci Limited.
 Copyright 2016 Oliver Smart.

 For help on HOLE suite see  http://www.holeprogram.org/

 HOLE release 2.2.005 (07 August 2016)

 usage subject to Apache License, Version 2.0
 see http://www.apache.org/licenses/LICENSE-2.0

 Control variables read:
```

For a start, we could build HOLE on the fly instead of downloading the old binaries. Even better for us (and the community) would be a conda package.

IAlibay:
Thanks for working this out @orbeckst, do you happen know if there's a testsuite out there to check compilation validity? I can't seem to see anything in the source.

I guess our own tests should be enough to check if something went wrong with the compilation, so it probably doesn't matter.

maintainer/certs.crt
richardjgowers: I have no idea what this is, so probably not?

richardjgowers (commented):
Maybe it's just sour grapes, but this integrates nicer on github than travis did.. This looks really good.

https://codecov.io/gh/MDAnalysis/mdanalysis/pull/3040/tree

^ The coverage seems to be a little weird where some of the package is flattened out.  I don't think that's a showstopper though

maintainer/certs.crt
IAlibay: I believe @lilyminium added it, I'll let her make the call here :)

IAlibay:
> ^ The coverage seems to be a little weird where some of the package is flattened out. I don't think that's a showstopper though

@richardjgowers I'm not sure what you mean here, the coverage seems the same as what's on develop at the moment, but I'm probably not looking at the right thing?

orbeckst:
No testsuite that I know off. One would probably use the gramicidin example that is described in the docs but I think I used that already for the MDA tests. It would, of course, be good to have tests for HOLE so that it’s clear that when the MDA tests fail, it’s a problem with MDA and not with HOLE. But life is short and time is precious... so unless the MDA tests fail right away I’d leave HOLE tests for another day.

maintainer/certs.crt
lilyminium: Sorry I should have deleted it, I added it as part of trying to fix ca-certificates

.github/workflows/gh-ci.yaml
```diff
@@ -0,0 +1,279 @@
...
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: setup_osx
+      if: startsWith(matrix.os, 'macOS')
+      run: |
+        # Set OS specific vars and compiler flags
+        echo "OS_NAME=osx" >> $GITHUB_ENV
+        # TODO: work out why this is necessary (from CI helpers)
+        echo "MACOSX_DEPLOYMENT_TARGET=10.9" >> $GITHUB_ENV
```
IAlibay, line 74: Anyone with a MacOS device know why this is necessary? I've taken this from the astropy ci-helpers, not having it seems to cause some pip compilation errors (see: https://github.com/MDAnalysis/mdanalysis/runs/1462154629?check_suite_focus=true).

However having it also yields these types of warnings:
`ld: warning: dylib (/usr/local/miniconda/envs/test/lib/libc++.dylib) was built for newer macOS version (10.9) than being linked (10.7)`

Considering the tests pass, unless someone knows a quick answer here, I'm tempted to just leave it for now and someone with a macOS device can fix it at a later date?

.github/workflows/gh-ci.yaml
```diff
@@ -0,0 +1,279 @@
...
+        cd package && python setup.py build_ext --inplace
+
+    - name: build_docs
+      run: |
+        cd package && python setup.py build_sphinx
+
+    - name: deploy_docs
+      if: github.event_name != 'pull_request'
+      run: |
+        # place the deploy call here
```
IAlibay, line 213: Fixing the deployment script to go from travis dependencies to gh actions is going to take extra work. I propose we do this in a separate PR to accelerate the deployment of the base CI workflow.

testsuite/MDAnalysisTests/analysis/test_hole2.py
```diff
@@ -480,7 +480,7 @@ def test_over_order_parameters(self, hole):
             assert key == rmsd
 
         idx = np.argsort(op)
-        arr = np.array(list(hole.profiles.values()))
+        arr = np.array(list(hole.profiles.values()), dtype=object)
```
IAlibay, line 483: I didn't explain this change; this is because numpy was complaining of a VisibledeprecationWarning on running the test ( I assume we weren't picking this up in Travis because the ci helpers tend to pin the numpy version to something conservative).

maintainer/install_hole.sh
```diff
@@ -31,6 +31,7 @@
 set -o errexit -o nounset
 
 SCRIPT=$(basename $0)
+MAINTAINER=$(cd "$(dirname "$0")"; pwd -P)
```
IAlibay, line 34: I'm going to leave these changes here for now in case we revert to using the pre-built binaries. We can clean up in a future PR.

IAlibay:
@MDAnalysis/coredevs this should be ready to review with the aim to merge.

As stated in the comments I think we should deal with docs deployment & the numpy-dev checks in separate PRs. Aside from this, the rest of the old travis functionality should be there (except for ARM support).

At the moment, the only thing I couldn't work out is why this happens when we build MDA ([see here for the full trace](https://github.com/MDAnalysis/mdanalysis/runs/1464963931?check_suite_focus=true#step:9:1206)) even though we install mmtf-python explicitly in the conda deps:

```
Installed /usr/share/miniconda/envs/test/lib/python3.8/site-packages/MDAnalysis-2.0.0.dev0-py3.8-linux-x86_64.egg
Processing dependencies for MDAnalysis==2.0.0.dev0
Searching for mmtf-python>=1.0.0
Reading https://pypi.org/simple/mmtf-python/
Downloading https://files.pythonhosted.org/packages/af/d2/aabefd607ff89e35704574205b382a351f531885cc98bfd91baaf192a9fe/mmtf_python-1.1.2-py2.py3-none-any.whl#sha256=a00136e9798e84f58f509a8bc341834e497169c1079bd757c12bc1a4ce736b57
Best match: mmtf-python 1.1.2
Processing mmtf_python-1.1.2-py2.py3-none-any.whl
Installing mmtf_python-1.1.2-py2.py3-none-any.whl to /usr/share/miniconda/envs/test/lib/python3.8/site-packages
Adding mmtf-python 1.1.2 to easy-install.pth file
```

IAlibay:
Coverage seems to have dropped by 0.08%, but I can't seem to understand why since tests are running as normal, I'll temporarily disable the fancy codecov action and see if that fixes things.

.github/workflows/gh-ci.yaml
RMeli: Related to https://github.com/conda/conda-build/issues/3682?

IAlibay: Probably? I've read the thread twice and I honestly can't seem to capture what's being argued there 🙃 

Any ideas of potential downsides to including this?

orbeckst: Is `/usr/local/miniconda/envs/test` where the GH actions install miniconda? Just curious about the path.

The error sounds almost as if `MACOSX_DEPLOYMENT_TARGET=10.9` is not correctly picked up because somewhere it is being linked against something that was build with the 10.7 SDK. Are we linking against anything outside the anaconda distribution? 

If the tests just run then I'd leave as is for the time being (and raise an issue so that it's not forgotten).

.github/workflows/gh-ci.yaml
```diff
@@ -0,0 +1,279 @@
...
+
+    - name: setup_miniconda
+      uses: conda-incubator/setup-miniconda@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+        auto-update-conda: true
+        channel-priority: flexible
+        channels: biobuilds, conda-forge
+        add-pip-as-python-dependency: true
+        # TODO: mamba causes pip to segfault, switch when fixed
```
orbeckst, line 103: How sad!

.github/workflows/gh-ci.yaml
```diff
@@ -0,0 +1,279 @@
...
+        # NOTE: vars need to be re-assigned
+        # NOTE: set matrix.numpy to pin to a specific numpy version
+        conda_deps="${{ matrix.numpy }} ${MDA_CONDA_${{ matrix.run_type }}_DEPS}"
+        pip_deps=${MDA_PIP_${{ matrix.run_type }}_DEPS}
+        conda install ${conda_deps}
+        pip install ${pip_deps}
+
+        # NOTE: set setuptools version if running asv test
+        if [ ${{ matrix.name }} = "asv_check" ]; then
+          pip install asv "setuptools<45.0.0"
```
orbeckst, line 121: Does asv not work with recent setuptools?

.github/workflows/gh-ci.yaml
```diff
@@ -0,0 +1,279 @@
...
+      if : matrix.install_hole
+      run: |
+        # We manually build hole2 to avoid OS incompatibilities
+        git clone https://github.com/MDAnalysis/hole2.git
+        cd hole2/src
+        source ../source.apache
+        (make FC=${FC}) && (make PREFIX=${HOME}/hole2 FC=${FC} install)
+        source ../source.unset
+        echo "HOLE_BINDIR=${PWD}/../exe" >> $GITHUB_ENV
+        echo "${PWD}/../exe" >> $GITHUB_PATH
```
orbeckst, line 144: With the installation step, you know the bin dir:
```suggestion
        echo "HOLE_BINDIR=${HOME}/hole2/bin" >> $GITHUB_ENV
        echo "${HOME}/hole2/bin" >> $GITHUB_PATH
```

.github/workflows/gh-ci.yaml
```diff
@@ -0,0 +1,279 @@
...
+
+    - name: install_mda
+      run: |
+        (cd package/ && python setup.py develop) && (cd testsuite/ && python setup.py install)
```
orbeckst, line 148: How about doing a proper install instead of develop?
```suggestion
        (cd package/ && python setup.py install) && (cd testsuite/ && python setup.py install)
```
This can catch the occasional mistake in missing files.

.github/workflows/gh-ci.yaml
```diff
@@ -0,0 +1,279 @@
...
+    - name: install_deps
+      run: |
+        conda_deps="${{ env.MDA_CONDA_MIN_DEPS }} ${{ env.MDA_CONDA_EXTRA_DEPS}}"
+        pip_deps="${{ env.MDA_PIP_MIN_DEPS}} ${{ env.MDA_PIP_EXTRA_DEPS }} sphinx==1.8.5 sphinx-sitemap sphinx_rtd_theme msmb_theme==1.2.0"
+        conda install ${conda_deps}
+        pip install ${pip_deps}
+
+    - name: install_mda
+      run: |
+        cd package && python setup.py build_ext --inplace
```
orbeckst, line 204: Why --inplace instead of a normal installation?

.github/workflows/gh-ci.yaml
orbeckst: Is this primarily @lilyminium 's domain? 

Raise new issue...

.github/workflows/gh-ci.yaml
```diff
@@ -0,0 +1,279 @@
...
+    if: "github.repository == 'MDAnalysis/mdanalysis'"
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: setup_miniconda
+      uses: conda-incubator/setup-miniconda@v2
+      with:
+        python-version: 3.6
```
orbeckst, line 227: bump to 3.7?

.github/workflows/gh-ci.yaml
```diff
@@ -0,0 +1,279 @@
...
```
orbeckst, line 258: bump to 3.7?

maintainer/install_hole.sh
orbeckst: I think you can remove the `install_hole` script. Perhaps we should pin the download from our hole repo (as we're pullling from master) to a tag or commit so that we ensure we know which version of (customized) hole we're building. But otherwise I don't see us going back to the binaries.

The next thing would be to use a conda-forge package.

orbeckst (approved):
You've done outstanding work on moving from travis to gh actions, +1000 Karma/XP/... 🙇 .

I have a bunch of questions/comments but nothing that I consider a show stopper. I'd rather put it into practice and raise issues as we go along than fiddling.

.github/workflows/gh-ci.yaml
lilyminium: This is on us, it's https://github.com/MDAnalysis/mdanalysis/blob/6803cf6c5aa62744099803145f3ac9398389244c/package/setup.py#L288-L292

Do we have good reason to support 10.7? It's now [10 years old.](https://en.wikipedia.org/wiki/Mac_OS_X_Lion)

.github/workflows/gh-ci.yaml
lilyminium: Done :-) https://github.com/MDAnalysis/mdanalysis/issues/3046

.github/workflows/gh-ci.yaml
lilyminium: This is outdated now and instead downloads a different certificate file. I don't think that will change quickly, I hope, and ideally the URL will not get hijacked (I guess we could upload our own file for this?) If Apple changes anything like adding sensible default certificates, we could change it back to use the system ones.

.github/workflows/gh-ci.yaml
IAlibay: > Is /usr/local/miniconda/envs/test where the GH actions install miniconda? Just curious about the path.

Yeah, the paths are a little bit all over the place. I think the miniconda action defaults to a system-wide install, so it's `/usr/share/miniconda/envs/test` on the ubuntu builds, and `/usr/local/miniconda/envs/test` on MacOS.

As a follow up to:

> If the tests just run then I'd leave as is for the time being (and raise an issue so that it's not forgotten).

and

> Do we have good reason to support 10.7? It's now 10 years old.

I agree with @orbeckst that an issue be raised & a follow-on PR be made for this issue. Currently where we are at is that if we don't specify `MACOSX_DEPLOYMENT_TARGET` things fails, and this current workflow offers a direct translation of what we had happening on Travis CI through the astropy CI helpers.

That is to say, at this current point in time our CI is probably no worse or better than it was before we lost access to Travis.

Given that I don't have** access to MacOS device, further debugging here is probably going to have to fall to someone that does.

** I guess technically I could boot up a VMBox instance, but I'd prefer not.

.github/workflows/gh-ci.yaml
IAlibay: This was a literal copy of the travis file, it looks like it's a leftover from a py2.7 pin (so we'll probably need it for `master`/`1.0.x`), but I'll try remove it here.

.github/workflows/gh-ci.yaml
IAlibay: I was doing this until: https://github.com/MDAnalysis/mdanalysis/pull/3040#issuecomment-734985358

The issue is that doing an `install` results in a lack of coverage for the `.pyx` files. I'll see if doing an `--inplace` call might work here.

.github/workflows/gh-ci.yaml
IAlibay: Honestly no idea, it's one of those things I just copied over from the initial travis yaml file. I'll see if we can live without it.

.github/workflows/gh-ci.yaml
IAlibay: I'm sure we can get it to work with a little bit of playing around (or at least we can work out what to report back to the mamba devs). I'll open a separate issue for this.

.github/workflows/gh-ci.yaml
IAlibay: With further reading, I don't think that'll work. If someone has an idea here I can implement it, otherwise I suggest we go with the `develop` setup for now?

IAlibay:
So this is slightly problematic, it looks like Travis still runs even though I turned off the "use credits" toggle... 

Unless anyone has any ideas, I guess I'll have to disable Travis in this PR.

.github/workflows/gh-ci.yaml
IAlibay: fixed (apologies didn't commit suggestion to avoid triggering CI too much)

.github/workflows/gh-ci.yaml
IAlibay: Done, I've also bumped up everything except from the minimum dep and the minimum numpy version builds. My reasoning for the latter two is that they are specifically around to test the "edge case with the least amount of features".

maintainer/install_hole.sh
IAlibay: I've removed the file. Creating a specific tag/release would make sense, but since the current code is in an ongoing PR, I don't know if it's worth waiting on that one (I'd hope that upstream approves it and then releases a minor version)?

IAlibay:
> Unless anyone has any ideas, I guess I'll have to disable Travis in this PR.

Thanks to @richardjgowers for the simple "just rename it" solution here. Unless there's any objections, I'd like to keep it around for now so that we can directly fix it rather than remove it and have to find it in history later?

Aside from that, I think everything "works" for now?

.github/workflows/gh-ci.yaml
orbeckst: Hm, ok, I bow to facts. Perhaps add a comment to the file as to why we're doing develop install, just so that future me is not asking the same question again?

IAlibay: I have a feeling it is fixeable, but it's probably one of those obvious things I can't find in the setuptools docs (I really need to sit down and learn it all properly one day) 🙃 I'll tag it as `TODO`, I do agree that using `install` everywhere would be much better.

.github/workflows/gh-ci.yaml
IAlibay: So it looks like it relates back to: https://github.com/MDAnalysis/mdanalysis/commit/80bc2a2a9cc826627eca2691e425c284fb76269f and https://github.com/MDAnalysis/mdanalysis/issues/386#issuecomment-139747327. I've tried iterating through a couple of different options, if we switch to `python setup.py build_sphinx -E` in the next step (instead of `python setup.py build_sphinx`), then we can use `python setup.py develop` here (note that `install` won't work because of cython [issues](https://github.com/IAlibay/mdanalysis/runs/1484029180?check_suite_focus=true)).

I have changed it to using python setup.py develop` since it's more in line with all the other things we do. The only thing I'm not sure about is if/how this would affect docs deployment.

.github/workflows/gh-ci.yaml
```diff
@@ -0,0 +1,281 @@
...
+
+    - name: build_docs
+      run: |
+        cd package && python setup.py build_sphinx -E
+
+    - name: deploy_docs
+      if: github.event_name != 'pull_request'
+      run: |
+        # place the deploy call here
+        echo "Oh, maple syrup roast parsnips [Richard Gowers]"
```
lilyminium, line 216: 🥕

orbeckst: Parsnips are white!

(and delicious)

orbeckst:
@IAlibay go head here, merge PR #3051 afterwards to develop, and then backport to master?

orbeckst:
Hooray!!!!!!!! THANK YOU, @IAlibay !

Can you inform the crowds aka developer list regarding this change? 

What do we have to do for open PRs?

IAlibay:
> Can you inform the crowds aka developer list regarding this change?

Sure, can do.
 
> What do we have to do for open PRs?

Just update the feature branch with the contents of develop and that should do the trick.

PR #3044, opened 2020-11-25
An "accurate and reliable" RDKitConverter

cbouy:
Hi @IAlibay @orbeckst @richardjgowers 

I made a few changes to the RDKitConverter so that it more accurately infers bond orders and charges in small molecules.
I tested this new version on the ChEMBL27 dataset and it managed to retrieve the original molecule or a resonance structure in 99.1% of the 1.7M compounds (the GSoC version was around 90%).
I can share the notebook that I used to run this ChEMBL benchmark, either in one of you other repos or I create a new one, as you prefer.

 - Fixes monatomic ion charges that were always negative
 - Fixes more cases of conjugated systems
 - Fixes tetravalent phosphorus
 - Fixes some edge cases with sulfur and nitrogen
 - Modifies the bond inferring tests to try matching with resonance structures
 - Adds tests for ions, all standard amino-acids and all standard nucleic-acids
 - Adds tests on molecules encountered in the ChEMBL 27 dataset
 - Fixes atoms in the RDKit molecule that weren't always in the same order as in the atomgroup
 - Fixes atom name formatting when writing a PDB file with RDKit
 - Fixes sanitization errors that occurred when using `NoImplicit=False`
 - The `names` property is now available through `atom.GetProp("_MDAnalysis_name")`, while `atom.GetMonomerInfo().GetName()` returns the PDB-formatted version (with spaces).

All of these changes do not take "efficiency" in consideration, so it probably slows down the code a bit, but I think they are all necessary. If you identify things that could be optimized please let me know.

Their are additional changes that I could make that on one hand would slow down the code even more but on the other hand make it even more reliable. See #3339 for examples of failing molecules.

I could fix them by adding more SMIRKS/reaction SMARTS, but considering the fact that these patterns are rare and would slow down the conversion I don't think they are worth adding to the code.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


richardjgowers:
Hey @cbouy I think we're more interested in correctness rather than speed.  Speed is something we can tweak on later once we've established what the code should do.

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -512,9 +512,25 @@ def _infer_bo_and_charges(mol):
     R-C(-O)-O the first oxygen read will receive a double bond and the other
     one will be charged. It will also affect more complex conjugated systems.
     """
+    atoms = sorted([a for a in mol.GetAtoms() if a.GetAtomicNum() > 1],
```
lilyminium, line 515: Will this cause issues for dummy atoms etc? e.g. ``rdkit.Chem.MolFromSmiles(r"C(=C(/C([2*:2]))N)(\[1*:1])"))``

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -512,9 +512,25 @@ def _infer_bo_and_charges(mol):
...
+    MONATOMIC_CATION_CHARGES = {
+        3: 1, 11: 1, 19: 1, 37: 1, 47: 1, 55: 1,
+        12: 2, 20: 2, 29: 2, 30: 2, 38: 2, 56: 2,
+        26: 2, # Fe could also be +3
+        13: 3,
+    }
 
-    for atom in sorted(mol.GetAtoms(), reverse=True,
-                       key=lambda a: _get_nb_unpaired_electrons(a)[0]):
+    for atom in atoms:
```
lilyminium, line 526: Could you please add tests for case where an atom degree is 0?

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -546,15 +562,16 @@ def _infer_bo_and_charges(mol):
...
-
-            # if the atom still has unpaired electrons
-            nue = _get_nb_unpaired_electrons(atom)[0]
-            if nue > 0:
-                # transform it to a negative charge
-                atom.SetFormalCharge(-nue)
+                    # go to next atom if one of the valences is complete
+                    nue = _get_nb_unpaired_electrons(atom)
+                    if any([n == 0 for n in nue]):
+                        break
```
lilyminium, line 568: Could you please add a test for this case where common_nue != 0?

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -546,15 +562,16 @@ def _infer_bo_and_charges(mol):
...
+
+            # if atom valence is still not filled
+            nue = _get_nb_unpaired_electrons(atom)
+            if not any([n == 0 for n in nue]):
```
lilyminium, line 572: Could you add a test for this please?

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -546,15 +562,16 @@ def _infer_bo_and_charges(mol):
...
+                # transform nue to charge
+                atom.SetFormalCharge(-nue[0])
```
lilyminium, line 574: Could you add a test for this please?

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -605,52 +622,53 @@ def _standardize_patterns(mol, max_iter=200):
...
- 
+
     """
 
     # standardize conjugated systems
     _rebuild_conjugated_bonds(mol, max_iter)
 
+    # list of sanitized reactions
+    reactions = []
+    for name, rxn in [
```
lilyminium, line 655: You don't seem to use ``name``. Maybe they could be comments.

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -680,27 +698,22 @@ def _run_reaction(reaction, reactant):
...
-    rxn = AllChem.ReactionFromSmarts(reaction)
-    for n in range(n_matches):
-        products = rxn.RunReactants((reactant,))
+    for n in range(reactant.GetNumAtoms()):
+        reactant.UpdatePropertyCache(strict=False)
+        Chem.Kekulize(reactant)
+        products = reaction.RunReactants((reactant,))
         # only keep the first product
         if products:
             product = products[0][0]
```
lilyminium, line 707: Is this guaranteed to be the "main" product?

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -735,61 +749,83 @@ def _rebuild_conjugated_bonds(mol, max_iter=200):
...
     n_matches = len(set([match[0]
                          for match in mol.GetSubstructMatches(pattern)]))
     if n_matches == 0:
         # nothing to standardize
         return
-    # check if there's an even number of anion-*=* patterns
-    elif n_matches % 2 == 0:
-        end_pattern = Chem.MolFromSmarts("[*-;!O]-[*+0]=[*+0]-[*-]")
-    else:
+    elif n_matches == 1:
```
lilyminium, line 764: Could you add a test for this please?

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -735,61 +749,83 @@ def _rebuild_conjugated_bonds(mol, max_iter=200):
...
         # as a last resort, the only way to standardize is to find a nitrogen
         # that can accept a double bond and a positive charge
         # or a carbonyl that will become an enolate
-        end_pattern = Chem.MolFromSmarts(
-            "[*-;!O]-[*+0]=[*+0]-[$([#7;X3;v3]),$([#6+0]=O)]")
+        end_pattern = odd_end_pattern
+    else:
```
lilyminium, line 769: Could you add a test for this please?

package/MDAnalysis/coordinates/RDKit.py
```diff
@@ -735,61 +749,83 @@ def _rebuild_conjugated_bonds(mol, max_iter=200):
...
+            ) or (
+                term_atom.GetAtomicNum() == 7
+                and term_atom.GetFormalCharge() == 1
+            ):
                 for neighbor in term_atom.GetNeighbors():
                     bond = mol.GetBondBetweenAtoms(anion2, neighbor.GetIdx())
-                    if neighbor.GetAtomicNum() == 8 and bond.GetBondTypeAsDouble() == 2:
+                    if (neighbor.GetAtomicNum() == 8 and
+                        bond.GetBondTypeAsDouble() == 2):
                         bond.SetBondType(Chem.BondType.SINGLE)
```
lilyminium, line 794: Could you please add a comment saying that you're converting this double bond to a single bond because (I guess) it's an edge case in RDKit? Just to make it clear it's on purpose.

And same for the cases below.

lilyminium (changes requested):
Hi @cbouy, this looks awesome. Could you please add some tests for the lines that codecov points out, and address the comment by the PEP8 bot?

package/MDAnalysis/coordinates/RDKit.py
cbouy: I haven't considered dummy atoms when making the converter. It will very likely break the bond-order inferring since it needs the valence(s) of an atom to work properly but it's `-1` for dummies. Not sure if it would be useful or not

cbouy:
A couple more changes:
- atoms in the RDKit molecule are now in the same order as in the atomgroup: the BO inferring code could scramble atoms a bit but I found a convenient way to fix this. Now `mol.GetAtomWithIdx(i)` corresponds to the atom in `atomgroup[i]` as most people would expect
- fixing atom name formatting: this is only useful when converting the RDKit molecule to a PDB file, but I needed it (with the above fix) while playing with [py3Dmol](https://github.com/3dmol/3Dmol.js/tree/master/py3Dmol)
- sanitization is skipped if not possible: could happen when you have a degree 4 nitrogen atom but you were skipping BO/charge inferring so no charge was assigned to it, RDKit would complain about the uncharged v4 N during sanitization.
- a bit of performance boost: because I'm using reactions to fix some things when inferring BO, the properties that are stored in each atom had to be transferred "manually" after each reaction and it was painfully slow. I reorganized the code a bit to only transfer the `_MDAnalysis_index` property after each reaction, and do the complete transfer of properties once we have a "clean" molecule. Also the function to set properties has been changed to avoid running an `if... elif... else` for each atomic property.

orbeckst:
> Once all this is done I'll likely benchmark the BO inferring code on the latest ChEMBL, were should I upload the corresponding notebook?

You can always make a gist and share it. However, I think starting a repo (MDA_RDKit_benchmarks – come up with a better name) might actually be better as it has some room to grow and also gives you the opportunity to deposit data/derived data and eventually archive/publish data via Zenodo. The RDKit functionality is of interest to a sizable part of the community so starting to document and benchmark it will likely be a growing effort. 

As a member of the MDAnalysis org, you should be able to create a repository under MDAnalysis. Don't forget to add a license when you create the repo — GPLv3 is guaranteed to be a correct choice when you import MDAnalysis; a permissive BSD or MIT license will likely not be challenged by anyone.

package/MDAnalysis/converters/RDKit.py
```diff
@@ -522,9 +563,25 @@ def _infer_bo_and_charges(mol):
     R-C(-O)-O the first oxygen read will receive a double bond and the other
     one will be charged. It will also affect more complex conjugated systems.
     """
+    atoms = sorted([a for a in mol.GetAtoms() if a.GetAtomicNum() > 1],
+                   reverse=True,
+                   key=lambda a: _get_nb_unpaired_electrons(a)[0])
+
+    MONATOMIC_CATION_CHARGES = {
```
orbeckst, line 570: document data structure — is it atomic number: charge in e?

package/MDAnalysis/converters/RDKit.py
```diff
@@ -733,6 +795,7 @@ def _rebuild_conjugated_bonds(mol, max_iter=200):
     Since ``anion-*=*`` is the same as ``*=*-anion`` in terms of SMARTS, we can
     control that we don't transform the same triplet of atoms back and forth by
     adding their indices to a list.
+    This function also handles conjugated systems with triple bonds.
```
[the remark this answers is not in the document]
orbeckst: Would it be more expensive to work on a copy of `mol` and return a new `mol`?

orbeckst (commented):
I only had a very shallow look at things. On the documentation side this is looking pretty good. Perhaps some general documentation at the top of the module that outlines your overall approach, high level ideas, and algorithms, like a Methods section in a paper.

I like the comments in the helper functions and that they appear in the docs. I like comments in the code but as @lilyminium pointed out somewhere, add more, especially to explain _why_ something is being done. 

Overall, excellent addition.

package/MDAnalysis/converters/RDKit.py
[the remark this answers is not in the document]
cbouy: It should be fine but I fail to see the point of working on a copy

cbouy:
I've added the link to my RDKit UGM presentation (both video and slides) so that there's an alternative and more visual explanation to how the converter works.

cbouy:
For tackling the remaining known failures (molecules containing `R=[N+](-[O-])-R`, `R-[N+]#N`, `R-N=[N+]=[N-]`, `R-[O+]=R`, `R-[N+]#[C-]`, `R-[N-]-R`) should I stick to this PR or open a new one?
I feel like this PR is already getting pretty lengthy and sprawling (my fault) so I'd rather get this essential part merged, then run and publish the benchmark, and see later for the rest as I'm not sure how available I'll be in the near future

lilyminium:
@cbouy yeah I'd make a new branch/PR.

orbeckst:
Could you please open an issue for the remaining known failures https://github.com/MDAnalysis/mdanalysis/pull/3044#issuecomment-848331158, @cbouy ? Comments get lost, issues can get addressed (eventually). They also provide documentation for what we _know_ is not working properly and can serve as a rallying point for users who experience problems. Thanks.

orbeckst:
An additional way to document the failing patterns would be to include tests that XFAIL. Then in the next step, one "just" has to add code to make these tests pass.

IAlibay:
@cbouy @orbeckst @lilyminium what do we want to do with this re: 2.0? I'd like to get 2.0 out ASAP, can this be pushed to 2.1 instead? We could include a warning in the docstring of the converter that would go somewhere along the lines of "this is experimental, improved accuracy is coming in a future release"?

orbeckst:
An "experimental" warning should be sufficient, perhaps with a link to this issue.

IAlibay:
@cbouy @lilyminium @orbeckst what needs to be done here to get this merged on time for 2.1.0?

If any @MDAnalysis/coredevs want to take over pushing this along it would be great.

cbouy:
@IAlibay I think I needed to find one or two more example molecules to have perfect coverage.
Is there a deadline for 2.1.0?

IAlibay:
> @IAlibay I think I needed to find one or two more example molecules to have perfect coverage. Is there a deadline for 2.1.0?

It's pretty much "as soon as" - I think the original deadline was 2 months ago 😅

cbouy:
All right, I'll give it a shot in the coming days then

cbouy:
Aaaaand... I think that's it! Ready for review
salt bae

IAlibay:
@lilyminium @orbeckst can you check that your comments have been addressed here?

orbeckst (changes requested):
I don't have the expertise to give this a thorough review so I am commenting in general.

- Please add a CHANGELOG entry.
- Are there still cases that fail? If so, can you document these cases somehow — either mention in the docs or include XFAIL tests?

package/MDAnalysis/converters/RDKit.py
```diff
@@ -325,9 +330,9 @@ def convert(self, obj, cache=True, NoImplicit=True, max_iter=200,
                 # assign coordinates
                 conf = Chem.Conformer(mol.GetNumAtoms())
                 for atom in mol.GetAtoms():
-                    idx = atom.GetIntProp("_MDAnalysis_index")
+                    idx = atom.GetIdx()
```
lilyminium, line 353: Just commenting to think out loud -- this is fine because `_MDAnalysis_index` is set sequentially in `atomgroup_to_mol`.

package/MDAnalysis/converters/RDKit.py
```diff
@@ -435,9 +440,16 @@ def atomgroup_to_mol(ag, NoImplicit=True, max_iter=200, force=False):
         # infer bond orders and formal charges from the connectivity
         _infer_bo_and_charges(mol)
         mol = _standardize_patterns(mol, max_iter)
+        # reorder atoms to match MDAnalysis
+        order = np.argsort([atom.GetIntProp("_MDAnalysis_index")
+                            for atom in mol.GetAtoms()])
+        mol = Chem.RenumberAtoms(mol, order.astype(int).tolist())
```
lilyminium: Question: Is this necessary? Shouldn't _Mdanalysis_index have been set sequentially above in the `enumerate` call?

lilyminium: Also, `MDAnalysis_index` is a bit ambiguous in the sense that I don't really know if it's the `atom.index` property or index in that particular atom group. Possibly out of scope for this PR.

package/MDAnalysis/converters/RDKit.py
```diff
@@ -475,30 +487,68 @@ def _add_mda_attr_to_rdkit(attr, value, mi):
...
+            # more than two --> cut to 4 characters
             else:
-                value = "{:>2.2}{:<2.2}".format(symbol, number)
+                symbol, number = name
+                value = f"{symbol + number:>4.4}"
+        # no number in the name
         else:
-            # no number in the name
-            value = " {:<}".format(name[0])
+            value = f" {name[0]:<3.3}"
```
lilyminium: I'm guessing this is for RDKit's PDB writer, but it's IMO suboptimal to edit the names like this. It'd be nice to set another property with `SetProp` without this character limit, if this property is used for conversion from RDKit into MDAnalysis as well.

package/MDAnalysis/converters/RDKit.py
```diff
@@ -693,27 +774,30 @@ def _run_reaction(reaction, reactant):
...
-    for n in range(n_matches):
-        products = rxn.RunReactants((reactant,))
-        # only keep the first product
+    # repeat the reaction until all matching moeities have been transformed
+    # note: this loop is meant to be ended by a `break` statement
+    # but let's avoid using `while` loops just in case
+    for n in range(reactant.GetNumAtoms()):
+        reactant.UpdatePropertyCache(strict=False)
+        Chem.Kekulize(reactant)
+        products = reaction.RunReactants((reactant,))
```
lilyminium, line 795: Would this be a good spot for `RunReactantInPlace`? Would that then save the property updating?

package/MDAnalysis/core/selection.py
```diff
@@ -679,12 +679,10 @@ def _apply(self, group):
...
         matches = mol.GetSubstructMatches(pattern, useChirality=True)
-        # convert rdkit indices to mdanalysis'
-        indices = [
-            mol.GetAtomWithIdx(idx).GetIntProp("_MDAnalysis_index")
-            for match in matches for idx in match]
+        # flatten all matches and remove duplicated indices
+        indices = np.unique([idx for match in matches for idx in match])
         # create boolean mask for atoms based on index
-        mask = np.in1d(range(group.n_atoms), np.unique(indices))
+        mask = np.in1d(range(group.n_atoms), indices)
```
lilyminium, line 685: LGTM. this works because the entire group is converted and therefore RDKit index should == MDAnalysis_index (relative to `group`).

package/MDAnalysis/topology/guessers.py
```diff
@@ -518,7 +516,6 @@ def guess_gasteiger_charges(atomgroup):
     mol = atomgroup.convert_to("RDKIT")
     from rdkit.Chem.rdPartialCharges import ComputeGasteigerCharges
     ComputeGasteigerCharges(mol, throwOnParamFailure=True)
-    atoms = sorted(mol.GetAtoms(),
-                   key=lambda a: a.GetIntProp("_MDAnalysis_index"))
-    return np.array([atom.GetDoubleProp("_GasteigerCharge") for atom in atoms],
+    return np.array([atom.GetDoubleProp("_GasteigerCharge")
+                     for atom in mol.GetAtoms()],
                     dtype=np.float32)
```
lilyminium, line 521: LGTM

testsuite/MDAnalysisTests/converters/test_rdkit.py
```diff
@@ -204,13 +189,12 @@ def test_monomer_info(self, pdb, sel_str, atom_index):
...
         umol = sel.convert_to.rdkit(NoImplicit=False)
         atom = umol.GetAtomWithIdx(atom_index)
-        mda_index = atom.GetIntProp("_MDAnalysis_index")
         mi = atom.GetMonomerInfo()
 
         for mda_attr, rd_attr in RDATTRIBUTES.items():
             rd_value = getattr(mi, "Get%s" % rd_attr)()
             if hasattr(sel, mda_attr):
-                mda_value = getattr(sel, mda_attr)[mda_index]
+                mda_value = getattr(sel, mda_attr)[atom_index]
```
lilyminium: I find this somewhat hard to read, to be honest. I don't think this class is inherited, so could you please just type out the attributes you expect the AtomGroup to have?

lilyminium: i.e.

```
assert sel.altLocs[atom_index] == monomer_info.GetAltLoc()
assert sel.names[atom_index] == monomer_info.GetName().strip()
```

lilyminium (changes requested):
Sorry, this is a partial review. I think everything looks good to me so far, but I need a bit of time to wrap my head around the chemoinformatics part.

package/MDAnalysis/converters/RDKit.py
lilyminium: I might also have missed sometihng but I don't fully understand the left 3-column justification if there's no number in the name -- is this an RDKit quirk? Otherwise you could maybe just use the `center` function: `f"{value.center(4):4.4}"`

package/MDAnalysis/converters/RDKit.py
```diff
@@ -618,60 +690,69 @@ def _standardize_patterns(mol, max_iter=200):
...
-    return mol
+    # reassign all properties
+    _transfer_properties(mol, newmol)
+
+    # fix bonds with "crossed" stereo
+    for bond in newmol.GetBonds():
+        if bond.GetStereo() == Chem.BondStereo.STEREOANY:
+            bond.SetStereo(Chem.BondStereo.STEREONONE)
+
+    return newmol
```
lilyminium: Is this because the reactions might have changed altered the stereochemistry? Do you mind commenting that? What would be the harm in leaving them as Stereoany?

package/MDAnalysis/converters/RDKit.py
lilyminium: Oh wait, I'm guessing this is because the RDKit reactions will have messed up the order. Would you mind adding a comment to that effect?

testsuite/MDAnalysisTests/converters/test_rdkit.py
```diff
@@ -397,9 +378,36 @@ def test_cache(self):
...
+        template = Chem.AddHs(mol)
+        for atom in template.GetAtoms():
+            atom.SetIsAromatic(False)
+            atom.SetFormalCharge(0)
+            atom.SetNoImplicit(True)
+        for bond in template.GetBonds():
+            bond.SetIsAromatic(False)
+            bond.SetBondType(Chem.BondType.SINGLE)
+        template.UpdatePropertyCache(strict=False)
+        return template
```
lilyminium: Does template have a special meaning here? If not, could you please rename this to something like `remove_bond_orders_and_charges` or something a bit more descriptive for the tests?

testsuite/MDAnalysisTests/converters/test_rdkit.py
```diff
@@ -484,82 +498,188 @@ def test_set_atom_property(self, attr, value, getter):
...
+        "[O-2]",
+        "[Na+].[Cl-]",
+    ])
+    def test_ions(self, smi):
+        ref = Chem.MolFromSmiles(smi)
+        mol = self.mol_to_template(ref)
+        _infer_bo_and_charges(mol)
+        mol = _standardize_patterns(mol)
+        Chem.SanitizeMol(mol)
+        assert mol.HasSubstructMatch(ref) and ref.HasSubstructMatch(mol)
```
lilyminium: Test LGTM

testsuite/MDAnalysisTests/converters/test_rdkit.py
```diff
@@ -484,82 +498,188 @@ def test_set_atom_property(self, attr, value, getter):
...
+            ref = Chem.MolFromSequence(input_str)
+        template = self.mol_to_template(ref)
         # go through each possible starting atom
         for a in template.GetAtoms():
             smi = Chem.MolToSmiles(template, rootedAtAtom=a.GetIdx())
             m = Chem.MolFromSmiles(smi, sanitize=False)
             for atom in m.GetAtoms():
                 atom.SetFormalCharge(0)
+                atom.SetIsAromatic(False)
                 atom.SetNoImplicit(True)
```
lilyminium: Doesn't `mol_to_template` already do this?

testsuite/MDAnalysisTests/converters/test_rdkit.py
```diff
@@ -484,82 +498,188 @@ def test_set_atom_property(self, attr, value, getter):
...
             m.UpdatePropertyCache(strict=False)
             _infer_bo_and_charges(m)
             m = _standardize_patterns(m)
```
lilyminium: This shows up in a few different places. Would it be easier to test and use if there was an overall wrapping function e.g. below (probably with a better name)?

```
def _cheminformagic(rdmol):
    _infer_bo_and_charges(rdmol)
    _standardize_patterns(rdmol)
```

testsuite/MDAnalysisTests/converters/test_rdkit.py
```diff
@@ -484,82 +498,188 @@ def test_set_atom_property(self, attr, value, getter):
...
-                m), (f"(input) {Chem.MolToSmiles(ref)} != "
-                     f"{Chem.MolToSmiles(m)} (output) root atom {a.GetIdx()}")
+            match = m.HasSubstructMatch(ref) and ref.HasSubstructMatch(m)
+            if not match:
+                # try resonance structures for charged conjugated systems
+                for mol in Chem.ResonanceMolSupplier(m, maxStructs=20):
+                    match = (mol.HasSubstructMatch(ref) and
+                             ref.HasSubstructMatch(mol))
+                    if match:
+                        break
```
lilyminium: Would it be difficult to break out the test cases where you run into resonance problems, and explicitly test that resonance structure, in a separate test? i.e. `def test_order_independent_resonance`

testsuite/MDAnalysisTests/converters/test_rdkit.py
```diff
@@ -484,82 +498,188 @@ def test_set_atom_property(self, attr, value, getter):
...
-
+        if input_type == "smi":
+            ref = Chem.MolFromSmiles(input_str)
+        elif input_type == "na":
+            ref = Chem.MolFromSequence(input_str, flavor=9)
+        else:
+            ref = Chem.MolFromSequence(input_str)
+        template = self.mol_to_template(ref)
         # go through each possible starting atom
         for a in template.GetAtoms():
```
lilyminium: This test has a lot of logic in it, to the point where the test itself almost needs a test. Could you turn the `for` loop into a separate function, e.g. `_assert_order_independent`, and separate the smi/na/aa molecules into separate tests?

lilyminium: Or you could just use smiles for the acids, unless testing the RDKit sequence is important / it's likely to change internally in coming releases.

testsuite/MDAnalysisTests/converters/test_rdkit.py
```diff
@@ -484,82 +498,188 @@ def test_set_atom_property(self, attr, value, getter):
...
+            atom.SetIntProp("_MDAnalysis_index", ix)
+            atom.SetProp("dummy", f"foo_{ix}")
+            old[ix] = {"_MDAnalysis_index": ix, "dummy": f"foo_{ix}"}
+        _infer_bo_and_charges(mol)
+        newmol = _standardize_patterns(mol)
+        new = {}
+        for a in newmol.GetAtoms():
+            ix = a.GetIntProp("_MDAnalysis_index")
+            new[ix] = {"_MDAnalysis_index": ix,
+                       "dummy": a.GetProp("dummy")}
```
lilyminium, line 561:
```suggestion
                       "dummy": a.GetProp(f"foo_{ix}")}
```
Otherwise I'm not sure this tests much.

testsuite/MDAnalysisTests/converters/test_rdkit.py
```diff
@@ -471,8 +479,14 @@ def test_standardize_patterns(self, smi, out):
...
     @pytest.mark.parametrize("attr, value, getter", [
         ("index", 42, "GetIntProp"),
+        ("index", np.int8(42), "GetIntProp"),
+        ("index", np.int16(42), "GetIntProp"),
         ("index", np.int32(42), "GetIntProp"),
         ("index", np.int64(42), "GetIntProp"),
+        ("index", np.uint8(42), "GetIntProp"),
+        ("index", np.uint16(42), "GetIntProp"),
+        ("index", np.uint32(42), "GetIntProp"),
+        ("index", np.uint64(42), "GetIntProp"),
```
lilyminium, line 517: At this rate it might be worth just writing a separate test for getting `index` specifically.

testsuite/MDAnalysisTests/converters/test_rdkit.py
```diff
@@ -484,82 +498,188 @@ def test_set_atom_property(self, attr, value, getter):
...
                 atom.SetNoImplicit(True)
             m.UpdatePropertyCache(strict=False)
             _infer_bo_and_charges(m)
             m = _standardize_patterns(m)
             Chem.SanitizeMol(m)
             m = Chem.RemoveHs(m)
-            assert m.HasSubstructMatch(ref) and ref.HasSubstructMatch(
-                m), (f"(input) {Chem.MolToSmiles(ref)} != "
-                     f"{Chem.MolToSmiles(m)} (output) root atom {a.GetIdx()}")
+            match = m.HasSubstructMatch(ref) and ref.HasSubstructMatch(m)
```
lilyminium: This is short, but it might still be nice to make it a function with a descriptive name, like `assert_is_isomorphic` or something. It took me a minute to work out that this is the actual test where you put in a template and get out the right structure. A docstring would be extremely helpful for this test too.

package/MDAnalysis/converters/RDKit.py
```diff
@@ -618,60 +690,69 @@ def _standardize_patterns(mol, max_iter=200):
...
+    for rxn in [
+        "[C-;X2;H0:1]=[O:2]>>[C+0:1]=[O:2]",  # Cterm
+        "[N-;X2;H1;$(N-[*^3]):1]>>[N+0:1]",  # Nterm
+        "[#6-:1]-[#6:2]=[O:3]>>[#6+0:1]=[#6:2]-[O-:3]",  # keto-enolate
+        "[C-;v3:1]-[#7+0;v3;H2:2]>>[#6+0:1]=[#7+:2]",  # ARG
+        "[#6+0;H0:1]=[#6+0:2]-[#7;X3:3]-[#6-;X3:4]"
+        ">>[#6:1]=[#6:2]-[#7+:3]=[#6+0:4]",  # HIP
+        "[S;D4;!v6:1]-[*-:2]>>[S;v6:1]=[*+0:2]",  # sulfone
+        "[#7+0;X3:1]-[*-:2]>>[#7+:1]=[*+0:2]",  # charged-N
+    ]:
```
lilyminium, line 749: Screen Shot 2022-03-02 at 2 08 46 am

I think this all LGTM, although I needed to convince myself that the `[C-]-N` to `[C]=[N+]` transformations generally applicable.

cbouy:
Thanks a ton for reviewing this again @lilyminium and @orbeckst 👍
I'll give it a go in the coming weeks

package/MDAnalysis/converters/RDKit.py
cbouy: > Also, `MDAnalysis_index` is a bit ambiguous in the sense that I don't really know if it's the `atom.index` property or index in that particular atom group.

It's the index in that specific atomgroup, I'll add a note on that.


> Oh wait, I'm guessing this is because the RDKit reactions will have messed up the order. Would you mind adding a comment to that effect?

Yup!

package/MDAnalysis/converters/RDKit.py
cbouy: I think a better approach would be to just reuse the `PDBWriter._deduce_PDB_atom_name` method instead of reinventing the wheel. It requires residue names to work though, so I'll need to refactor my code a bit but it's for the best.

Should I add a `_MDAnalysis_name` property to keep the original name just in case?

> I don't fully understand the left 3-column justification if there's no number in the name -- is this an RDKit quirk? 

There's no specific PDB or RDKit rule for atom names without numbers, it's just that I originally organized my code in "names with numbers" vs "names without numbers" because I thought that was a more pragmatic approach, although there's likely a much better way to do it.

> Otherwise you could maybe just use the `center` function: `f"{value.center(4):4.4}"`

`OXT` would become `OXT␣` instead of `␣OXT`

package/MDAnalysis/converters/RDKit.py
cbouy: I think this isn't needed anymore, I introduced it when the converter was just reading the topology (so very early in the GSoC project). Since then I added parsing coordinates and assigning stereo from the coordinates so I'll just remove it and run the tests.

package/MDAnalysis/converters/RDKit.py
cbouy: Yes, although it's been introduced in RDKit v2021.09.1 so I'm not sure if it's a good idea to implement it in the code right now.  
It will very likely speed things up a ton (and simplify the codebase), but we can focus on optimization later I think ?

testsuite/MDAnalysisTests/converters/test_rdkit.py
cbouy: you're right, only `atom.SetNoImplicit(True)` is needed

testsuite/MDAnalysisTests/converters/test_rdkit.py
cbouy: I think this would mostly concern the molecules with a charge next to a conjugated system, which is a good portion of the tests here, but I'm not sure how separating the tests is going to make things more readable. It would IMO make things more difficult when/if we later add other tests.
I think a good option is to reorganize the test code with more functions with descriptive names, as you suggested in other comments

testsuite/MDAnalysisTests/converters/test_rdkit.py
cbouy: Nope the test is fine, the name of the property is "dummy" and its value should be "foo_{ix}", so to retrieve the value we just call `atom.GetProp("dummy")`

cbouy:
Okay I should be done with this one now!

I realized that there was a better way to test if the mol matches one of the resonance structure, so I implemented this change.

I added the XFAIL tests as @orbeckst suggested

Thanks again for reviewing this!!

cbouy:
I'm also preparing the benchmark on ChEMBL for the converter, should I create a new `RDKitConverter-benchmark` repository inside MDAnalysis then?

orbeckst:
Yes, create the repo under the MDAnalysis org. (I _think_ you have permission to do so but if not say something.)

package/MDAnalysis/converters/RDKit.py
```diff
@@ -53,7 +53,7 @@
     infers the structures of approximately 90% of the `ChEMBL27`_ dataset.
```
orbeckst: Should this be updated to 
> 99.1% of the 1.7M compounds in the CHEMBL27 dataset

package/CHANGELOG
```diff
@@ -13,13 +13,23 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" http://semver.org
 
 ------------------------------------------------------------------------------
-??/??/?? IAlibay
+??/??/?? IAlibay, cbouy
 
  * 2.2.0
 
 Fixes
+  * When converting an AtomGroup to an RDKit molecule (PR #3044):
```
orbeckst, line 21: Does this fix #3339 ? If so, the issue number should appear here. Even if it's only a partial fix — issue numbers are generally more important than PRs although we typically cite _both_ when possible.

package/MDAnalysis/converters/RDKit.py
```diff
@@ -53,7 +53,7 @@
...
     Work is currently ongoing on further improving this and updates to the
     converter are expected in future releases of MDAnalysis.
-    Please see `Pull Request #3044`_ for more details.
+    Please see `Pull Request #3044`_ and `Issue #3339`_ for more details.
```
orbeckst: I think the link to the PR can be removed. Better write some text that describes the current capabilities and lists the remaining known limitations. If there's an issue for these limitations, link to the issue.

package/MDAnalysis/converters/RDKit.py
```diff
@@ -262,9 +267,32 @@ class RDKitConverter(base.ConverterBase):
...
+
+    Their are some molecules containing specific patterns that the converter
+    cannot currently tackle correctly. See
+    `Issue #3339 <https://github.com/MDAnalysis/mdanalysis/issues/3339>`__ for
+    more info.
 
     .. versionadded:: 2.0.0
 
+    .. versionchanged:: 2.2.0
+        See the full changelog for more details:
```
orbeckst: Please don't reference the CHANGELOG. The entry should be self-contained and provide the most important points. I also don't think that item lists properly format. They seem to be typeset separately. Just write prose over multiple lines. 

I don't think that the SanitizationError fix requires a "versionchanged" entry — this sounds like a bug fix to me. Mostly the versionchanged tells users if they need to be aware of changes in usage or output. (Although, if in dount, provide rather more information than less ;-) ).

package/MDAnalysis/converters/RDKit.py
orbeckst: Is the whole warning still necessary or can it become a Note or just text?

package/MDAnalysis/converters/RDKit.py
```diff
@@ -277,8 +305,8 @@ def convert(self, obj, cache=True, NoImplicit=True, max_iter=200,
 
         Parameters
         -----------
-        obj : :class:`~MDAnalysis.core.groups.AtomGroup` or :class:`~MDAnalysis.core.universe.Universe`
-
+        obj : :class:`~MDAnalysis.core.groups.AtomGroup` or
+            :class:`~MDAnalysis.core.universe.Universe`
```
orbeckst: has to be on one line to format correctly, see https://mdanalysis--3044.org.readthedocs.build/en/3044/documentation_pages/converters/RDKit.html#MDAnalysis.converters.RDKit.RDKitConverter.convert

orbeckst: It might work if you just say
```
obj : `AtomGroup` or `Universe`
```
Try it and see if it links correctly.

testsuite/MDAnalysisTests/converters/test_rdkit.py
```diff
@@ -529,37 +581,128 @@ def test_reassign_props_after_reaction(self, rdmol, product, name):
...
         smi = "[C-]C=CC=CC=CC=CC=CC=C[C-]"
         mol = Chem.MolFromSmiles(smi)
         with pytest.warns(UserWarning,
                           match="reasonable number of iterations"):
             _rebuild_conjugated_bonds(mol, 2)
+
+    @pytest.mark.parametrize("smi", [
+        "[Li+]", "[Na+]", "[K+]", "[Rb+]", "[Ag+]", "[Cs+]",
+        "[Mg+2]", "[Ca+2]", "[Cu+2]", "[Zn+2]", "[Sr+2]", "[Ba+2]",
+        "[Al+3]", "[Fe+2]",
```
orbeckst, line 662: What happens for `"[Fe+3]"` or `"[Cu+1]"`?

orbeckst (changes requested):
Looks very impressive. I could only comment on documentation aspects and one question about cations.

lilyminium (approved):

lilyminium:
Aside from @orbeckst's comments the code looks great -- thank you for putting up with the super slow reviews! Agreed with Oliver, the guessing code is very impressive and it'll be a fantastic solution for so many people trying to create molecular graphs out of lossy input like xyz files.

orbeckst:
@cbouy I don't know how busy you are at the moment but if you could have a look at my minor comments then we could merge this big PR really soon™.

testsuite/MDAnalysisTests/converters/test_rdkit.py
orbeckst: If nothing else, document known failure cases or make choices explicit in the docs. E.g., it's fine to say "Any iron ion will be guessed to be Fe3+" etc. We just need to be clear about what happens.

cbouy:
@orbeckst I'm in the middle of moving out but I should be able to find some time after Monday! 
Thanks again to both of you!

package/MDAnalysis/converters/RDKit.py
cbouy: > Should this be updated to
> 
> > 99.1% of the 1.7M compounds in the CHEMBL27 dataset

Yes! good catch

> Is the whole warning still necessary or can it become a Note or just text?

I think it's necessary to keep the warning, just to catch people's attention that it's not a perfect solution and there are a few cases that are known to fail. I'll update this part a bit though.

package/CHANGELOG
cbouy: No, #3339 keeps track of the cases that this PR cannot fix

testsuite/MDAnalysisTests/converters/test_rdkit.py
cbouy: > What happens for `"[Fe+3]"` or `"[Cu+1]"`?

Those will always be converted to Fe2+ and Cu2+ with the code in this PR. We could solve this by looking at naming conventions for these ions but that's for another PR I guess. I'll add it to issue #3339 though

package/MDAnalysis/converters/RDKit.py
orbeckst: The back-tic notation did not automagically generate links so your previous solution was better. I'd revert to it.

orbeckst (approved):
Thanks for answering my questions — looks good.

(There's one change that I would revert so that links are generated in the docs but that's not blocking.)

Great contribution!!

orbeckst:
@lilyminium or @IAlibay can you please be in charge of eventually merging? Thanks!

cbouy:
Reverted the docstring thingy (and fixed a typo)!

Thanks a ton @orbeckst @lilyminium and @IAlibay for your help!!

supernatural

orbeckst:
@lilyminium @IAlibay is something holding up the merge? I just don't know, otherwise I'd press the shiny green button myself.

(Sorry to be onerous, I just want to see this code in before anything else happened.)

IAlibay:
Sorry just really busy atm, I'll have a quick glance through over the weekend and then merge.

lilyminium:
No -- just had another quick glance and everything still LGTM! @IAlibay did you still want to have a look? Otherwise I can merge.

IAlibay:
> No -- just had another quick glance and everything still LGTM! @IAlibay did you still want to have a look? Otherwise I can merge.

Go for it!

orbeckst:
Hooray!!!!

PR #3051, opened 2020-12-02
Add gh-action docs ♨️🥕

### LINKED ISSUE #3046: Deploy docs via GH actions

Travis is dead to us (https://github.com/MDAnalysis/mdanalysis/issues/3036), but it's currently how we deploy docs.

Deploy docs via GH actions.

I have already tried and failed to use ReadTheDocs due errors that crop up from the structure of the MDA repo (where the normal top-level repo is in "package/"); I will re-investigate but at the time, the RTD build structure wasn't flexible enough to account for this quirk. It would be nice to get an RTD pipeline going even if we don't use it for actual docs, as it can add a check to GH PRs.

## Additional context ##

Follow on from https://github.com/MDAnalysis/mdanalysis/pull/3040

It would be neat if we could inspect for doc diffs and only deploy when that happens. Not certain if that is possible.

lilyminium:
Fixes #3046 

 - Changes deployment scripts to get run by gh-actions
 - This could likely be neater and nicer (e.g. not exporting all the environment variables), but I think these are the minimal changes required. 

[Built dev / 2.0.0 docs here](https://minium.com.au/mdanalysis/dev/)

Targeted to #3040 but could target to develop instead of we prefer to keep this separate.

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


.github/workflows/gh-ci.yaml
```diff
@@ -212,8 +212,16 @@ jobs:
     - name: deploy_docs
       if: github.event_name != 'pull_request'
       run: |
-        # place the deploy call here
-        echo "Oh, maple syrup roast parsnips [Richard Gowers]"
+        python -c 'import os,sys,fcntl; flags = fcntl.fcntl(sys.stdout, fcntl.F_GETFL); fcntl.fcntl(sys.stdout, fcntl.F_SETFL, flags&~os.O_NONBLOCK);'
```
IAlibay, line 215: I know this has been around forever, but mostly out of curiosity, what does it actually do?

.github/workflows/gh-ci.yaml
```diff
@@ -212,8 +212,16 @@ jobs:
...
+        cd package/MDAnalysis
+        export VERSION=$(python -c "import version; print(version.__version__)")
+        cd -
+        export GH_DOC_BRANCH=$GITHUB_REF
```
IAlibay, line 219: not that it really matters (especially since I think you still need to do the `cd -` either way), but these could be all put inside an `env:` block if you happened to think it was cleaner that way.

maintainer/deploy_docs.sh
```diff
@@ -44,12 +44,12 @@ cd ${MDA_DOCDIR} || die "Failed to 'cd ${MDA_DOCDIR}'. Run from the top level of
 mkdir ../${VERSION} && mv * ../${VERSION}
 
 git init
-git config user.name "${GIT_CI_USER}"
-git config user.email "${GIT_CI_EMAIL}"
+git config user.name github-actions
+git config user.email github-actions@github.com
```
IAlibay, line 48: I can't find the gh actions details on this, do you have a reference for these? I keep finding "github-action@users.noreply.github.com", but that's in random github repos :/

IAlibay (approved):
Overall lgtm, just a couple of questions.

I saw some commits linking to your own repo, I'm assuming that was testing & it works?

.github/workflows/gh-ci.yaml
orbeckst: I think there used to be a comment somewhere... related to hanging builds? @lilyminium had explained it to me and I promptly forgot.

maintainer/deploy_docs.sh
orbeckst: Do you think it matters? Didn't we use in the past "TravisCI@mdanlysis.org" – if anyone wants the commits associated with their name then they could put in theirs, e.g. lilyminium@users.noreply.github.com.

IAlibay: I guess probably not? This whole area of using gh actions to push to repos just seems sparingly documented (unless I'm missing a really important page), particularly when it comes to examples. But I guess you're right, git doesn't really care if the email address is real.

.github/workflows/gh-ci.yaml
```diff
@@ -212,8 +212,16 @@ jobs:
...
+        export GH_TOKEN=${{ secrets.GITHUB_TOKEN }}
```
orbeckst, line 220: Does GH-actions obscure encrypted tokens in the logs?

.github/workflows/gh-ci.yaml
orbeckst, line 219: Every env var (except the oauth2 token) could also become a commandline argument for the `deploy_docs.sh` script. However, as long as it's working the way it is I don't see much need to change it.

maintainer/deploy_docs.sh
```diff
@@ -44,12 +44,12 @@ cd ${MDA_DOCDIR} || die "Failed to 'cd ${MDA_DOCDIR}'. Run from the top level of
...
 
 mv ../${VERSION} $VERSION
 
-git remote add upstream "https://${GH_TOKEN}@${GH_REPOSITORY}"
+git remote add upstream "https://github-actions:${GH_TOKEN}@${GH_REPOSITORY}"
```
orbeckst, line 52: A user name is needed now? 

Does it in any way relate to the `user.name` config?

If this is the case then I would make the script more robust by also passing the name (as env var or commandline argument); I would avoid having hard-coded dependencies between the gh-actions script and the deploy script.

.github/workflows/gh-ci.yaml
orbeckst: Is this the old oauth2 token or a new one?

orbeckst (approved):
If it works then it's all cool.

See comments for potential improvements.

I'd be ok to have it as a PR against develop as opposed to increase #3040 .

IAlibay:
> I'd be ok to have it as a PR against develop as opposed to increase #3040 . EDIT: ... up to you and @IAlibay

I don't mind, I don't think it really matters either way. If #3040 is ready to merge now (was there anything else left @orbeckst?), then I'd say let's just go ahead and squash merge that one and then have this one target `develop`. My reasons mainly being a) if for some reason we don't merge this soon-ish, at least other PRs can be opened b) we could then more cleanly introduce a small change to the docs here and check that it worked as intended?

orbeckst:
I agree with you, @IAlibay : merge #3040 and then add this on top.

(We will have to backport both to master, right? But that's another 🍯  of 🐟 )

IAlibay:
> I agree with you, @IAlibay : merge #3040 and then add this on ~top~ EDIT develop.
> 
> (We will have to backport both to master, right? But that's another 🍯 of 🐟 )

Yup, it's on my todo list to open the master PR next :) I'd like to get 1.0.1 out there asap.

IAlibay:
Sorry I accidentally closed by deleting the gh-actions-ci branch 😱 I've re-added it for now just in case.

.github/workflows/gh-ci.yaml
lilyminium: I don't like those because I can never remember which order things are in 😅 Regardless it's probably better than cluttering up the environment. I think I will actually take @IAlibay's suggestion and move things to env / move the whole deployment within `gh-ci.yaml`, unless there are objections; the general pros for having a separate script (modularity, being able to deploy docs outside of CI) are somewhat negated by the heavy dependence on environment variables set in the CI.

.github/workflows/gh-ci.yaml
lilyminium: The [``secrets.GITHUB_TOKEN`` is automatically created by GH](https://docs.github.com/en/free-pro-team@latest/actions/reference/authentication-in-a-workflow#about-the-github_token-secret). But, [no](https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets#limits-for-secrets), Github does not redact secrets if they are printed.

maintainer/deploy_docs.sh
lilyminium: @IAlibay you're right I do see [the noreply](https://github.com/marketplace/actions/configure-git-credentials) credentials more often; I'm not sure where I stole the original from.

.github/workflows/gh-ci.yaml
lilyminium: So yeah this could be [an issue](https://github.community/t/how-to-see-my-git-secrets/123668/5) if someone is really determined to get it and knows how to google "how to see git secrets"

lilyminium: Update: Yes, you can echo `{{ secrets.GITHUB_TOKEN }}`, but it's a) [single-use](https://docs.github.com/en/free-pro-team@latest/actions/reference/authentication-in-a-workflow#about-the-github_token-secret) and b) read-only on forked repositories. This PR from my [alt account](https://github.com/lilyminium/mdanalysis/runs/1491001169#step:7:2796) fails to write to lilyminium/mdanalysis:gh-pages with a permission error.

lilyminium:
Superseded by #3053.

PR #3053, opened 2020-12-03
Deploy docs via GitHub Actions CI

### LINKED ISSUE #3046: Deploy docs via GH actions

lilyminium:
Fixes #3046

 - Supercedes #3051
 - Substitutes a job in gh-ci.yaml for the original deploy docs suite
 - This removes the need to pass variables around
 - Added comments

Commits referencing my site / repo were to test the deployment of docs, [as built here](https://minium.com.au/mdanalysis/dev/).

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


.github/workflows/gh-ci.yaml
```diff
@@ -211,9 +212,57 @@ jobs:
 
     - name: deploy_docs
       if: github.event_name != 'pull_request'
+      env:
```
lilyminium, line 215: Moved deployment to within this file because it saved a lot of variable checking and possibly insecure token passing

.github/workflows/gh-ci.yaml
```diff
@@ -211,9 +212,57 @@ jobs:
...
+        GH_USER: github-actions
+        GH_EMAIL: "github-action@users.noreply.github.com"
+        GH_REPOSITORY: "github.com/${{ github.repository }}.git"
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        URL: https://docs.mdanalysis.org
+
       run: |
-        # place the deploy call here
-        echo "Oh, maple syrup roast parsnips [Richard Gowers]"
```
orbeckst, line 216: Almost a shame to loose this citation ;-)

.github/workflows/gh-ci.yaml
```diff
@@ -211,9 +212,57 @@ jobs:
...
+        # set up environment variables
+        # cannot execute bash to make variables in env section
+        # export URL for the Python script $UPDATE_JSON
+        export URL
+        export VERSION=$(cd package/MDAnalysis; python -c 'import version; print(version.__version__)')
+        UPDATE_JSON=$(pwd)/maintainer/update_json_stubs_sitemap.py
+        BRANCH="${GITHUB_REF#refs/heads/}"
+
+        # the below turns off blocking as it causes large writes to stdout to fail
+        # (see https://github.com/travis-ci/travis-ci/issues/4704)
```
orbeckst, line 232: Thanks for the comment. Btw, I think what we're doing here we _enabled_ blocking mode again, in case it was switched to non-blocking, at least that's how I read https://github.com/travis-ci/travis-ci/issues/4704#issuecomment-348435959 and `flags&~os.O_NONBLOCK`.

Was this someone we only needed for Travis or is this a general problem, i.e., does gh-actions also have something weird running that switches to non-blocking mode?

orbeckst: P.S.: Even if we don't strictly need it here anymore, I would only ever comment the line and not delete it, just in case we need it ever again.

orbeckst (approved):
Nice, thank you!

Minor comments inline.

.github/workflows/gh-ci.yaml
lilyminium: Thanks for pointing that out @orbeckst; you're right, gh-actions did not have this problem the 2 times I ran the deployment (where Travis was previously failing every time). I've commented it out and left a note.

IAlibay (approved):
lgtm :)

PR #3058, opened 2020-12-07
Mol2 nobonds

### LINKED ISSUE #3057: MOL2 Reader and Parser don't accept files without bonds
## Expected behavior ##

Reading a MOL2 file that doesn't have any bonds (with only Na+ and Cl- atoms for instance) should be fine, as it is the case for PDB files.

## Actual behavior ##

It raises an Exception: `The mol2 block has no bonds`

The problem concerns both the MOL2Parser and MOL2Reader:
https://github.com/MDAnalysis/mdanalysis/blob/cf75dd6fe20e935c95fef57b0029382883ab0283/package/MDAnalysis/topology/MOL2Parser.py#L129-L131
https://github.com/MDAnalysis/mdanalysis/blob/cf75dd6fe20e935c95fef57b0029382883ab0283/package/MDAnalysis/coordinates/MOL2.py#L192-L194

I think those exceptions should both be removed.
Finally in the MOL2Parser, the Bonds attribute should only be added if bonds are present (just like in the PDBParser with #2832). Basically adding an `if bonds:` before this line:
https://github.com/MDAnalysis/mdanalysis/blob/cf75dd6fe20e935c95fef57b0029382883ab0283/package/MDAnalysis/topology/MOL2Parser.py#L183

And for the MOL2Writer, just writing an empty `@<TRIPOS>BOND` section if `obj.bonds` doesn't exist should do it?
https://github.com/MDAnalysis/mdanalysis/blob/cf75dd6fe20e935c95fef57b0029382883ab0283/package/MDAnalysis/coordinates/MOL2.py#L327-L338

I'll make a PR next week unless someone wants to give it a go before that

cbouy:
Fixes #3057 

 - MOL2 files without bonds can now be read and written

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


cbouy:
woops made a mess with git again

PR #3060, opened 2020-12-07
Add ReadTheDocs configuration for PR builds

lilyminium:
~Fixes #~

 - Adds RTD configuration
 - Add `environment.yml` for package installation
 - Imports MDAnalysis at the top of `conf.py` to facilitate non-breaking import

This would be handy for testing PR builds of docs, as in the user guide: https://github.com/MDAnalysis/UserGuide/pull/119 (thanks @richardjgowers for authenticating RTD for this!)

We could *also* just deploy docs via ReadTheDocs, if we're willing to accept a small ad:

Screen Shot 2020-12-08 at 9 49 54 am

Pros:
- This is more easily maintainable for docs deployment if devs ever [get hit by a bus](https://en.wikipedia.org/wiki/Bus_factor). 
- RTD comes with its own versioning system, including `stable` and `latest` endpoints. Docs are easily searchable and can be built into PDF and Epubs.
- Many, many projects use RTD (e.g. [GSD](https://gsd.readthedocs.io/en/stable/))
- We can point the URL to our own domain or just use mdanalysis.readthedocs.io if we wish.

Cons:
- Not as flexible. We couldn't do shenanigans with scripts as we do in gh-actions currently (although most of those shenanigans are replicating features that RTD does for you anyway). There does not seem to be a place to put custom commands; you can only install MDA with pip or setuptools, and then it runs `conf.py` for you.
- Small ad.
- Unlikely, but the continued existence of RTD is not guaranteed.

These and other cons may be why similar packages such as mdtraj have chosen to [roll their own solutions](https://github.com/mdtraj/mdtraj/blob/master/azure-pipelines.yml) for docs.

Example build for this PR: https://mdanalysislilyminium.readthedocs.io/en/rtd/

This could replace the docs building test in GH Actions, but I haven't yet broken docs to see how informative the error messages are.

Thoughts @MDAnalysis/coredevs?

PR Checklist
------------
 - [ ] Tests?
 - [x] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/maintainer/conda/environment.yml
+++ b/maintainer/conda/environment.yml
@@ -1,12 +1,9 @@
 name: mda-dev
 channels:
   - defaults
-  - biobuilds
   - conda-forge
 dependencies:
-  - biopython
   - chemfiles
-  - clustalw==2.1
   - codecov
   - cython
   - griddataformats
@@ -17,26 +14,19 @@ dependencies:
   - matplotlib==3.2.2
   - mmtf-python
   - mock
-  - netcdf4
   - networkx
   - numpy>=1.17.3
   - psutil
   - pytest
   - scikit-learn
   - scipy
-  - nbsphinx
-  - nglview
   - pip
   - plotly
-  - pybtex
-  - rdkit>=2020.03.1
-  - seaborn>=0.7.0
   - sphinx==1.8.5
   - tidynamics>=1.0.0
   - tqdm>=4.43.0
   - sphinxcontrib-bibtex
   - sphinx_rtd_theme
-  - tabulate
   - pip:
       - duecredit
       - parmed
--- a/maintainer/conda/environment.yml
+++ b/maintainer/conda/environment.yml
@@ -21,7 +21,6 @@ dependencies:
   - scikit-learn
   - scipy
   - pip
-  - plotly
   - sphinx==1.8.5
   - tidynamics>=1.0.0
   - tqdm>=4.43.0
--- a/maintainer/conda/environment.yml
+++ b/maintainer/conda/environment.yml
@@ -1,0 +1,46 @@
+name: mda-dev
+channels:
+  - defaults
+  - biobuilds
+  - conda-forge
+dependencies:
+  - biopython
+  - chemfiles
+  - clustalw==2.1
+  - codecov
+  - cython
+  - griddataformats
+  - gsd
+  - h5py==2.10.0
+  - hypothesis
+  - joblib>=0.12
+  - matplotlib==3.2.2
+  - mmtf-python
+  - mock
+  - netcdf4
+  - networkx
+  - numpy>=1.17.3
+  - psutil
+  - pytest
+  - scikit-learn
+  - scipy
+  - nbsphinx
+  - nglview
+  - pip
+  - plotly
+  - pybtex
+  - rdkit>=2020.03.1
+  - seaborn>=0.7.0
+  - sphinx==1.8.5
+  - tidynamics>=1.0.0
+  - tqdm>=4.43.0
+  - sphinxcontrib-bibtex
+  - sphinx_rtd_theme
+  - tabulate
+  - pip:
+      - duecredit
+      - parmed
+      - msmb_theme==1.2.0
+      - sphinx-sitemap==1.0.2
+      - git+https://github.com/MDAnalysis/mdanalysis@develop#egg=mdanalysis&subdirectory=package
+      - git+https://github.com/MDAnalysis/mdanalysis@develop#egg=MDAnalysisTests&subdirectory=testsuite
--- a/package/doc/sphinx/source/conf.py
+++ b/package/doc/sphinx/source/conf.py
@@ -15,6 +15,7 @@ import sys
 import os
 import platform
 import datetime
+import MDAnalysis as mda
 import msmb_theme  # for little versions pop-up
 # https://sphinx-rtd-theme.readthedocs.io/en/stable/
 import sphinx_rtd_theme
@@ -68,7 +69,7 @@ master_doc = 'index'
 # (take the list from AUTHORS)
 # Ordering: (1) Naveen (2) Elizabeth, then all contributors in alphabetical order
 #           (last) Oliver
-author_list = __import__('MDAnalysis').__authors__
+author_list = mda.__authors__
 authors = u', '.join(author_list[:-1]) + u', and ' + author_list[-1]
 project = u'MDAnalysis'
 now = datetime.datetime.now()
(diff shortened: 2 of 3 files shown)


maintainer/conda/environment.yml
```diff
@@ -0,0 +1,36 @@
...
+  - mmtf-python
+  - mock
+  - networkx
+  - numpy>=1.17.3
+  - psutil
+  - pytest
+  - scikit-learn
+  - scipy
+  - pip
+  - plotly
```
lilyminium, line 24: This is not a full list of dependencies (and actually, `plotly` shouldn't be here). When all dependencies were given, RTD errored out with a memory and time error on installation. Even now it's quite slow; if it happens more in the future we could probably remove everything but cython, numpy, and the sphinx/msmb_theme packages.

package/doc/sphinx/source/conf.py
```diff
@@ -68,7 +69,7 @@
 # (take the list from AUTHORS)
 # Ordering: (1) Naveen (2) Elizabeth, then all contributors in alphabetical order
 #           (last) Oliver
-author_list = __import__('MDAnalysis').__authors__
+author_list = mda.__authors__
```
lilyminium, line 72: Removing `__import__` was necessary for RTD to import MDAnalysis properly without dying because it couldn't find the C extensions.

orbeckst:
How does branding work, i.e., can we serve from *mdanalyis.org*?

lilyminium:
@orbeckst [yes](https://docs.readthedocs.io/en/stable/custom_domains.html#custom-domain-support), we can keep the ``docs.mdanalysis.org`` domain.

orbeckst:
Your RTD shows "RTD" as the version information in the top left corner. What would this look like for a develop build (where we see 2.0.0-dev)? 

It also does not show the ad.

orbeckst:
Note that if we were to make this change, we would need to update the section https://www.mdanalysis.org/pages/privacy/#who-will-your-information-be-shared-with of our Privacy Policy. In particular, if the ads include trackers of any kinds then we might need [GPRD](https://en.wikipedia.org/wiki/General_Data_Protection_Regulation) banners etc.

lilyminium:
> EDIT: that was my adblocker working as it is supposed to...

🤔 Clearly I need an upgrade.

> Your RTD shows "RTD" as the version information in the top left corner. What would this look like for a develop build (where we see 2.0.0-dev)?

This is a branch build, so it's "rtd" as the name of the branch and in the version selector.

Screen Shot 2020-12-08 at 11 57 30 am

On ``develop`` [it looks like](https://mdanalysislilyminium.readthedocs.io/en/develop/) (note for both this and RTD, it shows 2.0.0 in the little menu down the bottom until you click it to get the pop-out. I think that's a holdover from msmb_theme):
Screen Shot 2020-12-08 at 12 11 49 pm

And this is what the banner it shows for a pull request build (from the user guide PR): Both the "version" and the URL are the PR number, and it warns appropriately that it's a pull request build.

Screen Shot 2020-12-08 at 12 03 53 pm


Aaaand lucky last this is what "latest" points to [automatically on the UserGuide](https://mdanalysisuserguide.readthedocs.io/en/latest/), where the default branch is master:
Screen Shot 2020-12-08 at 12 05 28 pm

Hopefully I haven't caused any legal issues by having these extra user guide builds currently on ReadTheDocs? I can take them down.

> In particular, if the ads include trackers of any kinds

Ghostery tells me it's blocking a Google Analytics tracker. uBlock finds these domains:

Screen Shot 2020-12-08 at 12 01 17 pm

orbeckst:
> Hopefully I haven't caused any legal issues by having these extra user guide builds currently on ReadTheDocs? I can take them down.

I doubt that anybody would want to sue us/NF – there's no 💰 to be had. So I don't worry about these builds, which are not even linked from mdanalysis.org.

orbeckst:
Doc builds from PRs would be a major plus of this approach – does this work automatically? Can we then look at the doc build of an open PR to check the doc rendering?

orbeckst:
The Google Analytics tracker is annoying. Does this come from RTD??

lilyminium:
> Doc builds from PRs would be a major plus of this approach – does this work automatically? Can we then look at the doc build of an open PR to check the doc rendering?

Yes, if it works in this repo as it [does for the UserGuide](https://github.com/MDAnalysis/UserGuide/pull/119), then it becomes another check (so we could replace the Github Action doc building check) that updates for each push. Clicking "Details" when it works takes you to the built docs, when it fails takes you to the error message. If it becomes an issue to have a separate RTD build of the docs, I think we could actually turn off re-building on merge into branches too.

Screen Shot 2020-12-08 at 12 45 01 pm

> The Google Analytics tracker is annoying. Does this come from RTD??

Yes, [but we can disable it](https://docs.readthedocs.io/en/latest/guides/google-analytics.html#disabling-google-analytics-on-your-project)! I'm just disabling it on my fork now to see if the tracker goes away. The normal MDAnalysis project already has it disabled.

lilyminium:
Maybe it was cached; the Google Analytics tracker has disappeared now. If we just use it to check PRs and leave the builds hosted on readthedocs.io, I assume we wouldn't have to update our privacy policy anyhow?

package/doc/sphinx/source/conf.py
orbeckst: Well, if this works... no idea how this is different from `__import__()`.

maintainer/conda/environment.yml
```diff
@@ -0,0 +1,35 @@
...
+  - tidynamics>=1.0.0
+  - tqdm>=4.43.0
+  - sphinxcontrib-bibtex
+  - sphinx_rtd_theme
+  - pip:
+      - duecredit
+      - parmed
+      - msmb_theme==1.2.0
+      - sphinx-sitemap==1.0.2
+      - git+https://github.com/MDAnalysis/mdanalysis@develop#egg=mdanalysis&subdirectory=package
```
orbeckst, line 34: tricky!

orbeckst (approved):
Nice work, the amount of sleuthing that you must have done to figure out all the little things... thank you.

Is the next step to try and switch over the CNAME?

orbeckst:
Does this enable RTD only for PRs or also for the published docs? I thought only the former but just thought I'd ask again...

If this works then we can consider RTD for the published docs if this is what people want to do.

lilyminium:
> Does this enable RTD only for PRs or also for the published docs? I thought only the former but just thought I'd ask again...

Just the former. It does build the master branch as well by default, but we are not merging into master... 

The steps to point the develop docs to `docs.mdanalysis.org/develop` would be
* change the CNAME to point to RTD
* change the root URL on readthedocs.org to `docs.mdanalysis.org/`

so ideally no further PR will be needed. To point the main docs to `docs.mdanalysis.org/` we'd need to merge into master in addition to the above.

orbeckst:
Thanks!

PR #3061, opened 2020-12-08
Port GH actions and other fixes to master

lilyminium:
In aid of #2768.

Possibly this will help fix the last remaining obstacle, #3004?

Added all commits to `develop` that did not involve a `.. versionchanged:: 2.0.0` docstring. Please let me know if any were premature.

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


.github/workflows/gh-ci.yaml
```diff
@@ -0,0 +1,330 @@
...
+
+jobs:
+  main_tests:
+    if: "github.repository == 'MDAnalysis/mdanalysis'"
+    runs-on: ${{ matrix.os }}
+    strategy:
+        fail-fast: false
+        matrix:
+          os: [ubuntu-latest, ]
+          python-version: [3.6, 3.7, 3.8]
```
IAlibay, line 31: 1.0.x still covers py2.7, we'll need to add that to the matrix + the dependencies that come with it (six, etc...).

lilyminium: Thanks, good catch.

lilyminium:
The current thing I'm working on is that python 2.7 takes forever to install dependencies -- I haven't let it time out yet but it's been 1 hr+ stuck at that step. Not sure why because everything installs fine on my own device.

.github/workflows/gh-ci.yaml
```diff
@@ -0,0 +1,357 @@
...
+  MDA_CONDA_MIN_DEPS: "pip pytest mmtf-python biopython networkx cython matplotlib scipy griddataformats hypothesis gsd codecov"
+  MDA_CONDA_EXTRA_DEPS: "seaborn"
+  MDA_CONDA_PY3_DEPS: "netcdf4 scikit-learn joblib>=0.12 clustalw=2.1 chemfiles mock"
+  MDA_CONDA_PY27_DEPS: "clustalw=2.1 netcdf4=1.3.1 six=1.15.0 funcsigs<=1.0.2 mock<=3.0.5"
+  MDA_CONDA_PY35_DEPS: "netcdf4 scikit-learn joblib>=0.12 clustalw=2.1 mock<=3.0.5"
+  MDA_PIP_MIN_DEPS: 'coveralls coverage<5 pytest-cov pytest-xdist'
+  MDA_PIP_EXTRA_DEPS: 'duecredit parmed'
+  MDA_PIP_PY27_DEPS: "setuptools<45.0.0 chemfiles scikit-learn joblib>=0.12"
+  MDA_PIP_PY3_DEPS: ""
+  MDA_PIP_PY35_DEPS: ""
```
lilyminium, line 26: This and adding the 2.7 and 3.5 jobs are the primary changes I made. Py3.5 and 2.7 seemed to require a lot of pinning in gh actions for conda to resolve dependencies.

.github/workflows/gh-ci.yaml
```diff
@@ -0,0 +1,357 @@
...
+          conda-version: ["latest", ]
+          include:
+            - name: python-2.7
+              os: ubuntu-latest
+              python-version: 2.7
+              install_hole: true
+              codecov: true
+              numpy: numpy=1.16
+              extra_deps: "EXTRA PY27"
+              conda-version: 4.7.10
```
lilyminium, line 48: Pinning conda didn't really help in resolving dependencies but I found astropy's reason for pinning persuasive (conda is not always so stable from update to update)

.github/workflows/gh-ci.yaml
```diff
@@ -0,0 +1,357 @@
...
+        pip_deps="${MDA_PIP_MIN_DEPS}"
+
+        for dep in ${{ matrix.extra_deps }}; do
+          conda_var="MDA_CONDA_${dep}_DEPS"
+          pip_var="MDA_PIP_${dep}_DEPS"
+          echo "${!conda_var}"
+          conda_deps="$conda_deps ${!conda_var}"
+          echo $conda_deps
+          pip_deps="$pip_deps ${!pip_var}"
+        done
```
lilyminium, line 142: I changed this part from @IAlibay's original setup because setting up all the environment variables was a little bit of a pain. I am however happy to go back to the original construct now that we've figured out which dependencies to list.

.disabled-travis.yml
```diff
@@ -111,6 +111,31 @@ jobs:
            INSTALL_HOLE="false"
            CODECOV="true" SETUP_CMD="${PYTEST_FLAGS} --cov=MDAnalysis"
 
+    - os: linux
```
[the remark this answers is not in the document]
IAlibay: Perhaps @tylerjereddy has a better view of whether it fits in a near-future 1.0.1 release?

IAlibay (approved):
Thanks for working on this @lilyminium 

I have a few comments (particularly re: aarch64 inclusion), but I don't think any of them are blocking so I'll set it to approve.

.github/workflows/gh-ci.yaml
lilyminium: I think it broke for a 3.7 job?

.github/workflows/gh-ci.yaml
lilyminium: So conda never refused to resolve dependencies in the 1-2 hours I would let it run. In addition, on both my linux and mac machines it resolved the original unpinned versions easily within a couple minutes. Moreover, Travis didn't require pinning. I'm more inclined to think that this is a GH Action quirk than anything else. On the other hand, seeing as we *know* those versions work, I am up for it.

.github/workflows/gh-ci.yaml
lilyminium: Could do -- we could also have `environment.yml` files for pythons 2.7, 3.5, fully featured, and minimal. That also saves users from having to work this stuff out.

.github/workflows/gh-ci.yaml
IAlibay: How odd, is this the same version that the ci helpers were pinning to?

.github/workflows/gh-ci.yaml
IAlibay: So the reason I chose to avoid env files in `develop` is that they cause really big slowdowns for some reason, conda setup time was going from ~ 3 mins to ~ 15 mins. That and mamba 100% hates pip in environment files.

I'd prefer stay away from them on `develop`, but if we need them to clean up the matrix here, I don't have any objections (as long as CI actually completes).

.github/workflows/gh-ci.yaml
IAlibay: To be honest this sounds more like miniconda trickery more than GH issues :/ I don't know, I guess we can't release a new version that goes beyond tested limits.

lilyminium:
Superseded by #3067.

PR #3069, opened 2020-12-10
Add del_TopologyAttr

lilyminium:
Fixes #3068

Changes made in this Pull Request:
 - Add `del_TopologyAttr` that effectively undoes `add_TopologyAttr`

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?
 - [x] Issue raised/referenced?

PR #3071, opened 2020-12-10
Install MDAnalysis on ReadTheDocs via pip

### LINKED ISSUE #3070: ReadTheDocs isn't building the modified version of docs in pull requests
## Expected behavior ##

#3069 has a `del_TopologyAttr` method documented in the built docs by RTD.

## Actual behavior ##

It does not, because I mistakenly install the `develop` version of MDAnalysis instead of the current version. This makes the RTD check fairly useless.

``` python
import MDAnalysis as mda

u = mda.Universe(PSF, DCD)

....

```

- Which operating system?

lilyminium:
Fixes #3070 

 - install via `pip install package/` to build current docs.
 - Works, as tested on #3069 https://readthedocs.org/projects/mdanalysis/builds/12528298/

Not sure why `setuptools` (`python package/setup.py install`) doesn't work, but, it doesn't.

Sorry @orbeckst 🤦‍♀️

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -17,6 +17,9 @@ sphinx:
 # Optionally set the version of Python and requirements required to build your docs
 python:
   version: 3.7
+  install:
+    - method: setuptools
+      path: package
 
 conda:
   environment: maintainer/conda/environment.yml
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -18,7 +18,7 @@ sphinx:
 python:
   version: 3.7
   install:
-    - method: setuptools
+    - method: pip
       path: package
 
 conda:
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -16,7 +16,6 @@ sphinx:
 
 # Optionally set the version of Python and requirements required to build your docs
 python:
-  version: 3.7
   install:
     - method: pip
       path: package
--- a/maintainer/conda/environment.yml
+++ b/maintainer/conda/environment.yml
@@ -18,6 +18,7 @@ dependencies:
   - numpy>=1.17.3
   - psutil
   - pytest
+  - python==3.7
   - scikit-learn
   - scipy
   - pip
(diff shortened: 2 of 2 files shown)


cbouy:
If I may "intrude", since I'm in the process of using readthedocs for another project: RTD will use whatever you specified in the conda environment.yml file to set up the python env, and ignore both `version` and `requirements` parameters under the `python` parameter in the readthedocs.yml file. So you should probably set the python version in the env.yml file and remove the version line in the readthedocs file if you really want a specific version.

lilyminium:
Thank you @cbouy!

richardjgowers (approved):

orbeckst:
When I looked at the link https://readthedocs.org/projects/mdanalysis/builds/12528298/ I see this confusing message

> The project ParmEd doesn't have a valid webhook set up, commits won't trigger new builds for this project. See the project integrations for more information. 

Why is **ParmEd** mentioned here?

image

orbeckst (approved):
It's amazing how much nitty-gritty detail there is to "just make stuff work". Thank you so much for finding it out and _actually_ making it work.

lilyminium:
Thank you for the reviews @richardjgowers and @orbeckst! So I _think_ that RTD kindly gives you that message if your ParmEd project is not set up with a webhook, even if you're viewing another project -- could you have a look at a different project to see if that's the case? For what it's worth this is how it appears to me:

image

I think this may also only happen if you are logged in and an owner of MDAnalysis and a project called ParmEd at the time.

orbeckst:
You’re right – I see it on my login page, too. Hence has no bearing on your PR.

Sorry for making you explain this to me.


(The parmed was my emergency solution when parmed docs disappeared.)

lilyminium:
Thank you for checking! I did see more than a few of these messages when I was trying to work out ReadTheDocs myself 😅

PR #3073, opened 2020-12-10
Remove appveyor

### LINKED ISSUE #3065: drop AppVeyor CI
AppVeyor has been a great service for us, especially when it was the only one that provided Windows testing. The CI landscape has changed and AppVeyor has been consistently slower than GH-actions and Azure Pipelines. Furthermore, ci-helpers stopped active support #3017 .

I would suggest we remove AppVeyor testing from our CI on both 
- [x] develop and
- [ ] master.

Coverage might drop because AppVeyor uses conda installation whereas our Azure Pipelines use pip and are more limited in the available packages (e.g., RDKit is not on pip). 

Nevertheless, we can make the change now and raise a new issue to establish a Azure Pipeline with conda if necessary.

IAlibay:
Fixes #3065 

 - Quite simply; removes appveyor from MDA

I'll raise another issue about adding a conda run to the azure matrix (or we can add Windows to GH actions).

PR Checklist
------------
 - [x] Tests?
 - Docs?
 - CHANGELOG updated?
 - Issue raised/referenced?


IAlibay:
How odd, the webhook is still active even if you don't have the yaml file around 🙃 

I can delete it through, but I guess I should wait to hear any dissenting opinions on the issue/PR first?

tylerjereddy (approved):
Makes sense, Oli was in favor of this based on the matching issue and sounds like there is a plan in place to have a `conda`-based Windows run somewhere to make up for the dependency testing.

Probably safe to get rid of the Appveyor web hook too

IAlibay:
I've gone ahead deleted the webhook now, we just need to remove the yaml file from `master` too.

PR #3085, opened 2020-12-25
MAINT: simplify guessers regex

tylerjereddy:
* the `SYMBOLS` regex in `guessers.py` does not require
any escape sequences because the metacharacters are inactive
in the character class (this includes the range metacharacter
when placed at the start or end of the character class)


lilyminium (approved):
Thanks @tylerjereddy!

PR #3072, opened 2020-12-10
Fix syntax warning over comparison of literals using is.

tirkarthi:
Fixes #3066 

Changes made in this Pull Request:

Fix syntax warning over using is.

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?
 - [x] Issue raised/referenced?

PR #3086, opened 2020-12-27
Changing test_util to fix 'is' warning

aditya-kamath:
Fixes #

 - changed the test_util file to correct 'is' SyntaxWarning

PR Checklist
------------
 - [ x] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/testsuite/MDAnalysisTests/lib/test_util.py
+++ b/testsuite/MDAnalysisTests/lib/test_util.py
@@ -1474,7 +1474,7 @@ class TestStaticVariables(object):
 
         @static_variables(foo=0, bar={'test': x})
         def myfunc():
-            assert myfunc.foo is 0
+            assert myfunc.foo == 0
             assert type(myfunc.bar) is type(dict())
             if 'test2' not in myfunc.bar:
                 myfunc.bar['test2'] = "a"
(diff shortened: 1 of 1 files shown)


aditya-kamath:
Hi,
I am a student starting out as a developer. I see this issue already has a pull request, but I just want to try submitting my version. I would gladly take your comments

IAlibay:
Thanks for your PR @aditya-kamath. As you mentioned yourself, #3072 is already opened with the aim of addressing this issue. The approach we take here is that unless there is an agreed upon reason for a PR to supersede another, then the original PR is regarded as the one that we aim to merge into the code base. Should #3072 happen to stall (i.e. due to inactivity) we will then consider alternate PRs for the issue.

Whilst the author of #3072 has yet to address our review comments, we are also in the midst of the winter break. Hence, unless there is a pressing reason to address a given issue, we will likely wait until mid January before considering recent PRs as stalled.

P.S. To avoid PR collision, and to ensure your efforts are best used, we do encourage folks to first comment on existing issues/PRs to judge whether or not a given contribution is appropriate.

aditya-kamath:
Thank you @IAlibay. I live in Canada, so it is the winter break for me too. I am utilizing this time to test out my ability to contribute to some organizations with an ultimate aim for a GSOC proposal. I can take on a few 'starter' issues, however I don't want to interrupt your holidays; Perhaps I can work on some open issues without any PRs in the meantime

lilyminium:
Hi @aditya-kamath, thank you for having a go. There are definitely many starter issues that have not yet been addressed. You are very welcome to make a PR for these! We just ask that you please be understanding if we are not quick to review -- 2020 has been a long year and we are all taking a break :-) Similarly, do not feel pressured to address or reply to things immediately over this period, of course.

I will just close this for now, pending updates on #3072.

PR #3088, opened 2020-12-30
MAINT: char class regex improve

tylerjereddy:
* avoid the overhead of a regex character class
when that character class has only a single character
(i.e., serves no purpose)

* there is only one instance of this in MDA codebase
discovered by my [scraping code](https://github.com/tylerjereddy/regex-improve)

* for a longer explanation see my similar changes in
NumPy codebase:
https://github.com/numpy/numpy/pull/18083


orbeckst (approved):
Thanks @tylerjereddy !

Issue #2958, opened 2020-09-29
RDKIT tests sometimes fails


IAlibay:
## Expected behavior ##

All tests pass

## Actual behavior ##

In some cases (I believe @lilyminium first mentioned this, and I've seen this occur a few times on appveyor), the `TestRDKitConverter.test_assign_coordinates` can fail: https://ci.appveyor.com/project/orbeckst/mdanalysis/builds/35434315/job/4fl3rfbyde60rpqp

## Current version of MDAnalysis ##

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) current develop
- Which version of Python (`python -V`)? In the captured case: 3.6
- Which operating system? In the captured case: Win10 64

## Notes ##

This test failure is not deterministic, and I'm not sure what causes it to be triggered, maybe just hardware?

IAlibay:
It's done it again but this time on travis: https://travis-ci.com/github/MDAnalysis/mdanalysis/jobs/393115290

Again python 3.6...

IAlibay:
Another error, different test failing, not python 3.6: https://travis-ci.com/github/MDAnalysis/mdanalysis/jobs/393257863

Also tagging @cbouy who might have a better idea as to why we are getting random failures.

cbouy:
For `test_assign_coordinates`, I was gonna say it looks like the precision is at fault but now I just don't understand the shape mismatch `(shapes (42, 3), (1877, 3) mismatch)`... 42 is a rather specific number that I sometimes use in tests so maybe it has something to do with the converter cache somehow ? 42 also appears in the travis fail

IAlibay:
See here for another occasional RDKIT test failure: https://github.com/MDAnalysis/mdanalysis/pull/2983#issuecomment-707899696

IAlibay:
Here's another failure from #2997 

```
_______________ TestRDKitConverter.test_monomer_info[resid 1-0] ________________
self = <MDAnalysisTests.coordinates.test_rdkit.TestRDKitConverter object at 0x7f35d867ad68>
pdb = <Universe with 1877 atoms>, sel_str = 'resid 1', atom_index = 0
    @pytest.mark.parametrize("sel_str, atom_index", [
        ("resid 1", 0),
        ("resname LYS and name NZ", 1),
        ("resid 34 and altloc B", 2),
    ])
    def test_monomer_info(self, pdb, sel_str, atom_index):
[...]

PR #3091, opened 2021-01-06
Emit warnings for empty Ag and user grid in DensityAnalysis

### LINKED ISSUE #3055: DensityAnalysis with empty selection
Hello, I am trying to use DensityAnalysis for an atomselection that is initially empty but which becomes filled over the course of the trajectory. DensityAnalysis fails if the selection is initially empty despite using updating=True:

eg commands used:
```python
site=u.select_atoms("byres resname LIG and around 6 protein", updating=True)
D=DensityAnalysis(site, delta=1).run()
```
I have got around by filling with a random atom but obviously not ideal.

aditya-kamath:
Fixes #3055

 - Displays UserWarning when an empty atomgroup is given to DensityAnalysis

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ x] CHANGELOG updated?
 - [ x] Issue raised/referenced?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -21,8 +21,9 @@ The rules for this file:
   * 2.0.0
 
 Fixes
-  * UserWarning/ValueError displayed when empty atomgroup is given 
-    to DensityAnalysis (Issue #3055)
+  * ValueError raised when empty atomgroup is given to DensityAnalysis
+    without a user defined grid. UserWarning displayed when user defined 
+    grid is provided. (Issue #3055)
   * Fixed import of several `__all__` lists (PR #3013)
   * atommethods (_get_prev/next_residues_by_resid) returns empty residue group 
     when given empty residue group (Issue #3089)
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -16,13 +16,15 @@ The rules for this file:
 ??/??/?? tylerjereddy, richardjgowers, IAlibay, hmacdope, orbeckst, cbouy,
          lilyminium, daveminh, jbarnoud, yuxuanzhuang, VOD555, ianmkenney,
          calcraven，xiki-tempula, mieczyslaw, manuel.nuno.melo, PicoCentauri,
-         hanatok, rmeli, aditya-kamath
+         hanatok, rmeli, aditya-kamath, tirkarthi
 
   * 2.0.0
 
 Fixes
+
   * UserWarning displayed when empty atomgroup is given to DensityAnalysis (Issue #3055)
-  * atommethods(_get_prev/next_residues_by_resid) returns empty residue group when given empty residue group (Issue #3089)
+  * atommethods (_get_prev/next_residues_by_resid) returns empty residue group 
+    when given empty residue group (Issue #3089)
   * MOL2 files without bonds can now be read and written (Issue #3057)
   * Write `CONECT` record only once in multi-model PDB files (Issue #3018)
   * Add '.nc' as a recognised extension for the NCDFWriter (Issue #3030)
@@ -77,6 +79,7 @@ Fixes
     (Issue #2934)
   * Fix tests for analysis.bat that could fail when run in parallel and that
     would create a test artifact (Issue #2979, PR #2981)
+  * Fix syntax warning over comparison of literals using is (Issue #3066)
 
 Enhancements
   * Added automatic selection class generation for TopologyAttrs,
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -21,8 +21,8 @@ The rules for this file:
   * 2.0.0
 
 Fixes
-
-  * UserWarning displayed when empty atomgroup is given to DensityAnalysis (Issue #3055)
+  * UserWarning displayed when empty atomgroup is given 
+    to DensityAnalysis (Issue #3055)
   * atommethods (_get_prev/next_residues_by_resid) returns empty residue group 
     when given empty residue group (Issue #3089)
   * MOL2 files without bonds can now be read and written (Issue #3057)
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -21,7 +21,7 @@ The rules for this file:
   * 2.0.0
 
 Fixes
-  * UserWarning displayed when empty atomgroup is given 
+  * UserWarning/ValueError displayed when empty atomgroup is given 
     to DensityAnalysis (Issue #3055)
   * Fixed import of several `__all__` lists (PR #3013)
   * atommethods (_get_prev/next_residues_by_resid) returns empty residue group 
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -21,6 +21,7 @@ The rules for this file:
   * 2.0.0
 
 Fixes
+  * atommethods(_get_prev/next_residues_by_resid) returns empty residue group when given empty residue group (Issue #3089)
   * MOL2 files without bonds can now be read and written (Issue #3057)
   * Write `CONECT` record only once in multi-model PDB files (Issue #3018)
   * Add '.nc' as a recognised extension for the NCDFWriter (Issue #3030)
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -21,6 +21,7 @@ The rules for this file:
   * 2.0.0
 
 Fixes
+  * UserWarning displayed when empty atomgroup is given to DensityAnalysis (Issue #3055)
   * atommethods(_get_prev/next_residues_by_resid) returns empty residue group when given empty residue group (Issue #3089)
   * MOL2 files without bonds can now be read and written (Issue #3057)
   * Write `CONECT` record only once in multi-model PDB files (Issue #3018)
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -16,7 +16,7 @@ The rules for this file:
 ??/??/?? tylerjereddy, richardjgowers, IAlibay, hmacdope, orbeckst, cbouy,
          lilyminium, daveminh, jbarnoud, yuxuanzhuang, VOD555, ianmkenney,
          calcraven，xiki-tempula, mieczyslaw, manuel.nuno.melo, PicoCentauri,
-         hanatok, rmeli
+         hanatok, rmeli, aditya-kamath
 
   * 2.0.0
 
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -23,6 +23,7 @@ The rules for this file:
 Fixes
   * UserWarning displayed when empty atomgroup is given 
     to DensityAnalysis (Issue #3055)
+  * Fixed import of several `__all__` lists (PR #3013)
   * atommethods (_get_prev/next_residues_by_resid) returns empty residue group 
     when given empty residue group (Issue #3089)
   * MOL2 files without bonds can now be read and written (Issue #3057)
@@ -158,6 +159,7 @@ Changes
     `__call__` (Issue #2860, PR #2859)
   * deprecated ``analysis.helanal`` module has been removed in favour of
     ``analysis.helix_analysis`` (PR #2929)
+  * Move water bridge analysis from hbonds to hydrogenbonds (Issue #2739 PR #2913)
 
 Deprecations
 
--- a/package/MDAnalysis/analysis/density.py
+++ b/package/MDAnalysis/analysis/density.py
@@ -399,8 +399,10 @@ class DensityAnalysis(AnalysisBase):
                 smin = np.min(coord, axis=0)
                 smax = np.max(coord, axis=0)
             except ValueError as err:
-                msg = ("No atoms in selection at current frame,"
-                       " running user grid")
+                msg = ("No atoms in AtomGroup at input time frame. "
+                       "Grid for density could not be automatically"
+                       " generated. A user defined grid was found"
+                       " and is being used as the density grid")
                 warnings.warn(msg)
                 logger.warning(msg)
                 smin = self._gridcenter
@@ -418,14 +420,15 @@ class DensityAnalysis(AnalysisBase):
             # periodic boundaries' but that gets complicated when the box
             # rotates due to RMS fitting.
             try:
-                smin = np.min(coord, axis=0)
-                smax = np.max(coord, axis=0)
+                smin = np.min(coord, axis=0) - self._padding
+                smax = np.max(coord, axis=0) + self._padding
             except ValueError as err:
-                errmsg = ("No atoms in selection at current frame"
-                          " and no user defined grid")
+                errmsg = ("No atoms in AtomGroup at input time frame. "
+                          "Grid for density could not be automatically"
+                          " generated. If this is expected, a user"
+                          " defined grid will need to be "
+                          "provided instead.")
                 raise ValueError(errmsg) from err
-            smin = smin - self._padding
-            smax = smax + self._padding
         BINS = fixedwidth_bins(self._delta, smin, smax)
         arange = np.transpose(np.vstack((BINS['min'], BINS['max'])))
         bins = BINS['Nbins']
--- a/package/MDAnalysis/analysis/density.py
+++ b/package/MDAnalysis/analysis/density.py
@@ -849,5 +849,4 @@ class Density(Grid):
             grid_type = 'density'
         else:
             grid_type = 'histogram'
-        return '<Density ' + grid_type + ' with ' + \
-            str(self.grid.shape) + ' bins>'
+            return f"<Density {grid_type} with {self.grid.shape} bins>"
--- a/package/MDAnalysis/analysis/density.py
+++ b/package/MDAnalysis/analysis/density.py
@@ -228,10 +228,10 @@ class DensityAnalysis(AnalysisBase):
     around 4.0 (protein and not name H*)", i.e., the water oxygen atoms that
     are within 4 Å of the protein heavy atoms) then create an
     :class:`~MDAnalysis.core.groups.UpdatingAtomGroup` (see Examples).
-    
+
     If the 'AtomGroup' instance does not contain any selection of atoms and
     user defined box limits are provided, it is necessary to ensure that the
-    provided limits encompass all atoms in selection on future frames.   
+    provided limits encompass all atoms in selection on future frames.
 
     Examples
     --------
--- a/package/MDAnalysis/analysis/density.py
+++ b/package/MDAnalysis/analysis/density.py
@@ -229,12 +229,12 @@ class DensityAnalysis(AnalysisBase):
     are within 4 Å of the protein heavy atoms) then create an
     :class:`~MDAnalysis.core.groups.UpdatingAtomGroup` (see Examples).
 
-    If the 'AtomGroup' instance does not contain any selection of atoms and 
+    If the 'AtomGroup' instance does not contain any selection of atoms and
     'updating' is set to true, an empty :class: UpdatingAtomGroup is returned.
-    If an instance of this class is passed onto DensityAnalysis without any 
-    user-defined grid, DensityAnalysis will fail because it cannot predict 
-    selection bounds for future frames in the trajectory. In such a situation, 
-    user defined box limits should be provided ensuring that the provided 
+    If an instance of this class is passed onto DensityAnalysis without any
+    user-defined grid, DensityAnalysis will fail because it cannot predict
+    selection bounds for future frames in the trajectory. In such a situation,
+    user defined box limits should be provided ensuring that the provided
     limits encompass all atoms in selection on future frames.
 
     Examples
@@ -720,8 +720,8 @@ class Density(Grid):
         """
         # all this unit crap should be a class...
         try:
-            for unit_type, value in u.items():     
-                if value is None: # check here, too iffy to use dictionary[None]=None
+            for unit_type, value in u.items():
+                if value is None:  # check here, too iffy to use dictionary[None]=None
                     self.units[unit_type] = None
                     continue
                 try:
--- a/package/MDAnalysis/analysis/density.py
+++ b/package/MDAnalysis/analysis/density.py
@@ -107,7 +107,7 @@ atomgroup.
       After the analysis (see the :meth:`~DensityAnalysis.run` method), the resulting density is
       stored in the :attr:`density` attribute as a :class:`Density` instance.
 
-   .. automethod:: _set_user_grid
+   .. automethod:: _prepare, _set_user_grid
 
 
 Density object
@@ -400,6 +400,9 @@ class DensityAnalysis(AnalysisBase):
         UserWarning
             if AtomGroup is empty and a user defined grid is provided
 
+        .. versionadded:: 1.0.0    
+        .. versionchanged:: 2.0.0
+           Now a method of :class:`DensityAnalysis`.
         """
         coord = self._atomgroup.positions
         if self._gridcenter is not None:
--- a/package/MDAnalysis/analysis/density.py
+++ b/package/MDAnalysis/analysis/density.py
@@ -384,18 +384,7 @@ class DensityAnalysis(AnalysisBase):
         self._ydim = ydim
         self._zdim = zdim
 
-    def _checkinput(self):
-        if len(self._atomgroup) == 0:
-            msg = ("No atoms in selection over whole trajectory")
-            warnings.warn(msg)
-            logger.warning(msg)
-            return 1
-        else:
-            return 0
-
     def _prepare(self):
-        if _checkinput(self) == 1:
-            return
         coord = self._atomgroup.positions
         if self._gridcenter is not None:
             # Issue 2372: padding is ignored, defaults to 2.0 therefore warn
@@ -406,8 +395,15 @@ class DensityAnalysis(AnalysisBase):
                 logger.warning(msg)
             # Generate a copy of smin/smax from coords to later check if the
             # defined box might be too small for the selection
-            smin = np.min(coord, axis=0)
-            smax = np.max(coord, axis=0)
+            try: 
+                smin = np.min(coord, axis=0)
+                smax = np.max(coord, axis=0)
+            except ValueError as err:
+                msg = ("No atoms in selection at current frame, running user grid")
+                warnings.warn(msg)
+                logger.warning(msg)
+                smin = self._gridcenter 
+                smax = self._gridcenter 
             # Overwrite smin/smax with user defined values
             smin, smax = self._set_user_grid(self._gridcenter, self._xdim,
                                              self._ydim, self._zdim, smin,
@@ -420,8 +416,20 @@ class DensityAnalysis(AnalysisBase):
             # ideal solution would use images: implement 'looking across the
             # periodic boundaries' but that gets complicated when the box
             # rotates due to RMS fitting.
-            smin = np.min(coord, axis=0) - self._padding
-            smax = np.max(coord, axis=0) + self._padding
+            try:
+                smin = np.min(coord, axis=0) 
+                smax = np.max(coord, axis=0) 
+            except ValueError as err:
+                msg = ("No atoms in selection at current frame, running user grid")
+                warnings.warn(msg)
+                logger.warning(msg)
+                smin = self._gridcenter 
+                smax = self._gridcenter  
+                smin, smax = self._set_user_grid(self._gridcenter, self._xdim,
+                                                 self._ydim, self._zdim, smin,
+                                                 smax)
+            smin = smin - self._padding
+            smax = smax + self._padding
         BINS = fixedwidth_bins(self._delta, smin, smax)
         arange = np.transpose(np.vstack((BINS['min'], BINS['max'])))
         bins = BINS['Nbins']
@@ -436,8 +444,6 @@ class DensityAnalysis(AnalysisBase):
         self.density = None
 
     def _single_frame(self):
-        if _checkinput(self) == 1:
-            return
         h, _ = np.histogramdd(self._atomgroup.positions,
                               bins=self._bins, range=self._arange,
                               normed=False)
@@ -448,8 +454,6 @@ class DensityAnalysis(AnalysisBase):
         self._grid += h
 
     def _conclude(self):
-        if _checkinput(self) == 1:
-            return
         # average:
         self._grid /= float(self.n_frames)
         density = Density(grid=self._grid, edges=self._edges,
--- a/package/MDAnalysis/analysis/density.py
+++ b/package/MDAnalysis/analysis/density.py
@@ -513,8 +513,8 @@ class DensityAnalysis(AnalysisBase):
             raise ValueError("xdim, ydim, and zdim must be numbers") from err
 
         # Set min/max by shifting by half the edge length of each dimension
-        umin = gridcenter - xyzdim / 2
-        umax = gridcenter + xyzdim / 2
+        umin = gridcenter - xyzdim/2
+        umax = gridcenter + xyzdim/2
 
         # Here we test if coords of selection fall outside of the defined grid
         # if this happens, we warn users they may want to resize their grids
@@ -720,9 +720,8 @@ class Density(Grid):
         """
         # all this unit crap should be a class...
         try:
-            for unit_type, value in u.items():
-                # check here, too iffy to use dictionary[None]=None
-                if value is None:
+            for unit_type, value in u.items():     
+                if value is None: # check here, too iffy to use dictionary[None]=None
                     self.units[unit_type] = None
                     continue
                 try:
@@ -790,8 +789,7 @@ class Density(Grid):
         """
         if unit == self.units['length']:
             return
-        cvnfact = units.get_conversion_factor(
-            'length', self.units['length'], unit)
+        cvnfact = units.get_conversion_factor('length', self.units['length'], unit)
         self.edges = [x * cvnfact for x in self.edges]
         self.units['length'] = unit
         self._update()  # needed to recalculate midpoints and origin
@@ -842,8 +840,8 @@ class Density(Grid):
         if unit == self.units['density']:
             return
         try:
-            self.grid *= units.get_conversion_factor(
-                'density', self.units['density'], unit)
+            self.grid *= units.get_conversion_factor('density',
+                                                     self.units['density'], unit)
         except KeyError:
             errmsg = (f"The name of the unit ({unit} supplied) must be one "
                       f"of:\n{units.conversion_factor['density'].keys()}")
@@ -855,4 +853,4 @@ class Density(Grid):
             grid_type = 'density'
         else:
             grid_type = 'histogram'
-        return f"<Density {grid_type} with {self.grid.shape} bins>"
+        return '<Density ' + grid_type + ' with ' + str(self.grid.shape) + 'bins>'
--- a/package/MDAnalysis/analysis/density.py
+++ b/package/MDAnalysis/analysis/density.py
@@ -228,7 +228,10 @@ class DensityAnalysis(AnalysisBase):
     around 4.0 (protein and not name H*)", i.e., the water oxygen atoms that
     are within 4 Å of the protein heavy atoms) then create an
     :class:`~MDAnalysis.core.groups.UpdatingAtomGroup` (see Examples).
-
+    
+    If the 'AtomGroup' instance does not contain any selection of atoms and
+    user defined box limits are provided, it is necessary to ensure that the
+    provided limits encompass all atoms in selection on future frames.   
 
     Examples
     --------
@@ -400,9 +403,9 @@ class DensityAnalysis(AnalysisBase):
                 smax = np.max(coord, axis=0)
             except ValueError as err:
                 msg = ("No atoms in AtomGroup at input time frame. "
-                       "Grid for density could not be automatically"
-                       " generated. A user defined grid was found"
-                       " and is being used as the density grid")
+                       "This may be intended; please ensure that "
+                       "your grid selection covers the atomic "
+                       "positions you wish to capture.")
                 warnings.warn(msg)
                 logger.warning(msg)
                 smin = self._gridcenter
@@ -676,7 +679,7 @@ class Density(Grid):
 
         parameters = kwargs.pop('parameters', {})
         if (len(args) > 0 and isinstance(args[0], str) or
-                isinstance(kwargs.get('grid', None), str)):
+            isinstance(kwargs.get('grid', None), str)):
             # try to be smart: when reading from a file then it is likely that
             # this is a density
             parameters.setdefault('isDensity', True)
--- a/package/MDAnalysis/analysis/density.py
+++ b/package/MDAnalysis/analysis/density.py
@@ -369,6 +369,7 @@ class DensityAnalysis(AnalysisBase):
     .. versionchanged:: 2.0.0
        :func:`_set_user_grid` is now a method of :class:`DensityAnalysis`.
     """
+
     def __init__(self, atomgroup, delta=1.0,
                  metadata=None, padding=2.0,
                  gridcenter=None,
@@ -383,11 +384,17 @@ class DensityAnalysis(AnalysisBase):
         self._ydim = ydim
         self._zdim = zdim
 
-    def _prepare(self):
-        if len(self._atomgroup)==0:
+    def _checkinput(self):
+        if len(self._atomgroup) == 0:
             msg = ("No atoms in selection over whole trajectory")
             warnings.warn(msg)
             logger.warning(msg)
+            return 1
+        else:
+            return 0
+
+    def _prepare(self):
+        if _checkinput(self) == 1:
             return
         coord = self._atomgroup.positions
         if self._gridcenter is not None:
@@ -429,11 +436,8 @@ class DensityAnalysis(AnalysisBase):
         self.density = None
 
     def _single_frame(self):
-        if len(self._atomgroup)==0:
-            msg = ("No atoms in selection over whole trajectory")
-            warnings.warn(msg)
-            logger.warning(msg)
-            return    
+        if _checkinput(self) == 1:
+            return
         h, _ = np.histogramdd(self._atomgroup.positions,
                               bins=self._bins, range=self._arange,
                               normed=False)
@@ -444,10 +448,7 @@ class DensityAnalysis(AnalysisBase):
         self._grid += h
 
     def _conclude(self):
-        if len(self._atomgroup)==0:
-            msg = ("No atoms in selection over whole trajectory")
-            warnings.warn(msg)
-            logger.warning(msg)
+        if _checkinput(self) == 1:
             return
         # average:
         self._grid /= float(self.n_frames)
@@ -502,8 +503,8 @@ class DensityAnalysis(AnalysisBase):
             raise ValueError("xdim, ydim, and zdim must be numbers") from err
 
         # Set min/max by shifting by half the edge length of each dimension
-        umin = gridcenter - xyzdim/2
-        umax = gridcenter + xyzdim/2
+        umin = gridcenter - xyzdim / 2
+        umax = gridcenter + xyzdim / 2
 
         # Here we test if coords of selection fall outside of the defined grid
         # if this happens, we warn users they may want to resize their grids
@@ -672,7 +673,7 @@ class Density(Grid):
 
         parameters = kwargs.pop('parameters', {})
         if (len(args) > 0 and isinstance(args[0], str) or
-            isinstance(kwargs.get('grid', None), str)):
+                isinstance(kwargs.get('grid', None), str)):
             # try to be smart: when reading from a file then it is likely that
             # this is a density
             parameters.setdefault('isDensity', True)
@@ -710,7 +711,8 @@ class Density(Grid):
         # all this unit crap should be a class...
         try:
             for unit_type, value in u.items():
-                if value is None:  # check here, too iffy to use dictionary[None]=None
+                # check here, too iffy to use dictionary[None]=None
+                if value is None:
                     self.units[unit_type] = None
                     continue
                 try:
@@ -778,7 +780,8 @@ class Density(Grid):
         """
         if unit == self.units['length']:
             return
-        cvnfact = units.get_conversion_factor('length', self.units['length'], unit)
+        cvnfact = units.get_conversion_factor(
+            'length', self.units['length'], unit)
         self.edges = [x * cvnfact for x in self.edges]
         self.units['length'] = unit
         self._update()  # needed to recalculate midpoints and origin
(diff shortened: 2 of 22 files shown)


hmacdope:
Hey @aditya-kamath, would you be able to provide a test that triggers the warnings you have inserted?

Also using autopep8 to fix style issues will save you a lot of pain.

aditya-kamath:
Yeah, I can write a test for it. I just hope that this fix does not interfere with other functionalities of DensityAnalysis as I don't fully understand it yet. 

Thanks for the heads up on autopep8! :)

package/MDAnalysis/analysis/density.py
```diff
@@ -434,6 +444,11 @@ def _single_frame(self):
         self._grid += h
 
     def _conclude(self):
+        if len(self._atomgroup)==0:
+            msg = ("No atoms in selection over whole trajectory")
+            warnings.warn(msg)
+            logger.warning(msg)
+            return
```
tylerjereddy, line 451: I wonder if three duplications of the same code block is enough to justify abstraction to a single i.e., function. Maybe borderline if the `return` has to remain outside of the abstraction I suppose.

On a more basic level, just asking--the code will still produce the original failure/exception (is that what it was doing?) and this will simply add a warning to clarify?

aditya-kamath: This might sound very primitive but my logic was to return out of the function and warn the user of no atoms until the atomgroup gets updated with atleast one atom. In the Issue, the user wanted to avoid the program crashing and would provide a dummy atom to keep the density frames generating. I tried to fix this by preventing the crash at the initializing, frame generation and conclusion. Perhaps a more efficient way is to make a function to check for no atoms and return.

aditya-kamath: Again, my understanding of the function process may not be very accurate so I apologize for any incorrect arguments I make :P

package/MDAnalysis/analysis/density.py
```diff
@@ -383,7 +384,18 @@ def __init__(self, atomgroup, delta=1.0,
...
+            msg = ("No atoms in selection over whole trajectory")
+            warnings.warn(msg)
+            logger.warning(msg)
+            return 1
+        else:
+            return 0
+
     def _prepare(self):
+        if _checkinput(self) == 1:
+            return
```
IAlibay, line 398: Couple of points to consider here;

1. Is this the right place to be putting this? Given the context of the original issue, and the need to make it so that the user defined grid still works even if there are no atoms in the frame, would placing this elsewhere be more appropriate?

2. Doing a `return` post-warning is a non-safe way to handle things. It's worth considering what the difference between a "warning" and an "error" is. If you're issuing a warning, you are telling users that non-expected behaviour is likely to occur (and why) but that the code will carry on. If you raise an error, you are telling users that something wrong/unexpected has occurred and you are aborting further code execution. In this case, if you are attempting to abort execution (as you are via `return`), then an error (of appropriate type) should be raised instead.

3. I think the way this is written probably needs to be changed, but it's worth mentioning anyways. Doing a return equality check on method return that essentially returns a `Boolean` is a non-pythonic way to handle things. If you're checking for `True`/`False`, the best way to do things is to have the method you're calling return `True`/`False` and then just do:
```python
if method(input):
    true_condition
else:
    false_condition
```

package/MDAnalysis/analysis/density.py
```diff
@@ -383,7 +384,18 @@ def __init__(self, atomgroup, delta=1.0,
         self._ydim = ydim
         self._zdim = zdim
 
+    def _checkinput(self):
+        if len(self._atomgroup) == 0:
+            msg = ("No atoms in selection over whole trajectory")
```
IAlibay, line 389: Is this a correct statement? If you're passed an `UpdatingAtomGroup`, does checking only the first frame tell you that there will be no atoms covering this particular selection for the entirety of the trajectory?

package/MDAnalysis/analysis/density.py
```diff
@@ -383,7 +384,18 @@ def __init__(self, atomgroup, delta=1.0,
         self._ydim = ydim
         self._zdim = zdim
 
+    def _checkinput(self):
+        if len(self._atomgroup) == 0:
```
IAlibay, line 388: It's worth revisiting the original issue and checking what we need here. It's correct that we want to avoid building a grid on an empty selection when we aren't using a user defined grid, but when we are, we want to be able to let users pass an empty selection.

It's worth considering where in the code the edge case is failing (I'll let you work it out, but I believe it might be `smin` and `smax`), and how you could incorporate such a check within this context.

As @hmacdope mentioned, it's worth looking to apply a test driven development approach here. Write some tests that would check the new behaviour of the code post fix, and then use those as ways to verify if your code changes are indeed addressing the issue.

IAlibay (changes requested):
In addition to the other comments, please see below for some comments / areas which could be improved.

package/MDAnalysis/analysis/density.py
aditya-kamath: Hi, Thanks for all the suggestions! I will try to clean all this up. I think I need to approach this issue another way. I'll try to catch this issue from the smin,smax step.

package/MDAnalysis/analysis/density.py
aditya-kamath: Yeah, perhaps this should be - "No atoms in selection in frame #number"

package/MDAnalysis/analysis/density.py
```diff
@@ -369,6 +369,7 @@ class DensityAnalysis(AnalysisBase):
     .. versionchanged:: 2.0.0
        :func:`_set_user_grid` is now a method of :class:`DensityAnalysis`.
     """
+
```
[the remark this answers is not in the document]
aditya-kamath: I credit this change to - autopep8

package/MDAnalysis/analysis/density.py
```diff
@@ -657,7 +673,7 @@ def __init__(self, *args, **kwargs):
 
         parameters = kwargs.pop('parameters', {})
         if (len(args) > 0 and isinstance(args[0], str) or
-            isinstance(kwargs.get('grid', None), str)):
+                isinstance(kwargs.get('grid', None), str)):
```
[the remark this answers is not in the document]
aditya-kamath: I ran this file through autopep8.

package/MDAnalysis/analysis/density.py
```diff
@@ -827,4 +845,5 @@ def __repr__(self):
             grid_type = 'density'
         else:
             grid_type = 'histogram'
-        return '<Density ' + grid_type + ' with ' + str(self.grid.shape) + ' bins>'
+        return '<Density ' + grid_type + ' with ' + \
+            str(self.grid.shape) + ' bins>'
```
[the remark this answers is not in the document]
aditya-kamath: Thanks for this! Noted

package/MDAnalysis/analysis/density.py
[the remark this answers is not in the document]
IAlibay: The choice to use an automatic formatter is kind of a personal choice (I don't because of issues like these, I prefer having fine grained control about how I will format things). If you do use one though, particularly in highly collaborative projects like MDAnalysis, it's on you to then check if the changes made were appropriate.

That being said, whilst my PEP8 linters don't identify issues with this, it's possible that there is some. Given that we have a lot of heterogeneity in whether or not we have a blank line between the docstring and `__init__`, is there something in the PEP8 documentation that indicates whether or not this is necessary?

aditya-kamath:
Thanks to all the reviewers for their suggestions! I will now try to work on cleaning up the code in general and I see that I have a  messy commit history. I aim to clean it up too.

IAlibay:
> Thanks to all the reviewers for their suggestions! I will now try to work on cleaning up the code in general and I see that I have a messy commit history. I aim to clean it up too.

Messy commit histories don't usually matter for small fixes like these, eventually we will squash merge anyways, so it'll only show up a single commit on MDAnalysis.

Before I forget, probably something to take into account is the branch you make your changes on. I noticed that you are using the `develop` branch of your own fork. This is eventually going to be problematic and your local commit history is going to diverge from the main `develop` on MDAnalysis. This will end up causing you merge problems in the future.

What I would suggest in the future is to instead create an appropriately named feature branch on your own fork, which you then make your changes on. See the following for a bit more information: https://userguide.mdanalysis.org/stable/contributing_code.html#creating-a-branch

aditya-kamath:
@IAlibay , the merging was exactly what I was facing issues with. Thanks for the info!

package/MDAnalysis/analysis/density.py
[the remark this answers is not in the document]
aditya-kamath: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings 
I found this which says it is recommended to add a line after all docstrings.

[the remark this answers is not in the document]
aditya-kamath: That looks like it's for PEP 257 though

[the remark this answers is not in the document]
IAlibay: That's fine. The importance here was to have a rationale to explain why a change has been made rather than relying on the accuracy of automated tools.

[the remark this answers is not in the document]
aditya-kamath: Yes, probably not a good idea to make blind changes..

aditya-kamath:
@IAlibay @hmacdope , I've made the changes requested and added a test. Let me know if there's other changes..

package/MDAnalysis/analysis/density.py
```diff
@@ -408,8 +416,20 @@ def _prepare(self):
...
-            smin = np.min(coord, axis=0) - self._padding
-            smax = np.max(coord, axis=0) + self._padding
+            try:
+                smin = np.min(coord, axis=0) 
+                smax = np.max(coord, axis=0) 
+            except ValueError as err:
+                msg = ("No atoms in selection at current frame, running user grid")
+                warnings.warn(msg)
+                logger.warning(msg)
+                smin = self._gridcenter
```
IAlibay, line 426: Can you detail the logic here? This shouldn't work since you can't get here if `self._gridcenter` is not `None`?

package/MDAnalysis/analysis/density.py
```diff
@@ -394,8 +395,15 @@ def _prepare(self):
                 logger.warning(msg)
             # Generate a copy of smin/smax from coords to later check if the
             # defined box might be too small for the selection
-            smin = np.min(coord, axis=0)
-            smax = np.max(coord, axis=0)
+            try:
```
hmacdope, line 398: I could be wrong but I don't see how `np.min` can ever be `False` and therefore trigger the below `ValueError` unless `self._atomgroup.positions` is empty. Although this is the desired behaviour, IMO this is not the best way to test it. 

@IAlibay is there a canonical way of checking for an empty `atomgroup`?

package/MDAnalysis/analysis/density.py
```diff
@@ -827,4 +849,4 @@ def __repr__(self):
             grid_type = 'density'
         else:
             grid_type = 'histogram'
-        return '<Density ' + grid_type + ' with ' + str(self.grid.shape) + ' bins>'
+            return f"<Density {grid_type} with {self.grid.shape} bins>"
```
hmacdope, line 852: We don't use fstrings for compatibility reasons.

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -120,7 +120,8 @@ class DensityParameters(object):
     topology = TPR
     trajectory = XTC
     delta = 2.0
-    selections = {'static': "name OW",
+    selections = {'none': "byres resname LIG and around 8 protein", #this selection yields nothing
```
hmacdope, line 123: Not the best way to choose an empty selection. Although it is tagged as `none` the `byres resname LIG and around 8 protein` sounds plausible. I would do something that is more obvious like `None` or `"resname None"`

hmacdope (changes requested):

package/MDAnalysis/analysis/density.py
lilyminium: Which way would you prefer -- an `if len(coord)`? I actually prefer ``try/except`` here because the case where there are no atoms seems much rarer than the opposite, but I haven't put much thought into this.

richardjgowers: AGs can coerce to bool, so empty is false. Lily is right wrt EAFP

hmacdope: Okay sure, I agree and on second look the error message details what it is meant to be checking so all good.

package/MDAnalysis/analysis/density.py
aditya-kamath: Ooh, perhaps it should fail here; Since no user defined grid is present in this case!

aditya-kamath:
@IAlibay @hmacdope, I've finally cleaned up the trailing issues. This should be better.

package/MDAnalysis/analysis/density.py
```diff
@@ -408,8 +417,15 @@ def _prepare(self):
...
-            smin = np.min(coord, axis=0) - self._padding
-            smax = np.max(coord, axis=0) + self._padding
+            try:
+                smin = np.min(coord, axis=0)
+                smax = np.max(coord, axis=0)
+            except ValueError as err:
+                errmsg = ("No atoms in selection at current frame"
+                          " and no user defined grid")
+                raise ValueError(errmsg) from err
+            smin = smin - self._padding
```
IAlibay, line 427: why is this not being done on line 421 and 422 respectively as per the old code? Either smin/smax are going to be defined or you're going to get the ValueError?

package/MDAnalysis/analysis/density.py
```diff
@@ -408,8 +417,15 @@ def _prepare(self):
             # ideal solution would use images: implement 'looking across the
             # periodic boundaries' but that gets complicated when the box
             # rotates due to RMS fitting.
-            smin = np.min(coord, axis=0) - self._padding
-            smax = np.max(coord, axis=0) + self._padding
+            try:
+                smin = np.min(coord, axis=0)
+                smax = np.max(coord, axis=0)
+            except ValueError as err:
+                errmsg = ("No atoms in selection at current frame"
```
IAlibay, line 424: This needs to be a bit more explicit. Think about what users would understand if faced with this error.

Couple of pointers:
1) "selection" is ambiguous, would AtomGroup be more appropriate
2) "no user defined grid" -> ok, but what should users be doing?

package/MDAnalysis/analysis/density.py
```diff
@@ -394,8 +395,16 @@ def _prepare(self):
                 logger.warning(msg)
             # Generate a copy of smin/smax from coords to later check if the
             # defined box might be too small for the selection
-            smin = np.min(coord, axis=0)
-            smax = np.max(coord, axis=0)
+            try:
+                smin = np.min(coord, axis=0)
+                smax = np.max(coord, axis=0)
+            except ValueError as err:
+                msg = ("No atoms in selection at current frame,"
```
IAlibay, line 402: As per the other `msg` comment I left, this probably needs to be more informative to users. What does this mean, what are you defaulting to?

This may not be intended behaviour so you need to make sure that users know this is happening and why.

package/MDAnalysis/analysis/density.py
```diff
@@ -394,8 +395,16 @@ def _prepare(self):
                 logger.warning(msg)
             # Generate a copy of smin/smax from coords to later check if the
             # defined box might be too small for the selection
-            smin = np.min(coord, axis=0)
-            smax = np.max(coord, axis=0)
+            try:
```
IAlibay, line 398: Just a thought here.

I can't think of a way to clean this up to have fewer lines of repeated code, however it might still be cleaner to move this part of the check to `_set_user_grid`?

Thoughts (tagging @hmacdope too). Ignore if you both don't think this is necessary.

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -254,6 +255,17 @@ def test_ValueError_userdefn_xdim_type(self, universe):
                 delta=self.delta, xdim="MDAnalysis", ydim=10.0, zdim=10.0,
                 gridcenter=self.gridcenters['static_defined']).run(step=5)
 
+    def test_warn_noatomgroup(self, universe):
+        regex = ("No atoms in selection at current frame, running user grid")
+        with pytest.warns(UserWarning, match=regex):
```
IAlibay, line 260: it would be a lot cleaner here if you could split this into two tests please.

IAlibay (changes requested):
Couple more comments.

We should also think about documentation; what could be added here to make users aware of this behaviour and what they should be doing when they pass an empty atomgroup?

package/MDAnalysis/analysis/density.py
aditya-kamath: "please enter grid values" - I'll add this

package/MDAnalysis/analysis/density.py
aditya-kamath: My bad, that's a unnecessary step. I'll remove it

package/MDAnalysis/analysis/density.py
aditya-kamath: "defaulting to user defined grid" - is this better?

package/MDAnalysis/analysis/density.py
aditya-kamath: Hmm, this might be tricky.

package/MDAnalysis/analysis/density.py
IAlibay: It depends on the context of the full sentence, but probably not? You'll always end up using the user defined grid, so just saying you are "defaulting" to it doesn't make sense.

See my example above, re: an explicit user message.

package/MDAnalysis/analysis/density.py
aditya-kamath: Okay, the way I think this can be done is to have a len(atomgroup) check done in _set_user_grid to see if there's an atom group first and then calculate smin and smax. We will also need to move the padding function into _set_user_grid as well. But tbh, I feel like this will only help with the abstraction of the script rather than how fast the code runs.

package/MDAnalysis/analysis/density.py
```diff
@@ -763,7 +783,8 @@ def convert_length(self, unit='Angstrom'):
         """
         if unit == self.units['length']:
             return
-        cvnfact = units.get_conversion_factor('length', self.units['length'], unit)
```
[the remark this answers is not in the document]
IAlibay: I'm not really following what happened here. Why do you think these changes are causing issues?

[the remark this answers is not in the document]
aditya-kamath: There are a few changes which were made by autoPEP8 and not a part of fixing the error. The only failing check of CodeCov is because these changes are not covered by tests..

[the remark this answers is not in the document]
IAlibay: It's up to you, codecov is complaining because it's gone from one line of untested code to two. The best solution here would be to go ahead and add a test for this, but otherwise I think we can live with just a style change.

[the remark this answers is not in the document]
aditya-kamath: Sounds good! Thanks @IAlibay :)

package/MDAnalysis/analysis/density.py
hmacdope: IMO I think it can stay outside `set_user_grid` as it makes clear what is happening in main control flow for `_prepare`.

package/MDAnalysis/analysis/density.py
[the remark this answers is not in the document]
hmacdope: Looks bizzare to me. I would revert this.

package/MDAnalysis/analysis/density.py
```diff
@@ -394,8 +395,18 @@ def _prepare(self):
...
+                msg = ("No atoms in AtomGroup at input time frame. "
+                       "Grid for density could not be automatically"
+                       " generated. A user defined grid was found"
+                       " and is being used as the density grid")
+                warnings.warn(msg)
+                logger.warning(msg)
+                smin = self._gridcenter
+                smax = self._gridcenter
             # Overwrite smin/smax with user defined values
             smin, smax = self._set_user_grid(self._gridcenter, self._xdim,
```
hmacdope, line 411: @IAlibay what are your thoughts on emitting a warning if `self._gridcenter` is not set but `self._xdim, sel._ydim` etc is not? I know its mentioned in docs, but IMO it looks easy to have your grid being ignored as incomplete even if not the intention.

hmacdope (changes requested):
Hey @aditya-kamath, this is looking good. Mostly just a few small nitpicks from me and a question for @IAlibay. :)

package/MDAnalysis/analysis/density.py
IAlibay: I'm not sure I understand what you mean here. If X/y/zdim are not set you'll get a ValueError in _set_user_grid?

package/MDAnalysis/analysis/density.py
```diff
@@ -394,8 +395,18 @@ def _prepare(self):
...
-            smin = np.min(coord, axis=0)
-            smax = np.max(coord, axis=0)
+            try:
+                smin = np.min(coord, axis=0)
+                smax = np.max(coord, axis=0)
+            except ValueError as err:
+                msg = ("No atoms in AtomGroup at input time frame. "
+                       "Grid for density could not be automatically"
+                       " generated. A user defined grid was found"
+                       " and is being used as the density grid")
```
IAlibay, line 405: My example message for the other warning can't be directly copied here, the context is different. It's incorrect to say that you were attempting to "automatically generate" the grid since it was manually set by the user. Please amend this to better fit the context of the error.

package/MDAnalysis/analysis/density.py
```diff
@@ -408,8 +419,16 @@ def _prepare(self):
             # ideal solution would use images: implement 'looking across the
             # periodic boundaries' but that gets complicated when the box
             # rotates due to RMS fitting.
-            smin = np.min(coord, axis=0) - self._padding
-            smax = np.max(coord, axis=0) + self._padding
+            try:
+                smin = np.min(coord, axis=0) - self._padding
+                smax = np.max(coord, axis=0) + self._padding
+            except ValueError as err:
+                errmsg = ("No atoms in AtomGroup at input time frame. "
```
IAlibay, line 426: I've mentioned this elsewhere but I think that comment got overlooked; this needs documenting in the docstring.

IAlibay (changes requested):

package/MDAnalysis/analysis/density.py
aditya-kamath: `msg = ("No atoms in AtomGroup at input time frame. "`
                 `           "Density will contain a grid constructed with`
                 `           "user defined values ")`

package/MDAnalysis/analysis/density.py
aditya-kamath: I'll add it to the examples docstring of the `DensityAnalysis` class. :)

package/MDAnalysis/analysis/density.py
hmacdope: I meant the other way round. I could be missing something, but AFAIK if `gridcenter` is not set but `xdim`, `ydim` and `zdim` are set we never reach `_set_user_grid()` and `xdim`, `ydim` and `zdim` are ignored silently? 

Envisioning a warning like `"user defined grid coordinates are set but no grid centre is present, 
 user defined grid coordinates are being ignored."`?

IAlibay: I see, maybe we can change the `if` check to `any` that covers gridcenter and the dims? Doing so should trigger the type checking in _set_user_grid.

Ideally I'd prefer not adding extra error/warning emitting unless we move all this to _set_user_grid and revamp the error checking there. If we go down that route, I'd prefer we deal with this in a separate PR though.

If we choose to go down the `any` route, which should be the simplest to deal with, then I'm happy with us doing this here, but we'll need an extra test probably.

lilyminium: Idea: if none of them are set, default to box size if box exists?

IAlibay: That's dangerous imho, not only because it adds an extra default condition, but also because box size isn't an easy / cheap thing to define under NPT conditions. It'd essentially require an extra loop across the trajectory to get the max box dimensions etc... We can certainly have that option but imho it'd be better to have it as a non-default.

Re xyzdim I'd vote that our current approach is the simplest, fail unless you provide numbers. 

I agree that not setting grid center when xyzdim  are set shouldnt be a silent default either.

package/MDAnalysis/analysis/density.py
IAlibay: I think it's worth considering why we are emitting the warning. Is it reasonable to be warning folks that we're following expected operation (i.e. they asked to use a user defined grid, is it informative to have us tell them we're doing this)?

Rather than thinking about what the code is doing next, it's best to phrase the warning in terms of the consequences of not having any atoms in the atomgroup. What does this mean in terms of the user grid? How does it impact us checking if the user grid is large enough to accommodate the selected atoms?

aditya-kamath: Hmm, that would mean that the user should have made sure that they enter a grid large enough to accomodate atoms in the next frames.

aditya-kamath: `msg = ("No atoms in AtomGroup at input time frame. "`
             `           "Please ensure grid values are in a large range to`
             `           "accommodate all selected atoms ")`

IAlibay: I'm not sure "large range" is the right set of words here. Maybe something like "This be may be intended; please ensure that your grid selection covers the atomic positions you wish to capture."?

package/MDAnalysis/analysis/density.py
```diff
@@ -228,7 +228,10 @@ class DensityAnalysis(AnalysisBase):
     around 4.0 (protein and not name H*)", i.e., the water oxygen atoms that
     are within 4 Å of the protein heavy atoms) then create an
     :class:`~MDAnalysis.core.groups.UpdatingAtomGroup` (see Examples).
-
+    
+    If the 'AtomGroup' instance does not contain any selection of atoms and
```
IAlibay, line 232: Considering the original issue, does this reflect what should be documented here?

It's not wrong to tell users to make sure that grids are static and therefore should fit a selection over a trajectory (from a quick glance over the docstring, this isn't included, so it's not wrong to include it), but that's a separate point from the one being made about empty atomgroups.

Based on the changes done in this pull request, what would you tell a new user looking to do a density calculation with an starting empty atomgroup to do? I.e. what parts of the code aren't going to work, and which will?

aditya-kamath: If the 'AtomGroup' instance does not contain any selection of atoms and
user defined box limits are provided, density is prepared in expectation 
for atoms in future frames in the trajectory.

aditya-kamath: The user need to understand that without any atoms, they just get a grid with atoms to be added later on. Hopefully that's what the above message says

IAlibay: Yes, but I think we need to be more informative.

If we break it down, the message should really contain the following:
1) Sentence that defines an empty :class:`UpdatingAtomGroup` in this case (i.e. no atoms in selection at the current frame during class construction).
2) What happens if you pass a :class:`UpdatingAtomGroup` which is "empty"  under default conditions? That is to say, not a user-defined grid. Why does this happen?
3) How do you circumvent this issue? i.e. using a user-defined grid including any additional considerations about  doing this.

aditya-kamath: I think going back to the original issue clarifies this: 
1+2)  For an atomselection that is initially empty but which becomes filled over the course of the trajectory. DensityAnalysis will fail if the selection is initially empty despite using updating=True:

3) A quick solution to the above problem can be to provide a user-defined grid. However, it should capture all the atoms in future frames.

aditya-kamath:
Hi @IAlibay, sorry, I have'nt gotten to this PR in a while! I've been busy with school work.. 
I'm thinking of wrapping this up and completing any remaining fixes to be done. I've updated the Docstring, please have a look and  let me know your thoughts! I'd be happy to take feedback..

package/MDAnalysis/analysis/density.py
```diff
@@ -229,6 +229,13 @@ class DensityAnalysis(AnalysisBase):
     are within 4 Å of the protein heavy atoms) then create an
     :class:`~MDAnalysis.core.groups.UpdatingAtomGroup` (see Examples).
 
+    If the 'AtomGroup' instance does not contain any selection of atoms and 
+    'updating' is set to true, an empty :class: UpdatingAtomGroup is returned.
+    If an instance of this class is passed onto DensityAnalysis without any 
+    user-defined grid, DensityAnalysis will fail because it cannot predict 
+    selection bounds for future frames in the trajectory. In such a situation, 
+    user defined box limits should be provided ensuring that the provided 
+    limits encompass all atoms in selection on future frames.
```
orbeckst, line 238: This text does not reflect what the code does, or at least that's not clear. The only change that you made is to emit warnings, right? 

State what the problem can be (empty `atomgroup` – note write this in reST as
```reST
When the `atomgroup` is empty when :class:`DensityAnalysis` is initialized ...
```
so that you're referring to variables in the doc string). Then state what happens (warning emitted). Finally explain what the user should be doing (i.e., repeat what the warning says but elaborate). Imagine that you're the user of the function: What do you need to know when you

1. want to run DensityAnalysis for the first time
2. when you run into an error and think "maybe the docs tell me what to do because this error message was very confusing"

Rewrite with the above in mind.

The sentence 
> "If the 'AtomGroup' instance does not contain any selection of atoms and  'updating' is set to true, an empty :class: UpdatingAtomGroup is returned."

is confusing because the problem will occur as soon as `atomgroup` is empty, regardless of being updating or not. It's just that the situation is more likely to happen with an `UpdatingAtomGroup`.


Maybe @TBGAnsell has some suggestions what they _would_ have liked to see (see #3055)? Comments welcome!


For exceptions that are raised and that need explanation, add a section (after Returns)

```reST
Raises
------
ValueError: raised when ....
```

package/MDAnalysis/analysis/density.py
```diff
@@ -394,8 +402,18 @@ def _prepare(self):
...
+                smax = np.max(coord, axis=0)
+            except ValueError as err:
+                msg = ("No atoms in AtomGroup at input time frame. "
+                       "This may be intended; please ensure that "
+                       "your grid selection covers the atomic "
+                       "positions you wish to capture.")
+                warnings.warn(msg)
+                logger.warning(msg)
+                smin = self._gridcenter
+                smax = self._gridcenter
```
orbeckst, line 416: Why are `smin` and `smax` assigned the grid center coordinates?

Please explain the logic here. What consequences does it have downstream? Does the code fail further downstream if the user has *not* supplied a proper grid selection?

package/MDAnalysis/analysis/density.py
orbeckst: Add it to the *Raises* section. This needs to be documented and explained.

package/CHANGELOG
```diff
@@ -21,6 +21,8 @@ The rules for this file:
   * 2.0.0
 
 Fixes
+  * UserWarning displayed when empty atomgroup is given
```
orbeckst, line 24: you also raise a `ValueError` – please document

orbeckst (changes requested):
Thanks for your hard work on the PR. Documentation still needs to be clearer.

Maybe the original author of issue #3055 (@TBGAnsell ) would also like to chime in and suggest improvements?

package/MDAnalysis/analysis/density.py
aditya-kamath: I've given it some thought and updated the Docstring. I tried to keep it as simple as possible so that a user will know what to do! :)

package/MDAnalysis/analysis/density.py
aditya-kamath: Oh, that was done because `_set_user_grid` wanted a preassigned range from the coordinates of `AtomGroup` in selection. But there is no `AtomGroup` in this case, so I initialize a zero range grid at the grid center and take the range from the user-defined grid values.

package/MDAnalysis/analysis/density.py
TBGAnsell: Hello, the message here seems clear to me. One further suggestion could be to output the frames in which the density analysis fails, similar to the frames reported when H-bonds are not detected in PropkaTraj when skip_failure=True. This could be useful for an initial quick scan to see eg: how much equilibration time is needed for the ligand to reach the site.

aditya-kamath: Okay, maybe I'll include the frame number in the warning message!

IAlibay: The warning will only get called once at construction, so it won't do what @TBGAnsell is looking for.

That being said, whilst I'm not against the idea of including the frame number in the warning / error, I am somewhat opposed to checking every frame for the presence of atoms in the selection. My view is twofold:
1) An empty UpdatingAtomGroup at a given frame isn't necessarily a failure (unlike for PropkaTraj where we know that it's a failure [note: skip_failure captures a non-specific propka 3.x failure, not just solely the failure to detect any specific thing]).
2) Since we aren't currently checking for it, it will add extra checking code that probablyl won't be cheap. In my opinion, if folks want to do a quick scan it's much cheaper/cleaner to just do a quick loop over your updatingatomgroup.

That being said, I'm willing to default to @orbeckst's opinion on the matter.

orbeckst: Thank you @TBGAnsell for your comments — much appreciated!

I agree with @IAlibay reservations about checking every frame because of likely performance reasons. If a "scanning" feature is desired then this can be a new issue (please raise it and we discuss there). Let's keep this one focused on the original problem.

package/MDAnalysis/analysis/density.py
orbeckst: Could you do
```python
# need to be set for _set_user_grid() but will be
# overriden by user-defined values
smin = smax = None
```
to make clearer what's happening? If `None` is not possible (even though that would be the best value to indicate that we are not using these values), just setting to 0 or `[0, 0, 0]` would still be cleaner.

aditya-kamath: It will trigger this warning in set_user_grid if I do that.
`        if any(smin < umin) or any(smax > umax):`
`            msg = ("Atom selection does not fit grid --- "`
`                   "you may want to define a larger box")`
`            warnings.warn(msg)`
 `           logger.warning(msg)`

 The check is meant for regular situations to compare the atom selection with user-defined grid.

 In the case of no atoms in selection, I just want it to pass through. So setting smin and smax at the gridcenter will always pass the comparison with a user defined box.

package/MDAnalysis/analysis/density.py
```diff
@@ -229,6 +229,11 @@ class DensityAnalysis(AnalysisBase):
     are within 4 Å of the protein heavy atoms) then create an
     :class:`~MDAnalysis.core.groups.UpdatingAtomGroup` (see Examples).
 
+    DensityAnalysis will fail when the `AtomGroup` instance does not contain
+    any selection of atoms, even when `updating` is set to True. In such a
+    situation, user defined box limits can be provided to generate
+    a `Density`. Although, it should be ensured that the provided
+    grid limits encompass atoms to be selected on all trajectory frames.
```
orbeckst, line 236:
```suggestion
    a `Density` although it remains the user's responsibility to ensure that the 
    provided grid limits encompass atoms to be selected in all trajectory frames.
```

package/MDAnalysis/analysis/density.py
```diff
@@ -229,6 +229,11 @@ class DensityAnalysis(AnalysisBase):
     are within 4 Å of the protein heavy atoms) then create an
     :class:`~MDAnalysis.core.groups.UpdatingAtomGroup` (see Examples).
 
+    DensityAnalysis will fail when the `AtomGroup` instance does not contain
```
orbeckst, line 232:
```suggestion
    :class:`DensityAnalysis` will fail when the :class:`AtomGroup` instance does not contain
```

package/CHANGELOG
```diff
@@ -21,6 +21,8 @@ The rules for this file:
   * 2.0.0
 
 Fixes
+  * UserWarning/ValueError displayed when empty atomgroup is given 
+    to DensityAnalysis (Issue #3055)
```
orbeckst, line 25: A `ValueError` is not displayed but raised. Make clearer when the warning is shown and when ValueError is raised.

package/MDAnalysis/analysis/density.py
```diff
@@ -394,8 +410,18 @@ def _prepare(self):
...
+                smax = np.max(coord, axis=0)
+            except ValueError as err:
+                msg = ("No atoms in AtomGroup at input time frame. "
+                       "This may be intended; please ensure that "
+                       "your grid selection covers the atomic "
+                       "positions you wish to capture.")
+                warnings.warn(msg)
+                logger.warning(msg)
+                smin = self._gridcenter
+                smax = self._gridcenter
```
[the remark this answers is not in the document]
orbeckst: If it doesn't work then _add a comment_ to explain why this code is doing something non-sensical.

orbeckst (changes requested):
Thanks, looking better, but please address my comments.

package/MDAnalysis/analysis/density.py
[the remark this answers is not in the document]
aditya-kamath: We can do it this way! I get the intuition that it will trigger an additional warning. But I may be wrong..

package/MDAnalysis/analysis/density.py
```diff
@@ -384,6 +390,16 @@ def __init__(self, atomgroup, delta=1.0,
...
+        """Function to prepare Density
+
+        Raises
+        ------
+        ValueError
+            if AtomGroup is empty and no user defined grid is provided
+        UserWarning
+            if AtomGroup is empty and a user defined grid is provided
+
+        """
```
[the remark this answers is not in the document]
aditya-kamath: I guess I forgot to add a 'r' symbol which might be what triggers building the docs. My next commit will include this! :)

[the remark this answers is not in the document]
aditya-kamath: Hi @orbeckst, as I am just starting out with this kind of development. I am not sure how the webpage docs are built with the repository code? Perhaps there is a package I should refer to see how it's built so that I can ensure this comment is included in the webpage as well?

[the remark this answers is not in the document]
lilyminium: @aditya-kamath you can check updates to the PR in the [ReadTheDocs](https://mdanalysis--3091.org.readthedocs.build/en/3091/documentation_pages/analysis/density.html#MDAnalysis.analysis.density.DensityAnalysis) build. Otherwise, you could ``cd package/doc/sphinx``, type ``make html``, and your output will be in ``../html``.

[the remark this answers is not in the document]
aditya-kamath: Thanks! I think this issue is related to sphinx docs. Hopefully my next commit will compile this to the docs build.

orbeckst:
@aditya-kamath the doc build is failing, please have a look.

package/MDAnalysis/analysis/density.py
```diff
@@ -107,7 +107,7 @@
       After the analysis (see the :meth:`~DensityAnalysis.run` method), the resulting density is
       stored in the :attr:`density` attribute as a :class:`Density` instance.
 
-   .. automethod:: _set_user_grid
+   .. automethod:: _prepare, _set_user_grid
```
orbeckst, line 110: these might have to be separate lines

package/MDAnalysis/analysis/density.py
```diff
@@ -229,6 +229,12 @@ class DensityAnalysis(AnalysisBase):
     are within 4 Å of the protein heavy atoms) then create an
     :class:`~MDAnalysis.core.groups.UpdatingAtomGroup` (see Examples).
 
+    :class:`DensityAnalysis` will fail when the :class:`AtomGroup` instance
+    does not contain any selection of atoms, even when `updating` is set to
+    True. In such a situation, user defined box limits can be provided to
```
orbeckst, line 234: Mark up Python
```suggestion
    ``True``. In such a situation, user defined box limits can be provided to
```

package/MDAnalysis/analysis/density.py
orbeckst, line 110: Remove `_prepare`: this should not be documented as it's an internal method. Move the docs into the class doc string. This is the standard way how we document Analysis classes.

`_set_user_grid()` is a special case because it does all kinds of things and one might need to know about it.

package/MDAnalysis/analysis/density.py
```diff
@@ -384,6 +391,19 @@ def __init__(self, atomgroup, delta=1.0,
...
+        Raises
+        ------
+        ValueError
+            if AtomGroup is empty and no user defined grid is provided
+        UserWarning
+            if AtomGroup is empty and a user defined grid is provided
+
+        .. versionadded:: 1.0.0    
+        .. versionchanged:: 2.0.0
+           Now a method of :class:`DensityAnalysis`.
```
orbeckst, line 405: Move this doc string into the class doc string. See comment above.

orbeckst: Also, remove the `Now a method of :class:DensityAnalysis.` comment as it does not make sense. `_prepare` is always in a AnalysisClass.

Your change is for 2.0.0 so do not add any versionadded for 1.0.0.

package/MDAnalysis/analysis/density.py
```diff
@@ -394,8 +414,18 @@ def _prepare(self):
...
+                smax = np.max(coord, axis=0)
+            except ValueError as err:
+                msg = ("No atoms in AtomGroup at input time frame. "
+                       "This may be intended; please ensure that "
+                       "your grid selection covers the atomic "
+                       "positions you wish to capture.")
+                warnings.warn(msg)
+                logger.warning(msg)
+                smin = self._gridcenter
+                smax = self._gridcenter
```
orbeckst, line 428: Add a comment explaining the hacky assignment.

package/MDAnalysis/analysis/density.py
orbeckst: Document in a comment, please.

package/MDAnalysis/analysis/density.py
```diff
@@ -827,4 +865,4 @@ def __repr__(self):
             grid_type = 'density'
         else:
             grid_type = 'histogram'
-        return '<Density ' + grid_type + ' with ' + str(self.grid.shape) + ' bins>'
+        return f"<Density {grid_type} with {self.grid.shape} bins>"
```
orbeckst, line 868: Did you test this line?

orbeckst (changes requested):
Please see comments and fix doc build. Thanks!

package/MDAnalysis/analysis/density.py
aditya-kamath: Thanks again @orbeckst for your patience and advice! My next commit will make these changes..

package/MDAnalysis/analysis/density.py
aditya-kamath: No, perhaps I'll have to think about a test for this. I'd leave it alone, but it looks like autoPEP8 felt that the line was very long!

package/MDAnalysis/analysis/density.py
aditya-kamath: Done! :)

package/MDAnalysis/analysis/density.py
aditya-kamath: Hi @orbeckst, I've removed this change. However, a check seems to be failing and I'm not sure if it is a concern.(?)

IAlibay: The test failure was down to a non-related RDKIT test (it's a flaky test that fails sometimes on CI), I've restarted the jobs, hopefully they should pass this time.

aditya-kamath:
Hi @orbeckst , I'm just wondering if we can extend this PR to project 5 in the GSoC proposal list? I will have a basic idea on the DensityAnalysis module after this PR and I'm thinking about getting started with understanding other aspects of densities to include in my proposal..

orbeckst:
You could make a general density analysis (for RMS-fitted PBC systems) an application of the work in project 5 – that would be neat because it cannot really be done with any tools that I know.

However, let's not extend _this_ PR. It's much better to make PRs as small and self-contained as possible: easier to debug, much easier to review (and the main bottleneck is typically reviewer time/attention). You'd open a new PR for a new feature.

orbeckst:
@IAlibay is the [failure](https://github.com/MDAnalysis/mdanalysis/pull/3091/checks?check_run_id=2024455399#step:10:283)
```
TestRDKitConverter.test_assign_coordinates 
```
spurious or serious?

orbeckst (approved):
Thanks for addressing my comments.

I assume that we'll sort out the failing tests somehow, but as they are not related to your code I'll approve for now.

IAlibay:
> @IAlibay is the [failure](https://github.com/MDAnalysis/mdanalysis/pull/3091/checks?check_run_id=2024455399#step:10:283)
> 
> ```
> TestRDKitConverter.test_assign_coordinates 
> ```
> 
> spurious or serious?

So we've been treating it as spurious (see: https://github.com/MDAnalysis/mdanalysis/issues/2958), and generally re-starting CI fixes things. However, we should really get to the bottom of it before we release 2.0. The mismatch is really large :/ I'm hoping one of the RKDIT PRs that are still to be merged might fix this.

package/MDAnalysis/analysis/density.py
IAlibay: @hmacdope do you want to raise a separate issue for this? I say we go ahead with merging this PR as is and we can do the necessary changes to further validate user inputs for the grids?

IAlibay (approved):
lgtm as long as CI returns green!

IAlibay:
@hmacdope do you want to re-review/approve?

package/MDAnalysis/analysis/density.py
hmacdope: Ack I'll raise an issue.

hmacdope (approved):
LGTM. Congrats @aditya-kamath.

hmacdope (approved):
LGTM. Congrats @aditya-kamath.

hmacdope (approved):
LGTM. Congrats @aditya-kamath.

hmacdope:
Sorry for review spam, unintended.

Issue #2985, opened 2020-10-15
Some tests use fixtures whith scopes that are at odds with hypothesis


jbarnoud:
## Expected behavior ##

Running the tests does not throw warnings.

## Actual behavior ##

Hypothesis warns about the scope of some fixture potentially causing issues with generated test cases.

```
MDAnalysisTests/formats/test_libdcd.py::test_written_remarks_property
  /home/jon/miniconda3/envs/py38/lib/python3.8/site-packages/hypothesis/extra/pytestplugin.py:172: HypothesisDeprecationWarning: MDAnalysisTests/formats/test_libdcd.py::test_written_remarks_property uses the 'tmpdir' fixture, which is reset between function calls but not between test cases generated by `@given(...)`.  You can change it to a module- or session-scoped fixture if it is safe to reuse; if not we recommend using a context manager inside your test function.  See https://docs.pytest.org/en/latest/fixture.html#sharing-test-data for details on fixture scope.
    note_deprecation(

MDAnalysisTests/formats/test_libdcd.py::test_written_remarks_property
  /home/jon/miniconda3/envs/py38/lib/python3.8/site-packages/hypothesis/extra/pytestplugin.py:172: HypothesisDeprecationWarning: MDAnalysisTests/formats/test_libdcd.py::test_written_remarks_property uses the 'dcd' fixture, which is reset between function calls but not between test cases generated by `@given(...)`.  You can change it to a module- or session-scoped fixture if it is safe to reuse; if not we recommend using a context manager inside your test function.  See https://docs.pytest.org/en/latest/fixture.html#sharing-test-data for details on fixture scope.
    note_deprecation(
```

## Code to reproduce the behavior ##

`pytest formats/test_libdcd.py`

## Current version of MDAnalysis ##

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) 2.0.0-dev0 64a7c051f1a51ff2a89b9da9257a21cea0ef00fa
- Which version of Python (`python -V`)? Python 3.8.5
- Which operating system? Ubuntu 20.04 on WSL

lilyminium:
This now reliably fails.

See: https://github.com/MDAnalysis/mdanalysis/pull/3091/checks?check_run_id=1674311671#step:10:596

and https://github.com/MDAnalysis/mdanalysis/pull/2664/checks?check_run_id=1715357196#step:10:296

```
[...]

PR #3094, opened 2021-01-17
Copy README into package

IAlibay:
Fixes pypi upload issues

This is a bit of a silly one, setup.py requires README.rst to create the long format description for pypi. However the README.rst doesn't get uploaded to PyPi so you end up with a case where you can't build off it.

The only quick solution I could find was to have a copy of that README.rst within `package` that will then get uploaded.

Thoughts / quick review @MDAnalysis/coredevs we would like to push a release as soon as.

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


IAlibay:
tests are failing because of #2985

richardjgowers:
@IAlibay you'll also have to include this in MANIFEST for it to get included in the sdist

IAlibay:
Done. I'm still not happy with the idea of having a copy of the main readme in package. I have changed the name so as to make it less confusing in case end users look at the source on pypi.

orbeckst (approved):
thanks!

PR #3097, opened 2021-01-18
Change tmpdir fixture to tmpdir_factory (#3095)

lilyminium:
Master version of #3095 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/testsuite/MDAnalysisTests/formats/test_libdcd.py
+++ b/testsuite/MDAnalysisTests/formats/test_libdcd.py
@@ -26,6 +26,7 @@ import platform
 
 import hypothesis.strategies as strategies
 from hypothesis import example, given
+import hypothesis
 import numpy as np
 
 from numpy.testing import (assert_array_almost_equal, assert_equal,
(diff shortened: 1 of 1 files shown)


testsuite/MDAnalysisTests/formats/test_libdcd.py
```diff
@@ -322,14 +323,18 @@ def write_dcd(in_name, out_name, remarks='testing', header=None):
                     and sys.maxsize <= 2**32) or
                     platform.machine() == 'aarch64',
                    reason="occasional fail on 32-bit windows and ARM")
+# occasionally fails due to unreliable test timings
+@hypothesis.settings(deadline=None)  # see Issue 3096
```
IAlibay, line 327: Does this remove the need for the xfail above?

lilyminium: Hmm let's give it a try

IAlibay: I guess we're still not testing on ARM64, so it'd just be the 32bit Windows stuff.

orbeckst:
Did the 3.8 https://github.com/MDAnalysis/mdanalysis/pull/3097/checks?check_run_id=1724495378 fail because if issues in the citation processing?

lilyminium:
@orbeckst possibly, I think it may be an intermittent error?

testsuite/MDAnalysisTests/formats/test_libdcd.py
lilyminium: @IAlibay no idea why but it makes the Travis Py2.7 test fail again, so the xfail can be left for `develop` instead....

Issue #3101, opened 2021-01-22
Failed compilation using pip on macOS Big Sur


tclick:
## Expected behavior ##

MDAnalysis should install correctly when using `pip install mdanalysis`

## Actual behavior ##

    clang -Wno-unused-result -Wsign-compare -Wunreachable-code -DNDEBUG -g -fwrapv -O3 -Wall -I/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include -I/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include -I/Users/tclick/.pyenv/versions/3.9.1/lib/python3.9/site-packages/numpy/core/include -IMDAnalysis/lib/formats/include -I/Users/tclick/.pyenv/versions/3.9.1/include/python3.9 -c MDAnalysis/lib/formats/libdcd.c -o build/temp.macosx-11.1-x86_64-3.9/MDAnalysis/lib/formats/libdcd.o -std=c99 -ffast-math -O3 -funroll-loops -fsigned-zeros
    In file included from MDAnalysis/lib/formats/libdcd.c:629:
    In file included from /Users/tclick/.pyenv/versions/3.9.1/lib/python3.9/site-packages/numpy/core/include/numpy/arrayobject.h:4:
    In file included from /Users/tclick/.pyenv/versions/3.9.1/lib/python3.9/site-packages/numpy/core/include/numpy/ndarrayobject.h:12:
    In file included from /Users/tclick/.pyenv/versions/3.9.1/lib/python3.9/site-packages/numpy/core/include/numpy/ndarraytypes.h:1822:
    /Users/tclick/.pyenv/versions/3.9.1/lib/python3.9/site-packages/numpy/core/include/numpy/npy_1_7_deprecated_api.h:17:2: warning: "Using deprecated NumPy API, disable it with "          "#define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION" [-W#warnings]
    #warning "Using deprecated NumPy API, disable it with " \
     ^
    In file included from MDAnalysis/lib/formats/libdcd.c:635:
    MDAnalysis/lib/formats/include/readdcd.h:698:14: warning: unused variable 'tmbuf' [-Wunused-variable]
      struct tm *tmbuf;
                 ^
    MDAnalysis/lib/formats/include/readdcd.h:697:10: warning: unused variable 'cur_time' [-Wunused-variable]
      time_t cur_time;
             ^
[...]

PR #3102, opened 2021-01-22
Adds Big Sur to ci matrix

IAlibay:
Tests #3101 

Just checking how things go if we just blindly add Big Sur to our CI matrix

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/.github/workflows/gh-ci.yaml
+++ b/.github/workflows/gh-ci.yaml
@@ -39,6 +39,12 @@ jobs:
               run_type: FULL
               install_hole: true
               codecov: true
+            - name: macOS_11
+              os: macos-11.0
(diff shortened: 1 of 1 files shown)


IAlibay:
From https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on 

"Note: The macOS 11.0 virtual environment is currently provided as a private preview only. Any users or organizations that are already using this runner can continue using it, but we're not accepting any further users or organizations at this time. The macos-latest YAML workflow label still uses the macOS 10.15 virtual environment."

I guess we can't add it to CI for now then. -- closing

PR #3104, opened 2021-01-23
Fix python_requires

### LINKED ISSUE #3103: python_requires not working?
We are currently testing the release of 1.0.1 on testpypi (https://test.pypi.org/project/MDAnalysis/1.0.2/ please ignore the 1.0.2 tag, we originally uploaded 1.0.1 but that was broken so we've had to bump up the version number).

Currently `setup.py` on master has the following:
https://github.com/MDAnalysis/mdanalysis/blob/658a73d127716c6e1f5d4bcfd6dd5b12ecf71cef/package/setup.py#L629

My understanding is that this should not allow py3.5 to get installed (although I think technically we are meant to still support 3.5 in 1.0.x?). However, for some reason it's still possible for me to pip install it from a py3.5.6 environment.

Any ideas @MDAnalysis/coredevs ?

IAlibay:
Fixes #3103 

 - Fixes `python_requires` in setup.py by specifying the minor version
 - Also sets the max python version to <3.9 (we aren't testing 3.9 for master right now, so if we release it, it's best we avoid allowing users to install 3.9+).

# To do

For some reason we don't let 3.5 on `master` anymore. I was under the impression that we still allowed py3.5 on 1.0.x. I think @orbeckst made the change (from the blame history)?

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ env:
     - MAIN_CMD="pytest ${PYTEST_LIST}"
     - SETUP_CMD="${PYTEST_FLAGS} --cov=MDAnalysis"
     - BUILD_CMD="pip install -e package/ && (cd testsuite/ && python setup.py build)"
-    - CONDA_MIN_DEPENDENCIES="mmtf-python biopython networkx cython matplotlib scipy griddataformats hypothesis gsd codecov"
+    - CONDA_MIN_DEPENDENCIES="pip<21.0 mmtf-python biopython networkx cython matplotlib scipy griddataformats hypothesis gsd codecov"
     - CONDA_PY2_DEPENDENCIES="six funcsigs"
     - CONDA_STANDARD_DEPENDENCIES="${CONDA_MIN_DEPENDENCIES} seaborn>=0.7.0 clustalw=2.1 netcdf4==1.3.1 scikit-learn joblib>=0.12 tqdm>=4.43.0"
     - CONDA_DEPENDENCIES="${CONDA_STANDARD_DEPENDENCIES} mock chemfiles"
--- a/package/setup.py
+++ b/package/setup.py
@@ -626,7 +626,7 @@ if __name__ == '__main__':
                         ],
           },
           ext_modules=exts,
-          python_requires=">=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3,!=3.4.*,!=3.5.*,<3.9",
+          python_requires=">=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3,!=3.4.*,<3.9",
           # all standard requirements are available through PyPi and
           # typically can be installed without difficulties through setuptools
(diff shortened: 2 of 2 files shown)


richardjgowers (approved):

richardjgowers (approved):

orbeckst:
I don't remember why I removed 3.5 in PR #2738. You are correct in that I shouldn't have done it because for 1.0.1 we need to support whatever 1.0.0 supported, and in 1.0.0 we did not remove support; in 0.20.0 we removed support for >2.7 <3.5.

The metadata also still showed support for 3.5.

Thanks for catching my mistake.

orbeckst (approved):
👍

IAlibay:
Interesting something about this change is causing py3.5 to break 😕

orbeckst:
The pip version is not up to it (from https://travis-ci.com/github/MDAnalysis/mdanalysis/jobs/474702826)
```
  File "/home/travis/miniconda/envs/test/lib/python3.5/runpy.py", line 193, in _run_module_as_main

    "__main__", mod_spec)

  File "/home/travis/miniconda/envs/test/lib/python3.5/runpy.py", line 85, in _run_code

    exec(code, run_globals)

  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/pip/__main__.py", line 21, in <module>

    from pip._internal.cli.main import main as _main

  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/pip/_internal/cli/main.py", line 60

    sys.stderr.write(f"ERROR: {exc}")

                                   ^

SyntaxError: invalid syntax
```
A f-string is used in this pip version.

I wonder why does conda install a pip version that is incompatible with the python version? 

(Maybe I ran into the same thing before and couldn't take it anymore and upped the version...)

IAlibay:
Ooo that's really strange, it was running 20.3 just this morning when I re-ran one of the failed jobs, but it's picking up 21 now...

😫 yeah looks like they released 21 just 9h ago.. 

Related issue: https://github.com/pypa/pip/issues/9500

orbeckst:
I sometimes deeply despise new software...

F2MZK8A6N70W (approved):

IAlibay:
@orbeckst New release bugs are just a good way keep us on our toes (and burn off HPC allocations) 🙃 

Pinning on CI seems to do the trick here. I'm assuming upstream will yank/amend accordingly, so it hopefully shouldn't impact 1.0.1.

py2.7 CI stalled again 🤯 I've restarted it so hopefully that completes....

IAlibay:
py2.7 CI is very flaky, had to keep restarting until it finally worked. I don't think there's much we can do here since we're at a point where py2.7 conda installs is a long mess of frozen solves :/ If someone thinks it's relevant please do open an issue though

PR #3108, opened 2021-01-24
Use absolute file paths in ITPParser

### LINKED ISSUE #3037: ITPParser doesn't like some relative paths
## Actual behavior ##

Relative paths (when going "upwards"/`../`) in the ITPParser seem to fail.

I think that all is needed is for `find_path` to have an `abspath` call early on, as absolute paths seem to work fine.

```bash
cp file.itp ./
mkdir something
cd something
```

``` python
import MDAnalysis as mda

u = mda.Universe('../file.itp', format='ITP')

....

```

- Which operating system?

aditya-kamath:
Fixes #3037

Changes made in this Pull Request
- ITPParser now accepts relative paths 

PR Checklist
------------
 - [] Tests?
 - [] Docs?
 - [x] CHANGELOG updated?


package/CHANGELOG
```diff
@@ -21,6 +21,7 @@ The rules for this file:
   * 2.0.0
 
 Fixes
+  * ITPParser now accepts relative paths
```
IAlibay, line 24: Can you include the issue number here please?

Can you also add yourself to the author list above and the `AUTHORS` file?

richardjgowers (changes requested):
Needs a test

package/MDAnalysis/topology/ITPParser.py
```diff
@@ -268,7 +268,7 @@ def find_path(self, path):
             path = path.name
         except AttributeError:
             pass
-
+        path = os.path.abspath(path)
```
IAlibay, line 271: More of a style thing, but can this be just done on line 268 rather than assigning path twice in a row?

IAlibay (changes requested):
Couple of extra things on top of @richardjgowers' comment

package/CHANGELOG
aditya-kamath: Oh, I think I am already on the authors list on both places..

IAlibay: Sorry about that, just the issue number then.

IAlibay:
This might need some rethinking, it's breaking CI: https://github.com/MDAnalysis/mdanalysis/pull/3108/checks?check_run_id=1757898726#step:10:999

aditya-kamath:
Hmm, I guess I should understand how these path strings are manipulated in general. This did seem to work when I tried to load a .ITP file from another folder.

package/MDAnalysis/topology/ITPParser.py
```diff
@@ -268,7 +268,7 @@ def find_path(self, path):
             path = path.name
```
tylerjereddy, line 268: A slightly different change seems to allow the testsuite to pass while still allowing relative file path handling in my hands:

```diff
--- a/package/MDAnalysis/topology/ITPParser.py
+++ b/package/MDAnalysis/topology/ITPParser.py
@@ -265,7 +265,7 @@ class GmxTopIterator:
             current_file = self.current_file
 
         try:
-            path = path.name
+            path = os.path.abspath(path.name)
         except AttributeError:
             pass
```

I think it addresses Irfan's comment as well

aditya-kamath:
@richardjgowers, To add a test, would it be okay to make another folder and show that it can access an ITP file? I'm not able to figure out a better test..

IAlibay:
@MDAnalysis/gsoc-mentors can we get a follow-up review here?

@aditya-kamath can you fix the merge conflict and also maybe provide some example code for what you mean in https://github.com/MDAnalysis/mdanalysis/pull/3108#issuecomment-768661986? (that way it's easier for us to comment on how best to change it).

aditya-kamath:
I guess I'm just asking what kind of test should I provide in this case? I want to show that this parser can now access paths outside root directory. I tried this manually and it works but I'm thinking in a pytest context. :)

hmacdope:
In a pytest context you could try just providing any relative path to the ITP you are meant to be testing. This includes the ITPs in the `/MDAnalysisTests/data/` folder. That's probably a good place to start.

aditya-kamath:
I've added a test, and it seems to work fine when I do `pytest test_itp.py`. But, this seems to fail? I've also checked that it does indeed work with relative paths.

testsuite/MDAnalysisTests/topology/test_itp.py
```diff
@@ -335,3 +335,8 @@ def test_missing_endif(self):
         with pytest.raises(IOError):
             with self.parser(ITP_no_endif) as p:
                 top = p.parse(include_dir=GMX_DIR)
+
+class Testrelativepath:
+
+    def test_relpath(self):
+        u = mda.Universe('../data/gromacs_ala10.itp', format='ITP')
```
IAlibay, line 342: Where is this itp file being generated?

IAlibay (changes requested):
> I've added a test, and it seems to work fine when I do `pytest test_itp.py`. But, this seems to fail? I've also checked that it does indeed work with relative paths.

See here: https://github.com/MDAnalysis/mdanalysis/pull/3108/checks?check_run_id=2089681447#step:10:366

It doesn't seem like you're creating the temporary ITP file?

testsuite/MDAnalysisTests/topology/test_itp.py
aditya-kamath: Okay, I get it. I guess I need to load this file first..!

lilyminium: @aditya-kamath Python treats files with relative paths, as relative to whichever directory you've called the test from. Users can call `pytest` from anywhere in their system, which is why we make all our paths absolute in the `datafiles.py` file and import them when we want to use our test files. It might be easier for you to write a temporary ITP file using [pytest's tools](https://docs.pytest.org/en/stable/tmpdir.html) that only exists for the scope of this function or module, and try to read that.

aditya-kamath: @lilyminium , This really helps. Now, I understand how to write this test! :)

aditya-kamath: I've now created a temporary folder where an itp file is written and have accessed it with the parser. Although, I am still not completely sure if this asserts that itp parser can access relative locations.?

testsuite/MDAnalysisTests/topology/test_itp.py
```diff
@@ -335,3 +335,14 @@ def test_missing_endif(self):
         with pytest.raises(IOError):
             with self.parser(ITP_no_endif) as p:
                 top = p.parse(include_dir=GMX_DIR)
+
+class Testrelativepath:
+
+    def test_relpath(self, tmp_path):
+        content="[ atoms ]\n" +\
```
lilyminium, line 342: It would be neater here to use `"""` multiline strings. I don't think the indentation will matter to the ITPParser.

testsuite/MDAnalysisTests/topology/test_itp.py
```diff
@@ -335,3 +335,14 @@ def test_missing_endif(self):
...
+                "     1      H      1    SOL    HW1      1       0.41    1.00800"
+        d = tmp_path / "sub"
+        d.mkdir()
+        p = d / "test.itp"
+        p.write_text(content)
+        u = mda.Universe(tmp_path / "sub/test.itp", format='ITP')
```
lilyminium, line 348: I think this would still be an absolute path. Instead, I would suggest using the `tmpdir` fixture, creating a new directory inside that temporary one, and making sure that the relative path you test looks something like "../test.itp".

package/MDAnalysis/topology/ITPParser.py
```diff
@@ -265,10 +265,9 @@ def find_path(self, path):
             current_file = self.current_file
 
         try:
-            path = path.name
+            path = os.path.abspath(path.name)
```
lilyminium, line 268: This try/except is to check whether `path` is a string (which triggers the AttributeError) or a `Path` object from `pathlib` (so Path.path is the string version). Does your fix still work if a relative string path is passed in instead, triggering the AttributeError?

By the way, your test should check for both, just in case :)

package/CHANGELOG
```diff
@@ -21,6 +21,7 @@ The rules for this file:
   * 2.0.0
 
 Fixes
+  * ITPParser now accepts relative paths (Issue #3037)
```
lilyminium, line 24:
```suggestion
  * ITPParser now accepts relative paths (Issue #3037, PR #3108)
```

lilyminium (changes requested):
Thank you for fixing this, @aditya-kamath. I just have a few comments :)

package/MDAnalysis/topology/ITPParser.py
aditya-kamath: Hi @lilyminium, that's a good observation! However, if the user isn't providing a string and triggering this `AttributeError`, it's probably doing the right thing by just passing and eventually raising an IO error. I can write a test to check if an IOerror is raised when a relative string is passed but not as a string.

lilyminium: Hi @aditya-kamath, 

> if the user isn't providing a string and triggering this AttributeError

If a string is provided and the AttributeError is triggered, does your fix still work? The IOError test is probably not necessary. However, your test should check that your fix applies to both a relative `Path` (https://docs.python.org/3/library/pathlib.html) and a relative string.

> (so Path.path is the string version)

This was a typo, sorry. `Path.name` returns the base name of the Path, as a string.

aditya-kamath: Hmm, the answer to if it will work when there's an Attribute error is no! :( 
I'm assuming that the situation when this will happen is when os.path.abspath is not able to convert a relative string to an absolute path. I'll write the two situations in my test to see if there's a situation when this might happen! :D

testsuite/MDAnalysisTests/topology/test_itp.py
aditya-kamath: Hi @lilyminium, I've been looking at pytest tools and os.paths. I see that there's a command called relative_to in the pathlib module which computes a relative path to a specified directory. My approach would have been to make a tmpdir and set that as 'observer' directory and use that command (relative_to) to find the path that it sees. 

As it turns out, the relative_to command only works when computing paths from parent directory reference, not sibling directory. So, I can't compute something like '../dir'. The second thing is even if I manually create a relative string which tmpdir would use  to read my temporary itp file, I am not sure how to ask pytest to use my tmpdir while performing this test. 

I might have missed something here, especially because I am new to pytest, but I am happy to hear any advice you have here!

lilyminium: Hi @aditya-kamath,

If you go into your temporary directory, write a file called "test.itp", make a new subdirectory, and change into that directory, you should be able to test that file with the path "../test.itp". You might like the tmpdir.as_cwd context manager for that.

aditya-kamath: Great advice! I figured the tests out. Just a note that I've done the reverse of what you might expect. Instead of testing a relative path and then converting it to a relative string. I've first tested a relative string and then followed it up by converting it to a path. Hope that's alright!

lilyminium:
Thanks for adding the new tests, @aditya-kamath. It's unexpected that they pass, as I expect not accounting for the [AttributeError](https://github.com/MDAnalysis/mdanalysis/pull/3108#discussion_r602647621) to break the test. On further investigation, the ITPParser could always deal with relative paths when used to create a Universe. I tested this by reverting your `os.path.abspath` change; the tests still pass. However, what it cannot deal with is relative paths when they're `#include`d in earlier ITPs. That means the test case should actually be a file like this:

```
#include "../sub1/atoms.itp"
        
[ atoms ]
 1      H      1    SOL    HW1      1       0.41    1.00800
```

with a corresponding `atoms.itp`. This means that instead of writing a temporary file, you could actually just include a new ITP file in the [data directory](https://github.com/MDAnalysis/mdanalysis/tree/develop/testsuite/MDAnalysisTests/data/gromacs/gromos54a7_edited.ff) and as a variable in [datafiles.py](https://github.com/MDAnalysis/mdanalysis/blob/develop/testsuite/MDAnalysisTests/datafiles.py) that references an upper-level ITP file. It's up to you which one you choose.

@richardjgowers did you first find the bug when one of your ITP files `#include`d an upper level ITP file, or was the relative path included in some other way?

aditya-kamath:
@lilyminium , I tried #imcluding another file from the test file and it failed giving me an OS error 
test_relstring - OSError: Failed to load from the topology file ../sub1/test.itp with parser <class 'MDAnalysis.topology.ITPParser.ITPParser'

lilyminium:
Hi @aditya-kamath, it's easier for everyone to diagnose the error if you push to the PR and let the tests run with the full error trace :)

aditya-kamath:
You can check the failed test now!

lilyminium:
Right, that looks like the original issue in #3037. It means your fix isn't working for strings.

aditya-kamath:
Ah shoot! I'll look into this in more detail. Perhaps there's something up with the way these itp files are calling each other.

aditya-kamath:
Hi @lilyminium, I went back and reverted my fix and it seems that the tests (when they did not refer other .itp files) fail. I guess this fix is fixing something but not everything. What this fixes right now is the ITP parser being able to load files from relative paths.

 You are however right when you say that if that itp file tries to include another .itp file this fails. In this case, I'd actually recommend what @hejamu proposes in PR #3204  which is to have the ITP parser check for and find other .ITP files and include them. 

I'm not sure how useful this fix is to have the itparser be able to load relative files when they include more files, but it does fix something :P

hejamu:
Hi @aditya-kamath,

what I was trying to say with my comments in the other PR was just that the parser could continue with simple issuing a warning to the user in cases where the include file really can't be found. This is the case if one inherits data from elsewhere and does not have all forcefield data, or if newer versions of Gromacs don't ship specific files anymore.

The Parser could easily determine if it has enough information to build the system, a missing molecule definition for example already breaks the parser (which would need to be done in a controlled manner), allowing simple system analysis on older or incomplete systems.

But this is out of the scope for both PRs, I just wanted to get this thought out there to be discussed.

lilyminium:
> In this case, I'd actually recommend what @hejamu proposes in PR #3204 which is to have the ITP parser check for and find other .ITP files and include them.

@hejamu's fix is for a different problem to this PR. In fact, #3204 actually depends on this problem being fixed. As it stands, if a file `#include`s a relative path, and that file does actually exist, the current ITPParser will not find it because it is not checking absolute paths.

I looked and you had the right idea earlier with `os.path.abspath`. While @IAlibay's suggestion to move it into the try/except saved some lines of code, it didn't account for when the try/except actually triggered the error before calling `abspath`. You should consider moving it back.

aditya-kamath:
Okay, sounds good! I'll try to get the parser working for all these cases. The next steps may be to emit warnings for broken files...

aditya-kamath:
Btw, @lilyminium , I still believe this fix is working well as is. I tested it to see if it could parse files which `#include`d other files and while it could not do it before, it seems to be able to parse relative paths which '#include other paths. I've written a test for this which creates three temporary folders and while located in one of the folders, tries to parse an itp file in it's sibling directory which #includes a second itp file. I've run pytest fine with this but it's breaking CI so I guess I missed something..

lilyminium:
@aditya-kamath your tests still pass even without your changes. It's often good to come up with a falsifying example to start off with. Can you get something like this test to pass?

```python
def test_relative_path(self, tmpdir):
        test_itp_content = '#include "../atoms.itp"'
        atoms_itp_content = """
        [ moleculetype ]
        UNK 3

        [ atoms ]
        1      H      1    SOL    HW1      1       0.41    1.00800
        """

        with tmpdir.as_cwd():
            with open("atoms.itp", "w") as f:
                f.write(atoms_itp_content)
            subdir = tmpdir.mkdir("subdir")
            with subdir.as_cwd():
                with open("test.itp", "w") as f:
                    f.write(test_itp_content)
                subsubdir = subdir.mkdir("subsubdir")
                with subsubdir.as_cwd():
                    u = mda.Universe("../test.itp")
                    assert len(u.atoms) == 1
```

Having a look at how the function gets the output `dir_path` may help.

aditya-kamath:
Thanks for the test @lilyminium . It seems to pass on my computer, I've added it to this branch so let's see if it passes the checks here.

aditya-kamath:
All checks passed with the new test, there might yet be a case situation we are missing!

lilyminium:
Awesome, thank you @aditya-kamath -- I think this fix covers most cases then! Could you please just address the PEP8 comments by the bot in https://github.com/MDAnalysis/mdanalysis/pull/3108#issuecomment-797022010 and then I think this is good to go!

Sorry for the delay in replying, got caught up in other things. Please just ping me when you're happy.

aditya-kamath:
Great! :D
I've fixed most of the PEP8 errors. The last couple of them is because I've used a class to put the test functions in. This is not really necessary, mostly for a organized code structure.

testsuite/MDAnalysisTests/topology/test_itp.py
```diff
@@ -366,3 +366,57 @@ def test_missing_endif(self):
         with pytest.raises(IOError):
             with self.parser(ITP_no_endif) as p:
                 top = p.parse(include_dir=GMX_DIR)
+
+class TestRelativePath:
+
+
```
lilyminium, line 372:
```suggestion
```

testsuite/MDAnalysisTests/topology/test_itp.py
```diff
@@ -366,3 +366,57 @@ def test_missing_endif(self):
         with pytest.raises(IOError):
             with self.parser(ITP_no_endif) as p:
                 top = p.parse(include_dir=GMX_DIR)
+
```
lilyminium, line 369:
```suggestion


```

lilyminium (changes requested):
The bot is just asking you to add an extra line / remove some lines in the class :)

aditya-kamath:
It's looking good @lilyminium!

lilyminium (approved):

lilyminium:
@richardjgowers @IAlibay could you guys have another look at this? If not, I'll dismiss the reviews as stale.

lilyminium:
Thank you @aditya-kamath for your patience on this on this very helpful fix!

PR #3114, opened 2021-02-02
[maint] numpy 1.20 type deprecations

### LINKED ISSUE #3113: np.int, np.float etc now deprecated
## Expected behavior ##

Tests pass

## Actual behavior ##

They appear to be failing because numpy 1.20 has deprecated `np.int` and `np.float`.

https://github.com/MDAnalysis/mdanalysis/pull/3013/checks?check_run_id=1810195527#step:10:537

We should switch to the built-in `int`/`float` types, or choose the more precise `np.int32/64` etc.

``` python
import MDAnalysis as mda

u = mda.Universe(PSF, DCD)

....

```

- Which operating system?

IAlibay:
Fixes #3113 

 - test_helix_analysis int->int32 (to be fair int16 would be suitable here...)

Running as draft to see if there's extra stuff I'm missing since grep isn't always very helpful...

PR Checklist
------------
 - Tests?
 - Docs?
 - CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/coordinates/RDKit.py
+++ b/package/MDAnalysis/coordinates/RDKit.py
@@ -480,9 +480,9 @@ def _add_mda_attr_to_rdkit(attr, value, mi):
 
 def _set_atom_property(atom, attr, value):
     """Saves any attribute and value into an RDKit atom property"""
-    if isinstance(value, (float, np.float)):
+    if isinstance(value, (float, np.floating)):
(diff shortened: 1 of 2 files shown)


IAlibay:
Looks like I need to fix `_set_atom_property` in the RDKIT converter too.

richardjgowers (approved):

IAlibay:
Thanks for the review @richardjgowers, CI is coming back green now so I'll merge.

Issue #3052, opened 2020-12-02
MAINT: add gh CI cron job for numpy-dev (and others)


IAlibay:
In our Travis config we had a CI run for the dev version of numpy which was added by @tylerjereddy.

This is currently not present in our github actions CI workflow. It would probably be good to it back, possibly with other types of jobs we want to run, e.g.:

- a full sweep of different gcc/clang versions? #2727 
- different versions of ubuntu (currently we target 18.04 LTS, 16.04 LTS and 20.04 LTS are available) and macOS (currently we target 10.15, 11.0 is available)

PR #3126, opened 2021-02-19
Update to chemfiles 0.10

Luthaf:
 - Chemfiles coordinate reader updated to the API changes in chemfiles 0.10 (in practice, the only relevant change here is a different constructor for UnitCell).

Chemfiles 0.10 contains multiple fixes to different format readers; and add support for crystalographic CIF files, and extended XYZ (cf https://wiki.fysik.dtu.dk/ase/ase/io/formatoptions.html#extxyz)

PR Checklist
------------
 - [X] Tests?
 - [X] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/.github/workflows/gh-ci.yaml
+++ b/.github/workflows/gh-ci.yaml
@@ -15,7 +15,7 @@ defaults:
 
 env:
   MDA_CONDA_MIN_DEPS: "pip pytest==6.1.2 mmtf-python biopython networkx cython matplotlib-base scipy griddataformats hypothesis gsd codecov threadpoolctl openmm"
-  MDA_CONDA_EXTRA_DEPS: "seaborn>=0.7.0 clustalw=2.1 netcdf4 scikit-learn joblib>=0.12 chemfiles tqdm>=4.43.0 tidynamics>=1.0.0 rdkit>=2020.03.1 h5py"
+  MDA_CONDA_EXTRA_DEPS: "seaborn>=0.7.0 clustalw=2.1 netcdf4 scikit-learn joblib>=0.12 chemfiles>=0.9 tqdm>=4.43.0 tidynamics>=1.0.0 rdkit>=2020.03.1 h5py"
   MDA_PIP_MIN_DEPS: 'coveralls coverage<5 pytest-cov pytest-xdist'
   MDA_PIP_EXTRA_DEPS: 'duecredit parmed'
 
@@ -111,7 +111,7 @@ jobs:
         add-pip-as-python-dependency: true
         mamba-version: "*"
         architecture: x64
-        
+
     - name: install_deps
       env:
         MDA_CONDA_FULL_DEPS: "${{ env.MDA_CONDA_MIN_DEPS }} ${{ env.MDA_CONDA_EXTRA_DEPS }}"
@@ -128,7 +128,7 @@ jobs:
         if [ ${{ matrix.name }} = "asv_check" ]; then
           pip install asv
         fi
-    
+
     - name: check_setup
       run: |
         # Check OS and python setup
--- a/.github/workflows/gh-ci.yaml
+++ b/.github/workflows/gh-ci.yaml
@@ -107,6 +107,10 @@ jobs:
         python-version: ${{ matrix.python-version }}
         auto-update-conda: true
         channel-priority: flexible
+        # The order of these channel matters: both provide "fundamental"
+        # software (curl, ssh, ...), but the version in biobuilds are very
+        # outdated. This can lead to problem when loading shared libraries,
+        # see https://github.com/MDAnalysis/mdanalysis/pull/3126#issuecomment-813112080
         channels: conda-forge, biobuilds
         add-pip-as-python-dependency: true
         mamba-version: "*"
--- a/package/MDAnalysis/coordinates/chemfiles.py
+++ b/package/MDAnalysis/coordinates/chemfiles.py
@@ -81,6 +81,7 @@ Helper functions
 """
 from distutils.version import LooseVersion
 import warnings
+import numpy as np
 
 from . import base
 
@@ -394,7 +395,10 @@ class ChemfilesWriter(base.WriterBase):
 
         lengths = ts.dimensions[:3]
         angles = ts.dimensions[3:]
-        if angles[0] == 0.0 and angles[1] == 0.0 and angles[2] == 0.0:
+
+        # if there is no cell information in this universe, still pass a valid
+        # cell to chemfiles
+        if np.all(ts.dimensions == 0.0):
             angles = [90, 90, 90]
 
         if chemfiles.__version__.startswith("0.9"):
--- a/package/MDAnalysis/coordinates/chemfiles.py
+++ b/package/MDAnalysis/coordinates/chemfiles.py
@@ -342,7 +342,7 @@ class ChemfilesWriter(base.WriterBase):
         constructor.
 
         If `obj` contains velocities, and the underlying format supports it, the
-        velocities are writen to the file. Writing forces is unsupported at the
+        velocities are written to the file. Writing forces is unsupported at the
         moment.
 
         Parameters
@@ -356,21 +356,23 @@ class ChemfilesWriter(base.WriterBase):
            Deprecated support for Timestep argument has now been removed.
            Use AtomGroup or Universe as an input instead.
         """
-        if hasattr(obj, "atoms"):
+        try:
+            atoms = obj.atoms
+        except AttributeError:
+            errmsg = "Input obj is neither an AtomGroup or Universe"
+            raise TypeError(errmsg) from None
+        else:
             if hasattr(obj, "universe"):
                 # For AtomGroup and children (Residue, ResidueGroup, Segment)
                 ts_full = obj.universe.trajectory.ts
-                if ts_full.n_atoms == obj.atoms.n_atoms:
+                if ts_full.n_atoms == atoms.n_atoms:
                     ts = ts_full
                 else:
                     # Only populate a time step with the selected atoms.
-                    ts = ts_full.copy_slice(obj.atoms.indices)
+                    ts = ts_full.copy_slice(atoms.indices)
             elif hasattr(obj, "trajectory"):
                 # For Universe only --- get everything
                 ts = obj.trajectory.ts
-        else:
-            errmsg = "Input obj is neither an AtomGroup or Universe"
-            raise TypeError(errmsg) from None
 
         frame = self._timestep_to_chemfiles(ts)
         frame.topology = self._topology_to_chemfiles(obj, len(frame.atoms))
--- a/package/MDAnalysis/coordinates/chemfiles.py
+++ b/package/MDAnalysis/coordinates/chemfiles.py
@@ -23,9 +23,9 @@
 Reading trajectories with *chemfiles* --- :mod:`MDAnalysis.coordinates.chemfiles`
 =================================================================================
 
-MDAnalysis interoperates with the `chemfiles`_ library. The *chemfiles* C++ library 
+MDAnalysis interoperates with the `chemfiles`_ library. The *chemfiles* C++ library
 supports an expanding set of file formats, some of which are not natively supported by
-MDAnalysis. Using the *CHEMFILES* reader you can use  `chemfiles`_ for the low-level 
+MDAnalysis. Using the *CHEMFILES* reader you can use  `chemfiles`_ for the low-level
 file reading. Check the list of `chemfile-supported file formats <formats>`_.
 
 .. _chemfiles: https://chemfiles.org
@@ -37,7 +37,7 @@ file reading. Check the list of `chemfile-supported file formats <formats>`_.
 Using the CHEMFILES reader
 --------------------------
 
-When reading, set the ``format="CHEMFILES"`` keyword argument and I/O is delegated to 
+When reading, set the ``format="CHEMFILES"`` keyword argument and I/O is delegated to
 `chemfiles`_. For example::
 
    >>> import MDAnalysis as mda
@@ -82,7 +82,7 @@ Helper functions
 from distutils.version import LooseVersion
 import warnings
 
-from . import base, core
+from . import base
 
 try:
     import chemfiles
@@ -94,6 +94,7 @@ except ImportError:
 
     class MockTrajectory:
         pass
+
     chemfiles = types.ModuleType("chemfiles")
     chemfiles.Trajectory = MockTrajectory
 else:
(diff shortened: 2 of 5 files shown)


package/MDAnalysis/coordinates/chemfiles.py
```diff
@@ -333,7 +347,12 @@ def _timestep_to_chemfiles(self, ts):
         if ts.has_velocities:
             frame.add_velocities()
             frame.velocities[:] = ts.velocities[:]
-        frame.cell = chemfiles.UnitCell(*ts.dimensions)
+
```
richardjgowers, line 350: If this is the only real change, could we have a try/except here and allow 0.9 and 0.10?

Luthaf: This is the only change in the part of the API used in MDA yes.

I could add a version check here, but for the future I'd like to not keep old versions around too long, and drop 0.9 when 0.11 comes out 😃

IAlibay: Either way this goes, can the minimum version be pinned in the gh actions CI workflow? That way we can be sure we're picking up the right thing.

Luthaf:
@IAlibay 

> Either way this goes, can the minimum version be pinned in the gh actions CI workflow? That way we can be sure we're picking up the right thing.

I've pinned chemfiles to `== 0.9.3` in CI, do you think there should be automated tests with both 0.9.3 and 0.10.x?

IAlibay:
> @IAlibay
> 
> > Either way this goes, can the minimum version be pinned in the gh actions CI workflow? That way we can be sure we're picking up the right thing.
> 
> I've pinned chemfiles to `== 0.9` in CI, do you think there should be automated tests with both 0.9.3 and 0.10.x?

Probably the best thing here is to do >=0.9 and then if we want to test legacy versions we can create a cron job for it when we get around to implementing #3052 (at least in my opinion, ideally the only legacy version we _really_ want to test at every push is `numpy`, otherwise we'll quickly end up with a giant matrix).

Luthaf:
The codecov failure is weird, but other than that this should be ready to go!

IAlibay:
@Luthaf Apologies for the delay. I thought it was probably easier to merge #3128 first so that any changes could then be updated to reflect those being done here (e.g. version of the docs being linked to). Could you resolve the conflicts and then we can aim for a merge?

Luthaf:
Rebased! Let's see what CI says.

Luthaf:
The python 3.6 runner fails with `OSError: libssl.so.1.0.0: cannot open shared object file: No such file or directory` when trying to load `libchemfiles.so`. Is there any difference in the environment between the 3.6 builders and the other ones? I'll investigate a bit more where the dependency on libssl comes from, since chemfiles itself is not using it.

IAlibay:
> The python 3.6 runner fails with `OSError: libssl.so.1.0.0: cannot open shared object file: No such file or directory` when trying to load `libchemfiles.so`. Is there any difference in the environment between the 3.6 builders and the other ones? I'll investigate a bit more where the dependency on libssl comes from, since chemfiles itself is not using it.

How strange. We've been having some occasional issues with gh actions and py3.6 lately. I have restarted CI to see if somehow that fixes it.

IAlibay:
It seems like a consistent error. Strangely enough it's not happening on the 3.6 runner with numpy 1.16.0.
I guess there might be some kind of dependency clash here that is causing chemfiles to fail?

Luthaf:
I think I found the origin of the issue here: libchemfiles links to libnetcdf, which links to libcurl. For a yet unknown reason, the conda build pulls in libssh2 from the biobuilds channel, which is linked by libcurl. libssh2 links to libopenssl.so.1.0.0, which is not installed (instead conda-forge gives us libopenssl.so.1.1.1). Given that the last update to libssh2 in biobuilds is [more than 3 years old](https://github.com/biobuilds/biobuilds/tree/master/libssh2/1.8.0),  I think it is best to stick with the conda-forge version.

It looks like the biobuilds channel is used only for clustalw, is this right? It looks like removing biobuilds from the main list of channels and instead using the channel as a one shot (`conda install -c biobuilds clustalw=2.1`) fixes the issue, and conda picks the newer llibssh2/openssl. I'll re-run the full CI with this configuration.

.github/workflows/gh-ci.yaml
```diff
@@ -15,165 +15,170 @@ defaults:
...
-              install_hole: false
-              codecov: false
-              numpy: numpy=1.16.0
-            - name: asv_check
-              os: ubuntu-latest
-              python-version: 3.7
-              run_type: FULL
-              install_hole: false
-              codecov: false
+      fail-fast: false
```
IAlibay, line 30: @Luthaf can you un-shift all these? It makes it impossible to see where the changes have been introduced.

Luthaf: sorry, auto-formatter =/ I'll undo these!

.github/workflows/gh-ci.yaml
```diff
@@ -15,165 +15,170 @@ defaults:
...
+          gfortran -v
+          echo "FC=gfortran" >> $GITHUB_ENV
+
+      - name: setup_miniconda
+        uses: conda-incubator/setup-miniconda@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+          auto-update-conda: true
+          channel-priority: flexible
+          channels: conda-forge
```
IAlibay, line 104: Just a thought on maybe a way to do this a bit more cleanly, if you reverse the channel order (or set it to `conda-forge, defaults, biobuilds`) and set the channel-priority to strict, does this fix the issue?

Luthaf: I'll look into this tomorrow =)

Luthaf: Just inverting the channels seems to work

Luthaf:
The failed build is due to an HTTP error while fetching conda package. Could you restart it?

Luthaf:
Everything looks ok now, thanks for the restart! I'm not sure what to do about the codecov patch coverage error: most of the "not covered" lines are related to codepath for chemfiles 0.9

.github/workflows/gh-ci.yaml
IAlibay: Sorry... can I ask you to re-undo the auto-formatter again? I realise we should have shifted everything by 4 originally, but at this point it's going to make git blame difficult :(

.github/workflows/gh-ci.yaml
```diff
@@ -15,165 +15,163 @@ defaults:
...
+          gfortran -v
+          echo "FC=gfortran" >> $GITHUB_ENV
+
+      - name: setup_miniconda
+        uses: conda-incubator/setup-miniconda@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+          auto-update-conda: true
+          channel-priority: flexible
+          channels: conda-forge, biobuilds
```
IAlibay, line 101: I'm glad that fixes it, it _should_ make maintaining the thing easier in the long run.

Probably something we need to look into in the long run (probably in that mamba PR I got open somewhere), is to see if having strict channel priority speeds up the setup time.

package/MDAnalysis/coordinates/chemfiles.py
```diff
@@ -161,8 +164,9 @@ def __init__(self, filename, chemfiles_format="", **kwargs):
 
         """
         if not check_chemfiles_version():
-            raise RuntimeError("Please install Chemfiles > {}"
-                               "".format(MIN_CHEMFILES_VERSION))
+            raise RuntimeError(
```
IAlibay, line 167: So this one we'll only ever trigger if we go out of our way to install chemfiles < 0.9 right? can we maybe use mock/patch to spoof a non-acceptable value here?

package/MDAnalysis/coordinates/chemfiles.py
```diff
@@ -343,15 +357,15 @@ def _write_next_frame(self, obj):
...
+            if hasattr(obj, "universe"):
                 # For AtomGroup and children (Residue, ResidueGroup, Segment)
                 ts_full = obj.universe.trajectory.ts
                 if ts_full.n_atoms == obj.atoms.n_atoms:
                     ts = ts_full
                 else:
                     # Only populate a time step with the selected atoms.
-                    ts = ts_full.copy_slice(atoms.indices)
-            elif hasattr(obj, 'trajectory'):
+                    ts = ts_full.copy_slice(obj.atoms.indices)
```
IAlibay, line 367: Can we add a test to cover this?

package/MDAnalysis/coordinates/chemfiles.py
```diff
@@ -375,7 +389,17 @@ def _timestep_to_chemfiles(self, ts):
...
             frame.velocities[:] = ts.velocities[:]
-        frame.cell = chemfiles.UnitCell(*ts.dimensions)
+
+        lengths = ts.dimensions[:3]
+        angles = ts.dimensions[3:]
+        if angles[0] == 0.0 and angles[1] == 0.0 and angles[2] == 0.0:
+            angles = [90, 90, 90]
+
+        if chemfiles.__version__.startswith("0.9"):
+            frame.cell = chemfiles.UnitCell(*lengths, *angles)
```
IAlibay, line 399: In this case, imho it's acceptable for now to just not cover it. Adding various versions down in CI is going to eventually become a pain.

Can you open an issue about covering this so we don't forget? I was working on a weekly cron CI job somewhere else, I can probably just add a workflow that installs older versions of chemfiles.

.github/workflows/gh-ci.yaml
IAlibay: As an extra thought, could you add a comment above this to clarify that the order here matters (with a reference to this PR)?

IAlibay (changes requested):
Thanks @Luthaf sorry about all the CI related queries 😅 couple more things

package/MDAnalysis/coordinates/chemfiles.py
Luthaf: I'm not sure how this code would be triggered from the high level API. I think this was copy/pasted from the XYZ writer, I can change the code to use the same one as the current version of the XYZ writer, and take inspiration from there.

.github/workflows/gh-ci.yaml
Luthaf: Really sorry about that. I'll check how to disable the formatter for mdanalysis code!

package/MDAnalysis/coordinates/chemfiles.py
IAlibay: It looks like a case where you pass an atomgroup which only represents a portion of your full system, but I'm not 100% sure.

Either way, I fully support any changes that a) improves line coverage, b) makes it simpler for us to work out what the code is doing 🚀 

If you think you prefer addressing it in a separate PR that's also an option.

package/MDAnalysis/coordinates/chemfiles.py
Luthaf: Test added with a mock value here: https://github.com/MDAnalysis/mdanalysis/pull/3126/files#diff-01fe53b03a7506a787f8d550d36cd7bb68074c1f1c3e3f373d76a7ab40ade459R43-R58

IAlibay: Coverage is still not working for this 😕 I'm not very versed in the mixing of unittest and pytest, could this be part of the issue?

Luthaf: I finally got some time to update the test to pytest, hope this fixes it!

IAlibay: 😕 this is still happening, I'll try to dig into what's going on a little bit later. Worst case we just merge as-is and add coverage back later.

testsuite/MDAnalysisTests/coordinates/test_chemfiles.py
```diff
@@ -23,22 +23,42 @@
...
+
+class TestChemfileVersion(object):
+    def test_version_check(self):
+        # Make sure the version check works as intended
+        import chemfiles
+
+        actual_version = chemfiles.__version__
+        chemfiles.__version__ = "0.8.3"
+        assert not check_chemfiles_version()
+
```
IAlibay, line 50: Ok I see what's going on here (and why coverage isn't improving), technically we could use `mock` but given that 2.0 is around the corner we can just go with a quick solution and implement a more elegant one later if necessary.

Something like this should cover the two missing sections:

```suggestion
        with pytest.raises(RuntimeError, match='Please install Chemfiles > 0.9'):
            ChemfilesReader('test.nc')

        with pytest.raises(RuntimeError, match='Please install Chemfiles > 0.9'):
            ChemfilesWriter('test.nc')

```

IAlibay (approved):
@Luthaf  I'll approve as-is, if you can add the suggested test that'd be great, if it proves more complicated than just that addition then let's just go with this and we can fix coverage some other time.

testsuite/MDAnalysisTests/coordinates/test_chemfiles.py
Luthaf: oh, right my bad, I was testing the wrong code =(. I'll integrate this today!

orbeckst: @Luthaf would you be able to make any changes by end of Friday? We're trying to get 2.0 out and will freeze develop on 2021-05-07.

Luthaf: Done, I hope the coverage will be good!

package/MDAnalysis/coordinates/chemfiles.py
```diff
@@ -375,7 +391,17 @@ def _timestep_to_chemfiles(self, ts):
         if ts.has_velocities:
             frame.add_velocities()
             frame.velocities[:] = ts.velocities[:]
-        frame.cell = chemfiles.UnitCell(*ts.dimensions)
+
+        lengths = ts.dimensions[:3]
+        angles = ts.dimensions[3:]
+        if angles[0] == 0.0 and angles[1] == 0.0 and angles[2] == 0.0:
```
IAlibay, line 397: @Luthaf I'm actually not sure what's being done here, do you have a quick TLDR you can provide?

Luthaf: This might be me being overly defensive. Does MDA guarantee angles to have valid values? If so this check can probably be removed.

Luthaf: Yes, chemfiles is fine with lengths of 0 but requires angles to be 90 in this case. I try to be generous in what code accept, so checking only for angles gives the maximal flexibility, but I'm also fine with checking if `ts.dimensions` is `[0., 0., 0., 0., 0., 0.]` and rejecting more wrong/missing cells.

Luthaf: So what's the decision here? Should I change this to check for either `[0., 0., 0., 0., 0., 0.]` or `None`?

richardjgowers:
I’m fixing this behaviour in general in another PR, so either is fine
because it’ll get changed later.

richardjgowers (approved):

PR #3130, opened 2021-02-21
Fixed issue -3129

### LINKED ISSUE #3129: abspaths slipping into SOURCES and breaking windows conda build
https://github.com/conda-forge/mdanalysis-feedstock/pull/28

^^ On the 1.0.1 builds Windows is failing because some of the sources in the sdist are given as absolute paths.

This is a good description:

https://github.com/pyFFTW/pyFFTW/pull/250/files

It looks like we started trying to use `abspath` to get RTD to work here:

https://github.com/MDAnalysis/mdanalysis/pull/2749

probably because of our slightly unusual repo structure

richardjgowers:
Fixes #3129

I think we made this change while trying to get RTD to work, but then didn't go all the way with that, is this correct @lilyminium ?  I.e. am I breaking the doc build by doing this?

@IAlibay this is targetting master as it's hopefully to fix the 1.0.2 release


lilyminium (approved):
@richardjgowers yes, I think the important `abspath` calls are in `conf.py` now. 

It looks like all the failing tests are due to more warnings than expected, possibly due to the `np.int` etc deprecation.

IAlibay:
@richardjgowers how do you want to proceed with this? Do we want to this do a 1.0.1 release that doesn't work on Windows or do you want to mark 1.0.1 as broken on pypi and we do a 1.0.2 release instead?

IAlibay:
> It looks like all the failing tests are due to more warnings than expected, possibly due to the `np.int` etc deprecation.

This is confusing, the numpy version hasn't changed (it's pinned at a max of 1.16) why is it suddenly throwing new warnings?

tylerjereddy:
> This is confusing, the numpy version hasn't changed (it's pinned at a max of 1.16) why is it suddenly throwing new warnings?

For Azure it isn't pinned in the dep install stage of CI. For gh-actions I likewise see NumPy `1.20.1` getting pulled in by conda in `install_deps` stage. Tricky business, but one could consider runtime version check on NumPy if you really want to reject about some threshold. I think build time is using `1.16.x` in both cases.

IAlibay:
I've updated the branch against current master. Assuming it returns fine, I suggest that we squash-merge as is. I'll try to test on Windows later on.

PR #3132, opened 2021-02-22
WIP: Issue 1738

### LINKED ISSUE #1738: consistently indicate absence of box information `Timestep.dimensions`
### Expected behaviour
When a trajectory does not contain a unit cell, then `u.trajectory.ts.dimensions` should have one value that always indicates the absence of box information, for example `None` or `[0, 0, 0, 0, 0, 0]`

### Actual behaviour
Readers decide for themselves how they handle missing box information. For example:
* XYZReader: `array([ 0., 0., 0., 0., 0., 0.], dtype=float32)`
* DCDReader: `array([  0.,   0.,   0.,  90.,  90.,  90.], dtype=float32)`

See discussion https://github.com/MDAnalysis/mdanalysis/pull/1733#issuecomment-351234725

``` python
import MDAnalysis as mda
from MDAnalysis.tests.datafiles import XYZ, PSF, DCD

xyz = mda.Universe(XYZ)
print(xyz.trajectory.ts.dimensions)

dcd = mda.Universe(PSF, DCD)
print(dcd.trajectory.ts.dimensions)
```

0.17.0-dev

### LINKED ISSUE #2210: Change Memory layout of positions from Fortran- to C-contiguous
Currently, all functions that operate (especially in-place) on atom positions obtain a copy of atom positions via `AtomGroup.positions`, regardless of whether positions are modified in- or out-of-place, or they are merely input data that is not altered. In some (if not many) cases, a tremendous speed-up could be achieved if the respective functions could obtain a *view* instead of a copy of an AtomGroup's positions. At the moment, this is not possible for two reasons:
1. `AtomGroup.positions` gets its data from `Universe.trajectory.ts.positions` via fancy indexing, and fancy indexing means copying.
2. More importantly, `Universe.trajectory.ts.positions` refers to `coordinates.base.Timestep._pos`, which is a Fortran-contiguous array. Therefore, potentially obtained views of that array are unsuitable for use with C-level functions.

Change `coordinates.base.Timestep.order` from `'F'` to `'C'` and provide a way of obtaining a (if possible, C-contiguous) view of an AtomGroup's positions.

Invent computers with two-dimensional main memory...

**Additional context**
Two candidates that come to my mind in terms of speed-up are `*Group.center*()` and `*Group.wrap()`
when the index array `*Group.atoms.ix` is a (contiguous) range.

A possible drawback of changing the positions/velocities/forces memory layout is that reading and (probably to a lesser extent) writing of DCD trajectories will be a bit slower since DCD stores positions etc. as Fortran-contiguous arrays.

If you have any additional concerns, please let me know!

### LINKED ISSUE #2698: Inconsistent handling of Universe dimensions
## Expected behavior ##

I expect an Universe created from a file with meaningless or unspecified cell dimensions to have a consistent null value for the dimensions attribute.

## Actual behavior ##

Loading the same cryo-EM structure (with meaningless cell dimensions) from MMTF and PDB files results in different values of the dimensions attribute.

``` python

In [2]: u1 = mda.Universe("2bbm.pdb.gz")
/Users/rmeli/Documents/git/software/mdanalysis/package/MDAnalysis/coordinates/PDB.py:432: UserWarning: 1 A^3 CRYST1 record, this is usually a placeholder. Unit cell dimensions will not be set.

In [3]: u1.dimensions
Out[3]: array([0., 0., 0., 0., 0., 0.], dtype=float32)

In [4]: u2 = mda.Universe("2BBM.mmtf.gz")

In [5]: u2.dimensions

In [6]: u2.dimensions is None
Out[6]: True
```

_Note:_ The test files are added in PR #2685 

- Which version are you using? 0.20.2-dev0'
- Which version of Python (`python -V`)? Python 3.7.6

richardjgowers:
Fixes #1738
Fixes #2698 
Fixes #2210

 - `ts.dimensions` gives a **view** of the box info
 - `atomgroup.dimensions` gives a **copy** of the box info
 - nobody gets to touch `ts._unitcell` any more (use `ts.dimensions`)
 - `ts.dimensions` can be `None` for no box info, was previously often all zeros, but not consistently.  This means blindly calling e.g. `distance_array(..., box=ag.dimensions)` will just work.

I've also removed a lot of classes.  Previously many formats defined their own `Timestep` object, then stored their box in a native format in `_unitcell`, then did their own conversion when `.dimensions` was called.  I think it's easier to maintain if a `Reader` does the conversion to `[x,y,z,alpha,beta,gamma]`.  This might be a bit contentious, so an alternative would be we can define one of two Timestep classes, one where the box is stored (i.e. `ts._unitcell`) as a 3x3 matrix.  This would mean 2 classes and maybe better roundtripping?

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


package/MDAnalysis/coordinates/TRJ.py
```diff
@@ -662,10 +659,13 @@ def _read_frame(self, frame):
             ts._forces[:] = (self.trjfile.variables['forces'][frame] *
                              self.scale_factors['forces'])
         if self.periodic:
-            ts._unitcell[:3] = (self.trjfile.variables['cell_lengths'][frame] *
-                                self.scale_factors['cell_lengths'])
-            ts._unitcell[3:] = (self.trjfile.variables['cell_angles'][frame] *
-                                self.scale_factors['cell_angles'])
+            unitcell = np.zeros(6)
```
tylerjereddy, line 662: maybe this should be obvious, but did you use this instead of `np.empty()` because of a preference for having `0` as a placeholder in the absence of the appropriate unit cell info?

It is such a small array that the performance difference is likely irrelevant of course.

richardjgowers: Yeah internally (`ts._unitcell`) we're still using zeros as the way to indicate no box

package/MDAnalysis/coordinates/GRO.py
```diff
@@ -223,15 +216,20 @@ def _read_first_frame(self):
         if len(unitcell) == 3:
             # special case: a b c --> (a 0 0) (b 0 0) (c 0 0)
             # see Timestep.dimensions() above for format (!)
-            self.ts._unitcell[:3] = unitcell
+            self.ts.dimensions = np.r_[unitcell, [90., 90., 90.]]
         elif len(unitcell) == 9:
-            self.ts._unitcell[:] = unitcell  # fill all
+            self.ts.dimensions = _gmx_to_dimensions(unitcell)
         else:  # or maybe raise an error for wrong format??
             warnings.warn("GRO unitcell has neither 3 nor 9 entries --- might be wrong.")
```
IAlibay, line 223: I know this is historical, but shouldn't we just throw an error if this happens? "not sure about this" doesn't fill one with much confidence.

package/MDAnalysis/coordinates/GRO.py
```diff
@@ -456,20 +454,22 @@ def write(self, obj):
                     ))
 
             # Footer: box dimensions
-            if np.allclose(ag.dimensions[3:], [90., 90., 90.]):
-                box = self.convert_pos_to_native(
-                    ag.dimensions[:3], inplace=False)
+            if (ag.dimensions is None or
+                np.allclose(ag.dimensions[3:], [90., 90., 90.])):
+                if ag.dimensions is None:
+                    box = np.zeros(3)
```
IAlibay, line 460: Is this the behaviour we want? Tagging @jbarnoud here.

package/MDAnalysis/coordinates/PDBQT.py
```diff
@@ -166,12 +166,13 @@ def _read_first_frame(self):
         self.ts = self._Timestep.from_coordinates(
             coords,
             **self._ts_kwargs)
-        self.ts._unitcell[:] = unitcell
+        self.ts.dimensions = unitcell
```
IAlibay, line 169: How closely does PDBQT follow the PDB standard? I.e. when reading a 1,1,1,90,90,90 do we need to assign the box dimension to None (probably not to be tackled here, just thinking about whether we need to fix this later).

package/MDAnalysis/coordinates/PDBQT.py
```diff
@@ -166,12 +166,13 @@ def _read_first_frame(self):
...
         self.ts.frame = 0  # 0-based frame number
         if self.convert_units:
             # in-place !
             self.convert_pos_from_native(self.ts._pos)
-            self.convert_pos_from_native(self.ts._unitcell[:3])
+            if not self.ts.dimensions is None:
```
IAlibay, line 174:
```suggestion
            if self.ts.dimensions is not None:
```
pep8

package/MDAnalysis/coordinates/PQR.py
```diff
@@ -145,12 +145,13 @@ def _read_first_frame(self):
         self.n_atoms = len(coords)
         self.ts = self._Timestep.from_coordinates(
             coords, **self._ts_kwargs)
-        self.ts._unitcell[:] = unitcell
+        self.ts.dimensions = unitcell
```
IAlibay, line 148: Am I misreading this or is `dimensions` always `None` here?

package/MDAnalysis/coordinates/base.py
```diff
@@ -425,6 +420,15 @@ def __eq__(self, other):
             if not (self.positions == other.positions).all():
                 return False
 
+        if self.dimensions is None:
+            if not other.dimensions is None:
```
IAlibay, line 424:
```suggestion
            if other.dimensions is not None:
```

package/MDAnalysis/coordinates/base.py
```diff
@@ -425,6 +420,15 @@ def __eq__(self, other):
             if not (self.positions == other.positions).all():
                 return False
 
+        if self.dimensions is None:
```
IAlibay, line 423: Can we add tests for this?

package/MDAnalysis/lib/util.py
```diff
@@ -2386,6 +2386,8 @@ def check_box(box):
          in :mod:`~MDAnalysis.lib.c_distances`.
        * Removed obsolete box types ``tri_box`` and ``tri_vecs_bad``.
     """
+    if box is None:
```
IAlibay, line 2389: Can we add a test here? (should be easy enough to trigger manually)

testsuite/MDAnalysisTests/coordinates/base.py
```diff
@@ -633,570 +642,6 @@ def test_write_not_changing_ts(self, ref, universe, tmpdir):
                 assert_timestep_almost_equal(copy_ts, universe.trajectory.ts)
 
 
-class BaseTimestepTest(object):
```
IAlibay, line 636: Just checking since I have yet to wake up properly - we are moving this because custom Timestep objects have now been removed right?

testsuite/MDAnalysisTests/coordinates/test_netcdf.py
```diff
@@ -200,9 +200,10 @@ def test_dt(self, u):
         assert_almost_equal(ref, u.trajectory.ts.dt, self.prec)
 
     def test_box(self, u):
-        ref = np.array([0., 0., 0., 0., 0., 0.], dtype=np.float32)
+        #ref = np.array([0., 0., 0., 0., 0., 0.], dtype=np.float32)
```
IAlibay, line 203: Do we need these commented out or can we just remove them?

IAlibay (changes requested):
For the most part this all seems sensible. Couple of comments (a few short test requests, and the rest is just me not having woken up properly yet -- sorry it's probably all silly questions).

@MDAnalysis/coredevs can we get all hands on reviewing this please? I can't vouch for all the parsers, so if we can at least pick up the ones we know we can make sure this is all fine quickly.

testsuite/MDAnalysisTests/coordinates/base.py
richardjgowers: Yeah we used to have many Timesteps and this was the base set of tests for them.  Now we've just got one Timestep so you don't need to generalise the tests like this, there's just a single testsuite for Timestep.

package/MDAnalysis/coordinates/DLPoly.py
```diff
@@ -36,18 +36,8 @@
...
-
-    @property
-    def dimensions(self):
-        return core.triclinic_box(*self._unitcell)
-
-    @dimensions.setter
-    def dimensions(self, new):
-        self._unitcell[:] = core.triclinic_vectors(new)
+# legacy reasons
+Timestep = base.Timestep
```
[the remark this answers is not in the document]
richardjgowers: It's just being defensive about the namespace.  Maybe someone for some reason was importing from here.  Probably not though.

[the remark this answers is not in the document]
IAlibay: It'd be nice to have a way out of having them in the future, that being said 2.0.0 is for breaking things right?

package/MDAnalysis/analysis/bat.py
```diff
@@ -372,7 +372,7 @@ def _single_frame(self):
         # The rotation axis is a normalized vector pointing from atom 0 to 1
         # It is described in two degrees of freedom
         # by the polar angle and azimuth
-        if (self._root.dimensions[:3] == 0).any():
+        if self._root.dimensions is None:
```
lilyminium, line 375: I think `(self._root.dimensions[:3] == 0).any()` asks if any of x, y, z are 0, not just if it's unassigned. Is this a pertinent difference @daveminh ?

lilyminium (changes requested):
I had a look at: 
* bat
* dcd
* dlpoly
* gro
* parmed

Will review more if I have time :)

package/MDAnalysis/coordinates/GRO.py
IAlibay: I see, in any case, no matter what we do here we need to warn users of what we are doing.

Personally I'm still in the "if there's no standardised way to deal with this, then crash out" (folks can always add their own dimensions at their own discretion). But I'm happy with whatever the consensus is.

mnmelo: See that at this point in the code we're just dealing with the writing of box coordinates when `ts.dimensions` is `None`. And as long as there's no tracking of input box validity, I don't think writing `ts.dimensions` that are `None` should ever be a problem to warn users about -- unless a given Writer has no canonical or sensible way to represent the no-box case (and the GROWriter does have).

For the record, while I don't oppose changing Readers to be more permissive with invalid boxes, and tracking input validity, I am also not for it. It may bring robustness, but do we really want to raise sloppy users that can't format their inputs right? Invalid boxes may even become mistakenly seen as a sanctioned extension to input formats. Stuff for other discussions.

package/MDAnalysis/analysis/bat.py
richardjgowers: `== 0) .any()` was the old idiom for unassigned.  This keeps the behaviour the same.

package/MDAnalysis/coordinates/DCD.py
```diff
@@ -246,7 +244,8 @@ def _frame_to_ts(self, frame, ts):
         ts.positions = frame.xyz
 
         if self.convert_units:
-            self.convert_pos_from_native(ts.dimensions[:3])
+            if ts.dimensions is not None:
```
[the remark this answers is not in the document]
richardjgowers: Could do, but yeah would change that functions signature a bit

package/MDAnalysis/coordinates/DCD.py
```diff
@@ -416,7 +415,10 @@ def _write_next_frame(self, ag):
                 errmsg = "Input obj is neither an AtomGroup or Universe"
                 raise TypeError(errmsg) from None
         xyz = ts.positions.copy()
-        dimensions = ts.dimensions.copy()
+        try:
+            dimensions = ts.dimensions.copy()
+        except AttributeError:
+            dimensions = np.zeros(6)
```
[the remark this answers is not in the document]
richardjgowers: This should match the old behaviour, so it's not a visible change.

package/MDAnalysis/coordinates/DLPoly.py
```diff
@@ -170,9 +158,11 @@ def _read_next_timestep(self, ts=None):
         if not line.startswith('timestep'):
             raise IOError
         if not self._imcon == 0:
-            ts._unitcell[0] = self._file.readline().split()
-            ts._unitcell[1] = self._file.readline().split()
-            ts._unitcell[2] = self._file.readline().split()
+            unitcell = np.zeros((3, 3))
+            unitcell[0][:] = self._file.readline().split()
```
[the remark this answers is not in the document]
richardjgowers: adds a stylish elan to the line.  But yeah nothing.

package/MDAnalysis/coordinates/DLPoly.py
```diff
@@ -170,9 +158,11 @@ def _read_next_timestep(self, ts=None):
         if not line.startswith('timestep'):
             raise IOError
         if not self._imcon == 0:
-            ts._unitcell[0] = self._file.readline().split()
-            ts._unitcell[1] = self._file.readline().split()
-            ts._unitcell[2] = self._file.readline().split()
+            unitcell = np.zeros((3, 3))
```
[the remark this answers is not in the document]
richardjgowers: Only if the file has dimensions (related to `imcon`)

package/MDAnalysis/coordinates/ParmEd.py
```diff
@@ -115,8 +115,6 @@ def _read_first_frame(self):
         if self.filename.box is not None:
```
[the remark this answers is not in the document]
richardjgowers: done

package/MDAnalysis/coordinates/memory.py
```diff
@@ -190,52 +190,43 @@
 import warnings
 
 from . import base
+from .base import Timestep
 
+# These methods all pass in an existing *view* onto a larger array
+def _replace_positions_array(ts, new):
+    """Replace the array of positions
 
-class Timestep(base.Timestep):
```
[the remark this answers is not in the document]
richardjgowers: From an OO approach, if the methods are only used in one place, for one purpose, it doesn't make sense to create a whole new object for that.  See also the maintenance of having multiple objects.

Longer term, when we want to play with the performance of these it's much better to only have to worry about one class.

richardjgowers:
@mnmelo wrt changing dimensions and only having some frames with values etc, that's the responsibility of the Reader, that's not really changed by this PR from what I can see.

richardjgowers:
It looks like there's a couple questions about zero boxes for GRO and DCD, but in terms of getting this PR merged I think I haven't changed any default behaviour so I think they're for a future issue.

package/MDAnalysis/analysis/bat.py
IAlibay: I'm not seeing any other match to that pattern in `develop`? I'd agree with @lilyminium that this is either intentional or it was a typo to begin with.

pinging @daveminh again.

package/MDAnalysis/coordinates/DCD.py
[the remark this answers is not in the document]
IAlibay: What happens here if we write a timestep with `dimensions` set to `None`? That should cause an `IndexError` on line 430 right?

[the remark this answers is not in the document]
IAlibay: I've raised issue #3303 to deal with this.

package/MDAnalysis/coordinates/DMS.py
```diff
@@ -124,10 +105,12 @@ def dict_factory(cursor, row):
...
         self.ts.frame = 0  # 0-based frame number
 
-        self.ts._unitcell = unitcell
+        self.ts.dimensions = triclinic_box(unitcell['x'], unitcell['y'], unitcell['z'])
+
         if self.convert_units:
             self.convert_pos_from_native(self.ts._pos)  # in-place !
-            self.convert_pos_from_native(self.ts._unitcell)  # in-place ! (all are lengths)
+            if self.ts.dimensions is not None:
+                self.convert_pos_from_native(self.ts.dimensions[:3])  # in-place !
```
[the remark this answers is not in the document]
IAlibay: Raised issue #3304 for this, we can deal with it later.

package/MDAnalysis/coordinates/TRZ.py
```diff
@@ -591,7 +571,11 @@ def _write_next_frame(self, obj):
                           "".format(", ".join(faked_attrs)))
 
         # Convert other stuff into our format
-        unitcell = triclinic_vectors(ts.dimensions).reshape(9)
+        if ts.dimensions is not None:
+            unitcell = triclinic_vectors(ts.dimensions).reshape(9)
+        else:
+            unitcell = np.zeros(9, dtype=np.float32)
```
IAlibay, line 577: This behaviour should be at least documented, maybe warned too (I can't find any information about the TRZ file format..). I'll raise an issue...

package/MDAnalysis/coordinates/base.py
IAlibay: @richardjgowers if you don't want to add tests for this please open an issue for it.

package/MDAnalysis/coordinates/base.py
```diff
@@ -840,15 +846,21 @@ def triclinic_dimensions(self):
...
+            return core.triclinic_vectors(self.dimensions)
 
     @triclinic_dimensions.setter
     def triclinic_dimensions(self, new):
         """Set the unitcell for this Timestep as defined by triclinic vectors
 
         .. versionadded:: 0.11.0
         """
-        self.dimensions = core.triclinic_box(*new)
+        if new is None:
```
IAlibay, line 860: If you don't want to add a test for this, please raise an issue for it.

testsuite/MDAnalysisTests/coordinates/test_mol2.py
```diff
@@ -214,4 +214,7 @@ def test_mol2_universe_write(tmpdir):
         u2 = mda.Universe(outfile)
 
         assert_almost_equal(u.atoms.positions, u2.atoms.positions)
-        assert_almost_equal(u.dimensions, u2.dimensions)
+        if u.dimensions is None:
+            assert u2.dimensions is None
+        else:
+            assert_almost_equal(u.dimensions, u2.dimensions)
```
IAlibay, line 220: Is this ever triggered? From what I understand `dimensions` is always `None` right? If we're not sure, can we raise an issue for this please?

IAlibay (changes requested):
I've already done so for a few cases, but can you just raise issues for stuff that can't be dealt with here @richardjgowers ?

Ideally the smaller changes (i.e. `from .base import Timestep`) needs to be dealt with here because otherwise we'll lose track of them.

mnmelo:
@richardjgowers, I totally agree. That was my point (that I perhaps didn't express well): if we ever decide to act on invalid boxes -- something definitely for another PR, not this one -- that's a burden to fall on the Reader.

As far as I'm concerned, this is good to go regarding that point.

package/MDAnalysis/analysis/bat.py
IAlibay: raised #3326

package/MDAnalysis/coordinates/PQR.py
```diff
@@ -145,12 +145,13 @@ def _read_first_frame(self):
...
         self.ts.frame = 0  # 0-based frame number
         if self.convert_units:
             # in-place !
             self.convert_pos_from_native(self.ts._pos)
-            self.convert_pos_from_native(self.ts._unitcell[:3])
+            if not self.ts.dimensions is None:
+                self.convert_pos_from_native(self.ts.dimensions[:3])
```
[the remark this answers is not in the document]
IAlibay: See #3327

package/MDAnalysis/coordinates/TRZ.py
IAlibay: See #3328

IAlibay (approved):
I've moved all the extra bits to new issues - @lilyminium can you approve?

lilyminium (approved):

IAlibay:
Thanks @richardjgowers - squash merging.

PR #3134, opened 2021-03-01
Fixes RMSD weights_groupselections docstring

### LINKED ISSUE #2796: Inaccurate docstring in RMSD
## Expected behavior ##

If I pass ``weights_groupselections=False`` or just leave it out (as it is the default), then it uses the weights from ``weights``.

https://github.com/MDAnalysis/mdanalysis/blob/31929685389638232b8f22a2ee9cc63bd659f8d5/package/MDAnalysis/analysis/rms.py#L382-L386

## Actual behavior ##

It just sets it all to ``None``.

```python
   ...: from MDAnalysis.tests.datafiles import PSF, DCD, CRD
   ...: from MDAnalysis.analysis import rms

In [2]: u = mda.Universe(PSF, DCD)  # closed AdK (PDB ID: 1AKE)

In [3]: bb = u.select_atoms('backbone')

In [4]: sc = u.select_atoms('not backbone')
In [5]: r = rms.RMSD(u, u, select='backbone', groupselections=['not backbone'],
   ...:    ...: weights=bb.resids, weights_groupselections=False).run()

In [6]: r.weights_groupselections
Out[6]: [None]
```

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) 1.0.0
- Which operating system? MacOS Mavericks

z3y50n:
Fixes #2796 

------------

 - Updated the docstring to describe actual behavior of `weights_groupselections=False`

PR Checklist
------------
- [ ] Tests?

 - [x] Docs?
 - [ ] CHANGELOG updated?


package/MDAnalysis/analysis/rms.py
```diff
@@ -380,7 +380,10 @@ def __init__(self, atomgroup, reference=None, select='all',
              corresponding atom in `select`, and assumes ``None`` for `groupselections`.
 
         weights_groupselections : False or list of {"mass", ``None`` or array_like} (optional)
-             1. ``False`` will apply imposed weights to `groupselections` from ``weights`` option.
+             1. ``False`` will apply imposed weights to `groupselections` from
+             ``weights``option if `weights` is iterable. Otherwise will assume
```
lilyminium, line 384: I think you might have this the wrong way around!

https://github.com/MDAnalysis/mdanalysis/blob/ff7ffa10901e2df2be12c3d3dd78e4e0a262e90e/package/MDAnalysis/analysis/rms.py#L573-L577

lilyminium (changes requested):
Thank you for helping with this docstring @z3y50n! I think you may have misread the code -- however, it is a good idea to mention the ``'mass'`` option.

package/MDAnalysis/analysis/rms.py
```diff
@@ -380,7 +380,10 @@ def __init__(self, atomgroup, reference=None, select='all',
              corresponding atom in `select`, and assumes ``None`` for `groupselections`.
 
         weights_groupselections : False or list of {"mass", ``None`` or array_like} (optional)
-             1. ``False`` will apply imposed weights to `groupselections` from ``weights`` option.
+             1. ``False`` will apply imposed weights to `groupselections` from
+             ``weights`` option if `weights` is not iterable. Otherwise will assume
```
lilyminium, line 384: Thanks for making that change, @z3y50n. Could you actually make this a little more detailed, though -- it's not intended to be just any iterable, but values of None or "mass". The current message could imply that users can pass in any old value.

lilyminium: Also, I think this is your first pull request with us -- please add yourself to the AUTHORS and CHANGELOG :-)

lilyminium (changes requested):

lilyminium:
Hi @z3y50n, are you still keen to work on this? :)

z3y50n:
Yes I am sorry I delayed that much I was busy! I plan to fix this of course sorry for taking much time

package/MDAnalysis/analysis/rms.py
z3y50n: So from what I understand it should say something like `if weights is not of the appropriate value`? Because in the [code](https://github.com/z3y50n/mdanalysis/blob/4ea0b6f244a7ac572d9fbf1e208ffbf27734a1da/package/MDAnalysis/analysis/rms.py#L577) it just checks if `weights` is `iterable`

lilyminium: > Because in the code it just checks if weights is iterable

@z3y50n that's true, but consider what kind of non-iterable weights are valid for the code that [executes immediately after](https://github.com/z3y50n/mdanalysis/blob/4ea0b6f244a7ac572d9fbf1e208ffbf27734a1da/package/MDAnalysis/analysis/rms.py#L582-L589). Does it make sense to assign an array of weights that are all the same float?

IAlibay:
@z3y50n just a reminder that the deadline for applications is April 13th. If you are still interested in applying for GSoC we should aim to get this merged as soon as possible.

Feel free to also share your draft application on the mailing list too and we'll try to give feedback where possible.

package/MDAnalysis/analysis/rms.py
```diff
@@ -380,7 +380,10 @@ def __init__(self, atomgroup, reference=None, select='all',
              corresponding atom in `select`, and assumes ``None`` for `groupselections`.
 
         weights_groupselections : False or list of {"mass", ``None`` or array_like} (optional)
-             1. ``False`` will apply imposed weights to `groupselections` from ``weights`` option.
+             1. ``False`` will apply imposed weights to `groupselections` from
+             ``weights`` option if ``weights`` is either ``"mass"`` or ``None``. 
+             Otherwise will assume a list of length equal to length of 
+             `groupselections` filled with ``None`` values.
```
lilyminium, line 386: This is surprising, could you show me where it does that? I'm not convinced this sentence is needed.

lilyminium (changes requested):

package/MDAnalysis/analysis/rms.py
z3y50n: I guess [this](https://github.com/z3y50n/mdanalysis/blob/0546385e68da94295e430a9e2c2fcc5ebc681ad8/package/MDAnalysis/analysis/rms.py#L579-L580) is the part. The otherwise part is referring to the `self.weights` not being "mass" or "None" (Basically when it is any iterable). Also this is the behavior you get from issue #2796.

lilyminium: I see, thanks!

lilyminium:
Thank you very much @z3y50n!

PR #3135, opened 2021-03-02
Implements cache validation at the Universe level

### LINKED ISSUE #2376: why is (cached) fragments slow?

mnmelo:
Objects keep their own caches, but those can now be set to be validated against a centralized registry under object.universe. This simplifies the centralized invalidation of object caches.

This PR implements a mix of the ideas floated in #2376, #3005 and in the [mailing list discussion](https://groups.google.com/g/mdnalysis-devel/c/Sj8AkV9P1No/m/gVETgr1dAwAJ): caches are still stored under an object's own `_cache` but now, if asked to, the cache retrieval can check `universe._cache['_valid'][key]` for cache validity. This solution, compared to storing the actual cache under `universe`, prevents garbage collection from getting blocked when an object references itself in a cache.

Applied to fragment caching and added asv benchmark (`FragmentCaching.time_find_cached_fragments`). Already large speedup in fragment accession (Fixes #2376):
```
=============================== ============ =============
         universe_type           no caching     caching
------------------------------- ------------ -------------
 large_fragment_small_solvents   275±0.9ms    1.56±0.01μs
         large_fragment           167±9ms     1.56±0.01μs
         polymer_chains          47.9±0.2ms   1.58±0.01μs
=============================== ============ =============
```
Tests include cache invalidation when bonds are added/removed.

Also added asv's benchmarks/results subdir to `.gitignore`

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/lib/util.py
+++ b/package/MDAnalysis/lib/util.py
@@ -202,6 +202,7 @@ import warnings
 import functools
 from functools import wraps
 import textwrap
+import weakref
 
 import mmtf
 import numpy as np
@@ -1496,10 +1497,18 @@ def conv_float(s):
         return s
 
 
-def cached(key):
+# A dummy, empty, cheaply-hashable object class to use with weakref caching.
+# (class object doesn't allow weakrefs to its instances, but user-defined
+#  classes do)
+class _CacheKey:
+    pass
+
+
+def cached(key, universe_validation=False):
     """Cache a property within a class.
 
-    Requires the Class to have a cache dict called ``_cache``.
+    Requires the Class to have a cache dict :attr:`_cache` and, with
+    `universe_validation`, a :attr:`universe` with a cache dict :attr:`_cache`.
 
     Example
     -------
@@ -1513,23 +1522,54 @@ def cached(key):
            @property
            @cached('keyname')
            def size(self):
-               # This code gets ran only if the lookup of keyname fails
-               # After this code has been ran once, the result is stored in
+               # This code gets run only if the lookup of keyname fails
+               # After this code has been run once, the result is stored in
                # _cache with the key: 'keyname'
-               size = 10.0
+               return 10.0
+
+           @property
+           @cached('keyname', universe_validation=True)
+           def othersize(self):
+               # This code gets run only if the lookup
+               # id(self) is not in the validation set under
+               # self.universe._cache['_valid']['keyname']
+               # After this code has been run once, id(self) is added to that
+               # set. The validation set can be centrally invalidated at the
+               # universe level (say, if a topology change invalidates specific
+               # caches).
+               return 20.0
 
 
     .. versionadded:: 0.9.0
 
+    .. versionchanged::2.0.0
+        Added the `universe_validation` keyword.
     """
 
     def cached_lookup(func):
         @wraps(func)
         def wrapper(self, *args, **kwargs):
             try:
+                if universe_validation:  # Universe-level cache validation
+                    u_cache = self.universe._cache.setdefault('_valid', dict())
+                    # A WeakSet is used so that keys from out-of-scope/deleted
+                    # objects don't clutter it.
+                    valid_caches = u_cache.setdefault(key, weakref.WeakSet())
+                    try:
+                        if self._cache_key not in valid_caches:
+                            raise KeyError
+                    except AttributeError:  # No _cache_key yet
+                        # Must create a reference key for the validation set.
+                        # self could be used itself as a weakref but set()
+                        # requires hashing it, which can be slow for AGs. Using
+                        # id(self) fails because ints can't be weak-referenced.
+                        self._cache_key = _CacheKey()
+                        raise KeyError
                 return self._cache[key]
             except KeyError:
                 self._cache[key] = ret = func(self, *args, **kwargs)
+                if universe_validation:
+                    valid_caches.add(self._cache_key)
                 return ret
 
         return wrapper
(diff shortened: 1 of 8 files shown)


richardjgowers:
@mnmelo if I've read this correctly.. caches are still stored on the AtomGroup, but there's also a check against `u._caches['_valid']` to see if the cache on a Group is still trustworthy?

WRT the CacheKey class, is there a reason we can't use `id(ag)` to hash an AtomGroup?

Would it be possible to instead store the cache on Universe, so there'd be something like `u._caches['fragments'][id(ag)]` for a given `ag`?  Then there wouldn't have to be any weakset or extra classes.  This technically "leaks" integers into `u._caches` but I can't see this becoming a problem.

mnmelo:
Yup, it is mostly as you understood (but see the 2nd paragraph below).

I would have preferred to use `id(self)`, as you suggest, but weakrefs don't work for ints. The validity registry _could_ be made a regular set, using `id(self)` for hashing, but would leak ints, as you predict. I tried to avoid that leaking, not only for memory, but to keep the validity registry set small so that lookups are fast. However, the [set lookup complexity](https://stackoverflow.com/questions/7351459/time-complexity-of-python-set-operations) seems to be O(1), so I may have prematurely optimized. If you're ok with int leakages (sounds wrong, in principle), then I can simplify this to a regular, non-weak, set that uses `id(self)`.

As to storing the cache itself under universe, see that that not only leaks ints _but also AtomGroups and suchlike_, which is something heavier (imagine a `select_atoms` being repeated every frame for a couple thousand frames). Hence my initial strategy of weakrefs to hold the caches under Universe. And even with weakrefs, I ran into possible garbage collection blocks when caching under universe and the ag itself is part of its own cache (don't know if it's a possibility, but prefer to future-proof it now).

Storing the cache with the object solves garbage collection locks, allowing always for a clean collection of the ag. And using the validity registry still allows for full flexibility in cache invalidation.

mnmelo:
Also see that if what you want is to remove the ugly, dummy `class _CacheKey` it can be deployed in-line when first needed on an ag:
```
self._cache_key = type('_CacheKey', (), dict())()
```
although I suspect this is less performant when used repeatedly.

richardjgowers (approved):
@mnmelo yes you're right.  I'd missed that the *values* are also AtomGroups, so if we store everything on Universe it's annoying to get things to garbage collect.

The `_CacheKey` confused me for a bit until I realised it's a bit like a self-destructing int

package/MDAnalysis/lib/util.py
```diff
@@ -1513,23 +1522,54 @@ class A(object):
...
         @wraps(func)
         def wrapper(self, *args, **kwargs):
             try:
+                if universe_validation:  # Universe-level cache validation
+                    u_cache = self.universe._cache.setdefault('_valid', dict())
+                    # A WeakSet is used so that keys from out-of-scope/deleted
+                    # objects don't clutter it.
+                    valid_caches = u_cache.setdefault(key, weakref.WeakSet())
+                    try:
+                        if self._cache_key not in valid_caches:
```
richardjgowers, line 1559: As a comment, one day it might be worthwhile to put this `self._cache_key` setting behaviour into `Group.__hash__` so that every AtomGroup natively hashes quickly/cleanly/weakly.

mnmelo: Great idea; that seems super clean! I can implement it right away if you're ok with having both functionalities added in the same PR.

richardjgowers: I went over and checked and we already have a `__hash__` defined, and it looks like it's actually hashing `ix` which isn't going to be fast, but technically necessary for two equivalent AGs to hash identically.  For this lookup I'd rather we quickly hash the AG and maybe have duplicate caches

mnmelo: Ah, yea. I hadn't checked `__hash__` but I had initially tried to use the ag itself as the key, and got the same slow performance (I just assumed the entire object was being hashed instead of just `_ix`, which probably boils down to a similar time penalty).

PR #3138, opened 2021-03-05
WIP/discussion optimisation of distance header and compiler flags

hmacdope:
Hi all.
Something we discovered over in distopia is that the use of `size_t` based loops along with the compiler flag `march=native, mtune=native` can give massive performance increases. I have done a little bit of refactoring in to this effect to test performance and see benchmark improvements (on select benchmarks) between 5-40% (on my linux box). Thought I would bring this up to be discussed.

The downside of using `march=native` and `mtune=native` is that you produce specialised assembly for the machine you build on, meaning you have to build and run on the same machine. This makes it unsuitable for a downloaded binary.  We may be able to control this with various options. 

Please let me know what you think. Other people benchmarking on their various machines would be greatly appreciated if possible.

### DIFF
--- a/benchmarks/asv.conf.json
+++ b/benchmarks/asv.conf.json
@@ -21,6 +21,7 @@
 
     // The Pythons you'd like to test against.  If not provided, defaults
     // to the current version of Python used to run `asv`.
+    //"pythons": ["3.6"],
     "pythons": ["2.7"],
     // The matrix of dependencies to test.  Each key is the name of a
     // package (in PyPI) and the values are version numbers.  An empty
--- a/benchmarks/benchmarks/analysis/distances.py
+++ b/benchmarks/benchmarks/analysis/distances.py
@@ -15,7 +15,7 @@ class DistancesBench(object):
     """Benchmarks for MDAnalysis.analysis.distances
     functions.
     """
-
+    timeout = 200
     # TODO: eventually we should include box / pbc
     # unit cell information in the benchmarks
     params = (10, 100, 1000, 10000)
--- a/package/MDAnalysis/lib/include/calc_distances.h
+++ b/package/MDAnalysis/lib/include/calc_distances.h
@@ -33,9 +33,8 @@ typedef float coordinate[3];
 
 static void minimum_image(double* x, float* box, float* inverse_box)
 {
-  int i;
   double s;
-  for (i=0; i<3; i++) {
+  for (size_t i=0; i<3; i++) {
     if (box[i] > FLT_EPSILON) {
       s = inverse_box[i] * x[i];
       x[i] = box[i] * (s - round(s));
@@ -107,7 +106,8 @@ static void _ortho_pbc(coordinate* coords, int numcoords, float* box)
         return;
     }
 
-    int i, j, s;
+    int s;
+    size_t i, j;
     float crd;
     // inverse box for multi-box shifts:
     const double inverse_box[3] = {1.0 / (double) box[0], \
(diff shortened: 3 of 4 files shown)


IAlibay:
So throwing in my 2 cents, I'd be rather against a switch to march=native/mtune=native if it breaks our ability to run across multiple platforms on a single compilation.

Just on a "will this break my day to day work"; that would be disastrous given that I work on heterogeneous commodity clusters, especially ones that are about to buy AMD chips (with an existing pool of Intel chips that have non uniform AVX* capabilities).

It's also not infrequent that folks will have clusters where the headnode differs from the compute nodes (an example of this was the recently defunct ARCHER). Given that compute nodes will tend to not have external network access, it's not uncommon for folks to install on the headnode, which may prove problematic?

hmacdope:
@IAlibay I agree completely that we should never break cross platform compatibility. It's more a question of whether we can/should activate this in situations where it is appropriate. 😄

IAlibay:
I guess it's down to what you classify as an appropriate case for enabling it. If provided as an optional thing that "experts" can turn on under specific "this will break in these cases" instructions, I have no problems against that.

jbarnoud:
I agree with @IAlibay here. As an advanced user, I'll be happy to have an option that makes things a bit faster. However, the main target must be simplicity for the users.

hmacdope:
Sorry for spam but I would be keen to get MDA thoughts from @richardjgowers @orbeckst.

Also @tylerjereddy may know how this is handled when c/c++ is compiled from a scipy/numpy POV?

richardjgowers:
We could keep this module very simple then have an optional import of
distopia monkey patch the mda internals.

orbeckst:
If you want to enable native with an option in setup.cfg for expert users then I don't have a problem as long as
- default keeps running on multiple archs
- default is not slower than before
- docs that make clear what the implications of performance tuning are

I don't understand https://github.com/MDAnalysis/mdanalysis/pull/3138#issuecomment-791938219 and how this immediately relates to distopia. Wasn't the idea to make distopia a separate package that gets imported?

richardjgowers:
@orbeckst we could make distopia do some monkey-patching (like seaborn used to) so that `import distopia` changes what `mda.lib.distances.distance_array` is.  Then we can have an optional import that enables aggressive options, even inside things which think they are calling mda distance libraries (like hbond_analysis).

orbeckst:
Is there a reason to not make distopia a required dependency instead of monkey-patching?

tylerjereddy:
Closely related to the suggestions above about having a "portable default" are the build flags/policies adopted by NumPy for NEP38 and described here in quite some detail: https://numpy.org/devdocs/reference/simd/simd-optimizations.html (I see discussion of the exact "native" stuff being discussed here too).

More loosely related, but SciPy is seriously considering migrating to the `meson` build system because of all the chaos with setuptools/distutils, etc. I tried advocating for CMake since many of us use it at work, but looks like another build system to learn may be coming to the ecosystem..

hmacdope:
Thanks so much @tylerjereddy. I didn't realise the numpy system used so much dispatch, but there you go.
We can let this sit for now, but just flagging that using distopia these questions will return. :)

hmacdope:
I should also say that I think we should try and match numpy supported CPU/SIMD features and policies as closely as possible, even if we want to do something different with them.

hmacdope:
For clarity the following code snippet in `setup.py` accomplishes the same thing we are discussing here. 

```  python 
   # allow using architecture specific instructions. This allows people to
    # build optimized versions of MDAnalysis.
    arch = config.get('march', default=False)
    if arch:
        extra_compile_args.append('-march={}'.format(arch))
```

Perhaps this should be more public or documented?

hmacdope:
This can probably be closed with the introduction of custom compiler arguments in #3429, as you can now use the `march` family of flags through that mechanism.

PR #3139, opened 2021-03-06
unpin for py3.6 and above

IAlibay:
As discussed with @richardjgowers it probably would be best if we had a 1.0.2 release that didn't pin the max versions for 3.6+ to ~ 4 numpy major releases behind.

Also completes fix for #3103

 - Optional pins in setup.py (let's see how CI reacts to this...)
 - Backport numpy warning fixes (ragged arrays and type aliases)

<s>To do:
 - backport numpy 1.20 fixes in develop
 - anything else that CI is going to complain about</s>

PR Checklist
------------
 - [x] Tests?
 - Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/lib/nsgrid.pyx
```diff
@@ -672,7 +672,7 @@ cdef class _NSGrid(object):
 
         cdef ns_int i, cellindex = -1
         cdef ns_int ncoords = coords.shape[0]
-        cdef ns_int[:] beadcounts = np.empty(self.size, dtype=np.int)
+        cdef ns_int[:] beadcounts = np.empty(self.size, dtype=int)
```
IAlibay, line 675: the nsgrid tests still fail because I've not re-generated the c files, but it's not worth doing this until #3133 is fixed

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -536,11 +536,11 @@ def count_by_type(self):
         acceptor atoms in a hydrogen bond.
         """
 
-        d = self.u.atoms[self.hbonds[:, 1].astype(np.int)]
-        a = self.u.atoms[self.hbonds[:, 3].astype(np.int)]
+        d = self.u.atoms[self.hbonds[:, 1].astype(int)]
```
richardjgowers, line 539: Looking at these (and in general when an integer is made for indexing something), I think we want `np.intp` here

IAlibay: done (I think this was the only case it's a thing) - note it does mean that the change here diverges from `develop`, do we want to "reverse backport" it?

richardjgowers (approved):

IAlibay:
thanks @richardjgowers, I was testing 2.7 and got a notice that matplotlib does not like < 3.7 for 3.4+ (I guess they released rc1 on pypi recently).

What do we want to do about this @MDAnalysis/coredevs ? We can pin <3.4 for all version in setup.py but that kinda goes against the idea of the unpinning done in this PR.

tylerjereddy:
Pinning the max version of `pytest` used in our CI for `master` branch backporting seems sensible to me given our limited bandwidth and it is just a maintenance branch for occasional backports at this point, if I understand correctly.

I'm not sure we ever fully understood the issue--I think we were just happy to find a workaround for the strange test failure.

I'm looking a little more closely at `test_PDB_metals()` locally and two things stand out to me:
- this test is the **only** usage of the `pytest.approx` construct in our entire testsuite, and given that is only comparing floating point numbers it definitely seems worth trying the more common `assert_allclose()` instead, though perhaps just for `develop` branch unpinning at this point
- there's also a `@pytest.mark.filterwarnings("error")` decorator on this particular test, without an explanatory comment (bit of a stretch but Python warnings handling is not always threadsafe)--also the **only** usage of this construct in our entire testsuite

IAlibay:
Thanks for looking at this @tylerjereddy, I agree it's probably not worth the effort for the current `master` (although we should fix this before 2.0). I'll go ahead with the merge on this one.

PR #3152, opened 2021-03-13
Add kwarg to index TPRParser from 0 or 1

### LINKED ISSUE #2364: TPRParser should not index from 0
**Expected behavior**

My expected behaviour if is that if I select a residue in `gmx make_ndx` by residue number 1, and I select a residue with `u.select_atoms('resid 1')`, they give me the same residue.

**Actual behavior**

I would get different residues, because GROMACS indexes from 1, but TPRParser indexes from 0. TPR files may index from 0, but they are typically created from human-readable .top or .itp files that are indexed from 1.

```python
>>> from MDAnalysis.tests.datafiles import TPR
>>> tpr = mda.Universe(TPR)
>>> tpr.atoms.ids
array([    0,     1,     2, ..., 47678, 47679, 47680])
>>> tpr.atoms.resids
array([    0,     0,     0, ..., 11299, 11300, 11301])
```

- Which operating system? MacOS Mojave

lilyminium:
Fixes #2364 

 - Added keyword argument to index TPRParser from 0 or 1
 - Set the default to index from 1, changing the default behaviour

~This is easily undoable. I read the discussion in #2364 as supporting both including a kwarg and a warning. If that is not consensus, please let me know. I'll update CHANGELOG once we have this nailed down :-)~

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -133,8 +133,8 @@ Enhancements
     explicit in the topology (Issue #2468, PR #2775)
   
 Changes
-  * TPRParser now loads TPR files with `tpr_resid_from_one=True`
-    by default (PR #3152)
+  * TPRParser now loads TPR files with `tpr_resid_from_one=True` by default, 
+    which starts TPR resid indexing from 1 (instead of 0 as in 1.x) (Issue #2364, PR #3152)
   * Introduces encore specific C compiler arguments to allow for lowering of
     optimisations on non-x86 platforms (Issue #1389, PR #3149)
   * Continuous integration uses mamba rather than conda to install the
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -133,6 +133,8 @@ Enhancements
     explicit in the topology (Issue #2468, PR #2775)
   
 Changes
+  * TPRParser now loads TPR files with `tpr_resid_from_one=True`
+    by default (PR #3152)
   * Introduces encore specific C compiler arguments to allow for lowering of
     optimisations on non-x86 platforms (Issue #1389, PR #3149)
   * Continuous integration uses mamba rather than conda to install the
--- a/package/MDAnalysis/topology/tpr/utils.py
+++ b/package/MDAnalysis/topology/tpr/utils.py
@@ -286,7 +286,7 @@ def extract_box_info(data, fver):
     return obj.Box(box, box_rel, box_v)
 
 
-def do_mtop(data, fver, tpr_resid_from_one=False):
+def do_mtop(data, fver, tpr_resid_from_one=True):
     # mtop: the topology of the whole system
     symtab = do_symtab(data)
     do_symstr(data, symtab)  # system_name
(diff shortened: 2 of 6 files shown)


jbarnoud (approved):
Looks good to me. I am not sure the deprecation message is very clear, though. I would phrase the message to say that from version 2.0, resid will start at 1 rather than 0.

jbarnoud (changes requested):
I just realized it is missing an entry in the changelog and a "versionchanged".

lilyminium:
Thanks @jbarnoud, I've updated CHANGELOG now. I also realised this PR is to develop, so I removed the deprecation warning and changed the default. #3153 is the "first" addition of this to master with the deprecation warning and the status quo default of ``tpr_resid_from_one=False``.

package/MDAnalysis/topology/tpr/utils.py
```diff
@@ -46,6 +46,8 @@
 
 The module also contains the :func:`do_inputrec` to read the TPR header with.
 """
+import warnings
+
```
orbeckst, line 50: I don't think `warnings` is needed for the PR against develop. In 2.0.0 we will index TPR resids from 1 by default.

package/CHANGELOG
```diff
@@ -128,6 +128,8 @@ Enhancements
     explicit in the topology (Issue #2468, PR #2775)
   
 Changes
+  * TPRParser now loads TPR files with `tpr_resid_from_one=True`
+    by default (PR #3152)
```
orbeckst, line 132:
```suggestion
  * TPRParser now loads TPR files with `tpr_resid_from_one=True` by default, 
    which starts TPR resid indexing from 1 (instead of 0 as in 1.x) (Issue #2364, PR #3152)
```

package/MDAnalysis/topology/tpr/utils.py
```diff
@@ -284,7 +286,7 @@ def extract_box_info(data, fver):
     return obj.Box(box, box_rel, box_v)
 
 
-def do_mtop(data, fver):
+def do_mtop(data, fver, tpr_resid_from_one=True):
```
orbeckst, line 289: I feel the default should remain `False` because that is what the TPR actually contains. IMHO the TPR functions themselves could/should stay as faithful to the original format as possible. Assume someone wrote their own TPR parser with our functions. With the new default you break their code needlessly, with the old defaults nothing changes and our own parser just makes the well-documented choice to index from 1.

orbeckst: I can be convinced otherwise (ie to let default `True` stand) but wanted to discuss this point first before approving.

orbeckst (changes requested):
Thank you! I was first confused by the PR description but then saw that this is one against develop.

I have minor improvements and one bigger discussion point regarding the default value inside the low-level TPR reader code.

package/MDAnalysis/topology/tpr/utils.py
lilyminium: Others may want to chime in (maybe the most knowledgeable, @jbarnoud) but I agree with your argument.

orbeckst: I am going to approve because it looks sensible from my point but let @jbarnoud also say something before merging.

orbeckst (approved):

orbeckst:
@lilyminium the issue description is not correct, is it?

> Set the default to index from 0, i.e. status quo

*This* PR (against develop) *does* change the default indexing to 1-based.

orbeckst:
Note that anything referring to 1.0.2 should become 1.1.0 (see comment on #3153 ).

orbeckst:
@jbarnoud any last comments? Otherwise I'll merge tomorrow.

jbarnoud (approved):

package/MDAnalysis/topology/tpr/utils.py
jbarnoud: TPR indeed index from 0, but it is a bit moot because you do not read a TPR directly. Instead, you end up reading your gro and ITP files to make sense of your TPR so consistency there seems the right thing to do.

jbarnoud: My previous objection was that it would break the existing code, but the deprecation in 1.1.0 should be enough.

PR #3153, opened 2021-03-13
Add tpr_resid_from_one keyword to TPRParser

### LINKED ISSUE #2364: TPRParser should not index from 0

lilyminium:
Fixes #2364

Master version of #3152.

 - Add tpr_resid_from_one keyword to TPRParser
 - Set default to ``False``, keeping current status quo
 - Emits a warning for ``tpr_resid_from_one=False`` indicating that default behaviour will change in 2.0

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -15,7 +15,7 @@ The rules for this file:
 ------------------------------------------------------------------------------
 ??/??/??  richardjgowers, IAlibay, tylerjereddy, xiki-tempula, lilyminium
 
-  * 1.0.2
+  * 1.1.0
 
 Fixes
   * Replaces decreated NumPy type aliases and fixes NEP 34 explicit ragged
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -174,6 +174,8 @@ Fixes
   * Contact Analysis class respects PBC (Issue #2368)
 
 Enhancements
+  * Added `tpr_resid_from_one` keyword to select whether TPRParser
+    indexes resids from 0 or 1 (Issue #2364, PR #3153)
   * Added support for FHI-AIMS input files (PR #2705)
   * vastly improved support for 32-bit Windows (PR #2696)
   * Added methods to compute the root-mean-square-inner-product of subspaces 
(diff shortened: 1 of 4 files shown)


package/MDAnalysis/topology/TPRParser.py
```diff
@@ -171,12 +171,24 @@ class TPRParser(TopologyReaderBase):
...
+            files index from 1.
+
         Returns
         -------
         structure : dict
+
+
+        .. versionchanged:: 1.0.2
+            Added the ``tpr_resid_from_one`` keyword to control if
+            resids are indexed from 0 or 1. Default ``False``.
```
orbeckst, line 191: Becomes 1.1.0 ...

orbeckst (changes requested):
Thank you.

The CHANGELOG entries are in the wrong spot (1.0.0?).

The introduction of the `tpr_resid_from_one` keyword forces us to make this 1.1.0 according to https://semver.org/ (ping @IAlibay ).

package/MDAnalysis/topology/TPRParser.py
lilyminium: I thought we were going to leave 1.x at the 1.0.x states? It's becoming increasingly difficult to deal with Python 2.7, as evidenced by the recent issues with CI. What about putting the deprecation in 2.0 and then making the final change in a later release?

Tagging any interested @MDAnalysis/coredevs.

package/CHANGELOG
```diff
@@ -13,7 +13,7 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" http://semver.org
 
 ------------------------------------------------------------------------------
-??/??/??  richardjgowers, IAlibay, tylerjereddy, xiki-tempula,
+??/??/??  richardjgowers, IAlibay, tylerjereddy, xiki-tempula, lilyminium
 
   * 1.0.2
```
IAlibay, line 18: To avoid confusion can you change this one to 1.1.0 please?
```suggestion
  * 1.1.0
```

We'll change the version tags on release, but this will just confuse folks doing any PRs.

lilyminium: Ok, done :)

orbeckst (approved):

PR #3154, opened 2021-03-13
Fixed issue -3148 (#3154) Warn user grid ignored DensityAnalysis

### LINKED ISSUE #3148: DensityAnalysis user defined grid coordinates are silently ignored if a grid centre is not provided.
## Expected behavior ##
If grid dimensions (`xdim, ydim, zdim`) are provided to `DensityAnalysis` but a grid centre (`gridcenter`) is not set,the user should receive an exception or warning that the grid dimensions are being ignored. This is noted in the docs, but a warning should probably be issued. 

## Actual behavior ##

If grid dimensions (`xdim, ydim, zdim`) are provided to `DensityAnalysis` but a grid centre (`gridcenter`) is not set, the grid dimensions are silently ignored. 

``` python
import MDAnalysis as mda
from MDAnalysis.tests.datafiles import TPR, XTC
from MDAnalysis.analysis import density

u = mda.Universe(TPR, XTC)
ow = u.select_atoms("name OW")
D = density.DensityAnalysis(ow, xdim=1.0,ydim=1.0,zdim=1.0)
D.run()
print(D.density.edges)

....

```

- 2.0.0-dev0
-3.7.5
- Which operating system?
-Ubuntu 19.10 (Eoan Ermine)

biogen98:
Fixes #3148

 -  Added a warning message when the user provided grid dimensions are ignored.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/analysis/density.py
```diff
@@ -443,6 +443,13 @@ def _prepare(self):
                           " defined grid will need to be "
                           "provided instead.")
                 raise ValueError(errmsg) from err
+            if any([self._xdim, self._ydim, self._zdim]):
```
IAlibay, line 446: I'm going to let the other devs weigh in here too, but I'm personally not super fond of adding extra warnings in what is already a warning/error heavy piece of code.

Personally I'd favour we instead make line 403 a call to any for the dims and gridcenter, then if any of them are not set they just trigger the errors from the checks in `_set_user_grid`.

Part of my rationale here is also that if folks are setting any of these values manually and missing any of the others, then it's probably unintentional and we should be failing rather than warning.

IAlibay:
Thanks for working on this @biogen98 :)

Aside from my comment (maybe @hmacdope, @fiona-naughton or @lilyminium can weigh in here). We also will need tests for the edge case being covered here. If you want to implement a test now, it can be easily be switched from testing for a warning to an error if we decide to go down that route.

package/MDAnalysis/analysis/density.py
biogen98: I see what you mean, this warning could be misleading in some cases. I will wait for response from the other devs, but I'm happy to change it.

biogen98:
Thanks for the fast response! 

What edge case are you referring too? Is a test needed to see if a user only inputs 1 or 2 grid dimension?
I am not very familiar with writing test, but I will look at the docs. 

p.s. It is normal that pytest throws errors regarding to the topology attribute right? This is my first time contributing, so just making sure that I did not break anything.

IAlibay:
> What edge case are you referring too? Is a test needed to see if a user only inputs 1 or 2 grid dimensions?

Yeah, so the main "edge case" this code is trying to cover is the case where we pass values to any of the dim values but we don't assign gridcenter. So what we would want to do is replicate this to make sure the right warning/error is thrown.

> I am not very familiar with writing test, but I will look at the docs.

As an initial start I'd have a look here:

https://github.com/MDAnalysis/mdanalysis/blob/3f589e375434fd6a9b16c7da2cb1342cc3088b0e/testsuite/MDAnalysisTests/analysis/test_density.py#L242-L256

You essentially want to do something in between these two tests where you pass `None` to gridcenter, but you pass values to one or all of the dimension arguments.

package/MDAnalysis/analysis/density.py
hmacdope: I agree with @IAlibay, in that the checks belong with the other grid checks in `_set_user_grid`.

IAlibay: Ok let's do that then.

biogen98: @IAlibay One question, when making line 403 a call to any for the dims and gridcenter, the error "ValueError: gridcenter must be a 3D coordinate" is thrown. Would adding a specific error to `_set_user_grid` be over complicating things?
I personally think this would make things more clear.

IAlibay: What'd you mean by "specific error"? I probably would just expand on that error message, just say f"gricenter is set to {gridcenter}: to use a user-defined grid, it must be a 3D coordinate" (or something better written but along those lines)?

biogen98: Oh, I was a bit unclear. I referred to the case where xyzdims are provided but the gridcenter isn't. When making line 403 a call to any for the dims and gridcenter, the above ValueError is thrown in this case. I thought that an error of the form "Grid dimensions are provided but no grid center is." would be more informative for the user. What are your thoughts on this?

IAlibay: We can add an extra user input validation in `_set_user_grid` if it makes things easier to read/more intuitive. If it can be worked within the context of our existing check then that's even better (i.e. if it's a check after a `try` then the cost is near 0).

biogen98:
@IAlibay I added the extra user input validation in `_set_user_grid', not sure if this is the way u suggested.
This is the best way I came up with, if there is a better way please let me know.

package/CHANGELOG
```diff
@@ -131,7 +131,10 @@ Enhancements
     'protein' selection (#2751 PR #2755)
   * Added an RDKit converter that works for any input with all hydrogens
     explicit in the topology (Issue #2468, PR #2775)
-  
+  * Added warning message and test for when user provided
+    grid dimensions are ignored. (Issue #3148, PR #3154)
```
fiona-naughton, line 135: Can be updated to reflect it'll now raise an error. I'd also clarify the root cause as being when grid center wasn't defined.

fiona-naughton (dismissed):
Looks good to me - just the changelog message to update, but then I think we're good to go

package/MDAnalysis/analysis/density.py
```diff
@@ -512,6 +513,10 @@ def _set_user_grid(gridcenter, xdim, ydim, zdim, smin, smax):
         except ValueError as err:
             errmsg = "Non-number values assigned to gridcenter"
             raise ValueError(errmsg) from err
+        if np.isnan(gridcenter).all() and any([xdim, ydim, zdim]):
```
IAlibay, line 516: It's late in the day so I might be missing something, but the first half of this if is essentially the try condition above right?

If so, it maybe cleaner to make the errmsg value just be conditional of `if any([xdim, ydim, zdim])`?

biogen98: I think the first half is indeed necessary to verify if in the input `gridcenter == None`. In this case the above exception is not raised, but a NumPy array with one `NaN` value is created.
I agree that otherwise it would be cleaner.

IAlibay: Then It might be better to put it alongside the if below? Otherwise you end up with an edge case where someone could pass an array of 3 `nan`s (which is possible unintentionally) through and it won't trigger anything?

biogen98: I added the extra check to throw a more informative error to the user in case grid dimensions where provided but no gridcenter. Putting the `any([xdim, ydim, zdim])` alongside the if statement below would reverse this.
If the user inputs grid dimensions my check will catch the edge case, only if the grid dimensions are also missing it wont trigger anything. 
Is in this case catching the edge case more important then an informative error? I assume adding an extra check would be messy.

IAlibay: > Is in this case catching the edge case more important then an informative error?
Ideally the answer here is to try to get a solution that covers as much as we can identify. You can't guarantee user behaviour, so when identified edge cases are best covered imho.

My suggestion above was to make the if on the dims a sub-condition of catching either nans or a non-3D shape, but if it's too messy..

> I assume adding an extra check would be messy.

Nothing stops us from re-formatting the way we do the checks 🙃 

How about:

```python
# Move up the `None` check
if any(x is None for x in [gridcenter, xdim, ydim, zdim]):
    # something along the lines that one or more of these are not set
    raise ValueError(errmsg)
# We assume the input should be an ndarray, so the old try/except isn't that useful
if (gridcenter.shape != (3,)
    or np.isnan(gridcenter).any()
    or not np.issubdtype(gridcenter.dtype, np.floating):
        raise ValueError(errmsg)
try:
    xyzdim ....
```

(or some more creative way of doing this)

This has the nice side-effect that if gridcenter is set but the dimensions are not, then you don't accidentally create an ndarray of NaNs.

It introduces that big `if` case in the middle, but it's not exactly really hard to read (at least to me).

That being said, it's late in the day for me, so I'm probably overthinking/being too pedantic. If @fiona-naughton or @hmacdope think it's reasonable as-is, I'd be happy to go forward with it.

hmacdope: @IAlibay, I agree, the flow control and error handling is clear here. This layout jams the gridcenter, xdim, ydim, zdim checks in together which is the actual expected behaviour. Thanks for patience @biogen98.

biogen98: > That being said, it's late in the day for me, so I'm probably overthinking/being too pedantic.
@IAlibay I appreciate the suggestions, you raised good points. I will take a look at a nice way to implement it, was just a bit hesitant to change too much of the code I guess.

biogen98:
I added a separate check for `NaN` values to also include the grid dimensions, now all the edge cases are covered.

package/CHANGELOG
```diff
@@ -131,9 +131,13 @@ Enhancements
     'protein' selection (#2751 PR #2755)
   * Added an RDKit converter that works for any input with all hydrogens
     explicit in the topology (Issue #2468, PR #2775)
-  
+  * Added a ValueError raised when not given a gridcenter while providing
```
hmacdope, line 134: *Nitpick:* Can you mention the unit in which you made the changes (DensityAnalysis)?

package/MDAnalysis/analysis/density.py
```diff
@@ -400,7 +400,8 @@ def __init__(self, atomgroup, delta=1.0,
 
     def _prepare(self):
         coord = self._atomgroup.positions
-        if self._gridcenter is not None:
+        if (self._gridcenter is not None or
```
hmacdope, line 403: So now setting any of the values triggers this path. I think that is a good change and should be the expected behaviour. @IAlibay you happy with this?


We should also confirm that no changes need to be made in the docs to reflect this.

package/MDAnalysis/analysis/density.py
```diff
@@ -507,17 +508,20 @@ def _set_user_grid(gridcenter, xdim, ydim, zdim, smin, smax):
...
-            errmsg = "Non-number values assigned to gridcenter"
-            raise ValueError(errmsg) from err
-        if gridcenter.shape != (3,):
-            raise ValueError("gridcenter must be a 3D coordinate")
+            raise ValueError("Gridcenter must be a 3D coordinate") from err
         try:
             xyzdim = np.array([xdim, ydim, zdim], dtype=np.float32)
         except ValueError as err:
             raise ValueError("xdim, ydim, and zdim must be numbers") from err
+        if any(np.isnan(gridcenter)) or any(np.isnan(xyzdim)):
```
hmacdope, line 522: Another good change.

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -233,20 +233,28 @@ def test_warn_userdefn_smallgrid(self, universe):
...
                 gridcenter=self.gridcenters['error1']).run(step=5)
 
     def test_ValueError_userdefn_gridcenter_type(self, universe):
         # Test gridcenter includes non-numeric strings
-        with pytest.raises(ValueError, match="Non-number values assigned to gridcenter"):
+        with pytest.raises(ValueError, match="Gridcenter must be a 3D coordinate"):
             D = density.DensityAnalysis(
                 universe.select_atoms(self.selections['static']),
                 delta=self.delta, xdim=10.0, ydim=10.0, zdim=10.0,
                 gridcenter=self.gridcenters['error2']).run(step=5)
```
hmacdope, line 249: Could you also add a test for the `NaN` check code-path?

package/MDAnalysis/analysis/density.py
```diff
@@ -507,17 +508,20 @@ def _set_user_grid(gridcenter, xdim, ydim, zdim, smin, smax):
            Now a staticmethod of :class:`DensityAnalysis`.
         """
```
hmacdope, line 509: These updates to error checking need to be documented in the `Raises` section of the `DensityAnalysis` docstring, as they are now out of date.

hmacdope (changes requested):
Looking good so far, just a few things to change/confirm including some updates to the docs.

package/MDAnalysis/analysis/density.py
biogen98: I think removing the mention that xdim, ydim and zdim if gridenter is `None` would be enough. Instead of just ignoring it, now an error will be thrown.

hmacdope: LGTM

hmacdope (approved):
Looks good to me, thanks for making changes to docs and adding the `NaN` test. :)

IAlibay (changes requested):
thanks :) just some comments on the changelog, and I think we might need to expand the docstring for gridcenter to make the options clearer to users.

package/MDAnalysis/analysis/density.py
```diff
@@ -202,14 +202,11 @@ class DensityAnalysis(AnalysisBase):
             3 element numpy array detailing the x, y and z coordinates of the
```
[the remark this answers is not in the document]
hmacdope: Yep, could do with a little bit more explanation around what you should set if you want to set a `gridcenter` etc etc.

package/CHANGELOG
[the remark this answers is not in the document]
biogen98: Will do! Thanks for pointing it out.

package/MDAnalysis/analysis/density.py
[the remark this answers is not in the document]
biogen98: I personally agree that this is maybe not that clear. Adding a paragraph to the NOTES section is option?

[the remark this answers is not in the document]
hmacdope: I think better in NOTES as can be more detailed there about option *combinations* rather than just options themselves. @IAlibay ?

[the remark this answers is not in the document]
IAlibay: Sure, maybe just a couple of words to direct users to `NOTES` for more information? As long as y'all think it makes sense to users I don't have large preferences here. We explain it in examples I think, so notes can just refer to that probably.

hmacdope:
@IAlibay @fiona-naughton, have your requested changes been addressed?

hmacdope:
@biogen98 I am satisfied that the issue is resolved. Thanks for your patience as we are quite overloaded at the moment.

If no one has any more comments I will merge soon (sorry for spam @IAlibay, @fiona-naughton).

IAlibay (approved):

IAlibay:
I've dismissed @fiona-naughton's review as now-completed. Please proceed with the merge @hmacdope

PR #3157, opened 2021-03-14
Write out chainID to PDB instead of segID

### LINKED ISSUE #3144: ChainID is overwritten by segid
## Expected behavior ##

I can round-trip chainID information from files.

## Actual behavior ##

ChainIDs get overwritten by SegmentIDs, even though the latter is not actually in the original PDB specification. Because the default segment name is "system", this also means that the default chainID is "M" and that's kind of weird.

``` python
>>> import MDAnalysis as mda
>>> u = mda.Universe(PDB)
>>> u.atoms.chainIDs = "x"
>>> u.atoms.write("chain_x.pdb")
>>> u = mda.Universe("chain_x.pdb")
>>> u.atoms.chainIDs
array(['M', 'M', 'M', ..., 'M', 'M', 'M'], dtype=object)
```

And converting it to RDKit and using that to write it out ends up with weird spacing. :(

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) current develop
- Which operating system?

HenryKobin:
Fixes #3144

 - When writing to a file, a check will be made for if chainIDs exist, and if they do they will be used instead of being defaulted to the last letter of the segid. If there is no chainID present, then segid's last letter will be used (like how this was before).

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?


package/CHANGELOG
```diff
@@ -16,16 +16,17 @@ The rules for this file:
 ??/??/?? tylerjereddy, richardjgowers, IAlibay, hmacdope, orbeckst, cbouy,
          lilyminium, daveminh, jbarnoud, yuxuanzhuang, VOD555, ianmkenney,
          calcraven，xiki-tempula, mieczyslaw, manuel.nuno.melo, PicoCentauri,
-         hanatok, rmeli, aditya-kamath, tirkarthi
+         hanatok, rmeli, aditya-kamath, tirkarthi, Henry Kobin
```
lilyminium, line 19:
```suggestion
         hanatok, rmeli, aditya-kamath, tirkarthi, HenryKobin
```

The CHANGELOG asks for your GitHub username. Instead, please add your name to the AUTHORS files.

package/CHANGELOG
```diff
@@ -16,16 +16,17 @@ The rules for this file:
...
 
   * 2.0.0
 
 Fixes
+  * Fixed chainID being overwritten by segid during file write. (Issue #3144)
```
lilyminium, line 24: Could you please be more specific and mention the PDBParser?

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -1051,6 +1051,8 @@ def _write_timestep(self, ts, multiframe=False):
            Writing now only uses the contents of the elements attribute
            instead of guessing by default. If the elements are missing,
            empty records are written out (Issue #2423).
+           Atoms are now checked for a chainID instead of being overwritten
+           by the last letter of the SegmentID (Issue #3144).
```
lilyminium, line 1055:
```suggestion
           by the last letter of the `segid` (Issue #3144).
```

Try to use the actual names of the attributes.

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -1051,6 +1051,8 @@ def _write_timestep(self, ts, multiframe=False):
            Writing now only uses the contents of the elements attribute
            instead of guessing by default. If the elements are missing,
            empty records are written out (Issue #2423).
+           Atoms are now checked for a chainID instead of being overwritten
```
lilyminium, line 1054: Could you be more specific? e.g. What happens if there is no chainID?

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -1080,13 +1082,18 @@ def get_attr(attrname, default):
         altlocs = get_attr('altLocs', ' ')
         resnames = get_attr('resnames', 'UNK')
         icodes = get_attr('icodes', ' ')
+        chainids = get_attr('chainIDs', ' ')
```
lilyminium, line 1085: How will you distinguish between not having a chainID (``not hasattr(obj, "chainID")`` and having an empty chainID?

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -1080,13 +1082,18 @@ def get_attr(attrname, default):
...
         resids = get_attr('resids', 1)
         occupancies = get_attr('occupancies', 1.0)
         tempfactors = get_attr('tempfactors', 0.0)
         atomnames = get_attr('names', 'X')
         elements = get_attr('elements', ' ')
         record_types = get_attr('record_types', 'ATOM')
+        chainids = (
+            [id[:4] if (chainids[index] != '') else segids[index][-1:]
+            for index, id in enumerate(chainids)]
+            ) # check for a chainID, if no chainID default to segid
```
lilyminium: I find this not ideal -- there may be instances where the chain is empty in a file that otherwise has L/M/N chains. In these cases, we wouldn't want to muddy the waters by assigning only *some* atoms to the last letter of the `segid`. Instead, it would be more beneficial to simply check if the atoms have the chainID attribute, and if not, use the segid for all of them.

> id[:4]

Could you explain why you decided to limit the chainID to 4 letters? In the PDB spec, chainIDs are only 1 character long.

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -1107,13 +1114,15 @@ def get_attr(attrname, default):
...
             vals['resName'] = resnames[i][:4]
-            vals['chainID'] = segids[i][-1:]
             vals['resSeq'] = util.ltruncate_int(resids[i], 4)
             vals['iCode'] = icodes[i][:1]
             vals['pos'] = pos[i]  # don't take off atom so conversion works
             vals['occupancy'] = occupancies[i]
             vals['tempFactor'] = tempfactors[i]
             vals['segID'] = segids[i][:4]
+            vals['chainID'] = (
+                chainids[i][:4] if (chainids[i][:4] != ' ') else segids[i][-1:]
```
lilyminium, line 1124: My concerns regarding the 4-letter chainIDs and the confusion between empty and missing chainIDs is relevant here, too.

testsuite/MDAnalysisTests/coordinates/test_pdb.py
```diff
@@ -463,6 +463,15 @@ def test_segid_chainid(self, universe2, outfile):
         u_pdb = mda.Universe(outfile)
         assert u_pdb.segments.chainIDs[0][0] == ref_id
 
+    def test_segid_chainid_overwrite(self, universe3, outfile):
+        """check if chainID gets round-tripped after file write (issue #3144)"""
+        ref_id = 'x'
+        u = universe3
```
lilyminium, line 469: This line feels a little redundant.

testsuite/MDAnalysisTests/coordinates/test_pdb.py
```diff
@@ -463,6 +463,15 @@ def test_segid_chainid(self, universe2, outfile):
...
+        u.atoms.chainIDs = ref_id
+        u.atoms.write(outfile)
+        u_pdb = mda.Universe(outfile)
+        assert u_pdb.atoms.chainIDs[0] == ref_id
```
lilyminium, line 473: You could check  that all the atom chainIDs equal the ref_id with ``assert_equals``.

lilyminium (changes requested):
Thanks for fixing this @HenryKobin! Just a few small comments :-)

Please add yourself to the AUTHORS file, as this is your first contribution to MDAnalysis.

package/MDAnalysis/coordinates/PDB.py
HenryKobin: this makes sense. Considering right now

HenryKobin: Yeah, I think I overthought this one. hasattr would be a much better use case. Also, I don't have a great reason for limiting to 4 characters, I think I misread. Working on the change :+1:

HenryKobin:
@lilyminium  I'm  having trouble passing [this test](https://github.com/MDAnalysis/mdanalysis/blob/cbc5b6e0ea7ec76a330d3452d579466a830a48f3/testsuite/MDAnalysisTests/coordinates/test_pdb.py#L1083). It appears this file does have a chainID property, however they are all blank. I was under the impression I should only set the chainIDs to the SegmentID default if they don't have the chainID property. Am I misunderstanding?

IAlibay:
Just copying over from the discord.

The PDB standard is surprisingly quite clear about what a chain ID can be in a PDB file:
"Non-blank alphanumerical character is used for chain identifier."
(see: http://www.wwpdb.org/documentation/file-format-content/format33/sect9.html#ATOM)

In my opinion we shouldn't be allowing the writing of blank chain IDs. In an ideal world we would just throw an error and avoid writing out of standard, but I assume this might break a lot of things downstream. Probably the easiest answer here is to set the chainID to the SegmentID default if it exists but doesn't conform to the PDB standard?

I'm sure others have views on this though, I am _a bit_ of a PDB standard purist.

HenryKobin:
Got it. I'll wait and see what the consensus is, and move forward from there

lilyminium:
Well, "non-blank" is pretty unambiguous 😅. I think, following the spirit of #2627 (where elements are read from the file), we could go for an all-or-nothing approach: if even one chainID is blank, we ignore them all and write the last letter of the segid. What do you think @HenryKobin @IAlibay?

IAlibay:
> Well, "non-blank" is pretty unambiguous 😅

My interpretation is that it means `str.isalnum() == True`, i.e. non-whitespace alphanumeric. Maybe I'm misinterpreting something though? The PDB folks are quite responsive on twitter, so we can ping them if needed.

> I think, following the spirit of #2627 (where elements are read from the file), we could go for an all-or-nothing approach: if even one chainID is blank, we ignore them all and write the last letter of the segid. What do you think @HenryKobin @IAlibay?
> 
> Edit: similar to the PDB elements reading, this would likely need a warning too because users probably haven't memorised the PDB spec.

Honestly this is a bit of a difficult one, residues are modular so I can see folks setting some residues and leaving others blank. Personally I wouldn't default to segid[-1] at all (I think it's just easier to default to 'U' or something like that, i.e. like what things like openbabel do), although I assume there is some historical reason for doing so (@jbarnoud might have some ideas on this?).

I should note that we now handle PDB element reading in a non-all-or-nothing approach (see #3001). I'm ok with going all-or-nothing here to keep the PR simple, but we'll need a long term solution that a) prevents us ever defaulting to writing out empty strings, b) defaults to a predictable value, c) allows modularity.

HenryKobin:
@IAlibay If I switch to all or nothing a couple tests will break, so I might have to alter those to pass (they are ensuring that segID is being used to create a chainID).

I like the idea of checking if there are any non alphanumeric ids in the list and defaulting to something expected if this occurs (although I am not an end user of the tool, so I don't know what the end user would want).

```
        if not any([id.isalnum() for id in chainids]):
            chainids = np.array(['U'] * len(atoms))
```

something like the above

package/MDAnalysis/coordinates/PDB.py
HenryKobin: @lilyminium I just realized i've been marking these conversations as resolved as I've addressed the change. Is that bad practice? Should I be letting you mark these as resolved or should I resolve them as they are addressed? Thanks.

lilyminium: What you're doing now is good :-)

HenryKobin: Thanks @lilyminium ! :smile:  Is there anything else I need to do at this point, or are we just waiting on confirmation of if they want to do anything more than just use the chainID (if the attribute exists and is all alphanumeric) else last char of segid?

jbarnoud:
I do not know what the best behaviour should be. As I see it, the issue is that chains and segments are pretty arbitrary concepts and they are not consistent between parsers/formats.

If a chain is there, it should be used, regardless of what it is. The only constraint is that it should not be more than one character. "alpha-numerical" is ambiguous, even if you stay in the realm of ASCII characters, `¼` and `°`  are alpha-numerical.

Defaulting to the last character of the segment name makes some assumptions. It assumes that:
* if the segment name changes from one "chain" to the next, the last letter reflects that
* the last character will not be the same for 2 different chains.
Basically, it assumes that the segment name encodes the chain name: `segA`, `segB`, etc...

One solution I implemented for other projects was to name chains `A`, `B`, `C`... and changing the chain name for new segments. If we go for that, I'd try to respect the chain ids that are already in the Universe if any: this means starting the sequence after the largest chain id already in use.

HenryKobin:
@jbarnoud so would a solution perhaps be to check if a chainID attribute exists, and if it is anything other than 1 character long or non existent to find all unique segment names and set the chainID to A,B,C... changing each time the segment name changes?

jbarnoud:
It is indeed what I would do rather than using the segment name. It is worth seeing what the others think about it, though. There are very good arguments to keep it simple and just name the chain `A`, `X`, or `U`: somehow what I suggest is some sort of a guesser for chain ids and it is historically not always the best of ideas. Because the chainID is an attribute of the atom and NOT an attribute of the segment, every clever ideas based on segments can encounter nasty corner cases.

So to summarize:
* we probably want to keep it simple
* if we want to make it clever, I'll go for what I suggested above

It is nothing but my own opinion, though; @lilyminium or @IAlibay may think differently.

jbarnoud:
For alpha-numerical, I would take any non-empty single character as valid and emit a warning if it is not in the range `[a-zA-Z0-9]` (which is not what `isalnum` look for).

lilyminium:
@HenryKobin, thank you for your patience. I think consensus amongst @jbarnoud, @IAlibay and I is that:

- if the chainID for all atoms exists and a valid, non-empty, [a-zA-Z0-9] character, use that
- if the chainID for *some* atoms is missing, raise a warning [and use "X"] -- @jbarnoud was this right?
- if the chainID is > 1 character for any atoms, raise a warning
- if the chainID attribute is missing or is empty for all atoms, use "X" as a placeholder value

What do you think?

jbarnoud:
> if the chainID for _some_ atoms is missing, raise a warning [and use "X"] -- @jbarnoud was this right?

Use "X" for the atoms that do not have a chainID.

HenryKobin:
@lilyminium  I think it makes sense. I just wanna make sure i'm following, though.

`if the chainID for some atoms is missing, raise a warning and use "X" -- @jbarnoud was this right?`
`Use "X" for the atoms that do not have a chainID.`

So is the plan is to replace individual chainID's with "X" if they are missing, not to change _every_ chainID to "X" if one is missing, correct? 

`if the chainID is > 1 character for any atoms, raise a warning` 
In this case I should raise a warning and additionally use "X", right? 

Thanks for clarifying. I'm not very familiar with many of these atom concepts.

HenryKobin:
Just brainstorming, would something like this be overkill?
Of course I would like to avoid adding a new function, but perhaps it could be useful in the future depending on how granular our validation becomes. I could also try list comprehension, but that doesn't feel as pythonic since we are checking against multiple factors, so it could become hard to read.
```
        def validate_chainids(chainids, default):
            """Validate each atom's chainID 

            chainids - np array of chainIDs
            default - default value in case chainID is considered invalid
            """
            invalid_length_ids = False
            invalid_char_ids = False
            missing_ids = False

            for (i, id) in enumerate(chainids):
                if id is None:
                    missing_ids = True
                    chainids[i] = default
                elif len(id) > 1:
                    invalid_length_ids = True
                    chainids[i] = default
                elif not id.isalnum():
                    invalid_char_ids = True
                    chainids[i] = default

            if invalid_length_ids:
                warnings.warn("Found chainIDs with invalid length"
                              " Corresponding atoms will use default value of '{}'"
                              "".format(default))
            if invalid_char_ids:
                warnings.warn("Found chainIDs using unnaccepted character"
                              " Corresponding atoms will use default value of '{}'"
                              "".format(default))
            if missing_ids:
                warnings.warn("Found missing chainIDs"
                              " Corresponding atoms will use default value of '{}'"
                              "".format(default))
            return chainids

        chainids = validate_chainids(chainids, "X")
```

HenryKobin:
image

I had to alter a couple of tests to reflect the new method of handing chainID, however this one test in particular fails and I can't figure out why it would be failing... As far as I can tell nothing my code is doing is changing what's failing here? Unless I'm misunderstanding the relationship between segid and chainID. @lilyminium would appreciate input on this whenever you have the chance :face_with_head_bandage:

lilyminium:
Sorry for the long silence, @HenryKobin. I would think that if segIDs are not present, writing a blank value instead would be most expected. In reality this would not happen often, as segIDs are always present in our current format readers ( https://userguide.mdanalysis.org/stable/topology_system.html#canonical-attributes).

lilyminium:
> Just brainstorming, would something like this be overkill?
> 
> Of course I would like to avoid adding a new function, but perhaps it could be useful in the future depending on how granular our validation becomes. I could also try list comprehension, but that doesn't feel as pythonic since we are checking against multiple factors, so it could become hard to read.
> 
> ```
> 
>         def validate_chainids(chainids, default):
> 
>             """Validate each atom's chainID
> 
> 
> 
>             chainids - np array of chainIDs
> 
>             default - default value in case chainID is considered invalid
> 
>             """
> 
>             invalid_length_ids = False
> 
>             invalid_char_ids = False
> 
>             missing_ids = False
> 
> 
> 
>             for (i, id) in enumerate(chainids):
> 
>                 if id is None:
> 
>                     missing_ids = True
> 
>                     chainids[i] = default
> 
>                 elif len(id) > 1:
> 
>                     invalid_length_ids = True
> 
>                     chainids[i] = default
> 
>                 elif not id.isalnum():
> 
>                     invalid_char_ids = True
> 
>                     chainids[i] = default
> 
> 
> 
>             if invalid_length_ids:
> 
>                 warnings.warn("Found chainIDs with invalid length"
> 
>                               " Corresponding atoms will use value of '{}'"
> 
>                               "".format(default))
> 
>             if invalid_char_ids:
> 
>                 warnings.warn("Found chainIDs using unnaccepted character"
> 
>                               " Corresponding atoms will use value of '{}'"
> 
>                               "".format(default))
> 
>             if missing_ids:
> 
>                 warnings.warn("Found missing chainIDs"
> 
>                               " Corresponding atoms will use value of '{}'"
> 
>                               "".format(default))
> 
>             return chainids
> 
> 
> 
>         chainids = validate_chainids(chainids, "X")
> 
> ```

That is a very specific check. It is verbose but, as far as I can tell, not much slower than just checking `if id is None or not id.isalnum() or len(id) !=1`. I'm happy with it, but could be persuaded otherwise if @Ialibay and @jbarnoud disagree. 

One thing I'll note though is that `id` is a built in function in Python and it's preferable to not overwrite it.

HenryKobin:
No worries @lilyminium , I've been moving apartments so it's been hectic over here as well. I agree that using `id` is probably not best practice for that function. I'll change it to `chainid`.  As for the segid situation, I can set it so that it uses `' '` instead of defaulting to chainids (if no one objects)

IAlibay:
Sorry I've not been keeping up here. @HenryKobin just wanted to check what we can do to push this towards merging. Is there anything on our end that we still need to decide on?

HenryKobin:
Hey @IAlibay , @lilyminium  sorry for the delay. I've noticed that changing the segids to use `' '` instead of defaulting to chainIDs may have some farther reaching consequences, as this causes a few tests to fail for reasons a bit unknown to me. I would appreciate any advice in this regard. Thank you

package/MDAnalysis/topology/PDBParser.py
```diff
@@ -288,10 +290,6 @@ def _parseatoms(self):
             warnings.warn("Serial numbers went over 100,000.  "
                           "Higher serials have been guessed")
 
-        # If segids not present, try to use chainids
-        if not any(segids):
-            segids = chainids
```
lilyminium, line 293: If you put this back in, that second and third failing tests should pass. I think that this is expected behaviour from most users, that differences in chain are conceptually similar to differences in segment -- better not to change this part.

package/MDAnalysis/topology/PDBParser.py
```diff
@@ -360,7 +358,7 @@ def _parseatoms(self):
             attrs.append(Segids(segids))
         else:
             n_segments = 1
-            attrs.append(Segids(np.array(['SYSTEM'], dtype=object)))
+            attrs.append(Segids(np.array([' '], dtype=object)))
```
lilyminium, line 361: Instead of using spaces, use an empty string for blank. That being said, the reason the first failing test is failing is that some functions use "segid xx and resid yy" to select atoms -- so if `xx=''` or `xx=' '`, the string "segid  and resid yy" breaks the parser. You could change the code that the test shows is failing, but I would be wary of more things failing that tests aren't showing. If you change the default value back to "SYSTEM", that's not too much of an issue -- the original issue was just about writing out segids.

lilyminium (commented):
@HenryKobin looking at the errors, you're right that messing with *reading* segids has far-reaching consequences. Unlike the chainID, which is just a topology attribute (essentially a label for each atom), segments are structurally important in defining different parts of the molecule. It's probably simpler to focus this issue on accurately writing out segids, and we can look at changing default segid later as this will require some thought.

HenryKobin:
@lilyminium you're right! Whoops. I mixed up reading with writing... :disappointed: 

I reverted the changes. I did have to alter one test, though. Since we are defaulting to 'X' for a chainID if it is missing, there are always chainids -  so segids will default to `'X'` when it parses the file. The test was formerly expecting `'SYSTEM'`.

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -1081,13 +1083,51 @@ def get_attr(attrname, default):
...
+
+            chainids - np array of chainIDs
+            default - default value in case chainID is considered invalid
+            """
+            invalid_length_ids = False
+            invalid_char_ids = False
+            missing_ids = False
+
+            for (i, chainid) in enumerate(chainids):
+                if chainid is None:
```
lilyminium, line 1105: Is it ever None? Isn't the default value above " "?

testsuite/MDAnalysisTests/coordinates/test_pdb.py
```diff
@@ -455,13 +455,30 @@ def get_MODEL_lines(filename):
             # test number (only last 4 digits)
             assert int(line[10:14]) == model % 10000
 
-    def test_segid_chainid(self, universe2, outfile):
-        """check whether chainID comes from last character of segid (issue #2224)"""
-        ref_id = 'E'
-        u = universe2
+    def test_chainid_validated(self, universe3, outfile):
```
lilyminium, line 458: Could you please use `pytest.mark.parametrize` (https://docs.pytest.org/en/stable/parametrize.html) to turn this into three tests with different arguments? There are a few examples around if you ctrl+F in the tests.

lilyminium (changes requested):
Thanks @HenryKobin, looking through I just have a question about None and if you can parametrize the test :-)

package/MDAnalysis/coordinates/PDB.py
HenryKobin: You're correct, it shouldn't ever be None.

testsuite/MDAnalysisTests/coordinates/test_pdb.py
HenryKobin: Good idea!

HenryKobin:
@lilyminium , I made the requested changes.

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -1081,13 +1083,51 @@ def get_attr(attrname, default):
         resnames = get_attr('resnames', 'UNK')
         icodes = get_attr('icodes', ' ')
         segids = get_attr('segids', ' ')
+        chainids = get_attr('chainIDs', ' ')
```
lilyminium, line 1086:
```suggestion
        chainids = get_attr('chainIDs', '')
```

Your `validate_chainids` function checks for the empty string "", not the space.

lilyminium (changes requested):
Thanks @HenryKobin, if you just switch that space to the empty string I think it's good to go!

HenryKobin:
@lilyminium committed your suggested changes! Thanks!

lilyminium (approved):

lilyminium:
Thank you @HenryKobin for fixing this bug, my PDB files will be much happier for it! :-)

HenryKobin:
@lilyminium thank you for your patience and assistance in walking me through everything! 😁

Issue #2821, opened 2020-07-03
u.select_atoms(...).bonds should return only bonds with atoms *inside* of the selection


cbouy:
No necessarily a bug, but I think it's worth mentioning: 
When making a selection, should the resulting bonds only contain bonds where both atoms are in the selection, or should it return all bonds where at least one atom of the atomgroup is part of the bond ?

The current behavior is to return all bonds where at least one atom of the atomgroup is part of the bond

## Code to reproduce the behavior ##

``` python
>>> import MDAnalysis as mda
>>> from rdkit import Chem

>>> mol = Chem.MolFromSequence("GAG")
>>> u = mda.Universe(mol)
>>> u.select_atoms("resid 2").indices
array([4, 5, 6, 7, 8])

>>> u.select_atoms("resid 2").bonds.indices
array([[2, 4],
       [4, 5],
       [5, 6],
       [5, 8],
       [6, 7],
       [6, 9]], dtype=int32)

```
![image](https://user-images.githubusercontent.com/27850535/86459000-6f2f5000-bd26-11ea-88ad-25c22c3dfad9.png)

Should it return only the green bonds, or should it return the green bonds and the yellow ones (between atoms 2-4 and 6-9) as well ?
The current behavior is green+yellow.

## Current version of MDAnalysis ##

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`)
2.0.0-dev0
- Which version of Python (`python -V`)?
Python 3.6.10
- Which operating system?
Ubuntu 18.04.4 LTS through a VM (WSL2)

lilyminium:
Try `atomgroup_intersection(strict=True)` -- not sure if there's a quicker way to do this. Maybe turning `bonds` etc into functions (`ag.bonds(strict=True)`) instead of properties would be easier.

```python
ag = u.select_atoms('protein')
bonds = ag.bonds.atomgroup_intersection(ag, strict=True)
```

cbouy:
Great, thank you! Yes it looks a bit heavy to be honest, but I wouldn't expect users to use this everyday so...
I'm not sure if changing to `ag.bonds()` is a good idea, tbh as a new user I wouldn't expect `bonds` to be a method, it's a bit confusing. Maybe adding a `get_bonds()` with parameters could work ?

lilyminium:
[...]

Issue #1264, opened 2017-03-27
Dihedrals and selections


mimischi:
When creating a selection, e.g., `backbone = u.select_atoms('backbone')` and then looking at all the dihedral angles of that selection, MDAnalysis also returns dihedrals between atoms pairs that should not be presented.

### Initial code

```python
import numpy as np
import MDAnalysis as mda
from MDAnalysis.tests.datafiles import PSF, DCD

u = mda.Universe(PSF, DCD)
backbone = u.select_atoms('backbone')

backbone
<AtomGroup with 855 atoms>
```

### Expected behaviour

Only list dihedral angles between atoms in the selection (here: backbone). Total number of dihedrals: 425.

```python
# Extract indiced of dihedral angles phi and psi 
dihedral_indices = np.zeros([2 * (len(backbone.residues) - 1)], dtype=np.int)

j = 0
for i, dih in enumerate(backbone.dihedrals):
    if dih.atoms[0].name == 'C' and dih.atoms[-1].name == 'C' or dih.atoms[0].name == 'N' and dih.atoms[-1].name == 'N': 
        dihedral_indices[j] = i
        j += 1

for dih in backbone.dihedrals[dihedral_indices]:
    print dih.atoms

<AtomGroup [<Atom 1: N of type 56 of resname MET, resid 1 and segid 4AKE>, <Atom 5: CA of type 22 of resname MET, resid 1 and segid 4AKE>, <Atom 18: C of type 20 of resname MET, resid 1 and segid 4AKE>, <Atom 20: N of type 54 of resname ARG, resid 2 and segid 4AKE>]>
<AtomGroup [<Atom 18: C of type 20 of resname MET, resid 1 and segid 4AKE>, <Atom 20: N of type 54 of resname ARG, resid 2 and segid 4AKE>, <Atom 22: CA of type 22 of resname ARG, resid 2 and segid 4AKE>, <Atom 42: C of type 20 of resname ARG, resid 2 and segid 4AKE>]>
<AtomGroup [<Atom 20: N of type 54 of resname ARG, resid 2 and segid 4AKE>, <Atom 22: CA of type 22 of resname ARG, resid 2 and segid 4AKE>, <Atom 42: C of type 20 of resname ARG, resid 2 and segid 4AKE>, <Atom 44: N of type 54 of resname ILE, resid 3 and segid 4AKE>]>
[... truncated ...]
[...]

PR #3160, opened 2021-03-14
Added get_connections

lilyminium:
Relates to #1264
Relates to #2821 

 - Added `get_connections` method for Components and Groups to control which connections get returned
 - Added ``outside`` keyword to `get_atoms` argument of the Connection topology attributes, to easily toggle the behaviour of Groups.bonds, Groups.angles, etc. This is set to ``True`` to keep status quo for 1.1.0. It will be set to False in 2.0.0.
 - Added deprecation warning for the Groups.bonds, Groups.angles, etc. methods

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/core/groups.py
+++ b/package/MDAnalysis/core/groups.py
@@ -388,6 +388,41 @@ class _MutableBase(object):
                 err += 'Did you mean {match}?'.format(match=match)
             raise AttributeError(err)
 
+    def get_connections(self, typename, outside=True):
+        """
+        Get bonded connections between atoms as a
+        :class:`~MDAnalysis.core.topologyobjects.TopologyGroup`.
+
+        Parameters
+        ----------
+        typename : str
+            group name. One of {"bonds", "angles", "dihedrals",
+            "impropers", "ureybradleys", "cmaps"}
+        outside : bool (optional)
+            Whether to include connections involving atoms outside
+            this group.
+
+        Returns
+        -------
+        TopologyGroup
+            containing the bonded group of choice, i.e. bonds, angles,
+            dihedrals, impropers, ureybradleys or cmaps.
+
+        .. versionadded:: 1.1.0
+        """
+        # AtomGroup has handy error messages for missing attributes
+        ugroup = getattr(self.universe.atoms, typename)
+        if not len(ugroup):
+            return ugroup
+        func = np.any if outside else np.all
+        try:
+            indices = self.atoms.ix_array
+        except AttributeError:  # if self is an Atom
+            indices = self.ix_array
+        seen = [np.in1d(col, indices) for col in ugroup._bix.T]
+        mask = func(seen, axis=0)
+        return ugroup[mask]
+
 
 class _ImmutableBase(object):
     """Class used to shortcut :meth:`__new__` to :meth:`object.__new__`.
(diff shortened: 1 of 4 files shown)


testsuite/MDAnalysisTests/core/test_groups.py
```diff
@@ -1466,3 +1466,85 @@ def test_decorator(self, compound, pbc, unwrap):
...
         else:
             assert_equal(self.dummy_funtion(compound=compound, pbc=pbc, unwrap=unwrap), 0)
+
+
+@pytest.fixture()
+def tpr():
+    return mda.Universe(TPR)
+
+
+class TestGetConnections(object):
```
IAlibay, line 1476: Could we get some tests to cover at the very least the two cases that were outlined in #1264 and #2821?

lilyminium: Do you mean those specific files and selections, or do you mean using the .bonds and .dihedrals attributes? The latter will have to wait until 2.0

richardjgowers (approved):
Lgtm.

package/MDAnalysis/core/groups.py
```diff
@@ -388,6 +388,37 @@ def __getattr__(self, attr):
...
+        Returns
+        -------
+        TopologyGroup
+            containing the bonded group of choice, i.e. bonds, angles,
+            dihedrals, impropers, ureybradleys or cmaps.
+
+        .. versionadded:: 1.1.0
+        """
+        # AtomGroup has handy error messages for missing attributes
+        ugroup = getattr(self.universe.atoms, typename)
```
richardjgowers, line 414: So we're not using `atomgroup_intersection` here because this method can handle not just things in a `TopologyGroup`?

package/MDAnalysis/core/topologyattrs.py
```diff
@@ -2250,15 +2250,42 @@ def _bondDict(self):
...
         try:
             unique_bonds = set(itertools.chain(
                 *[self._bondDict[a] for a in ag.ix]))
         except TypeError:
             # maybe we got passed an Atom
             unique_bonds = self._bondDict[ag.ix]
-        bond_idx, types, guessed, order = np.hsplit(
-            np.array(sorted(unique_bonds), dtype=object), 4)
+        unique_bonds = np.array(sorted(unique_bonds), dtype=object)
+        if not outside:
```
cbouy, line 2275: Why not 
```
if outside: 
    <deprecation code>
else:
    <np.all>
```
It makes it more readable imo

lilyminium: I started with ``strict=False`` 😅

This will make it easier to add to 2.0, though, where we will not have the deprecation warning.

package/MDAnalysis/core/groups.py
lilyminium: I think it was more to avoid having to distinguish between Atoms and AtomGroups (etc) with ``ix_array``, as ``atomgroup_intersection`` only accepts AtomGroups

lilyminium: I just realised that `Residue.bonds` is not existing functionality, so I added tests.

lilyminium:
@IAlibay I forget where you asked me this, but the reason `get_atoms` doesn't use `get_connections` is it results in recursive logic. I think `get_atoms` is used to create the TopologyGroup that `get_connections` works off.

richardjgowers (approved):

PR #3162, opened 2021-03-14
Use match to check hole2 warnings

lilyminium:
~Fixes #~

Tests are failing in #3160 because tests for hole2 check how many warnings are generated. This has been a pain in the past as we add and remove warnings.

 - Removes the length of the warning check
 - Instead, checks for the desired warning by checking the message

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/testsuite/MDAnalysisTests/analysis/test_hole2.py
+++ b/testsuite/MDAnalysisTests/analysis/test_hole2.py
@@ -203,13 +203,10 @@ class TestHole(object):
 
     def test_output_level(self, tmpdir):
         with tmpdir.as_cwd():
-            with pytest.warns(UserWarning) as rec:
+            with pytest.warns(UserWarning, match="needs to be < 3"):
                 profiles = hole2.hole(self.filename,
                                       random_seed=self.random_seed,
                                       output_level=100)
-            assert len(rec) == 1
-            assert 'needs to be < 3' in rec[0].message.args[0]
-            # no profiles
             assert len(profiles) == 0
 
     def test_keep_files(self, tmpdir):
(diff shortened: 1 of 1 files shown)


IAlibay:
536 extra errors is problematic, we need to get to the bottom of that first?

I can't imagine it being healthy for users to have a wall of warnings show up?

lilyminium:
@IAlibay I think it's because we get the bonds for each atom. There's 264 atoms, and the DeprecationWarning in #3160 is raised each time we get the bonds and it's not an empty group. That'll be a pain for writing PDB files out, I guess -- although I think the default treatment of multiple warnings is to only print out the first instance.

IAlibay:
> @IAlibay I think it's because we get the bonds for each atom. There's 264 atoms, and the DeprecationWarning in #3160 is raised each time we get the bonds and it's not an empty group. That'll be a pain for writing PDB files out, I guess -- although I think the default treatment of multiple warnings is to only print out the first instance.

Yikes :( does this incur an additional performance cost?

lilyminium:
> Yikes :( does this incur an additional performance cost?

Incrementally more than the cost associated with getting the bonds atom-by-atom. I could probably update the writer to just call bonds once.

IAlibay:
If it's a tiny cost increase I have no issues, I think it's worth pinging @jbarnoud and @orbeckst who might have stronger views on warnings & performance.

orbeckst (approved):
I am happy with _this_ PR; checking exactly what you're looking for than just checking numbers of warnings is better in any case.

The root cause issue of having all these additional warnings requires another issue. Let's keep things simple.

orbeckst:
@IAlibay please merge when you think it's ready.

IAlibay:
> I am happy with _this_ PR; checking exactly what you're looking for than just checking numbers of warnings is better in any case.
> 
> The root cause issue of having all these additional warnings requires another issue. Let's keep things simple.

Ok let's go with this then.

We should think of a way to test how many warnings get emitted during our tests (i.e. something like codecov for warnings), that way we can tell just how much of an impact adding warnings have.

PR #3165, opened 2021-03-14
Adds test for issue 2670 [for develop]

### LINKED ISSUE #2670: capped distance fails for very small (<0.1A) boxes

``` python
import MDAnalysis as mda

boxsize = 0.01
box = np.array([boxsize, boxsize, boxsize, 60., 60., 60], dtype=np.float32)

u = mda.Universe(PDB)
u.dimensions = box

mda.lib.nsgrid.FastNS(boxsize / 10., u.atoms.positions, box=u.dimensions)
# this hangs the interpreter then crashes
```

Doesn't seem to fail when boxsize is 0.1, so not likely to actually impact anyone.

v0.21 dev

- Which operating system?

IAlibay:
Fixes #2670

 - Adds the same test being added to `master`
 
Note:
 - This should fail at the moment, but it's a good reminder to make sure that we merge the new nsgrid code into `develop`.

PR Checklist
------------
 - [x] Tests?
 - Docs?
 - CHANGELOG updated?

### DIFF
--- a/testsuite/MDAnalysisTests/lib/test_nsgrid.py
+++ b/testsuite/MDAnalysisTests/lib/test_nsgrid.py
@@ -272,3 +272,19 @@ def test_around_overlapping():
     #                                         box=u.dimensions)
     # assert np.count_nonzero(np.any(dist <= 0.0, axis=0)) == 48
     assert u.select_atoms('around 0.0 index 0:11').n_atoms == 48
+
+
+def test_issue_2670():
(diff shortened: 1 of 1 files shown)


richardjgowers:
I think because this causes a segfault it won't "fail" but just explode horribly.

IAlibay:
There's been a lot of changes to what's going on in NSGrid, and I'm not sure this is still relevant. I'm going to close this placeholder, let's just remember to make sure we forwardport the NSGrid fixes.

Issue #3031, opened 2020-11-18
Remove support for ts as an input to the ParmedConverter


IAlibay:
## Context

As discussed in https://github.com/MDAnalysis/mdanalysis/pull/3029#issuecomment-729047942 the ParmedConverter currently takes either an atomgroup/universe or timestep as an input. However as of #2754, support for passing timestep objects to writers was removed.

Given this, it might make sense to remove support for `ts`?

Maybe we should deprecate for 1.0.x and remove for 2.0?

avitomar12:
Hey @IAlibay  Is it possible to give reference to the code for this issue?

IAlibay:
@avitomar12 we are referring to: https://github.com/MDAnalysis/mdanalysis/blob/6282385644d672cf7023584347e35048935ab429/package/MDAnalysis/coordinates/ParmEd.py#L137

Particularly the class' ability to take timestep objects as an input.

@lilyminium should we try to get the deprecation warning through in 1.1.0 for this?

lilyminium:
I don't know how many people actually use this code instead of just the convert_to function. Did we have a deprecation warning when we removed ts from the other writers?

IAlibay:
Yup: https://github.com/MDAnalysis/mdanalysis/blob/4365f3b07c1bf2ebcf16424b26162102954c5b90/package/MDAnalysis/coordinates/base.py#L2225-L2230

lilyminium:
Okies, sounds good. I'll do that and the chainID deprecation.

kitsiosvas:
Hi, @IAlibay is this issue taken? I would like to work on this if not.

IAlibay:
@kitsiosvas as far as I am aware there are no PRs open on for this issue. Whilst @avitomar12 indicated they may want to work on it ten days ago, MDAnalysis has decided to take a PR-first approach (i.e. the first PR open is the one we will review, see: https://github.com/MDAnalysis/mdanalysis/wiki/GSoC-FAQ#can-i-claim-an-issue-to-work-on). So please do feel free to open a PR addressing the issue.

IAlibay:
This was taken care of for 1.1. Needs forward porting.

kitsiosvas:
@IAlibay Apologies for my delay on working for this issue. I want to inform you that I will not be able to work on this. If anyone else wants to take it, I haven't opened a PR yet so it's free. Again, apologies.

IAlibay:
[...]

PR #3172, opened 2021-03-16
Deprecate ParmedConverter taking Timestep objects

lilyminium:
~Fixes #~
Relates to #3031 

 - Fixed ParmedConverter reading Timestep objects, at least the AttributeErrors
 - Deprecated ParmedConverter reading Timestep objects

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/coordinates/ParmEd.py
```diff
@@ -211,9 +215,21 @@ def convert(self, obj):
         atom_kwargs = []
         for atom, name, resname, xyz, vel in zip(ag_or_ts, names, resnames, 
                                                  positions, velocities):
-            akwargs = {'name': name}
-            chain_seg = {'segid': atom.segid}
-            for attrname in ('mass', 'charge', 'type',
+            try:
```
IAlibay, line 218: I'll be honest, I'm not sure why this is being added here, was `ts` just not working to begin with? If so, do we even need to deprecate it (there's 0 end-user impact except from us removing a dead code path right?)

lilyminium: But I want it to work (or at least not die) in 1.1 :-(

IAlibay: 🙃 we're going to need to test that the outcome makes sense then.. or you can burn it with fire.

lilyminium: Too much work for something that's a little silly (making a template molecule with atoms called X, masses of 0, resids of 1...) and will be removed next release anyway.

package/MDAnalysis/coordinates/ParmEd.py
```diff
@@ -211,9 +212,21 @@ def convert(self, obj):
...
```
IAlibay, line 215: Are these changes necessary if we're killing `ts`?

package/MDAnalysis/coordinates/ParmEd.py
```diff
@@ -174,9 +174,10 @@ def convert(self, obj):
             ag_or_ts = obj.atoms
         except AttributeError:
             if isinstance(obj, base.Timestep):
-                ag_or_ts = obj.copy()
+                raise ValueError("Writing Timesteps to ParmEd "
+                                 "objects is not supported")
             else:
-                raise_from(TypeError("No Timestep found in obj argument"), None)
+                raise_from(TypeError("No atoms found in obj argument"), None)
```
IAlibay, line 180: Can you add a test to cover this TypeError please?

IAlibay (changes requested):
Couple of minor things

package/MDAnalysis/coordinates/ParmEd.py
lilyminium: I guess I'm anticipating changes in API that don't guarantee resids/masses/whatever, but we can cross that bridge when we come to it. I'll revert it.

package/MDAnalysis/coordinates/ParmEd.py
```diff
@@ -174,9 +174,10 @@ def convert(self, obj):
             ag_or_ts = obj.atoms
         except AttributeError:
             if isinstance(obj, base.Timestep):
-                ag_or_ts = obj.copy()
+                raise ValueError("Writing Timesteps to ParmEd "
+                                 "objects is not supported")
             else:
-                raise_from(TypeError("No Timestep found in obj argument"), None)
+                raise TypeError("No atoms found in obj argument") from None
```
IAlibay, line 180: raise_from is needed here unfortunately - py2.7 things

IAlibay (changes requested):
Still gotta live that py2.7 life :(

package/MDAnalysis/coordinates/ParmEd.py
lilyminium: 🐍🐍🐍🐍🐍🐍🐍

IAlibay (approved):

PR #3173, opened 2021-03-16
Deprecate using segid for chainid

lilyminium:
~Fixes #~

Relates to #3144

 - Deprecates using segid for chainid

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -30,6 +30,8 @@ Changes
   * Maximum pinned versions in setup.py removed for python 3.6+ (PR #3139)
 
 Deprecations
+  * Deprecated using the last letter of the segid as the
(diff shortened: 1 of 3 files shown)


IAlibay (approved):
lgtm but it's going to be super noisy (can't think of a better way though).

@MDAnalysis/coredevs any objections on the extra warning noise here?

orbeckst:
LGTM to me too and I don’t think super-noisy is a problem. If you’re writing PDB files and you keep using 1.1 then sooner or later you’ll likely figure out how to silence all these deprecation warnings... because you’ll have them for a very, very long time (if you don’t switch to 2.x for whatever reasons).

PR #3177, opened 2021-03-16
Re-enable mamba in CI

### LINKED ISSUE #3048: Enable mamba in github actions CI
As discussed in https://github.com/MDAnalysis/mdanalysis/pull/3040#discussion_r532887435 we currently seem unable to enable mamba in gh actions CI as it leads to strange `pip` segfaults.

Considering this was possible in Travis, I assume that there might be a solution here (possibly pinning to a given mamba version).

I'm raising this issue so that we can look into this more after #3040 is merged.

IAlibay:
Fixes #3048 

 - Enable mamba for GH actions CI

Testing on some cron jobs I'll push to MDA later, it looks like it works now (not sure if it's that stable, but I guess we can always disable again if it fails).

PR Checklist
------------
 - Tests?
 - Docs?
 - CHANGELOG updated?


IAlibay:
The gains here are modest (~ minute and a half).

Personally I think we should go ahead with it anyways, it adds up to ~ 10 mins of CPU time which is a non-significant environmental impact.

lilyminium (approved):
LGTM.
By the way, mdtraj has nuked conda/mamba for `nix`. That could be something we could look at if installation remains slow.

https://github.com/mdtraj/mdtraj/blob/master/.github/workflows/main.yaml

IAlibay:
Thanks @lilyminium I'm going to temporarily switch to draft though because I think my latest commit caused a performance regression.

> By the way, mdtraj has nuked conda/mamba for nix. That could be something we could look at if installation remains slow.

Yeah, for now we should be fine, we're sub 1 minute for dependency installs on linux. Annoyingly MacOS takes much longer.

orbeckst:
nix looks interesting and is available for Linux and macOS. 

Is it a disadvantage to recommend, say, conda for users but use another package manager for the bulk of our testing?

IAlibay:
Alright, I think this is as optimised as I can get it.

Main changes are:
- addition of mamba in places where we do lots of dependency installs
- setting the number of macOS processors to 3 instead of 2 (this doesn't have a noticeable influence since the performance differs a lot between runs, but in some cases it can make it run a bit faster)
- Switching the channel priorities around (which is something that's being introduced in the chemfiles PR)

## Quick timing summary:

### Linux:

#### full: ~ 2 mins less compared to current develop
miniconda setup: ~ 1 min
deps install: ~ 1 min

#### min: ~ 1 min less compared to current develop
miniconda_setup: ~ 1 min
deps install: ~ 20s

#### build_docs: ~ 3 mins less compared to current develop
setup_miniconda: ~ 1 min
install_deps: ~ 1 min

#### pylint: ~ same as current develop
setup_miniconda: 46s
deps install: 5s

#### pypi_check: same as current develop
setup_miniconda: 32s
deps install: 10s

### macOS:

#### full: for setup ~ 7 mins less compared to current develop
setup_miniconda: 1m 9s
install_deps: 1m 43s
run_tests: ~ 15 - 25 mins (same as current develop? sometimes faster?)

lilyminium (approved):
LGTM. Thanks for commenting why it's a conda install in pypi_check.

PR #3178, opened 2021-03-18
Removes deprecated sklearn arguments from encore

### LINKED ISSUE #2986: ENCORE uses deprecated features of sklearn
## Expected behavior ##

Running the tests throws no warning, and the library does not use deprecated features of the libraries it depends on.

## Actual behavior ##

Running the tests warns about ENCORE using deprecated features of scikit-learn:

```
MDAnalysisTests/analysis/test_encore.py: 13 warnings
  /home/jon/miniconda3/envs/py38/lib/python3.8/site-packages/sklearn/cluster/_kmeans.py:932: FutureWarning: 'precompute_distances' was deprecated in version 0.23 and will be removed in 0.25. It has no effect
    warnings.warn("'precompute_distances' was deprecated in version "

MDAnalysisTests/analysis/test_encore.py: 13 warnings
  /home/jon/miniconda3/envs/py38/lib/python3.8/site-packages/sklearn/cluster/_kmeans.py:938: FutureWarning: 'n_jobs' was deprecated in version 0.23 and will be removed in 0.25.
    warnings.warn("'n_jobs' was deprecated in version 0.23 and will be"

MDAnalysisTests/analysis/test_encore.py::TestEncoreClustering::test_clustering_AffinityPropagation_direct
MDAnalysisTests/analysis/test_encore.py::TestEncoreClustering::test_clustering_two_different_methods
MDAnalysisTests/analysis/test_encore.py::TestEncoreClustering::test_sklearn_affinity_propagation
  /home/jon/miniconda3/envs/py38/lib/python3.8/site-packages/sklearn/cluster/_affinity_propagation.py:146: FutureWarning: 'random_state' has been introduced in 0.23. It will be set to None starting from 0.25 which means that results will differ at every function call. Set 'random_state' to None to silence this warning, or to 0 to keep the behavior of versions <0.23.
    warnings.warn(("'random_state' has been introduced in 0.23. "
```

`pytest analysis/test_encore.py`

thadanipaarth:
Fixes #2986 

Removed depreciated parameters 'n_jobs' and 'precompute_distances' of sklearn.cluster.KMeans and Added new parameter 'random_state' of sklearn.cluster.AffinityPropagation

PR Checklist
------------
 - [ ] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/analysis/encore/clustering/ClusteringMethod.py
```diff
@@ -170,7 +170,7 @@ class AffinityPropagation(ClusteringMethod):
 
         def __init__(self,
                      damping=0.9, preference=-1.0,
-                     max_iter=500, convergence_iter=50,
+                     max_iter=500, convergence_iter=50, random_state=0,
```
lilyminium, line 173: Why is the default random state 0? [The scikit-learn default is "warn".](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.AffinityPropagation.html)

package/MDAnalysis/analysis/encore/clustering/ClusteringMethod.py
```diff
@@ -195,6 +195,13 @@ def __init__(self,
                 (default is 50). Parameter in the Affinity Propagation for
                 clustering.
 
+            random_state : int, RandomState instance or None, optional
+                Used to control the starting state(default is 0). If an
+                integer is given, it fixes the seed for the new random number
+                generator. Use of integer is recommended to mimic the function
+                call. None uses the global random state instance from
```
lilyminium, line 202: Given that we already pass ``kwargs`` into AffinityProgation, I can only see giving a default to `random_state` causing trouble for us in the future. In scikit-learn 0.23, the default setting "warn" warns us that the default will change from "warn" to `None` in 1.0. 

https://github.com/scikit-learn/scikit-learn/blob/95119c13af77c76e150b753485c662b7c52a41a2/sklearn/cluster/_affinity_propagation.py#L147

Instead, it would be more useful to document `**kwargs`.

package/CHANGELOG
```diff
@@ -16,11 +16,14 @@ The rules for this file:
 ??/??/?? tylerjereddy, richardjgowers, IAlibay, hmacdope, orbeckst, cbouy,
          lilyminium, daveminh, jbarnoud, yuxuanzhuang, VOD555, ianmkenney,
          calcraven，xiki-tempula, mieczyslaw, manuel.nuno.melo, PicoCentauri,
-         hanatok, rmeli, aditya-kamath, tirkarthi
+         hanatok, rmeli, aditya-kamath, tirkarthi, thadanipaarth
 
   * 2.0.0
 
 Fixes
+  * Removed depreciated parameters 'n_jobs' and 'precompute_distances' of
```
lilyminium, line 24:
```suggestion
  * Removed deprecated parameters `n_jobs` and `precompute_distances` of
```

lilyminium (changes requested):
Hi @thadanipaarth, thanks for looking at this issue. I'm not sure you need to add the `random_state` parameter, as the scikit-learn API is still in flux. I don't think PR needs a test, but I'll defer to the opinions of other @MDAnalysis/gsoc-mentors?

As this is your first contribution, please add yourself to AUTHORS. :-)

thadanipaarth:
> Hi @thadanipaarth, thanks for looking at this issue. I'm not sure you need to add the `random_state` parameter, as the scikit-learn API is still in flux. I don't think PR needs a test, but I'll defer to the opinions of other @MDAnalysis/gsoc-mentors?
> 
> As this is your first contribution, please add yourself to AUTHORS. :-)

Hi @lilyminium, thanks for reviewing. I had misinterpreted the PR Checklist, this PR will not need a test. I am working on the other changes and will update it as soon as possible.

package/CHANGELOG
thadanipaarth: Done!

package/MDAnalysis/analysis/encore/clustering/ClusteringMethod.py
thadanipaarth: If the default state is set as "warn", a warning will be generated and the random_state variable will be set to 0(https://github.com/scikit-learn/scikit-learn/blob/95119c13af77c76e150b753485c662b7c52a41a2/sklearn/cluster/_affinity_propagation.py#L147). Hence, it will not make a difference to the best of my understanding.

The primary purpose of setting the default value to 0 is to mimic the function's behaviour before adding the random_state parameter.

lilyminium: Perhaps consider the purpose of the warning.

> Hence, it will not make a difference to the best of my understanding.
> 
> The primary purpose of setting the default value to 0 is to mimic the function's behaviour before adding the random_state parameter.

Was the warning emitted before adding the random_state parameter?

thadanipaarth: No, the warning was not emitted before adding the random_state parameter, i.e. Scikit-Learn 0.23. If we go through the warning, it only acknowledges the new parameter `random_state` and specifies its future default value. 

Now to remove the warning(as mentioned in the Issue), we can either pass a constant `random_state` parameter value or add a `random_state` parameter in the class constructor and set the default value explicitly to 0(to mimic the behaviour of previous versions) or None(In this case, calling the function multiple times will produce different results). Please suggest.

package/MDAnalysis/analysis/encore/clustering/ClusteringMethod.py
thadanipaarth: If we go with the above approach and include the `random_state` parameter in `**kwargs`, and do not provide any `random_state` default value. We can only solve this part of the issue by providing a constant value to `random_state` parameter for the `sklearn.cluster.AffinityPropagation()` function call.

package/MDAnalysis/analysis/encore/clustering/ClusteringMethod.py
lilyminium: I think it's best to get the original encore developers' opinions here: cc @mtiberti, @wouterboomsma 

In scikit-learn v0.23 (the latest one) the `random_state` keyword was introduced to control the seed of the random number generator in AffinityPropagation. Up until now the seed was hardcoded as 0. In the yet-to-be released scikit-learn 1.0, the seed defaults to `None`, generating a random initial seed. The current default value of `random_state` in 0.23 is "warn", which warns the user of this upcoming change and then sets the seed to 0. By using a default value of 0, we would be removing this warning.

As sklearn has historically used a seed of 0, Encore's sklearn clustering has historically used 0. With this change in scikit-learn API, we need to now consider whether `encore` should also change.

IMO given that Encore already implements its own C version of AffinityPropagation as default, people who choose to use the sklearn version are doing so because they like sklearn and expect sklearn defaults. However, people may have come to expect certain results from the way we wrap scikit-learn. What do you think, @mtiberti, @wouterboomsma (and to some extent all @MDAnalysis/coredevs ?)

> Now to remove the warning(as mentioned in the Issue)

I see now that this is in the original issue. Correct me if you disagree @jbarnoud, but I don't see this warning as urgently requiring removal; `random_state` is not actually a deprecated feature but just a warning generated due to transitional API by scikit-learn. That is to say, it is not MDAnalysis being backwards that is generating this warning, but scikit-learn deeming it important to warn people about this change.

package/MDAnalysis/analysis/encore/clustering/ClusteringMethod.py
lilyminium: @thadanipaarth I elaborated above, but as you can tell by the length of the comment, the value of `random_state` is a somewhat important choice. Whichever way we choose to go on the default value, I think the documentation will have to be more specific. For example: "starting state" of what? Is the random number generator "new", or has scikit-learn just always hardcoded the seed? Would the use of *any* integer mimic the historical performance of Encore, or just 0? And which function call does the value of 0 mimic? These are all questions that users may have, and that we should be able to answer or at least link to the scikit-learn documentation for more information.

thadanipaarth: Yes, I completely got the point. I will update the documentation in detail once we decide on the final approach.

package/MDAnalysis/analysis/encore/clustering/ClusteringMethod.py
lilyminium: Pinging @MDAnalysis/coredevs @mtiberti @wouterboomsma just because the GSOC student deadline is in 10 days

thadanipaarth: Eagerly waiting for your response!

orbeckst: I would go with https://github.com/MDAnalysis/mdanalysis/pull/3178#discussion_r603497072
> the way that I would solve the warnings in tests, and also to make tests pass, is to seed it with 0 in the tests.

For the rest of the code, I'd eventually go with whatever scikit-learn does.

thadanipaarth: @lilyminium I have made the required changes. Please have a look at it.

package/MDAnalysis/analysis/encore/clustering/ClusteringMethod.py
```diff
@@ -195,6 +195,14 @@ def __init__(self,
...
                 clustering.
 
+            random_state : int, RandomState instance or None, optional
+		Used to control the random number generator used, for the
+		randomization which is part of scikit-learn algorithm(default
+		is scikit-learn's default).Use an int(unique) for reproducible
+		results across function calls. Previously(version<0.23), this
+		parameter was hardcoded to 0 by scikit-learn. Refer its
+		documentation for more details.
+
```
lilyminium, line 205: Please don't mix tabs and spaces. This is detailed documentation, but I'm not sure it's necessary to be documented -- you could just document `kwargs` as being sent to scikit.

package/CHANGELOG
```diff
@@ -17,11 +17,14 @@ The rules for this file:
...
          calcraven，xiki-tempula, mieczyslaw, manuel.nuno.melo, PicoCentauri,
          hanatok, rmeli, aditya-kamath, tirkarthi, LeonardoBarneschi, hejamu,
-         biogen98, orioncohen
+         biogen98, orioncohen, thadanipaarth
 
   * 2.0.0
 
 Fixes
+  * Removed deprecated parameters `n_jobs` and `precompute_distances` of
+    sklearn.cluster.KMeans and Added new parameter `random_state` of
```
lilyminium, line 26: Could you please now remove this from the CHANGELOG?

lilyminium (changes requested):
Thanks @thadanipaarth, you're basically done! Could you please just remove the new `random_state` documentation and fix the PEP8 errors noted by the bot? https://github.com/MDAnalysis/mdanalysis/pull/3178#issuecomment-802000919

package/CHANGELOG
thadanipaarth: Done!

package/MDAnalysis/analysis/encore/clustering/ClusteringMethod.py
thadanipaarth: Done! I will take care regarding this from next time :)

thadanipaarth:
Hi @lilyminium, I noticed that one test occasionally fails after today's changes (Failing test-test_rdkit.py::TestRDKitConverter::test_assign_coordinates). I am sure that this not due to the changes made in this PR.

package/MDAnalysis/analysis/encore/clustering/ClusteringMethod.py
```diff
@@ -195,6 +195,8 @@ def __init__(self,
                 (default is 50). Parameter in the Affinity Propagation for
                 clustering.
 
+            **kwargs being sent to scikit-learn.
```
lilyminium, line 198:
```suggestion
            **kwargs being sent to scikit-learn.
            **kwargs: Other keyword arguments are passed to :class:`sklearn.cluster.AffinityPropagation`.
```

The "sent to scikit-learn" was more casual phrasing. Something like the above is more suitable.

lilyminium (changes requested):
Yeah that RDKit error has nothing to do with this PR @thadanipaarth -- it's a [known bug](https://github.com/MDAnalysis/mdanalysis/issues/2958), unfortunately.

package/MDAnalysis/analysis/encore/clustering/ClusteringMethod.py
thadanipaarth: Yeah this would be better.

package/MDAnalysis/analysis/encore/clustering/ClusteringMethod.py
```diff
@@ -195,6 +195,9 @@ def __init__(self,
                 (default is 50). Parameter in the Affinity Propagation for
                 clustering.
 
+            **kwargs : Other keyword arguments are passed to : class:
```
lilyminium, line 198:
```suggestion
            **kwargs (optional): Other keyword arguments are passed to :class:
```

You can't have a space in the `:class:` tag for Sphinx.


Anyway, oops, that didn't turn out looking so good. https://mdanalysis--3178.org.readthedocs.build/en/3178/documentation_pages/analysis/encore/clustering.html?highlight=affinitypropagation#MDAnalysis.analysis.encore.clustering.ClusteringMethod.AffinityPropagation

thadanipaarth: The above changes were not producing the desired result. I have made the changes according to me. Please have a look at it! In my opinion, it looks good. 

Docs build: https://mdanalysis--3178.org.readthedocs.build/en/3178/documentation_pages/analysis/encore/clustering.html

lilyminium: Thanks, that looks much better.

lilyminium (approved):

lilyminium:
Thank you @thadanipaarth, congratulations on your first commit to MDAnalysis! :D

Issue #3181, opened 2021-03-21
#2345 not fixed in NSGrid


zemanj:
## Expected behavior ##

The grid-based distance search in `NSGrid` works for an fcc crystal in a triclinic box.

## Actual behavior ##

When calculating coordination numbers for all atoms in the correctly reproduced fcc-system from #2345, the NSGrid-based result is wrong. The test `testsuite/MDAnalysisTests/lib/test_nsgrid.py::test_issue_2670` passes only because it loads the system from a pdb file, which lacks the coordinate precision required to reproduce the issue.

## Code to reproduce the behavior ##

``` python
from collections import defaultdict, Counter
from numpy.testing import assert_allclose
import numpy as np
import MDAnalysis as mda
from MDAnalysisTests.datafiles import SURFACE_PDB, SURFACE_TRR
from MDAnalysis.lib.distances import distance_array, apply_PBC
from MDAnalysis.lib.mdamath import triclinic_box, triclinic_vectors
from MDAnalysis.lib.nsgrid import FastNS
from ase.build import fcc111

#search radius:
cutoff = 2.9

print("selected cutoff: {}".format(cutoff))

#generate triclinic fcc slab system with ASE:
ase_system = fcc111('Ag', size=(5, 5, 4), vacuum=10.0)
ase_box = triclinic_box(*ase_system.cell)
ase_positions = ase_system.positions.astype(np.float32)

#load similar system from corresponding PDB file:
u = mda.Universe(SURFACE_PDB)
pdb_box = u.dimensions
pdb_positions = u.atoms.positions


# Assert that systems are almost equal. PDB rounds to 3 decimal places so that
# the difference must not be larger than 0.0005:
assert_allclose(ase_box, pdb_box, atol=0.0005)
assert_allclose(ase_positions, pdb_positions, atol=0.0005)

#print boxes:
print()
print("ASE box: [{}]"
      "".format(" ".join(["{:15.10f}"] * len(ase_box))).format(*ase_box))
print("PDB box: [{}]"
      "".format(" ".join(["{:15.10f}"] * len(pdb_box))).format(*pdb_box))

#ASE bruteforce distance search:
ase_da = distance_array(ase_positions, ase_positions, box=ase_box)
ase_bf_cn = []
for i in range(ase_da.shape[0]):
    ase_bf_cn.append(len([ix for ix in range(ase_da.shape[1])
                 if 0.0 < ase_da[i, ix] < cutoff]))
[...]

PR #3182, opened 2021-03-21
fixed NSGrid bug #2345

### LINKED ISSUE #2345: `FastNS` yields wrong results on simple test case

zemanj:
Fixes #2345 #3181

 - Replaced coordinate wrapper `nsgrid.pyx::coordintoprimarycell()` by `calc_distances.h::_ortho_pbc()` and `calc_distances.h::_triclinic_pbc()`
 - Added full-precision coordinates for the test case reproducing #2345 `test_nsgrid.py::test_issue_2345()`
 - Fixed test case `test_nsgrid.py::test_issue_2670()` which only used to work because of the inexact coordinate wrapping in `nsgrid.pyx::coordintoprimarycell()`

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/lib/nsgrid.pyx
```diff
@@ -548,14 +533,21 @@ cdef class FastNS(object):
             raise ValueError("search_coords must have a shape of (n, 3), got "
                              "{}.".format(search_coords.shape))
 
+        tmpcoords = search_coords.copy()
```
richardjgowers, line 536: I'm not a huge fan of doing a numpy copy here. Can we keep the wrapping of individual positions inside the loop (and call with n=1)?

package/MDAnalysis/lib/nsgrid.pyx
```diff
@@ -424,32 +429,11 @@ cdef class FastNS(object):
...
-            coord[0] += self.triclinic_dimensions[XY]
-
-        adj_cx = coord[2] * self.cellsize[XZ] + coord[1] * self.cellsize[XY]
-        while ((coord[0] - adj_cx) >= self.triclinic_dimensions[XX]):
-            coord[0] -= self.triclinic_dimensions[XX]
-        while ((coord[0] - adj_cx) < 0):
-            coord[0] += self.triclinic_dimensions[XX]
+        # Make sure cell coordinate indices are within the primary unit cell
+        # (better safe than sorry):
+        xyz[0] %= self.ncells[0]
```
richardjgowers, line 434: It should be possible to not need this right? I think the wrapping we're importing has a slightly different style so the above code requires this.

richardjgowers (changes requested):

richardjgowers (approved):

package/MDAnalysis/lib/nsgrid.pyx
zemanj: Yeah I thought about this, too. I did this because I felt a bit uncomfortable with having a function call in a loop, and I thought that an O(n)  memory footprint should be affordable. However, if the function call gets inlined by the compiler, copying the array will certainly be slower. So, in the end, you are probably right but we'll have to measure the difference to be sure.

package/MDAnalysis/lib/nsgrid.pyx
zemanj: I thought this through and the answer is yes and no:
* I wrote the code in `_ortho_pbc()` so that folded coordinates  will always lie within the simulation box so that the condition  is numerically *guaranteed* to be fulfilled (you probably remember because you reviewed the respective PR back then). So, yes, for orthorhombic systems the modulo operations are entirely redundant.
* I had the same in mind at the time I implemented the corresponding operations in `_triclinic_pbc()`, but it turns out that for triclinic boxes, the same guarantee cannot be given for all dimensions simultaneously (at least I think not without seriously impacting performance). The reason for this is that in systems with more than one non-orthogonal box angle, shifting coordinates by one box vector can result in changing more than one dimension of these coordinates. Thus, it can happen that you end up in a situation where eliminating the error in one dimension re-introduces the same problem in another dimension, and vice versa. Nevertheless, the current implementation of `_triclinic_pbc()` at least minimizes this error and guarantees to eliminate it in at least one dimension. Anyway, in the end, the modulo operations are still required!

zemanj: I should probably change the code comment from "better safe than sorry" to "this is strictly required for triclinic boxes".

package/MDAnalysis/lib/nsgrid.pyx
zemanj: Ah, there's one more thing: Distinguishing between orthorhombic and general triclinic coordinate folding potentially makes the code a bit faster, and I wanted to avoid checking for `self.triclinic` in the loop. For large systems, branch prediction will most likely eliminate the impact, but it might make function inlining less likely (that's purely speculative, though).

zemanj: BTW I think this is not a numpy copy operation but a to-C-contiguous [Memoryview copy](https://cython.readthedocs.io/en/latest/src/userguide/memoryviews.html#c-and-fortran-contiguous-copies). Don't know if there's a huge difference since both operations require the GIL. Using `tmpcoords[:, :] = search_coords` does not require the GIL but will throw a `ValueError` at runtime if `search_coords` is non-contiguous, so that's not an option.

richardjgowers: It's more I'd rather have 3 floats on the stack than O(n) on the heap

package/MDAnalysis/lib/nsgrid.pyx
richardjgowers: Not sure I follow here. We represent triclinic boxes as a 3x3 matrix which is upper triangular sparse, i.e. `[lx,0,0], [xy,ly,0], [xz,yz,lz]`.  If we do z first, then the remaining 2 dimensions can't affect z, so `xyz[2]` is fixed and won't be affected by `xyz[0/1]`.  Were you getting failures without the modulo operations?

zemanj: I got failures without modulo operations when using `coordintoprimarycell()` because with that function it definitely happens that coordinates can lie outside of the box. With `_triclinic_pbc()` I did not observe any failures without the modulo operations but I left them in as a precaution. After all, they won't affect the correctness of the results and their performance impact is currently not even measurable. I'll have to check the code of `_triclinic_pbc()` once again before I can give a definite statement on whether the modulos are needed.

package/MDAnalysis/lib/nsgrid.pyx
zemanj: Is it because you want to avoid the entailed syscalls (and later garbage collection) or only because of the memory footprint? I don't expect the latter to be of any concern. From a performance point of view it doesn't make any difference.
Anyway, I went ahead and tested different scenarios with triclinic and orthorhombic boxes for the ADK protein system and compared in- and out-of-loop wrapping. The current approach seems fastest, but within error bars they're all the same speed. So it seems reasonable to go back to the single-coordinate in-loop wrapping.

The bottleneck is in fact the minimum image distance computation:
image
Perf log for a triclinic box. The minimum image computation is responsible for basically 100% of the `FastNS.search()` runtime.

image
Same for orthorhombic boxes, the minimum image computation causes almost 100% of the `FastNS.search()` runtime.

We should consider a lazy minimum image implementation leveraging the fact that we're always dealing with wrapped coordinates.

package/MDAnalysis/lib/nsgrid.pyx
zemanj: Ok I got your point, and it's exactly what we're doing in `_triclinic_pbc()`. You're right that by working our way up through the matrix from the z-row to the x-row, operations involving the y-row won't affect the z-component and operations involving the x-row won't affect the y- and z-components. From that point of view we should be safe here.
However, I'm still concerned because `_triclinic_pbc()` uses mixed-precision in the sense that the single-precision coordinates are checked to lie inside the box by comparing with the respective upper and lower bounds of the box, and that comparison is done in double precision. In the `FastNS` implementation, however, determining the cell xyz indices is done in single precision, which (I suppose) might lead to different results if coordinates are close to any of the box surfaces.
So I'd still opt for leaving the modulo operations in there as they really won't hurt but instead guarantee that the algorithm works correctly.

package/MDAnalysis/lib/nsgrid.pyx
zemanj: I just tried out a lazy orthorhombic minimum image distance computation and it yields a speedup of about 22 %.
In that scenario the in-loop wrapping is a tiny bit faster so that's the way to go.
So, thank you for insisting! :smile: :+1:

PR #3184, opened 2021-03-21
Create Universe from empty RDKit ROMol

LeonardoBarneschi:
Fixes #
A Universe created from an ROMol with no atoms returns now a Universe with 0 atoms

 - included a try/except block in  RDKitParser.py
 - included corresponding test in test_universe.py

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/topology/RDKitParser.py
+++ b/package/MDAnalysis/topology/RDKitParser.py
@@ -179,7 +179,7 @@ class RDKitParser(TopologyReaderBase):
 
         try:
             atom = mol.GetAtomWithIdx(0)
-        except:
+        except RuntimeError:
             top = Topology(n_atoms=0, n_res=0, n_seg=0,
                            attrs=None,
                            atom_resindex=None,
--- a/package/MDAnalysis/topology/RDKitParser.py
+++ b/package/MDAnalysis/topology/RDKitParser.py
@@ -177,8 +177,16 @@ class RDKitParser(TopologyReaderBase):
         occupancies = []
         tempfactors = []
 
+        try:
+            atom = mol.GetAtomWithIdx(0)
+        except:
+            top = Topology(n_atoms=0, n_res=0, n_seg=0,
+                           attrs=None,
+                           atom_resindex=None,
+                           residue_segindex=None)
+            return top
+
         # check if multiple charges present
-        atom = mol.GetAtomWithIdx(0)
         if atom.HasProp('_GasteigerCharge') and (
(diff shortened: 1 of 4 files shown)


package/MDAnalysis/topology/RDKitParser.py
```diff
@@ -177,8 +177,16 @@ def parse(self, **kwargs):
         occupancies = []
         tempfactors = []
 
+        try:
+            atom = mol.GetAtomWithIdx(0)
+        except:
```
tylerjereddy: As the auto-linter suggests, you may want to be more specific here so that you don't accidentally intercept other important exceptions.

Looking at the matching issue, you may be able to use `except RuntimeError` to be more specific.

testsuite/MDAnalysisTests/core/test_universe.py
```diff
@@ -210,6 +210,10 @@ def test_universe_topology_class_with_coords(self):
         assert_equal(u.trajectory.n_frames, u2.trajectory.n_frames)
         assert u2._topology is u._topology
 
+    def test_universe_empty_ROMol(self):
+        Chem = pytest.importorskip("rdkit.Chem")
+        mol = Chem.Mol()
+        u = mda.Universe(mol, format="RDKIT")
```
tylerjereddy, line 216: I suppose you could also check that the universe is empty--has `0` atoms for example.

package/MDAnalysis/topology/RDKitParser.py
LeonardoBarneschi: Hi Tyler, I fixed that as soon as I have done the pull request

testsuite/MDAnalysisTests/core/test_universe.py
LeonardoBarneschi: Yes, I included the asserts also.

lilyminium (changes requested):
Thanks for looking at this, @LeonardoBarneschi. The fix and test looks good to me. However, could you please update CHANGELOG? As is your first time contributing to MDAnalysis, please also add yourself to AUTHORS.

lilyminium (approved):

lilyminium:
Thank you @LeonardoBarneschi! 🥳

LeonardoBarneschi:
More than happy to have contributed!!!!

Il sab 27 mar 2021, 18:09 Lily Wang ***@***.***> ha scritto:

PR #3188, opened 2021-03-24
Add sort method to AtomGroup

### LINKED ISSUE #2976: Add sort method to the atomgroup
I have an edgy case, where I have two small molecules, they share a common substructure. I want to superimpose them based on the common substructure. The difficulty is that the common substructure doesn't have the same order of atoms.

Given a `mapping` dictionary which maps atoms from molecule 1 (key) to molecule 2 (value), the superimposition is done like this, where I need to manually reorder the ordered atomgroup.

```python
from MDAnalysis.analysis.align import alignto
substructure_1 = []
substructure_2 = []
for id1, id2 in mapping.items():
    substructure_1.append(molecule_1.select_atoms('bynum {}'.format(id1))
    substructure_2.append(molecule_2.select_atoms('bynum {}'.format(id2))
substructure_1 = sum(substructure_1)
substructure_2 = sum(substructure_2)
alignto(substructure_1, substructure_2)

# move to the original rank
new_substructure_1 = []
for id in sorted(substructure_1.ids):
    new_substructure_1.append(substructure_1.select_atoms('bynum {}'.format(id))
substructure_1 = sum(new_substructure_1)
alignto(molecule_1, substructure_1, select='bynum {}'.format(' '.join(substructure_1.ids)))
```

So if the atomgroup has a sort method, I could then just sort the atomgroup with one line. `substructure_1.sort(key=lambda atom: atom.id)`

```python
from MDAnalysis.analysis.align import alignto
substructure_1 = []
substructure_2 = []
for id1, id2 in mapping.items():
    substructure_1.append(molecule_1.select_atoms('bynum {}'.format(id1))
    substructure_2.append(molecule_2.select_atoms('bynum {}'.format(id2))
substructure_1 = sum(substructure_1)
substructure_2 = sum(substructure_2)
alignto(substructure_1, substructure_2)

# move to the original rank
substructure_1.sort(key=lambda atom: atom.id)
alignto(molecule_1, substructure_1, select='bynum {}'.format(' '.join(substructure_1.ids)))
```

1ut:
Fixes #2976

Added sort method to the atomgroup.

- Parameters
        A function for the parameter of each atom to use for sorting. i.e. lambda atom: atom.id

- Returns
       Sorted atomgroup.

I'm new to mdanalysis so I would be grateful if you could inform me when you find any mistakes.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


richardjgowers (changes requested):
Hi @1ut thanks for starting on this.  We'll also need a test written in test_atomgroup.py to make sure that this code is working as it should.

package/MDAnalysis/core/groups.py
```diff
@@ -3324,6 +3324,56 @@ def write(self, filename=None, file_format=None,
...
+        .. versionadded:: 2.0.0
+        """
+        if len(keys) == 0:
+            raise NameError("argument 'key' is not defined. ")
+        for key, value in keys.items():
+            if key == "key":
+                n = len(self.atoms)
+                idx = np.zeros(n)
+                for i in range(n):
+                    idx[i] = value(self.atoms[i])
```
richardjgowers, line 3370: I don't like the pattern of providing a lambda that is applied to each atom.  Mostly because it will scale poorly and AtomGroups are often large.  I think a better approach would be to take the name of an attribute (eg `"ids"`, `"masses"`) and have a `getattr` call to build the `idx` array here.  This can then still use the argsort afterwards.

richardjgowers (changes requested):

1ut:
I've rewritten the code. I appreciate your kind suggestion.

package/CHANGELOG
```diff
@@ -133,6 +133,7 @@ Enhancements
     'protein' selection (#2751 PR #2755)
   * Added an RDKit converter that works for any input with all hydrogens
     explicit in the topology (Issue #2468, PR #2775)
+  * Added sort method to the atomgroup (Issue #2976)
```
IAlibay, line 136: Note these entries should be newest first.

Also please do add yourself to the author list for 2.0 (above) and AUTHORS.

IAlibay (changes requested):
Couple of quick thoughts.

testsuite/MDAnalysisTests/core/test_atomgroup.py
```diff
@@ -1674,3 +1675,25 @@ def test_partial_timestep(self, universe):
...
                                 self.prec,
                                 err_msg="Partial timestep coordinates wrong")
+
+
+class TestAtomGroupSort(object):
+    """Tests the AtomGroup.sort attribute"""
+
+    @pytest.fixture()
+    def u(self):
+        return mda.Universe(PDB_helix)
```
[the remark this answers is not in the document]
richardjgowers: generally using `Universe.empty()` is better for tests

testsuite/MDAnalysisTests/core/test_atomgroup.py
```diff
@@ -1674,3 +1675,25 @@ def test_partial_timestep(self, universe):
...
+
+    @pytest.fixture()
+    def ag(self, u):
+        return u.atoms[[3, 2, 1, 0]]
+
+    @pytest.fixture()
+    def agsort(self, ag):
+        return ag.sort("ids")
+
+    def test_sort(self, agsort):
```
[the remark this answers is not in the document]
hmacdope: @IAlibay sorry to jump the gun a bit here, but would we want to restrict to a list of `sortable` attributes or check there is no collision (scales badly)?

richardjgowers:
What would be unsortable?

testsuite/MDAnalysisTests/core/test_atomgroup.py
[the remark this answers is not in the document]
1ut: Add tests for TestAtomGroupSort for non-unique attribute using atomgroup from Universe.empty().

https://github.com/MDAnalysis/mdanalysis/pull/3188/commits/a535599aaafcbacec22b804ced24204bda3f3157

package/CHANGELOG
1ut: Added my name to author list for 2.0 and AUTHORS.

https://github.com/MDAnalysis/mdanalysis/pull/3188/commits/67bef3a3e1e2d76872dfae0b5b678d4802397877
https://github.com/MDAnalysis/mdanalysis/pull/3188/commits/aeca48da8de3be1a8570168792a9c2e937ee2c7f

1ut:
I assume no one wants to sort atomgroup by characters like names of residues. I'll add lines to check if all elements of the given array are numeric if needed.

IAlibay:
> I assume no one wants to sort atomgroup by characters like names of residues. I'll add lines to check if all elements of the given array are numeric if needed.

There might be a rare use case, but I agree it's probably not happening.

Imho it might be safer to only allow for cases where we can predict the behaviour. @richardjgowers thoughts on this? The alternative here is just to consider the atomgroup as a numpy array-esque class, and go with the philosophy that all we are doing is just replicating upstream numpy sort behaviour.

@hmacdope re - collisions, that's part of the reason I brought it up, i.e. can we see what happens in those cases (which is essentially the `masses` case). If we think it's too unpredictable to have collisions, we could restrict the attributes to unique ones (i.e. ix and ids), but personally I'm in favour of going "if it replicates numpy behaviour, then it's fine".

richardjgowers:
Nothing wrong with sorting strings? No need to complicate this; if the user
asks for sorted by resname then we can treat them like adults and sort by
resname rather than springing an error on them?

IAlibay:
> Nothing wrong with sorting strings? No need to complicate this; if the user asks for sorted by resname then we can treat them like adults and sort by resname rather than springing an error on them?

Makes sense, let's just go with a simple sort then, unless @hmacdope has a specific reason against it.

richardjgowers:
@1ut I've noticed you've opened a new PR #3195 - are you going to finish this or otherwise we can close this?

1ut:
@richardjgowers 
I didnt intend to give up this issue. Should I close PR #3195 ?

1ut:
I noticed that argsort doesn't work properly to sort atomgroup when given multiple dimensional array, so I made the argument "keyfunc" that receives a function to process multiple dimension array that the attribute returns and to return 1D array as a key for sorting.

1ut:
@IAlibay 
> Apologies, I'm likely missing something obvious here. Do we have a use case for sorting on non-1D arrays?
Atomgroup.positions returns 2D arrays. When given a key "positions", atomgroup.sort of previous code didn't raise any error and returned atomgroup with destructed .poisitions. 
> Similarly, idx (as set in your code), is a Universe attribute right? Is there a normal case for this being a 0 dimension numpy array?
No. So I removed the lines for checking if idx is 0 dimension array. 
> I agree that there should be a check for non-"number of atoms" sized attributes (off the top of my head, technically someone could try to pass something like "bonds" here and the whole thing would throw (at the moment) an AttributeError). We probably have to be a bit more general about our checking here.
I added line to check if number of atoms and idx have a same size.

package/MDAnalysis/core/groups.py
```diff
@@ -3324,6 +3324,66 @@ def write(self, filename=None, file_format=None,
...
+            >>> ag.ids
+            array([1 2 3 4])
+            >>> ag.ix
+            array([0 1 2 3])
+
+        .. versionadded:: 2.0.0
+        """
+        idx = getattr(self.atoms, key)
+        if len(idx) != len(self.atoms):
+            raise ValueError("The array returned by the attribute '{}' "
```
IAlibay, line 3367: Please add a test to cover this.

package/MDAnalysis/core/groups.py
```diff
@@ -3324,6 +3324,66 @@ def write(self, filename=None, file_format=None,
...
+                             "must have a same length as the number of "
+                             "atoms".format(key))
+        if idx.ndim == 1:
+            order = np.argsort(idx, kind='stable')
+        elif idx.ndim > 1:
```
IAlibay, line 3372: I'm not fully convinced this is in the scope of what we want to cover with this function, but I'll defer to @richardjgowers' opinion though.

package/MDAnalysis/core/groups.py
```diff
@@ -3324,6 +3324,66 @@ def write(self, filename=None, file_format=None,
...
+        Example
+        -------
+
+        .. code-block:: python
+
+            >>> import MDAnalysis as mda
+            >>> from MDAnalysisTests.datafiles import PDB_small
+            >>> u = mda.Universe(PDB_small)
+            >>> ag = u.atoms[[3, 2, 1, 0]]
+            >>> ag.ids
```
IAlibay, line 3353: In my opinion, the inclusion of both `ids` and `ix` in this docstring example is confusing for users. It feels like it hints that the sort method sorts by `ids` because that's the attribute you are showing off first. For the sake of simplicity, I would suggest just having `ag.ix`.

package/MDAnalysis/core/groups.py
```diff
@@ -3324,6 +3324,66 @@ def write(self, filename=None, file_format=None,
...
+
+        .. versionadded:: 2.0.0
+        """
+        idx = getattr(self.atoms, key)
+        if len(idx) != len(self.atoms):
+            raise ValueError("The array returned by the attribute '{}' "
+                             "must have a same length as the number of "
+                             "atoms".format(key))
+        if idx.ndim == 1:
+            order = np.argsort(idx, kind='stable')
```
IAlibay, line 3371: The `kind` type will need to be documented in the docstring, probably as a Note. Some advanced users might need to know what kind of sort behaviour they should be expecting.

testsuite/MDAnalysisTests/core/test_atomgroup.py
```diff
@@ -1674,3 +1674,70 @@ def test_partial_timestep(self, universe):
...
+    ]
+
+    @pytest.mark.parametrize("inputs, expected", test_data, ids=test_ids)
+    def test_sort(self, ag, inputs, expected):
+        agsort = ag.sort(inputs)
+        assert np.array_equal(expected, agsort.ix)
+
+    def test_sort_position(self, ag):
+        ag.positions = (-np.arange(21)).reshape(7, 3)
+        with pytest.raises(ValueError):
```
IAlibay, line 1735: Can we have some `match` arguments to raises here so we check that we are picking up the right error?

package/MDAnalysis/core/groups.py
```diff
@@ -3324,6 +3324,66 @@ def write(self, filename=None, file_format=None,
...
+                                "a proper function that returns 1D array "
+                                "as the attribute '{}' returns multiple "
+                                "dimension array.".format(key))
+            sortkeys = keyfunc(idx)
+            if sortkeys.ndim != 1:
+                raise ValueError("The function you assigned to the argument "
+                                 "'keyfunc':{} doesn't return 1D array."
+                                 .format(keyfunc))
+            order = np.argsort(sortkeys, kind='stable')
+        agsorted = self.atoms[order]
```
IAlibay, line 3384: nitpicking, but this seems a bit unnecessary if the next line is just to return this. It's probably understandable enough to just return `self.atoms[order]`?

IAlibay (changes requested):
Thanks @1ut, just a couple of comments to address.

package/MDAnalysis/core/groups.py
1ut: Added test for it. 
https://github.com/MDAnalysis/mdanalysis/pull/3188/commits/26b377323c113b0343517ac5473c573d381b60df

package/MDAnalysis/core/groups.py
1ut: Fixed.
https://github.com/MDAnalysis/mdanalysis/pull/3188/commits/a577298b4985b533233bc52a3a7bf75323055312

package/MDAnalysis/core/groups.py
1ut: https://github.com/MDAnalysis/mdanalysis/pull/3188/commits/3582d3a9768d20d2414b1d63a4fea2b7e60b2ea0

Added a description for the kind type.

testsuite/MDAnalysisTests/core/test_atomgroup.py
1ut: https://github.com/MDAnalysis/mdanalysis/pull/3188/commits/f6146d90640eb77d37d2e1090b52a82b1f1d5e47

Added match arguments.

package/MDAnalysis/core/groups.py
1ut: https://github.com/MDAnalysis/mdanalysis/pull/3188/commits/2f1577bfae5a9a1b166d43e7dede359059c9a7de

Fixed.

package/MDAnalysis/core/groups.py
1ut: I thought someone may want to sort by extracts of multiple dimensional arrays like the x-axis of positions. However, I also feel it should just raise an error when `idx.ndim` is not equal to 1.

IAlibay:
I've fixed the merge conflicts so CI can run.

@hmacdope and @richardjgowers can you weigh in on the "sorting > 1 D arrays"?

richardjgowers (approved):
Lgtm

package/MDAnalysis/core/groups.py
```diff
@@ -3324,6 +3324,66 @@ def write(self, filename=None, file_format=None,
...
+        """Return stably sorted Atomgroup by a key that specifies
+           an attribute of atomgroup.
+
+        Parameters
+        ----------
+        keys: str
+            The name for attribute of Atomgroup. (e.g. "ids", "ix". default="ix".)
+        keyfunc: function
+            A function that returns 1 dimension array, a key for sorting the
+            atomgroup from multiple dimension array that the attribute returns.
```
IAlibay, line 3337:
```suggestion
            A function to convert multidimensional arrays to a single
            dimension. This 1D array will be used as the sort key and
            is required when sorting with an ``AtomGroup`` attribute
            key which has multiple dimensions. Note: this argument
            is ignored when the attribute is one dimensional.
```

package/MDAnalysis/core/groups.py
```diff
@@ -3324,6 +3324,66 @@ def write(self, filename=None, file_format=None,
...
+        This sort is stable as it is implemented by
+        `numpy.argsort(kind='stable')`.
+
+        .. versionadded:: 2.0.0
+        """
+        idx = getattr(self.atoms, key)
+        if len(idx) != len(self.atoms):
+            raise ValueError("The array returned by the attribute '{}' "
+                             "must have a same length as the number of "
+                             "atoms".format(key))
```
IAlibay, line 3370:
```suggestion
                             "must have the same length as the number of "
                             "atoms in the input AtomGroup".format(key))
```

package/MDAnalysis/core/groups.py
```diff
@@ -3324,6 +3324,66 @@ def write(self, filename=None, file_format=None,
...
+            order = np.argsort(idx, kind='stable')
+        elif idx.ndim > 1:
+            if keyfunc is None:
+                raise NameError("You have to assign the argument 'keyfunc' "
+                                "a proper function that returns 1D array "
+                                "as the attribute '{}' returns multiple "
+                                "dimension array.".format(key))
+            sortkeys = keyfunc(idx)
+            if sortkeys.ndim != 1:
+                raise ValueError("The function you assigned to the argument "
```
IAlibay, line 3381:
```suggestion
                raise ValueError("The function assigned to the argument "
```

package/MDAnalysis/core/groups.py
```diff
@@ -3324,6 +3324,66 @@ def write(self, filename=None, file_format=None,
...
+        idx = getattr(self.atoms, key)
+        if len(idx) != len(self.atoms):
+            raise ValueError("The array returned by the attribute '{}' "
+                             "must have a same length as the number of "
+                             "atoms".format(key))
+        if idx.ndim == 1:
+            order = np.argsort(idx, kind='stable')
+        elif idx.ndim > 1:
+            if keyfunc is None:
+                raise NameError("You have to assign the argument 'keyfunc' "
```
IAlibay, line 3375: I think this message needs reversing in order to be more user friendly.

"The {} attribute returns a multidimensional array. In order to sort it, a function returning a 1D array (to be used as the sort key) must be passed to the `keyfunc` argument"

IAlibay (changes requested):
Nearly there, just some comments on the documentation text.

testsuite/MDAnalysisTests/core/test_atomgroup.py
```diff
@@ -1674,3 +1674,74 @@ def test_partial_timestep(self, universe):
...
+    ]
+
+    @pytest.mark.parametrize("inputs, expected", test_data, ids=test_ids)
+    def test_sort(self, ag, inputs, expected):
+        agsort = ag.sort(inputs)
+        assert np.array_equal(expected, agsort.ix)
+
+    def test_sort_position(self, ag):
+        ag.positions = (-np.arange(21)).reshape(7, 3)
+        with pytest.raises(ValueError, match=r"The array returned by the "
```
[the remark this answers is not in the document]
1ut: https://github.com/MDAnalysis/mdanalysis/pull/3188/commits/52b677a50c2977a0df07f40ba462c230f01974fb

Fixed.

package/MDAnalysis/core/groups.py
1ut: Fixed. 

https://github.com/MDAnalysis/mdanalysis/pull/3188/commits/e0bef58681c2fba19490ceb1d964740f60518e2d
https://github.com/MDAnalysis/mdanalysis/pull/3188/commits/338417e2ea1c71d908d22037412a8d7d5c014b72

1ut:
@IAlibay 
Thank you for your kind suggestions for documentations. I committed all changes.

package/MDAnalysis/core/groups.py
```diff
@@ -3324,6 +3324,82 @@ def write(self, filename=None, file_format=None,
 
         raise ValueError("No writer found for format: {}".format(filename))
 
+    def sort(self, key='ix', keyfunc=None):
+        """Returns a sorted ``AtomGroup`` using a specified attribute as \
```
IAlibay, line 3328: Unecessary `\`?

package/MDAnalysis/core/groups.py
```diff
@@ -3324,6 +3324,82 @@ def write(self, filename=None, file_format=None,
...
+            >>> import MDAnalysis as mda
+            >>> from MDAnalysisTests.datafiles import PDB_small
+            >>> u = mda.Universe(PDB_small)
+            >>> ag = u.atoms[[3, 2, 1, 0]]
+            >>> ag.ix
+            array([3 2 1 0])
+            >>> ag = ag.sort()
+            >>> ag.ix
+            array([0 1 2 3])
+            >>> ag.positions
```
IAlibay, line 3362: Awesome thanks for adding this example too!

package/MDAnalysis/core/groups.py
```diff
@@ -3324,6 +3324,82 @@ def write(self, filename=None, file_format=None,
...
+        keyfunc: callable, optional
+            A function to convert multidimensional arrays to a single
+            dimension. This 1D array will be used as the sort key and
+            is required when sorting with an ``AtomGroup`` attribute
+            key which has multiple dimensions. Note: this argument
+            is ignored when the attribute is one dimensional.
+
+        Returns
+        ----------
+        Group
```
IAlibay, line 3345: Looking at the docs (https://mdanalysis--3188.org.readthedocs.build/en/3188/documentation_pages/core/groups.html?highlight=sort#MDAnalysis.core.groups.AtomGroup.sort), this sets the "Return type" to "Group". We probably want to have this be a :class:`AtomGroup`?

```suggestion
        :class:`AtomGroup`
```

package/MDAnalysis/core/groups.py
```diff
@@ -3324,6 +3324,82 @@ def write(self, filename=None, file_format=None,
...
+                   [-11.921,  26.307,  10.41 ],
+                   [-11.447,  26.741,   9.595],
+                   [-12.44 ,  27.042,  10.926]], dtype=float32)
+
+        Note
+        ----------
+        This uses a stable sort as implemented by
+        `numpy.argsort(kind='stable')`.
+
+        .. versionadded:: 2.0.0
```
IAlibay, line 3379: Looking at the docs, the versionadded is being placed inside the `Note` box. We'll need two blank lines here.
```suggestion

        .. versionadded:: 2.0.0
```

package/MDAnalysis/core/groups.py
```diff
@@ -3324,6 +3324,82 @@ def write(self, filename=None, file_format=None,
...
+            if keyfunc is None:
+                raise NameError("The {} attribute returns a multidimensional "
+                                "array. In order to sort it, a function "
+                                "returning a 1D array (to be used as the sort "
+                                "key) must be passed to the keyfunc argument"
+                                .format(key))
+            sortkeys = keyfunc(idx)
+            if sortkeys.ndim != 1:
+                raise ValueError("The function assigned to the argument "
+                                 "'keyfunc':{} doesn't return 1D array."
```
IAlibay, line 3398:
```suggestion
                                 "'keyfunc':{} doesn't return a 1D array."
```

testsuite/MDAnalysisTests/core/test_atomgroup.py
```diff
@@ -1674,3 +1674,82 @@ def test_partial_timestep(self, universe):
...
+        assert np.array_equal(expected, agsort.ix)
+
+    def test_sort_bonds(self, ag):
+        with pytest.raises(ValueError, match=r"The array returned by the "
+                           "attribute.*"):
+            ag.sort("bonds")
+
+    def test_sort_positions_2D(self, ag):
+        with pytest.raises(ValueError, match=r"The function assigned to"
+                           ".*"):
```
IAlibay, line 1743: you shouldn't need the `.*` here (or above/below), it _should_ by default by a substring of the error message being raised.

IAlibay (changes requested):
Nearly there, just a couple of small things, mainly docstring changes to make sure the generated docs look nice.

package/MDAnalysis/core/groups.py
1ut: Fixed.
https://github.com/MDAnalysis/mdanalysis/pull/3188/commits/33776558e9dc6ce659dfc57537d98aeb80209ea7

testsuite/MDAnalysisTests/core/test_atomgroup.py
1ut: https://github.com/MDAnalysis/mdanalysis/pull/3188/commits/c2f734260ee05bd387572144b5a0a15463cf53e0

Fixed.

IAlibay (approved):
Assume CI returns green, lgtm!

hmacdope (approved):
great work @1ut, lgtm also 👍

PR #2869, opened 2020-07-24
adding H5MD-format writer

edisj:
Fixes #2866

Changes made in this Pull Request:
 - adds `class H5MDWriter` to `H5MD.py` 

This is a continuation from #2787

Some info:
[H5MD documentation](https://nongnu.org/h5md/index.html) - the writer is following these file format guidelines
[pyh5md github](https://github.com/pdebuyl/pyh5md) - A lot of the code is borrowed (with credit attributed to the author) from this library

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?
 - [x] Issue raised/referenced?

PR #3189, opened 2021-03-24
adding H5MDWriter to H5MD.py

### LINKED ISSUE #2866: Adding H5MDWriter to H5MD.py

Currently in my pull request ([PR#2787](https://github.com/MDAnalysis/mdanalysis/pull/2787)), we've only implemented an H5MD file reader. I'd like to implement a writer as well using [H5PY's](http://docs.h5py.org/en/stable/index.html) libraries for writing HDF5 files. [H5MD](https://nongnu.org/h5md/h5md.html) files follow specific rules for the structure of groups and datasets. There's already a package, [pyh5md](https://github.com/pdebuyl/pyh5md), that can write H5MD files nicely, but it's no longer being maintained by its author, so we've decided to stick to H5PY for writing the files.

We've been given permission from Pierre de Buyl, the author of pyh5md, to borrow code from his library and implement it into H5MDWriter. I've already begun a draft that's mostly working (uploaded to my github [here](https://github.com/Becksteinlab/h5md_examples/blob/master/testing/h5mdwriter/H5MD.py#L522-L966)), where I modeled it after `NCDFWriter` and added/tweaked code from pyh5md. I would like to open a pull request as soon as the reader is merged.

We could write the writer from scratch without using pyh5md, but that sounds a lot harder.

## Additional context ##

[H5MD format](https://nongnu.org/h5md/h5md.html)
[NCDFWriter](https://github.com/MDAnalysis/mdanalysis/blob/develop/package/MDAnalysis/coordinates/TRJ.py#L736-L1077) - the writer I used as a template to write mine
[pyh5md code](https://github.com/pdebuyl/pyh5md/blob/master/pyh5md/h5md_module.py) - used `TimeElement` class, `create_box()`, `default_chunks()`, and `element()`

edisj:
Fixes #2866 

 - adds `class H5MDWriter` to `H5MD.py` 

This is a continuation from #2787 and #2869. I closed #2869 because I tried rebasing to current develop and ended up making a mess of things, so it was easier to make a new pull request.

At the moment, the writer has all of the functionality it needs. It writes all of the data from the `timestep` (positions, velocities, etc), it writes everything from the `ts.data` dictionary that isn't `'time', 'step', 'dt'` into the 'observables' group. It can write datasets with chunked and/or compressed configurations, and files can be opened with MPI drivers.

The current to-do list is:
- ~~Benchmark the writer on multiple nodes with different configurations (chunked, compression)~~
- Have the code reviewed and clean things up
- ~~Figure out what to do the shape argument for `step` and `time` datasets when the number of frames written isn't known by the writer. This could be solved by chunking, but I'm not sure what the default chunk size should be for these datasets.~~

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -826,3 +826,301 @@ def __setstate__(self, state):
...
+        ts : :class:`Timestep`
+             :class:`Timestep` instance containing coordinates to
+             be written to trajectory file
+
+        Returns
+        -------
+        bool
+            Return ``True`` if `ts` contains a valid simulation box
+        """
+        return np.all(ts.dimensions > 0)
```
IAlibay, line 1010: Just passing by (strangely enough was thinking of #2698 and saw this pop up), going by line 622, wouldn't a no-dimensions case set `ts.dimensions` to `None` (i.e. this should cause a `TypeError`)?

IAlibay:
@edisj @orbeckst what's the target milestone for this? (i.e. is this a long term goal or are you planning 2.0/2.1?)

orbeckst:
It's not crucial for 2.0 and could go into 2.1 (at this point there will likely be a citation for a SciPy paper); if @edisj wants to work double-time to attempt to squeeze it into a 2.0 then I won't object, though.

edisj:
Hi @IAlibay , I'll push to have it done by 2.0 (May 10th, right?). I'm still testing some things with HPC benchmarks before updating this PR

IAlibay:
> Hi @IAlibay , I'll push to have it done by 2.0 (May 10th, right?).

So unless I misunderstood our workshop targets, the 10th is really the day it needs to be on conda-forge. I think we'd have to put down the code freeze at the latest the Friday before, i.e. the May 7th.

edisj:
Hey @orbeckst , ready for review! :) (also @IAlibay if  wanted to take a look)

I've made a lot of changes over the last few days and think it's complete other than  maybe a couple exceptions to add and documentation to edit. I'll finish up the tests for codecov tomorrow

There are a few parts of the code I'm not sure about, I'll open some discussions...

IAlibay (changes requested):
Sorry only had time to have a quick glance over, so it's mostly superficial comments.

I'll probably wait until you've added the extra tests before I re-review.

testsuite/MDAnalysisTests/coordinates/test_h5md.py
```diff
@@ -493,9 +358,325 @@ def test_no_units(h5md_file, ref, tmpdir):
...
+        with h5md_file as f:
+            u = mda.Universe(TPR_xvf, h5md_file)
+
+    def test_wrong_driver(self):
+        with pytest.raises(ValueError):
+            u = mda.Universe(TPR_xvf, H5MD_xvf,
+                             driver='wrong_driver', comm="MPI.COMM_WORLD")
+
+    def test_open_with_driver(self):
+        u = mda.Universe(TPR_xvf, H5MD_xvf, driver="core")
```
orbeckst, line 374: `assert` something here

testsuite/MDAnalysisTests/coordinates/test_h5md.py
```diff
@@ -493,9 +358,325 @@ def test_no_units(h5md_file, ref, tmpdir):
                 del g['particles/trajectory/position/time'].attrs['unit']
                 del g['particles/trajectory/box/edges/value'].attrs['unit']
         with pytest.raises(ValueError):
-            u = mda.Universe(TPR_xvf, outfile, format="H5MD")
+            u = mda.Universe(TPR_xvf, outfile)
+
+    @pytest.mark.xfail(reason='Issue #2884')
+    def test_open_filestream(self, h5md_file):
+        with h5md_file as f:
+            u = mda.Universe(TPR_xvf, h5md_file)
```
orbeckst, line 366: `assert` something here

testsuite/MDAnalysisTests/coordinates/test_h5md.py
```diff
@@ -493,9 +358,325 @@ def test_no_units(h5md_file, ref, tmpdir):
                 del g['particles/trajectory/position/time'].attrs['unit']
                 del g['particles/trajectory/box/edges/value'].attrs['unit']
         with pytest.raises(ValueError):
-            u = mda.Universe(TPR_xvf, outfile, format="H5MD")
+            u = mda.Universe(TPR_xvf, outfile)
```
orbeckst, line 361: `assert` something here

testsuite/MDAnalysisTests/coordinates/test_h5md.py
```diff
@@ -329,84 +280,28 @@ def test_unknown_time_unit(h5md_file, ref, tmpdir):
...
-            with h5py.File(outfile, 'w') as g:
-                f.copy(source='particles', dest=g)
-                f.copy(source='h5md', dest=g)
-                g['particles/trajectory/velocity/value'].resize((3, 10000, 3))
-        with pytest.raises(ValueError):
-            u = mda.Universe(TPR_xvf, outfile, format='H5MD')
-
+                del g['particles/trajectory/position']
+        u = mda.Universe(TPR_xvf, outfile)
+        assert u.trajectory.units['length'] == 'nm'
```
orbeckst, line 292: Also check that values are correct?

testsuite/MDAnalysisTests/coordinates/test_h5md.py
```diff
@@ -493,9 +358,325 @@ def test_no_units(h5md_file, ref, tmpdir):
...
+                             driver='wrong_driver', comm="MPI.COMM_WORLD")
+
+    def test_open_with_driver(self):
+        u = mda.Universe(TPR_xvf, H5MD_xvf, driver="core")
 
 
 @pytest.mark.skipif(not HAS_H5PY, reason="h5py not installed")
-def test_open_with_driver():
-    u = mda.Universe(TPR_xvf, H5MD_xvf, driver="core")
+class TestH5MDWriter_2(object):
```
orbeckst, line 378: more specific naming, please

orbeckst: and why can't it be under `TestH5MDWriter` ? ;-)

testsuite/MDAnalysisTests/coordinates/test_h5md.py
```diff
@@ -493,9 +358,325 @@ def test_no_units(h5md_file, ref, tmpdir):
...
+        assert_equal(3, box.attrs['dimension'])
+
+    @pytest.mark.parametrize('pos, vel, force', (
+            (True, False, False),
+            (True, True, False),
+            (True, False, True),
+            (True, True, True),
+            (False, True, True),
+            (False, False, True),
+            #(False, False, False)
```
orbeckst, line 425: What is the behavior in this case? It should be well-defined and tested.

testsuite/MDAnalysisTests/coordinates/test_h5md.py
```diff
@@ -493,9 +358,325 @@ def test_no_units(h5md_file, ref, tmpdir):
...
+                    convert_units=convert_units) as W:
+            for ts in universe.trajectory:
+                W.write(universe)
+
+        uw = mda.Universe(TPR_xvf, outfile)
+        uw_units = uw.trajectory.units.values()
+        for u1, u2 in zip(units, uw_units):
+            assert_equal(u1, u2)
+
+    #@pytest.mark.parametrize('chunks', (None, (1, 1000, 3)))
```
orbeckst, line 564: What does `chunks = None` do at the moment? Trying different chunks looks like a good test.

testsuite/MDAnalysisTests/coordinates/test_h5md.py
```diff
@@ -493,9 +358,325 @@ def test_no_units(h5md_file, ref, tmpdir):
...
+        chunks = (1, 1000, 3)
+        with Writer(outfile,
+                    universe.atoms.n_atoms,
+                    chunks=chunks) as W:
+            for ts in universe.trajectory:
+                W.write(universe)
+
+        uw = mda.Universe(TPR_xvf, outfile)
+        dset = uw.trajectory._particle_group['position/value']
+        assert_equal(dset.chunks, chunks)
```
orbeckst, line 575: Should you also assert actual values to make sure that they end up in the correct position in the file?

testsuite/MDAnalysisTests/coordinates/test_h5md.py
```diff
@@ -493,9 +358,325 @@ def test_no_units(h5md_file, ref, tmpdir):
...
+            for ts in universe.trajectory:
+                W.write(universe)
+
+        uw = mda.Universe(TPR_xvf, outfile)
+        dset = uw.trajectory._particle_group['position/value']
+        assert_equal(dset.compression, filter)
+        assert_equal(dset.compression_opts, opts)
+
+    @pytest.mark.parametrize('driver, comm', (('core', None),
+                                              ('mpio', 'comm')))
```
orbeckst, line 596: The try/except inside to distinguish these two cases looks unnecessary complicated and introduces additional uncertainty (the blanket `except`). Why not make two different tests?

testsuite/MDAnalysisTests/coordinates/test_h5md.py
```diff
@@ -493,9 +358,325 @@ def test_no_units(h5md_file, ref, tmpdir):
...
+                        universe.atoms.n_atoms,
+                        driver=driver,
+                        comm=comm) as W:
+                for ts in universe.trajectory:
+                    W.write(universe)
+
+            uw = mda.Universe(TPR_xvf, outfile, driver=driver, comm=comm)
+            file = uw.trajectory._file
+            assert_equal(file.driver, driver)
+        except:
```
orbeckst, line 611: That's not good — avoid catch-all excepts as they can obscure real problems.

testsuite/MDAnalysisTests/coordinates/test_h5md.py
```diff
@@ -493,9 +358,325 @@ def test_no_units(h5md_file, ref, tmpdir):
...
+                    compression_opts=opts) as W:
+            for ts in universe.trajectory:
+                W.write(universe)
+
+        uw = mda.Universe(TPR_xvf, outfile)
+        dset = uw.trajectory._particle_group['position/value']
+        assert_equal(dset.compression, filter)
+        assert_equal(dset.compression_opts, opts)
+
+    @pytest.mark.parametrize('driver, comm', (('core', None),
```
orbeckst, line 595: I assume that 'comm' on the next line is a mock-up for the MPI communicator. If so, perhaps either use a proper `mock` object or write the string like "mock MPICOMM", just to make sure that this is clear. Sometimes people look at the tests for how to use code so when you don't use the code correctly (on purpose) it can get confusing.

testsuite/MDAnalysisTests/coordinates/test_h5md.py
```diff
@@ -493,9 +358,325 @@ def test_no_units(h5md_file, ref, tmpdir):
...
+        except:
+            with pytest.raises(ValueError):
+                with Writer(outfile,
+                            universe.atoms.n_atoms,
+                            driver=driver,
+                            comm=comm) as W:
+                    for ts in universe.trajectory:
+                        W.write(universe)
+
+    def test_timestep_not_modified_by_writer(self, universe, Writer, outfile):
```
orbeckst, line 620: Is this a concern?''

I am generally all in favor of testing everything but if there's no possibility that this happens, I'd remove it to make the tests pass a shred faster.

testsuite/MDAnalysisTests/coordinates/test_h5md.py
```diff
@@ -493,9 +358,325 @@ def test_no_units(h5md_file, ref, tmpdir):
...
+            W.write(universe)
+
+        assert_equal(
+            ts._pos,
+            x,
+            err_msg="Positions in Timestep were modified by writer.")
+        assert_equal(
+            ts.time, time, err_msg="Time in Timestep was modified by writer.")
+
+    """def test_gaps(self, universe, Writer, outfile):
```
orbeckst, line 640: You can mark tests as XFAIL if that's a known problem. It serves as documentation and a good way to motivate future work — all one has to do is to make the XFAILS pass!

testsuite/MDAnalysisTests/data/coordinates/README.md
```diff
@@ -44,10 +44,7 @@ plain text using Gromacs utilities.
 test.h5md
 ---------
 ## Creation
-Converted test.trr to test.h5md with MDAnalysis and pyh5md
-using the 'create_h5md_data.py' script.
-See https://github.com/pdebuyl/pyh5md for pyh5md repository.
+Written with MDAnalysis using the 'create_h5md_data.py' script
```
orbeckst, line 47: Does the h5md file now have good layout?

testsuite/MDAnalysisTests/data/coordinates/README.md
```diff
@@ -44,10 +44,7 @@ plain text using Gromacs utilities.
...
 
 ## Validation
-Manually examined contents of test.trr with MDAnalysis and test.h5md
-with h5py and checked that the contents were the same.
+Manually examined contents of test.h5md with MDAnalysis.
```
orbeckst, line 50: Validation should use tools outside MDAnalysis.... otherwise we might be self consistent but wrong.

orbeckst (changes requested):
That's a pretty hefty PR. Sorry, it took me a while and there's probably stuff I overlooked but please see comments/questions.

Apart from my comments, please update CHANGELOG.

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -183,10 +184,13 @@
...
+.. autoclass:: H5MDWriter
+   :members:
+
 """
 
-import numpy as np
 import MDAnalysis as mda
+import numpy as np
 from . import base, core
 from .base import Timestep
```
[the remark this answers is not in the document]
edisj: Wow I didn't realize I had to hit "submit review" for my comments to show up, I thought I had already replied to your review comments sorry :(

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -726,6 +746,433 @@ def __setstate__(self, state):
...
+                try:
+                    self.units[key[:-4]] = self.new_units[key]
+                except KeyError:
+                    raise RuntimeError(f"{key} is not a unit recognizable by"
+                                        " MDAnalyis. Allowed units are:"
+                                       f" {self._unit_translation_dict.keys()}"
+                                        " For more information on units, see"
+                                        " `MDAnalyis units`_.")
+
+        """if self.convert_units:
```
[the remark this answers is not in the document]
edisj: Will uncomment and add tests

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -726,6 +746,451 @@ def __setstate__(self, state):
...
+        for key, value in self.new_units.items():
+            if value is not None:
+                if value not in self._unit_translation_dict[key]:
+                    raise RuntimeError(f"{value} is not a unit recognizable by"
+                                        " MDAnalyis. Allowed units are:"
+                                       f" {self._unit_translation_dict.keys()}"
+                                        " For more information on units, see"
+                                        " `MDAnalyis units`_.")
+                else:
+                    self.units[key] = self.new_units[key]
```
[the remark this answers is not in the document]
edisj: I'm pretty sure `self.units` has to be in MDA units since all of the `convert_*_to_native()` methods pull from `self.units.values()` to get conversion factors

testsuite/MDAnalysisTests/coordinates/test_h5md.py
edisj: should this be asserting something? I thought it was just trying to catch the value error

orbeckst: You're right, I must have overlooked the raises.

testsuite/MDAnalysisTests/coordinates/test_h5md.py
orbeckst: As discussed in our meeting, leave it as it seems to be useful and done for other writers.

testsuite/MDAnalysisTests/coordinates/test_h5md.py
edisj: two separate tests makes more sense

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -726,6 +746,451 @@ def __setstate__(self, state):
...
+
+        """
+
+        self._step.resize(self._step.shape[0]+1, axis=0)
+        try:
+            self._step[-1] = ts.data['step']
+        # step must exist in h5md file
+        except(KeyError):
+            self._step[-1] = ts.frame
+        self._time.resize(self._time.shape[0]+1, axis=0)
```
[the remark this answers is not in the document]
edisj: I've done a little bit of digging but can't seem to find any case where resizing with a particular chunk layout doesn't work. I'll keep looking though

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -726,6 +746,451 @@ def __setstate__(self, state):
...
+        _has = {'position': self.has_positions,
+                'velocity': self.has_velocities,
+                'force': self.has_forces}
+
+        for group, value in _has.items():
+            if value:
+                self._step = self.traj.require_dataset(f'{group}/step',
+                                                       shape=(0,),
+                                                       maxshape=(None,),
+                                                       dtype=np.int32)
```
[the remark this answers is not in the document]
edisj: From what I can tell MDA uses the regular python int for both `ts.data['step']` and `ts.frame`. Is there any datatype conversion between int and numpy.int32?

testsuite/MDAnalysisTests/data/coordinates/README.md
edisj: yes, it's following the (1, n_atoms, 3) chunk scheme

testsuite/MDAnalysisTests/coordinates/test_h5md.py
edisj: new updates let you write contiguous dsets, made a new test for this case

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -726,6 +746,433 @@ def __setstate__(self, state):
...
+        self._comm = comm
+        self.n_atoms = n_atoms
+        self.chunks = (1, n_atoms, 3) if chunks is None else chunks
+        self.compression = compression
+        self.compression_opts = compression_opts
+        self.convert_units = convert_units
+        self.h5md_file = None
+
+        # check which datasets are to be written
+        self.has_positions = kwargs.get('positions', False)
```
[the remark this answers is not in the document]
orbeckst: Yes, that's a good idea. We still need the catch-all `**kwargs` for anything else that we do not care about but having explicit kwarg names in the signature (+ defaults) is good.

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -726,6 +777,635 @@ def __setstate__(self, state):
...
+        software that wrote the file [``MDAnalysis``]
+    creator_version : str (optional)
+        version of software that wrote the file
+        [:attr:`MDAnalysis.__version__`]
+
+
+    Note
+    ----
+    Parellel writing with the use of a MPI communicator and the ``'mpio'``
+    HDF5 driver is currently not supported.
```
orbeckst, line 893: Move to after the synopsis/before Parameters. If it doesn't format as a box, use
```reST
.. note::

    Parallel writing ...
```

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -726,6 +777,635 @@ def __setstate__(self, state):
...
+        :class:`AtomGroup`
+    IOError
+        when ``n_atoms`` of the :class:`Universe` or :class:`AtomGroup`
+        being written does not match ``n_atoms`` passed as an argument
+        to the writer
+    ValueError
+        when any of the optional ``timeunit``, ``lengthunit``,
+        ``velocityunit``, or ``forceunit`` keyword arguments are
+        unrecognizable by MDAnalysis
+
```
orbeckst, line 921: Put the long text section here. See numpy doc.

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -726,6 +777,635 @@ def __setstate__(self, state):
...
+    def __init__(self,
+                 filename,
+                 n_atoms,
+                 driver=None,
+                 comm=None,
+                 convert_units=True,
+                 chunks=None,
+                 compression=None,
+                 compression_opts=None,
+                 **kwargs):
```
orbeckst, line 983: Make all documented kwargs explicit instead of `**kwargs`. Just leave `**kwargs` to catch any unprocessed kwargs.

This will help a lot with the documentation and seeing defaults.

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -726,6 +777,635 @@ def __setstate__(self, state):
...
+
+        By default, it fills the :attr:`self.units` dictionary by copying the
+        units dictionary of the parent file. Because H5MD files have no
+        standard unit restrictions, users may pass a kwarg in ``(timeunit,
+        lengthunit, velocityunit, forceunit)`` to the writer so long as
+        MDAnalysis has a conversion factor for it (:exc:`ValueError` raised if
+        it does not). These custom unit arguments must be in
+        `MDAnalysis notation`_. If custom units are supplied from the user,
+        :attr`self.units[unit]` is replaced with the corresponding
+        ``unit`` argument.
```
orbeckst, line 1074: This text will not be visible in the docs even though it contains useful information. Please summarize it as part of the section on units in the writer notes.

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -726,6 +777,635 @@ def __setstate__(self, state):
...
+        """
+
+        # use units of parent file as native units
+        try:
+            # ag == AtomGroup
+            self.units = ag.universe.trajectory.units.copy()
+        except AttributeError:
+            # ag == Universe
+            self.units = ag.trajectory.units.copy()
+
```
orbeckst, line 1085: Replace this whole thing with
```python
self.units = ag.universe.trajectory.units.copy()
```

Should always work (tested in 2.0.0b):
```python
from
In [2]: from MDAnalysis.tests import datafiles as data

In [4]: u = mda.Universe(data.TPR, data.XTC)

In [5]: u.universe.trajectory.units
Out[5]: {'time': 'ps', 'length': 'nm'}

In [6]: u.atoms.universe.trajectory.units
Out[6]: {'time': 'ps', 'length': 'nm'}
```

orbeckst: And it will increase coverage because you don't have to explicitly test the two cases.

(We'll assume that it works because hopefully we're testing elsewhere writing with ag and Universe.)

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -726,6 +777,635 @@ def __setstate__(self, state):
...
+        # set user input units
+        for key, value in self._new_units.items():
+            if value is not None:
+                if value not in self._unit_translation_dict[key]:
+                    raise ValueError(f"{value} is not a unit recognizable by"
```
orbeckst, line 1090: recognized

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -726,6 +777,635 @@ def __setstate__(self, state):
...
+
+    def _open_file(self):
+        """Opens file with `H5PY`_ library and fills in metadata from kwargs.
+
+        :attr:`self.h5md_file` becomes file handle that links to root level.
+
+        """
+
+        if self._comm is not None:  # pragma: no cover
+            # ValueError raised if communicator is passed
```
orbeckst, line 1114: say where he ValueError is raised, currently a bit confusing because the raise is not here

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -726,6 +777,635 @@ def __setstate__(self, state):
...
+        """initializes all datasets that will be written to by
+        :meth:`_write_next_timestep`
+
+        Note
+        ----
+        :exc:`NoDataError` is raised if no positions, velocities, or forces are
+        found in the input trajectory. While the H5MD standard allows for this
+        case, :class:`H5MDReader` cannot currently read files without at least
+        one of these three groups. A future change to both the reader and
+        writer will allow this case.
```
orbeckst, line 1149: Does not show up in docs. Summarize in main writer docs (and leave it here, too).

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -726,6 +777,635 @@ def __setstate__(self, state):
...
+
+
+        """
+
+        # for keeping track of where to write in the dataset
+        self._counter = 0
+
+        # ask the parent file if it has postions, velocities, and forces
```
orbeckst, line 1157: positions

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -726,6 +777,635 @@ def __setstate__(self, state):
...
+                                                    maxshape=(None,),
+                                                    dtype=np.int32)
+            self._time = self._traj.require_dataset('box/edges/time',
+                                                    shape=(0,),
+                                                    maxshape=(None,),
+                                                    dtype=np.float32)
+            self._set_attr_unit(self._edges, 'length')
+            self._set_attr_unit(self._time, 'time')
+        else:
+            self._traj['box'].attrs['boundary'] = 3*['none']
```
orbeckst, line 1194: comment that it _must_ be "none" and not `None` according to ... (h5 or h5md?)

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -726,6 +777,635 @@ def __setstate__(self, state):
...
+                                   dtype=data.dtype)
+        if 'step' not in self._obsv[group]:
+            self._obsv[f'{group}/step'] = self._step
+        if 'time' not in self._obsv[group]:
+            self._obsv[f'{group}/time'] = self._time
+
+    def _set_attr_unit(self, dset, unit):
+        """helper function to set a 'unit' attribute for an HDF5 dataset"""
+
+        if self.units[unit] is None:
```
orbeckst, line 1292: Is there really no test case that exercises this branch?

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -726,6 +777,635 @@ def __setstate__(self, state):
...
+
+        The first dimension of each dataset is extended by +1 and
+        then the data is written to the new slot.
+
+        Note
+        ----
+        Writing H5MD files with fancy trajectory slicing where the Timestep
+        does not increase monotonically such as ``u.trajectory[[0,1,2,1]]``
+        or ``u.trajectory[[0,1,2,0,1,2]]`` raises a :exc:`ValueError` as this
+        violates the rules of the step dataset in the H5MD standard.
```
orbeckst, line 1312: Add this information to the main docs (as this is not visible).

package/MDAnalysis/coordinates/H5MD.py
orbeckst: You have a lot of great docs hidden in private methods. Many describe limitations or caveats with the format or writer. The notes section should include this information, too.

package/MDAnalysis/coordinates/H5MD.py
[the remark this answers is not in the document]
orbeckst: Add a comment to the code here along the lines of "should work with ANY chunk layout", at least then we have a record that we thought about it. If you find any evidence, put links/citation in the comment.

orbeckst (changes requested):
First half of a review (did not have enough time to finish). Mainly docs.

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -148,6 +175,7 @@
 .. _`Build h5py from sources`: https://docs.h5py.org/en/stable/mpi.html#building-against-parallel-hdf5
 .. _`H5MD notation`: https://nongnu.org/h5md/modules/units.html
 .. _`MDAnalysis notation`: https://userguide.mdanalysis.org/units.html
+.. _`MDAnalyis units`: https://userguide.mdanalysis.org/units.html
```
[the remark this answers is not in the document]
edisj: fixed

orbeckst:
@edisj see the remaining comments, please. If you're addressing it then just resolve it. If you want to discuss or disagree then add a comment and leave it open.

It would also be nice to use *duecredit* to automatically generate citations.

edisj:
Hi @orbeckst , just pushed the changes to duecredit and added that final test. All comments should be addressed! I'm going through to add a couple more details in the docs regarding writing files with `n_frames` and with contiguous layouts.

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -1002,13 +1000,19 @@ class H5MDWriter(base.WriterBase):
...
+               description="Specifications of the H5MD standard",
+               path=__name__, version='1.1')
     def __init__(self, filename, n_atoms, n_frames=None, driver=None,
                  comm=None, convert_units=True, chunks=None, compression=None,
                  compression_opts=None, positions=True, velocities=True,
                  forces=True, timeunit=None, lengthunit=None,
                  velocityunit=None, forceunit=None, author='N/A',
                  author_email=None, creator='MDAnalysis',
-                 creator_version=mda.__version__, **kwargs):
+                 creator_version=mda_version, **kwargs):
```
orbeckst, line 1015: `mda.__version__` looks ok to me, why did you change it?

package/CHANGELOG
```diff
@@ -23,9 +23,13 @@ The rules for this file:
...
 
 Fixes
+<<<<<<< HEAD
+  * Fixed missing array flatten preventing PCA from running when mean positions
+=======
   * Fixed sometimes wrong sorting of atoms into fragments when unwrapping
     (Issue #3352, PR #3353)
   * Fixed missing array flatten preventing PCA from running when mean positions  
+>>>>>>> 65319046c49f4bce5e49949b7ab451039adcbd67
     provided and bugs causing tests to incorrectly pass (Issue #2728, PR #3296)
```
orbeckst, line 33: fix merge !!

testsuite/MDAnalysisTests/coordinates/test_writer_api.py
```diff
@@ -52,6 +52,9 @@ def test_ts_error(writer, tmpdir):
         fn = str(tmpdir.join('out.xtc'))
     elif writer == mda.coordinates.LAMMPS.DATAWriter:
         pytest.skip("DATAWriter requires integer atom types")
+    elif writer == mda.coordinates.H5MD.H5MDWriter:
+        pytest.skip("H5MDWriter requires h5py to be installed, "
+                    "fails MIN dependency test")
```
orbeckst, line 57: You are _always_ excluding this test, even if hdf5 is installed. The API tests should run in full tests.

Only skip the tests if you cannot import hdf5.

orbeckst (changes requested):
initial comments to change, sorry short on time

testsuite/MDAnalysisTests/coordinates/test_writer_api.py
```diff
@@ -52,10 +52,9 @@ def test_ts_error(writer, tmpdir):
         fn = str(tmpdir.join('out.xtc'))
     elif writer == mda.coordinates.LAMMPS.DATAWriter:
         pytest.skip("DATAWriter requires integer atom types")
-    elif writer == mda.coordinates.H5MD.H5MDWriter:
-        if not mda.coordinates.H5MD.HAS_H5PY:
-            pytest.skip("H5MDWriter requires h5py to be installed, "
-                        "fails MIN dependency test")
+    elif writer == mda.coordinates.H5MD.H5MDWriter and not mda.coordinates.H5MD.HAS_H5PY:
+        pytest.skip("H5MDWriter requires h5py to be installed, "
+                    "fails MIN dependency test")
```
orbeckst, line 57: It's not a fail, just a skip. Just say "skipping for MIN dependency test"

testsuite/MDAnalysisTests/coordinates/test_writer_api.py
```diff
@@ -83,14 +82,17 @@ def test_write_with_atomgroup(writer, tmpdir):
         pytest.skip("MOL2 only writes MOL2 back out")
     elif writer == mda.coordinates.LAMMPS.DATAWriter:
         pytest.skip("DATAWriter requires integer atom types")
-    elif writer == mda.coordinates.H5MD.H5MDWriter:
-        if not mda.coordinates.H5MD.HAS_H5PY:
-            pytest.skip("macOS and ubuntu, 3.6, MIN tests fail")
+    elif writer == mda.coordinates.H5MD.H5MDWriter and not mda.coordinates.H5MD.HAS_H5PY:
+        pytest.skip("macOS and ubuntu, 3.6, MIN tests fail")
```
orbeckst, line 86: It's not a fail, just a skip. Just say "skipping for MIN dependency test"

testsuite/MDAnalysisTests/coordinates/test_writer_api.py
```diff
@@ -83,14 +82,17 @@ def test_write_with_atomgroup(writer, tmpdir):
...
     else:
         fn = str(tmpdir.join('out.traj'))
 
-    with writer(fn, n_atoms=u.atoms.n_atoms) as w:
-        w.write(u.atoms)
+    if writer == mda.coordinates.H5MD.H5MDWriter:
+        with writer(fn, n_atoms=u.atoms.n_atoms, convert_units=False) as w:
```
orbeckst, line 91: add a comment explaining why special case necessary here

testsuite/MDAnalysisTests/coordinates/test_writer_api.py
```diff
@@ -106,11 +108,14 @@ def test_write_with_universe(writer, tmpdir):
         pytest.skip("MOL2 only writes MOL2 back out")
     elif writer == mda.coordinates.LAMMPS.DATAWriter:
         pytest.skip("DATAWriter requires integer atom types")
-    elif writer == mda.coordinates.H5MD.H5MDWriter:
-        if not mda.coordinates.H5MD.HAS_H5PY:
-            pytest.skip("macOS and ubuntu, 3.6, MIN tests fail")
+    elif writer == mda.coordinates.H5MD.H5MDWriter and not mda.coordinates.H5MD.HAS_H5PY:
+        pytest.skip("macOS and ubuntu, 3.6, MIN tests fail")
```
orbeckst, line 112: It's not a fail, just a skip. Just say "skipping for MIN dependency test"

orbeckst: Also, remove 3.6, this might change in the future. ... and the ubuntu/macOS
```suggestion
        pytest.skip("skipping H5MDWriter test because h5py is not installed")
```

testsuite/MDAnalysisTests/coordinates/test_writer_api.py
orbeckst: Also, remove 3.6, this might change in the future. ... and the ubuntu/macOS
```suggestion
        pytest.skip("skipping H5MDWriter test because h5py is not installed")
```

testsuite/MDAnalysisTests/coordinates/test_writer_api.py
orbeckst, line 57: Also, remove 3.6, this might change in the future. ... and the ubuntu/macOS
```suggestion
        pytest.skip("skipping H5MDWriter test because h5py is not installed")
```

testsuite/MDAnalysisTests/coordinates/test_writer_api.py
```diff
@@ -106,11 +108,14 @@ def test_write_with_universe(writer, tmpdir):
...
     else:
         fn = str(tmpdir.join('out.traj'))
 
-    with writer(fn, n_atoms=10) as w:
-        w.write(u)
+    if writer == mda.coordinates.H5MD.H5MDWriter:
+        with writer(fn, n_atoms=u.atoms.n_atoms, convert_units=False) as w:
```
orbeckst, line 117: please explain why special case needed (comment)

orbeckst (changes requested):
minor doc/comment changes

orbeckst:
@edisj is there anything missing as far as you can see? Is it worth's @IAlibay 's time to have a final look?

I only had one last set of minor requests, but I also have seen the whole thing multiple times so a second pair of eyes would be welcome.

Can you think of anyone else who should review the hdf5 code, @edisj ?

edisj:
> @edisj is there anything missing as far as you can see? Is it worth's @IAlibay 's time to have a final look?
> 
> I only had one last set of minor requests, but I also have seen the whole thing multiple times so a second pair of eyes would be welcome.
> 
> Can you think of anyone else who should review the hdf5 code, @edisj ?

After scanning through the pull request again I see I never updated `testsuite/MDAnalysisTests/data/coordinates/README.md` to say it was validated outside MDAnalysis with h5py, I'll fix that...

Maybe @lilyminium or @IAlibay can scan through it? In the dev discord a while ago, we discussed what the default data to write from the `ts` should be. If you two have some spare time to scan through to make sure it's consistent with the other writers I would appreciate it. Also, what are your thoughts on the documentation? Is it understandable? And is the logic of the design decisions clear?

package/MDAnalysis/coordinates/H5MD.py
edisj: instead of importing all of `mda`, I did: `from MDAnalysis import __version__ as mda_version`, since the only thing I needed to import `mda` for was `mda.__version__`. Is one way preferable over the other?

IAlibay:
> Maybe @lilyminium or @IAlibay can scan through it?
Sure I'll have a look through tomorrow morning.

P.S. Self-assigned because it's the only way to fool my notification filters into thinking it has work to do 🙄

package/MDAnalysis/coordinates/H5MD.py
orbeckst: You will immediately import all of mda with `from MDAnalysis import __version__` — it makes no difference. Therefore, I'd keep it simple and clear (as it was previously). Generally, I am not a fan of `from x import y` because it is messy. "Namespaces are a honking good idea. Let's have more of those" is one of the Python Zens... IIRC.

testsuite/MDAnalysisTests/data/coordinates/create_h5md_data.py
```diff
@@ -1,81 +1,41 @@
...
-            velocities.append(uni.trajectory.ts.velocities,
-                              ts.frame, time=ts.time)
-            forces.append(uni.trajectory.ts.forces,
-                          ts.frame, time=ts.time)
-            data_step.append(uni.trajectory.ts.data['step'],
-                             ts.frame, time=ts.time)
-            data_lambda.append(uni.trajectory.ts.data['lambda'],
-                               ts.frame, time=ts.time)
 
 def main():
```
IAlibay, line 34: How is this being called? It's unclear to me as to how this is being used. i.e. do we need to have a `main` or is it just being called by something else?

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -726,6 +772,670 @@ def __setstate__(self, state):
...
+    def __init__(self, filename, n_atoms, n_frames=None, driver=None,
+                 comm=None, convert_units=True, chunks=None, compression=None,
+                 compression_opts=None, positions=True, velocities=True,
+                 forces=True, timeunit=None, lengthunit=None,
+                 velocityunit=None, forceunit=None, author='N/A',
+                 author_email=None, creator='MDAnalysis',
+                 creator_version=mda_version, **kwargs):
+
+        if not HAS_H5PY:
+            raise RuntimeError("H5MDWriter: Please install h5py")
```
IAlibay, line 1018: We have the machinery to mock the inability to import of a module. Can we do this to test this?

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -726,6 +772,670 @@ def __setstate__(self, state):
...
+                self._edges[i] = self.convert_pos_to_native(self._edges[i])
+        if self._has['velocity']:
+            if self.units['velocity'] is not None:
+                self._vel[i] = self.convert_velocities_to_native(self._vel[i])
+        if self._has['force']:
+            if self.units['force'] is not None:
+                self._force[i] = self.convert_forces_to_native(self._force[i])
+
+    @property
+    def has_positions(self):
```
IAlibay, line 1412: any ideas why the setters aren't getting covered here?

IAlibay (changes requested):
Having a deeper look but here are some initial comments from a quick scan.

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -726,6 +772,670 @@ def __setstate__(self, state):
...
+                                 "with no units, set ``convert_units=False``.")
+
+    def _open_file(self):
+        """Opens file with `H5PY`_ library and fills in metadata from kwargs.
+
+        :attr:`self.h5md_file` becomes file handle that links to root level.
+
+        """
+
+        if self._comm is not None:  # pragma: no cover
```
IAlibay, line 1139: If this is currently an unreachable code path, would just having this commented out instead be better? (it's not cleaner, but then at least there's no unaccessible code?

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -726,6 +772,670 @@ def __setstate__(self, state):
...
+            # ValueError raised in __init__() if communicator is passed
+            # MPI parallel writing not yet supported
+            self.h5md_file = h5py.File(name=self.filename,
+                                       mode='w',
+                                       driver=self._driver,
+                                       comm=self._comm)
+        elif self._driver is not None:
+            self.h5md_file = h5py.File(name=self.filename,
+                                       mode='w',
+                                       driver=self._driver)
```
IAlibay, line 1149: The default value for `driver` is `None` anyways, could this not be squashed into the one call?

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -726,6 +772,670 @@ def __setstate__(self, state):
...
+        self._traj = self.h5md_file['particles/trajectory']
+        self.data_keys = [
+            key for key in ts.data.keys() if key not in self.data_blacklist]
+        if self.data_keys:
+            self._obsv = self.h5md_file.require_group('observables')
+
+        # box group is required for every group in 'particles'
+        self._traj.require_group('box')
+        self._traj['box'].attrs['dimension'] = 3
+        if ts.dimensions is not None and np.all(ts.dimensions > 0):
```
IAlibay, line 1202: Just checking, is the ts.dimensions == 0 as an empty box behaviour also being replicated on read?

IAlibay (changes requested):
--> review through to line 1240 of H5MD.py

Sorry I'm going to have to break up my reviews into chunks. Here is the first part of my review, I'll follow up with the rest probably tomorrow at this rate.

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -726,6 +772,670 @@ def __setstate__(self, state):
...
+        Note
+        ----
+        Writing H5MD files with fancy trajectory slicing where the Timestep
+        does not increase monotonically such as ``u.trajectory[[2,1,0]]``
+        or ``u.trajectory[[0,1,2,0,1,2]]`` raises a :exc:`ValueError` as this
+        violates the rules of the step dataset in the H5MD standard.
+
+        """
+
+        i = self._counter
```
IAlibay, line 1343: A more descriptive counter here might be worth it. I'm not in the "i,j,k,l are terrible" camp, but at the same time if they get created outside of a loop it's really easy for someone to accidentally do `for i in range(N)` somewhere in your code and forget that you had `i` already declared somewhere else.

testsuite/MDAnalysisTests/coordinates/test_h5md.py
```diff
@@ -77,336 +75,226 @@ def test_get_writer_2(self, ref, reader, tmpdir):
...
+    def test_no_group(self, h5md_file, outfile, attr, group):
         with h5md_file as f:
             with h5py.File(outfile, 'w') as g:
                 f.copy(source='particles', dest=g)
                 f.copy(source='h5md', dest=g)
-                del g['particles/trajectory/force']
-        u = mda.Universe(TPR_xvf, outfile, format='H5MD')
+                del g[f'particles/trajectory/{group}']
+        u = mda.Universe(TPR_xvf, outfile)
         with pytest.raises(NoDataError):
```
IAlibay, line 256: Wherever possible, can we have a `match` for the associated error/warning we are matching?
This makes sure that we are testing the error we actually want to test (you'd think this wouldn't be needed, but in practice it happens more often than you'd think).

IAlibay (changes requested):
Second half of review.

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -613,13 +634,19 @@ def _copy_to_data(self):
                 self.ts.data[key] = self._file['observables'][key][
                     'value'][self._frame]
 
-        # pulls 'time' out of first available parent group
+        # pulls 'time' and 'step' out of first available parent group
         for name, value in self._has.items():
             if value:
                 if 'time' in self._particle_group[name]:
                     self.ts.data['time'] = self._particle_group[name][
```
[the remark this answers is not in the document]
orbeckst: `ts.data` is actually where per-step information lives... and `ts.time = ...` goes through the property.setter. But I agree, `ts.time = ...` is clearer and more uniform across the code base.

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -613,13 +634,19 @@ def _copy_to_data(self):
...
                         'time'][self._frame]
                     break
+        for name, value in self._has.items():
```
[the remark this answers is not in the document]
orbeckst: I think, technically speaking, `time` and `step` can be in different groups so the `break` for one would not be correct for the other. --- or did I get this wrong @edisj ?

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -726,6 +772,670 @@ def __setstate__(self, state):
...
+    ----
+    Parellel writing with the use of a MPI communicator and the ``'mpio'``
+    HDF5 driver is currently not supported.
+
+    Note
+    ----
+    :exc:`NoDataError` is raised if no positions, velocities, or forces are
+    found in the input trajectory. While the H5MD standard allows for this
+    case, :class:`H5MDReader` cannot currently read files without at least
+    one of these three groups. A future change to both the reader and
```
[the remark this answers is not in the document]
orbeckst: Maybe do not promise "a future change" and just acknowledge the limitation. Raise an issue and wait for anyone actually wanting empty h5md trajectories (eg for storing observables).

package/MDAnalysis/coordinates/H5MD.py
orbeckst: Check out PR #3294 where @IAlibay mocks out netcdf.

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -726,6 +772,670 @@ def __setstate__(self, state):
...
+    n_atoms : int
+        number of atoms in trajectory
+    n_frames : int (optional)
+        number of frames to be written in trajectory
+    driver : str (optional)
+        H5PY file driver used to open H5MD file
+    comm : :class:`MPI.Comm` (optional)
+        MPI communicator used to open H5MD file
+        Must be passed with `'mpio'` file driver.
+        This argument is currently disabled.
```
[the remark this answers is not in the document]
orbeckst: On reflection I agree with @IAlibay .  Keep it lean and promise as little as possible. 

Remove the docs for `comm` and the code that raises the exception. You can raise an issue and copy the text/code there for future use.

package/MDAnalysis/coordinates/H5MD.py
orbeckst: Still to do

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -726,6 +772,670 @@ def __setstate__(self, state):
...
+                 creator_version=mda_version, **kwargs):
+
+        if not HAS_H5PY:
+            raise RuntimeError("H5MDWriter: Please install h5py")
+        self.filename = filename
+        if n_atoms == 0:
+            raise ValueError("H5MDWriter: no atoms in output trajectory")
+        self._driver = driver
+        self._comm = comm
+        if self._comm is not None or self._driver == 'mpio':
```
orbeckst, line 1024: remove (with the parallel promise)

package/MDAnalysis/coordinates/H5MD.py
orbeckst: I think `i` is fine here because it makes the rest of the code more readable.

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -726,6 +772,670 @@ def __setstate__(self, state):
...
+                obs = self._obsv[f'{key}/value']
+                obs.resize(obs.shape[0]+1, axis=0)
+                obs[i] = ts.data[key]
+
+        if self.convert_units:
+            self._convert_units(i)
+
+        self._counter += 1
+
+    def _convert_units(self, i):
```
orbeckst, line 1392: Can you rename this to `_convert_dataset_with_units()` or something similarly expressive? Because for a second I though this was one of the standard `_convert_*` methods and wondered why you would want to convert the integer `i`.

package/MDAnalysis/coordinates/H5MD.py
orbeckst: They are never set through the setter in this code.

Will need extra tests.

orbeckst (changes requested):
Minor things & mostly agreeing with @IAlibay (except that I think `i` can stay ;-) ).

You'll have to ask @IAlibay how fast you need to be in order to try and slip this into 2.0.0.

package/MDAnalysis/coordinates/H5MD.py
edisj: good call, I added a parameter in the "no dimensions" test and it passes

package/MDAnalysis/coordinates/H5MD.py
edisj: it can, looks much better!

package/MDAnalysis/coordinates/H5MD.py
edisj: I opted to remove it completely until parallel is enabled

package/MDAnalysis/coordinates/H5MD.py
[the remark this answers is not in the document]
IAlibay: I think I was thinking more of a `continue` statement, but I can see how this might be a pain so I'm happy with the way it is.

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -687,6 +714,25 @@ def _reopen(self):
...
+
+        .. versionadded:: 2.0.0
+
+        """
+        if n_atoms is None:
+            n_atoms = self.n_atoms
+        return H5MDWriter(
+            filename,
+            n_atoms=n_atoms,
+            convert_units=self.convert_units,
```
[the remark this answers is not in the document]
edisj: I went ahead and mimicked the way you handled kwargs in NCDF. By default it now writes with the settings of the file being read and users should be able to overwrite them

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -687,6 +714,25 @@ def _reopen(self):
...
+        --------
+        :class:`H5MDWriter`  Output class for the H5MD format
+
+
+        .. versionadded:: 2.0.0
+
+        """
+        if n_atoms is None:
+            n_atoms = self.n_atoms
+        return H5MDWriter(
```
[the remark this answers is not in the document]
edisj: I agree, fixed now

package/MDAnalysis/coordinates/H5MD.py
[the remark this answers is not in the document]
edisj: That's right, `time` is not a required dataset, so some groups may contain it and some may not

package/MDAnalysis/coordinates/H5MD.py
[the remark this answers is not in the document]
edisj: changed to `ts.time` instead of `ts.data['time']`

testsuite/MDAnalysisTests/coordinates/test_h5md.py
edisj: I'll go through and add matches

package/MDAnalysis/coordinates/H5MD.py
edisj: I like `i` since it's always used as a counter, but I see your point in that since _it's always used as a counter_ someone might accidently overwrite it. I'll leave it for now?

testsuite/MDAnalysisTests/data/coordinates/create_h5md_data.py
edisj: It's meant to be called by a person, not within the package itself. I copied `testsuite/MDAnalysisTests/data/coordinates/create_data.py` which does the same thing to create synthetic trajectories

edisj: Unless I'm misunderstanding how `testsuite/MDAnalysisTests/data/coordinates/create_data.py` is meant to be used...

testsuite/MDAnalysisTests/coordinates/test_h5md.py
```diff
@@ -1,13 +1,14 @@
 import pytest
 from numpy.testing import assert_almost_equal, assert_equal
 import numpy as np
+import sys
```
orbeckst, line 4: Not needed, I think.
```suggestion
```

testsuite/MDAnalysisTests/coordinates/test_h5md.py
```diff
@@ -457,13 +457,23 @@ def test_step_from_frame(self, universe, Writer, outfile):
...
+    def test_has_setter(self, universe, Writer, outfile):
+        with Writer(outfile, universe.atoms.n_atoms) as W:
+            W.has_forces = False
+            W.has_velocities = False
+            W.write(universe)
+
+        uw = mda.Universe(TPR_xvf, outfile)
+        assert_equal(uw.trajectory.ts.has_positions, True)
+        assert_equal(uw.trajectory.ts.has_velocities, False)
+        assert_equal(uw.trajectory.ts.has_forces, False)
```
orbeckst, line 476: Instead of `assert_equal` (which works for everything but particularly arrays) I would just use the simpl
```python
assert uw.trajectory.ts.has_positions is True
```
(for boolean, `is` is ok, but in general `==` is safer).

testsuite/MDAnalysisTests/coordinates/test_h5md.py
```diff
@@ -457,13 +457,23 @@ def test_step_from_frame(self, universe, Writer, outfile):
         for step, frame in zip(steps, frames):
             assert_equal(step, frame)
 
-    def test_has_setter(self, universe, Writer, outfile):
+    def test_has_property(self, universe, Writer, outfile):
         with Writer(outfile, universe.atoms.n_atoms) as W:
             W.write(universe)
             assert_equal(W.has_positions, W._has['position'])
             assert_equal(W.has_velocities, W._has['velocity'])
             assert_equal(W.has_forces, W._has['force'])
```
orbeckst: See other comment
```python
assert W.has_positions == W._has['position']
```

testsuite/MDAnalysisTests/coordinates/test_h5md.py
lilyminium: Interesting facts:

```python
>>> np.bool(1) is True
True
>>> np.ones(1, dtype=bool)[0] is True
False
>>> np.ones(1, dtype=bool)[0] == True
True
```

;)

orbeckst: Scary.

And odd:

```python
>>> isinstance(np.bool_, bool)
False
```

lilyminium: Yeah the basic gist is that `np.bool_` is an actual type, as opposed to the Python bool which is a subclass of int. [There's a warning on the docs if you actively look for it.](https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.bool_). It's just funny when `assert True is True` failing.

```python
>>> isinstance(True, int)
True
```

Screen Shot 2021-08-01 at 1 49 18 pm

orbeckst (approved):

IAlibay:
I can try to re-review a bit later today, although if you think all my comments were addressed please do go ahead with the merge @orbeckst

orbeckst:
Your point about "check exception messages" was a good one — I am currently watching @edisj find out which exceptions are actually triggered .... so _my_ comments had been addressed, yours should be done in the next hour.

edisj:
Hi @IAlibay , I just pushed my final changes, all comments should be addressed. All error tests now have a `match`. I hope there's still time to make it into 2.0.0...

If you have time to review, there are a couple of decisions made while talking to @orbeckst that might need a little justification:
- Removed the `has_*` setters from the writer because we realized that in order to actually write a test to activate them, we had to write code that wouldn't ever be used for an H5MD trajectory, so the setter itself was pointless
- Passed defaults (compression, compression_opts, driver) from input H5MD into `H5MDReader.Writer()`, but intentionally left out `chunks`. This is because H5MDWriter chooses `(1, n_atoms, 3)` by default, and accidently choosing a bad chunk shape can be really bad for performance. I added a detailed note in the docstring of `H5MDReader.Writer()` to explain the situation and also say how to manually set the chunk shape if you wanted to.

Thanks again for your review :D, if there's anything else that remains let me know

testsuite/MDAnalysisTests/data/coordinates/create_h5md_data.py
IAlibay: I'll be honest, that one exists from before my time so I don't know either 😅 I'm happy with it as-is if there's a use case for it.

IAlibay (commented):
Thanks @edisj I'm just having a quick look through now. Just doing this quick pre-review comment to include this duecredit thing - not sure if it needs testing or not ?

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -352,6 +377,12 @@ class H5MDReader(base.ReaderBase):
         }
     }
 
+    @due.dcite(Doi("10.25080/majora-1b6fd038-005"),
```
IAlibay, line 380: Or I guess my comment didn't get added to my review - was wondering if this needed to be added to test_duecredit in some way?

edisj: I did test it manually with a script that used H5MD and it works. I asked @orbeckst whether a unit test was necessary and he said no

IAlibay: Ok, sounds fair, wouldn't have thought it would be a blocking thing either way anyways. @orbeckst do we maybe want to raise an issue for the future to check if we should eventually add to test_duecredit? (sounds like an easy GSoC starter?)

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -726,6 +801,642 @@ def __setstate__(self, state):
...
+    the default value. To write a file without any one of these datsets,
+    set ``positions=False``, ``velocities=False``, or ``forces=False``.
+
+    .. rubric:: Units
+
+    The H5MD format is very flexible with regards to units, as there is no
+    standard defined unit for the format. For this reason, :class:`H5MDWriter`
+    does not enforce any units. The units of the written trajectory can be set
+    explicitly with the keyword arguments `lengthunit`, `velocityunit`, and
+    `forceunit`. If units are not explicitly specified, they are set to the
```
IAlibay, line 930:
```suggestion
    explicitly with the keyword arguments ``lengthunit``, ``velocityunit``, and
    ``forceunit``. If units are not explicitly specified, they are set to the
```

Do we want double ticks here? There's double ticks above for other parameters I _think_?

IAlibay (approved):
Just the one comment from a quick scan though. I'll directly approve now since it's just a documentation style thing.

package/MDAnalysis/coordinates/H5MD.py
orbeckst: I didnt' think we tested any of the other due citations. BUT: I was wrong! 

@edisj if you could add an entry in [testsuite/MDAnalysisTests/utils/test_duecredit.py](https://github.com/MDAnalysis/mdanalysis/blob/develop/testsuite/MDAnalysisTests/utils/test_duecredit.py) then that would be the dot on the i.

(I don't know if we have really tested all duecredits. THat's something to check & easy enough)

package/MDAnalysis/coordinates/H5MD.py
orbeckst: I _think_ that when we refer to parameters in docstrings we use single ticks, i.e., the reST default role. If it's anything else code-like for which there's no reST role in sphinx then we use the "monospace" double-backticks.

(I didn't look at the code context here so I don't know if `lengthunit` and friends are parameters/kwargs of a method or no.)

package/MDAnalysis/coordinates/H5MD.py
edisj: duecredit test is in

package/MDAnalysis/coordinates/H5MD.py
edisj: They're parameters of H5MDWriter's `__init__()`, so I _think_ they should get double back-ticks, or not?

orbeckst: single back ticks

edisj: I've been way too liberal with doubles... fixing now

edisj:
@IAlibay @orbeckst just merged with current develop for good measure, but I think the final points regarding the duecredit test and the backticks are resolved

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -70,9 +76,40 @@
...
+
+.. code-block:: python
+
+    import MDAnalysis as mda
+    u = mda.Universe("topology.tpr", "trajectory.h5md")
+    with mda.Writer("output.h5md", n_atoms=u.trajectory.n_atoms) as W:
+        for ts in u.trajectory:
+            W.write(u)
+
+To write an H5MD file with contiguous datasets, you must specifly the
```
richardjgowers, line 97: Specifly 🪰

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -70,9 +76,40 @@
...
+number of frames to be written and set ``chunks=False``:
```
richardjgowers, line 98: Maybe a quick one line hint as to why someone would want to do this

package/MDAnalysis/coordinates/base.py
```diff
@@ -107,6 +107,12 @@
    .. autoattribute:: dimensions
    .. autoattribute:: triclinic_dimensions
    .. autoattribute:: volume
+   .. attribute:: data
+
+      :class:`dict` that holds arbitrary per Timestep data
+
+      .. versionadded:: 0.11.0
```
richardjgowers, line 114: 2.0?

richardjgowers (commented):
Looks good, just a couple doc quibbles that can slide

package/MDAnalysis/coordinates/base.py
IAlibay: This is a retrospective addition of docs for `data`, which would match release 0.11.0.

package/MDAnalysis/coordinates/H5MD.py
IAlibay: There's a note for it on line 948, although not sure if it fully covers what you're asking @richardjgowers ?

package/MDAnalysis/coordinates/H5MD.py
IAlibay:
```suggestion
To write an H5MD file with contiguous datasets, you must specify the
```
I'll just merge it in myself.

IAlibay:
Seems to be everything, I'll merge once CI returns green - thanks for all the work @edisj 🎉

orbeckst:
Congratulations @edisj — PR with >200 comments and what turned out to be one of the more complicated formats.

Given that this could be considered, together with the SciPy paper, the capstone of your NSF REU experience, would you want to write a blog post for https://www.mdanalysis.org/blog/ where you summarize your experience and achievements?

PR #3202, opened 2021-04-03
Fixes sphzone, sphlayer, cyzone, cylayer empty selections

### LINKED ISSUE #2915: make empty sphzone / cyzone selections consistent with 'around'
Since sphzone will do a center of geometry calculation first, if an empty atomgorup is passed to the sphzone, a not very useful error message will be given.

```
    water = u.select_atoms( 'resname SOL and sphzone 3 (resid 12 and name HE)')
  File "/Users/grte2001/GitHub/mdanalysis/package/MDAnalysis/core/universe.py", line 677, in select_atoms
    return self.atoms.select_atoms(*args, **kwargs)
  File "/Users/grte2001/GitHub/mdanalysis/package/MDAnalysis/core/groups.py", line 2863, in select_atoms
    selections[0].apply(self))
  File "/Users/grte2001/GitHub/mdanalysis/package/MDAnalysis/core/selection.py", line 136, in apply
    rsel = self.rsel.apply(group)
  File "/Users/grte2001/GitHub/mdanalysis/package/MDAnalysis/core/selection.py", line 168, in apply
    return func(self, group)
  File "/Users/grte2001/GitHub/mdanalysis/package/MDAnalysis/core/selection.py", line 333, in apply
    ref = sel.center_of_geometry().reshape(1, 3).astype(np.float32)
ValueError: cannot reshape array of size 0 into shape (1,3)
```

~An error message which says the selection is empty could be given.~

EDIT: Following the discussion below and on #3026 we want these selections to behave like `around`, i.e., just return an empty AtomGroup. — @orbeckst 

Just return an empty atomgroup.

orioncohen:
Fixes #2915 

 - created a new decorator to check for empty sub-expressions in sphzone, sphlayer, cyzone, and cylayer.
 - added testing to confirm the fix
 - updated CHANGELOG
 - updated AUTHORS

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -22,7 +22,8 @@ The rules for this file:
   * 2.0.0
 
 Fixes
-  * Fixed 'sphzone' behavior to be consistent with 'around' (Issue #2915)
+  * Fixed 'sphzone', 'sphlayer', 'cyzone', and 'cylayer' to return empty if the
+   zone/layer is empty, consistent with 'around' (Issue #2915)
   * A Universe created from an ROMol with no atoms returns now a Universe 
     with 0 atoms (Issue #3142)
   * ValueError raised when empty atomgroup is given to DensityAnalysis
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -17,7 +17,7 @@ The rules for this file:
          lilyminium, daveminh, jbarnoud, yuxuanzhuang, VOD555, ianmkenney,
          calcraven，xiki-tempula, mieczyslaw, manuel.nuno.melo, PicoCentauri,
          hanatok, rmeli, aditya-kamath, tirkarthi, LeonardoBarneschi, hejamu,
-         biogen98
+         biogen98, orioncohen
 
   * 2.0.0
 
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -22,6 +22,7 @@ The rules for this file:
   * 2.0.0
 
 Fixes
+  * Fixed 'sphzone' behavior to be consistent with 'around' (Issue #2915)
   * A Universe created from an ROMol with no atoms returns now a Universe 
     with 0 atoms (Issue #3142)
   * ValueError raised when empty atomgroup is given to DensityAnalysis
--- a/package/MDAnalysis/core/selection.py
+++ b/package/MDAnalysis/core/selection.py
@@ -352,6 +352,8 @@ class SphericalLayerSelection(DistanceSelection):
     def apply(self, group):
         indices = []
         sel = self.sel.apply(group)
+        if len(sel) == 0:
+            return group[[]]
         box = self.validate_dimensions(group.dimensions)
         periodic = box is not None
         ref = sel.center_of_geometry().reshape(1, 3).astype(np.float32)
@@ -396,7 +398,8 @@ class CylindricalSelection(Selection):
     @return_empty_on_apply
     def apply(self, group):
         sel = self.sel.apply(group)
-
+        if len(sel) == 0:
+            return group[[]]
         # Calculate vectors between point of interest and our group
         vecs = group.positions - sel.center_of_geometry()
 
--- a/package/MDAnalysis/core/selection.py
+++ b/package/MDAnalysis/core/selection.py
@@ -217,6 +217,20 @@ def return_empty_on_apply(func):
         return func(self, group)
     return apply
 
+def return_empty_on_empty_selection(func):
+    """
+    Decorator to return empty AtomGroups from the apply() function
+    if entire the selection is empty or a sub-expression is empty.
+    """
+    @functools.wraps(func)
+    def apply(self, group):
+        if len(group) == 0:
+            return group
+        if len(self.sel.apply(group)) == 0:
+            return group[[]]
+        return func(self, group)
+    return apply
+
 class _Selectionmeta(type):
     def __init__(cls, name, bases, classdict):
         type.__init__(type, name, bases, classdict)
@@ -348,12 +362,10 @@ class SphericalLayerSelection(DistanceSelection):
         self.exRadius = float(tokens.popleft())
         self.sel = parser.parse_expression(self.precedence)
 
-    @return_empty_on_apply
+    @return_empty_on_empty_selection
     def apply(self, group):
         indices = []
         sel = self.sel.apply(group)
-        if len(sel) == 0:
-            return group[[]]
         box = self.validate_dimensions(group.dimensions)
         periodic = box is not None
         ref = sel.center_of_geometry().reshape(1, 3).astype(np.float32)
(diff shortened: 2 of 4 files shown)


orioncohen:
Could someone help me understand what codecov is doing and how I can fix the error it is throwing?

lilyminium (changes requested):
Thanks for looking at this, @orioncohen. LGTM. As this is your first contribution to MDAnalysis, please add yourself to AUTHORS -- please also add your username to the top of CHANGELOG :-)

lilyminium:
Thanks @orioncohen. Looking more closely, though, the issue actually mentions cyzone and (in a comment lower down) sphlayer. As a more general fix could you look at those too? In fact it may be more elegant to just modify the `return_empty_on_apply` decorator or make a new decorator, to follow the Don't Repeat Yourself philosophy of Python.

orioncohen:
Hi @lilyminium. I've added a fix for cyzone, cylayer, and sphlayer too. I tried creating a new decorator to test if the selection is empty. The problem is that we need to check the contents of `sel` after calling`sel = self.sel.apply(group)`, so the decorator needs to reference an internal method of the object. The best workaround I can think of is to add a decorator method to the Selection base class but that appears to be a meta-class. I ended up going with the naive solution, which required adding 4 more lines of code to fix cyzone, cylayer, and sphlayer.

lilyminium:
> so the decorator needs to reference an internal method of the object. The best workaround I can think of is to add a decorator method to the Selection base class but that appears to be a meta-class. 

Your solution does fix the problem, but I'm surprised you can't just decorate the `apply()` methods of the relevant Selection classes the same way that `return_empty_on_apply` has been used.

orioncohen:
I should have taken a closer look when I was less sleepy! I implemented this with a decorator as you suggested. I added a new decorator rather than modifying `return_empty_on_apply` because not all of the selection methods call `self.sel.apply(group)`. I also moved around some testing to be a bit more consistent.

package/MDAnalysis/core/selection.py
```diff
@@ -348,7 +363,7 @@ def __init__(self, parser, tokens):
         self.exRadius = float(tokens.popleft())
         self.sel = parser.parse_expression(self.precedence)
 
-    @return_empty_on_apply
+    @return_empty_on_empty_selection
```
lilyminium, line 366: Rather than replacing `return_empty_on_apply`, you can usually stack decorators like

```
@return_empty_on_apply
@return_empty_on_empty_selection
```

package/MDAnalysis/core/selection.py
```diff
@@ -217,6 +217,21 @@ def apply(self, group):
...
 
+
+def return_empty_on_empty_selection(func):
+    """
+    Decorator to return empty AtomGroups from the apply() function
+    if entire the selection is empty or a sub-expression is empty.
+    """
+    @functools.wraps(func)
+    def apply(self, group):
+        if len(group) == 0:
```
lilyminium, line 228: If you stack decorators as I commented above, you could avoid replicating the functionality of `return_empty_on_apply`.

package/MDAnalysis/core/selection.py
```diff
@@ -217,6 +217,21 @@ def apply(self, group):
...
+            return group
+        if len(self.sel.apply(group)) == 0:
+            return group[[]]
+        return func(self, group)
```
lilyminium, line 232: @richardjgowers will know efficient this is and whether the following suggestion is a good idea, but calling `self.sel.apply(group)` twice (once in the decorator, once in the function it's decorating) seems a little wasteful. Could you do `sel = self.sel.apply(group)` and pass `sel` in as an extra keyword argument? This would involve modifying the function signature of the relevant `apply`s, which might be a bad idea.

testsuite/MDAnalysisTests/core/test_atomselections.py
```diff
@@ -209,6 +209,8 @@ def test_sphzone(self, universe, selstr):
     def test_cylayer(self, universe, selstr):
         sel = universe.select_atoms(selstr)
         assert_equal(len(sel), 88)
+        empty = universe.select_atoms('cylayer 4.0 6.0 10 -10 name NOT_A_NAME')
```
lilyminium, line 212: Thanks, these are thorough tests -- could you make each of these separate though? That way each function tests one specific functionality.

lilyminium (changes requested):
Thanks @orioncohen, the decorator is much nicer than before! I just have some design comments. @richardjgowers you know more than I do about selections, so please weigh in if I've said anything dumb.

testsuite/MDAnalysisTests/core/test_atomselections.py
orioncohen: Yep! just fixed this!

package/MDAnalysis/core/selection.py
orioncohen: Implemented this too!

package/MDAnalysis/core/selection.py
orioncohen: I could definitely just pass `sel` in as an extra keyword. I don't like that it would make the function signature of `apply` inconsistent between classes though. I could also go back to implementing it in the functions, but that's very repetitive. This seems like largely a stylistic choice, so just let me know how I should proceed!

richardjgowers: Yeah I think Lily is right that this is neater but less efficient.  I'd say duplicate the code a few times

orioncohen: Done. I let me know if there is anything else I should do before this gets merged.

package/CHANGELOG
```diff
@@ -17,11 +17,12 @@ The rules for this file:
          lilyminium, daveminh, jbarnoud, yuxuanzhuang, VOD555, ianmkenney,
          calcraven，xiki-tempula, mieczyslaw, manuel.nuno.melo, PicoCentauri,
          hanatok, rmeli, aditya-kamath, tirkarthi, LeonardoBarneschi, hejamu,
-         biogen98
+         biogen98, orioncohen
 
   * 2.0.0
 
 Fixes
+  * Fixed 'sphzone' behavior to be consistent with 'around' (Issue #2915)
```
lilyminium, line 25: Could you now update this CHANGELOG message to be more informative?

lilyminium (changes requested):
Other than the CHANGELOG message, LGTM!

package/CHANGELOG
orioncohen: Done! Thanks for the help Lily! (loved your PyCon talk btw)

lilyminium: Thank you @oriencohen! :-)

lilyminium (approved):

PR #3210, opened 2021-04-03
speed-up FastNS

zemanj:
Fixes nothing, just improves `FastNS` performance

 - use branching instead of boolean multipliers in `_minimum_image_ortho_lazy()` to gain 23.4 % speedup

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/lib/include/calc_distances.h
+++ b/package/MDAnalysis/lib/include/calc_distances.h
@@ -50,9 +50,17 @@ inline void _minimum_image_ortho_lazy(double* x, float* box, float* half_box)
     *
     * Assumes that the maximum separation is less than 1.5 times the box length.
     */
-    for (int i = 0; i < 3; i++) {
-        x[i] -= (x[i] > half_box[i]) * box[i];
-        x[i] += (x[i] <= -half_box[i]) * box[i];
+    for (int i = 0; i < 3; ++i) {
+        if (x[i] > half_box[i]) {
+            x[i] -= box[i];
+        }
+        else
+        {
+            if (x[i] <= -half_box[i])
+            {
+                x[i] += box[i];
+            }
+        }
     }
 }
 


zemanj:
I noticed that the code for `_minimum_image_ortho_lazy()` I used in PR #3186 is not ideal because I scraped it from an old AVX2-vectorized code that I wrote a while ago that used a different memory layout for coordinates. For the memory layout we use here, branching is more efficient than the previously employed boolean multipliers, regardless of using "-march=native" or the like.

**Code used for performance test:**
```python
import numpy as np
import MDAnalysis as mda
from MDAnalysis.lib.nsgrid import FastNS
from MDAnalysis.lib.mdamath import triclinic_vectors
u = mda.Universe(PDB)  # triclinic box with 47681 atoms
u.dimensions = np.concatenate((np.diag(triclinic_vectors(u.dimensions)),
                               np.array([90.0, 90.0, 90.0], dtype=np.float32)))
# test small cut-off:
fns = FastNS(cutoff=5.0, coords=u.atoms.positions, box=u.dimensions, pbc=True)
%timeit -r 10 -n 10 fns.search(u.atoms.positions)
# test larger cut-off:
fns = FastNS(cutoff=10.0, coords=u.atoms.positions, box=u.dimensions, pbc=True)
%timeit -r 10 -n 10 fns.search(u.atoms.positions)
```
**Output before this PR:**
```
293 ms ± 365 µs per loop (mean ± std. dev. of 10 runs, 10 loops each)
2.31 s ± 5.64 ms per loop (mean ± std. dev. of 10 runs, 10 loops each)
```
**Output after this PR:**
```
224 ms ± 1.18 ms per loop (mean ± std. dev. of 10 runs, 10 loops each)
1.77 s ± 4.3 ms per loop (mean ± std. dev. of 10 runs, 10 loops each)
```
The resulting speedup is 23.4 % for both the small and the large cut-off.

zemanj:
BTW does anyone know why codecov/project is usually just a very slow random number generator?

IAlibay:
> BTW does anyone know why codecov/project is usually just a very slow random number generator?

I think @mnmelo brought this up on some issue somewhere. From what I understand the patch is a nightmare because you essentially have a ton of different CI runs with different coverages piling on top of each other. The project eventually returns a reasonable number, but whatever the bot says remains completely wrong.

I need to sit down at some point and work it out, but as far as I can tell if we just run by the post-all CI project number it's accurate (and then you have to actually go on codecov to get the real breakdown).

zemanj:
Thanks @IAlibay for the explanation. Looks like codecov calmed down already :smile:

package/MDAnalysis/lib/include/calc_distances.h
```diff
@@ -50,9 +50,17 @@ inline void _minimum_image_ortho_lazy(double* x, float* box, float* half_box)
     *
     * Assumes that the maximum separation is less than 1.5 times the box length.
```
IAlibay, line 51: As a thought, it would be good to have your findings somewhere (maybe like a line here to say branching is faster?). I think it'll save time next time to know someone already went ahead and looked at whether branching made this better/worse.

IAlibay (approved):
Not much here to review so easy enough to approve :)

I'll still let @richardjgowers make the final decision on merging this to `master` (given he's got the reigns on 1.1 NSGrid stuff).

package/MDAnalysis/lib/include/calc_distances.h
zemanj: In general I agree, but in this case I'm a bit hesitant because i) the function did not exist before my previous PR #3186, and ii) such things are never irrefutable truths because technology changes and such a comment will be misleading if (for whatever reason) in the future, boolean multipliers are faster than branching. Plus I didn't try CMOVs, and there's probably more things one could try out.

zemanj:
Thanks @IAlibay and @richardjgowers, that was lightning fast :zap:

PR #3211, opened 2021-04-04
tweak after Manel called out my bad code

richardjgowers:
@mnmelo this PR targets your branch, I think this change removes the functional call you said was slow.  I wasn't 100% sure how you were timing things, so if you could double check the performance, but for me I got a performance boost.


mnmelo:
Seems good! I'll just merge and report performance measurements back in the discussion of #3169.

mnmelo:
(I also changed the description of the squashed merge to something a bit less self-flagellating, but if you want to we're still in time to do plenty of mea culpas when #3169 gets merged ;)

PR #3223, opened 2021-04-08
Coverage for diffusionmap raised to 100% [Issue #3208]

### LINKED ISSUE #3208: Improve diffusionmap coverage
The coverage for diffusionmap is only 96.55% covered, with lines 312 and 338 uncovered.

Add tests to raise this to 100% (the warning maybe hard to trigger, but maybe a dummy single atom system could be constructed to trigger the frame limit?

## Additional context ##
This is an issue aimed at providing a simple issue for first time contributing GSoC students to work on.

ojeda-e:
Fixes #3208 
Coverage for diffusionmap is 100% covered, including lines 312 and 338.

 - Raised value error when not a Universe by parsing trajectory only, with the `pytest.raises(ValueError,  match=<pattern>)` approach.
 - Raised warning when `n_frames > 5000` by manually patching `_dist_matrix`  instead of parsing long trajectory. 
 It takes a very long time to test/run using a dummy trajectory with `n_frames>5000` even when selecting 1 atom only.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/testsuite/MDAnalysisTests/analysis/test_diffusionmap.py
+++ b/testsuite/MDAnalysisTests/analysis/test_diffusionmap.py
@@ -81,14 +81,16 @@ def test_transform(u, dmap):
     diffusion_space = dmap.transform(n_eigenvectors, 1)
     assert diffusion_space.shape == (eigvects.shape[0], n_eigenvectors)
 
+
 def test_long_traj(u):
     with pytest.warns(UserWarning, match='The distance matrix is very large'):
         dmap = diffusionmap.DiffusionMap(u)
-        dmap._dist_matrix.run()
+        dmap._dist_matrix.run(stop=1)
         dmap._dist_matrix.n_frames = 5001
-        assert dmap.run()
+        dmap.run()
 
-def test_universe(u):
+
+def test_not_universe_error(u):
     trj_only = u.trajectory
-    with pytest.raises(ValueError, match='U is not a Universe'):       
-        assert diffusionmap.DiffusionMap(trj_only)
+    with pytest.raises(ValueError, match='U is not a Universe'):    
+        diffusionmap.DiffusionMap(trj_only)
--- a/testsuite/MDAnalysisTests/analysis/test_diffusionmap.py
+++ b/testsuite/MDAnalysisTests/analysis/test_diffusionmap.py
@@ -92,5 +92,5 @@ def test_long_traj(u):
 
 def test_not_universe_error(u):
     trj_only = u.trajectory
-    with pytest.raises(ValueError, match='U is not a Universe'):    
+    with pytest.raises(ValueError, match='U is not a Universe'):
         diffusionmap.DiffusionMap(trj_only)
--- a/testsuite/MDAnalysisTests/analysis/test_diffusionmap.py
+++ b/testsuite/MDAnalysisTests/analysis/test_diffusionmap.py
@@ -81,7 +81,7 @@ def test_transform(u, dmap):
     diffusion_space = dmap.transform(n_eigenvectors, 1)
     assert diffusion_space.shape == (eigvects.shape[0], n_eigenvectors)
 
-def test_long_traj(u, tmpdir):
+def test_long_traj(u):
     with pytest.warns(UserWarning, match='The distance matrix is very large'):
         dmap = diffusionmap.DiffusionMap(u)
         dmap._dist_matrix.run()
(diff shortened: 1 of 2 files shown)


testsuite/MDAnalysisTests/analysis/test_diffusionmap.py
```diff
@@ -80,3 +80,16 @@ def test_transform(u, dmap):
     dmap.run()
     diffusion_space = dmap.transform(n_eigenvectors, 1)
     assert diffusion_space.shape == (eigvects.shape[0], n_eigenvectors)
+
+def test_long_traj(u, tmpdir):
```
tylerjereddy, line 84: Is the `tmpdir` fixture actually used for something here? I don't see it being used?

If there is an artifact being produced, I believe `with tmpdir.as_cwd():` is a popular choice with the core team.

ojeda-e: Ugh, no sorry. I forgot to erase that from the trial I ran using a dummy trajectory with 5000 frames. The test run with `u` only.

testsuite/MDAnalysisTests/analysis/test_diffusionmap.py
```diff
@@ -80,3 +80,15 @@ def test_transform(u, dmap):
     dmap.run()
     diffusion_space = dmap.transform(n_eigenvectors, 1)
     assert diffusion_space.shape == (eigvects.shape[0], n_eigenvectors)
+
+def test_long_traj(u):
+    with pytest.warns(UserWarning, match='The distance matrix is very large'):
+        dmap = diffusionmap.DiffusionMap(u)
+        dmap._dist_matrix.run()
```
IAlibay, line 87: So it's not exactly a big thing (it only makes execution go down from 555 ms of CPU time to 27.7 ms), but since we only really care about the user warning, it might make sense to just put a `stop=1` argument on that `run` call.

testsuite/MDAnalysisTests/analysis/test_diffusionmap.py
```diff
@@ -80,3 +80,15 @@ def test_transform(u, dmap):
...
+        dmap._dist_matrix.n_frames = 5001
+        assert dmap.run()
+
+def test_universe(u):
+    trj_only = u.trajectory
+    with pytest.raises(ValueError, match='U is not a Universe'):       
+        assert diffusionmap.DiffusionMap(trj_only)
```
IAlibay, line 94: As above, you shouldn't need the `assert` here.

IAlibay (changes requested):
Thanks @ojeda-e, just a couple of suggestions.

testsuite/MDAnalysisTests/analysis/test_diffusionmap.py
```diff
@@ -80,3 +80,15 @@ def test_transform(u, dmap):
     dmap.run()
     diffusion_space = dmap.transform(n_eigenvectors, 1)
     assert diffusion_space.shape == (eigvects.shape[0], n_eigenvectors)
+
```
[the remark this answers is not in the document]
ojeda-e: Thanks, fixed.

testsuite/MDAnalysisTests/analysis/test_diffusionmap.py
```diff
@@ -80,3 +80,15 @@ def test_transform(u, dmap):
...
+def test_long_traj(u):
+    with pytest.warns(UserWarning, match='The distance matrix is very large'):
+        dmap = diffusionmap.DiffusionMap(u)
+        dmap._dist_matrix.run()
+        dmap._dist_matrix.n_frames = 5001
+        assert dmap.run()
+
+def test_universe(u):
```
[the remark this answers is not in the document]
ojeda-e: Yes, agreed. Updated.

testsuite/MDAnalysisTests/analysis/test_diffusionmap.py
ojeda-e: Thanks @IAlibay, I was getting confused with the usage of assertions here but your comments helped me to get the point. Your feedback is very appreciated.

testsuite/MDAnalysisTests/analysis/test_diffusionmap.py
ojeda-e: Thanks, that's a good point. Will keep it in mind.

ojeda-e:
Hi @IAlibay, I went through the comments. Would you please review it again? The pep8 is now solved. 
Thanks ☺️

IAlibay (approved):

IAlibay:
Thanks @ojeda-e, squash-merging.

PR #3232, opened 2021-04-14
MDAnalysis.analysis.align.alignto won't reindex the atomgroup as is documented

### LINKED ISSUE #2977: MDAnalysis.analysis.align.alignto won't reindex the atomgroup as is documented
## Expected behavior ##

In the doc (https://docs.mdanalysis.org/1.0.0/documentation_pages/analysis/align.html#MDAnalysis.analysis.align.alignto), it is said:

```
Warning

The atom order for mobile and reference is only preserved when select is either “all” or None. In any other case, a new selection will be made that will sort the resulting AtomGroup by index and therefore destroy the correspondence between the two groups. It is safest not to mix ordered AtomGroups with selection strings.
```

So I would imagine that is `select` keyword is used, the atomgroup will be sorted by index before passing for alignment, which is not the case.

## Actual behavior ##

The atom group won't be sorted even when the select is used.

```python
>>> from MDAnalysisTests.datafiles import PDB_helix
>>> from MDAnalysis.analysis.align import alignto
>>> mobile = u.atoms[:4]
>>> mobile.ids
array([1, 2, 3, 4])
>>> mobile.ix
array([0, 1, 2, 3])
>>> ref = sum([u.atoms[3], u.atoms[2], u.atoms[1], u.atoms[0]])
>>> ref.ids
array([4, 3, 2, 1])
>>> ref.ix
array([3, 2, 1, 0])
>>> alignto(mobile, ref, select='bynum 1-4')
  File "<input>", line 1, in <module>
  File "~/GitHub/mdanalysis/package/MDAnalysis/analysis/align.py", line 492, in alignto
    match_atoms=match_atoms)
  File "~/GitHub/mdanalysis/package/MDAnalysis/analysis/align.py", line 1367, in get_matching_atoms
    raise SelectionError(errmsg)
MDAnalysis.exceptions.SelectionError: Inconsistent selections, masses differ by more than 0.1; mis-matching atoms are shown above.
>>> alignto(mobile, ref, select='bynum 1-3')
  File "<input>", line 1, in <module>
  File "~/GitHub/mdanalysis/package/MDAnalysis/analysis/align.py", line 492, in alignto
    match_atoms=match_atoms)
  File "~/GitHub/mdanalysis/package/MDAnalysis/analysis/align.py", line 1367, in get_matching_atoms
    raise SelectionError(errmsg)
MDAnalysis.exceptions.SelectionError: Inconsistent selections, masses differ by more than 0.1; mis-matching atoms are shown above.
```

So mobile and ref are essentially the same things with a different order. Using the select should sort them and thus, give them the same order, which is not the case. Related to #2976 

- Which operating system? osx

1ut:
Fix issue #2977
Related to PR #3188

 - Made MDAnalysis.analysis.align.alignto to reindex given mobile and reference atoms before passing for alignment.

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?


testsuite/MDAnalysisTests/analysis/test_align.py
```diff
@@ -386,6 +386,11 @@ def test_alignto_partial_universe(self, universe, reference):
         assert_array_almost_equal(segB_bound.positions, segB_free.positions,
                                   decimal=3)
 
+    def test_alignto_sort(self, universe):
+        mobile = universe.atoms[:4]
+        ref = universe.atoms[[3, 2, 1, 0]]
+        assert align.alignto(mobile, ref, select='bynum 1-4') == (0.0, 0.0)
```
tylerjereddy, line 392: I sometimes get a bit nervous on exact equality checks with floating point calculations, though probably not a big deal here. `np.assert_allclose()` may be slightly less risky, though I suppose one could argue the values really should be very close to identical.

tylerjereddy (commented):
For completeness, I checked that the example code in the matching issue still fails on `develop` and works as expected using this feature branch. (verified in part because a different source file/universe is used for the test here)

So, it seems to be the behavior the docs describe for this case, if I understand correctly.

testsuite/MDAnalysisTests/analysis/test_align.py
1ut: Fixed the code. I didn't take the margin of error of floating-point calculations seriously. 
https://github.com/MDAnalysis/mdanalysis/pull/3232/commits/109714eece66c6f231f8b8362ea4cce4d767bb6c

orbeckst (approved):
In principle this looks very good to me. It fixes the original issue #2977 . 

At the issue I posed the question if we actually _want_ the sorting behavior or if we rather prefer the new behavior with selections maintaining resorted AtomGroups. 

If the consensus is to keep the currently documented behavior then it can be merged as is.

orbeckst:
@lilyminium will your PR #3368 make this PR superfluous?

IAlibay:
> In principle this looks very good to me. It fixes the original issue #2977 .
> 
> At the issue I posed the question if we actually _want_ the sorting behavior or if we rather prefer the new behavior with selections maintaining resorted AtomGroups.
> 
> If the consensus is to keep the currently documented behavior then it can be merged as is.

Tagging @lilyminium here, how does this change with #3368 ?

lilyminium:
> @lilyminium will your PR #3368 make this PR superfluous?

Yeah sorry, there's a test for #2977 in it too.

orbeckst:
@1ut thank you for having provided this PR. The underlying root cause of the problem #2977 was a pretty big thing and needed to be fixed comprehensively, which was done in in PR #3368. Therefore, we will close this PR without merging. Again, thank you for the effort — you gave us a good backup solution that we could have used immediately if we had had other unforeseen problems. We hope to see more contributions from you in the future!

PR #3194, opened 2021-03-29
Deprecation warning for results attribute

PicoCentauri:
Following the discussion in #3115 this PR adds a deprecation warning stating that the results attribute of LinearDensity, PersistenceLength and GNMAnalysis will change in MDAnalysis version 2.0.

PR #3233, opened 2021-04-16
RFC: Add results class for storing analysis results

### LINKED ISSUE #3115: Common AnalysisBase framework for storing results
I'm working together with @joaomcteixeira to implement a [command line interface](https://github.com/PicoCentauri/mda_cli) for MDAnalysis. Our implementation is mostly finished 🥳. Currently, we are preparing the repo for moving to the mda organization and make it ready for CI, PYPI, Conda, etc. 

However, one important part is still missing. For saving the data a common structure for the results of each analysis class would be helpful. We already started a discussion but since some structural changes to the Analysis Framework and most likely to all analysis classes are necessary I move the discussion.

@orbeckst suggested three possibilities and I will use the rdf class as an example for the API. The current implementation for running and accessing the results looks like 

```python
rdf = InterRDF(ag1, ag2).run()
plt.plot(rdf.bins, rdf.rdf)
```

## 1. Reserved attribute names
For example, `self.results` could always be a data structure containing computed data. This data structure could either be 
a dictionary 

```python
rdf = InterRDF(ag1, ag2).run()
plt.plot(rdf.results["bins"], rdf.results["rdf"])
```

or attributes

```python
rdf = InterRDF(ag1, ag2).run()
plt.plot(rdf.results.bins, rdf.results.rdf)
```

## 2. Trailing underscore attributes (like sklearn-style)
In scikit-learn, computed variables get an underscore affixed (e.g. `.results_`).

```python
rdf = InterRDF(ag1, ag2).run()
plt.plot(rdf.bins_, rdf.rdf_)
```

## 3. Flexible annotation
A dictionary containing a list of attributes for computed data. This dictionary could look like

```python
self._annotation = {'bins': 'OUTPUT', 'rdf': 'OUTPUT',}
```
 
Here, `OUTPUT` could mark anything that is always computed after `run()` and should be saved. Possible other tags such as "OPTIONAL" for something that might be computed by an auxiliary method. With this approach only a few changes to the classes themselves are necessary. Their python api stays untouched as shown above.

---

My favorite is the first approach using attributes. It combines all results into a structure common for all classes. For me as a user this also handy. I can quickly access all results without always checking the documentation for the correct name.

Regardless of the method we choose, the MDACLI will find 'results' attributes and save them. If a numpy array, we could use `np.save()` to CSV, if it's a `pd.DataFrame`, we use `to_csv()`...

If you have better ideas we are of course open for discussion.

PicoCentauri:
Fixes #3115 

This PR adds a `_Results` class to the `AnalysisBase` for similarly storing results across all analysis modules. One could think about using Python's "new" [dataclass](https://docs.python.org/3/library/dataclasses.html#module-dataclasses) decorator when Py3.6 is dropped.

I would like to know what you think about the approach and I'm open to discussing improvements or other suggestions.
The docs definitely need some polishing.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/analysis/__init__.py
```diff
@@ -27,6 +27,20 @@
 
 The :mod:`MDAnalysis.analysis` sub-package contains various recipes and
 algorithms that can be used to analyze MD trajectories.
+If not stated differently, an analysis conducted by the available modules
+always follows the same structure
+
+1. Initiliaze the object prevously imported.
```
tylerjereddy, line 33: `Initiliaze` -> `Initialize`
`prevously` -> `previously`

package/MDAnalysis/analysis/__init__.py
```diff
@@ -27,6 +27,20 @@
...
+2. Run the analysis for specific trajectory slices
+3. Acces the analysis from the `results` attribute (if available)
```
tylerjereddy, line 35: `Acces` -> `Access`

package/MDAnalysis/analysis/__init__.py
```diff
@@ -27,6 +27,20 @@
...
+
+ .. code-block:: python
+
+    from MDAnalysis.anlysis import AnalysisModule
```
tylerjereddy, line 39: `MDAnalysis.anlysis` -> `MDAnalysis.analysis`

package/MDAnalysis/analysis/base.py
```diff
@@ -40,17 +40,63 @@
 logger = logging.getLogger(__name__)
 
 
+class _Results(object):
+    r"""Class storing results obtained from an analysis.
+
+    The class is stores all results obatined from
```
tylerjereddy, line 46: `is stores` -> `stores`

package/MDAnalysis/analysis/base.py
```diff
@@ -198,21 +248,22 @@ def run(self, start=None, stop=None, step=None, verbose=None):
...
-    >>> rot = AnalysisFromFunction(rotation_matrix, trajectory, mobile, ref).run()
-    >>> print(rot.results)
+        rot = AnalysisFromFunction(rotation_matrix, trajectory, 
+                                   mobile, ref).run()
+        print(rot.results)
+
+    Attributes
+    ----------
+    results : asarray
+        calculation results for each frame opf the underlaying function
```
tylerjereddy, line 265: `opf` -> `of`

package/MDAnalysis/analysis/base.py
```diff
@@ -40,17 +40,63 @@
 logger = logging.getLogger(__name__)
 
 
+class _Results(object):
```
tylerjereddy, line 43: I think we can omit the `object` inheritance for `develop` branch Python 3.x

package/MDAnalysis/analysis/lineardensity.py
```diff
@@ -171,38 +204,38 @@ def _single_frame(self):
...
 
     def _conclude(self):
         k = 6.022e-1  # divide by avodagro and convert from A3 to cm3
 
         # Average results over the  number of configurations
         for dim in ['x', 'y', 'z']:
-            for key in ['pos', 'pos_std', 'char', 'char_std']:
-                self.results[dim][key] /= self.n_frames
+            for key in self._keys:
+                self._results[f"{dim}_{key}"] /= self.n_frames
```
tylerjereddy, line 216: I see a bit of somewhat-complex discussion about backward compatibility in the matching issue. Maybe a summary of what is going on here would be useful?

You want to use `self.results` to hold an instance of the new common `_Results` class, so you are displacing the old use of `self.results`, but also changing its data structure?

I guess the usual questions about needed deprecations come up (depends on the policy adopted for MDA 2.0 I suppose).

PicoCentauri: > You want to use self.results to hold an instance of the new common _Results class, so you are displacing the old use of self.results, but also changing its data structure?

I'm not sure if I got your question right but I'm always using the new data structure. Everything is saved to the attributes of the new _Results instance as defined in L145.

> I guess the usual questions about needed deprecations come up (depends on the policy adopted for MDA 2.0 I suppose).

 We already gave a warning that the structure of these modules will change in MDA 2.0 (PR https://github.com/MDAnalysis/mdanalysis/pull/3194)

package/MDAnalysis/analysis/lineardensity.py
```diff
@@ -81,6 +111,9 @@ class LinearDensity(AnalysisBase):
 
     .. versionchanged:: 1.0.0
        Changed `selection` keyword to `select`
+
+    .. versionchanged:: 2.0.0
+        Changed structure of the the `results` dictionary to an object.
```
tylerjereddy, line 116: Everything is an object in Python

tylerjereddy: Also, double use of "the"

package/MDAnalysis/analysis/lineardensity.py
```diff
@@ -109,14 +140,16 @@ def __init__(self, select, grouping='atoms', binsize=0.25, **kwargs):
         self.nbins = bins.max()
         slices_vol = self.volume / bins
 
-        self.keys = ['pos', 'pos_std', 'char', 'char_std']
+        self._keys = ['pos', 'pos_std', 'char', 'char_std']
+        # Create an alias for the results object to save writing
```
tylerjereddy, line 144: What do you mean by "save writing"? Reducing IO operations, or you just want to re-use the old code in some way? There used to be a `self.results` object, but there is now a new `_Results` instance stored in `self.results`. You're saying that using `self._results` allows you to use more concise code than working with the instance directly, or to make less changes to the old code?

package/MDAnalysis/analysis/lineardensity.py
tylerjereddy: To be fair, that warning was only merged in 20 days ago.

tylerjereddy: It looks like MDA core devs are in favor in any case--I'd probably lean a bit more conservatively, but fair enough.

tylerjereddy: > I'm not sure if I got your question right but I'm always using the new data structure.

Ok, I think I found/still find the use of the alias with the underscore a bit confusing maybe.

package/MDAnalysis/analysis/lineardensity.py
PicoCentauri: The latter one. I want to make as less changes to the old code as possible. If this is confusing I can also remove this attribute.

package/MDAnalysis/analysis/lineardensity.py
PicoCentauri: > To be fair, that warning was only merged in 20 days ago.

Sorry, for the delay but it took us a while to figure out how to structure the changes.

> It looks like MDA core devs are in favor in any case--I'd probably lean a bit more conservatively, but fair enough.

You would like to stick to dictionaries instead of named attributes? I love dictionaries if it is your own project and you are the only one accessing their items. If other people are involved these named attributes are convenient  allowing autocompletion is most IDEs.

> Ok, I think I found/still find the use of the alias with the underscore a bit confusing maybe.
 Write

As stated above, I'm totally fine with removing the alias.

tylerjereddy: >  If other people are involved these named attributes are convenient allowing autocompletion is most IDEs.

I'm just saying that autocompletion on its own probably doesn't justify breaking user code without a pretty long/advertised deprecation period on the latest stable release. If you're writing a command line tool you could always write a shim layer to make it more convenient for your purposes.

I think the devs were "ok" with the breaking change though because of the bump to 2.0, so my nitpicking is probably irrelevant.

PicoCentauri: I get your point and I'm usually also against breaking changes. Of course we could write a parser checking different  types but a universal attribute for all classes sounds reasonable for breaking some code - at least for me.

I don't think this is nitpicking. A RFC PR is made for being discussed :).

IAlibay:
Just going to ping all the @MDAnalysis/coredevs this is central to the 2.0 release so the more eyes the better.

package/MDAnalysis/analysis/base.py
```diff
@@ -198,21 +256,22 @@ def run(self, start=None, stop=None, step=None, verbose=None):
...
 
-    >>> rot = AnalysisFromFunction(rotation_matrix, trajectory, mobile, ref).run()
-    >>> print(rot.results)
+        rot = AnalysisFromFunction(rotation_matrix, trajectory,
+                                   mobile, ref).run()
+        print(rot.results)
+
+    Attributes
+    ----------
+    results : asarray
```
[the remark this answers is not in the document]
orbeckst: How does AnalysisFromFunction now store results? This should be consistent with normal AnalysisBase classes. Can you explain a bit more?

If necessary, introduce `results.values` and `results.times` (and maybe `results.frames` ) to always store the times from the frames and whatever was computed. `results.frames` could be useful when times are unreliable or frames are needed for further indexing, e.g., when you want to get a specific trajectory frame based on a value such as in this contrived example where I want to do postprocessing on all frames where I detected that a residue had a certain z position:

```python
zpos = AnalysisFromFunction(lambda ag: ag.residue[0]., u, u.select_atoms("resname ASP and resid 157")).run(start=1000, step=100)
above_zero = zpos.results.values > 0
above_zero_frames = zpos.results.frames[above_zero]
for ts in u.trajectory[above_zero_frames]:
   print(ts)
   # do some additional analysis only on the selected frames
```

orbeckst (changes requested):
* I think the `_Results` class could be made more dict like, which would also help with integrating it into the existing classes.
* I am not clear how AnalysisFromFunction will use the new results class.

package/MDAnalysis/analysis/base.py
```diff
@@ -40,17 +40,71 @@
...
+            str_repr += ", ".join(attribute_list)
+        else:
+            str_repr += ", ".join(attribute_list[:3])
+            str_repr += " ... " + ", ".join(attribute_list[-3:])
+
+        return str_repr + ">"
+
+    def __eq__(self, other):
+        if self.__dict__.keys() != other.__dict__.keys() :
+            raise TypeError("Can't compare results from with different"
```
mnmelo, line 88:
```suggestion
            raise TypeError("Can't compare results with different"
```

package/MDAnalysis/analysis/base.py
```diff
@@ -40,17 +40,71 @@
...
+                            " attributes.")
+        return self.__dict__ == other.__dict__
+
+    def __ne__(self, other):
```
mnmelo, line 92: Defining `__ne__` as this here seems unneeded, as the [datamodel](https://docs.python.org/3/reference/datamodel.html#object.__ne__) already delegates it by default to be the negation of the result of `__eq__`.

package/MDAnalysis/analysis/base.py
```diff
@@ -40,17 +40,71 @@
 logger = logging.getLogger(__name__)
 
 
+class _Results:
```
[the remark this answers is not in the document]
PicoCentauri: > Is this just a Namespace that is linked to an AnalysisClass? Is the link necessary or could it just be a simple container?

I don't know if we want the class to be available in public. In my opinion it should only live `AnalysisClasses`...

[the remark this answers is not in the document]
PicoCentauri: > Note there's also "[Bunch](https://scikit-learn.org/stable/modules/generated/sklearn.utils.Bunch.html)" in scikit-learn that does similar things, but it's essentially a `dict` that allows attrib access.

Ahh this looks like a nice structure 🤩. I will change the style of the class accordingly.

package/MDAnalysis/analysis/gnm.py
```diff
@@ -239,7 +254,10 @@ def __init__(self,
         self.u = universe
         self.select = select
         self.cutoff = cutoff
-        self.results = []  # final result
+        self.results = _Results(self)
+        self.results.times = []
+        self.results.eigenvalues = []
+        self.results.eigenvectors = []
```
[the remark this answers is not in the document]
PicoCentauri: Done.

package/MDAnalysis/analysis/gnm.py
```diff
@@ -268,9 +286,10 @@ def _generate_output(self, w, v, outputobject, time, matrix,
                         w[list_map[1]],
                         item[1],
                         file=oup)
-        outputobject.append((time, w[list_map[1]], v[list_map[1]]))
-        # outputobject.append((time, [ w[list_map[i]] for i in range(nmodes) ],
-        # [ v[list_map[i]] for i in range( nmodes) ] ))
+
+        outputobject.times.append(time)
+        outputobject.eigenvalues.append(w[list_map[1]])
+        outputobject.eigenvectors.append(v[list_map[1]])
```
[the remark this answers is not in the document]
PicoCentauri: Done.

PicoCentauri:
I'm wondering and don't see a reason why `GNMAnalysis` is not inheriting from `AnalysisBase`. I could adjust this here or create a different PR to fix this.

package/MDAnalysis/analysis/base.py
PicoCentauri: Done.

package/MDAnalysis/analysis/polymer.py
```diff
@@ -187,6 +187,8 @@ class PersistenceLength(AnalysisBase):
        The run method now automatically performs the exponential fit
     .. versionchanged:: 1.0.0
        Deprecated :meth:`PersistenceLength.perform_fit` has now been removed.
+    .. versionchanged:: 2.0.0
+        Results are stored in attribute names and not as a numpy.ndarray.
```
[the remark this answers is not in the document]
PicoCentauri: Done.

package/MDAnalysis/analysis/lineardensity.py
```diff
@@ -55,19 +55,49 @@ class LinearDensity(AnalysisBase):
...
+    results.z_pos : numpy.ndarray
+           mass density in z direction
+    results.z_pos_std : numpy.ndarray
+           standard deviation of the mass density in z direction
+    results.z_char : numpy.ndarray
+           charge density in z direction
+    results.z_char_std : numpy.ndarray
+           standard deviation of the charge density in z direction
+    results.z_slice_volume : float
+           volume of bin in z direction
```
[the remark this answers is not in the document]
PicoCentauri: :)

package/MDAnalysis/analysis/base.py
```diff
@@ -198,21 +256,22 @@ def run(self, start=None, stop=None, step=None, verbose=None):
...
-    >>> def rotation_matrix(mobile, ref):
-    >>>     return mda.analysis.align.rotation_matrix(mobile, ref)[0]
+        def rotation_matrix(mobile, ref):
+            return mda.analysis.align.rotation_matrix(mobile, ref)[0]
 
-    >>> rot = AnalysisFromFunction(rotation_matrix, trajectory, mobile, ref).run()
-    >>> print(rot.results)
+        rot = AnalysisFromFunction(rotation_matrix, trajectory,
+                                   mobile, ref).run()
+        print(rot.results)
```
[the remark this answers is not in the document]
PicoCentauri: `results.timeseries` sounds good.

package/MDAnalysis/analysis/base.py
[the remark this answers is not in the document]
PicoCentauri: Good idea. `times` and `frames` are already saved as `self.times` and `self.frames`. I will just link them to results.

orbeckst:
> I'm wondering and don't see a reason why GNMAnalysis is not inheriting from AnalysisBase. I could adjust this here or create a different PR to fix this.

History... very old code.

Would be awesome to have it AnalysisBase-derived but better in a different PR – small, self-contained PRs are more easily reviewed and merged.

orbeckst:
If you raise an issue then you can attach the PR to the issue for making GNM analysis canonical.

package/MDAnalysis/analysis/base.py
```diff
@@ -40,17 +40,59 @@
...
+        super().__init__(kwargs)
+
+    def __setattr__(self, key, value):
+        self[key] = value
+
+    def __dir__(self):
+        return self.keys()
+
+    def __getattr__(self, key):
+        if key in self.keys():
```
richardjgowers, line 77: This should probably be a try/except, otherwise you're looking up the key twice.

package/MDAnalysis/analysis/base.py
```diff
@@ -40,17 +40,59 @@
 logger = logging.getLogger(__name__)
 
 
+class Results(dict):
```
richardjgowers, line 43: I'd rather this was `_Results` so we're not committed to supporting use of this externally.

orbeckst: I will become a hard requirement in the Analysis API if we go forward with this PR — I don't see a way around making this explicit and well defined. And once we do this, we might as well make it public.

In order to be compatible with mdacli, all analysis classes _must_ put results into this data structure (right @PicoCentauri @joaomcteixeira ?). In order to ensure that all our classes are properly interoperable we will have to require that `Results` (or rather the instance `AnalysisBases.results`) be used by everyone — including future MDKits.

package/MDAnalysis/analysis/base.py
PicoCentauri: Good point. I changed this.

package/MDAnalysis/analysis/base.py
PicoCentauri: My first thought was also to make it private but Oliver is right. The CLI will scan for the existence and only take modules into account with `AnalysisBases.results`. Without a well defined documentation it is probably hard for developers to implement results in their classes.

orbeckst:
With GNM now AnalysisBase-based #3244, could you please update the PR, @PicoCentauri ?

orbeckst:
I am pretty much ready to approve this PR and the use of the `Results` class for holding results.

@MDAnalysis/coredevs if you see issues with this approach, now is the time to make a blocking review.

package/MDAnalysis/analysis/__init__.py
```diff
@@ -27,6 +27,20 @@
...
+2. Run the analysis for specific trajectory slices
+3. Access the analysis from the `results` attribute (if available)
+
+ .. code-block:: python
+
+    from MDAnalysis.anlaysis import AnalysisModule
+
+    analysis_obj = AnalysisModule(<trajectory>, ...)
+    analysis_obj.run(<start>, <stop>, <step>)
+    print(analysis_obj.results)
```
orbeckst, line 43: We typically do not use "placeholder" syntax but just write descriptive variable names.

Also, is this text actually rendered anywhere in the docs?

package/MDAnalysis/analysis/base.py
```diff
@@ -40,17 +40,60 @@
...
+    r"""Container object for storing results.
+
+    Results are extend dictionaries by enabling values to be accessed by key,
+    `results["value_key"]`, or by an attribute, `results.value_key`. 
+    They store all results obatined from an analysis after the ``run`` call.
```
orbeckst, line 48: obtained

package/MDAnalysis/analysis/base.py
```diff
@@ -40,17 +40,60 @@
 logger = logging.getLogger(__name__)
 
 
+class Results(dict):
+    r"""Container object for storing results.
+
+    Results are extend dictionaries by enabling values to be accessed by key,
```
orbeckst, line 46:
```suggestion
    Results are dictionaries that provide two ways in which can values be accessed:
```

package/MDAnalysis/analysis/base.py
```diff
@@ -40,17 +40,60 @@
...
+    `results["value_key"]`, or by an attribute, `results.value_key`.
```
orbeckst, line 47: In reST, python code needs double backticks
```suggestion
    by dict key ``results["value_key"]`` or by object attribute, ``results.value_key``. 
```

package/MDAnalysis/analysis/base.py
```diff
@@ -40,17 +40,60 @@
...
 
 
+class Results(dict):
+    r"""Container object for storing results.
+
+    Results are extend dictionaries by enabling values to be accessed by key,
+    `results["value_key"]`, or by an attribute, `results.value_key`. 
+    They store all results obatined from an analysis after the ``run`` call.
+
+    The current implementation is similar to the `Bunch` class in `sklearn`.
```
orbeckst, line 50: double back ticks, and you can use
```suggestion
    The current implementation is similar to the :class:`sklearn.utils.Bunch` class in `scikit-learn`_.
  
    .. _`scikit-learn`: https://scikit-learn.org/
     
```

package/MDAnalysis/analysis/base.py
orbeckst: **IMPORTANT:** For the API it's important to say what the expectations for code are. Add here your API for Results. For example
* MDAnalysis does _not_ prescribe the names of the keys. This is up to the analysis class and must be documented by each analysis.
* MDAnalysis expects the following canonical set of keys to always  be present:  frames, times, .... where frames means this, times means that, etc (and yes, I know that times and frames are in AnalysisBase, I was trying to find quick examples)

package/MDAnalysis/analysis/base.py
orbeckst: See comment below: define the API in the docs.

package/MDAnalysis/analysis/base.py
orbeckst: If there are expected keys then the `Results` class should know about it. If there are no prescriptions then it's fine as it is.

package/MDAnalysis/analysis/base.py
```diff
@@ -40,17 +40,60 @@
...
+
+
 class AnalysisBase(object):
-    """Base class for defining multi frame analysis
+    r"""Base class for defining multi frame analysis
 
     The class it is designed as a template for creating multiframe analyses.
     This class will automatically take care of setting up the trajectory
     reader for iterating, and it offers to show a progress meter.
+    Computed results are stored inside the `results` attribute.
```
orbeckst, line 90: Use mark up
```suggestion
    Computed results are stored inside the :attr:`results` attribute.
```

package/MDAnalysis/analysis/base.py
```diff
@@ -40,17 +40,60 @@
...
+    >>> results.c = [1, 2, 3, 4]
+    >>> results['c']
+    [1, 2, 3, 4]
+    """
+
+    def __init__(self, **kwargs):
+        super().__init__(kwargs)
+
+    def __setattr__(self, key, value):
+        self[key] = value
```
orbeckst, line 71: Should you filter for any `value in dir(dict)` so that people do not define keys that they cannot get with `getattr`? In particular "values" would be an obvious one...

package/MDAnalysis/analysis/base.py
```diff
@@ -198,25 +248,33 @@ def run(self, start=None, stop=None, step=None, verbose=None):
...
 class AnalysisFromFunction(AnalysisBase):
-    """
-    Create an analysis from a function working on AtomGroups
+    r"""Create an :class:`AnalysisBase` from a function working on AtomGroups
 
-    Attributes
-    ----------
-    results : ndarray
-        results of calculation are stored after call to ``run``
+    .. code-block:: python
```
orbeckst, line 253: This should still be under an Example heading.

And Examples come after Attributes etc, according to https://userguide.mdanalysis.org/stable/contributing_code.html#guidelines-for-writing-docstrings and the numpydoc standard https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard

package/MDAnalysis/analysis/base.py
```diff
@@ -198,25 +248,33 @@ def run(self, start=None, stop=None, step=None, verbose=None):
...
+    results.timeseries : asarray
+        Results for each frame of the underlaying function
+        stored after call to ``run``.
 
     Raises
     ------
     ValueError : if ``function`` has the same kwargs as ``BaseAnalysis``
+
+    .. versionchanged:: 2.0.0
+        Former `results` are now stored as `results.timeseries`
```
orbeckst, line 277: use markup
```suggestion
        Former :attr:`results` are now stored as :attr:`results.timeseries`
```
It won't link but be formatted consistently.

package/MDAnalysis/analysis/base.py
```diff
@@ -261,33 +319,57 @@ def __init__(self, function, trajectory=None, *args, **kwargs):
...
+        self.results.timeseries = np.asarray(self.results.timeseries)
 
 
 def analysis_class(function):
-    """
-    Transform a function operating on a single frame to an analysis class
+    r"""Transform a function operating on a single frame to an 
+    :class:`AnalysisBase` class.
+
+    For an usage in a library we recommend the following style
```
orbeckst, line 338: For use in the library ...

Please move to Example section at the end of the docs.

(I now realize that you've been following the existing docs, which probably predate our adherence to numpydocs formatting style.)

package/MDAnalysis/analysis/base.py
```diff
@@ -261,33 +319,57 @@ def __init__(self, function, trajectory=None, *args, **kwargs):
...
+        rot = RotationMatrix(u.trajectory, mobile, ref).run(step=2)
+        print(rot.results.timeseries)
+
+    Attributes
+    ----------
+    results.frames : numpy.ndarray
+            simulatiom frames taken for evaluation
+    results.times : numpy.ndarray
+            simulatiom times taken for evaluation
+    results.timeseries : asarray
```
orbeckst, line 363: What does "asarray" mean? Why no ndarray?

package/MDAnalysis/analysis/base.py
```diff
@@ -261,33 +319,57 @@ def __init__(self, function, trajectory=None, *args, **kwargs):
...
+        stored after call to ``run``.
+
+    Raises
+    ------
+    ValueError : if ``function`` has the same kwargs as ``BaseAnalysis``
 
-    >>> rot = RotationMatrix(u.trajectory, mobile, ref).run(step=2)
-    >>> print(rot.results)
+    .. versionchanged:: 2.0.0
+        Former `results` are now stored as `results.timeseries`
```
orbeckst, line 372: please use markup

package/MDAnalysis/analysis/gnm.py
```diff
@@ -217,6 +221,15 @@ class GNMAnalysis(object):
           `Bonus_groups` is contained in `selection` as this could lead to
           double counting. No checks are applied. Default is ``None``.
 
+    Attributes
```
orbeckst, line 224: please update against newly merged code

package/MDAnalysis/analysis/gnm.py
```diff
@@ -228,6 +241,9 @@ class GNMAnalysis(object):
     .. versionchanged:: 1.0.0
        Changed `selection` keyword to `select`
 
+    .. versionchanged:: 2.0.0
+        Results are stored as attributes `times`, `eigenvalues` and 
+        `eigenvectors` of the `results` attribute.
```
orbeckst, line 246: please use role markup (or no markup but consistently – I like markup better)

package/MDAnalysis/analysis/polymer.py
```diff
@@ -187,6 +187,8 @@ class PersistenceLength(AnalysisBase):
        The run method now automatically performs the exponential fit
     .. versionchanged:: 1.0.0
        Deprecated :meth:`PersistenceLength.perform_fit` has now been removed.
+    .. versionchanged:: 2.0.0
+       Former `results` are now stored as `results.bond_autocorrelation`
```
orbeckst, line 191: consistent markup please

testsuite/MDAnalysisTests/analysis/test_base.py
```diff
@@ -33,6 +33,23 @@
 from MDAnalysisTests.util import no_deprecated_call
 
 
+class Test_Results:
```
orbeckst, line 36: If you filter forbidden keys then you'll also need a test for it.

orbeckst (changes requested):
I am generally happy with the approach. 

The two major things I see
- document what API should contain (or that it contain anything)
- prevent Results to have keys that collide with dict methods/attributes

Otherwise I have various minor doc things – please see comments.

testsuite/MDAnalysisTests/analysis/test_base.py
PicoCentauri: Currrently, I don't know any that we forbid.

package/MDAnalysis/analysis/__init__.py
PicoCentauri: Actually no. At least I didn't found this.

package/MDAnalysis/analysis/base.py
```diff
@@ -40,17 +41,89 @@
...
+    >>> results['c']
+    [1, 2, 3, 4]
+
+    Raises
+    ------
+    ValueError
+        If a to assigned attribute has the same name as a default dictionary
+        attribute.
+
+    ValueError
```
orbeckst, line 78: TypeError

package/MDAnalysis/analysis/base.py
```diff
@@ -40,17 +41,89 @@
...
+    >>> results['a']
+    3
+    >>> results.c = [1, 2, 3, 4]
+    >>> results['c']
+    [1, 2, 3, 4]
+
+    Raises
+    ------
+    ValueError
+        If a to assigned attribute has the same name as a default dictionary
```
orbeckst, line 75: If a attribute would have the same name ...

package/MDAnalysis/analysis/base.py
```diff
@@ -198,25 +279,36 @@ def run(self, start=None, stop=None, step=None, verbose=None):
...
-    >>> def rotation_matrix(mobile, ref):
-    >>>     return mda.analysis.align.rotation_matrix(mobile, ref)[0]
+    .. code-block:: python
+
+    def rotation_matrix(mobile, ref):
+        return mda.analysis.align.rotation_matrix(mobile, ref)[0]
+
+    rot = AnalysisFromFunction(rotation_matrix, trajectory,
+                                mobile, ref).run()
+    print(rot.results.timeseries)
```
orbeckst, line 303: I think needs to be indented to format correctly.

testsuite/MDAnalysisTests/analysis/test_base.py
```diff
@@ -33,6 +33,42 @@
...
+            results.c
+
+    def test_dir(self, results):
+        assert list(results.__dir__()) == ["a", "b"]
+
+    @pytest.mark.parametrize('key', dir(dict))
+    def test_existing_dict_attr(self, results, key):
+        msg = f"'{key}' is a protected dictionary attribute"
+        with pytest.raises(TypeError, match=key):
+         results[key] = None
```
orbeckst, line 56: indent 4 (PEP8?)

orbeckst (changes requested):
Looking very good, just minor doc/style comments.

orbeckst:
I didn't have time to see why the tests fail but obviously they should pass, too.

package/MDAnalysis/analysis/__init__.py
```diff
@@ -27,6 +27,20 @@
...
+2. Run the analysis for specific trajectory slices
+3. Access the analysis from the `results` attribute (if available)
+
+ .. code-block:: python
+
+    from MDAnalysis.anlaysis import AnalysisModule
+
+    analysis_obj = AnalysisModule(trajectory, ...)
+    analysis_obj.run(start_frame, stop_frame, step)
+    print(analysis_obj.results)
```
PicoCentauri, line 43: @orbeckst I could move some text to package/doc/sphinx/source/documentation_pages/analysis_modules.rst and remove everything else since its doubled and not rendered anywhere in the docs.

package/MDAnalysis/analysis/__init__.py
```diff
@@ -27,6 +27,20 @@
 
 The :mod:`MDAnalysis.analysis` sub-package contains various recipes and
 algorithms that can be used to analyze MD trajectories.
+If not stated differently, an analysis conducted by the available modules
```
lilyminium, line 30:
```suggestion
Unless stated otherwise, an analysis using the available modules
```

package/MDAnalysis/analysis/__init__.py
lilyminium: Yes please, it's a bit strange that this module is so well documented and then appears nowhere.

package/MDAnalysis/analysis/__init__.py
```diff
@@ -27,6 +27,20 @@
...
+always follows the same structure
+
+1. Initialize the object previously imported.
```
lilyminium, line 33: Maybe add a step where you import it?

package/MDAnalysis/analysis/__init__.py
```diff
@@ -27,6 +27,20 @@
...
+2. Run the analysis for specific trajectory slices
```
lilyminium, line 34:
```suggestion
2. Run the analysis, optionally for specific trajectory slices
```

package/MDAnalysis/analysis/__init__.py
```diff
@@ -27,6 +27,20 @@
...
+3. Access the analysis from the `results` attribute (if available)
```
lilyminium, line 35: I thought the point of this PR was to standardize things so it was always available?

package/MDAnalysis/analysis/__init__.py
```diff
@@ -27,6 +27,20 @@
...
+
+ .. code-block:: python
+
+    from MDAnalysis.anlaysis import AnalysisModule
+
+    analysis_obj = AnalysisModule(trajectory, ...)
+    analysis_obj.run(start_frame, stop_frame, step)
```
lilyminium, line 42:
```suggestion
    from MDAnalysis.analysis import ExampleAnalysisModule  # (e.g. RMSD)

    analysis_obj = ExampleAnalysisModule(universe, ...)
    analysis_obj.run(start_frame, stop_frame, step)
```

I think it's best to be clear. Usually analysis modules operate on universes, right? There's a `universe.trajectory` attribute that would be best not to mix up with it.

package/MDAnalysis/analysis/base.py
lilyminium: Is it best to subclass dict or UserDict?

package/MDAnalysis/analysis/base.py
```diff
@@ -40,18 +41,111 @@
 logger = logging.getLogger(__name__)
 
 
+class Results(dict):
+    r"""Container object for storing results.
+
+    Results are dictionaries that provide two ways in which can values be
+    accessed: by dict key ``results["value_key"]`` or by object attribute,
+    ``results.value_key``. They store all results obatined from an analysis
```
lilyminium, line 49:
```suggestion
    ``results.value_key``. They store all results obtained from an analysis
```

package/MDAnalysis/analysis/base.py
```diff
@@ -40,18 +41,111 @@
...
+    after the ``run`` call.
```
lilyminium, line 50:
```suggestion
    after calling :func:`run()`.
```

package/MDAnalysis/analysis/base.py
```diff
@@ -40,18 +41,111 @@
 logger = logging.getLogger(__name__)
 
 
+class Results(dict):
+    r"""Container object for storing results.
+
+    Results are dictionaries that provide two ways in which can values be
```
lilyminium, line 47:
```suggestion
    Results are dictionaries that provide two ways by which can values be
```

package/MDAnalysis/analysis/base.py
```diff
@@ -40,18 +41,111 @@
...
+    3
+    >>> results.c = [1, 2, 3, 4]
+    >>> results['c']
+    [1, 2, 3, 4]
+
+
+    Raises
+    ------
+    ValueError
+        If a to assigned attribute has the same name as a default dictionary
```
lilyminium:
```suggestion
        If an assigned attribute has the same name as a default
```

? I'm not sure what this is referring to

package/MDAnalysis/analysis/base.py
```diff
@@ -40,18 +41,111 @@
...
+                  and re.match("^[a-zA-Z0-9_]*$", key)):
+            raise TypeError(f"'{key}' is not able to be accessed by attribute")
+
+    def __init__(self, **kwargs):
+        for key in kwargs.keys():
+            self._validate_key(key)
+        super().__init__(kwargs)
+
+    def __setattr__(self, key, value):
+        self._validate_key(key)
```
lilyminium, line 98: Do you need to validate your key again? doesn't self[key] just call `__setitem__` which validates it anyway?

package/MDAnalysis/analysis/base.py
```diff
@@ -40,18 +41,111 @@
...
+
+    def _validate_key(self, key):
+        if key in dir(dict):
+            raise TypeError(f"'{key}' is a protected dictionary attribute")
+        elif not (isinstance(key, str) and key[0].isalpha()
+                  and re.match("^[a-zA-Z0-9_]*$", key)):
+            raise TypeError(f"'{key}' is not able to be accessed by attribute")
+
+    def __init__(self, **kwargs):
+        for key in kwargs.keys():
```
lilyminium, line 93:
```suggestion
        for key in kwargs:
```

package/MDAnalysis/analysis/base.py
```diff
@@ -40,18 +41,111 @@
...
+        try:
+            return self[key]
+        except KeyError as err:
+            raise AttributeError("'Results' object has no "
+                                 f"attribute '{key}'") from err
+
+
 class AnalysisBase(object):
-    """Base class for defining multi frame analysis
+    r"""Base class for defining multi frame analysis
```
lilyminium, line 117:
```suggestion
    r"""Base class for defining multi-frame analysis
```

package/MDAnalysis/analysis/base.py
```diff
@@ -40,18 +41,111 @@
...
 
     The class it is designed as a template for creating multiframe analyses.
```
lilyminium, line 119:
```suggestion
    The class is designed as a template for creating multi-frame analyses.
```

package/MDAnalysis/analysis/base.py
```diff
@@ -40,18 +41,111 @@
...
     This class will automatically take care of setting up the trajectory
     reader for iterating, and it offers to show a progress meter.
+    Computed results are stored inside the :attr:`results` attribute.
 
     To define a new Analysis, `AnalysisBase` needs to be subclassed
-    `_single_frame` must be defined. It is also possible to define
-    `_prepare` and `_conclude` for pre and post processing. See the example
-    below.
+    :meth:`_single_frame` must be defined. It is also possible to define
```
lilyminium, line 125:
```suggestion
    and :meth:`_single_frame` must be defined. It is also possible to define
```

package/MDAnalysis/analysis/base.py
```diff
@@ -40,18 +41,111 @@
...
+    :meth:`_prepare` and :meth:`_conclude` for pre and post processing.
```
lilyminium, line 126:
```suggestion
    :meth:`_prepare` and :meth:`_conclude` for pre- and post-processing.
```

package/MDAnalysis/analysis/base.py
```diff
@@ -40,18 +41,111 @@
...
+    All results should be stored as attributes of the :class:`Results`
+    container.
+
+    Parameters
+    ----------
+    trajectory : mda.Reader
```
lilyminium, line 132: Could you be clearer with the class path here -- like MDAnalysis.coordinates.base.ReaderBase ?

package/MDAnalysis/analysis/base.py
```diff
@@ -40,18 +41,111 @@
...
+        A trajectory Reader
+    verbose : bool, optional
+        Turn on more logging and debugging
 
+    Attributes
+    ----------
+    times: np.ndarray
+        array of Timestep times. Only exists after calling run()
```
lilyminium:
```suggestion
        array of Timestep times. Only exists after calling :meth:`run`
```

package/MDAnalysis/analysis/base.py
```diff
@@ -40,18 +41,111 @@
...
+    frames: np.ndarray
+        array of Timestep frame indices. Only exists after calling run()
```
lilyminium:
```suggestion
        array of Timestep frame indices. Only exists after calling :meth:`run`
```

package/MDAnalysis/analysis/base.py
```diff
@@ -65,53 +159,43 @@ def _prepare(self):
...
                # Data structures can be set up at this time
-               self.result = []
+               self.results.example_result = []
 
            def _single_frame(self):
                # REQUIRED
                # Called after the trajectory is moved onto each new frame.
-               # store result of `some_function` for a single frame
-               self.result.append(some_function(self._ag, self._parameter))
+               # store a example_result of `some_function` for a single frame
```
lilyminium, line 172:
```suggestion
               # store an example_result of `some_function` for a single frame
```

package/MDAnalysis/analysis/base.py
```diff
@@ -65,53 +159,43 @@ def _prepare(self):
...
 
-    Afterwards the new analysis can be run like this.
+    Afterwards the new analysis can be run like this:
 
     .. code-block:: python
 
        na = NewAnalysis(u.select_atoms('name CA'), 35).run(start=10, stop=20)
-       print(na.result)
+       print(na.results.example_result)
+       # results can also accessed by key
```
lilyminium, line 195:
```suggestion
       # results can also be accessed by key
```

package/MDAnalysis/analysis/base.py
```diff
@@ -198,46 +281,58 @@ def run(self, start=None, stop=None, step=None, verbose=None):
...
+        arguments for ``function``
+    **kwargs : dict
+        arguments for ``function`` and ``AnalysisBase``
 
     Attributes
     ----------
-    results : ndarray
-        results of calculation are stored after call to ``run``
+    results.frames : numpy.ndarray
+            simulatiom frames taken for evaluation
```
lilyminium, line 312:
```suggestion
            simulation frames used in analysis
```

package/MDAnalysis/analysis/base.py
```diff
@@ -198,46 +281,58 @@ def run(self, start=None, stop=None, step=None, verbose=None):
...
+    results.times : numpy.ndarray
+            simulatiom times taken for evaluation
```
lilyminium, line 314:
```suggestion
            simulation times used in analysis
```

package/MDAnalysis/analysis/base.py
```diff
@@ -198,46 +281,58 @@ def run(self, start=None, stop=None, step=None, verbose=None):
...
+    results.timeseries : numpy.ndarray
+            Results for each frame of the underlaying function
```
lilyminium, line 316:
```suggestion
            Results for each frame of the wrapped function,
```

package/MDAnalysis/analysis/base.py
```diff
@@ -198,46 +281,58 @@ def run(self, start=None, stop=None, step=None, verbose=None):
...
 
-    >>> rot = AnalysisFromFunction(rotation_matrix, trajectory, mobile, ref).run()
-    >>> print(rot.results)
 
     Raises
     ------
-    ValueError : if ``function`` has the same kwargs as ``BaseAnalysis``
-    """
+    ValueError
+        if ``function`` has the same ``kwargs`` as ``BaseAnalysis``
```
lilyminium, line 334: AnalysisBase?

package/MDAnalysis/analysis/base.py
```diff
@@ -261,33 +356,68 @@ def __init__(self, function, trajectory=None, *args, **kwargs):
...
+    function : callable
+        function to evaluate at each frame
 
-    >>> def rotation_matrix(mobile, ref):
-    >>>     return mda.analysis.align.rotation_matrix(mobile, ref)[0]
-    >>> RotationMatrix = analysis_class(rotation_matrix)
+    Attributes
+    ----------
+    results.frames : numpy.ndarray
+            simulatiom frames taken for evaluation
```
lilyminium, line 394:
```suggestion
            simulation frames used in analysis
```

package/MDAnalysis/analysis/base.py
```diff
@@ -261,33 +356,68 @@ def __init__(self, function, trajectory=None, *args, **kwargs):
...
+    results.times : numpy.ndarray
+            simulatiom times taken for evaluation
```
lilyminium, line 396:
```suggestion
            simulation times used in analysis
```

package/MDAnalysis/analysis/base.py
```diff
@@ -261,33 +356,68 @@ def __init__(self, function, trajectory=None, *args, **kwargs):
...
+    results.timeseries : numpy.ndarray
+            Results for each frame of the underlaying function
```
lilyminium, line 398:
```suggestion
            Results for each frame of the wrapped function,
```

package/MDAnalysis/analysis/base.py
```diff
@@ -261,33 +356,68 @@ def __init__(self, function, trajectory=None, *args, **kwargs):
...
+            stored after call to :meth:`AnalysisFromFunction.run`.
+
+    Examples
+    --------
 
-    It can also be used as a decorator:
+    For an use in a library we recommend the following style
```
lilyminium, line 404:
```suggestion
    For use in a library, we recommend the following style
```

package/MDAnalysis/analysis/base.py
```diff
@@ -261,33 +356,68 @@ def __init__(self, function, trajectory=None, *args, **kwargs):
...
+            return mda.analysis.align.rotation_matrix(mobile, ref)[0]
+
+        rot = RotationMatrix(u.trajectory, mobile, ref).run(step=2)
+        print(rot.results.timeseries)
+
+
+    Raises
+    ------
+    ValueError
+        if ``function`` has the same ``kwargs`` as ``BaseAnalysis``
```
lilyminium, line 427: AnalysisBase?

package/MDAnalysis/analysis/base.py
```diff
@@ -318,7 +448,8 @@ def _filter_baseanalysis_kwargs(function, kwargs):
 
     Raises
     ------
-    ValueError : if ``function`` has the same kwargs as ``BaseAnalysis``
+    ValueError
+        if ``function`` has the same ``kwargs`` as ``BaseAnalysis``
```
lilyminium, line 463: AnalysisBase?

package/MDAnalysis/analysis/gnm.py
```diff
@@ -204,27 +208,28 @@ class GNMAnalysis(AnalysisBase):
...
+          double counting. No checks are applied.
 
     Attributes
     ----------
-    results : list
-        GNM results per frame:
-            results = [(time,eigenvalues[1],eigenvectors[1]),
-                       (time,eigenvalues[1],eigenvectors[1]), ...]
+    results.times : numpy.ndarray
+            simulatiom times taken for evaluation
```
lilyminium, line 228:
```suggestion
            simulation times used in analysis
```

package/MDAnalysis/analysis/gnm.py
```diff
@@ -238,7 +243,9 @@ class GNMAnalysis(AnalysisBase):
        Changed `selection` keyword to `select`
 
     .. versionchanged:: 2.0.0
-       Use :class:`~MDAnalysis.analysis.AnalysisBase` as parent class.
+       Use :class:`~MDAnalysis.analysis.AnalysisBase` as parent class and
+       store results as attributes ``times``, ``eigenvalues`` and
+       ``eigenvectors`` of the `results` attribute.
```
lilyminium, line 248:
```suggestion
       ``eigenvectors`` of the ``results`` attribute.
```

package/MDAnalysis/analysis/gnm.py
```diff
@@ -251,7 +258,10 @@ def __init__(self,
         self.u = universe
         self.select = select
         self.cutoff = cutoff
-        self.results = []  # final result
+        self.results = Results()
+        self.results.times = []
```
lilyminium, line 262: Shouldn't `times` be handled in the base analysis class?

package/MDAnalysis/analysis/gnm.py
```diff
@@ -357,19 +372,27 @@ class closeContactGNMAnalysis(GNMAnalysis):
...
           If set to "size" (the default) then weight the contact by
           :math:`1/\sqrt{N_i N_j}` where :math:`N_i` and :math:`N_j` are the
           number of atoms in the residues :math:`i` and :math:`j` that contain
           the atoms that form a contact.
 
+    Attributes
+    ----------
+    results.times : numpy.ndarray
+            simulatiom times taken for evaluation
+    results.eiegenvalues : numpy.ndarray
```
lilyminium, line 391:
```suggestion
    results.eigenvalues : numpy.ndarray
```

package/MDAnalysis/analysis/gnm.py
```diff
@@ -357,19 +372,27 @@ class closeContactGNMAnalysis(GNMAnalysis):
...
     weights : {"size", None} (optional)
           If set to "size" (the default) then weight the contact by
           :math:`1/\sqrt{N_i N_j}` where :math:`N_i` and :math:`N_j` are the
           number of atoms in the residues :math:`i` and :math:`j` that contain
           the atoms that form a contact.
 
+    Attributes
+    ----------
+    results.times : numpy.ndarray
+            simulatiom times taken for evaluation
```
lilyminium, line 390:
```suggestion
            simulation times used in analysis
```

package/MDAnalysis/analysis/gnm.py
```diff
@@ -357,19 +372,27 @@ class closeContactGNMAnalysis(GNMAnalysis):
...
+    results.eiegenvalues : numpy.ndarray
+            calculated eigenvalues
+    results.eiegenvectors : numpy.ndarray
```
lilyminium, line 393:
```suggestion
    results.eigenvectors : numpy.ndarray
```

package/MDAnalysis/analysis/lineardensity.py
```diff
@@ -50,24 +50,58 @@ class LinearDensity(AnalysisBase):
...
+          Show detailed progress of the calculation if set to ``True``
 
     Attributes
     ----------
-    results : dict
-          Keys 'x', 'y', and 'z' for the three directions. Under these
-          keys, find 'pos', 'pos_std' (mass-weighted density and
-          standard deviation), 'char', 'char_std' (charge density and
-          its standard deviation), 'slice_volume' (volume of bin).
+    results.x.dim : int
```
lilyminium, line 57: Pedantically I'd change all the "in [xyz] direction"s below to "along [xyz] axis"s but up to you :-)

package/MDAnalysis/analysis/lineardensity.py
```diff
@@ -50,24 +50,58 @@ class LinearDensity(AnalysisBase):
...
+    results.z.char_std : numpy.ndarray
+           standard deviation of the charge density in z direction
+    results.z.slice_volume : float
+           volume of bin in z direction
 
     Example
     -------
-    First create a LinearDensity object by supplying a selection,
-    then use the :meth:`run` method::
+    First create a `LinearDensity` object by supplying a selection,
```
lilyminium, line 96:
```suggestion
    First create a ``LinearDensity`` object by supplying a selection,
```

Use double backticks for fun monospace in rendered docs -- single backticks if followed by a sphinx directive like `:meth:`. I believe single backticks italicise instead.

lilyminium (changes requested):
I saw you updated while I was reviewing so not all comments may be applicable @PicoCentauri. I've left comments on the docs :)

package/MDAnalysis/analysis/base.py
```diff
@@ -40,20 +41,118 @@
...
+    TypeError
+        If a key is not of type ``str`` and therefore is not able to be
+        accessed by attribute.
+    """
+
+    def _validate_key(self, key):
+        if key in dir(dict):
+            raise TypeError(f"'{key}' is a protected dictionary attribute")
+        elif not (isinstance(key, str) and key[0].isalpha()
+                  and re.match("^[a-zA-Z0-9_]*$", key)):
```
lilyminium, line 89: What does the regex do that `key[0].isalpha()` doesn't?

lilyminium: I imagine you'd also run into trouble if people sprinkled "~*=-" etc. after the first character.

lilyminium: Can you combine the isalpha into the regex with ``re.match("^[a-zA-Z_][a-zA-Z0-9_]*$")`` ?

lilyminium:
For the failing test try escaping the [.

```diff
--- a[
+++ a\[
```

package/MDAnalysis/analysis/base.py
PicoCentauri: Yes, good idea. But I think I got some errors with spaces...

package/MDAnalysis/analysis/base.py
lilyminium: https://treyhunner.com/2019/04/why-you-shouldnt-inherit-from-list-and-dict-in-python/#What%E2%80%99s_the_alternative_to_inheriting_from_list_and_dict?

package/MDAnalysis/analysis/base.py
PicoCentauri: True!

package/MDAnalysis/analysis/base.py
PicoCentauri: Thanks for figuring out. I will adjust the parent class.

package/MDAnalysis/analysis/__init__.py
PicoCentauri: No it wasn't I added these lines.

package/MDAnalysis/analysis/__init__.py
PicoCentauri: Okay copy everything or just use the beginning of the description. The modules are linked anyway right?

package/MDAnalysis/analysis/gnm.py
PicoCentauri: Yes we could remove it.

package/MDAnalysis/analysis/lineardensity.py
PicoCentauri: Good point.

package/MDAnalysis/analysis/base.py
orbeckst: Thanks @lilyminium , I learned something new.

package/MDAnalysis/analysis/base.py
```diff
@@ -85,8 +86,8 @@ class in `scikit-learn`_.
     def _validate_key(self, key):
         if key in dir(dict):
             raise TypeError(f"'{key}' is a protected dictionary attribute")
-        elif not (isinstance(key, str) and key[0].isalpha()
-                  and re.match("^[a-zA-Z0-9_]*$", key)):
+        elif not (isinstance(key, str)
+                  and re.match("^[a-zA-Z_][a-zA-Z0-9_]*$", key)):
             raise TypeError(f"'{key}' is not able to be accessed by attribute")
```
orbeckst: Is `TypeError` the appropriate exception here or should it be `ValueError` (because you provided an incorrect value)?

PicoCentauri:
First of all: Sorry that this PR blew up that much. :(

Finally, I restructured the docs of the Analysis module and included the `__init__.py` of MDAnalysis.analysis. I hope the docs are cleared and structured enough...

Now to the bottom line of this PR. I solved the infinite recursion problem when inheriting from `UserDict` instead of `dict` with an ugly workaround. If you have better ideas I'm happy to change my weird construct. However, since we now use a `UserDict` there appears a new problem for our container. A `UserDict` is only a wrapper for a dictionary that lives inside `self.data`. In general, this is no problem but here we also want to allow accessing keys and attributes resulting in the following problem:

 Constructions like 
```python
 >>> results = Results()
 >>> results["data"] = 1 
```

or 

```python
>>> Results(data=1)
```

are allowed but leading to inconsistencies when `data` is tried to access by attribute 

```python

 >>> results.data
 {'data': 1}
```

This is an artifact of the wrapper class `UserDict`, but I don't how to get around it. Denying a `data` attribute by adding a statement to the `_validate_key` method fails since the `init` method needs to define such an attribute. I tried several ways but was not able to found a way to allow the creation in the init and deny the change later. Any suggestions are welcome.

orbeckst:
I read the long post on UserDict, too, and from the reading it seemed like the way to go. Alternatively, one could use ABCs (mappable) as stated there.

However, if deriving from `dict` does not have problems with `update`, `setdefault`, `pop` etc then I don't see an immediate problem with going with `dict`. (If there really are issues, the underlying implementation can be fixed later, as long as the interface remains the same.) Note that sklearn [bunch](https://scikit-learn.org/stable/modules/generated/sklearn.utils.Bunch.html) also [derives from dict](https://github.com/scikit-learn/scikit-learn/blob/15a949460/sklearn/utils/__init__.py#L61).

(Btw, bunch has a special `__setstate__` method for pickles ... I assume this is a workaround for some of their earlier problems with pickling (according to their comments) but did you look into pickling `Results`? Is this even something we need to be concerned about?)

orbeckst:
> First of all: Sorry that this PR blew up that much. :(

Look at it that way: A RFC with < 50 comments is not successful — so your RFC is a success. ;-)

cbouy:
Hey, here's a way to fix some of the problems you encountered with `UserDict`:
```python
class Results(UserDict):
    def _validate_key(self, key):
        if key in dir(UserDict) or (key == "data" and self._dict_frozen):
            raise AttributeError(f"'{key}' is a protected dictionary attribute")
        elif isinstance(key, str) and not key.isidentifier():
            raise ValueError(f"'{key}' is not a valid attribute")

    def __init__(self, **kwargs):
        self._dict_frozen = False
        for key in kwargs:
            self._validate_key(key)
        super().__init__(**kwargs)
        self._dict_frozen = True

    def __setitem__(self, key, item):
        self._validate_key(key)
        super().__setitem__(key, item)

    def __setattr__(self, attr, value):
        self._validate_key(attr)
        super().__setattr__(attr, value)
    
    def __getattr__(self, attr):
        try:
            return self.data[attr]
        except KeyError as err:
            raise AttributeError("'Results' object has no "
                                 f"attribute '{attr}'") from err
```

The `self._dict_frozen` attribute will lock the `data` attribute only after the `init` method is done:
```python

r = Results(a=1, b=2)
r
# {'a': 1, 'b': 2}

r["4ttr1but3"] = "foo"
# ValueError: '4ttr1but3' is not a valid attribute

r["data"] = 1
# AttributeError: 'data' is a protected dictionary attribute

r.data = 1
# AttributeError: 'data' is a protected dictionary attribute
```

Also with your current implementation of `Results`, the `data` had become a class attribute instead of an instance attribute, so all instances of `Results` would share the same dict, which would be quite problematic.

Does the `Results` class needs to add the attributes set to be visible in the `data` dict as well ? i.e:
```python
r.c = 3
r
# {'a': 1, 'b': 2, 'c': 3}
```

PS: I took the liberty of changing the errors raised by `_validate_key` for consistency, and used the `str.isidentifier` method which does the same thing as your regex.

PicoCentauri:
Thanks for the fix @cbouy! It is actually what I indented. And yes set attributes should also be visible in the data dict. I will take a look.

package/MDAnalysis/analysis/base.py
```diff
@@ -65,53 +169,49 @@ def _prepare(self):
...
-        ----------
-        trajectory : mda.Reader
-            A trajectory Reader
-        verbose : bool, optional
-           Turn on more logging and debugging, default ``False``
 
+    .. versionchanged:: 1.0.0
+        Support for setting ``start``, ``stop``, and ``step`` has been
+        removed. These should now be directly passed to
+        :meth:`AnalysisBase.run`.
```
orbeckst, line 207: Add a versionchanged 2.0.0 for addition of the results attr.

package/MDAnalysis/analysis/base.py
```diff
@@ -40,20 +40,124 @@
...
+    >>> results['b']
+    2
+    >>> results.b
+    2
+    >>> results.a = 3
+    >>> results['a']
+    3
+    >>> results.c = [1, 2, 3, 4]
+    >>> results['c']
+    [1, 2, 3, 4]
```
orbeckst, line 82: add a versionadded 2.0.0 (note that you need two empty lines above)

package/MDAnalysis/analysis/__init__.py
```diff
@@ -22,11 +22,26 @@
...
+ .. code-block:: python
+
+    from MDAnalysis.analysis import ExampleAnalysisModule  # (e.g. RMSD)
+
+    analysis_obj = ExampleAnalysisModule(universe, ...)
+    analysis_obj.run(start_frame, stop_frame, step)
+    print(analysis_obj.results)
 
 If you use them please check if the documentation mentions any specific caveats
 and also if there are any published papers associated with these algorithms.
```
orbeckst, line 47: Please remove the whole section *Available analysis modules* — this list is not maintained anymore and it's confusing when it shows up in the docs.

orbeckst (changes requested):
Looking good, just some doc changes from my end.

orbeckst:
@PicoCentauri in order to get this into 2.0.0, this PR _must_ be merged by Friday at the very latest. Could you please address the few remaining issues? Thanks.

orbeckst:
It looks as if all of @lilyminium 's comments were addressed. If need be, I can dismiss her review (she said so on Discord).

PicoCentauri:
Thank for your suggestion @orbeckst. I implemented the doc changes. While removing the *available analysis* section I merged the docstring of the analysis' `__init__.py` into `analysis_modules.rst. For me this seemed more reasonable as an extra documentation page with only one paragraph.

orbeckst (approved):

orbeckst:
@lilyminium if you have a few minutes to spare to do a final look at this PR then please do. Otherwise I am happy to merge it tonight (so that we can add `Results` to other analysis classes – see https://github.com/MDAnalysis/mdanalysis/issues/3115#issuecomment-832753553 ).

@IAlibay , you said on discord you found some duplicate authorship/names — I didn't understand what that was about. Anything that needs to be fixed here?

IAlibay:
> @IAlibay , you said on discord you found some duplicate authorship/names — I didn't understand what that was about. Anything that needs to be fixed here?

O yeah, nothing major, there's just two `PicoCentauri` entries in the CHANGELOG for 2.0. We can just remove it when we update a few of the other `AnalysisBase` things to use `Results` (i.e. there's no point to triggering CI again just for that).

PicoCentauri:
> O yeah, nothing major, there's just two `PicoCentauri` entries in the CHANGELOG for 2.0. We can just remove it when we update a few of the other `AnalysisBase` things to use `Results` (i.e. there's no point to triggering CI again just for that).

Sorry, that was probably me that messed this up.

orbeckst:
@PicoCentauri and @joaomcteixeira – the new AnalysisBase is in 2.0.0 !

As @IAlibay said: let's get at least the most used AnalysisBase classes to also use .results (in a backward compatible way) so that we can teach the new way at the PRACE workshop.

joaomcteixeira:
Fantastic! All kudos for this implementation to @PicoCentauri. Congratz! :tada:

Issue #3176, opened 2021-03-16
Compiling with clang compilers on Ubuntu runners causes OpenMP linking issues


IAlibay:
## Expected behavior ##

Compiling by setting CC and CXX appropriately to capture the clang compilers should work on both MacOS and Ubuntu.

## Actual behavior ##

I'm in the midst of setting up a gh action cron job to do a compiler sweep for both clang and gcc, Turns out that if I attempt to set CC=clang and CXX=clang++ and then install MDAnalysis via setup I get OpenMP linking issues. It's not immediately clear to me if the issue is down to the gh action runners or generally an issue with Ubuntu.

Here is an example failed job: https://github.com/IAlibay/mdanalysis/runs/2109402006?check_suite_focus=true

hmacdope:
Looks like it might need `-fopenmp`.

hmacdope:
Some things appear to be being compiled with gcc.

`gcc -pthread -shared -B /usr/share/miniconda/envs/test/compiler_compat -L/usr/share/miniconda/envs/test/lib -Wl,-rpath=/usr/share/miniconda/envs/test/lib -Wl,--no-as-needed -Wl,--sysroot=/ build/temp.linux-x86_64-3.8/home/runner/work/mdanalysis/mdanalysis/package/MDAnalysis/lib/c_distances.o -lm -o build/lib.linux-x86_64-3.8/MDAnalysis/lib/c_distances.cpython-38-x86_64-linux-gnu.so
`

IAlibay:
yeah I noticed that, the gcc calls seems to be related to the areas where openmp is getting picked up.

IAlibay:
It looks like it might be down to our attempts to pick up OpenMP (so in the MacOS builds we build a serial version only), so something in that code path is ignoring the wishes of CC/CXX

hmacdope:
Is openmp pulled from conda/pip or apt?

IAlibay:
Here's an interesting one, looks like even when using gcc (in this case CC=gcc-7), you end up defaulting to "gcc" for the openmp sections: https://github.com/IAlibay/mdanalysis/runs/2109401928?check_suite_focus=true#step:8:119

hmacdope:
Perhaps `fopenmp=libomp` will work.

PR #3234, opened 2021-04-18
BLD: handle gcc on MacOS

### LINKED ISSUE #3109: BLD: compiler recognition on MacOS
## Expected behavior ##

Build system automagically handles both `clang` and  "genuine" `gcc` on MacOS. This command should work with `spack` built `gcc` compiler toolchain loaded on macOS 10.15.7: `python setup.py develop`

## Actual behavior ##

The build system thinks it is using `clang` but it is actually using `gcc` built with `clang` by `spack` (not the "alias" you would normally have on a "normal" mac machine).

```
gcc: error: unrecognized command line option '-stdlib=libc++'
error: command '/Users/treddy/github_projects/spack/opt/spack/darwin-catalina-x86_64/clang-11.0.0-apple/gcc-8.3.0-fxj4m5svuktydrjgtfkbxyifibwf6wsi/bin/gcc' failed with exit status 1
```

`python setup.py develop`

Python 3.7.4, though not likely relevant. Latest `develop` branch.

Hacky work around, that hints at part of code involved:

```diff
--- a/package/setup.py
+++ b/package/setup.py
@@ -286,9 +286,9 @@ def extensions(config):
     cpp_extra_link_args=[]
     # needed to specify c++ runtime library on OSX
     if platform.system() == 'Darwin' and using_clang():
-        cpp_extra_compile_args.append('-stdlib=libc++')
+        #cpp_extra_compile_args.append('-stdlib=libc++')
         cpp_extra_compile_args.append('-mmacosx-version-min=10.9')
-        cpp_extra_link_args.append('-stdlib=libc++')
+        #cpp_extra_link_args.append('-stdlib=libc++')
         cpp_extra_link_args.append('-mmacosx-version-min=10.7')
```

tylerjereddy:
* gracefully handle the case where `gcc` toolchain
in use on MacOS has been built from source using `clang`
by `spack` (so it really is `gcc` in use, not `clang`)

* we could try to add regression testing, but a few problems:
  - `using_clang()` is inside `setup.py`, which probably
can't be safely imported because it has unguarded statements/
code blocks that run right away
  - testing build issues is typically tricky with mocking, etc.
(though in this case, probably just need to move `using_clang()`
somewhere else and then test it against a variety of compiler
metadata strings)

Fixes #3109

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/setup.py
+++ b/package/setup.py
@@ -253,7 +253,19 @@ def using_clang():
     compiler = new_compiler()
     customize_compiler(compiler)
     compiler_ver = getoutput("{0} -v".format(compiler.compiler[0]))
-    return 'clang' in compiler_ver
+    if 'Spack GCC' in compiler_ver:
+        # when gcc toolchain is built from source with spack
+        # using clang, the 'clang' string may be present in
+        # the compiler metadata, but it is not clang
+        is_clang = False
+    elif 'clang' in compiler_ver:
+        # by default, Apple will typically alias gcc to
+        # clang, with some mention of 'clang' in the
+        # metadata
+        is_clang = True
+    else:
+        is_clang = False
+    return is_clang
 
 
 def extensions(config):


package/setup.py
```diff
@@ -253,7 +253,19 @@ def using_clang():
     compiler = new_compiler()
     customize_compiler(compiler)
     compiler_ver = getoutput("{0} -v".format(compiler.compiler[0]))
-    return 'clang' in compiler_ver
+    if 'Spack GCC' in compiler_ver:
```
IAlibay, line 256: So I have no issues with the setup file as-is, but I do wonder if this is opening a pandora's box in supporting the peculiarities for various build tools (for example, does EasyBuild also need handling differently?).

Is this the same way numpy/scipy is handling compiler identification @tylerjereddy ?

tylerjereddy: No, SciPy uses `scipy/_build_utils/compiler_helper.py`, which is a module dedicated to compiler handling. As discussed in the matching issue, they implement a small `try_compile()` function to only allow flags to get appended to the compile line if the compiler genuinely supports them.

We could redesign to use something like that--that may be biting off a bit more than I was hoping to chew here, though we could probably even copy-paste a decent bit of the code.

Here's an example of how it works:

```python
    if sys.platform == 'darwin':
        # Set min macOS version
        min_macos_flag = '-mmacosx-version-min=10.9'
        if has_flag(cc, min_macos_flag):
            args.append(min_macos_flag)
            ext.extra_link_args.append(min_macos_flag)
```

That `has_flag()` function checks that the compile flag works "for real" with the compiler:

```python
def has_flag(compiler, flag, ext=None):
    """Returns True if the compiler supports the given flag"""
    return try_compile(compiler, flags=[flag], ext=ext)
```

orbeckst: Is this starting to re-implement what automake and cmake build systems do?

orbeckst (approved):
I don't see any specific harm coming from being more correct in how clang is identified.

@IAlibay if you (or anyone else) has a particular view on how to proceed then please block the PR with a "Request changes".

tylerjereddy:
> Is this starting to re-implement what automake and cmake build systems do?

Yeah, basically--this discussion is also relevant: https://github.com/scipy/scipy/issues/13615

I wonder if the Python ecosystem will stabilize on a common solution eventually.

IAlibay (approved):
I probably should have just approved in my first review 😅 

Thanks for the insights @tylerjereddy, maybe we should open a separate issue to target a refresh of the build system based on what scipy does? (if it somehow fixes #3176 at the same time, then even better!).

tylerjereddy:
Sounds good, your skepticism is justified--this isn't exactly elegant either. And we have in-house expertise with CMake, so some of our dystopia project assembly-level folks may share my irritation with having to learn meson on top of CMake, but anyway..

PR #3236, opened 2021-04-20
Fix permission error test to not be language centric

### LINKED ISSUE #3221: Tests for "Permission denied" fail on non-English computers
Hello, I just installed MDAnalysis, the installation was fine, without any error. The problem is when I run the pytest commands, these are: 

## When I run this, it causes 2 errors. ##
pytest --disable-pytest-warnings --pyargs MDAnalysisTests
Screenshot_20210406_144410

In the end this comes out.
Screenshot_20210406_165548

 ## When I install the pytest puglin, there is no problem either.  ##
pip install pytest-xdist
Screenshot_20210406_171345

 ## But when I run the test, I get 3 errors.  ##
pytest --disable-pytest-warnings --pyargs MDAnalysisTests --numprocesses 4
Screenshot_20210406_172652

In the end this comes out.
Screenshot_20210406_174544

## About installation; ##
I am using a virtual machine with linux OS (Debian 10). I have checked the installation of python, its pytest library and pip. I have also performed all the steps in order. The errors as you can see in the image mainly say "Permission denied". This is my first time participating in a project on this topic and I really want to implement it and learn more about molecular dynamics. I would appreciate if someone can guide me to solve this.

 ## Versions:  ##
MDAnalysis version: I had problems running the command to find out the version.
Python Version: Python 3.7.3
OS Version: Debian 10.9
Kernel Version: Linux 4.19.0-16-amd64

Screenshot_20210406_185544

IAlibay:
Fixes #3221 

 - Changes check to be on PermissionError (which reflects what we are testing) and the error number [13].

PR Checklist
------------
 - [x] Tests?
 - CHANGELOG updated? N/A

### DIFF
--- a/testsuite/MDAnalysisTests/core/test_universe.py
+++ b/testsuite/MDAnalysisTests/core/test_universe.py
@@ -24,6 +24,7 @@ import pickle
 
 import os
 import subprocess
+import errno
 
 from io import StringIO
 
@@ -163,7 +164,7 @@ class TestUniverseCreation(object):
                 os.chmod(temp_file, 0o200)
 
             # Issue #3221 match by PermissionError and error number instead
-            with pytest.raises(PermissionError, match="13"):
+            with pytest.raises(PermissionError, match=f"Errno {errno.EACCES}"):
                 mda.Universe('permission.denied.tpr')
 
     def test_load_new_VE(self):
--- a/testsuite/MDAnalysisTests/core/test_universe.py
+++ b/testsuite/MDAnalysisTests/core/test_universe.py
@@ -161,12 +161,10 @@ class TestUniverseCreation(object):
                                 shell=True)
             else:
                 os.chmod(temp_file, 0o200)
-            try:
+
+            # Issue #3221 match by PermissionError and error number instead
+            with pytest.raises(PermissionError, match="13"):
                 mda.Universe('permission.denied.tpr')
-            except IOError as e:
-                assert 'Permission denied' in str(e.strerror)
-            else:
-                raise AssertionError
 
     def test_load_new_VE(self):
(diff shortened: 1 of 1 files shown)


IAlibay:
This _should_ be good

testsuite/MDAnalysisTests/core/test_universe.py
```diff
@@ -161,12 +161,10 @@ def test_Universe_invalidpermissionfile_IE_msg(self, tmpdir):
                                 shell=True)
             else:
                 os.chmod(temp_file, 0o200)
-            try:
+
+            # Issue #3221 match by PermissionError and error number instead
+            with pytest.raises(PermissionError, match="13"):
```
orbeckst, line 166: Is there a way that we can test for the error code itself as a number instead of checking that the string is included?

orbeckst: You could use
```python
import errno

with pytest.raises(PermissionError, match=str(errno.EACCES)):
```

because I get
```python
>>> errno.errorcode[13]
'EACCES'
```

orbeckst (approved):
Looks good as it is & improves over the previous.

If possible I'd like `errno.EACCES` instead of "13", assuming this works on all tested systems.

testsuite/MDAnalysisTests/core/test_universe.py
IAlibay: Sounds good, I've added `Errno` to the match too so we know we're not just capturing a random number. (I'm assuming the `Errno 13` match is language agnostic, which seems to be the case in #3221, but I don't have a non-English system to test against at the moment).

PR #3240, opened 2021-04-23
Remove ParmEd Timestep writing "support"

### LINKED ISSUE #3031: Remove support for ts as an input to the ParmedConverter

lilyminium:
Fixes #3031 

 - Develop version of #3172

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/coordinates/ParmEd.py
+++ b/package/MDAnalysis/coordinates/ParmEd.py
@@ -172,11 +172,7 @@ class ParmEdConverter(base.ConverterBase):
             # make sure to use atoms (Issue 46)
             ag_or_ts = obj.atoms
         except AttributeError:
-            if isinstance(obj, base.Timestep):
-                raise ValueError("Writing Timesteps to ParmEd "
-                                 "objects is not supported")
-            else:
-                raise TypeError("No atoms found in obj argument") from None
+            raise TypeError("No atoms found in obj argument") from None
 
         # Check for topology information
         missing_topology = []
--- a/package/MDAnalysis/coordinates/ParmEd.py
+++ b/package/MDAnalysis/coordinates/ParmEd.py
@@ -173,9 +173,10 @@ class ParmEdConverter(base.ConverterBase):
             ag_or_ts = obj.atoms
         except AttributeError:
             if isinstance(obj, base.Timestep):
-                ag_or_ts = obj.copy()
+                raise ValueError("Writing Timesteps to ParmEd "
+                                 "objects is not supported")
             else:
-                raise_from(TypeError("No Timestep found in obj argument"), None)
+                raise TypeError("No atoms found in obj argument") from None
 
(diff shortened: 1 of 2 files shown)


package/MDAnalysis/coordinates/ParmEd.py
```diff
@@ -173,9 +173,10 @@ def convert(self, obj):
             ag_or_ts = obj.atoms
         except AttributeError:
             if isinstance(obj, base.Timestep):
-                ag_or_ts = obj.copy()
+                raise ValueError("Writing Timesteps to ParmEd "
```
IAlibay, line 176: Since passing timesteps never really worked, and was removed in 1.1, is the logic here really required (i.e. do we really need to specifically call out passing a timestep)?

testsuite/MDAnalysisTests/coordinates/test_parmed.py
```diff
@@ -286,3 +287,17 @@ class TestParmEdConverterPDB(BaseTestParmEdConverter):
...
+    with pytest.raises(ValueError, match=err):
+        c = ParmEdConverter()
+        c.convert(u.trajectory.ts)
+
+
+def test_incorrect_object_passed_typeerror():
+    err = "No atoms found in obj argument"
+    with pytest.raises(TypeError, match=err):
+        c = ParmEdConverter()
+        c.convert("we still don't support emojis :(")
```
IAlibay, line 303: Do we not? I thought py3+ was UTF-8 compliant?

IAlibay (approved):
thanks @lilyminium lgtm, although two questions.

I'll let you choose whether you think it's worth addressing either :)

IAlibay:
How does the changelog work here? This is a forward port of the 1.1 changes right? Are we going to add the 1.1 changelog to 2.0 or does it exist in two different branches that will never see each other?

@orbeckst @richardjgowers thoughts here?

orbeckst:
The 1.1 changes need to be integrated into the 2.0 CHANGELOG once 1.1 is released.

package/MDAnalysis/coordinates/ParmEd.py
lilyminium: Probably not, I just copied over the changes from 1.1

IAlibay:
Windows' lack of support for emojis is deeply upsetting :(

IAlibay:
@lilyminium I'll let you deal with changelog & merging.

PR #3241, opened 2021-04-23
Patch chi1_sel to match all gamma atoms

### LINKED ISSUE #2044: Residue.chi1_angle() selection is incomplete
**Expected behavior**

`Residue.chi1_angle()`  https://github.com/MDAnalysis/mdanalysis/blob/388346d38370d833b48152f61405a18942d25251/package/MDAnalysis/core/topologyattrs.py#L614 should find all chi1 angles for all residues for standard PDB files from the Protein Databank.

**Actual behavior**

@hfmull noticed in https://github.com/MDAnalysis/mdanalysis/pull/2033/files#r208769338 that the standard selection is incomplete. He implemented a custom selection in the new `analysis.dihedrals.Janin` class. 

Need example file.

See #3238 for example.

develop 0.18.1-dev

lilyminium:
Fixes #2044

 - Changes CG name to '*G *G1 and not name H*'
 - A bit hacky

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/core/topologyattrs.py
+++ b/package/MDAnalysis/core/topologyattrs.py
@@ -1098,7 +1098,7 @@ class Atomnames(_AtomStringAttr):
     transplants[ResidueGroup].append(('omega_selections', omega_selections))
 
     def chi1_selection(residue, n_name='N', ca_name='CA', cb_name='CB',
-                       cg_name='CG'):
+                       cg_name='*G *G1 and not name H*'):
         """Select AtomGroup corresponding to the chi1 sidechain dihedral N-CA-CB-CG.
 
         Parameters
@@ -1125,7 +1125,7 @@ class Atomnames(_AtomStringAttr):
         .. versionadded:: 0.7.5
         """
         names = [n_name, ca_name, cb_name, cg_name]
-        ags = [residue.atoms[residue.atoms.names == n] for n in names]
+        ags = [residue.atoms.select_atoms(f"name {n}") for n in names]
         if any(len(ag) != 1 for ag in ags):
             return None
         return sum(ags)
--- a/package/MDAnalysis/core/topologyattrs.py
+++ b/package/MDAnalysis/core/topologyattrs.py
@@ -1099,7 +1099,18 @@ class Atomnames(_AtomStringAttr):
 
     def chi1_selection(residue, n_name='N', ca_name='CA', cb_name='CB',
                        cg_name='CG CG1 OG OG1 SG'):
-        """Select AtomGroup corresponding to the chi1 sidechain dihedral N-CA-CB-CG.
+        r"""Select AtomGroup corresponding to the chi1 sidechain dihedral ``N-CA-CB-*G.``
+        The gamma atom is taken to be the heavy atom in the gamma position. If more than one
+        heavy atom is present (e.g. CG1 and CG2), the one with the lower number is used (CG1).
+
+        .. warning::
+
+            This numbering of chi1 atoms here in following with the IUPAC 1970 rules\.
+            However, it should be noted that analyses which use dihedral angles may have
+            different definitions. For example, the
+            :class:`MDAnalysis.analysis.dihedrals.Janin` class does not incorporate
+            amino acids where the gamma atom is not carbon, into its chi1 selections.
+
 
         Parameters
         ----------
--- a/package/MDAnalysis/core/topologyattrs.py
+++ b/package/MDAnalysis/core/topologyattrs.py
@@ -1098,7 +1098,7 @@ class Atomnames(_AtomStringAttr):
     transplants[ResidueGroup].append(('omega_selections', omega_selections))
 
     def chi1_selection(residue, n_name='N', ca_name='CA', cb_name='CB',
-                       cg_name='*G *G1 and not name H*'):
+                       cg_name='CG CG1 OG OG1 SG'):
         """Select AtomGroup corresponding to the chi1 sidechain dihedral N-CA-CB-CG.
 
         Parameters
--- a/package/MDAnalysis/core/topologyattrs.py
+++ b/package/MDAnalysis/core/topologyattrs.py
@@ -1105,7 +1105,7 @@ class Atomnames(_AtomStringAttr):
 
         .. warning::
 
-            This numbering of chi1 atoms here in following with the IUPAC 1970 rules\.
+            This numbering of chi1 atoms here in following with the IUPAC 1970 rules.
             However, it should be noted that analyses which use dihedral angles may have
             different definitions. For example, the
             :class:`MDAnalysis.analysis.dihedrals.Janin` class does not incorporate
--- a/testsuite/MDAnalysisTests/core/test_atomgroup.py
+++ b/testsuite/MDAnalysisTests/core/test_atomgroup.py
@@ -868,9 +868,7 @@ class TestDihedralSelections(object):
         rssel = [r.chi1_selection() for r in resgroup]
         assert_equal(rgsel, rssel)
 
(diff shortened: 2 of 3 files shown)


package/MDAnalysis/core/topologyattrs.py
```diff
@@ -1098,7 +1098,7 @@ def omega_selections(residues, c_name='C', n_name='N', ca_name='CA'):
     transplants[ResidueGroup].append(('omega_selections', omega_selections))
 
     def chi1_selection(residue, n_name='N', ca_name='CA', cb_name='CB',
-                       cg_name='CG'):
+                       cg_name='*G *G1 and not name H*'):
```
orbeckst, line 1101: Feels a bit hacky – can't we be more explicit and say CG CG1 CG2 ... or whatever the official definition might be? Ultimately this will only work for standard residues anyway (at least in default) so as long as it works for typical PDB structures and MD structures closely following PDB standards, I think it will be ok.

testsuite/MDAnalysisTests/core/test_atomgroup.py
```diff
@@ -706,6 +707,11 @@ def GRO():
     def PSFDCD():
         return mda.Universe(PSF, DCD)
 
+    @staticmethod
+    @pytest.fixture()
+    def TPR():
+        return mda.Universe(TPR)
+
```
orbeckst, line 714: Just to include a CHARMM protein as well:         
```suggestion

    @staticmethod
    @pytest.fixture()
    def PSF():
        return mda.Universe(PSF, DCD)   # load DCD to reduce warnings...
```

orbeckst (changes requested):
Thanks @lilyminium ! 

I think I'd prefer a more explicit selection if possible, but can be swayed. What was your rationale for using wildcards for the gamma position? 

For tests: _optional_: also test on a PSF with CHARMM ff names (which are mostly the same as for the OPLS-AA TPR except for His) — and I think the TPR didn't have one amino acid but I forgot which one was missing... although, both are AdK so if AdK doesn't have a Trp (not sure....) then it's not in PSF either.

package/MDAnalysis/core/topologyattrs.py
lilyminium: I did go around hunting for a chi1 definition, since I always understood it to be to whichever heavy atom was hanging off C-beta (so therefore ill-defined for branched aa's). Both this [EBI practical](https://www.ebi.ac.uk/pdbe/modval5) and [garlic](http://www.mlb.co.jp/linux/science/garlic/doc/commands/dihedrals.html) specify CG, CG1, OG1 or SG for the canonical amino acids -- no CG2. We can be explicit, yes. I leaned towards flexibility because of that recent question with the custom DNA residue and people using more and more ncAAs. However the arguments in the method are so easily modified that maybe the defaults should be more strictly sensible.

package/MDAnalysis/core/topologyattrs.py
```diff
@@ -1098,7 +1098,7 @@ def omega_selections(residues, c_name='C', n_name='N', ca_name='CA'):
     transplants[ResidueGroup].append(('omega_selections', omega_selections))
 
     def chi1_selection(residue, n_name='N', ca_name='CA', cb_name='CB',
-                       cg_name='CG'):
+                       cg_name='CG CG1 OG OG1 SG'):
         """Select AtomGroup corresponding to the chi1 sidechain dihedral N-CA-CB-CG.
```
orbeckst, line 1102: Given your research into chi1 definitions, could you _please_ record your reasoning here and perhaps write the chi1 as N-CA-CB-*G where *G = is the heavy atom in the gamma position?

The Janin 1978 paper https://doi.org/10.1016/0022-2836(78)90408-4 also talks about chi1 but I didn't have time to find the exact definition. Something about IUPAC 1970...(?)

orbeckst (approved):
Thanks for addressing my comments.

I have a suggestion for improving the docs; whatever you consider appropriate works for me.

package/MDAnalysis/core/topologyattrs.py
orbeckst: Which reminds me: our Janin analysis and the chi1 selection here need to be consistent.

lilyminium: We will need to decide if we want to keep to the stricter definition where the gamma can only be "C", then (I am leaning... yes). I don't think the IUPAC 1970 document actually specifies that the gamma atom must be carbon, but a number of secondary sources citing IUPAC 1970 have rolled with that definition (and for Janin particularly, because of energy differences). The Janin class ignores residues without a carbon in that gamma position.

Janin 1978:

> Our calculations apply to residues with carbon atoms in the y-position. Sulphur (Cys) and oxygen atoms (Ser, Thr) have lesser effects (Tonnuswamy & Sasisekharan, 1971a). For side-chains with branched Cb atoms (Ile, Val, Thr) the energy barriers created by the two y-substituents add up

IUPAC-IUB Commission 1969 Draft
> 4.5.2 Torsion angles are denoted by x and are specified by two (or three) superscripts, the first one (or two), in the situation
> ```
>						D
>						|
> 				A-B-C--E
>						|
>						F
> ```
> indicating the bond B-C about which the angle is measured, and the last indicating whether the angle is measured relative to D, E, or F. The principal torsion angle is defined by Rule 2.2.1, and if there is no ambiguity the last superscript may be omitted in referring to it.
> Thus in valine $\chi_i^{2, 1}$ refer to the torsion angles specifying $C_i^{y1}$ and $C_i^{y2}$ [...]

^^ so IUPAC reckons we need to be specific with which chi1 angle to use, and Janin ignores branched cases and non-Carbon cases.

orbeckst: Thanks for doing the research. I didn't realize that Janin does not do "all" chi1.

Given that branched amino acids have more than a chi1, can we consistently use one and state that? (Then we could later update our Janin analysis to also collect those... or at least be clear that we follow the original paper and exclude branched aa.)

I feel that when using chi1 you really want to know what the sidechain does and not getting any answer for Thr or Cys feels odd.

lilyminium: I added a warning that the Janin class differs from the `chi1_selection()` function. Turns out the Janin class already has a note saying that by default these certain residues are ignored, although it attributes them all to not having a chi2 angle (which true, they do not), not because the paper defines it as carbon-only.

package/MDAnalysis/core/topologyattrs.py
```diff
@@ -1098,8 +1098,19 @@ def omega_selections(residues, c_name='C', n_name='N', ca_name='CA'):
...
-                       cg_name='CG'):
-        """Select AtomGroup corresponding to the chi1 sidechain dihedral N-CA-CB-CG.
+                       cg_name='CG CG1 OG OG1 SG'):
+        r"""Select AtomGroup corresponding to the chi1 sidechain dihedral ``N-CA-CB-*G.``
+        The gamma atom is taken to be the heavy atom in the gamma position. If more than one
+        heavy atom is present (e.g. CG1 and CG2), the one with the lower number is used (CG1).
+
+        .. warning::
+
+            This numbering of chi1 atoms here in following with the IUPAC 1970 rules\.
```
orbeckst, line 1108: Is there a reason for the backslash escaped period, @lilyminium ?

I would have changed it but maybe there's something I am missing.

lilyminium: Accident, sorry.

orbeckst:
Thanks @lilyminium !

PR #3242, opened 2021-04-25
Hbond excl fix

### LINKED ISSUE #2987: HydrogenBondAutoCorrel uses deprecated features of numpy
## Expected behavior ##

Running the tests throws no warning, and the library does not use deprecated features of the libraries it depends on.

## Actual behavior ##

Running the tests warns about HydrogenBondAutoCorrel using deprecated features of numpy:

```
MDAnalysisTests/analysis/test_hydrogenbondautocorrel.py::TestHydrogenBondAutocorrel::test_continuous_excl
MDAnalysisTests/analysis/test_hydrogenbondautocorrel.py::TestHydrogenBondAutocorrel::test_continuous_excl
MDAnalysisTests/analysis/test_hydrogenbondautocorrel.py::TestHydrogenBondAutocorrel::test_intermittent_excl
MDAnalysisTests/analysis/test_hydrogenbondautocorrel.py::TestHydrogenBondAutocorrel::test_intermittent_excl
  /home/jon/dev/mdanalysis/package/MDAnalysis/analysis/hbonds/hbond_autocorrel.py:430: DeprecationWarning: elementwise comparison failed; this will raise an error in the future.
    pair = np.delete(pair, np.where(pair==exclude), 0)

MDAnalysisTests/analysis/test_hydrogenbondautocorrel.py::TestHydrogenBondAutocorrel::test_continuous_excl
MDAnalysisTests/analysis/test_hydrogenbondautocorrel.py::TestHydrogenBondAutocorrel::test_intermittent_excl
  <__array_function__ internals>:5: DeprecationWarning: Calling nonzero on 0d arrays is deprecated, as it behaves surprisingly. Use `atleast_1d(cond).nonzero()` if the old behavior was intended. If the context of this warning is of the form `arr[nonzero(cond)]`, just use `arr[cond]`.
```

`pytest analysis/test_hydrogenbondautocorrel.py`

richardjgowers:
Fixes #2987

 - 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -28,6 +28,7 @@ Fixes
     (Issues #2229 #2345 #2919, PR #2937).
   * Removed deprecation warning from numpy in hbond_autocorrel
     (Issue #2987 PR #3242)
+  * Fixed bug in exclusion matrix of hbond_autocorrel (PR #3242)
 
 Changes
   * Maximum pinned versions in setup.py removed for python 3.6+ (PR #3139)
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -26,6 +26,8 @@ Fixes
     array warnings (PR #3139, backports PRs #2845 and #2834)
   * Fixed several issues with NSGrid and triclinic boxes not finding some pairs.
     (Issues #2229 #2345 #2919, PR #2937).
+  * Removed deprecation warning from numpy in hbond_autocorrel
+    (Issue #2987 PR #3242)
 
 Changes
   * Maximum pinned versions in setup.py removed for python 3.6+ (PR #3139)
--- a/package/MDAnalysis/analysis/hbonds/hbond_autocorrel.py
+++ b/package/MDAnalysis/analysis/hbonds/hbond_autocorrel.py
@@ -432,7 +432,7 @@ class HydrogenBondAutoCorrel(object):
         # 2d array of all distances
         pair = capped_distance(self.h.positions, self.a.positions, max_cutoff=self.d_crit, box=box,
                                return_distances=False)
-        if not self.exclusions is None:
+        if self.exclusions is not None:
             pair = pair[~ _in2d(pair, self.exclusions)]
 
         hidx, aidx = np.transpose(pair)
--- a/package/MDAnalysis/analysis/hbonds/hbond_autocorrel.py
+++ b/package/MDAnalysis/analysis/hbonds/hbond_autocorrel.py
@@ -322,7 +322,7 @@ class HydrogenBondAutoCorrel(object):
                              "length.  Try using `find_hydrogen_donors`.")
 
         if exclusions is not None:
-            if not len(exclusions[0]) == len(exclusions[1]):
+            if len(exclusions[0]) != len(exclusions[1]):
                 raise ValueError(
                     "'exclusion' must be two arrays of identical length")
             self.exclusions = np.column_stack((exclusions[0], exclusions[1])).astype(np.intp)
--- a/package/MDAnalysis/analysis/hbonds/hbond_autocorrel.py
+++ b/package/MDAnalysis/analysis/hbonds/hbond_autocorrel.py
@@ -115,7 +115,7 @@ as so
 
 
 The keyword **exclusions** allows a tuple of array addresses to be provided,
-(Hidx, Aidx),these pairs of hydrogen-acceptor are then not permitted to be
+(Hidx, Aidx), these pairs of hydrogen-acceptor are then not permitted to be
 counted as part of the analysis.  This could be used to exclude the
 consideration of hydrogen bonds within the same functional group, or to perform
 analysis on strictly intermolecular hydrogen bonding.
@@ -215,6 +215,7 @@ import warnings
 
 from MDAnalysis.lib.log import ProgressBar
 from MDAnalysis.lib.distances import capped_distance, calc_angles, calc_bonds
+from MDAnalysis.lib._cutil import _in2d
 from MDAnalysis.core.groups import requires
 
 from MDAnalysis.due import due, Doi
@@ -320,11 +321,13 @@ class HydrogenBondAutoCorrel(object):
             raise ValueError("Donors and Hydrogen groups must be identical "
                              "length.  Try using `find_hydrogen_donors`.")
 
-        self.exclusions = exclusions
-        if self.exclusions:
-            if not len(self.exclusions[0]) == len(self.exclusions[1]):
+        if not exclusions is None:
+            if not len(exclusions[0]) == len(exclusions[1]):
                 raise ValueError(
                     "'exclusion' must be two arrays of identical length")
+            self.exclusions = np.column_stack((exclusions[0], exclusions[1])).astype(np.intp)
+        else:
+            self.exclusions = None
 
         self.bond_type = bond_type
         if self.bond_type not in ['continuous', 'intermittent']:
@@ -427,11 +430,10 @@ class HydrogenBondAutoCorrel(object):
         box = self.u.dimensions if self.pbc else None
 
         # 2d array of all distances
-        pair, d = capped_distance(self.h.positions, self.a.positions, max_cutoff=self.d_crit, box=box)
-        if self.exclusions:
-            # set to above dist crit to exclude
-            exclude = np.column_stack((self.exclusions[0], self.exclusions[1]))
-            pair = np.delete(pair, np.where(pair==exclude), 0)
+        pair = capped_distance(self.h.positions, self.a.positions, max_cutoff=self.d_crit, box=box,
+                               return_distances=False)
+        if not self.exclusions is None:
+            pair = pair[~ _in2d(pair, self.exclusions)]
 
         hidx, aidx = np.transpose(pair)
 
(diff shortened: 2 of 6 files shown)


package/MDAnalysis/analysis/hbonds/hbond_autocorrel.py
```diff
@@ -427,11 +430,10 @@ def _single_run(self, start, stop):
...
         # 2d array of all distances
-        pair, d = capped_distance(self.h.positions, self.a.positions, max_cutoff=self.d_crit, box=box)
-        if self.exclusions:
-            # set to above dist crit to exclude
-            exclude = np.column_stack((self.exclusions[0], self.exclusions[1]))
-            pair = np.delete(pair, np.where(pair==exclude), 0)
+        pair = capped_distance(self.h.positions, self.a.positions, max_cutoff=self.d_crit, box=box,
+                               return_distances=False)
+        if not self.exclusions is None:
+            pair = pair[~ _in2d(pair, self.exclusions)]
```
IAlibay, line 436: I'm surprised there was no suitable numpy operation for this :/

package/MDAnalysis/lib/_cutil.pyx
```diff
@@ -90,6 +91,47 @@ def unique_int_1d(np.intp_t[:] values):
...
+
+    .. versionadded:: 1.1.0
+    """
+    cdef object out
+    cdef ssize_t i
+    cdef cset[pair[np.intp_t, np.intp_t]] hits
+    cdef pair[np.intp_t, np.intp_t] p
+
+    if not arr1.shape[1] == 2 or not arr2.shape[1] == 2:
+        raise ValueError("Both arrays must be (n, 2) arrays")
```
IAlibay, line 116: Test for this please.

package/MDAnalysis/lib/_cutil.pyx
```diff
@@ -90,6 +91,47 @@ def unique_int_1d(np.intp_t[:] values):
     return np.array(result)
 
 
+@cython.boundscheck(False)
+def _in2d(np.intp_t[:, :] arr1, np.intp_t[:, :] arr2):
```
IAlibay, line 95: We probably should have a direct test for this rather than just indirectly testing via hbond_autocorrel (it's a useful function that probably would have other uses).

IAlibay (changes requested):
Couple of tests & logic comments.

Also my cython is probably too green to probably review stuff, if someone with more experience can have a look that'd be great - pinging @MDAnalysis/coredevs

IAlibay:
Forgot to mention, as far as I understand the old tests weren't picking up that this code path wasn't working. If that's the case, are we sure they are now working properly?

testsuite/MDAnalysisTests/lib/test_cutil.py
```diff
@@ -64,3 +66,21 @@ def test_find_fragments(edges, ref):
...
+
+    assert_equal(result, np.array([False, True, False]))
+
+
+@pytest.mark.parametrize('arr1,arr2', [
+    (np.array([1, 2, 3], dtype=np.intp), np.array([[1, 2], [3, 4]], dtype=np.intp)),
+    (np.array([[1, 2], [3, 4]], dtype=np.intp), np.array([1, 2, 3], dtype=np.intp)),
+])
+def test_in2d_VE(arr1, arr2):
+    with pytest.raises(ValueError):
```
IAlibay, line 85:
```suggestion
    with pytest.raises(ValueError, match="Both arrays must be (n, 2) arrays"):
```

package/MDAnalysis/lib/_cutil.pyx
IAlibay: Weird codecov is still throwing a warning here :/

IAlibay (changes requested):
Thanks for the tests. For some reason codecov isn't recognising the `ValueError` test, just proposing a match to make sure it's actually picking up the right thing.

package/MDAnalysis/analysis/hbonds/hbond_autocorrel.py
tylerjereddy: The below passes the analysis suite locally without using a custom Cython function. Maybe double check with Richard if his mixture of Cython/C++ is more efficient -- probably any advantage is related to the integer only nature of the data structure?

```diff
--- a/package/MDAnalysis/analysis/hbonds/hbond_autocorrel.py
+++ b/package/MDAnalysis/analysis/hbonds/hbond_autocorrel.py
@@ -210,6 +210,7 @@ from six import raise_from
 
 import numpy as np
 import scipy.optimize
+from scipy.spatial.distance import cdist
 
 import warnings
 
@@ -433,7 +434,7 @@ class HydrogenBondAutoCorrel(object):
         pair = capped_distance(self.h.positions, self.a.positions, max_cutoff=self.d_crit, box=box,
                                return_distances=False)
         if not self.exclusions is None:
-            pair = pair[~ _in2d(pair, self.exclusions)]
+            pair = pair[~ (cdist(pair, self.exclusions)==0).any(axis=1)]
 
         hidx, aidx = np.transpose(pair)
```

package/MDAnalysis/lib/_cutil.pyx
```diff
@@ -90,6 +91,47 @@ def unique_int_1d(np.intp_t[:] values):
...
+    cdef ssize_t i
+    cdef cset[pair[np.intp_t, np.intp_t]] hits
+    cdef pair[np.intp_t, np.intp_t] p
+
+    if not arr1.shape[1] == 2 or not arr2.shape[1] == 2:
+        raise ValueError("Both arrays must be (n, 2) arrays")
+
+    for i in range(arr2.shape[0]):
+        p = pair[np.intp_t, np.intp_t](arr2[i, 0], arr2[i, 1])
+        hits.insert(p)
```
tylerjereddy, line 120: Not sure if we need a comment or two about the C++ magic happening here; I suppose we are slowly growing expertise with devs familiar with C++ (I'm not one of them).

package/MDAnalysis/analysis/hbonds/hbond_autocorrel.py
richardjgowers: You could use cdist but it's odd to use an O(n2) solution when you can just use a set

tylerjereddy: I don't have a strong view on it--if this genuinely a performance-critical part of the code maybe the custom maintenance burden is worth it.

richardjgowers:
WRT codecov, I'm guessing Cython linenumbers and linetracing isn't a match made in heaven.

package/CHANGELOG
```diff
@@ -26,6 +26,8 @@ Fixes
     array warnings (PR #3139, backports PRs #2845 and #2834)
   * Fixed several issues with NSGrid and triclinic boxes not finding some pairs.
     (Issues #2229 #2345 #2919, PR #2937).
+  * Removed deprecation warning from numpy in hbond_autocorrel
```
IAlibay, line 29: Given the changes introduced here, did the previous exclusions code give the right behaviour? If not, we should warn users of behaviour changes here.

package/MDAnalysis/analysis/hbonds/hbond_autocorrel.py
```diff
@@ -427,11 +430,10 @@ def _single_run(self, start, stop):
...
 
         # 2d array of all distances
-        pair, d = capped_distance(self.h.positions, self.a.positions, max_cutoff=self.d_crit, box=box)
-        if self.exclusions:
-            # set to above dist crit to exclude
-            exclude = np.column_stack((self.exclusions[0], self.exclusions[1]))
-            pair = np.delete(pair, np.where(pair==exclude), 0)
+        pair = capped_distance(self.h.positions, self.a.positions, max_cutoff=self.d_crit, box=box,
+                               return_distances=False)
+        if not self.exclusions is None:
```
orbeckst, line 435: Equivalent and more common in Python:
```suggestion
        if self.exclusions is not None:
```

package/CHANGELOG
lilyminium: I think (@richardjgowers please chime in) that -- the old and new code has different behaviour, *but*, it "shouldn't" result in users noticing different results. In the new `_in2d` function, the ordering of the H-donor and H-acceptor indices matters when matching exclusions. In the old one it doesn't. So, in this new code, it matters if users mess up which atom is the donor and which is the acceptor when inputting `exclusions`.

lilyminium: @IAlibay did you mean warn users with a real warning or in the CHANGELOG?

richardjgowers: So originally this used to work on a dense matrix, so the input of two tuples of "coordinates" into the dense matrix could be applied directly.  

When we switched to a sparse matrix (i.e. capped_dist) I think we started doing it wrong because `pair == exclude` isn't masking out pairs as expected.  So you might get excluded pairs in the output.

This new code uses a "set of tuples" to exclude pairs so will work.

IAlibay: > @IAlibay did you mean warn users with a real warning or in the CHANGELOG?

Sorry I meant just a changelog entry. I know most don't read it, but it is a useful record for folks to keep track of stuff that might have affected their work.

IAlibay (approved):
lgtm!

PR #3244, opened 2021-04-26
Switch GNMAnalysis to AnalysisBase

### LINKED ISSUE #3243: Switch GNMAnalysis to AnalysisBase.

The `GNMAnalysis` currently uses its own logic for the trajectory analysis. The AnalysisBase class provides the 
same  functionality and should be used as a parent class.

 1. Change the parent class of  `GNMAnalysis` to `AnalysisBase`
https://github.com/MDAnalysis/mdanalysis/blob/fa074966bfe21c85bde8e4b61cd29ef8aecd692a/package/MDAnalysis/analysis/gnm.py#L192

2.  Move the code in the `run` method that is outside the loop to `_prepare` and everything that is inside to `_single_frame`
https://github.com/MDAnalysis/mdanalysis/blob/fa074966bfe21c85bde8e4b61cd29ef8aecd692a/package/MDAnalysis/analysis/gnm.py#L309

PicoCentauri:
Fixes #3243 

Switch GNMAnalysis to AnalysisBase.

I don't think a `versionchanged` info is necessary here. For the user there are no breaking changes.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


orbeckst (approved):
This looks like an uncontroversial, easy LGTM to me.

I looked at the 80% coverage for the diff and I don't think it's an issue. The only thing that's not tested is the case when the SVD fails and we skip a frame with an ugly print message.

Thanks @PicoCentauri .

@IAlibay would it be ok to squeeze it into 2.0? I'll add it to the milestone but leave it to you to remove it or to merge.

package/MDAnalysis/analysis/gnm.py
```diff
@@ -306,49 +313,29 @@ def generate_kirchoff(self):
...
-                matrix,
-                ReportVector=self.ReportVector,
-                counter=ts.frame)
+
+    def _single_frame(self):
+        matrix = self.generate_kirchoff()
+        try:
+            u, w, v = np.linalg.svd(matrix)
+        except np.linalg.LinAlgError:
+            print("\nFrame skip at", self._ts.time,
```
IAlibay, line 324: How about a warning not a print statement here?
Also needs coverage.

package/MDAnalysis/analysis/gnm.py
```diff
@@ -306,49 +313,29 @@ def generate_kirchoff(self):
...
+                  "(SVD failed to converge). Cutoff", self.cutoff)
+            return
```
IAlibay, line 326: Needs coverage.

IAlibay (changes requested):
Couple of comments, I realise the `print` statement is a leftover of the previous code, but it would be much better if we could easily test it.

@orbeckst if one of the coredevs is happy to ensure that this gets merged on time for the 2.0 deadline then I have no objections. My comment on discord were more intented as a "please prioritize reviewing efforts on what is on the current to-do list" rather than a blanket code freeze.

package/MDAnalysis/analysis/gnm.py
PicoCentauri: I changed the statement to print. SVD is quite robust and I haven't found a simple test case quickly. Maybe you guys have an idea?

IAlibay: Is it possible to just patch the return of np.linalg.svd to trigger the exception on call?

If not, I'd be happy with a pragma no cover.

PicoCentauri: Maybe there is a way. Let me check tomorrow.

IAlibay: Probably something like this (this is probably not the best/prettiest way to do it)?

```python
import pytest
import numpy as np
from unittest.mock import patch


def run_svd():
    a = np.random.randn(9, 6) + 1j*np.random.randn(9,6)
    return np.linalg.svd(a, full_matrices=True)


def test_linalgexcept():
    with patch('numpy.linalg.svd', side_effect=np.linalg.LinAlgError('SVD error')):
        with pytest.raises(np.linalg.LinAlgError):
            run_svd()
```

PicoCentauri: That looks nice. I took the case from `test_reload_offsets_if_offsets_readin_fails`

testsuite/MDAnalysisTests/analysis/test_gnm.py
```diff
@@ -80,6 +81,18 @@ def test_generate_kirchoff(universe):
...
 
+def test_gnm_SVD_fail(universe):
+    gnm = mda.analysis.gnm.GNMAnalysis(universe)
+    # Initilize attributes like self._ts 
+    # necessary for print the warning
+    gnm.run(stop=1)
+
+    with patch.object(np.linalg, "svd") as np_load_mock:
+        np_load_mock.side_effect = np.linalg.LinAlgError
+        with pytest.warns(UserWarning):
```
IAlibay, line 92: I'd add a match on the warning here (MDA throws a ton of warnings all the time).

package/MDAnalysis/analysis/gnm.py
```diff
@@ -217,17 +219,22 @@ class GNMAnalysis(object):
...
+
     See Also
     --------
     :class:`closeContactGNMAnalysis`
 
-
     .. versionchanged:: 0.16.0
        Made :meth:`generate_output` a private method :meth:`_generate_output`.
 
     .. versionchanged:: 1.0.0
```
IAlibay, line 236: Could you also add suitable versionchanged entries?
The impact on user is low but it might interest developers.

testsuite/MDAnalysisTests/analysis/test_gnm.py
```diff
@@ -80,6 +81,18 @@ def test_generate_kirchoff(universe):
...
+            assert gnm._single_frame() is None
```
IAlibay, line 93: Either this is True right? We probably don't need an assert here, but if you want it, would it be better to just assert on a zero length results list?

IAlibay (changes requested):
Couple final comments (docs + matching on the warning text).

testsuite/MDAnalysisTests/analysis/test_gnm.py
```diff
@@ -80,6 +81,18 @@ def test_generate_kirchoff(universe):
        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])
 
 
+def test_gnm_SVD_fail(universe):
+    gnm = mda.analysis.gnm.GNMAnalysis(universe)
+    # Initilize attributes like self._ts
```
IAlibay, line 86: Forgot to add this in my review, I guess auto-pep8 is complaining about it
```suggestion
    # Initilize attributes like self._ts
```

testsuite/MDAnalysisTests/analysis/test_gnm.py
PicoCentauri: You are right. No we don't necessarily need the assert here.

testsuite/MDAnalysisTests/analysis/test_gnm.py
PicoCentauri: True :)

package/MDAnalysis/analysis/gnm.py
PicoCentauri: My local docs look a bit weird. Lets wait how the online version look like...

Screenshot 2021-04-28 at 08 57 18

package/MDAnalysis/analysis/gnm.py
```diff
@@ -217,17 +219,25 @@ class GNMAnalysis(object):
...
+    Attributes
+    ----------
+    results : list
+        GNM results per frame:
+            results = [(time,eigenvalues[1],eigenvectors[1]),
+                       (time,eigenvalues[1],eigenvectors[1]), ...]
+
     See Also
     --------
     :class:`closeContactGNMAnalysis`
```
IAlibay, line 232:
```suggestion


```

That should be cause of the doc issue, the two empty lines are needed here to tell sphinx you are exiting from the `see also` block.

package/MDAnalysis/analysis/gnm.py
```diff
@@ -402,18 +395,13 @@ def __init__(self,
                  cutoff=4.5,
```
IAlibay, line 395: Can you also add a versionchanged here too please?

IAlibay (changes requested):
Docs do indeed build weirdly, looks like just a missing empty line though.

package/MDAnalysis/analysis/gnm.py
PicoCentauri: Of course.

IAlibay (approved):
lgtm - @orbeckst are you happy with the recent changes?

orbeckst:
Looks good to me. Thank you both!

Please go ahead, Irfan

orbeckst:
@IAlibay I'm jumping in and merging it — I think we're all happy with the state of the PR.

PR #3247, opened 2021-04-28
Restructured WaterOrientationalRelaxation

### LINKED ISSUE #2719: Make waterdynamics stop behaviour the same as AnalysisBase

As detailed in https://github.com/MDAnalysis/mdanalysis/pull/2525#issuecomment-608264856 several waterdynamics methods currently have an inclusive `tf` (i.e. final frame) behaviour instead of the exclusive behaviour exhibited by AnalysisBase's `stop`. In order to offer users a consistent API, it would be advantageous to make everything exclusive.

Change the behaviour of the final frame definitions in waterdynamics to be exclusive.

The ideal alternative here is to switch waterdynamics over to AnalysisBase.

## Additional context ##
PR #2525

PicoCentauri:
Partly Fixes MDAnalysis/waterdynamics#22 

The `WaterOrientationalRelaxation` was restructured to follow the `start`, `stop` and `step` attributes of the `AnalysisBase`.
The documentation was also improved and the example was adjusted to a pure water system. I also moved everything 
into the class documentation.

It seems that the results after my restructuring are okay but maybe somebody who knows how these correlation function should look takes a look at the results.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/analysis/waterdynamics.py
+++ b/package/MDAnalysis/analysis/waterdynamics.py
@@ -332,10 +332,10 @@ class WaterOrientationalRelaxation(AnalysisBase):
     r"""Water orientation relaxation analysis
 
     Class evaluating the water orientational relaxation proposed by Yu-ling
-    Yeh and Chung-Yuan Mou [Yeh1999_]. 
+    Yeh and Chung-Yuan Mou [Yeh1999_].
     `WaterOrientationalRelaxation` (WOR) indicates
-    "how fast" water molecules are rotating or changing direction. 
-    If WOR is very stable we can assume that water molecules are 
+    "how fast" water molecules are rotating or changing direction.
+    If WOR is very stable we can assume that water molecules are
     rotating/changing direction very slow, on the
     other hand, if WOR decay very fast, we can assume that water molecules are
     rotating/changing direction very fast. The results are
@@ -344,7 +344,7 @@ class WaterOrientationalRelaxation(AnalysisBase):
     .. math::
         C_{\hat u}(\tau)=\langle \mathit{P}_2[\mathbf{\hat{u}}(t_0)\cdot\mathbf{\hat{u}}(t_0+\tau)]\rangle
 
-    where :math:`P_2=(3x^2-1)/2` is the second-order Legendre polynomial 
+    where :math:`P_2=(3x^2-1)/2` is the second-order Legendre polynomial
     and :math:`\hat{u}` is a unit vector along HH, OH or dipole vector.
 
     Parameters
@@ -373,11 +373,11 @@ class WaterOrientationalRelaxation(AnalysisBase):
 
         import matplotlib.pyplot as plt
         import MDAnalysis as mda
-        from MDAnalysis.analysis.waterdynamics import WaterOrientationalRelaxation as WOR
+        from MDAnalysis.analysis.waterdynamics import WaterOrientationalRelaxation
 
         u = mda.Universe(waterPSF, waterDCD)
         select = "byres name OH2"
-        wor = WOR(u, select)
+        wor = WaterOrientationalRelaxation(u, select)
         wor.run(0, 7, 1)
 
         # Plot the results
@@ -414,7 +414,7 @@ class WaterOrientationalRelaxation(AnalysisBase):
 
     .. versionchanged:: 2.0.0
        `t0`, `tf` and `dtmax` are renamed to start, stop, step as attributes
-       of the `run` method providing a consistent behaviour with other 
+       of the `run` method providing a consistent behaviour with other
        modules.
     """
 
@@ -425,11 +425,11 @@ class WaterOrientationalRelaxation(AnalysisBase):
 
     def _compare_molecules(self, selection, cur_frame, other_frame):
         """
-        Compare the molecules in the `cur_frame` selection to the 
+        Compare the molecules in the `cur_frame` selection to the
         `other_frame` selection and
         select only those particles that are repeated in both frame. This is to
-        consider only those molecules that remain in the selection after 
-        a certain time has elapsed. The result is a list with the indices of 
+        consider only those molecules that remain in the selection after
+        a certain time has elapsed. The result is a list with the indices of
         the atoms.
         """
         a = set(selection[cur_frame])
@@ -438,7 +438,7 @@ class WaterOrientationalRelaxation(AnalysisBase):
 
     def _get_repeated_indices(self, selection, frame):
         """
-        Indicates the comparation between the current frame and all 
+        Indicates the comparation between the current frame and all
         following frames.
         The results is a list of lists with all the repeated index per frame.
         Ex: frame=1, so compare frames (1,2),(2,3),(3,4)...
@@ -522,7 +522,6 @@ class WaterOrientationalRelaxation(AnalysisBase):
         else:
             return 0, 0, 0
 
-
     def _get_mean_one_point(self, universe, selections):
         """
         This function gets one point of the plot C_vec vs t. It uses the
@@ -536,7 +535,10 @@ class WaterOrientationalRelaxation(AnalysisBase):
         n = 0
 
         for i in range(self.start, self.stop // self.frame - 1):
-            a = self._get_one_delta_point(universe, repeated_indices, i - 1, cum_frames)
+            a = self._get_one_delta_point(universe,
+                                          repeated_indices,
+                                          i - 1,
+                                          cum_frames)
             sumDeltaOH += a[0]
             sumDeltaHH += a[1]
             sumDeltadip += a[2]
@@ -560,14 +562,14 @@ class WaterOrientationalRelaxation(AnalysisBase):
         self.dip = []
 
         # List storing the atoms inside selection for every frame
-        self.selection_out = []
+        self.sel_atoms = []
         for _ in self._trajectory:
-            self.selection_out.append(self.universe.select_atoms(self.selection))
+            self.sel_atoms.append(self.universe.select_atoms(self.selection))
 
     def _single_frame(self):
         self.frame = self._frame_index + self.step
-        results_ts = self._get_mean_one_point(self.universe, 
-                                              self.selection_out)
+        results_ts = self._get_mean_one_point(self.universe,
+                                              self.sel_atoms)
         self.OH.append(results_ts[0])
         self.HH.append(results_ts[1])
         self.dip.append(results_ts[2])
@@ -924,7 +926,7 @@ class SurvivalProbability(object):
        Using the MDAnalysis.lib.correlations.py to carry out the intermittency
        and autocorrelation calculations.
        Changed `selection` keyword to `select`.
-       Removed support for the deprecated `t0`, `tf`, and `dtmax` keywords. 
+       Removed support for the deprecated `t0`, `tf`, and `dtmax` keywords.
        These should instead be passed to :meth:`SurvivalProbability.run` as
        the `start`, `stop`, and `tau_max` keywords respectively.
        The `stop` keyword as passed to :meth:`SurvivalProbability.run` has now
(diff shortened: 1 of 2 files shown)


package/MDAnalysis/analysis/waterdynamics.py
```diff
@@ -388,79 +316,147 @@
...
         """
-        Indicates the comparation between all the t+dt.
-        The results is a list of list with all the repeated index per frame
-        (or time).
-        Ex: dt=1, so compare frames (1,2),(2,3),(3,4)...
-        Ex: dt=2, so compare frames (1,3),(3,5),(5,7)...
-        Ex: dt=3, so compare frames (1,4),(4,7),(7,10)...
+        Compare the molecules in the `cur_frame` selection to the
+        `other_frame` selection and
+        select only those particles that are repeated in both frame. This is to
```
tylerjereddy, line 430: `frame` -> `frames`

package/MDAnalysis/analysis/waterdynamics.py
```diff
@@ -388,79 +316,147 @@
...
+        a certain time has elapsed. The result is a list with the indices of
+        the atoms.
+        """
+        a = set(selection[cur_frame])
+        b = set(selection[other_frame])
+        return sorted(list(a.intersection(b)))
+
+    def _get_repeated_indices(self, selection, frame):
+        """
+        Indicates the comparation between the current frame and all
```
tylerjereddy, line 441: `comparation` -> `comparison` I think; interestingly, `comparation` does appear to be a word related to `comparison`, but largely considered obsolete from what I find online

package/MDAnalysis/analysis/waterdynamics.py
```diff
@@ -388,79 +316,147 @@
...
+        following frames.
+        The results is a list of lists with all the repeated index per frame.
```
tylerjereddy, line 443: `index` -> `indices`?

testsuite/MDAnalysisTests/analysis/test_waterdynamics.py
```diff
@@ -41,18 +41,23 @@ def universe():
...
 
 
 def test_WaterOrientationalRelaxation_zeroMolecules(universe):
-    wor_zero = waterdynamics.WaterOrientationalRelaxation(
-        universe, SELECTION2, 0, 5, 2)
-    wor_zero.run()
-    assert_almost_equal(wor_zero.timeseries[1], (0.0, 0.0, 0.0))
+    wor = waterdynamics.WaterOrientationalRelaxation(universe, SELECTION2)
+    wor.run(0, 5, 1)
+    assert_almost_equal(wor.dip[1], (0.0, 0.0, 0.0))
```
tylerjereddy, line 60: Doesn't have to be changed here, but worth noting that `assert_almost_equal` is discouraged now: https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_almost_equal.html

tylerjereddy (commented):
I added a few drive-by copy edit and test suggestions, but maybe you want to avoid adding more changes to keep the diff small for now.

package/MDAnalysis/analysis/waterdynamics.py
PicoCentauri: Huh interesting. I changed it.

testsuite/MDAnalysisTests/analysis/test_waterdynamics.py
PicoCentauri: Thanks for the hint. This is an easy change.

orbeckst:
Are these changes are backwards-compatible?

PicoCentauri:
In some way:

* `t0`, the argument for the initial step was silently ignored. 
* `dtmax`, was the argument for the maximal step size considered for the correlation function. During my tests it seemed that only  `dtmax=2`worked. Therefore, I removed the argument `dtmax`.

orbeckst:
It sounds as if these changes get an entry under "Fixes" in the CHANGELOG. Then this PR can go in anytime after 2.0.0.

alejob:
> In some way:
> 
> * `t0`, the argument for the initial step was silently ignored.
> * `dtmax`, was the argument for the maximal step size considered for the correlation function. During my tests it seemed that only  `dtmax=2`worked. Therefore, I removed the argument `dtmax`.


> In some way:
> 
> * `t0`, the argument for the initial step was silently ignored.
> * `dtmax`, was the argument for the maximal step size considered for the correlation function. During my tests it seemed that only  `dtmax=2`worked. Therefore, I removed the argument `dtmax`.


Hi there!

 `dtmax` is an important part of WOR analysis, this because is the time windows where the water molecules relax. Did this change pass the unitary test?

package/MDAnalysis/analysis/waterdynamics.py
```diff
@@ -388,79 +316,147 @@
...
 
     .. versionadded:: 0.11.0
 
     .. versionchanged:: 1.0.0
        Changed `selection` keyword to `select`
+
+    .. versionchanged:: 2.0.0
+       `t0`, `tf` and `dtmax` are renamed to start, stop, step as attributes
+       of the `run` method providing a consistent behaviour with other
+       modules.
```
orbeckst, line 418: We did not deprecate t0, tf, dtmax in 1.1.1 — they are still all documented as standard
https://github.com/MDAnalysis/mdanalysis/blob/1e456e85be62fc1d7d97140b60ab2983310af386/package/MDAnalysis/analysis/waterdynamics.py#L685-L696

This means that for 2.x, it still needs to be possible to use the old code. But you can make them optional kwargs
```python
def __init__(self, universe, select, t0=None, tf=None, dtmax=None):
```
If they are not None, issue a deprecation warning. Fail with `TypeError("If the deprecated t0, tf, and dtmax are used, all of them are required")` if not all three of them are set (as the original code would have done).

You might have to subclass `run()` to substitute t0, tf, dtmax for start, stop, step.

package/MDAnalysis/analysis/waterdynamics.py
```diff
@@ -388,79 +316,147 @@
...
     """
 
-    def __init__(self, universe, select, t0, tf, dtmax, nproc=1):
```
orbeckst, line 438: Document removal of `nproc` as a versionchanged ("Removed optional ``nproc`` kwarg that was not used.") and in CHANGELOG Changes.

I think we can get away without deprecations for this one.

package/MDAnalysis/analysis/waterdynamics.py
orbeckst: If `run()` replaces start/stop/step then issue a deprecation warning, too.

package/MDAnalysis/analysis/waterdynamics.py
```diff
@@ -388,79 +316,147 @@
...
+        a certain time has elapsed. The result is a list with the indices of
+        the atoms.
+        """
+        a = set(selection[cur_frame])
+        b = set(selection[other_frame])
+        return sorted(list(a.intersection(b)))
+
+    def _get_repeated_indices(self, selection, frame):
+        """
+        Indicates the comparasion between the current frame and all
```
orbeckst, line 441: comparison

package/MDAnalysis/analysis/waterdynamics.py
```diff
@@ -388,79 +316,147 @@
...
+    time correlation functions given by:
 
     .. math::
         C_{\hat u}(\tau)=\langle \mathit{P}_2[\mathbf{\hat{u}}(t_0)\cdot\mathbf{\hat{u}}(t_0+\tau)]\rangle
 
-    where :math:`P_2=(3x^2-1)/2` is the second-order Legendre polynomial and :math:`\hat{u}` is
-    a unit vector along HH, OH or dipole vector.
-
+    where :math:`P_2=(3x^2-1)/2` is the second-order Legendre polynomial
+    and :math:`\hat{u}` is a unit vector along HH, OH or dipole vector.
```
orbeckst, line 348: Please add a warning that this class assumes a 3-point water model where each water residue contains O as the first entry and hydrogen 1 and 2 as second and third entry.

This warning should have been here from the beginning. Even better would be a more robust approach to choosing O and H1, H2 but that's not the job of this PR.

package/MDAnalysis/analysis/waterdynamics.py
```diff
@@ -521,79 +517,67 @@ def _getOneDeltaPoint(self, universe, repInd, i, t0, dt):
...
+        repeated_indices = self._get_repeated_indices(selections, self.frame)
+        cum_frames = 0
+        sumDeltaOH = 0
+        sumDeltaHH = 0
+        sumDeltadip = 0
+        n = 0
 
-        for j in range(totalFrames // dt - 1):
-            a = self._getOneDeltaPoint(universe, repInd, j, sumsdt, dt)
+        for i in range(self.start, self.stop // self.frame - 1):
```
orbeckst, line 537: I don't understand why `self.stop // self.frame - 1`. Can you add a comment here explaining what this loop does?

package/MDAnalysis/analysis/waterdynamics.py
```diff
@@ -388,79 +316,147 @@
...
     universe : Universe
       Universe object
     selection : str
       Selection string for water [‘byres name OH2’].
-    t0 : int
-      frame  where analysis begins
-    tf : int
-      frame where analysis ends
-    dtmax : int
-      Maximum dt size, `dtmax` < `tf` or it will crash.
```
orbeckst, line 429: @alejob what is `dtmax` supposed to be doing?

Is it the same as a stride when you subselect frames from a trajectory as in `trajectory[start:stop:step]`? If not,  please explain.

alejob: It is not the same as trajectory[start:stop:step].

First, we need to define `dt`. `dt` is the delta time in which WOR is calculated. For instance, if `dt=5` this means that WOR will be calculated comparing frame 1 with frame 5, and then comparing frame 6 with 10, then 11 with 15, and so on. Then we get the mean of these values and get the final value for `dt=5`. If `dt=8`, WOR will be calculated between frame 1 and frame 8, between 9 and 16 and repeat all the process. Now, `dtmax` indicates the maximum value that `dt` could have, and indicates that all the minor `dt` will be calculated, in other words, if you plot the output of this analysis, `dtmax` is the maximum value of the `x` axis. 

Here an example modified from the docs:
If you have that `t0 = 0`, `tf = 1000` and `dtmax = 20`,  we create 20 windows
timesteps (20 values in the x axis), the first window is created with 1000
timestep average (1000/1), the second window is created with 500 timestep
average(1000/2), the third window is created with 333 timestep average (1000/3)
and so on. Obtaining the mean of many timesteps gives us a better statistic for the WOR value, _i.e_, a smoother curve.

Because of this fix `dtmax` in `dtmax=2` has non sense, `dtmax` should be chosen by the user.

orbeckst:
@PicoCentauri is this PR something you might want to come back to?

PicoCentauri:
Currently, I have to no time to look into this in great details. When we had the last round of nice feedback I had the feeling that the [current state of the class is physically broken](https://github.com/MDAnalysis/mdanalysis/pull/3247/files#r632271000) and that a transition to the AnalysisBase is not as straightforward as I thought in the beginning.

IAlibay:
> Currently, I have to no time to look into this in great details. When we had the last round of nice feedback I had the feeling that the [current state of the class is physically broken](https://github.com/MDAnalysis/mdanalysis/pull/3247/files#r632271000) and that a transition to the AnalysisBase is not as straightforward as I thought in the beginning.

@PicoCentauri this is on the 2.2.0 milestone, from this comment shall I assume that this is not going to make it? (I've not been keeping up with the review comments sorry).

If so, which future milestone shall I add this to? 2.3.0 or 3.0.0?

PicoCentauri:
Probabaly 3.0.0

PR #3249, opened 2021-04-30
Coverage density raised to 100%  [issue #3207]

### LINKED ISSUE #3207: Increase density code test coverage
The current density code only has 82.04% test coverage (see [here](https://codecov.io/gh/MDAnalysis/mdanalysis/src/develop/package/MDAnalysis/analysis/density.py) )
The general aim is that we should get 100% coverage for everything.

Tests to cover the uncovered areas

## Additional context
This is an issue aimed at providing a simple issue for first time contributing GSoC students to work on.

ojeda-e:
Fixes #3207 
Density code test coverage raised to 100%.

- Test coverage added in lines
-- 771-774 
-- 738-739
-- 754-755
-- 803-808
-- 810-869 (`convert_density`).

(No Docs needed)

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -14,7 +14,7 @@ The rules for this file:
 
 ------------------------------------------------------------------------------
 ??/??/?? IAlibay, melomcr, mdd31, ianmkenney, richardjgowers, hmacdope,
-         orbeckst, scal444, p-j-smith, edisj, Atharva7K
+         orbeckst, scal444, p-j-smith, edisj, Atharva7K, ojeda-e
 
  * 2.1.0
 
--- a/testsuite/MDAnalysisTests/analysis/test_density.py
+++ b/testsuite/MDAnalysisTests/analysis/test_density.py
@@ -22,6 +22,7 @@
 #
 import numpy as np
 import pytest
+import copy
 import sys
 import warnings
 from unittest.mock import Mock, patch
@@ -33,7 +34,7 @@ import gridData.OpenDX
 import MDAnalysis as mda
 from MDAnalysis.analysis import density
 
-from MDAnalysisTests.datafiles import TPR, XTC, GRO, PDB_full
+from MDAnalysisTests.datafiles import TPR, XTC
 from MDAnalysisTests.util import block_import
 
 
@@ -59,6 +60,13 @@ class TestDensity(object):
         d = density.Density(h, edges, parameters={'isDensity': False},
                             units={'length': 'A'})
         d.make_density()
+        return d
+
+    @pytest.fixture()
+    def D1(self, h_and_edges):
+        h, edges = h_and_edges
+        d = density.Density(h, edges, parameters={},
+                            units={})
         return d
 
     def test_shape(self, D):
@@ -105,6 +113,76 @@ class TestDensity(object):
         with pytest.raises(ValueError):
             D._check_set_unit(units)
 
+    def test_check_set_unit_density_none(self, D):
+        units = {'density': None}
+        assert D._check_set_unit(units) == None
+
+    def test_check_set_unit_density_none(self, D):
+        del D.units['density']
+        D._check_set_unit({})
+        assert D.units['density'] == None
+        assert D._check_set_unit(D.units) == None
+
+    def test_parameters_isdensity(self, D):
+        with pytest.warns(UserWarning, match='Running make_density()'):
+            D.make_density()
+
+    def test_check_convert_density_grid_not_density(self, D1):
+        unit = 'nm^{-3}'
+        with pytest.raises(RuntimeError, match="The grid is not a density"):
+            D1.convert_density(unit)
+
+    def test_check_convert_density_value_error(self, D):
+        unit = 'A^{-2}'
+        with pytest.raises(ValueError, match="The name of the unit"):
+            D.convert_density(unit)
+
+    def test_check_convert_density_units_same_density_units(self, D):
+        unit = 'A^{-3}'
+        D_orig = copy.deepcopy(D)
+        D.convert_density(unit)
+        assert D.units['density'] == D_orig.units['density'] == unit
+        assert_almost_equal(D.grid, D_orig.grid)
+
+    def test_check_convert_density_units_density(self, D):
+        unit = 'nm^{-3}'
+        D_orig = copy.deepcopy(D)
+        D.convert_density(unit)
+        assert D.units['density'] == 'nm^{-3}'
+        assert_almost_equal(D.grid, 10**3 * D_orig.grid)
+
+    def test_convert_length_same_length_units(self, D):
+        unit = 'A'
+        D_orig = copy.deepcopy(D)
+        D.convert_length(unit)
+        assert D.units['length'] == D_orig.units['length'] == unit
+        assert_almost_equal(D.grid, D_orig.grid)
+
+    def test_convert_length_other_length_units(self, D):
+        unit = 'nm'
+        D_orig = copy.deepcopy(D)
+        D.convert_length(unit)
+        assert D.units['length'] == unit
+        assert_almost_equal(D.grid, D_orig.grid)
+
+    def test_repr(self, D, D1):
+        assert str(D) == '<Density density with (3, 4, 5) bins>'
+        assert str(D1) == '<Density histogram with (3, 4, 5) bins>'
+
+    def test_check_convert_length_edges(self, D):
+        D1 = copy.deepcopy(D)
+        unit = 'nm'
+        D.convert_length(unit)
+        for prev_edge, conv_edge in zip(D1.edges, D.edges):
+            assert_almost_equal(prev_edge, 10*conv_edge)
+
+    def test_check_convert_density_edges(self, D):
+        unit = 'nm^{-3}'
+        D_orig = copy.deepcopy(D)
+        D.convert_density(unit)
+        for new_den, orig_den in zip(D.edges, D_orig.edges):
+            assert_almost_equal(new_den, orig_den)
+
     @pytest.mark.parametrize('dxtype',
                              ("float", "double", "int", "byte"))
     def test_export_types(self, D, dxtype, tmpdir, outfile="density.dx"):
@@ -116,6 +194,7 @@ class TestDensity(object):
             data = dx.components['data']
         assert data.type == dxtype
 
+
 class DensityParameters(object):
     topology = TPR
     trajectory = XTC
@@ -126,7 +205,7 @@ class DensityParameters(object):
                   'solute': "protein and not name H*",
                   }
     references = {'static':
-                      {'meandensity': 0.016764271713091212, },
+                  {'meandensity': 0.016764271713091212, },
                   'static_sliced':
                       {'meandensity': 0.016764270747693617, },
                   'static_defined':
@@ -150,6 +229,7 @@ class DensityParameters(object):
     def universe(self):
         return mda.Universe(self.topology, self.trajectory, tpr_resid_from_one=False)
 
+
 class TestDensityAnalysis(DensityParameters):
     def check_DensityAnalysis(self, ag, ref_meandensity,
                                     tmpdir, runargs=None, **kwargs):
@@ -162,8 +242,8 @@ class TestDensityAnalysis(DensityParameters):
 
             D2 = density.Density(self.outfile)
             assert_almost_equal(
-                    D.results.density.grid, D2.grid, decimal=self.precision,
-                    err_msg="DX export failed: different grid sizes"
+                D.results.density.grid, D2.grid, decimal=self.precision,
+                err_msg="DX export failed: different grid sizes"
             )
 
     @pytest.mark.parametrize("mode", ("static", "dynamic"))
@@ -298,7 +378,7 @@ class TestDensityAnalysis(DensityParameters):
 
     def test_warn_results_deprecated(self, universe):
         D = density.DensityAnalysis(
-                universe.select_atoms(self.selections['static']))
+        universe.select_atoms(self.selections['static']))
         D.run(stop=1)
         wmsg = "The `density` attribute was deprecated in MDAnalysis 2.0.0"
         with pytest.warns(DeprecationWarning, match=wmsg):
@@ -323,7 +403,7 @@ class TestGridImport(object):
         # mock gridData in case there are testing scenarios where
         # it is not available
         mock = Mock()
-        with patch.dict('sys.modules', {'gridData':mock}):
+        with patch.dict('sys.modules', {'gridData': mock}):
             try:
                 import MDAnalysis.analysis.density
             except ImportError:
--- a/testsuite/MDAnalysisTests/analysis/test_density.py
+++ b/testsuite/MDAnalysisTests/analysis/test_density.py
@@ -113,18 +113,23 @@ class TestDensity(object):
         with pytest.raises(ValueError):
             D._check_set_unit(units)
 
-    def test_check_set_unit_density_none(self, D):
+    def test_check_set_density_none(self, D1):
         units = {'density': None}
-        assert D._check_set_unit(units) is None
+        D1._check_set_unit(units)
+        assert D1.units['density'] is None
+
+    def test_check_set_density_not_in_units(self, D1):
+        del D1.units['density']
+        D1._check_set_unit({})
+        assert D1.units['density'] is None
 
     def test_parameters_isdensity(self, D):
         with pytest.warns(UserWarning, match='Running make_density()'):
             D.make_density()
 
     def test_check_convert_density_grid_not_density(self, D1):
-        unit = 'nm^{-3}'
         with pytest.raises(RuntimeError, match="The grid is not a density"):
-            D1.convert_density(unit)
+            D1.convert_density()
 
     def test_check_convert_density_value_error(self, D):
         unit = 'A^{-2}'
@@ -226,10 +231,11 @@ class DensityParameters(object):
 
 class TestDensityAnalysis(DensityParameters):
     def check_DensityAnalysis(self, ag, ref_meandensity,
-                                    tmpdir, runargs=None, **kwargs):
+                              tmpdir, runargs=None, **kwargs):
         runargs = runargs if runargs else {}
         with tmpdir.as_cwd():
-            D = density.DensityAnalysis(ag, delta=self.delta, **kwargs).run(**runargs)
+            D = density.DensityAnalysis(
+                ag, delta=self.delta, **kwargs).run(**runargs)
             assert_almost_equal(D.results.density.grid.mean(), ref_meandensity,
                                 err_msg="mean density does not match")
             D.results.density.export(self.outfile)
@@ -372,7 +378,7 @@ class TestDensityAnalysis(DensityParameters):
 
     def test_warn_results_deprecated(self, universe):
         D = density.DensityAnalysis(
-        universe.select_atoms(self.selections['static']))
+            universe.select_atoms(self.selections['static']))
         D.run(stop=1)
         wmsg = "The `density` attribute was deprecated in MDAnalysis 2.0.0"
         with pytest.warns(DeprecationWarning, match=wmsg):
--- a/testsuite/MDAnalysisTests/analysis/test_density.py
+++ b/testsuite/MDAnalysisTests/analysis/test_density.py
@@ -115,13 +115,7 @@ class TestDensity(object):
 
     def test_check_set_unit_density_none(self, D):
         units = {'density': None}
-        assert D._check_set_unit(units) == None
-
-    def test_check_set_unit_density_none(self, D):
-        del D.units['density']
-        D._check_set_unit({})
-        assert D.units['density'] == None
-        assert D._check_set_unit(D.units) == None
+        assert D._check_set_unit(units) is None
 
     def test_parameters_isdensity(self, D):
         with pytest.warns(UserWarning, match='Running make_density()'):
--- a/package/MDAnalysis/analysis/density.py
+++ b/package/MDAnalysis/analysis/density.py
@@ -865,7 +865,7 @@ class Density(Grid):
 
         """
         if not self.parameters['isDensity']:
-            errmsg = 'The grid is not a density so converty_density() makes no sense.'
+            errmsg = 'The grid is not a density so convert_density() makes no sense.'
             logger.fatal(errmsg)
             raise RuntimeError(errmsg)
         if unit == self.units['density']:


PicoCentauri (approved):
LGTM 🚀

package/CHANGELOG
```diff
@@ -23,6 +23,7 @@ The rules for this file:
   * 2.0.0
 
 Fixes
+  * Test coverage density code increased to 100% (Issue #3207)
```
IAlibay, line 26:
```suggestion
  * Test coverage of the density code increased to 100% (Issue #3207) 
```

Sorry 😅 I know it's pedantic.

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -94,6 +111,10 @@ def test_check_set_unit_keyerror(self, D):
         with pytest.raises(ValueError):
             D._check_set_unit(units)
 
+    def test_check_set_unit_none_value(self, D):
+        units = {'density': None}
+        D._check_set_unit(units)
```
IAlibay, line 116: Sorry it's late and I probably am missing what we're doing here but shouldn't we be checking that the units have been set properly?

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -105,6 +126,35 @@ def test_check_set_unit_nolength(self, D):
         with pytest.raises(ValueError):
             D._check_set_unit(units)
 
+    def test_check_set_unit_nodensity(self, D):
+        del D.units['density']
+        units = {'length': 'A'}
+        D._check_set_unit(units)
```
IAlibay, line 132: As above

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -105,6 +126,35 @@ def test_check_set_unit_nolength(self, D):
...
+
+    def test_convert_length(self, D):
+        D.convert_length()
```
IAlibay, line 135: What are we testing here?

IAlibay (changes requested):
Sorry I'm probably missing something really simple so please feel free to tell me I'm wrong.

It looks like we are calling convert methods but not checking their return? (i.e. I'm not actually sure what we are testing except that the code paths don't crash).

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -105,6 +126,35 @@ def test_check_set_unit_nolength(self, D):
...
+        with pytest.raises(ValueError, match='The name of the unit'):
+            D.convert_density()
+
+    def test_convert_density_units(self, D):
+        unit = 'nm^{-3}'
+        D.convert_density(unit)
+
+    def test_repr_(self, D, hist):
+        D.__repr__()
+        hist.__repr__()
```
tylerjereddy, line 156: Similar to Irfan's other comments--this just produces two strings so doesn't do much apart from flushing code paths. Making some kind of assertion about the contents of the strings seems like it might be in order for this case?

ojeda-e:
Hi @IAlibay, sorry for the long delay.  

I think I changed the code as requested:

- Function `test_check_set_unit_density_none` added. 
Coverage raised from 82 -> 84% in lines 757-758.

- Function `test_parameters_isdensity` added. 
Coverage raised from 84 -> 86% in lines 789-792.

- Function `test_check_convert_density_grid_not_density` added. 
Coverage raised from 86 -> 89% in lines 869-871.

- Function `test_check_convert_density_value_error` added. 
Coverage raised from 89 -> 92% in lines 874-880.

- Function `test_check_convert_density_units_same_density_units` and
`test_check_convert_density_units_density` added. 
Coverage raised from 92 -> 93% in lines 872 and 881.

- Function `test_convert_length_same_length_units` added. 
Coverage raised from 93 -> 95% in line 822.

- Function `test_parameters_is_Density` added. 
Coverage raised from 95 -> 97% in line 883.

-Function `test_check_convert_length` added.
Coverage raised from 97 -> 100%; lines 824-827

About the last function, `test_check_convert_length`, I think you may want me to add an assertion for self.edges. I tried using `assert_almost_equal` for the two bumpy arrays, however, it didn't work. I got a warning about defining `dtype=object`. If you want that line added,I would like to have some suggestions of how to approach it from your side.
   
Thanks.

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -105,6 +112,67 @@ def test_check_set_unit_nolength(self, D):
...
+    def test_check_convert_density_units_same_density_units(self, D):
+        unit = 'A^{-3}'
+        D.convert_density(unit)
+        assert unit == D.units['density']
+
+    # 93 -> 93%; line 881
+    def test_check_convert_density_units_density(self, D):
+        unit = 'nm^{-3}'
+        D.convert_density(unit)
+        assert D.units['density'] == unit
```
IAlibay, line 155: Rather than just checking that the units have been set. It would be good to also check that numerically the conversion has been made in a way that would be expected.

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -105,6 +112,67 @@ def test_check_set_unit_nolength(self, D):
...
+        assert unit == D.units['length']
+
+    # 95 -> 97%; line 884
+    def test_parameters_is_Density(self, D, D1):
+        D.__repr__()
+        D1.__repr__()
+        assert 'isDensity' in D1.parameters
+
+    # 97 -> 100%; lines 824-827
+    def test_check_convert_length(sef, D):
```
IAlibay, line 170: Same as above, we probably should be checking that the edges, origin, etc... are all as expected?

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -105,6 +112,67 @@ def test_check_set_unit_nolength(self, D):
...
+        unit='nm'
+        D.convert_length(unit)
+        D._update()
```
IAlibay, line 173: Doesn't `convert_length` already call `_update` if it's not the same units?

IAlibay (changes requested):
Sorry I don't have time for a full review today. Here are some quick comments.
Where possible we should ensure that the right numerical operations have been done on the density.

testsuite/MDAnalysisTests/analysis/test_density.py
ojeda-e: In the previous submission I had a question regarding the `numpy_array_almost_equal` to check the elements you mentioned are the same since I was getting the following error:
`numpy.VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.`
According to the documentation, `numpy_array_almost_equal` should work for numpy arrays. I got confused and was asking for some guide to approach this.

I think I fixed it by comparing element by element using `zip` as you can see for edges in `test_check_convert_density` and `test_check_convert_length` of the most recent commit (lines 168-169, and 175-176) . Can you please confirm that's a valid approach before adding to origin, etc?
Thanks.

testsuite/MDAnalysisTests/analysis/test_density.py
ojeda-e: Thanks, you're correct. I rewrote all the test for `__repr__`.

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -105,6 +112,69 @@ def test_check_set_unit_nolength(self, D):
...
+
+    def test_check_convert_length_edges(self, D):
+        D1 = D
+        new_unit = 'nm'
+        D.convert_length(new_unit)
+        for prev_edge, conv_edge in zip(D1.edges, D.edges):
+            assert_almost_equal(prev_edge, conv_edge)
+
+    def test_check_convert_density_edges(self, D):
+        D1 = D
```
IAlibay, line 172: Please double check, but I'm pretty sure this is a shallow copy, so it won't work as a comparison if you're modifying D in place (same for the test above).

testsuite/MDAnalysisTests/analysis/test_density.py
IAlibay: If you can give us a example of what you were trying to use the `assert_almost_equal` on (note, if I remember correctly it's best to use that over `array_almost_equal` @lilyminium had some explanation about this somewhere), I can probably offer some insight into what's going on.

For the `grid`, `origin`, and `delta` attributes, assert_almost_equal should work as-is.

For the `edges` attribute you'll probably have to iterate (I can't remember if there's a better way to deal with that kind of list of arrays), something like what you're doing would work.

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -105,6 +112,69 @@ def test_check_set_unit_nolength(self, D):
...
+    def test_convert_length_same_length_units(self, D):
+        unit = 'A'
+        D.convert_length(unit)
+        assert unit == D.units['length']
+        unit='nm'
+        D.convert_length(unit)
+        assert unit == D.units['length']
+
+    def test_parameters_is_Density(self, D, D1):
+        D.__repr__()
```
IAlibay, line 159: re-echoing Tyler's comment here. The `__repr__` is essentially a means of allowing a class to have a printable string.

Here I would create a density with and without `isDensity` and then do:

assert str(density_grid) = '<Density density with insert_shape_here bins>'
assert str(not_density_grid) = '<Density histogram with insert_shape_here bins>'

testsuite/MDAnalysisTests/analysis/test_density.py
IAlibay: Note: it would be more than reasonable [if not even suggested] to have this test as part of the same test where you compare the grid value change (same for the convert_length stuff).

testsuite/MDAnalysisTests/analysis/test_density.py
ojeda-e: An example of the behaviour is below
```
    def test_check_fail(self, D, h_and_edges):
        h, edges = h_and_edges
        assert_almost_equal(edges, D.edges)
    
    def test_check_no_fail(self, D, h_and_edges):
        h, edges = h_and_edges
        for i, j in zip(edges, D.edges):
            assert_almost_equal(i, j)
```
The error I get is
```
E       numpy.VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.

../../../opt/anaconda3/envs/mdanalysis-dev/lib/python3.7/site-packages/numpy/core/_asarray.py:102: VisibleDeprecationWarning
```

The assertion is intended for the `edges`, which is a numpy array. [The documentation](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_almost_equal.html) gives an example of comparison of two numpy arrays. 
Printing `edges` and `D.edges` reads as
```
[array([ 0.        ,  3.33333333,  6.66666667, 10.        ]), array([ 0. ,  2.5,  5. ,  7.5, 10. ]), array([ 0.,  2.,  4.,  6.,  8., 10.])]
[array([ 0.        ,  3.33333333,  6.66666667, 10.        ]), array([ 0. ,  2.5,  5. ,  7.5, 10. ]), array([ 0.,  2.,  4.,  6.,  8., 10.])]

```
Hence my question if the approach by using `zip` to compare individual elements is valid. I get that in this case `edges` is list of arrays. Which I eventually could transform to `np.array(edges)`, but even in that case `assert_almost_equal` fails and prints the same error.

testsuite/MDAnalysisTests/analysis/test_density.py
ojeda-e: Is this better? (`grid` is missing, though)

```
    def test_check_convert_length(self, D, h_and_edges, bins):
        midpoints = [0.5*(bin[:-1] + bin[1:]) for bin in bins]
        deltas = np.array([self.Lmax])/np.array(self.nbins)
        origin = [mp[0] for mp in midpoints]
        h, edges = h_and_edges
        new_unit = 'nm'
        D.convert_length(new_unit)

        for prev_edge, conv_edge in zip(edges, D.edges):
            assert_almost_equal(prev_edge*0.1,  conv_edge)
        
        for prev_origin, conv_origin in zip(origin, D.origin):
            assert_almost_equal(prev_origin*0.1,  conv_origin)
        
        for prev_delta, conv_delta in zip(deltas, D.delta):
            assert_almost_equal(prev_delta*0.1,  conv_delta)
```

testsuite/MDAnalysisTests/analysis/test_density.py
ojeda-e: Sure, I considered to move it to same function that check convert_density but was hesitating. Thanks.

testsuite/MDAnalysisTests/analysis/test_density.py
ojeda-e: Yes, my bad, now it's gone.

testsuite/MDAnalysisTests/analysis/test_density.py
ojeda-e: Ok, no, I would have never seen that myself. Thanks.

testsuite/MDAnalysisTests/analysis/test_density.py
orbeckst: The zip approach is good: the elements of the list are simply arrays of different lengths; I can't think of a more sensible way to do the tests than what you did.

testsuite/MDAnalysisTests/analysis/test_density.py
orbeckst: See my comment with `deepcopy` on how you could also check that the values changed:
```python
...
assert D.units['density'] ==  unit
assert_almost_equal(D.grid, 10**3 * D_orig.grid)   # or write 1e3 * ... but I wanted to make the factor clear
```

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -105,6 +112,69 @@ def test_check_set_unit_nolength(self, D):
...
+    def test_check_convert_density_units_density(self, D):
+        unit = 'nm^{-3}'
+        D.convert_density(unit)
+        assert D.units['density'] == unit
+
+    def test_convert_length_same_length_units(self, D):
+        unit = 'A'
+        D.convert_length(unit)
+        assert unit == D.units['length']
+        unit='nm'
```
orbeckst, line 154: pep8 spaces
```suggestion
        unit = 'nm'
```

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -105,6 +112,69 @@ def test_check_set_unit_nolength(self, D):
...
+
+    def test_check_convert_density_units_density(self, D):
+        unit = 'nm^{-3}'
+        D.convert_density(unit)
+        assert D.units['density'] == unit
+
+    def test_convert_length_same_length_units(self, D):
+        unit = 'A'
+        D.convert_length(unit)
+        assert unit == D.units['length']
```
orbeckst, line 153: slightly more readable if you put the reference on the RHS
```suggestion
        assert D.units['length'] == unit
```

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -105,6 +112,69 @@ def test_check_set_unit_nolength(self, D):
...
+        unit='nm'
+        D.convert_length(unit)
+        assert unit == D.units['length']
```
orbeckst, line 156: switch LHS with RHS

testsuite/MDAnalysisTests/analysis/test_density.py
orbeckst, line 154: When you switch to `nm`, you're not testing "same length unit" any more, are you? Then it would be better to make it a separate test.

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -105,6 +112,69 @@ def test_check_set_unit_nolength(self, D):
...
+        D.convert_density(unit)
+        assert unit == D.units['density']
+
+    def test_check_convert_density_units_density(self, D):
+        unit = 'nm^{-3}'
+        D.convert_density(unit)
+        assert D.units['density'] == unit
+
+    def test_convert_length_same_length_units(self, D):
+        unit = 'A'
```
orbeckst, line 151: Also test the _value_ of the grid array.

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -105,6 +112,69 @@ def test_check_set_unit_nolength(self, D):
...
+        D.convert_length(unit)
+        assert unit == D.units['length']
+        unit='nm'
+        D.convert_length(unit)
+        assert unit == D.units['length']
+
+    def test_parameters_is_Density(self, D, D1):
```
orbeckst, line 158: rename `test_repr`.

testsuite/MDAnalysisTests/analysis/test_density.py
orbeckst: You can use `copy.deepcopy()` before changing the Density.

`Density` could really have a `Density.copy()` method to do it by itself, but that's for another day.

orbeckst (changes requested):
As stated by previous reviewers, we want to test _values_ in addition to metadata changes. One possible approach is to make deep copies (see comments). Please address the outstanding comments and then we can merge this — thank you!!

orbeckst:
@ojeda-e I merged the latest develop (in CHANGELOG I added your name to 2.1.0 but I removed the entry itself because users don't care about test coverage in the CHANGELOG). If you could address the remaining comments then we could probably get this in quickly.

testsuite/MDAnalysisTests/analysis/test_density.py
ojeda-e: Thanks. Separate test added.

package/CHANGELOG
```diff
@@ -38,21 +38,7 @@ Deprecations
   * 2.0.0
 
 Fixes
-  * Added an `rdkit_kwargs` parameter to pass arguments to the RDKitConverter
```
orbeckst: The CHANGELOG got messed up, possibly by force pushing.

orbeckst (changes requested):
Code looks good, just CHANGELOG got messed up and you need to add entries under 2.1.0. Perhaps pull develop and then rebase your branch onto the latest develop?

ojeda-e:
Sorry about it, @orbeckst.
I think it's fixed now although I don't fully understand the change shown in the [last line](https://github.com/MDAnalysis/mdanalysis/pull/3249/files#diff-4ac8d9ef37778aec14402a8cba558978f084f8359a6fbb7f494c7be8b8e6c5ccR2770) of the CHANGELOG.
The file from the `develop` branch has no line at the end. 
Thanks.

orbeckst:
Now you're missing from CHANGELOG, you should be listed as a 2.1.0 author. Do you want to update the CHANGELOG before I merge?

Just add the new line ending in the last line in the CHANGELOG back, it seems your editor removed it eagerly in 
57d66ce ; the smaller the diff the better and then at least someone else's editor will not auto-add it back. Teach your tools to only change things that _you_ want to edit (they don't always know better ;-) ).

orbeckst (approved):

orbeckst:
@IAlibay could you please check here if all of your comments have been addressed? Thanks!

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -105,6 +113,78 @@ def test_check_set_unit_nolength(self, D):
         with pytest.raises(ValueError):
             D._check_set_unit(units)
 
+    def test_check_set_unit_density_none(self, D):
+        units = {'density': None}
+        assert D._check_set_unit(units) == None
+
+    def test_check_set_unit_density_none(self, D):
+        units = {'density': None}
```
IAlibay: The test logic here isn't clear to me.

1) You start with units of {'length': 'A', 'density': 'A^{-3}'}
2) Then you set units to None again
3) Then you delete the density units
4) Then you test that passing an empty dictionary will yield units of None for 'density'

The only thing that properly gets tested here is number 4 in that list (3 is kinda meant to test 4). What's the point for actions 1 and 2, why not just delete the units from the start? If the point is to check what happens when you pass None, then should there be a second assert?

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -105,6 +113,78 @@ def test_check_set_unit_nolength(self, D):
...
+        unit = 'A^{-3}'
+        D_orig = copy.deepcopy(D)
+        D.convert_density(unit)
+        assert D.units['density'] == D_orig.units['density'] == unit
+        assert_almost_equal(D.grid, D_orig.grid)
+
+    def test_check_convert_density_units_density(self, D):
+        unit = 'nm^{-3}'
+        D_orig = copy.deepcopy(D)
+        D.convert_density(unit)
```
IAlibay, line 149: Probably should add an assert here that checks that the unit was set properly (that would match the same test behaviour as the test just above).

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -105,6 +113,78 @@ def test_check_set_unit_nolength(self, D):
...
+    def test_convert_length_other_length_units(self, D):
+        unit = 'nm'
+        D_orig = copy.deepcopy(D)
+        D.convert_length(unit)
+        assert D.units['length'] == unit
+        assert_almost_equal(D.grid, D_orig.grid)
+
+    def test_repr(self, D, D1):
+        D.__repr__()
+        D1.__repr__()
```
IAlibay: str(D) should just call `__repr__` in the background right? I don't _think_ these two calls are needed?

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -298,7 +381,7 @@ def test_ValueError_noatomgroup(self, universe):
 
     def test_warn_results_deprecated(self, universe):
         D = density.DensityAnalysis(
-                universe.select_atoms(self.selections['static']))
+            universe.select_atoms(self.selections['static']))
```
IAlibay, line 381: The visual indentation here is probably worse? Is this the result of an auto-formatter? Personally I'd prefer we stick with the previous formatting for this line.

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -150,20 +231,22 @@ class DensityParameters(object):
...
+
 class TestDensityAnalysis(DensityParameters):
     def check_DensityAnalysis(self, ag, ref_meandensity,
-                                    tmpdir, runargs=None, **kwargs):
+                              tmpdir, runargs=None, **kwargs):
         runargs = runargs if runargs else {}
         with tmpdir.as_cwd():
-            D = density.DensityAnalysis(ag, delta=self.delta, **kwargs).run(**runargs)
+            D = density.DensityAnalysis(
+                ag, delta=self.delta, **kwargs).run(**runargs)
```
IAlibay, line 238: IMHO this is visually worse than what it was before. If it's to avoid PEP8 issues, I'd at least space indent the arguments here.

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -126,7 +207,7 @@ class DensityParameters(object):
                   'solute': "protein and not name H*",
                   }
     references = {'static':
-                      {'meandensity': 0.016764271713091212, },
+                  {'meandensity': 0.016764271713091212, },
```
IAlibay, line 207: The indentation here is no longer consistent with the rest of the dictionary, unless there's a good reason, I'd just stick with what it was?

IAlibay (changes requested):
Couple of quick comments.

orbeckst:
@ojeda-e can I motivate you to address the requested changes to get the PR done and into the 2.1.0 release?

testsuite/MDAnalysisTests/analysis/test_density.py
ojeda-e: Thanks for pointing this out, indentation was unintended. Fixed here and lines below.

testsuite/MDAnalysisTests/analysis/test_density.py
ojeda-e: Thanks, _I think_ I followed your suggestions and change the test:
1. deleted units, 
2. assert units,
3. assert check_set_units.

ojeda-e:
Thanks for the motivation @orbeckst. ⚡ 
@IAlibay, I hope I addressed all your comments, sent a request to have your review.

package/CHANGELOG
```diff
@@ -13,7 +13,7 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" http://semver.org
 
 ------------------------------------------------------------------------------
-??/??/?? IAlibay, melomcr, mdd31, ianmkenney, richardjgowers, hmacdope
+??/??/?? IAlibay, melomcr, mdd31, ianmkenney, richardjgowers, hmacdope, ojeda-e
```
IAlibay: This is going to have to spillover to the next line next time someone adds an entry - we should just break it into two lines now to save time later.

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -105,6 +113,76 @@ def test_check_set_unit_nolength(self, D):
         with pytest.raises(ValueError):
             D._check_set_unit(units)
 
+    def test_check_set_unit_density_none(self, D):
+        units = {'density': None}
+        assert D._check_set_unit(units) == None
```
IAlibay: I might be missing the point of what this is doing here.
`_check_set_unit` has no return right? So asserting None is doesn't do anything?
Shouldn't you be asserting that `D.units['density'] is None`? (note that `is` is better than `==` for `None` checking).

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -105,6 +113,76 @@ def test_check_set_unit_nolength(self, D):
...
+
+    def test_check_set_unit_density_none(self, D):
+        del D.units['density']
+        D._check_set_unit({})
+        assert D.units['density'] == None
+        assert D._check_set_unit(D.units) == None
```
IAlibay: Which bit of the code is being tested by passing the same units to itself?

Also - same issue as previous comment, unless I'm missing something obvious `_check_set_unit` has no return so you need to assert on `D.units`.

IAlibay (changes requested):
Apologies, couple more things I didn't catch on to before it seems.

IAlibay:
@ojeda-e just a reminder that there are two final review queries to address on this PR.

testsuite/MDAnalysisTests/analysis/test_density.py
ojeda-e: Yes, sorry about that duplicated `test_check_set_unit_density_none`, this one was supposed to be deleted. The unit test is intended to cover lines 772-773 in `density.py`.

ojeda-e:
My apologies for making the progress in this PR so stupidly slow @IAlibay, also thanks for spotting the duplicated unit test. Pushed the changes you pointed out. Thanks.

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -105,6 +113,70 @@ def test_check_set_unit_nolength(self, D):
         with pytest.raises(ValueError):
             D._check_set_unit(units)
 
+    def test_check_set_unit_density_none(self, D):
+        units = {'density': None}
+        assert D._check_set_unit(units) is None
```
IAlibay: Sorry I'm going to have to query this again, I don't see how asserting `None` on a function that returns nothing does anything else but flush the code path?

Is this just checking lines 755-756?

If so this is what I would do here:

```suggestion
        D._check_set_unit(units)
        assert D.units['density'] is None
```

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -105,6 +113,70 @@ def test_check_set_unit_nolength(self, D):
...
+
+    def test_parameters_isdensity(self, D):
+        with pytest.warns(UserWarning, match='Running make_density()'):
+            D.make_density()
+
+    def test_check_convert_density_grid_not_density(self, D1):
+        unit = 'nm^{-3}'
```
IAlibay: nitpicking but is passing a unit necessary here? Shouldn't kwarg default also trigger this error?

IAlibay (changes requested):
I think we're circling around the same test, I've provided a suggestion for what I _think_ `test_check_set_unit_density_none` should be doing instead. Once dealt with, this PR will be good to go.

testsuite/MDAnalysisTests/analysis/test_density.py
ojeda-e: Yes, you're correct, thanks for catching this.

ojeda-e:
Thanks @IAlibay for clarifying.

Before I submit the changes, I would like to explain what I did and check if I am doing something wrong once again.
There are two cases in `density.py` that involve the key `density` in units dict.

The first one is when `density` is a key in units dict, but has value `None` (line 756), which can be tested with:
```python3
    def test_check_set_density_none(self, D1):
        units = {'density': None}
        D1._check_set_unit(units)
        assert D1.units['density'] is None
```

The second one is when `density` is not a key in units dict (lines 773). To test these lines, I added 
```python3
    def test_check_set_density_not_in_units(self, D1):
        del D1.units['density']
        D1._check_set_unit({})
        assert D1.units['density'] is None
```
Sounds/looks correct?

IAlibay:
> Thanks @IAlibay for clarifying.
> 
> Before I submit the changes, I would like to explain what I did and check if I am doing something wrong once again. There are two cases in `density.py` that involve the key `density` in units dict.
> 
> The first one is when `density` is a key in units dict, but has value `None` (line 756), which can be tested with:
> 
> ```python
>     def test_check_set_density_none(self, D1):
>         units = {'density': None}
>         D1._check_set_unit(units)
>         assert D1.units['density'] is None
> ```
> 
> The second one is when `density` is not a key in units dict (lines 773). To test these lines, I added
> 
> ```python
>     def test_check_set_density_not_in_units(self, D1):
>         del D1.units['density']
>         D1._check_set_unit({})
>         assert D1.units['density'] is None
> ```
> 
> Sounds/looks correct?

Yup, those two tests sound perfect! Thanks @ojeda-e !

IAlibay (approved):
lgtm (assuming CI returns green)

IAlibay:
Codecov reports 100% coverage in density.py, thanks @ojeda-e !

PR #3250, opened 2021-05-01
removed autogenerated C/CPP files from version control

richardjgowers:
Currently we have the .c/.cpp files that cython generates tracked on master but not develop.  The reason for this is so that Cython isn't a dependency for releases, but this is annoying to work with/less easy to automate.

This patch removes them from version control.  I'll open a separate issue where we can make `python sdist` (the thing which makes the tarball which pip/conda work off) add the Cython generated files in.


IAlibay:
Just in case anyone gets trigger happy, please hold until we finish the 1.1.1 release (probably ~ 30 mins)

Issue #3261, opened 2021-05-06
make AnalysisBase-derived classes use .results


orbeckst:
## Is your feature request related to a problem? ##

With PR #3233 merged, we established a new API for AnalysisBase-derived analysis classes: results should be stored in the `.results` attribute, which is a dict that makes keys available as attributes. Any classes that have not been using results should start using it _in a backwards compatible manner_.

## Describe the solution you'd like ##

Add `.results` to all Analysis-classes in such a way that access to the old documented attributes still works for the full 2.x life cycle.

### Classes
The following classes should be upgraded (roughly in order of importance) --- raise individual issues and add them here; add more classes to be upgraded as necessary
- [x] BAT (get this done in 2.0.0 since it was added then = no deprecation needed) #3272
- [x] helix_analysis (as above needs, to get done before 2.0.0 gets released)  #3268
- [x] rms.RMSD PR #3274
- [x] rms.RMSF PR #3274
- [x] align.AlignTraj and align.AverageStructure #3278 
- [x] hydrogenbonds.HydrogenBondAnalysis #3271
- [x] contacts.ContactAnalysis PR #3264
- [x] density.DensityAnalysis PR #3263
- [x] pca.PCA #3275
- [x] dihedrals.Ramachandran PR #3266
- [x] dihedrals.Dihedrals PR #3266
- [x] dihedrals.Janin PR #3266
- [x] msd.EinsteinMSD PR #3265
- [x] rdf.InterRDF #3276
- [x] rdf.InterRDF_s #3276
- [x] hydrogenbonds.WaterBridgeAnalysis #3270
- [x] hole2.HoleAnalysis #3269
- [x] DiffusionMap.DistanceMatrix #3290 
- [x] polymer.PersistenceLength #3291
- [ ] ...

### Approach
Generally follow the approach taken for DensityAnalysis in PR #3263

#### Code
Assume that the old code in a class `ExampleAnalysis(AnalysisBase)` stored a result in attribute `foo`. 

* change to using `.results`:
  * You do _not_ need to set up `self.results = Results()` because `AnalysisBase` does it.
  * You must change any attribute assignments to be in `self.results` instead of `self`.  For example, `self.foo = data` becomes `self.results.foo = data`. 
[...]

PR #3263, opened 2021-05-06
Make DensityAnalysis use the Results class

IAlibay:
Towards #3261 

 - deprecate `.density` in favour of `.results.density`.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


IAlibay:
forgot to update the docs properly -- in progress, also the `deprecate` decorator is just making the docs messy, so I'll just manually add the warning.

PicoCentauri (changes requested):

package/MDAnalysis/analysis/density.py
```diff
@@ -476,7 +483,15 @@ def _conclude(self):
                           units={'length': "Angstrom"},
                           parameters={'isDensity': False})
         density.make_density()
-        self.density = density
+        self.results.density = density
+
+    @property
+    def density(self):
```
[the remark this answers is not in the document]
IAlibay: Yeah it enforces docstring which is just not very human friendly :(
For the sake of a couple of extra lines I'm happy just adding the messages by hand.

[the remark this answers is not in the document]
PicoCentauri: Yes, but this still looks good and as a good template for the other classes.

package/MDAnalysis/analysis/density.py
```diff
@@ -457,7 +463,8 @@ def _prepare(self):
         self._edges = edges
         self._arange = arange
         self._bins = bins
-        self.density = None
+        self.results = Results()
```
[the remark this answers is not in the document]
IAlibay: I had missed that completely 😅 cheers

package/MDAnalysis/analysis/density.py
```diff
@@ -457,7 +467,7 @@ def _prepare(self):
         self._edges = edges
         self._arange = arange
         self._bins = bins
-        self.density = None
+        self.results.density = None
```
PicoCentauri, line 470: Should we also remove this line.`self.results.density` is defined in L489? Or leave it as a kind of comment?

IAlibay: Yeah that's a good point, I guess it never really did anything. We might as well remove it.

PicoCentauri (approved):

package/MDAnalysis/analysis/density.py
```diff
@@ -208,12 +211,16 @@ class DensityAnalysis(AnalysisBase):
...
-    Returns
-    -------
-    :class:`Density`
+    Attributes
+    ----------
+    results.density : :class:`Density`
             A :class:`Density` instance containing a physical density of units
             :math:`Angstrom^{-3}`.
 
+    density : :class:`Density`
```
orbeckst, line 220: You can put the deprecated reST directly here. Not sure if we have a policy that says it must go at the end. It would make sense to have it here.

package/MDAnalysis/analysis/density.py
```diff
@@ -384,6 +391,9 @@ class DensityAnalysis(AnalysisBase):
     .. versionadded:: 1.0.0
     .. versionchanged:: 2.0.0
        :func:`_set_user_grid` is now a method of :class:`DensityAnalysis`.
+    .. deprecated:: 2.0.0
```
orbeckst, line 394: Move to the attribute?

orbeckst (changes requested):
Looking good. Only required thing is another CHANGELOG entry under deprecations. 

Thanks for creating a template for the other cases!

orbeckst:
Please also resolve the conflicts. Thank you!!

orbeckst:
Going forward, is there an improvement we could make to our deprecate decorator?

There is a little “deprecated” package https://python-deprecated.readthedocs.io/en/latest/api.html  with a similar decorator. Maybe we check it out and adapt if it does anything better than ours  (the source code is only a single file).

package/MDAnalysis/analysis/density.py
IAlibay: Thanks for the review. Done, I'll do the same for the other two PRs.

IAlibay:
> Going forward, is there an improvement we could make to our deprecate decorator?
> 
> There is a little “deprecated” package https://python-deprecated.readthedocs.io/en/latest/api.html with a similar decorator. Maybe we check it out and adapt if it does anything better than ours (the source code is only a single file).

From this PR, the main issue is that it didn't want to play well with not having a docstring (I don't remember this being a problem before, I think it might be a clash with `@property`?). The default message isn't super user friendly (at least in my opinion) either, but that is overridable via `message` (although at that point you've written as many lines as it would take to just write your own warning message..).

orbeckst (approved):

orbeckst:
Maybe defining a doc string for the attribute first would help. But ultimately, just doing it explicitly is not wrong and if it's clearer then all the better. 

Perhaps there's a clever helper that could do something like
```python
@deprecated_results_attribute(doc="Density contains a :class:`Density` instance blabla")
def density(self):
    pass
```
which takes `func.__name__` to index `self.results` and adds the docs with the deprecation.

(But time's short so simple & fast is a virtue...)

PR #3273, opened 2021-05-06
Switch BAT to use Results

### LINKED ISSUE #3272: update analysis.bat.BAT to use .results
Follow the new AnalysisBase API and store results in analysis.bat.BAT in `.results` instead of class attributes.

See omnibus issue #3261 for background and details. See PR #3263 as an example (DensityAnalysis) but note that no deprecations are needed (new in 2.0.0)
- [ ] put all generated data into `self.results` instead of `self`
- [ ] keep managed attributes in place with deprecation message

IAlibay:
Fixes #3272
Towards #3261 

 - BAT is new in 2.0.0, so it's a direct switch from `bat` to `results.bat`.
 - Did a bit of cleaning up whilst I was there, including; a misindented code block and added a test.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -174,6 +174,8 @@ Enhancements
     checking if it can be used in parallel analysis. (Issue #2996, PR #2950)
 
 Changes
+  * `bat.BAT` now uses the `analysis.base.Results` class to store the `bat`
+    results attribute (Issue #3261)
   * `contacts.Contacts` now stores data using the `Contacts.results.timeseries`
     attribute. `Contacts.timeseries` is now deprecated (Issue #3261)
   * `DensityAnalysis` now uses the `results.density` attribute for storing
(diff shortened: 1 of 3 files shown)


package/CHANGELOG
```diff
@@ -174,6 +174,8 @@ Enhancements
     checking if it can be used in parallel analysis. (Issue #2996, PR #2950)
 
 Changes
+  * `bat.BAT` now uses the `analysis.base.Results` class to store the `bat`
+    results attribute (Issue #3261, #3272)
```
orbeckst, line 178: Technically not needed because bat wasn't released – but I think I said in the past "if it has an issue then list it"... so leave it in

orbeckst: I changed the issue text to include CHANGELOG entry

orbeckst (approved):
lgtm, thanks

PR #3279, opened 2021-05-06
Move results attributes of helix_analysis to Results

### LINKED ISSUE #3261: make AnalysisBase-derived classes use .results

### LINKED ISSUE #3267: Incorrect versionadded for helix_analysis?
The helix_analysis docstring says that it was added in version 1.0.0

https://github.com/MDAnalysis/mdanalysis/blob/2c714fccd95d8b05f807e38175557e398199b1d6/package/MDAnalysis/analysis/helix_analysis.py#L32

However, it's not (unless I'm missing something) in 1.1.1 (or at least current `master`).

### LINKED ISSUE #3268: update helix_analysis to use .results
Follow the new AnalysisBase API for the new helix_analysis and store results in `.results` instead of class attributes.

This should be done for 2.0.0 #3254 so that we don't have to start this new analysis module with deprecations in place.

See omnibus issue #3261 for background and details. See PR #3263 as an example (DensityAnalysis) but note that _no deprecations are necessary_ — just change where the results are stored and update the docs.

- [ ] use attributes in`self.results` instead of `self`
- [ ] update docs

lilyminium:
Fixes #3268
Fixes #3267

 - changed versionchanged to the right number
 - Moved attributes to `results`

I did *not* add a deprecation warning because this analysis is new in 2.0.0.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/analysis/helix_analysis.py
+++ b/package/MDAnalysis/analysis/helix_analysis.py
@@ -29,7 +29,7 @@ HELANAL --- analysis of protein helices
 :Year: 2020
 :Copyright: GNU Public License v3
 
-.. versionadded:: 1.0.0
+.. versionadded:: 2.0.0
 
 This module contains code to analyse protein helices using the
 HELANAL_ algorithm
@@ -63,7 +63,9 @@ You can pass in a single selection::
     helanal = hel.HELANAL(u, select='name CA and resnum 161-187')
     helanal.run()
 
-    print(helanal.summary)
+All computed properties are available in ``.results``::
+
+    print(helanal.results.summary)
 
 Alternatively, you can analyse several helices at once by passing
 in multiple selection strings::
@@ -309,42 +311,42 @@ class HELANAL(AnalysisBase):
 
     Attributes
     ----------
-    local_twists : array or list of arrays
+    results.local_twists : array or list of arrays
         The local twist angle from atom i+1 to i+2.
         Each array has shape (n_frames, n_residues-3)
-    local_nres_per_turn : array or list of arrays
+    results.local_nres_per_turn : array or list of arrays
         Number of residues per turn, based on local_twist.
         Each array has shape (n_frames, n_residues-3)
-    local_axes : array or list of arrays
+    results.local_axes : array or list of arrays
         The length-wise helix axis of the local window.
         Each array has shape (n_frames, n_residues-3, 3)
-    local_heights : array or list of arrays
+    results.local_heights : array or list of arrays
         The rise of each local helix.
         Each array has shape (n_frames, n_residues-3)
-    local_helix_directions : array or list of arrays
+    results.local_helix_directions : array or list of arrays
         The unit vector from each local origin to atom i+1.
         Each array has shape (n_frames, n_residues-2, 3)
-    local_origins :array or list of arrays
+    results.local_origins :array or list of arrays
         The projected origin for each helix.
         Each array has shape (n_frames, n_residues-2, 3)
-    local_screw_angles : array or list of arrays
+    results.local_screw_angles : array or list of arrays
         The local screw angle for each helix.
         Each array has shape (n_frames, n_residues-2)
-    local_bends : array or list of arrays
+    results.local_bends : array or list of arrays
         The angles between local helix axes, 3 windows apart.
         Each array has shape (n_frames, n_residues-6)
-    all_bends : array or list of arrays
+    results.all_bends : array or list of arrays
         The angles between local helix axes.
         Each array has shape (n_frames, n_residues-3, n_residues-3)
-    global_axis : array or list of arrays
+    results.global_axis : array or list of arrays
         The length-wise axis for the overall helix. This points at
         the first helix window in the helix, so it runs opposite to
         the direction of the residue numbers.
         Each array has shape (n_frames, 3)
-    global_tilts : array or list of arrays
+    results.global_tilts : array or list of arrays
         The angle between the global axis and the reference axis.
         Each array has shape (n_frames,)
-    summary : dict or list of dicts
+    results.summary : dict or list of dicts
         Summary of stats for each property: the mean, the sample
         standard deviation, and the mean absolute deviation.
     """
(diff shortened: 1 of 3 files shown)


package/MDAnalysis/analysis/helix_analysis.py
```diff
@@ -418,49 +420,49 @@ def _prepare(self):
         for key, dims in self.attr_shapes.items():
             empty = [self._zeros_per_frame(
                 dims, n_positions=n) for n in n_res]
-            setattr(self, key, empty)
+            setattr(self.results, key, empty)
```
orbeckst, line 423: fyi: Could now be written as
```suggestion
            self.results[key] = empty
```

package/MDAnalysis/analysis/helix_analysis.py
```diff
@@ -418,49 +420,49 @@ def _prepare(self):
...
+        self.results.global_axis = [self._zeros_per_frame((3,)) for n in n_res]
+        self.results.all_bends = [self._zeros_per_frame((n-3, n-3)) for n in n_res]
 
     def _single_frame(self):
         _f = self._frame_index
         for i, ag in enumerate(self.atomgroups):
             results = helix_analysis(ag.positions, ref_axis=self.ref_axis)
             for key, value in results.items():
-                attr = getattr(self, key)
+                attr = getattr(self.results, key)
```
orbeckst, line 433: fyi
```suggestion
                attr = self.results[key]
```

package/MDAnalysis/analysis/helix_analysis.py
```diff
@@ -418,49 +420,49 @@ def _prepare(self):
...
         global_attrs = ['global_axis', 'global_tilts', 'all_bends']
         attrnames = list(self.attr_shapes.keys()) + global_attrs
         # summarise
-        self.summary = []
+        self.results.summary = []
         for i in range(len(self.atomgroups)):
             stats = {}
             for name in attrnames:
-                attr = getattr(self, name)
+                attr = getattr(self.results, name)
```
orbeckst, line 453: fyi
```suggestion
                attr = self.results[name]
```

package/MDAnalysis/analysis/helix_analysis.py
```diff
@@ -418,49 +420,49 @@ def _prepare(self):
...
-            self.summary.append(stats)
+            self.results.summary.append(stats)
 
         # flatten?
         if len(self.atomgroups) == 1 and self._flatten:
             for name in attrnames + ['summary']:
-                attr = getattr(self, name)
-                setattr(self, name, attr[0])
+                attr = getattr(self.results, name)
+                setattr(self.results, name, attr[0])
```
orbeckst, line 465:
```suggestion
                self.results[name] = attr[0]
```

package/MDAnalysis/analysis/helix_analysis.py
```diff
@@ -418,49 +420,49 @@ def _prepare(self):
...
                                'abs_dev': dev.mean(axis=0)}
-            self.summary.append(stats)
+            self.results.summary.append(stats)
 
         # flatten?
         if len(self.atomgroups) == 1 and self._flatten:
             for name in attrnames + ['summary']:
-                attr = getattr(self, name)
-                setattr(self, name, attr[0])
+                attr = getattr(self.results, name)
```
orbeckst, line 464:
```suggestion
                attr = self.results[name]
```

testsuite/MDAnalysisTests/analysis/test_helix_analysis.py
```diff
@@ -368,11 +368,11 @@ def test_regression_values(self):
 
         for key, value in HELANAL_SINGLE_DATA.items():
             if 'summary' in key:
-                data = getattr(ha, key.split()[0])
+                data = getattr(ha.results, key.split()[0])
```
orbeckst, line 371:
```suggestion
                data = ha.results[key.split()[0]]
```

testsuite/MDAnalysisTests/analysis/test_helix_analysis.py
```diff
@@ -368,11 +368,11 @@ def test_regression_values(self):
...
                 calculated = [data.mean(), data.std(ddof=1),
                               np.fabs(data-data.mean()).mean()]
             else:
-                calculated = getattr(ha, key)[0]
+                calculated = getattr(ha.results, key)[0]
```
orbeckst, line 375:
```suggestion
                calculated = ha.results[key][0]
```

orbeckst (approved):
lgtm, just suggestions how to make use of the fact that Results is a dict and get rid of ugly {get,set}attr.

orbeckst:
Something is not right in how the results are stored now. I am not sure what, I just looked at the failures and they are typically an extra (or missing) level of `[ .. ]` in the results. Maybe I made a mistake in my quick attr <-> dict translation.

package/MDAnalysis/analysis/helix_analysis.py
```diff
@@ -461,8 +461,8 @@ def _conclude(self):
         # flatten?
         if len(self.atomgroups) == 1 and self._flatten:
             for name in attrnames + ['summary']:
-                attr = getattr(self.results, name)
-                setattr(self.results, name, attr[0])
+                attr = self.results[name]
+                self.results[name] = attr[0]
```
orbeckst, line 465: Is this really
```python
self.results[name] = self.results[name][0]
```
?

lilyminium: It is. I think this is an issue with the Results class. The below fails in Results.

```python
def __setitem__(self, key, item):
        self._validate_key(key)
        super().__setitem__(key, item)
        assert self[key] is getattr(self, key)
```

lilyminium: See [this falsifying test](https://github.com/MDAnalysis/mdanalysis/pull/3281/commits/cb2c34edfebbc079ac530515474fe51dfd3fa44e) for an example.

orbeckst:
I assume you need PR #3281 merged before restarting this one?

orbeckst (changes requested):
I'll block the PR until the tests pass (which requires solving #3282 )

PR #3281, opened 2021-05-07
alias __setattr__ to __setitem__ in Results

### LINKED ISSUE #3282: Current Results class has a conflict in key validation between setattr and setitem
## Expected behavior ##

I can modify the *same value* in a Results class both by using `__setitem__` and `__setattr__`. e.g.

```python
def test_setattr_modify_item(self, results):
        setattr(results, "myattr", 0)
        assert results.myattr == 0
        results["myattr"] = 3
        assert results.myattr == 3
```

## Actual behavior ##

The [above test fails](https://github.com/MDAnalysis/mdanalysis/pull/3281/commits/cb2c34edfebbc079ac530515474fe51dfd3fa44e). This only happens when actions follow a specific order:

- modify a value with setattr
- then modify that value with setitem
- then try to get back the modified value with getattr/ibute

This is because in the current Results code, `__setattr__` actually sets the attribute. When this happens, attribute access happens with `__getattribute__` rather than `__getattr__`. This means that while `__setitem__` is modifying the key/value pair in `Results.data`, there is also an attribute that *does not* get modified. This attribute is what gets returned on attribute access.

While it may be easy to just tell users not to mix-and-match the setattr/setitem behaviour, it's apparently natural enough that it's causing the helix_analysis tests to fail (https://github.com/MDAnalysis/mdanalysis/pull/3279).

``` python
import MDAnalysis as mda

u = mda.Universe(PSF, DCD)

....

```

- Which operating system?

lilyminium:
Fixes #3282

 - 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/analysis/base.py
+++ b/package/MDAnalysis/analysis/base.py
@@ -111,13 +111,6 @@ class Results(UserDict):
             raise AttributeError("'Results' object has no "
                                  f"attribute '{attr}'") from err
 
-    def __delattr__(self, attr):
-        try:
-            del self[attr]
-        except KeyError as err:
-            raise AttributeError("'Results' object has no "
-                                 f"attribute '{attr}'") from err
-
 
 class AnalysisBase(object):
     r"""Base class for defining multi-frame analysis
--- a/package/MDAnalysis/analysis/base.py
+++ b/package/MDAnalysis/analysis/base.py
@@ -94,7 +94,8 @@ class Results(UserDict):
         elif isinstance(key, str) and not key.isidentifier():
             raise ValueError(f"'{key}' is not a valid attribute")
 
-    def __init__(self, **kwargs):
+    def __init__(self, *args, **kwargs):
+        kwargs = dict(*args, **kwargs)
         if "data" in kwargs.keys():
             raise AttributeError(f"'data' is a protected dictionary attribute")
         self.__dict__["data"] = {}
--- a/package/MDAnalysis/analysis/base.py
+++ b/package/MDAnalysis/analysis/base.py
@@ -111,6 +111,13 @@ class Results(UserDict):
             raise AttributeError("'Results' object has no "
                                  f"attribute '{attr}'") from err
 
+    def __delattr__(self, attr):
+        try:
+            del self[attr]
+        except KeyError as err:
+            raise AttributeError("'Results' object has no "
+                                 f"attribute '{attr}'") from err
+
 
 class AnalysisBase(object):
     r"""Base class for defining multi-frame analysis
--- a/package/MDAnalysis/analysis/base.py
+++ b/package/MDAnalysis/analysis/base.py
@@ -31,7 +31,6 @@ from collections import UserDict
 import inspect
 import logging
 import itertools
-import functools
 
 import numpy as np
 from MDAnalysis import coordinates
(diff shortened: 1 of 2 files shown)


package/MDAnalysis/analysis/base.py
```diff
@@ -84,8 +84,11 @@ class in `scikit-learn`_.
 
     .. versionadded:: 2.0.0
     """
+
+    data = None
+
     def _validate_key(self, key):
-        if key in dir(UserDict) or (key == "data" and self._dict_frozen):
+        if key in dir(self):
```
PicoCentauri, line 91: This is nice now!

package/MDAnalysis/analysis/base.py
```diff
@@ -94,28 +97,25 @@ def _validate_key(self, key):
...
         if "data" in kwargs.keys():
             raise AttributeError(f"'data' is a protected dictionary attribute")
-
-        self._dict_frozen = False
-        for key in kwargs:
-            self._validate_key(key)
-        super().__init__(**kwargs)
-        self._dict_frozen = True
+        self.__dict__["data"] = {}
+        self.update(kwargs)
```
PicoCentauri, line 101: [`UserDict`](https://github.com/python/cpython/blob/7424105a02bbc209e032752636bae30ccb556402/Lib/collections/__init__.py#L1040-L1048) allow to initlize the instance with passing a dictionary. This does not work anymore. I suggest you copy the four lines 
```python
if dict is not None:
    self.update(dict)
if kwargs:
    self.update(kwargs)
```

from cpython, adjust the arguments, add a test and everything is fine.

PicoCentauri (changes requested):
This is a smart PR!

package/MDAnalysis/analysis/base.py
lilyminium: Are you happier with [my somewhat lazier solution? :) ](https://github.com/MDAnalysis/mdanalysis/pull/3281/commits/619d9374851c4b3c8e45f3a95f0181b080288285)

orbeckst (approved):
Looks good to me.

(I was debating if we should have a test showing that two instances do not interfere with each other but that seems un-necessary with the current implementation — it just makes use of the way that Python resolves class and instance attribute access.)

orbeckst:
@PicoCentauri are you happy with the changes?

@PicoCentauri @lilyminium please merge when ready.

package/MDAnalysis/analysis/base.py
```diff
@@ -84,38 +84,39 @@ class in `scikit-learn`_.
 
     .. versionadded:: 2.0.0
     """
+
+    data = None
+
```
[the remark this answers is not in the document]
lilyminium: Woops, that's actually an unnecessary remnant of earlier code. Writing into `__dict__` is about the hackiest, most guaranteed way to make sure you're changing the object instance without calling any dunder methods.

lilyminium:
Should I update CHANGELOG? The Results PR itself was very recent.

package/MDAnalysis/analysis/base.py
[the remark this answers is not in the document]
orbeckst: I thought you needed this for the check on `__init__` that no existing attrs were included. But given that you add `data` to `__dict__` before anything else happens, this is not needed.

orbeckst:
I'd say no — normally we update CHANGELOG for anything that has an associated issue. But in this case this is a hotfix and the CHANGELOG entry would be more confusing than anything else.

cbouy (approved):
Sorry didn't get much time but all looks good!

package/MDAnalysis/analysis/base.py
PicoCentauri: Totally!

PR #3292, opened 2021-05-07
Use Results in Hole2

### LINKED ISSUE #3269: update hole2 analysis to use .results
Follow the new AnalysisBase API and store results in HoleAnalysis in `.results` instead of class attributes.

This should be done for 2.0.0 #3254 so that we don't have to start this new analysis module with deprecations in place.

See omnibus issue #3261 for background and details. See PR #3263 as an example (DensityAnalysis) but note that _no deprecations are necessary_ — just change where the results are stored and update the docs.

- [ ] use attributes in`self.results` instead of `self`
- [ ] update docs

lilyminium:
Fixes #3269 

 - Uses Results
 - Tidies some code (e.g. self.frames is now part of AnalysisBase)

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/analysis/hole2/hole.py
```diff
@@ -513,14 +517,49 @@ class HoleAnalysis(AnalysisBase):
...
+        Alias of :attr:`results.profiles`
+
+        .. deprecated:: 2.0.0
+            This will be removed in MDAnalysis 3.0.0. Please use
+            :attr:`results.profiles` instead.
 
     .. versionadded:: 1.0
 
+
+
```
PicoCentauri, line 562:
```python
    .. versionchanged:: 2.0.0
        :attr:`sphpdbs`, :attr:`outfiles` and :attr:`profiles `
        are now stored in a :class:`MDAnalysis.analysis.base.Results` 
        instance.
```

PicoCentauri (changes requested):
Just one minor doc addition.

package/MDAnalysis/analysis/hole2/hole.py
orbeckst: please add the versionchanged

package/CHANGELOG
```diff
@@ -175,6 +175,8 @@ Enhancements
     checking if it can be used in parallel analysis. (Issue #2996, PR #2950)
 
 Changes
+  * `analysis.hole2.HoleAnalysis` now stores ``sphpdbs``, ``outfiles``
```
orbeckst, line 178: also needs an entry in Deprecations

orbeckst (changes requested):
only docs (what @PicoCentauri noted) and CHANGELOG

orbeckst (approved):

orbeckst:
Waiting for CI... for CHANGELOG :-p.

(And depending on which other of the results PRs goes in, we're going to do it again... unless this one is first.)

PR #3293, opened 2021-05-07
Use read_direct() in H5MDReader

### LINKED ISSUE #2924: Optimizing H5MDReader's use of dataset indexing and __getitem__()

Following some benchmarks we've done on `H5MDReader`'s serial performance vs other file formats, we found that it lagged behind. Here's a graph:

A line profiler elucidated that the largest contributor to H5MD's slow I/O speeds was the `__getitem__()` method in dataset.py:
h5mdprofile

which makes things a little tricky because `H5MDReader` indexes a lot of datasets. Every time a frame is called, the positions, velocity, forces, box dimensions, and any datasets in the 'observables' group are indexed, which will call `__getitem__()`

1. Trim down some of the superfluous indexing in `H5MDReader`
2. Maybe give the user a prompt to suggest not to load velocity and force data if they're not using it, which would save a lot of I/O time

None for now, will update

## Additional context ##

Here's a link to the report from my REU where the results come from: https://figshare.com/articles/journal_contribution/REU_Report_pdf/12861215

edisj:
 - use the `h5py.Dataset.read_direct()` method instead of using NumPy slicing to read the data from the dataset. 

The `read_direct()` method is generally faster, as the NumPy style slicing creates an intermediate array that is discarded during the I/O process. 

I ran some benchmarks on a 113GB, ~90k frame trajectory file on ASU's HPC cluster Agave and found that using `read_direct` performed better than the slicing in the frame-chunked and contiguous formats:

image

The above plot also shows that the slow serial I/O times were caused by the chunk layout of the file on disk, where the "H5MD_default" bar refers to the file that was chunked with h5py's auto-chunking algorithm. "H5MD_chunked" refers to a custom chunk layout I chose when writing the file, where the chunk shape was (1, n_atoms, 3), and "H5MD_contiguous" refers to the file with no chunking applied.

EDIT: 
I also compared the H5MDReader with a properly chunked file compared to other file formats. For the benchmark, the I/O consisted of reading a 90k frame trajectory and computing the RMSD of the alpha carbons at every frame. The plot below shows the mean I/O time across 10 repeats. For the "H5MD_chunked" bar, 5 of the runs were with `read_direct()`, and 5 were with the old numpy style slicing. If you're interested, all the benchmark scripts can be found here https://github.com/edisj/Agave/blob/main/benchmarks/serial/scripts/benchmark.py

 
image

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG?


edisj:
I still need to change the `_get_frame_dataset()` method before review

orbeckst:
Then switch your PR to Draft.

orbeckst:
@edisj for the second graph (comparison between h5md and other formats) it is not clear why there are two bars for DCD, XTC, TRR. Either explain (edit your text) or redraw the figure.

edisj:
If all tests pass, this is ready for review :D  @orbeckst or anyone else passing by

orbeckst:
Sorry @edisj this might have to sit for a while — various deadlines and pretty much everything here is trying to get the 2.0.0 beta out.

orbeckst:
Thanks for updating the figures; very promising that h5md is almost as fast as DCD. @yuxuanzhuang might be interested as he's also been looking at parallel performance.

orbeckst:
Can you please merge with the latest develop to make sure it works with the recent changes?

It’s otherwise straightforward and looks good.

orbeckst (approved):
LGTM; this is an internal change that does not really need a CHANGELOG entry.

edisj:
Sounds good, thanks for the review

orbeckst (changes requested):
Sorry, have to reverse. You are addressing an issue so please add an entry to  CHANGELOG. This can go under Enhancements.

orbeckst:
@IAlibay can this PR me merged once the CHANGELOG entry is added or do you want to hold off? It's not user-visible and purely performance improvement without other side-effects known.

edisj:
@orbeckst CHANGELOG updated

orbeckst (approved):

orbeckst:
Thank you @edisj , I am happy with it.

I am assigning the PR to @IAlibay and he can merge it when it fits into the schedule.

IAlibay:
> @IAlibay can this PR me merged once the CHANGELOG entry is added or do you want to hold off? It's not user-visible and purely performance improvement without other side-effects known.

Yeah I don't see any problems with merging now.

PR #3185, opened 2021-03-21
Issue 2327 - Warn of upcoming scale_factor changes

IAlibay:
Towards #2327 

Changes made in this Pull Request:
 - Adds FutureWarning & docstring related to the upcoming changes to the way in which we write `scale_factor` entries with the NCDFWriter.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?
 - [x] Issue raised/referenced?

PR #3294, opened 2021-05-08
Add scale_factor write support for NetCDF writer

### LINKED ISSUE #2327: Improving the NCDFWriter behaviour (units + scale_factors)

IAlibay:
Fixes #2327 completes #3185 

I doubt this will make it into the beta release (unless someone feels like reviewing it tonight), but it should make it into the final release as it was a promised changed in 1.1.1.

 - Adds `scale_factor` writing support to NCDFWriter.
 - Changes the default `scale_factor` value for velocity writing to 20.455 in order to match the AMBER default

From ambertools' AmberNetdf.F90
```FORTRAN
    double precision, parameter :: velocityScale = 20.455d0
```
 - Cleans up the NCDFReader code a bit (still could be improved, but at least now it's more readable).

For context, here is the AMBER NetCDF convention: http://ambermd.org/netcdf/nctraj.xhtml

PR Checklist
------------
 - [x] Tests? -- finishing
 - [x] Docs?
 - [x] CHANGELOG updated? -- to do


package/MDAnalysis/coordinates/TRJ.py
```diff
@@ -731,18 +745,38 @@ def Writer(self, filename, **kwargs):
...
         """
         n_atoms = kwargs.pop('n_atoms', self.n_atoms)
         kwargs.setdefault('remarks', self.remarks)
-        kwargs.setdefault('dt', self.dt)
+        kwargs.setdefault('convert_units', self.convert_units)
+        kwargs.setdefault('velocities', self.has_velocities)
+        kwargs.setdefault('forces', self.has_forces)
+        for key in self.scale_factors:
+            kwargs.setdefault(f'scale_{key}', self.scale_factors[key])
         return NCDFWriter(filename, n_atoms, **kwargs)
```
lilyminium, line 780: What's the reasoning for setdefault instead of sticking it in the function signature?

IAlibay: Might be misunderstanding what you mean here, but the setdefault is there to allow users to override the NCDF file defaults (i.e. if you don't pass anything it'll use whatever defaults it had on reading the NCDF file, but you can override those by passing stuff to `Write`)

IAlibay:
How odd, I can't replicate this locally, https://github.com/MDAnalysis/mdanalysis/pull/3294/checks?check_run_id=2535134574#step:10:548

lilyminium (changes requested):
Waiting on new tests to finish running for diagnostic purposes before addressing those :)

package/MDAnalysis/coordinates/TRJ.py
```diff
@@ -1059,27 +1143,49 @@ def _write_next_timestep(self, ts):
             unitcell = self.convert_dimensions_to_unitcell(ts)
 
         # write step
-        self.trjfile.variables['coordinates'][self.curr_frame, :, :] = pos
-        self.trjfile.variables['time'][self.curr_frame] = time
+        # coordinates
+        self.trjfile.variables['coordinates'][
```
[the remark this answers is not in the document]
IAlibay: Everything to avoid pep8 🙃 I can clean it up though if it's unreadable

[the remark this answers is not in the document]
lilyminium: You could try https://github.com/MDAnalysis/mdanalysis/pull/3294#discussion_r628770930 ?

package/MDAnalysis/coordinates/TRJ.py
```diff
@@ -1029,6 +1102,13 @@ def _write_next_frame(self, ag):
 
         return self._write_next_timestep(ts)
 
+    @staticmethod
+    def _scale(variable, scale_factor):
+        if scale_factor is None or scale_factor == 1:
+            return variable
+        else:
+            return variable / scale_factor
+
```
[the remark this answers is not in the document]
IAlibay: Due to `mmap` I _think_ reassigning `value` is problematic. I can look into doing the assignment in the function it really makes things clearer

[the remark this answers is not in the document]
lilyminium: Does assigning `values` to a different variable name inside the function get around that?

package/MDAnalysis/coordinates/TRJ.py
```diff
@@ -937,18 +994,25 @@ def _init_netcdf(self, periodic=True):
         coords = ncfile.createVariable('coordinates', 'f4',
                                        ('frame', 'atom', 'spatial'))
         setattr(coords, 'units', 'angstrom')
+        if self.scale_factors['coordinates']:
+            setattr(coords, 'scale_factor', self.scale_factors['coordinates'])
```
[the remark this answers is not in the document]
IAlibay: I was under the impression that it did for netcdf files (maybe it does and I've forgotten what exactly), but it seems to work either way. I'd prefer not rock the boat here though. If we move from setattr to direct attribute assignment we should do it for everything and at that point we might as well refactor everything.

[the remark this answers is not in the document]
lilyminium: I guess I'd assume that both `x.y = 3` and `setattr(x, y, 3)` call `x.__setattr__(y, 3)` anyway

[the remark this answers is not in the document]
IAlibay: 🤷🏽 can refactor to a single style later (either way the whole thing needs a bigger cleanup, probably when I get around to doing NCRST reading)

package/MDAnalysis/coordinates/TRJ.py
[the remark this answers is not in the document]
IAlibay: I can't remember the exact details of `mmap` but iirc the TLDR was "the least amount of things that reference the underlying python object the better".

Can just add an extra line and not having to worry about it though 🙃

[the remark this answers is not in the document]
IAlibay: Sorry my bad, I messed up, this is write not read (probably should have named the functions differently...).

Modifying `value` will modify things like the timestep positions inplace (if self.convert_units is False), it's not an issue since it'll get rewritten once you move away from that timestep, but the original intent was to avoid doing any inplace modification if possible (see `convert_units` stuff).

Anyways, still doable, just needs that extra `else` 🙃

IAlibay:
Ok complication number 1:

`maskandscale`: "whether or not to do automatic conversion based on the value of `scale_factor`

In `scipy` this is False by default, in `netcdf4` this is True by default....

So as far as I can tell when we fixed reading (because it's scipy only reading), then we were fine (i.e. auto scaling was off so we just applied scaling ourselves). But with writing, if we write with netcdf4 then it's on, if we write without it then it's off... unless we enforce it to be off....

IAlibay:
So one way to make the NetCDF4 and scipy default behaviours the same is just to enable `maskandscale` when creating the scipy netcdf file.

@tylerjereddy do you happen to have any insights here on a) performance overheads of using numpy masked arrays, b) potential issues in using masked arrays (i.e. can it lead to the unintentional hiding of corrupted netcdf files).

IAlibay:
@MDAnalysis/coredevs this is the last thing we definitely promised in 2.0, can we try to get this reviewed/merged as a priority so we can finally release 2.0?

package/MDAnalysis/coordinates/TRJ.py
```diff
@@ -411,7 +412,6 @@ class NCDFReader(base.ReaderBase):
     Current limitations:
 
     * only trajectories with time in ps and lengths in Angstroem are processed
-    * scale_factors are supported on read but are currently not kept/used when
       writing
```
orbeckst, line 415: Delete the "writing", too
```suggestion
```

orbeckst (commented):
reviewed first half, looking good so far

package/MDAnalysis/coordinates/TRJ.py
```diff
@@ -768,18 +809,24 @@ class NCDFWriter(base.WriterBase):
         name of output file
     n_atoms : int
         number of atoms in trajectory file
-    start : int (optional)
```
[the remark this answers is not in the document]
orbeckst: https://docs.mdanalysis.org/2.0.0-dev0/documentation_pages/coordinates/init.html#trajectory-writer-class seems to say that start/stop/step should be available. Not sure if this still makes sense (you know, after >5 years when the trajectory API was probably updated last time).

WriterBase https://github.com/MDAnalysis/mdanalysis/blob/55ffd255013dac427be3a8f4cf7962a4a886b712/package/MDAnalysis/coordinates/base.py#L2204 is not clear about what it requires, the API docs are really what counts.

This would be a good time to clean up the trajectory API, too, at least the Writer part.

package/MDAnalysis/coordinates/TRJ.py
```diff
@@ -1130,6 +1228,11 @@ class NCDFPicklable(scipy.io.netcdf.netcdf_file):
     mmap : None or bool, optional
         Whether to mmap `filename` when reading. True when `filename`
         is a file name, False when `filename` is a file-like object.
+    version : {1, 2}, optional
```
orbeckst, line 1231: Not clear what this is and why it's needed. Link to other docs available?

orbeckst: The sentence is ambiguous: Could be read as if it is putting "netcdf file version" into "read/write" mode.

testsuite/MDAnalysisTests/coordinates/test_netcdf.py
```diff
@@ -876,6 +887,133 @@ def test_write_u(self, pos, vel, force, tmpdir, u1, u2):
         u.trajectory.close()
 
 
+class TestNCDFWriterScaleFactors:
+    """Class to check that the NCDF Writer properly handles the
+    application of scale factors"""
+
+    @pytest.fixture()
+    def outfile(self, tmpdir):
+        return str(tmpdir) + 'ncdf-write-scale.ncdf'
```
orbeckst, line 896: Doesn't
```python
return str(tmpdir / 'beetroots.ncdf')
```
also work?

testsuite/MDAnalysisTests/coordinates/test_netcdf.py
```diff
@@ -876,6 +887,133 @@ def test_write_u(self, pos, vel, force, tmpdir, u1, u2):
...
+        return str(tmpdir) + 'ncdf-write-scale2.ncdf'
+
+    @pytest.fixture()
+    def universe(self):
+        return mda.Universe(PRM_NCBOX, TRJ_NCBOX)
+
+    def get_scale_factors(self, ncdfile):
+        """Get a dictionary of scale factors stored in netcdf file"""
+        sfactors = {}
+        with netcdf.netcdf_file(ncdfile, mmap=False) as f:
```
orbeckst, line 909: Just curious, why is `mmap=False` needed?

testsuite/MDAnalysisTests/coordinates/test_netcdf.py
```diff
@@ -876,6 +887,133 @@ def test_write_u(self, pos, vel, force, tmpdir, u1, u2):
...
+            assert_almost_equal(universe.atoms.velocities,
+                                universe3.atoms.velocities, 4)
+            assert_almost_equal(universe.atoms.forces,
+                                universe3.atoms.forces, 4)
+
+
+class TestScipyScaleFactors(TestNCDFWriterScaleFactors):
+    """As above, but netCDF4 is disabled since scaleandmask is different
+    between the two libraries"""
+    @pytest.fixture(autouse=True)
```
orbeckst, line 1006: So much magic in pytest!

package/MDAnalysis/coordinates/TRJ.py
```diff
@@ -1130,6 +1228,11 @@ class NCDFPicklable(scipy.io.netcdf.netcdf_file):
...
+        Sets the netcdf file version to read / write. 1 is classic, 2 is
+        65-bit offset format. Default is 1.
+    maskandscale : bool, optional
```
orbeckst, line 1234: Why, what? Links ... and when would one want to use it?

Are `version` and `maskandscale` options that users should play with?

orbeckst (approved):
This looks sensible to me. Kudos for writing these tests!

I have some minor comments regarding docs — I am missing some knowledge about netcdf details. Assuming that I am somewhat representing an average user here, the question is if the available options should carry warning signs that mere mortals should leave them at defaults. Links to other docs would also be helpful.

As I mentioned in a comment, we should remove a whole bunch of stuff from the [API docs for Writers](https://docs.mdanalysis.org/2.0.0-dev0/documentation_pages/coordinates/init.html#trajectory-writer-class), namely the `start`, `stop`, `step` and `dt` kwargs/attributes should not be required. That's another issue, though.

testsuite/MDAnalysisTests/coordinates/test_netcdf.py
IAlibay: Yeah I can just use f-strings too, I can update (I think I just copy pasted from some fixture in another class somewhat 😅 )

package/MDAnalysis/coordinates/TRJ.py
IAlibay: So this class essentially is just a wrapper around https://docs.scipy.org/doc/scipy/reference/generated/scipy.io.netcdf_file.html with a couple of extra methods to make it pickle-able. I just added the extra options because a) maskandscale needed to be propagated (although it's False by default currently I wanted to make sure we could be explicit here in case it changed in the future to match netCDF4's True-by-default), b) there might be some obscure use case where someone will want to use this for something else and will want the ability to use the other parameters.

Rather than detailing things too much (and also at the risk of just ending up copying the scipy docs verbatim), should we just link to https://docs.scipy.org/doc/scipy/reference/generated/scipy.io.netcdf_file.html ?

package/MDAnalysis/coordinates/TRJ.py
IAlibay: "Sets the netcdf version. 1 is classic, 2 is the 64-bit offset format"? (also I notice the 65 bit typo)

It's less descriptive but as mentioned in the other comment we can link to the scipy docs for more information? (not that they are more descriptive with this parameter...)

testsuite/MDAnalysisTests/coordinates/test_netcdf.py
IAlibay: Honestly the main reason here is "an abundance of caution", I have no idea how the mmap implementation this plays with pytest-xdist so I thought it best to avoid it where possible.

The performance increase is somewhat minimal (given how slow the other parts are), so we're looking at ~ 100 microsecs lost, disabling mmap (at least on my machine it goes from 800 microsec to 900).

I'd probably prefer just being cautious here, but can turn it back on if you don't agree (I'm happy to drop a comment above saying I'm probably being unreasonably cautious 🙃 ).

package/MDAnalysis/coordinates/TRJ.py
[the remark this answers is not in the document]
IAlibay: So looking at the other main writers (DCD, XDR, LAMMPS), it doesn't seem like there is anything that uses start/step/delta (or even has it as a parameter for `_init_`). The closest we have is `step` and `dt` for the DCDWriter, although worth nothing that it's not immediately clear to me that `step` actually does anything!

It's possible I'm missing something, but since it doesn't look like other writers don't also meet the API's requirements to have these (either as parameters to init or as class attributes), I'd suggest we just clean up the API docs in a separate PR instead?

testsuite/MDAnalysisTests/coordinates/test_netcdf.py
orbeckst: _Maybe_ add a comment anyone else looking at it and wondering. I am not worry about the performance, more along the lines of using tests as implementation examples and then not being sure about why something is done.

testsuite/MDAnalysisTests/coordinates/test_netcdf.py
orbeckst: `tmpdir` is actually py.path object, not pathlib so I am not even sure if it actually overloads `/`. It might only use `tmpdir.join("cabbage.ncdf")`.

You know what, just leave it. Eventually we get MDA to read path-like objects and then we rewrite it anyway.

package/MDAnalysis/coordinates/TRJ.py
orbeckst: Linking to scipy docs is good.

Should we even allow setting the format? Doesn't ncdf always use version=2? According to http://ambermd.org/netcdf/nctraj.xhtml

> **Trajectory/restart files shall use 64 bit offsets**
> This can be accomplished by setting a flag during file creation; refer to API docs for details.

package/MDAnalysis/coordinates/TRJ.py
orbeckst: Link, but also say that for normal ncdf behavior it should be left at the default

package/MDAnalysis/coordinates/TRJ.py
IAlibay: So interestingly we don't set version on read, but we do on write. I need to double check if it defaults to 2.... otherwise we got a potential bug on our hands.

package/MDAnalysis/coordinates/TRJ.py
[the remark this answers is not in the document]
orbeckst: yes, separate PR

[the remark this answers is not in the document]
orbeckst: See #3392 3392

package/MDAnalysis/coordinates/TRJ.py
IAlibay: Ok yeah so the version_byte is just picked up from the file on read, so it _shouldn't_ be an issue that we don't set the flag when reading. Either way we throw an error if the version_byte is wrong, so that's fine.

testsuite/MDAnalysisTests/coordinates/test_netcdf.py
IAlibay: ok, I'll leave it

IAlibay:
That should have address the review comments @orbeckst - let me know what you think

@lilyminium do you want to re-review and/or are you ok with your review being dismissed?

package/MDAnalysis/coordinates/TRJ.py
```diff
@@ -1123,13 +1220,25 @@ class NCDFPicklable(scipy.io.netcdf.netcdf_file):
...
     Parameters
     ----------
     filename : str or file-like
         a filename given a text or byte string.
     mmap : None or bool, optional
         Whether to mmap `filename` when reading. True when `filename`
         is a file name, False when `filename` is a file-like object.
+    version : {1, 2}, optional
+        Sets the netcdf file version to read / write. 1 is classic, 2 is
+        65-bit offset format. Default is 1.
```
orbeckst, line 1239:
```suggestion
        64-bit offset format. Default is 1 (but note AMBER ncdf *requires* 2).
```

Or change the default to 2, it's AMBER reader after all.

orbeckst (approved):
The 65 bit -> 64 bit fix still needs to be added. @IAlibay you can use my suggestion or do it yourself. I won't hold up proceedings. Thanks for addressing my comments.

package/MDAnalysis/coordinates/TRJ.py
IAlibay: Oops missed that thanks!. Personally I prefer we keep true to the original scipy class - just in case we end up getting confused in the future (there are other MD engines / workflows that use netcdf out there [see Yank] that don't necessarily stick to the way AMBER does things [not checked that version_byte yank uses]).

lilyminium (approved):

lilyminium:
Sorry -- LGTM

PR #3285, opened 2021-05-07
Use Results class for PCA Analysis

PicoCentauri:
Fixes #3275 

Changes made in this Pull Request:
Move the results of the PCA analysis to the new results class

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?
 - [x] Issue raised/referenced?

PR #3296, opened 2021-05-09
Allow PCA to run with given reference mean values

### LINKED ISSUE #2728: PCA test sometimes fails...
I get [this test](https://github.com/MDAnalysis/mdanalysis/blob/develop/testsuite/MDAnalysisTests/analysis/test_pca.py#L166) failing locally when I run in parallel:

```
=============================================================== FAILURES ===============================================================
___________________________________________________________ test_given_mean ____________________________________________________________
[gw0] linux -- Python 3.6.10 /home/richard/miniconda3/envs/mda/bin/python

pca_aligned = <MDAnalysis.analysis.pca.PCA object at 0x7f7cda848588>, u = <Universe with 3341 atoms>

    def test_given_mean(pca_aligned, u):
        pca = PCA(u, select=SELECTION, align=False,
                  mean=pca_aligned._mean).run()
>       assert_almost_equal(pca_aligned.cov, pca.cov, decimal=5)
E       AssertionError: 
E       Arrays are not almost equal to 5 decimals
E       
E       Mismatched elements: 900 / 900 (100%)
E       Max absolute difference: 1.0380358
E       Max relative difference: 106.43334522
E        x: array([[ 6.85943e-02,  6.14962e-03, -2.85235e-02, -2.46964e-03,
E                4.93089e-03, -1.45766e-02,  6.77579e-03,  6.73560e-03,
E               -6.87083e-03, -1.39547e-02,  4.34639e-03, -1.46411e-02,...
E        y: array([[ 1.10663e+00, -4.76013e-01,  2.59657e-01,  8.86885e-01,
E               -4.62178e-01,  3.13851e-01,  7.73986e-01, -4.43201e-01,
E                2.00352e-01,  4.87414e-01, -3.32454e-01,  1.01439e-01,...

analysis/test_pca.py:169: AssertionError
```

This is when I run `pytest -n 12` from the MDAnalysisTests/ directory.  If I run `pytest -n 12 analysis/` it doesn't fail any more.  Varying core count seems to make it fail/not fail, so it feels like something unthreadsafe is happening?

I've tried reducing the scope of the fixture and this still fails.

fiona-naughton:
Fixes #2728

 - Fix shape of positions array from provided reference atomgroup so the PCA will actually run
 - Update the corresponding test to actually run with given mean values

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


testsuite/MDAnalysisTests/analysis/test_pca.py
```diff
@@ -162,11 +164,22 @@ def test_calculate_mean(pca_aligned, u, u_aligned):
 
 
 def test_given_mean(pca_aligned, u):
-    pca = PCA(u, select=SELECTION, align=False,
-              mean=pca_aligned._mean).run()
+    # create dummy atomgroup to populate with the mean values
+    num_atoms = len(u.select_atoms(SELECTION))
+    ag = mda.Universe.empty(num_atoms, trajectory=True).select_atoms('all')
+    ag.positions = pca_aligned.mean.reshape((num_atoms, 3))
```
IAlibay, line 170: according to the docs we should just be able to use `mean_atoms` back as the atomgroup with the mean positions right?

fiona-naughton: From the current docs, yes - but, per discussion in #3285, it seems that `mean_atoms` does not actually contain the mean positions. If we fix that, then we could just use `mean_atoms` here;  I decided to do it roundabout like this for now with a view of getting the more 'urgent' changes merged rather than debating exactly what `mean_atoms` should be doing (imo the doc changes in #3285 reflect what `mean_atoms` seem to actually be storing atm so it's not a big after that goes through). 

All that being said, there is still the option to fix `mean_atoms` here (or in 3285), if that seems better? (Personally, it feels like both `means` and `mean_atoms` would be better as position arrays rather than atomgroups, but I'm not super familiar with pca and how people use this part of the input/output so others should weigh in first)

IAlibay: I'd vote for just fixing it properly now (as a coordinate array) and not going through a deprecation cycle.

orbeckst: Works for me.

fiona-naughton:
`mean_atoms` has been removed and I switched `mean` to an array. I decided against allowing atomgroups to be passed in for now and depreciating, since part of the original error means that setting `mean` didn't work anyway, so that implies people weren't using it anyway (without patching the code themselves) - but I'm open to if others think it's better we do.

package/MDAnalysis/analysis/pca.py
```diff
@@ -217,10 +222,15 @@ def _prepare(self):
         self._n_atoms = self._atoms.n_atoms
 
         if self._mean is None:
-            self.mean = np.zeros(self._n_atoms*3)
+            self.mean = np.zeros((self._n_atoms,3))
```
orbeckst, line 225: pep 8: space after comma
```suggestion
            self.mean = np.zeros((self._n_atoms, 3))
```

testsuite/MDAnalysisTests/analysis/test_pca.py
```diff
@@ -151,20 +153,28 @@ def test_cosine_content():
...
+def test_given_mean(pca, u):
     pca = PCA(u, select=SELECTION, align=False,
-              mean=pca_aligned._mean).run()
-    assert_almost_equal(pca_aligned.cov, pca.cov, decimal=5)
+              mean=pca.mean).run()
+    assert_almost_equal(pca.cov, pca.cov, decimal=5)
+
+
+def test_wrong_num_given_mean(u):
+    wrong_mean = [[0,0,0],[1,1,1]]
```
orbeckst, line 174: pep 8
```suggestion
    wrong_mean = [[0,0,0], [1,1,1]]
```

orbeckst (approved):
Excellent!

(I'm just fixing 2 pep8 things)

orbeckst:
@IAlibay I'll merge this unless you object or CI is unhappy.

orbeckst:
Sigh, not sure what sphinx is unhappy about:
```
/home/runner/work/mdanalysis/mdanalysis/package/MDAnalysis/analysis/pca.py:docstring of MDAnalysis.analysis.pca.PCA.transform:5:Inline literal start-string without end-string.
```

package/MDAnalysis/analysis/pca.py
```diff
@@ -308,14 +316,15 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
             Iteration stops *before* this frame number, which means that the
             trajectory would be read until the end.
         step : int, optional
-            Number of frames to skip over for PCA transform. If set to ``None``
+            Include every `step` frames in the PCA transform. If set to ``None``
             (the default) then every frame is analyzed (i.e., same as
             ``step=1``).
```
IAlibay, line 321:
```suggestion
            Include every `step` frames in the PCA transform. If set to
            ``None`` (the default) then every frame is analyzed (i.e., same as
            ``step=1``).
```

I'm running out weird movie-based PEP8 messages to throw.

testsuite/MDAnalysisTests/analysis/test_pca.py
```diff
@@ -151,20 +153,28 @@ def test_cosine_content():
...
+def test_given_mean(pca, u):
     pca = PCA(u, select=SELECTION, align=False,
-              mean=pca_aligned._mean).run()
-    assert_almost_equal(pca_aligned.cov, pca.cov, decimal=5)
+              mean=pca.mean).run()
+    assert_almost_equal(pca.cov, pca.cov, decimal=5)
+
+
+def test_wrong_num_given_mean(u):
+    wrong_mean = [[0,0,0], [1,1,1]]
```
IAlibay, line 174:
```suggestion
    wrong_mean = [[0, 0, 0], [1, 1, 1]]
```
One does not simply PEP8 into M(ordor)DA?

testsuite/MDAnalysisTests/analysis/test_pca.py
```diff
@@ -151,20 +153,28 @@ def test_cosine_content():
...
+    with pytest.raises(ValueError) as exc:
+         pca = PCA(u, select=SELECTION, mean=wrong_mean).run()
+    assert 'Number of atoms in' in str(exc.value)
```
IAlibay, line 177:
```suggestion
    with pytest.raises(ValueError, match='Number of atoms in'):
        pca = PCA(u, select=SELECTION, mean=wrong_mean).run()
```
> Hello there
General Ken(pep8)-Obi!

(re-structuring because the indentations are don't match up and this is kinda neater)

IAlibay (approved):
Overall lgtm thanks @fiona-naughton, please do merge in the following PEP8 fixes

(I lack sleep, so now all PEP8 messages are movie meme-based).

orbeckst:
Thank you @fiona-naughton for fixing this one!

PR #3297, opened 2021-05-09
Unpin pytest

### LINKED ISSUE #3082: Unpin pytest on CI
Follow up from #3081 

`pytest` was pinned due to Windows OS failures, probably due to some upstream issues.

I'm opening this issue so that we remember to look into what happened here and aim to unpin as soon as we can.

IAlibay:
Fixes #3082

 - Changed the test_pdb_metal filterwarning to just error out if it sees a mass guessing warning

Can't replicate the azure failure locally, let's see what happens...

PR Checklist
------------
 - [x] Tests?
 - Docs? N/A
 - CHANGELOG updated? N/A

### DIFF
--- a/.github/workflows/gh-ci.yaml
+++ b/.github/workflows/gh-ci.yaml
@@ -14,7 +14,7 @@ defaults:
     shell: bash -l {0}
 
 env:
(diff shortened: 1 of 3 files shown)


IAlibay:
Seems like it just works now (not sure if that filterwarning fix just did the trick [it did on *nix, but I don't know if that was the same issue in Windows]).

PR #3299, opened 2021-05-13
WIP: Master merge

richardjgowers:
This merges the 1.x branch into develop again.

PR #3300, opened 2021-05-13
Add concurrency group

IAlibay:
<s>Fixes #</s>

Adds a concurrency group to the github action CI workflow which automatically kills workflows from the same PR that have already started working (see: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency).

This is quite useful, because we often have collaborative PRs were we get lots of reviewer commits (see #3299) or PRs where folks are trying things iteratively, and that tends to clog up CI. It also allows us to be a lot less wasteful, so we can feel a little bit less guilty about our carbon footprints :)

 - Adds concurrency group to CI

This _might_ fix me having to manually cancel action runs all the time...

PR Checklist
------------
 - Tests?
 - Docs?
 - CHANGELOG updated?
 - Issue raised/referenced?

### DIFF
--- a/.github/workflows/gh-ci.yaml
+++ b/.github/workflows/gh-ci.yaml
@@ -9,6 +9,10 @@ on:
       - develop
       - master
 
+concurrency: 
+  group: ${{ github.head_ref }}
+  cancel-in-progress: true
+
 defaults:
   run:
     shell: bash -l {0}
--- a/.github/workflows/gh-ci.yaml
+++ b/.github/workflows/gh-ci.yaml
@@ -9,7 +9,6 @@ on:
       - develop
       - master
 
-# comment to trigger CI
 concurrency: 
   group: ${{ github.head_ref }}
   cancel-in-progress: true
--- a/.github/workflows/gh-ci.yaml
+++ b/.github/workflows/gh-ci.yaml
@@ -9,6 +9,7 @@ on:
       - develop
       - master
 
+# comment to trigger CI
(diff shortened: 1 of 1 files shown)


IAlibay:
I was really trying to hold off on that 100th PR to make it some kind of meaningful thing, but I'll settle for saving time by not having to manually cancel workflows :)

orbeckst:
I think @richardjgowers 100th as a 1-line change. You'd be in good company.

.github/workflows/gh-ci.yaml
```diff
@@ -9,6 +9,10 @@ on:
       - develop
       - master
 
+concurrency: 
+  group: ${{ github.head_ref }}
+  cancel-in-progress: true
```
orbeckst, line 14: You directly followed https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-concurrency-to-cancel-any-in-progress-job-or-run so I assume this will work.

But I just want to check: this will only cancel jobs related to _this_ PR, not anything in other PRs, right?

IAlibay: Yup, just in case though I'll cycle the CI again just to make sure it doesn't kill the currently running H5MD workflow.

IAlibay: image

Seems to work fine :)

orbeckst (approved):
lgtm

Happy 💯 PR!

PR #3308, opened 2021-05-16
Consistent ` : ` in Analysis modules

PicoCentauri:
Adds a space before `:` for some docs and removed and unused parameter in 
helix_analysis


package/MDAnalysis/analysis/helix_analysis.py
```diff
@@ -377,10 +376,7 @@ def __init__(self, universe, select='name CA', ref_axis=[0, 0, 1],
             counter = 0
             if len(groups) > 1:
                 msg = 'Your selection {} has gaps in the residues.'.format(s)
-                if split_residue_sequences:
-                    msg += ' Splitting into {} helices.'.format(len(groups))
-                else:
-                    groups = [ag.resindices]
+                groups = [ag.resindices]
```
IAlibay, line 379: @lilyminium @fiona-naughton please review - I assume there was intent behind this?

orbeckst:
Is the additional space before the colon required for mdacli or is it just a "Schönheitskorrektur"?

Is it needed by sphinx/napoleon?

PicoCentauri:
> Is the additional space before the colon required for mdacli or is it just a "Schönheitskorrektur"?

Both :), mdacli parses a space a double dot and a space. I know this is not a perfect parser, but the easiest we come up with. I think this styling is also the preferred one from sphinx/napoleon...

orbeckst:
It is pretty difficult to ensure correct doc formatting in PRs. I am certainly in favor of making our docs consistent. But ultimately the mdacli parser should be able to deal with any syntactically correct reST.

On that note, given that sphinx/napoleon _do_ parse the doc strings in a semantically meaningful way, perhaps a future goal could be to use the same parsing functionality as this would ensure that you really get the parameters as documented.

package/MDAnalysis/analysis/helix_analysis.py
orbeckst: The tests fail because of this change.

PicoCentauri:
I totally agree. I checked the sphinx/napoleon repo and tried to cut out their parser but I failed. A proper parser is something on my todo list.

package/MDAnalysis/analysis/helix_analysis.py
PicoCentauri: I checked the code again and the `groups` definition makes sense here. Sorry for the confusion. I added a proper doc entry.

package/MDAnalysis/analysis/helix_analysis.py
```diff
@@ -306,6 +306,8 @@ class HELANAL(AnalysisBase):
         of best fit, and the local screw angles.
     flatten_single_helix : bool, optional
         Whether to flatten results if only one selection is passed.
+    split_residue_sequences : bool, optional
+        Wether to split residues into a list of same or consecutive helices.
```
IAlibay, line 310: I'm not sure this is suitably informative for what it does. We should mention that this only is of use when `len(groups) > 1`. An example in the module docstring would be great if possible.

PicoCentauri: Yes, this would be nice. However, I'm not familiar with the class... Maybe @lilyminium can help?

PicoCentauri: Yes, let's do it this way.

IAlibay (approved):
I'll merge, please do remember to open that issue @PicoCentauri

PR #3312, opened 2021-05-20
DL_POLY HISTORY may contain cells even for non-periodic simulations

### LINKED ISSUE #3314: DL_POLY HISTORY may contain cells even for non-periodic simulations
## Expected behavior ##

If the DL_POLY HISTORY file starts like:

```
Test molecule                        
         0         0      2520                  101               509446
timestep         0      2520 0 0            0.000700            0.000000
       63.9987726600        0.0000000000        0.0000000000            
        0.0000000000       60.0000030000        0.0000000000            
        0.0000000000        0.0000000000       63.1569481800           
CT               1     12.011500      0.115000      0.000000            
   -0.5783592591        0.4774311051         18.27086368                
CA               2     12.011500     -0.115000      0.000000            
    -1.263279574        0.8116030738         16.92369799                
CA               3     12.011500     -0.115000      0.000000            
```

running:

`u = mda.Universe('HISTORY', format='HISTORY')`

should produce a Universe with 2520 atoms. The cell parameters might differ between timeframes.

## Actual behavior ##

Raises:

`ValueError: invalid literal for int() with base 10: '0.0000000000'`.

This behaviour happens because `HistoryReader` in `coordinates/DLPoly.py` skips the cell dimensions only `if not imcom == 0`. `imcom` is the second integer in the second line of the file, i.e. zero. Such incom value implies no periodic boundary conditions (see DLPOLY manual). For this reason, it is reasonable for the HistoryReader to assume there are no cell dimensions and that the next line under the `timestep ...` line already defines an atom. So it assumes that the `       63.9987726600        0.0000000000        0.0000000000     ` line is atom definition and tries to read `0.0000000000` as atom index (integer).

Such assumption makes sense, as non-periodic simulations do not need a cell. However, if (for whatever reason) the simulation used an ensemble with constant volume or pressure, a cell is nonetheless generated and appended for each timestep.  Instead of using `imcom == 0` as a flag for cell parameters being printed, the parser should actively check whether cell parameters are present or not.

Can be reproduced with a HISTORY file containing:

```Test molecule                        
         0         0      3                  101               509446
timestep         0      3 0 0            0.000700            0.000000
       63.9987726600        0.0000000000        0.0000000000            
        0.0000000000       60.0000030000        0.0000000000            
        0.0000000000        0.0000000000       63.1569481800           
CT               1     12.011500      0.115000      0.000000            
   -0.5783592591        0.4774311051         18.27086368                
CA               2     12.011500     -0.115000      0.000000            
    -1.263279574        0.8116030738         16.92369799                
CA               3     12.011500     -0.115000      0.000000     ```

``` python
import MDAnalysis as mda

u = mda.Universe('HISTORY', format='HISTORY')

```

MDAnalysis 1.0.0, Python 3.7, Win 10 64-bit.

fiszczyp:
Fixes #3314 

DL_POLY HISTORY files can still have cell, even if the boundary conditions are switched off (imcom=0). 

 - Created a new attribute `_has_cell` (akin to `_has_velocity`, etc.) to use as a flag instead of `not imcom == 0` for DL_POLY trajectory HistoryReader.
 - Changed one `if not all(...)` to `if not np.all(...)` as I think it is deprecated and was throwing errors.
 - Not so relevant for the DL_POLY topology HistoryParser, but also removed the `not imcom == 0` check there and instead just checked once the line length corresponds to a correct atom entry to skip all the cell parameters.

PR Checklist
------------
 - [X] Tests?
 - [X] Docs?
 - [x] CHANGELOG updated?


richardjgowers (changes requested):
Hey @fiszczyp thanks for the fix. To include this we'll also need a test written, you might be able to alter one of the existing test files for this.

testsuite/MDAnalysisTests/datafiles.py
```diff
@@ -501,6 +501,7 @@
 DLP_HISTORY = resource_filename(__name__, "data/dlpoly/HISTORY")
 DLP_HISTORY_order = resource_filename(__name__, "data/dlpoly/HISTORY_order")
 DLP_HISTORY_minimal = resource_filename(__name__, "data/dlpoly/HISTORY_minimal")
+DLP_HISTORY_minimal_cell = resource_filename(__name__, "data/dlpoly/HISTORY_minimal_cell")
```
IAlibay, line 504: This also needs adding to the `__all__` list (see lines 39->209)

package/MDAnalysis/topology/DLPolyParser.py
```diff
@@ -129,16 +129,14 @@ def parse(self, **kwargs):
...
-                inf.readline()
-                inf.readline()
-                inf.readline()
 
             names = []
             ids = []
 
             line = inf.readline()
+            while not len(line.split()) == 5:
+                line = inf.readline()
```
richardjgowers, line 138: I think this has the ability to infinitely loop if a (non HISTORY) file never has 5 tokens on a line.  I think once we reach end-of-file `readline()` is going to return an empty string (i.e. not even a newline char), can we add a check for that and raise an error if we reach the end of file?

package/MDAnalysis/coordinates/DLPoly.py
```diff
@@ -216,31 +227,17 @@ def _read_frame(self, frame):
...
-        """Read the number of frames, and the offset for each frame
-
-        offset[i] - returns the offset in bytes to seek into the file to be
-                    just before the frame starts
-        """
-        offsets = self._offsets = []
+        # Second line is traj_key, imcom, n_atoms, n_frames, n_records
+        offsets = []
 
         with open(self.filename, 'r') as f:
```
richardjgowers, line 233: The reason for `self._n_frames` was to cache the number of frames so it's only calculated once.  This needs to be added back in.

richardjgowers (changes requested):
thanks for the test, couple of comments

testsuite/MDAnalysisTests/datafiles.py
fiszczyp: Thanks, missed that!

package/MDAnalysis/coordinates/DLPoly.py
fiszczyp: Actually, a question about the expected behaviour if a *wrong* HISTORY file is passed. DL_POLY HISTORY files have the `n_frames` added to the header line on top, so from the test files `DLP_HISTORY_minimal`:

```
DL_POLY: Minimal example of HISTORY
         0         0       3                    3                 2606
```

it can be parsed as:
*trajectory key:* `0` - only coordinates saved
*imcom:* `0` - non periodic
*n_atoms:* `3` - three atoms
*n_frames:* `3` - three frames
*n_records:* `2606` - number of records (typically lines, wrong here).

The `HistoryReader()` takes `n_atoms` from there. I also made it take `n_frames` from there, but it might be that someone just truncated the trajectory file and the information might be wrong. Should we instead just go through the file updating the number of frames as it reads? That is:

`n_frames = len(offsets)`.

I think it might be more bullet proof, as I can totally imagine when someone would remove some last/first 50 frames without updating a header, but not touch the number of atoms at all.

fiszczyp: Re caching, I don't think the `HistoryReader()` was doing that but it makes sense to add, thanks!

package/MDAnalysis/topology/DLPolyParser.py
fiszczyp: It now raises EOFError, thanks!

richardjgowers (approved):

richardjgowers:
thanks @fiszczyp can you also add yourself to AUTHORS and we can merge this

fiszczyp:
> thanks @fiszczyp can you also add yourself to AUTHORS and we can merge this

Done, thanks!

PR #3324, opened 2021-05-22
Passing arguments to the RDKitConverter from `select_atoms`

### LINKED ISSUE #3319: "smarts" selection keywork issue an advice I cannot follow if the structure has no hydrogens
## Expected behavior ##

When selecting atoms using the `smarts` keyword on a system that does not have hydrogen it either:

* just works
* tells me it cannot work
* tells me how to make it work

## Actual behavior ##

I get an error message that gives me an advice that I cannot follow:
```
AttributeError: No hydrogen atom could be found in the topology, but the converter requires all hydrogens to be explicit. You can use the parameter ``NoImplicit=False`` when using the converter to allow implicit hydrogens and disable inferring bond orders and charges. You can also use ``force=True`` to ignore this error.
```

I cannot pass the suggested keywords to `select_atoms`.

``` python
import MDAnalysis as mda
u = mda.fetch_mmtf('3cs9')

# MMTF does not fill the elements attribute
u.add_TopologyAttr('elements', u.atoms.types)
u.select_atoms('smarts c3ccncc3c4nc(Nc1cc(NC(=O)c2ccccc2)ccc1)ncc4')
```

- Which version are you using? '2.0.0-dev0'
- Which version of Python (`python -V`)? '3.7.6 | packaged by conda-forge | (default, Jan  7 2020, 21:48:41) [MSC v.1916 64 bit (AMD64)]'
- Which operating system? Windows

cbouy:
Fixes #3319 

 - Adds the `rdkit_kwargs` parameter to the `select_atoms` method for passing arguments to the RDKitConverter

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


richardjgowers:
Can we call this smarts_kwargs instead?

cbouy:
@richardjgowers Hmmm I see your point, this is only needed when using SMARTS so it makes sense to call it smarts_kwargs
But if we later add other RDKit-based selections, rdkit_kwargs makes more sense. Also the docs specifies that the `smarts` selection uses RDKit so it shouldn't be too surprising for users. And most importantly, smarts_kwargs makes it sound like you're tweaking how the SMARTS search is performed which might be confusing for users?

package/MDAnalysis/core/groups.py
```diff
@@ -2716,6 +2717,9 @@ def select_atoms(self, sel, *othersel, periodic=True, rtol=1e-05,
           force the selection to be re evaluated each time the Timestep of the
           trajectory is changed.  See section on **Dynamic selections** below.
           [``True``]
+        rdkit_kwargs : dict (optional)
+          Arguments passed to the RDKitConverter when using selection based on
```
IAlibay, line 2721: :class:`converters.RDKitConverter` (or whatever sphinx highlighting looks better)?

IAlibay:
I _think_ I agree with @cbouy here. In fact `rdkit_kwargs` isn't clear enough imho that these are kwargs for the converter not the kwargs that are passed directly to rdkit itself.

cbouy:
@richardjgowers @IAlibay so would `rdkitconverter_kwargs` work for you?

testsuite/MDAnalysisTests/core/test_atomselections.py
```diff
@@ -575,6 +575,13 @@ def test_invalid_smarts_sel_raises_error(self, u2):
         with pytest.raises(ValueError, match="not a valid SMARTS"):
             u2.select_atoms("smarts foo")
 
+    def test_passing_args_to_converter(self):
+        u = mda.Universe.from_smiles("O=C=O")
+        with pytest.raises(AttributeError, match="No hydrogen atom"):
```
richardjgowers, line 580: this should be two tests

richardjgowers:
Sorry to be a pain but no.  If I'm getting an error with a smarts
selection, I don't think it's immediately obvious that I need to start
playing with a kwarg called "rdkit_something".  Yes, RDKit is how we
implement it (and will likely be for a while) but that's only obvious to a
developer.

IAlibay:
> Sorry to be a pain but no. If I'm getting an error with a smarts selection, I don't think it's immediately obvious that I need to start playing with a kwarg called "rdkit_something". Yes, RDKit is how we implement it (and will likely be for a while) but that's only obvious to a developer.

Whilst I agree with you, `smarts_kwargs` is also kinda non-descriptive too no?
If I'm passing a `smarts_kwargs` I'd expect to be passing kwargs that affect the way in which the `smarts` is being calculated. In this case we're passing converter specific kwargs (that really don't do anything about how the smarts work, but how much we tolerate the presence of hydrogens).

My stance here is that the kwargs name has to point the users to which part of the docs they should be reading (it's advanced user stuff, and shouldn't really be something everyday users should be worrying about).

testsuite/MDAnalysisTests/core/test_atomselections.py
cbouy: That first test was just me making sure I reproduce the same error as in the original issue, it's not really needed now so I'll delete it

cbouy:
How about `smarts_converter` (or `smarts_converter_kwargs`) ?
It's explicit that it will help with whatever smarts-related problems the user is having, and is descriptive enough to (hopefully) not be confused with tweaking how the smarts matching is made.
Also the error message only mentions `converter` and not `smarts` so users might not even understand that the issue with their selection comes from there if they are using different types of selections. `smarts_converter` should catch the attention of people who know the issue comes from the smarts selection, and users who have no clue about anything whatsoever
```
AttributeError: No hydrogen atom could be found in the topology, but the converter requires all hydrogens to be
explicit. You can use the parameter ``NoImplicit=False`` when using the converter to allow implicit hydrogens and
disable inferring bond orders and charges. You can also use ``force=True`` to ignore this error.
```
Or is it worse ?😆

IAlibay:
@richardjgowers @cbouy can we get this fixed on a quick turnaround for 2.0? As far as I'm aware we're essentially feature incomplete if we don't have this fixed?

I think most of the issue here is "what are we naming the thing"?

richardjgowers (approved):
I think we can use rdkit_kwargs here since it's documented well

testsuite/MDAnalysisTests/core/test_atomselections.py
IAlibay: I _think_ this is the only thing that needs fixing, I guess it's just a case of deleting these two lines?

IAlibay (changes requested):
@cbouy if you can just delete these two lines then we can merge this :)

IAlibay (approved):
Thanks @cbouy ! I'll merge once CI returns green.

PR #3338, opened 2021-05-27
Adds python_requires >= 3.6 to develop

### LINKED ISSUE #3333: [2.0.0 release] fix setup.py requires
Moving from 1.x -> 2.0, we lost the python_requires entry. We should add one back so that we can prevent folks from trying to install MDA 2.0 on py<3.6.

IAlibay:
Fixes #3333 

Simple one, we're just adding a python_requires back so that folks don't try to install with py3.5 or py2.7.

 - adds back a `python_requires` to setup.py

PR Checklist
------------
 - Tests?
 - Docs?
 - CHANGELOG updated?

### DIFF
--- a/package/setup.py
+++ b/package/setup.py
@@ -635,9 +635,11 @@ if __name__ == '__main__':
                         ],
           },
           ext_modules=exts,
+          python_requires='>=3.6',
           requires=['numpy (>=1.16.0)', 'biopython (>= 1.71)', 'mmtf (>=1.0.0)',
                     'networkx (>=1.0)', 'GridDataFormats (>=0.3.2)', 'joblib',
                     'scipy (>=1.0.0)', 'matplotlib (>=1.5.1)', 'tqdm (>=4.43.0)',
+                    'threadpoolctl',
                     ],
           # all standard requirements are available through PyPi and
           # typically can be installed without difficulties through setuptools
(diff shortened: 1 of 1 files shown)


orbeckst (approved):
Does what it tries to fix. LGTM

(Good catch on missing threadpoolctl, I hope nothing else is missing)

IAlibay:
> Does what it tries to fix. LGTM
> 
> (Good catch on missing threadpoolctl, I hope nothing else is missing)

So GSD is missing from that list, but both it and threadpoolctl are in `install_requires`, so for most cases it's fine? I'll raise an issue though.

IAlibay:
Actually @orbeckst is `requires` just deprecated? I found this comment in your PR for #2738:

"- removed deprecated 'requires' in setup.py (see
  https://setuptools.readthedocs.io/en/latest/setuptools.html#developer-s-guide)"

But I'm not seeing what it refers to. If we can confirm it's no longer needed I'll just delete it out here.

IAlibay:
Ok yeah, it is deprecated, see: https://setuptools.readthedocs.io/en/latest/history.html?highlight=requires#id359

I'll just remove it.

orbeckst (approved):
Good that you found out about ditching `requires` — I had totally forgotten against my PR against 1.0. All good here.

PR #3344, opened 2021-05-31
ChainReaders now apply their own transformations

### LINKED ISSUE #3343: Transformations not applied to some frames with ChainReader
## Expected behavior ##
If an `Universe` is created with more than one trajectory file and a `MDAnalysis.transformations.fit.fit_rot_trans` transformation is added to it, it is expected that all frames get fitted on-the-fly when we iterate over the trajectory.

## Actual behavior ##
Whenever a `ChainReader` is used to load the trajectories, the transformation isn't applied to some frames. It seems that there is a pattern on which frames are fitted and which are not. I reproduced this behavior with independent XTC and DCD files.

This code results in some frames with low and others with high RMSD values. It was expected that all frames had low RMSD.
``` python
import MDAnalysis as mda
from MDAnalysis.analysis.rms import rmsd
from MDAnalysis.transformations.fit import fit_rot_trans

reference = mda.Universe(PSF, DCD)
simul = mda.Universe(PSF, [DCD])  # Force the usage of a ChainReader (wrong behavior)
# simul = mda.Universe(PSF, DCD, DCD)  # Wrong behavior
# simul = mda.Universe(PSF, [DCD, DCD])  # Wrong behavior
# simul = mda.Universe(PSF, DCD)  # Correct behavior

# Un-fit the reference
reference.atoms.rotateby(100, [0, 0, 1])
reference.atoms.translate([10, 10, 10])

# Fitting transform
ref_sel = reference.select_atoms("name CA")
fit_sel = simul.select_atoms("name CA")
transform = fit_rot_trans(fit_sel, ref_sel)
simul.trajectory.add_transformations(transform)

for ts in simul.trajectory[:5]:
    print(rmsd(fit_sel.positions, ref_sel.positions))

```
Output:
```
27.879789573782062
0.42571286245739176
27.965311103398278
0.7381344995787231
28.00443710176041
```

If we don't use the `ChainReader` by un-commenting the following line, all frames get fitted.
``` python
simul = mda.Universe(PSF, DCD)  # Expected behavior
```
Output:
```
1.280204933315509e-06
0.4234302855368412
0.5936585854007296
0.7368312603393679
0.8277202818678987
```

If we keep using the `ChainReader`, but explicitly apply the transformation in the trajectory loop, all frames get fitted.
``` python
for ts in simul.trajectory[:5]:
    ts = transform(ts)
    print(rmsd(fit_sel.positions, ref_sel.positions))
```
Output:
```
1.3684785310354626e-06
0.4234302377147153
0.5936586440093501
0.7368312953807447
0.8277203745278706
```

Just for completeness, here is a case without the fitting stuff and the correct behavior (the example trajectory was already fitted).
``` python
import MDAnalysis as mda
from MDAnalysis.analysis.rms import rmsd

simul = mda.Universe(PSF, [DCD])
ref_sel = reference.select_atoms("name CA")
fit_sel = simul.select_atoms("name CA")

for ts in simul.trajectory[:5]:
    print(rmsd(fit_sel.positions, ref_sel.positions))
```
Output:
```
0.0
0.4257129087755455
0.5951654476464195
0.7384635729638336
0.8350828168683272
```

The above codes were tested on versions 1.1.1 and 2.0.0-dev0.
- Which version of Python ()?
Versions 3.6.9 and 3.8.5.
- Which operating system?
Ubuntu 18.04 and 20.04

mnmelo:
Closes #3343

Takes into account over-transformation of `SingleFrameReaders`, which
never re-create their `ts` when re-accessed.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


mnmelo:
To prevent over-transformation of `SingleFrameReaders` this now executes an `active_reader.rewind()` if the `active_reader` is a `SingleFrameReader`. It works, but might penalize performance because the reader will read the Timestep at least twice.

I'm exploring the possibility that the ChainReader stores an unmodified Timestep along with the sub-reader.

jbarnoud:
How does that work with the memory reader? This was the main reason the ChainReader is currently differing the transformations for the sub-readers.

mnmelo:
Good point. Will check. Tests passed, though this may just mean we need more tests for these conditions 😉

mnmelo:
I dug deeper and, from what I could understand, @jbarnoud, there should be no conflict with the MemoryReader. Here's the several cases that I see may happen when loading a Universe with a ChainReader:
1. `in_memory` requested at Universe creation time and `transformations` passed: in this case coordinates are first read into memory, and only then transformations added and applied. It will always be the MemoryReader (which has, by then, replaced the ChainReader) that will apply them, and this shouldn't differ from current behavior;
2. `transfer_to_memory` issued after the Universe is loaded with transformations: in this case timesteps will be transformed by the ChainReader, then read into memory. This seems to also be the current behavior.
3. `add_transformations` issued after the trajectory is in memory: this is essentially the same order of events as during Universe initialization.

Did you have any specific conflicting use-case in mind that I can double check? Perhaps there may be contrived corner cases where one of a ChainReader's sub-readers is a MemoryReader (is this even possible?). Was this what you were thinking about?

jbarnoud:
The case that I expect to break is if you have a memory reader within the chain reader. That could happen if one of the trajectories in the chain is an array of coordinates. Then you have a special case with the chain reader and a special case for the memory reader, question is : how do these special case work with each other? On 7 Jun 2021 02:51, Manuel Nuno Melo ***@***.***> wrote:
I dug deeper and, from what I could understand, @jbarnoud, there should be no conflict with the MemoryReader. Here's the several cases that I see may happen when loading a Universe with a ChainReader:
in_memory requested at Universe creation time and transformations passed: in this case coordinates are first read into memory, and only then transformations added and applied. It will always be the MemoryReader (which has, by then, replaced the ChainReader) that will apply them, and this shouldn't differ from current behavior;transfer_to_memory issued after the Universe is loaded with transformations: in this case timesteps will be transformed by the ChainReader, then read into memory. This seems to also be the current behavior.add_transformations issued after the trajectory is in memory: this is essentially the same order of events as during Universe initialization.
Did you have any specific conflicting use-case in mind that I can double check? Perhaps there may be contrived corner cases where one of a ChainReader's sub-readers is a MemoryReader (is this even possible?). Was this what you were thinking about?

mnmelo:
Ok, so maybe not so much a corner case.

In that situation we have indeed a problem, in that the ChainReader risks applying the same transform multiple times, cumulatively, on the MemoryReader. I'm trying to make the ChainReader aware of this and somehow keep track of the transformed sub-frames.

mnmelo:
The tests are passing, but there are some problems connecting to Codecov from GH Actions CI (a consequence of the fastly crash?). I keep re-running the GH Actions CI tests but there's always one or two that can't complete that post-test step. Would a rerun of all CI help? (how can I even trigger it?)

IAlibay:
> The tests are passing, but there are some problems connecting to Codecov from GH Actions CI (a consequence of the fastly crash?). I keep re-running the GH Actions CI tests but there's always one or two that can't complete that post-test step. Would a rerun of all CI help? (how can I even trigger it?)

@mnmelo I didn't get to see what happened. Maybe let's see what happens when this current round finishes running? There's a few options to re-run all the CI (the easiest is just to close and re-open the PR), but it's not necessary at the moment.

IAlibay:
Seemed to have worked this time. We can ignore the codecov project report (since it's just affected by the reduced number of lines).

orbeckst:
Given the slew of ChainReader-related issues, I'd like to breath some new life into this PR...

PR #3353, opened 2021-06-10
Fixed wrong atom sorting in _split_by_compound_indices

mnmelo:
Closes #3353

 `_split_by_compound_indices` was not properly propagating the sorting when sorting was needed.

Tested initially on an example LAMMPS structure provided by @hejamu, but found that `1hvr.pdb` (`MDAnalysisTests.datafiles.CONECT`) also elicited the same problem when unwrapping. The implemented test uses the latter, but the fix does work for the LAMMPS case too. 

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -23,6 +23,8 @@ The rules for this file:
   * 2.0.0
 
 Fixes
+  * Fixed sometimes wrong sorting of atoms into fragments when unwrapping
+    (Issue #3352, PR #3353)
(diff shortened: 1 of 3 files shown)


mnmelo:
(I pinged you, @richardjgowers, because this is essentially a fix over #3005, which you reviewed).

richardjgowers (approved):

PR #874, opened 2016-06-10
Plugin to display the open files upon test failure

jbarnoud:
Add a plugin that list open files when a test fails. This aims at testing the hypothesis that we have to many opened file descriptors, and that it could cause #853.

The plugin is enabled by default. It displays the path to the file that
have a open file handler. To disable the plugin, run mda_nosetests with
the --no-open-files option.

The commit also introduces a test that fails if the MDA_FAILURE_TEST
environment variable is set. This test comes handy to test nose behavior
upon failure.

The open-files plugin requires psutil. That requirement has been added
to setup.py.

PR #3354, opened 2021-06-19
Adds macos-11 runner

IAlibay:
Tries enabling macOS-11 runner (apparently we might have gotten access to the private preview now)

 - CI now does bigsur 3.9 and catalina 3.6 checks [max/min if you wish]
 - setup now makes 10.9 the minimum macos deployment version
 - get rid of `psutil` as an MDAnalysisTests dependency

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/setup.py
+++ b/package/setup.py
@@ -301,9 +301,9 @@ def extensions(config):
     # needed to specify c++ runtime library on OSX
     if platform.system() == 'Darwin' and using_clang():
         cpp_extra_compile_args.append('-stdlib=libc++')
-        cpp_extra_compile_args.append('-mmacosx-version-min=10.14')
+        cpp_extra_compile_args.append('-mmacosx-version-min=10.9')
         cpp_extra_link_args.append('-stdlib=libc++')
-        cpp_extra_link_args.append('-mmacosx-version-min=10.14')
+        cpp_extra_link_args.append('-mmacosx-version-min=10.9')
 
     # Needed for large-file seeking under 32bit systems (for xtc/trr indexing
     # and access).
--- a/package/setup.py
+++ b/package/setup.py
@@ -301,9 +301,9 @@ def extensions(config):
     # needed to specify c++ runtime library on OSX
     if platform.system() == 'Darwin' and using_clang():
         cpp_extra_compile_args.append('-stdlib=libc++')
-        cpp_extra_compile_args.append('-mmacosx-version-min=10.9')
+        cpp_extra_compile_args.append('-mmacosx-version-min=10.14')
         cpp_extra_link_args.append('-stdlib=libc++')
-        cpp_extra_link_args.append('-mmacosx-version-min=10.7')
+        cpp_extra_link_args.append('-mmacosx-version-min=10.14')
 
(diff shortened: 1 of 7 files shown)


IAlibay:
Huzzah 🎉 

```
Operating System
  macOS
  11.4
  20F71
```

IAlibay:
Looks like deps are going to be a problem 🤔

```
Best match: psutil 5.8.0
Processing psutil-5.8.0.tar.gz
Writing /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/easy_install-ord_n98e/psutil-5.8.0/setup.cfg
Running psutil-5.8.0/setup.py -q bdist_egg --dist-dir /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/easy_install-ord_n98e/psutil-5.8.0/egg-dist-tmp-ece2t05m
No eggs found in /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/easy_install-ord_n98e/psutil-5.8.0/egg-dist-tmp-ece2t05m (setup script problem?)
error: The 'psutil>=4.0.2' distribution was not found and is required by MDAnalysisTests
Error: Process completed with exit code 1.
```

IAlibay:
It looks like we maybe don't need psutil anymore? (seems like something pre-pytest move?)

@jbarnoud this seems like psutil was added in one of your old PRs #874 - any ideas here?

jbarnoud:
> @jbarnoud this seems like psutil was added in one of your old PRs #874 - any ideas here?

We aren't testing for open files since we moved to pytest.

IAlibay:
For some reason in #2184 the macos minimum version for library linking was set to 10.7 which mismatched the 10.9 minimum version we were setting in other places.

Given that py3.9 now has a minimum of 10.9, I've changed everything to 10.9. In the long run we should consider whether we want to start ditching older versions of macos (not that it actually affects us too much).

Aside from all this, the CI run just demonstrates that big sur was compatible for `x86_64`, so I'm not adding a changelog entry (unless we feel it's useful user info).

The next thing will be to look into cross-compiling `universe2` stuff, but I think we'll need to think about re-working CI and setup for that [see discord] (plus testing on M1 without an M1 machine is going to be a pain).

package/setup.py
```diff
@@ -303,7 +303,7 @@ def extensions(config):
         cpp_extra_compile_args.append('-stdlib=libc++')
         cpp_extra_compile_args.append('-mmacosx-version-min=10.9')
         cpp_extra_link_args.append('-stdlib=libc++')
-        cpp_extra_link_args.append('-mmacosx-version-min=10.7')
+        cpp_extra_link_args.append('-mmacosx-version-min=10.9')
```
richardjgowers, line 306: does this bump the minimum required macos?

IAlibay: I don't think it does since we were using a minimum compile version of 10.9 anyways [line 304].

In any case:
1) Mavericks was released in 2013 and was EOL in 2016.
2) Python 3.9 only supports 10.9+ anyways.

richardjgowers (approved):

PR #3356, opened 2021-06-21
fixes #3307

### LINKED ISSUE #3307: Document TRZ writer behaviour when dimensions is None
Follow up from #3132 

When there is no timestep dimension, the TRZ writer will write out a box of zeroes. I can't find any reference for the TRZ format, but this behaviour probably needs to be both documents and warned.

richardjgowers:
writing a trz file with dimensions=None issues a warning

Fixes #3307

 - 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/coordinates/TRZ.py
+++ b/package/MDAnalysis/coordinates/TRZ.py
@@ -571,6 +571,8 @@ class TRZWriter(base.WriterBase):
         if ts.dimensions is not None:
             unitcell = triclinic_vectors(ts.dimensions).reshape(9)
         else:
+            warnings.warn("Timestep didn't have dimensions information, "
+                          "box will be written as all zero values")
             unitcell = np.zeros(9, dtype=np.float32)
 
         try:
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -284,6 +284,7 @@ Changes
   * Move water bridge analysis from hbonds to hydrogenbonds (Issue #2739 PR #2913)
   * `threadpoolctl` is required for installation (Issue #2860)
   * Added OpenMM coordinate and topology converters (Issue #2863, PR #2917)
+  * Writing a TRZ file with no box information now raises a warning (Issue #3307)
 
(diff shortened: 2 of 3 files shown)


package/MDAnalysis/coordinates/TRZ.py
```diff
@@ -571,6 +571,8 @@ def _write_next_frame(self, obj):
         if ts.dimensions is not None:
             unitcell = triclinic_vectors(ts.dimensions).reshape(9)
         else:
+            warnings.warn("Timestep didn't have dimensions information, "
```
IAlibay, line 574: We probably should have self-consistency here, if `None` == np.zeros(6), then `np.allclose(box, np.zeros(6))` should read in as `None` ? (my understanding is that currently it just sets a zero sized unitcell?)

IAlibay (changes requested):
Thanks, just the one comment [I know it's a near-zero use reader/writer, but we should really read the same way we write]

package/MDAnalysis/coordinates/TRZ.py
richardjgowers: added a test for this, it's already implemented like this

IAlibay: My bad I forgot about : https://github.com/MDAnalysis/mdanalysis/blob/4154e6304a38c076ef008907ed393d36ba5958a8/package/MDAnalysis/coordinates/base.py#L795

Side note - shouldn't that be an `allclose` [if we're comparing floats] ?

IAlibay (approved):
I'll approve - the only thing I'll leave up to you is whether you want to document the behaviour somewhere, maybe an entry in the class notes?

[I'm not going to request changes for it since it's a very low use class]

richardjgowers:
I think because we're literally looking for zeros all() is ok here.  This
isn't a value which is computed and could be subject to rounding errors.

PR #2333, opened 2019-08-21
Unhardcode LammpsDumpParser

lilyminium:
**Fixes below:**

[Amir can't load their LammpsDump file because atom lines have more than 5 fields.](https://groups.google.com/forum/?pli=1#!topic/mdnalysis-discussion/XzkQ-scJxLc)

```
>>> cell = mda.Universe(path, format='LAMMPSDUMP', atom_style='id type x y z xs ys zs')


ValueError                                Traceback (most recent call last)
/anaconda3/lib/python3.7/site-packages/MDAnalysis/core/universe.py in __init__(self, *args, **kwargs)
    289                     with parser(self.filename) as p:
--> 290                         self._topology = p.parse(**kwargs)
    291                 except (IOError, OSError) as err:
/anaconda3/lib/python3.7/site-packages/MDAnalysis/topology/LAMMPSParser.py in parse(self, **kwargs)
    609             for i in range(natoms):
--> 610                 idx, atype, _, _, _ = fin.readline().split()
    611 
ValueError: too many values to unpack (expected 5)
```

**Changes made in this Pull Request:**
```diff
-                  idx, atype, _, _, _ = fin.readline().split()
+                  idx, atype, *_ = fin.readline().split()
```

I know nothing about Lammps formats, so maybe atom lines having >5 fields is a sign that something is very wrong -- but then an informative error could be raised. Or possibly it should have at least 5 fields (for x, y, z)?

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?
 - [ ] Issue raised/referenced?

PR #3360, opened 2021-07-06
Fix reading of lammpsdump ATOM card. 

### LINKED ISSUE #3358: LAMMPS DumpReader does not return the correct coordinates for unwrapped trajectories.
## Expected behavior ##
Unscaled coordinates written with LAMMPS in a `lammpstrj` dump format should be parsed correctly. 

## Actual behavior ##
The DumpReader Reader does not check what convention coordinates are written in. (xs,ys,zs) vs (xu,yu,zu). The transform `distances.transform_StoR(ts.positions, ts.dimensions)` is also called in `_read_next_timestep` regardless of whether the coordinates are scaled or not, resulting in insanely large coordinates if unscaled (xu,yu,zu) coordinates are input. 

``` python
import MDAnalysis as mda
import numpy as np
from numpy import testing

n_atoms = 1536
# do it by reading the file by hand
def parse_lammpstrj(file, n_atoms):
    with open(file, "r") as f:
        lines = f.readlines()
        atomsegs = []
        for i, line in enumerate(lines):
            if "ITEM: ATOMS" in line:
                atomseg = lines[i+1:i+n_atoms+1]
                atomsegs.append(atomseg)
        coords = np.empty([len(atomsegs),n_atoms,3])
        for frame in range(len(atomsegs)):
            atom_frame = atomsegs[frame]
            for atom in range(len(atom_frame)):
                crd_line = atom_frame[atom]
                idx, _, xu, yu, zu = crd_line.split()
                xu = float(xu)
                yu = float(yu)
                zu = float(zu)
                coords[frame, atom,:] = [xu,yu,zu]
        return coords

coords = parse_lammpstrj("wat_unwraped.lammpsdump",n_atoms)

# do it with MDA
u = mda.Universe("wat_unwraped.lammpsdump")
coords_mda = np.empty([len(u.trajectory),n_atoms,3])
for i,ts in enumerate(u.trajectory):
    coords_mda[i,:,:] = ts.positions

np.testing.assert_allclose(coords_mda,coords)
....

```

- Which version are you using: 2.0.0-dev0
- Which version of Python: 3.7.5
- Which operating system: Ubuntu Eoan Ermine

hmacdope:
Fixes #3358
Supersedes #2333 
 - Allow coordinates in different conventions to be read by LAMMPSDumpReader
 - Unhardcode LAMMPSDumpReader to expect a set number of columns
 - S->R transform with ` distances.transform_StoR` is no longer applied by default, only if the coordinates are in the `scaled (xs,ys,zs)` or `scaled_unwrapped (xsu, ysu, zsu)` convention
 - On the way to allowing reading of other ts data from the LAMMPS dump file.

PR Checklist
------------
 - [x] Tests?
 - [X] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -552,16 +561,57 @@ def _read_next_timestep(self):
...
+        col_ids = {}  # column index of each attribute
+        for i, attr in enumerate(attrs):
+            col_ids[attr] = i
+
+        # check for ids and what type of coordinate convention
+        ids = "id" in col_ids.keys()
+        # keys = convention values = col_ids or False if not present
+        coord_dict = {"unscaled": False, "scaled": False, "unwrapped": False,
+                      "scaled_unwrapped":False}
+        if "x" and "y" and "z" in col_ids.keys():  # unscaled
```
richardjgowers, line 575: Won't this be doing the Boolean logic on the left before the "in" check? I think you need something gross like "x in col AND y in col"

package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -552,16 +561,57 @@ def _read_next_timestep(self):
...
+        if "xu" and "yu" and "zu" in col_ids.keys():  # unwrapped
+            coord_dict["unwrapped"] = [col_ids["xu"], col_ids["yu"],
+                                       col_ids["zu"]]
+        if "xsu" and "ysu" and "zsu" in col_ids.keys():  # scaled unwrapped
+            coord_dict["scaled_unwrapped"] = [col_ids["xsu"], col_ids["ysu"],
+                                              col_ids["zsu"]]
+
+        # this should only trigger on first read of "ATOM" card, after which it
+        # is fixed to the guessed value.
+        if self.coordinate_convention == "guess":
```
richardjgowers, line 590: Maybe mention that the default is unscaled?

package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -552,16 +561,57 @@ def _read_next_timestep(self):
...
+            if ids:
+                indices[i] = fields[col_ids["id"]]
+            x, y, z = [fields[i] for i in coord_cols]
+            ts.positions[i] = x, y, z
 
         order = np.argsort(indices)
         ts.positions = ts.positions[order]
-        # by default coordinates are given in scaled format, undo that
-        ts.positions = distances.transform_StoR(ts.positions, ts.dimensions)
+        if self.coordinate_convention == "scaled":
```
richardjgowers, line 612: Or scaled unwrapped

testsuite/MDAnalysisTests/coordinates/test_lammps.py
```diff
@@ -434,7 +434,8 @@ def u(self, tmpdir, request):
                 with gzip.GzipFile(f, 'wb') as fout:
                     fout.write(data)
 
-        yield mda.Universe(f, format='LAMMPSDUMP')
+        yield mda.Universe(f, format='LAMMPSDUMP',
+                           coordinate_convention='guess')
```
richardjgowers, line 438: I'd call this lammps coordinate convention to make it clear it's a format specific option?

richardjgowers (changes requested):
What happens when multiple coordinate columns are present?

package/MDAnalysis/coordinates/LAMMPS.py
hmacdope: Doh...

package/MDAnalysis/coordinates/LAMMPS.py
hmacdope: Does scaled_unwrapped use the same S->R? I wasn't sure so i left it out.

testsuite/MDAnalysisTests/coordinates/test_lammps.py
hmacdope: Roger roger

package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -449,19 +450,27 @@ def write(self, selection, frame=None):
 class DumpReader(base.ReaderBase):
     """Reads the default `LAMMPS dump format`_
 
-    Expects trajectories produced by the default 'atom' style dump.
-
-    Will automatically convert positions from their scaled/fractional
-    representation to their real values.
+    Supports coordinates in the "unscaled" (x,y,z), "scaled" (xs,ys,zs),
```
IAlibay: Link to format docs if available?

package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -449,19 +450,27 @@ def write(self, selection, frame=None):
...
 
-    def __init__(self, filename, **kwargs):
+    def __init__(self, filename, coordinate_convention="guess", **kwargs):
         super(DumpReader, self).__init__(filename, **kwargs)
 
         root, ext = os.path.splitext(self.filename)
+        if coordinate_convention in self._conventions:
+            self.coordinate_convention = coordinate_convention
+        else:
+            raise ValueError("coordinate convention incorrectly specified")
```
IAlibay, line 476: Please do make sure all the things that codecov complain about get appropriate tests.

package/MDAnalysis/topology/LAMMPSParser.py
```diff
@@ -584,9 +584,9 @@ def _parse_box(self, header):
 
 
 class LammpsDumpParser(TopologyReaderBase):
-    """Parses Lammps ascii dump files in 'atom' format
-
-    Only reads atom ids.  Sets all masses to 1.0.
+    """Parses Lammps ascii dump files in 'atom' format.
+
```
IAlibay, line 588:
```suggestion

```

package/MDAnalysis/topology/LAMMPSParser.py
```diff
@@ -584,9 +584,9 @@ def _parse_box(self, header):
...
 
 class LammpsDumpParser(TopologyReaderBase):
-    """Parses Lammps ascii dump files in 'atom' format
-
-    Only reads atom ids.  Sets all masses to 1.0.
+    """Parses Lammps ascii dump files in 'atom' format.
+    
+    Sets all masses to 1.0.
 
     .. versionadded:: 0.19.0
```
IAlibay, line 591: add versionchanged?

package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -123,6 +123,7 @@
 """
 import os
 import numpy as np
+import warnings
```
IAlibay, line 126:
```suggestion
```
Is warnings being used anywhere?

testsuite/MDAnalysisTests/datafiles.py
```diff
@@ -137,6 +137,7 @@
     "LAMMPShyd", "LAMMPShyd2",
     "LAMMPSdata_deletedatoms",  # with deleted atoms
     "LAMMPSDUMP",
+    "LAMMPSDUMP_long",  # lammpsdump file with extra columns
```
IAlibay, line 140: might be worth defining what extra columns somewhere

IAlibay (changes requested):
Couple of quick comments, probably needs more tests?

package/MDAnalysis/coordinates/LAMMPS.py
hmacdope: Yeah more tests needed for sure.

testsuite/MDAnalysisTests/datafiles.py
hmacdope: I might generate some new test data to cover a few of the other coordinate conventions now that you mention it.

package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -449,19 +449,31 @@ def write(self, selection, frame=None):
...
+    "scaled", "unwrapped", "scaled_unwrapped" and whichever set of coordinates 
+    is detected first will be used. If coordinates are given in the scaled
+    coordinate convention (xs,ys,zs) or scaled unwrapped coordinate convention
+    (xsu,ysu,zsu) they will automatically be converted from their 
+    scaled/fractional representation to their real values.
 
     .. versionadded:: 0.19.0
     """
     format = 'LAMMPSDUMP'
+    _conventions = ["guess", "unscaled", "scaled", "unwrapped",
```
richardjgowers, line 465: rather than guess, this is more like auto right?

package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -552,16 +564,59 @@ def _read_next_timestep(self):
 
         indices = np.zeros(self.n_atoms, dtype=int)
 
-        f.readline()  # ITEM ATOMS etc
-        for i in range(self.n_atoms):
-            idx, _, xs, ys, zs = f.readline().split()
+        atomline = f.readline()  # ITEM ATOMS etc
+        attrs = atomline.split()[2:]  # attributes on coordinate line
+        col_ids = {}  # column index of each attribute
```
richardjgowers, line 569: can just make this a one liner, `col_ids = {i: attr for i, attr in enumerate(attrs)}`

package/MDAnalysis/coordinates/LAMMPS.py
richardjgowers: Yeah I'd want to test to make sure but I'm assuming scaled_unwrapped can be 1.5 to mean it's 0.5 * boxlength + 1 boxlength

testsuite/MDAnalysisTests/coordinates/test_lammps.py
```diff
@@ -507,3 +509,86 @@ def test_atom_reordering(self, u, reference_positions):
...
+        atom340_pos2_unscaled = [4.41947, 35.4403, 2.25115]
+        atom340_pos3_unscaled = [4.48989, 0.360633, 2.63623]
+        return np.asarray([atom340_pos1_unscaled, atom340_pos2_unscaled,atom340_pos3_unscaled])
+    
+    def test_unscaled_reference(self, universes, reference_unscaled_positions):
+        atom_340 = universes["unscaled"].atoms[339]
+        for i,ts_u in enumerate(universes["unscaled"].trajectory[0:3]):
+            assert_almost_equal(atom_340.position, reference_unscaled_positions[i,:], decimal=5)
+    
+    def test_scaled_reference(self, universes, reference_unscaled_positions):
```
hmacdope, line 555: @richardjgowers @IAlibay question surrounding the accuracy of the `StoR` transform from `lib.distances`. Is this function known to be ill conditioned or something? This only seems to match to 1 dp.

package/MDAnalysis/coordinates/LAMMPS.py
hmacdope: Correct in that it uses same S->R.

testsuite/MDAnalysisTests/coordinates/test_lammps.py
hmacdope: Now a separate issue #3361

hmacdope:
This is ready for re review :)

p-j-smith:
This looks like a really nice improvement to the LAMMPS parser. I have a few general comments that shouldn't necessarily be addressed in this PR but could be nice to add at some point:

- Often, in the LAMMPS data files, the dimensions of the box go from -L/2 to L2, rather then from 0 to L. The last time I checked, there was no mention in the documentation that MDAnalysis will store only the extent of the box, implicitly assuming the box goes from 0 to L. It's not really an issue, but may be worth warning users. For example, if I didn't know this was the case, I wouldn't expect `u.atoms.wrap()` to change to the coordinates of my atoms if they were already wrapped in a box that goes from -L/2 to L/2.

- There are other important optional flags in the LAMMPS data file: `nx`, `ny`, and `nz`. For each atom, these flags (signed integers) specify the number of times each periodic boundary has been crossed, and can be used for nojump unwrapping of coordinates. It would be nice to be able to make use of these with mdanalysis.

- The data file in LAMMPS contains no atom names or residue names. However, there is a convention (used by moltemplate and I think the VMD LAMMPS data parser, provided by topotools) that can be used for supplying this information. It's been a while since I used LAMMPS, but I think the convention is to add this information in a comment at the end of the line for each ATOM entry. It would be cool if MDAnalysis followed the same convention because using these attributes is so much more intuitive than working with arbitrarily numbered atom types (arbitrarily numbered by LAMMPS, not MDAnalysis). It is possible to manually add this information after creating the universe, but it would be nice to not need to.

hmacdope:
Thanks @p-j-smith I will raise some issues.

 I think a partial solution to pulling in extra data would be to be able to pull any column with a kwarg, eg:
```python
mda.Universe(coords, format='LAMMPSDUMP', additional_data='nx')
```
however I think this would need to be standardised across all readers / a solution may already exist.
Ping @MDAnalysis/coredevs how do any of the other readers handle arbitrary timestep info?

RE box, perhaps we should emit a warning of some kind, but where I guess is the key question. seems excessive to warn on coordinate read. Does that then become a job for `u.atoms.wrap` etc etc?

RE atom and residue names, should be an easy enough fix if someone can provide a data file in the appropriate format. :)

p-j-smith:
> I will raise some issues.

Thanks! I probably should have done that rather than bombarding you here with these unrelated comments.

> I think a partial solution to pulling in extra data would be to be able to pull any column with a kwarg, eg: 

> `mda.Universe(coords, format='LAMMPSDUMP', additional_data='nx')`

LAMMPS uses an `image` flag, and if this is set the `True` then `nx`, `ny`, and `nz` are all written to file. But passing `additional_data='image'` is probably not a very good description for non-LAMMPS users.

> RE box, perhaps we should emit a warning of some kind, but where I guess is the key question. seems excessive to warn on coordinate read. Does that then become a job for u.atoms.wrap etc etc?

Perhaps a warning could be emitted when the universe is created?

> RE atom and residue names, should be an easy enough fix if someone can provide a data file in the appropriate format. :)

I'll see if I can dig one out from somewhere :)

package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -552,16 +567,56 @@ def _read_next_timestep(self):
...
+            coord_cols = coord_dict[self.lammps_coordinate_convention]
 
-            indices[i] = idx
-            ts.positions[i] = xs, ys, zs
+        for i in range(self.n_atoms):
+            fields = f.readline().split()
+            if ids:
+                indices[i] = fields[col_ids["id"]]
+            x, y, z = [fields[i] for i in coord_cols]
+            ts.positions[i] = x, y, z
```
IAlibay, line 613: Probably can just reduce this to one line and add a comment line instead?

package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -552,16 +567,56 @@ def _read_next_timestep(self):
...
 
         order = np.argsort(indices)
         ts.positions = ts.positions[order]
-        # by default coordinates are given in scaled format, undo that
-        ts.positions = distances.transform_StoR(ts.positions, ts.dimensions)
+        if (self.lammps_coordinate_convention == "scaled") or (self.lammps_coordinate_convention == "scaled_unwrapped"):
```
IAlibay, line 617: Please do deal with all the PEP8 issues.

package/MDAnalysis/topology/LAMMPSParser.py
IAlibay: Any reason for not adding the versionchanged here?

testsuite/MDAnalysisTests/coordinates/test_lammps.py
```diff
@@ -507,3 +509,99 @@ def test_atom_reordering(self, u, reference_positions):
...
+def test_open_incorrect_convention_fails():
+    with pytest.raises(ValueError, match="coordinate convention incorrectly specified"):
+        mda.Universe(LAMMPSDUMP, format='LAMMPSDUMP',
+                     lammps_coordinate_convention="42")
+
+
+@pytest.mark.parametrize("convention", ["auto", "unscaled", "scaled",
+                                        "unwrapped", "scaled_unwrapped"])
+def test_open_all_convention(convention):
+    mda.Universe(LAMMPSDUMP_allcoords, format='LAMMPSDUMP',
```
IAlibay, line 530: an assert of some kind would be good here

testsuite/MDAnalysisTests/coordinates/test_lammps.py
IAlibay: 1 d.p. sounds problematic, you kinda have to try to get that bad a float match 😕 
Was this the same issue with the old StoR transformation or is it now only becoming an issue with this code?

IAlibay (changes requested):
Sorry don't have much time to spend on this, but here are some quick comments

p-j-smith:
> RE atom and residue names, should be an easy enough fix if someone can provide a data file in the appropriate format. :)

So I was slightly misremembering things. It is possible to provide an atom name, but not residue name. The atom names are defined using comments in the `masses` section. I've attached a LAMMPS data file as an example. I should add though, after looking into it again, I'm not sure how widely used this convention is. Perhaps there's not wide-enough usage of it to make it worth implementing, especially if other LAMMPS users have never requested this feature before.

[popc-pope-psm.data.zip](https://github.com/MDAnalysis/mdanalysis/files/6817114/popc-pope-psm.data.zip)

testsuite/MDAnalysisTests/coordinates/test_lammps.py
hmacdope: Yeah no changes made to `StoR()` here, I have raised an issue #3361, as will need to validate on a toy system.

package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -552,16 +567,56 @@ def _read_next_timestep(self):
...
+        attrs = atomline.split()[2:]  # attributes on coordinate line
+        col_ids = {attr: i for i, attr in enumerate(attrs)}  # column ids
+
+        # check for ids and what type of coordinate convention
+        ids = "id" in col_ids.keys()
+        # keys = convention values = col_ids or False if not present
+        coord_dict = {"unscaled": False, "scaled": False, "unwrapped": False,
+                      "scaled_unwrapped": False}
+        keys = col_ids.keys()
+        if ("x" in keys) and ("y" in keys) and ("z" in keys):  # unscaled
```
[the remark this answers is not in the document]
hmacdope: 👍

hmacdope:
Thanks for the feedback @IAlibay  @p-j-smith  and @lilyminium.

Answers to a few comments below:

* I think we can deal with extra data to pull from the data file if and when we want it, as it stands this PR gives you all the machinery.
* WRT atom/residue names, perhaps if i knew more about the LAMMPS ecosystem I would be better placed to make a judgement call. However, I am inclined to not read them if it is not what the lammps format dictates. A bit pedantic I know. :/ The comments arn't hurting anyone as they are not currently being read so no need for an error or warning?
*WRT a box, we can warn if the box spans are negative but only on `_read_next_timestep`, as this is the only place the box is read. Is there a graceful way to avoid this do we think? Or am I being stupid.

package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -474,6 +493,10 @@ def _reopen(self):
         self.ts = self._Timestep(self.n_atoms, **self._ts_kwargs)
         self.ts.frame = -1
 
+    def _get_column(self, ids, col_ids):
```
richardjgowers, line 496: This is just a convenience function, I'd just inline it with a lambda where it's used

package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -552,16 +575,52 @@ def _read_next_timestep(self):
...
+            fields = f.readline().split()
+            if ids:
+                indices[i] = fields[col_ids["id"]]
+            ts.positions[i] = [fields[dim] for dim in coord_cols]
 
         order = np.argsort(indices)
         ts.positions = ts.positions[order]
-        # by default coordinates are given in scaled format, undo that
-        ts.positions = distances.transform_StoR(ts.positions, ts.dimensions)
+        if (self.lammps_coordinate_convention == "scaled" or
```
richardjgowers, line 620: `.startswith('scaled')`?

package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -552,16 +575,52 @@ def _read_next_timestep(self):
...
+
+        # this should only trigger on first read of "ATOM" card, after which it
+        # is fixed to the guessed value. Auto proceeds unscaled -> scaled
+        # -> unwrapped -> scaled_unwrapped
+        if self.lammps_coordinate_convention == "auto":
+            match = [var for var in self._conventions[1:] if coord_dict[var]]
+            if match:
+                self.lammps_coordinate_convention = match[0]
+            else:
+                raise ValueError("no coordinate information detected")
```
richardjgowers, line 603: needs a test for this I guess?

package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -552,16 +575,52 @@ def _read_next_timestep(self):
...
+            ["xu", "yu", "zu"], col_ids)  # unwrapped
+        coord_dict["scaled_unwrapped"] = self._get_column(
+            ["xsu", "ysu", "zsu"], col_ids)  # scaled unwrapped
+
+        # this should only trigger on first read of "ATOM" card, after which it
+        # is fixed to the guessed value. Auto proceeds unscaled -> scaled
+        # -> unwrapped -> scaled_unwrapped
+        if self.lammps_coordinate_convention == "auto":
+            match = [var for var in self._conventions[1:] if coord_dict[var]]
+            if match:
```
richardjgowers, line 600: try/except and catch the IndexError if match[0] doesn't exist?

richardjgowers (commented):
The negative lower bound is definitely worth handling in a future PR

hmacdope:
@richardjgowers I have raised an issue for -L/2 -> L/2 box bound discussion. #3363

package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -449,19 +452,35 @@ def write(self, selection, frame=None):
...
-    def __init__(self, filename, **kwargs):
+    def __init__(self, filename, lammps_coordinate_convention="auto",
+                 **kwargs):
         super(DumpReader, self).__init__(filename, **kwargs)
 
         root, ext = os.path.splitext(self.filename)
+        if lammps_coordinate_convention in self._conventions:
+            self.lammps_coordinate_convention = lammps_coordinate_convention
+        else:
+            raise ValueError("coordinate convention incorrectly specified")
```
orbeckst, line 482: The error message could list (1) what was incorrectly specified and (2) what the allowed values are.

package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -550,18 +570,57 @@ def _read_next_timestep(self):
             alpha = beta = gamma = 90.
         ts.dimensions = xlen, ylen, zlen, alpha, beta, gamma
 
+        _get_column = lambda ids, col_ids : [col_ids[i] for i in ids if
```
orbeckst, line 573: Gratuitous use of `lambda` ;-) — you could have also used
```python
def _get_column(...):
   return ...
```
Defining in-block functions is not frowned upon, to my knowledge.

package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -550,18 +570,57 @@ def _read_next_timestep(self):
...
+                                             all(id in col_ids.keys() for
```
orbeckst, line 574: `.keys()` might be superfluous here.

package/MDAnalysis/coordinates/LAMMPS.py
orbeckst: And it could even be a static method.

package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -550,18 +570,57 @@ def _read_next_timestep(self):
...
-            idx, _, xs, ys, zs = f.readline().split()
+        atomline = f.readline()  # ITEM ATOMS etc
+        attrs = atomline.split()[2:]  # attributes on coordinate line
+        col_ids = {attr: i for i, attr in enumerate(attrs)}  # column ids
+        keys = col_ids.keys()
+        # check for ids and what type of coordinate convention
+        ids = "id" in keys
+        # keys = convention values = col_ids or False if not present
+        coord_dict = {conv: False for conv in self._conventions[1:]}
+        coord_dict['unscaled'] = _get_column(
```
orbeckst, line 587: Will `coord_dict` contain the column numbers? Perhaps add a comment to explain the parsing logic.

package/MDAnalysis/topology/LAMMPSParser.py
```diff
@@ -584,10 +584,9 @@ def _parse_box(self, header):
 
 
 class LammpsDumpParser(TopologyReaderBase):
-    """Parses Lammps ascii dump files in 'atom' format
-
-    Only reads atom ids.  Sets all masses to 1.0.
-
+    """Parses Lammps ascii dump files in 'atom' format.
+    Sets all masses to 1.0.
+    .. versionchanged:: 2.0.0
```
orbeckst, line 589: Does this format correctly? Normally we have two blank lines above `.. versionadded`

package/MDAnalysis/topology/LAMMPSParser.py
```diff
@@ -584,10 +584,9 @@ def _parse_box(self, header):
 
 
 class LammpsDumpParser(TopologyReaderBase):
-    """Parses Lammps ascii dump files in 'atom' format
-
-    Only reads atom ids.  Sets all masses to 1.0.
-
+    """Parses Lammps ascii dump files in 'atom' format.
```
orbeckst, line 587: Blank line after first line, see numpy doc standard https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard

package/CHANGELOG
```diff
@@ -23,6 +23,9 @@ The rules for this file:
   * 2.0.0
 
 Fixes
+  * LAMMPSDumpReader can now read coordinates in all the different LAMMPS
```
orbeckst, line 26: I think that's an enhancement, not a bug fix.

orbeckst (commented):
General comments inline.

orbeckst:
I didn't see anything in this PR regarding storing additional data. If you want to look at another format that is very versatile: The H5MDReader loads any additional data into the `ts.data` dict, which I think is the standard way to associate additional per-timestep data. @edisj can tell you more.

hmacdope:
Thanks for the review @orbeckst, I will address comments ASAP. :)

testsuite/MDAnalysisTests/coordinates/test_lammps.py
```diff
@@ -254,7 +255,7 @@ def test_LAMMPSDCDReader_set_dt(self, u, dt=1500.):
         assert_almost_equal(u.trajectory[iframe].time, iframe * dt,
                             err_msg="setting time step dt={0} failed: "
                                     "actually used dt={1}".format(
-                                dt, u.trajectory._ts_kwargs['dt']))
+            dt, u.trajectory._ts_kwargs['dt']))
```
IAlibay, line 258: Not that it matters, but I'm not sure this doesn't make the visual indentation worse 🤔

testsuite/MDAnalysisTests/coordinates/test_lammps.py
```diff
@@ -507,3 +509,117 @@ def test_atom_reordering(self, u, reference_positions):
...
+
+@pytest.mark.parametrize("convention,result",
+                         [("auto", "unscaled"), ("unscaled", "unscaled"),
+                          ("scaled", "scaled"), ("unwrapped", "unwrapped"),
+                          ("scaled_unwrapped", "scaled_unwrapped")])
+def test_open_all_convention(convention, result):
+    u = mda.Universe(LAMMPSDUMP_allcoords, format='LAMMPSDUMP',
+                     lammps_coordinate_convention=convention)
+    assert(u.trajectory.lammps_coordinate_convention == result)
+
```
IAlibay, line 537:
```suggestion


```

testsuite/MDAnalysisTests/coordinates/test_lammps.py
```diff
@@ -507,3 +509,117 @@ def test_atom_reordering(self, u, reference_positions):
...
+def test_no_coordinate_info():
+    with pytest.raises(ValueError,match="no coordinate information detected"):
```
IAlibay, line 539:
```suggestion
    with pytest.raises(ValueError, match="no coordinate information detected"):
```

testsuite/MDAnalysisTests/coordinates/test_lammps.py
```diff
@@ -507,3 +509,117 @@ def test_atom_reordering(self, u, reference_positions):
...
+        u = mda.Universe(LAMMPSDUMP_nocoords, format='LAMMPSDUMP',
+                     lammps_coordinate_convention="auto")
```
IAlibay, line 541:
```suggestion
                         lammps_coordinate_convention="auto")
```

testsuite/MDAnalysisTests/coordinates/test_lammps.py
```diff
@@ -507,3 +509,117 @@ def test_atom_reordering(self, u, reference_positions):
...
+
```
IAlibay, line 542:
```suggestion

```

IAlibay (changes requested):
Still quite a lot of PEP8 violations, I've highlighted a few here.

hmacdope:
note @IAlibay that I have left the PEP8 violations in for `datafiles.py` as it seems to be the standard in that file. LMK if you would prefer it indented. :)

package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -552,16 +579,51 @@ def _read_next_timestep(self):
...
+        # coord_dict keys = convention keys = column data or False
+        coord_dict = {conv: False for conv in self._conventions[1:]}
+        coord_dict['unscaled'] = self._get_column(
+            ["x", "y", "z"], col_ids)  # unscaled
+        coord_dict['scaled'] = self._get_column(
+            ["xs", "ys", "zs"], col_ids)  # scaled
+        coord_dict["unwrapped"] = self._get_column(
+            ["xu", "yu", "zu"], col_ids)  # unwrapped
+        coord_dict["scaled_unwrapped"] = self._get_column(
+            ["xsu", "ysu", "zsu"], col_ids)  # scaled unwrapped
```
lilyminium, line 597: Does this one-liner need to be a static method, accessible to users? And does `coord_dict` need the "un/scaled/_unwrapped" labels if it's None? How about using a for loop and making it clear what the column names are this function, like

```python
COORDTYPE_COLUMN_NAMES = {
	"unscaled": ["x", "y", "z"],
	"scaled": ["xs", "ys", "zs"],
	...,
}
for coordtype, column_names in COORDTYPE_COLUMN_NAMES.items():
	if all(axis in attrs for axis in column_names):
		coord_dict[coordtype] = [attrs[axis] for axis in column_names]
```

package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -449,19 +452,39 @@ def write(self, selection, frame=None):
...
         super(DumpReader, self).__init__(filename, **kwargs)
 
         root, ext = os.path.splitext(self.filename)
+        if lammps_coordinate_convention in self._conventions:
+            self.lammps_coordinate_convention = lammps_coordinate_convention
+        else:
+            raise ValueError("coordinate convention"
+                             f"{lammps_coordinate_convention} incorrectly "
+                             "specified, please specify one of auto, unscaled"
+                             "scaled, unwrapped or scaled_unwrapped")
```
lilyminium, line 486: I think this will result in "coordinate **conventionBADCONVENTION** incorrectly specified, please specify one of auto, **unscaledscaled**, unwrapped or scaled_unwrapped"

I bolded the missing spaces. Could you please also mention the `lammps_coordinate_convention` keyword in the error message and make it clear that the options are strings, like the below?

```python
			option_string = "'auto', '" + "', '".join(COORDTYPE_COLUMN_NAMES) + "'"
            raise ValueError("lammps_coordinate_convention="
							 f"'{lammps_coordinate_convention}' is not a valid option. "
							 f"Please choose one of {option_string}"
```

package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -552,16 +579,51 @@ def _read_next_timestep(self):
...
+        coord_dict["unwrapped"] = self._get_column(
+            ["xu", "yu", "zu"], col_ids)  # unwrapped
+        coord_dict["scaled_unwrapped"] = self._get_column(
+            ["xsu", "ysu", "zsu"], col_ids)  # scaled unwrapped
+
+        # this should only trigger on first read of "ATOM" card, after which it
+        # is fixed to the guessed value. Auto proceeds unscaled -> scaled
+        # -> unwrapped -> scaled_unwrapped
+        if self.lammps_coordinate_convention == "auto":
+            match = [var for var in self._conventions[1:] if coord_dict[var]]
```
lilyminium, line 603: If you order `COORDTYPE_COLUMN_NAMES` in the desired checking order, you could take the resulting `list(coord_dict)[0]` instead.

package/MDAnalysis/coordinates/LAMMPS.py
hmacdope: It was inline, richard suggested a lambda and oliver suggested a staticmethod, although I may have misinterpreted. Happy to condense it down, but hopefully we can settle on this one.

package/MDAnalysis/coordinates/LAMMPS.py
hmacdope: self._conventions is ordered in the correct checking order as is. :)

package/MDAnalysis/coordinates/LAMMPS.py
hmacdope: okay :+1:

package/MDAnalysis/coordinates/LAMMPS.py
lilyminium: Yes, I just meant that it didn't need to exist if you use the dict I suggested.

hmacdope: Ah I see.

hmacdope:
I think I have addressed everyone's comments :)

hmacdope:
Do people have any further comments? :)

package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -552,16 +582,46 @@ def _read_next_timestep(self):
...
+        atomline = f.readline()  # ITEM ATOMS etc
+        attrs = atomline.split()[2:]  # attributes on coordinate line
+        col_ids = {attr: i for i, attr in enumerate(attrs)}  # column numbers
+        keys = col_ids.keys()
+        # check for ids and what type of coordinate convention
+        ids = "id" in keys
+        # coord_dict keys = convention keys = column data or False
+        coord_dict = {conv: False for conv in self._conventions[1:]}
+        for coordtype, col_names in self._coordtype_column_names.items():
+            if all(axis in col_ids for axis in col_names):
```
IAlibay, line 594: What happens if it's partially missing? Do we just ignore those entries? Should we be raising an error or is the standard allowing for this?

IAlibay: Can the order be reversed or badly arranged? What do we do if it's y/z/x instead of x/y/z?

IAlibay (changes requested):
I think we're nearly there, but I'm a little bit hesitant since I'm not sure how modular the standard is, can coordinates be in the wrong order? Or does it not matter - does the way we loop through coord_cols on setting positions enforce ordering?

@MDAnalysis/coredevs if you have a chance to look here, maybe we can sneak this into 2.0.

package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -449,19 +452,46 @@ def write(self, selection, frame=None):
 class DumpReader(base.ReaderBase):
     """Reads the default `LAMMPS dump format`_
 
-    Expects trajectories produced by the default 'atom' style dump.
+    Supports coordinates in the LAMMPS "unscaled" (x,y,z), "scaled" (xs,ys,zs),
+    "unwrapped" (xu,yu,zu) and "scaled_unwrapped" (xsu,ysu,zsu) coordinate
+    conventions (see https://docs.lammps.org/dump.html for more details).
+    If no coordinate convention is provided, or if "auto" is selected,
```
lilyminium, line 458: Then why have both, why not just have `None` guess the order?

package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -449,19 +452,46 @@ def write(self, selection, frame=None):
...
+    one will be guessed. Auto proceeds in the order "unscaled",
```
lilyminium, line 459:
```suggestion
    one will be guessed. Guessing checks whether the coordinates fit each convention in the order "unscaled",
```

package/MDAnalysis/coordinates/LAMMPS.py
lilyminium: I don't think the code actually supports `None`, by the way -- it checks if the variable is in a list of strings. Perhaps

```suggestion
    If `lammps_coordinate_convention='auto'` (default),
```

package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -552,16 +582,46 @@ def _read_next_timestep(self):
...
+        attrs = atomline.split()[2:]  # attributes on coordinate line
+        col_ids = {attr: i for i, attr in enumerate(attrs)}  # column numbers
+        keys = col_ids.keys()
+        # check for ids and what type of coordinate convention
+        ids = "id" in keys
+        # coord_dict keys = convention keys = column data or False
+        coord_dict = {conv: False for conv in self._conventions[1:]}
+        for coordtype, col_names in self._coordtype_column_names.items():
+            if all(axis in col_ids for axis in col_names):
+                coord_dict[coordtype] = [col_ids[axis] for axis in col_names]
```
lilyminium, line 595: This seems quite convoluted to me, and I suspect @IAlibay, given his question. Perhaps you could simplify it and add more explicit variable names, and define the variables where they are locally used? You could only populate the dictionary for column indices if they are found, instead of setting them to False to start with. e.g. (just a suggestion):

```python
attr_to_col_ix = {x: i for i, x in enumerate(attrs)}
convention_to_col_ix = {}
for cv_name, cv_col_names in self._coordtype_column_names.items():
    try:
        convention_to_col_ix[cv_name] = [attr_to_col_ix[x] for x in cv_col_names]
    except KeyError:
        pass
if self.lammps_coordinate_convention == "auto":
    try:
        # this will automatically select in order of priority
        # as defined in the original coordtype_column_names dict
        self.lammps_coordinate_convention = list(convention_to_col_ix)[0]
    except IndexError:
        raise ValueError("No coordinate information detected")
elif not self.lammps_coordinate_convention in convention_to_col_ix:
    raise ValueError(f"No coordinates following {self.lammps_coordinate_convention} found in timestep")

coord_cols = convention_to_col_ix[self.lammps_coordinate_convention]
```

package/MDAnalysis/coordinates/LAMMPS.py
lilyminium: The column indices means that the order is agnostic. What this code does is:
* ignore the convention if it's partially missing (e.g. only y and z are provided)
* store the order of the columns in `coord_dict`
* which then gets moved to `coord_cols`.

package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -552,16 +582,46 @@ def _read_next_timestep(self):
...
+                f"{self.lammps_coordinate_convention} present in timestep")
+        else:
+            coord_cols = coord_dict[self.lammps_coordinate_convention]
 
-            indices[i] = idx
-            ts.positions[i] = xs, ys, zs
+        for i in range(self.n_atoms):
+            fields = f.readline().split()
+            if ids:
+                indices[i] = fields[col_ids["id"]]
```
lilyminium, line 617: This seems to be the only use of the `ids` variable, and it's far enough below that it's easy to forget what `ids` is by the time you get here. It's also a little convoluted in checking. Could you just not use the `ids` variable?

```python
try:
    indices[i] = attr_to_col_ix["id"]
except KeyError:
    pass
```

or 

```python
if "id" in col_ids:
   indices[i] = fields[col_ids["id"]]
```

IAlibay: I'd probably vote for setting id just above the `for` loop? (just to avoid the need to do lots of `KeyError`s if it's missing or doing the extra `in` operation for natom times.

package/MDAnalysis/coordinates/LAMMPS.py
IAlibay: That sounds sensible then.

IAlibay (approved):
lgtm

IAlibay:
@richardjgowers do you want to review / approve since you still have a blocking review?

richardjgowers (approved):

PR #3368, opened 2021-07-25
Add asunique, change unique, change atom selection behaviour

### LINKED ISSUE #2977: MDAnalysis.analysis.align.alignto won't reindex the atomgroup as is documented

### LINKED ISSUE #3364: groups.unique does not sort the group
## Expected behavior ##

Related to https://github.com/MDAnalysis/mdanalysis/issues/2977#issuecomment-882558185

The documentation implies that `atomgroup.unique` will return a sorted atomgroup, so I expect it to. However, there is a confusing contradiction in the documentation.

```python
    def unique(self):
        """An :class:`AtomGroup` containing sorted and unique
        :class:`Atoms<Atom>` only.

        If the :class:`AtomGroup` is unique, this is the group itself.
```

## Actual behavior ##

If the atomgroup is already unique, as per docs, the original *unsorted* atomgroup is returned. @coredevs is this the desired behaviour? Should we change it or add a new `sorted` attribute?

The `.unique` attribute is primarily used in `select_atoms` and wrapping-related methods, i.e. `bsphere`, `wrap`, `unwrap`, `translate`, and `rotate`. I don't think being ordered is important in any of those -- but being sorted is probably important to `select_atoms`? Are any users relying on `unique` being sorted, or conversely not being sorted?

```python
>>> import MDAnalysis as mda
>>> u = mda.Universe(PDB)
>>> ag = u.atoms[[2, 1, 0]]
>>> ag.unique.ix
array([2, 1, 0])
```

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) 2.0.0-dev0
- Which operating system?

lilyminium:
Fixes #3364
Fixes #2977 

 - Add test for current failure
 - Add .asunique. This function (well, ._asunique) now does all the grunt work in making unique groups and caching
 - Change .unique to always return a copy of the sorted unique group
 - Add .issorted
 - .unique is no longer cached. _cache['unique'] no longer exists
 - .sorted_unique and .unsorted_unique are now cached
 - .unsorted_unique and .sorted_unique can be the same group, which can be self
 - Added .cache_properties to save on carpal tunnel
 - uniqueness and sorted properties now propagate to children. They did not use to (see example code). The new function is called `_set_unique_caches_from`.

```python
sg = u.segments[[0, 0]]
assert not sg.isunique
res = sg.residues
assert not res.isunique  # old behaviour: isunique did not get cached in the child group
```

Demo:

```python
>>> import MDAnalysis as mda
>>> u = mda.Universe(PDB)
>>> ag = u.atoms[:5]
>>> ag_unique = ag[[4, 0, 3, 1, 2]]
>>> ag_dupe = ag[[4, 4, 0, 4, 3, 1, 4, 1, 2, 3]]
>>> ag_unique.ix
array([4, 0, 3, 1, 2])
>>> ag_unique.select_atoms("prop index > 2").ix
array([3, 4])
>>> ag_unique.select_atoms("prop index > 2", sort=False).ix
array([4, 3])
>>> ag_dupe.select_atoms("prop index > 2").ix
array([3, 4])
>>> ag_dupe.select_atoms("prop index > 2", sort=False).ix
array([4, 3])
>>> ag_unique.unique is ag_unique
False
>>> ag_unique.asunique() is ag_unique
True
>>> ag_unique.asunique(sorted=True) is ag_unique
False
>>> ag_dupe.asunique().ix
array([4, 0, 3, 1, 2])
>>> ag_dupe.asunique(sorted=True).ix
array([0, 1, 2, 3, 4])
>>> ag_dupe.issorted
False
>>> ag_unique.issorted
False
>>> ag_unique.unique.issorted
True
```

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


testsuite/MDAnalysisTests/core/test_atomselections.py
```diff
@@ -1363,3 +1363,13 @@ def test_error_selection_for_strange_dtype():
     with pytest.raises(ValueError, match="No base class defined for dtype"):
         MDAnalysis.core.selection.gen_selection_class("star", "stars",
                                                       dict, "atom")
+
+@pytest.mark.parametrize("sel, ix", [
+    ("name N", [5, 335, 451]),
+    ("resname GLU", [5, 6, 7, 8, 335, 451]),
+])
+def test_default_selection_on_ordered_unique_group(u_pdb_icodes, sel, ix):
+    # ordered ag => unordered again
```
zemanj, line 1372: Somehow I don't get what that comment means...

testsuite/MDAnalysisTests/core/test_atomselections.py
```diff
@@ -1363,3 +1363,13 @@ def test_error_selection_for_strange_dtype():
...
+    base_ag = u_pdb_icodes.atoms[[335, 5, 451, 8, 7, 6]]
+    ag = base_ag.select_atoms(sel)
+    assert_equal(ag.ix, ix)
```
zemanj, line 1375: Not sure if we usually put newlines at the EOF

lilyminium: We do, I just wait for PEP8 to complain at me before I do that kind of thing though.

testsuite/MDAnalysisTests/core/test_atomselections.py
lilyminium: Just reminding myself what the test does: it takes a manually ordered atomgroup and sorts it again. I can get rid of it

lilyminium:
I'm going to wait for tests to fail before I add the new code in, but preemptively, I might change `isunique` to use `bincount` because I think it's faster and fewer lines of code:

Screen Shot 2021-07-25 at 2 27 16 pm

testsuite/MDAnalysisTests/core/test_atomselections.py
zemanj: Thanks, now I got it! :smile:
I (wrongly) interpreted "ordered" as synonymous with "sorted".

package/MDAnalysis/core/groups.py
```diff
@@ -3566,21 +3609,50 @@ def unique(self):
...
+
+
+        .. versionadded:: 2.0.0
+        """
+        if self.isunique:
+            if not sorted or self.issorted:
+                return self
+        if sorted:
+            return self.unique
+        _unique = self.universe.residues[unique_int_1d_unsorted(self.ix)]
```
lilyminium, line 3655: This gets really repetitive but unless we want to give up different docstrings, probably can't fix it. Otherwise a `._universe_group` property would be *very* handy.

package/MDAnalysis/core/selection.py
```diff
@@ -172,30 +172,34 @@ def __init__(cls, name, bases, classdict):
...
 
 class LogicOperation(object, metaclass=_Operationmeta):
-    def __init__(self, lsel, rsel):
+    def __init__(self, lsel, rsel, parser):
         self.rsel = rsel
         self.lsel = lsel
+        self.parser = parser
+
+    def apply(self, *args, **kwargs):
+        return self._apply(*args, **kwargs).asunique(sorted=self.parser.sort)
```
lilyminium, line 181: Sorry for the large diffs here -- having `.unique` in 21 different lines seemed repetitive, so I renamed them all to `_apply` and apply the uniqueness catch in `apply()`.

zemanj:
Nice trick using ``np.bincount``, I guess it wins in the majority of relevant cases against the old implementation. :+1: 
The following is probably a corner case:
```python
import numpy as np
import MDAnalysis as mda

def original_isunique(arr):
    s_ix = np.sort(arr)
    mask = s_ix[1:] == s_ix[:-1]
    return not np.count_nonzero(mask)

def np_bincount(arr):
    return not np.count_nonzero(np.bincount(arr) > 1)

u = mda.Universe(PDB)
ix = u.select_atoms("resname ILE or name NA").ix  # most atoms with rather low ix and a few atoms with high ix
%timeit original_isunique(ix)
# 9.39 µs ± 37.1 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
%timeit np_bincount(ix)
# 51.8 µs ± 129 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

package/MDAnalysis/core/groups.py
```diff
@@ -725,6 +726,20 @@ def dimensions(self):
     def dimensions(self, dimensions):
         self.universe.trajectory.ts.dimensions = dimensions
 
+    @staticmethod
+    def _create_unique_group(_unique):
```
zemanj, line 730: I'm a bit unhappy with the function name here as it doesn't create a group but only sets its caches.
Plus you could set the ``issorted`` property here as well, I suppose.

lilyminium:
> corner case

That's a pretty plausible case, though. Using the inbuilt `unique_int_1d` might work best across each case:

```python
import MDAnalysis as mda
import numpy as np
from MDAnalysis.lib.util import unique_int_1d, unique_int_1d_unsorted

ix = u.select_atoms("protein or name NA").ix  # high + low
ix2 = u.select_atoms("protein").ix  # mostly low

def original_isunique(arr):
    s_ix = np.sort(arr)
    mask = s_ix[1:] == s_ix[:-1]
    return not np.count_nonzero(mask)

def int_method(arr):
    return not len(unique_int_1d(arr)) == len(arr)

def np_bincount(arr):
    return np.count_nonzero(np.bincount(arr) > 1)
```

Screen Shot 2021-07-25 at 6 00 00 pm

lilyminium:
@Luthaf I *think* the failing tests are because `chemfiles` isn't installing properly / is installing other things (chemfiles-lib, chemfiles-python). I'll see if pinning chemfiles to 0.10.0 fixes it, but in the meantime, do you have any insight?

package/MDAnalysis/core/groups.py
```diff
@@ -725,6 +726,20 @@ def dimensions(self):
...
+        # Since we know that _unique is a unique AtomGroup, we set its
+        # uniqueness caches from here:
+        _unique._cache['isunique'] = True
+        _unique._cache['unique'] = _unique
+        return _unique
+
+    @property
+    @cached('issorted')
+    def issorted(self):
```
zemanj, line 739: I think this can be done without sorting. Also, ``np.argmin`` has a lazy implementation that short-circuits for boolean arrays (in contrast to ``np.all``). So the following should be faster, especially for unsorted arrays (not thoroughly tested, though):
```python
def issorted(self):
    mask = self.ix[:-1] <= self.ix[1:]
    try:
        return mask[0] and mask.argmin() == 0
    except IndexError:
        # Empty atom groups are sorted, I guess...
        return True
```

IAlibay:
> @Luthaf I _think_ the failing tests are because `chemfiles` isn't installing properly / is installing other things (chemfiles-lib, chemfiles-python). I'll see if pinning chemfiles to 0.10.0 fixes it, but in the meantime, do you have any insight?

Carrying this on in #3369,

@Luthaf I suspect it's because chemfiles-python 0.10.1 hasn't been released yet, but chemfiles-lib has, so there's a mismatch between the two (basing myself on https://github.com/chemfiles/chemfiles/issues/415)?

Luthaf:
> I suspect it's because chemfiles-python 0.10.1 hasn't been released yet, but chemfiles-lib has, so there's a mismatch between the two (basing myself on chemfiles/chemfiles#415)?

Yes, exactly. This should not have happened, but it looks like chemfiles-python is too restrictive in the version it looks for (version 0.10.0 should be fine with chemfiles-lib 0.10.1). I'll look into this, but in the meantime I just merged the update for chemfiles-python to 0.10.1 (https://github.com/conda-forge/chemfiles-python-feedstock/pull/17), it should be avaialable in ~30 min

Luthaf:
Took a while (~1h30) to propagate in conda, but 0.10.1 should be available from both conda & pip now!

IAlibay:
> Took a while (~1h30) to propagate in conda, but 0.10.1 should be available from both conda & pip now!

Awesome thanks!

IAlibay:
fingers crossed if you unpin chemfiles & update against `develop` CI _should_ be fixed... 🙏🏽

package/MDAnalysis/core/groups.py
lilyminium: Ah, clever, thank you! If anyone else was wondering whether `np.any` short-circuits with an inverted mask: oddly, apparently not.

```python
>>> x = np.zeros(100000, dtype=bool)
>>> timeit.timeit(x.any)
3.772446749999972
>>> timeit.timeit(lambda: x[x.argmax()])
2.405966604000241
>>> x[1] = True
>>> timeit.timeit(x.any)
3.6103378889997657
>>> timeit.timeit(lambda: x[x.argmax()])
0.4463869180003712
```

lilyminium:
There was an inherent contradiction between caching `unique` and consistently returning a copy for `.unique`. So, now, `unsorted_unique` and `sorted_unique` are cached and can be `self`. `.unique` is no longer cached but always returns `sorted_unique[:]`. There might be less disruptive approaches.

package/MDAnalysis/core/groups.py
```diff
@@ -745,21 +767,73 @@ def isunique(self):
...
+                return self
+            if self.issorted:
+                self._cache['unsorted_unique'] = self
+                self._cache['sorted_unique'] = self
+                return self
+
+        if sorted:
+            if 'unsorted_unique' in self._cache and self.unsorted_unique.issorted:
+                self._cache['sorted_unique'] = self.unsorted_unique
+                return self.unsorted_unique
```
orbeckst, line 800: With all this complicated logic of what's cached, sorted, unique etc it would be good to have this `if` clause tested, too.

package/MDAnalysis/core/groups.py
```diff
@@ -745,21 +767,73 @@ def isunique(self):
...
+            mask = np.zeros_like(self.ix)
+            for i, x in enumerate(indices):
+                values = np.where(self.ix == x)[0]
+                mask[values] = i
+            self._unique_restore_mask = mask
+
+        issorted = int_array_is_sorted(indices)
+        if issorted and 'sorted_unique' in self._cache:
+            self._cache['unsorted_unique'] = self.sorted_unique
+            return self.sorted_unique
```
orbeckst, line 827: test?

Or explicitly `pragma: nocover` (or whatever it was) to be explicit with code that might bite us later...

package/MDAnalysis/core/groups.py
```diff
@@ -1581,7 +1655,7 @@ def wrap(self, compound="atoms", center="com", box=None, inplace=True):
         # atoms:
         atoms = self.atoms
         if not self.isunique:
-            _atoms = atoms.unique
+            _atoms = atoms.sorted_unique
```
orbeckst, line 1658: State explicitly in the docs that atoms will be sorted? (Makes sense to me.)

Btw, can you please fix the docs a few lines above
```reST
:meth:`MDanalysis.lib.distances.apply_PBC`
```
to
```reST
:meth:`MDAnalysis.lib.distances.apply_PBC`
```

package/MDAnalysis/core/groups.py
```diff
@@ -2561,24 +2643,52 @@ def unique(self):
...
+        group = self.sorted_unique[:]
+        group.cache_properties(isunique=True, issorted=True,
+                               sorted_unique=group, unsorted_unique=group)
+        return group
+
+    def asunique(self, sorted=False):
+        """Return a :class:`AtomGroup` containing unique
+        :class:`Atoms<Atom>` only, with optional sorting.
+
+        If the :class:`AtomGroup` is unique, this is the group itself.
```
orbeckst, line 2669: add *Parameters* section for *sorted : bool* and the *Returns*.

package/MDAnalysis/core/groups.py
```diff
@@ -2561,24 +2643,52 @@ def unique(self):
...
+           >>> ag2.ix
+           array([2, 1, 0])
+           >>> ag3 = ag.asunique(sorted=True)
+           >>> ag3 is ag
+           False
+           >>> ag3.ix
+           array([0, 1, 2])
+           >>> u.atoms[[2, 1, 1, 0, 1]].asunique(sorted=False)
+           array([2, 1, 0])
+
```
orbeckst, line 2687: Needs two blank lines?

package/MDAnalysis/core/groups.py
```diff
@@ -3566,23 +3671,48 @@ def unique(self):
...
+        group.cache_properties(isunique=True, issorted=True,
+                               sorted_unique=group, unsorted_unique=group)
+        return group
+
+
+    def asunique(self, sorted=False):
+        """Return a :class:`ResidueGroup` containing unique
+        :class:`Residues<Residue>` only, with optional sorting.
+
+        If the :class:`ResidueGroup` is unique, this is the group itself.
```
orbeckst, line 3694: add doc sections as above

package/MDAnalysis/core/groups.py
```diff
@@ -3712,22 +3829,48 @@ def unique(self):
...
+        group.cache_properties(isunique=True, issorted=True,
+                               sorted_unique=group, unsorted_unique=group)
+        return group
+
+
+    def asunique(self, sorted=False):
+        """Return a :class:`ResidueGroup` containing unique
+        :class:`Residues<Residue>` only, with optional sorting.
+
+        If the :class:`SegmentGroup` is unique, this is the group itself.
```
orbeckst, line 3852: doc sections as above

package/doc/sphinx/source/documentation_pages/selections.rst
```diff
@@ -33,6 +33,26 @@ Almost all the basic CHARMM selections work.
...
+    For example, the below code will return the first, second, and
+    sixth atom in ``ag``::
+
+        >>> ag = u.select_atoms("name N")
+        >>> ag2 = ag[[5, 1, 0]]
+        >>> ag3 = ag2.select_atoms("name N")
+        >>> np.all(ag3.ix == ag2.ix)
+        False
+
+    You can turn this off with the ``sort`` keyword::
```
orbeckst, line 48: You can turn off the sorting behavior with ...

orbeckst (commented):
only minor comments; I'll leave proper approval to @zemanj and others that are deeper involved in how selections etc work

IAlibay:
@lilyminium @zemanj how are we doing here? Any close to merging?

package/MDAnalysis/core/groups.py
lilyminium: I don't think it matters, as below the positions get re-ordered back to the input.

```
if not self.isunique:
	positions = positions[restore_mask]
```

richardjgowers (changes requested):

package/MDAnalysis/core/groups.py
```diff
@@ -3566,23 +3684,59 @@ def unique(self):
...
-        _unique._cache['isunique'] = True
-        _unique._cache['unique'] = _unique
-        return _unique
+        group = self.sorted_unique[:]
+        group.cache_properties(isunique=True, issorted=True,
+                               sorted_unique=group, unsorted_unique=group)
+        return group
+
+
+    def asunique(self, sorted=False):
```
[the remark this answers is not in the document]
lilyminium: `_asunique` does check/set caches for `sorted_unique` and `unsorted_unique`.

package/MDAnalysis/core/groups.py
```diff
@@ -1809,15 +1883,26 @@ def copy(self):
...
+        for i, prop in enumerate(['issorted', 'isunique'], 1):
+            try:
+                self._cache[prop] = value = other._cache[prop]
+            except KeyError:
+                pass
+            else:
+                x += (i * value)
+        if x >= 2:  # isunique=True
+            self._cache['unsorted_unique'] = self
+        if x == 3:  # isunique=True and issorted=True
```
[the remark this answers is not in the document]
lilyminium: Got rid of it in favour of explicit logic

package/MDAnalysis/core/groups.py
```diff
@@ -745,21 +767,73 @@ def isunique(self):
...
+            _unique = group[unique_ix]
+            _unique.cache_properties(isunique=True, issorted=True,
+                                     sorted_unique=_unique,
+                                     unsorted_unique=_unique)
+            self._cache['sorted_unique'] = _unique
+            return _unique
+        
+
+        indices = unique_int_1d_unsorted(self.ix)
+        if set_mask:
```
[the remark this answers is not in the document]
lilyminium: Raised #3387

richardjgowers (approved):

IAlibay:
@orbeckst @zemanj - any final comments? If not we should try to get this merged ASAP

zemanj:
I didn't check the atom selection stuff thoroughly but from my side this looks pretty solid.
@lilyminium I guess you're in the best position to decide whether to squash or not.

IAlibay (approved):
Had a rather brief scan through, but lgtm

IAlibay:
Thanks for the awesome work here @lilyminium !

PR #3369, opened 2021-07-26
Pin pytest-cov, remove coveralls

IAlibay:
Fixes CI issues seen in #3368 + remove coveralls from CI

Changes made in this Pull Request:
 - pins pytest-cov due to some deps resolution issues seen with the pip install - see #3372
 - removes unused coveralls components (completes #1980)

 
Acknowledgements
----------------------
Thanks to @Luthaf for working out the issue with the chemfiles compatibility. This is been handled upstream so no further changes are required.

PR Checklist
------------
 - Tests?
 - Docs?
 - CHANGELOG updated?
 - Issue raised/referenced?

PR #3375, opened 2021-07-29
On loading the same XTC offset in parallel (WIP)

### LINKED ISSUE #3230: Error loading XDR files with corrupted offset files
## Expected behavior ##

The offset file should be regenerated when the offset is corrupted. 

## Actual behavior ##

Unclear ValueError is raised.
```python
~/anaconda3/envs/yzhuang/lib/python3.8/site-packages/numpy/lib/npyio.py in load(file, mmap_mode, allow_pickle, fix_imports, encoding)
    442             # Try a pickle
    443             if not allow_pickle:
--> 444                 raise ValueError("Cannot load file containing pickled data "
    445                                  "when allow_pickle=False")
    446             try:

ValueError: Cannot load file containing pickled data when allow_pickle=False
```

```bash
#/usr/bin/bash
# generate an empty npz file
touch .md.xtc_offsets.npz
```
``` python
import MDAnalysis as mda

u = mda.Universe('./md.pdb', './md.xtc')
....

```

## Possible solution ##
Add an ValueError exception catch for `read_numpy_offsets`.

## Note ##
This happened when I tried to process ~100 different trajectories parallelly, but it may not be related to #1988. Might just because I killed the same job beforehand and the offset files ended up corrupted. I will update if that's the case.

yuxuanzhuang:
Fixes #3230, #1988

 - Add a `Shared_memory` (across all processes) object to monitor the status of offset building
 - Add a flag `offset_creator` to indicate whether the offset should be built on this instance.
 - Add a ValueError exception for corrupted npz file (fix #3004)

## Test Code
```python
# delete the existing offset file first

import MDAnalysis as mda
from joblib import Parallel, delayed

def read_trajectory():
    return  mda.coordinates.XTC.XTCReader('./test.xtc')
u_s = Parallel(n_jobs=12)(delayed(read_trajectory)() for i in range(12))
```

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


IAlibay:
@yuxuanzhuang thanks for working on this

Unfortunately couple of things
1. multiprocessing.shared_memory is a py3.8+ thing :(
2. I'm going to cancel the doc build, it's stalled - fixed in #3369 (updating against develop will do the trick here)

yuxuanzhuang:
Yeah I just realized it's a 3.8 thing when I checked how test went---It's a bummer. I will search if there is anything else that could achieve the same function.

orbeckst:
Can you just use file locking instead of the shared memory lock? Then that might also work in an MPI or dask-distributed context on a shared file system. IIRC, you'd make a temp directory and create the lock file inside that directory to be safe. Maybe just look into using [fasteners](https://pypi.org/project/fasteners/).

This S/O post https://stackoverflow.com/a/34124007 recommends [oslo.concurrency](https://docs.openstack.org/oslo.concurrency/latest/) which built with `fasteners`.(Generally, light-weight deps are better for MDA.)

richardjgowers:
Yeah file locking seems a more robust way to get this done.

yuxuanzhuang:
The current fix should work for this scenario with/without offset in the first place:
```python
def read_trajectory():
    return  mda.coordinates.XTC.XTCReader('./test.xtc')
u_s = Parallel(n_jobs=12)(delayed(read_trajectory)() for i in range(12))
```
Only one process can enter the  `_load_offsets`function at the same time so that no other processes will create/read the unfinished offset file. 
I am not sure what we should do if users explicitly set `XTCReader('./test.xtc', refresh_offset=True)` in the code above. Should we only refresh_offset once? I guess that should be the users' intention. Also need to check how it could be implemented.

.travis.yml
```diff
@@ -25,7 +25,7 @@ matrix:
         - conda activate base
         - conda update --yes conda
         - conda install --yes pip pytest==6.1.2 mmtf-python biopython networkx cython matplotlib-base "scipy>=1.6" griddataformats hypothesis gsd codecov -c conda-forge -c biobuilds
-        - pip install pytest-xdist threadpoolctl
+        - pip install pytest-xdist threadpoolctl fasteners
```
IAlibay: Needs to also go in the github actions yaml file.

yuxuanzhuang: Thanks! That's the one I always forget :)

orbeckst: Is there no conda package?

orbeckst: ignore me ... just saw that you included one

package/MDAnalysis/coordinates/XDR.py
```diff
@@ -82,7 +84,7 @@ def read_numpy_offsets(filename):
     """
     try:
         return {k: v for k, v in np.load(filename).items()}
-    except IOError:
+    except (ValueError, IOError):
```
orbeckst, line 89: Add comment to explain when we might see ValueError.

package/MDAnalysis/coordinates/XDR.py
```diff
@@ -181,6 +183,11 @@ def _load_offsets(self):
         """load frame offsets from file, reread them from the trajectory if that
         fails"""
```
orbeckst: add to docs that file locking is used

package/MDAnalysis/coordinates/XDR.py
```diff
@@ -181,6 +183,11 @@ def _load_offsets(self):
...
         fname = offsets_filename(self.filename)
+        lock_name = offsets_filename(self.filename,
+                                     ending='lock')
+
+        lock = fasteners.InterProcessLock(lock_name)
+        lock.acquire()
```
orbeckst: Is there a way to run this in a [`with` block as a context manager](https://fasteners.readthedocs.io/en/latest/examples.html#interprocess-locks)?

orbeckst: Or use the [decorator](https://fasteners.readthedocs.io/en/latest/api/process_lock.html#decorators) on the whole method?

IAlibay:
Quick question - is this an extra dependency that we only need for py3.7? If so we should raise an issue + add comments to remind us to switch away from it when we release a 3.8+ only version (~ 6-8 months from now I think according to NEP29)

orbeckst:
fasteners would stay; it also seems more general because it uses file locks instead of memory (which would only work on the same node)

package/MDAnalysis/coordinates/XDR.py
yuxuanzhuang: I don't think it's possible to use the decorator since the lock file path is dependent on the trajectory `filename` of each instance.

testsuite/MDAnalysisTests/parallelism/test_multiprocessing.py
```diff
@@ -125,6 +127,23 @@ def test_universe_unpickle_in_new_process():
     assert_equal(ref, res)
 
 
+def create_universe():
+    return mda.Universe(GRO, XTC)
+
+
+def test_creating_multiple_universe_without_offset():
```
yuxuanzhuang: The tested XTC file is way too short to induce a race scenario here (at least on my computer).

package/MDAnalysis/coordinates/XDR.py
```diff
@@ -179,53 +185,67 @@ def close(self):
...
-            # we tripped over some old offset formated file
-            pass
-
-        if not (ctime_ok and size_ok and n_atoms_ok):
-            warnings.warn("Reload offsets from trajectory\n "
-                          "ctime or size or n_atoms did not match")
-            self._read_offsets(store=True)
-        else:
-            self._xdr.set_offsets(data['offsets'])
+            filelock = fasteners.InterProcessLock(lock_name)
```
yuxuanzhuang: This part of the code looks rather ugly for the condition when the lock file cannot be created. Looking for advice on what we should do. For now, I am just creating a lock file inside `tmp` folder.

package/setup.py
```diff
@@ -597,6 +597,7 @@ def long_description(readme):
           'matplotlib>=1.5.1',
           'tqdm>=4.43.0',
           'threadpoolctl',
+          'fasteners'
```
orbeckst: add a comma to reduce noise in the next diff ;-)

package/CHANGELOG
```diff
@@ -139,6 +139,10 @@ Fixes
     would create a test artifact (Issue #2979, PR #2981)
   * Fix syntax warning over comparison of literals using is (Issue #3066)
   * new `Results` class now can be pickled/unpickled. (PR #3309)
+  * Fix the competition when generating offsets from multiple processes by
+    using a interprocess file lock. (Issue #1988, PR #3375)
+  * Fix the error when trying to load a corrupted offset file (Issue #3230
+    PR #3375)
```
orbeckst: This needs to go into the 2.1.0 section (together with your name).

package/MDAnalysis/coordinates/XDR.py
orbeckst: I think if we cannot write to the same directory as the trajectory then we could also not write to the offsets file.

In this case we should just treat it as if we cannot do anything with the offset file and do the slow "read and build an in-memory offset".

The problems with choosing /tmp as a fall back
* will not work over multiple nodes (/tmp are probably node-local)
* preferred location of /tmp might actually be in $TMPDIR (some clusters)
* windows?????

If you want to use a fallback at all then you need to find a save way to find the canonical tmp directory, e.g., whatever `tmpfile` is doing.

package/MDAnalysis/coordinates/XDR.py
```diff
@@ -179,53 +185,67 @@ def close(self):
...
+            filelock.acquire()
+            filelock.release()
+        except PermissionError:
+            head, tail = split(self.filename)
+            filelock = fasteners.InterProcessLock('/tmp/' + tail + '.lock')
+
+        with filelock:
+            if not isfile(fname):
```
orbeckst, line 205: Stupid question: what can `fname` be if not a file?

I know that this is old code but I don't understand it.

package/MDAnalysis/coordinates/XDR.py
```diff
@@ -179,53 +185,67 @@ def close(self):
...
+            # if offsets file read correctly, data will be a dictionary of offsets
+            # if not, data will be False
+            # if False, offsets should be read from the trajectory
+            # this warning can be avoided by loading Universe like:
+            # u = mda.Universe(data.TPR, data.XTC, refresh_offsets=True)
+            # refer to Issue #1893
+            data = read_numpy_offsets(fname)
+            if not data:
+                warnings.warn("Reading offsets from {} failed, "
+                              "reading offsets from trajectory instead\n"
```
orbeckst: add period at end before newline

package/MDAnalysis/coordinates/XDR.py
```diff
@@ -179,53 +185,67 @@ def close(self):
...
+                              "Consider setting 'refresh_offsets=True' "
+                              "when loading your Universe".format(fname))
```
orbeckst: add trailing period

package/MDAnalysis/coordinates/XDR.py
```diff
@@ -179,53 +185,67 @@ def close(self):
...
-                          "ctime or size or n_atoms did not match")
-            self._read_offsets(store=True)
-        else:
-            self._xdr.set_offsets(data['offsets'])
+            filelock = fasteners.InterProcessLock(lock_name)
+            filelock.acquire()
+            filelock.release()
+        except PermissionError:
+            head, tail = split(self.filename)
+            filelock = fasteners.InterProcessLock('/tmp/' + tail + '.lock')
```
orbeckst: I would replace with
```python
warnings.warn(f"Cannot write lock/offset file in same location as {self.filename}. Using slow offset calculation.")
self._read_offsets(store=True)
return
```

package/MDAnalysis/coordinates/XDR.py
```diff
@@ -179,53 +185,67 @@ def close(self):
...
-
-        if not (ctime_ok and size_ok and n_atoms_ok):
-            warnings.warn("Reload offsets from trajectory\n "
-                          "ctime or size or n_atoms did not match")
-            self._read_offsets(store=True)
-        else:
-            self._xdr.set_offsets(data['offsets'])
+            filelock = fasteners.InterProcessLock(lock_name)
+            filelock.acquire()
+            filelock.release()
```
orbeckst: Does this need to be done inside the try/except PermissionError, i.e., when do we know that we cannot create the lock?

Is there a chance to kill the process during `filelock.acquire()` and never release? Wouldn't it be safer to test with
```python
with filelock:
   pass
```
which will ensure that the lock is released.

testsuite/MDAnalysisTests/coordinates/test_xdr.py
```diff
@@ -819,6 +820,14 @@ def test_persistent_offsets_readonly(self, tmpdir):
         self._reader(filename)
         assert_equal(os.path.exists(XDR.offsets_filename(filename)), False)
 
+        # test if the offset lock file is created in tmp folder
+        head, tail = split(filename)
+        assert_equal(os.path.exists('/tmp/' + tail + '.lock'), True)
```
orbeckst: See comments on creating fall back locks in /tmp.

testsuite/MDAnalysisTests/coordinates/test_xdr.py
```diff
@@ -819,6 +820,14 @@ def test_persistent_offsets_readonly(self, tmpdir):
...
+
+    def test_offset_lock_created(self):
+        assert_equal(os.path.exists(XDR.offsets_filename(self.filename,
```
orbeckst: just use
```python
assert os.path.exists(XDR.offsets_filename(self.filename, ending='lock'))
```

testsuite/MDAnalysisTests/parallelism/test_multiprocessing.py
orbeckst: What does this test currently assert? It should `assert` something.

orbeckst: Can you think of a way to trigger the condition?  Longer file (can be created on the fly) or many more simultaneous processes? 

It might always be a flaky test that does not reliably trigger.

package/MDAnalysis/coordinates/XDR.py
yuxuanzhuang: I assume it checks the existence of `fname` :)

yuxuanzhuang:
Thanks for the review.

> The race condition test is difficult. Maybe someone else has an idea how to use mock.patch creatively to change the XDR code in one instance to stall?

I could look into it.

Also adding a test in terms of #3230.

package/MDAnalysis/coordinates/XDR.py
orbeckst: Ok... I would have used [os.path.exists()](https://docs.python.org/3/library/os.path.html#os.path.exists)  — but thanks for reminding me. No need for any changes here.

testsuite/MDAnalysisTests/parallelism/test_multiprocessing.py
```diff
@@ -125,6 +127,23 @@ def test_universe_unpickle_in_new_process():
...
+    #  test if they can be created without generating
+    #  the offset simultaneously.
+    try:
+        os.remove(XDR.offsets_filename(XTC))
+    except OSError:
+        pass
+    p = multiprocessing.Pool(2)
+    _ = [p.apply(create_universe)
+            for i in range(3)]
+    p.close()
```
orbeckst: assert something

orbeckst (changes requested):
Looks really good. One of the tests still needs to assert something.

There's still the question how to test the race condition. If you can't think of a good way to test the latter then add some comments to both the code and the tests, explaining what the code should do, and that it currently is not tested, and reference the issue. In this way, if  it turns out that the problem persists, we can reopen issues and know where to look.

richardjgowers (changes requested):
This looks like the kind of test that needs `mock` to fake the error that a race condition would trigger, to check that our recovery is working,

orbeckst:
@richardjgowers you're the *King of Mocks*, do you have a rough suggestion how to go about it?
Mock Tutrle

orbeckst:
Ping @richardjgowers for any ideas how to test this situation.

yuxuanzhuang:
Sorry. I dabbled into the mocking world a while ago, but got distracted and didn't update the PR.

Currently, I [test](https://github.com/MDAnalysis/mdanalysis/pull/3375/files#diff-1ab5d1d68ab899054f45ffa165fe389468f2c1d412f4886dd1ffbce2f1491381R774) the racing condition by mocking **ValueError** in `numpy.load`, which was the error raised in the racing condition before. 

I still leave the testing in multiprocessing there (and add a related comment) even though it does not really raise any error. Let me know it does not make sense to include some meaningless test, then I could remove it:)

testsuite/MDAnalysisTests/parallelism/test_multiprocessing.py
orbeckst: Is this test useful if it does not normally trigger the problem?

testsuite/MDAnalysisTests/parallelism/test_multiprocessing.py
```diff
@@ -134,15 +134,18 @@ def create_universe():
 def test_creating_multiple_universe_without_offset():
     #  test if they can be created without generating
     #  the offset simultaneously.
+    #  The tested XTC file is way too short to induce a race scenario
     try:
         os.remove(XDR.offsets_filename(XTC))
```
orbeckst: This remove interferes with other tests that may work with the XTC trajectory. You'd need to make a temporary copy first for this test.

testsuite/MDAnalysisTests/coordinates/test_xdr.py
```diff
@@ -750,7 +750,16 @@ def test_nonexistant_offsets_file(self, traj):
             saved_offsets = XDR.read_numpy_offsets(outfile_offsets)
             assert_equal(saved_offsets, False)
 
-    def test_reload_offsets_if_offsets_readin_fails(self, trajectory):
+    def test_nonexistant_offsets_file(self, traj):
```
orbeckst: spelling: nonexistent ([existant](https://en.wiktionary.org/wiki/existant) is a noun)

orbeckst (changes requested):
I like the mock tests but I also want to hear @richardjgowers 's view.

There's a test for multiprocessing `test_creating_multiple_universe_without_offset()` where you added a comment that the trajectory is too short to trigger the original bug. Did I understand this correctly? If so, then the test is not really doing anything useful and I would just remove it, unless there are other reasons for keeping it that I don't see.

Did you add `fasteners` to the github actions? I don't see any changes to our workflow files. 

(@IAlibay do we add new dependencies to CHANGELOG? I don't think so but I wanted to ask. In any case, they will need to be added to the conda meta data for conda-forge eventually.)

yuxuanzhuang:
@orbeckst  
> unless there are other reasons for keeping it that I don't see.

No, not really. Just keep it as a potential issue/reference for now.

orbeckst:
Then you could label the test as XPASS and add more commentary on why the test is kept here.

orbeckst:
As it turns out, there's no `pytest.mark.xpass`, I don't think we need to [add a custom mark](https://docs.pytest.org/en/stable/how-to/mark.html). 

I've been dabbling with the PR to address my remaining points so that we can get on with it. I will push an update and then this can be discussed.

testsuite/MDAnalysisTests/parallelism/test_multiprocessing.py
```diff
@@ -125,6 +141,21 @@ def test_universe_unpickle_in_new_process():
...
+def test_creating_multiple_universe_without_offset(temp_xtc, ncopies=3):
+    #  test if they can be created without generating
+    #  the offset simultaneously.
+    #  The tested XTC file is way too short to induce a race scenario but the
+    #  test is included as documentation for the scenario that used to create
+    #  a problem (see PR #3375 and issues #3230, #1988)
+
+    args = (GRO, str(temp_xtc))
+    with multiprocessing.Pool(2) as p:
+        universes = [p.apply(mda.Universe, args) for i in range(ncopies)]
```
orbeckst: @yuxuanzhuang why are we using [`apply`](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool.apply), which blocks, and not `apply_async`?

testsuite/MDAnalysisTests/coordinates/test_xdr.py
orbeckst: fixed in [94611d7](https://github.com/MDAnalysis/mdanalysis/pull/3375/commits/94611d79eab7d7e289410ce2649e184581a8f471)

testsuite/MDAnalysisTests/parallelism/test_multiprocessing.py
orbeckst: fixed in [94611d7](https://github.com/MDAnalysis/mdanalysis/pull/3375/commits/94611d79eab7d7e289410ce2649e184581a8f471)

orbeckst:
I force pushed to remove merges from develop (from my earlier commit) — they would mess up a squash-merge here, I think.

orbeckst (approved):

orbeckst:
@richardjgowers could you have a look if your concerns were addressed?

@IAlibay this PR adds *fasteners* as a new core dependency.

IAlibay:
> @richardjgowers could you have a look if your concerns were addressed?
> 
> @IAlibay this PR adds *fasteners* as a new core dependency.

I'm noticing it's pip installed in Travis CI, does fasteners have a conda-forge package we can use there?

Also please add to requirements.txt.

orbeckst:
@IAlibay how do you find out if there's a fasteners package suitable for ppc64le  — by looking at https://anaconda.org/conda-forge/fasteners I can't tell unless we can use noarch. (For arm64 we are apparently using pip only anyway.)

Similarly. the travis.yml also contains threadpoolctl as pip — is there a conda-forge package?

orbeckst:
> Also please add to requirements.txt.

It was already in requirements.txt, but had been added to the end. I sorted it alphabetically.

testsuite/MDAnalysisTests/parallelism/test_multiprocessing.py
orbeckst: ping @yuxuanzhuang

orbeckst:
@richardjgowers , can you please have a quick look to see if you're happy with the changes?

@IAlibay , please have a look if the CI contains everything needed.

It would be nice to have this PR merged eventually.

testsuite/MDAnalysisTests/parallelism/test_multiprocessing.py
yuxuanzhuang: Okay, I may have misunderstood `apply` as opposed to `apply_async`. I have changed accordingly.

orbeckst: Honestly, I don't really understand it myself, as I haven't really played with it. That's why I asked you ;-). Do you think that `apply_async` is the better way to go? Does it trigger the test without the code fixes?

yuxuanzhuang: I think `apply` never runs the tasks in parallel (I don't even know why it exists right now:<) It is just a wrapper of `apply_async().get()` [under the hood](https://github.com/python/cpython/blob/main/Lib/multiprocessing/pool.py#L352).

```python
import multiprocessing
import os

def f():
    time.sleep(1)
    return 0

n_works = 4

with multiprocessing.Pool(n_works) as pool:
    multiple_workers = [pool.apply_async(f, ()) for i in range(n_works)]
    print([res.get() for res in multiple_workers])
1.7 s

with multiprocessing.Pool(n_works) as pool:
    print([pool.apply(f, ()) for i in range(n_works)])
4.1 s
```
So the answer is `apply_async` is definitely what we want to test here. It also doesn't trigger the test when the trajectory is short.

orbeckst: Thank you for doing the research. I am very happy with the change then.

orbeckst:
ping @richardjgowers and @IAlibay — this is really just needs a quick glance from you to be done. Thanks.

richardjgowers (approved):
Yeah LGTM as long as you're happy with the new dependency

IAlibay:
Thanks @yuxuanzhuang ! 🎉

yuxuanzhuang:
Thanks!

PR #3376, opened 2021-07-29
Increase minimum numpy version for aarch64 CI.

### LINKED ISSUE #3374: MDA compilation crash with aarch64
For some reason our latest aarch64 builds are reproducibly failing on aarch64, see: https://travis-ci.com/github/MDAnalysis/mdanalysis/jobs/526646721#L1562-L1582

Issue seems to coincide with the release of cython 0.29.24.

@tylerjereddy sorry to bug you here, are you seeing anything similar upstream?

IAlibay:
Fixes #3374 

 - Increase minimum numpy version for aarch64 build to 1.19.2
 - Remove unused old travis file

PR Checklist
------------
 - Tests?
 - Docs?
 - CHANGELOG updated?


.travis.yml
```diff
@@ -44,7 +44,7 @@ matrix:
       group: edge
       if: type = cron
       before_install:
-        - python -m pip install cython numpy scipy
+        - python -m pip install cython "numpy>=1.19.2" scipy
```
IAlibay, line 47: We could technically pin the minimum numpy version on setup.py for aarch64, but I'm not sure it's a great idea (or at least it'll get messy soon).

I'm thinking we should just label aarch64 as experimental and then publish the tested versions, thoughts @MDAnalysis/coredevs ?

tylerjereddy: If it fixes the issue, it makes sense based on the release notes for that version of NumPy: https://github.com/numpy/numpy/releases/tag/v1.19.2

There are ARM and Cython fixes mentioned.

Actually, I just took a closer look--SciPy does the exact same NumPy version pin for `aarch64` right here: https://github.com/scipy/scipy/blob/master/pyproject.toml#L26

I guess MDA could lift some of the version pins it needs from upstream `pyproject.toml` if we adopt one at some point.

tylerjereddy (approved):

tylerjereddy:
Thanks Irfan; I'm almost tempted to try adding a `pyproject.toml` just for the `aarch64` shims (leaving other platforms alone/to `setup.py` for now) so that the source builds of MDA with i.e., `pip install .` work reliably on that platform.

The only ARM we shim we have for building (`setup.py`) at the moment is for encore compiler optimization:

```
encore_compile_args.append('-O1')
```

PR #3382, opened 2021-08-04
WIP: nearest N particle search

richardjgowers:
Fixes #

 - 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/lib/include/cubeiter.h
+++ b/package/MDAnalysis/lib/include/cubeiter.h
@@ -1,0 +1,129 @@
+#ifndef DISTOPIA_CUBEITER_H
+#define DISTOPIA_CUBEITER_H
+
+/*
+ * usage:
+ * ConcentricCubeIterator ci();
+ *
+ * ci.x, ci.y, ci.z  // all 0,0,0 at this point
+ * ci.next();  // will return false, only one cube in central shell
+ * ci.expand();  // moves onto 3x3 shell of cubes
+ * while (ci) {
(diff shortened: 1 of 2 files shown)


richardjgowers:
@hmacdope I think if you get your head round how the results container is keeping track of the "best" N results, then how the cubeiter works the actual search function isn't that nasty tbh. It's currently failing because of a double free somewhere, so I think I've got my cython wrong somewhere but I'll track that down.

cbouy:
I'm a bit late to the party but doesn't the [KDTree](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.cKDTree.html) [query](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.cKDTree.query.html) from scipy already do what you need?

richardjgowers:
Yes, but KDTree won’t (ever) do periodic boundaries

package/MDAnalysis/lib/distances.py
```diff
@@ -1136,6 +1137,56 @@ def _nsgrid_capped_self(reference, max_cutoff, min_cutoff=None, box=None,
...
+    """
+    if N <= 0:
+        raise ValueError("N must be positive")
+
+    # TODO: Something clever about automatically choosing this distance
+    if grid_spacing is None:
+        grid_spacing = 4.0
+
+    if box is None:
+        raise ValueError("TODO")
```
orbeckst, line 1183: use KDTree as suggested by @cbouy ?

Issue #3394, opened 2021-08-19
LAMMPS trajectories should set atoms.names equal to element names


orioncohen:
## Is your feature request related to a problem? ##
Whenever I use MDAnalysis to handle LAMMPS trajectories, I have to overwrite the atom names and set them equal to the elements. Otherwise, when writing .xyz files, the elements are not properly assigned.

## Describe the solution you'd like ##
I would like the LAMMPS trajectory to automatically guess elements based on mass and assign them to atom.names. (and fail gracefully)

## Describe alternatives you've considered ##
Alternatively, the XYZ writer could be modified to automatically assign element types if masses are present but names are missing.

IAlibay:
As far as I know the XYZWriter now writes out elements from `atoms.elements` instead of using `atoms.names` if the former is available.

The best thing here would probably be to make it so that whatever favourite LAMMPS format feeds correct elements to `atoms.elements` instead?

Is there another more stable property we could use for element getting here? There have been many decisions over the years about "guessing elements by default", and the general view is that we want to avoid doing that as much as possible. Whilst masses can be a reasonable direct translation to elements in normal cases, in _a lot_ of edge cases they can be problematic (CG, hydrogen mass repartitioning, united atom force fields, etc...).

Tagging @lilyminium who has been part of a lot of these conversations too.

orioncohen:
Ahh ok, that is probably the case. I haven't checked out the code for the XYZWriter. Will take a look.

Got it, those are tricky. In that case, maybe assign elements to atoms if and only if all atom masses can be matched with an element?

 Or, add an optional to guess elements as a keyword argument? Currently the XYZWriter basically fails for unmodified LAMMPS trajectories, which is frustrating.

lilyminium:
Yeah, we have a lot of issues about guessing. I'm guessing LAMMPS trajectories don't have a trivial way to obtain elements, like giving particles atomic numbers? What does LAMMPS do about types, are they named in a way that makes it easy to guess elements? I notice the topology parser has a `massdict` argument for assigning masses via types, which basically sounds like the inverse of what you want to do.

orioncohen:
It may be possible to set atom types in LAMMPS that are compatible with the MDAnalysis reader (I'm no LAMMPS super-user) but in all the LAMMPS simulations I have done, LAMMPS just assigns the types as integers from 1...N. So sadly, pretty useless.

[...]

PR #3395, opened 2021-08-19
Atom naming​, issue #3394

orionarcher:
This PR aims to address issue #3394. However I am unclear on the best way to append a new attribute or what the best approach is.

So far, this PR only implements a guesser, it does not address the issue of assigning and Atom attribute.

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/topology/guessers.py
+++ b/package/MDAnalysis/topology/guessers.py
@@ -178,8 +178,9 @@ def guess_atom_type(atomname):
     return guess_atom_element(atomname)
 
 
-NUMBERS = re.compile(r'[0-9]') # match numbers
+NUMBERS = re.compile(r'[0-9]')  # match numbers
 SYMBOLS = re.compile(r'[*+-]')  # match *, +, -
+
 
 def guess_atom_element(atomname):
     """Guess the element of the atom from the name.
@@ -223,6 +224,16 @@ def guess_atom_element(atomname):
 
         # if it's numbers
         return no_symbols
+
+
+def guess_atom_element_from_mass(mass_guess, tol=0.1):
+    for element, mass in tables.masses.items():
+        if abs(mass - mass_guess) < tol:
+            return element
+    raise ValueError(f'the molecular weight: {mass_guess} amu, is not valid')
+
+
+guess_atom_elements_from_masses = np.vectorize(guess_atom_element_from_mass)
 
 
 def guess_bonds(atoms, coords, box=None, **kwargs):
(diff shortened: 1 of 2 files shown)


package/MDAnalysis/topology/guessers.py
```diff
@@ -225,6 +226,16 @@ def guess_atom_element(atomname):
...
 
 
+def guess_atom_element_from_mass(mass_guess, tol=0.1):
+    for element, mass in tables.masses.items():
+        if abs(mass - mass_guess) < tol:
+            return element
+    raise ValueError(f'the molecular weight: {mass_guess} amu, is not valid')
+
+
+guess_atom_elements_from_masses = np.vectorize(guess_atom_element_from_mass)
```
IAlibay, line 236: Whilst vectorize is nice, I believe it does not provide any actual performance improvements instead is just a for loop around the other function.

Since these are user-facing functions, I'd suggest writing a proper function call with docstring that folks can call if they need to.

package/MDAnalysis/topology/guessers.py
```diff
@@ -225,6 +226,16 @@ def guess_atom_element(atomname):
         return no_symbols
 
 
+def guess_atom_element_from_mass(mass_guess, tol=0.1):
+    for element, mass in tables.masses.items():
```
IAlibay, line 230: Doing this for every element is going to end up being pretty tedious & expensive. Is there a way we could just store a "rounded to the given precision" dictionary somewhere and just round the input value?

IAlibay (commented):
So following from my comment in #3394, I'm not against the idea of implementing a guesser based on mass. I don't think we should be applying it by default, but modifying existing readers through an additional method to TopologyReaderBase or just adding a guess_elements method to Universe are two potential options I could be on board with.

package/MDAnalysis/topology/guessers.py
richardjgowers: `int(mass)` is probably good enough for the guessing going on here.  It would be nice to get the `O(log(N))` dict lookup rather than looping through the dict.

testsuite/MDAnalysisTests/topology/test_guessers.py
```diff
@@ -81,6 +81,11 @@ def test_guess_types(self):
     def test_guess_atom_element(self):
         assert guessers.guess_atom_element('MG2+') == 'MG'
 
+    def test_guess_atom_elements_from_masses(self):
+        assert guessers.guess_atom_element_from_mass(12.01) == 'C'
```
richardjgowers, line 85: the failing case also needs to be handled.

package/MDAnalysis/topology/guessers.py
```diff
@@ -225,6 +226,16 @@ def guess_atom_element(atomname):
...
+        if abs(mass - mass_guess) < tol:
+            return element
+    raise ValueError(f'the molecular weight: {mass_guess} amu, is not valid')
```
richardjgowers, line 233: I think other guessers don't raise Errors but instead warn and return empty/default values

richardjgowers (changes requested):
Thanks for taking this on @orioncohen !

IAlibay:
@orioncohen - guess elements from mass (or just generally more guessing tbh) is something that'd be quite cool to have available in MDA. Is this something you're still willing to work on?

orionarcher:
Hi @IAlibay, thanks for the bump, this is something I'm still interested in pushing through but I will not have time to work on it until March.

orionarcher:
I no longer use LAMMPS and don't have the bandwidth to work on this right now, so closing PR. Thanks for the feedback regardless.

PR #3396, opened 2021-08-21
pre-release changes for 2.0.0

IAlibay:
Towards the 2.0.0 release

 - Updates the version to 2.0.0 and ties off the CHANGELOG for this version

Notes:
- Please hold on merge until #3189 and possibly #3360 are merged
- Once this is merged we will have a merge freeze on `develop` until we have finalised the 2.0.0 release (just so we don't have to make any changes to the code that will put `master` out of synch with `develop`).

PR Checklist
------------
 -  Tests?
 - Docs?
 - [x] CHANGELOG updated?


IAlibay:
So the plan for release is:

1) Once we're happy with having completed 2.0.0, merge this into `develop`
2) Merge `develop` into `master` (standard merge)
3) Upload `master` to test-pypi
4) test on all platforms we can
5) release on pypi
6) release on conda (assuming we can also do an aarch64 and Power9 release with that)

richardjgowers:
@IAlibay I think this commit can go onto master instead of develop, then this PR can be the commit which starts 2.x-dev0, which is probably the next commit after the head of master if that makes sense

IAlibay:
> @IAlibay I think this commit can go onto master instead of develop, then this PR can be the commit which starts 2.x-dev0, which is probably the next commit after the head of master if that makes sense

@richardjgowers I'm not sure this will work? If we do that wouldn't the `master` branch will always be 1 commit ahead of `develop`?

My thought was; 1) we merge this, 2) we merge `develop` into `master` so the two are equal, 3) once 2.0.0 passes the testpypi things, we then open a PR to quick off 2.1.0-dev0?

That was `master` and `develop` always stay the same?

richardjgowers:
@IAlibay yeah I just don't think this PR is necessary, you can merge this into master locally, do the release, then the next PR onto develop is a few commits after this one (but this commit is still in the timeline)

IAlibay:
Ok here's what I'm proposing (and have already started doing):

1) merge develop into master so they are both equal (done directly via CLI)
2) add 2.0.0 release commit to master (will close this PR and reposition it to target `master`)
3) do release
4) PR of master feature branch into develop with the 2.1.0 stuff (that way we don't end up with an extra commit on `master`)

PR #3400, opened 2021-08-22
Start 2.1.0-dev0

### LINKED ISSUE #3254: Release 2.0.0
The target release date for 2.0.0 is the 10th of May.

This means that we want a code freeze that comes into effect next Friday the 7th of May.**

Please see the following for a list of all current things that need to get done: https://github.com/MDAnalysis/mdanalysis/milestone/16

If there's anything that isn't on there, please do bring it up here and/or add it to the milestone directly.

** Note: I'm not convinced this is realistic - I would still like to advocate for the possibility of a release candidate release by the 10th so that we have some spare time to get things done.

IAlibay:
Fixes #3254

 - Starts off 2.1.0-dev0
 - Raises minimum python version to 3.7 and numpy to 1.18 in accordance to current NEP29 support. If we have an idea of when the release is going to be we can provide a more accurate minimum version here. Ideally I'd like 2.1.0 to be out faster than normal so we can add the RDKit fixes.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/core/universe.py
+++ b/package/MDAnalysis/core/universe.py
@@ -68,21 +68,9 @@ import sys
 # trigger an MPI warning because importing the uuid module triggers a call to
 # os.fork(). This happens if MPI_Init() has been called prior to importing
 # MDAnalysis. The problem is actually caused by the uuid module and not by
-# MDAnalysis itself. Python 3.7 fixes the problem. However, for Python < 3.7,
-# the uuid module works perfectly fine with os.fork() disabled during import.
-# A clean solution is therefore simply to disable os.fork() prior to importing
-# the uuid module and to re-enable it afterwards.
+# MDAnalysis itself. Python 3.7 fixes the problem.
 import os
-
-# Windows doesn't have os.fork so can ignore
-# the issue for that platform
-if sys.version_info >= (3, 7) or os.name == 'nt':
-    import uuid
-else:
-    _os_dot_fork, os.fork = os.fork, None
-    import uuid
-    os.fork = _os_dot_fork
-    del _os_dot_fork
+import uuid
 
 from .. import _TOPOLOGY_ATTRS, _PARSERS
 from ..exceptions import NoDataError
--- a/.github/workflows/gh-ci.yaml
+++ b/.github/workflows/gh-ci.yaml
@@ -35,7 +35,7 @@ jobs:
         fail-fast: false
         matrix:
           os: [ubuntu-latest, ]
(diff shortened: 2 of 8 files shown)


IAlibay:
## Important! Please do not squash-merge! This includes the `master` commits that will bring the two branches back to parity.

IAlibay:
Not really thought about what we'll do if we want to do a bugfix for 2.0.0, I guess we can just do a feature branch from `master`?

package/MDAnalysis/core/universe.py
```diff
@@ -68,21 +68,9 @@
 # trigger an MPI warning because importing the uuid module triggers a call to
 # os.fork(). This happens if MPI_Init() has been called prior to importing
 # MDAnalysis. The problem is actually caused by the uuid module and not by
-# MDAnalysis itself. Python 3.7 fixes the problem. However, for Python < 3.7,
-# the uuid module works perfectly fine with os.fork() disabled during import.
-# A clean solution is therefore simply to disable os.fork() prior to importing
-# the uuid module and to re-enable it afterwards.
+# MDAnalysis itself. Python 3.7 fixes the problem.
```
IAlibay, line 71: We could remove the whole comment if we want, but since we don't test it, it might be good to keep for historical / in case it happens again?

IAlibay:
@MDAnalysis/coredevs we need to get this moving along so that we can get dev back up.

IAlibay:
The conversation to be had here is if we want to wait to apply NEP29 until after we know we don't want a bugfix for 2.0.0? If so I can revert some of the changes here and keep 3.6 support until we drop it later?

package/MDAnalysis/core/universe.py
orbeckst: I’d keep it.

orbeckst (approved):
Let’s do this in forward looking mode. (The alternative would have been to set the version to 2.0.1 and change it as soon as a new feature goes in.)

Let’s try to make PRs focused so that we could easily backport them to a patch branch on master. Or if we only have fixed on this branch when merging into master we can just change the version information there and then change develop.

IAlibay:
> Let’s do this in forward looking mode. (The alternative would have been to set the version to 2.0.1 and change it as soon as a new feature goes in.)

Sounds good to me. If anything I'd like to get 2.1.0 out faster (especially so we could get #3044 out there).

lilyminium (approved):

Issue #3406, opened 2021-08-26
H5PY 3.4.0 causes H5MD writer tests fail with PermissionError on Windows (investigate temporary pin)


IAlibay:
## Expected behavior ##

All tests pass.

## Actual behavior ##

For the Windows runners TestH5MDWriterWithRealTrajectory.test_write_with_drivers[core] fails due to a permission error in writing the H5MD file.

## Code to reproduce the behavior ##

See: https://dev.azure.com/mdanalysis/mdanalysis/_build/results?buildId=2576&view=logs&j=af49e0b6-b01c-50c5-fac7-b346db9f76f5&t=6e9d53c4-1948-52bf-f8ab-45d3eec15dc0&l=821

orbeckst:
Do we have any inkling as to why?

@edisj ?

edisj:
I’ll be home in ~1.5 hrs, can take a look then

IAlibay:
I wonder if this is linked to the 3.4.0 release (it got released 4 days ago), no issues linked to this issue at this point though.

IAlibay:
See #3407 - seems like an issue with the 3.4.0 release of h5py. If someone has the spare capacity / time we should report it upstream.

With regards to moving forward, it's probably best to just pin to <3.4.0?

edisj:
That seems like the best short term fix to me. There was a bug fix listed here https://docs.h5py.org/en/latest/whatsnew/3.4.html relating to the `core` and `stdio` drivers and using `mode='a'`, however `H5MDWriter` only uses `mode='w'`. I looked through a few of the PRs linked and I haven't found anything to do specifically with Windows or `mode='w'`

orbeckst:
Did we ever reproduce the failure locally?

Has anyone the time to bisect h5py releases to pinpoint the ones that fail? We could exclude known bad releases for #3711 (or separately).

IAlibay:
It doesn't look like this is a thing anymore? Closing.

PR #3407, opened 2021-08-26
xfail on h5md write nt test, more py3.6->3.7 changes

IAlibay:
Temporarily addresses #3406

 - xfail on windows NT for `TestH5MDWriterWithRealTrajectory.test_write_with_drivers`
 - more py3.6 -> py3.7 changes

PR Checklist
------------
 - [x] Tests?
 - Docs?
 - CHANGELOG updated?

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -54,7 +54,7 @@ jobs:
       pytest-xdist
       scikit-learn
       scipy
-      h5py
+      "h5py<3.4.0"
       tqdm
       threadpoolctl
     displayName: 'Install dependencies'
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -26,9 +26,6 @@ jobs:
         Python37-32bit-full:
           PYTHON_VERSION: '3.7'
           PYTHON_ARCH: 'x86'
-        Python36-64bit-full:
-          PYTHON_VERSION: '3.6'
-          PYTHON_ARCH: 'x64'
         Python37-64bit-full:
           PYTHON_VERSION: '3.7'
           PYTHON_ARCH: 'x64'
--- a/package/MDAnalysis/due.py
+++ b/package/MDAnalysis/due.py
@@ -57,16 +57,7 @@ try:
     # https://github.com/MDAnalysis/mdanalysis/pull/1822#issuecomment-373009050
     import sys
     import os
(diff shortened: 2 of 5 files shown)


IAlibay:
codecov seems to be doing it's weird "fail for no reason" thing - this _should_ be fine?

IAlibay:
Let's just try pinning h5py (see if it's related to 3.4.0)

orbeckst (approved):
Please also update the setup.py metadata to remove 3.6.

Otherwise looks ok but we should have an open issue for the failures. I don't really want to have to restrict the h5py versions needlessly.

IAlibay:
@orbeckst, which metadata? I was pretty sure I had already removed it previously but I'm probably looking at the wrong thing?

Re: issue - I'm planning to keep #3406 open (probably renamed) so that we can keep an eye on h5py and un-pin as soon as we can.

orbeckst:
I didn’t look into setup.py when I wrote my comment (and still haven’t, sorry). I meant the Python version that is set in the PyPi metadata. If that’s already >= 3.7 then there’s nothing else to be done. I haven’t kept up with the details of NEP29 — is 2.1 going to be 3.7-3.9?

> Am 8/26/21 um 18:30 schrieb Irfan Alibay ***@***.***>:
> 
> @orbeckst, which metadata? I was pretty sure I had already removed it previously but I'm probably looking at the wrong thing?

IAlibay:
> I meant the Python version that is set in the PyPi metadata. 

Ah ok, yeah I deleted that in the previous PR, I had just forgotten to delete out all the things here.

>  I haven’t kept up with the details of NEP29 — is 2.1 going to be 3.7-3.9?

If we were to release today, that would be the case. If we release in 6 months, we probably would be 3.8-4.0. I'm setting it to a minimum of "what would happen if we released tomorrow", and then we can progressively push the support as time goes by?

orbeckst:
Very sensible, I agree with your approach. 

> Am 8/27/21 um 08:31 schrieb Irfan Alibay ***@***.***>:
> 
> If we were to release today, that would be the case. If we release in 6 months, we probably would be 3.8-4.0. I'm setting it to a minimum of "what would happen if we released tomorrow", and then we can progressively push the support as time goes by?

IAlibay:
Should be everything, I'll squash-merge so we don't stall other PRs.

PR #3411, opened 2021-09-01
change error message numpy version 1.16 -> 1.18

IAlibay:
Fixes https://github.com/MDAnalysis/mdanalysis/pull/3398#issuecomment-909854722

 - Change old error message numpy version

PR Checklist
------------
 - Tests?
 - Docs?
 - CHANGELOG updated?
 - Issue raised/referenced?

### DIFF
--- a/package/setup.py
+++ b/package/setup.py
@@ -189,7 +189,7 @@ def get_numpy_include():
         import numpy as np
     except ImportError:
         print('*** package "numpy" not found ***')
-        print('MDAnalysis requires a version of NumPy (>=1.16.0), even for setup.')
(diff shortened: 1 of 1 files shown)


tylerjereddy (approved):
That looks right--of course, for bleeding edge versions of Python or on some platforms that were only recently supported like ARM and Mac M1, we probably need even newer NumPy, but I suspect that can eventually be encoded in the `pyproject.toml` stuff.

tylerjereddy:
I'll just put this in instead of suggesting that we use a variable to maintain that version in a single place (vs. error message, setup reqs and install reqs locations), since writing more shim code in `setup.py` probably isn't the best solution long-term anyway.

Thanks!

PR #3415, opened 2021-09-06
Added frames processing to AnalysisBase

### LINKED ISSUE #1985: Allow analysis classes to run on arbitrary selections of frames
It would be convenient to be able to run some analyses on an arbitrary set of frames. I recently encounter a case where I wanted to build an RSMD matrix with `mda.analysis.diffusionmap.DistanceMatrix` on a selection of frame I obtained from clustering. For now, the only ways I could find to do that was to either write a new trajectory (on file or in memory), or to hack a `SelectionReader`. Adding a `frames` argument to run alonside `start`, `stop`, and `step` (see #1979) would allow to pass an array of frame indices, or a mask, or an already indexed trajectory using #1934. The interface would be the same as for `ag.write`.

```python
# API as envisioned

u = mda.Universe(TPR, XTC)
ag = u.select_atoms(...)

labels = my_clustering_method(ag)  # returns np.array( [1, 1, 0, 0, 2, ...])
frames_group0 = u.trajectory[labels == 0]

mat0 = MySuperAnalysis(ag).run(frames=frames_group0)
# or
mat0 = MySuperAnalysis(ag).run(frames=(labels == 0))

# Should prevent non-sensical calls such as mixing `frames` with `start`, `stop`, `step`.
mat0 = MySuperAnalysis(ag).run(frames=frames_group0, step=3)  # ValueError
```

Some analyses rely on the time step to be constant, so the use of the `frames` must be limited for these analyses. An slices trajectory can be used, but not an indexed one. In these case, a `NotImplementedError` should be raised. This could be achieved by using a class attribute to switch off arbitrary frame selections:

```python
class MyConstentDTAnalysis(AnalysisBase):
    needs_constant_dt = True

   ...
```

rsexton2:
Partially addresses #1985

 - Make analysis base select arbitrary frames given by an array of integers (i.e., an array of frame indices) or booleans.
 
TODO (possibly), see #1985 
- accept FrameIterators

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/analysis/base.py
```diff
@@ -281,25 +295,31 @@ def run(self, start=None, stop=None, step=None, verbose=None):
...
         step : int, optional
             number of frames to skip between each analysed frame
+        frames : array_like, optional
+            array of integers or booleans to slice trajectory
         verbose : bool, optional
             Turn on verbosity
+       
+
+        .. versionchanged:: 2.1.0
+            Added frames processing
```
IAlibay: I'd make this a bit more explicit "Added ability to iterate through trajectory by passing a list of frame indices".

package/MDAnalysis/analysis/base.py
```diff
@@ -270,7 +283,8 @@ def _conclude(self):
         """
         pass  # pylint: disable=unnecessary-pass
 
-    def run(self, start=None, stop=None, step=None, verbose=None):
+    def run(self, start=None, stop=None, step=None, verbose=None,
```
IAlibay: nitpicking, but I'd place frames before verbose here, mostly a visual thing but it looks nicer to have start/stop/step/frame bunched up together

package/MDAnalysis/analysis/base.py
```diff
@@ -238,17 +239,29 @@ def _setup_frames(self, trajectory, start=None, stop=None, step=None):
...
 
         .. versionchanged:: 1.0.0
             Added .frames and .times arrays as attributes
+
+        .. versionchanged:: 2.1.0
+            Added frames processing
         """
         self._trajectory = trajectory
-        start, stop, step = trajectory.check_slice_indices(start, stop, step)
+        if frames is not None:
```
IAlibay, line 348: Appropriate tests will be necessary here.

package/MDAnalysis/analysis/base.py
```diff
@@ -238,17 +239,29 @@ def _setup_frames(self, trajectory, start=None, stop=None, step=None):
...
+            if any([start, stop, step]):
+                raise ValueError("start/stop/step cannot be combined with frames")
+            slicer = frames
+        else:
+            start, stop, step = trajectory.check_slice_indices(start, stop, step)
+            slicer = slice(start, stop, step)
+        self._sliced_trajectory = trajectory[slicer]
```
IAlibay, line 357: I _think_ using a frameiterator changes the behaviour of the iteration.

So if I do:

```python
import MDAnalysis as mda

u = mda.Universe(TPR, XTC)

iterator = u.trajectory[[0,1,2]]

for ts in iterator:
    pass

print(u.trajectory.ts)
```
Ouput: < Timestep 2 with unit cell dimensions [79.98359 79.98359 79.98358 60.      60.      90.     ] >

Whilst:

```python
import MDAnalysis as mda

u = mda.Universe(TPR, XTC)

for ts in u.trajectory[:3]
    pass

print(u.trajectory.ts)
```
Ouput: < Timestep 0 with unit cell dimensions [80.017006 80.017006 80.017006 60.       60.       90.      ] >

This is a general thing with the way we iterate through a trajectory based on an index selection (not limited to this code here), and we should probably fix this? But I'd say it's probably a bad idea for the two approaches differ like this for AnalysisBase.

package/MDAnalysis/analysis/base.py
```diff
@@ -238,17 +239,29 @@ def _setup_frames(self, trajectory, start=None, stop=None, step=None):
...
         self.start = start
         self.stop = stop
         self.step = step
-        self.n_frames = len(range(start, stop, step))
+        self.n_frames = len(self._sliced_trajectory)
         self.frames = np.zeros(self.n_frames, dtype=int)
```
IAlibay, line 362: I can see this becoming confusing real quickly, we're offering `frames` as an input argument to run and then offering a different thing contained in the `frames` user-facing variable. Thoughts?

Maybe we should call the input argument `indices` instead?

package/MDAnalysis/analysis/base.py
```diff
@@ -281,25 +295,31 @@ def run(self, start=None, stop=None, step=None, verbose=None):
             stop frame of analysis
         step : int, optional
             number of frames to skip between each analysed frame
+        frames : array_like, optional
+            array of integers or booleans to slice trajectory
```
IAlibay, line 396: We'll need some documentation / examples for this.

IAlibay (changes requested):
Couple of initial comments.

Main thing here is that I'm a little bit apprehensive about the differing behaviour about which frame you end up with at the end of for the frame iteration depending on if you use start/stop/step or a list of indices.

package/MDAnalysis/analysis/base.py
IAlibay: I've opened a separate issue, I'm not actually sure if it's intended behaviour.

package/MDAnalysis/analysis/base.py
orbeckst: I see your point. Although in principle `self.frames` should be equivalent to what the user inputs in `frames` in the sense that running a second time `a.run(frames=a.frames)` should give the identical answer (assuming that `ts.frame` is the true frame index), which I like. If the input `frames` was a boolean array or a `slice()` then `self.frames` will be an int array of frame indices but _should_ be functionally equivalent.

I am not too fond of calling the input `frames` kwarg `indices` because (I think) we typically use it for indexing atoms. (?)

package/MDAnalysis/analysis/base.py
orbeckst: Thanks for opening  #3416.

As a band-aid solution we could add a `self._trajectory.rewind()` after the `for i, ts in enumerate(self._sliced_trajectory)` loop. This would ensure consistent behavior of all AnalysisBase classes. It would just need a comment that references issue #3416 so that we can remove it again if that gets fixed/changed.

package/MDAnalysis/analysis/base.py
orbeckst: (I think that was me suggesting to keep the diff minimal... but you're absolutely right that frames should be with the other frame selections)

orbeckst:
@rsexton2 please ping me when you're ready (remove draft status) for review and I am happy to have a look; while you're hacking away at tests, please already address @IAlibay's comments, which are all very good & relevant. Also have a look at what PEP8speaks complains about, please.

rsexton2:
@orbeckst I added tests but still need to test slicing by a boolean array.
I think I addressed most of @IAlibay comments in the changes and fixed pep8 things

package/MDAnalysis/analysis/base.py
rsexton2: I changed the description to be more explicit, thank you.

package/MDAnalysis/analysis/base.py
rsexton2: I switched these around

package/MDAnalysis/analysis/base.py
rsexton2: I added tests to ensure frames isn't used with start, stop or step

package/MDAnalysis/analysis/base.py
IAlibay: I would preferentially add the more explicit versionchanged text here rather than under `_setup_frames` since `run` is exposed in the docs but `_setup_frames` isn't.

rsexton2: I added the versionchanged text to run

orbeckst:
For some reason the CI was waiting for approval — sorry @rsexton2 , I didn't notice.

Given that you have added tests, do you want to set the status to "ready to review", i.e., remove the draft status?

package/MDAnalysis/analysis/base.py
orbeckst: @IAlibay can you be convinced to leave the input `frames` as originally proposed or do you have strong opinions on the matter?

testsuite/MDAnalysisTests/analysis/test_base.py
```diff
@@ -197,6 +197,45 @@ def test_start_stop_step(u, run_kwargs, frames):
     assert_almost_equal(an.times, frames+1, decimal=4, err_msg=TIMES_ERR)
 
 
+@pytest.mark.parametrize('run_kwargs, frames', [
+    ({'frames': [4, 5, 6, 7, 8, 9]}, np.arange(4, 10)),
+    ({'frames': [0, 2, 4, 6, 8]}, np.arange(0, 10, 2)),
+    ({'frames': [4, 6, 8]}, np.arange(4, 10, 2)),
+    ({'frames': [True, True, False, True, False, True, True, False, True,
```
orbeckst, line 205: Also add `[0, 3, 4, 3, 5]` with a repeated frame.

Otherwise I like that you're checking `an.frames` !

testsuite/MDAnalysisTests/analysis/test_base.py
```diff
@@ -197,6 +197,45 @@ def test_start_stop_step(u, run_kwargs, frames):
...
+def test_frame_slice(run_kwargs, frames):
+    u = mda.Universe(TPR, XTC)  # dt = 100
+    an = FrameAnalysis(u.trajectory).run(**run_kwargs)
+    assert an.n_frames == len(frames)
+    assert_equal(an.found_frames, frames)
+    assert_equal(an.frames, frames, err_msg=FRAMES_ERR)
+
+
+@pytest.mark.parametrize('run_kwargs, frames', [
+    ({'start': 4, 'frames': [4, 5, 6, 7, 8, 9]}, np.arange(4, 10)),
```
orbeckst: What happens when you have `start=0, stop=0, step=0` ?

testsuite/MDAnalysisTests/analysis/test_base.py
```diff
@@ -197,6 +197,45 @@ def test_start_stop_step(u, run_kwargs, frames):
...
+])
+def test_frame_slice(run_kwargs, frames):
+    u = mda.Universe(TPR, XTC)  # dt = 100
+    an = FrameAnalysis(u.trajectory).run(**run_kwargs)
+    assert an.n_frames == len(frames)
+    assert_equal(an.found_frames, frames)
+    assert_equal(an.frames, frames, err_msg=FRAMES_ERR)
+
+
+@pytest.mark.parametrize('run_kwargs, frames', [
```
orbeckst: remove the frames part from the parametrization. You never check against it, so this is just noise and confuses.

testsuite/MDAnalysisTests/analysis/test_base.py
```diff
@@ -197,6 +197,45 @@ def test_start_stop_step(u, run_kwargs, frames):
...
+    ({'start': 4, 'frames': [4, 5, 6, 7, 8, 9]}, np.arange(4, 10)),
+    ({'stop': 6, 'frames': [0, 1, 2, 3, 4, 5]}, np.arange(0, 6)),
+    ({'step': 2, 'frames': [0, 2, 4, 6, 8]}, np.arange(0, 10, 2)),
+    ({'start': 4, 'stop': 7, 'frames': [4, 5, 6]}, np.arange(4, 8)),
+    ({'stop': 6, 'step': 2, 'frames': [0, 2, 4, 6]}, np.arange(0, 7, 2)),
+    ({'start': 4, 'step': 2, 'frames': [4, 6, 8]}, np.arange(4, 10, 2)),
+])
+def test_frame_fail(u, run_kwargs, frames):
```
orbeckst: Remove the reference value that is never tested
```suggestion
def test_frame_fail(u, run_kwargs):
```

orbeckst (changes requested):
This looks already very good and your test coverage is great!

I only found minor things — I'll approve once you addressed my comments.

The only major remaining issue is some example docs as requested by @IAlibay and I will let him review that aspect (i.e., all reviewers must approve).

testsuite/MDAnalysisTests/analysis/test_base.py
rsexton2: Is the proper behavior that it will run frame 3 twice?

testsuite/MDAnalysisTests/analysis/test_base.py
rsexton2: done

testsuite/MDAnalysisTests/analysis/test_base.py
rsexton2: done

testsuite/MDAnalysisTests/analysis/test_base.py
rsexton2: This slipped through.
I changed _setup_frames to fail if start/stop/step aren't all None

testsuite/MDAnalysisTests/analysis/test_base.py
orbeckst: yes

orbeckst (approved):

.github/pr_path_labeler.yml
```diff
@@ -0,0 +1,26 @@
+# Labeler triggered by .github/workflows/labels.yml
```
IAlibay: Strange, why are these files here? Did the merge not go right?

package/CHANGELOG
```diff
@@ -13,13 +13,14 @@ The rules for this file:
...
 ------------------------------------------------------------------------------
-??/??/?? IAlibay, melomcr
+??/??/?? IAlibay, melomcr, rsexton2
 
  * 2.1.0
 
 Fixes
 
 Enhancements
+  * Added frames processing to AnalysisBase.run (Issue #1985)
```
IAlibay: Can we use something similar to the versionchanged entry here? Whilst I'm sure it's not something most people read, it is where we direct folks to read about new changes, so being a bit more descriptive doesn't hurt.

package/MDAnalysis/analysis/base.py
IAlibay: Sure I'm fine with this as a temporary measure. We should implement the fix in this PR so that we don't forget.

package/MDAnalysis/analysis/base.py
IAlibay: It's a little bit messy but I'm not too fussed.

I'm not sure what (if anything) would need adding, but do we want to alter the attributes description for frame to make this clearer?

testsuite/MDAnalysisTests/analysis/test_base.py
```diff
@@ -197,6 +197,47 @@ def test_start_stop_step(u, run_kwargs, frames):
     assert_almost_equal(an.times, frames+1, decimal=4, err_msg=TIMES_ERR)
```
IAlibay, line 199: It'd be too much to ask to add a test for every analysis method, but could we maybe throw in an extra test for RMSD that shows that passing frames works with that as expected?

IAlibay (changes requested):
Sorry only had time for a quick glance at it. Do ping me when you have the documentation written.

.github/pr_path_labeler.yml
orbeckst: I'm not sure why they showed up. _Should_ be harmless when squash-merged, right?

orbeckst:
@IAlibay I assigned the PR to you as you're the last one with outstanding change requests.

I don't know why some of the commits that are already in master show up here, perhaps they came from pushing the "update with develop" button (or whatever it's called when GitHub tells you a branch is out of date). I _think_ it will be fine when squash-merging but if there are any concerns then possibly rebasing this branch against and updated develop??

.github/pr_path_labeler.yml
rsexton2: I rebased against an updated develop and squashed everything into 1 commit

package/CHANGELOG
rsexton2: Added more explicit description

package/MDAnalysis/analysis/base.py
rsexton2: I added the line to rewind with a comment referencing the issue

package/MDAnalysis/analysis/base.py
rsexton2: Added an example here with two uses of 'frames'

testsuite/MDAnalysisTests/analysis/test_base.py
rsexton2: I added a test to test_rmsd.py

package/MDAnalysis/analysis/base.py
rsexton2: @IAlibay How should I alter the attribute description for `frames`? 
It is currently "array of Timestep frame indices. Only exists after calling AnalysisBase.run"
Something like "array of Timestep frame indices. Only exists after calling AnalysisBase.run. If the 'frames' 
option is used in .run(), the input for 'frames' will be returned"

package/MDAnalysis/analysis/base.py
```diff
@@ -281,26 +298,51 @@ def run(self, start=None, stop=None, step=None, verbose=None):
...
+                self._sliced_trajectory,
                 verbose=verbose)):
             self._frame_index = i
             self._ts = ts
             self.frames[i] = ts.frame
             self.times[i] = ts.time
-            # logger.info("--> Doing frame {} of {}".format(i+1, self.n_frames))
             self._single_frame()
+        # rewind added for behavioral consistency, Issue #3416
+        self._trajectory.rewind()
```
IAlibay: Thanks, could you just add a quick test to check that this is working as intended for now? (nothing fancy, just a quick check that the trajectory rewinds back to 0 will do)

package/MDAnalysis/analysis/base.py
```diff
@@ -281,26 +298,51 @@ def run(self, start=None, stop=None, step=None, verbose=None):
...
         step : int, optional
             number of frames to skip between each analysed frame
+        frames : array_like, optional
+            array of integers or booleans to slice trajectory
         verbose : bool, optional
             Turn on verbosity
+
+
+    Example
+    -------
```
IAlibay: Should this be indented by two the right?

package/MDAnalysis/analysis/base.py
```diff
@@ -281,26 +298,51 @@ def run(self, start=None, stop=None, step=None, verbose=None):
...
+       import MDAnalysis as mda
+       from MDAnalysisTests.datafiles import TPR, XTC
+       from MDAnalysis.analysis import rms
+
+       u = mda.Universe(TPR, XTC)
+
+       rmsd = rms.RMSD(u, select='name CA')
+       # A list of frames to run the analysis on can be passed
```
IAlibay: Could you also throw in an example using start/stop/step? That way we have every use case here?

package/MDAnalysis/analysis/base.py
IAlibay: With hindsight, let's leave the attribute as it is. I'm not really sure how to amend (or if it's at all possible) it without making it more complex.

IAlibay (changes requested):
Thanks, just a couple of hopefully final changes.

package/MDAnalysis/analysis/base.py
rsexton2: done

package/MDAnalysis/analysis/base.py
rsexton2: You're right, I fixed it

package/MDAnalysis/analysis/base.py
rsexton2: done

IAlibay (approved):
lgtm assuming CI returns green

rsexton2:
git bisect says commit c35458b2b004a34a595c1c881be826cb9e0df460 is the one that makes the test testsuite/MDAnalysisTests/analysis/test_rdf.py::test_double_run FAILED

Locally testsuite/MDAnalysisTests/analysis/test_rdf.py::test_count FAILED
but not in the CI

```
c35458b2b004a34a595c1c881be826cb9e0df460 is the first bad commit
commit c35458b2b004a34a595c1c881be826cb9e0df460
Author: Ricky Sexton <rsexton2@asu.edu>
Date:   Fri Sep 10 15:49:19 2021 -0700

    Added rmsd test and 'frames' examples in docs

 package/CHANGELOG                              |  3 ++-
 package/MDAnalysis/analysis/base.py            | 18 ++++++++++++++++++
 testsuite/MDAnalysisTests/analysis/test_rms.py |  7 +++++++
 3 files changed, 27 insertions(+), 1 deletion(-)
```

The only thing that is added here is the rewind.
I'm not sure how this is affecting InterRDF

rsexton2:
@richardjgowers @VOD555 
Do you have any idea what is happening with InterRDF and trajectory.rewind() ?

IAlibay:
Just to follow up from discord - the issue seems to be from SingleFrameBase implementing an `_iter_` that doesn't rewind, which means that FrameIteratorAll rewinds on multi-frame trajectories but not on single frame readers.

In this particularly case - prior to this PR the `sels` fixture in test_rdf changes the coordinates and then keeps them that way. Whilst now the rewind specifically resets the timestep coordinates, hence yielding the test failure.

orbeckst:
What do we need to do so that this PR can move forward?

orbeckst:
As outlined in https://github.com/MDAnalysis/mdanalysis/pull/3415#issuecomment-922109775, iteration for single frame "trajectories" behaves differently from the way our multi-frame iterators _ought to_ behave (see also #3416 ). 

Could someone with proper understanding of the problem (@richardjgowers or @IAlibay ?) please raise an issue with a conclusion of what we will do about it? (I don't think it belongs into #3416, does it?) Then we can fix that problem and hopefully this PR can then be merged without additional changes.

IAlibay:
> What do we need to do so that this PR can move forward?

Assuming we want to change the behaviour of singleframe, a quick fix here would be to just change the failing test to not use a function scoped fixture and instead just do the frame manipulation twice. Either way we'll need to do that once we implement the singleframe change/fix.

I'll raise an issue.

jbarnoud:
I only had a very brief look, but since this PR already slices the trajectory, what makes it challenging to accept an already sliced trajectory as well?

If this PR goes through without accepting a sliced trajectory, the description of the PR should be edited so it does not close the original issue.

orbeckst:
In the original discussion following https://github.com/MDAnalysis/mdanalysis/issues/1985#issuecomment-913865103 we had decided to start out conservatively with only explicit start/stop/step. I agree that being able to hand over FrameIterators would make it really complete. 

It makes sense exploring expanding this PR once we get the tests to pass, which requires #3642 IIRC.

orbeckst:
For right now I changed the issue description to reflect the code that is present.

orbeckst:
PR #3642 needs to get merged before we can finalize this PR.

package/AUTHORS
```diff
@@ -171,6 +171,7 @@ Chronological list of authors
   - Marcelo C. R. Melo
   - Mark D. Driver
   - Kevin Boyd
+
```
IAlibay:
```suggestion
```
We don't usually have an empty line between year

package/MDAnalysis/analysis/base.py
```diff
@@ -238,17 +239,32 @@ def _setup_frames(self, trajectory, start=None, stop=None, step=None):
...
         step : int, optional
             number of frames to skip between each analysed frame
+        frames : array_like, optional
+            array of integers or booleans to slice trajectory
 
 
         .. versionchanged:: 1.0.0
             Added .frames and .times arrays as attributes
+
+        .. versionchanged:: 2.1.0
```
IAlibay:
```suggestion
        .. versionchanged:: 2.2.0
```

package/MDAnalysis/analysis/base.py
```diff
@@ -281,26 +298,54 @@ def run(self, start=None, stop=None, step=None, verbose=None):
...
+                self._sliced_trajectory,
                 verbose=verbose)):
             self._frame_index = i
             self._ts = ts
             self.frames[i] = ts.frame
             self.times[i] = ts.time
-            # logger.info("--> Doing frame {} of {}".format(i+1, self.n_frames))
             self._single_frame()
+        # rewind added for behavioral consistency, Issue #3416
+        self._trajectory.rewind()
```
IAlibay:
```suggestion
```
_should_ be fixed now

IAlibay (commented):
I think just these and this is ready to go (going to self-merge to speed things up).

IAlibay:
Assuming CI returns green I'm happy for this to get merged - @orbeckst do you want to have a final look?

orbeckst:
Please go ahead at your leisure — don’t have the bandwidth and don’t want to hold up things. Thank you for looking after this one and the related PR!!

lilyminium (approved):
LGTM, thanks @rsexton2, @IAlibay, @orbeckst.

package/MDAnalysis/analysis/base.py
```diff
@@ -281,25 +298,54 @@ def run(self, start=None, stop=None, step=None, verbose=None):
...
+            array of integers or booleans to slice trajectory
         verbose : bool, optional
             Turn on verbosity
+
+
+        Example
+        -------
+
+        .. code-block:: python
+
```
orbeckst:
```suggestion

```

IAlibay:
@orbeckst is this good to re-review or are you still working on the docs?

orbeckst:
Done with docs, good for most final of final reviews (I hope)—feel free to just edit as you see fit. 

The main thing was to move a specific example out of run() because it would be really confusing for all tools to reference RMSD in their inherited run().

orbeckst:
Can we merge this one ... finally??

IAlibay:
> Can we merge this one ... finally??

Sorry it's on my task list, I'll have a final look in the morning

package/MDAnalysis/analysis/base.py
```diff
@@ -20,11 +20,105 @@
...
+.. _`Writing your own trajectory analysis`:
+   https://userguide.mdanalysis.org/stable/examples/analysis/custom_trajectory_analysis.html
+
+
+Classes
+-------
+
+The :class:`Results` and :class:`AnalysisBase` classes are the essential
+building blocks for almost all MDAnalysis tools in the
+:mod:`MDAnalysis.analysis` module. They are supposed to be easily useable and
```
IAlibay:
```suggestion
:mod:`MDAnalysis.analysis` module. They aim to be easily useable and
```

IAlibay (approved):
lgtm just the one word change which I'll commit

IAlibay:
Thanks @rsexton2 ! Sorry it took so long to merge, but it fixed some pretty big bugs in MDAnalysis along the way! 🎉

PR #3424, opened 2021-09-23
[MAINT] removed --py3k argument for pylint

### LINKED ISSUE #3422: CI pylint fails because removed --py3k option is used
## Expected behavior ##

CI (pylint) runs

## Actual behavior ##

pylin failed with error 
```
pylint: error: no such option: --py3k
```
see https://github.com/MDAnalysis/mdanalysis/pull/3421#issuecomment-926026947

see https://github.com/MDAnalysis/mdanalysis/pull/3421#issuecomment-926026947

orbeckst:
Fixes #3422

- removed `pylint --py3k` option from GH CI workflow (pylint is not used in Azure or Travis) 
- py3k checker was removed in pylint 2.11
  https://pylint.pycqa.org/en/latest/whatsnew/2.11.html#removed-checkers

PR Checklist
------------
 - n/a Tests?
 - n/a Docs?
 - n/a CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/converters/RDKit.py
+++ b/package/MDAnalysis/converters/RDKit.py
@@ -452,7 +452,7 @@ def set_converter_cache_size(maxsize):
         conversions in memory. Using ``maxsize=None`` will remove all limits
         to the cache size, i.e. everything is cached.
     """
-    global atomgroup_to_mol
+    global atomgroup_to_mol   # pylint: disable=global-statement
     atomgroup_to_mol = lru_cache(maxsize=maxsize)(atomgroup_to_mol.__wrapped__)
 
 
@@ -638,7 +638,7 @@ def _standardize_patterns(mol, max_iter=200):
     +---------------+------------------------------------------------------------------------------+
     | nitro         | ``[N;X3;v3:1](-[O-;X1:2])-[O-;X1:3]>>[N+:1](-[O-:2])=[O+0:3]``               |
     +---------------+------------------------------------------------------------------------------+
- 
+
     """
 
     # standardize conjugated systems
--- a/package/MDAnalysis/coordinates/null.py
+++ b/package/MDAnalysis/coordinates/null.py
@@ -60,4 +60,3 @@ class NullWriter(base.WriterBase):
         except AttributeError:
             errmsg = "Input obj is neither an AtomGroup or Universe"
             raise TypeError(errmsg) from None
-        pass
--- a/package/MDAnalysis/core/topologyattrs.py
+++ b/package/MDAnalysis/core/topologyattrs.py
@@ -2343,7 +2343,6 @@ class _ConnectionTopologyAttrMeta(_TopologyAttrMeta):
     to return only the connections within the atoms in the group.
     """
     def __init__(cls, name, bases, classdict):
-        type.__init__(type, name, bases, classdict)
         attrname = classdict.get('attrname')
 
         if attrname is not None:
--- a/package/MDAnalysis/core/topologyattrs.py
+++ b/package/MDAnalysis/core/topologyattrs.py
@@ -2343,6 +2343,8 @@ class _ConnectionTopologyAttrMeta(_TopologyAttrMeta):
     to return only the connections within the atoms in the group.
     """
     def __init__(cls, name, bases, classdict):
+        super().__init__(name, bases, classdict)
+
         attrname = classdict.get('attrname')
 
         if attrname is not None:
@@ -2354,8 +2356,6 @@ class _ConnectionTopologyAttrMeta(_TopologyAttrMeta):
             method = MethodType(intra_connection, cls)
             prop = property(method, None, None, method.__doc__)
             cls.transplants[AtomGroup].append((f"intra_{attrname}", prop))
-
-        super().__init__(name, bases, classdict)
 
 
 class _Connection(AtomAttr, metaclass=_ConnectionTopologyAttrMeta):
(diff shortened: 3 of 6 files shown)


IAlibay:
That's a lot fewer lint failures than I expected.

edit: at least RDKit we know for sure is fine (so we'll need to get pylint to ignore the global statement)

orbeckst:
If you have an idea how to fix/ignore the three issues then we can do it in this PR and start with a clean pylint.

package/MDAnalysis/coordinates/null.py
```diff
@@ -60,4 +60,3 @@ def _write_next_frame(self, obj):
         except AttributeError:
             errmsg = "Input obj is neither an AtomGroup or Universe"
             raise TypeError(errmsg) from None
-        pass
```
IAlibay, line 63: Doesn't seem like this was required anyways?

package/MDAnalysis/converters/RDKit.py
```diff
@@ -452,7 +452,7 @@ def set_converter_cache_size(maxsize):
         conversions in memory. Using ``maxsize=None`` will remove all limits
         to the cache size, i.e. everything is cached.
     """
-    global atomgroup_to_mol
+    global atomgroup_to_mol # pylint: disable=global-statement
```
IAlibay: I believe we said this was a justified global call.

IAlibay (commented):
Note sure about the third one.

Don't really understand why pylint is complaining about: https://github.com/MDAnalysis/mdanalysis/blob/develop/package/MDAnalysis/core/topologyattrs.py#L2346

package/MDAnalysis/coordinates/null.py
orbeckst: yep

package/MDAnalysis/converters/RDKit.py
orbeckst: I think so... caches are iffy.

IAlibay:
If the `type.__init__` call is fine, we can just add `# pylint: disable=W0233` on that line and it'll just ignore the warning.

orbeckst:
[W0233](https://pylint.pycqa.org/en/latest/technical_reference/features.html?highlight=W0233#classes-checker-messages): Used when an `__init__` method is called on a class which is not in the direct ancestors for the analysed class.

This metaclass `_ConnectionTopologyAttrMeta(_TopologyAttrMeta)` is ultimately derived from `type` https://github.com/MDAnalysis/mdanalysis/blob/39850215eae5a3e2c9d28a24dbc670646e305f79/package/MDAnalysis/core/topologyattrs.py#L287 so I am not quite sure why it triggers.

Who wrote that metaclass magic and might be able to comment? Is there a more pythonic way to do this?

orbeckst:
From blame, the most likely authors are @richardjgowers and @lilyminium — would you mind having a quick look here and comment on why pylint is complaining about the line 
```python
class _ConnectionTopologyAttrMeta(_TopologyAttrMeta):
    def __init__(cls, name, bases, classdict):
        type.__init__(type, name, bases, classdict)     # <---- pylint W0233
        attrname = classdict.get('attrname')

```
see https://github.com/MDAnalysis/mdanalysis/blob/39850215eae5a3e2c9d28a24dbc670646e305f79/package/MDAnalysis/core/topologyattrs.py#L2346

Thanks.

IAlibay:
> [W0233](https://pylint.pycqa.org/en/latest/technical_reference/features.html?highlight=W0233#classes-checker-messages): Used when an `__init__` method is called on a class which is not in the direct ancestors for the analysed class.
> 
> This metaclass `_ConnectionTopologyAttrMeta(_TopologyAttrMeta)` is ultimately derived from `type` https://github.com/MDAnalysis/mdanalysis/blob/39850215eae5a3e2c9d28a24dbc670646e305f79/package/MDAnalysis/core/topologyattrs.py#L287 so I am not quite sure why it triggers.
> 
> Who wrote that metaclass magic and might be able to comment? Is there a more pythonic way to do this?
> 
> 

@lilyminium I believe? If it works as intended we can just ignore it ?

lilyminium:
I think you could just remove `type.__init__`.

orbeckst:
ok, will try

orbeckst:
Looking good locally
```
$ pytest -n 4 --disable-pytest-warnings --durations=50 testsuite/MDAnalysisTests
======================================================= test session starts =======================================================
platform darwin -- Python 3.7.10, pytest-6.2.5, py-1.10.0, pluggy-1.0.0
...
====================== 17187 passed, 107 skipped, 2 xfailed, 2 xpassed, 45878 warnings in 432.02s (0:07:12) =======================
```

package/MDAnalysis/core/topologyattrs.py
```diff
@@ -2343,7 +2343,6 @@ class _ConnectionTopologyAttrMeta(_TopologyAttrMeta):
     to return only the connections within the atoms in the group.
     """
     def __init__(cls, name, bases, classdict):
-        type.__init__(type, name, bases, classdict)
```
IAlibay, line 2346: I guess the call just gets inherited from the parent class anyways (given we are calling super below)?

@richardjgowers - looks like you're the original author of the parent class, does it all make sense to you?

IAlibay (approved):
lgtm

IAlibay:
@richardjgowers - please review here, we need to get this merged so that CI stops being in failed status.

orbeckst:
ping @richardjgowers please!

package/MDAnalysis/core/topologyattrs.py
richardjgowers: This seems to work

richardjgowers (approved):

orbeckst:
Thanks @IAlibay and @richardjgowers !

PR #3425, opened 2021-09-24
Itpparserchargefix

### LINKED ISSUE #3419: Issue reading custom GROMACS TOP/ITP files in MDAnalysis 2.0.0
## Expected behaviour ##

I'm using MDAnalysis for processing coarse grained simulations run in GROMACS. I need to assign the correct topology to systems to enable further processing. The easiest way to do this in MDAnalysis was to read in this information from the top/itp files to apply to the pdb/trr files from gromacs. This worked in MDAnalysis 1.1.1 (however charges were not detected).

## Actual behaviour ##

In MDAnalysis 2.0.0 a Value error from reading charges in top/itp files is returned.

> Traceback (most recent call last):
> 
>   File "/home/mark/anaconda3/envs/1bpa2/lib/python3.8/site-packages/MDAnalysis/core/universe.py", line 122, in _topology_from_file_like
>     topology = p.parse(**kwargs)
> 
>   File "/home/mark/anaconda3/envs/1bpa2/lib/python3.8/site-packages/MDAnalysis/topology/ITPParser.py", line 584, in parse
>     attrs.append(Attr(np.array(vals, dtype=dtype)))
> 
> ValueError: could not convert string to float: '-'
> 
> 
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
> 
>   File "/tmp/ipykernel_7090/573278789.py", line 1, in <module>
>     mdanalysis.Universe("polymer.top", "examplegromacs.pdb",  topology_format="ITP", include_dir="./")
> 
>   File "/home/mark/anaconda3/envs/1bpa2/lib/python3.8/site-packages/MDAnalysis/core/universe.py", line 336, in __init__
>     topology = _topology_from_file_like(self.filename,
> 
>   File "/home/mark/anaconda3/envs/1bpa2/lib/python3.8/site-packages/MDAnalysis/core/universe.py", line 137, in _topology_from_file_like
>     raise ValueError(
> 
> ValueError: Failed to construct topology from file polymer.top with parser <class 'MDAnalysis.topology.ITPParser.ITPParser'>.
> Error: could not convert string to float: '-'

I added some extra logging statements to localise the issue, and I found the behaviour stems from ITPParser when it is reading charges into the array (detailed output in the attached log file [mdanalysisissue.log](https://github.com/MDAnalysis/mdanalysis/files/7208420/mdanalysisissue.log)).
 The setting of the array in [line 534](https://github.com/MDAnalysis/mdanalysis/blob/39850215eae5a3e2c9d28a24dbc670646e305f79/package/MDAnalysis/topology/ITPParser.py#L534) seems to cause the issue- the datatype is set to '<U1' which later causes only the '-' sign of any residues to be read (later causing the observed issue when type cast to a float). 

This breaks my current analysis workflow. I'm upgrading to 2.0.0 to try and utilise the parallelisation support you have added.

I found that setting dtype=object in the array created on line 534 resolves this issue. If the issue is confirmed how would I go about a pull request/ getting this change committed into develop?

``` python
import MDAnalysis as mda

u = mda.Universe("polymer.top", "examplegromacs.pdb",  topology_format="ITP", include_dir="./")


....

```
Example coarse grained files attached that cause the issue. Any residues with a negative charge result in the given error.
[mdanalysisissueexamplefiles.zip](https://github.com/MDAnalysis/mdanalysis/files/7208390/mdanalysisissueexamplefiles.zip)

mdanalysis.__version__
Out[18]: '2.0.0'

Python 3.8.11

- Which operating system?

Ubuntu 20.04

mdd31:
Fixes #3419 

 - 

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


mdd31:
@orbeckst  Assuming the CI passes the changes are ready for review.

richardjgowers (changes requested):
Thanks @mdd31 this looks good. Could you add a small test which would have failed previously and now passes?

mdd31:
@richardjgowers Yes, hopefully the added test case is suitable? Any modifications required please tell me.

package/CHANGELOG
```diff
@@ -38,6 +38,7 @@ Deprecations
   * 2.0.0
 
 Fixes
+  * Fix ITPParser charge reading (Issue #3419).
```
richardjgowers: 2.0 is released already so this'll need to go in the 2.1 section

richardjgowers (changes requested):

package/CHANGELOG
mdd31: Should be  fixed in https://github.com/MDAnalysis/mdanalysis/pull/3425/commits/a69cd8cd3695aea4309a5b800fac384ee301d605

testsuite/MDAnalysisTests/topology/test_itp.py
```diff
@@ -184,11 +185,40 @@ def test_charge_parse(self, universe):
...
+        assert_almost_equal(universe.atoms[0].charge, -1.0)
+        assert_almost_equal(universe.atoms[1].charge, 0)
+        assert_almost_equal(universe.atoms[2].charge, 0)
+        assert_almost_equal(universe.atoms[3].charge, -1.)
+
+    def test_mass_parse_or_guess(self, universe):
+        assert_almost_equal(universe.atoms[0].mass, 100.0)
+        assert_almost_equal(universe.atoms[1].mass, 100.0)
+        assert_almost_equal(universe.atoms[2].mass, 100.0)
+        assert_almost_equal(universe.atoms[3].mass, 100.0)
```
tylerjereddy: since it is used a lot here, perhaps worth noting that `assert_allclose` is recommended these days per the docs:
https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_almost_equal.html

IAlibay: @tylerjereddy - do you think that warrants a proper issue being raised? i.e. is there a possibility that assert_almost_equal will eventually be deprecated?

I'll raise a userguide issue since we recommend assert_almost_equal at the moment.

mdd31: Ok, I'll update the tests I've added and also convert the other ones in the test_itp to use `assert_allclose`. I was just following the existing convention in the file but happy to change it.

testsuite/MDAnalysisTests/datafiles.py
```diff
@@ -557,6 +557,7 @@
 ITP = resource_filename(__name__, 'data/gromacs_ala10.itp')
 ITP_nomass = resource_filename(__name__, 'data/itp_nomass.itp')
 ITP_atomtypes = resource_filename(__name__, 'data/atomtypes.itp')
```
richardjgowers, line 560: Ok the final test failing is that this needs to be added to ‘__all__’ (for reasons I’m unsure of)

richardjgowers, line 560: Ok the final test failing is that this needs to be added to ‘__all__’ (for reasons I’m unsure of)

mdd31: I'm hoping I've fixed it in the latest push- It looks like it needed adding to the list __all__ defined at the start of  datafiles.py

richardjgowers (approved):

richardjgowers:
Thanks @mdd31 for fixing this, the ITP parser is relatively new and fixes like this are very important to make it useful for people!

PR #3429, opened 2021-09-27
Fix arch option in setup.py to work properly + not trigger on MSVC

### LINKED ISSUE #3428: March=native setup.py flag not setup correctly and will not work with MSVC
## Expected behavior ##

You should be able to build architecture optimised code with the setup.py `arch` option.
Additionally this option is poorly documented not covered in CI AFAIK.

## Actual behavior ##

setup.cfg has string quotes making the option invalid 

`march="native"` 👎 

`march=native` 👍 

Additionally `-march` is not a valid flag for MSVC, instead the `-arch:` flag must be used 

Uncomment `march="native"` in setup.cfg and install with `setup.py`

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`)  2.1.0-dev0
- Which version of Python (`python -V`)? 3.8.8

hmacdope:
Fixes #3428

 -  ~~fixes~~ removes march option in setup.cfg
 - ~~checks for MSVC (WIP??)~~
 - ~~only allows if not on MSVC~~
 - allow any additional compiler flags to be used for build excluding encore

Questions:
- ~~Do we want to check if the **flag** `march` is accepted rather than using the compiler identity? E.g what about icc, pgi etc etc~~
- ~~Do we want to improve documentation of this option?~~
- Do we want to add this option to CI? My guess is probably not?

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ matrix:
       name: "Power PC"
       os: linux
       arch: ppc64le
-      if: type = cron
+      #if: type = cron
       before_install:
         - wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-ppc64le.sh -O miniconda.sh
         - mkdir $HOME/.conda
@@ -42,7 +42,7 @@ matrix:
       dist: focal
       virt: vm
       group: edge
-      if: type = cron
+      #if: type = cron
       before_install:
         - python -m pip install cython "numpy>=1.19.2" scipy
         - python -m pip install --no-build-isolation hypothesis matplotlib pytest pytest-cov pytest-xdist tqdm threadpoolctl
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ matrix:
       name: "Power PC"
       os: linux
       arch: ppc64le
-      #if: type = cron
+      if: type = cron
       before_install:
         - wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-ppc64le.sh -O miniconda.sh
         - mkdir $HOME/.conda
@@ -42,7 +42,7 @@ matrix:
       dist: focal
       virt: vm
       group: edge
-      #if: type = cron
+      if: type = cron
       before_install:
         - python -m pip install cython "numpy>=1.19.2" scipy
         - python -m pip install --no-build-isolation hypothesis matplotlib pytest pytest-cov pytest-xdist tqdm threadpoolctl
--- a/package/setup.cfg
+++ b/package/setup.cfg
@@ -7,7 +7,7 @@
 #use_cython=False
 #keep_cythonized=True
 ## enable cpu specific optimizations
-#march="native"
+#march=native
 [wheel]
 universal = 1
 
--- a/package/setup.cfg
+++ b/package/setup.cfg
@@ -6,8 +6,8 @@
 ## Uncomment the following two lines to get the _release_ behavior.
 #use_cython=False
 #keep_cythonized=True
-## enable cpu specific optimizations
-march=native
+## any additional compiler flags
+#extra_cflags = 
 [wheel]
 universal = 1
 
--- a/package/setup.cfg
+++ b/package/setup.cfg
@@ -6,7 +6,7 @@
 ## Uncomment the following two lines to get the _release_ behavior.
 #use_cython=False
 #keep_cythonized=True
-## any additional compiler flags
+## any additional compiler flags for core C routines, excluding ENCORE
 #extra_cflags = 
 [wheel]
 universal = 1
--- a/package/setup.cfg
+++ b/package/setup.cfg
@@ -7,7 +7,7 @@
 #use_cython=False
 #keep_cythonized=True
 ## enable cpu specific optimizations
-#march=native
+march=native
 [wheel]
 universal = 1
 
--- a/package/setup.py
+++ b/package/setup.py
@@ -296,26 +296,14 @@ def extensions(config):
     else:
         encore_compile_args.append('-O3')
 
-    # allow using architecture specific instructions. This allows people to
-    # build optimized versions of MDAnalysis. Do here so not included in encore
-    arch = config.get('march', default=False)
-    if arch:
-        if using_msvc():
-            # MSVC doesn't have a good equivalent for the march flag
-            pass
-        else:
-            if platform.machine() == 'x86_64':
-                extra_compile_args.append('-march={}'.format(arch))
-            elif platform.machine() == 'arm':
-                # arm prefers the use of  mcpu only
-                extra_compile_args.append('-mcpu={}'.format(arch))
-            elif platform.machine() == 'ppc64le':
-                # PowerPC prefers the use of mcpu and mtune
-                extra_compile_args.append('-mcpu={}'.format(arch))
-                extra_compile_args.append('-mtune={}'.format(arch))
-            else:
-                # platform specific optimizations skipped
-                pass
+    # allow using custom c/c++ flags and architecture specific instructions.
+    # This allows people to build optimized versions of MDAnalysis.
+    # Do here so not included in encore
+    extra_cflags = config.get('extra_cflags', default=False)
+    if extra_cflags:
+        flags = extra_cflags.split()
+        for flag in flags:
+            extra_compile_args.append(flag)
 
     cpp_extra_compile_args = [a for a in extra_compile_args if 'std' not in a]
     cpp_extra_compile_args.append('-std=c++11')
--- a/package/setup.py
+++ b/package/setup.py
@@ -307,7 +307,7 @@ def extensions(config):
         else:
             if platform.machine() == "x86_64":
                 extra_compile_args.append('-march={}'.format(arch))
-            elif: platform.machine == "arm":
+            elif platform.machine == "arm":
             `   # arm prefers the use of  mcpu only
                 extra_compile_args.append('-mcpu={}'.format(arch))
             else:
--- a/package/setup.py
+++ b/package/setup.py
@@ -307,7 +307,7 @@ def extensions(config):
         else:
             if platform.machine() == "x86_64":
                 extra_compile_args.append('-march={}'.format(arch))
-            elif platform.machine == "arm":
+            elif platform.machine() == "arm":
                 # arm prefers the use of  mcpu only
                 extra_compile_args.append('-mcpu={}'.format(arch))
             else:
--- a/package/setup.py
+++ b/package/setup.py
@@ -310,8 +310,8 @@ def extensions(config):
                 # arm prefers the use of  mcpu only
                 extra_compile_args.append('-mcpu={}'.format(arch))
             elif platform.machine() == 'ppc64le':
-                #PowerPC prefers the use of mcpu and mtune
-                extra_compile_args.append('-mcpu={}'.format(arch))        
+                # PowerPC prefers the use of mcpu and mtune
+                extra_compile_args.append('-mcpu={}'.format(arch))
                 extra_compile_args.append('-mtune={}'.format(arch))
             else:
                 # platform specific optimizations skipped
--- a/package/setup.py
+++ b/package/setup.py
@@ -46,7 +46,6 @@ Google groups forbids any name that contains the string `anal'.)
 from setuptools import setup, Extension, find_packages
 from distutils.ccompiler import new_compiler
 from distutils.sysconfig import customize_compiler
-from distutils.msvccompiler import MSVCCompiler
 import codecs
 import os
 import sys
--- a/package/setup.py
+++ b/package/setup.py
@@ -270,13 +270,9 @@ def using_clang():
 
 def using_msvc():
     """Will we be using an MSVC compiler?"""
-    # how portable is this?
-    compiler = new_compiler()
-    customize_compiler(compiler)
-    if isinstance(compiler, MSVCCompiler):
-        is_msvc = True
-    else:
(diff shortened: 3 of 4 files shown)


hmacdope:
Tagging @IAlibay for CI question. :)

Also @tylerjereddy, how is the MSVC check done upstream in numpy/scipy? I couldn't see if there was a consistent and compact way it is done. 

General question: test for compiler identity or flag support?

tylerjereddy:
> Also @tylerjereddy, how is the MSVC check done upstream in numpy/scipy? I couldn't see if there was a consistent and compact way it is done.

SciPy is on the verge of switching to `meson` for builds, but digging through the code a bit I see stuff like this:

image

That seems to be a mixture of both approaches (looking for compiler by name and also using a `try..` style function to conservatively add a flag after.

NumPy seems to have bunch of stuff--after all, it provides the `numpy.distutils` build system proper. This one stands out for being clean and simple and seems to be used in some prominent places (NumPy core itself and the build for `random`, and both are used *tons* of course):

image

That appears to be in response to problems with `mingw` compiling issues on Windows. Part of me wonders if you could use one compiler to build your Python interpreter and another one to build MDAnalysis (just trying to figure out why one *wouldn't* be able to use that), but if it used in those crucial places in NumPy I suppose it is likely to be robust.

package/setup.py
```diff
@@ -284,7 +295,11 @@ def extensions(config):
     # build optimized versions of MDAnalysis.
     arch = config.get('march', default=False)
     if arch:
-        extra_compile_args.append('-march={}'.format(arch))
+        if using_msvc():
+            # MSVC doesn't have a good equivalent for the march flag
+            pass
+        else:
+            extra_compile_args.append('-march={}'.format(arch))
```
tylerjereddy: FWIW, I don't have any particular objection to the approach from a quick glance.. CI would fail without this shim I guess?

I don't know if long-term dependence on `distutils` is really desirable but that's another issue. Maybe a `try..` style approach would be more robust, but not sure if we want to vendor one from upstream or can easily use another dep to provide that

package/setup.py
```diff
@@ -267,6 +268,16 @@ def using_clang():
         is_clang = False
     return is_clang
 
+def using_msvc():
+    """Will we be using an MSVC compiler?"""
+    # how portable is this?
```
tylerjereddy: I suppose checking if you're on Windows platform here/somewhere is fairly cheap as an extra guard, though not sure it would really help to have that extra filter.

IAlibay:
So I know the march option existed, albeit undocumented & broken, but I do have to query its need / existence:
1. Our tests are quite flaky on -O3 already, do we have any sense on if they'd pass with more advanced instructions
2. Would using it actually provide a noticeable benefit to runtime execution speed? My understanding is that the majority of our costs were "things moving in memory".

The latter is pretty important, if there's no benefits then maybe we should kill off the option rather than try to fix it?

hmacdope:
Fair enough to question whether worth fixing. 

AFAIK The use of instructions **shouldn't** change the level of accuracy on most things. AFAIK no standard breaking or unsafe floating point arithmetic is allowed by AVX or SSE or whatever. 

 Let me do some tests on my work computer RE speed improvements and get back to you.

package/setup.py
hmacdope: For clarity march build is disabled by being commented out `setup.cfg` in regular build so CI is normally fine.

hmacdope:
Thanks @tylerjereddy and @IAlibay :)

package/setup.py
hmacdope: 👍

hmacdope:
@tylerjereddy @IAlibay Apologies for screenshot rather than graph, but appears as though performance gains are significant, esp for distances when you can use AVX/AVX2. This mirrors what I found last time and also what @richardjgowers and I have found over at distopia, that the compiler autovectorisation is a significant improvement.

Here we can see that some of the distance calculations are almost 3X faster with `march=native` enabled, run on my AVX2 capable workstation.

MDA_bench

With that in mind I would probably advocate for continued life of the `march` option but would of course value any opinion for @MDAnalysis/coredevs and others.

IAlibay:
Do all the tests still pass?

hmacdope:
Only a single Encore failure on Ubuntu MacOS and Win appears fine (see most recent CI). Tests are flakier on my local attempts but can try and get to bottom of that.

hmacdope:
Was a problem with my test env, all tests seem to pass. :)

hmacdope:
@IAlibay thanks for the heads up, removing `march=native` from the encore flags fixed the only failure.

IAlibay:
> @IAlibay thanks for the heads up, removing `march=native` from the encore flags fixed the only failure.

I suspect even with this you would still trigger failures on aarch64 and ppc64le, but we run CI for those on a weekly cron job. As long as you don't end up doing a lot of commits, if you want to temporarily change the travis yaml file to remove the cron job limitation, I _think_ it should trigger a build.

hmacdope:
Thats a great idea @IAlibay, I will try and find the right thingo and confirm with you. Once enable I won't add any more commits until people have had a chance to discuss.

hmacdope:
ARM = OK 

but flags have  different meaning, prefers mcpu=native see:  https://community.arm.com/developer/tools-software/tools/b/tools-software-ides-blog/posts/compiler-flags-across-architectures-march-mtune-and-mcpu. 

PowerPC=no build

 needs "mcpu=native  mtune=native" instead

hmacdope:
Fixed by adding x86 and arm specific paths only.

hmacdope:
Did we reach a consensus on supporting this option? 3x speedup for distances is quite attractive (to me). Key will be documenting an "advanced" install option, which I am happy to do, just need to know either way.

richardjgowers:
Definitely writing a guide on how to achieve this with a manual install is
an easy win.

hmacdope:
I will add an example to the user guide, and update the installation wiki page post merge.

hmacdope:
See https://github.com/MDAnalysis/UserGuide/pull/177 for docs.  Also note that changes to CI and actual enabling of the march=native option will be removed, but are here for now to show CI passing.

IAlibay:
> See [MDAnalysis/UserGuide#177](https://github.com/MDAnalysis/UserGuide/pull/177) for docs. Also note that changes to CI and actual enabling of the march=native option will be removed, but are here for now to show CI passing.

Can you raise an issue about this? I'll eventually add it to our weekly cron job at some point.

package/setup.cfg
```diff
@@ -7,7 +7,7 @@
 #use_cython=False
 #keep_cythonized=True
 ## enable cpu specific optimizations
-#march="native"
+march=native
```
IAlibay: march options are True or False internally, setting "native" just seems counter-intuitive here, (i.e. what does native mean to a user - True or False?)

package/setup.py
```diff
@@ -295,6 +297,24 @@ def extensions(config):
...
+    # build optimized versions of MDAnalysis. Do here so not included in encore
+    arch = config.get('march', default=False)
+    if arch:
+        if using_msvc():
+            # MSVC doesn't have a good equivalent for the march flag
+            pass
+        else:
+            if platform.machine() == "x86_64":
+                extra_compile_args.append('-march={}'.format(arch))
+            elif platform.machine() == "arm":
```
IAlibay: if ppc64le needs "mcpu=native mtune=native" why not add an extra branch for this here?

.travis.yml
```diff
@@ -14,7 +14,7 @@ matrix:
       name: "Power PC"
       os: linux
       arch: ppc64le
-      if: type = cron
+      #if: type = cron
```
IAlibay: Once you're done please remember to uncomment these.

package/setup.py
```diff
@@ -267,6 +267,12 @@ def using_clang():
         is_clang = False
     return is_clang
```
IAlibay: I know we don't in most places and I would ask we don't attempt to format the entirety of setup.py in one go, but if we're adding things we should try to keep to pep8 if we can (slowly we'll eventually fix the whole thing).

package/setup.py
```diff
@@ -280,11 +286,7 @@ def extensions(config):
         extra_compile_args.extend(['-Wall', '-pedantic'])
         define_macros.extend([('DEBUG', '1')])
 
-    # allow using architecture specific instructions. This allows people to
-    # build optimized versions of MDAnalysis.
-    arch = config.get('march', default=False)
-    if arch:
-        extra_compile_args.append('-march={}'.format(arch))
+
```
IAlibay: unnecessary extra line?

IAlibay (changes requested):
Couple of comments

package/setup.cfg
hmacdope: I think this option is set to `native` to allow the use of a specific cpu arch if you would like with a list of allowed options here:
https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html
eg if you could set it to `broadwell` if you so  desired.

Do we want to allow this? or too opaque and only allow "native" to be selected.

package/setup.py
hmacdope: good idea

package/setup.cfg
IAlibay: Yeah but right now you could set it to "parsnips" and it would still get interpreted as a bool and do the same thing as if you set it to whatever else right?

If we want to support more things I'm not against it but we should a) rename to "compiler_opts" or similar, b) change at least the loading of that specific option to not be a bool.

hmacdope: Its not a bool even though it looks like one:

```python
arch = config.get('march', default=False)
```
stores the string `"native"` so the `if` tests `if "native"`:

The value is used again to select the compiler option with 
```python
 extra_compile_args.append('-mtune={}'.format(arch))
```
 so supplying `march=parsnips` gives a compiler error `unknown target CPU parsnip` :)

IAlibay: In that case we shouldn't have the branch, just change the use case to use whatever computer flags users want to feed in?

Edit: here I mean `mtune` etc... and all get the whole flag passed rather than just the option

IAlibay: if we go that way then we can reduce the code footprint to ~ 2-3 lines, it's a power user feature anyways, so folks shouldn't be messing with it unless they actually understand how their compilers work

hmacdope: I guess in my mind the branching was to handle/encourage correct inter-platform variation in the compiler flags for 'mtune', 'march' and 'mcpu' which all take the same family of input args.

IAlibay: Since it's advanced stuff I would recommend that being in docs/userguide. In my opinion, it's kinda silly to make it an advanced feature but to also a) handhold users, b) make it restrictive in choice

hmacdope: I think perhaps the main motivation for handholding was encouraging an easy performance gain. Perhaps a flag to add **any** extra compiler flags you want is a good idea instead and then a detailed user guide example?

richardjgowers: I'm not quite keeping track of everything here, but I think it's probably best to allow *any* c(xx)flags that a user passes in to be passed along to the compiler. Then user guide can provide an example that march=native is a good idea.

IAlibay: yeah that's what I'm advocating for too

hmacdope: Roger roger, I think thats a good all round solution.

hmacdope: It's also extensible for any kinds of future compiler optimisations we want to apply.

package/setup.py
```diff
@@ -268,6 +268,13 @@ def using_clang():
     return is_clang
 
 
+def using_msvc():
```
hmacdope: This is not used, but may be a useful future addition?

IAlibay: unless it's necessary now or you know you'll need it in the next PR - please remove it

it's always easier to add code than work out why it exists & remove it

package/setup.py
```diff
@@ -295,6 +296,15 @@ def extensions(config):
     else:
         encore_compile_args.append('-O3')
 
+    # allow using custom c/c++ flags and architecture specific instructions.
+    # This allows people to build optimized versions of MDAnalysis.
+    # Do here so not included in encore
```
IAlibay, line 294: Can you add a short version of this to the setup.cfg comment too please? (i.e. that it doesn't affect encore compilation flags)

package/setup.py
```diff
@@ -295,6 +296,15 @@ def extensions(config):
...
+    extra_cflags = config.get('extra_cflags', default=False)
+    if extra_cflags:
+        flags = extra_cflags.split()
+        for flag in flags:
+            extra_compile_args.append(flag)
```
IAlibay: couldn't this just be done in one line using list.extend()?

i.e.

```python
extra_compile_args.extend(flags)
```

IAlibay (changes requested):
two small things and this should be done

package/setup.py
hmacdope: Yep for sure.

package/setup.py
hmacdope: Roger

package/setup.py
hmacdope: Yep for sure

IAlibay:
There's something wrong with codecov at the moment but I'm not seeing what anything in either actions or codecov's status updates.. I restarted the runners we'll see what happens

IAlibay (approved):
Assuming CI returns green lgtm, thanks @hmacdope !

PR #3432, opened 2021-10-01
TRRWriter, use ts.data['step'] if it exists

### LINKED ISSUE #3350: TRRWriter doesn't use ts.data['step'] when copying TRR file
I'm not entirely sure if this is a bug because I'm not entirely sure what `ts.data['step']` is _supposed_ to be. But in any case, I think copying a file by writing the entire thing to a new file should maintain all of the data in the original file, which in this case it's not.

## Expected behavior ##

Writing an entire trajectory to a new file should maintain all of the data.

## Actual behavior ##

The `ts.data['step']` information is lost when copying a TRR file by writing with MDAnalysis

``` python
import MDAnalysis as mda
from MDAnalysis.tests.datafiles import TPR_xvf, TRR_xvf

u = mda.Universe(TPR_xvf, TRR_xvf)

for ts in u.trajectory:
    print (ts.data['step'])
0
25000
50000

with mda.Writer('copied_file.trr', n_atoms=u.trajectory.n_atoms) as W:
    for ts in u.trajectory:
        W.write(u)

copied_u = mda.Universe(TPR_xvf, 'copied_file.trr')

# now has different steps
for ts in copied_u.trajectory:
    print (ts.data['step'])
0
1
2

```

2.0.0-dev0
Python 3.8.5
- Which operating system?
Windows 10 Home

ianmkenney:
If ts.data['step'] exists, use this for the output
step value. Otherwise, use ts.frame, consistent
with previous behavior.

Fixes #3350 

 - use `ts.data.get('step', ts.frame)` instead of `ts.frame` for the step variable

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/coordinates/TRR.py
+++ b/package/MDAnalysis/coordinates/TRR.py
@@ -41,13 +41,18 @@ class TRRWriter(XDRBaseWriter):
     """Writer for the Gromacs TRR format.
 
     The Gromacs TRR trajectory format is a lossless format. The TRR format can
-    store *velocoties* and *forces* in addition to the coordinates. It is also
+    store *velocities* and *forces* in addition to the coordinates. It is also
     used by other Gromacs tools to store and process other data such as modes
     from a principal component analysis.
 
     If the data dictionary of a :class:`Timestep` contains the key
     'lambda' the corresponding value will be used as the lambda value
     for written TRR file.  If ``None`` is found the lambda is set to 0.
+
+    If the data dictionary of a :class:`Timestep` contains the key
+    'step' the corresponding value will be used as the step value for
+    the written TRR file. If the dictionary does not contain 'step', then
+    the step is set to the :class:`Timestep` frame attribute.
 
     """
 
@@ -75,6 +80,9 @@ class TRRWriter(XDRBaseWriter):
         .. versionchanged:: 2.0.0
            Deprecated support for Timestep argument has now been removed.
            Use AtomGroup or Universe as an input instead.
+        .. versionchanged:: 2.1.0
+           When possible, TRRWriter assigns `ts.data['step']` to `step` rather
+           than `ts.frame`.
         """
         try:
             ts = ag.ts
@@ -124,7 +132,7 @@ class TRRReader(XDRBaseReader):
     """Reader for the Gromacs TRR format.
 
     The Gromacs TRR trajectory format is a lossless format. The TRR format can
-    store *velocoties* and *forces* in addition to the coordinates. It is also
+    store *velocities* and *forces* in addition to the coordinates. It is also
     used by other Gromacs tools to store and process other data such as modes
     from a principal component analysis.
 
--- a/package/MDAnalysis/coordinates/TRR.py
+++ b/package/MDAnalysis/coordinates/TRR.py
@@ -105,7 +105,7 @@ class TRRWriter(XDRBaseWriter):
                 self.convert_forces_to_native(forces)
 
         time = ts.time
-        step = ts.frame
+        step = ts.data.get('step', ts.frame)
 
         if self._convert_units:
             dimensions = self.convert_dimensions_to_unitcell(ts, inplace=False)
--- a/testsuite/MDAnalysisTests/coordinates/test_xdr.py
+++ b/testsuite/MDAnalysisTests/coordinates/test_xdr.py
@@ -29,7 +29,9 @@ import os
 import shutil
 import subprocess
 
-from numpy.testing import (assert_equal, assert_almost_equal)
+from numpy.testing import (assert_equal,
+                           assert_almost_equal,
+                           assert_allclose)
 
 from MDAnalysisTests import make_Universe
 from MDAnalysisTests.datafiles import (
@@ -442,6 +444,38 @@ class TestTRRWriter(_GromacsWriter):
             else:
                 with pytest.raises(mda.NoDataError):
                     getattr(written_ts, 'velocities')
+
+    def test_data_preservation(self, universe, Writer, outfile):
+
+        with Writer(outfile, universe.atoms.n_atoms, dt=universe.trajectory.dt) as W:
+            for ts in universe.trajectory:
+                W.write(universe)
+
+        uw = mda.Universe(GRO, outfile)
+
+        assert np.isclose(ts.data['time'], 0.0)
+        assert ts.data['step'] == 0
+        assert np.isclose(ts.data['lambda'], 0.0)
+        assert np.isclose(ts.data['dt'], 100.0)
+
+        # check that the data are identical for each time step
+        for orig_ts, written_ts in zip(universe.trajectory, uw.trajectory):
+            # data lengths must be the same
+            assert len(written_ts.data) == len(orig_ts.data)
+
+            # check that the keys exist in both dictionaries
+            for k in orig_ts.data:
+                assert k in written_ts.data
+
+            err_msg = ('mismatch between '
+                       'original and written trajectory at '
+                       f'frame {orig_ts.frame} vs {written_ts.frame}')
+
+            # check that each value is the same
+            for k in orig_ts.data:
+                assert_allclose(orig_ts.data[k],
+                                written_ts.data[k],
+                                err_msg=err_msg)
 
 
 class _GromacsWriterIssue101(object):
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -13,12 +13,13 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" http://semver.org
 
 ------------------------------------------------------------------------------
-??/??/?? IAlibay, melomcr, mdd31 
+??/??/?? IAlibay, melomcr, mdd31, ianmkenney
 
  * 2.1.0
 
 Fixes
   * Fix ITPParser charge reading (Issue #3419).
+  * TRRWriter preserves timestep data (Issue #3350).
   
 Enhancements
 


hmacdope (changes requested):
Thanks for the PR @ianmkenney!
 Would you be able to add a test to cover the writing behaviour?

ianmkenney:
@hmacdope Test added.

richardjgowers (changes requested):
Ch ch ch ch CHANGELOG

testsuite/MDAnalysisTests/coordinates/test_xdr.py
```diff
@@ -443,6 +443,21 @@ def test_gaps(self, universe, Writer, outfile):
...
+        with Writer(outfile, universe.atoms.n_atoms, dt=universe.trajectory.dt) as W:
+            for ts in universe.trajectory:
+                W.write(universe)
+
+        uw = mda.Universe(GRO, outfile)
+
+        # check that the data are identical for each time step
+        for orig_ts, written_ts in zip(universe.trajectory, uw.trajectory):
+            assert orig_ts.data == written_ts.data, ('mismatch between '
+                'original and written trajectory at '
```
hmacdope: Could you please use an fstring rather than % format here :)

testsuite/MDAnalysisTests/coordinates/test_xdr.py
```diff
@@ -443,6 +443,21 @@ def test_gaps(self, universe, Writer, outfile):
...
+    def test_data(self, universe, Writer, outfile):
+        with Writer(outfile, universe.atoms.n_atoms, dt=universe.trajectory.dt) as W:
+            for ts in universe.trajectory:
+                W.write(universe)
+
+        uw = mda.Universe(GRO, outfile)
+
+        # check that the data are identical for each time step
+        for orig_ts, written_ts in zip(universe.trajectory, uw.trajectory):
+            assert orig_ts.data == written_ts.data, ('mismatch between '
```
hmacdope: This will also pass if both are 'None' or empty. Can we also check that 'ts.data' contains the data we expect? 

Follow up question is there data in ts.data for the original file?

hmacdope (changes requested):
Just a nitpick and a question :)

hmacdope:
Also need to address PEP8.

testsuite/MDAnalysisTests/coordinates/test_xdr.py
ianmkenney: `ts.data` contains this information. I was just patching the behavior described in #3350 with the solution described by https://github.com/MDAnalysis/mdanalysis/issues/3350#issuecomment-864844782 which indicates that `ts.data['step']` might not exist in some cases. I can check the contents of the test file, if that's what you mean.

```
{'time': 0.0, 'step': 0, 'lambda': 0.0, 'dt': 100.00000762939453}
{'time': 100.00000762939453, 'step': 50000, 'lambda': 0.0, 'dt': 100.00000762939453}
{'time': 200.00001525878906, 'step': 100000, 'lambda': 0.0, 'dt': 100.00000762939453}
{'time': 300.0, 'step': 150000, 'lambda': 0.0, 'dt': 100.00000762939453}
{'time': 400.0000305175781, 'step': 200000, 'lambda': 0.0, 'dt': 100.00000762939453}
{'time': 500.0000305175781, 'step': 250000, 'lambda': 0.0, 'dt': 100.00000762939453}
{'time': 600.0, 'step': 300000, 'lambda': 0.0, 'dt': 100.00000762939453}
{'time': 700.0000610351562, 'step': 350000, 'lambda': 0.0, 'dt': 100.00000762939453}
{'time': 800.0000610351562, 'step': 400000, 'lambda': 0.0, 'dt': 100.00000762939453}
{'time': 900.0000610351562, 'step': 450000, 'lambda': 0.0, 'dt': 100.00000762939453}
```

ianmkenney:
@hmacdope how offensive do you find this to be? I can also make this change in other parts of the file if you want.

```
n_atoms = universe.atoms.n_atoms                                                                                                                                                    
dt = universe.trajectory.dt                                                                                                                                                         
                                                                                                                                                                          
with Writer(outfile, n_atoms, dt) as W:                                                                                                                                             
    for ts in universe.trajectory:                                                                                                                                                  
    W.write(universe)  
```

testsuite/MDAnalysisTests/coordinates/test_xdr.py
ianmkenney: Sure

testsuite/MDAnalysisTests/coordinates/test_xdr.py
hmacdope: Sorry I should have been clearer, my bad, was just checking that `ts.data` has something in it in the test file. All sorted.

hmacdope:
You mean as a replacement for:

```python
    def test_data(self, universe, Writer, outfile):
        with Writer(outfile, universe.atoms.n_atoms, dt=universe.trajectory.dt) as W:
            for ts in universe.trajectory:
                W.write(universe)
```
I kind of like the way that it is now, as everything is nice and encapsulated in its owning object (no unnecessary variables), but it is your code and if you think its clearer the second way then go for it.

 Other people may also have an opinion @lilyminium?

Perhaps leave the rest of the file the way it is. We generally avoid cosmetic changes for their own sake, as often original writers have made something look the way it does for a reason (often one that you may not expect), so always safer to leave something in its original state.

ianmkenney:
@hmacdope I have the same opinion as you regarding the extra variables. I'd also rather contribute something consistent. That said, I've also updated the CHANGELOG.

testsuite/MDAnalysisTests/coordinates/test_xdr.py
```diff
@@ -443,6 +443,26 @@ def test_gaps(self, universe, Writer, outfile):
...
 
+    def test_data_preservation(self, universe, Writer, outfile):
+
+        with Writer(outfile, universe.atoms.n_atoms, dt=universe.trajectory.dt) as W:
+            for ts in universe.trajectory:
+                W.write(universe)
+
+        uw = mda.Universe(GRO, outfile)
+
+        assert ts.data['time'] is not None
```
IAlibay: There's no cost here between checking that the contents exist or that they match a certain value. Can we do the latter for these instead?

testsuite/MDAnalysisTests/coordinates/test_xdr.py
```diff
@@ -443,6 +443,26 @@ def test_gaps(self, universe, Writer, outfile):
...
+        assert ts.data['step'] is not None
+        assert ts.data['lambda'] is not None
+        assert ts.data['dt'] is not None
+
+        # check that the data are identical for each time step
+        for orig_ts, written_ts in zip(universe.trajectory, uw.trajectory):
+            err_msg = ('mismatch between '
+                       'original and written trajectory at '
+                       f'frame {orig_ts.frame} vs {written_ts.frame}')
+            assert orig_ts.data == written_ts.data, err_msg
```
IAlibay: are we likely to be comparing floats here? if so we should use an almost_equal for the data components that are floats?

package/MDAnalysis/coordinates/TRR.py
```diff
@@ -105,7 +105,7 @@ def _write_next_frame(self, ag):
                 self.convert_forces_to_native(forces)
 
         time = ts.time
-        step = ts.frame
+        step = ts.data.get('step', ts.frame)
```
IAlibay, line 116: Can you add a versionchanged entry please?

Can you maybe update the blurb on line 48? Whilst you're here, can you fix the typo on line 44 ("velocoties").

IAlibay (changes requested):
Couple of comments - mostly updating the docs.

richardjgowers (approved):
I'm happy once @IAlibay is happy

testsuite/MDAnalysisTests/coordinates/test_xdr.py
ianmkenney: Checking with the values found in the first frame rather than `not None`

testsuite/MDAnalysisTests/coordinates/test_xdr.py
ianmkenney: using `assert_equal` on the dictionary itself

package/MDAnalysis/coordinates/TRR.py
ianmkenney: Fixed two typos ('velocoties'), added versionchanged for 2.1.0, and appended to the docstring following line 48 as an example.

testsuite/MDAnalysisTests/coordinates/test_xdr.py
```diff
@@ -443,6 +443,26 @@ def test_gaps(self, universe, Writer, outfile):
...
+        assert ts.data['step'] == 0
+        assert np.isclose(ts.data['lambda'], 0.0)
+        assert np.isclose(ts.data['dt'], 100.0)
+
+        # check that the data are identical for each time step
+        for orig_ts, written_ts in zip(universe.trajectory, uw.trajectory):
+            err_msg = ('mismatch between '
+                       'original and written trajectory at '
+                       f'frame {orig_ts.frame} vs {written_ts.frame}')
+            assert_equal(orig_ts.data, written_ts.data, err_msg=err_msg)
```
IAlibay: I think you want assert_almost_equal for floats

IAlibay (changes requested):
Just the one final thing

testsuite/MDAnalysisTests/coordinates/test_xdr.py
hmacdope: Note that `assert_almost_equal` is being phased out in favour of `assert_allclose` but makes no difference here AFAIK. Im happy either way.

ianmkenney: Just going to use `assert_allclose`

IAlibay (approved):

IAlibay:
thanks for dealing with this @ianmkenney

IAlibay:
@hmacdope you have a blocking review - since you've commented since I'll wait until tomorrow morning for your approval, after that I'll squash-merge (and dismiss blocking reviews if necessary).

hmacdope (approved):
LGTM! Thanks @ianmkenney

Issue #3387, opened 2021-08-15
Easy inverse index array


lilyminium:
## Is your feature request related to a problem? ##

PR #3368 creates an inverse index array of unsorted indices with this Python loop:

```python

            mask = np.zeros_like(self.ix)
            for i, x in enumerate(indices):
                values = np.where(self.ix == x)[0]
                mask[values] = i
            self._unique_restore_mask = mask
```

## Describe the solution you'd like ##

This could be done very easily and much faster in Cython instead, in the ``lib`` module.

SanjanaSogimatt:
Hey, I am new to open source so it would be great if I can help to solve this issue.

IAlibay:
@SanjanaSogimatt please do go ahead and open up a PR with code to solve this issue.

SanjanaSogimatt:
Hey actually I'm new to open source. Can you give me some suggestions on how to get started? It would be really a great help

IAlibay:
@SanjanaSogimatt please have a look at our userguide on contributing to MDAnalysis: https://userguide.mdanalysis.org/stable/contributing.html

If you open up a pull request which addresses this issue we can guide you through the process of merging your code into our codebase.

lilyminium:
I added a comment elaborating on this problem here: https://github.com/MDAnalysis/mdanalysis/pull/3433#issuecomment-940087191

kainszs:
Hi,

I am applying to GSoC and would like to work on this issue. I didn't see the issue among the open PRs, but apparently several people have already worked on it. 

I've been wanting to get to grips with cython for a while and this is a nice opportunity.

Is it possible for us to flag this issue so that no one else is working on it at the same time? I have an important exam next week, so I will probably use up the remaining time to work on it.

How much time should I plan for the PR if I want to have the code merged by April 2nd?

I look forward to your feedback.

Best,
Kai

Syedmahmood777:
Hey, I am new to contributing, can I solve this or is it already solved?

bobbysharma05:
Hey, I want to work on this issue. Can you please assigned it to me.

PR #3433, opened 2021-10-10
Easy inverse index array #3387

SanjanaSogimatt:
Fixes #

 - Functions used to fix the bug
 - There are two methods used and one is commented out 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


SanjanaSogimatt:
Hey, I want to know is my code good or like I would like to improve it.

SanjanaSogimatt:
Hello, I wanted to know how to improve my code.

tylerjereddy (commented):
This doesn't reference any MDAnalysis issue, or adjust any MDAnalysis source code, so I'm going to close it and suggest browsing our issues/docs if you'd like to contribute.

If you'd like feedback on improving your generic Python code you might try something like StackOverflow.

lilyminium:
Hi @SanjanaSogimatt,

Thanks for having a look at this issue and for linking it in the title. As Tyler points out, it would have been more beneficial to add the issue number after the `Fixes #` line as that enables GitHub to link the pull request like in the screenshot below -- this makes it much easier for reviewers to track the discussion, especially in a large project like MDAnalysis.

Screen Shot 2021-10-11 at 7 31 15 am


Issue #3387 describes how PR #3368 adds this code to the library:

https://github.com/MDAnalysis/mdanalysis/blob/149eb504e35670568201af28306fa04462f6c100/package/MDAnalysis/core/groups.py#L811-L815

What this code does is create an *inverse* array, akin to what you get below (where `np` stands for [numpy](https://numpy.org/)):

`sorted_unique_array, inverse_array = np.unique(my_list, return_inverse=True)`

What this `inverse_array` does is:

`sorted_unique_array[inverse_array] == my_list`


This is not the same thing as *reversing* an array.

The reason I couldn't use the easy `np.unique` command is because `np.unique` sorts the output `sorted_unique_array`. If you go to the `groups.py` file by clicking on the link, you'll see that the indices that are used in the function are specifically *unsorted*, but unique. https://github.com/MDAnalysis/mdanalysis/blob/149eb504e35670568201af28306fa04462f6c100/package/MDAnalysis/core/groups.py#L809

So for example:

[ 1, 5, 3, 3, 6,] is *un*sorted and *not* unique
[ 1, 3, 3, 5, 6,] is sorted and *not* unique
[ 1, 5, 3, 6,] is *un*sorted and unique
[ 1, 3, 5, 6,] is sorted and unique

So what the original code in #3387 does is create an inverse array for the third scenario there, i.e. `[ 1, 5, 3, 6,]`. The ideal goal is that if `[ 1, 5, 3, 3, 6,]` is the input, then

```np.array([ 1, 5, 3, 6,])[inverse_array] == [ 1, 5, 3, 3, 6,]```.

However, I think that the original code in #3387 is clumsy and slow. The goal is to rewrite it in [Cython](https://cython.org/) to make it faster. An example of a similar function would be here:

https://github.com/MDAnalysis/mdanalysis/blob/149eb504e35670568201af28306fa04462f6c100/package/MDAnalysis/lib/_cutil.pyx#L51-L91

To fix Issue #3387, the solution would need to:

1. Implement a Cython function in _cutil.pyx, the file linked there
2. Show that it's faster by benchmarking the speedup. The original code would need to be pulled out and turned into its own function for straightforward comparison
3. Add a test to make sure that the Cython function works
4. Call this Cython function from groups.py, replacing the [original code](https://github.com/MDAnalysis/mdanalysis/blob/149eb504e35670568201af28306fa04462f6c100/package/MDAnalysis/core/groups.py#L811-L815)
5. Open a pull request and verify that the testsuite passes

If you're happy to do this then please reopen this PR, or open another one, when you're ready!

PR #3444, opened 2021-10-24
Issue 3443 coordinateless objects

### LINKED ISSUE #3443: Custom Topology readers can't create topology-only Universe

``` python
class Thingy:
    def __init__(self, val):
	self.v = val


class ThingyParser(TopologyReaderBase):
    format='THINGY'

    @staticmethod
    def _format_hint(thing):
        return isinstance(thing, Thingy)

    def parse(self, **kwargs):
        return mda.core.topology.Topology(n_atoms=10)

class TestOnlyTopology:
    def test_only_top(self):
        t = Thingy(20)

	u = mda.Universe(t)

        assert len(u.atoms) == 10

```

This should work (using the logic from [here](https://github.com/MDAnalysis/mdanalysis/blob/develop/package/MDAnalysis/core/universe.py#L338)) and provide a Universe without a Reader attached.  Instead it raises a TypeError from [here](https://github.com/MDAnalysis/mdanalysis/blob/develop/package/MDAnalysis/lib/util.py#L988)

- Which operating system?

richardjgowers:
Fixes #3443 

 - fixes the check on if a filename (here an object) can also be a coordinate file as well as topology

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -13,14 +13,16 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" http://semver.org
 
 ------------------------------------------------------------------------------
-??/??/?? IAlibay, melomcr, mdd31, ianmkenney
+??/??/?? IAlibay, melomcr, mdd31, ianmkenney, richardjgowers
 
  * 2.1.0
 
 Fixes
   * Fix ITPParser charge reading (Issue #3419).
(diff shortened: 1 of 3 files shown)


lilyminium (approved):
LGTM!

PR #3445, opened 2021-10-28
Rs chirality

richardjgowers:
Fixes #

 - now reads R/S chirality from RDKit molecules

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -26,6 +26,7 @@ Fixes
 Enhancements
   * Add option for custom compiler flags for C/C++ on build and remove march
     option from setup.cfg (Issue #3428, PR #3429).
+  * Added R/S Chirality support to RDKit parsed molecules
 
 Changes
   * Dropped python 3.6 support and raised minimum numpy version to 1.18.0
--- a/package/MDAnalysis/converters/RDKitParser.py
+++ b/package/MDAnalysis/converters/RDKitParser.py
@@ -73,16 +73,12 @@ from ..core.topology import Topology
 logger = logging.getLogger("MDAnalysis.converters.RDKitParser")
 
 
-def _rdkit_chiral_to_RS(val):
-    # convert RDKit chirality tag to R/S/''
-    from rdkit import Chem
-
-    if val == Chem.ChiralType.CHI_TETRAHEDRAL_CW:
-        return 'R'
-    elif val == Chem.ChiralType.CHI_TETRAHEDRAL_CCW:
-        return 'S'
-    else:
-        return ''
+def _rdkit_atom_to_RS(atom):
+    """Fetches RDKit chiral tags"""
+    try:
+        return atom.GetProp("_CIPCode")
+    except KeyError:
+        return ""
 
 
 class RDKitParser(TopologyReaderBase):
--- a/package/MDAnalysis/converters/RDKitParser.py
+++ b/package/MDAnalysis/converters/RDKitParser.py
@@ -79,7 +79,7 @@ def _rdkit_chiral_to_RS(val):
 
     if val == Chem.ChiralType.CHI_TETRAHEDRAL_CW:
         return 'R'
-    elif val == Chem.ChiralType.CHI_TETRAHEDAL_CCW:
+    elif val == Chem.ChiralType.CHI_TETRAHEDRAL_CCW:
         return 'S'
     else:
         return ''
--- a/package/MDAnalysis/converters/RDKitParser.py
+++ b/package/MDAnalysis/converters/RDKitParser.py
@@ -60,6 +60,7 @@ from ..core.topologyattrs import (
     Resids,
     Resnums,
     Resnames,
+    RSChirality,
     Segids,
     AltLocs,
     ChainIDs,
@@ -70,6 +71,18 @@ from ..core.topologyattrs import (
 from ..core.topology import Topology
 
 logger = logging.getLogger("MDAnalysis.converters.RDKitParser")
+
+
+def _rdkit_chiral_to_RS(val):
+    # convert RDKit chirality tag to R/S/''
+    from rdkit import Chem
+
+    if val == Chem.ChiralType.CHI_TETRAHEDRAL_CW:
+        return 'R'
+    elif val == Chem.ChiralType.CHI_TETRAHEDAL_CCW:
+        return 'S'
+    else:
+        return ''
 
 
 class RDKitParser(TopologyReaderBase):
@@ -85,6 +98,7 @@ class RDKitParser(TopologyReaderBase):
      - Bonds
      - Resids
      - Resnums
+     - RSChirality
      - Segids
 
     Guesses the following:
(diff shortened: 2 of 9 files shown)


cbouy:
I'll have a proper look next week but overall looking good!
For the tests you could directly `Universe.from_smiles` ?

package/MDAnalysis/converters/RDKitParser.py
```diff
@@ -72,6 +73,18 @@
...
+def _rdkit_chiral_to_RS(val):
+    # convert RDKit chirality tag to R/S/''
+    from rdkit import Chem
+
+    if val == Chem.ChiralType.CHI_TETRAHEDRAL_CW:
+        return 'R'
+    elif val == Chem.ChiralType.CHI_TETRAHEDRAL_CCW:
+        return 'S'
+    else:
+        return ''
```
cbouy: So it turns out that `Chem.ChiralType.CHI_TETRAHEDRAL_CW` and CCW are based on the atoms order, which are not guaranteed to match with the CIP order according to [this thread](https://www.mail-archive.com/rdkit-discuss@lists.sourceforge.net/msg09255.html).

However, the chiral tags are available as an atom property when reading a molecule made from an isomeric SMILES or if `Chem.AssignStereochemistryFrom3D` was called, so everything is not lost!

```suggestion
def _rdkit_atom_to_RS(atom):
    """Fetches RDKit chiral tags"""
    try:
        return atom.GetProp("_CIPCode")
    except KeyError:
        return ""
```

package/MDAnalysis/converters/RDKitParser.py
```diff
@@ -124,6 +138,8 @@ class RDKitParser(TopologyReaderBase):
     | atom.GetMonomerInfo().GetName()             | names                   |
     | atom.GetProp('_TriposAtomName')             |                         |
     +---------------------------------------------+-------------------------+
+    | atom.GetChiralTag()                         | chiralities             |
```
cbouy:
```suggestion
    | atom.GetProp('_CIPCode')                    | chiralities             |
```

package/MDAnalysis/converters/RDKitParser.py
```diff
@@ -204,6 +223,7 @@ def parse(self, **kwargs):
             elements.append(atom.GetSymbol())
             masses.append(atom.GetMass())
             aromatics.append(atom.GetIsAromatic())
+            chiralities.append(_rdkit_chiral_to_RS(atom.GetChiralTag()))
```
cbouy:
```suggestion
            chiralities.append(_rdkit_atom_to_RS(atom))
```

package/MDAnalysis/core/selection.py
```diff
@@ -567,6 +567,29 @@ def _apply(self, group):
...
+
+        self.values = vals
+
+    @return_empty_on_apply
+    def _apply(self, group):
+        attr = getattr(group, self.field)
+
+        mask = (attr == self.values[0])
+        for v in self.values[1:]:
+            mask |= (attr == v)
```
cbouy: Any reason not to use np.isin ?
```suggestion
        mask = np.isin(attr, self.values)
```

testsuite/MDAnalysisTests/core/test_atomselections.py
```diff
@@ -1405,3 +1406,36 @@ def test_unique_selection_on_ordered_group(u_pdb_icodes, sel, sort, ix):
...
+
+    m = Chem.MolFromSmiles(smi)
+    u = mda.Universe(m)
+
+    assert hasattr(u.atoms, 'chiralities')
+
+    assert u.atoms[0].chirality == ''
+    assert u.atoms[1].chirality == chirality
+
+    assert_equal(u.atoms[:3].chiralities, np.array(['', chirality, ''], dtype='U1'))
```
cbouy, line 1426: Isn't this a bit redundant? Or you want a specific test for the dtype?

package/MDAnalysis/core/selection.py
cbouy: Oh wait I'm not sure if self.values will be a string like `'R S'` or a list like `['R', 'S']`. It needs to be a list to work with np.isin

cbouy (changes requested):
Could you also add a "roundabout" test ?
You start with an isomeric smiles to create an rdkit molecule, gather the chiral tags with `_rdkit_atom_to_RS` in an array, then convert the molecule to a Universe, then convert the Universe to an RDKit molecule, gather the chiral tags again and compare them.

package/MDAnalysis/core/selection.py
richardjgowers: `self.values` will be a list, as we consume all tokens until the next keyword.

testsuite/MDAnalysisTests/core/test_atomselections.py
richardjgowers: Yeah getting an array for an AtomGroup is a ilttle different to single values from an Atom, worth checking both.

richardjgowers:
thanks @cbouy !

cbouy (commented):
I think it's just missing some docs in `AtomGroup.select_atoms` and then we should be good!

richardjgowers:
Cool done thanks @cbouy , @MDAnalysis/coredevs probably good to go now!

IAlibay:
I've restarted CI to see if the py3.9 MacOS failures were just random.

package/CHANGELOG
```diff
@@ -26,6 +26,7 @@ Fixes
 Enhancements
   * Add option for custom compiler flags for C/C++ on build and remove march
     option from setup.cfg (Issue #3428, PR #3429).
+  * Added R/S Chirality support to RDKit parsed molecules
```
IAlibay, line 29: tsk tsk tsk - newest entry first

package/MDAnalysis/core/selection.py
```diff
@@ -567,6 +567,27 @@ def _apply(self, group):
...
+        self.values = vals
+
+    @return_empty_on_apply
+    def _apply(self, group):
+        attr = getattr(group, self.field)
+
+        mask = np.isin(attr, self.values)
+
+        return group[mask]
+
```
IAlibay, line 590:
```suggestion


```

package/MDAnalysis/core/topologyattrs.py
```diff
@@ -1957,6 +1957,13 @@ def _gen_initial_values(na, nr, ns):
         return np.zeros(na, dtype=bool)
 
 
+class RSChirality(AtomAttr):
+    "R/S chirality"
```
IAlibay: Weird - are neither of these covered by tests?

IAlibay:
```suggestion
    """R/S chirality"""
```
Mostly to keep to the style.

IAlibay (changes requested):
Couple of tiny things - main this is I can't work out why codecov is complaining about coverage for those two topologyattr sections. I'll restart CI, if it's still happening afterwards could you look into it @richardjgowers ?

package/MDAnalysis/core/topologyattrs.py
richardjgowers: L1961 is a docstring, so I think this is just a codecov snafu

IAlibay (approved):
Still not sure what's up with that coverage issue, but 🤷🏽‍♂️

PR #3454, opened 2021-11-07
Make HydrogenBondAnalysis handle empty selection strings

### LINKED ISSUE #3453: HydrogenBondAnalysis cannot handle empty selection strings
## Expected behavior ##

If I pass an empty selection string to `HydrogenBondAnalysis`, no hydrogen bonds will be found.

## Actual behavior ##

If I pass an empty selection string to `HydrogenBondAnalysis`, hydrogen bonding atoms are guessed (using `guess_donors`, `guess_hydrogens`, and `guess_acceptors`), and a non-zero number of hydrogen bonds is found.

This is because `if not self.donors_sel` is checked rather than `if self.donors_sel is None` (the former will evaluate to False for an empty string).

Thanks for Dina Sharon for raising this issue in the [mailing list](https://groups.google.com/g/mdnalysis-discussion/c/NiptD2G4pko/m/cCyeHHP4AAAJ).

``` python
import MDAnalysis as mda

u = mda.Universe(PSF, DCD)

    universe=u,
    hydrogens_sel='',
    acceptors_sel='',
)

hbonds.run()

print(hbonds.count_by_time().sum())

```

Prints `10858`.

2.1.0-dev0
Python 3.8.0
- Which operating system?
Ubuntu 20.04.2 LTS

p-j-smith:
Fixes #3453

 - in `HydrogenBondAnalysis`, acceptors and donors are guessed `if self.acceptors_sel is None` and `if self.donors_sel is None`, rather than e.g. `if not self.acceptors_sel`.
- tests added that pass empty selection strings to `HydrogenBondAnalysis`

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
+++ b/package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
@@ -328,6 +328,10 @@ class HydrogenBondAnalysis(AnalysisBase):
         if isinstance(acceptors_sel, str) and not acceptors_sel:
             warnings.warn(msg.format("acceptors_sel"))
 
+        self.donors_sel = donors_sel
+        self.hydrogens_sel = hydrogens_sel
+        self.acceptors_sel = acceptors_sel
+
         # If hydrogen bonding groups are selected, then generate
         # corresponding atom groups
         if between is not None:
--- a/package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
+++ b/package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
@@ -320,7 +320,7 @@ class HydrogenBondAnalysis(AnalysisBase):
         msg = ("{} is an empty selection string - no hydrogen bonds will "
                "be found. This may be intended, but please check your "
                "selection."
-        )
+               )
         if isinstance(donors_sel, str) and not donors_sel:
             warnings.warn(msg.format("donors_sel"))
         if isinstance(hydrogens_sel, str) and not hydrogens_sel:
--- a/package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
+++ b/package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
@@ -317,16 +317,16 @@ class HydrogenBondAnalysis(AnalysisBase):
         self.u = universe
         self._trajectory = self.u.trajectory
 
-        msg = ("{} is an empty selection string - no hydrogen bonds will "
+        msg = ("{}_sel is an empty selection string - no hydrogen bonds will "
                "be found. This may be intended, but please check your "
                "selection."
                )
-        if isinstance(donors_sel, str) and not donors_sel:
-            warnings.warn(msg.format("donors_sel"))
-        if isinstance(hydrogens_sel, str) and not hydrogens_sel:
-            warnings.warn(msg.format("hydrogens_sel"))
-        if isinstance(acceptors_sel, str) and not acceptors_sel:
-            warnings.warn(msg.format("acceptors_sel"))
+        for sel, sel_string in zip(
+            [donors_sel, hydrogens_sel, acceptors_sel],
+            ['donors', 'hydrogens', 'acceptors']
+        ):
+            if isinstance(sel, str) and not sel:
+                warnings.warn(msg.format(sel_string))
 
         self.donors_sel = donors_sel
         self.hydrogens_sel = hydrogens_sel
--- a/package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
+++ b/package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
@@ -316,9 +316,17 @@ class HydrogenBondAnalysis(AnalysisBase):
 
         self.u = universe
         self._trajectory = self.u.trajectory
-        self.donors_sel = donors_sel
-        self.hydrogens_sel = hydrogens_sel
-        self.acceptors_sel = acceptors_sel
+
+        msg = ("{} is an empty selection string - no hydrogen bonds will "
+               "be found. This may be intended, but please check your "
+               "selection."
+        )
+        if isinstance(donors_sel, str) and not donors_sel:
+            warnings.warn(msg.format("donors_sel"))
+        if isinstance(hydrogens_sel, str) and not hydrogens_sel:
+            warnings.warn(msg.format("hydrogens_sel"))
+        if isinstance(acceptors_sel, str) and not acceptors_sel:
+            warnings.warn(msg.format("acceptors_sel"))
 
         # If hydrogen bonding groups are selected, then generate
         # corresponding atom groups
--- a/package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
+++ b/package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
@@ -439,7 +439,7 @@ class HydrogenBondAnalysis(AnalysisBase):
 
         # We need to know `hydrogens_sel` before we can find donors
         # Use a new variable `hydrogens_sel` so that we do not set `self.hydrogens_sel` if it is currently `None`
-        if not self.hydrogens_sel:
+        if self.hydrogens_sel is None:
             hydrogens_sel = self.guess_hydrogens()
         else:
             hydrogens_sel = self.hydrogens_sel
@@ -512,7 +512,7 @@ class HydrogenBondAnalysis(AnalysisBase):
         """
 
         # If donors_sel is not provided, use topology to find d-h pairs
-        if not self.donors_sel:
+        if self.donors_sel is None:
 
             # We're using u._topology.bonds rather than u.bonds as it is a million times faster to access.
             # This is because u.bonds also calculates properties of each bond (e.g bond length).
@@ -583,9 +583,9 @@ class HydrogenBondAnalysis(AnalysisBase):
         self.results.hbonds = [[], [], [], [], [], []]
 
         # Set atom selections if they have not been provided
-        if not self.acceptors_sel:
+        if self.acceptors_sel is None:
             self.acceptors_sel = self.guess_acceptors()
-        if not self.hydrogens_sel:
+        if self.hydrogens_sel is None:
             self.hydrogens_sel = self.guess_hydrogens()
 
         # Select atom groups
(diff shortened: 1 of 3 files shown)


testsuite/MDAnalysisTests/analysis/test_hydrogenbonds_analysis.py
```diff
@@ -552,3 +552,28 @@ def test_count_by_type(self, h):
...
+        kwargs = {
+            'donors_sel': '',
+            'hydrogens_sel': '',
+            'acceptors_sel': '',
+        }
+
+        h = HydrogenBondAnalysis(universe, **kwargs)
+        h.run()
+
+        return h
```
tylerjereddy: Minor point--is it likely this fixture will be used again? Just wondering about the value of the abstraction/fixture magic, but probably doesn't matter much.

p-j-smith: Yeah good point, I can't see it being used gain so I've removed the fixture

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -316,9 +316,17 @@ def __init__(self, universe,
 
         self.u = universe
         self._trajectory = self.u.trajectory
-        self.donors_sel = donors_sel
-        self.hydrogens_sel = hydrogens_sel
-        self.acceptors_sel = acceptors_sel
```
IAlibay, line 321: These aren't being allocated anymore so you don't get them further below.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -316,9 +316,17 @@ def __init__(self, universe,
...
+
+        msg = ("{} is an empty selection string - no hydrogen bonds will "
+               "be found. This may be intended, but please check your "
+               "selection."
+        )
+        if isinstance(donors_sel, str) and not donors_sel:
```
IAlibay: I feel like this could be squashed into a `for` loop, thoughts?

IAlibay (changes requested):

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
p-j-smith: I've put those lines into a for loop now. It's slightly less readable to me, but that might just be me, and I'm happy to leave it this way

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -316,6 +316,18 @@ def __init__(self, universe,
...
+
+        msg = ("{}_sel is an empty selection string - no hydrogen bonds will "
+               "be found. This may be intended, but please check your "
+               "selection."
+               )
+        for sel, sel_string in zip(
+            [donors_sel, hydrogens_sel, acceptors_sel],
+            ['donors', 'hydrogens', 'acceptors']
+        ):
+            if isinstance(sel, str) and not sel:
```
IAlibay: Do you also want to test for the presence of blank space here? if so probably worth evaluating sel.strip()

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -316,6 +316,18 @@ def __init__(self, universe,
 
         self.u = universe
         self._trajectory = self.u.trajectory
+
+        msg = ("{}_sel is an empty selection string - no hydrogen bonds will "
+               "be found. This may be intended, but please check your "
+               "selection."
+               )
+        for sel, sel_string in zip(
```
IAlibay: So the line optimising person in me thinks that this would be a fun way to squash a few more lines:

```python
for sel in ['donors_sel', 'hydrogens_sel', 'acceptors_sel']:
    var = eval(sel)
    setattr(self, sel, var)
    if isinstance(var, str) and not var.strip():
        warnings.warn(msg.format(sel))
```

Whilst I _think_ it's readable to me, maybe others would disagree - @lilyminium thoughts?

lilyminium: If you didn't want to mess with `eval` you could achieve a similar effect by examining `self`. e.g. (below setting self.donors_sel etc):

```
for sel in ['donors_sel', 'hydrogens_sel', 'acceptors_sel']:
    val = getattr(self, sel)
    if isinstance(val, str) and not val:
        # do the warning
```

lilyminium: Should this only be checking for the empty string? Or should you check for all cases where nothing is selected (e.g. `name NOT_A_REAL_ATOM`?) Otherwise it seems to me this PR could work just by changing the check to `if self.xx_sel is None: guess` as you do below, without the warning, and let the selection parser handle the empty string however it does.

p-j-smith: I think it should only check for an empty string, because some selections might have empty atom groups at some frames but not others (e.g. those that include `around`). Also, relying on the selection parser would give the following response:

`"Empty string to select atoms, empty group returned."`

whereas it might be useful for the user to know which selection string was empty.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -316,10 +316,20 @@ def __init__(self, universe,
...
         self.hydrogens_sel = hydrogens_sel
         self.acceptors_sel = acceptors_sel
 
+        msg = ("{} is an empty selection string - no hydrogen bonds will "
+               "be found. This may be intended, but please check your "
+               "selection."
+               )
+        for sel in ['donors_sel', 'hydrogens_sel', 'acceptors_sel']:
+            val = getattr(self, sel)
+            if isinstance(val, str) and not val:
```
IAlibay, line 330: I think my previous comment may have been lost in the discussions - I think you should be using `val.strip()` here to also catch the edge case of someone passing `" "` (which will evaluate to `True`)

IAlibay (changes requested):

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
p-j-smith: Ah yes, sorry about that! I'm pretty busy now (I have my viva tomorrow), but I'll sort it out either at the weekend or early next week

IAlibay: Since it's not a large change, if you don't mind I'll just push a quick update to deal with this w/ associated test. That way I can push this in to the 2.1.0 release.

edit: best of luck with the viva!

p-j-smith: That's would be great if you could take care of it.

Thank you!

p-j-smith: Thanks for merging the latest develop into this branch.


Rather than doing `sel.strip()` here, I've stripped the strings of whitespace when assigning them as instance attributes. This is because `select_atoms` [doesn't strip the whitespace](https://github.com/MDAnalysis/mdanalysis/blob/develop/package/MDAnalysis/core/groups.py#L3133-L3138), and so passing a whitespace-only string to it raises a `SelectionError`:

`Unknown selection token: 'None'`.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -316,9 +316,19 @@ def __init__(self, universe,
...
         self._trajectory = self.u.trajectory
-        self.donors_sel = donors_sel
-        self.hydrogens_sel = hydrogens_sel
-        self.acceptors_sel = acceptors_sel
+
+        self.donors_sel = donors_sel.strip() if donors_sel is not None else donors_sel
+        self.hydrogens_sel = hydrogens_sel.strip() if hydrogens_sel is not None else hydrogens_sel
+        self.acceptors_sel = acceptors_sel.strip() if acceptors_sel is not None else acceptors_sel
+
+        msg = ("{} is an empty selection string - no hydrogen bonds will "
```
IAlibay, line 324: I'll be honest - I'm still not sure in what scenario this would be a case that would be expected.

However - in the interest of getting this done I'll merge this but keep the underlying issue open. We can discuss this further there.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
IAlibay: There's possibly a viewpoint where getting the `SelectionError` might be a better thing, but I'm not sure it's worth holding up this PR.

testsuite/MDAnalysisTests/analysis/test_hydrogenbonds_analysis.py
```diff
@@ -552,3 +552,34 @@ def test_count_by_type(self, h):
...
+        return MDAnalysis.Universe(waterPSF, waterDCD)
+
+    msg = ("{} is an empty selection string - no hydrogen bonds will "
+           "be found. This may be intended, but please check your "
+           "selection."
+           )
+
+    @pytest.mark.parametrize('seltype',
+            ['donors_sel', 'hydrogens_sel', 'acceptors_sel'])
+    def test_empty_sel(self, universe, seltype):
```
IAlibay, line 571: I just squashed these tests to save some lines.

IAlibay (approved):
thanks @p-j-smith 

As mentioned in the comments - I might keep the underlying issue open just so we can discuss if a warning or an error should be thrown here (I'm sure I've forgotten some discussion we've already had about this).

I have also squashed some of the tests (just an aesthetic change really).

edit: ha! I guess I just forgot what we discussed: https://github.com/MDAnalysis/mdanalysis/issues/3453#issuecomment-963083372 🙃

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
lilyminium: Yeah probably not worth holding up the PR, but IMO whitespace etc should be addressed in the selection parser itself -- raised #3487 for this.

IAlibay: Thanks for raising the issue @lilyminium !

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
p-j-smith: Yeah that's true. If there's an empty selection string then there's no point in running the analysis - no hydrogen bonds can possible be found...

p-j-smith:
thanks @IAlibay !

PR #3458, opened 2021-11-11
Add minimal deps py3.10 to linux CI matrix

### LINKED ISSUE #3457: Add python 3.10 support
As per every year - a new python version is out. We should add it to the test matrix and see if anything breaks.

IAlibay:
Towards #3457 

 - Add minimal py3.10 build to linux CI matrix
   - Currently following dependencies are a problem for py3.10; `chemfiles-python`, `openmm`, `rdkit`
 - Replace use of biobuild channel with bioconda (better maintained?)

PR Checklist
------------
 - [x] Tests?
 - Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/.github/workflows/gh-ci.yaml
+++ b/.github/workflows/gh-ci.yaml
@@ -125,7 +125,7 @@ jobs:
         # software (curl, ssh, ...), but the version in biobuilds are very
         # outdated. This can lead to problem when loading shared libraries,
         # see https://github.com/MDAnalysis/mdanalysis/pull/3126#issuecomment-813112080
-        channels: conda-forge, biobuilds
+        channels: conda-forge, bioconda
         add-pip-as-python-dependency: true
         mamba-version: "*"
         architecture: x64
--- a/.github/workflows/gh-ci.yaml
+++ b/.github/workflows/gh-ci.yaml
@@ -22,7 +22,7 @@ defaults:
 
 env:
   MDA_CONDA_MIN_DEPS: "pip pytest mmtf-python biopython networkx cython matplotlib-base scipy griddataformats hypothesis gsd codecov threadpoolctl"
-  MDA_CONDA_EXTRA_DEPS: "seaborn>=0.7.0 clustalw=2.1 netcdf4 scikit-learn joblib>=0.12 tqdm>=4.43.0 tidynamics>=1.0.0 h5py"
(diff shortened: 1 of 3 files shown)


IAlibay:
Looks like some of our deps aren't resolving with py3.10 yet... it's unclear as to which, I'll give 3.10 on a minimal build a go.

IAlibay:
I'm going to put this on hold - looks like there's some issue with one or more deps, might have more luck later on.

IAlibay:
Build works fine with minimal deps but fails when attempting to do a full deps install - not sure which dep is breaking

testsuite/MDAnalysisTests/topology/test_top.py
```diff
@@ -493,19 +493,9 @@ def test_flag_index_error(self):
...
-                    warnings.filterwarnings(
-                            action='ignore',
-                            category=ImportWarning)
-                u = mda.Universe(parm)
-
-        assert len(record) == len(errmsgs)
-        # Assumes errmsgs list is in order of occurence
-        for i, msg in enumerate(errmsgs):
-            assert msg in str(record[i].message.args[0])
+            u = mda.Universe(parm)
```
IAlibay, line 496: With hindsight, this test testing a limited number of warnings is both unnecessary and in my opinion somewhat excessive. My suggestion here is just to drop this and go for a simpler "these warnings are raised" check.

.github/workflows/gh-ci.yaml
```diff
@@ -119,7 +119,7 @@ jobs:
         # software (curl, ssh, ...), but the version in biobuilds are very
         # outdated. This can lead to problem when loading shared libraries,
         # see https://github.com/MDAnalysis/mdanalysis/pull/3126#issuecomment-813112080
-        channels: conda-forge, biobuilds
+        channels: conda-forge, bioconda
```
IAlibay, line 128: As far as I'm aware biobuild still works, but it hasn't been updated since 2019? (https://github.com/biobuilds/biobuilds), if purely on a security basis, it probably is better to just use the better maintained bioconda here?

IAlibay:
@MDAnalysis/coredevs this should be good to go.

Only adding py3.10 to CI as a minimal dependency check. Note: we also face the same warning issues with dependencies using the `six` library as seen in #3461 (not limited to Win 10 unfortunately).

.github/workflows/gh-ci.yaml
orbeckst: I don't think that there's anything that ties us to biobuilds.

orbeckst (approved):
looks fine to me, and your CI is green

PR #3461, opened 2021-11-14
CI: add Win 3.10

tylerjereddy:
Deals with the Windows portion of gh-3457

 - adds Windows Python 3.10 testing to Azure CI to prevent any regressions there moving forward
 - for the moment, we skip `h5py` install on Python 3.10 and we no longer pin the version of `gsd`
 - we temporarily suppress the following Python 3.10 warning observed on Windows for a third-party library that is still importing `six`: `ImportWarning('_SixMetaPathImporter.find_spec() not found`
 - add the option to skip one test that had an obligate requirement for `h5py`, which seems more consistent anyway because we already have a bunch of tests that skip conditionally with `HAS_H5PY`
 - in my hands, this allows the test suite to pass on Windows w/ Python `3.10`; I do suggest that we open an issue to revert any of the minor workarounds in use here as the `3.10` dependencies/ecosystem matures in the future

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -47,7 +47,7 @@ jobs:
   - script: python -m pip install --upgrade pip setuptools wheel
     displayName: 'Install tools'
   - script: >-
-      python -m pip install
+      python -m pip install --only-binary=h5py
       cython
       hypothesis
       matplotlib
@@ -57,20 +57,17 @@ jobs:
       pytest-xdist
       scikit-learn
       scipy
+      h5py
       tqdm
       threadpoolctl
     displayName: 'Install dependencies'
-  - script: >-
-      python -m pip install --only-binary=h5py
-      "h5py<3.4.0"
-    displayName: 'Install h5py for Python < 3.10'
-    condition: and(succeeded(), ne(variables['PYTHON_VERSION'], '3.10'))
   # TODO: recent rdkit is not on PyPI
   - script: >-
       python -m pip install
       biopython
       chemfiles>=0.9
       duecredit
+      gsd
       joblib
       GridDataFormats
       mmtf-python
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -35,6 +35,9 @@ jobs:
         Python39-64bit-full:
           PYTHON_VERSION: '3.9'
           PYTHON_ARCH: 'x64'
+        Python310-64bit-full:
+          PYTHON_VERSION: '3.10'
+          PYTHON_ARCH: 'x64'
   steps:
   - task: UsePythonVersion@0
     inputs:
@@ -44,7 +47,7 @@ jobs:
   - script: python -m pip install --upgrade pip setuptools wheel
     displayName: 'Install tools'
   - script: >-
-      python -m pip install --only-binary=h5py
+      python -m pip install
       cython
       hypothesis
       matplotlib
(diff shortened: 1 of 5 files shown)


tylerjereddy:
The codecov failure looks spurious (we probably should never fail CI on codecov anyway, there can be various reasons not to want that).

The PEP8 complaint is probably debatable--feel free to push in an adjustment for that if you don't like the visual indents or whatever.

IAlibay:
@tylerjereddy is the py3.10 h5py skip a known issue?

tylerjereddy:
> @tylerjereddy is the py3.10 h5py skip a known issue?

@IAlibay At the time I opened this PR there were no Windows Python 3.10 binary wheels up on PyPI. I believe there now are, so you could just try removing that, or I can do that in the next few days if you want.

azure-pipelines.yml
```diff
@@ -54,17 +57,20 @@ jobs:
...
       scipy
-      "h5py<3.4.0"
       tqdm
       threadpoolctl
     displayName: 'Install dependencies'
+  - script: >-
+      python -m pip install --only-binary=h5py
+      "h5py<3.4.0"
+    displayName: 'Install h5py for Python < 3.10'
+    condition: and(succeeded(), ne(variables['PYTHON_VERSION'], '3.10'))
```
tylerjereddy: we could try removing this condition, though we'll need to unpin `h5py` to get the binary on newer Python, and not sure how that will work out..

tylerjereddy:
@IAlibay I've simplified the diff a bit here--I'm almost tempted to `xfail` the single Windows h5py-related test failures for now. Some kind of permissions issue.

We also need to get on top of the Windows version bump in CI--I think we have other issues/PRs for that, but version 2016 is going away soon I think.

testsuite/MDAnalysisTests/coordinates/test_chainreader.py
```diff
@@ -345,7 +348,17 @@ def test_easy_trigger_warning(self, tmpdir):
         sequences = ([0, 1, 2, 3], [2, 3, 4, 5], [4, 5, 6, 7])
         utop, fnames = build_trajectories(folder, sequences=sequences,)
         with no_warning(UserWarning):
-            mda.Universe(utop._topology, fnames, continuous=True)
+            with warnings.catch_warnings():
+                # for windows Python 3.10 ignore:
+                # ImportWarning('_SixMetaPathImporter.find_spec() not found
+                # TODO: remove when we no longer have a dependency
```
IAlibay, line 354: Seems sensible to me.

Could this be raised as an issue so we remember to deal with this in the future?

IAlibay (changes requested):
Only putting this as a "request changes" with the ask that we raise some issues + the h5py xfail.

Regarding h5py - I don't think this is something we can fix immediately (it looks like something we'll need to deal with upstream), so I agree with the xfail. Just if we can also raise an issue to remind us to deal with it, that'd be awesome.

testsuite/MDAnalysisTests/coordinates/test_chainreader.py
tylerjereddy: see gh-3485 for the reminder; also, note that the `xfail` has been restored from previously being commented out--so this test has clearly been a pain point in the past (we could also be more granular with the `xfail`, since it only occurs for one of the parameters I think, but since the `xfail` was previously used more broadly I've simply restored it for Windows).

IAlibay (approved):
Thanks @tylerjereddy !

PR #3468, opened 2021-11-28
MAINT: update Cython imports

### LINKED ISSUE #3455: MAINT: updating Cython imports
Related to this for newer versions of Cython/NumPy: https://github.com/scipy/scipy/pull/14813

There are a few places we could probably add in `np.import_array()` just to be safe later on. Compare the output of `git grep -E -i 'cimport numpy'` and `git grep -E -i 'import_array'` to see the spots where we might add it in.

I don't think it is urgent though, and to be fair MDA already does the right thing in a few places based on those greps. Probably just a useful guard to add at some point.

tylerjereddy:
* use `np.import_array()` to prevent convoluted issues with
newer versions of Cython when using `cimport numpy ..` in
Cython sources

* in short, Cython and NumPy try to coordinate to do the right
thing with interwoven header inclusions, but this explicit call
prevents any confusion between the two

Fixes #3455

 - 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/analysis/encore/clustering/affinityprop.pyx
+++ b/package/MDAnalysis/analysis/encore/clustering/affinityprop.pyx
@@ -32,6 +32,8 @@ import numpy
 cimport numpy
 cimport cython
 
+numpy.import_array()
+
 cdef extern from "ap.h":
     int CAffinityPropagation(float*, int, float, int, int, bint, long*)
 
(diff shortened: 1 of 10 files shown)


richardjgowers (approved):
LGTM

Issue #3469, opened 2021-11-29
Increase SMARTS query maximum matches


orionarcher:
The rdkit smarts query has maximum matches set to 1000. I believe that is too low, many simulations have many thousands of molecules and it prevents SMARTs from being used to select a class of them. I'd like to bump the maximum up to 10000. That still prevents wasted computation but is better for large simulations. I have implemented this in PR #3470.

It would also be possible to add a max_matches keyword to the rdkit_kwargs.

PR #3470, opened 2021-11-29
increased the maximum number of matches from an rdkit smarts query

orionarcher:
Related to Issue #3469.

 - increased the maximum number of matches from a SMARTS selection query from 1000 to n_atoms.
 - separated kwargs for `GetSubstructMatch` and `convert_to`. 
 - updated documentation to clarify kwarg usage.

PR Checklist
------------
 - [x] Tests?
 - [x] Update main docs
 - [x] Docs?
 - [x] CHANGELOG updated?


lilyminium:
10k is also somewhat arbitrary. The OpenFF toolkit uses the largest unsigned int, could that work too? (`np.iinfo(np.uintc).max`)

IAlibay:
I'm not super into the idea of hardcoding stuff like this in. @lilyminium did OFF do any related benchmarks on the cost of this or was it just done for the sake of completeness without performance in mind?

If this is something that will regularly be used maybe we should just expose the argument to users?

lilyminium:
@IAlibay I doubt it, and tbh it's difficult to hit with small molecules anyway. We should probably just expose the argument. IIRC OpenFF has merged this too now, `max_matches=None` is what gets you the largest unsigned int. 

Edit: worth noting that 1000 is RDKit's own default limit, and I like setting it to the highest possible number as a default instead.

orionarcher:
Thank you for the feedback. I set the default value to the largest unsigned int, added an optional `max_matches` argument to the `rdkit_kwargs`, wrote a test to confirm behavior, and added some documentation and an example in the docstring of `atoms.select_atoms`.

> I'm not super into the idea of hardcoding stuff like this in.

As @lilyminium pointed out, it's going to be hardcoded to 1000 by default. I tend to think a higher hardcoding is preferable.

Please let me know if this needs anything else!

richardjgowers:
It's probably worth holding off and seeing what the person who put the limit in thinks (@cbouy )

The comparisons to OFF are a little misleading, a design difference between OFF and MDA is that OFF templates molecules and will only need to perform matches on the unique molecule types (e.g. there's one water stereotype that water molecules point to), whereas MDA just brute forces this.  This will mean that things like this selection scale differently, and OFF will likely be better at these things.

IAlibay:
> It's probably worth holding off and seeing what the person who put the limit in thinks (@cbouy )

Not that I don't want @cbouy's opinion on this (given that it's his original code), but in his defense isn't `GetSubstructMatches` an RDKit method? (i.e. it's just a built-in default)

IAlibay:
Seems like test failures are chemfiles related - seems like the feedstock got updated just an hour ago so this PR probably just ran at a bad time. Let's see if it re-runs a bit later in the day.

cbouy:
I didn't enforce any limit, I just mentioned in the docs that RDKit's default is 1000 unique matches.

Exposing `max_matches` is a good idea, although I would keep the RDKit syntax for it: `maxMatches`.
I don't particularly like camel-case in python but since users will have to type `rdkit_kwargs=...` I think it's better to stay consistent with RDKit, as we've done with other RDKit-based functions.

As for a good default value, why not use `maxMatches=mol.GetNumAtoms()` instead? This way you consistently get the maximum number of actual matches and not an arbitrary number.

orionarcher:
I agree with both @cbouy suggestions, I just implemented them in the PR.

lilyminium:
> As for a good default value, why not use maxMatches=mol.GetNumAtoms() instead? This way you consistently get the maximum number of actual matches and not an arbitrary number.

is uniquify True by default? Otherwise symmetry can complicate things a lot.

package/MDAnalysis/core/selection.py
```diff
@@ -655,8 +655,16 @@ def _apply(self, group):
         pattern = Chem.MolFromSmarts(self.pattern)
         if not pattern:
             raise ValueError(f"{self.pattern!r} is not a valid SMARTS query")
+        if "maxMatches" in self.rdkit_kwargs:
+            max_matches = self.rdkit_kwargs.pop("maxMatches")
+        else:
+            max_matches = group.n_atoms
```
IAlibay: Could we use the default from pop instead here?
```suggestion
        max_matches = self.rdkit_kwargs.pop("maxMatches", group.n_atoms)
```

orionarcher:
> is uniquify True by default? Otherwise symmetry can complicate things a lot.

Uniquify defaults to 1. I'm a RDkit novice so I'm not sure what that implies.

Another thought, currently `useChirality` is set as True in `getSubstructMatches`, maybe we should expose that in rdkit_kwargs too?

package/MDAnalysis/core/selection.py
```diff
@@ -655,8 +655,16 @@ def _apply(self, group):
         pattern = Chem.MolFromSmarts(self.pattern)
         if not pattern:
             raise ValueError(f"{self.pattern!r} is not a valid SMARTS query")
+        if "maxMatches" in self.rdkit_kwargs:
```
IAlibay: Not sure if it's somewhere else, but maybe we should document the rdkit_kwargs option if it isn't already?

IAlibay (changes requested):

lilyminium:
> Uniquify defaults to 1.

Sounds True-y.

> Another thought, currently useChirality is set as True in getSubstructMatches, maybe we should expose that in rdkit_kwargs too?

Yeah, I'm probably in favour of exposing all possible kwargs just because it leaves it up to the user.
Possibly everything in rdkit_kwargs not used in MDA could just get passed to RDKit? Or have another kwargs dict passed to RDKit?

IAlibay:
As discussed off-github with @lilyminium personally I think if we're going down the "optional kwargs for everything", then we might want to separate this out into "kwargs for the converter" and "kwargs for GetSubstructMatches"?

package/MDAnalysis/core/selection.py
orionarcher: I was unfamiliar with this syntax. Way cleaner! Thanks!

package/MDAnalysis/core/selection.py
orionarcher: I added documentation to `select_atoms`, is there anywhere else this should be documented?

package/MDAnalysis/core/selection.py
```diff
@@ -655,8 +655,14 @@ def _apply(self, group):
         pattern = Chem.MolFromSmarts(self.pattern)
         if not pattern:
             raise ValueError(f"{self.pattern!r} is not a valid SMARTS query")
+        use_chirality = self.rdkit_kwargs.pop("useChirality", True)
+        max_matches = self.rdkit_kwargs.pop("maxMatches", group.n_atoms)
```
IAlibay: my comment was more to have a separate kwarg from `rdkit_kwargs` so that folks don't get confused because what's being passed to what - thoughts @lilyminium ?

The useChirality being inverse of the default here is kind of a pain :/

lilyminium: Yeah, I guess something like `rdkit_kwargs` and `substruct_match_kwargs` could work better -- I defer to @cbouy's opinions on casing and spelling of the latter. it is nice to be consistent but usually means I just google "rdkit substruct match" instead of "mdanalysis substruct match".

> The useChirality being inverse of the default here is kind of a pain :/

@cbouy could you comment on this? I also usually set it to True when matching, I haven't really thought through the argument for just defaulting to RDKit's defaults when possible, vs. returning what we think are the best results by default.

orionarcher: Personally, I'd lean towards keeping a single `rdkit_kwargs` variable and popping out all possible keywords for `getSubstructMatches`, which is only four. 

As is, it's not clear whether `rdkit_kwargs` is being passed to `convert_to` or `getSubstructMatches`. If we are calling the variable `rdkit_kwargs`, which we are locked into at this point, I think it makes sense that it would supply all the `rdkit` functions used. Though I see how this is a little less clean, I think a bit of documentation could allay any confusion.

Just my two cents, I'm happy to implement either approach. Interested to hear @cbouy's take.

cbouy: I think I'd rather have a `smarts_kwargs` that gets passed to RDKit's `GetSubstructMatches` and a `rdkit_kwargs` (or another more descriptive name) that get's passed to `convert_to` instead of having a kwarg that combines both. Sounds more future-proof to me.

@lilyminium Users might not know that when they type `u.select_atom("smarts ...")` the underlying function being called is `GetSubstructMatch`, which is why I would go with `smarts_kwargs` instead of `substruct_match_kwargs`.

Googling "mdanalysis smarts" will take users to the docs page so @orioncohen should add a link to RDKit's [SubstructMatch docs](https://www.rdkit.org/docs/source/rdkit.Chem.rdchem.html#rdkit.Chem.rdchem.Mol.GetSubstructMatches) somewhere in the `select_atoms` docstring.

Also, we had started a similar discussion in #3324 and we ended up using `rdkit_kwargs` but I don't consider it to be set in stone. I'm not sure if converters other than RDKit are being used for making selections as of now, but I guess it will likely happen in the future so renaming it to `converter_kwargs` should be out of the question.
As @richardjgowers pointed out in the other PR, people using a SMARTS query might not know we implemented it with RDKit and that we need to convert the atomgroup to an RDKit molecule (even if it's documented) so `rdkit_kwargs` might not be descriptive enough. `rdkit_converter_kwargs` is a bit long, so `rdkit_converter` maybe ?

As for `useChirality`, I probably went for the non-default because we didn't expose the args for substructure matching at that time so I went for the option that would work for everyone. Also, not sure why it's not the default in RDKit actually, maybe a performance reason ? Anyway we could remove it now, and indicate in the docs that chirality is ignored by default when performing the substructure search.

orionarcher: Good point. It is definitely more future proof. I went with `smarts_kwargs` and `rdkit_kwargs` for now. I also added a link to the RDkit documentation as you suggested.

Currently I have `useChirality` set to `True`, but I'll change it based on the outcome of this discussion. If it is set to `True` and a user supplies a stereochemistry agnostic query, such as "CCC(C)O" instead of "CC[C@H](C)O" or "CC[C@@H](C)O", will that select both stereoisomers?

cbouy: Yes it selects both

orionarcher:
I've added a couple more tests and CI is now passing! 

Based on the discussion above, `rdkit_kwargs` and `smarts_kwargs` are now separate arguments to `select_atoms`. I updated the documentation and clarified the difference between the two.

The one remaining issue seems to be whether the `useChirality` kwarg for `getSubstructMatches` should be `True` or `False`. The RDkit default is `False` the current default in MDA is `True`. For now, I left the default as `True`, but I'm happy to implement either. I am interested to hear others opinions on this.

package/CHANGELOG
```diff
@@ -29,6 +30,8 @@ Fixes
 Enhancements
   * Add option for custom compiler flags for C/C++ on build and remove march
     option from setup.cfg (Issue #3428, PR #3429).
+  * increase smarts atom selection limit from 1000 -> n_atoms. Add
```
IAlibay: Changelog entries should be sorted newest first (see line 7)

package/MDAnalysis/core/groups.py
```diff
@@ -2977,7 +2977,15 @@ def select_atoms(self, sel, *othersel, periodic=True, rtol=1e-05,
             smarts *SMARTS-query*
                 select atoms using Daylight's SMARTS queries, e.g. ``smarts
                 [#7;R]`` to find nitrogen atoms in rings. Requires RDKit.
-                All matches (max 1000) are combined as a unique match
+                All matches are combined as a unique match. Uses two sets of
+                kwargs: rdkit_kwargs is passed to `RDKitConverter.convert()`
+                and smarts_kwargs is passed to RDKit's [GetSubstructMatches](
+                https://www.rdkit.org/docs/source/
+                rdkit.Chem.rdchem.html#rdkit.Chem.rdchem.Mol.GetSubstructMatches).
+                The useChirality kwarg is True by default.
```
IAlibay: I would expand on this and make it clear that this is a keyword argument of GetSubstructMatches.

Also personally I would avoid using "kwargs" directly in docs if avoidable. It assumes a certain amount of technical knowledge, which folks newer to Python may not have.

package/MDAnalysis/core/groups.py
```diff
@@ -2977,7 +2977,15 @@ def select_atoms(self, sel, *othersel, periodic=True, rtol=1e-05,
             smarts *SMARTS-query*
                 select atoms using Daylight's SMARTS queries, e.g. ``smarts
                 [#7;R]`` to find nitrogen atoms in rings. Requires RDKit.
-                All matches (max 1000) are combined as a unique match
+                All matches are combined as a unique match. Uses two sets of
+                kwargs: rdkit_kwargs is passed to `RDKitConverter.convert()`
+                and smarts_kwargs is passed to RDKit's [GetSubstructMatches](
```
IAlibay: I don't think the formatting is working here (https://mdanalysis--3470.org.readthedocs.build/en/3470/documentation_pages/core/groups.html?highlight=select_atoms#MDAnalysis.core.groups.AtomGroup.select_atoms). Please check and fix. Also might be worth re-organising / re-writing this to be a bit more readable.

package/MDAnalysis/core/groups.py
```diff
@@ -3147,7 +3155,8 @@ def select_atoms(self, sel, *othersel, periodic=True, rtol=1e-05,
                                                    periodic=periodic,
                                                    atol=atol, rtol=rtol,
                                                    sorted=sorted,
-                                                   rdkit_kwargs=rdkit_kwargs)
+                                                   rdkit_kwargs=rdkit_kwargs,
```
IAlibay, line 3215: Please add a versionchanged entry for this.

package/MDAnalysis/core/selection.py
```diff
@@ -656,7 +657,12 @@ def _apply(self, group):
         if not pattern:
             raise ValueError(f"{self.pattern!r} is not a valid SMARTS query")
         mol = group.convert_to("RDKIT", **self.rdkit_kwargs)
-        matches = mol.GetSubstructMatches(pattern, useChirality=True)
+        # override GetSubstructMatches default values
+        if "useChirality" not in self.smarts_kwargs:
```
IAlibay: Using setdefault here might be cleaner?

package/MDAnalysis/core/selection.py
```diff
@@ -656,7 +657,12 @@ def _apply(self, group):
...
+            self.smarts_kwargs["useChirality"] = True
+        if "maxMatches" not in self.smarts_kwargs:
+            self.smarts_kwargs["maxMatches"] = group.n_atoms
```
IAlibay: I think someone is going to have to convince me about setting goup.n_atoms here as the default instead of just using the standard default. Particularly in high symmetry,  low N molecules, do we run the risk of setting this too low?

It's best we be as clear and future proof as possible here so we minimise user impact.

IAlibay (changes requested):
Couple of comments

package/MDAnalysis/core/selection.py
orionarcher: I don't think the default limit of 1000 matches is appropriate for MD simulations, and I doubt that's what the RDkit writes had in mind when they set it. 

I often use the smarts query to select a class of molecules by querying for a unique functional group. It's nice to do that without thinking about system size. IMO the default of 1000 leads to inconsistent behavior, where in some cases I will have no issues but when the simulations is slightly larger I will hit an arbitrary cutoff.

A default that scales with system sizes seems more future proof to me than a low cutoff.

package/MDAnalysis/core/selection.py
orionarcher: Good suggestion, implementing

package/MDAnalysis/core/groups.py
orionarcher: not sure what you mean?

package/MDAnalysis/core/selection.py
IAlibay: I'm going to have to ask for some quantitative specifics here please.

I get that for large systems this is an issue but what about small ones - does a symmetric low N case cause problems? are there other edge cases we can think of? Every time we change a default we need to 2 minor versions (indeed we can't implement this change in 2.1.0 without warning users). If we change defaults I'd also like it so that users know exactly the bounds of what they are getting themselves into. (I know this sounds harsh but please don't take it as such, it's just that folks rely on this for science so we need to be as clear as possible in our decisions)

package/MDAnalysis/core/groups.py
IAlibay: In the docstring above - add a versionchanged entry

package/MDAnalysis/core/groups.py
orionarcher: Will do. I noticed that the [main selection docs](https://mdanalysis--3470.org.readthedocs.build/en/3470/documentation_pages/selections.html) are not updated. Where is that text written? I can't seem to track it down.

package/MDAnalysis/core/groups.py
orionarcher: working on it...

package/MDAnalysis/core/selection.py
orionarcher: Ahhh, I see your point regarding low N molecules now. For a more backwards compatible default, we could use `max(1000, group.n_atoms)`?

The two downsides I can think of for a higher default are a) wasted compute for high-symmetry systems and b) it could disrupt a users workflow if they are relying on the hard cap of 1000. For a) I believe a cap of `n_atoms` would prevent too much wasted compute. For b) hopefully adequate warning would prevent any trouble. Otherwise, I can't think of any cases where increasing the selection cap would have a detrimental. I can more quantitatively benchmark a) if needed.

orionarcher:
Happy new year everyone! 

I just pushed a commit that will change the default `maxMatches` value back to 1000, as @IAlibay suggested. This is the value that was used previously, so this PR shouldn't break any users workflows, even in edge cases. 

Please let me know if there are any other changes I should make.

IAlibay:
@MDAnalysis/coredevs can someone take over here? I'd like to get whatever we want to do here into 2.1.0, but have other things to address first.

The only thing I would ask is that if we _think_ we want to change the default, we will need to advertise it with a warning.

One other potential option here is to check if the number of matches is lower than the number of molecules and then throw a warning?

orbeckst:
@orioncohen I had a quick look at the PR and I want to understand what it is now about since the default was not changed. Is this now adding the ability to pass additional kwargs through? 

Could you please briefly summarize and resolve the conflicts. It looks to me that it shouldn't be too difficult to get the PR in, especially if there are no changes that change default behavior.

orionarcher:
@orbeckst correct, this PR now serves to add additional kwargs. Namely, it creates the `smarts_kwargs` argument in `select_atoms`. It allows users to tailor the behavior of the `smarts` keyword by passing arguments to RDKit's `GetSubstructMatch` function. Previously, there was no way to override the default values. This has been causing me many issues as I've tried to select specific atoms and regularly hit the default 1000 match limit. I think this will be broadly useful to users who want to use `smarts` on large systems.

I'll resolve the conflicts next week and push the changes.

orbeckst:
@orioncohen any chance to make some time in your schedule to finish up this PR? Would be good to put a bow on it.

orionarcher:
Thanks for the bump @orbeckst. I merged in the develop branch and this should be good to go.

package/MDAnalysis/core/selection.py
```diff
@@ -1408,7 +1411,7 @@ def expect(self, token):
                 "".format(self.tokens[0], token))
 
     def parse(self, selectstr, selgroups, periodic=None, atol=1e-08,
-              rtol=1e-05, sorted=True, rdkit_kwargs=None):
+              rtol=1e-05, sorted=True, rdkit_kwargs=None, smarts_kwargs=None):
```
IAlibay, line 1432: `smarts_kwargs` needs a docs parameters entry

package/MDAnalysis/core/selection.py
```diff
@@ -678,7 +679,9 @@ def _apply(self, group):
         if not pattern:
             raise ValueError(f"{self.pattern!r} is not a valid SMARTS query")
         mol = group.convert_to("RDKIT", **self.rdkit_kwargs)
-        matches = mol.GetSubstructMatches(pattern, useChirality=True)
+        self.smarts_kwargs.setdefault("useChirality", True)
+        self.smarts_kwargs.setdefault("maxMatches", 1000)
```
IAlibay: So re-reading the various comments I actually quite like your suggestion (https://github.com/MDAnalysis/mdanalysis/pull/3470#discussion_r766880225) to use `max(1000, n_atoms)` as the default, I would happy with that being implemented here and skipping the warning / deprecation with the excuse that it's technically a "bug" in large systems.

package/MDAnalysis/core/groups.py
```diff
@@ -3010,7 +3014,18 @@ def select_atoms(self, sel, *othersel, periodic=True, rtol=1e-05,
             smarts *SMARTS-query*
                 select atoms using Daylight's SMARTS queries, e.g. ``smarts
                 [#7;R]`` to find nitrogen atoms in rings. Requires RDKit.
-                All matches (max 1000) are combined as a unique match
+                All matches are combined as a single unique match. The `smarts`
+                selection accepts two sets of key word arguments from
+                `select_atoms()`: the rdkit_kwargs are passed internally to
```
IAlibay: probably will want double ticks around `rdkit_kwargs` and `smart_kwargs`

As you mentioned this is not being updated in the main docs, I _think_ you need to directly edit: doc/sphinx/source/documentation_pages/selections.rst

(not sure why there are two versions of this, probably historical things)

IAlibay (changes requested):
Couple of things, overall lgtm, however I would be ok with swapping to the default to max(1000, n_atoms) as you suggested

package/MDAnalysis/core/selection.py
orionarcher: I might actually suggest raising it to `max(1000, 10 * n_atoms)`. When using smarts selection I've found that `n_atoms` is often too low to capture all matches. While 10 * n_atoms is a bit arbitrary, I believe it would work better for many users. We could also set it to some extremely high number and give users responsibility for not using a selection like `CC` that would be very slow.

package/MDAnalysis/core/groups.py
orionarcher: Thanks, I was wondering where I needed to make those changes!

package/MDAnalysis/core/selection.py
IAlibay: Do we have a sense for how frequently this happens at n_atoms vs 10 * n_atoms, and how large a cost each order of magnitude adds? Is there a way this could easily be benchmarked?

If we think we'll still get a lot of problem cases unless we have to go to a number that is prohibitively expensive, we might want to inspect the return number of matches and raise a warning if maxMatch == len(matches)

orionarcher: I performed a quick benchmarking and the match number seems to have a negligible impact on the overall time calling `select_atoms`. The system I explored has ~8000 atoms

```
u.select_atoms(f'smarts C', smarts_kwargs={'maxMatches': 10})
```
and 
```
u.select_atoms(f'smarts C', smarts_kwargs={'maxMatches': 100000})
```
the former returned 10 atoms in 718 ms and the latter returned 2035 atoms in 728 ms, which is correct. It seems that the python code is much slower than the substructure matching written in C. Except for very large systems, I doubt the substructure matching will be an issue.

orionarcher: I'm not sure how frequently `n_atoms` is insufficient. I wasn't able to reproduce the issue on the system I have on hand, but I have encountered it in the past.

IAlibay: Ok yeah that makes sense it'll just be a loop that terminates once there are no further matches so you won't get any difference in performance unless `maxMatches < all possible matches`.

I _think_  we should do the following:

1. We set `maxMatches` to some arbitrary high value if we think that's beneficial in most cases (10 * n_atoms if you want for now).
2. We put up a warning in the docs detailing a) the issue, explaining what happens when you run out of allowable matches, b) how to fix it, c) the fact that this is currently under refinement and the maximum number of matches may grow / reduce in future versions.
3. Issue a warning if the number of matches == maxMatches that tells users that a maximum number of matches was returned and they may need to increase the number of matches if they want. (@jbarnoud @lilyminium I'm terrible at adding extra warnings - you both have stronger views on these things, are you for/against this?)

orionarcher: That all sounds good to me. A warning seems wise because a) `matches == maxMatches` almost certainly is not the intended behavior, and b) otherwise it would be a silent error that would be tricky to track down.

orionarcher: I just pushed an implementation of 1-3.

package/MDAnalysis/core/groups.py
```diff
@@ -3166,6 +3186,9 @@ def select_atoms(self, sel, *othersel, periodic=True, rtol=1e-05,
             Added the *smarts* selection. Added `atol` and `rtol` keywords
             to select float values. Added the ``sort`` keyword. Added
             `rdkit_kwargs` to pass parameters to the RDKitConverter.
+        .. versionchanged:: 2.0.1
```
IAlibay:
```suggestion
        .. versionchanged:: 2.2.0
```

orionarcher:
I implemented @IAlibay's suggested changes and fixed the conflicts with CHANGELOG. I'd love to get this merged if there are no more changes!

IAlibay:
Thanks @orioncohen, I'll try to review it in a few

package/MDAnalysis/core/selection.py
```diff
@@ -678,7 +679,17 @@ def _apply(self, group):
         if not pattern:
             raise ValueError(f"{self.pattern!r} is not a valid SMARTS query")
         mol = group.convert_to("RDKIT", **self.rdkit_kwargs)
-        matches = mol.GetSubstructMatches(pattern, useChirality=True)
+        self.smarts_kwargs.setdefault("useChirality", True)
+        self.smarts_kwargs.setdefault("maxMatches", len(group) * 10)
```
IAlibay: I thought we had agreed on `max(1000, n_atoms*10)` so as not to change the default in cases where you have few atoms?

IAlibay (changes requested):
I've not read all the docs fully yet, but this will need addressing first.

richardjgowers (approved):
LGTM barring Irfan's comment about the default number of matches

IAlibay (approved):
Ok that _should_ be fine now.

@richardjgowers do you want to have a quick glance at the changes I made? Then I think it's good to go.

IAlibay:
Thanks @orioncohen ! Sorry for taking over at the end, I just wanted to get this through for the 2.2.0 release.

orionarcher:
Thanks for wrapping it up @IAlibay, glad this is merged! I've been moving the past couple days so I was AFK.

PR #3475, opened 2021-12-06
Pbc dist bench

### LINKED ISSUE #3206: Distance benchmarks should cover all box types
The benchmarks for `lib.distances` currently use a `None` box indicating no periodic boundaries.  The benchmarks should also include both orthogonal and triclinic boxes.

Miro-Astore:
Fixes #3206 and #2794

Rewrote the whole benchmark since last pull request so  made a new one. 

   Added, for relevant functions, benchmarks for testing orthogonal and triclinic
   PBC settings
   Rewrote all distance benchmarks to use a universe of randomly generated 
   atoms. This lets the benchmarks scale to arbitrary sizes.

   It also makes realistic atom densities which are important for the contact 
   matrix distance functions. If the universe is too dense then a matrix with
   too many connections is made and the benchmark will time out. As is the 
   10 000 atom option requires a modification to timeout for asv to run the 
   benchmark.

I had to split the DistancesBench class into 2 classes to avoid
 uncessary reruns of the benchmars which do not take pbc type as an argument. Hence I 
 created BetweenBench.

The 2 classes and the distribution of testing functions is

BetweenBench

    time_between

DistancesBench
    time_distance_array
    time_distance_array_pre_allocated
    time_self_distance_array
    time_self_distance_array_pre_allocated
    time_dist
    time_dist_offsets
    time_contact_matrix
    time_contact_matrix_sparse

Splitting the classes like this was necessary because asv does not
allow us to easily skip certain combinations of parameters at the
function level, only in the setup()
function of the class.


benchmarks/benchmarks/analysis/distances.py
```diff
@@ -1,6 +1,12 @@
 import MDAnalysis
 import numpy as np
 
+
+try:
+    from MDAnalysis.lib import mdamath
+except:
```
tylerjereddy: Is there any realistic type of exception you're expecting here apart from `ImportError` for some old versions of MDAnalysis? In other words, why catch all exceptions here? It looks like, to be fair, I wrote some of the generic `except` into this file 4 years ago, but not sure that makes it a good idea.

benchmarks/benchmarks/analysis/distances.py
```diff
@@ -11,120 +17,236 @@
...
+            self.box_dims = self.u.dimensions
+
+        elif pbc_type == 'triclinic':
+            # making a triclinic box with the same volume as the cube
+            deg_alpha = np.float32(60)
+            deg_beta = np.float32(60)
+            deg_gamma = np.float32(60)
+            alpha = np.deg2rad(deg_alpha)
+            beta = np.deg2rad(deg_beta)
+            gamma = np.deg2rad(deg_gamma)
```
tylerjereddy: I can't really think of a reason for all this repetition with `np.float32(60)` 3 times, etc. For example, `np.deg2rad()` takes a `dtype` argument anyway, and if there is only one value couldn't we just use a single `angle = np.deg2rad(60, dtype=np.float32)` variable or similar?

benchmarks/benchmarks/analysis/distances.py
```diff
@@ -11,120 +17,236 @@
...
-        distances.between(group=self.ag3,
-                          A=self.ag1,
-                          B=self.ag2,
-                          distance=15.0)
+        distances.contact_matrix(coord=self.ag1.positions,
+                                 cutoff=15.0,
+                                 returntype='numpy',
+                                 box=self.box_dims)
+
+    def time_contact_matrix_sparse(self, num_atoms, box_dims):
```
tylerjereddy: You have benchmark functions in the same class with different parameters, but normally the same parameters are used for each. In fact, it is extra confusing when I look at the output (see below) because `box_dims` gets replaced by the true parameter name (`pbc_type`) it seems, which would not be very useful for an analysis of a performance regression (a user/developer would have to realise that an override/hack has happened, if that's what is happening here...).

Running this locally:

```
[ 12.50%] ··· analysis.distances.DistancesBench.time_contact_matrix_sparse                                                                                                                                                                                                                                                ok
[ 12.50%] ··· =========== ========== ============ ===========
              --                        pbc_type             
              ----------- -----------------------------------
               num_atoms     None     orthogonal   triclinic 
              =========== ========== ============ ===========
                   10      466±0μs     473±0μs      425±0μs  
                  100      33.3±0ms    34.5±0ms     35.1±0ms 
                  1000     1.51±0s     2.81±0s      2.47±0s  
                 10000     25.9±0s     35.4±0s      27.6±0s  
              =========== ========== ============ ===========
```

benchmarks/benchmarks/analysis/distances.py
Miro-Astore: Yeah I was just following you haha. Happy to take it out.

benchmarks/benchmarks/analysis/distances.py
Miro-Astore: Yeah this made more sense when I was debugging my code with different angle values. I'll change it.

benchmarks/benchmarks/analysis/distances.py
Miro-Astore: See commit [d8d42c8](https://github.com/MDAnalysis/mdanalysis/pull/3475/commits/d8d42c8a3add4d984a15bcfde522efbf453ebd2c)

Took me a bit to figure out what you meant but I completely agree that this is a clearer way of writing it. Please let me know of any other suggestions you have.

Miro-Astore:
Hi, just wondering if there's anything that still needs fixing for this pull request to be merged.

orbeckst (changes requested):
As I understand your changes, you changed the underlying data for DistanceBench — correct me if I am wrong. My concern with changing a benchmark case but keeping the name the same will break the history so looking at the time series of performances may now show an artificial increase or decrease in performance that has nothing to do with the code itself.

My opinion is that we should
* keep old benchmarks _as they are_ or retire them
* add new benchmarks with new names

@tylerjereddy @hmacdope @IAlibay @richardjgowers if you have different opinions I am all ears. This looks like something we should be deciding now.

orbeckst:
Sorry @Miro-Astore for the delay in reviewing. I think we have to first decide how we want to progress in cases like this one where the benchmarks themselves are changed.

Miro-Astore:
No problem. I will be happy to adapt the code to whatever framework the development team decide on.

orbeckst:
After a quick discussion on the developer list there are at least two developers (@richardjgowers and I) who think that we need to do what I said in my comment:
* keep old benchmarks as they are
* create new ones for anything new

Can you please change your PR in this way, @Miro-Astore ? Thank you very much! (I do look forward to seeing the results!)

orbeckst:
Please ping me when you need me to review again, I try to do it reasonably quickly.

orbeckst:
@hmacdope if you have time to look at any updates then that would also be very helpful.

hmacdope:
> @hmacdope if you have time to look at any updates then that would also be very helpful.

I'll get onto it asap

hmacdope:
@orbeckst @Miro-Astore agree that we probably need to keep original benchmark names as otherwise performance history is not really linear.

Miro-Astore:
No problem. I'll get to it after the BPS meeting. If any of you are here I'll be sure to say hi!

benchmarks/benchmarks/analysis/distances.py
```diff
@@ -11,120 +11,236 @@
...
+        self.u.atoms.positions = cube_side_length * random_samples
+
+        # splitting atom groups into 3 groups with the same number of atoms
+        self.ag1 = self.u.atoms[:num_atoms]
+        self.ag2 = self.u.atoms[num_atoms:-num_atoms]
+        self.ag3 = self.u.atoms[-num_atoms:]
         self.allocated_array_2D = np.empty((num_atoms, num_atoms),
-                                            dtype=np.float64)
+                                           dtype=np.float64)
         self.array_shape_1D = int(num_atoms * (num_atoms - 1) / 2.)
```
hmacdope, line 70: nitpick but would prefer `num_atoms * (num_atoms - 1) // 2`

benchmarks/benchmarks/analysis/distances.py
```diff
@@ -11,120 +11,236 @@
...
+
+    def setup(self, num_atoms, pbc_type):
+
+        # Some functions are performance dependent on a realistic
+        # atom density. So we make a new universe with random atom
+        # positions, density derived from an example universe.
+
+        # Make fake universe with 2*num_atoms because some of our benchmarks
+        # require 2 atom groups.
+        universe_num_atoms = num_atoms*2
```
hmacdope: I understand why this was set up like this but is it possible to avoid the code repetition of a large block here?

Personally I would make `DistancesBench` and `BetweenBench` subclass  something (perhaps `CoordinateBench`) that sets this all up as its `__init__()`. 

Just a suggestion. You can then do the `DistancesBench`  PBC specific stuff in setup. :)

benchmarks/benchmarks/analysis/distances.py
```diff
@@ -11,120 +11,236 @@
...
+
+        np.random.seed(17809)
+        self.coords_1 = np.random.random_sample((num_atoms,
+                                                 3)).astype(np.float32)
+        np.random.seed(9008716)
+        self.coords_2 = np.random.random_sample((num_atoms,
+                                                 3)).astype(np.float32)
+        self.allocated_array_2D = np.empty((num_atoms, num_atoms),
+                                           dtype=np.float64)
+        self.array_shape_1D = int(num_atoms * (num_atoms - 1) / 2.)
```
hmacdope, line 285: same as above `num_atoms * (num_atoms - 1) // 2`

hmacdope (changes requested):
Looking good in general , just a few things. 

Also worth considering is having a benchmark with a significant portion of the coordinates outside the primary cell.

 I could be wrong but for triclinic boxes I think this is slightly more expensive due to need to call `_triclinic_pbc ` which has heaps of branching logic. For orthorhombic boxes I am fairly sure all the code paths are the same.

benchmarks/benchmarks/analysis/distances.py
hmacdope: Actually ignore me

benchmarks/benchmarks/analysis/distances.py
hmacdope: ignore this.

Miro-Astore:
I think my latest commits need a bit more organisation sorry. I'll ping you when they're ready for reveiw. 

> Also worth considering is having a benchmark with a significant portion of the coordinates outside the primary cell.

Just a small note that I don't think this is a good idea because the original reason behind the bug in https://github.com/MDAnalysis/mdanalysis/issues/2794 was because contact matrix calculations  made massive matrices which cause MemoryErrors when lots of atoms are very close.

If we make atoms outside the box it could cause the same error because then they are wrapped into the smaller box they would closely overlap. I'll try and test it out at some point.

jbarnoud:
> If we make atoms outside the box it could cause the same error because then they are wrapped into the smaller box they would closely overlap. I'll try and test it out at some point.

In real systems, atoms outside the box do not wrap closely to atoms inside it. This is because the interactions are computed following theminimum image convention so atoms that are close to each other are "seen" as such regardless of either that are in the box or not.

To have a test system with atoms outside the box, take a system with all the atoms in the box, then translate some of these atoms by a full box in one dimension or another. When wrapped back inside the box, they go back where they started which should avoid the issue you describe.

hmacdope:
@Miro-Astore are you likely to push on this in the near future? Only asking because I am looking into adding some benchmarks as part of our CZI work, but don't want to steal your thunder. :)

Miro-Astore:
Hi Hugo, thanks for checking in!

Sorry for letting this go a bit dark.

I am planning to finish off this PR. I'm gonna be submitting my thesis in a
few weeks and was planning to push the final changes for this once I've
submitted.

Best, Miro

hmacdope:
Okay that is awesome news! Feel free to ping me for a review. :)

hmacdope:
Bump @Miro-Astore, I am happy to take over here otherwise :)

Miro-Astore:
I'm actually just finishing a project this week. So i'll (finally) get back to this on the weekend. Thanks Hugo!

Miro-Astore:
Hi @hmacdope  I think i've finished the required changes and I'm ready for another review. It's obviously been awhile but from reading @orbeckst  's comments on this I've just gone and split up everything into a pbc class. Emperically I'm noticing that there is indeed a difference for some functions between the no pbc option and the orthorhombic versions. 

Thanks for being patient with me everybody. Let me know if you'd like more changes.

hmacdope:
@Miro-Astore the diff seems enormous for some reason?

Miro-Astore:
Sorry about that, I think I've done something silly and merged the current development branch because my current one was so old. It seems to have created a problem. Can you just look at the ` benchmarks/benchmarks/analysis/distances.py  ` script or would you like me to try and fix the weird diffs ?

hmacdope:
You will need to sort out diffs before merge is possible. @MDAnalysis/coredevs I forget what the fix is here.

orbeckst:
Yes, please fix the patch. You also have conflicts. I'd go back to [087cb46](https://github.com/MDAnalysis/mdanalysis/pull/3475/commits/087cb464532300a20b9d844dfcdf8c57f0dbc6ac). Then I'd try merging from GitHub. Once the merge is done cleanly, you can put the changes from https://github.com/MDAnalysis/mdanalysis/pull/3475/commits/1eb6bdde098a05bb07ea5c42cf2272202fb701ca back.

(Perhaps you can try a rebase ... but be careful with any history rewriting. If in doubt, ask first.)

Miro-Astore:
Hi, sorry for the messy commit history. I think this is now ready?

.gitignore
```diff
@@ -1,54 +0,0 @@
-# Ignore python bytecoded files
```
orbeckst: please don't delete our .gitignore

benchmarks/benchmarks/analysis/distances.py
```diff
@@ -14,44 +14,70 @@
...
+        desired_box_volume = universe_num_atoms*density
+        # making a cube with orthogonal basis vectors
+        cube_side_length = np.cbrt(desired_box_volume)
+        sides = [float(cube_side_length)]*3
+        ortho_angles = [float(90)]*3
+        self.u.dimensions = np.array(sides + ortho_angles)
+
+        # filling the cube with random points.
         np.random.seed(17809)
-        self.coords_1 = np.random.random_sample((num_atoms, 3)).astype(
```
orbeckst, line 27: As much as I like the idea of getting a sensible density in principle, it's more important to keep benchmarks comparable. Don't change the data.

Instead, make a copy of DistancesBench (maybe DistancesBench2) where you generate the data in the new way.

Then compare the two. Does it make a difference?

benchmarks/benchmarks/analysis/distances.py
```diff
@@ -116,51 +162,187 @@ def time_contact_matrix_sparse(self, num_atoms):
         memory implementation of contact_matrix intended
         for larger systems.
         """
-        distances.contact_matrix(
-            coord=self.coords_1, cutoff=15.0, returntype="sparse", box=None
-        )
+        distances.contact_matrix(coord=self.ag1.positions,
+                                 cutoff=15.0,
+                                 returntype='sparse',
+                                 box=None)
```
orbeckst, line 172: two empty lines

benchmarks/benchmarks/analysis/distances.py
```diff
@@ -116,51 +162,187 @@ def time_contact_matrix_sparse(self, num_atoms):
...
+    timeout = 180
+    params = ([10, 100, 1000, 10000], ['orthogonal', 'triclinic'])
+    param_names = ['num_atoms', 'pbc_type']
 
-    def time_dist(self, num_atoms):
+
+    def setup(self, num_atoms, pbc_type):
+
+        # This setup function is virtually identical to the one
+        # in the DistancesBench, we just have to do some extra
```
orbeckst: DistancesBench2

benchmarks/benchmarks/analysis/distances.py
```diff
@@ -116,51 +162,187 @@ def time_contact_matrix_sparse(self, num_atoms):
...
-        distances.contact_matrix(
-            coord=self.coords_1, cutoff=15.0, returntype="sparse", box=None
-        )
+        distances.contact_matrix(coord=self.ag1.positions,
+                                 cutoff=15.0,
+                                 returntype='sparse',
+                                 box=None)
+class DistancesBenchPBC(object):
+    """Benchmarks for MDAnalysis.analysis.distances
+    functions.
```
orbeckst: with PBC

benchmarks/benchmarks/analysis/distances.py
```diff
@@ -116,51 +162,187 @@ def time_contact_matrix_sparse(self, num_atoms):
...
+        # positions, density derived from an example universe.
+
+        # Make fake universe with 3*num_atoms because some of our benchmarks
+        # require 3 atom groups.
+        universe_num_atoms = num_atoms*3
+        self.u = MDAnalysis.Universe.empty(universe_num_atoms, trajectory=True)
+        # calculating reasonable density for an atomistic system.
+        ex_universe = MDAnalysis.Universe(GRO)
+        density = ex_universe.coord.volume/ex_universe.atoms.n_atoms
+        desired_box_volume = universe_num_atoms*density
```
orbeckst, line 223: space around operators

benchmarks/benchmarks/analysis/distances.py
```diff
@@ -116,51 +162,187 @@ def time_contact_matrix_sparse(self, num_atoms):
...
+        # making a cube with orthogonal basis vectors
+        cube_side_length = np.cbrt(desired_box_volume)
+        sides = [float(cube_side_length)]*3
+        ortho_angles = [float(90)]*3
```
orbeckst, line 227: space around operators

benchmarks/benchmarks/analysis/distances.py
```diff
@@ -116,51 +162,187 @@ def time_contact_matrix_sparse(self, num_atoms):
...
-    def time_between(self, num_atoms):
-        """Benchmark determination of subgroup
-        of atomgroup that is within a specific
-        distance of two other atomgroups.
-        """
-        distances.between(
-            group=self.ag3, A=self.ag1, B=self.ag2, distance=15.0
-        )
 
-    def time_calc_bonds(self, num_atoms):
```
orbeckst, line 146: add it back

benchmarks/benchmarks/analysis/distances.py
```diff
@@ -116,51 +162,187 @@ def time_contact_matrix_sparse(self, num_atoms):
...
+        a single numpy array using the default arguments
+        to contact_matrix.
         """
-        mda.lib.distances.calc_bonds(self.coords_1, self.coords_2)
+        distances.contact_matrix(coord=self.ag1.positions,
+                                 cutoff=15.0,
+                                 returntype='numpy',
+                                 box=self.box_dims)
 
-    def time_calc_angles(self, num_atoms):
```
orbeckst, line 152: add it back

benchmarks/benchmarks/analysis/distances.py
```diff
@@ -116,51 +162,187 @@ def time_contact_matrix_sparse(self, num_atoms):
...
         """
-        mda.lib.distances.calc_angles(
-            self.coords_1, self.coords_2, self.coords_3
-        )
+        distances.contact_matrix(coord=self.ag1.positions,
+                                 cutoff=15.0,
+                                 returntype='sparse',
+                                 box=self.box_dims)
 
-    def time_calc_dihedrals(self, num_atoms):
```
orbeckst, line 160: add it back

orbeckst (changes requested):
Very brief/terse review
- The original DistancesBench needs to have the old data generation. If there's a need for a version with the new realistic density then make a copy.
- You removed the time_calc_* methods. They should be added back, _at least_ to DistancesBench.
- Add .gitignore back.

Miro-Astore:
Hi sorry this is taking so long, and for all the trouble. I think I'm ready for another round of reviews. 

(accidentally made a giant commit again but i've reset it. Ping me if there's an issue. ) @orbeckst

Miro-Astore:
> Hi sorry this is taking so long, and for all the trouble. I think I'm ready for another round of reviews.
> 
> (accidentally made a giant commit again but i've reset it. Ping me if there's an issue. ) @orbeckst

bump

.gitignore
orbeckst, line 1: Please don't delete lines from the file. They _might_ be there for a reason.

If some clean-up is needed, do a separate mini PR.

benchmarks/benchmarks/analysis/distances.py
```diff
@@ -101,66 +131,245 @@ def time_self_distance_array_pre_allocated(self, num_atoms):
...
+            coord=self.ag1.positions, cutoff=15.0, returntype="numpy", box=None
         )
 
+    def time_contact_matrix_sparse(self, num_atoms):
+        """Benchmark calculation of contacts within
+        a single numpy array using the slower reduced
+        memory implementation of contact_matrix intended
+        for larger systems.
+        """
+        distances.contact_matrix(coord=self.ag1.positions,
```
orbeckst, line 169: Can you briefly explain how your changes address #2794 ? Does the `numatom == 10000` case not time out anymore? 

Perhaps we can just [skip this case](https://asv.readthedocs.io/en/stable/writing_benchmarks.html#skipping-benchmarks)
```python
from asv_runner.benchmarks.mark import skip_for_params

@skip_for_params([(10000, 'numatoms')])
def time_contact_matrix_sparse(self, num_atoms):
   ...
```

benchmarks/benchmarks/analysis/distances.py
```diff
@@ -14,44 +14,74 @@
...
+        universe_num_atoms = num_atoms*3
+        self.u = MDAnalysis.Universe.empty(universe_num_atoms, trajectory=True)
+        # calculating reasonable density for an atomistic system.
+        ex_universe = MDAnalysis.Universe(GRO)
+        density = ex_universe.coord.volume/ex_universe.atoms.n_atoms
+        desired_box_volume = universe_num_atoms*density
+        # making a cube with orthogonal basis vectors
+        cube_side_length = np.cbrt(desired_box_volume)
+        sides = [float(cube_side_length)]*3
+        ortho_angles = [float(90)]*3
```
orbeckst, line 40: PEP 8 — space around operators

benchmarks/benchmarks/analysis/distances.py
```diff
@@ -14,44 +14,74 @@
...
 
     def setup(self, num_atoms):
+
+        # Some functions are performance dependent on a realistic
+        # atom density. So we make a new universe with random atom
+        # positions, density derived from an example universe.
+
+        # Make fake universe with 3*num_atoms because some of our benchmarks
+        # require 3 atom groups.
+        universe_num_atoms = num_atoms*3
```
orbeckst, line 31: PEP 8 — space around operators

benchmarks/benchmarks/analysis/distances.py
```diff
@@ -14,44 +14,74 @@
...
+        self.u = MDAnalysis.Universe.empty(universe_num_atoms, trajectory=True)
+        # calculating reasonable density for an atomistic system.
+        ex_universe = MDAnalysis.Universe(GRO)
+        density = ex_universe.coord.volume/ex_universe.atoms.n_atoms
+        desired_box_volume = universe_num_atoms*density
```
orbeckst, line 36: PEP 8 — space around operators

benchmarks/benchmarks/analysis/distances.py
```diff
@@ -101,66 +131,245 @@ def time_self_distance_array_pre_allocated(self, num_atoms):
...
+        # in the DistancesBench class above, we just have to do some
+        # extra work to set up different periodic boundaries.
+
+        # Some functions are performance dependent on a realistic
+        # atom density. So we make a new universe with random atom
+        # positions, density derived from an example universe.
+
+        # Make fake universe with 3*num_atoms because some of our benchmarks
+        # require 3 atom groups.
+        universe_num_atoms = num_atoms*3
```
orbeckst, line 218: PEP 8 — space around operators

benchmarks/benchmarks/analysis/distances.py
```diff
@@ -101,66 +131,245 @@ def time_self_distance_array_pre_allocated(self, num_atoms):
...
+        self.u = MDAnalysis.Universe.empty(universe_num_atoms, trajectory=True)
+        # calculating reasonable density for an atomistic system.
+        ex_universe = MDAnalysis.Universe(GRO)
+        density = ex_universe.coord.volume/ex_universe.atoms.n_atoms
+        desired_box_volume = universe_num_atoms*density
```
orbeckst, line 223: PEP 8 — space around operators

benchmarks/benchmarks/analysis/distances.py
orbeckst: PEP 8 — space around operators

orbeckst (changes requested):
Looks very good; I only have one question (see comments) and please do some PEP8 formatting. Thanks!

orbeckst:
@hmacdope can you also have a look if your previous concerns have been addressed?

benchmarks/benchmarks/analysis/distances.py
hmacdope: This is fixed now @Miro-Astore thanks.

Miro-Astore:
@hmacdope 
if i fix up the syntax errors is this ready to roll?

orbeckst:
@Miro-Astore sorry ... hadn't seen you question.

If you address my comments then we should be good to go. Please ping me when you're done! Thank you!!

PR #3480, opened 2021-12-08
[hold pub] added function to calculate angle between base planes

### LINKED ISSUE #3478: angle between the planes of two nucleobases

I wanted to analyze the nucleic acids for the angle between the base planes and could not find anything simple and quick.

It would be great to have a function that takes in the resids for the two bases and returns the angle between the planes of the bases.

Adding it to the analysis module for nucleic acids sounds like best option.

aakognole:
Fixes #3478 

 - added function to calculate angle between base planes

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/analysis/nuclinfo.py
```diff
@@ -763,3 +763,54 @@ def pseudo_dihe_baseflip(universe, bp1, bp2, i,
...
+
+    Returns
+    -------
+    float
+          angle in degrees
+
+
+    .. versionadded:: 1.0.1
+    """
+    baseatoms = { 'CYT': ['N1', 'N3', 'C5'],
```
IAlibay: I'm not sure I agree with the premise of having hardcoded atom names and residues for things like this. It is always best for things like this to leave the selection to users and instead ask that folks pass in either directly coordinates or atomgroups.

package/MDAnalysis/analysis/nuclinfo.py
```diff
@@ -763,3 +763,54 @@ def pseudo_dihe_baseflip(universe, bp1, bp2, i,
     pseudo = mdamath.dihedral(x[0] - x[1], x[1] - x[2], x[2] - x[3])
     pseudo = np.rad2deg(pseudo) % 360
     return pseudo
+
```
IAlibay, line 774: Please do attend to the various PEP8 issues highlighted by pep8speaks https://github.com/MDAnalysis/mdanalysis/pull/3480#issuecomment-989321451

package/MDAnalysis/analysis/nuclinfo.py
```diff
@@ -763,3 +763,54 @@ def pseudo_dihe_baseflip(universe, bp1, bp2, i,
...
+    segid2 : str (optional)
+        segid of b2
+
+    Returns
+    -------
+    float
+          angle in degrees
+
+
+    .. versionadded:: 1.0.1
```
IAlibay:
```suggestion
    .. versionadded:: 2.1.0
```

IAlibay (changes requested):
Hi @aakognole,

Thanks for opening this PR.

This is not the type of system I generally deal with and it's hard for me to conceptualise just how common an operation this is, and therefore a) if it should be included in the core library, and b) where it should be included.

If this is a common analysis, then this may be better implemented as an [AnalysisBase](https://docs.mdanalysis.org/2.0.0/documentation_pages/analysis/base.html?highlight=analysisbase) class rather than a function. If it's more of a method that can be generalised across various systems then maybe it should go into mdamath instead.

I am hoping one of the other @MDAnalysis/coredevs (maybe @orbeckst) has a better idea here?

package/MDAnalysis/analysis/nuclinfo.py
orbeckst: Admittedly, this follows the style of the (very old) nuclinfo module. It would be consistent with the legacy code that's already there.

orbeckst:
@aakognole do you have one or two paper citations where this analysis was used and/or described?

aakognole:
Hi @orbeckst 

Yes, the angle between the base planes can be important criteria in determining weather the bases are considered stacked or not. I am working on something where I need to consider this as a colvar.

Here are a few references
https://doi.org/10.1016/0263-7855(95)00086-0
https://doi.org/10.1021/jp209986y

testsuite/MDAnalysisTests/analysis/test_nuclinfo.py
```diff
@@ -174,3 +174,13 @@ def test_hydroxyl(u, seg, i, expected_value):
...
+
+
+@pytest.mark.parametrize('b1, b2, seg1, seg2, expected_value', (
+        (2, 3, 'RNAA', 'RNAA', 15.490582),
+        (10, 11, 'RNAA', 'RNAA', 43.71941),
+        (7, 17, 'RNAA', 'RNAA', 168.37481),
+))
+def test_angle_between_base_planes(u, b1, b2, seg1, seg2, expected_value):
+    val = nuclinfo.angle_between_base_planes(u, b1, b2, seg1, seg2)
+    assert_almost_equal(val, expected_value, decimal=3)
```
tylerjereddy: maybe `assert_allclose()` for new code: https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_almost_equal.html

aakognole: Okay. Would you like me to move to `assert_allclose()` for all the tests in nuclinfo? 

Although I think it is reasonable to stick with `assert_almost_equal()` since we are not looking for precision beyond 3 decimals.

tylerjereddy: No, just here for now; I can't think of a reason to use the discouraged approach in new code.

aakognole: Updated. Thanks.

orbeckst:
Is the *Gabb et al, Journal of MolecularGraphics14:6-1 I. 1996* paper the original paper where they described the base plane angle the first time (tau_ij in the paper) ?

If so, please include the reference in the documentation, the [userguide has an entry on our doc style](https://userguide.mdanalysis.org/stable/contributing_code.html#working-with-the-code-documentation), in particular the [numpy doc string standard](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard). Add the citation itself to [doc/sphinx/source/documentation_pages/references.rst](https://github.com/MDAnalysis/mdanalysis/blob/develop/package/doc/sphinx/source/documentation_pages/references.rst) instead of a references section in the function docs.

As an example look at [AtomGroup.asphericity](https://docs.mdanalysis.org/stable/documentation_pages/core/groups.html#MDAnalysis.core.groups.AtomGroup.asphericity).

aakognole:
Hi @orbeckst 

> Is the _Gabb et al, Journal of MolecularGraphics14:6-1 I. 1996_ paper the original paper where they described the base plane angle the first time (tau_ij in the paper) ?
> 
Yes, this is the reference I could trace down for base plane angle analysis used for stacking. However the angle_between_base_planes in the paper is _theta_ij_ and not _tau_ij_. The _tau_ij_ refers to a different angle not used frequently. I would also refer to Jafilan et. al. paper since it has developed on this analysis and gives simpler stacking coordinate based on the angle_between_base_planes.

> If so, please include the reference in the documentation, the [userguide has an entry on our doc style](https://userguide.mdanalysis.org/stable/contributing_code.html#working-with-the-code-documentation), in particular the [numpy doc string standard](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard). Add the citation itself to [doc/sphinx/source/documentation_pages/references.rst](https://github.com/MDAnalysis/mdanalysis/blob/develop/package/doc/sphinx/source/documentation_pages/references.rst) instead of a references section in the function docs.
> 
> As an example look at [AtomGroup.asphericity](https://docs.mdanalysis.org/stable/documentation_pages/core/groups.html#MDAnalysis.core.groups.AtomGroup.asphericity).

Thanks for the guidelines. I will add the references accordingly.

orbeckst:
Very good, ping me when you need me to look at it.

aakognole:
@orbeckst please take a look at the changes. Thanks.

orbeckst (changes requested):
Please integrate the new functions in the docs. You need to mention them explicitly at the top of the file. Possibly add a new section under which the new functions fits if none of the other sections work.

Check that the docs are visible in the RTD build https://mdanalysis--3480.org.readthedocs.build/en/3480/documentation_pages/analysis/nuclinfo.html

orbeckst (approved):
Looks good. Thank you @aakognole – sorry it took such a long time.

orbeckst:
@IAlibay do you want to have a final look and/or wait for Azure to run again (in which case, please kick it once the brownout is over). Cheers!

package/MDAnalysis/analysis/nuclinfo.py
```diff
@@ -763,3 +769,62 @@ def pseudo_dihe_baseflip(universe, bp1, bp2, i,
     pseudo = mdamath.dihedral(x[0] - x[1], x[1] - x[2], x[2] - x[3])
     pseudo = np.rad2deg(pseudo) % 360
     return pseudo
+
+
+def angle_between_base_planes(universe, b1, b2, seg1="SYSTEM", seg2="SYSTEM"):
```
IAlibay: I realise this is in keeping with the other functions in this module, but segments is kind of a weird one because not everything really fully supports it - given the recent discussions on "chainID" vs "segID" would it make sense to switch from being resid/system for each base to just "a selection string for that base"? Would that be easier?

package/MDAnalysis/analysis/nuclinfo.py
```diff
@@ -763,3 +769,62 @@ def pseudo_dihe_baseflip(universe, bp1, bp2, i,
...
+             'URA': ['N1', 'N3', 'C5'],
+             'ADE': ['N9', 'N1', 'N7'],
+             'GUA': ['N9', 'N1', 'N7']}
+    # select residues
+    bf1 = universe.select_atoms(
+        "(segid {0!s} and resid {1!s})".format(seg1, b1))
+    bf2 = universe.select_atoms(
+        "(segid {0!s} and resid {1!s})".format(seg2, b2))
+    # get residue names
+    resn1 = bf1.atoms.resnames[0]
```
IAlibay, line 850: even if we disregard my comment above, I believe resids are not guaranteedto be unique (I always forget - am I right here @lilyminium ?) Would it make sense to do a check for > 1 residues in this atomgroup and throw a warning that we're taking the first residue if that's the case?

package/MDAnalysis/analysis/nuclinfo.py
```diff
@@ -763,3 +769,62 @@ def pseudo_dihe_baseflip(universe, bp1, bp2, i,
...
+    resn2 = bf2.atoms.resnames[0]
+    # extract positions for specific atoms of each base
+    c11 = bf1.select_atoms("name {0!s}".format(batms[resn1][0])).positions[0]
```
IAlibay: This will throw a KeyError if the residue is not part of batms and an IndexError if the names are wrong (since you're doing a positions call on a zero-width AtomGroup).

In both cases this will be quite incomprehensible to users if it fails. Could this be reformatted to add some try/excepts to make it clear to users as to why things fail (if they do)?

IAlibay (changes requested):
Couple of comments here - please feel free to disregard my comment re: switching away from using segID (pinging @orbeckst) if it's too much of a change.

I would definitely like to see a bit more usage safety checking on the selections though.

package/MDAnalysis/analysis/nuclinfo.py
aakognole: @IAlibay Thanks for your comments. I believe a combination of segid and resid should point to a unique residue in case of polynucleotides.

package/MDAnalysis/analysis/nuclinfo.py
aakognole: In this case we can only have the 5 residues that make up the DNA and RNA systems. We already inform the users that this analysis module is based on CHARMM specific names for nucleic acid residues and atom names. Again, the user just needs to check that the segid and resid are correct.

package/MDAnalysis/analysis/nuclinfo.py
IAlibay: I am willing to defer to @orbeckst here, but personally I really am not sure that this is the case for all topology types we support. Segments are very charmm-centric, and `resid` is non-unique. If we use the PDB format as an example, wouldn't this easily trip up if you had an altloc for a residue with resid N?

Given the number of `select_atoms` calls we're doing here, performance is really not going to be limited to a simple check on `len(resn1.residues)`.

package/MDAnalysis/analysis/nuclinfo.py
orbeckst: Normally I'd agree with you. However, for historical reasons all other functions do the same. In the interest of a common and consistent API I'd keep it and rather open an issue to rework the whole thing.

package/MDAnalysis/analysis/nuclinfo.py
IAlibay: This is putting a lot of onus on users to follow our docs and fully understand everything we mean. Personally, given that wrapping in a try/except here has nearly no cost and will immensely clarify failures, I fail to see how this is not worth it.

Again happy to defer to the opinions of @orbeckst as the original approving reviewer here.

orbeckst: I agree with @IAlibay here — even though it's the user's job to put in the right selection, telling them if they don't is a better approach than just failing with unintelligible stack traces (because then _we_ have to explain on the mailing list what the error means). 

The other functions don't have these additional checks but your function could do better than the old ones. (And this is not an API change but usability improvement.)

package/MDAnalysis/analysis/nuclinfo.py
IAlibay: That's fair, I'm happy to go with that philosophy.

package/MDAnalysis/analysis/nuclinfo.py
orbeckst: Valid argument, @IAlibay .

package/MDAnalysis/analysis/nuclinfo.py
aakognole: Please take a look. Thanks!

package/MDAnalysis/analysis/nuclinfo.py
```diff
@@ -25,17 +25,18 @@
 Nucleic acid analysis --- :mod:`MDAnalysis.analysis.nuclinfo`
 =============================================================
 
-:Author: Elizabeth Denning
-:Year: 2011
+:Authors: Elizabeth Denning, Abhishek A. Kognole
```
IAlibay, line 28: Only just picked this up - @orbeckst I've never been super clear on how these authorship entries worked in MDA, is this the correct way to address this? (and also do we have a clear view on when we add someone and when we don't?)

package/MDAnalysis/analysis/nuclinfo.py
```diff
@@ -763,3 +770,84 @@ def pseudo_dihe_baseflip(universe, bp1, bp2, i,
...
+    bf2 = universe.select_atoms(
+        "(segid {0!s} and resid {1!s})".format(seg2, b2))
+    if len(bf1.residues) > 1:
+        warnings.warn(
+            "Found more than one residues with same resid {0!s} !".format(b1)
+            "Using first residue in the selection for calculation.")
+    if len(bf2.residues) > 1:
+        warnings.warn(
+            "Found more than one residues with same resid {0!s} !".format(b2)
+            "Using first residue in the selection for calculation.")
```
IAlibay: I think this is incorrect syntax (i.e. adding a `format` in the middle of a string continuation).

If anything this needs to be fixed - probably using f-strings here would be better?

Aside from that your construct isn't wrong but I think we can clear this up a little bit, I'll let you make the choice if you want to squash it down to fewer lines with the following loop:

```suggestion
    for ag, segid, res in [(bf1, seg1, b1), (bf2, seg2, b2)]:
        if len(ag.residues) > 1:
            warnings.warn(f"Found more than one residues with resid {res} and "
                          f"segid {segid}. Using first residue in selection.")
```

package/MDAnalysis/analysis/nuclinfo.py
```diff
@@ -763,3 +770,84 @@ def pseudo_dihe_baseflip(universe, bp1, bp2, i,
...
+        c11 = bf1.select_atoms("name {0!s}".format(batms[resn1][0])).positions[0]
+        c12 = bf1.select_atoms("name {0!s}".format(batms[resn1][1])).positions[0]
+        c13 = bf1.select_atoms("name {0!s}".format(batms[resn1][2])).positions[0]
+        c21 = bf2.select_atoms("name {0!s}".format(batms[resn2][0])).positions[0]
+        c22 = bf2.select_atoms("name {0!s}".format(batms[resn2][1])).positions[0]
+        c23 = bf2.select_atoms("name {0!s}".format(batms[resn2][2])).positions[0]
+    except KeyError:
+        print("Residue names may be incorrect!"
+              "Please use either ADE/GUA/CYT/URA/THY or A/G/C/U/T format.")
+        raise
```
IAlibay: So the approach we would take here would be to raise the message with the error. Print calls can often be lost amongst a stack trace.

```suggestion
        errmsg = ("Residue names may be incorrect! "
                  "Please use either ADE/GUA/CYT/URA/THY or A/G/C/U/T format.")
        raise KeyError(errmsg)
```

package/MDAnalysis/analysis/nuclinfo.py
```diff
@@ -763,3 +770,84 @@ def pseudo_dihe_baseflip(universe, bp1, bp2, i,
...
+    except IndexError:
+        print("Atom names may be incorrect!"
+              "Make sure base atom names follow CHARMM format.")
+        raise
```
IAlibay: Similarly to the above:

```suggestion
        errmsg = ("Atom names may be incorrect! "
                  "Make sure base atom names follow CHARMM format.")
        raise IndexError(errmsg)
```

testsuite/MDAnalysisTests/analysis/test_nuclinfo.py
```diff
@@ -174,3 +174,13 @@ def test_hydroxyl(u, seg, i, expected_value):
...
+
+
+@pytest.mark.parametrize('b1, b2, seg1, seg2, expected_value', (
+        (2, 3, 'RNAA', 'RNAA', 15.490582),
+        (10, 11, 'RNAA', 'RNAA', 43.71941),
+        (7, 17, 'RNAA', 'RNAA', 168.37481),
+))
+def test_angle_between_base_planes(u, b1, b2, seg1, seg2, expected_value):
+    val = nuclinfo.angle_between_base_planes(u, b1, b2, seg1, seg2)
+    assert_allclose(val, expected_value, rtol=1e-3, atol=0)
```
IAlibay, line 191: Could you add in tests for the warnings being raised and the two exceptions please?

package/MDAnalysis/analysis/nuclinfo.py
```diff
@@ -763,3 +770,84 @@ def pseudo_dihe_baseflip(universe, bp1, bp2, i,
...
+        c12 = bf1.select_atoms("name {0!s}".format(batms[resn1][1])).positions[0]
+        c13 = bf1.select_atoms("name {0!s}".format(batms[resn1][2])).positions[0]
+        c21 = bf2.select_atoms("name {0!s}".format(batms[resn2][0])).positions[0]
+        c22 = bf2.select_atoms("name {0!s}".format(batms[resn2][1])).positions[0]
+        c23 = bf2.select_atoms("name {0!s}".format(batms[resn2][2])).positions[0]
+    except KeyError:
+        print("Residue names may be incorrect!"
+              "Please use either ADE/GUA/CYT/URA/THY or A/G/C/U/T format.")
+        raise
+    except IndexError:
```
IAlibay, line 870: It's quite rare to have two `except` clauses, but I think it's justified here.

IAlibay (changes requested):

package/MDAnalysis/analysis/nuclinfo.py
orbeckst: No real guide lines and I would be liberal with it for right now.

Truth be told, I'd actually like to get rid of these entries (which come from the earliest days...) to make the docs more uniform and because AUTHORS is really the most appropriate part for noting contributions as it makes clear that this is a collective work.

So, I'd just leave it as is, with the understanding that all of these entries will maybe suddenly disappear .

IAlibay: Yeah sounds good.

> Truth be told, I'd actually like to get rid of these entries (which come from the earliest days...) to make the docs more uniform and because AUTHORS is really the most appropriate part for noting contributions as it makes clear that this is a collective work.

That's also the feeling I have (hence the query here - I've just never reviewed a case of someone adding to the notice so I've just ignored them until now). That being said, we'd probably have to poll the individuals that wrote these notices to begin with, I'll raise an issue.

fiona-naughton:
I had a quick look at the [Gabb et al. paper](https://www.sciencedirect.com/science/article/abs/pii/0263785595000860), and it looks like you're defining the base pane in a different way (using atoms N1/N3/C5 or N9/N1/N7 rather than geometric center+two atoms). If we're assuming planarity of the rings it shouldn't affect the result, but is there are particular reason for you choice (is there a standard way it's usually done?). Might be worth explicitly mentioning how you calculate it here, since it differs from the cited paper.

(Also, the second paper ([Jafilin et al](https://pubs.acs.org/doi/full/10.1021/jp209986y)), seems to be about developing a stacking coordinate that uses the angle-between-base-planes rather than about the angle-between-base-planes calculation itself - including it with please-cite here seems unnecessary to me? )

aakognole:
> I had a quick look at the [Gabb et al. paper](https://www.sciencedirect.com/science/article/abs/pii/0263785595000860), and it looks like you're defining the base pane in a different way (using atoms N1/N3/C5 or N9/N1/N7 rather than geometric center+two atoms). If we're assuming planarity of the rings it shouldn't affect the result, but is there are particular reason for you choice (is there a standard way it's usually done?). Might be worth explicitly mentioning how you calculate it here, since it differs from the cited paper.

You are right, it is not exactly the same way. Assuming the planarity of the rings and using a modified selection of atoms to be consistent across the bases gives correct direction of the normal to the base plane. Mentioning the difference is a good idea. Thanks!

> (Also, the second paper ([Jafilin et al](https://pubs.acs.org/doi/full/10.1021/jp209986y)), seems to be about developing a stacking coordinate that uses the angle-between-base-planes rather than about the angle-between-base-planes calculation itself - including it with please-cite here seems unnecessary to me? )

Okay, yes although it is a good example of how the angle is used in defining the stacking coordinate, it may not be necessary to cite it.

IAlibay:
> You are right, it is not exactly the same way. Assuming the planarity of the rings and using a modified selection of atoms to be consistent across the bases gives correct direction of the normal to the base plane. Mentioning the difference is a good idea. Thanks!

Is this used/validated anywhere else? Ideally we would want to justify to users why we're taking a shortcut here. As far as I can tell doing either the Cremer Pople based approach or the geometric center + 2 atoms isn't necessarily that much more complicated than what is done here, so it would be good to justify reasoning.

package/MDAnalysis/analysis/nuclinfo.py
```diff
@@ -763,3 +770,80 @@ def pseudo_dihe_baseflip(universe, bp1, bp2, i,
...
+    except KeyError:
+        errmsg = ("Residue names may be incorrect! "
+                  "Please use either ADE/GUA/CYT/URA/THY or A/G/C/U/T format.")
+        raise KeyError(errmsg)
+    except IndexError:
+        errmsg = ("Atom names may be incorrect! "
+                  "Make sure base atom names follow CHARMM format.")
+        raise IndexError(errmsg)
+    # get normals to the planes of the bases
+    n1 = mdamath.normal(c12-c11, c13-c11)
```
richardjgowers: This isn't PBC aware and probably should be. See minimize_vectors which recently got added to lib.

richardjgowers (changes requested):

package/MDAnalysis/analysis/nuclinfo.py
aakognole: good point. Usually users recenter the trajectories before analysis and this may not be always necessary. So do you think in this case the box input to minimize_vectors should be optional or not?

package/MDAnalysis/analysis/nuclinfo.py
```diff
@@ -763,3 +771,118 @@ def pseudo_dihe_baseflip(universe, bp1, bp2, i,
...
+                  "Please use either ADE/GUA/CYT/URA/THY or A/G/C/U/T format.")
+        raise KeyError(errmsg)
+    except IndexError:
+        errmsg = ("Atom names may be incorrect! "
+                  "Make sure base atom names follow CHARMM format.")
+        raise IndexError(errmsg)
+    # check box
+    if box is None:
+        box = universe.dimensions
+        if box is None:
```
richardjgowers, line 877: Don't invent a box if there isn't one, it's incredibly hard to guess a box. Better to only do the call to minimize if there's a box, otherwise trust the vectors to be correct.

richardjgowers (changes requested):

testsuite/MDAnalysisTests/analysis/test_nuclinfo.py
aakognole: Added.

orbeckst:
@aakognole from the discussion above https://github.com/MDAnalysis/mdanalysis/pull/3480#issuecomment-1019553272 I gather that the method you propose here is different from the one in cited references. Is the approach here your own invention then? 

Did you use it in one of your published papers or are you about to publish a paper with it?

aakognole:
> @aakognole from the discussion above [#3480 (comment)](https://github.com/MDAnalysis/mdanalysis/pull/3480#issuecomment-1019553272) I gather that the method you propose here is different from the one in cited references. Is the approach here your own invention then?

Hi @orbeckst . There is hardly any invention here. It is simpler yet effective approximation of finding the normal to the base planes and getting angle between them to use further. 

> Did you use it in one of your published papers or are you about to publish a paper with it?

Not published yet, but I am working on it. You may put merging this PR on hold until then. Thanks.

IAlibay:
> Not published yet, but I am working on it. You may put merging this PR on hold until then. Thanks.

Just to keep you updated on this @aakognole . We are currently having some internal discussions amongst the core developer team on how to go forward with cases like this in the future.

Essentially the consensus is that we prefer not adding experimental (especially pre-publication) methods to the core library. Our aim in the near future will be to encourage folks to share new methods via an MDAKit (i.e. a self-contained python package that lives downstream from the core library).

In this particular case we are in agreement that we would at least like to wait until this is properly validated (i.e. published) _before_ we make a decision on merging. To this end I have added [hold pub] to the PR title. Once it is published, please do ping us again.

**Note:** Our approach to having `[hold pub]` labelled PRs may change in the near future once we have stricter rules on how long PRs can remain open.

IAlibay:
On a more reviewer 3 type of comment - my general understanding is that the only way to prove that this approximation works is to compare it against one of the previously published methods (i.e. as defined in the Gabb et al. paper). In that case, I would ask that this validation method (which I assume is already written in a similar way?) also be added as an option for users to call & compare against.

IAlibay:
@aakognole @orbeckst - any ideas on the status of this PR?

orbeckst:
I think the primary question is if @aakognole published a paper where they applied their method.

aakognole:
@orbeckst Thanks for your consideration. But this work was not published and I have moved on to another role. You may close the pull request.

IAlibay:
Thanks for updating us @aakognole. Please do reopen the PR if the situation changes.

PR #3482, opened 2021-12-18
More details on Dihedral doc

PicoCentauri:
Improve the doc of the Dihedral module. The changes are cosmetic but allow us to use the Dihedral module in the CLI since [PR82](https://github.com/MDAnalysis/mdacli/pull/82) adds handling of a List of AtomGroups. 

The type of the `atomgroups` attribute follows the [type hint](https://www.python.org/dev/peps/pep-0484/) recommendation.

### DIFF
--- a/package/MDAnalysis/analysis/dihedrals.py
+++ b/package/MDAnalysis/analysis/dihedrals.py
@@ -296,7 +296,7 @@ class Dihedral(AnalysisBase):
         """Parameters
         ----------
         atomgroups : list[AtomGroup]
-            a list of :class:`~MDAnalysis.core.groups.AtomGroup` for which 
+            a list of :class:`~MDAnalysis.core.groups.AtomGroup` for which
             the dihedral angles are calculated
 
         Raises
--- a/package/MDAnalysis/analysis/dihedrals.py
+++ b/package/MDAnalysis/analysis/dihedrals.py
@@ -295,8 +295,9 @@ class Dihedral(AnalysisBase):
     def __init__(self, atomgroups, **kwargs):
         """Parameters
         ----------
-        atomgroups : list
-            a list of atomgroups for which the dihedral angles are calculated
+        atomgroups : list[AtomGroup]
+            a list of :class:`~MDAnalysis.core.groups.AtomGroup` for which 
+            the dihedral angles are calculated
 
         Raises
         ------


IAlibay:
Just to check - this doesn't really follow the NumPy docstring standard right?

If so, I'm not sure it's a super great idea to start changing bits of our docstring style whilst we wait for type hinting to be better supported 🤔 Thoughts @lilyminium @orbeckst ? I'd prefer we put in sub-optimal type hinting than change bits we're eventually going to change back (if that makes sense).

orbeckst:
It renders sensibly [in the dihedrals docs](https://mdanalysis--3482.org.readthedocs.build/en/3482/documentation_pages/analysis/dihedrals.html)

image

so I am not too worried. It seems like a backwards-compatible change. Even if we don't do it everywhere, it will be fine.

If we decide to enhance the numpy doc string standard with optional [pep 484](https://www.python.org/dev/peps/pep-0484/) type hints then we would need to document it for us, though.

PicoCentauri:
Just as another remark: My proposed syntax only "works" in the interpreter for 3.9+. 

3.5 < 3.9 only allows for ```List[AtomGroup]``` (after typing was imported).

lilyminium:
@PicoCentauri for actual type hinting, not docs, right? Does documentation render alright with `List[AtomGroup]`?

PicoCentauri:
@lilyminium not as good as compared to `list[Atomgroup]`

Screenshot 2021-12-22 at 17 37 37

richardjgowers (commented):
I think this is a clear improvement, especially since it's rendering nicely

lilyminium (approved):
Yeah, it looks like a strict improvement -- thanks @PicoCentauri.

@IAlibay pointed out that `list[AtomGroup]` might not be the solution we go with immediately, and @PicoCentauri's mentioned that it's a 3.9+ feature wrt to actual type hinting. IMO it'll be a while before 3.9 is our minimum version, so we'll need to figure out if we want to start type-hinting with `typing` before switching, or just wait until 3.9+ to do hints. However, using this 3.9 syntax for docs right now isn't a bad idea -- I especially like that `list` links to the actual list type.

IAlibay (approved):
As long as we're happy that this is the future of our docs then I'm happy. I will ask that a separate issue be raised about updating our userguide / wiki to match this new style.

orbeckst:
@PicoCentauri I am merging.

Could you please *raise an [issue in the UserGuide](https://github.com/MDAnalysis/UserGuide/issues)* that reminds us to update our Style docs (and link to this PR). Thanks.

Issue #3043, opened 2020-11-25
Hole2 skipifs either improperly applied or given the wrong reason


IAlibay:
## Expected behavior ##

All test_hole2.py tests pass if skipifs reasons are met.

#### sys.version_info checks:

`test_plot3D`, `test_plot3D_rmax` and `test_plot3D_order_parameters` all state "get_data_3d requires 3.1 or higher" as skipifs reasons. However the check is `skipif(sys.version_info < (3,1)` not `skipif(sys.version_info > (3,1)`.

edit: now that we support 3.6+ only, we can just remove these checks.

#### rlimits_missing check:

`test_hole_module_fd_closure` shouldn't be skipped if `resource.getrlimit` can be called. This is the case on my local machine, but the test is still being skipped.

I have tested fixing the sign for the sys.version_info checks and removing the rlimits_missing check and the tests all run fine on my local machine.

## Code to reproduce the behavior ##

`pytest test_hole2.py`

## Current version of MDAnalysis ##

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) 2.0-dev0
- Which version of Python (`python -V`)? 3.8
- Which operating system? Ubuntu 20.04 LTS

IAlibay:
Addressed in #3489

PR #3489, opened 2021-12-29
Fixes frame indexing in create_vmd_surface

### LINKED ISSUE #3476: Suggestion of the bugfix hole.py create_vmd_surface function
The function create_vmd_surface is not working if you compute hole analysis starting from a specific frame.

hole.py Line 799:   ` for i, sphpdb in zip(self.frames, self.sphpdbs[self.frames])`
leads to index mismatch between two arrays if self.frames[0] != 0.

The fix that worked for me `for i, sphpdb in zip(self.frames, self.sphpdbs[np.indices(self.frames.shape, sparse = True)])`

Not sure if this is the most elegant solution, but it seems to work :)

IAlibay:
Fixes #3476 

So this _fixes_ the indexing issue, but unfortunately I a) don't have proper access to VMD right now, and b) my knowledge of VMD scripting is more than rusty. - (for @lilyminium or another @MDAnalysis/coredevs  who might know better on how this all works / is able to check) is the way `set triangles` is invoked right now allowing this to work on a non-contiguous trajectory in VMD?

 - changes the loop over self.frames to index self.results.sphpdbs properly
 - swaps internally the use of self.outfiles and self.sphpdbs instead of self.results.outfiles and self.results.sphpdbs

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/analysis/hole2/hole.py
+++ b/package/MDAnalysis/analysis/hole2/hole.py
@@ -83,6 +83,8 @@ The class can be set-up and run like a normal MDAnalysis analysis class::
     from MDAnalysis.tests.datafiles import MULTIPDB_HOLE
     from MDAnalysis.analysis import hole2
 
+    u = mda.Universe(MULTIPDB_HOLE)
+
     ha = hole2.HoleAnalysis(u, executable='~/hole2/exe/hole') as h2:
     ha.run()
     ha.create_vmd_surface(filename='hole.vmd')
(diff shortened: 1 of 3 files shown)


orbeckst:
I am trying out your patch with VMD 1.9.4a48:

```python
import MDAnalysis as mda
from MDAnalysis.tests import datafiles
from MDAnalysis.analysis import hole2

u = mda.Universe(datafiles.MULTIPDB_HOLE)
BINDIR = "/Users/oliver/hole2/exe"
h2 = hole2.HoleAnalysis(u, executable=BINDIR + "/hole", sos_triangle=BINDIR + "/sos_triangle", sph_process=BINDIR + "/sph_process")
h2.run(start=1, stop=9, step=2)
h2.create_vmd_surface(filename="hole_1_7_2.vmd")
```
This creates hole profiles for frames 1, 3, 5, 7
```python
>>> h2.results.outfiles
array(['hole001.out', 'hole003.out', 'hole005.out', 'hole007.out'],
      dtype=object)
```

Create subsampled trajectory and initial frame for visualization
```python
u.atoms.write("frame0.pdb")
u.atoms.write("frames_1_7_2.trr", frames=list(range(1, 9, 2)))
```

Load in VMD, delete frame 0 (to get a trajectory with the original frames 1, 3, 5, 7).

**ERROR**: VMD cannot load the sliced surface trajectory
```tcl
(tmp) 40 % source hole_1_7_2.vmd
can't read "vmd_frame(-1)": no such element in array
```
The problem is that in VMD trajectories always start at 0 and are indexed with consecutive integer frames.

I'll see if I can change the code.

package/MDAnalysis/analysis/hole2/hole.py
```diff
@@ -915,7 +915,7 @@ def create_vmd_surface(self, filename='hole.vmd', dot_density=15,
                 pass
 
             tri = '{ { ' + ' } { '.join(list(map(' '.join, shapes))) + ' } }'
-            frames.append('set triangles({i}) '.format(i=i) + tri)
+            frames.append(f'set triangles({frame}) ' + tri)
```
orbeckst: Need to index with `i` for VMD, see https://github.com/MDAnalysis/mdanalysis/pull/3489#issuecomment-1003202993.

orbeckst (changes requested):
Currently fails in VMD, see https://github.com/MDAnalysis/mdanalysis/pull/3489#issuecomment-1003202993.

I'll see if I can quickly fix and test it.

package/MDAnalysis/analysis/hole2/hole.py
IAlibay: Does it make sense to just do that, output N "consecutive frames" (which may have been picked non-consecutively) and let users deal with the mismatch?

orbeckst: The only way that VMD will "play" the surface movie is by writing consecutive frames. It's the user's responsibility to write the appropriate trajectory as in my example in the comment.

orbeckst:
Currently two other hole tests fail...
```
FAILED MDAnalysisTests/analysis/test_hole2.py::TestHole::test_correct_input - assert False
FAILED MDAnalysisTests/analysis/test_hole2.py::TestHole::test_input_options - assert False
```
Not sure what's happening... looking into it. Sigh.

orbeckst:
crap... my emacs python mode deletes trailing whitespace at the end of the line, which killed tests...

orbeckst:
I can't be bothered to dig into configuring my emacs right now — according to https://stackoverflow.com/questions/14164292/my-emacs-deletes-trailing-white-space-how-can-i-disable-this-behaviour and https://emacs.stackexchange.com/questions/21868/disable-removal-of-trailing-whitespace-in-emacs and https://www.emacswiki.org/emacs/DeletingWhitespace this is too much effort.

Workaround solution:
```
vim test_hole2.py
```
:-p

orbeckst (approved):

orbeckst:
@IAlibay i approved the PR. Please have a look yourself and recruit another reviewer if necessary. I leave it to you to merge when you’re happy.

IAlibay (commented):
Thanks @orbeckst lgtm - I'm going to hand off the final merge decision here to @lilyminium (mostly because I don't fully understand what's going on with those plot3d skip-if deleted tests [it's actually an issue I raised quite some time ago! #3043 ]).

lilyminium:
Thanks @IAlibay, I'll go over this tomorrow -- USA -> AUS was harder than the other way around.

lilyminium:
> The problem is that in VMD trajectories always start at 0 and are indexed with consecutive integer frames.

There are a couple alternative or additional solutions:

* write empty triangles for the frames not in the analysis. That's fairly easy to do, as we have access to the underlying `_trajectory` and therefore know the length
* write out an additional `hole_surface.pdb` trajectory of the analysed frames, to go with the `hole_surface.vmd` of analysed frames currently written in this PR. (However, the written atoms will have to be all the atoms in the universe, as the analysed AtomGroup is an UpdatingAtomGroup. The default selection is "protein" so not sure if this is necessary either)

What do you think? I'm somewhat inclined to think that people will prefer the latter and it's more usable, but I don't really use HOLE much myself. @mavstrikova, do you have any opinions on this?

lilyminium:
Holding off on official approval and merge until further discussion of the question raised in the previous comment, but this looks good to me -- thank you @orbeckst and @IAlibay! I did find an unrelated bug in #3493 .

orbeckst:
I would keep it as simple as possible. I would not automatically write out a trajectory, especially not in the terrible multi-PDB format. I added the docs so that users could do trajectory writing if they wanted to. 

I would also not write a trajectory with triangles at the skipped step as this increases the file for anyone who wants to write their own trajectory. 

To accommodate everyone we would need many additional switches to tune the output behavior. That needs to be tested…

Instead of overengineering and guessing what users would want I’d rather keep it simple and make sure it works. 

In summary: I’d merge as is.

IAlibay:
So I _think_ the main use case we should cover here is:

Creating a HOLE vmd surface with start=10, stop=20, step=5, and then load the full trajectory in VMD whilst also defining the same start/stop/step inputs on file selection (in the molecule file browser) and then also load the newly created surface.

I don't have the ability to test this, but my understanding from @orbeckst's comments is that the way it's currently implemented in this PR is the most likely way to work for this?

I would also agree with @orbeckst that whilst @lilyminium's options would work, they add a certain extra level of complexity on top (plus in the case of writing empty frames kinda breaks the idea of MDA's start/stop/step?).

orbeckst:
> Creating a HOLE vmd surface with start=10, stop=20, step=5, and then load the fully trajectory in VMD but also defined the same start/stop/step inputs on file selection (in the molecule file browser) and then also loaded the newly created surface.

This works if one remembers that VMD's *Last* is **inclusive** so for the following
```python
>>> u.trajectory
<PDBReader ~/mdanalysis/testsuite/MDAnalysisTests/data/1grm_elNemo_mode7.pdb.bz2 with 11 frames of 264 atoms>
>>> with hole2.HoleAnalysis(u, executable=BINDIR+"/hole", sos_triangle=BINDIR+"/sos_triangle", sph_process=BINDIR+"/sph_process") as h2:
        h2.run(start=1, stop=9, step=2)
        h2.create_vmd_surface(filename="hole_192.vmd")
>>> w = mda.Universe("t_192.trr")
>>> w.trajectory
<TRRReader t_192.trr with 4 frames of 264 atoms>
>>> [ts.frame for ts in w.trajectory]
[0, 1, 2, 3]
```
you need to load the original trajectory with 
* First: 1
* **Last: 7** (and not 9!)
* Stride: 2

VMD_load_with_stride
 
Then you can load the surface file
```tcl
vmd > source hole_192.vmd
```
just fine.

orbeckst:
I am adding something to the docs.

orbeckst:
I updated the docs with a force-push.

orbeckst:
Not sure what the codcov error is about
```
[2022-01-03T17:55:52.383Z] ['error'] There was an error running the uploader: Error uploading to https://codecov.io: Error: There was an error fetching the storage URL during POST: 404 - Not Found - {'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}
```

IAlibay:
> Not sure what the codcov error is about
> 
> ```
> [2022-01-03T17:55:52.383Z] ['error'] There was an error running the uploader: Error uploading to https://codecov.io: Error: There was an error fetching the storage URL during POST: 404 - Not Found - {'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}
> ```

Looks like there's something going on with GH Actions - some weird timeout errors on random runners. I've restarted CI, let's see what happens.

lilyminium:
Failing test is due to using `assert_equal` to compare floats in testing mdamath.angle; this should probably be `assert_allclose`.

orbeckst:
Alright, I'll procrastinate a little bit and fix it... thanks for identifying the issue @lilyminium

IAlibay:
> Alright, I'll procrastinate a little bit and fix it... thanks for identifying the issue @lilyminium

Already got a PR open for this @orbeckst (#3495)

lilyminium (approved):

Issue #3442, opened 2021-10-20
Add march=native build to cron job


hmacdope:
## Is your feature request related to a problem? ##

Enabling of an advanced "march=native" build `setup.cfg` Issue #3428 PR #3429 means that it should be tested, including on ARM and PowerPC in the weekly cron job.

## Describe the solution you'd like ##

Add build(s) with `march=native` to the weekly cron job. We may need to have a git patch apply a change to setup.cfg OR have some way of enabling the march=native build independent of setup.cfg, eg a command line flag to setup.py.

IAlibay:
The lazy person in me is just going to do a sed call, that's the least painful way.

hmacdope:
@IAlibay @orbeckst is there a canonical way to pass installation options on the command line ? I think having this as a flag rather than a "comment this back in" is probably cleaner.

richardjgowers:
ENV variables? Not sure if they’re officially supported but sniffing for
C_FLAGS would seem sane.

IAlibay:
env variables are a bit meh IMHO. I'm pretty sure you can pass it via the setuptools CLI by subclassing something, I'll have to have a poke through some docs later.

IAlibay:
I guess our current setup is built for env variables, so it's probably best to stick with that, it's in the setup docs so I assume they _should_ be officially supported.

IAlibay:
To expand on this - right now (on current develop) if you were to assign the env variable "MDA_MARCH=True" then it would trigger the march variable.

hmacdope:
I am going to close this for now, as we decided in #3429 to allow an option for **any** compiler flags rather than just `march=native`, making this issue obsolete.

IAlibay:
I'm not sure I agree we should close - we're still going to document the use of march=native as a flag right? So we should test it...

hmacdope:
I guess my understanding of discussion in #3429 was that we were going to rely on the user to **specify** any additional compiler flags (it would be blank by default).

 But I see now that if we **document** the use of a flag specifically, we should test for it (if that is an okay use of CI resources?).

Sorry for the perhaps premature close.

IAlibay:
>  (if that is an okay use of CI resources?).

It's a reminder for a weekly Cron job we'll probably add in the near future, it's not a waste of CI if it's the one job once a week IMHO.

hmacdope:
[...]

Issue #1727, opened 2017-12-08
Add pip install cron job


richardjgowers:
We currently assume that the package and all dependencies are installable via pip.  We need to set up a cron job on travis which checks this assumption

See:
https://github.com/MDAnalysis/mdanalysis/pull/1693#issuecomment-349390539

palnabarun:
Hi Richard, Can I work on this?

PR #3490, opened 2021-12-29
Cleanup of CI actions & cron job

### LINKED ISSUE #3216: Add CI runner with older versions of chemfiles
> Can you open an issue about covering this so we don't forget? I was working on a weekly cron CI job somewhere else, I can probably just add a workflow that installs older versions of chemfiles.

_Originally posted by @IAlibay in https://github.com/MDAnalysis/mdanalysis/pull/3126#r607320587_

IAlibay:
Fixes #3216 #1727
Towards #3052 #3442

 - Implement composite actions for similar steps across multiple workflow jobs
   - Adds a fully tune-able composite workflow for conda/pip installing, allowing us to point mutate any dependency whilst still having every default defined in a single file.
 - A variety of cron jobs
   - pip only install
   - install with numpy & scipy nightly dev versions
   - < 10.0 version of chemfiles-python
   - "march=native mtune=native" cflags build

To do:
 - Make gh-ci-cron.yaml purely a cron workflow (after review).

PR Checklist
------------
 - Tests?
 - Docs?
 - CHANGELOG updated?


IAlibay:
@MDAnalysis/coredevs should be finished with this one.

It's a pretty big overhaul of CI - but for the most part the "big changes" are just moving things over to composite actions to make things easier to re-use.

All that's left to do is amend the trigger on gh-ci-cron.yaml to just be a cron job (kept as it currently is so we can make sure CI actually works).

orbeckst (approved):
I gave it a cursory glance and I like the modularization.

If no-one else with sufficient GitHub-foo has time to review, I'd say let's try it...

IAlibay:
> If no-one else with sufficient GitHub-foo has time to review, I'd say let's try it...

Thanks @orbeckst, I was hoping maybe @fiona-naughton or @lilyminium might also want to have a look through. I think we might want to publish (in the near future) some of these composite actions to streamline downstream develop builds in MDAKits.

orbeckst:
More eyes would be good, especially with a view towards using it for the MDAkit cookie cutter. 

When I had opened the PR in the browser I hadn’t seen your comment and I then came back to the tab at the end of the day.

lilyminium:
Modularising GH actions makes the main CI file much tidier, thank you. The default definitions of dependencies is long but allows for much easier mixing and matching. I did just make a PR making the dependency environment variables more list-like and therefore easier to diff; they're also printed in the log as part of the env.

> I think we might want to publish (in the near future) some of these composite actions to streamline downstream develop builds in MDAKits.

That also sounds like a good idea. The cookiecutter probably won't need all of them, and this makes it way easier to split out. (One model that many projects follow successfully is installing test dependencies from an `environment.yaml`, for example; and that also helps with having a yaml file on hand for people building their own development environments.)

Actually, a GH Action installing `develop` might be handy for downstream projects.

.github/actions/build-src/action.yaml
```diff
@@ -0,0 +1,69 @@
...
+        which pip
+        pip list
+        conda info
+        conda list
+
+    - name: build_hole
+      if: ${{ inputs.build-hole == 'true' }}
+      shell: bash -l {0}
+      run: |
+        # We manually build hole2 to avoid OS incompatibilities
```
richardjgowers, line 43: Separate issue, but isn't this something a hole2 conda package would solve?

.github/actions/build-src/action.yaml
```diff
@@ -0,0 +1,69 @@
...
+        source ../source.unset
+        echo "HOLE_BINDIR=${HOME}/hole2/bin" >> $GITHUB_ENV
+        echo "${HOME}/hole2/bin" >> $GITHUB_PATH
+
+    - name: build_mda_main
+      shell: bash -l {0}
+      run: |
+        # TODO: using install instead of develop here causes coverage to drop
+        # for .pyx file. If possible a solution for this should be found.
+        cd package/ && python setup.py develop
```
richardjgowers, line 56: I think this is because with pyx files the pyx source files don't get put into installation (as they become .so and unrecognisable?), whereas a develop install leaves everything "in tree" and more visible to coverage.  Is there a specific problem with a develop install?

.github/workflows/gh-ci-cron.yaml
```diff
@@ -0,0 +1,203 @@
...
+        build-tests: true
+        build-docs: false
+
+    - name: run_tests
+      run: |
+        pytest -n $numprocs testsuite/MDAnalysisTests --disable-pytest-warnings --durations=50
+
+
+  # Issue 3216
+  old-chemfiles:
```
richardjgowers: I'm not sure we should be maintaining support for older chemfiles versions.  The package is pre 1.0 and will continually be unstable, so we're signing up for a lot of work.

richardjgowers (approved):
LGTM, might be worth cherry-picking @lilyminium 's patch onto this to avoid a merge snafu

.github/actions/build-src/action.yaml
IAlibay: Yeah - I want to say @orbeckst mentioned looking into that at some point (or was asking for someone to look into that?)

If I remember correctly part of the limitation here is that we wanted to update the upstream repository (since MDAnalysis' version is a fork)? That being said, it's been a year and that PR (https://github.com/osmart/hole2/pull/2) hasn't been merged. I have raised an issue on the MDA's hole2 repo.

.github/actions/build-src/action.yaml
IAlibay: Not that I'm aware of, this was a TODO carried over from our original actions build. The pyx source file explanation makes sense, I've changed the comment to just say why we use `develop` instead of `install`.

.github/workflows/gh-ci-cron.yaml
IAlibay: Currently our code paths enable support for both pre and post chemfiles 0.10. I agree that we probably shouldn't waste time trying to support old pre-1.0 here, in which case we should just drop support for it (in a separate PR).

I've raised a separate PR. Let's keep this in for now and remove it in the follow-up PR?

IAlibay:
Just going to let @fiona-naughton review this before switching things to cron-only and merging.

fiona-naughton (approved):
sorry for the delay - lgtm!

PR #3491, opened 2021-12-29
Add millisecond unit

### LINKED ISSUE #3474: `millisecond` conversion
```
  File "/home/joao/anaconda3/envs/taurenmddev/lib/python3.8/site-packages/MDAnalysis/units.py", line 363, in convert
    raise ValueError(errmsg) from None
ValueError: unit 'ms' not recognized.
It must be one of Angstrom, A, angstrom, Å, nm, nanometer, pm, picometer, fm, femtometer, Angstrom^{-3}, A^{-3}, Å^{-3}, nm^{-3}, nanometer^{-3}, Molar, SPC, TIP3P, TIP4P, water, ps, picosecond, fs, femtosecond, ns, nanosecond, second, sec, s, AKMA, e, Amber, C, As, Angstrom/ps, A/ps, Å/ps, Angstrom/picosecond, angstrom/picosecond, Angstrom/fs, Angstrom/femtosecond, angstrom/femtosecond, angstrom/fs, Angstrom/AKMA, nm/ps, nanometer/ps, nanometer/picosecond, nm/ns, pm/ps, m/s, kJ/(mol*Angstrom), kJ/(mol*A), kJ/(mol*Å), kJ/(mol*nm), Newton, N, J/m, kcal/(mol*Angstrom), kJ/mol, kcal/mol, J, eV.
```

Would it be interesting to implement `millisecond` (`ms`) conversion in `MDAnalaysis.units.py` ?

Atharva7K:
Fixes #3474

- Added support for millisecond unit. 

PR Checklist
------------
 - [X] Tests?
 - [ ] Docs?
 - [X] CHANGELOG updated?


Atharva7K:
I noticed that some combinations  using `lengthUnit ` and `timeUnit `are missing in the `speedUnit_factor`  dictionary. Shall I add them?

Atharva7K:
One test is failing apparently. Any guidance on how to resolve it will be highly appreciated.

IAlibay (changes requested):
Thanks for starting this @Atharva7K 

The failing test is codecov saying that there are new code lines without test coverage. Could you add tests to cover the new additions? There's a relevant test file under `testsuite/MDAnalysisTests/utils/test_units.py` where tests for units are included. We probably will need a new entry to TestConversion for velocities.

With regards to the missing combinations for speedUnit - I'm not sure, these really exist to handle conversion of units in various file formats (as far as I understand the history of MDAnalysis' units module). I don't really know of a file format that deals in nm/fs or nm/ms (indeed I'm not sure I know of a file format that deals in milliseconds either). That being said I'm probably missing a use case - maybe @orbeckst or @joaomcteixeira might want to weigh in on this?

edit: please remember to add yourself to the AUTHORS and also add an entry to CHANGELOG.

package/MDAnalysis/units.py
```diff
@@ -239,6 +239,7 @@
     'ps': 1.0, 'picosecond': 1.0,  # 1/1.0
     'fs': 1e3, 'femtosecond': 1e3,  # 1/1e-3,
     'ns': 1e-3, 'nanosecond': 1e-3,  # 1/1e3,
+    'ms': 1e-9, 'millisecond': 1e-9,  # 1/1e9,
```
joaomcteixeira, line 242: Since you reach this far, why don't you add also microsecond? :relaxed: 

Good job! :+1:

joaomcteixeira:
> With regards to the missing combinations for speedUnit - I'm not sure, these really exist to handle conversion of units in various file formats (as far as I understand the history of MDAnalysis' units module). I don't really know of a file format that deals in nm/fs or nm/ms (indeed I'm not sure I know of a file format that deals in milliseconds either). That being said I'm probably missing a use case - maybe @orbeckst or @joaomcteixeira might want to weigh in on this?

I don't know also. BUT, if the implementation is as easy as adding more lines describing the combinations, I would add them. In this particular case, thinking for the future I think it does not hurt.

package/MDAnalysis/units.py
Atharva7K: On it. Thanks for the review!

joaomcteixeira (commented):
This is cool! I guess you only need to add the tests now :-)

IAlibay:
@Atharva7K - due to the way our github organisation is currently set up, continuous integration workflows have to be approved for new contributors.

In order to avoid excess carbon footprint, I'll wait until tests are implemented to approve the workflow. Do ping us when this is done or if you need more help for this.

Atharva7K:
> @Atharva7K - due to the way our github organisation is currently set up, continuous integration workflows have to be approved for new users.
> 
> In order to avoid excess carbon footprint, I'll wait until tests are implemented to approve the workflow. Do ping us when this is done or if you need more help for this.

I completely understand. Working on the tests, will complete them soon

Atharva7K:
I've implemented the tests @IAlibay . I've tested using random combination of units since there were a lot of possibilities :). If you feel this was a bad idea, please tell me I will then add specific tests or more if you want.

Also, is it necessary to do repetitive calculations again while testing. For instance, the value 1.0 * 1e3 is being calculated thrice during angstrom/femtosecond speed factor calculation. Doesn't it make more sense to cache the value and use that instead?

IAlibay:
> I've implemented the tests @IAlibay . I've tested using random combination of units since there were a lot of possibilities :). If you feel this was a bad idea, please tell me I will then add specific tests or more if you want.
> 
> Also, is it necessary to do repetitive calculations again while testing. For instance, the value 1.0 * 1e3 is being calculated thrice during angstrom/femtosecond speed factor calculation. Doesn't it make more sense to cache the value and use that instead?

Thanks, I've approved CI.

re: cache-ing - honestly the cost of doing a single float multiplication (even considering how slow python is), is so infinitesimally (probably a couple of CPU cycles), that it's really not worth thinking about. You really wouldn't see any gains trying to cache it. In fact, given that we run tests in parallel, you'd probably lose out on more CPU cycles trying to invoke a synchronisation.

IAlibay:
>  I've tested using random combination of units since there were a lot of possibilities :). If you feel this was a bad idea, please tell me I will then add specific tests or more if you want.

We'll see what codecov returns with, but in my opinion ideally you should invoke at least each new entry once. It doesn't need to be a complete all-to-all conversion matrix, but at least we want to test at least once that each entry behaves in a way we think is normal.

Atharva7K:
> > I've tested using random combination of units since there were a lot of possibilities :). If you feel this was a bad idea, please tell me I will then add specific tests or more if you want.
> 
> We'll see what codecov returns with, but in my opinion ideally you should invoke at least each new entry once. It doesn't need to be a complete all-to-all conversion matrix, but at least we want to test at least once that each entry behaves in a way we think is normal.

Since codecov is failing again, I'll add more test cases. I'll need some days to do so though, busy with some other stuff at the moment. Also thanks for the clarification regarding cache-ing.

Atharva7K:
@IAlibay there are unexpected coverage changes in `TRZ.py` file even though I haven't touched it. Can you please help me resolve this issue?

testsuite/MDAnalysisTests/utils/test_units.py
```diff
@@ -96,6 +100,14 @@ def test_energy(self, quantity, unit1, unit2, ref):
     def test_force(self, quantity, unit1, unit2, ref):
         self._assert_almost_equal_convert(quantity, unit1, unit2, ref)
 
+    @pytest.mark.parametrize('quantity, unit1, unit2, ref', (
+        (1, 'A/ps', 'm/s', 1e-10/1e-12),
+        (1, 'A/ps', 'nm/ps', 0.1),
+        (1, 'A/ps', 'pm/ps', 1e2),
```
IAlibay, line 106: Can you add some entries to cover Angstrom/microsecond, Angstrom/millisecond, Angstrom/AKMA, Angstrom/femtosecond please?

Just converting between these and A/ps is fine (doesn't need to be all-to-all).

IAlibay (changes requested):
The TRZ codecov return is because your feature branch is a couple of commits behind the main MDAnalysis `develop` branch. It doesn't affect anything so you don't need any further changes to address that.

Beyond the review comment I just made, if you can also add yourself to AUTHORS and relevant entries to CHANGELOG please.

testsuite/MDAnalysisTests/utils/test_units.py
Atharva7K: Okay I will add the mentioned tests, update AUTHORS and CHANGELOG too. Also, my apologies  for bothering you over the silly codecov return I didn't realise it had to do with my feature branch being behind, I will be more careful from next time. Thanks for the review.

package/AUTHORS
```diff
@@ -171,6 +171,7 @@ Chronological list of authors
   - Marcelo C. R. Melo
   - Mark D. Driver
   - Kevin Boyd
+  - Atharva Kulkarni
```
IAlibay, line 175: It's now 2022, a new entry needs to be added
```suggestion
2022
  - Atharva Kulkarni
```

package/CHANGELOG
```diff
@@ -32,6 +32,7 @@ Fixes
     created using add_Residue and add_Segment (Issue #3437)
 
 Enhancements
+  * Add support for millisecond and microsecond units (Issue #3473, PR #3491)
```
IAlibay, line 35: You also need to add yourself on line 17.

IAlibay (changes requested):
Thanks @Atharva7K just a final set of comments re: CHANGELOG and AUTHORS

IAlibay (approved):
Thanks @Atharva7K 👍🏽

PR #3496, opened 2022-01-04
Issue#3493

### LINKED ISSUE #3493: exe_err is never imported in hole2/hole.py
## Expected behavior ##

`exe_err` is used to raise errors four times if various executables aren't found, but it's never actually imported in hole.py. I imagine this hasn't come up yet due to poor test coverage.

``` python
>>> import MDAnalysis as mda
>>> from MDAnalysis.tests.datafiles import MULTIPDB_HOLE
>>> from MDAnalysis.analysis import hole2
>>> u = mda.Universe(MULTIPDB_HOLE)
/Users/lily/pydev/mdanalysis/package/MDAnalysis/topology/PDBParser.py:317: UserWarning: Element information is missing, elements attribute will not be populated. If needed these can be guessed using MDAnalysis.topology.guessers.
  warnings.warn("Element information is missing, elements attribute "
>>> h2 = hole2.HoleAnalysis(u, executable="~/dev")
  File "/Users/lily/pydev/mdanalysis/package/MDAnalysis/analysis/hole2/hole.py", line 727, in __init__
    raise OSError(errno.ENOENT, exe_err.format(name=hole,
NameError: name 'exe_err' is not defined
```

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) PR #3489 
- Which operating system?

Atharva7K:
Fixes #3493

 - Added `exe_error` import in hole.py

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


Atharva7K:
Please guide me on where to add the tests.

Atharva7K:
Apparently, I was working on #3474 and forgot to switch to the develop branch before starting a new branch for this issue, that's why the commits from#3474 are visible here. Shall I close this PR and start a new one for this issue? Or there's something else we can do? Apologies for the inconvenience caused, I'm stil learning git.

lilyminium:
@Atharva7K that's a perfectly understandable slip. There are several approaches you can take:

* Delete this branch and start a new one from develop, as you mentioned
* (more advanced) use git rebase to edit the commit history of this PR. I usually do this interactively by going back to an early commit you want to keep, e.g. develop `git rebase -i 79104dc14cf8567af9b349e42ea39a051d87d0a3`. Then you can delete each line of a commit that you don't want.

Regarding tests, have a look at `mdanalysis/testsuite/MDAnalysisTests/analysis` -- the test suite structure reflects the structure of the actual package, so the hole analysis tests should be in there.

Atharva7K:
> @Atharva7K that's a perfectly understandable slip. There are several approaches you can take:
> 
>     * Delete this branch and start a new one from develop, as you mentioned
> 
>     * (more advanced) use git rebase to edit the commit history of this PR. I usually do this interactively by going back to an early commit you want to keep, e.g. develop `git rebase -i 79104dc14cf8567af9b349e42ea39a051d87d0a3`. Then you can delete each line of a commit that you don't want.
> 
> 
> Regarding tests, have a look at `mdanalysis/testsuite/MDAnalysisTests/analysis` -- the test suite structure reflects the structure of the actual package, so the hole analysis tests should be in there.

@lilyminium thanks for the solution, I've edited the commit history using the latter method, I will push the changes after I've added the tests to prevent running CI again. Also, the hole analysis test consists of many classes, can you please guide me on which class to add the test under? If you can also elaborate a bit more on how I can generate the correct error messages for the tests, that'd be great.  Thanks for the help.

IAlibay:
>Also, the hole analysis test consists of many classes, can you please guide me on which class to add the test under? If you can also elaborate a bit more on how I can generate the correct error messages for the tests, that'd be great. Thanks for the help.

_Re: where to add the tests_

My suggestion here is to look at where the other similar tests are being done. There are two main components to hole2, the `hole` method and the `HoleAnalysis` class. One option here is to make sure that you add tests to the existing class that is most relevant in testing that specific component. The other option, if you are testing lots of similar things, is to create a new test class specifically checking for these errors.

_Re: generating the correct error messages_

Since this is a `GSoC starter` labelled issue, it's not clear to me if this is a normal pull requests or an attempt to apply to GSoC in the future. If it's the latter case it somewhat defeats the purpose of it being a `GSoC starter` issue if we tell you how to specifically add these tests. Maybe the best way here is for you to tell us how you've attempted to trigger the errors to this point and what isn't working?

lilyminium:
@Atharva7K Just to add onto @IAlibay's reply -- there are a number of tests elsewhere in the test suite for checking that a correct error, or error message, is raised; you could have a look at those by searching "with pytest.raises". For triggering the error, when pull requests are made to fix an issue, it's usually a good idea to have a look at the specific problem encountered in the issue to ensure that it is fixed. To check that you've written a good test, it's common to undo the fix, check that the test fails, and then redo the fix to check that the test now passes.

Atharva7K:
Thank you for the guidance @lilyminium and @IAlibay . Just to make sure that I've understood the problem I have to solve correctly: I have to write tests to ensure that an `OSError` is raised with the appropriate error message in four cases:

- when the `hole` executable isn't found when the hole method is called.
- when `hole`,  `sos_triangle` and `sph_process` executables are not found while the `HoleAnalysis` object is created.

Please correct me If my understanding is wrong.

IAlibay:
> Thank you for the guidance @lilyminium and @IAlibay . Just to make sure that I've understood the problem I have to solve correctly: I have to write tests to ensure that an `OSError` is raised with the appropriate error message in four cases:
> 
> * when the `hole` executable isn't found when the hole method is called.
> * when `hole`,  `sos_triangle` and `sph_process` executables are not found while the `HoleAnalysis` object is created.
> 
> Please correct me If my understanding is wrong.

Yes, that sounds about right to me.

Atharva7K:
> > Thank you for the guidance @lilyminium and @IAlibay . Just to make sure that I've understood the problem I have to solve correctly: I have to write tests to ensure that an `OSError` is raised with the appropriate error message in four cases:
> > 
> > * when the `hole` executable isn't found when the hole method is called.
> > * when `hole`,  `sos_triangle` and `sph_process` executables are not found while the `HoleAnalysis` object is created.
> > 
> > Please correct me If my understanding is wrong.
> 
> Yes, that sounds about right to me.

Thanks for the reply I've added the tests. Now, do I also have to test for the absence of  ` sos_triangle` and `sph_process` executables too? Because that would require the hole executable to be in the codebase somewhere as the error will only be thrown by the absence of these files once hole is found. If that's the case then I have to add the hole binary somewhere in the testsuite and I am not sure whether it will be a good practice or not.
Or in a completely different case, my implementation of the tests is wrong and I will need to think of some other approach.

Also, In order to simulate the absence of these files, I have used a dummy path so that the `FileNotFound` error will be thrown for the tests. If you think this was a bad idea, please say so, I will then have to use a different approach.

IAlibay:
> Thanks for the reply I've added the tests. Now, do I also have to test for the absence of ` sos_triangle` and `sph_process` executables too? Because that would require the hole executable to be in the codebase somewhere as the error will only be thrown by the absence of these files once hole is found. If that's the case then I have to add the hole binary somewhere in the testsuite and I am not sure whether it will be a good practice or not. Or in a completely different case, my implementation of the tests is wrong and I will need to think of some other approach.

You don't need to include the hole2 binary, this is built and added to the environment directly by our continuous integration workflow. This is why the other tests work fine (for example the `TestHole` or `TestHoleAnalysis` classes).

testsuite/MDAnalysisTests/analysis/test_hole2.py
```diff
@@ -98,6 +98,23 @@ def test_symlink_file(self, tmpdir):
...
+class TestOserror(object):
+    u = mda.Universe(MULTIPDB_HOLE)
+
+    errmsg1 = exe_err.format(name='dummy_path', kw='executable')
+
+    errmsg2 = exe_err.format(name='None', kw='executable')
+
+    def test_hole_oserror(self):
+        with pytest.raises(OSError, match=self.errmsg1):
+            hole = hole2.hole(PDB_HOLE, executable = 'dummy_path')
```
IAlibay: Is there a reason for using different executable paths for these two tests?

testsuite/MDAnalysisTests/analysis/test_hole2.py
```diff
@@ -98,6 +98,23 @@ def test_symlink_file(self, tmpdir):
             assert not fixed.endswith(long_name)
 
 
+class TestOserror(object):
+    u = mda.Universe(MULTIPDB_HOLE)
+
+    errmsg1 = exe_err.format(name='dummy_path', kw='executable')
```
IAlibay: In this case it would be clearer for the error messages to be defined within the individual test functions, there's really no need for these to be defined by the whole class.

testsuite/MDAnalysisTests/analysis/test_hole2.py
```diff
@@ -98,6 +98,23 @@ def test_symlink_file(self, tmpdir):
             assert not fixed.endswith(long_name)
 
 
+class TestOserror(object):
+    u = mda.Universe(MULTIPDB_HOLE)
```
IAlibay: You should use a fixture here rather than defining the universe this way, see lines 238 to 240.

testsuite/MDAnalysisTests/analysis/test_hole2.py
```diff
@@ -98,6 +98,23 @@ def test_symlink_file(self, tmpdir):
...
+
+    errmsg2 = exe_err.format(name='None', kw='executable')
+
+    def test_hole_oserror(self):
+        with pytest.raises(OSError, match=self.errmsg1):
+            hole = hole2.hole(PDB_HOLE, executable = 'dummy_path')
+
+    def test_hole_analysis_oserror(self):
+        with pytest.raises(OSError, match=self.errmsg2):
+            h1 = hole2.HoleAnalysis(self.u, executable=
```
IAlibay: Please do address the PEP8 errors as outlined by: https://github.com/MDAnalysis/mdanalysis/pull/3496#issuecomment-1007433066

IAlibay (changes requested):
Here are some initial review comments.

Atharva7K:
> Here are some initial review comments.

@IAlibay  I really really appreciate you for investing your time in the review, my code involved many malpractices thus I learnt some good ones through your feedback. I have been bothering you a lot lately over some silly issues :).

Now, regarding the tests for `sph_process`: due to the way the `HoleAnalysis` class is written, the code will not throw an `OSError` if I use a dummy path as the code overrides the dummy path and tries to find the executable in the hole2  directory with the correct path and eventually finds the file hence the error is not thrown. So the only way to throw an error for the tests, If I am not wrong, is to actually delete the executable and then run the test.

IAlibay:
> Now, regarding the tests for `sph_process`: due to the way the `HoleAnalysis` class is written, the code will not throw an `OSError` if I use a dummy path as the code overrides the dummy path and tries to find the executable in the hole2 directory with the correct path and eventually finds the file hence the error is not thrown. So the only way to throw an error for the tests, If I am not wrong, is to actually delete the executable and then run the test.

I'm going to ping @lilyminium here but it looks like that's actually a "bug" in the codebase. The behaviour is inconsistent between `sos_triangle` and `sph_process`:

https://github.com/MDAnalysis/mdanalysis/blob/0e630776b0294041d12f8f498add6af6a9d68879/package/MDAnalysis/analysis/hole2/hole.py#L798-L811

Personally I'd change line `807` to use `sph_process` instead of `'sph_process'`, but let's see what @lilyminium thinks.

testsuite/MDAnalysisTests/analysis/test_hole2.py
```diff
@@ -264,6 +264,25 @@ def test_deprecated_warning(self, hole, attrname):
             value = getattr(hole, attrname)
             assert value is hole.results[attrname]
 
+
+class TestOserror(BaseTestHole):
```
IAlibay: I wouldn't subclass `BaseTestHole` here, it adds a set of tests via `test_deprecated_warning` which aren't necessary to re-test here. Instead just add a fixture that does something like this:

```python
@pytest.fixture()
def universe(self):
    return mda.Universe(MULTIPDB_HOLE)
```

IAlibay (changes requested):

lilyminium:
@IAlibay yup, your fix sounds good. I assume `sph_process` is passed in as a kwarg?

Anyone who's successfully installed `hole` should have both executables in `exe`, but always good to be flexible.

IAlibay:
> @IAlibay yup, your fix sounds good. I assume `sph_process` is passed in as a kwarg?

Yeah.

Atharva7K:
Should I update the changelog with the HoleAnalysis bug fix?

lilyminium:
@Atharva7K sure, go ahead, but the Windows tests are failing -- ``test_sos_triangle_oserror`` and ``test_sph_process_oserror``. It looks like they're triggering early, on the actual ``hole`` executable.

testsuite/MDAnalysisTests/analysis/test_hole2.py
```diff
@@ -264,6 +264,34 @@ def test_deprecated_warning(self, hole, attrname):
             value = getattr(hole, attrname)
             assert value is hole.results[attrname]
 
+
+class TestOserror(object):
```
lilyminium: If you make this its own class, please follow PEP8. Python 3+ also doesn't need to directly subclass `object` for a normal class.

Elsewhere in the tests there should be a `pytest` skip condition (`skipif`) that checks if the `hole` executable is found. If not, then the HOLE tests don't run. This isn't triggering tests on Linux so far, but to be safe, could you please add that condition here so that the tests only run if HOLE is found? (It might also be the reason for the failures on Windows).

testsuite/MDAnalysisTests/analysis/test_hole2.py
```diff
@@ -264,6 +264,34 @@ def test_deprecated_warning(self, hole, attrname):
...
+
+    def test_hole_method_oserror(self):
+        errmsg = exe_err.format(name='dummy_path', kw='executable')
+        with pytest.raises(OSError, match=errmsg):
+            hole = hole2.hole(PDB_HOLE, executable='dummy_path')
+
+    def test_hole_oserror(self, universe):
+        errmsg = exe_err.format(name='None', kw='executable')
+        with pytest.raises(OSError, match=errmsg):
+            h = hole2.HoleAnalysis(universe, executable='dummy_path')
```
lilyminium: The actual message here doesn't seem to make much sense (``HOLE binary None not found. None must be on the PATH, or the path must provided with the keyword argument: executable``) -- I think this is yet another sneaky bug! (Sorry....)

Could you please change this so the message actually reads something like ``HOLE binary dummy_path not found. None must be on the PATH, or the path must provided with the keyword argument: executable`` ? Probably lines 792-795:

```
        hole = util.which(executable)
        if hole is None:
            raise OSError(errno.ENOENT, exe_err.format(name=hole,
                                                       kw='executable'))
```

lilyminium (changes requested):

testsuite/MDAnalysisTests/analysis/test_hole2.py
Atharva7K: @lilyminium thanks for the review, I've done all the suggested changes

testsuite/MDAnalysisTests/analysis/test_hole2.py
```diff
@@ -264,6 +264,36 @@ def test_deprecated_warning(self, hole, attrname):
             value = getattr(hole, attrname)
             assert value is hole.results[attrname]
 
+
+@pytest.mark.skipif(executable_not_found('hole'),
+                    reason="Test skipped because HOLE not found")
+class TestOSError():
```
IAlibay:
```suggestion
class TestOSError:
```
Don't need the `()` here.

IAlibay (approved):
I'll approve and directly commit the one suggestion (removing `()` from the class declaration).

@lilyminium can you re-review please?

lilyminium (approved):
Thanks @Atharva7K fixing the original and new bugs found here! I also committed a newline to the test_hole2.py file to fall in line with the PEP8 issues raised in https://github.com/MDAnalysis/mdanalysis/pull/3496#issuecomment-1007433066.

The code all looks good to me!

Atharva7K:
> Thanks @Atharva7K fixing the original and new bugs found here! I also committed a newline to the test_hole2.py file to fall in line with the PEP8 issues raised in [#3496 (comment)](https://github.com/MDAnalysis/mdanalysis/pull/3496#issuecomment-1007433066).
> 
> The code all looks good to me!

My apologies that you had to do the PEP8 fix, I didn't notice the PEP8 issues, won't happen again. Thanks

IAlibay:
Thanks @Atharva7K - squash merging this.

PR #3499, opened 2022-01-06
Chemfiles updates

### LINKED ISSUE #3216: Add CI runner with older versions of chemfiles

### LINKED ISSUE #3492: DeprecationWarning: deprecated distutils Version classes
This is popping up in CI - there's at least two uses of LooseVersion in our codebase that probably needs fixing (chemfiles.py and setup.py).

```
/home/runner/work/mdanalysis/mdanalysis/package/MDAnalysis/coordinates/chemfiles.py:108: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
  MIN_CHEMFILES_VERSION = LooseVersion("0.9")
```

### LINKED ISSUE #3498: Drop support for chemfiles < 0.10
Related to #3490 

Given that we now only support py3.7+ it would make sense to just drop support for old versions of chemfiles and simplify our code paths.

Luthaf:
Fixes #3492
Fixes #3498
Fixes #3216 

 - Use `packaging.version.Version` instead of deprecated `distutils.version.LooseVersion`
- Drop support for chemfiles 0.9

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/coordinates/chemfiles.py
+++ b/package/MDAnalysis/coordinates/chemfiles.py
@@ -79,10 +79,9 @@ Helper functions
 .. autofunction:: check_chemfiles_version
 
 """
-from distutils.version import LooseVersion
 import numpy as np
 import warnings
-import numpy as np
+from packaging.version import Version
 
 from . import base
 
@@ -105,10 +104,10 @@ else:
 
 #: Lowest version of chemfiles that is supported
 #: by MDAnalysis.
-MIN_CHEMFILES_VERSION = LooseVersion("0.9")
+MIN_CHEMFILES_VERSION = Version("0.9")
 #: Lowest version of chemfiles that is *not supported*
 #: by MDAnalysis.
-MAX_CHEMFILES_VERSION = LooseVersion("0.11")
+MAX_CHEMFILES_VERSION = Version("0.11")
 
 
 def check_chemfiles_version():
@@ -126,7 +125,7 @@ def check_chemfiles_version():
             "Try installing with 'pip install chemfiles'"
         )
         return False
-    version = LooseVersion(chemfiles.__version__)
+    version = Version(chemfiles.__version__)
     wrong = version < MIN_CHEMFILES_VERSION or version >= MAX_CHEMFILES_VERSION
     if wrong:
         warnings.warn(
@@ -401,7 +400,7 @@ class ChemfilesWriter(base.WriterBase):
         else:
             lengths = ts.dimensions[:3]
             angles = ts.dimensions[3:]
-            
+
         if chemfiles.__version__.startswith("0.9"):
             frame.cell = chemfiles.UnitCell(*lengths, *angles)
         else:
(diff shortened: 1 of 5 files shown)


package/MDAnalysis/coordinates/chemfiles.py
```diff
@@ -126,7 +125,7 @@ def check_chemfiles_version():
             "Try installing with 'pip install chemfiles'"
         )
         return False
-    version = LooseVersion(chemfiles.__version__)
+    version = Version(chemfiles.__version__)
     wrong = version < MIN_CHEMFILES_VERSION or version >= MAX_CHEMFILES_VERSION
     if wrong:
         warnings.warn(
```
IAlibay: Why do we have both a warning and a subsequent error doing the same thing? isn't it easier/cleaner to just do the error?

testsuite/MDAnalysisTests/coordinates/test_chemfiles.py
```diff
@@ -45,7 +45,7 @@ def test_version_check(self):
         import chemfiles
 
         actual_version = chemfiles.__version__
-        chemfiles.__version__ = "0.8.3"
+        chemfiles.__version__ = "0.9.3"
         assert not check_chemfiles_version()
 
         with pytest.raises(RuntimeError, match="Please install Chemfiles > 0.9"):
```
IAlibay: matches need updating here?

Luthaf:
CI errors seems unrelated, they are exception from the `parse_latex` function of the `citeproc` module

IAlibay (changes requested):
Thanks for working on this @Luthaf.

Can you elaborate on where you are seeing these CI errors? I'm seeing this in the tests "RuntimeError: Please install Chemfiles > 0.10" which seems like somehow the versions aren't getting checked properly?

testsuite/MDAnalysisTests/coordinates/test_chemfiles.py
IAlibay: Ok the failures are specifically due to this test. The test fails early on the first `pytest.raises` so the version doesn't get set back to the correct value.

I have put a PR against your feature branch with a monkeypatch rewrite that should avoid this issue.

package/MDAnalysis/coordinates/chemfiles.py
Luthaf: I think the idea was to be able to send a warning on import (and error on usage), but it is not implemented. I'll change it to only error instead

Luthaf:
> Can you elaborate on where you are seeing these CI errors?

My bad, I clicked on a random failed runner and got what looked like unrelated errors (https://github.com/MDAnalysis/mdanalysis/runs/4727027488?check_suite_focus=true).

IAlibay (approved):
Thanks @Luthaf !

PR #3503, opened 2022-01-16
Unpinning CI dependencies

### LINKED ISSUE #3372: Investigate pins for various CI components (pytest-cov, coverage, msmb_theme)
Follow up from #3369 
Related to #3224 

Our docs build infrastructure is starting to be a bit flaky, particularly in #3369 we've had to pin pytest-cov due to some dependency resolution issues we were having when installing it at the same time as the sphinx components.

We should aim to check a) if all the pins are still necessary, b) what steps we should take to remove these pins.

My understanding is that some of this is related to the fact that we can't upgrade to the latest sphinx (see #3224), so we probably will want to fix that first.

IAlibay:
Fixes #3372

Now that we've fixed sphinx issues, let's see if we can unhardcode CI dependencies - fixing this is a priority to allow folks downstream to properly build develop version of MDA.

 - unpin coverage and pytest-cov

PR Checklist
------------
 - Tests?
 - Docs?
 - CHANGELOG updated?


.coveragerc
```diff
@@ -6,6 +6,7 @@ omit =
     */MDAnalysis/tests/*
     */legacy/*
     */due.py
+concurrency = thread, multiprocessing
```
IAlibay, line 9: Since we use some multiprocessing in our code, we should include this I think (see: https://coverage.readthedocs.io/en/latest/config.html#run-concurrency)

.github/workflows/gh-ci.yaml
```diff
@@ -109,7 +109,7 @@ jobs:
       run: |
         PYTEST_FLAGS="--disable-pytest-warnings --durations=50"
         if [ ${{ matrix.codecov }} = "true" ]; then
-          PYTEST_FLAGS="${PYTEST_FLAGS} --cov=MDAnalysis --cov-report=xml"
+          PYTEST_FLAGS="${PYTEST_FLAGS} --cov-config=.coveragerc --cov=MDAnalysis --cov-report=xml"
```
IAlibay, line 112: So it's not strictly necessary but we have multiple coveragerc files knocking about in the repo (I have another PR open looking to fix that) and I want to make 100% sure we're picking up the right thing.

IAlibay:
This is a reasonable simple one that was giving us headaches with py37, but now that we ditched it, seems like CI is happy

IAlibay:
```
INTERNALERROR> coverage.exceptions.DataError: Conflicting file tracer name for '/home/runner/work/mdanalysis/mdanalysis/package/MDAnalysis/lib/_cutil.pyx': 'Cython.Coverage.Plugin' vs ''
```
this will need a bit more thought I think

.coveragerc
orbeckst: thanks for linking the docs, I agree

.github/workflows/gh-ci.yaml
orbeckst: Better explicit than implicit.

(I was thinking even `./.coveragerc` but that seems too much.)

orbeckst (approved):
LGTM & CI is happy.

PR #3511, opened 2022-01-24
Fix OpenMM parser not working when elements is not set

### LINKED ISSUE #3317: OpenMM parser fails if the elements are not set in the OpenMM topology
## Expected behavior ##

An OpenMM topology created from an input file without elements will not have elements defined. This should not be a problem for the OpenMM parser that would then create a universe without the `elements` topology attribute.

## Actual behavior ##

The OpenMM topology parser fails to create a universe.

``` python
import MDAnalysis as mda
from MDAnalysis.tests.datafiles import PDB  # the file miss the element symbol column
from simtk.openmm import app

openmm_pdb = app.PDBFile(PDB)
u = mda.Universe(openmm_pdb)
```

This raises the following exception:

```
<ipython-input-10-afab9f57734a> in <module>
      4 
      5 openmm_pdb = app.PDBFile(PDB)
----> 6 u = mda.Universe(openmm_pdb)

c:\users\dr19502\dev\mdanalysis\package\MDAnalysis\core\universe.py in __init__(self, topology, all_coordinates, format, topology_format, transformations, guess_bonds, vdwradii, in_memory, in_memory_step, *coordinates, **kwargs)
    336             topology = _topology_from_file_like(self.filename,
    337                                                 topology_format=topology_format,
--> 338                                                 **kwargs)
    339 
    340         if topology is not None:

c:\users\dr19502\dev\mdanalysis\package\MDAnalysis\core\universe.py in _topology_from_file_like(topology_file, topology_format, **kwargs)
    120     try:
    121         with parser(topology_file) as p:
--> 122             topology = p.parse(**kwargs)
    123     except (IOError, OSError) as err:
    124         # There are 2 kinds of errors that might be raised here:

c:\users\dr19502\dev\mdanalysis\package\MDAnalysis\converters\OpenMMParser.py in parse(self, **kwargs)
    183         except AttributeError:
    184             omm_topology = self.filename.topology
--> 185         top = self._mda_topology_from_omm_topology(omm_topology)
    186 
    187         return top

c:\users\dr19502\dev\mdanalysis\package\MDAnalysis\converters\OpenMMParser.py in _mda_topology_from_omm_topology(self, omm_topology)
    110         atomnames = [a.name for a in omm_topology.atoms()]
    111         chainids = [a.residue.chain.id for a in omm_topology.atoms()]
--> 112         elements = [a.element.symbol for a in omm_topology.atoms()]
    113         atomtypes = [a.element.symbol for a in omm_topology.atoms()]
    114         masses = [a.element.mass._value for a in omm_topology.atoms()]

c:\users\dr19502\dev\mdanalysis\package\MDAnalysis\converters\OpenMMParser.py in <listcomp>(.0)
    110         atomnames = [a.name for a in omm_topology.atoms()]
    111         chainids = [a.residue.chain.id for a in omm_topology.atoms()]
--> 112         elements = [a.element.symbol for a in omm_topology.atoms()]
    113         atomtypes = [a.element.symbol for a in omm_topology.atoms()]
    114         masses = [a.element.mass._value for a in omm_topology.atoms()]

AttributeError: 'NoneType' object has no attribute 'symbol'
```

- Which version are you using? '2.0.0-dev0'
- Which version of Python (`python -V`)? '3.7.6 | packaged by conda-forge | (default, Jan  7 2020, 21:48:41) [MSC v.1916 64 bit (AMD64)]'
- Which operating system? Windows

Atharva7K:
Fixes #3317

 Fixed `OpenMMParser` not working when the elements attribute is not set in the OpenMM topology.

PR Checklist
------------
 - [X] Tests?
 - [X] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -75,6 +75,8 @@ Deprecations
  * 2.1.0
 
 Fixes
+  * Fixed OpenMMParser not working when some or none of the elements are set
+    (Issue #3317, PR #3511)
   * Use uint64_t loop counters in C level distance functions to avoid overflow
     for large arrays (Issue #3512, PR #3513).
   * Prevents attempts to close an already closed NamedStream (Issue #3386)
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -16,11 +16,13 @@ The rules for this file:
 ??/??/?? IAlibay, BFedder, inomag, Agorfa, aya9aladdin, shudipto-amin, cbouy,
          HenokB, umak1106, tamandeeps, Mrqeoqqt, megosato, AnirG, rishu235,
          mtiberti, manishsaini6421, Sukeerti1, robotjellyzone, markvrma, alescoulie,
-         mjtadema, PicoCentauri
+         mjtadema, PicoCentauri, Atharva7K
 
  * 2.2.0
 
 Fixes
+  * Fixed `OpenMMTopologyParser` not working when some or none of the elements
+    are present in the openmm topology (Issue #3317, PR #3511).
   * Remove None from filenames in analysis.hole.HoleAnalysis (PR #3650)
   * Fixed Numpy deprecation warnings about elementwise comparisons.
     (Issue #3535, PR #3619)
@@ -57,7 +59,7 @@ Enhancements
     charges.
   * New optimized AnalysisBase derived Watson-Crick distance analysis class in
     analysis.nucleicacids
-  * CRD extended format can now be explicitly requested with the `extended` 
+  * CRD extended format can now be explicitly requested with the `extended`
     keyword (Issue #3605)
 
 Changes
@@ -75,8 +77,6 @@ Deprecations
  * 2.1.0
 
 Fixes
-  * Fixed OpenMMParser not working when some or none of the elements are set
-    (Issue #3317, PR #3511)
   * Use uint64_t loop counters in C level distance functions to avoid overflow
     for large arrays (Issue #3512, PR #3513).
   * Prevents attempts to close an already closed NamedStream (Issue #3386)
--- a/package/MDAnalysis/converters/OpenMMParser.py
+++ b/package/MDAnalysis/converters/OpenMMParser.py
@@ -125,11 +125,12 @@ class OpenMMTopologyParser(TopologyReaderBase):
             atomtypes = [a.element.symbol for a in omm_topology.atoms()]
             masses = [a.element.mass._value for a in omm_topology.atoms()]
 
-        except:
+        except AttributeError:
             attrs = [
                 Atomids(np.array(atomids, dtype=np.int32)),
                 Atomnames(np.array(atomnames, dtype=object)),
-                Bonds(bonds, types=bond_types, order=bond_orders, guessed=False),
+                Bonds(bonds, types=bond_types, order=bond_orders,
+                      guessed=False),
                 ChainIDs(np.array(chainids, dtype=object)),
                 Resids(resids),
                 Resnums(resnums),
@@ -142,7 +143,8 @@ class OpenMMTopologyParser(TopologyReaderBase):
                 Atomids(np.array(atomids, dtype=np.int32)),
                 Atomnames(np.array(atomnames, dtype=object)),
                 Atomtypes(np.array(atomtypes, dtype=object)),
-                Bonds(bonds, types=bond_types, order=bond_orders, guessed=False),
+                Bonds(bonds, types=bond_types, order=bond_orders,
+                      guessed=False),
                 ChainIDs(np.array(chainids, dtype=object)),
                 Elements(np.array(elements, dtype=object)),
                 Masses(np.array(masses, dtype=np.float32)),
--- a/package/MDAnalysis/converters/OpenMMParser.py
+++ b/package/MDAnalysis/converters/OpenMMParser.py
@@ -118,6 +118,9 @@ class OpenMMTopologyParser(TopologyReaderBase):
            * When partial elements are present, values from available elements
              are used whereas the absent elements are assigned an empty string
              with their atomtype set to 'X' and mass set to 0.0.
+           * For abnormal elements (i.e not in SYM2Z) their symbol is
+             used as it is for atomtypes but an empty string is used for
+             elements.
 
         """
         atom_resindex = [a.residue.index for a in omm_topology.atoms()]
@@ -149,10 +152,13 @@ class OpenMMTopologyParser(TopologyReaderBase):
         atomtypes = []
         for a in omm_topology.atoms():
             elem = a.element
-            if elem is not None and elem.symbol.capitalize() in SYMB2Z:
-                validated_elements.append(elem.symbol.capitalize())
-                masses.append(elem.mass._value)
+            if elem is not None:
+                if elem.symbol.capitalize() in SYMB2Z:
+                    validated_elements.append(elem.symbol.capitalize())
+                else:
+                    validated_elements.append('')
                 atomtypes.append(elem.symbol.capitalize())
+                masses.append(elem.mass._value)
             else:
                 wmsg = (f"Unknown element {elem} found for some atoms. "
                         f"These have been given an empty element record "
--- a/package/MDAnalysis/converters/OpenMMParser.py
+++ b/package/MDAnalysis/converters/OpenMMParser.py
@@ -56,8 +56,11 @@ Classes
 """
 
 import numpy as np
+import warnings
 
 from ..topology.base import TopologyReaderBase
+from ..topology.tables import SYMB2Z
+from ..topology.guessers import guess_masses, guess_types
 from ..core.topology import Topology
 from ..core.topologyattrs import (
     Atomids,
@@ -106,7 +109,6 @@ class OpenMMTopologyParser(TopologyReaderBase):
         top : MDAnalysis.core.topology.Topology
 
         """
-
         atom_resindex = [a.residue.index for a in omm_topology.atoms()]
         residue_segindex = [r.chain.index for r in omm_topology.residues()]
         atomids = [a.id for a in omm_topology.atoms()]
@@ -119,53 +121,57 @@ class OpenMMTopologyParser(TopologyReaderBase):
         bonds = [(b.atom1.index, b.atom2.index) for b in omm_topology.bonds()]
         bond_orders = [b.order for b in omm_topology.bonds()]
         bond_types = [b.type for b in omm_topology.bonds()]
+        elements = [None if a.element is None else a.element.symbol
+                    for a in omm_topology.atoms()]
 
-        try:
-            elements = [a.element.symbol for a in omm_topology.atoms()]
-            atomtypes = [a.element.symbol for a in omm_topology.atoms()]
-            masses = [a.element.mass._value for a in omm_topology.atoms()]
+        attrs = [
+            Atomids(np.array(atomids, dtype=np.int32)),
+            Atomnames(np.array(atomnames, dtype=object)),
+            Bonds(bonds, types=bond_types, order=bond_orders, guessed=False),
+            ChainIDs(np.array(chainids, dtype=object)),
+            Resids(resids),
+            Resnums(resnums),
+            Resnames(resnames),
+            Segids(segids),
+        ]
 
-        except AttributeError:
-            attrs = [
-                Atomids(np.array(atomids, dtype=np.int32)),
-                Atomnames(np.array(atomnames, dtype=object)),
-                Bonds(bonds, types=bond_types, order=bond_orders,
-                      guessed=False),
-                ChainIDs(np.array(chainids, dtype=object)),
-                Resids(resids),
-                Resnums(resnums),
-                Resnames(resnames),
-                Segids(segids),
-            ]
-
+        if not any(elements):
+            atomtypes = guess_types(atomnames)
+            attrs.append(Atomtypes(np.array(atomtypes, dtype=object),
+                         guessed=True))
+            warnings.warn("Element information is missing, elements attribute "
+                          "will not be populated. If needed these can be "
+                          "guessed using MDAnalysis.topology.guessers.")
         else:
-            attrs = [
-                Atomids(np.array(atomids, dtype=np.int32)),
-                Atomnames(np.array(atomnames, dtype=object)),
-                Atomtypes(np.array(atomtypes, dtype=object)),
-                Bonds(bonds, types=bond_types, order=bond_orders,
-                      guessed=False),
-                ChainIDs(np.array(chainids, dtype=object)),
-                Elements(np.array(elements, dtype=object)),
-                Masses(np.array(masses, dtype=np.float32)),
-                Resids(resids),
-                Resnums(resnums),
-                Resnames(resnames),
-                Segids(segids),
-            ]
+            validated_elements = []
+            for elem in elements:
+                if elem is not None and elem.capitalize() in SYMB2Z:
+                    validated_elements.append(elem.capitalize())
+                else:
+                    wmsg = (f"Unknown element {elem} found for some atoms. "
+                            f"These have been given an empty element record. "
+                            f"If needed they can be guessed using "
+                            f"MDAnalysis.topology.guessers.")
+                    warnings.warn(wmsg)
+                    validated_elements.append('')
 
-        finally:
-            n_atoms = len(atomids)
-            n_residues = len(resids)
-            n_segments = len(segids)
-            top = Topology(
-                n_atoms,
-                n_residues,
-                n_segments,
-                attrs=attrs,
-                atom_resindex=atom_resindex,
-                residue_segindex=residue_segindex,
-            )
+            attrs.append(Elements(np.array(validated_elements, dtype=object)))
+        atomtypes = [elem for elem in validated_elements]
+        attrs.append(Atomtypes(np.array(atomtypes, dtype=object)))
+        masses = guess_masses(atomtypes)
+        attrs.append(Masses(np.array(masses, dtype=np.float32)))
+
+        n_atoms = len(atomids)
+        n_residues = len(resids)
+        n_segments = len(segids)
+        top = Topology(
+            n_atoms,
+            n_residues,
+            n_segments,
+            attrs=attrs,
+            atom_resindex=atom_resindex,
+            residue_segindex=residue_segindex,
+        )
 
         return top
 
@@ -208,4 +214,3 @@ class OpenMMAppTopologyParser(OpenMMTopologyParser):
         top = self._mda_topology_from_omm_topology(omm_topology)
 
         return top
-
--- a/package/MDAnalysis/converters/OpenMMParser.py
+++ b/package/MDAnalysis/converters/OpenMMParser.py
@@ -76,11 +76,6 @@ from ..core.topologyattrs import (
     Segids,
 )
 
-try:
-    from openmm.unit import daltons
-except ImportError:
-    from simtk.unit import daltons
-
 
 class OpenMMTopologyParser(TopologyReaderBase):
     format = "OPENMMTOPOLOGY"
@@ -127,6 +122,18 @@ class OpenMMTopologyParser(TopologyReaderBase):
              as it is for atomtypes but an empty string is used for elements.
 
         """
+
+        try:
+            from openmm.unit import daltons
+        except ImportError:
+            try:
+                from simtk.unit import daltons
+            except ImportError:
+                msg = ("OpenMM is required for the OpenMMParser but "
+                       "it's not installed. Try installing it with \n"
+                       "conda install -c conda-forge openmm")
+                raise ImportError(msg)
+
         atom_resindex = [a.residue.index for a in omm_topology.atoms()]
         residue_segindex = [r.chain.index for r in omm_topology.residues()]
         atomids = [a.id for a in omm_topology.atoms()]
--- a/package/MDAnalysis/converters/OpenMMParser.py
+++ b/package/MDAnalysis/converters/OpenMMParser.py
@@ -108,18 +108,22 @@ class OpenMMTopologyParser(TopologyReaderBase):
         -------
         top : MDAnalysis.core.topology.Topology
 
+        Note
+        ----
+        When none of the elements are present in the openmm topolgy, their
+        atomtypes are guessed using their names and their masses are
+        then guessed using their atomtypes.
+
+        When partial elements are present, values from available elements
+        are used whereas the absent elements are assigned an empty string
+        with their atomtype set to 'X' and mass set to 0.0.
+
+        For elements with invalid and unreal symbols, the symbol is used
+        as it is for atomtypes but an empty string is used for elements.
 
         .. versionchanged:: 2.2.0
-           * The parser now works when element attribute is missing in some or
-             all the atoms.
-           * When none of the elements are present their atomtypes are guessed
-             using their names and their masses are then guessed using their
-             atomtypes.
-           * When partial elements are present, values from available elements
-             are used whereas the absent elements are assigned an empty string
-             with their atomtype set to 'X' and mass set to 0.0.
-           * For elements with invalid and unreal symbols, the symbol is used
-             as it is for atomtypes but an empty string is used for elements.
+           The parser now works when element attribute is missing in some or
+           all the atoms.
 
         """
 
--- a/package/MDAnalysis/converters/OpenMMParser.py
+++ b/package/MDAnalysis/converters/OpenMMParser.py
@@ -171,7 +171,6 @@ class OpenMMTopologyParser(TopologyReaderBase):
                 atomtypes.append(elem.symbol)
                 masses.append(elem.mass.value_in_unit(daltons))
             else:
-                warn = True
                 validated_elements.append('')
                 masses.append(0.0)
                 atomtypes.append('X')
@@ -180,7 +179,7 @@ class OpenMMTopologyParser(TopologyReaderBase):
             if any(validated_elements):
                 warnings.warn("Element information missing for some atoms. "
                               "These have been given an empty element record ")
-                if any([i == 'X' for i in atomtypes]):
+                if any(i == 'X' for i in atomtypes):
                     warnings.warn("For absent elements, atomtype has been  "
                                   "set to 'X' and mass has been set to 0.0. "
                                   "If needed these can be guessed using "
--- a/package/MDAnalysis/converters/OpenMMParser.py
+++ b/package/MDAnalysis/converters/OpenMMParser.py
@@ -76,6 +76,11 @@ from ..core.topologyattrs import (
     Segids,
 )
 
+try:
+    from openmm.unit import daltons
+except ImportError:
+    from simtk.unit import daltons
+
 
 class OpenMMTopologyParser(TopologyReaderBase):
     format = "OPENMMTOPOLOGY"
@@ -118,9 +123,8 @@ class OpenMMTopologyParser(TopologyReaderBase):
            * When partial elements are present, values from available elements
              are used whereas the absent elements are assigned an empty string
              with their atomtype set to 'X' and mass set to 0.0.
-           * For abnormal elements (i.e not in SYM2Z) their symbol is
-             used as it is for atomtypes but an empty string is used for
-             elements.
+           * For elements with invalid and unreal symbols, the symbol is used
+             as it is for atomtypes but an empty string is used for elements.
 
         """
         atom_resindex = [a.residue.index for a in omm_topology.atoms()]
@@ -150,32 +154,36 @@ class OpenMMTopologyParser(TopologyReaderBase):
         validated_elements = []
         masses = []
         atomtypes = []
+        warn = False
         for a in omm_topology.atoms():
             elem = a.element
             if elem is not None:
                 if elem.symbol.capitalize() in SYMB2Z:
-                    validated_elements.append(elem.symbol.capitalize())
+                    validated_elements.append(elem.symbol)
                 else:
                     validated_elements.append('')
-                atomtypes.append(elem.symbol.capitalize())
-                masses.append(elem.mass._value)
+                atomtypes.append(elem.symbol)
+                masses.append(elem.mass.value_in_unit(daltons))
             else:
-                wmsg = (f"Unknown element {elem} found for some atoms. "
-                        f"These have been given an empty element record "
-                        f"with their atomtype set to 'X' "
-                        f"and their mass set to 0.0. "
-                        f"If needed they can be guessed using "
-                        f"MDAnalysis.topology.guessers.")
-                warnings.warn(wmsg)
+                warn = True
                 validated_elements.append('')
                 masses.append(0.0)
                 atomtypes.append('X')
 
+        if warn:
+            wmsg = (f"Element information missing for some atoms. "
+                    f"These have been given an empty element record "
+                    f"with their atomtype set to 'X' "
+                    f"and their mass set to 0.0. "
+                    f"If needed they can be guessed using "
+                    f"MDAnalysis.topology.guessers.")
+            warnings.warn(wmsg)
+
         if not any(validated_elements):
             atomtypes = guess_types(atomnames)
             masses = guess_masses(atomtypes)
-            warnings.warn("Element information is missing, elements "
-                          "attribute will not be populated. "
+            warnings.warn("Element information is missing for all the atoms. "
+                          "Elements attribute will not be populated. "
                           "Atomtype attribute will be guessed using atom "
                           "name and mass will be guessed using atomtype."
                           "See MDAnalysis.topology.guessers.")
--- a/package/MDAnalysis/converters/OpenMMParser.py
+++ b/package/MDAnalysis/converters/OpenMMParser.py
@@ -161,7 +161,6 @@ class OpenMMTopologyParser(TopologyReaderBase):
         validated_elements = []
         masses = []
         atomtypes = []
-        warn = False
         for a in omm_topology.atoms():
             elem = a.element
             if elem is not None:
@@ -177,26 +176,29 @@ class OpenMMTopologyParser(TopologyReaderBase):
                 masses.append(0.0)
                 atomtypes.append('X')
 
-        if warn:
-            wmsg = (f"Element information missing for some atoms. "
-                    f"These have been given an empty element record "
-                    f"with their atomtype set to 'X' "
-                    f"and their mass set to 0.0. "
-                    f"If needed they can be guessed using "
-                    f"MDAnalysis.topology.guessers.")
-            warnings.warn(wmsg)
+        if not all(validated_elements):
+            if any(validated_elements):
+                warnings.warn("Element information missing for some atoms. "
+                              "These have been given an empty element record ")
+                if any([i == 'X' for i in atomtypes]):
+                    warnings.warn("For absent elements, atomtype has been  "
+                                  "set to 'X' and mass has been set to 0.0. "
+                                  "If needed these can be guessed using "
+                                  "MDAnalysis.topology.guessers.")
+                attrs.append(Elements(np.array(validated_elements,
+                                               dtype=object)))
 
-        if not any(validated_elements):
-            atomtypes = guess_types(atomnames)
-            masses = guess_masses(atomtypes)
-            warnings.warn("Element information is missing for all the atoms. "
-                          "Elements attribute will not be populated. "
-                          "Atomtype attribute will be guessed using atom "
-                          "name and mass will be guessed using atomtype."
-                          "See MDAnalysis.topology.guessers.")
+            else:
+                atomtypes = guess_types(atomnames)
+                masses = guess_masses(atomtypes)
+                wmsg = ("Element information is missing for all the atoms. "
+                        "Elements attribute will not be populated. "
+                        "Atomtype attribute will be guessed using atom "
+                        "name and mass will be guessed using atomtype."
+                        "See MDAnalysis.topology.guessers.")
+                warnings.warn(wmsg)
         else:
             attrs.append(Elements(np.array(validated_elements, dtype=object)))
-
         attrs.append(Atomtypes(np.array(atomtypes, dtype=object)))
         attrs.append(Masses(np.array(masses)))
 
--- a/package/MDAnalysis/converters/OpenMMParser.py
+++ b/package/MDAnalysis/converters/OpenMMParser.py
@@ -176,7 +176,7 @@ class OpenMMTopologyParser(TopologyReaderBase):
 
             attrs.append(Elements(np.array(validated_elements, dtype=object)))
             attrs.append(Masses(np.array(masses)))
-            atomtypes = [elem if len(elem) != 0 else 'X'
+            atomtypes = [elem if elem else 'X'
                          for elem in validated_elements]
             attrs.append(Atomtypes(np.array(atomtypes, dtype=object)))
 
--- a/package/MDAnalysis/converters/OpenMMParser.py
+++ b/package/MDAnalysis/converters/OpenMMParser.py
@@ -112,9 +112,6 @@ class OpenMMTopologyParser(TopologyReaderBase):
         atomids = [a.id for a in omm_topology.atoms()]
         atomnames = [a.name for a in omm_topology.atoms()]
         chainids = [a.residue.chain.id for a in omm_topology.atoms()]
-        elements = [a.element.symbol for a in omm_topology.atoms()]
-        atomtypes = [a.element.symbol for a in omm_topology.atoms()]
-        masses = [a.element.mass._value for a in omm_topology.atoms()]
         resnames = [r.name for r in omm_topology.residues()]
         resids = [r.index + 1 for r in omm_topology.residues()]
         resnums = resids.copy()
@@ -123,32 +120,50 @@ class OpenMMTopologyParser(TopologyReaderBase):
         bond_orders = [b.order for b in omm_topology.bonds()]
         bond_types = [b.type for b in omm_topology.bonds()]
 
-        n_atoms = len(atomids)
-        n_residues = len(resids)
-        n_segments = len(segids)
+        try:
+            elements = [a.element.symbol for a in omm_topology.atoms()]
+            atomtypes = [a.element.symbol for a in omm_topology.atoms()]
+            masses = [a.element.mass._value for a in omm_topology.atoms()]
 
-        attrs = [
-            Atomids(np.array(atomids, dtype=np.int32)),
-            Atomnames(np.array(atomnames, dtype=object)),
-            Atomtypes(np.array(atomtypes, dtype=object)),
-            Bonds(bonds, types=bond_types, order=bond_orders, guessed=False),
-            ChainIDs(np.array(chainids, dtype=object)),
-            Elements(np.array(elements, dtype=object)),
-            Masses(np.array(masses, dtype=np.float32)),
-            Resids(resids),
-            Resnums(resnums),
-            Resnames(resnames),
-            Segids(segids),
-        ]
+        except:
+            attrs = [
+                Atomids(np.array(atomids, dtype=np.int32)),
+                Atomnames(np.array(atomnames, dtype=object)),
+                Bonds(bonds, types=bond_types, order=bond_orders, guessed=False),
+                ChainIDs(np.array(chainids, dtype=object)),
+                Resids(resids),
+                Resnums(resnums),
+                Resnames(resnames),
+                Segids(segids),
+            ]
 
-        top = Topology(
-            n_atoms,
-            n_residues,
-            n_segments,
-            attrs=attrs,
-            atom_resindex=atom_resindex,
-            residue_segindex=residue_segindex,
-        )
+        else:
+            attrs = [
+                Atomids(np.array(atomids, dtype=np.int32)),
+                Atomnames(np.array(atomnames, dtype=object)),
+                Atomtypes(np.array(atomtypes, dtype=object)),
+                Bonds(bonds, types=bond_types, order=bond_orders, guessed=False),
+                ChainIDs(np.array(chainids, dtype=object)),
+                Elements(np.array(elements, dtype=object)),
+                Masses(np.array(masses, dtype=np.float32)),
+                Resids(resids),
+                Resnums(resnums),
+                Resnames(resnames),
+                Segids(segids),
+            ]
+
+        finally:
+            n_atoms = len(atomids)
+            n_residues = len(resids)
+            n_segments = len(segids)
+            top = Topology(
+                n_atoms,
+                n_residues,
+                n_segments,
+                attrs=attrs,
+                atom_resindex=atom_resindex,
+                residue_segindex=residue_segindex,
+            )
 
         return top
 
--- a/package/MDAnalysis/converters/OpenMMParser.py
+++ b/package/MDAnalysis/converters/OpenMMParser.py
@@ -60,7 +60,7 @@ import warnings
 
 from ..topology.base import TopologyReaderBase
 from ..topology.tables import SYMB2Z
-from ..topology.guessers import guess_masses, guess_types
+from ..topology.guessers import guess_types, guess_masses
 from ..core.topology import Topology
 from ..core.topologyattrs import (
     Atomids,
@@ -109,14 +109,15 @@ class OpenMMTopologyParser(TopologyReaderBase):
         top : MDAnalysis.core.topology.Topology
 
 
-        .. versionchanged:: 2.1.0
+        .. versionchanged:: 2.2.0
            * The parser now works when element attribute is missing in some or
              all the atoms.
-           * When none of the elements are present their atomtypes are guessed using
-             their names and their masses are then guessed using their atomtypes.
-           * When partial elements are present, values from available elements are used
-             whereas the absent elements are assigned an empty string with their
-             atomtype set to 'X' and mass set to 0.0.
+           * When none of the elements are present their atomtypes are guessed
+             using their names and their masses are then guessed using their
+             atomtypes.
+           * When partial elements are present, values from available elements
+             are used whereas the absent elements are assigned an empty string
+             with their atomtype set to 'X' and mass set to 0.0.
 
         """
         atom_resindex = [a.residue.index for a in omm_topology.atoms()]
@@ -131,8 +132,6 @@ class OpenMMTopologyParser(TopologyReaderBase):
         bonds = [(b.atom1.index, b.atom2.index) for b in omm_topology.bonds()]
         bond_orders = [b.order for b in omm_topology.bonds()]
         bond_types = [b.type for b in omm_topology.bonds()]
-        elements = [None if a.element is None else a.element.symbol
-                    for a in omm_topology.atoms()]
 
         attrs = [
             Atomids(np.array(atomids, dtype=np.int32)),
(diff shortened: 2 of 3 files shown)


package/MDAnalysis/converters/OpenMMParser.py
```diff
@@ -120,32 +117,52 @@ def _mda_topology_from_omm_topology(self, omm_topology):
...
-
-        top = Topology(
-            n_atoms,
-            n_residues,
-            n_segments,
-            attrs=attrs,
-            atom_resindex=atom_resindex,
-            residue_segindex=residue_segindex,
-        )
+        try:
```
IAlibay: Any reason for using a `try/except` here instead of an `if/else` based on the presence of `element` as an attribute? Would the latter not be more readable / lead to fewer duplicated lines?

lilyminium: I would also wonder if it's possible for only some elements in the `omm_topology` to be present, much like a PDB file. If that's the case, I'd want my Universe to behave the same as from a PDB file: contain specified elements, with empty strings for the others.

Link to the relevant PDB code: https://github.com/MDAnalysis/mdanalysis/blob/d593993eb5b3ab06ecc3562cb7572605544f77d0/package/MDAnalysis/topology/PDBParser.py#L311-L336

Atharva7K:
My apologies if the code is not upto the mark :)

IAlibay:
Thanks for working on this @Atharva7K - before review though, could you add some tests please?

Atharva7K:
@IAlibay the codecov checks are taking longer than usual ,I am not sure whether something is wrong or not but thought I should let you know.

package/MDAnalysis/converters/OpenMMParser.py
```diff
@@ -103,41 +106,61 @@ def _mda_topology_from_omm_topology(self, omm_topology):
...
             ChainIDs(np.array(chainids, dtype=object)),
-            Elements(np.array(elements, dtype=object)),
-            Masses(np.array(masses, dtype=np.float32)),
             Resids(resids),
             Resnums(resnums),
             Resnames(resnames),
             Segids(segids),
         ]
 
+        if not any(elements):
```
IAlibay, line 135: As far as I can tell you don't have with a system that doesn't have elements so this codepath is never triggered.

package/MDAnalysis/converters/OpenMMParser.py
```diff
@@ -103,41 +106,61 @@ def _mda_topology_from_omm_topology(self, omm_topology):
...
+                         guessed=True))
+            warnings.warn("Element information is missing, elements attribute "
+                          "will not be populated. If needed these can be "
+                          "guessed using MDAnalysis.topology.guessers.")
+        else:
+            validated_elements = []
+            for elem in elements:
+                if elem is not None and elem.capitalize() in SYMB2Z:
+                    validated_elements.append(elem.capitalize())
+                else:
```
IAlibay, line 147: Similarly, you don't have a test with a case with only partial elements

testsuite/MDAnalysisTests/converters/test_openmm_parser.py
```diff
@@ -96,6 +96,25 @@ def test_segids(self, top):
         else:
             assert top.segids.values == []
 
+    def test_elements(self, top):
+        if hasattr(top, 'elements'):
```
IAlibay: rather than looking for the attributes of the topology, it would be safer (and probably a better test) to look at whether or not "elements" is in "expected_attrs".

Also please note that adding these only tests the present of elements, you should also be adding tests for when there aren't elements (or partial elements, if that's something we want to support).

package/MDAnalysis/converters/OpenMMParser.py
```diff
@@ -103,41 +106,61 @@ def _mda_topology_from_omm_topology(self, omm_topology):
...
+                            f"These have been given an empty element record. "
+                            f"If needed they can be guessed using "
+                            f"MDAnalysis.topology.guessers.")
+                    warnings.warn(wmsg)
+                    validated_elements.append('')
+
+            attrs.append(Elements(np.array(validated_elements, dtype=object)))
+        atomtypes = [elem for elem in validated_elements]
+        attrs.append(Atomtypes(np.array(atomtypes, dtype=object)))
+        masses = guess_masses(atomtypes)
```
IAlibay: Mass guessing is very expensive and also can be quite inaccurate. Why make the change from reading the `atom.element.mass._value` ?

package/MDAnalysis/converters/OpenMMParser.py
```diff
@@ -103,41 +106,61 @@ def _mda_topology_from_omm_topology(self, omm_topology):
...
+                else:
+                    wmsg = (f"Unknown element {elem} found for some atoms. "
+                            f"These have been given an empty element record. "
+                            f"If needed they can be guessed using "
+                            f"MDAnalysis.topology.guessers.")
+                    warnings.warn(wmsg)
+                    validated_elements.append('')
+
+            attrs.append(Elements(np.array(validated_elements, dtype=object)))
+        atomtypes = [elem for elem in validated_elements]
```
IAlibay: What happens here if `validated_elements` isn't assigned?

IAlibay (changes requested):
> @IAlibay the codecov checks are taking longer than usual ,I am not sure whether something is wrong or not but thought I should let you know.

This can happen sometimes, not too sure what's caused it.

In any case the codecov report above is correctly behaving: https://github.com/MDAnalysis/mdanalysis/pull/3511#issuecomment-1020281198

Your tests aren't sufficiently covering the changes made here.

Please see below for some initial comments.

Atharva7K:
Thanks for the review @IAlibay  I'll do the changes, the codecov checks are still running, I think you should stop the if you can.

IAlibay:
> Thanks for the review @IAlibay I'll do the changes, the codecov checks are still running, I think you should stop the if you can.

It's not something you can just stop afaik (unless I've forgotten some bot action). It doesn't matter anyways it'll reset itself once a new commit is made.

Atharva7K:
@IAlibay  I am having issues with the tests for partial and no elements. For the tests to work, I have to  modify the values of the elements attribute in an openmm topology by setting  some/all of the elements to None and then call the `_mda_topology_from_omm_topology` method with this topology. But since this method can be only called through the `parse` method, it's not possible to modify the topology in `parse` and then use it for the tests. If you can suggest some hints/workaround to solve this issue, that'd be great.

IAlibay:
> @IAlibay I am having issues with the tests for partial and no elements. For the tests to work, I have to modify the values of the elements attribute in an openmm topology by setting some/all of the elements to None and then call the `_mda_topology_from_omm_topology` method with this topology. But since this method can be only called through the `parse` method, it's not possible to modify the topology in `parse` and then use it for the tests. If you can suggest some hints/workaround to solve this issue, that'd be great.

It's hard for me to work out what you have tried based on this.

Have you attempted to replicate the use case highlighted in the original issue? That should at least cover the `partial elements` use case right?

Then you _should_ be able to modify the openmm topology in-place to get a `no elements` case? (i.e. loop over the atoms in the openmm topology and just assign atom.element to `None`)

Atharva7K:
> > @IAlibay I am having issues with the tests for partial and no elements. For the tests to work, I have to modify the values of the elements attribute in an openmm topology by setting some/all of the elements to None and then call the `_mda_topology_from_omm_topology` method with this topology. But since this method can be only called through the `parse` method, it's not possible to modify the topology in `parse` and then use it for the tests. If you can suggest some hints/workaround to solve this issue, that'd be great.
> 
> It's hard for me to work out what you have tried based on this.
> 
> Have you attempted to replicate the use case highlighted in the original issue? That should at least cover the `partial elements` use case right?
> 
> Then you _should_ be able to modify the openmm topology in-place to get a `no elements` case? (i.e. loop over the atoms in the openmm topology and just assign atom.element to `None`)

Thanks for the reply @IAlibay I've made the changes, one check is failing through and I'm not sure what the check does and what's the reason for failure. Can you please have a look  so that I can address it.

IAlibay:
@Atharva7K sorry about the delay in reviewing, the core team is quite busy at the moment.

Just before we carry on, I realise you are now a second time contributor, but MDAnalysis is currently in the process of evaluating a potential change in its license. As part of this we are asking that all new-ish contributors introduce themselves (with an email address that they regularly monitor) to the [MDAnalysis developer list](https://groups.google.com/group/mdnalysis-devel) before we accept their contributions. Please include your github handle and a link to this pull request. The aim is that we then have a means of contacting the developers should we decide to change our license in the near future.

Since you mentioned being interested in taking part in GSoC 2022, please note that communicating with the developers list is highly encouraged. [Our FAQ contains more information](https://github.com/MDAnalysis/mdanalysis/wiki/GSoC-FAQ#do-you-want-to-see-my-work-and-learn-more-about-me). Please note that the FAQ has not been updated since 2021 and its contents are not yet considered final until the GSoC application window opens on the 4th of April.

Atharva7K:
> @Atharva7K sorry about the delay in reviewing, the core team is quite busy at the moment.
> 
> Just before we carry on, I realise you are now a second time contributor, but MDAnalysis is currently in the process of evaluating a potential change in its license. As part of this we are asking that all new-ish contributors introduce themselves (with an email address that they regularly monitor) to the [MDAnalysis developer list](https://groups.google.com/group/mdnalysis-devel) before we accept their contributions. Please include your github handle and a link to this pull request. The aim is that we then have a means of contacting the developers should we decide to change our license in the near future.
> 
> Since you mentioned being interested in taking part in GSoC 2022, please note that communicating with the developers list is highly encouraged. [Our FAQ contains more information](https://github.com/MDAnalysis/mdanalysis/wiki/GSoC-FAQ#do-you-want-to-see-my-work-and-learn-more-about-me). Please note that the FAQ has not been updated since 2021 and its contents are not yet considered final until the GSoC application window opens on the 4th of April.

Thanks for the information @IAlibay, I've sent an introductory email and I've gone through the FAQ. I completely understand you are busy, we can continue with this review when you get time :)

jbarnoud (changes requested):
This looks good. Thank you. There are only minor things to fix before this PR can be merged:

* [x] I see the CHANGELOG ticked in the first message of the PR, but I do not see the file updated
* [ ] The new behaviour should be documented in the module's docstring
* [x] The change in behaviour should be documented with a `.. versionchanged::` note in the docstring of `_mda_topology_from_omm_topology`
* [x] Fix the merge conflict

Beyond that, I think it is all good :)

package/MDAnalysis/converters/OpenMMParser.py
```diff
@@ -103,41 +106,66 @@ def _mda_topology_from_omm_topology(self, omm_topology):
...
+                            f"If needed they can be guessed using "
+                            f"MDAnalysis.topology.guessers.")
+                    warnings.warn(wmsg)
+                    validated_elements.append('')
+                    masses.append(0.0)
+
+            attrs.append(Elements(np.array(validated_elements, dtype=object)))
+            attrs.append(Masses(np.array(masses)))
+            atomtypes = [elem for elem in validated_elements]
+            attrs.append(Atomtypes(np.array(atomtypes, dtype=object)))
```
IAlibay: The behavior for masses and atomtypes is very different based on these two cases. Can this be documented in the docstring of this class (as per @jbarnoud's request) and possibly clarify a little bit in the warning message?

IAlibay: Currently in discussions with @jbarnoud but maybe we shouldn't be setting `atomtype` records to empty strings here. Since atomtypes are kind of a require attribute (for now), it's might make more sense to replace the empty positions with something like atomname or just any selectable character?

Thoughts @MDAnalysis/coredevs ?

jbarnoud: I would rather use a common type for all the non-assigned atoms. This way a user can select all of them at once. I would happily suggest going non-ascii here with something like ❌ but that might be controversial :stuck_out_tongue:

IAlibay: I'd be happy to go with the more traditional `X` 🙃

Atharva7K: > I would rather use a common type for all the non-assigned atoms. This way a user can select all of them at once. I would happily suggest going non-ascii here with something like x but that might be controversial stuck_out_tongue

I'd like the controversy :)  but settled for conventional `X` upon @IAlibay's request. Anyway thanks for the review, I've done the changes.

Quick question: afaik the PDB parser too uses empty string for absent elements. Should i make the changes there too?

jbarnoud: > Quick question: afaik the PDB parser too uses empty string for absent elements. Should i make the changes there too?

The disprepency is annoying, indeed, but elements and types are different for various reasons. Changing how elements behave is likely to break downstream code so we will avoid doing it.

IAlibay: I think there's a misunderstanding here - default elements should be '' and default atomtypes should be X.

The choice for '' in elements was the result of long discussions on the matter so I'd prefer not restart those at the moment.

Atharva7K: Apologies for the misunderstanding, I've made the corrections

> I think there's a misunderstanding here - default elements should be '' and default atomtypes should be X.
> 
> The choice for '' in elements was the result of long discussions on the matter so I'd prefer not restart those at the moment.

Apologies for the misunderstanding, I've made the corrections

Atharva7K:
Hey @IAlibay, you've been reviewing my code for more than a month now or should I say cleaning up my mess :). Any suggestions for improvement or things that you find annoying or anything else in general? Thanks

lilyminium:
Hi @Atharva7K, you're doing great, it's always challenging figuring out a new library (or two, including OpenMM) or entering a new domain like molecular dynamics. Getting a feel for things like how to write tests and working with Python objects in general are just things you pick up with experience. Contributing to open source is a pretty good way to get experience! :-)

testsuite/MDAnalysisTests/converters/test_openmm_parser.py
```diff
@@ -130,6 +150,44 @@ class TestOpenMMTopologyParser(OpenMMTopologyBase):
...
+                    "with their atomtype set to 'X' "
+                    "and their mass set to 0.0. "
+                    "If needed they can be guessed using "
+                    "MDAnalysis.topology.guessers.")
+            with pytest.warns(UserWarning, match=wmsg):
+                mda_top = self.parser(PDB) \
+                          ._mda_topology_from_omm_topology(omm_topology)
+
+    def test_with_no_elements(self, top):
+        self.expected_attrs.remove('elements')
```
lilyminium: As `expected_attrs` is a class attribute, I think this will remove "elements" from the base class and cause potential trouble for any further subclasses of OpenMMTopologyBase. Could you please just create re-define `expected_attrs`? I think that also means that another check will run that the resulting Universe does *not* have elements.

testsuite/MDAnalysisTests/converters/test_openmm_parser.py
```diff
@@ -130,6 +150,44 @@ class TestOpenMMTopologyParser(OpenMMTopologyBase):
...
+        wmsg = ("Element information is missing, elements attribute "
+                "will not be populated. "
+                "Atomtype attribute will be guessed using atom "
+                "name and mass will be guessed using atomtype."
+                "See MDAnalysis.topology.guessers."
+                )
+        with pytest.warns(UserWarning, match=wmsg):
+            mda_top = self.parser(PDB) \
+                      ._mda_topology_from_omm_topology(omm_topology)
+        self.expected_attrs.append('elements')
```
lilyminium: This is probably not necessary.

testsuite/MDAnalysisTests/converters/test_openmm_parser.py
```diff
@@ -130,6 +150,44 @@ class TestOpenMMTopologyParser(OpenMMTopologyBase):
...
+            omm_topology = self.ref_filename
+            wmsg = ("Unknown element None found for some atoms. "
+                    "These have been given an empty element record "
+                    "with their atomtype set to 'X' "
+                    "and their mass set to 0.0. "
+                    "If needed they can be guessed using "
+                    "MDAnalysis.topology.guessers.")
+            with pytest.warns(UserWarning, match=wmsg):
+                mda_top = self.parser(PDB) \
+                          ._mda_topology_from_omm_topology(omm_topology)
```
lilyminium: Could you please make this a separate class to the test for no elements, and then actually check a couple output values? e.g. if `mda_top.elements[9] == "X"`.

testsuite/MDAnalysisTests/converters/test_openmm_parser.py
```diff
@@ -101,6 +102,25 @@ def test_segids(self, top):
         else:
             assert top.segids.values == []
 
+    def test_elements(self, top):
+        if 'elements' in self.expected_attrs:
+            assert len(top.elements.values) == self.expected_n_atoms
+            assert isinstance(top.elements.values, np.ndarray)
+            assert all(isinstance(elem, str) for elem in top.elements.values)
```
lilyminium, line 113: Hmm, I suppose I'm not sure how much this kind of test adds. The resname/resnum/segids test if there's an empty list or an array; a corresponding `else` here would be helpful too. Especially for the some/no elements subclasses below.

package/MDAnalysis/converters/OpenMMParser.py
```diff
@@ -102,42 +105,81 @@ def _mda_topology_from_omm_topology(self, omm_topology):
...
         segids = [str(c.index) for c in omm_topology.chains()]
         bonds = [(b.atom1.index, b.atom2.index) for b in omm_topology.bonds()]
         bond_orders = [b.order for b in omm_topology.bonds()]
         bond_types = [b.type for b in omm_topology.bonds()]
-
-        n_atoms = len(atomids)
-        n_residues = len(resids)
-        n_segments = len(segids)
+        elements = [None if a.element is None else a.element.symbol
+                    for a in omm_topology.atoms()]
```
lilyminium: You don't really use the contents of this list. Could you reorder this to:

* construct a validated_elements list and masses at the same time
* Check the values and give the appropriate warning if needed
* Guess/construct the atomtypes?

package/MDAnalysis/converters/OpenMMParser.py
```diff
@@ -102,42 +105,81 @@ def _mda_topology_from_omm_topology(self, omm_topology):
...
+                            f"and their mass set to 0.0. "
+                            f"If needed they can be guessed using "
+                            f"MDAnalysis.topology.guessers.")
+                    warnings.warn(wmsg)
+                    validated_elements.append('')
+                    masses.append(0.0)
+
+            attrs.append(Elements(np.array(validated_elements, dtype=object)))
+            attrs.append(Masses(np.array(masses)))
+            atomtypes = [elem if len(elem) != 0 else 'X'
```
lilyminium: One of my favourite things about Python is [truthiness](https://www.freecodecamp.org/news/truthy-and-falsy-values-in-python/)! `if len(elem) != 0` and `if elem` accomplish pretty much the same thing.

lilyminium (changes requested):

Atharva7K:
Really appreciate the detailed review. I'm busy with exams right now so I'm gonna need some time to work on this :).

Atharva7K:
I'm almost done with the changes though when I fetched upstream code and did a rebase and tried to run the tests I'm getting the following error:
```
ImportError while importing test module '/home/ubuntu/mdanalysis/testsuite/MDAnalysisTests/converters/test_openmm_parser.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../../../../anaconda3/envs/mdanalysis-dev/lib/python3.8/importlib/__init__.py:127: in import_module
test_openmm_parser.py:26: in <module>
    import MDAnalysis as mda
../../../package/MDAnalysis/__init__.py:191: in <module>
    from .lib import log
../../../package/MDAnalysis/lib/__init__.py:37: in <module>
    from . import distances  # distances relies on mdamath
../../../package/MDAnalysis/lib/distances.py:108: in <module>
    from .c_distances import (_UINT64_MAX,
E   ImportError: cannot import name '_UINT64_MAX' from 'MDAnalysis.lib.c_distances' (/home/ubuntu/mdanalysis/package/MDAnalysis/lib/c_distances.cpython-38-x86_64-linux-gnu.so)

```
I'm not completely sure but this may have something to do with @orbeckst 's recent [commit](https://github.com/MDAnalysis/mdanalysis/pull/3511/commits/3d0896ca15baba936ddffdcf663cc62b41673743). I want to make some final changes before I push but It's not possible if I cannot run the tests. Can someone please help me with this? Thanks

orbeckst:
I think you need to [re-install the development version](https://userguide.mdanalysis.org/stable/contributing_code.html#build-mdanalysis-develop) because the merge changed some of the C-code. Try 
```
pip install -e package/
```
from the top of the checked-out repository.

testsuite/MDAnalysisTests/converters/test_openmm_parser.py
Atharva7K: I'm not able to figure out what the corresponding else should have, since elements is not expected as an attribute, what am I supposed to test? Do you mean to say I should test if there's an empty list or array like the tests you mentioned. Also, I modified the tests for types and masses to check for an empty array like resname/resnum etc, I hope it's fine.

lilyminium: What if the OpenMM input has no elements?

package/MDAnalysis/converters/OpenMMParser.py
Atharva7K: The list is actually being used to test whether all the elements are set or not, if some of the elements are present the code is pretty much already doing the things you requested. Do you want me to make further modifications please let me know

testsuite/MDAnalysisTests/converters/test_openmm_parser.py
Atharva7K: > What if the OpenMM input has no elements?

Umm in that case the the parsed topology too shouldn't have elements, is that right?

lilyminium: Yes -- and it would be good to have a test for this :)

Atharva7K: Okay this was so obvious :), anyway thanks for the hint. Will add the tests

Atharva7K:
Hmm all the tests are running fine locally but they're failing in the checks this might have something to do with the order in which the tests are run in the checks because the elements attribute is present till the no elements test is run so the unexpected attribute test fails . I think I have to redefine the `top` fixture with no elements in the subclass to make sure elements is not set for all the tests. Is this a good idea?

jbarnoud:
The failing tests look unrelated to the PR. There seem to be something wrong about file system stuff on windows?

Atharva7K:
> The failing tests look unrelated to the PR. There seem to be something wrong about file system stuff on windows?

Wondering the same but not sure :)

IAlibay:
since it was specific to the one runner I've restarted CI, let's see what happens.

package/MDAnalysis/converters/OpenMMParser.py
```diff
@@ -105,42 +108,81 @@ def _mda_topology_from_omm_topology(self, omm_topology):
         -------
         top : MDAnalysis.core.topology.Topology
 
-        """
 
+        .. versionchanged:: 2.1.0
```
IAlibay:
```suggestion
        .. versionchanged:: 2.2.0
```

package/MDAnalysis/converters/OpenMMParser.py
```diff
@@ -105,42 +108,81 @@ def _mda_topology_from_omm_topology(self, omm_topology):
...
-            Masses(np.array(masses, dtype=np.float32)),
             Resids(resids),
             Resnums(resnums),
             Resnames(resnames),
             Segids(segids),
         ]
 
+        if not any(elements):
+            atomtypes = guess_types(atomnames)
+            attrs.append(Atomtypes(np.array(atomtypes, dtype=object)))
```
IAlibay: There's no point duplicating lines here and in the `else` block, I would move the appending of Atomtypes and Masses until after the `if else` block.

package/MDAnalysis/converters/OpenMMParser.py
```diff
@@ -105,42 +108,81 @@ def _mda_topology_from_omm_topology(self, omm_topology):
...
+            attrs.append(Masses(np.array(masses)))
+            warnings.warn("Element information is missing, elements "
+                          "attribute will not be populated. "
+                          "Atomtype attribute will be guessed using atom "
+                          "name and mass will be guessed using atomtype."
+                          "See MDAnalysis.topology.guessers.")
+        else:
+            validated_elements = []
+            masses = []
+            for a in omm_topology.atoms():
```
IAlibay: Looping over this twice seems rather inefficient, is there anything we can do to avoid this?

As per Liily's comment https://github.com/MDAnalysis/mdanalysis/pull/3511/files#r810629598

You should be able to re-order this so that you loop over the atoms once, pick up validated elements, and then if all the elements are bad, guess the aotmtypes & masses.

package/MDAnalysis/converters/OpenMMParser.py
```diff
@@ -105,42 +108,81 @@ def _mda_topology_from_omm_topology(self, omm_topology):
...
+                elem = a.element
+                if elem is not None and elem.symbol.capitalize() in SYMB2Z:
```
IAlibay: Do we have a case for thinking that OpenMM will ever had non-normal elements? (i.e. those that won't fit in SYMB2Z) the element attribute is not None? If not, then we don't need to check inside of SYMB2Z?

testsuite/MDAnalysisTests/converters/test_openmm_parser.py
```diff
@@ -130,6 +155,72 @@ class TestOpenMMTopologyParser(OpenMMTopologyBase):
...
+        "chainIDs",
+    ]
+
+    @pytest.fixture()
+    def top(self, filename):
+        omm_topology = self.ref_filename
+        for a in omm_topology.atoms():
+            a.element = None
+        return self.parser(PDB) \
+                   ._mda_topology_from_omm_topology(omm_topology)
```
IAlibay: This line continuation breaks is more of a PEP8 break than having a it on a single line.

Additionally do you really need to redefine `top` this way? Couldn't you just do:

```python
ref_filename = app.PDBFile(PDB).topology
for a in ref_filename.atoms():
    a.element = None
```
Replacing line 159?

testsuite/MDAnalysisTests/converters/test_openmm_parser.py
```diff
@@ -130,6 +155,72 @@ class TestOpenMMTopologyParser(OpenMMTopologyBase):
...
+
+        wmsg = ("Element information is missing, elements attribute "
+                "will not be populated. "
+                "Atomtype attribute will be guessed using atom "
+                "name and mass will be guessed using atomtype."
+                "See MDAnalysis.topology.guessers."
+                )
+        with pytest.warns(UserWarning, match=wmsg):
+            mda_top = self.parser(PDB) \
+                      ._mda_topology_from_omm_topology(omm_topology)
```
IAlibay: I don't think this is necessary either, you should be able to directly parse the modified topology (proposed in my previous comment). Also the name of this test should be better. Would this work?

```suggestion
    def test_no_elements_warn(self, top):
        wmsg = ("Element information is missing, elements attribute "
                "will not be populated. "
                "Atomtype attribute will be guessed using atom "
                "name and mass will be guessed using atomtype."
                "See MDAnalysis.topology.guessers.")

        with pytest.warns(UserWarning, match=wmsg):
            self.parser(self.filename)
```

testsuite/MDAnalysisTests/converters/test_openmm_parser.py
```diff
@@ -130,6 +155,72 @@ class TestOpenMMTopologyParser(OpenMMTopologyBase):
...
+    def test_with_partial_elements(self, top):
+        if 'elements' in self.expected_attrs:
+            omm_topology = self.ref_filename
+            wmsg = ("Unknown element None found for some atoms. "
+                    "These have been given an empty element record "
+                    "with their atomtype set to 'X' "
+                    "and their mass set to 0.0. "
+                    "If needed they can be guessed using "
+                    "MDAnalysis.topology.guessers.")
+            with pytest.warns(UserWarning, match=wmsg):
```
IAlibay: see my previous comment regarding calling parser this way.

package/CHANGELOG
```diff
@@ -33,10 +33,12 @@ Deprecations
  * 2.1.0
 
 Fixes
+  * Fixed OpenMMParser not working when some or none of the elements are set
```
IAlibay: Please move to the 2.2.0 section

IAlibay (changes requested):

package/MDAnalysis/converters/OpenMMParser.py
lilyminium: > The list is actually being used to test whether all the elements are set or not

Yeah, I just think that this list does not need to be constructed and you could work the check into the remainder of the code. I suggested an outline above. It's not that the code is *wrong*, just that redundant code always adds maintenance burden.

package/MDAnalysis/converters/OpenMMParser.py
lilyminium: I think you could.

```
>>> from openmm.app.element import Element
>>> Element(0, "*", "*", 0)
<Element *>
```

That being said, I would prefer not to lose any weird element information I had. I guess it would go into `Atomtype` and not `Element` if we want to ensure all our elements are valid.

lilyminium (changes requested):

testsuite/MDAnalysisTests/converters/test_openmm_parser.py
Atharva7K: Calling the parser that way doesn't raise any warnings for testing so I've replaced the call with a more workable one

package/MDAnalysis/converters/OpenMMParser.py
```diff
@@ -123,24 +133,55 @@ def _mda_topology_from_omm_topology(self, omm_topology):
...
+                wmsg = (f"Unknown element {elem} found for some atoms. "
+                        f"These have been given an empty element record "
+                        f"with their atomtype set to 'X' "
+                        f"and their mass set to 0.0. "
+                        f"If needed they can be guessed using "
+                        f"MDAnalysis.topology.guessers.")
+                warnings.warn(wmsg)
+                validated_elements.append('')
+                masses.append(0.0)
+                atomtypes.append('X')
```
lilyminium: I know we said to use an "X" in AtomTypes for empty elements, but could you please actually use the elem.symbol in Atomtypes? There could be weird elements in there and it would be best to keep that information. A relevant test would probably require manually constructing a small OpenMM topology with manually defined elements like below:

```
>>> from openmm.app.element import Element
>>> Element(0, "*", "*", 0)
<Element *>

```

package/MDAnalysis/converters/OpenMMParser.py
```diff
@@ -123,24 +133,55 @@ def _mda_topology_from_omm_topology(self, omm_topology):
...
+        validated_elements = []
+        masses = []
+        atomtypes = []
+        for a in omm_topology.atoms():
+            elem = a.element
+            if elem is not None and elem.symbol.capitalize() in SYMB2Z:
+                validated_elements.append(elem.symbol.capitalize())
+                masses.append(elem.mass._value)
+                atomtypes.append(elem.symbol.capitalize())
+            else:
```
lilyminium: Nice and efficient restructuring, thank you!

lilyminium (changes requested):

package/MDAnalysis/converters/OpenMMParser.py
Atharva7K: Sorry I've been offline for many days, was busy with exams. 
re - using elem.symbol, for this case what if the elem itself is None? I've changed my code to accept weird types that are not in SYMB2Z but when the elem is None, should we stick to using 'X', should we guess it or do something else?

lilyminium: I think the logic flow should probably go like this to be consistent with other parsers and not lose information -- (@IAlibay you're probably more in touch with this than I am, any comments or emoji reacts?):

- if element is present:
  - if element is valid, use that symbol for both AtomTypes and Elements
  - if element is invalid, use the symbol for AtomType but have an empty '' for Element
- if element is absent:
  - use 'X' for AtomType and empty '' for Element

IAlibay: Yeah that sounds reasonable

package/MDAnalysis/converters/OpenMMParser.py
```diff
@@ -105,16 +108,26 @@ def _mda_topology_from_omm_topology(self, omm_topology):
...
+             all the atoms.
+           * When none of the elements are present their atomtypes are guessed
+             using their names and their masses are then guessed using their
+             atomtypes.
+           * When partial elements are present, values from available elements
+             are used whereas the absent elements are assigned an empty string
+             with their atomtype set to 'X' and mass set to 0.0.
+           * For abnormal elements (i.e not in SYM2Z) their symbol is
+             used as it is for atomtypes but an empty string is used for
+             elements.
```
lilyminium: Instead of `SYM2Z`, could you please just clarify that the symbols are not real, valid elements?

package/MDAnalysis/converters/OpenMMParser.py
```diff
@@ -123,24 +136,58 @@ def _mda_topology_from_omm_topology(self, omm_topology):
...
+        masses = []
+        atomtypes = []
+        for a in omm_topology.atoms():
+            elem = a.element
+            if elem is not None:
+                if elem.symbol.capitalize() in SYMB2Z:
+                    validated_elements.append(elem.symbol.capitalize())
+                else:
+                    validated_elements.append('')
+                atomtypes.append(elem.symbol.capitalize())
```
lilyminium: Could this please not be `capitalize()`d -- we can retain the capitalisation of the input element, instead.

package/MDAnalysis/converters/OpenMMParser.py
```diff
@@ -123,24 +136,58 @@ def _mda_topology_from_omm_topology(self, omm_topology):
...
+                masses.append(elem.mass._value)
+            else:
+                wmsg = (f"Unknown element {elem} found for some atoms. "
```
lilyminium: This will always come out to `Unknown element None`, right? But that doesn't quite make sense -- there is just *no* element. Could you please change the error message to make that clear?

package/MDAnalysis/converters/OpenMMParser.py
```diff
@@ -123,24 +136,58 @@ def _mda_topology_from_omm_topology(self, omm_topology):
...
+                        f"These have been given an empty element record "
+                        f"with their atomtype set to 'X' "
+                        f"and their mass set to 0.0. "
+                        f"If needed they can be guessed using "
+                        f"MDAnalysis.topology.guessers.")
+                warnings.warn(wmsg)
```
lilyminium: This warning will trigger for every invalid atom -- that sounds like it could get overwhelming. MDAnalysis already has a lot of warnings as it is. Could you please only warn once, if any element is not valid?

testsuite/MDAnalysisTests/converters/test_openmm_parser.py
```diff
@@ -130,6 +160,82 @@ class TestOpenMMTopologyParser(OpenMMTopologyBase):
...
+
+class TestOpenMMTopologyParserWithPartialElements(OpenMMTopologyBase):
+    ref_filename = app.PDBFile(PDB).topology
+    expected_n_atoms = 47681
+    expected_n_residues = 11302
+    expected_n_segments = 1
+    expected_n_bonds = 25533
+
+    def test_with_partial_elements(self, top):
+        if 'elements' in self.expected_attrs:
```
lilyminium: I don't understand why this if is here, could you please explain to me? Shouldn't you know whether or not there are elements in this PDB file?

Edit: if this is the same file as above, it looks like there are no elements... so this test doesn't look like it runs?

testsuite/MDAnalysisTests/converters/test_openmm_parser.py
```diff
@@ -130,6 +160,82 @@ class TestOpenMMTopologyParser(OpenMMTopologyBase):
...
+        "ids",
+        "names",
+        "resids",
+        "resnames",
+        "masses",
+        "bonds",
+        "chainIDs",
+    ]
+
+    def test_no_elements_warn(self, top):
```
lilyminium: This test doesn't use the `top` fixture.

testsuite/MDAnalysisTests/converters/test_openmm_parser.py
```diff
@@ -130,6 +160,82 @@ class TestOpenMMTopologyParser(OpenMMTopologyBase):
...
+
+
+class TestOpenMMTopologyParserWithPartialElements(OpenMMTopologyBase):
+    ref_filename = app.PDBFile(PDB).topology
+    expected_n_atoms = 47681
+    expected_n_residues = 11302
+    expected_n_segments = 1
+    expected_n_bonds = 25533
+
+    def test_with_partial_elements(self, top):
```
lilyminium: This test doesn't use the `top` fixture.

testsuite/MDAnalysisTests/converters/test_openmm_parser.py
```diff
@@ -130,6 +160,82 @@ class TestOpenMMTopologyParser(OpenMMTopologyBase):
...
+                "will not be populated. "
+                "Atomtype attribute will be guessed using atom "
+                "name and mass will be guessed using atomtype."
+                "See MDAnalysis.topology.guessers.")
+
+        with pytest.warns(UserWarning, match=wmsg):
+            mda_top = self.parser(self.ref_filename).parse()
+
+
+class TestOpenMMTopologyParserWithPartialElements(OpenMMTopologyBase):
```
lilyminium: Does this class need to subclass `OpenMMTopologyBase`? Do you use any of the base class tests or fixtures?

testsuite/MDAnalysisTests/converters/test_openmm_parser.py
```diff
@@ -130,6 +160,82 @@ class TestOpenMMTopologyParser(OpenMMTopologyBase):
...
+                    "If needed they can be guessed using "
+                    "MDAnalysis.topology.guessers.")
+            with pytest.warns(UserWarning, match=wmsg):
+                mda_top = self.parser(self.ref_filename).parse()
+                assert mda_top.types.values[3344] == 'X'
+                assert mda_top.types.values[3388] == 'X'
+                assert mda_top.elements.values[3344] == ''
+                assert mda_top.elements.values[3388] == ''
+
+    def test_abnormal_elements(self, top):
```
lilyminium: This test doesn't use the `top` fixture. Also, it doesn't use `self` at all, does it need to be in this class?

package/MDAnalysis/converters/OpenMMParser.py
```diff
@@ -123,24 +136,58 @@ def _mda_topology_from_omm_topology(self, omm_topology):
...
+        atomtypes = []
+        for a in omm_topology.atoms():
+            elem = a.element
+            if elem is not None:
+                if elem.symbol.capitalize() in SYMB2Z:
+                    validated_elements.append(elem.symbol.capitalize())
+                else:
+                    validated_elements.append('')
+                atomtypes.append(elem.symbol.capitalize())
+                masses.append(elem.mass._value)
```
lilyminium: I suddenly have a horrible thought.... does anyone know how common it is to specify masses in non-amu units? @jbarnoud or @ahy3nz?

Regardless, I think it's probably better to avoid "private" attributes (prefixed by an underscore) and use the public API, i.e. `elem.mass.value_in_unit(unit.daltons)`

testsuite/MDAnalysisTests/converters/test_openmm_parser.py
```diff
@@ -130,6 +160,82 @@ class TestOpenMMTopologyParser(OpenMMTopologyBase):
...
+
+        assert mda_top.types.values[0] == 'Ag'
+        assert mda_top.types.values[1] == '*'
+        assert mda_top.types.values[2] == '?'
+        assert mda_top.elements.values[0] == 'Ag'
+        assert mda_top.elements.values[1] == ''
+        assert mda_top.elements.values[2] == ''
+        assert mda_top.masses.values[0] == 107.86822
+        assert mda_top.masses.values[1] == (0*daltons)._value
+        assert mda_top.masses.values[2] == (6*daltons)._value
```
lilyminium: Could you please just compare these to the actual floats?

testsuite/MDAnalysisTests/converters/test_openmm_parser.py
```diff
@@ -130,6 +160,82 @@ class TestOpenMMTopologyParser(OpenMMTopologyBase):
...
+    def test_abnormal_elements(self, top):
+        omm_top = Topology()
+        bad1 = Element(0, "*", "*", 0*daltons)
+        bad2 = Element(0, "?", "?", 6*daltons)
+        silver = Element.getBySymbol('Ag')
+        chain = omm_top.addChain()
+        res = omm_top.addResidue('R', chain)
+        omm_top.addAtom(name='Ag', element=silver, residue=res)
+        omm_top.addAtom(name='Bad1', element=bad1, residue=res)
+        omm_top.addAtom(name='Bad2', element=bad2, residue=res)
```
lilyminium: Thanks for adding this test of invalid elements! Could you please just add an atom without an element as well, to test that :)

lilyminium (changes requested):
Thanks for the changes @Atharva7K, just a few more comments!

package/MDAnalysis/converters/OpenMMParser.py
Atharva7K: I should've warned only once in the first place. The execution time for the tests reduced significantly after the redundant warnings were removed :)

testsuite/MDAnalysisTests/converters/test_openmm_parser.py
Atharva7K: That if indeed is redundant as partial elements are present in that file, I wasn't thinking clearly when I added it, anyway I've removed it so the tests run.

testsuite/MDAnalysisTests/converters/test_openmm_parser.py
Atharva7K: Now that you pointed it out, it doesn't need to subclass. Got rid of many redundant tests after removing it.

testsuite/MDAnalysisTests/converters/test_openmm_parser.py
Atharva7K: It did actually used `self` to call the parser, but I've moved the test to a separate function

Atharva7K:
@lilyminium , the tests are failing due to openmm and simtk `ModuleNotFoundError`. Is something wrong with my code like am I not supposed to import openmm in the parser file or something else?

package/MDAnalysis/converters/OpenMMParser.py
```diff
@@ -73,6 +76,11 @@
     Segids,
 )
 
+try:
+    from openmm.unit import daltons
+except ImportError:
+    from simtk.unit import daltons
+
```
lilyminium: @Atharva7K the tests are failing here because in minimal environments, OpenMM is not installed. Where do you use daltons? It would be better to just import it there.

Atharva7K: Umm I'm using daltons in the OpenMMParser to convert the mass using `elem.mass.value_in_unit(unit.daltons)` as you requested but I'm not sure how that can be done without importing daltons in the parser file. But since amu and dalton is [equal](https://github.com/openmm/openmm/blob/be19e0222ddf66f612016a3c1f687161a53c2396/wrappers/python/openmm/unit/unit_definitions.py#L213) in OpenMM, will it be fine to use the private `mass._value` like before?

lilyminium: Yes, you'll still need to import `daltons`, but doing it at the top level without a safeguard is leading to the error. There are two options here:

* add an `except ImportError` to the `simtk.units` that `pass`es if OpenMM is not installed, e.g. https://github.com/MDAnalysis/mdanalysis/blob/459c04ea86fc9f885d0c259e831a735111c249ce/package/MDAnalysis/converters/RDKit.py#L97-L101
* import `daltons` in the function where you use it

Currently it's erroring because the converter class is imported into the main package even when OpenMM is not installed, which runs into this top-level import. There's a similar function-level import in the RDKit package: https://github.com/MDAnalysis/mdanalysis/blob/459c04ea86fc9f885d0c259e831a735111c249ce/package/MDAnalysis/converters/RDKit.py#L322-L338

testsuite/MDAnalysisTests/converters/test_openmm_parser.py
```diff
@@ -130,6 +159,89 @@ class TestOpenMMTopologyParser(OpenMMTopologyBase):
...
+                f"with their atomtype set to 'X' "
+                f"and their mass set to 0.0. "
+                f"If needed they can be guessed using "
+                f"MDAnalysis.topology.guessers.")
+        with pytest.warns(UserWarning, match=wmsg):
+            mda_top = self.parser(self.ref_filename).parse()
+            assert mda_top.types.values[3344] == 'X'
+            assert mda_top.types.values[3388] == 'X'
+            assert mda_top.elements.values[3344] == ''
+            assert mda_top.elements.values[3388] == ''
```
lilyminium: From these two tests I'm assuming that `PDB` has some elements, but not all. It would be better to swap these tests then -- have `test_with_partial_elements` be in the class that subclasses  `OpenMMTopologyBase` without editing the OpenMM `ref_filename`. Editing this class attribute for the entire class `TestOpenMMTopologyParserWithNoElements` of tests is not great practice, as it might affect the other tests or future tests that we add to the class as well, or to any in the string of base classes (`OpenMMTopologyBase` and `ParserBase`).

Then, have a separate function that edits the OpenMM topology to have no elements and test that the right error message is raised. Only create and edit the OpenMM system inside the function to keep changes contained.

lilyminium (changes requested):
Thanks for fixing that @Atharva7K -- just one final comment on how the tests are arranged, the rest looks good to go!

testsuite/MDAnalysisTests/converters/test_openmm_parser.py
Atharva7K: Umm just to be clear, I have to do the following:
1) Move `test_with_partial_elements` in a class that subclasses ` OpenMMTopologyBase` (with no modification to `ref_filename`.
3) Make a separate function to test with no elements(by creating and testing the topology in the function itself)
4) Get rid of one redundant class that remains after all this.

Does this sound correct?

lilyminium: Yeah, that sounds about right -- the goal is to have tests that make sense. For example, it makes sense for the partial elements check to happen on a file that only has partial elements, during a whole suite of other tests first defined in ` OpenMMTopologyBase` or `ParserBase`. It makes a little less sense to edit the PDB file for the entire suite of tests just to accommoodate one test -- that can happen inside one function and be isolated, to avoid potential undesired side effects on the other tests in the class.

lilyminium (approved):
Fantastic, thank you! The fix looks great -- just waiting for CI to turn green to merge :-)

Atharva7K:
> Fantastic, thank you! The fix looks great -- just waiting for CI to turn green to merge :-)

Thanks! I really appreciate the time you invested in the PR, learnt good practices from you that'll help with future contributions.
I think codecov will complain again because the lines that raise error when openmm is not found are not covered by tests.

package/MDAnalysis/converters/OpenMMParser.py
```diff
@@ -123,24 +147,62 @@ def _mda_topology_from_omm_topology(self, omm_topology):
...
+                masses.append(0.0)
+                atomtypes.append('X')
+
+        if warn:
+            wmsg = (f"Element information missing for some atoms. "
+                    f"These have been given an empty element record "
+                    f"with their atomtype set to 'X' "
+                    f"and their mass set to 0.0. "
+                    f"If needed they can be guessed using "
+                    f"MDAnalysis.topology.guessers.")
```
IAlibay: If there's no variable substitution then f-strings aren't worth it.

```suggestion
            wmsg = ("Element information missing for some atoms. "
                    "These have been given an empty element record "
                    "with their atomtype set to 'X' "
                    "and their mass set to 0.0. "
                    "If needed they can be guessed using "
                    "MDAnalysis.topology.guessers.")
```

package/MDAnalysis/converters/OpenMMParser.py
```diff
@@ -105,16 +108,37 @@ def _mda_topology_from_omm_topology(self, omm_topology):
...
+             atomtypes.
+           * When partial elements are present, values from available elements
+             are used whereas the absent elements are assigned an empty string
+             with their atomtype set to 'X' and mass set to 0.0.
+           * For elements with invalid and unreal symbols, the symbol is used
+             as it is for atomtypes but an empty string is used for elements.
+
         """
 
+        try:
```
IAlibay, line 130: Ideally we need a test for this, however this will need reformatting the openmm parser tests to not directly skip on import.

For the sake of getting this merged, can I ask that you raise covering this part of the code as a separate issue please?

package/MDAnalysis/converters/OpenMMParser.py
```diff
@@ -123,24 +147,62 @@ def _mda_topology_from_omm_topology(self, omm_topology):
...
+                    validated_elements.append('')
+                atomtypes.append(elem.symbol)
+                masses.append(elem.mass.value_in_unit(daltons))
+            else:
+                warn = True
+                validated_elements.append('')
+                masses.append(0.0)
+                atomtypes.append('X')
+
+        if warn:
```
IAlibay: I'm trying to follow the logic here, maybe I'm missing something obvious but doesn't this end up in some cases giving two conflicting warnings, and in others no warnings at all when there should be one?

So in the case of two conflicting warnings, technically you will set `warn` to `True` if all the elements are None right? So this warning here is triggered, then the warning below? So now users get told "hey your element information is going to be available as empty records", and then straight afterwards "hey I won't populate your element information".

In the case of no warning at all, you could hit line 171 as many times as you want and never generate a user warning. Users never know they are missing elements and carry on eventually getting downstream errors whenever they try to use something that needs elements.

Rather than using and setting a `warn` variable how about the following pseudo-code?
Note: it's not pretty, and calls to any and all aren't free, but hopefully they don't cost too much? (would need checking)

```python
if not all(validated_elements):
    if any(validated_elements):
        warn about missing element info for some atoms
        append elements
        if any([i=='X' for i in atomtypes]):
            warn about atomtypes and masses
    else:
        guess types & masses and warn about lack of elements

**Review by IAlibay — CHANGES_REQUESTED** [review]
Overall looks good to me, but unfortunately there appears to be an issue with the warning logic which will either cause confusion for users or fail to warn users in some cases.

File: package/MDAnalysis/converters/OpenMMParser.py (hunk 40 above)
Comment (Atharva7K) [reply package/MDAnalysis/converters/OpenMMParser.py]: Thanks for pointing that out, it was a potential issue. I timed the previous commit and the current one using asv and apparently using any/all  doesn't cost too much.

File: package/MDAnalysis/converters/OpenMMParser.py (hunk 39 above)
Comment (Atharva7K) [reply package/MDAnalysis/converters/OpenMMParser.py:130]: I've raised #3637 to address this, let me know if everything is covered. Am I supposed to submit a PR for this issue or it's meant for other contributors?

File: package/MDAnalysis/converters/OpenMMParser.py [hunk 41]
```diff
@@ -123,24 +147,64 @@ def _mda_topology_from_omm_topology(self, omm_topology):
...
+        validated_elements = []
+        masses = []
+        atomtypes = []
+        for a in omm_topology.atoms():
+            elem = a.element
+            if elem is not None:
+                if elem.symbol.capitalize() in SYMB2Z:
+                    validated_elements.append(elem.symbol)
+                else:
+                    validated_elements.append('')
```
richardjgowers, line 174: when is this branch reached?

package/MDAnalysis/converters/OpenMMParser.py
```diff
@@ -123,24 +147,64 @@ def _mda_topology_from_omm_topology(self, omm_topology):
...
+                atomtypes.append(elem.symbol)
+                masses.append(elem.mass.value_in_unit(daltons))
+            else:
+                warn = True
```
richardjgowers: when is the warn variable used?

package/MDAnalysis/converters/OpenMMParser.py
```diff
@@ -123,24 +147,64 @@ def _mda_topology_from_omm_topology(self, omm_topology):
...
+                validated_elements.append('')
+                masses.append(0.0)
+                atomtypes.append('X')
+
+        if not all(validated_elements):
+            if any(validated_elements):
+                warnings.warn("Element information missing for some atoms. "
+                              "These have been given an empty element record ")
+                if any([i == 'X' for i in atomtypes]):
```
richardjgowers: use a generator expressions `any(i == 'X' for i in atomtypes)`

package/MDAnalysis/converters/OpenMMParser.py
Atharva7K: Forgot to remove it from the last commit, sorry

package/MDAnalysis/converters/OpenMMParser.py
Atharva7K: That branch is reached in case of invalid elements. We use empty string for elements but use the element symbol as is for atomtype . [Context](https://github.com/MDAnalysis/mdanalysis/pull/3511#discussion_r843332961)

package/MDAnalysis/converters/OpenMMParser.py
```diff
@@ -105,16 +108,37 @@ def _mda_topology_from_omm_topology(self, omm_topology):
...
+           * The parser now works when element attribute is missing in some or
+             all the atoms.
+           * When none of the elements are present their atomtypes are guessed
+             using their names and their masses are then guessed using their
+             atomtypes.
+           * When partial elements are present, values from available elements
+             are used whereas the absent elements are assigned an empty string
+             with their atomtype set to 'X' and mass set to 0.0.
+           * For elements with invalid and unreal symbols, the symbol is used
+             as it is for atomtypes but an empty string is used for elements.
```
IAlibay: Move these to a `Notes` section just above instead. Keep versionchanged brief and instead provide the information that's going to be useful going forward in the docstring.

IAlibay (changes requested):
Just the one final thing from me, @richardjgowers do you want to give this another look?

IAlibay (approved):
Thanks for all the work on this @Atharva7K !

IAlibay:
@jbarnoud you have a blocking review - please re-review.

jbarnoud (approved):

jbarnoud:
Thank you @Atharva7K ! :tada:

PR #3513, opened 2022-01-29
Refactor distance header to use wider integer representation 

### LINKED ISSUE #3512: Segmentation fault for large vectors in distances.distance_array
## Expected behavior ##

MDAnalysis.lib.distances.distance_array should work for large vectors.

## Actual behavior ##
MDAnalysis.lib.distances.distance_array has a segmentation fault for vectors of length >50000.
I think this may be due to integer overflow in the array indexing in the C code in lib/include/calc_distances.h, the line:
  `*(distances+i*numconf+j) = sqrt(rsq);`

``` python
import numpy as np
import MDAnalysis


N=50000
print(N,N*N)
a=np.random.rand(N,3)
b=np.random.rand(N,3)

d=MDAnalysis.lib.distances.distance_array(a,b)
```
Output for me is:

```
50000 2500000000
Segmentation fault
```

- Which version are you using?  2.0.0
- Which operating system? Linux (HPE Cray Linux)

hmacdope:
Fixes #3512

 - refactor `calc_distances.h` to use uint_64 indexing to avoid overflows. 

Looking for some inital feedback and we can work from there. 
Still to discuss `int64_t` vs `uint64_t`. Leave your thoughts!

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


jbarnoud:
Thank you for tackling this!

Could you add a check in the python part to make sure the resulting matrix is not too large for the index type? We should be clear for a while with int64, but changing the type does not prevent the original to come back after when simulation will become even larger.

hmacdope:
Did we decide on `in64_t` vs `uint64_t` in the end ? A switch between the two in this PR is easy but is probably worth considering.

jbarnoud:
> Did we decide on `in64_t` vs `uint64_t` in the end ? A switch between the two in this PR is easy but is probably worth considering.

The question was whether or not `int64` is faster than `uint64` because of compiler optimisations. I guess the only way to know would be to run a quick benchmark. Otherwise, we are just imagining answers.

richardjgowers:
Merge as is and worry about signs later?

IAlibay:
Agreed with @jbarnoud - also the specific influence of uint vs int is likely instruction specific iirc. Like uint has bounds checks in C whilst int doesn't, but if I remember correctly division requires fewer instructions with uint? It's likely all very meaningless if our bottleneck is always feeding the data in.

IAlibay (changes requested):
I realise it's just going to be a test that just passes, but can we add a check that we actually fixed the original issue please?

The author of the issue provided a pretty simple synthetic test for it, we should just have that lying around to make sure we're not hitting a limitation somewhere else eventually.

hmacdope:
> > Did we decide on `in64_t` vs `uint64_t` in the end ? A switch between the two in this PR is easy but is probably worth considering.
> 
> The question was whether or not `int64` is faster than `uint64` because of compiler optimisations. I guess the only way to know would be to run a quick benchmark. Otherwise, we are just imagining answers.

possibly too instruction set, and compiler flag specific to provide a general answer I think.  I can try cook something up on quickbench as a reference. 

I'll add in a test to the calling python code that @jbarnoud  suggested: N*M < MAX_TYPE

The original issue just has a segfault so will have to test explicitly.

hmacdope:
See [here](https://quick-bench.com/q/lPCgExi-aOl_L-3vKFHyhv6ESaA) for an example benchmark. Almost exactly the same and without `benchmark::DoNotOptimize` (designed to mimic python memory managment leading into call) they compile down to almost the same asm.

jbarnoud:
> See [here](https://quick-bench.com/q/lPCgExi-aOl_L-3vKFHyhv6ESaA) for an example benchmark. Almost exactly the same and without `benchmark::DoNotOptimize` (designed to mimic python memory managment leading into call) they compile down to almost the same asm.

I would say that if there is no perf difference between signed and unsigned, then we should go for unsigned. Let's not waste good bits on negative numbers we won't use.

richardjgowers (approved):

IAlibay:
I _think_ all we need is that test and we're good to go here?

jbarnoud:
We also need the bound checks, then it should be good to go.

IAlibay:
> We also need the bound checks, then it should be good to go.

Sorry this is what I meant by tests, or at least tests of the checks :P

hmacdope:
Problem:  I could be missing something obvious, but how do we test that the limit is caught in a test without actually allocating an array that is way way way too large for memory. 

By my calculations to overflow `uint64_t` with `calc_distance_array` you need an array that is at least:

 sqrt(0xffffffffffffffff) = 4294967296 elements = 17179869184 bytes (float32) = ~ 17GB. (hopefully my maths holds up). 

My guess is that this is too memory hungry for most people.

IAlibay:
> Problem:  I could be missing something obvious, but how do we test that the limit is caught in a test without actually allocating an array that is way way way too large for memory. 
> 
> By my calculations to overflow `uint64_t` with `calc_distance_array` you need an array that is at least:
> 
>  sqrt(0xffffffffffffffff) = 4294967296 elements = 17179869184 bytes (float32) = ~ 17GB. (hopefully my maths holds up). 
> 
> My guess is that this is too memory hungry for most people. 

Not checked the new bounds checking code but can you just patch the function in test to override the necessary variables?

hmacdope:
The size is pulled off the array directly and not passed to the function directly. See below snippet. Perhaps there is some pytest hack im not aware of?

```python
    refnum = reference.shape[0]
    distnum = refnum * (refnum - 1) // 2
    # check resulting array will not overflow UINT64_MAX
    if distnum >_UINT64_MAX:
        raise ValueError(f"Size of resulting array {distnum} elements larger than size of maximum integer")
```

jbarnoud:
Since the check appends on the python side, we have quite a lot of flexibility. I would create a mock array with a dummy shape attribute.

hmacdope:
I'm working on this with Mock but I'm still getting used to mock itself.

jbarnoud:
If mock proves itself annoying, you should be able to create a `FakeArray` class with only a `shape` attribute.

hmacdope:
> If mock proves itself annoying, you should be able to create a `FakeArray` class with only a `shape` attribute. 

The problem is there is a bunch of type checking, with isinstance and also as part of some other calls but I'm sure I can figure it out :)

jbarnoud:
> The problem is there is a bunch of type checking, with isinstance and also as part of some other calls but I'm sure I can figure it out :)

Just for reference, this is the nasty decorator to please: https://github.com/MDAnalysis/mdanalysis/blob/16edabd9ff14f9f0d908c8b2411dfa6b691cacbc/package/MDAnalysis/lib/util.py#L1932

jbarnoud:
That may be of use to pass the decorator: https://stackoverflow.com/a/33024739

jbarnoud:
> That may be of use to pass the decorator: https://stackoverflow.com/a/33024739

I got some luck with the following snippet:

```python
import MDAnalysis as mda
import numpy as np

class FakeArray(np.ndarray):
    shape = (23, 9)

a = FakeArray([1, 2, 3])
b = FakeArray([4, 5, 6])

print(a.shape, b.shape)

mda.lib.distances.distance_array.__wrapped__(a, b)
```

jbarnoud (commented):
From what I see, it is missing a changelog entry. Otherwise, lgtm.

hmacdope:
Let me double check I have got the limits correct then I'll add changelog entry :)

hmacdope:
Its worth noting that other things **may** go wrong before this warning is triggered. IE to actually get a **result** distance array larger than `UINT64_MAX`, you would need to have coordinate arrays  of size `UINT64_MAX*3` when flat which I am fairly sure is larger than `size_t` on most platforms?  I'm not sure how numpy handles this, but regardless memory will rule this out 99.9999% of the time anyway.

jbarnoud:
This should at least fix the segfault that was reported.

jbarnoud (approved):

jbarnoud:
@IAlibay if you are happy with the changes, you can merge

package/MDAnalysis/lib/distances.py
```diff
@@ -66,6 +66,7 @@
 .. autofunction:: augment_coordinates(coordinates, box, r)
 .. autofunction:: undo_augment(results, translation, nreal)
 """
+from asyncio import LimitOverrunError
```
IAlibay: Is this import being used anywhere?
```suggestion
```

IAlibay (changes requested):
The one comment for me, is that error from asyncio actually being used anywhere?

hmacdope:
Good spot @IAlibay thanks!

IAlibay (approved):
lgtm

PR #3518, opened 2022-02-04
Fix import warning

### LINKED ISSUE #3506: Deprecation of simtk.openmm
## Expected behavior ##

MDAnalysis has no upstream deprecation warnings

## Actual behavior ##

Using the OpenMM converter yields the following deprecation warning:

`Warning: importing 'simtk.openmm' is deprecated.  Import 'openmm' instead.`

``` python
import MDAnalysis as mda
from openmm import app
from MDAnalysisTests.datafiles import PDBX

u = mda.Universe(app.PDBxFile(PDBX))
```

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) 2.1.0-dev0
- Which operating system? Ubuntu 20.04 LTS

yangtcai:
Fixes #3506 

 - 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/converters/OpenMM.py
+++ b/package/MDAnalysis/converters/OpenMM.py
@@ -85,7 +85,7 @@ class OpenMMSimulationReader(base.SingleFrameReaderBase):
         """Can this reader read *thing*?
         """
         try:
-            from simtk.openmm.app import Simulation
+            from openmm.app import Simulation
         except ImportError:
             return False
         else:
@@ -150,7 +150,7 @@ class OpenMMAppReader(base.SingleFrameReaderBase):
         """Can this reader read *thing*?
         """
         try:
-            from simtk.openmm import app
+            from openmm import app
         except ImportError:
             return False
         else:
--- a/package/MDAnalysis/converters/OpenMM.py
+++ b/package/MDAnalysis/converters/OpenMM.py
@@ -87,7 +87,10 @@ class OpenMMSimulationReader(base.SingleFrameReaderBase):
         try:
             from openmm.app import Simulation
         except ImportError:
-            return False
+            try:
+                from simtk.openmm.app import Simulation
+            except ImportError:
+                return False
         else:
             return isinstance(thing, Simulation)
 
@@ -152,7 +155,10 @@ class OpenMMAppReader(base.SingleFrameReaderBase):
         try:
             from openmm import app
         except ImportError:
-            return False
+            try:
+                from simtk.openmm import app
+            except ImportError:
+                return False
         else:
             return isinstance(thing, (app.PDBFile, app.Modeller,
                 app.PDBxFile))


IAlibay:
So just going to ping @jbarnoud and @lilyminium who probably have stronger opinions than I on this - I think we should try to support older versions of OpenMM by trying to still do the simtk import as a fallback, but it does make for a rather ugly import try/except, thoughts?

Edit: with hindsight it's two lines it's not that bad

package/MDAnalysis/converters/OpenMM.py
```diff
@@ -85,7 +85,7 @@ def _format_hint(thing):
         """Can this reader read *thing*?
         """
         try:
-            from simtk.openmm.app import Simulation
```
IAlibay, line 88: Ideally we would like to support older versions which require the simtk import. Could you re-arrange this to account for that?

jbarnoud:
> So just going to ping @jbarnoud and @lilyminium who probably have stronger opinions than I on this - I think we should try to support older versions of OpenMM by trying to still do the simtk import as a fallback, but it does make for a rather ugly import try/except, thoughts?

Agreed. At least for a while, it would be good to maintain compatibility with older versions of OpenMM. Especially since that support is not that expensive to keep.

IAlibay (changes requested):
Thanks @yangtcai please see below.

package/MDAnalysis/converters/OpenMM.py
jbarnoud: In practice, it would be something in the line of:

* try importing directly from `openmm`
* if it fails, fall back to importing from `simtk`
* if that fails too, we do not have openmm

yangtcai: Thanks a lot, I have pushed another commit to keep compatibility, PTAL

IAlibay:
Hi,

MDAnalysis is currently in the process of evaluating a potential change in its license. As part of this we are asking that all new contributors introduce themselves (with an email address that they regularly monitor) to the [MDAnalysis developer list](https://groups.google.com/group/mdnalysis-devel) before we accept their contributions. Please include your github handle and a link to this pull request. The aim is that we then have a means of contacting the developers should we decide to change our license in the near future.

If you are contributing with interest to take part in GSoC 2022, please note that communicating with the developers list is highly encouraged. [Our FAQ contains more information](https://github.com/MDAnalysis/mdanalysis/wiki/GSoC-FAQ#do-you-want-to-see-my-work-and-learn-more-about-me). Please note that the FAQ has not been updated since 2021 and its contents are not yet considered final until the GSoC application window opens on the 4th of April.

On behalf of the MDAnalysis core development team

yangtcai:
Thanks, I have sent introduction in the mail list, I wish it would be a wonderful trip with MDAnalysis. :DD

IAlibay (approved):
Thanks @yangtcai !

PR #3527, opened 2022-02-13
BLD, MAINT: hoist packaging import

tylerjereddy:
* hoist the import of `packaging` outside of a Cython-focused
`try`/`except` block to avoid a deceptive error message when
`packaging` is missing, per discussion at:
https://github.com/MDAnalysis/mdanalysis/pull/3398#issuecomment-1037371853

* we may also want to consider removing the external dependency
completely and vendoring the pertinent code since it is so small, but
maybe not this close to a release; an example:
https://github.com/scipy/scipy/blob/main/scipy/_lib/_pep440.py

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/setup.py
+++ b/package/setup.py
@@ -46,6 +46,7 @@ Google groups forbids any name that contains the string `anal'.)
 from setuptools import setup, Extension, find_packages
 from distutils.ccompiler import new_compiler
 from distutils.sysconfig import customize_compiler
+from packaging.version import Version
 import codecs
 import os
 import sys
@@ -83,7 +84,6 @@ try:
     import Cython
     from Cython.Build import cythonize
     cython_found = True
-    from packaging.version import Version
 
     required_version = "0.16"
     if not Version(Cython.__version__) >= Version(required_version):


package/setup.py
```diff
@@ -83,7 +84,6 @@
     import Cython
     from Cython.Build import cythonize
     cython_found = True
-    from packaging.version import Version
```
IAlibay, line 86: So thinking about it a bit more, this explicitly makes `packaging` a core requirement for user builds right?

I guess the question is - where do we stand on eventually making Cython a core dependency (rather than us shipping the C/C++ files)? If we want to keep things the way we usually do, then I think we should possibly try/except around the packaging import and just set it to either say that either cython or packaging isn't found?

IAlibay:
@MDAnalysis/coredevs please review - this will technically add a dependency. I don't personally think `packaging` is that big that it will cause issues (it gets pulled in by one or more of our core dependencies anyways).

jbarnoud:
When is `packaging` installed?

IAlibay:
> When is `packaging` installed? 

So my understanding is that if you pip install it'll pull it in as a build dependency (based on the project.toml), otherwise on a python setup.py call you'll need to have it installed (i.e. it'll need to be under `requires` in setup.py).

I'm not sure which, but one of our core dependencies already pulls it in so we had added it in without ever really caring too much until now.

IAlibay:
Poking around a little bit, seems like we've already bought into using `packaging` internally as a core dependency.

For example: https://github.com/MDAnalysis/mdanalysis/blob/3b7590e5e9de93371c26572f0d4dc35a31d891cd/package/MDAnalysis/coordinates/chemfiles.py#L84

I think we should just move ahead with putting `packaging` down as both a `setup_requires` and a `install_requires` entry?

jbarnoud:
If we imply it anyway, we should have it explicit.

tylerjereddy:
> please review - this will technically add a dependency.

Huh? MDAnalysis added the `packaging` dependency months ago, I'm not adding a dependency here, I'm just making the error message a bit less useless when `packaging` is absent.

package/setup.py
tylerjereddy: > So thinking about it a bit more, this explicitly makes packaging a core requirement for user builds right?

What? I just moved the import so the error message would be better when `packaging` is absent--it was the MDA team who added this dependency months ago.

IAlibay:
> > please review - this will technically add a dependency.
> 
> Huh? MDAnalysis added the `packaging` dependency months ago, I'm not adding a dependency here, I'm just making the error message a bit less useless when `packaging` is absent.

@tylerjereddy - yeah you're right, but the problem we face is that `packaging` was "added" implicitly rather than explicitly in #3499 (if that makes seinse - I only realised this in https://github.com/MDAnalysis/mdanalysis/pull/3527#issuecomment-1038359006).

Sorry about the confusing/confused messages, essentially my final point is that we probably should reflect this dependency in `setup.py`?

tylerjereddy:
> Sorry about the confusing/confused messages, essentially my final point is that we probably should reflect this dependency in setup.py?

Honestly, it looks like the team accidentally added a *runtime* dependency and not just a *build time* dependency, which is a bit more serious, even for the average consumer of a pre-built binary for example.

Given the increased gravity of a *runtime* dependency addition, and one that wasn't really discussed with the team IIUC, I'd probably vote for vendoring the module I mention above, which may also have the added benefit of avoiding chicken-and-egg problems with needing an external package to make (versioning) decisions about other external packages in the setup code.

I'm not sure that any of the possible decisions really make a difference to this PR though--regardless of where the machinery comes from, and what else you tell `setup.py` to do, the import for version handling still seems appropriately placed at the top-level, where the default `ImportError` would probably be clear enough without the Cython error message confusion.

It looks like there may have been an intention to make `packaging` a build time dependency only when building a development branch, but not for the `sdist` official releases placed on PyPI. This is rather subtle though--you can only get away with that as long as you only need version checking on Cython and nothing else and always have pre-built Cython in the `sdist`. I suppose you could still make `packaging` an optional build-time dependency for git checkouts only, and a mandatory run-time dependency, which is the strongest argument I can see against this PR, but I doubt you're gaining much if it is already a mandatory runtime dependency anyway.

So, it still seems to me like this is one of the cases where vendoring a small chunk of external code may actually be the right choice to avoid some head-spinning dependency issues, and then a top-level import should always be safe anyway.

tylerjereddy:
One other point is that I'm not sure why we have such a low minimum version of Cython in `setup.py`--I'd say we should feel free to bump that up quite a bit since it is a build time dependency only and even at the very base of the ecosystem in NumPy the minimum version is way higher: https://github.com/numpy/numpy/blob/main/pyproject.toml#L8

Plus I've already set `Cython>=0.28` in `pyproject.toml`. Do we support the 3.x series already? These may be more appropriate in separate issues...

jbarnoud:
I agree that the vendoring of `version` is out of the scope of that PR. I opened #3532 to track that specific issue.

The cython version issue is likely also a separate issue. I opened #3533 about it.

jbarnoud (approved):
I approve the PR but wait for @IAlibay to give a green light as well before merging. Let's try to limit the scope of each PR.

IAlibay:
> Honestly, it looks like the team accidentally added a runtime dependency and not just a build time dependency, which is a bit more serious, even for the average consumer of a pre-built binary for example.

Yeah I think this is on me as the approver of that PR (I forgot `packaging` wasn't a built-in package).

Here's what I propose we do:

1 - In this PR add `packaging` as a listed setup.py + pyproject.toml `setup_requires` and `install_requires`
2 - Raise an issue to vendor `packaging` in the future.

I realise this is an issue that happened before this PR, but if we can fix it now we can just push out 2.1.0.

Does that sound sensible @tylerjereddy ?

IAlibay (changes requested):
putting a temporary halt on merge whilst we discuss next steps

IAlibay:
> > Honestly, it looks like the team accidentally added a runtime dependency and not just a build time dependency, which is a bit more serious, even for the average consumer of a pre-built binary for example.
> 
> Yeah I think this is on me as the approver of that PR (I forgot `packaging` wasn't a built-in package).
> 
> Here's what I propose we do:
> 
> 1 - In this PR add `packaging` as a listed setup.py + pyproject.toml `setup_requires` and `install_requires` 2 - Raise an issue to vendor `packaging` in the future.
> 
> I realise this is an issue that happened before this PR, but if we can fix it now we can just push out 2.1.0.
> 
> Does that sound sensible @tylerjereddy ?

We need to get this moving so I'll merge this and follow up with a PR that explicitly adds packaging as a dependency.

IAlibay (approved):

PR #3528, opened 2022-02-13
Towards PEP621

### LINKED ISSUE #3526: BLD: PEP518/pyproject.toml todo items
Irfan is planning a release soon, and if we do plan to upload a source distribution to PyPI for that release we may want to add a few things to `pyproject.toml` beyond the initial draft in gh-3398. If issues with that file do pop up please ping me directly (`@`) so I see the notification on my phone (a few too many GitHub emails to scan these days..).

- [ ] add a `[project]` section with some useful metadata (can mimic other popular Python projects for this most likely)
  - [ ] name
  - [ ] license
  - [ ] maintainers
  - [ ] requires-python
  - [ ] dependencies (this means "runtime" instead of "buildtime" I think, but just for metadata as I understand it, at least for `pip`)
  - [ ] readme (path to readme file I think)
  - [ ] classifiers (the PyPI homepage for the release will extract a bunch of metadata from here like supported operating systems, and the versions of Python we consider supported)
- [ ] `[project.optional-dependencies]` section
  - [ ] optional testing deps
  - [ ] optional doc building deps
  - [ ] optional development tools/linters, etc.
- [ ] `[project.urls]` for homepage, source repo, etc.

IAlibay:
Fixes #3526

 - Adds PEP621 items to pyproject.toml
 - Updated LICENSE in package to reflect additions to the main repo LICENSE
 - Removed pypi-description.rst (the 2.0.0 release was fine without it, so I'm assuming whatever made it necessary is no longer needed? - we'll see on release)
 - Updated setup.py to reflect that we're actually GPL v2+ (using poetry's naming convention here)

TODO:
 - Add pyproject.toml for testsuite

PR Checklist
------------
 - Tests?
 - Docs?
 - CHANGELOG updated?


package/pyproject.toml
```diff
@@ -12,3 +12,61 @@ requires = [
...
+version = 2.1.0-dev0
+license = {file = "LICENSE"}
+authors = [
+    {name = 'MDAnalysis Core Developers', email = 'mdanalysis@numfocus.org'}
+]
+maintainers = [
+    {name = 'MDAnalysis Core Developers', email = 'mdanalysis@numfocus.org'}
+]
+requires-python = ">=3.7"
+dependencies = [
```
IAlibay, line 42: I wasn't too sure how to deal with the optional deps here.

Tests -> should I just add MDAnalysisTests as the dep? How do you dynamically set the version number in pyproject.toml?
Extra deps ->
  - Not sure what the keyword should be here (it's unclear if you can just say "extra").
  - How do we deal with conda only deps? Does it matter?

package/pyproject.toml
```diff
@@ -12,3 +12,61 @@ requires = [
...
+    'Programming Language :: Python :: 3.7',
+    'Programming Language :: Python :: 3.8',
+    'Programming Language :: Python :: 3.9',
+    'Programming Language :: C',
+    'Topic :: Scientific/Engineering',
+    'Topic :: Scientific/Engineering :: Bio-Informatics',
+    'Topic :: Scientific/Engineering :: Chemistry',
+    'Topic :: Software Development :: Libraries :: Python Modules',
+]
+urls = {'Documentation': 'https://docs.mdanalysis.org/',
```
IAlibay, line 64: Seemed cleaner to have these here than under [project.urls]

package/pyproject.toml
```diff
@@ -12,3 +12,63 @@ requires = [
...
+keywords = [
+    "python", "science", "chemistry", "biophysics", "molecular-dynamics",
+    "computational-chemistry", "molecular-simulation", "analysis",
+    "trajectory-analysis",
+]
+classifiers = [
+    'Development Status :: 6 - Mature',
+    'Environment :: Console',
+    'Intended Audience :: Science/Research',
+    'License :: OSI Approved :: GNU General Public License v2 or later (GPLv2+)',
```
IAlibay, line 66: @orbeckst I'm correct in thinking we can do this change right?

(edit: actually meant to point to setup.py, but here too, and in testsuite)

testsuite/pyproject.toml
```diff
@@ -0,0 +1,56 @@
...
+license = {file = "LICENSE"}
+authors = [
+    {name = 'MDAnalysis Development Team', email = 'mdanalysis@numfocus.org'}
+]
+maintainers = [
+    {name = 'MDAnalysis Core Developers', email = 'mdanalysis@numfocus.org'}
+]
+requires-python = ">=3.7"
+dependencies = [
+    'MDAnalysis==2.1.0-dev0',
```
IAlibay: I'm not super into the idea of having _yet-another_ point where we need to update the version numbers. Is there a way around this?

IAlibay:
@tylerjereddy a lot of this is based on my likely poor understanding of PEP 621 and 631. Any advice / corrections would be greatly appreciated.

package/pyproject.toml
```diff
@@ -12,3 +12,63 @@ requires = [
     "setuptools",
     "wheel",
 ]
+
+[project]
+name = "MDAnalysis"
+version = "2.1.0-dev0"
```
tylerjereddy: Can just do `dynamic = ['version']` I think, to avoid having to update this.

package/pyproject.toml
tylerjereddy: `conda` only deps seems like a bad idea in general, not sure you can do much about them here..

As for `dependencies`, I think it should just be the *mandatory* runtime dependencies, and then `[project.optional-dependencies]` can contain sections for `test`, `doc`, and `dev` tools (I'd suggest just looking at i.e., the SciPy example `pyproject.toml` and adapting for our purposes, if you aren't already doing that).

package/pyproject.toml
```diff
@@ -12,3 +12,63 @@ requires = [
...
+    'Intended Audience :: Science/Research',
+    'License :: OSI Approved :: GNU General Public License v2 or later (GPLv2+)',
+    'Operating System :: POSIX',
+    'Operating System :: MacOS :: MacOS X',
+    'Operating System :: Microsoft :: Windows ',
+    'Programming Language :: Python',
+    'Programming Language :: Python :: 3',
+    'Programming Language :: Python :: 3.7',
+    'Programming Language :: Python :: 3.8',
+    'Programming Language :: Python :: 3.9',
```
tylerjereddy, line 72: Are we not supporting `3.10` fully just yet?

testsuite/pyproject.toml
```diff
@@ -0,0 +1,56 @@
...
+# Minimum requirements for the build system to execute
+requires = [
+    "packaging",
+    "setuptools",
+    "wheel",
+]
+
+[project]
+name = "MDAnalysisTests"
+version = "2.1.0-dev0"
```
tylerjereddy: I think we can avoid updating this by using `dynamic = ['version', 'description']`, or similar, based on me cheating and looking at other projects.

testsuite/pyproject.toml
tylerjereddy: Hmm, it is indeed pretty ugly from a design standpoint now to have separate main and test packages.

Is there as much of a rush to add a `pyproject.toml` for the test suite?

It would probably be nice to have the version dynamically set to match the same version of the main package. Not sure how to do that.

tylerjereddy: A bit out of scope, but what other major Python projects package their test suite entirely separately from the main project? That might be a good reference point for "how should we handle this?"

If we were to start from scratch, I'd almost certainly vote against having separate packages, and instead use one of the popular "lazy loading/lazy download" packages to pull in test assets if there's an internet connection, otherwise skipping and doing a minimal suite only.

IAlibay:
So long story short, it seems that the latest release of setuptools is essentially enforcing a ton of things related to pyproject.toml which it didn't use to.

To speed things up I'm ditching the addition of a pyproject.toml for testsuite for now. Let's try to get the main package one up to scratch first, then we address testsuite later.

package/pyproject.toml
IAlibay: Sounds good, for now let's just skip `[project.optional-dependencies]` since most are conda based, I'll raise an issue on merge of this PR with all the things that can be added at a later date to augment the pyproject.toml.

package/pyproject.toml
```diff
@@ -12,3 +12,64 @@ requires = [
     "setuptools",
     "wheel",
 ]
+
+[project]
+name = "MDAnalysis"
+dynamic = ['version', 'readme']
```
IAlibay: @tylerjereddy can I pick your brains on this?

Essentially adding a direct readme entry in the pyproject.toml is complicated because it accesses the README.rst which exists outside of the `package` directory. So I'm thinking we can possibly just leave it as `dynamic` since `setup.py` has set logic for gathering `long-description`, does that make sense to do?

edit: adding to dynamic still caused issues... not sure what the best solution here is at this point.

package/pyproject.toml
IAlibay: @orbeckst can I ping you on this again please?

package/pyproject.toml
IAlibay: Fixed now. We didn't at the time of opening this as we didn't have tests in place for all the optional deps, but at this point we might as well. I doubt we'll get full 3.10 upstream compliance this side of the year...

package/pyproject.toml
```diff
@@ -12,3 +12,64 @@ requires = [
...
+    {name = 'MDAnalysis Development Team', email = 'mdanalysis@numfocus.org'}
+]
+maintainers = [
+    {name = 'MDAnalysis Core Developers', email = 'mdanalysis@numfocus.org'}
+]
+requires-python = ">=3.7"
+dependencies = [
+    'numpy>=1.18.0',
+    'biopython>=1.71',
+    'networkx>=1.0',
```
RMeli: NetworkX 2 is from 2017. Is there a good reason to keep NetworkX 1 supported? Does this prevent the usage of new features in NetworkX 2?

IAlibay: Yeah we probably should up all our dependencies, especially as we move closer to NEP29. I'm not against upping the minimum networkx version, I doubt this prevents networkx 2 though - in fact our CI runners use it: https://github.com/MDAnalysis/mdanalysis/runs/5754334356?check_suite_focus=true#step:6:94

RMeli: Yes, I've only ever used NetworkX 2 with MDA I think, that why I'm wondering why NetworkX 1 is still around. My comment was about adding new features requiring NetworkX; if version 1 is kept, one would have to make sure the the change is backward-compatible (and maybe can't use a fancy new function or a function with different API).

IAlibay: I see what you mean, tbh I'd be up for just getting rid of networkx as a core dependency overall (I think we barely use it for some leaflet stuff). But yeah, we should bump it up to 2.0+

package/pyproject.toml
orbeckst: Sorry, must have overlooked the notifications. 

Yes, that's excellent. I think I wasn't aware that there was a classifier for "GPL v2-or-later".

testsuite/pyproject.toml
orbeckst: For 3.0, getting rid of MDAnalysisTests would be a good aim. One thing to solve would be to keep the correct versions of test files in sync with the tests. Maybe download from the correct tag or commit in the repo? That might get a bit more tricky when you want to add/change test files in a branch... anyway, that would be something to think about and to learn from other packages that presumably already do this.

Which lazy-loading packages do you have in mind?

Do you have an example of a package that pulls in test files when needed?

package/pyproject.toml
orbeckst: If we ever get a "lipids MDAKit" then we can remove the leafletfinder together with the networkx dependency. However, networkx might be useful for the GSOC project on "ring groups". 

In anhy case, bumping the minimum version really should be fine.

orbeckst (commented):
The license update to "GPL v2 or later" is perfect. ✅

IAlibay:
I'm adding this to the 2.4.0 milestone, with the way that things are going we should attempt to be fully pep621 compliant by end of year.

TODOs here are mostly updating things.

package/pyproject.toml
```diff
@@ -23,3 +23,90 @@ requires = [
...
+    "h5py>=2.10",
+    "chemfiles>=0.10",
+    "pyedr>=0.7.0",
+]
+analysis = [
+    "seaborn",
+    "scikit-learn",
+    "tidynamics>=1.0.0",
+]
+doc = [
```
IAlibay, line 94: missing test entry because passing in a dynamic `version` for MDAnalysisTests doesn't really work...

IAlibay:
@MDAnalysis/coredevs this is good to go as needed for 2.4.0, I recommend using scipy's pyproject.toml as a template to check against during review: https://github.com/scipy/scipy/blob/main/pyproject.toml

IAlibay:
@MDAnalysis/coredevs can I get a review here please? This is stalling the 2.4.0 release

richardjgowers (approved):

IAlibay:
Ta, two PRs to go.

PR #3531, opened 2022-02-13
Fix for leftover simtk imports in openmm tests

IAlibay:
Follow up to #3518 - we forgot about the test imports for openmm

 - try/else around tests with fallback to a pytest.skip()
 - Also add missing author information for #3518 - I guess I got distracted over the whole license thing

Notes:
 - Originally attempted to do a bit more of a fine-grained pytest.mark.skipif, but had some issues with `test_openmm_parser` with some weird pytest behaviour that just hated the decorator. Given that we don't really need the fine-grained behaviour here I'm defaulting to just doing the pytest.skip().

PR Checklist
------------
 - Tests?
 - Docs?
 - CHANGELOG updated?

### DIFF
--- a/testsuite/setup.cfg
+++ b/testsuite/setup.cfg
@@ -18,6 +18,7 @@ filterwarnings=
     ignore:Failed to guess the mass:UserWarning
     # Coordinates
     ignore:No coordinate reader found:UserWarning
+    ignore:Reader has no dt information, set to 1.0 ps
     # NamedStream warnings
     ignore:Constructed NamedStream:RuntimeWarning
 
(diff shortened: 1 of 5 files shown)


testsuite/setup.cfg
```diff
@@ -18,6 +18,7 @@ filterwarnings=
     ignore:Failed to guess the mass:UserWarning
     # Coordinates
     ignore:No coordinate reader found:UserWarning
+    ignore:Reader has no dt information, set to 1.0 ps
```
IAlibay, line 21: This puts the warnings at 0 for the openmm converter tests.

jbarnoud (approved):

jbarnoud:
Thank you :grin:

PR #3541, opened 2022-02-19
Accept AtomGroup alongwith Universe in the DistanceMatrix

### LINKED ISSUE #3486: DistanceMatrix accept AtomGroup

Is there any reason DistanceMatrix cannot accept an AtomGroup inplace of a Universe? This would help with atomgroups created through non-selection-string ways, which you then need to re-create with atom indices. e.g.
https://groups.google.com/g/mdnalysis-discussion/c/ad4HW_6w2MM/m/-h3bRDv0DgAJ?utm_medium=email&utm_source=footer 

Here: https://github.com/MDAnalysis/mdanalysis/blob/5bf6359a821c085a1039ca414ce51ca2820722fb/package/MDAnalysis/analysis/diffusionmap.py#L251-L253

```diff
- universe.trajectory
+ universe.universe.trajectory
```

AnirG:
Fixes #3486

 - Accepts `AtomGroup` instead of `Universe` in the DistanceMatrix.
 - Modified tests to account for the modified input parameter.  

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/analysis/diffusionmap.py
```diff
@@ -245,12 +246,13 @@ class DistanceMatrix(AnalysisBase):
          :class:`MDAnalysis.analysis.base.Results` instance.
 
     """
-    def __init__(self, universe, select='all', metric=rmsd, cutoff=1E0-5,
+    def __init__(self, atomGroup, select='all', metric=rmsd, cutoff=1E0-5,
```
richardjgowers: This is missing the point of the change. Currently a Universe and selection string are both passed to construct an AtomGroup. The desired change is for this selection to happen outside of the function, so these two arguments are replaced by just an AtomGroup. To not break things abruptly, this will therefore require for a while accepting either U+selection (and warning about the change) OR AG.

richardjgowers (changes requested):

package/MDAnalysis/analysis/diffusionmap.py
AnirG: I see. I thought we'd have AG along with `selection` anyways should the need arise. I'll make the necessary changes to accept AG along with `Universe` and `selection`.

AnirG:
@richardjgowers does it seem fine? please let me know if I should add any more tests or any modification is required.

testsuite/MDAnalysisTests/analysis/test_diffusionmap.py
```diff
@@ -69,6 +74,13 @@ def test_dist_weights(u):
                                 [.707, -.707, 0, 0]]), 2)
 
 
+def test_distvalues_ag_universe(u, ag):
+    dist_universe = diffusionmap.DistanceMatrix(u, select='backbone').run()
+    dist_ag = diffusionmap.DistanceMatrix(ag).run()
+    assert_array_almost_equal(dist_universe.results.dist_matrix,
```
tylerjereddy: For *new* tests we may as well use `assert_allclose` if the matrices contain floats per https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_almost_equal.html

testsuite/MDAnalysisTests/analysis/test_diffusionmap.py
```diff
@@ -93,7 +105,7 @@ def test_long_traj(u):
 
 
 def test_not_universe_error(u):
-    trj_only = u.trajectory
+    trj_only = u.universe.trajectory
```
tylerjereddy: can you briefly explain what this change is for? The `u` fixture returns a universe and now we need another `.universe` call?

AnirG: yeah, since `u` can also be an `AtomGroup`.  I should've also updated `match` as `match='U is not a Universe or AtomGroup'`. I will do that.

testsuite/MDAnalysisTests/analysis/test_diffusionmap.py
AnirG: sure, will update it.

AnirG: Maybe we should consider replacing `assert_array_almost_equal` in other places as well as per the above link. I've updated it for only this test function.

IAlibay: At this moment in time our approach is to only change tests we touch and make sure new tests use assert_allclose. Making large blanket changes makes it very hard to review PRs and also understand why changes were made based on history. If we need to do a blanket change we will do this in a specific PR in the future.

package/MDAnalysis/analysis/diffusionmap.py
```diff
@@ -248,7 +249,8 @@ class DistanceMatrix(AnalysisBase):
     def __init__(self, universe, select='all', metric=rmsd, cutoff=1E0-5,
```
richardjgowers, line 248: This change still needs documenting in the docstring for the class.

richardjgowers (changes requested):

testsuite/MDAnalysisTests/analysis/test_diffusionmap.py
IAlibay: @tylerjereddy is correct here, u.trajectory == u.universe.trajectory, there's no need to make this change, either way you are testing what happens if you pass a trajectory through (neither a universe or an atomgroup)

testsuite/MDAnalysisTests/analysis/test_diffusionmap.py
```diff
@@ -69,6 +74,13 @@ def test_dist_weights(u):
                                 [.707, -.707, 0, 0]]), 2)
 
 
+def test_distvalues_ag_universe(u, ag):
+    dist_universe = diffusionmap.DistanceMatrix(u, select='backbone').run()
+    dist_ag = diffusionmap.DistanceMatrix(ag).run()
```
IAlibay, line 75: Honestly as much as I like fixtures I'm not sure it's worth building a module-level fixture just for the one test, particularly when this can just be `u.select_atoms('backbone')`

IAlibay (changes requested):
thanks @AnirG just a couple of small additional comments (on top of the doc changes Richard asked for).

AnirG:
Hi, I'm not sure as to why the one codecov test failed.

AnirG:
@tylerjereddy Hi looks like even #3527 failed similar tests on macOS like I'm facing. How should that be taken care of?

IAlibay:
This is related to https://github.com/MDAnalysis/mdanalysis/issues/3550 and some further issues upstream - we are currently dealing with this, there is nothing to do about this in this specific PR.

AnirG:
Hi @IAlibay! Can you please suggest any more changes? I'm looking forward to taking up more issues once I get done with this one, and possibly would love to hear some general practices I might've missed here.

AnirG:
@IAlibay Hi! Do the final changes seem fine? Would make the necessary changes otherwise.

jbarnoud (changes requested):
I restarted the tests as they seem to have failed on an unrelated issue. Assuming the tests do pass now, I think all the comments have been addressed appropriately. Thank you.

A couple of things are missing:
* [ ] you need to update the CHANGELOG
* [ ] you need to add yourself to the AUTHORS file

We hadn't release the 2.1.0 version when you last modified your pull request, so you will need to merge the develop branch or rebase against it to edit the most up to date version of these files.

Also, could you introduce yourself on the mailing list? We are discussing changes that will require us to contact all the contributors so we try to make sure we know how to reach new contributors. Thank you. (If you are interested in the [Google Summer of Code](https://www.mdanalysis.org/2022/03/07/gsoc2022/), please say so explicitly; many people are introducing themselves at the moment in that context and we need to know who's a candidate and who's not :smiley:)

AnirG:
@jbarnoud sure, will do that.

package/MDAnalysis/analysis/diffusionmap.py
```diff
@@ -171,7 +172,7 @@ class DistanceMatrix(AnalysisBase):
 
     Parameters
     ----------
-    universe : `~MDAnalysis.core.universe.Universe`
+    universe : `~MDAnalysis.core.universe.Universe or ~MDAnalysis.core.groups.AtomGroup`
```
jbarnoud:
```suggestion
    universe : `~MDAnalysis.core.universe.Universe` or `~MDAnalysis.core.groups.AtomGroup`
```

package/MDAnalysis/analysis/diffusionmap.py
```diff
@@ -328,12 +330,12 @@ def __init__(self, u, epsilon=1, **kwargs):
             Parameters to be passed for the initialization of a
             :class:`DistanceMatrix`.
         """
-        if isinstance(u, Universe):
+        if isinstance(u, AtomGroup) or isinstance(u, Universe):
```
jbarnoud, line 344: The change does not appear in the docstring just above.

jbarnoud (commented):
I missed some tiny details during my previous readthrough. Once they are fixed, it should be good to go, assuming the other reviewers are satisfied.

package/MDAnalysis/analysis/diffusionmap.py
AnirG: Apologies! I have updated the docstrings.

package/CHANGELOG
```diff
@@ -28,7 +28,7 @@ Deprecations
 
 03/06/22 IAlibay, melomcr, mdd31, ianmkenney, richardjgowers, hmacdope,
          orbeckst, scal444, p-j-smith, edisj, Atharva7K, ojeda-e, jbarnoud,
-         yangtcai
+         yangtcai, AnirG
```
IAlibay: Version 2.1.0 has now been released, you'll need to update your branch and add your changelog entry to 2.2.0 instead.

package/MDAnalysis/analysis/diffusionmap.py
```diff
@@ -152,6 +152,7 @@
 """
 import logging
 import warnings
+from MDAnalysis import AtomGroup
```
IAlibay: Imports are usually grouped as "python built-ins", "external libraries", "internal library", see the PEP8 guide for more details: https://peps.python.org/pep-0008/#source-file-encoding

Can you move this to after line 159 please?

IAlibay (changes requested):
Couple of extra things, you'll need to also resolve conflicts (which I think are related to the recent 2.1.0 -> 2.2.0 change)

package/CHANGELOG
AnirG: Sorry, didn't realize 2.1.0 was out already. I have revoked the changes and have updated on 2.2.0.

package/MDAnalysis/analysis/diffusionmap.py
```diff
@@ -328,12 +332,12 @@ def __init__(self, u, epsilon=1, **kwargs):
             Parameters to be passed for the initialization of a
             :class:`DistanceMatrix`.
         """
-        if isinstance(u, Universe):
+        if isinstance(u, AtomGroup) or isinstance(u, Universe):
```
IAlibay, line 344: Quick question here - what happens if you pass an UpdatingAtomGroup, does the code still work?

AnirG: yes, it works fine. PFA example code snippet I tried on.
[example.zip](https://github.com/MDAnalysis/mdanalysis/files/8323412/example.zip)

package/CHANGELOG
```diff
@@ -65,7 +67,7 @@ Enhancements
   * Added MDAnalysis.lib.distances.minimize_vectors
 
 Changes
-  * `packaging` is now a core dependency for MDAnalysis
+ * `packaging` is now a core dependency for MDAnalysis
```
IAlibay: Not sure why this white space change exists.
```suggestion
  * `packaging` is now a core dependency for MDAnalysis
```

package/MDAnalysis/analysis/diffusionmap.py
```diff
@@ -248,7 +250,8 @@ class DistanceMatrix(AnalysisBase):
     def __init__(self, universe, select='all', metric=rmsd, cutoff=1E0-5,
                  weights=None, **kwargs):
         # remember that this must be called before referencing self.n_frames
-        super(DistanceMatrix, self).__init__(universe.trajectory, **kwargs)
+        super(DistanceMatrix, self).__init__(universe.universe.trajectory,
```
IAlibay, line 253: Do add a versionchanged entry after line 247.

package/MDAnalysis/analysis/diffusionmap.py
```diff
@@ -315,7 +318,8 @@ def __init__(self, u, epsilon=1, **kwargs):
         """
         Parameters
         -------------
-        u : MDAnalysis Universe or DistanceMatrix object
+        u : MDAnalysis Universe or MDAnalysis AtomGroup or
```
IAlibay: Do add a versionchanged entry after line 314.

package/MDAnalysis/analysis/diffusionmap.py
IAlibay: Could you add in a test that uses an UpdatingAtomGroup?

IAlibay (changes requested):
Two documentation things and a test request.

richardjgowers (approved):

package/MDAnalysis/analysis/diffusionmap.py
AnirG: sure, will add a separate test for updating atomgroups.

package/MDAnalysis/analysis/diffusionmap.py
```diff
@@ -303,20 +307,24 @@ class DiffusionMap(object):
 
     Methods
     -------
-    run()
+    run()line
```
IAlibay: typo?

package/MDAnalysis/analysis/diffusionmap.py
```diff
@@ -303,20 +307,24 @@ class DiffusionMap(object):
...
         Constructs an anisotropic diffusion kernel and performs eigenvalue
         decomposition on it.
     transform(n_eigenvectors, time)
         Perform an embedding of a frame into the eigenvectors representing
         the collective coordinates.
+    
+    .. versionchanged: 2.2.0
```
IAlibay: See here: https://mdanalysis--3541.org.readthedocs.build/en/3541/documentation_pages/analysis/diffusionmap.html#MDAnalysis.analysis.diffusionmap.DiffusionMap.transform

You'll need two empty lines for this to work as intended.
```suggestion

    .. versionchanged: 2.2.0
```

package/MDAnalysis/analysis/diffusionmap.py
```diff
@@ -303,20 +307,24 @@ class DiffusionMap(object):
...
+         :class: `DiffusionMap` now also accepts `AtomGroup`.
```
IAlibay: typo - extra space between class and classname
```suggestion
         :class:`DiffusionMap` now also accepts `AtomGroup`.
```

IAlibay (changes requested):
Couple of small extra things. Not sure why CI isn't triggering properly at the moment.

package/MDAnalysis/analysis/diffusionmap.py
AnirG: yeah my bad :(

package/MDAnalysis/analysis/diffusionmap.py
AnirG: hi, I'm not very sure why it still doesn't work as intended even after giving 2 empty lines.

package/MDAnalysis/analysis/diffusionmap.py
AnirG: Hi, I'm not very sure about what kind of test do we want here. Do I make a test out of the example.zip that I attached before and compare the DistMatrix constructed by updating atom group with just universe and selection string of the atom group?

AnirG: @IAlibay I did the above for updating AG checks. Please let me know if it is fine.

testsuite/MDAnalysisTests/analysis/test_diffusionmap.py
```diff
@@ -69,6 +69,30 @@ def test_dist_weights(u):
...
+    dist_universe = diffusionmap.DistanceMatrix(u, select='backbone').run()
+    ag = u.select_atoms('backbone')
+    dist_ag = diffusionmap.DistanceMatrix(ag).run()
+    assert_allclose(dist_universe.results.dist_matrix,
+                    dist_ag.results.dist_matrix)
+
+
+def test_distvalues_updating_ag(u):
+    selection = 'resname ALA'
+    ag = u.select_atoms(selection, updating=True)
```
IAlibay: I'm not really understanding how this checks that an updating selection works, maybe I'm missing something obvious though.

Since universes are static, there shouldn't be fewer or more atoms that match "resname ALA" within the lifetime of a trajectory. Instead of you should aim to analyse something dynamic, like "ALA residues within a given distance of residue N".

IAlibay (changes requested):

testsuite/MDAnalysisTests/analysis/test_diffusionmap.py
AnirG: yes my bad, I didn't give much attention to the atom selection. I was more concerned about if it worked when AG is an instance of UpdatingAtomGroup. I will update it such that selection is dynamic over all frames and the final selection is parsed over to `DistanceMatrix`. In my opinion, this test case is kinda redundant as we are checking if it's working for the final AG after all.

IAlibay: So the point of passing and UpdatingAtomGroup through and checking that everything still works as expected is because most Analysis methods in MDAnalysis use static defined numpy arrays. So if you end up with more or fewer entries than you'd expect, things can go wrong. Having a test that we can check works as expected is very important.

AnirG: @IAlibay I have a doubt here. Say, you make a selection as `selection = "around 5 resname ALA"`. For frame 1, let's assume this selection has 10 atoms, and for frame 50, it has 20 atoms. Do you expect RMSD(frame 1, frame 50) and RMSD(frame 50, frame 1) in the `DistanceMatrix` to be different since the selection is being updated dynamically due to the UpdatingAtomGroup? If this is what was implied here  https://github.com/MDAnalysis/mdanalysis/pull/3541/files/2799ad9cf2b9f12bf673a217985d781535f802a9#r831565393 , I believe it doesn't work this way. 
What can be done is package all these dynamic selections into one atomgroup and run our `DistanceMatrix` analysis on this AG (finally a static selection).

AnirG: @IAlibay Is my understanding correct?

IAlibay: @AnirG this is essentially the crux of what I was leading to with my original question on this, I was trying to make you think about the cases where a user passed in an UpdatingAtomgroup expecting that the there would be more / fewer atoms depending on the frames - would things work as expected?

Because of the way `metric` works, there is an expectation that the `i_ref` and `j_ref` arrays are of equal lengths (see https://github.com/MDAnalysis/mdanalysis/blob/009c467511210c442d15e4014f02e62e07dc9e85/package/MDAnalysis/analysis/diffusionmap.py#L185-L191), so as you observed above, this is unlikely to work.

It's fine in this case for the solution to just be "UpdatingAtomGroups are not accepted". Creating a complex solution involving the eventual generation of a static selection is probably not worth it.

The next steps here are then - what code changes are necessary here to avoid users passing UpdatinAtomGroups?

IAlibay: As @lilyminium just reminded me, https://github.com/MDAnalysis/mdanalysis/blob/e6f0fe1289d150747befd1f1773db5657760f2c5/package/MDAnalysis/analysis/diffusionmap.py#L256 essentially means that all UpdatingAtomGroups will be squashed to AtomGroups on class initilisation. Probably all you need here is to have a check on the input to see if it's an UpdatingAtomGroup and warn users that it is being converted to a static AtomGroup based on the current frame selection.

AnirG: @IAlibay thanks for clarifying, I'll make the necessary changes.

package/MDAnalysis/analysis/diffusionmap.py
```diff
@@ -243,12 +244,20 @@ class DistanceMatrix(AnalysisBase):
...
     def __init__(self, universe, select='all', metric=rmsd, cutoff=1E0-5,
                  weights=None, **kwargs):
         # remember that this must be called before referencing self.n_frames
-        super(DistanceMatrix, self).__init__(universe.trajectory, **kwargs)
+        super(DistanceMatrix, self).__init__(universe.universe.trajectory,
+                                             **kwargs)
+
+        if isinstance(universe, UpdatingAtomGroup):
+            wmsg = ("U must be a static AtomGroup. Parsing an updating AtomGroup "
+                    "will result in a static AtomGroup with the first frame "
```
IAlibay: Please double check this, but I'm pretty sure it'll freeze it at the current frame, so if someone did:

```python
u.trajectory[10]
uag = u.select_atoms('resname HOH and within 4 resname LIG', updating=True)
dmatrix = DistanceMatrix(ag)
```

Then you're doing uag.select_atoms('all') on the current frame, rather than the first frame.

```suggestion
                    "will result in a static AtomGroup with the current frame "
```

IAlibay (changes requested):
Just the one last thing, otherwise everything looks good to me thanks!

IAlibay:
Please do fix merge conflicts too.

IAlibay:
@jbarnoud if you can also have a re-review that'd be great

package/MDAnalysis/analysis/diffusionmap.py
AnirG: yes. you're right. It's the current frame.

IAlibay (approved):
lgtm, thanks for all the work @AnirG.

Just waiting on @jbarnoud to re-review.

jbarnoud (changes requested):
Sorry, I'll ask for one last small test, and it is all good.

Users may provide both an AtomGroup and the select argument. From what I read, it should work well with your implementation, but I would like to see a test for it. That test does not have to be complicated, just to use both an AtomGroup and the select argument. I would write it as the same as `test_distvalues_ag_universe` but with 

```python
ag = u.select_atoms('protein'')
dist_ag = diffusionmap.DistanceMatrix(ag, select='backbone').run()
```

jbarnoud:
Thank you for having taken care of the conflicts as well. I'll merge as soon as the CI is done and passing :)

AnirG:
@IAlibay @jbarnoud thanks a lot for your patience and guidance :) felt quite satisfying starting out on this. Will hopefully solve more issues as I use the package more and get a better understanding over time.

jbarnoud (approved):
:tada:

jbarnoud:
Congratulations! Thank you for addressing the issue :D

PR #3544, opened 2022-02-23
Add optional dtype argument to box routines

### LINKED ISSUE #3500: check_box and triclinic_vectors don't respect dtype
Both of these functions make a lot of assumptions about input dtype, and often "squash" down to float32.  These should instead preserve and maintain whatever precision they are given.

hmacdope:
Fixes #3500

 - Add an optional dtype argument to the box routines. 

 As we don't guarantee the input is an `ndarray` we can't pass the dtype on from the input array.  Regardless the compulsory squash to np.float32 has been removed and a dtype can be specified if required.

The previous behaviour is still preserved with `np.float32` being the default. 

PR Checklist
------------
 - [x] Tests?
 - [X] Docs?
 - [x] CHANGELOG updated?


richardjgowers (approved):
Looks like progress toward multi precision

IAlibay (changes requested):
test please

hmacdope:
> test please

sorry should have made draft

package/CHANGELOG
```diff
@@ -20,6 +20,8 @@ The rules for this file:
  * 2.1.0
 
 Fixes
+  * Add optional dtype argument to box checking routines to avoid forced
```
jbarnoud: This needs to move to 2.2.0.

jbarnoud (commented):
Could there be a test to make sure the type of the positions is not affected when a method like `AtomGroup.wrap`, that uses both positions and the box, is called?

hmacdope:
> Could there be a test to make sure the type of the positions is not affected when a method like `AtomGroup.wrap`, that uses both positions and the box, is called?

can do! I'll fix changelog also.

hmacdope:
@jbarnoud you make a really good point in that if an `np.float64` box is passed, any mixed precision box-coordinate operations can promote the result to `np.float64`, which may have flow on effects. I will investigate further.

hmacdope:
This is basically dependent on decisions in #3707

hmacdope:
Given discussion on #3545 #3707 and #3797 I think there is some support gathering to push forward on this. I will add the tests that @jbarnoud was asking for and we can go from there.

hmacdope:
It seems you can set the coordinates to be `np.float64` at least for the memoryreader, ill dig into it.

IAlibay:
Seems like tests were still broken when we last got to a review here, what needs to get done to get this merged @hmacdope ?

hmacdope:
@IAlibay @tylerjereddy I think the issue was thaty you can get coordinate array promotion, I will investigate.

Issue #3550, opened 2022-03-03
CI: macos runners segfault failing on openmm tests


IAlibay:
## Expected behavior ##

All CI tests work as expected.

## Actual behavior ##

CI is failing on macos runners with very unspecific segfaults.

See: https://github.com/MDAnalysis/mdanalysis/runs/5398923069?check_suite_focus=true as an example.

Apparently this might be related to: https://github.com/openmm/openmm/issues/3491

Although it is oddly timed with py3.10  openmm support having been added yesterday.

IAlibay:
Issue exists upstream: https://github.com/openmm/openmm/issues/3495

My next steps will be to pin to the 7.7.0 build 0 and see if that'll fix things temporarily. Hopefully a fix will come soon.

IAlibay:
build 1 has been marked as broken upstream

PR #3551, opened 2022-03-03
temporarily pin openmm to 7.7.0 build 0 for macos

IAlibay:
Temporary solution for #3550

 - Pin macos runners to 7.7.0 build 0 (build 1 is causing the failures). 

PR Checklist
------------
 - Tests?
 - Docs?
 - CHANGELOG updated?
 - Issue raised/referenced?

### DIFF
--- a/.github/workflows/gh-ci.yaml
+++ b/.github/workflows/gh-ci.yaml
@@ -52,7 +52,6 @@ jobs:
               full-deps: true
               openmm: openmm=7.7.0=py37hf0bdc88_0_khronos
               install_hole: true
-              openmm:
               codecov: false
             - name: minimal-ubuntu
               os: ubuntu-latest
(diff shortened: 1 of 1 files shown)


IAlibay:
@MDAnalysis/coredevs please prioritise for merge, it _should_ be the only thing left that's blocking getting 2.1.0 done.

jbarnoud:
I cannot tell if openmm is actually being installed in the non-macos jobs. I do not see it in the logs of the ubuntu jobs I looked at.

IAlibay:
> I cannot tell if openmm is actually being installed in the non-macos jobs. I do not see it in the logs of the ubuntu jobs I looked at.

How strange, I might have messed up the defaults on the action, should be easy enough to fix.

IAlibay:
That should do the trick, I always forget that defaults don't get triggered if you just pass nothing through (seems like it's only a thing if you don't specify the argument..)

IAlibay:
Maybe hold for a couple of hours, looks like they might pull build 1 from conda-forge.

jbarnoud:
The broken build is now tagged as such on conda-forge :tada:

PR #3563, opened 2022-03-15
BAT method Cartesian modifies input data issue resolved

### LINKED ISSUE #3501: BAT method Cartesian modifies input data
## Expected behavior ##

When calling the `Cartesian` method of a BAT class instance to convert BAT coordinates back to Cartesian, the input BAT data should not be modified.

## Actual behavior ##

"Non-primary" torsions in the original input BAT data are modified due to issues with scope and pass-by-reference.  The very easy fix is just to call `copy.deepcopy()` when defining the torsions from the input data here: https://github.com/MDAnalysis/mdanalysis/blob/4a6aee43515db6b954281efdf6d7092ec58452e2/package/MDAnalysis/analysis/bat.py#L502

``` python
import copy
import numpy as np

import MDAnalysis as mda
from MDAnalysis.analysis import bat

#Just load in a simple test molecule - you can use any molecule as long as it has non-primary torsions
prmtopFile = 'alanine-dipeptide.prmtop'
strucFile = 'alanine-dipeptide.pdb'
uni = mda.Universe(prmtopFile, strucFile)

#Create BAT analysis and convert input coordinates to BAT, noting that we deep copy the data
bat_analysis = bat.BAT(uni.select_atoms('all'))
bat_analysis.run()
bat_coords = copy.deepcopy(bat_analysis.bat[0])
print('Original BAT coordinates:\n', bat_coords)

#Now run test to convert back to Cartesian
print('Original Cartesian coordinates:\n', uni.trajectory[0].positions)
print('First transformation back to Cartesian:\n', bat_analysis.Cartesian(bat_coords))

#It works the first time, but let's check bat_coords and try again...
print('Modified BAT coordinates after one transformation:\n', bat_coords)
print('Transforming again is now incorrect (obviously):\n', bat_analysis.Cartesian(bat_coords))
print('Non-primary torsions keep having primaries added to them:\n', bat_coords)

```
[Dialanine.zip](https://github.com/MDAnalysis/mdanalysis/files/7840872/Dialanine.zip)

2.0.0-dev0 (looking at https://github.com/MDAnalysis/mdanalysis/blob/develop/package/MDAnalysis/analysis/bat.py it looks like the current code will still have the same issue)
python 3.7
BigSur, Centos7

Neel-Shah-29:
Fixes #3501 

 - "Non-primary" torsions in the original input BAT data are modified due to issues with scope and pass-by-reference.I have just called the `copy.deepcopy()` method to resolve this error.

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/analysis/bat.py
+++ b/package/MDAnalysis/analysis/bat.py
@@ -501,7 +501,7 @@ class BAT(AnalysisBase):
         bonds = bat_frame[9:n_torsions + 9]
         angles = bat_frame[n_torsions + 9:2 * n_torsions + 9]
         torsions = bat_frame[2 * n_torsions + 9:]
-        torsions=copy.deepcopy(torsions)
+        torsions = copy.deepcopy(torsions)
         # When appropriate, convert improper to proper torsions
         shift = torsions[self._primary_torsion_indices]
         shift[self._unique_primary_torsion_indices] = 0.
--- a/package/MDAnalysis/analysis/bat.py
+++ b/package/MDAnalysis/analysis/bat.py
@@ -501,7 +501,7 @@ class BAT(AnalysisBase):
         bonds = bat_frame[9:n_torsions + 9]
         angles = bat_frame[n_torsions + 9:2 * n_torsions + 9]
         torsions = bat_frame[2 * n_torsions + 9:]
-        copy.deepcopy()
+        torsions=copy.deepcopy(torsions)
         # When appropriate, convert improper to proper torsions
         shift = torsions[self._primary_torsion_indices]
         shift[self._unique_primary_torsion_indices] = 0.
--- a/package/MDAnalysis/analysis/bat.py
+++ b/package/MDAnalysis/analysis/bat.py
@@ -173,8 +173,9 @@ References
 """
 import logging
 import warnings
-import copy
+
 import numpy as np
+import copy
 
 import MDAnalysis as mda
 from .base import AnalysisBase
--- a/package/MDAnalysis/analysis/bat.py
+++ b/package/MDAnalysis/analysis/bat.py
@@ -173,7 +173,7 @@ References
 """
 import logging
 import warnings
-
+import copy
 import numpy as np
 
 import MDAnalysis as mda
@@ -500,6 +500,7 @@ class BAT(AnalysisBase):
         bonds = bat_frame[9:n_torsions + 9]
         angles = bat_frame[n_torsions + 9:2 * n_torsions + 9]
         torsions = bat_frame[2 * n_torsions + 9:]
+        copy.deepcopy()
         # When appropriate, convert improper to proper torsions
         shift = torsions[self._primary_torsion_indices]
         shift[self._unique_primary_torsion_indices] = 0.
--- a/testsuite/MDAnalysisTests/analysis/test_bat.py
+++ b/testsuite/MDAnalysisTests/analysis/test_bat.py
@@ -127,3 +127,8 @@ class TestBAT(object):
         errmsg = 'Dimensions of array in loaded file'
         with pytest.raises(ValueError, match=errmsg):
             R = BAT(selected_residues, filename=bat_npz)
+
+    def test_bat_input_data(self,torsions,bat_frame):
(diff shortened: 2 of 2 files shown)


package/MDAnalysis/analysis/bat.py
```diff
@@ -500,6 +501,7 @@ def Cartesian(self, bat_frame):
         bonds = bat_frame[9:n_torsions + 9]
         angles = bat_frame[n_torsions + 9:2 * n_torsions + 9]
         torsions = bat_frame[2 * n_torsions + 9:]
+        copy.deepcopy()
```
tylerjereddy: A few issues here:

1) you don't actually provide an argument to `deepcopy()`, so it isn't going to copy anything
2) you don't capture the return value, so even if you did deepcopy something, it wouldn't do anything useful in the control flow
3) There is no regression test for the issue--you could use something like `np.may_share_memory()` to check for input/output pass by reference issues for example.

Neel-Shah-29: Hey, According to third point as you said i need to add a regression test , So can you suggest in which file should i add the test regarding it. As i am a beginner i dont know the codebase, it would be great if you could help me with it. Thanks!

Neel-Shah-29:
Kindly review the PR and please suggest the solution to pass[ this](https://github.com/MDAnalysis/mdanalysis/pull/3563/checks?check_run_id=5566549363) check.

package/MDAnalysis/analysis/bat.py
jbarnoud: The tests for that modules are here: <https://github.com/MDAnalysis/mdanalysis/blob/develop/testsuite/MDAnalysisTests/analysis/test_bat.py>. That would be a good place for your tests.

IAlibay:
> Kindly review the PR and please suggest the solution to pass[ this](https://github.com/MDAnalysis/mdanalysis/pull/3563/checks?check_run_id=5566549363) check.

Codecov project checks failing are normal and non-blocking, they usually happen when you add extra lines.

As detailed by others, you'll need to add some tests before we can review. See https://github.com/MDAnalysis/mdanalysis/pull/3563#discussion_r827747411 for more details.

Neel-Shah-29:
Hey, I am not able to figure out what regression test that can be added to check the `copy.deepcopy()` function. I tried implementing it with `np.may_share_memory()` but the checks didn't pass. Can someone please guide me through how to write the regression test .

jbarnoud:
When writing a test, *think about what problem your PR is fixing for the user*. Your test should make sure the issue is fixed so future changes will have the test fail if they make the problem occur again.

IAlibay:
@Neel-Shah-29 - this PR exceeds our stale PR policy for gsoc-starter issues by ~ 10 days (see: https://github.com/MDAnalysis/mdanalysis/wiki/GSoC-FAQ#closing-stale-prs). Are you still planning to work on this? Is there any further information we can provide here?

testsuite/MDAnalysisTests/analysis/test_bat.py
```diff
@@ -127,3 +127,8 @@ def test_bat_incorrect_dims(self, bat_npz):
         errmsg = 'Dimensions of array in loaded file'
         with pytest.raises(ValueError, match=errmsg):
             R = BAT(selected_residues, filename=bat_npz)
+
+    def test_bat_input_data(self,torsions,bat_frame):
+        arr1 = np.array(torsions)
+        arr2 = np.array(bat_frame)
+        return np.may_share_memory(arr1,arr2)
```
orbeckst, line 134: The test does not directly test the problem described in #3501 . 

I would look at the original issue report and see if I can turn the report into a test. Your test should _fail_ with the code before your fix and pass with your fix.

orbeckst (changes requested):
The test needs to be specific for the issue.

orbeckst:
There hasn't been activity on this PR for 5 days since https://github.com/MDAnalysis/mdanalysis/pull/3563#issuecomment-1083099487 so in accordance with our GSOC stale PR policy I am closing the PR.

PR #3564, opened 2022-03-16
Issue #2417: ITPParser should not add angles to SETTLEd waters. 

### LINKED ISSUE #2417: Remove angle from [ settles ] constraints in ITPParser
**Expected behavior**

~If `[ settles ]` is a triangular constraint, shouldn't it add an angle as well as two bonds?~ **Edit: ITPParser only adds two bonds for SETTLE water, the way AMBER treats them.** 

**Actual behavior**

~TPRParser only adds two bonds.~

ITPParser adds two bonds *and* an angle.

https://github.com/MDAnalysis/mdanalysis/blob/3e8ea8e0744f3163ddbf7b4f5f66874dd26796a7/package/MDAnalysis/topology/ITPParser.py#L349-L365

~One or the other should change for consistent behaviour. Personally I think adding an angle makes sense, even if it is not very useful.~ It shouldn't add an angle.

There are existing tests that read `[ settles ]` directives, so changing the number of angles expected in those would suffice to test any solution to this issue.

Develop branch
- Which operating system?

BFedder:
Fixes #2417

 ITPParser no longer adds angles for water molecules with SETTLEd
   geometry. The tests were modified to reflect this.

Changes:
**MDAnalysis/topology/ITPParser.py**
- Molecule.parse_settles changed to no longer add angle

**MDAnalysisTests/topology/test_itp.py**
- TestITPNoKeywords now expects 0 angles
- TestITPKeywords changed to continue to expect 1 angle
- def test_angles_values moved from TestITPNoKeywords to TestITPKeywords
- TestNestedIfs now expects 0 angles
- TestReadTop's topology file has 3 SETTLEd water molecules and now correctly expects three fewer angles than before

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -22,6 +22,9 @@ Fixes
 Enhancements
 
 Changes
+  * ITPParser no longer adds an angle for water molecules that have the SETTLE
+    constraint applied to them. This mirrors the way the constraint is handled
+    in AMBER. (Issue #2417, PR #)
 
 Deprecations
 
@@ -36,7 +39,7 @@ Fixes
   * Use uint64_t loop counters in C level distance functions to avoid overflow
     for large arrays (Issue #3512, PR #3513).
   * Prevents attempts to close an already closed NamedStream (Issue #3386)
-  * Added missing exe_err import in hole.py. Fixed a bug in HoleAnalysis where 
+  * Added missing exe_err import in hole.py. Fixed a bug in HoleAnalysis where
     kwarg was passed instead of object to generate path (Issue #3493, PR #3496)
   * Fixes creation of vmd surfaces in hole2 when using non-contiguous frames
     with start/stop/step (Issue #3476)
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -24,7 +24,7 @@ Enhancements
 Changes
   * ITPParser no longer adds an angle for water molecules that have the SETTLE
     constraint applied to them. This mirrors the way the constraint is handled
-    in AMBER. (Issue #2417, PR #)
+    in AMBER. (Issue #2417, PR #3564)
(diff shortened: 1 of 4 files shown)


orbeckst (commented):
This is already looking very good. Can you please look at the PEP8 comments?

I'd like to have @jbarnoud and/or @lilyminium have a look, too.

If you haven't done so already, please introduce yourself on the [developer mailing list](https://groups.google.com/g/mdnalysis-devel) so that we have a current email address from you. Thank you!

package/CHANGELOG
```diff
@@ -22,6 +22,9 @@ Fixes
 Enhancements
 
 Changes
+  * ITPParser no longer adds an angle for water molecules that have the SETTLE
+    constraint applied to them. This mirrors the way the constraint is handled
+    in AMBER. (Issue #2417, PR #)
```
```suggestion
    in AMBER. (Issue #2417, PR #3564)
```

jbarnoud (approved):
Just a tiny fix for the changelog. Otherwise, looks good to me. This looks like a very clean fix.

jbarnoud (changes requested):
After a second look, it would be good to add a `.. versionchanged:` notice in the docstring of the `ITPParser` class. Just that small documentation fix and it is good to go.

BFedder:
Hello and thanks! 
@orbeckst, I have now addressed the PEP8 issues and introduced myself on the mailing list. 
@jbarnoud, thanks for the updated changelog - I have added the PR number now, and also added a `.. versionchanged::` note to `ITPParser`.

jbarnoud (approved):
Thank you! And congratulation on your first merged contribution :smile:

PR #3565, opened 2022-03-18
Provide ability to use AtomGroup as selection for Contact

### LINKED ISSUE #2666: Provide AtomGroup selections to Contacts
## Feature request ##
_Originally posted by @orbeckst in https://github.com/MDAnalysis/mdanalysis/issues/2368#issuecomment-545079547_

* Add AtomGroup functionality:
  - It is  clunky that I have to provide *selections* instead of AtomGroups for `Contacts(..., selections=...))` to indicate which groups the contacts are formed between. I should be able to use static AtomGroups (cannot be dynamic). See #2369 

* Tidy up the helper function:
  - The https://github.com/MDAnalysis/mdanalysis/blob/9372d093acebb78bc5b1a2b4dba815a4e3281b04/package/MDAnalysis/analysis/contacts.py#L464 helper function can just use `Universe.copy()`.

shudipto-amin:
Fixes #2666

Summary
Solves the GSoC starter issue [#2666](https://github.com/MDAnalysis/mdanalysis/issues/2666)

 - the class `mdanalysis/package/MDAnalysis/analysis/contacts.py::Contacts` is modified to accept static AtomGroups
 - the helper function `_new_selections()` in the same script is modified to use `Universe.copy()` **UPDATE: this has now been raised in new Issue #3570. This PR has reverted it back to the original.**
 - tests were added to `test_contacts.py` to ensure Contacts successfully uses static AtomGroups as well as selection texts.
 - tests were also added to ensure it fails when wrong types are provided for selection, including updating AtomGroups.

PR Checklist
------------
 - [x] Codecov coverage?
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -13,14 +13,16 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" http://semver.org
 
 ------------------------------------------------------------------------------
-??/??/?? IAlibay, BFedder, inomag, Agorfa, aya9aladdin
+??/??/?? IAlibay, BFedder, inomag, Agorfa, aya9aladdin, shudipto-amin
 
  * 2.2.0
 
 Fixes
   * Fixed DumpReader triclinic box dimensions reading. (Issue #3386, PR #3403)
   * Updated the badge in README of MDAnalysisTest to point to Github CI Actions
-
+  * Provides Contacts ability to use static AtomGroups as well as selection 
+    texts.
+  
 Enhancements
 
 Changes
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -22,7 +22,7 @@ Fixes
     (Issue #2666, PR #3565)
   * Fixed DumpReader triclinic box dimensions reading. (Issue #3386, PR #3403)
   * Updated the badge in README of MDAnalysisTest to point to Github CI Actions
-  
+ 
 Enhancements
 
 Changes
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -22,6 +22,10 @@ Fixes
   * Updated the badge in README of MDAnalysisTest to point to Github CI Actions
   * Provides Contacts ability to use static AtomGroups as well as selection 
     texts.
+  * Contacts can now accepts static AtomGroup as well as selection text. 
+    (Issue #2666, PR #3565)
+  * Helper function  `_new_selections()` in `contacts.py` now simply uses 
+    `Universe.copy()`. (Issue #2666, PR #3565)   
   
 Enhancements
 
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -22,7 +22,7 @@ Fixes
   * Updated the badge in README of MDAnalysisTest to point to Github CI Actions
   * Provides Contacts ability to use static AtomGroups as well as selection 
     texts.
-  * Contacts can now accepts static AtomGroup as well as selection text. 
+  * Contacts can now accept static AtomGroups as well as selection texts.
     (Issue #2666, PR #3565)
   * Helper function  `_new_selections()` in `contacts.py` now simply uses 
     `Universe.copy()`. (Issue #2666, PR #3565)   
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -18,14 +18,12 @@ The rules for this file:
  * 2.2.0
 
 Fixes
+  * Contacts can now accept static AtomGroups as well as selection texts.
+    (Issue #2666, PR #3565)
   * Fixed DumpReader triclinic box dimensions reading. (Issue #3386, PR #3403)
   * Updated the badge in README of MDAnalysisTest to point to Github CI Actions
   * Provides Contacts ability to use static AtomGroups as well as selection 
     texts.
-  * Contacts can now accept static AtomGroups as well as selection texts.
-    (Issue #2666, PR #3565)
-  * Helper function  `_new_selections()` in `contacts.py` now simply uses 
-    `Universe.copy()`. (Issue #2666, PR #3565)   
   
 Enhancements
 
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -22,8 +22,6 @@ Fixes
     (Issue #2666, PR #3565)
   * Fixed DumpReader triclinic box dimensions reading. (Issue #3386, PR #3403)
   * Updated the badge in README of MDAnalysisTest to point to Github CI Actions
-  * Provides Contacts ability to use static AtomGroups as well as selection 
-    texts.
   
 Enhancements
 
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -22,7 +22,7 @@ Fixes
     (Issue #2666, PR #3565)
   * Fixed DumpReader triclinic box dimensions reading. (Issue #3386, PR #3403)
   * Updated the badge in README of MDAnalysisTest to point to Github CI Actions
- 
+
 Enhancements
 
 Changes
--- a/package/MDAnalysis/analysis/contacts.py
+++ b/package/MDAnalysis/analysis/contacts.py
@@ -386,6 +386,8 @@ class Contacts(AnalysisBase):
     .. versionchanged:: 2.0.0
        :attr:`timeseries` results are now stored in a
        :class:`MDAnalysis.analysis.base.Results` instance.
+    .. versionchanged:: 2.2.0
+       :class: `Contacts` accepts both AtomGroup and string for `select`
     """
     # Error message for wrong selection type:
     _select_error_message = ("selection must be either string or a "
--- a/package/MDAnalysis/analysis/contacts.py
+++ b/package/MDAnalysis/analysis/contacts.py
@@ -389,7 +389,7 @@ class Contacts(AnalysisBase):
     """
     # Error message for wrong selection type:
     select_error_message = "selection must be either string or a " \
-                                    "static AtomGroup"
+                           "static AtomGroup"
 
     def __init__(self, u, select, refgroup, method="hard_cut", radius=4.5,
                  pbc=True, kwargs=None, **basekwargs):
--- a/package/MDAnalysis/analysis/contacts.py
+++ b/package/MDAnalysis/analysis/contacts.py
@@ -28,7 +28,7 @@ Native contacts analysis --- :mod:`MDAnalysis.analysis.contacts`
 This module contains classes to analyze native contacts *Q* over a
 trajectory. Native contacts of a conformation are contacts that exist
 in a reference structure and in the conformation. Contacts in the
-reference structure are always defined as being closer then a distance
+reference structure are always defined as being closer than a distance
 `radius`. The fraction of native contacts for a conformation can be
 calculated in different ways. This module supports 3 different metrics
 listed below, as well as custom metrics.
@@ -394,7 +394,7 @@ class Contacts(AnalysisBase):
         ----------
         u : Universe
             trajectory
-        select : tuple(string, string)
+        select : tuple(AtomGroup, AtomGroup) | tuple(string, string)
             two contacting groups that change over time
         refgroup : tuple(AtomGroup, AtomGroup)
             two contacting atomgroups in their reference conformation. This
@@ -437,8 +437,22 @@ class Contacts(AnalysisBase):
             self.fraction_contacts = method
 
         self.select = select
-        self.grA = u.select_atoms(select[0])
-        self.grB = u.select_atoms(select[1])
+        
+        select_error_message = "selection must be either string or AtomGroup"
+
+        if isinstance(select[0], str):
+            self.grA = u.select_atoms(select[0])
+        elif isinstance(select[0], AtomGroup):
+            self.grA = select[0]
+        else:
+            raise TypeError(select_error_message)
+
+        if isinstance(select[1], str):
+            self.grB = u.select_atoms(select[1])
+        elif isinstance(select[1], AtomGroup):
+            self.grB = select[1]
+        else:
+            raise TypeError(select_error_message)
         self.pbc = pbc
         
         # contacts formed in reference
--- a/package/MDAnalysis/analysis/contacts.py
+++ b/package/MDAnalysis/analysis/contacts.py
@@ -389,10 +389,6 @@ class Contacts(AnalysisBase):
     .. versionchanged:: 2.2.0
        :class: `Contacts` accepts both AtomGroup and string for `select`
     """
-    # Error message for wrong selection type:
-    _select_error_message = ("selection must be either string or a "
-                             "static AtomGroup. Updating AtomGroups "
-                             "are not supported.")
 
     def __init__(self, u, select, refgroup, method="hard_cut", radius=4.5,
                  pbc=True, kwargs=None, **basekwargs):
@@ -473,15 +469,18 @@ class Contacts(AnalysisBase):
 
     @staticmethod
     def _get_atomgroup(u, sel):
+        select_error_message = ("selection must be either string or a "
+                                "static AtomGroup. Updating AtomGroups "
+                                "are not supported.")
         if isinstance(sel, str):
             return u.select_atoms(sel)
         elif isinstance(sel, AtomGroup):
             if isinstance(sel, UpdatingAtomGroup):
-                raise TypeError(Contacts._select_error_message)
+                raise TypeError(select_error_message)
             else:
                 return sel
         else:
-            raise TypeError(Contacts._select_error_message)
+            raise TypeError(select_error_message)
 
     def _prepare(self):
         self.results.timeseries = np.empty((self.n_frames, len(self.r0)+1))
--- a/package/MDAnalysis/analysis/contacts.py
+++ b/package/MDAnalysis/analysis/contacts.py
@@ -392,18 +392,6 @@ class Contacts(AnalysisBase):
                              "static AtomGroup. Updating AtomGroups "
                              "are not supported.")
 
-    @staticmethod
-    def _get_atomgroup(u, sel):
-        if isinstance(sel, str):
-            return u.select_atoms(sel)
-        elif isinstance(sel, AtomGroup):
-            if isinstance(sel, UpdatingAtomGroup):
-                raise TypeError(Contacts._select_error_message)
-            else:
-                return sel
-        else:
-            raise TypeError(Contacts._select_error_message)
-
     def __init__(self, u, select, refgroup, method="hard_cut", radius=4.5,
                  pbc=True, kwargs=None, **basekwargs):
         """
@@ -480,6 +468,18 @@ class Contacts(AnalysisBase):
                 self.r0.append(distance_array(refA.positions, refB.positions,
                                                 box=self._get_box(refA.universe)))
                 self.initial_contacts.append(contact_matrix(self.r0[-1], radius))
+
+    @staticmethod
+    def _get_atomgroup(u, sel):
+        if isinstance(sel, str):
+            return u.select_atoms(sel)
+        elif isinstance(sel, AtomGroup):
+            if isinstance(sel, UpdatingAtomGroup):
+                raise TypeError(Contacts._select_error_message)
+            else:
+                return sel
+        else:
+            raise TypeError(Contacts._select_error_message)
 
     def _prepare(self):
         self.results.timeseries = np.empty((self.n_frames, len(self.r0)+1))
--- a/package/MDAnalysis/analysis/contacts.py
+++ b/package/MDAnalysis/analysis/contacts.py
@@ -387,7 +387,7 @@ class Contacts(AnalysisBase):
        :attr:`timeseries` results are now stored in a
        :class:`MDAnalysis.analysis.base.Results` instance.
     .. versionchanged:: 2.2.0
-       :class: `Contacts` accepts both AtomGroup and string for `select`
+       :class:`Contacts` accepts both AtomGroup and string for `select`
     """
 
     def __init__(self, u, select, refgroup, method="hard_cut", radius=4.5,
--- a/package/MDAnalysis/analysis/contacts.py
+++ b/package/MDAnalysis/analysis/contacts.py
@@ -510,7 +510,7 @@ class Contacts(AnalysisBase):
 
 def _new_selections(u_orig, selections, frame):
     """create stand alone AGs from selections at frame"""
-    u = u_orig.copy()
+    u = MDAnalysis.Universe(u_orig.filename, u_orig.trajectory.filename)
     u.trajectory[frame]
     return [u.select_atoms(s) for s in selections]
 
--- a/package/MDAnalysis/analysis/contacts.py
+++ b/package/MDAnalysis/analysis/contacts.py
@@ -387,6 +387,10 @@ class Contacts(AnalysisBase):
        :attr:`timeseries` results are now stored in a
        :class:`MDAnalysis.analysis.base.Results` instance.
     """
+    # Error message for wrong selection type:
+    select_error_message = "selection must be either string or a " \
+                                    "static AtomGroup"
+
     def __init__(self, u, select, refgroup, method="hard_cut", radius=4.5,
                  pbc=True, kwargs=None, **basekwargs):
         """
@@ -437,27 +441,25 @@ class Contacts(AnalysisBase):
             self.fraction_contacts = method
 
         self.select = select
-        
-        select_error_message = "selection must be either string or a static AtomGroup"
 
         if isinstance(select[0], str):
             self.grA = u.select_atoms(select[0])
         elif isinstance(select[0], AtomGroup):
             if isinstance(select[0], UpdatingAtomGroup):
-                raise TypeError(select_error_message)
+                raise TypeError(self.select_error_message)
             else:
                 self.grA = select[0]
         else:
-            raise TypeError(select_error_message)
+            raise TypeError(self.select_error_message)
 
         if isinstance(select[1], str):
             self.grB = u.select_atoms(select[1])
         elif isinstance(select[1], AtomGroup):
             if isinstance(select[1], UpdatingAtomGroup):
-                raise TypeError(select_error_message)
+                raise TypeError(self.select_error_message)
             self.grB = select[1]
         else:
-            raise TypeError(select_error_message)
+            raise TypeError(self.select_error_message)
 
         self.pbc = pbc
         
--- a/package/MDAnalysis/analysis/contacts.py
+++ b/package/MDAnalysis/analysis/contacts.py
@@ -218,7 +218,7 @@ import MDAnalysis
 import MDAnalysis.lib.distances
 from MDAnalysis.lib.util import openany
 from MDAnalysis.analysis.distances import distance_array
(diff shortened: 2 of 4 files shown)


testsuite/MDAnalysisTests/analysis/test_contacts.py
```diff
@@ -171,16 +171,64 @@ class TestContacts(object):
...
+        acidic = universe.select_atoms(self.sel_acidic)
+        basic = universe.select_atoms(self.sel_basic)
+
+        with_string = self._run_Contacts(universe)
+        with_atomgroup = self._run_Contacts(universe, sel_acidic=acidic, sel_basic=basic)
+       
+        assert_equal(with_atomgroup.grA, with_string.grA)
+        assert_equal(with_atomgroup.grB, with_string.grB) 
+
+    @pytest.mark.xfail
```
hmacdope: rather than using `xfail` it would be much better to use the `with pytest.raises()` construct with a match to the error message you are expecting (`select_error_message`).

testsuite/MDAnalysisTests/analysis/test_contacts.py
```diff
@@ -171,16 +171,64 @@ class TestContacts(object):
...
+    def test_select_updating_acid(self, universe):
+        uag = universe.select_atoms(self.sel_acidic, updating=True)
+        self._run_Contacts(universe, sel_acidic=uag)
+    
+    @pytest.mark.xfail
```
hmacdope: same here with pytest.raises

testsuite/MDAnalysisTests/analysis/test_contacts.py
```diff
@@ -171,16 +171,64 @@ class TestContacts(object):
...
+    def test_select_updating_basic(self, universe):
+        uag = universe.select_atoms(self.sel_basic, updating=True)
+        self._run_Contacts(universe, sel_basic=uag)
+
+    @pytest.mark.xfail
```
hmacdope: same here with pytest.raises

testsuite/MDAnalysisTests/analysis/test_contacts.py
```diff
@@ -171,16 +171,64 @@ class TestContacts(object):
     def universe():
         return mda.Universe(PSF, DCD)
 
-    def _run_Contacts(self, universe, start=None, stop=None, step=None, **kwargs):
+    def _run_Contacts(self, universe, sel_acidic=None, sel_basic=None, start=None, stop=None, step=None, **kwargs):
```
hmacdope: Can i just query why the choice of `sel_acidic` and `sel_basic`? perhaps I'm missing some context here.

package/CHANGELOG
```diff
@@ -13,6 +13,13 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" http://semver.org
 
 ------------------------------------------------------------------------------
+03/18/22 shudipto-amin
+
+Fixes Issue #2666
+* Contacts can now accept static AtomGroup as well as selection text.
+* helper function  `_new_selections()` in `contacts.py` no simply uses
```
hmacdope: `no` vs `now` typo

hmacdope (changes requested):
This looks like real progress! See my comments for changes :)

package/CHANGELOG
```diff
@@ -13,6 +13,13 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" http://semver.org
 
 ------------------------------------------------------------------------------
+03/18/22 shudipto-amin
```
IAlibay: Unfortunately this is not how the changelog should be updated.

1. Update your branch so you have the latest version of the changelog
2. Add yourself at the current line 23 (on the line starting with my username)
3. Put fixes as a bullet point under Fixes (currently at line 28).

shudipto-amin: Thanks for your instructions! I had some confusion about this part, which is clearer now. I updated the CHANGELOG accordingly, let me know if you have other corrections/suggestions.

testsuite/MDAnalysisTests/analysis/test_contacts.py
```diff
@@ -171,16 +171,64 @@ class TestContacts(object):
...
+        fails = True
+        for sel_acid, self_basic in selections:
+            try:
+                self._run_Contacts(universe, sel_acid=sel_acid, sel_basic=sel_basic)
+                fails = False
+                break
+            except:
+                pass
+        if fails:
+            raise TypeError
```
tylerjereddy: Perhaps a bit awkward to catch all/generic `Exception`s in a test and manually raise a `TypeError` inside the test. Perhaps following the guidance related to `pytest.raises()` will clear that up to test for the expected errors in the code proper, since manually raising an error in the testsuite isn't of much use to end users of course.

If some of the parameters fail and others succeed you can xfail on a case-by-case basis as well.

testsuite/MDAnalysisTests/analysis/test_contacts.py
shudipto-amin: Thanks for all your comments! Let me answer this while I work on implementing pystest.raises(). 
I initially thought of using `ag_acidic` and `ag_basic`, where `ag` is for AtomGroup, but then I realized it doesn't have to be an AtomGroup, and can be a selection text. I thought `sel` would be more general. However, now I realize that it can be easily confused with `self.sel_basic` and `self.sel_acidic`. Should I change the variable names to `ag_*` instead of `sel_*`?

testsuite/MDAnalysisTests/analysis/test_contacts.py
shudipto-amin: This is indeed a better approach! I updated the testing code to use `pytest.raises()`, and `pytest.mark.parametrize` to combine all expected failure tests in one function. As for referring to the `select_error_message`, I updated `contacts.Contacts` class to define it as a static attribute, and referred to it in `test_contacts.py` using `with pytest.raises(TypeError, match=contacts.Contacts.select_error_message)`. Is that okay, or is there a better way?

testsuite/MDAnalysisTests/analysis/test_contacts.py
hmacdope: yeah that makes sense.

testsuite/MDAnalysisTests/analysis/test_contacts.py
```diff
@@ -171,16 +171,61 @@ class TestContacts(object):
...
+        ]
+    )
+    def test_select_wrong_types(self, universe, ag1, ag2):
+        """test that Contacts fails if selection types are wrong"""
+        if ag1 == "USE UPDATING AG":
+            ag1 = universe.select_atoms(self.sel_acidic, updating=True)
+        if ag2 == "USE UPDATING AG":
+            ag2 = universe.select_atoms(self.sel_basic, updating=True)
+
+        sel_err_msg = contacts.Contacts.select_error_message
```
hmacdope: No need to duplicate here, can just use `contacts.Contacts.select_error_message` directly in the `pytest.raises expression`

hmacdope:
@shudipto-amin looking good! Would you be able to resolve all the comments you have addressed so I can see what is still outstanding?

testsuite/MDAnalysisTests/analysis/test_contacts.py
shudipto-amin: Thanks for you comment! I updated the code to use `@pytest.mark.parametrization` decorator, and I believe that should bring it in line with the style of `contacts.py`. I'm a bit uncomfortable with how I had to use the decorator, but maybe that's okay? Should we resolve?

package/MDAnalysis/analysis/contacts.py
```diff
@@ -387,14 +387,18 @@ class Contacts(AnalysisBase):
        :attr:`timeseries` results are now stored in a
        :class:`MDAnalysis.analysis.base.Results` instance.
     """
+    # Error message for wrong selection type:
+    select_error_message = "selection must be either string or a " \
+                                    "static AtomGroup"
```
IAlibay: Please make sure you also review the other PEP8 issues that pep8speaks has highlighted: https://github.com/MDAnalysis/mdanalysis/pull/3565#issuecomment-1072770416

package/CHANGELOG
```diff
@@ -13,13 +13,19 @@ The rules for this file:
...
 ------------------------------------------------------------------------------
-??/??/?? IAlibay, BFedder, inomag
+??/??/?? IAlibay, BFedder, inomag, shudipto-amin
 
  * 2.2.0
 
 Fixes
   * Updated the badge in README of MDAnalysisTest to point to Github CI Actions
-
+  * Provides Contacts ability to use static AtomGroups as well as selection
```
IAlibay: Entries should be newest first in the changelog

package/MDAnalysis/analysis/contacts.py
```diff
@@ -387,14 +387,18 @@ class Contacts(AnalysisBase):
        :attr:`timeseries` results are now stored in a
        :class:`MDAnalysis.analysis.base.Results` instance.
     """
+    # Error message for wrong selection type:
+    select_error_message = "selection must be either string or a " \
+                           "static AtomGroup"
```
IAlibay: MDAnalysis tries to follow PEP8 where possible, as part of it, the recommendation is that line continuation should [use implicit line continuation where possible](https://peps.python.org/pep-0008/#:~:text=The%20preferred%20way%20of%20wrapping,a%20backslash%20for%20line%20continuation).

```suggestion
    select_error_message = ("selection must be either string or a "
                            "static AtomGroup")
```

package/MDAnalysis/analysis/contacts.py
```diff
@@ -437,8 +441,26 @@ def __init__(self, u, select, refgroup, method="hard_cut", radius=4.5,
...
+            raise TypeError(self.select_error_message)
+
+        if isinstance(select[1], str):
+            self.grB = u.select_atoms(select[1])
+        elif isinstance(select[1], AtomGroup):
+            if isinstance(select[1], UpdatingAtomGroup):
+                raise TypeError(self.select_error_message)
+            self.grB = select[1]
+        else:
+            raise TypeError(self.select_error_message)
```
IAlibay: There's a lot of repeated code here, let's see if we can condense it a little bit:

How about something like:

```python
@staticmethod
def _get_atomgroup(u, entry):
    if isinstance(entry, str):
        return u.select_atoms(entry)
    elif:
        if isinstance(entry, AtomGroup):
            if isinstance(entry, UpdatingAtomGroup):
                errmsg = "UpdatingAtomGroups are not supported"
                raise TypeError(errmsg)
            else:
                return entry
    else:
        raise TypeError(f"unrecognised selection type {entry}")

# ... and now back in __init__

self.grA, self.grB = [_get_atomgroup(var) for var in select]
```

This reduces the number of lines and also means you can directly test the staticmethod without having to go via class creation.

package/MDAnalysis/analysis/contacts.py
```diff
@@ -492,7 +514,7 @@ def timeseries(self):
 
 def _new_selections(u_orig, selections, frame):
     """create stand alone AGs from selections at frame"""
-    u = MDAnalysis.Universe(u_orig.filename, u_orig.trajectory.filename)
+    u = u_orig.copy()
```
IAlibay: This technically changes the behaviour of q1q2, you could consider a case where `u` a trajectory which has been transferred in-memory and offset by a certain value (let's say you've shifted all the atomic positions by 10 A). By doing a copy on that object you now have a new trajectory that has the same shift in the atomic positions, whilst the old code would have led to a brand new universe that doesn't have this shift in atomic positions.

Whilst the copy behaviour is the one we are likely to want, introducing new changes that break behaviour without announcing it to users is a bad idea, because _someone_ might actually be relying on this behaviour (even though it's broken).

I realise @lilyminium pointed to fixing this in the original issue, but I don't agree that we should make this without warning users beforehand. I suspect further discussion will be needed here, so in the interest of saving time I would ask that we just not address this at this time and revisit it in the original issue before / fix it in a future PR.

testsuite/MDAnalysisTests/analysis/test_contacts.py
```diff
@@ -171,16 +171,62 @@ class TestContacts(object):
...
-    def _run_Contacts(self, universe, start=None, stop=None, step=None, **kwargs):
+    def _run_Contacts(
+        self, universe,
+        ag_acidic=None, ag_basic=None,
+        start=None, stop=None, step=None, **kwargs
+    ):
         acidic = universe.select_atoms(self.sel_acidic)
         basic = universe.select_atoms(self.sel_basic)
+        if ag_acidic is None:
+            ag_acidic = self.sel_acidic
```
IAlibay: I see what you're doing here, but I think it actually makes things harder to read, I would suggestion creating a new fixture that's just for dealing with atomgroups.

testsuite/MDAnalysisTests/analysis/test_contacts.py
```diff
@@ -171,16 +171,62 @@ class TestContacts(object):
...
+        [
+            (1, 2),
+            ([1], [2]),
+            (mda.Universe, sel_basic),
+            (sel_acidic, 2),
+            ("USE UPDATING AG", sel_basic),
+            (sel_acidic, "USE UPDATING AG")
+        ]
+    )
+    def test_select_wrong_types(self, universe, ag1, ag2):
```
IAlibay: If going by my suggestion above, you can just directly do tests that run over the staticmethod instead here, it should be a lot simpler a parametrize.

IAlibay (changes requested):
Some comments, the way you've done it isn't wrong/bad, but it could be refined a little bit to make things easier long term.

package/MDAnalysis/analysis/contacts.py
shudipto-amin: I see your point. From what you say, it does sound like it would be best to make a separate issue/PR out of it, and revert it back to original here. So should I go ahead and create a new issue for it and reference it here (as well as in #2666), or leave it to core developers to raise it?

shudipto-amin:
> @shudipto-amin looking good! Would you be able to resolve all the comments you have addressed so I can see what is still outstanding?

@hmacdope Thanks for your comments! I have resolved those conversations with the latest commits.

package/CHANGELOG
shudipto-amin: Done, I put it at the top.

package/MDAnalysis/analysis/contacts.py
shudipto-amin: Fixed in commits made since this comment.

package/MDAnalysis/analysis/contacts.py
shudipto-amin: Done! I added your suggested method in the Contacts class.

testsuite/MDAnalysisTests/analysis/test_contacts.py
shudipto-amin: Done, all tests are done on new static method `Contacts._get_atomgroup()`

shudipto-amin:
@hmacdope @IAlibay @tylerjereddy 
Thank you all for your constructive comments! I learned a lot from them, and I believe I addressed all the concerns and the PR is ready or almost ready to merge. I just have a couple of questions:
1. According to [this discussion](https://github.com/MDAnalysis/mdanalysis/pull/3565#discussion_r831316421) with IAlibay, we should address the change to `universe.copy()` in a separate PR/Issue. Should I open a new issue for that?
2. The codecov report varies from time to time (without me pushing any new commits). Sometimes it shows the diff to be about -4%, sometimes 0%. Is there anything I can do to inverstigate/solve this, if it is a problem at all?

IAlibay:
> @hmacdope @IAlibay @tylerjereddy Thank you all for your constructive comments! I learned a lot from them, and I believe I addressed all the concerns and the PR is ready or almost ready to merge. I just have a couple of questions

Thanks, one of the mentors will re-review over the next few days.

> 1. According to [this discussion](https://github.com/MDAnalysis/mdanalysis/pull/3565#discussion_r831316421) with IAlibay, we should address the change to `universe.copy()` in a separate PR/Issue. Should I open a new issue for that?

Please do, refer the original issue please.

> 2. The codecov report varies from time to time (without me pushing any new commits). Sometimes it shows the diff to be about -4%, sometimes 0%. Is there anything I can do to inverstigate/solve this, if it is a problem at all?

This is likely due to us needing to approve the majority of the gh action runners when you do a new commit. It _shouldn't_ be an issue but we will look into it as part of the review process.

testsuite/MDAnalysisTests/analysis/test_contacts.py
```diff
@@ -171,16 +171,43 @@ class TestContacts(object):
...
+
         return contacts.Contacts(
             universe,
             select=(self.sel_acidic, self.sel_basic),
             refgroup=(acidic, basic),
             radius=6.0,
             **kwargs).run(start=start, stop=stop, step=step)
 
+    @pytest.mark.parametrize("seltxt", [sel_acidic, sel_basic])
+    def test_select_valid_types(self, universe, seltxt):
```
hmacdope, line 188: Could you add one more test, almost the same as this one but where you actually `.run()` the contacts with a `string` and `AtomGroup` input and compare the results?


You have tested the internal `staticmethod` `Contacts.get_atomgroup()` but not the actual flow of data through the whole function when a `string ` vs `AtomGroup` is used.

hmacdope (changes requested):
Almost there @shudipto-amin just one more test IMO then should be good to go. Thanks for your hard work.

hmacdope:
Don't worry about codecov, it is just the runners not auto-triggering for first time contributors.

package/MDAnalysis/analysis/contacts.py
```diff
@@ -387,14 +387,31 @@ class Contacts(AnalysisBase):
        :attr:`timeseries` results are now stored in a
        :class:`MDAnalysis.analysis.base.Results` instance.
     """
+    # Error message for wrong selection type:
+    _select_error_message = ("selection must be either string or a "
+                             "static AtomGroup. Updating AtomGroups "
+                             "are not supported.")
+
+    @staticmethod
+    def _get_atomgroup(u, sel):
```
IAlibay: If possible could this be moved after `__init__`?

Usually folks inspecting classes will want to see the initialisation as the first thing so that they can quickly get an idea of what the class expects.

package/MDAnalysis/analysis/contacts.py
```diff
@@ -387,14 +387,31 @@ class Contacts(AnalysisBase):
        :attr:`timeseries` results are now stored in a
        :class:`MDAnalysis.analysis.base.Results` instance.
     """
```
IAlibay, line 391: Please do add a `versionchanged` entry that mentions that AtomGroups are now supported.

package/MDAnalysis/analysis/contacts.py
```diff
@@ -387,14 +387,31 @@ class Contacts(AnalysisBase):
...
+    # Error message for wrong selection type:
+    _select_error_message = ("selection must be either string or a "
```
IAlibay: This is now only needed within `_get_atomgroup` so I would just define it there rather than making it a class attribute.

testsuite/MDAnalysisTests/analysis/test_contacts.py
```diff
@@ -171,16 +171,65 @@ class TestContacts(object):
...
         return mda.Universe(PSF, DCD)
 
-    def _run_Contacts(self, universe, start=None, stop=None, step=None, **kwargs):
+    def _run_Contacts(
+        self, universe,
+        start=None, stop=None, step=None, **kwargs
+    ):
         acidic = universe.select_atoms(self.sel_acidic)
         basic = universe.select_atoms(self.sel_basic)
+
```
IAlibay: Improving readability is appreciated but I also don't think it's worth affecting the git history for an empty line.
```suggestion
```

IAlibay (changes requested):
Couple more things

package/MDAnalysis/analysis/contacts.py
shudipto-amin: Done

testsuite/MDAnalysisTests/analysis/test_contacts.py
shudipto-amin: Done! See [line 199:](https://github.com/MDAnalysis/mdanalysis/pull/3565/files#diff-c79b8e25558e1ba2f042d25eaa0e355502afba57cf03bf8c9b22e0c6103063a4R199)
```python
def test_contacts_selections(self, universe):
```

package/MDAnalysis/analysis/contacts.py
shudipto-amin: I actually do use it in [this line](https://github.com/MDAnalysis/mdanalysis/pull/3565/files#diff-c79b8e25558e1ba2f042d25eaa0e355502afba57cf03bf8c9b22e0c6103063a4R228) of `test_contacts.py`:
```python
       with pytest.raises(
            TypeError, match=contacts.Contacts._select_error_message
        ) as te:
            contacts.Contacts._get_atomgroup(universe, ag)
```
I can't think of another way to access that error message that doesn't require first calling a class/method. Is there some way, or do I just keep it as in internal attribute?

shudipto-amin (commented):
@IAlibay I have addressed all but one comment below, regarding `_select_error_message` attribute, which I believe is still needed for testing functions.

@hmacdope I added the workflow test as you requested, [see line 199](https://github.com/MDAnalysis/mdanalysis/pull/3565/files#diff-c79b8e25558e1ba2f042d25eaa0e355502afba57cf03bf8c9b22e0c6103063a4R199):
```python
    def test_contacts_selections(self, universe):
```

package/MDAnalysis/analysis/contacts.py
IAlibay: Honestly I think it's better to:
a) not have this as a class attribute
b) make the error messages within `_get_atomgroup` a bit more specific if you feel like it
c) just write a subsection of the message as part of the match in `test_select_wrong_types`

Convenience variables are great, but having them stick around for the single test just isn't worth it.

testsuite/MDAnalysisTests/analysis/test_contacts.py
```diff
@@ -181,6 +184,51 @@ def _run_Contacts(self, universe, start=None, stop=None, step=None, **kwargs):
...
+        """Test if Contacts can take both string and AtomGroup as selections.
+        """
+        aga = universe.select_atoms(self.sel_acidic)
+        agb = universe.select_atoms(self.sel_basic)
+
+        cag = contacts.Contacts(
+            universe,
+            select=(aga, agb),
+            refgroup=(aga, agb),
+        ).run()
```
IAlibay: It's very pedantic but I'm going to object to these a little bit, this is not very readable.

If you're going to call run on it in one line then please re-arrange so that it's easier to read.

Otherwise close the brackets at 208 and call `cag.run()` separately.

Same thing below please.

IAlibay (changes requested):

testsuite/MDAnalysisTests/analysis/test_contacts.py
shudipto-amin: Opted for rearrangement. Is this better?

package/MDAnalysis/analysis/contacts.py
shudipto-amin: Okay, done!

testsuite/MDAnalysisTests/analysis/test_contacts.py
IAlibay: Sorry my readability issue wasn't with the way the arguments were arranged but instead with the `).run()` being alone on a single line. The compactness advantage of calling `run()` at the same time as class construction just disappears in this case.

shudipto-amin: I think I get you now. I put `.run()` on a new line altogether in [this latest commit](https://github.com/MDAnalysis/mdanalysis/pull/3565/commits/55636bf304a590b0c6e52dfee76a13e0d4861a48#diff-c79b8e25558e1ba2f042d25eaa0e355502afba57cf03bf8c9b22e0c6103063a4R215).

package/MDAnalysis/analysis/contacts.py
```diff
@@ -386,15 +386,18 @@ class Contacts(AnalysisBase):
     .. versionchanged:: 2.0.0
        :attr:`timeseries` results are now stored in a
        :class:`MDAnalysis.analysis.base.Results` instance.
+    .. versionchanged:: 2.2.0
+       :class: `Contacts` accepts both AtomGroup and string for `select`
```
IAlibay: Unfortunately read the docs hasn't triggered but there shouldn't be a blank space between :class: and the classname.
```suggestion
       :class:`Contacts` accepts both AtomGroup and string for `select`
```

package/CHANGELOG
```diff
@@ -13,14 +13,16 @@ The rules for this file:
...
 
  * 2.2.0
 
 Fixes
+  * Contacts can now accept static AtomGroups as well as selection texts.
+    (Issue #2666, PR #3565)
   * Fixed DumpReader triclinic box dimensions reading. (Issue #3386, PR #3403)
   * Updated the badge in README of MDAnalysisTest to point to Github CI Actions
-
+
```
IAlibay: Can you keep this empty line but remove the whitespace from it?

testsuite/MDAnalysisTests/analysis/test_contacts.py
```diff
@@ -181,6 +184,51 @@ def _run_Contacts(self, universe, start=None, stop=None, step=None, **kwargs):
...
+    def test_select_valid_types(self, universe, seltxt):
+        """Test if Contacts._get_atomgroup() can take both string and AtomGroup
+         as selections.
+        """
+        ag = universe.select_atoms(seltxt)
+
+        ag_from_string = contacts.Contacts._get_atomgroup(universe, seltxt)
+        ag_from_ag = contacts.Contacts._get_atomgroup(universe, ag)
+
+        assert_equal(ag_from_string, ag_from_ag)
```
IAlibay: atomgroup comparison already calls np.array_equal in the background, doing a standard assert is orders of magnitude faster here (seconds to microseconds).
```suggestion
        assert ag_from_string == ag_from_ag
```

IAlibay (changes requested):
Just a few tiny things and then I'm good with this.

@hmacdope do you want to re-review please?

package/CHANGELOG
shudipto-amin:
```suggestion
 
```

shudipto-amin (commented):
Minor edits via suggestions.

hmacdope (approved):
Looks good to me @shudipto-amin. Thanks for all your hard work, we appreciate your contribution.

shudipto-amin:
@IAlibay and @hmacdope, once again thanks for the feedback. Is there anything else, or is this ready to merge?

IAlibay:
I think it's waiting on a review from me, will do in a little bit

IAlibay:
Cycling PR to restart gh action runners that had halted last Friday (gh infrastructure issues it appears)

IAlibay (approved):
ltgm, thanks for the work on this @shudipto-amin.

Issue #2508, opened 2020-02-06
Have LinearDensity support updating AtomGroups, more intuitive names, save bins


lilyminium:
**Is your feature request related to a problem? Please describe.**
1. [This mailing list issue](https://groups.google.com/forum/#!searchin/mdnalysis-discussion/density%7Csort:date/mdnalysis-discussion/hTcPz8jaE_A/gUa5arH8AgAJ) is a good use case of Linear Density but is foiled by the class not supporting UpdatingAtomGroups.

2. It is not intuitive to me that `LinearDensity.results['x']['pos']` refers to the mass-weighted density in g/cm^3 of the selection. `pos` makes me think of coordinates. `char` makes me think of characters.

3. If the class saves the bins from `np.histogram`, it would be far easier to take advantage of existing histogram plotting tools. Also, using `np.histogramdd` seems easier.

4. The output is just an array of floats and the documentation doesn't say which units the density is in. These are not the normal units of MDAnalysis (amu for mass, Angstrom for length, e for charge) so they need special documentation. Also, the units written in the header of the file output are inaccurate for the charge. See #2507 for more

**Describe the solution you'd like**
1. Move the code below to `_single_frame()` so it updates. 
```python
        # Get masses and charges for the selection
        try:  # in case it's not an atom
            self.masses = np.array([elem.total_mass() for elem in group])
            self.charges = np.array([elem.total_charge() for elem in group])
        except AttributeError:  # much much faster for atoms
            self.masses = self._ags[0].masses
            self.charges = self._ags[0].charges

        self.totalmass = np.sum(self.masses)
```

2. Call them `.mass_density` and `.charge_density` in their own NumPy arrays. Have the results as Numpy arrays instead of dictionaries.

3. Save the bins

4. Not change the units.

**Describe alternatives you've considered**

**Additional context**
Happy to do this at some point, if people are ok with breaking scripts that have coded the original `.results['x']['pos']` indexing into them

orbeckst:
This all sounds extremely sensible to me.

For 1.0 we could break the API but if there's a way to keep the old data structures around (maybe hide them behind managed attributes that create them on the fly from the new real data structures) then that would help users who just want their old stuff to still work with the last py2 release but set us up for a clean removal for 2.0.

With properties for the old attributes we can also easily issue deprecation warnings.

hmacdope:
Hey all,
I would love to work on this if possible?

lilyminium:
[...]

Issue #2507, opened 2020-02-06
Different units in analysis.LinearDensity


lilyminium:
**Expected behavior**

MDAnalysis keeps consistent units of Angstrom for length, or documents when it doesn't.

**Actual behavior**

`analysis.LinearDensity` converts lengths from A^3 to cm^3

**Code to reproduce the behavior**

https://github.com/MDAnalysis/mdanalysis/blob/009ec225f0408f4e4f1366bec2c5576b61c88160/package/MDAnalysis/analysis/lineardensity.py#L177

[This isn't in the docs.](https://www.mdanalysis.org/mdanalysis/documentation_pages/analysis/lineardensity.html)

**Currently version of MDAnalysis**

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`)
- Which version of Python (`python -V`)?
- Which operating system?

lilyminium:
This also makes the header written in the file inaccurate. (charges are indicated to have e/A^3 density)

```python
        header = ("1 coord [Ang] 2-7 mass density (x,sx,y,sz,z,sz) [g/cm^3]"
                  "8-13 charge density (x,sx,y,sz,z,sz) [e/A^3]\n Average "
                  "density: {} g/cm3".format(density))
```

orbeckst:
@mattihappy can you clarify what was intended?

orbeckst:
I would be in favor of using standard units, Å, and e/Å<sup>3</sup>.

For density, g/cm<sup>3</sup> is more sensible than u/Å<sup>3</sup>.

Note that the save functions were removed #1745 (specifically PR #2487) so this boils down to 
1. documenting what is actually done
2. deciding to change the output data without deprecation — do we want to and if so, what should be changed?

PR #3572, opened 2022-03-24
LinearDensity now working with residues/segments

### LINKED ISSUE #3571: LinearDensity does not work on residues or segments
## Expected behavior ##

LinearDensity can calculate mass and charge densities for groups of atoms, residues, segments and fragments

## Actual behavior ##

Running LinearDensity with grouping='residues' or grouping='segments' causes AttributeError
AttributeError: Residue has no attribute centroid.
AttributeError: Segment has no attribute centroid.

``` python
import MDAnalysis as mda
from MDAnalysis.tests.datafiles import waterPSF, waterDCD
from MDAnalysis.analysis.lineardensity import LinearDensity

u = mda.Universe(waterPSF, waterDCD)
selection = u.select_atoms("all")

ld = LinearDensity(selection, grouping="residues").run()    # Fails
ld = LinearDensity(selection, grouping="segments").run()    # Fails

```

- Which version are you using?  2.2.0-dev0
- Which version of Python (`python -V`)? 3.9.10

BFedder:
Fixes #3571

LinearDensity did not work with grouping='residues' and grouping='segments'.
This is now fixed, and more tests have been added.

Changes made:
MDAnalysis/analysis/lineardensity.py:
  - in _prepare(self):
    - try/except statement replaced by if statements checking the grouping
      specified for LinearDensity.
    - masses and charges are obtained in a way appropriate for each case.

  - in _single_frame(self):
    - 'positions' variable assignment changed from elem.centroid() to
      elem.atoms.centroid() so it works for residues/segments

  - in _conclude(self):
    - Floating point arithmetic issues could lead to negative radicands. A check
      was implemented to set numbers which are close to zero to precisely zero
      before the square root (for standard deviation) is calculated.

MDAnalysisTests/analysis/test_lineardensity.py:
  - renamed previous test function and added expected masses and charges
  - added test functions for the other groupings with their respective expected
    masses and charges

Regarding my suggested change to _conclude(self): 
I stuck with binsize=5 for the new tests for the other groupings for internal 
consistency in the tests. With this bin size, the positions of the centres of 
geometry are invariant, yielding standard deviations equal to zero. Floating point
imprecision made it possible for negative values on the order of e-13 to e-14 to 
appear and cause a warning from np.sqrt() though, and that is why I added the 
extra step of calculating the radicand and editing it where appropriate beforehand. 
Other solutions are to just ignore this and accept nan values in the result arrays, 
or change the bin size for the tests. I was hesitant to do the later to stay consistent
with the established test, but I'm happy to change this of course if that's the consensus.

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


package/MDAnalysis/analysis/lineardensity.py
```diff
@@ -146,13 +150,19 @@ def _prepare(self):
         group = getattr(self._ags[0], self.grouping)
 
         # Get masses and charges for the selection
-        try:  # in case it's not an atom
-            self.masses = np.array([elem.total_mass() for elem in group])
-            self.charges = np.array([elem.total_charge() for elem in group])
-        except AttributeError:  # much much faster for atoms
+        if self.grouping == "atoms":
```
IAlibay, line 161: rather than using 3 `if` it might be worth making this `if/elif/else`? Possibly throwing an error if you have an unknown grouping?

package/MDAnalysis/analysis/lineardensity.py
```diff
@@ -146,13 +150,19 @@ def _prepare(self):
...
             self.charges = self._ags[0].charges
 
+        if self.grouping in ["residues", "segments"]:
+            self.masses = np.array([elem.atoms.total_mass() for elem in group])
+            self.charges = np.array(
+                [elem.atoms.total_charge() for elem in group])
+
+        if self.grouping == "fragments":
+            self.masses = np.array([elem.total_mass() for elem in group])
+            self.charges = np.array([elem.total_charge() for elem in group])
```
IAlibay: So someone like @richardjgowers is probably going to tell me off for this, but the cost of accessing `.atoms.total_mass()` on an atomgroup is pretty much within error as calling `.total_mass()`.

Would it not make sense here to just use the same logic for residues, segments and fragments?

package/MDAnalysis/analysis/lineardensity.py
```diff
@@ -183,6 +194,7 @@ def _single_frame(self):
             key = 'char'
             key_std = 'char_std'
             # histogram for positions weighted on charges
+
```
IAlibay: unnecessary blank line addition?

package/MDAnalysis/analysis/lineardensity.py
```diff
@@ -191,6 +203,7 @@ def _single_frame(self):
             self.results[dim][key] += hist
             self.results[dim][key_std] += np.square(hist)
 
+
```
IAlibay: pep8 - single blank line between class methods?

IAlibay (changes requested):
Couple of quick comments on the first half of the code (didn't get pass _conclude for now)

BFedder:
Hi, and thanks for the feedback!
First off, I'll be better about checking PEP8 compliance before opening a PR next time... Sorry! 

I could change the code you mention above to the following to have if/elif/else, raise an exception in the case of unknown groupings, and use the same logic for residues, segments and fragments?

``` python
# Get masses and charges for the selection
if self.grouping == "atoms":
    self.masses = self._ags[0].masses
    self.charges = self._ags[0].charges

elif self.grouping in ["residues", "segments", "fragments"]:
    self.masses = np.array([elem.atoms.total_mass() for elem in group])
    self.charges = np.array(
        [elem.atoms.total_charge() for elem in group])

else:
    raise RuntimeError(
        f"{self.grouping} is not a valid value for grouping.")

```

IAlibay:
> First off, I'll be better about checking PEP8 compliance before opening a PR next time... Sorry!

Flake8 can be really useful here, especially if combined with something like VS code. That being said, even of the coredevs are terrible at keeping up with pep8 🙄 

> I could change the code you mention above to the following to have if/elif/else, raise an exception in the case of unknown groupings, and use the same logic for residues, segments and fragments?

Sure, at a quick glance that seems reasonable, although you might want to use an AttributeError instead of a RuntimeError here.

IAlibay:
I'm going to cycle the PR (opening and closing it), I'm not sure why gh actions is being silly today...

IAlibay:
well at least the webhooks got triggered, that should be enough for now until actions comes back (another day another gh service issue... https://www.githubstatus.com/)

package/MDAnalysis/analysis/lineardensity.py
```diff
@@ -199,10 +210,24 @@ def _conclude(self):
...
+            # For certain tests in testsuite, floating point imprecision
+            # can lead to negative radicands of tiny magnitude (yielding nan),
+            # rather than actual value of zero. radicand_pos and radicand_char
+            # are therefore tested for closeness to 0 first and changed where
+            # appropriate before the square root is calculated.
+            radicand_pos = self.results[dim][
+                'pos_std'] - np.square(self.results[dim]['pos'])
+            for i in range(len(radicand_pos)):
+                if np.isclose(radicand_pos, np.zeros(self.nbins))[i]:
+                    radicand_pos[i] = 0
```
RMeli:
```suggestion
            radicand_pos = np.where(np.isclose(radicand_pos, 0), 0, radicand_pos)
```
or something on those lines.

package/MDAnalysis/analysis/lineardensity.py
```diff
@@ -199,10 +210,24 @@ def _conclude(self):
...
+            self.results[dim]['pos_std'] = np.sqrt(radicand_pos)
+
+            radicand_char = self.results[dim][
+                'char_std'] - np.square(self.results[dim]['char'])
+            for i in range(len(radicand_char)):
+                if np.isclose(radicand_char, np.zeros(self.nbins))[i]:
+                    radicand_char[i] = 0
```
RMeli:
```suggestion
            radicand_char = np.where(np.isclose(radicand_char, 0), 0, radicand_char)
```
or something on those lines.

package/MDAnalysis/analysis/lineardensity.py
RMeli: Or even
```suggestion
            radicand_pos[radicand_pos < 0] = 0
```

package/MDAnalysis/analysis/lineardensity.py
RMeli: Or even
```suggestion
            radicand_char[radicand_char < 0] = 0
```

package/MDAnalysis/analysis/lineardensity.py
BFedder: Thanks @RMeli! I was sure there must be a better solution for this and your suggestion looks much cleaner. 
I think using your first suggestion with `np.where` has two advantages: 

- If an actual negative number that's not close to zero shows up at this point, that's an error that should be flagged up/ throw a warning
- positive numbers on the order of e-14 are not meaningful results and should be 0 as well

But your second suggestion looks like it would give better performance. I'm not sure how much of a difference this would make, could people weigh in here? If the first suggestion is not much slower than the second, than that's the one I would go for.

RMeli: Yes, precisely. The first option is safer, the second one is faster. 
```
In [1]: import numpy as np

In [2]: x = np.array([1.0, 1e-10, -1e-10, 0.1])

In [3]: %timeit y = np.where(np.isclose(x, 0), 0, x)
40.4 µs ± 325 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [4]: z = x.copy()

In [5]: %timeit z[z < 0] = 0
2.1 µs ± 17.1 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

```
In [6]: x = np.random.normal(size=10000)

In [7]: %timeit y = np.where(np.isclose(x, 0), 0, x)
90.5 µs ± 1.23 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [8]: z = x.copy()

In [9]: %timeit z[z < 0] = 0
6.72 µs ± 65.2 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

In scientific software, IMHO, correctness is far more important than speed. However, since this is the `_conclude` function, if the code executed before performs enough checks it might be safe to use the second option here (i.e. the only negative values only come from numerical errors).

IAlibay: Because we only have to do this once, I might vote for the safer option here, unless we can guarantee that negative numbers are always floating point issues (I've not really delved into what the cause is here, maybe that is the case).

BFedder: Thanks for this clear breakdown, RMeli! 

IAlibay, I have looked into the code again and I can't see a way for actual negative numbers to appear here. I would suggest the following, then:

`radicand_pos[radicand_pos < 1e-9] = 0`

This is fast and also takes care of floating point errors that give a tiny positive number. 

`x = np.array([1.0, 1e-10, -1e-10, 0.1])`
`%timeit x[x < 1e-9] = 0`
`1.07 µs ± 8.22 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)`

IAlibay: Yeah sure I can live with that

BFedder: Changed now.

orbeckst: Sorry for the drive-by comment: If you're only concerned with negative numbers then I'd keep @RMeli 's suggestion with `x[x<0]` instead of introducing an arbitrary cutoff (and 1e-9 is not necessarily small). Numerical roundoff is normal and if anything I'd want to take care of it at the very end of a calculation and not infer in the middle if something ought to be zero or not. The objective of "no negative numbers" is clear. The objective of "should be zero" is not. Thus I only would implement the first one.

package/MDAnalysis/analysis/lineardensity.py
```diff
@@ -146,13 +150,19 @@ def _prepare(self):
...
-        try:  # in case it's not an atom
-            self.masses = np.array([elem.total_mass() for elem in group])
-            self.charges = np.array([elem.total_charge() for elem in group])
-        except AttributeError:  # much much faster for atoms
+        if self.grouping == "atoms":
             self.masses = self._ags[0].masses
             self.charges = self._ags[0].charges
 
+        elif self.grouping in ["residues", "segments", "fragments"]:
+            self.masses = np.array([elem.atoms.total_mass() for elem in group])
```
PicoCentauri: The method `group.total_mass` and `total_charge` have a parameter for `compound` to provide grouping which is faster than list compregensions. Is there a reason not to use it here?

package/MDAnalysis/analysis/lineardensity.py
```diff
@@ -163,8 +173,9 @@ def _single_frame(self):
         if self.grouping == 'atoms':
             positions = self._ags[0].positions  # faster for atoms
         else:
-            # COM for res/frag/etc
-            positions = np.array([elem.centroid() for elem in self.group])
+            # Centre of geometry for residues, segments, fragments
+            positions = np.array(
+                [elem.atoms.centroid() for elem in self.group])
```
PicoCentauri: The center of geometry has no physical meaning. I would still stick to the center of mass. You can get it with 
`positions  = group.center_of_mass(compound=self.grouping)`

testsuite/MDAnalysisTests/analysis/test_lineardensity.py
```diff
@@ -25,15 +25,70 @@
 
 from MDAnalysisTests.datafiles import waterPSF, waterDCD
 from MDAnalysis.analysis.lineardensity import LinearDensity
-from numpy.testing import assert_almost_equal
+from numpy.testing import assert_allclose
 
 
-def test_serial():
+def test_grouping_atoms():
```
PicoCentauri: You may simplify duplicated code with `pytest.mark.parametrize`. Look for examples in the testsuite, there are plenty of them. 

Additionally, there are no tests for the actual calculated density. I would at least add a check for the mean of `results.x.pos` and `results.x.char`

PicoCentauri (changes requested):
Good work to fix this!

I have some code simplifications and one physical question/addition. For me the center of geometry does not make sense to use as a reference point. It has no physical meaning.

package/MDAnalysis/analysis/lineardensity.py
PicoCentauri: I am confused by his operation. Do have an example where this happens? I never had an issue with `nan` in calculating density profiles.

Edit: Actually, I also had them. I [usually](https://gitlab.com/maicos-devel/maicos/-/blob/main/maicos/modules/density.py#L358-360) convert negative number with `np.nan_to_num` to 0 after calculating the standard deviation.

package/MDAnalysis/analysis/lineardensity.py
```diff
@@ -146,13 +150,19 @@ def _prepare(self):
...
             self.masses = self._ags[0].masses
             self.charges = self._ags[0].charges
 
+        elif self.grouping in ["residues", "segments", "fragments"]:
+            self.masses = np.array([elem.atoms.total_mass() for elem in group])
+            self.charges = np.array(
+                [elem.atoms.total_charge() for elem in group])
+
+        else:
+            raise AttributeError(
```
IAlibay, line 170: Make sure this is covered by tests please.

package/MDAnalysis/analysis/lineardensity.py
IAlibay: Seems like one of those where the original code thought COM but actually attempted (and failed) to do COG.

I agree with COM here too.

This does need to be clearly signposted in the documentation for this class. Can you see if you could add this somewhere in the docstring, and also add a relevant example of using this for there various groupings too please?

IAlibay (changes requested):

package/MDAnalysis/analysis/lineardensity.py
BFedder: That's a good point, @orbeckst, thanks! I have changed this to `x[x<0]` now. 

@PicoCentauri, Ah, I did not know about `np.nan_to_num`, thank you! For now, I stuck with the other suggestion because it is slightly faster, but is there another advantage to nan_to_num?

package/MDAnalysis/analysis/lineardensity.py
BFedder: No reason that I can see, thanks for this suggestion!

package/MDAnalysis/analysis/lineardensity.py
BFedder: Yes, I went with COG because that's what was used previously. I have now changed it to COM and made use of `compound`, thanks. 

@IAlibay, I have changed the class docstring in the Parameters section and in the version history, but I am not sure what a useful example would be, because the user input has not changed and the output for the relevant groupings is only now available, so it should not break any pre-existing scripts. Could you please clarify what examples could help here?

PicoCentauri (approved):
Very well. LGTM

package/MDAnalysis/analysis/lineardensity.py
PicoCentauri: If it is faster it is always better. I am not aware of any other advantages compared to your solution.

package/MDAnalysis/analysis/lineardensity.py
IAlibay: Just the same as what is already under example but saying something along the lines of "if you wanted to do this per residue you could do it this way <code block>". It's a tiny addition but it can be quite a nice thing to read if you're not super sure how the grouping argument works.

testsuite/MDAnalysisTests/analysis/test_lineardensity.py
BFedder: I have now refactored the test suite to limit duplicate code, thanks for the suggestion. 

The reason I had not previously added tests for the calculated densities is that I am not sure how I would calculate these on my own. For now, I have taken the output of lineardensity.py and used that as placeholder tests, but that obviously does not do much in the way of actual testing. Do you have some advice on how I could get the "ground truth" data for the densities?

PicoCentauri: You are right. The current values are just magic numbers. These will probably fail when we do something with the code. Depending on how much effort you want to invest I propose the following:

Generate a trajectory based on random positions of particles in a cubic box. Draw the positions from a uniform distribution. Run the analysis and compare the densities. You could do three kinds of system: uncharged particles, charged particles and maybe charged dimers. This is probably beyond the scope of this PR but would be a valuable addition!

package/MDAnalysis/analysis/lineardensity.py
BFedder: I have added an example of this now

testsuite/MDAnalysisTests/analysis/test_lineardensity.py
BFedder: One reason why I stuck with binsize=5 with my new tests is that the original test, which also used this value, had expected values for `results.x.pos`. I thought that this would at least be partial validation for the restored functionality. But independently determined expected values are definitely still needed here. 

Thanks for the suggestion! For now, I need to focus on my GSoC application/proposal, but I would be interested in working on LinearDensity further. I originally started looking into this module because it looked like a good candidate for more tests on codecov, and that's when I encountered the bug fixed in this PR. It looks like between adding tests for the parallel analysis code, getting expected values for the tests, and issues #2507 and #2508, there is still quite a bit to be done here, and I'd be happy to look into it more after more work on my application. For now, I can open an issue describing the need for more test data, and this PR could get merged?

PicoCentauri: Yes take your time but opening an issue is a good idea!

BFedder: I have now opened issue #3603 for improved tests.

orbeckst:
@BFedder can you please resolve the conflicts?

orbeckst:
@IAlibay can you please check if your comments were addressed?

testsuite/MDAnalysisTests/analysis/test_lineardensity.py
orbeckst: Just a quick comment: "ground truth" tests are excellent when you can get them. However, "regression tests" are the next best thing: they will check that future changes do not change what was once considered working. For this PR I would be totally happy with regression tests, especially as the original code also used them.

Thanks for raising the issue, that's a good approach to doing incremental software development: do small working units and keep track of what needs to be done through issues.

package/MDAnalysis/analysis/lineardensity.py
```diff
@@ -79,6 +80,8 @@ class LinearDensity(AnalysisBase):
        ldens.run()
        print(ldens.results.x.pos)
 
+       # alternatively, specify grouping/binsize like this
+       ldens = LinearDensity(selection, grouping='residues', binsize=1.0)
```
IAlibay:
```suggestion

    Alternatively, the other types of groupings can be selected using the ``grouping``
    keyword. For example to calculated the density based on the :class:`ResidueGroup`s
    of the system:
    
    .. code-block:: python
       ldens = LinearDensity(selection, grouping='residues', binsize=1.0)
       ldens.run()

```

testsuite/MDAnalysisTests/analysis/test_lineardensity.py
```diff
@@ -22,18 +22,91 @@
...
+                          expected_masses_segments,
+                          expected_charges_segments,
+                          expected_xpos_segments,
+                          expected_xchar_segments),
+                         ("fragments",
+                          expected_masses_fragments,
+                          expected_charges_fragments,
+                          expected_xpos_fragments,
+                          expected_xchar_fragments)
+                         ])
```
IAlibay: PEP8 is a suggestion not an obligation - don't use this against me

```suggestion
@pytest.mark.parametrize("grouping, expected_masses, expected_charges, expected_xpos, expected_xchar", [
    ("atoms", expected_masses_atoms, expected_charges_atoms,
     expected_xpos_atoms, expected_xchar_atoms),
    ("residues", expected_masses_residues, expected_charges_residues,
     expected_xpos_residues, expected_xchar_residues),
    ("segments", expected_masses_segments, expected_charges_segments,
     expected_xpos_segments, expected_xchar_segments),
    ("fragments", expected_masses_fragments, expected_charges_fragments,
     expected_xpos_fragments, expected_xchar_fragments)
])
```

IAlibay (changes requested):
Just two small suggested changes - I'm directly suggesting them here since they are more "my style" than anything. Feel free to say no, I'll approve after.

BFedder:
Thanks, I have taken the suggestions on board

IAlibay (approved):
Thanks @BFedder 

@MDAnalysis/gsoc-mentors - I can't merge this as a GSoC PR due to conflicts of interest. Can someone please take over the responsibility of merging this?

orbeckst:
@PicoCentauri can you please have a final look and then squash-merge? Thanks.

PicoCentauri (approved):

PicoCentauri:
Thanks @BFedder for your fix!

PR #3573, opened 2022-03-24
Fast cythonized inverse array of unsorted, unique indice

### LINKED ISSUE #3387: Easy inverse index array

ghost:
Fixes #3387

 - Added cython implementation of inverse of unsorted, unqiue array
 - Replaced python code in groups.py with new cython function
 - Added tests confirming inverse correctly maps unique array onto original

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


ghost:
Artifical benchmarking shows speed up.
Randomly generated arrays of integers of varying sizes and varying number of unique values. ie:
`np.random.randint(1, n_unique, n_indices, dtype=np.intp)`
And comparing the new function with the old version pulled from groups.py
```
def current(index_array, indices):
    mask = np.zeros_like(index_array)  
    for i, x in enumerate(indices):  
        values = np.where(index_array == x)[0]  
    mask[values] = i 
    return mask
```  

I can add these result in the comments if they are sufficient.
Or is there a more realistic test that should be run?

testsuite/MDAnalysisTests/lib/test_cutil.py
```diff
@@ -45,6 +45,47 @@ def test_unique_int_1d(values):
...
+    
+    inverse = inverse_unique_contiguous_1d_array(array, unique_array)
+    ref = array
+    res = unique_array[inverse]
+    assert_equal(res, ref)
+    assert type(res) == type(ref)
+    assert res.dtype == ref.dtype
+    
+    # raise ValueError on none 1d inputs
+    with pytest.raises(ValueError) as e_info:
```
tylerjereddy: here and below `e_info` is not used? you could use the `match` keyword if you care about checking the error messages maybe

testsuite/MDAnalysisTests/lib/test_cutil.py
```diff
@@ -45,6 +45,47 @@ def test_unique_int_1d(values):
...
+    unique_array = array[np.sort(indices)]
+    
+    inverse = inverse_unique_contiguous_1d_array(array, unique_array)
+    ref = array
+    res = unique_array[inverse]
+    assert_equal(res, ref)
+    assert type(res) == type(ref)
+    assert res.dtype == ref.dtype
+    
+    # raise ValueError on none 1d inputs
```
tylerjereddy: `none` -> `non`?

testsuite/MDAnalysisTests/lib/test_cutil.py
```diff
@@ -45,6 +45,47 @@ def test_unique_int_1d(values):
...
+    [2, 3, 5, 7, ],  # unique, sorted
+    [5, 2, 7, 3, ],  # unique, unsorted
+    [1, 2, 2, 4, 4, 6, ],  # duplicates, sorted
+    [1, 2, 2, 6, 4, 4, ],  # duplicates, unsorted
+))
+def test_inverse_unique_contiguous_1d_array(arrays):
+    # Generate unique, unsorted arrays.
+    array = np.array(arrays, dtype=np.intp)
+    values, indices = np.unique(array, return_index=True)
+    unique_array = array[np.sort(indices)]
```
tylerjereddy: I'm having a hard time understanding why you are doing a sort operation here. `unique_array`, as defined for the Cython version of `inverse_unique_contiguous_1d_array` is (emphasis mine):

> the unique, **unsorted** values of full_arr.

So, shouldn't it work whether the values are sorted or not?

For example, this diff still passes the tests and is a bit easier for me to parse:

```diff
--- a/testsuite/MDAnalysisTests/lib/test_cutil.py
+++ b/testsuite/MDAnalysisTests/lib/test_cutil.py
@@ -57,9 +57,11 @@ def test_inverse_unique_contiguous_1d_array(arrays):
     # Generate unique, unsorted arrays.
     array = np.array(arrays, dtype=np.intp)
     values, indices = np.unique(array, return_index=True)
-    unique_array = array[np.sort(indices)]
+    unique_array = array[indices]
     
     inverse = inverse_unique_contiguous_1d_array(array, unique_array)
+    expected_inverse = np.unique(array, return_inverse=True)[1]
+    assert_equal(inverse, expected_inverse)
     ref = array
     res = unique_array[inverse]
     assert_equal(res, ref)
@@ -71,7 +73,7 @@ def test_inverse_unique_contiguous_1d_array(arrays):
         # change array shape
         array_2d = array.reshape((2,-1))
         values, indices = np.unique(array_2d, return_index=True)
-        unique_array = array[np.sort(indices)]
+        unique_array = array[indices]
         inverse_unique_contiguous_1d_array(array_2d, unique_array)
     
     # raise ValueError on non-contiguous data
@@ -79,7 +81,7 @@ def test_inverse_unique_contiguous_1d_array(arrays):
     if not array.data.contiguous: # empty array ([]) is still contiguous
       with pytest.raises(ValueError) as e_info:
           values, indices = np.unique(array, return_index=True)
-          unique_array = array[np.sort(indices)]
+          unique_array = array[indices]
           inverse = inverse_unique_contiguous_1d_array(array, unique_array)
```

I'm also not really sure if we need the block with:

```python
     ref = array
     res = unique_array[inverse]
     assert_equal(res, ref)
```

It might be sufficient to confirm that the MDAnalysis version of `inverse` matches the same functionality in NumPy (though this is more minor point I suppose).

tylerjereddy (commented):
For benchmarks we try to use airspeed velocity: https://asv.readthedocs.io/en/stable/

You can find some examples in our `benchmarks/` folder. It may also be useful to informally show that this new Cython stuff is faster than the NumPy version. I suppose it should be since it likely has less error handling/arguments/conditions to worry about on top of not having a sorting burden?

testsuite/MDAnalysisTests/lib/test_cutil.py
ghost: This hasn't been changed by me and was taken directly from the function which generates the unique indicies just prior to the change:
https://github.com/MDAnalysis/mdanalysis/blob/800298ad271a00762ba3f51629c1ce41b7bf4172/package/MDAnalysis/lib/util.py#L233-L235

As far as I understand it, np.unique returns the array **sorted** and the return_index returns the sorted array inverse, by sorting the sorted unique array by the inverse you then return the **unsorted**, unique array.
Now that I type that out I suppose the proper thing to do is to import that function to use, then any changes to it will also remain in the tests.

This new cython code then generates the inverse of the new unsorted array.

Now that I type that out I suppose the proper thing to do is to import that function to use, then any changes to that function will also be expressed in the tests.

Perhaps this might also be a target for future improvement to remove the redundent sorts (and this code) by implimenting a unique function that returns the unique array and its inverse without the two additional sorts (sorting in np.unique, sorting to unsorted the results).

There already exists a cython 1d unique function, but that relies on sorting first too:
https://github.com/MDAnalysis/mdanalysis/blob/800298ad271a00762ba3f51629c1ce41b7bf4172/package/MDAnalysis/lib/_cutil.pyx#L54-L57

Just to clarify the comment about the test you think just confirming that the original code block returns the same inverse array as the new one is sufficient, rather than testing that the inverse returns the original array?

testsuite/MDAnalysisTests/lib/test_cutil.py
ghost: I have just remove this.

testsuite/MDAnalysisTests/lib/test_cutil.py
ghost: Typos! Bane of my life.

testsuite/MDAnalysisTests/lib/test_cutil.py
```diff
@@ -46,6 +46,43 @@ def test_unique_int_1d(values):
...
 
+@pytest.mark.parametrize('arrays', (
+    [],  # empty array
+    [1, 1, 1, 1, ],  # all identical
+    [2, 3, 5, 7, ],  # unique, sorted
+    [5, 2, 7, 3, ],  # unique, unsorted
+    [1, 2, 2, 4, 4, 6, ],  # duplicates, sorted
+    [1, 2, 2, 6, 4, 4, ],  # duplicates, unsorted
+))
+def test_inverse_unique_contiguous_1d_array(arrays):
```
lilyminium: `arrays` implies multiple array, but as far as I can tell this is only the one array -- could you please rename this variable for readability?

testsuite/MDAnalysisTests/lib/test_cutil.py
```diff
@@ -46,6 +46,43 @@ def test_unique_int_1d(values):
...
+    # Generate unique, unsorted arrays.
+    array = np.array(arrays, dtype=np.intp)
+    unique_array = unique_int_1d_unsorted(array)
+
+    inverse = inverse_unique_contiguous_1d_array(array, unique_array)
+    ref = array
+    res = unique_array[inverse]
+    assert_equal(res, ref)
+    assert type(res) == type(ref)
+    assert res.dtype == ref.dtype
```
lilyminium: Instead of relying on the logic of `unique_int_1d_unsorted`, could you please just test against an expected value? You can pass that in as another `parameterize` argument. That makes it far more of a [unit test](https://smartbear.com/learn/automated-testing/what-is-unit-testing/) and therefore more robust (it doesn't rely on `unique_int_1d_unsorted` working properly) and also more readable. It should be fairly doable for these short arrays.

testsuite/MDAnalysisTests/lib/test_cutil.py
```diff
@@ -46,6 +46,43 @@ def test_unique_int_1d(values):
...
+
+    # raise ValueError on non 1d inputs
+    with pytest.raises(ValueError):
+        # change array shape
+        array_2d = array.reshape((2, -1))
+        values, indices = np.unique(array_2d, return_index=True)
+        unique_array = array[np.sort(indices)]
+        inverse_unique_contiguous_1d_array(array_2d, unique_array)
```
lilyminium: Could you please make this a separate test -- it doesn't need to run for each `parametrize` value.

testsuite/MDAnalysisTests/lib/test_cutil.py
```diff
@@ -46,6 +46,43 @@ def test_unique_int_1d(values):
...
+
+    # raise ValueError on non-contiguous data
+    array = np.array(arrays, dtype=np.intp)[::2]
+    if not array.data.contiguous:  # empty array ([]) is still contiguous
+        with pytest.raises(ValueError):
+            values, indices = np.unique(array, return_index=True)
+            unique_array = array[np.sort(indices)]
+            inverse = inverse_unique_contiguous_1d_array(array, unique_array)
```
lilyminium: See above comment about separate tests

lilyminium (changes requested):
Thank you for opening this pull request! I just had a couple comments about the tests :)

package/MDAnalysis/lib/_cutil.pyx
```diff
@@ -49,6 +49,70 @@ ctypedef cset[int] intset
 ctypedef cmap[int, intset] intmap
 
 
+@cython.boundscheck(False) # turn off bounds-checking for entire function 
+@cython.wraparound(False)  # turn off negative index wrapping for entire function 
+cdef np.intp_t[::1] _inverse_unique_unsorted_array(np.intp_t[::1] full_arr,
```
richardjgowers: Since we're modifying mask in place, I think it's clearer to have this with a void return type.

package/MDAnalysis/lib/_cutil.pyx
```diff
@@ -49,6 +49,70 @@ ctypedef cset[int] intset
...
+    full_arr: numpy.ndarray
+        1D array of dtype ``numpy.int64`` in which to find the non-unique values.
+    unique_arr: numpy.ndarray
+        1D array of dtype ``numpy.int64``, the unique, unsorted values of full_arr.
+
+    Returns
+    -------
+    numpy.ndarray
+        1D array of dtype ``numpy.int64``, the inverse that maps unique_arr onto full_arr.
+
```
richardjgowers, line 159: Could we include a trivial example here? It's not obvious from the name what this does

package/MDAnalysis/core/groups.py
```diff
@@ -822,10 +822,10 @@ def _asunique(self, group, sorted=False, set_mask=False):
 
         indices = unique_int_1d_unsorted(self.ix)
         if set_mask:
-            mask = np.zeros_like(self.ix)
-            for i, x in enumerate(indices):
-                values = np.where(self.ix == x)[0]
-                mask[values] = i
+            # self.ix must be contiguous for cython implementation.
```
richardjgowers, line 825: this could be relaxed if we removed the `[::1]` on the type annotations.  It might be overall quicker to not have to first copy the data...   More generally in another issue it might then be worth looking at enforcing `self.ix` being contiguous (rather than repeatedly converting) since it's used in this context so often.

package/MDAnalysis/lib/_cutil.pyx
```diff
@@ -49,6 +49,70 @@ ctypedef cset[int] intset
...
+    
+    cdef Py_ssize_t i = 0 
+    cdef Py_ssize_t j = 0
+    cdef int n_full = full_arr.shape[0]
+    cdef int n_unique = unique_arr.shape[0]
+
+    for i in range(n_unique):
+        for j in range(n_full):
+            if unique_arr[i] == full_arr[j]:
+                mask[j] = i
```
richardjgowers: if we rearrange these loops we can put a break statement inside and reduce the number of iterations right?  (i.e. reduce the number of comparisons from n x m)

richardjgowers (commented):
Really cool first contribution!

package/MDAnalysis/core/groups.py
ghost: From what I can tell np.ascontiguousarray only does a copy on the array if it isn't contiguous.
```
import numpy as np
print(np.__version__)   # 1.22.3
array = np.array([0,0,0,0], dtype=np.intp)    
noncontiguous_arr = array[::2]
contiguous_arr = np.ascontiguousarray(array[::2])
print(np.ascontiguousarray(contiguous_arr) is contiguous_arr)        # True
print(np.ascontiguousarray(noncontiguous_arr) is noncontiguous_arr)  # False
```

I have tried timing two different version, 

1. with the contiguous type removed(`[::1]` to `[:]`)  as suggested 
2. applying np.ascontiguousarray() onto the two function parameters.

In all cases tested option 2 was faster than 1 for both contiguous and noncontiguous input data.
image

package/MDAnalysis/lib/_cutil.pyx
ghost: Ah yes, missed that.
I was originally generating mask in that function but generating it outside and editing it inplace was quicker for some reason.

package/MDAnalysis/lib/_cutil.pyx
ghost: That is true, based based on the ordering of the unique values, you are guarenteed to shrink the range of the inner loop after each sweep over the values by just tracking the position of the first value in the array not yet inversed.
The same could be done at the end too although that might not pay off as the last value might be in the last position (ie `full_arr=[1, 1, 2, 2, 3, 3, 4, 4, 5, 5]`), it could subtantially help in the case of something more like `full_arr=[1, 2, 3, 4, 5, 1, 1, 1, 1, 1]` which would have the same `unique_arr`. 

Ill have a go at impletmenting something like this and timing it.

As it currently is used in group.py, I believe the inner loop can be change to `for j in range(i, n_full)` but this would completely remove the option of using this function on a shuffed/sorted `unique_arr`. Not sure why that might be done but it could come up in the future.

package/MDAnalysis/lib/_cutil.pyx
```diff
@@ -49,6 +49,70 @@ ctypedef cset[int] intset
...
+	1D array of dtype ``numpy.int64``, preallocated inverse array to be filled.
+
+    Returns
+    -------
+    numpy.ndarray
+        1D array of dtype ``numpy.int64``, the inverse that maps unique_arr onto full_arr.
+
+    """
+    
+    cdef Py_ssize_t i = 0
```
hmacdope: The use of `Py_ssize_t` here is correct and should really be done more in our cython layers. Just an observation.

Thanks for your contribution!

package/MDAnalysis/core/groups.py
richardjgowers: Cool thanks for benchmarking it

package/MDAnalysis/lib/_cutil.pyx
ghost: I didn't add any break statements but implemented a slightly more sophisticated inner loop range.
It just progressively attempts to shrink the inner loop range by ignoring values already processed.
It does have a flaw that it won't bound an array like `[1,2,1,...,1,2,1]` effectively as the inner 1 will 'block' the bounding.
The much more complicated attempt to resolve this, relying on the assumption that the `unique_arr` is sorted according to the order of occurance in the `full_arr` but it was actually slower. It effectively reduced the total number comparisons within the array but I guess due to the additional overhead, this wasn't worth it.

I forgot to set the facecolor when saving the plots. x-axis=full_arr size, y-axis=runtime accoring to timeit.timeit().
It is slower for non-interleaved `full_arr` values (ie `[1,1,2,3,4,5,6,6,7,...]`):
image

It is slower than just brute force for random `full_arr` values:
image

testsuite/MDAnalysisTests/lib/test_cutil.py
ghost: Just occured to me but what is the best practice for tests with a single set of parameters?
I used a `parameterize` here, but `test_ind2d` a little further down just hardcodes the values into the test.

lilyminium: Yeah just do that if you only want one test :)

package/MDAnalysis/lib/_cutil.pyx
```diff
@@ -37,8 +37,8 @@ from cython.operator cimport dereference as deref
 
 np.import_array()
 
-__all__ = ['unique_int_1d', 'make_whole', 'find_fragments',
-           '_sarrus_det_single', '_sarrus_det_multiple']
+__all__ = ['inverse_unique_contiguous_1d_array', 'unique_int_1d', 'make_whole',
+	   'find_fragments', '_sarrus_det_single', '_sarrus_det_multiple']
```
lilyminium: Could you please replace the tabs with spaces?

package/MDAnalysis/lib/_cutil.pyx
```diff
@@ -49,6 +49,119 @@ ctypedef cset[int] intset
 ctypedef cmap[int, intset] intmap
 
 
+@cython.boundscheck(False) # turn off bounds-checking for entire function 
+@cython.wraparound(False)  # turn off negative index wrapping for entire function 
+cdef void _inverse_unique_unsorted_contiguous_array_inplace(
+						np.intp_t[::1] full_arr,
+                				np.intp_t[::1] unique_arr,
+                                		np.intp_t[::1] mask
+						):
```
lilyminium: There are tabs here too

package/MDAnalysis/lib/_cutil.pyx
```diff
@@ -49,6 +49,119 @@ ctypedef cset[int] intset
...
+    """Find the inverse of a unique 1D array to map onto a non-unique 1D array.
+
+    Cython driver function. Loops though the unique array and finds the indices 
+    in the full array that correspond to the values in the unique array.
+
+    ie: np.array([1,2,3])[np.array(0,0,1,2,2)] == np.array([1,1,2,3,3])
+	       unique_arr[mask]                ==       full_arr
```
lilyminium: tabs

package/MDAnalysis/lib/_cutil.pyx
```diff
@@ -49,6 +49,119 @@ ctypedef cset[int] intset
...
+						np.intp_t[::1] full_arr,
+                				np.intp_t[::1] unique_arr,
+                                		np.intp_t[::1] mask
+						):
+    """Find the inverse of a unique 1D array to map onto a non-unique 1D array.
+
+    Cython driver function. Loops though the unique array and finds the indices 
+    in the full array that correspond to the values in the unique array.
+
+    ie: np.array([1,2,3])[np.array(0,0,1,2,2)] == np.array([1,1,2,3,3])
```
lilyminium, line 64: Could you reformat this as an example, i.e. using the Example section? This currently doesn't really tell me anything, as I don't know what `full_arr` and `unique_arr[mask]` are meant to correspond to (on the line below).

```
Example
--------
::
    full_arr = np.array([1, 1, 2, 3, 3])
    ...
```


Typically `Examples` will be under the `Parameters` and `Returns` sections of a docstring

package/MDAnalysis/lib/_cutil.pyx
```diff
@@ -49,6 +49,119 @@ ctypedef cset[int] intset
...
+	       unique_arr[mask]                ==       full_arr
+
+
+    Calculates the bounding region of unsorted values to reduce number of iterations,
+    more efficient on uninterleaved full_arr, where unique_arr is sorted according to
+    first occurance of the value in full_arr.
```
lilyminium: I think this means `unique_arr` is *unsorted* and *ordered* by first appearance. The bounding region and efficiency note also is a bit of an implementation detail and should appear last, after the requirement of contiguous arrays, which is much more important.

Also, occurance -> occurrence

package/MDAnalysis/lib/_cutil.pyx
lilyminium: Agreed -- using the `Example` section in the docstring I suggested above would be nice :)

package/MDAnalysis/lib/_cutil.pyx
```diff
@@ -49,6 +49,119 @@ ctypedef cset[int] intset
...
+                    lower_bound += 1  # increment until first non-matching value is found
+                if not bound_end:
+                    # set to position of first N matching values.
+                    new_upper_bound = j  
+                    bound_end = True  # stop incrementing after first found value.
+            else:
+                # stop start bounding once first non-matching value found.
+                init_val = False 
+                # restart end bounding if non-matching value found.
+                bound_end = False
```
lilyminium: How does this new bounds thing compare in speed to the following rearrangement of your initial loop?

```
for j in range(n_full):
    for i in range(n_unique):
        if unique_arr[i] == full_arr[j]:
            mask[j] = i
            break
```

testsuite/MDAnalysisTests/lib/test_cutil.py
```diff
@@ -46,6 +45,43 @@ def test_unique_int_1d(values):
     assert res.dtype == ref.dtype
 
 
+@pytest.mark.parametrize(['array', 'unique_array', 'inverse'],  (
+    ([], [], []),  # empty array
+    ([1, 1, 1, 1, ], [1,], [0, 0, 0, 0, ]),  # all identical
+    ([2, 3, 5, 7, ], [2, 3, 5, 7, ], [0, 1, 2, 3, ]),  # unique, sorted
+    ([5, 2, 7, 3, ], [5, 2, 7, 3, ], [0, 1, 2, 3, ],),  # unique, unsorted
+    ([1, 2, 2, 4, 4, 6, ], [1, 2, 4, 6, ], [0, 1, 1, 2, 2, 3, ]),  # duplicates, sorted
```
lilyminium:
```suggestion
    # duplicates, sorted
    ([1, 2, 2, 4, 4, 6], [1, 2, 4, 6], [0, 1, 1, 2, 2, 3]),
```

The PEP8 bot is complaining about some line lengths here -- one option could be to move the comments (like the `# duplicates, sorted`) ones to the lines above

lilyminium (changes requested):
Thank you for the changes you made to the tests -- they're so much easier to read and understand! This is a great contribution to the library, especially with the benchmarks you did showing the faster speeds. I just had some comments about the rewriting of the loop and docs.

package/MDAnalysis/lib/_cutil.pyx
ghost: I went down a rabbit hole massively overcomplicating things it seems.

hmacdope:
@mdaviesNCL are you thinking of applying for GSOC or Outreachy with MDAnalysis this year? I ask because if you are thinking of applying, you will need to introduce your on the developers mailing [list](https://groups.google.com/g/mdnalysis-devel). 

If not all good, its a good idea to introduce yourself on the mailing list regardless. :)

package/MDAnalysis/lib/_cutil.pyx
```diff
@@ -49,6 +49,106 @@ ctypedef cset[int] intset
...
+
+
+    Parameters
+    ----------
+    full_arr: numpy.ndarray
+        1D array of dtype ``numpy.int64``, the non-unique values.
+    unique_arr: numpy.ndarray
+        1D array of dtype ``numpy.int64``, the unique values of full_arr.
+    mask: numpy.ndarray
+        1D array of dtype ``numpy.int64``, inverse array to be filled.
```
lilyminium: Could you please move the notes in lines 67-69 into the `Parameters` description, that just keeps them more ordered and tells you everything you need to know about the necessary argument in one place.

package/MDAnalysis/lib/_cutil.pyx
```diff
@@ -49,6 +49,106 @@ ctypedef cset[int] intset
...
+
+        full_arr = np.array([1,2,3,4,1,2,3,4], dtype=np.intp)
+        unique_arr = unique_int_1d_unsorted(full_arr)
+        # unique_arr == np.array([1,2,3,4], dtype=np.intp)
+
+        inverse = inverse_unique_contiguous_1d_array(full_arr, unique_arr)
+        # inverse == np.array([0,1,2,3,0,1,2,3], dtype=np.intp)
+
+        rebuilt_arr = unique_arr[inverse]
+        # rebuilt_arr has the same elements as the original full_arr
```
lilyminium: Would this be clearer?

```
assert full_arr == unique_arr[inverse]
```

(although in actual execution you might just get numpy complaining about array equality -- you could import assert_equal if you want to be more technically correct)

lilyminium (changes requested):
Thanks for the update, @mdaviesNCL -- it's looking great. Just some minor doc comments.

Is this also your first contribution to MDAnalysis? Could you please add your name to the AUTHORS list here: https://github.com/MDAnalysis/mdanalysis/blob/develop/package/AUTHORS

And please update CHANGELOG with a description of the PR! It probably belongs in `Enhancements`: https://github.com/MDAnalysis/mdanalysis/blob/develop/package/CHANGELOG

package/MDAnalysis/lib/_cutil.pyx
lilyminium: This would still be nice.

package/MDAnalysis/lib/_cutil.pyx
```diff
@@ -49,6 +49,122 @@ ctypedef cset[int] intset
...
+    -------
+
+    Example
+    -------
+    Find the inverse of an array::
+
+        cdef full_arr = np.array([1,2,3,4,1,2,3,4], dtype=np.intp)
+        cdef unique_arr = np.array([1,2,3,4], dtype=np.intp)
+        cdef mask = np.empty(full_arr.shape[0], dtype=np.intp)
+    	_inverse_unique_unsorted_contiguous_array_inplace(full_arr,
```
lilyminium, line 91: Lonely tab here.

package/MDAnalysis/lib/_cutil.pyx
```diff
@@ -49,6 +49,122 @@ ctypedef cset[int] intset
...
+                                                          unique_arr,
+                                                          mask)
+       assert full_arr == unique_arr[mask]
+
+    """
+
+
+    """
```
lilyminium, line 99:
```suggestion
```

I think these extra quotes are breaking the compilation and therefore the test suite :)

package/MDAnalysis/lib/_cutil.pyx
```diff
@@ -49,6 +49,122 @@ ctypedef cset[int] intset
...
+                                                np.intp_t[::1] mask
+                                                ):
+    """Find the inverse of a unique 1D array to map onto a non-unique 1D array.
+
+    Cython driver function. Loops though the unique array and finds the indices
+    in the full array that correspond to the values in the unique array.
+
+    ie: np.array([1,2,3])[np.array(0,0,1,2,2)] == np.array([1,1,2,3,3])
+               unique_arr[mask]                ==       full_arr
+
```
lilyminium, line 66: I think this is now extraneous given the example below.

lilyminium (changes requested):
Aside from breaking CI, this looks pretty good to me now :) Could you please add your username to the CHANGELOG as well, in the username section before you described your fix?

@hmacdope @richardjgowers do you also want to re-review?

1krishnasharma:
Hi, I am new to open source. so, please tell me how can I start

package/MDAnalysis/lib/_cutil.pyx
```diff
@@ -49,6 +49,122 @@ ctypedef cset[int] intset
...
+    unique_arr: numpy.ndarray
+        1D contiguous array of dtype ``numpy.int64``,
+        the unique values of full_arr, can be unsorted.
+    mask: numpy.ndarray
+        1D contiguous array of dtype ``numpy.int64``,
+        inverse array to be changed inplace, the same length as full_arr.
+
+    Returns
+    -------
+
```
richardjgowers, line 83: Returns what? :)

package/MDAnalysis/core/groups.py
```diff
@@ -822,10 +822,10 @@ def _asunique(self, group, sorted=False, set_mask=False):
...
+            # crashes on slices of atomgroups otherwise.
+            mask = inverse_unique_contiguous_1d_array(
+                       np.ascontiguousarray(self.ix), indices)
```
richardjgowers, line 828: I'd just push the ascontiguous call into the function rather than leave the sharp edge dangling

lilyminium:
@1krishnasharma welcome to MDAnalysis! Are you a GSOC or Outreachy applicant? If so, please introduce yourself on our mailing list (https://groups.google.com/g/mdnalysis-devel)

As a first step, have a look through the user guide (https://userguide.mdanalysis.org/stable/index.html) to get familiar with MDAnalysis, particularly the [quick start guide](https://userguide.mdanalysis.org/stable/examples/quickstart.html) and [contributing to MDAnalysis](https://userguide.mdanalysis.org/stable/contributing.html).

If you are applying to GSOC: Have a look at [<https://www.mdanalysis.org/2022/03/07/gsoc2022/>](https://www.mdanalysis.org/2022/03/07/gsoc2022/). If you want to apply to the Google Summer of Code, you will need to make a contribution to MDAnalysis first; have a look at the starter issues: https://github.com/MDAnalysis/mdanalysis/labels/GSOC%20Starter

If you are applying to Outreachy: have a look at our wiki (https://github.com/MDAnalysis/mdanalysis/wiki/Outreachy) and FAQ (https://github.com/MDAnalysis/mdanalysis/wiki/Outreachy-FAQ). 

If you have any further questions, please ask them on the mailing list or Discord instead of here. This page is for a specific pull request that fixes an issue in MDAnalysis. All discussion should be related to the PR.

ghost:
This is probably something simple but I am having issues pushing the latest changes.
Specfically:
```
 ! [rejected]            cython-inverse-array -> cython-inverse-array (non-fast-forward)
error: failed to push some refs to 'https://github.com/mdaviesNCL/mdanalysis.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
```

I attempted fetch and rebase based on the contibution guide and this seemed successful but does not stop the issue.

richardjgowers:
@mdaviesNCL hmm odd, that is usually the error message for when upstream has had additions.  Worst case, you can rename your local branch to something "safe", re-check out this remote branch and cherry pick (`git cherry-pick <HASH>`)  in your local commits onto the "safe" branch.

richardjgowers:
@mdaviesNCL are you still having trouble pushing up changes?

Issue #3592, opened 2022-04-02
BLD: install issues on M1 mac


tylerjereddy:
`python3 setup.py install --user` on the gcc compile farm M1 machine manages to build MDAnalysis, but then it tries to install deps, and `gsd` cannot be built from source:

```
Searching for gsd>=1.4.0
Reading https://pypi.org/simple/gsd/
Downloading https://files.pythonhosted.org/packages/bd/65/55fec6a39f02aae7c460ce2bbceb7159ce2a3ba461738dd66a8b15a643b0/gsd-2.5.1.tar.gz#sha256=76bf228b1d8e95e7d6a334e8cc7712c0bd8c256148007f7ce88a489c21996593
Best match: gsd 2.5.1
Processing gsd-2.5.1.tar.gz
Writing /var/folders/47/rvvzxnmx61x7n_btlv6p4tn800017l/T/easy_install-ta94os9x/gsd-2.5.1/setup.cfg
Running gsd-2.5.1/setup.py -q bdist_egg --dist-dir /var/folders/47/rvvzxnmx61x7n_btlv6p4tn800017l/T/easy_install-ta94os9x/gsd-2.5.1/egg-dist-tmp-r5m66yf9
Compiling gsd/fl.pyx because it depends on /Users/treddy/Library/Python/3.8/lib/python/site-packages/Cython/Includes/libc/string.pxd.
[1/1] Cythonizing gsd/fl.pyx
package init file 'gsd/test/__init__.py' not found (or not a regular file)
warning: no files found matching 'ChangeLog.md'
gsd/fl.c:11582:3: warning: 'tp_print' is deprecated [-Wdeprecated-declarations]
  0, /*tp_print*/
  ^
/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.8/include/python3.8/cpython/object.h:260:5: note: 'tp_print' has been explicitly marked deprecated here
    Py_DEPRECATED(3.8) int (*tp_print)(PyObject *, FILE *, int);
    ^
/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.8/include/python3.8/pyport.h:515:54: note: expanded from macro 'Py_DEPRECATED'
#define Py_DEPRECATED(VERSION_UNUSED) __attribute__((__deprecated__))
                                                     ^
1 warning generated.
gsd/fl.c:11582:3: warning: 'tp_print' is deprecated [-Wdeprecated-declarations]
[...]

PR #3595, opened 2022-04-02
BLD: improve M1 build support

tylerjereddy:
Deals with part of MDAnalysis/mdanalysis#3592 

* the changes here allow me to build MDAnalysis and its
dependencies on an M1 Mac on the gcc compile farm using
`python3 -m pip install --user .`

* the full test suite passes, except for 1 extra `xfail`
I've added in here--which is also documented by MDAnalysis/mdaencore#38
(it has apparently been failing on "exotic platforms"
for almost 5 years)

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/testsuite/MDAnalysisTests/analysis/test_encore.py
+++ b/testsuite/MDAnalysisTests/analysis/test_encore.py
@@ -446,7 +446,7 @@ class TestEncoreClustering(object):
                                                           "results: {0}".format(cluster_collection)
 
     @pytest.mark.xfail(platform.machine() == "arm64" and platform.system() == "Darwin",
-                       reason="Fails on M1 Mac")
+                       reason="see gh-3599")
     def test_clustering_three_ensembles_two_identical(self, ens1, ens2):
         cluster_collection = encore.cluster([ens1, ens2, ens1])
         expected_value = 40
--- a/testsuite/MDAnalysisTests/analysis/test_encore.py
+++ b/testsuite/MDAnalysisTests/analysis/test_encore.py
@@ -29,6 +29,7 @@ import numpy as np
 import sys
 import os
 import warnings
+import platform
 
 import pytest
 from numpy.testing import assert_equal, assert_almost_equal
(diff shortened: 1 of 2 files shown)


testsuite/MDAnalysisTests/analysis/test_encore.py
```diff
@@ -444,6 +445,8 @@ def test_clustering_two_ensembles(self, ens1, ens2):
         assert len(cluster_collection) == expected_value, "Unexpected " \
                                                           "results: {0}".format(cluster_collection)
 
+    @pytest.mark.xfail(platform.machine() == "arm64" and platform.system() == "Darwin",
+                       reason="Fails on M1 Mac")
```
tylerjereddy: could also reference the GH issue proper here, I just didn't search for it until after I opened the PR

IAlibay:
> (it has apparently been failing on "exotic platforms" for almost 5 years)

It's not been an issue on ppc64le or aarch64 (non-M1) though, it might be good to supplement that original issue with a new one, it's not immediately obvious to me why this is failing, is `encore_compile_args` being set properly (to -O1 / equivalent) for M1 builds? https://github.com/MDAnalysis/mdanalysis/blob/009c467511210c442d15e4014f02e62e07dc9e85/package/setup.py#L287

tylerjereddy:
> is encore_compile_args being set properly (to -O1 / equivalent) for M1 builds?

I tried that--it actually made the situation worse on the M1. I can post what happens if you do that in an issue later maybe.

IAlibay:
> > is encore_compile_args being set properly (to -O1 / equivalent) for M1 builds?
> 
> I tried that--it actually made the situation worse on the M1. I can post what happens if you do that in an issue later maybe.

How very odd. Yeah if you could either open a new issue or add to the old one that'd be great. Otherwise I'm happy to merge this.

testsuite/MDAnalysisTests/analysis/test_encore.py
IAlibay: If possible that'd be great.

IAlibay (approved):

tylerjereddy:
Ok, I added a more detailed description of the M1 situation for this test in MDAnalysis/mdaencore#28, and revised this PR to reference that issue instead of vaguely referring to a test failure on M1.

IAlibay (approved):

PR #3596, opened 2022-04-03
Fix issue 2703 (On-the-fly transformation for principal component projection)

### LINKED ISSUE #2703: On-the-fly transformation for principal component projection
The `analysis.pca` module calculates the principal components (PC) but does not have a built-in way to generate a trajectory that is projected on a PC (or "filtered by a PC").

I would like to be able to generate and analyze a trajectory `X_i(t)` that was projected on the i-th PC `p_i`:
```
X_i(t) = [(X(t) - <X>) . p_i] p_i + <X>
```
where `X(t)` is the 3N coordinate vector (the original trajectory), `<X>` is the mean (that was also used for PCA) and `.` is the inner product of the deviation from the mean with the PC.

This projection can be implemented as a *on-the-fly* transformation, which will make it easy to write out a filtered trajectory or just analyze it.

Filtering could be built into the PCA class similar to the [PCA.transform()](https://www.mdanalysis.org/mdanalysis/documentation_pages/analysis/pca.html#MDAnalysis.analysis.pca.PCA.transform) method but then one might be limited to just writing out a trajectory.

One possibility is to add a `PCA.project_single_frame(ts)` method that does the projection for a single time step. This method could then be used directly for the otf-transformation.

## Unsolved issues ##
One potential issue is that often the PCA is calculated only for a subset of atoms such as the CA. How does one meaningfully extend it to, say, all protein atoms? Should one just apply the same projection to all atoms in a residue, at least if the PCA was residue based? What should one do (or how should one fail) if the PCA was performed just on an arbitrary subset of atoms? 

Does it make sense to apply the projection to only a protein and let solvent just move as in the original trajectory?

rishu235:
Fixes #2703 

 - Added a function for inverse-PCA transform in pca.py: PCA.project_single_frame
 - The function takes an integer/list of component(s) onto which the structure is projected
 - The function returns a transformation function which can be applied to a TimeStep object
 - Added a test function which runs PCA.project_single_frame for all PCA components

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


testsuite/MDAnalysisTests/analysis/test_pca.py
```diff
@@ -145,6 +145,15 @@ def test_transform_universe():
     pca_test.transform(u2)
 
 
+def test_project_single_frame():
+    u = mda.Universe(PSF, DCD)
+    PSF_pca = PCA(u, select='backbone')
+    pca_run = PSF_pca.run()
+    func = PSF_pca.project_single_frame()
+    u.trajectory.add_transformations(func)
+    u.trajectory[-1]
```
orbeckst: Test actual numbers, not just that the code doesn't crash.

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,42 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+        Returns
+        -------
+        Returns the projecting function which transforms a timestep
+        """
+        if components is None:
+            components = range(self.results.p_components.shape[1])
+
+        def wrapped(ts):
+            xyz = self._atoms.positions.ravel() - self._xmean
+            dot = np.dot(xyz, self._p_components[:, components])
```
orbeckst: use a more expressive variable name than `dot` (which could also be confused with the numpy function)

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,42 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+            self._atoms.positions = np.reshape(
+                (np.dot(dot, self._p_components[:, components].T)
+                 + self._xmean), (-1, 3)
```
orbeckst: Possibly do everything in one line (better performance?) with [np.linalg.multi_dot](https://numpy.org/doc/stable/reference/generated/numpy.linalg.multi_dot.html).

Also, if we can avoid transposes somehow then that will also speed up — perhaps using [np.einsum](https://numpy.org/doc/stable/reference/generated/numpy.einsum).

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,42 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+        """Runs an inverse-PCA and generates
+           a trajectory that is projected on selected PCs
+
+        Run PCA class before using this function.
+        To project the trajectory onto the first principal component,
+            pca = PCA(universe, select='backbone').run()
+            func = pca.project_single_frame(components=0)
+
+        To apply the projection on-the-fly, may add as a transformation
+            u.trajectory.add_transformations(func)
```
orbeckst: The synopsis should describe in words what the method does.

Then add an 
```resT
Example
---------
```
section after the *Returns* section.

See the general structure of docs https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,42 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+
+        Parameters
+        ----------
+        components : int, array, optional
+             Components to be projected onto.
+             The default ``None`` maps onto all components.
+
+        Returns
+        -------
+        Returns the projecting function which transforms a timestep
```
orbeckst: At the end, add two empty lines and then
```reST
 
 
.. versionadded:: 2.2.0
```

package/MDAnalysis/analysis/pca.py
orbeckst: Use reST markup for code.

orbeckst (changes requested):
Looks like a very good start. There are some conceptual issues to be resolved https://github.com/MDAnalysis/mdanalysis/issues/2703#issuecomment-1086948647 namely
- How to deal with the situation when the number of PCA atoms is not the number of atoms in the trajectory ("extrapolation"/"applying the PCA filter to whole residues")?
- What error/sanity checks to add?

In the meantime I have a number of smaller comments. 

Please also check that the docs look the way that you expect them to by looking at https://mdanalysis--3596.org.readthedocs.build/en/3596/

orbeckst:
Eventually we'll also ask you to add yourself to AUTHORS and add an entry for the CHANGELOG.

orbeckst (changes requested):
To make progress on the PR, please do the following
* address my comments from my previous review
* **focus on only two approaches for applying the transformation**
  1. all atoms are in the PCA mode so the displacement vector can be applied to all atoms
  2. transform _all_ atoms inside a residue that has exactly one displacement vector per residue (e.g. from CA) by the same vector
* raise an exception for all cases that aren't (i) or (ii).

This will produce a useful contribution but will leave open the door for more refined approaches.

package/MDAnalysis/analysis/pca.py
rishu235: > Possibly do everything in one line (better performance?) with [np.linalg.multi_dot](https://numpy.org/doc/stable/reference/generated/numpy.linalg.multi_dot.html).

The latest commit does everything in one line. However, there are two issues with [np.linalg.multi_dot](https://numpy.org/doc/stable/reference/generated/numpy.linalg.multi_dot.html). 

1) When the trajectory is projected onto a single component, the multi_dot method will fail since it requires the PC component matrix to be 2-D.
2) multi_dot is in fact slower because it tries to figure out the order of dot products by calculating the cost for matrix multiplication. np.dot(np.dot(A, B), C) already computes the optimized cost because, in our case, A and C are  1-D arguments. 
> Also, if we can avoid transposes somehow then that will also speed up — perhaps using [np.einsum](https://numpy.org/doc/stable/reference/generated/numpy.einsum).

I tested einsum and saw no significant difference in performance. I continue to use self.transpose method. I looked at the inverse-PCA implementation in some other standard codes and they all seem to use the transpose, either in the PCA transformation or in the inverse-PCA transformation, depending on the conventions for matrix axes.

rishu235:
@orbeckst, thank you for your comments. I believe, I have addressed your initial comments, except for the Examples in the docstring and CHANGELOG file.

I will implement the class ii approach, along with exception error and the two unresolved comments mentioned above.

Is there a way I can see what the docstring is going to look like on MDAnalysis webpage, before I push commits?

orbeckst:
You can build the docs yourself with
```
python setup.py build_sphinx
```
as described in the [User Guide: building documentation](https://userguide.mdanalysis.org/stable/contributing_code.html#building-code-documentation).

I often just push and the let ReadTheDocs build the docs as part of the PR checks, namely the one called **docs/readthedocs.org:mdanalysis** builds and shows the docs at https://mdanalysis--3596.org.readthedocs.build/en/3596/ (the link under [Details](https://mdanalysis--3596.org.readthedocs.build/en/3596/) for the check).

orbeckst:
@rishu235 please ping me (i.e., a comment with `@orbeckst`) when you need me to review again. (I am quite busy but I'll try to review within a day.)

package/MDAnalysis/analysis/pca.py
orbeckst: Thanks for considering everything, I appreciate your thorough and well-reasoned reply!

rishu235:
@orbeckst please review the changes.

The code can now extrapolate the projection, onto selected PCs, to non-PCA atoms. The extrapolation is turned off by default, to make the function faster. It also throws an error to the user, if the number of displacement vector is not exactly one per residue.

I introduced a line break into the existing line of authors in CHANGELOG to keep the line width consistent. I hope that's doesn't go against the rules.

orbeckst:
I'll review in a day. In the meantime please resolve conflicts relative to develop. Breaking the contributors line in CHANGELOG is good --- thanks! (When you resolve the merge conflict, add your name at the end and break the line so thta your under the max character per line (79?) that's stated at top of CHANGELOG).

I also saw selections for resid. These are expensive. Try to do it without them.

rishu235:
@orbeckst To apply the transformation per residue, don't we need a residue selection?

(Fixed) Please ignore the changes shown in overview.rst file. Those changes are from another merged PR by someone else. I may have messed up in rebasing/resolving. I'll try to fix that tomorrow.

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,118 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+            one PCA atom in each residue of group. The atomselection must
+            not have any atoms thats were not part of the PCA transformation.
+            The default ``None`` does not extrapolate the projection
+            to non-PCA atoms. anchor cannot be ``None``
+            if group is not ``None``.
+
+        Returns
+        -------
+        function
+            The projecting function which transforms a timestep.
```
orbeckst: Document the signature of the returned function here.

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,118 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+
+        or apply the projection to the universe::
+
+            u.trajectory.add_transformations(project)
+
+
+        .. versionadded:: 2.2.0
+        """
+        if not self._calculated:
+            raise ValueError('Call run() on the PCA before projecting')
```
orbeckst, line 530: I think that should be `RuntimeError` — unless we have done things differently elsewhere.

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,118 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+        if components is None:
+            components = range(self.results.p_components.shape[1])
+
+        if group is not None:
+            if anchor is None:
+                raise ValueError('anchor cannot be None if group is not None')
+
+            for i in group.residues.resnums:
+                n_anchor = self._atoms.select_atoms(
+                           f'resnum {i} and {anchor}').n_atoms
```
orbeckst: Doing a selection for every frame in a trajectory is quite slow. Given that this is a method of a class, you have the opportunity to save state. I would cache the selection. Perhaps we _can_ leave performance improvements for later. 

However, there should be a faster way to get your answer than iterating over residues. Something along the line of 
```python
anchors = group.select_atoms(anchor)
group.residues == anchors.residues 
```
should indicate that you cover all residues. To check that you have exactly one anchor per residue, you could use `anchors.groupby("resindices")` (see [groupby()](https://docs.mdanalysis.org/2.1.0/documentation_pages/core/groups.html#MDAnalysis.core.groups.AtomGroup.groupby)) but that still requires loops. Perhaps it is good enough to just check
```python
np.all(anchors.resindices == group.resindices)
```

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,118 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+            non_pca = group.subtract(self._atoms)
+
+        def wrapped(ts):
+
+            if group is not None:
+                for i in group.residues.resnums:
+                    non_pca.select_atoms(f'resnum {i}').positions -= (
+                        self._atoms.select_atoms(f'resnum {i} and {anchor}')
+                            .positions
+                    )
```
orbeckst: Avoid selections in inner loops!!

Instead of using selections to get residues, directly use the groups, possibly using group set operations. 
```python
non_pca_with_anchor = (non_pca.residues - anchor.residues).atoms
```
Write the operationas a single numpy operation  instead of per-residue loop , perhaps along the lines of
```python
non_pca_with_anchor.positions -= anchor.positions
```
... or something along the line.

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,118 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+                        self._p_components[:, components].T)
+                 + self._xmean), (-1, 3)
+            )
+
+            if group is not None:
+                for i in group.residues.resnums:
+                    non_pca.select_atoms(f'resnum {i}').positions += (
+                        self._atoms.select_atoms(f'resnum {i} and {anchor}')
+                            .positions
+                    )
```
orbeckst: Avoid selections and loops, see above.

orbeckst (changes requested):
My primary concern is the use of selections and loops over residues which will lead to very poor performance. I added some ideas how you might be able to get around it. Ultimately it might also be good to cache the selections such as the `anchor` atom group and the `non_pca` group but let's worry about the obvious things first.

orbeckst:
> To apply the transformation per residue, don't we need a residue selection?

I think it only makes sense to apply the transformations to the same residues that the pca atoms are in. I think your code is already doing that. It just has to raise an error if it cannot do it and in my review I probably did not consider all cases where this may happen. That's something to think about and possibly address in another round of changes.

orbeckst:
By the way, I do appreciate that you're tackling a challenging issue. I know that it might feel that it goes terribly slowly. We are concerned with keeping standards in MDAnalysis high so the process takes as long as it has to. I would encourage you to start with your application in parallel.

package/MDAnalysis/analysis/pca.py
rishu235: `pca.py` has five more instances of this error, all called `ValueError`

rishu235:
@orbeckst I have managed to avoid loops and selections in the wrapped function. There still are atomselections and one `for` loop _outside_ the wrapped function. Since this part is run only once per function call (as opposed to every frame), the code is much faster now.

Thanks for pointing me in the right direction. I was earlier looping through residues in each frame, since each residue has a unique translation vector. The displacement vectors of `anchor` atoms can be grouped together. But, it was not trivial to apply it to a group of non-PCA atoms, since each displacement vector is used for varying number of atoms in each residue.

I now use the fact that atomgroups remain the same through the frames. I calculate an extrapolation matrix, comprising of zeros and ones, which converts the `anchor` atoms' displacment vector (shape[0] = number of residues in `group`) into a displacement vector for the non-PCA atoms (shape[0] = number of atoms in `group`).

Thank you the encouraging words. I am indeed finding it difficult, with the end-of-term duties, to go back to the GSoC proposal while the PR is open. Every kind gesture goes a long way. At the same time, it is good to see a demand for high standard. It gives me more confidence as an MDAnalysis user.

package/MDAnalysis/analysis/pca.py
orbeckst: Ok, then we stay consistent and keep it that way.

Btw, please let reviewers resolve conversations, at least that's our custom in MDAnalysis. It helps reviewers to keep track of points that they raised.

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,129 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+                                   res.atoms.n_atoms - n_common)
+                matrix_extrapolate[cumul_atoms_old:cumul_atoms_new,
+                                   res_index] = 1
+                cumul_atoms_old = cumul_atoms_new
+
+        if components is None:
+            components = range(self.results.p_components.shape[1])
+
+        def wrapped(ts):
+
```
orbeckst: remove empty line or add a one-line doc string

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,129 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+            if not np.isin(group.resindices, anchors_res_ids).all():
+                raise ValueError("Some residues in 'group'" +
+                                 " do not have an 'anchor'")
+            if not anchors.issubset(self._atoms):
+                raise ValueError("Some 'anchors' are not part of PCA class")
+
+            non_pca = group - self._atoms
+            pca_res_indices, pca_res_counts = np.unique(
+                self._atoms.resindices, return_counts=True)
+
```
orbeckst, line 552: Add a comment that explains what problem you're solving here and what the key idea is. This will be important if someone else wants to debug this code (or learn from it). They need to know your intentions.

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,129 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+            matrix_extrapolate = np.zeros((non_pca.n_atoms, anchors.n_atoms))
+            cumul_atoms_old = 0
+            for res in group.residues:
+                res_index = res.resindex
```
orbeckst: I'd just use res.resindex to keep the code more concise, the time saving from caching the attribute access are not worth it, I think.

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,129 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+                cumul_atoms_old = cumul_atoms_new
+
+        if components is None:
+            components = range(self.results.p_components.shape[1])
+
+        def wrapped(ts):
+
+            if group is not None:
+                anchors_pos = np.matmul(matrix_extrapolate, anchors.positions)
+                non_pca.positions -= anchors_pos
```
orbeckst: Make it one calculation.

orbeckst: An you can just use `@` which is the recommended operator.

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,129 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+                                   res_index] = 1
+                cumul_atoms_old = cumul_atoms_new
+
+        if components is None:
+            components = range(self.results.p_components.shape[1])
+
+        def wrapped(ts):
+
+            if group is not None:
+                anchors_pos = np.matmul(matrix_extrapolate, anchors.positions)
```
orbeckst: This is a clever solution to partition the translation vector over residues, basically a broadcast with uneven partitions.

Just out of curiosity, does numpy have a function for that @tylerjereddy ?

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,129 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+
+            xyz = self._atoms.positions.ravel() - self._xmean
+            self._atoms.positions = np.reshape(
+                (np.dot(np.dot(xyz, self._p_components[:, components]),
+                        self._p_components[:, components].T)
+                 + self._xmean), (-1, 3)
+            )
+
+            if group is not None:
+                anchors_pos = np.matmul(matrix_extrapolate, anchors.positions)
```
orbeckst: Why do you have to recalculate anchors_pos? It should be the same as before, so you can just use the variable that you defined previously.

(And then there's an excellent reason to _not_ put everything on one line... I'll edit my comment)

package/MDAnalysis/analysis/pca.py
orbeckst: Actually, do _not_ make it one calculation but keep `anchors_pos` because you can reuse them for the backtransformation below! (At least if I understood the code correctly.)

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,129 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+
+        def wrapped(ts):
+
+            if group is not None:
+                anchors_pos = np.matmul(matrix_extrapolate, anchors.positions)
+                non_pca.positions -= anchors_pos
+
+            xyz = self._atoms.positions.ravel() - self._xmean
+            self._atoms.positions = np.reshape(
+                (np.dot(np.dot(xyz, self._p_components[:, components]),
```
orbeckst, line 574: I _think_ you can use @ instead of dot(). Check the docs.

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,129 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+                        self._p_components[:, components].T)
+                 + self._xmean), (-1, 3)
+            )
+
+            if group is not None:
+                anchors_pos = np.matmul(matrix_extrapolate, anchors.positions)
+                non_pca.positions += anchors_pos
+
```
orbeckst: remove blank

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,129 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+            return ts
+
```
orbeckst: remove blank

testsuite/MDAnalysisTests/analysis/test_pca.py
```diff
@@ -145,6 +145,101 @@ def test_transform_universe():
     pca_test.transform(u2)
 
 
+def test_project_no_pca_run(u, pca):
+    pca_class = PCA(u, select=SELECTION)
+    with pytest.raises(ValueError):
```
orbeckst: also check the error message with the kwarg of pytes.raises

testsuite/MDAnalysisTests/analysis/test_pca.py
```diff
@@ -145,6 +145,101 @@ def test_transform_universe():
...
+        func = pca_class.project_single_frame()
+
+
+def test_project_none_anchor(u, pca):
+    group = u.select_atoms('resnum 1')
+    with pytest.raises(ValueError):
```
orbeckst: also check the error message with the kwarg of pytes.raises

testsuite/MDAnalysisTests/analysis/test_pca.py
```diff
@@ -145,6 +145,101 @@ def test_transform_universe():
...
+        func = pca.project_single_frame(0, group=group, anchor=None)
+
+
+def test_project_more_anchor(u, pca):
+    group = u.select_atoms('resnum 1')
+    with pytest.raises(ValueError):
```
orbeckst: also check the error message with the kwarg of pytest.raises

testsuite/MDAnalysisTests/analysis/test_pca.py
```diff
@@ -145,6 +145,101 @@ def test_transform_universe():
...
+        project = pca.project_single_frame(0, group=group, anchor='backbone')
+
+
+def test_project_less_anchor(u, pca):
+    group = u.select_atoms('all')
+    with pytest.raises(ValueError):
```
orbeckst: also check the error message with the kwarg of pytest.raises

testsuite/MDAnalysisTests/analysis/test_pca.py
```diff
@@ -145,6 +145,101 @@ def test_transform_universe():
...
+        project = pca.project_single_frame(0, group=group, anchor='name CB')
+
+
+def test_project_improper_anchor(u):
+    pca = PCA(u, select='name CA').run()
+    group = u.select_atoms('all')
+    with pytest.raises(ValueError):
```
orbeckst: also check the error message with the kwarg of pytest.raises

testsuite/MDAnalysisTests/analysis/test_pca.py
```diff
@@ -145,6 +145,101 @@ def test_transform_universe():
...
+        project = pca.project_single_frame(0, group=group, anchor='name N')
+
+
+def test_project_compare_projections():
+    u = mda.Universe(PSF, DCD)
```
orbeckst: make the universe a fixture
```python
@pytest.fixture
def universe():
   return mda.Universe(PSF, DCD)

def test_xxxxx(universe):
    u_copy = universe.copy()
    ...
```

Same for the other ones below.

testsuite/MDAnalysisTests/analysis/test_pca.py
```diff
@@ -145,6 +145,101 @@ def test_transform_universe():
...
+    pca = PCA(u, select=SELECTION).run()
+    project0 = pca.project_single_frame(0)
+    project1 = pca.project_single_frame(1)
+
+    u.trajectory[0]
+    coord0 = project0(u.trajectory.ts).positions.copy()
+    u.trajectory[0]
+    coord1 = project1(u.trajectory.ts).positions.copy()
+
+    with pytest.raises(AssertionError):
```
orbeckst: delete the line, 
```suggestion
```
just use 
```python
assert_almost_equal(coord0, coord1, decimal=5)
```
You don't have to wrap the assert_* functions.

testsuite/MDAnalysisTests/analysis/test_pca.py
```diff
@@ -145,6 +145,101 @@ def test_transform_universe():
...
+    distances_original = (
+        mda.lib.distances.self_distance_array(group.positions.copy())
+    )
+    distances_new = (
+        mda.lib.distances.self_distance_array(project(group).positions.copy())
+    )
+
+    if case == 1:
+        assert_almost_equal(distances_original, distances_new, decimal=5)
+    if case == 2:
```
orbeckst: make it `elif`

testsuite/MDAnalysisTests/analysis/test_pca.py
```diff
@@ -145,6 +145,101 @@ def test_transform_universe():
...
+        with pytest.raises(AssertionError):
+            assert_almost_equal(distances_original, distances_new, decimal=5)
```
orbeckst: add an `else` that raises NotImplementedError to catch any test misconfigurations

testsuite/MDAnalysisTests/analysis/test_pca.py
```diff
@@ -145,6 +145,101 @@ def test_transform_universe():
...
+                          ('resnum 1 and name N CA CB', 2)))
+def test_projet_extrapolate_translation(distance_sel, case):
+    u = mda.Universe(PSF, DCD)
+    pca = PCA(u, select='backbone').run()
+    group = u.select_atoms(distance_sel)
+    project = pca.project_single_frame(0, group=group,
+                                       anchor='resnum 1 and name CA')
+
+    distances_original = (
+        mda.lib.distances.self_distance_array(group.positions.copy())
```
orbeckst: I think you can remove the copy() here and below as the data are not altered and a new data structure is returned.

testsuite/MDAnalysisTests/analysis/test_pca.py
```diff
@@ -145,6 +145,101 @@ def test_transform_universe():
...
+    )
+    distances_new = (
+        mda.lib.distances.self_distance_array(project(group).positions.copy())
+    )
+
+    if case == 1:
+        assert_almost_equal(distances_original, distances_new, decimal=5)
+    if case == 2:
+        with pytest.raises(AssertionError):
```
orbeckst:
```suggestion
```

orbeckst: Or do you want to say that you expect case 2 to fail? If so, then you should use xfail marks.

testsuite/MDAnalysisTests/analysis/test_pca.py
orbeckst: Or do you expect this situation to fail? Then mark the test xfail. 

But please explain what this test is supposed to show.

orbeckst (changes requested):
Neat idea with the matrix multiplication to generate the correct shape of the trasnlation vector array.

Please have a look at my comments. Thanks.

orbeckst:
One comment on how we do reviews in MDA: We leave comments open until the _reviewer_ closes them. You as someone addressing comments would add a comment ("done" or an explanation or sometimes nothing and just fix the code) and the reviewer would close them when they are satisfied. This makes reviewing A LOT easier. We maybe reviewing 4 PRs at the same time and it's sometimes difficult to keep track of everything. Everything that makes your reviewers' life easier is good and will help them to spend time on your PR.

Don't worry, part of the process is to learn the customs. And you're doing an excellent job at discussing the PR with me.

package/MDAnalysis/analysis/pca.py
orbeckst: I wasn't clear here, I think: The return value *is* a function, so your original docs were correct., What I meant was to add in the text ("The projecting function ...") to write something like: The resulting function `f` takes as input a `Timestep` ts, `f(ts)` and returns the Timestep. 

Or something like that. It should be clear that it can be used as a transformation.

orbeckst:
@rishu235 how big does the extrapolation matrix become and how much of a computational bottleneck will it become if you apply it to a protein with, say 20,000 atoms? Can you do a quick `%timeit` of numpy matrix multiplication with random matrices and vectors for a few sizes? 

I like your approach conceptually and accept it as a solution to the problem. If the performance becomes a problem for larger systems then we will open a new issue for performance improvement.

orbeckst:
I think the PR is almost done. Once you address my remaining comments I will have one last request, and I'll tell you right away:

Could you demonstrate your approach by making a movie of the test trajectory (PSF, DCD) when projected on the first PC? Add the code for how you do it to the documentation (eventually it can be moved to the User Guide in the [PCA](https://userguide.mdanalysis.org/stable/examples/analysis/reduced_dimensions/pca.html) section). 

If you're adventurous, you can also try it for https://www.mdanalysis.org/MDAnalysisData/adk_equilibrium.html .

package/MDAnalysis/analysis/pca.py
rishu235: added comments to explain the purpose of matrix_extrapolate below this line

package/MDAnalysis/analysis/pca.py
rishu235: The first operation results in a _scalar_, when only one PC is selected, and gives an error with @

package/MDAnalysis/analysis/pca.py
rishu235: added the `@` operator.

> ... keep `anchors_pos` because you can reuse them for the backtransformation below! (At least if I understood the code correctly.)

`anchors` are (required to be) part of `self` atomgroup, which is projected with the inverse-PCA in the next line. So, need to recalculate later.

rishu235:
> how big does the extrapolation matrix become and how much of a computational bottleneck will it become if you apply it to a protein with, say 20,000 atoms?

It does become quite expensive relative to other parts. The matrix has the size of (`n_atoms`,`n_residues`). For a thousand residues with twenty atoms, one matrix multiplication took 261 ms on my computer. This is compared to 1.9 ms for projecting the PCA atoms (equivalent random matrix operations rather). ~~I should at least avoid one of the two matrix multiplications in the wrapped function~~ (Pushed changes). For 100 residues, each with twenty atoms, one matrix multiplication takes 1.58 ms, while the PCA atoms took 205 µs.


> Could you demonstrate your approach by making a movie of the test trajectory (PSF, DCD) when projected on the first PC? Add the code for how you do it to the documentation (eventually it can be moved to the User Guide in the [PCA](https://userguide.mdanalysis.org/stable/examples/analysis/reduced_dimensions/pca.html) section).

Yes, I was already thinking of that. I also have a trajectory filtered by aligning each residue. Yes, it's very expensive, but it looks super cool! @orbeckst Could I post the three movies (raw, anchor_filtered, and align_filtered) on the issue #2703  page, instead of here, and that will serve the next person working on the issue?

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,128 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+                                   res.resindex] = 1
+                cumul_atoms_old = cumul_atoms_new
+
+        if components is None:
+            components = range(self.results.p_components.shape[1])
+
+        def wrapped(ts):
+            """Projects a timestep"""
+            if group is not None:
+                non_pca.positions -= matrix_extrapolate @ anchors.positions
```
rishu235: Note to self: Avoid this matrix multiplication. Save the old anchor coordinates to calculates the displacement vector.
Finally,

```
non_pca.positions += matrix_extrapolate @ (anchors.positions_old - anchor.positions_new)
```

package/MDAnalysis/analysis/pca.py
orbeckst: Thanks for the explanation, I missed the fact that anchors are also being transformed as part of atomgroup.

package/MDAnalysis/analysis/pca.py
orbeckst: I was wrong, thanks to the explanation above. Needs to be recalculated.

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,130 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+        Run PCA class before using this function. For backbone PCA, run::
+
+            pca = PCA(universe, select='backbone').run()
+
+        Obtain a transformation function to project the
+        backbone trajectory onto the first principal component::
+
+            project = pca.project_single_frame(components=0)
+
+        To projet onto the first two components, run::
```
orbeckst:
```suggestion
        To project onto the first two components, run::
```

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,130 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+            The default ``None`` does not extrapolate the projection
+            to non-PCA atoms.
+
+        Returns
+        -------
+        function
+            The resulting function f(ts) takes as input a
+            :class:`~MDAnalysis.coordinates.base.Timestep` ts,
+            and returns ts with the projected structure
+
```
orbeckst, line 451: I think we need to include a warning:
```
.. warning::

   The transformation function takes a :class:`Timestep` as input because this is required for using
   it as a :ref:`transformations`. However, the transformations are all applied on the group and atoms
   of the Universe that was used for the PCA. It is *expected* that the `ts` is from the same Universe
   but this is currently not checked.
```

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,130 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+            The AtomGroup containing atoms to be projected.
+            The projection applies to the residues in ``group``.
+            The atoms in the PCA class are not affected by this argument.
+            The default ``None`` does not extrapolate the projection
+            to non-PCA atoms.
+
+        anchor : string, optional
+            The string to select the PCA atom whose displacement vector
+            is applied to non-PCA atoms in a residue. The atomselection
+            must have exactly one PCA atom in each residue of ``group``.
```
orbeckst: Insert:

"The `anchor` selection is applied to `group`."

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,130 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+            if not anchors.issubset(self._atoms):
+                raise ValueError("Some 'anchors' are not part of PCA class")
+
+            non_pca = group - self._atoms
+            pca_res_indices, pca_res_counts = np.unique(
+                self._atoms.resindices, return_counts=True)
+
+            # matrix_extrapolate is later multiplied to anchors' coordinates.
+            # multiplication shapes the array as appropriate for non_pca atoms
+            matrix_extrapolate = np.zeros((non_pca.n_atoms, anchors.n_atoms))
```
orbeckst: Set the dtype to float32 or int — I don't know which one will be faster. Our coordinates are all float32 so we don't want float64 (default) unless we need it.

Or, if you really need float64 then be explicit about it.

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,130 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+                                               group=all, anchor='name CA')
+
+        Finally, apply the transformation function to a timestep::
+
+            project(u.trajectory.ts)
+
+        or apply the projection to the universe::
+
+            u.trajectory.add_transformations(project)
+
```
orbeckst, line 525: Can you add a section "Background" where you essentially write out the equations for the transformation?

Also write out the mathematics of your extrapolation trick.

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,130 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+                n_common = pca_res_counts[np.where(
+                           pca_res_indices == res.resindex)][0]
+                cumul_atoms_new = (cumul_atoms_old +
+                                   res.atoms.n_atoms - n_common)
+                matrix_extrapolate[cumul_atoms_old:cumul_atoms_new,
+                                   res.resindex] = 1
+                cumul_atoms_old = cumul_atoms_new
+
+        if components is None:
+            components = range(self.results.p_components.shape[1])
```
orbeckst: Using a `range` object works for indexing a np.array but is ~50x slower for me. I'd do
```suggestion
            components = np.arange(self.results.p_components.shape[1])
```
and store the array.

```python
In [1]: import numpy as np

In [2]: a = np.arange(100)

In [3]: a[range(4)]
Out[3]: array([0, 1, 2, 3])

In [4]: a[range(29)]
Out[4]:
array([ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11, 12, 13, 14, 15, 16,
       17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28])

In [5]: %timeit a[range(29)]
9.29 µs ± 61.7 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [6]: ix = np.arange(29)

In [7]: %timeit a[ix]
209 ns ± 7.5 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [8]: 9.29e-6/209e-9
Out[8]: 44.449760765550245
```

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,130 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+
+        def wrapped(ts):
+            """Projects a timestep"""
+            if group is not None:
+                anchors_coords_old = anchors.positions
```
orbeckst, line 570: (For a moment I thought you need a copy() here but then I remembered that AtomGroup.positions is *always* a new array because we use fancy indexing `Timestep.positions[ag.ix]`.)

All good!

package/MDAnalysis/analysis/pca.py
orbeckst: Well, if the current code is robust then we go with this code! Thanks for checking.

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,130 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+            xyz = self._atoms.positions.ravel() - self._xmean
+            self._atoms.positions = np.reshape(
+                (np.dot(np.dot(xyz, self._p_components[:, components]),
+                        self._p_components[:, components].T)
+                 + self._xmean), (-1, 3)
+            )
+
+            if group is not None:
+                non_pca.positions += (matrix_extrapolate @
+                                      (anchors.positions - anchors_coords_old))
```
orbeckst: Smart! 

I look forward to reading the equations in the Background section.

I have a feeling that we could get the rearrangement more cheaply but I still like the very clean formulation as a matrix multiplication.

testsuite/MDAnalysisTests/analysis/test_pca.py
```diff
@@ -41,6 +41,12 @@ def u():
     return mda.Universe(PSF, DCD)
 
 
+@pytest.fixture(scope='module')
```
orbeckst: I think this should be function scoped because the projection will change the underlying universe. Thus the tests cannot be safely run in parallel. Creating the universe is not terribly expensive so rather be safe (make a new one for every test) than sorry (trying to be fast but sometimes wrong).

testsuite/MDAnalysisTests/analysis/test_pca.py
```diff
@@ -145,6 +151,113 @@ def test_transform_universe():
...
+
+def test_project_invalid_anchor(u):
+    pca = PCA(u, select='name CA').run()
+    group = u.select_atoms('all')
+    with pytest.raises(ValueError) as exc:
+        project = pca.project_single_frame(0, group=group, anchor='name N')
+    assert "Some 'anchors' are not part of PCA class" in str(exc.value)
+
+
+@pytest.mark.xfail
```
orbeckst: Now I understand the test a bit better. You expect this to be different. Then you do _not_ use xfail (which we use when we know that something hasn't been fixed).

Instead, invert the assertion below.

testsuite/MDAnalysisTests/analysis/test_pca.py
```diff
@@ -145,6 +151,113 @@ def test_transform_universe():
...
+    # projections along different PCs should be different
+    pca = PCA(u_project, select=SELECTION).run()
+    project0 = pca.project_single_frame(0)
+    project1 = pca.project_single_frame(1)
+
+    u_project.trajectory[0]
+    coord0 = project0(u_project.trajectory.ts).positions
+    u_project.trajectory[0]
+    coord1 = project1(u_project.trajectory.ts).positions
+    assert_almost_equal(coord0, coord1, decimal=5)
```
orbeckst: Need to rewrite
```python
assert not np.allclose(coord0, coord1)
```

testsuite/MDAnalysisTests/analysis/test_pca.py
```diff
@@ -145,6 +151,113 @@ def test_transform_universe():
...
+
+def test_project_reconstruct_whole(u, u_project):
+    # structure projected along all PCs
+    # should be same as the original structure
+    pca = PCA(u_project, select=SELECTION).run()
+    project = pca.project_single_frame()
+
+    coord_original = u.trajectory.ts.positions
+    coord_reconstructed = project(u_project.trajectory.ts).positions
+    assert_almost_equal(coord_original, coord_reconstructed, decimal=5)
```
orbeckst: recommended (from https://numpy.org/doc/stable/reference/routines.testing.html )
```suggestion
    assert_allclose(coord_original, coord_reconstructed, rtol=1e-5)
```
We still have mostly assert_almost_equal but eventually we are going to replace them.

testsuite/MDAnalysisTests/analysis/test_pca.py
```diff
@@ -145,6 +151,113 @@ def test_transform_universe():
...
+    # if both projections are along the same PC(s).
+    pca = PCA(u_project, select=SELECTION).run()
+
+    project_first = pca.project_single_frame(n1)
+    project_second = pca.project_single_frame(n2)
+
+    u_project.trajectory[0]
+    coord1 = project_first(u_project.trajectory.ts).positions.copy()
+    coord2 = project_second(u_project.trajectory.ts).positions.copy()
+    assert_almost_equal(coord1, coord2, decimal=5)
```
orbeckst: use assert_allclose

testsuite/MDAnalysisTests/analysis/test_pca.py
```diff
@@ -145,6 +151,113 @@ def test_transform_universe():
...
+    coord_original = u.trajectory.ts.positions
+    coord_reconstructed = project(u_project.trajectory.ts).positions
+    assert_almost_equal(coord_original, coord_reconstructed, decimal=5)
+
+
+@pytest.mark.parametrize(
+    ("n1", "n2"),
+    [(0, 0), (1, 1), ([0, 1], [0, 1]),
+     pytest.param(0, 1, marks=pytest.mark.xfail),
+     pytest.param(1, 0, marks=pytest.mark.xfail)]
```
orbeckst: These two are always expected to fail. But if for some reason they pass, the tests will just show "XPASS" and not fail so these tests are not alerting us to any errors that might be introduced in the code. Just remove the two cases. If you want to test them because you're concerned that something could happen to the code where suddenly the two frame projections are the same, write a new test.

orbeckst: (You did use marks really nicely!)

testsuite/MDAnalysisTests/analysis/test_pca.py
orbeckst: How important is this test, i.e., when do you think it would fail to pass, i.e., what would have gone wrong if `coord1 == coord2`?

testsuite/MDAnalysisTests/analysis/test_pca.py
```diff
@@ -145,6 +151,113 @@ def test_transform_universe():
...
+    u_project.trajectory[0]
+    coord1 = project_first(u_project.trajectory.ts).positions.copy()
+    coord2 = project_second(u_project.trajectory.ts).positions.copy()
+    assert_almost_equal(coord1, coord2, decimal=5)
+
+
+@pytest.mark.parametrize(
+    "sel",
+    ['resnum 1 and name CA CB CG',
+     pytest.param('resnum 1 and name N CB CG', marks=pytest.mark.xfail)]
```
orbeckst: Is this something that should eventually work? Then leave it here with xfail.

However, this is one of the cases that you catch with a ValueError. Therefore, add a separate test `test_project_ValueError_toomany_anchors` or something like that. Importantly, the test should be testing (and making it clear) that it wants to see an exception because you're constructing the failure case.

testsuite/MDAnalysisTests/analysis/test_pca.py
```diff
@@ -145,6 +151,113 @@ def test_transform_universe():
...
+)
+def test_projet_extrapolate_translation(u_project, sel):
```
orbeckst: project --> project

orbeckst (changes requested):
Please see comments, sorry didn't have as much time to review as I wanted.

package/MDAnalysis/analysis/pca.py
rishu235: added below

package/MDAnalysis/analysis/pca.py
rishu235: Added before the Examples section, going by the section order at style guide at https://numpydoc.readthedocs.io/en/latest/format.html#notes

package/MDAnalysis/analysis/pca.py
rishu235: done

testsuite/MDAnalysisTests/analysis/test_pca.py
rishu235: Done. It's called u_fresh; gives a fresh universe to each function.

testsuite/MDAnalysisTests/analysis/test_pca.py
rishu235: verifies that the projections along two different PCs are different. Added "assert not np.allclose()" because we want it to detect future changes in the code, in case it returns the same timestep (or the same projection).

testsuite/MDAnalysisTests/analysis/test_pca.py
rishu235: no case numbers in the new version. changed second if to else.

testsuite/MDAnalysisTests/analysis/test_pca.py
rishu235: replaced this if by else

testsuite/MDAnalysisTests/analysis/test_pca.py
rishu235: Done. All the copy() are removed, except where the function is applied twice in succession.

testsuite/MDAnalysisTests/analysis/test_pca.py
rishu235: Removed parametrization of case number in this test, since expected bugs will be caught by other tests (twice_projection and compare_projection)

package/MDAnalysis/analysis/pca.py
rishu235: Done

package/MDAnalysis/analysis/pca.py
rishu235: Added

package/MDAnalysis/analysis/pca.py
rishu235: Set the dtype to np.int8, the performance is comparable to the default allocation. int may improve memory efficiency though.
I tested a few types on my machine, and the default float64 is as fast as/faster than float32.

package/MDAnalysis/analysis/pca.py
rishu235: Fixed. I think, it was a typo! An oversight that _seemed_ to work.

package/MDAnalysis/analysis/pca.py
rishu235: Added equations to show PCA transformation and inverse-PCA transformation. Explained the extrapolation trick too.

testsuite/MDAnalysisTests/analysis/test_pca.py
rishu235: Done.

testsuite/MDAnalysisTests/analysis/test_pca.py
rishu235: Will catch if the projection does nothing, or is insensitive to the `components` variable.

testsuite/MDAnalysisTests/analysis/test_pca.py
rishu235: Done

testsuite/MDAnalysisTests/analysis/test_pca.py
rishu235: Done

testsuite/MDAnalysisTests/analysis/test_pca.py
rishu235: Done

testsuite/MDAnalysisTests/analysis/test_pca.py
rishu235: Thank you :)

And now I've removed them all :/
because we don't want XPASS

testsuite/MDAnalysisTests/analysis/test_pca.py
rishu235: No, this should always fail.
The purpose is to show that anchor atom (and only the anchor atom) indeed works like an anchor. Removed the second test case though since the errors will be caught by other tests anyway.

testsuite/MDAnalysisTests/analysis/test_pca.py
rishu235: Done

testsuite/MDAnalysisTests/analysis/test_pca.py
orbeckst: Furthermore, `ATomGroup.positions` is already a copy!

package/MDAnalysis/analysis/pca.py
orbeckst: You can even make the matrix Boolean, which makes the matrix multiplication a bit slower compared to int8, at least when I played with 100x100 matrices. 

For me, the fastest appears to be everything float32 or float64:
```
bool @ f64: 14.9 µs ± 278 ns
bool @ f32: 13.8 µs ± 735 ns
f64 @ f64 : 3.4 µs ± 143 ns
f32 @ f64:  11 µs ± 232 ns
f32 @ f32 : 3.56 µs ± 269 ns
i8 @ f32 : 11.1 µs ± 411 ns
i8 @ f64 : 13.3 µs ± 2.23 µs
```
Given that speed is an issue, I would say we do not care about memory but ensure that **there is no implicit type casting** involved: Ensure that **all terms in the matrix multiplication have the same type**.

orbeckst:
```python
import numpy as np
rng = np.random.default_rng()
Mbool = rng.random((100, 100)) > 0.5
Mf64 = Mbool.astype(np.float64)
Mf32 = Mbool.astype(np.float32)
vf64 = rng.random(100)
vf32 = v.astype(np.float32)

%timeit Mbool @ vf64
%timeit Mf32 @ vf32
...
```

orbeckst (changes requested):
This looks really good, ❤️ it!

The only remaining discussion point is the speed of the matrix multiplication. I played around as detailed in the comment and it seems to me that we should avoid any kind of implicit type casting here. Could you please write the code so that the type of the extrapolation matrix is the same type as the displacement vector? I don't know what the type of the displacement vector is — I *assumed* float32 but maybe some upcasting to float64 happens somewhere. 

Prioritize speed over memory efficiency.

testsuite/MDAnalysisTests/analysis/test_pca.py
rishu235: Is it? This may require a further investigation of the `positions` method. I have a `.copy()` method in test_project_twice_projection (line 230) and I absolutely need it. ~~(Copy in line 231 can be removed though. I'm on it.)~~

I also compared the actual numbers. The test works as intended with`project_first(u_fresh.trajectory.ts).positions.copy()`, but not with `project_first(u_fresh.trajectory.ts).positions`

package/MDAnalysis/analysis/pca.py
rishu235: I see! I was playing with the dtype of only one matrix. No wonder, float32 was slower than float64, since the second matrix was implicitly defined as float64. Thank you!

Displacement vector is float32, so I've made `matrix_extrapolate` float32

rishu235:
@orbeckst Looping through 97 frames in a protein trajectory, testfiles (PSF, DCD) with 214 residues and 3341 atoms, takes

1) 13.3 ms ± 280 µs with no transformation
raw trajectory

2) 26.8 ms ± 3.44 ms with `project(0)`
inverse-PCA for the PCA atoms (backbone) only and non-PCA atoms untouched

3) 66.3 ms ± 8.21 ms with `project(0, group=all, anchor='name CA')`
inverse-PCA for the PCA atoms and extended to non-PCA atoms by anchoring to a PCA atom in each residue.

4) Not implemented yet
inverse-PCA for the PCA atoms and extended to non-PCA atoms by aligning the PCA-atoms in each residue.

I have made ten-second movies. You may have to download them for a good resolution. Please watch the [movies](https://uofc-my.sharepoint.com/:f:/g/personal/rishabh_shukla_ucalgary_ca/EpBePX_TWP9LpNjt4XwoCeIBkqfptRKANXmOMjscODyPpQ?e=yfWYy1) in the order (and after the fourth movie, remember to pick up your jaw from the floor!)

orbeckst:
The movies look very cool. Can you describe what the differences between them are? From the file names I only understand "raw".

EDIT: Reading your post again makes things clearer. I was confused by "inverse" -- for me this is just "projection on PC".   Is 4 taking the initial frame for the "non PCA atoms" to align?

orbeckst (changes requested):
I have a number of points to be addressed, after thinking things through and reading your docs in detail

# speeding up the extrapolation
I think I have an idea how to avoid the matrix multiplication, which should be much faster. Your matrix contains mostly zeros and what it really does is describe a way to *fancy index* the displacement vector `v`: Let's say you have
```python
>>> v = np.arange(4) * 10
array[0, 10, 20, 30]
```
and you want to make it 
```python
array([ 0,  0, 10, 10, 10, 20, 20, 30, 30, 30, 30, 30])
```
then you can either set up your matrix that starts as (EDIT: I incorrectly wrote the transpose of the matrix ... but you know what I mean)
```
1 0 0 0 0 0 0 ...
1 0 0 0 0 0 0 ...
0 1 0 0 0 0 0 ...
0 1 0 0 0 0 0 ...
0 1 0 0 0 0 0 ...
0 0 1 0 0 0 0 ...
...
```
and matrix multiply or you can fancy index:
```
v_expanded = v[[0, 0, 1, 1, 1, 2, 2, 3, 3, 3, 3, 3]]
```

Generating the fancy-index vector should be easy as you've essentially done it already.

Please test this approach and get some timings. I am pretty sure its faster and it definitely conserves memory.

# doc consistency
The User Guide's [PCA tutorial](https://userguide.mdanalysis.org/stable/examples/analysis/reduced_dimensions/pca.html#Visualising-projections-into-a-reduced-dimensional-space) already lays out the basic equations. In particular, what you call PC_score is called weight w_i there. It also uses **u_i** for PC_i. Could you please change your equations to be consistent with the User Guide. 

Note that the User Guide incorrectly bold-faces w_i(t), implying it is a vector. However, it's a scalar and should be set in standard math mode italics. (The equation for the projection with the outer produce is not correct because *for a given time t*, the multiplication is simply scalar-vector multiplication *w_i(t)* **u_i**. The Python implementation does correctly use `np.outer` because the weights `trans1` are _all_ times. )

orbeckst:
For generating the index vector for the expansion look into the various numpy index tricks such as [np.r_](https://numpy.org/doc/stable/reference/generated/numpy.r_.html) and similar index generation objects.

rishu235:
>Is 4 taking the initial frame for the "non PCA atoms" to align?

The fourth approach, not part of this PR (right?), aligns the raw PCA atoms to projected PCA atoms and applies this transformation to the non-PCA atoms. Basically, it accounts for displacement _and rotation_.

Fancy indexing is simpler and faster. The time taken to loop through 97 frames are now
- 13.3 ms ± 99.3 µs for raw
- 21.1 ms ± 749 µs for projected backbone
- 36.4 ms ± 1.04 ms for projected backbone and projection extended by anchoring to CA atoms

The first two are more or less the same, as expected. While, the third one is almost twice as fast.

Also made the equations consistent with [PCA tutorial](https://userguide.mdanalysis.org/stable/examples/analysis/reduced_dimensions/pca.html#Visualising-projections-into-a-reduced-dimensional-space).

orbeckst:
Great work. I might not get to reviewing over the weekend (work/life balance…) so please have some patience with me. Maybe another coredev has some comments.

rishu235:
Sure! Thank you for your prompt responses :)

> I might not get to reviewing over the weekend (work/life balance…)

I'm actually glad to hear that! Believe it or not, I had (almost) decided _not_ to apply for GSoC. Now, I'll think about it again.

orbeckst:
> I'm actually glad to hear that! Believe it or not, I had (almost) decided not to apply for GSoC. Now, I'll think about it again.

Please consider applying! You have shown that you have the initiative to tackle difficult problems, the creativity to come up with solutions, the humility to listen to advice and criticism, and the professionalism needed to engage in productive discussions that ultimately lead to great code that will be useful to many others.

I have very much enjoyed the discussion on this PR and I am thrilled to see it getting to completion. I have very little time in my day-to-day job so I am trying to make time when I can to review. Therefore, you'll get comments at odd hours, and sometimes the comments are fairly terse because that's all I can squeeze into the gap between two meetings or teaching. Perhaps I should have made clear that we don't expect you to respond immediately, we understand that people have lives. 

I really want to see your excellent work completed — and I would very much look forward reading a GSOC proposal from you. (Feel free to ask questions about the proposal on the developer list. You can also submit the proposal already and embed a link to a draft such as Google Doc or Overleaf.)

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,174 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+                raise ValueError("More than one 'anchor' found in residues")
+            if not np.isin(group.resindices, anchors_res_ids).all():
+                raise ValueError("Some residues in 'group'" +
+                                 " do not have an 'anchor'")
+            if not anchors.issubset(self._atoms):
+                raise ValueError("Some 'anchors' are not part of PCA class")
+
+            # non_pca has "all" the atoms in residues of `group`. This makes
+            # sure that extrapolation works on residues, not random atoms.
+            non_pca = group.residues.atoms[:] - self._atoms
```
orbeckst: No need for slicing
```suggestion
            non_pca = group.residues.atoms - self._atoms
```

orbeckst (approved):
Looking excellent, great work.

(You can raise an issue to add the fitting procedure for extrapolation and if you already have code then please open a PR. Reviewing might be a bit slow because we prioritized GSOC at the moment and you already have an excellent PR done.)

rishu235:
Thank you for the review and approval. Does this PR need approvals from other coredevs before merging?

> I have very much enjoyed the discussion on this PR and I am thrilled to see it getting to completion. I have very little time in my day-to-day job so I am trying to make time when I can to review. Therefore, you'll get comments at odd hours...

I wasn't sure how my previous comment would be read. But you do get the point. My recent disinterest is mainly on me (more blah blahs to come in the google group). I am hoping I would again be excited towards the application soon.

> I really want to see your excellent work completed — and I would very much look forward reading a GSOC proposal from you. 

I too have been enjoying the review process and I appreciate all the encouragement and support from you ❤️

> (You can raise an issue to add the fitting procedure for extrapolation and if you already have code then please open a PR. Reviewing might be a bit slow because we prioritized GSOC at the moment and you already have an excellent PR done.)

I have the quick-and-dirty code in a notebook, but it will take some effort to make it nice with the appropriate tests and documentation. I will work on that after a couple of weeks, I think.

Meanwhile, I will try to increase the overlap between my GSoC proposal and my PhD project, which is the main constraint with respect to time and priority. If the overlap looks good enough to me and my supervisor, I will run it by someone via the google group later this week.

orbeckst:
There's one failure in the 32 bit windows run https://dev.azure.com/mdanalysis/mdanalysis/_build/results?buildId=3458&view=logs&j=8c1244f7-4387-51eb-cf10-2a1290f32f77&t=6cc8a0ec-9d5a-5758-2f06-db7cff0afb7f

There's an implicit int64 -> int32 case (maybe in repeat()) that needs to be made explicit. 

Sorry, I'll have to raise this as a blocking review. Maybe @tylerjereddy or @IAlibay has a good idea.

package/MDAnalysis/analysis/pca.py
```diff
@@ -414,6 +414,174 @@ def transform(self, atomgroup, n_components=None, start=None, stop=None,
...
+
+            non_pca_atoms = np.array([], dtype=np.int64)
+            for res in group.residues:
+                # n_common is the number of pca atoms in a residue
+                n_common = pca_res_counts[np.where(
+                           pca_res_indices == res.resindex)][0]
+                non_pca_atoms = np.append(non_pca_atoms,
+                                          res.atoms.n_atoms - n_common)
+            # index_extrapolate records the anchor number for each non-PCA atom
+            index_extrapolate = np.repeat(np.arange(anchors.atoms.n_atoms),
```
orbeckst, line 561: There's one failure in the 32 bit windows run https://dev.azure.com/mdanalysis/mdanalysis/_build/results?buildId=3458&view=logs&j=8c1244f7-4387-51eb-cf10-2a1290f32f77&t=6cc8a0ec-9d5a-5758-2f06-db7cff0afb7f

There's an implicit int64 -> int32 case (maybe in repeat()) that needs to be made explicit.

orbeckst (changes requested):
We need to figure out why the 32 bit windows fails.

package/MDAnalysis/analysis/pca.py
rishu235: Fixed now.

If I understand correctly, it was because I had an array explicitly defined as np.int64, which later required an implicit conversion on a 32-bit machine. I changed the array creation line to int. The new declared array does not require a conversion later.

orbeckst: you solved it – test passes!

orbeckst (approved):

orbeckst:
I'll squash-merge the PR in about an hour.

(After getting a coffee ☕ and probably after my next meeting.)

orbeckst:
Congratulations @rishu235 , you added a really neat feature to MDAnalysis! 💯 🏅 

I closed #2703 because you solved the key problems but your idea of using structural alignment is really cool so _please_ just open a follow-up issue where you mention #2703 and this PR and then say what you would want to see next. You don't have to open a PR — you can do it if/whenever you want to, or someone else could attempt it and you could review it and give advice. Either way, we use issues as reminders for how to move forward.

Eventually we'd also want to update the [User Guide: PCA](https://userguide.mdanalysis.org/stable/examples/analysis/reduced_dimensions/pca.html) page and include an example in the [User Guide: On-the-fly transformations](https://userguide.mdanalysis.org/stable/trajectories/transformations.html).

PR #3606, opened 2022-04-06
Fix FrameIteratorIndices doesn't rewind (#3416)

### LINKED ISSUE #3416: FrameIteratorIndices doesn't rewind whilst FrameIteratorSliced does
Related to #3415 

## Expected behavior ##

See https://github.com/MDAnalysis/mdanalysis/pull/3415#discussion_r703080785

Iterating through FrameIteratorSliced should leave u.trajectory at the same point as if I iterated through FrameIteratorIndices.

## Actual behavior ##

The former rewinds whilst the latter doesn't.

I might be missing a pythonic reason (which I'd be happy to accept), just bringing this up because of discussions on #3415 

```python
import MDAnalysis as mda

u = mda.Universe(TPR, XTC)

iterator = u.trajectory[[0,1,2]]

for ts in iterator:
    pass

print(u.trajectory.ts)
```
Ouput: < Timestep 2 with unit cell dimensions [79.98359 79.98359 79.98358 60.      60.      90.     ] >

Whilst:

```python
import MDAnalysis as mda

u = mda.Universe(TPR, XTC)

for ts in u.trajectory[:3]
    pass

print(u.trajectory.ts)
```
Ouput: < Timestep 0 with unit cell dimensions [80.017006 80.017006 80.017006 60.       60.       90.      ] >

- Which version of Python (`python -V`)? 3.9

megosato:
Fixes #3416

- FrameIteratorIndices now rewinds back to 0 and is consistent with
  FrameIteratorSliced and FrameIteratorAll

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -29,6 +29,7 @@ Fixes
   * Fixed DumpReader triclinic box dimensions reading. (Issue #3386, PR #3403)
   * Updated the badge in README of MDAnalysisTest to point to Github CI Actions
   * Fixed FrameIteratorIndices doesn't rewind. (Issue #3416)
+
 Enhancements
   * Improves the RDKitConverter's accuracy (PR #3044): AtomGroups containing
     monatomic ion charges or edge cases with nitrogen, sulfur, phosphorus and
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -19,6 +19,8 @@ The rules for this file:
  * 2.2.0
 
 Fixes
+  * Fixed LinearDensity not working with grouping='residues' or 'segments'
+    (Issue #3571, PR #3572)
   * MOL2 can read files without providing optional columns:
     subst_id, subst_name and charge. (Issue #3385, PR #3598)
   * When converting an AtomGroup to an RDKit molecule (PR #3044):
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -13,8 +13,8 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" http://semver.org
 
 ------------------------------------------------------------------------------
-??/??/?? IAlibay, BFedder, inomag, Agorfa, aya9aladdin, shudipto-amin, cbouy, HenokB, umak1106
-
+??/??/?? IAlibay, BFedder, inomag, Agorfa, aya9aladdin, shudipto-amin, cbouy, 
+         HenokB, umak1106, megosato
  * 2.2.0
 
 Fixes
@@ -28,7 +28,7 @@ Fixes
     (Issue #2666, PR #3565)
   * Fixed DumpReader triclinic box dimensions reading. (Issue #3386, PR #3403)
   * Updated the badge in README of MDAnalysisTest to point to Github CI Actions
-
+  * Fixed FrameIteratorIndices doesn't rewind. (Issue #3416)
 Enhancements
   * Improves the RDKitConverter's accuracy (PR #3044): AtomGroups containing
     monatomic ion charges or edge cases with nitrogen, sulfur, phosphorus and
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -14,7 +14,8 @@ The rules for this file:
 
 ------------------------------------------------------------------------------
 ??/??/?? IAlibay, BFedder, inomag, Agorfa, aya9aladdin, shudipto-amin, cbouy, 
-         HenokB, umak1106, megosato
+         HenokB, umak1106, tamandeeps, megosato
+
  * 2.2.0
 
 Fixes
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -13,12 +13,14 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" http://semver.org
 
 ------------------------------------------------------------------------------
-??/??/?? IAlibay, BFedder, inomag, Agorfa, aya9aladdin, shudipto-amin, cbouy, 
-         HenokB, umak1106, tamandeeps, megosato
+??/??/?? IAlibay, BFedder, inomag, Agorfa, aya9aladdin, shudipto-amin, cbouy,
+         HenokB, umak1106, tamandeeps, Mrqeoqqt, megosato
 
  * 2.2.0
 
 Fixes
+  * MOL2 can read files without providing optional columns:
+    subst_id, subst_name and charge. (Issue #3385, PR #3598)
   * When converting an AtomGroup to an RDKit molecule (PR #3044):
     - the atoms are now in the same order
     - the output of `atom.GetMonomerInfo().GetName()` now follows the
--- a/testsuite/MDAnalysisTests/coordinates/base.py
+++ b/testsuite/MDAnalysisTests/coordinates/base.py
@@ -384,7 +384,7 @@ class BaseReaderTest(object):
 
         for i, ts in enumerate(transformed):
             assert_almost_equal(ts.positions, idealcoords[i], decimal=ref.prec)
-    
+
     def test_transformations_slice(self, ref, transformed):
         # Are the transformations applied when iterating over a slice of the trajectory?
         v1 = np.float32((1,1,1))
@@ -510,22 +510,15 @@ class MultiframeReaderTest(BaseReaderTest):
             assert_timestep_almost_equal(ts,
                                          ref.iter_ts(ref.aux_lowf_frames_with_steps[i]),
                                          decimal=ref.prec)
-
-    def test_iter_rewinds(self, reader):
-        for ts_indices in reader[[0, 1, 2]]:
+    @pytest.mark.parametrize("accessor", [
+              lambda traj: traj[[0, 1, 2]], 
+              lambda traj: traj[:3],
+              lambda traj: traj],
+              ids=["indexed", "sliced", "all"])
+    def test_iter_rewinds(self, reader, accessor):
+        for ts_indices in accessor(reader):
             pass
-        for ts_sliced in reader[:3]:
-            pass
-        assert_equal(ts_indices.frame, ts_sliced.frame,
-                     err_msg="FrameIteratorIndices and Sliced do not match")
-        assert_equal(ts_indices.frame, 0,
-                     err_msg="FrameIteratorIndices did not rewind")
-        assert_equal(ts_sliced.frame, 0,
-                     err_msg="FrameIteratorSliced did not rewind")
-        for ts_all in reader:
-            pass
-        assert_equal(ts_all.frame, 0,
-                     err_msg="FrameIteratorAll did not rewind")
+        assert_equal(ts_indices.frame, 0)
 
     #  To make sure we not only save the current timestep information,
     #  but also maintain its relative position.
--- a/testsuite/MDAnalysisTests/coordinates/base.py
+++ b/testsuite/MDAnalysisTests/coordinates/base.py
@@ -384,7 +384,7 @@ class BaseReaderTest(object):
 
         for i, ts in enumerate(transformed):
             assert_almost_equal(ts.positions, idealcoords[i], decimal=ref.prec)
-
+    
     def test_transformations_slice(self, ref, transformed):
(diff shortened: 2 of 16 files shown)


package/CHANGELOG
```diff
@@ -28,7 +28,7 @@ Fixes
     (Issue #2666, PR #3565)
   * Fixed DumpReader triclinic box dimensions reading. (Issue #3386, PR #3403)
   * Updated the badge in README of MDAnalysisTest to point to Github CI Actions
-
+  * Fixed FrameIteratorIndices doesn't rewind. (Issue #3416)
```
hmacdope, line 36: Can you leave a blank line here?

testsuite/MDAnalysisTests/coordinates/test_xdr.py
```diff
@@ -212,6 +212,20 @@ def go_beyond_EOF():
...
+        frames = []
+        iterator = u.trajectory[[0,1,2]]
+        for ts in iterator:
+            frames.append(ts.frame)
+        assert_equal(np.array(frames),
+                     np.arange(u.trajectory.n_frames)[[0,1,2]],
+                     err_msg="indexing did not produce the expected frames")
+        assert_equal(u.trajectory.ts.frame, 0,
+                     err_msg="Index Iterator did not rewind")
+
```
hmacdope: Too many blank lines here + some whitespace, see PEP8 complaint by the bot.

testsuite/MDAnalysisTests/coordinates/test_pdb.py
```diff
@@ -636,6 +636,8 @@ def test_iteration(self, multiverse):
             "iterated number of frames %d is not the expected number %d; "
             "trajectory iterator fails to rewind" %
             (len(frames), u.trajectory.n_frames))
+        assert_equal(u.trajectory.ts.frame, 0,
```
hmacdope: I think this is a good idea both here and in `test_xdr` and I am happy to leave them in but as we are dealing with behaviour from the `ProtoReader` base class, I think we will also need a test of the base class behaviour. 

 I would love to see a test in `testsuite/MDAnalysisTests/coordinates/base.py` called something like `test_iteration_rewinds` where you compare where the trajectory ends up from the different types of iteration. 

 Other people feel free to weigh in also.  :)

hmacdope (changes requested):
Great start! See my comments.

testsuite/MDAnalysisTests/coordinates/test_pdb.py
megosato: @hmacdope I added a `test_iteration_rewinds` to `testsuite/MDAnalysisTests/coordinates/base.py`. However, I'm not entirely sure it's what you were intending. Any feedback is appreciated! :) Thank you for your help!

hmacdope: Yep looks spot on to me!

hmacdope:
@megosato would you be able to resolve conflicts? 

@MDAnalysis/gsoc-mentors  would I be able to grab a second pair of eyes for a review? Possibly someone familiar with original issue :+1:

testsuite/MDAnalysisTests/coordinates/base.py
```diff
@@ -384,7 +384,7 @@ def test_transformations_2iter(self, ref, transformed):
 
         for i, ts in enumerate(transformed):
             assert_almost_equal(ts.positions, idealcoords[i], decimal=ref.prec)
-
+
```
orbeckst: stray whitespace

testsuite/MDAnalysisTests/coordinates/base.py
```diff
@@ -512,6 +511,22 @@ def test_iter_as_aux_lowf(self, ref, reader):
...
+        assert_equal(ts_indices.frame, ts_sliced.frame,
+                     err_msg="FrameIteratorIndices and Sliced do not match")
+        assert_equal(ts_indices.frame, 0,
+                     err_msg="FrameIteratorIndices did not rewind")
+        assert_equal(ts_sliced.frame, 0,
+                     err_msg="FrameIteratorSliced did not rewind")
+        for ts_all in reader:
+            pass
+        assert_equal(ts_all.frame, 0,
+                     err_msg="FrameIteratorAll did not rewind")
```
orbeckst: These are three different tests. Split it into three different methods to keep things simple. You might be able to write one test and then [parametrize](https://docs.pytest.org/en/6.2.x/parametrize.html) with three accessor functions (and different [test ids](https://docs.pytest.org/en/6.2.x/example/parametrize.html#paramexamples) so that it looks nice in the output)
```python
@pytest.mark.parametrize("accessor", [
          lambda traj: traj[[0, 1, 2]], 
          lambda traj: traj[:3],
          lambda traj: traj],
          ids=["indexed", "sliced", "all"])
def test_iter_rewinds(self, reader, accessor):
   for ts in accessor(reader):
        ...
   ...
```

testsuite/MDAnalysisTests/coordinates/test_pdb.py
orbeckst: How is this test different from the `test_rewind()` in the preceding method? Is this test really needed, i.e., **did this test FAIL before you made your change?**

If the answer is "no" then remove the line here as it is not really testing your new code.

testsuite/MDAnalysisTests/coordinates/test_pdb.py
```diff
@@ -645,6 +647,19 @@ def test_slice_iteration(self, multiverse):
         assert_equal(np.array(frames),
                      np.arange(u.trajectory.n_frames)[4:-2:4],
                      err_msg="slicing did not produce the expected frames")
+        assert_equal(u.trajectory.ts.frame, 0,
+                     err_msg="slice iterator did not rewind")
```
orbeckst: Same as previous comment: Is this test actually testing your code? Did it fail *before* you added your fix? If not, remove the test.

testsuite/MDAnalysisTests/coordinates/test_pdb.py
```diff
@@ -645,6 +647,19 @@ def test_slice_iteration(self, multiverse):
...
+    def test_indices_iteration(self, multiverse):
+        u = multiverse
+        frames = []
+        for ts in u.trajectory[[2, 3, 4]]:
+            frames.append(ts.frame)
+        assert_equal(np.array(frames),
+                     np.arange(u.trajectory.n_frames)[[2, 3, 4]],
+                     err_msg="indexing did not produce the expected frames")
+        assert_equal(u.trajectory.ts.frame, 0,
+                     err_msg="indices iterator did not rewind")
```
orbeckst: Did this test fail before you made your changes?

I assume the answer is "yes" but I'd like you to try it and report the result of the *failing test* in this issue comment.

Can you please also discuss why you think that this test is necessary (i.e., for the PDBReader)?

testsuite/MDAnalysisTests/coordinates/test_xdr.py
```diff
@@ -212,6 +213,18 @@ def go_beyond_EOF():
         with pytest.raises(StopIteration):
             go_beyond_EOF()
 
+    def test_index_iterator(self):
```
orbeckst: Same comment as for the PDBReader test:
- does it fail without your fix?
- why is this test necessary?

orbeckst (changes requested):
Thanks for the contribution. The fix itself looks good. I have some comments on the tests so please see there.

Please also add yourself to AUTHORS.

Finally, I suggest that you also work on your proposal in parallel. I think that this PR will get merged pretty soon, but it always go slower than we think, so don't wait for the PR.

testsuite/MDAnalysisTests/coordinates/test_pdb.py
megosato: Hi @orbeckst. Thank you for all the feedback! 

Before discussing with Hugo (above) I wasn't aware I could use the `base.py` file to test the base class functionality. So thats why I put the tests in the PDB reader and XDR reader tests

I'm happy to post the failing test result for this comment after I get off work. As well as remove the tests I added to `test_xdr` and `test_pdb` before I had a better understanding of the testing infrastructure.

testsuite/MDAnalysisTests/coordinates/base.py
megosato: Hi @orbeckst,

@hmacdope previously suggested "I would love to see a test in testsuite/MDAnalysisTests/coordinates/base.py called something like test_iteration_rewinds where you compare where the trajectory ends up from the different types of iteration."

which is what I was trying to do with the assert referenced below. I was wondering if you think I explicitly should compare the that they end up at the same frame or if its better to just test that all the iterators end up rewinding to 0?
```python
assert_equal(ts_indices.frame, ts_sliced.frame,
                     err_msg="FrameIteratorIndices and Sliced do not match")
```

testsuite/MDAnalysisTests/coordinates/test_pdb.py
megosato: @orbeckst Here is the failed test case: 
```python
___________________________________ TestMultiPDBReader.test_indices_iteration ___________________________________

self = <MDAnalysisTests.coordinates.test_pdb.TestMultiPDBReader object at 0x12eb53f10>, multiverse = <Universe with 392 atoms>

    def test_indices_iteration(self, multiverse):
        u = multiverse
        frames = []
        for ts in u.trajectory[[2, 3, 4]]:
            frames.append(ts.frame)
        assert_equal(np.array(frames),
                     np.arange(u.trajectory.n_frames)[[2, 3, 4]],
                     err_msg="indexing did not produce the expected frames")
>       assert_equal(u.trajectory.ts.frame, 0,
                     err_msg="indices iterator did not rewind")
E       AssertionError: 
E       Items are not equal: indices iterator did not rewind
E        ACTUAL: 4
E        DESIRED: 0

test_pdb.py:657: AssertionError
```

testsuite/MDAnalysisTests/coordinates/base.py
orbeckst: I check the most specific thing, namely that they rewind to zero.

testsuite/MDAnalysisTests/coordinates/test_pdb.py
orbeckst: That looks like a good failure.

Regarding the PDB and XDR: I am inclined to drop them *or* integrate them into some tests that check all readers (not sure if we have those).

orbeckst (approved):

orbeckst:
Looks very good to me — I'll leave everything else to you @hmacdope !

megosato:
Hi @hmacdope I was wondering if you have any insight on the GH Actions CI / pypi_check (pull_request) check? I'm not entirely sure where to start for this failure.

hmacdope:
Very strange, I'll investigate.

hmacdope:
Perhaps related to #3612, I might cycle PR open and closed to restart.

IAlibay:
Just pushed a merge against develop, _should_ fix the failing doc test.

hmacdope:
Sorry for delay @megosato should get to this tonight.

testsuite/MDAnalysisTests/coordinates/base.py
```diff
@@ -511,6 +510,15 @@ def test_iter_as_aux_lowf(self, ref, reader):
             assert_timestep_almost_equal(ts,
                                          ref.iter_ts(ref.aux_lowf_frames_with_steps[i]),
                                          decimal=ref.prec)
+    @pytest.mark.parametrize("accessor", [
+              lambda traj: traj[[0, 1, 2]],
```
hmacdope:
```suggestion
              lambda traj: traj[[0, 1, 2]],
```

hmacdope (changes requested):
Just this and the blank line PEP8 change and then all good to roll :)

hmacdope (approved):
Congratulations on your first contribution! Excellent work  :100: :1st_place_medal: :100:

PR #3611, opened 2022-04-07
Higher Performance nuclinfo

### LINKED ISSUE #3310: nuclinfo.wc_pair extremely slow
I'm using analysis.nuclinfo.wc_pair to profile base pair distances over long trajectories + long DNA sequences. It seems to be rather fast over ~1-100 executions, but when evaluated over a loop e.g., 
for t in all time_steps:
for i in all bases:
for j in all bases: 
distance[t,i,j]=nuclinfo.wc_pair(etc.)

When the total number of wc_pair evaluations reaches ~1e3 (rough guess), evaluation gets extremely slow, or indeed hangs and will not finish at all on my platform. (MDAnalysis 1.0.0 on python 3.8.5 on Windows). 

Any guidance at all on why this might be the case would be very much appreciated. It evaluates rather quickly for few-iterations, but seems to hang when queried repeatedly over thousands of base pairs and time steps. This is an extremely useful function for my work in-principle, but not if it keeps hanging for long trajectories and/or large DNA sequences.

### LINKED ISSUE #3600: nuclinfo module does not use AnalysisBase
## Expected behavior ##
Analysis modules follow standard behavior, i.e., use [AnalysisBase](https://docs.mdanalysis.org/stable/documentation_pages/analysis/base.html) when possible.

## Actual behavior ##
[nuclinfo](https://docs.mdanalysis.org/stable/documentation_pages/analysis/nuclinfo.html) consists of legacy code that predates AnalysisBase.

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) 2.2-dev

ALescoulie:
Fixes #3600, #3310

 - Add `AnalysisBase` derived classes for calculating different nucleic acid base pair distances.

I built a new class `NucPairDist` that calculates distances between atoms in nucleic acid. It has 3 subclasses `WCDist`, `MajorDist` and `MinorDist`. Those are indented to  improve the performance of their equivalents in nucinfo. They also support taking groups of base pairs and running the distance over a trajectory.

TODO:
- [x] Docs

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


ALescoulie:
Just realized that pandas is not included in MDAnalysis, I had been using tidy DataFrames to record my results

IAlibay:
I'll let the other @MDAnalysis/coredevs speak here too, but I'm not particularly keen on having pandas as a core dependency (in fact my current goal is to remove more dependencies from MDAnalysis). I would prefer if numpy arrays could be used to store results where possible.

ALescoulie:
@IAlibay Don't think switching to arrays will be too much trouble, I'm trying to store selection, distance and time results in an organized way so I deflated to a DataFrame.

orbeckst:
I second @IAlibay 's https://github.com/MDAnalysis/mdanalysis/pull/3611#issuecomment-1092257941 — keep it as a numpy array. I don't see pandas here to be essential. 

If you _really_ need to make data available then you could create `to_df()` that imports pandas only in the function and that can fail and tell you to install it. We have packages that are optional and then we follow that approach.

orbeckst:
Btw, pandas.DataFrame does not have great performance in my experience — you get convenience but not the speed of numpy arrays.

ALescoulie:
@orbeckst I don't use it during iteration, I just build it from a dictionary in the `_conclude` step, but I could just return a names list and a numpy array of distances and times. Overall is this like what you were wanting for #3600 .

orbeckst:
I understand; my point was that when you want to process your results further (e.g., calculate statistics) then this will be faster when the data are already an array than if it were a DataFrame. For people comfortable with pandas, the df is probably more convenient and organized. MDAnalysis has been embracing numpy arrays as the foundation for everything so it makes sense to use them for results, too. But I don't see anything wrong with adding a convenience function that presents the raw data as a DataFrame (although other devs may have different opinions on the matter).

ALescoulie:
@orbeckst and @IAlibay  looking at how my DataFrame (selection, time, dist as cols with results in rows) is structured, how do you think is best to structure the results into an array. I'm thinking a dictionary with selection number as keys.

orbeckst:
I'd say look at some of the other standard analysis, e.g., RMSD.

Also @PicoCentauri and @joaomcteixeira might also have an opinion on how to store data in `results`.

PicoCentauri:
I am also in favor of keeping things as `numpy` arrays. I don't see much benefit of a pandas data frame here. Additionally, the new classes will be part of the [CLI](https://github.com/MDAnalysis/mdacli), and the CLI's saving workflow is (currently) not able to handle data frames. The latter is maybe an issue of the CLI I have to fix though.

@ALescoulie I would also be very happy if you could test your modules from the command (once you have a proper docstring)!

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -0,0 +1,187 @@
...
+:Year: 2022
+:copyright: GNU Public Licence v3
+
+"""
+
+from typing import List, Tuple, Dict, Union, NamedTuple
+
+import numpy as np
+import pandas as pd
+from math import pi, sin, cos, atan2, sqrt, pow
```
orbeckst: Whenever I see `import math` I fear that there's code that could be made faster by using numpy array operations...

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -0,0 +1,187 @@
...
+
+import MDAnalysis as mda
+from MDAnalysis.lib import mdamath
+from MDAnalysis.analysis.base import AnalysisBase
+from package import MDAnalysis
+from package.MDAnalysis.exceptions import SelectionError
```
orbeckst: There's no package module. This shouldn't even work.... except if you're running from the top of the checked out git repo.

If you want to do development, install with
```
pip install -e package/
```

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -0,0 +1,187 @@
...
+    This makes passing in selections more organized.
+
+    :Arguments:
+    *segid*
+        The name of the segment the base is on
+
+    *base*
+        The number of the base pair in the segment
+
+    .. rubric:: Example
```
orbeckst: We use numpydoc formatting for doc strings. See https://userguide.mdanalysis.org/stable/contributing_code.html#working-with-the-code-documentation

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -0,0 +1,187 @@
...
+class NucPairDist(AnalysisBase):
+    """An class for developing analyses that run on a pair of base pairs """
+
+    _sel: List[Tuple[str, str]]
+    _unv: mda.Universe
+    _s1: mda.AtomGroup
+    _s2: mda.AtomGroup
+    _res_dict: Dict[str, List[Union[float, str]]]
+    _names: List[str]
+    results: pd.DataFrame
```
orbeckst: results has to be a [Results](https://docs.mdanalysis.org/stable/documentation_pages/analysis/base.html#MDAnalysis.analysis.base.Results) container to be compatible with AnalysisBase/mdacli

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -0,0 +1,187 @@
...
+:Author: Alia Lescoulie
+:Year: 2022
+:copyright: GNU Public Licence v3
+
+"""
+
+from typing import List, Tuple, Dict, Union, NamedTuple
+
+import numpy as np
+import pandas as pd
```
orbeckst: avoid pandas as a dependency

orbeckst (dismissed):
A few quick comments — I see it's in draft but this might already be helpful
- don't use pandas.DataFrame the container for the data, use our base.Results dict-like thing
- use numpydoc formatting for documentation
- eventually: update CHANGELOG and add yourself to AUTHORS

In general your idea looks good!

ojeda-e (commented):
One quick question @ALescoulie (And while this is a draft PR): is there any particular reason to use `assert_almost_equal` for your tests? I think MDAnalysis is using `assert_allclose` for float comparisons in tests. The [numpy docs](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_almost_equal.html) also suggests to use `assert_allclose` instead.

testsuite/MDAnalysisTests/analysis/test_nucleicacids.py
```diff
@@ -0,0 +1,66 @@
...
+# J. Comput. Chem. 32 (2011), 2319--2327, doi:10.1002/jcc.21787
+#
+
+import MDAnalysis as mda
+import pytest
+from MDAnalysis.analysis.nucleicacids import WCDist, MinorDist, MajorDist, BaseSelect
+from MDAnalysisTests.datafiles import RNA_PSF, RNA_PDB
+from numpy.testing import (
+    assert_almost_equal,
+    assert_allclose,
```
ojeda-e: and you are importing it here already

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: I'll use the numpy implementations, that was just copied from nucinfo.py I haven't implemented the other half of it using those operations yet

testsuite/MDAnalysisTests/analysis/test_nucleicacids.py
ALescoulie: I switched to `assert_all_close`, I was imitating an older set of tests in `test_nucinfo.py`

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: I got rid of it in favor of a Results object

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: I'm switching over to that

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: Ok, I'll start on the docs soon now that I fixed the some of the functionality of the code.

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: I got the imports fixed not sure why it was like that

ALescoulie:
I updated my code based of the feedback, I apologize for the delay, my systems programing class is keeping me really busy. I still need to write documentation for new module, but it retains the core functionality of `nucinfo.py` making measuring multiple groups and distance over time more efficient. Now I need to finish the docs.

ALescoulie:
@orbeckst do you want me to also update the phase and torsion functions from `nucinfo.py`

ALescoulie:
Looking at coverage, I need to write some tests for the selection checking, It should be simple to just try passing in a protein structure and make sure it rejects the residues for not being amino acids.

orbeckst:
> do you want me to also update the phase and torsion functions from nucinfo.py

It's up to you. I am ok with a partial transition of nuclinfo as a start. If you don't transition all functionality, please raise individual issues for missing parts. Having small issues is good because they make for good starter issues.

orbeckst:
Also have a look at PEP8 — as you're starting a new module, let it start out in a PEP8 compliant state.

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -0,0 +1,185 @@
...
+
+"""
+
+from typing import List, Tuple, Dict, Union, NamedTuple
+
+import numpy as np
+
+import MDAnalysis as mda
+from ..lib import mdamath
+from ..analysis.base import AnalysisBase, Results
```
orbeckst: Should work as 
```suggestion
from .base import AnalysisBase, Results
```

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -0,0 +1,185 @@
...
+
+        for s in self._sel:
+            # Checking number of selections and building mda.AtomGroups
+            if len(s) != 2:
+                raise ValueError("Each base pair selection given as a tuple of length 2")
+            try:
+                ag1.append(self._unv.select_atoms(s[0]))
+
+            except SelectionError:
+                raise SelectionError(f"Invalid selection in {s[0]}")
```
orbeckst: You might not need to catch these SelectionErrors (and thus not need to test them!) — isn't it sufficiently clear to let them happen?

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -0,0 +1,185 @@
...
+            self._res_dict['time'].append(self._ts.time)
+
+    def _conclude(self) -> None:
+        self.results['times'] = np.array(self._res_dict['time'])
+        self.results['selections'] = self._names
+        for i, n in enumerate(self._names):
+            self.results[i] = np.array(self._res_dict[n])
+
+
+class WCDist(NucPairDist):
```
orbeckst: I'd go for more expressive names such as *WatsonCrickDistance*.

testsuite/MDAnalysisTests/analysis/test_nucleicacids.py
```diff
@@ -0,0 +1,63 @@
...
+
+@pytest.fixture(scope='module')
+def u():
+    return mda.Universe(RNA_PSF, RNA_PDB)
+
+
+def test_wc_dist(u):
+    sel = [(BaseSelect('RNAA', 1), BaseSelect('RNAA', 2)),
+           (BaseSelect('RNAA', 22), BaseSelect('RNAA', 23))]
+    WC = WCDist(u, sel)
```
orbeckst: Is the user supposed to build selections in this way? That looks quite cumbersome. 

I think I'd prefer something where the user has to provide AtomGroups, e.g. for strand_1 and and strand_2 and the residues are matched up with zip(strand_1.residues, strand_2.residues). We can then provide a helper function to build these groups but the classes themselves don't have to much guessing baggage, except identifying the correct atoms for the distance.

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: Ok updating that. That makes sense to update.

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: Ok, I'll just remove that.

testsuite/MDAnalysisTests/analysis/test_nucleicacids.py
ALescoulie: I'm trying to get an idea of how you want this to work, so the user provides the strands, then the residue pairs are formed in a list of tuples. The classes would then take the residue selections, use the zipped list to find the specific residues than grab the specific atoms. The super class would just accept a list of AtomGroups then merge them into a single AtomGroup. I think to do this I'd write a helper function to select the strands of DNA from a Universe. Is this what you are envisioning?

The question I have it are how do you select a specific strand of strand of DNA?

ALescoulie: I mostly used the named tuple for the sake of organizing inputs.

hmacdope:
@ALescoulie I love the typed code! It looks amazing. 👍 I was going to add that you need a bunch of tests for the errors you are raising. Have a look at the lines codecov is complaining about. :)

testsuite/MDAnalysisTests/analysis/test_nucleicacids.py
orbeckst: >  so the user provides the strands, 

Yes, right now make it the user's responsibility to identify the WC pairs. `strand_1.residues[i]` will be considered base pairing with `strand_2.residues[i]`.

This is similar to the existing function, which asks the user to identify a single base pair via resid and segid. However, for your module we should move to an object-oriented approach and just work with AtomGroups instead of resid/segid selections.

> then the residue pairs are formed in a list of tuples. 

Yes.

> The classes would then take the residue selections, 

They take `strand_1.residues` and `strand_2.residues`

> use the zipped list to find the specific residues 

Yes.

> than grab the specific atoms. 

This is where the class does some actual work in order to identify the specific atoms that are required for the W-C base pair. Depending on the residue type, a different atom is chosen.

> The super class would just accept a list of AtomGroups then merge them into a single AtomGroup. 

I don't think that it would merge them into a single group. Ultimately you want to be able to use `calc_bonds()` on two arrays of positions. Each strand should have N residues and in the end, for each strand you should end up with an AtomGroup with N atoms, where the atom is either an N3 or and N1, depending on the nucleobase.

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -0,0 +1,185 @@
...
+        self._s2 = mda.AtomGroup([ag[0] for ag in ag2])
+
+    def _prepare(self) -> None:
+        self._names = [s[0] + s[1] for s in self._sel]
+        self._res_dict = {k: [] for k in self._names}
+        self._res_dict['time'] = []
+
+    def _single_frame(self) -> None:
+        dist = self._s1.positions - self._s2.positions
+        wc: np.ndarray = mdamath.pnorm(dist)
```
orbeckst: use [calc_bonds()](https://docs.mdanalysis.org/stable/documentation_pages/lib/distances.html?highlight=calc_bonds#MDAnalysis.lib.distances.calc_bonds) because it will properly take PBC into account

orbeckst:
You can reduce the scope of this PR and focus on the base class and just the WatsonCrick pair distance — the other functionality can then be added separately later. Just raise issues for the missing parts once the initial PR has been merged.

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def __init__(self, universe: mda.Universe, selections: List[Tuple[BaseSelect, BaseSelect]],
+                 n1_name: str = 'N1', n3_name: str = "N3", **kwargs) -> None:
+        sel_str: List[Tuple[str, str]] = []
+
+        for s in selections:
+            if universe.select_atoms(f" resid {s[0].resid} ").resnames[0] in ["DC", "DT", "U", "C", "T", "CYT", "THY", "URA"]:
+                a1, a2 = n3_name, n1_name
+            elif universe.select_atoms(f" resid {s[0].resid} ").resnames[0] in ["DG", "DA", "A", "G", "ADE", "GUA"]:
+                a1, a2 = n1_name, n3_name
```
richardjgowers: Worth bringing these hardcoded names out into a kwarg so someone with different names can still use the class?

richardjgowers (changes requested):
I think @orbeckst has already said this, but I'm much more in favour of writing analysis where the atom selecting happens outside of the analysis class/function... so


```python
ag1 = u.select_atoms('foo')
ag2 = u.select_atoms('bar')

analysis(ag1, ag2)
```
rather than

```python
analysis(u, 'foo', 'bar')
```

The former is much easier to debug if you're suspicious about what's happening inside the analysis class, and also caters for people who might have differently named systems.

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: Ok I think a good way to do this would be a kwarg for each base. I can pretty easily set that up.

ALescoulie:
> I think @orbeckst has already said this, but I'm much more in favour of writing analysis where the atom selecting happens outside of the analysis class/function... so

I agree, the one thing is needing to select different things depending on the base pair, for example selecting N1 or N3 depending on the base pair, so I think I'll do something where they select the base pairs and my just selects the specific atoms.

testsuite/MDAnalysisTests/analysis/test_nucleicacids.py
ALescoulie: I was thinking it would merge them into two atom groups which then are put into calc_bonds

richardjgowers (approved):
This is good to go as it stands.  I think as a modernisation I'd like to convert this from passing in selection strings to passing in AtomGroups, but this can be a separate step so we can enjoy the benefits of this code as-is.

ALescoulie:
@richardjgowers I'd like to make a few quick changes as well as add some docs then we can merge.

ALescoulie:
@richardjgowers I implemented the suggestions from @orbeckst, removed selections from inside the Classes aside from selecting atoms and added keyword arguments for each base pair. I still need to add documentation but am working on that now.

ALescoulie:
ok I wrote some docs. I made sure to keep the original citation from nuclinfo in since I essentially copied her system for solving the problem, just restructured it to a faster OOP approach.

ALescoulie:
I got the docs working and moved the selection outside the classes

richardjgowers (approved):

Issue #2542, opened 2020-02-19
rewrite AnalysisBase to be compatible with pmda.parallel.AnalysisBase


orbeckst:
**Expected behavior**

MDAnalysis and pmda share code and it is easy to parallelize existing analysis; we don't maintain duplicate code. For instance, I should be able to take a `_single_frame()` method for MDAnalysis directly in pmda.

**Actual behavior**

MDAnalysis.analysis.base.AnalysisBase and pmda.parallel.AnalysisBase are different because the serial version can easily share data via the class while the parallel version must have `_single_frame()` return the results for accumulation. The parallel version has `_reduce` to handle accumulation in a parallelizable fashion.

**Comments**

pmda is only one way to parallelize trajectory analysis. However, my understanding is that as soon as one wants to distribute work over processes that cannot communicate by shared memory one needs to change the way in which AnalysisBase works. Therefore, I think that adapting the pmda model is a step in the right direction.

IAlibay:
Is this still feasible for 2.0?

How is this affected by the proposed changes for mda_cli? Pinging @PicoCentauri here.

orbeckst:
Not feasible for 2.0

IAlibay:
@orbeckst shall we move this to 2.1 or do we think it's now become a 3.0 target?

orbeckst:
Do 3.0 for now and if we can get hands on deck we can move it up.

IAlibay:
Looks to me like we're going a different direction here -- closing.

RMeli:
Linking https://github.com/MDAnalysis/mdanalysis/issues/4158 for record.

PR #3617, opened 2022-04-10
Changed LinearDensity for Issue #2508

### LINKED ISSUE #2508: Have LinearDensity support updating AtomGroups, more intuitive names, save bins

BFedder:
Fixes #2508 (partially)

* LinearDensity now works with updating AtomGroups. A new test was
added to verify this.
* Result variables "pos" and "char" as well as their "std" were
renamed to "mass_density", "charge_density" and "stddev", respectively.
* The bin edges of np.histogram are now saved for xyz as "hist_bin_edges"

This PR addresses points 1, 2, and 3 raised by @lilyminium in #2508. It seems to me that discussion around point 4, which is also referenced in #2507, has not concluded yet. How should the units be handled?

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?


IAlibay (commented):
Just a quick pre-review comment - changing attribute names, particularly for results without warning is problematic as folks downstream might rely on scripts that work with specific attribute names that they expect.

We can only fully remove the old attributes in a major version change (the next one being 3.0.0). Since it's too early for a major version change, here we will need to have a way to deprecate the old results attributes, add the new ones, and warn users of the deprecation (either through warnings on access or at the very least through documentation).

IAlibay (changes requested):
Making my previous comment a blocking review since it will need addressing.

PicoCentauri:
Now we are working intensively on this class, I have three questions/things to discuss.

1. We should give the units of the density in the docstring. 

2. Why does `LinearDensity` calculate the mass AND the charge density at the same time? This will lead to errors for trajectories without charges making calculating only the mass density impossible. I would rather suggest adding an option like `dens` to calculate a specific type of density only. A suitable weighting attribute of an function like

```python
    if self.dens == "mass":
        self.weights = ag.masses
    elif dens == "number":
        self.weights = np.ones(ag.n_atoms)
    elif dens == "charge":
        self.weights = ag.charges
    else:
        raise ValueError(f"`{dens}` not supported. "
                         "Use `mass`, `number` or `charge`")
```

can be called in the every frame. The code should of course work for different grouping schemes.

3. Why does `LinearDensity` calculate all directions (x,y,z) at the same time? Most of the systems are either homogenous leading to the same density in all directions or have a planar symmetry in one direction making two directions meaningless. Additionally, there is the `DensityAnalysis` class for general volumetric analysis. Leaving out the calculations in two directions will also speed up calculations!

IAlibay:
@PicoCentauri given that this is still a follow-up GSoC candidate, I might ask that points 2 and 3 be raised as separate issues to be tackled in separate PRs if that's reasonable? (there's not much time before the GSoC deadline)

PicoCentauri:
@IAlibay. Of course I will open one after this is merged.

BFedder:
I have worked on a way to deprecate here and made progress, but ran into a problem. Guided by how other deprecations around the `Results` object are currently handled, I was hoping to make use of properties. That was problematic, though, because of how nested the dictionary is. Defining a getter for `results.x.pos` was difficult because of the `.`.  In the new commit, I got this to work by defining `class Results(Results)` and specifying properties there. Now, when accessing the data using attribute syntax (i.e. `ld.results.x.pos`) gives a DeprecationWarning and returns the contents of the new variable (i.e. `ld.results.x.mass_density`). I cannot get dictionary syntax to work, though... so `ld.results["x"]["pos"] fails. Could I ask for advice?

PicoCentauri:
Good effort. For me a general deprecation warning for the whole module would also be enough. What about @IAlibay?

IAlibay:
> Good effort. For me a general deprecation warning for the whole module would also be enough. What about @IAlibay?

What'd you mean here? A warning every time you use the module or just a warning in the docs?

PicoCentauri:
Probably both. 

I played around with the results class and I also haven't found a way to deprecate a dictionary entry.

BFedder:
Hi @PicoCentauri, sorry for the delay in addressing your previous response and thanks for your input. 

> We should give the units of the density in the docstring.

Yes, definitely, happy to do that. I'm also happy to change the units in this PR, that would then fully address #2507 and #2508. But I think the discussion around what to change the units to had not concluded when the issues were first raised. If we do not want to decide on the new units for now, I could just add the current ones to the docstring? 

>Why does LinearDensity calculate the mass AND the charge density at the same time? 
>Why does LinearDensity calculate all directions (x,y,z) at the same time?

Both very good points! Changing that would definitely make sense, but I'll second @IAlibay here and say that a separate issue might be good for these. Thanks for agreeing to raise it after this PR here is merged

BFedder:
So regarding deprecations, I could either:

- leave things as they currently are: warn users when accessing the result attributes, but accessing the dictionary is broken 
or 
- remove the deprecation warning on accessing the attribute (which fixes the dictionary accession) and only warn on loading the module and in the docstring

I'm guessing the latter is preferable because the dictionary isn't broken this way, right?

IAlibay:
> Probably both. 
> 
> I played around with the results class and I also haven't found a way to deprecate a dictionary entry.

Warnings every time we use a class might be a bit much. I'll ping @jbarnoud and @lilyminium who have particular views on warnings here.

Maybe we can limit it to the access of the results dictionary itself (if that's feasible)?

jbarnoud:
I don't think a deprecation should be able to break existing code, so breaking the access to the dict is not really an option if it was the intended way of doing things.

Did you consider creating a `DeprecationDict` that would inherit from `dict`?

BFedder:
Thanks @jbarnoud, that suggestion led me to find a solution here. Now, accessing the old result names both via attributes and via dictionaries warns about the deprecation. For version 3.0.0, all that needs to be done here is to delete the definition of Results(Results) in lineardensity.py and the corresponding test in test_lineardensity.py.

jbarnoud:
Youhou! I am glad the idea worked. I would suggest a couple of reactors:
* you could use a template for the deprecation message and avoid repeating it over and over
* in your `__getitem__` method, you could use a dictionary to link the deprecated name to the new one; combined with the template message, it can make for some more compact and more readable code

These are just suggestions, though.

PicoCentauri:
Nice! Following @jbarnoud and since this is not probably not the last time we have to deprecate attributes of `results` we could introduce a general function like 

```python
class ResultsDeprecated(Results):
    """Results class used for deprecating attributes.
    
    Parameters
    ----------
    attr_dict : dict
        A dictionary mapping the old attributes names to new ones.
    remove_version : str
        version when the attributes will be removed.
    """

    def __init__(self, attr_dict, remove_version, *args, **kwargs):
        super(ResultsDeprecated, self).__init__(*args, **kwargs)
        self._attr_dict = attr_dict
        self._remove_version = remove_version

    def _deprication_warning(self, key):
        warnings.warn(
            f"`{key}` is deprecated and will be removed in version "
            f"{self._remove_version}. Please use "
            f"`{self._attr_dict[key]}` instead.", DeprecationWarning)

    def __getitem__(self, key):
        if key != "_attr_dict" and key in self._attr_dict.keys():
            self._deprication_warning(key)
            key = self._attr_dict[key])

        return super(ResultsDeprecated, self).__getitem__(key)

    def __getattr__(self, attr):
        if attr != "_attr_dict" and attr in self._attr_dict.keys():
            self._deprication_warning(attr)
            attr = self._attr_dict[attr])

        return super(ResultsDeprecated, self).__getattr__(attr)
```

and put into `package/MDAnalysis/lib/util.py`? For the current case it could look like:

```python
attr_dict = {"pos": "mass_density", "pos_std": "mass_density_stddev",
                "char": "charge_density", "char_std": "charge_density_stddev"}


results = ResultsDeprecated(attr_dict, "3.0.0")
```

PicoCentauri:
> Yes, definitely, happy to do that. I'm also happy to change the units in this PR, that would then fully address https://github.com/MDAnalysis/mdanalysis/issues/2507 and https://github.com/MDAnalysis/mdanalysis/issues/2508. But I think the discussion around what to change the units to had not concluded when the issues were first raised. If we do not want to decide on the new units for now, I could just add the current ones to the docstring?

I suggest leave the units as they are for now but comment in the docstring. When we (maybe) start a deep restructuring  of the `LinearDensity` class (different types of densities, only one direction) we can change the units to the standard units of MDAnalysis.

BFedder:
Thanks for these suggestions! It is now indeed much nicer.
I have now also added information on the density units as well as a note that they will likely be changed in the future.

I for one think having a general `ResultsDeprecated(Results)` class would be great! I can open an issue for this if you'd like.

package/MDAnalysis/analysis/lineardensity.py
```diff
@@ -57,16 +88,20 @@ class LinearDensity(AnalysisBase):
...
-    results.x.pos_std : numpy.ndarray
+    results.x.mass_density : numpy.ndarray
+           mass density in g/cm^3 in [xyz] direction
+    results.x.mass_density_stddev : numpy.ndarray
            standard deviation of the mass density in [xyz] direction
-    results.x.char : numpy.ndarray
-           charge density in [xyz] direction
-    results.x.char_std : numpy.ndarray
+    results.x.charge_density : numpy.ndarray
+           charge density in (e * mol)/cm^3 in [xyz] direction
```
PicoCentauri: You could use math syntax here

```suggestion
           charge density in :math:`\mathrm{e \cdot mol \cdot cm^{-3}` in [xyz] direction
```

package/MDAnalysis/analysis/lineardensity.py
```diff
@@ -57,16 +88,20 @@ class LinearDensity(AnalysisBase):
     ----------
     results.x.dim : int
            index of the [xyz] axes
-    results.x.pos : numpy.ndarray
-           mass density in [xyz] direction
-    results.x.pos_std : numpy.ndarray
+    results.x.mass_density : numpy.ndarray
+           mass density in g/cm^3 in [xyz] direction
```
PicoCentauri: See my comment below

PicoCentauri (changes requested):
Yes much better now!

You can also add a general class within PR. It is not more code compared to what we have now. Just some additional tests that everything works as expected including a test that the warning is raised. Tests for your new Results class have to be written in any case (within this PR).  But I am also fine with moving the generalization to a another PR.

BFedder:
I have now tried implementing the general class in mda.lib.utils, but that creates a circular import problem (ImportError: cannot import name 'triclinic_box' from partially initialized module 'MDAnalysis.lib.mdamath' (most likely due to a circular import)) I believe because mda.analysis.base.Results imports mda.coordinates.core, which imports from mda.lib.mdamath just like util does. Fixing this import hierarchy and making sure that nothing breaks is a bit beyond what I have time for given how close the application deadline is.. Sorry! 

I have now changed the docstring to include `:math:` though, and my tests now not only check that the correct values are returned, but also that warnings are raised. 
Weirdly enough, pytest fails when the '\\' are present in the docstring (SyntaxError: invalid escape sequence) even though it's obviously in the docstring, and the code runs with no problem. Should I turn the docstring into a raw string to avoid this issue?

PicoCentauri:
> I have now tried implementing the general class in mda.lib.utils, but that creates a circular import problem (ImportError: cannot import name 'triclinic_box' from partially initialized module 'MDAnalysis.lib.mdamath' (most likely due to a circular import)) I believe because mda.analysis.base.Results imports mda.coordinates.core, which imports from mda.lib.mdamath just like util does. Fixing this import hierarchy and making sure that nothing breaks is a bit beyond what I have time for given how close the application deadline is.. Sorry!

No worries. We can look at this in an later PR.

> I have now changed the docstring to include :math: though, and my tests now not only check that the correct values are returned, but also that warnings are raised.
> Weirdly enough, pytest fails when the '\' are present in the docstring (SyntaxError: invalid escape sequence) even though it's obviously in the docstring, and the code runs with no problem. Should I turn the docstring into a raw string to avoid this issue?

A raw string string should work.

PicoCentauri:
Looks almost done. I only wonder about the units of the charge density are they really `e * mol * cm^{-3}`  and not `e * cm^{-3}`? At least the latter one is what I suggest from reading the code.

BFedder:
I looked at it again and this is how I understand it.
```python
def _conclude(self):
    k = 6.022e-1  # divide by avodagro and convert from A3 to cm3
    ...
    for dim in ['x', 'y', 'z']:
        norm = k * self.results[dim]['slice_volume']
        for key in self.keys:
            self.results[dim][key] /= norm
```

`k` is Avogadro times 10^{-24} cm^3/A^3
norm then converts the slice volume to cubic centimeters, so is in units of Avogadro * cm^3
results.x.charge_density holds charge values and is then then divided by norm to give the charge density in e/cm^3, but because norm also includes avogadro, that's where I think the * mol comes in. However, the more I look at this, the more confused I get... What I'm thinking now is that it would make the most sense to do 
```python 
    for dim in ['x', 'y', 'z']:
        norm = k * self.results[dim]['slice_volume']
        for key in ["mass_density", "mass_density_stddev"]:
            self.results[dim][key] /= norm
        for key in ["charge_density", "charge_density_stddev"]:
            self.results[dim][key] /= self.results[dim]['slice_volume']
````
That should return g/cm^3 for mass density and e/A^3 for charge density, right?

**Comment by PicoCentauri** [comment]
Yes your approach sounds good. Are the numbers for the mass and the charge density the same as before?

also I checked `k = constants.Avogadro * constants.centi**3/constants.angstrom**3` where constants is from scipy.

**Comment by BFedder** [comment]
The mass density numbers are unaffected, but the charge densities change by a factor of `k = 0.6022`. For example, what was [0., 0., 0., 2.2158751e-05, -2.2158751e-05, 0., 0., 0., 0., 0.] for `results.x.charge_density` would now be [ 0.,  0.,  0.,  1.33439999e-05, -1.33439999e-05,  0.,  0.,  0., 0.,  0.]

**Comment by PicoCentauri** [comment]
I see. This would be a breaking change and we decided to keep the units as they are. Then...lets keep with the old units which are e * mol * cm^{-3} right? The more we do on this class the more I have the feeling that we should write a new function from scratch...

**Comment by BFedder** [comment]
Good point. Yes, I do believe the charge densities are currently in units of e * mol * cm^{-3}.
I have added a few more comments to `LinearDensity._conclude()` to hopefully help the next person trying to work with this module and change the units

**Review by PicoCentauri — COMMENTED** [review]
Looks good for me. However codecov wants some additional tests. Could you address them.

BTW: I took my time in the train and wrote a [new LinearDensity](https://github.com/PicoCentauri/mdanalysis/commit/cde5659eee086ae5316cfa0c7414308f0aec44a2). If you want you can take a look. I will create a PR ones I have some tests and more docu...

**Comment by BFedder** [comment]
The lines codecov is complaining about are part of the definition of   `_add_other_results(self, other):`, which isn't currently used anywhere as far as I can tell. It's picked up here because I changed the results names as above. I believe this is safe to ignore for now because it doesn't seem to be implemented/used anyway, right? 

I like your new LinearDensity approach. It would indeed be cool to remove the limitation of a invariant orthorhombic simulation box. One addition to what you have now that I think could be useful is if the user could specify more than one dimension and more than one density type to be calculated per instance of LinearDensity. Although having multiple objects, say one for mass and one for charge density, would also work of course.

**Comment by PicoCentauri** [comment]
It is used in the parallel version of an analysis provided by [pmda](https://github.com/MDAnalysis/pmda) which is not part of the main library but should be tested.

The restriction to one dimension is a decision to keep the code simple. I would suggest that people who want to analysis more than dimension should use several instances. But, we can discuss this after this PR 😌

**Comment by BFedder** [comment]
I have rewritten `_add_other_results(self, other)` a bit to avoid code repetition there, and also wrote a test case to make sure the results of a split analysis are the same as those of an analysis over the entire Universe. I couldn't find any mention of `LinearDensity` in `pmda` though, so I'm not sure if I have implemented the tests correctly here.

File: package/CHANGELOG [hunk 3]
```diff
@@ -61,6 +62,11 @@ Enhancements
     keyword (Issue #3605)
 
 Changes
+  * LinearDensity's results container was changed. (Issue #2508, PR #3617)
```
IAlibay: Add in a one line comment under deprecations that the old results attributes are now deprecated?

package/MDAnalysis/analysis/lineardensity.py
```diff
@@ -57,16 +88,21 @@ class LinearDensity(AnalysisBase):
     ----------
     results.x.dim : int
            index of the [xyz] axes
-    results.x.pos : numpy.ndarray
-           mass density in [xyz] direction
-    results.x.pos_std : numpy.ndarray
+    results.x.mass_density : numpy.ndarray
```
IAlibay, line 93: Rather than replacing the old parameters completely, they should still be here but accompanied with a `.. deprecated` tag.

IAlibay: See here for example: https://github.com/MDAnalysis/mdanalysis/blob/3a03af01886977aa53d2a43086cae86c45038738/package/MDAnalysis/analysis/contacts.py#L373-L378

package/MDAnalysis/analysis/lineardensity.py
```diff
@@ -200,55 +251,54 @@ def _single_frame(self):
...
+                                           bins=self.nbins,
+                                           range=(0.0, max(self.dimensions)))
 
             self.results[dim][key] += hist
             self.results[dim][key_std] += np.square(hist)
+            self.results[dim]['hist_bin_edges'] = bin_edges
 
     def _conclude(self):
-        k = 6.022e-1  # divide by avodagro and convert from A3 to cm3
+        avogadro = 6.022e23  # unit: mol^{-1}
```
IAlibay: pick this up from units.py instead?

package/MDAnalysis/analysis/lineardensity.py
```diff
@@ -31,12 +31,43 @@
...
+    """From version 3.0.0 onwards, some entries in Results will be renamed. See
+    the docstring for LinearDensity for details. The Results class is defined
+    here to implement deprecation warnings for the user."""
+
+    _deprecation_dict = {"pos": "mass_density",
+                         "pos_std": "mass_density_stddev",
+                         "char": "charge_density",
+                         "char_std": "charge_density_stddev"}
+
+    def _deprication_warning(self, key):
```
IAlibay:
```suggestion
    def _deprecation_warning(self, key):
```

package/MDAnalysis/analysis/lineardensity.py
```diff
@@ -200,55 +251,54 @@ def _single_frame(self):
...
-            key_std = 'pos_std'
-            results[dim][key] += other[dim][key]
-            results[dim][key_std] += other[dim][key_std]
-
-            key = 'char'
-            key_std = 'char_std'
-            results[dim][key] += other[dim][key]
-            results[dim][key_std] += other[dim][key_std]
+            for key in self.keys:
+                self.results[dim][key] += other.results[dim][key]
```
IAlibay, line 334: I think I'm missing something here, shouldn't this be averaged?

Note: I'm probably missing something important, but if the answer here is "this never worked at intended" then I'm going to advocate for just removing this.

testsuite/MDAnalysisTests/analysis/test_lineardensity.py
```diff
@@ -47,53 +49,138 @@ def test_invalid_grouping():
...
     assert_allclose(ld.masses, expected_masses)
     assert_allclose(ld.charges, expected_charges)
     # rtol changed here due to floating point imprecision
-    assert_allclose(ld.results['x']['pos'], expected_xpos, rtol=1e-06)
-    assert_allclose(ld.results['x']['char'], expected_xchar)
+    assert_allclose(ld.results.x.mass_density, expected_xmass, rtol=1e-06)
+    assert_allclose(ld.results.x.charge_density, expected_xcharge)
+
+
+def make_testing_Universe():
```
IAlibay: Make this a pytest fixture?

testsuite/MDAnalysisTests/analysis/test_lineardensity.py
```diff
@@ -47,53 +49,138 @@ def test_invalid_grouping():
...
+        assert_allclose(ld.results.x.pos_std, ld.results.x.mass_density_stddev)
+        assert_allclose(ld.results.x.char, ld.results.x.charge_density)
+        assert_allclose(ld.results.x.char_std,
+                        ld.results.x.charge_density_stddev)
+        for key in testdict.keys():
+            assert_allclose(ld.results["x"][key],
+                            ld.results["x"][testdict[key]])
+
+    with warnings.catch_warnings():
+        warnings.simplefilter("error")
```
IAlibay: Use `@pytest.mark.filterwarnings` instead?

package/MDAnalysis/analysis/lineardensity.py
IAlibay: Had a look at the tests and this is a very weird way of parallelizing by doing it over chunks of the system rather than by frames.

This is not the standard way of doing things in MDA (or PMDA). I might advocate that we deprecate and remove this. We should just eventually implement a `_reduce` method instead in the future (as per: https://github.com/MDAnalysis/mdanalysis/issues/2542)

IAlibay (changes requested):
Thanks for the continued work here @BFedder - I've added some extra changes I would like to see here.

Re: `_add_other_results`, I only don't like the way this is meant to work, parallelism over frames is much more efficient / reasonable. That being said, changing this is a) outside the scope of this PR (in my opinion) and b) can't be done immediately (in case someone does actually use this somewhere).

I'll like other @MDAnalysis/coredevs chime in, but I would advocate for just deprecating the method for now. We can add in a `_reduce` method at a later point instead.

package/MDAnalysis/analysis/lineardensity.py
BFedder: I considered this but have decided against it here because in units.py, it's defined as `6.02214129e+23` which would change the results. I changed it this way to make what is happening clearer while not affecting the results of the calculation. Happy to instead use the definition from units.py though if this is preferred.

IAlibay: Not sure, @orbeckst you tend to have some stronger views on what should be in a minor versions vs a major one, can we call this a bug fix and make the switch for 2.2.0?

package/MDAnalysis/analysis/lineardensity.py
BFedder: I haven't seen this method used anywhere in MDA or PMDA, so it's quite possible the test I wrote does not mirror good usage. But I can add a deprecation warning here either way.

IAlibay: I suspect it is the intended usage, but yeah I don't think anyone actually uses it in the wild.

Unless @PicoCentauri disagrees, please do add the deprecation warning (you can use the decorator for it): https://github.com/MDAnalysis/mdanalysis/blob/f494c98dbc1c169854399ff6815f699718d903fe/package/MDAnalysis/lib/util.py#L2296-L2302 and just raise an issue to remove it & switch to a more standard `_reduce` over frames method.

package/MDAnalysis/analysis/lineardensity.py
orbeckst: I'd be in favor of using the consistent value from units.py. 

It does change results but it simply adds more significant digits so it doesn't warrant a major version in my opinion. I'd be ok calling it a fix.

package/MDAnalysis/analysis/lineardensity.py
PicoCentauri: I agree. If there is no usage in the library we can deprecate and remove it!

testsuite/MDAnalysisTests/analysis/test_lineardensity.py
BFedder: I am now using `MDAnalysisTests.util.no_deprecated_call` here instead to check that no deprecation warning is raised when accessing the new attribute names.

testsuite/MDAnalysisTests/analysis/test_lineardensity.py
```diff
@@ -47,53 +50,142 @@ def test_invalid_grouping():
...
+    # Check that no DeprecationWarning is raised with new attributes
+    with no_deprecated_call():
+        ld.results.x.mass_density
+        ld.results.x.mass_density_stddev
+        ld.results.x.charge_density
+        ld.results.x.charge_density_stddev
+
+
+# TODO: deprecated, remove in 3.0.0
+def test_parallel_analysis(testing_Universe):
```
IAlibay, line 177: Can you add a `pytest.warns` somewhere here to check that the deprecation warning is being triggered (unless you're checking this somewhere else and I've missed it).

IAlibay (changes requested):

IAlibay (approved):
Thanks @BFedder lgtm!

@PicoCentauri I'll let you re-review / merge.

PicoCentauri (approved):
Really nice! Thanks @BFedder

Issue #3535, opened 2022-02-15
DeprecationWarnings: numpy - elementwise comparison, calling nonzero on 0d arrays, `np.object`.


IAlibay:
Couple of numpy deprecation warnings we'll need to take care of.

This gets thrown in `hbond_autocorrel.py`: - done

```
hbond_autocorrel.py:435: DeprecationWarning: elementwise comparison failed; this will raise an error in the future.
    pair = np.delete(pair, np.where(pair==exclude), 0)
```

This gets throws in `test_hydrogenbond_autocorrel_deprecated.py`:

```
<__array_function__ internals>:180: DeprecationWarning: Calling nonzero on 0d arrays is deprecated, as it behaves surprisingly. Use `atleast_1d(cond).nonzero()` if the old behavior was intended. If the context of this warning is of the form `arr[nonzero(cond)]`, just use `arr[cond]`.
```

This deprecation warning gets thrown in `test_topologyattr.py`: - done

```
DeprecationWarning: `np.object` is a deprecated alias for the builtin `object`. To silence this warning, use `object` by itself. Doing this will not modify any behavior and is safe.
  Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
    dtype=np.object)
```

MDA version: 2.1.0-dev0

jbarnoud:
One more:

```
MDAnalysisTests/core/test_topologyattrs.py::TestAtomnames::test_next_emptyresidue
MDAnalysisTests/core/test_topologyattrs.py::TestAtomnames::test_prev_emptyresidue
  /home/jon/miniconda3/envs/work/lib/python3.8/site-packages/numpy/testing/_private/utils.py:397: DeprecationWarning: The truth value of an empty array is ambiguous. Returning False, but in future this will result in an error. Use `array.size > 0` to check that an array is not empty.
    if isdesnan and isactnan:
```

IAlibay:
Adding this to 2.2.0 milestone.

ojeda-e:
This looks like a `good first issue`. Would you please confirm @IAlibay? (Pinning you since you opened the issue)

IAlibay:
> This looks like a `good first issue`. Would you please confirm @IAlibay? (Pinning you since you opened the issue)   

Sure sounds good to me

IAlibay:
self assigning this as I'd like to get it done ASAP for 2.2.0

PR #3619, opened 2022-04-11
Updated elementwise comparison

Sukeerti1:
Fixes part of  #3535

 - 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


jbarnoud (changes requested):
Thank you for this contribution. I find the fix very elegant and, from what I see, the tests in `analysis/test_hydrogenbond*` do not raise the problematic warnings anymore.

There are a couple of small things to fix:
* could you fix the PEP8 warnings that the bot found?
* could you add your name in the AUTHORS file in the package directory?
* could you add your github handle and an entry in the CHANGELOG?

With this done, the pull request can be merged.

package/CHANGELOG
```diff
@@ -15,11 +15,12 @@ The rules for this file:
 ------------------------------------------------------------------------------
 ??/??/?? IAlibay, BFedder, inomag, Agorfa, aya9aladdin, shudipto-amin, cbouy,
          HenokB, umak1106, tamandeeps, Mrqeoqqt, megosato, AnirG, rishu235,
-         mtiberti, manishsaini6421
+         mtiberti, manishsaini6421, Sukeerti1
 
  * 2.2.0
 
 Fixes
+  * Updated Elementwise Comparison. (Issue #3535, PR #3619)
```
jbarnoud: This needs to be a bit more explicit. Something in the lines of "Fix Numpy deprecation warnings about elementwise comparisons."

package/MDAnalysis/analysis/hydrogenbonds/hbond_autocorrel.py
```diff
@@ -321,11 +321,15 @@ def __init__(self, universe,
...
-        if self.exclusions:
-            if not len(self.exclusions[0]) == len(self.exclusions[1]):
+        if exclusions is not None:
+            if len(exclusions[0]) != len(exclusions[1]):
                 raise ValueError(
-                    "'exclusion' must be two arrays of identical length")
+                        "'exclusion' must be two arrays of identical length")
+            self.exclusions = np.column_stack((exclusions[0], 
+                                               exclusions[1]))
+                                               .astype(np.intp)
```
jbarnoud: This syntax is invalid and causes the tests to break. Here, `.astype(np.intp)` is taken as a new instruction rather than the continuation of the previous line.

There are various ways to cut this line not to be too long. None is ideal but I would suggest something like:

```python
            self.exclusions = np.column_stack((
                exclusions[0], exclusions[1],
            )).astype(np.intp)
````

Note that I can cut the line where I did because python omits the end of lines until it closes the parentheses. You cannot cut where you did because the line is caught up on parentheses before the dot, and python does not loog ahead to see the dot.

jbarnoud (changes requested):
Something is broken and needs fixing, but then it should be good.

jbarnoud (approved):

jbarnoud:
Congratulations on your first merged contribution :tada: Thank you for addressing this issue.

PR #3620, opened 2022-04-11
MSD example where multiple replicates are combined

### LINKED ISSUE #3578: Documentation should include an MSD example where multiple replicates are combined.

Combining replicates when calculating MSDs should be common practice, however the documentation does not contain an explicit example of how to do this.  

The same cannot be achieved by concatenating the replicas in a single run as the jump between frame -1 and frame 0 of the next trajectory will lead to an artificial inflation of the MSD and hence any subsequent diffusion coefficient calculated.  A note to this effect should probably appear in the docs as well. 

The user will need to use combine the results of several separate calls to `MSD.run()` accessing the `.msd_by_particle ` attribute, concatenating the array along the particle axis and then averaging along the particle axis.

This might look something like 

```python
u1 = mda.Universe(traj, top)
MSD1 = msd.EinsteinMSD(u, select='all', msd_type='xyz', fft=True)
MSD1.run()

u2 = umda.Universe(traj, top)
MSD2 = msd.EinsteinMSD(u, select='all', msd_type='xyz', fft=True)
MSD2.run() 

combined_msds = np.concatenate(MSD1.msds_by_particle, MSD2.msds_by_particle, axis=...) # need to choose the right axis
average_msd = np.mean(combined_msds, axis=...) # need to choose the right axis. 

```

markvrma:
Fixes #3578 

 - Added an example where multiple replicates are combined in docs ( `MDAnalysis/analysis/msd.py` )

PR Checklist
------------
 - [ ] Docs
     - Added an example where multiple replicates are combined in docs
 - [ ] CHANGELOG updated


markvrma:
Due to merge issues that I can't track I'm closing this PR and opening #3630

PR #3635, opened 2022-04-13
Make CRD EXT format optional

### LINKED ISSUE #3605: Make CRD (CHARMM card file) EXT format optional for smaller systems.
Hi,

When writing CRD (CHARMM card) files, it would be useful to allow users to force the EXT format even for systems smaller than 99,999 particles. An optional argument could be quick and simple solution. Any thoughts?

mdpoleto:
Fixes issue #3605 

 - Inclusion of an boolean `crdext` input for the CRDWriter.write() function.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/coordinates/CRD.py
```diff
@@ -125,6 +125,8 @@ class CRDWriter(base.WriterBase):
 
     .. versionchanged:: 0.11.0
        Frames now 0-based instead of 1-based
+    .. versionchanged:: 0.12.0
```
orbeckst: This would be 2.2.0 (as it references the upcoming release)

package/MDAnalysis/coordinates/CRD.py
```diff
@@ -125,6 +125,8 @@ class CRDWriter(base.WriterBase):
...
+       CRD extended format now can be explicitely requested
```
orbeckst: explicitly requested with the `crdext` keyword

package/MDAnalysis/coordinates/CRD.py
```diff
@@ -151,12 +153,23 @@ def __init__(self, filename, **kwargs):
         ----------
         filename : str or :class:`~MDAnalysis.lib.util.NamedStream`
              name of the output file or a stream
+
+        crdext   : bool (optional)
```
orbeckst: just call the kwarg `extended` — writer can have their own kwargs and "ext" is closer to the actual format name

package/MDAnalysis/coordinates/CRD.py
```diff
@@ -151,12 +153,23 @@ def __init__(self, filename, **kwargs):
...
+             By default, noextended CRD format is used [``False``].
+             However, extended CRD format can be forced by
+             specifying `crdext` ``=True``.
```
orbeckst: add the versionchanged here (make sure to have two empty lines above the `..versionchanged`, otherwise the reST/sphinx formatting breaks)

package/MDAnalysis/coordinates/CRD.py
```diff
@@ -151,12 +153,23 @@ def __init__(self, filename, **kwargs):
...
 
         self.filename = util.filename(filename, ext='crd')
         self.crd = None
 
-    def write(self, selection, frame=None):
+        # account for explicit crd format, if requested
+        if "crdext" in kwargs.keys():
+            self.crdext = kwargs.pop('crdext')
+        else:
+            self.crdext = False   # if not requested, default to NOEXT format
```
orbeckst: Can be simplified as
```suggestion
        self.extended = kwargs.pop("extended", False)
```

package/MDAnalysis/coordinates/CRD.py
```diff
@@ -151,12 +153,23 @@ def __init__(self, filename, **kwargs):
...
+
+    def write(self, selection, frame=None, crdext=False):
```
orbeckst: I would think that `extended=False` is only really needed in init. Why would you want to override the format choice when writing?
```suggestion
    def write(self, selection, frame=None):
```

I am open for a discussion, though, just make your case for having the `extended` argument in write(), too.

package/MDAnalysis/coordinates/CRD.py
```diff
@@ -182,23 +195,37 @@ def write(self, selection, frame=None):
             except AttributeError:
                 frame = 0  # should catch cases when we are analyzing a single PDB (?)
 
+        # account for explicit crd format, if requested
+        if self.crdext:
+            crdext = self.crdext
+        else:
+            crdext = crdext   # if not requested, default to NOEXT format
+
```
orbeckst: I think this should all be removed and you use the class/init value in `self.extended`.
```suggestion
```

package/MDAnalysis/coordinates/CRD.py
```diff
@@ -182,23 +195,37 @@ def write(self, selection, frame=None):
...
         atoms = selection.atoms  # make sure to use atoms (Issue 46)
         coor = atoms.positions  # can write from selection == Universe (Issue 49)
 
         n_atoms = len(atoms)
         # Detect which format string we're using to output (EXT or not)
         # *len refers to how to truncate various things,
         # depending on output format!
+
         if n_atoms > 99999:
```
orbeckst: Can be simpified
```suggestion
        if self.extended or n_atoms > 99999:
```

package/MDAnalysis/coordinates/CRD.py
```diff
@@ -182,23 +195,37 @@ def write(self, selection, frame=None):
...
+            if crdext is True:
+                at_fmt = self.fmt['ATOM_EXT']
+                serial_len = 10
+                resid_len = 8
+                totres_len = 10
+            else:
+                at_fmt = self.fmt['ATOM']
+                serial_len = 5
+                resid_len = 4
+                totres_len = 5
```
orbeckst: revert
```suggestion
            at_fmt = self.fmt['ATOM']
            serial_len = 5
            resid_len = 4
            totres_len = 5
```

package/MDAnalysis/coordinates/CRD.py
```diff
@@ -241,7 +268,10 @@ def write(self, selection, frame=None):
             if n_atoms > 99999:
                 crd.write(self.fmt['NUMATOMS_EXT'].format(n_atoms))
             else:
-                crd.write(self.fmt['NUMATOMS'].format(n_atoms))
+                if crdext is True:
+                    crd.write(self.fmt['NUMATOMS_EXT'].format(n_atoms))
+                else:
+                    crd.write(self.fmt['NUMATOMS'].format(n_atoms))
```
orbeckst: simplify as above

orbeckst (changes requested):
Thanks for the PR @mdpoleto . I have a number of initial comments.
- Why do we need the extended kwarg in write() and init? Isn't init sufficient? Right now I'd remove the kwarg from write() but if you can convince me that it should be there I'll change my mind. Is there a specific reason why it needs to be in two places or do other writers do similar things?
- The code can be simplified, see comments (you can accept the suggestions to make your life easier).
- Please add tests for the new functionality.
- Please update CHANGELOG.
- Add yourself to AUTHORS.

mdpoleto:
@orbeckst Thanks for your comments and patience. This is my first PR to a big code like MDA, so I am still learning how to navigate the process. I am not sure how to add test cases for this new functionality. Could you provide some tips to that?

orbeckst:
Contributing to a medium sized package such as MDAnalysis is certainly a learning experience. In the User Guide we have a section [Contributing to the main codebase](https://userguide.mdanalysis.org/stable/contributing_code.html) that you should have a look at.  We're trying to help, and we thank you for _your_ patience in advance. The process can take a while because we strive to maintain high code quality so there will invariably be multiple rounds of reviews, often from different developers. We aren't making petty requests but we are all trying our best to make our code useful and robust for the few thousand users that rely on it for their work. 

Testing is therefore incredibly important; the User Guide has a section on [testing](https://userguide.mdanalysis.org/stable/testing.html#testing). The CRD format can be used both for **topology** information (through the CRDParser) and as a **coordinate** CRDReader and CRDWriter. You are modifying the **CRDWriter** so you need to add tests that check your new functionality. These tests are located in [testsuite/MDAnalysisTests/coordinates/test_crd.py](https://github.com/MDAnalysis/mdanalysis/blob/develop/testsuite/MDAnalysisTests/coordinates/test_crd.py). Look for tests for writing. Then start by adding your own test, modelled on the existing one, where you use your new feature. Check in your test that the output is correct. 

During development I run tests locally using `pytest` — the User Guide tells you more. 

When you push your changes to the branch, have a look at the *Checks* and the coverage. It can tell you which lines have not been tested.

package/CHANGELOG
```diff
@@ -34,6 +34,8 @@ Enhancements
     monatomic ion charges or edge cases with nitrogen, sulfur, phosphorus and
     conjugated systems should now have correctly assigned bond orders and
     charges.
+  * CRD extended format can now be explicitly requested with the `extended`
```
orbeckst, line 54: reference the issue as in (most) other entries

mdpoleto:
@orbeckst Thanks for all the info. I did read the "Contributing to the main codebase" section, which was very helpful. The testing section now was very helpful.

I included a test for the extended format and it ran just fine locally. I included it on my last commit. If anything else is needed, please let me know.

testsuite/MDAnalysisTests/coordinates/test_crd.py
```diff
@@ -66,11 +66,13 @@ def CRD_iter(fn):
...
-    def test_write_EXT(self):
-        # TODO: Write tests that use EXT output format
-        # Must have *lots* of atoms, maybe fake the system
-        # to make tests faster
-        pass
+    def test_write_EXT(self, u, outfile):
+        # Use the `extended` keyword to force the EXT format
+        try:
+            u.atoms.write(outfile, extended=True)
+        except Exception:
```
tylerjereddy: Allowing a test to fail "naturally" is probably the right thing to do instead of capturing all possible exceptions and then re-raising (there's no real advantage to the generic capture, and it may even make the failure harder to diagnose).

Some other thoughts here:

1) The test as written would produce an artifact in the user/developer space because it is not scoped to a temporary directory. There are examples of using the `tmpdir` `pytest` fixture in this module, so that the test files get written without leaving artifacts behind. If you do use the examples in this testing module as inspiration, I'd suggest avoiding the construct with `outfile = str(tmpdir) + '/out.crd' ` and favoring something like `os.path.join` to compose the path in a platform-portable way (we should probably fix that, but it is orthogonal to this PR)
2) Is there anything you could check/verify in the output file that gets written? For example, some string/binary data that is unique to the `EXT` format that would signal that MDAnalysis has done the right thing? Making an assertion on that would make the test much more robust.

mdpoleto: @tylerjereddy Thanks for your comments. I used `test_pdb.py` as inspiration to how the `outfile` variable is defined. Let me know if that solves the issue. 

Also, CRD files in extended format are expected to have a `EXT` string in the third line. I made an assertion for that existence that ran just fine locally. I hope that enhances the test.

package/MDAnalysis/coordinates/CRD.py
```diff
@@ -151,11 +154,24 @@ def __init__(self, filename, **kwargs):
...
+
+        extended : bool (optional)
+             By default, noextended CRD format is used [``False``].
+             However, extended CRD format can be forced by
+             specifying `extended` ``=True``.
+
+
+        .. versionchanged:: 2.2.0
+           CRD extended format can now be explicitly requested with the
+           `extended` keyword
```
orbeckst: You already have the version changed in the class docs. Perhaps we just mark the new argument as added here. Check if this formats correctly.

```suggestion

             .. versionadded:: 2.2.0
```

testsuite/MDAnalysisTests/coordinates/test_crd.py
```diff
@@ -66,11 +68,14 @@ def CRD_iter(fn):
...
-        # to make tests faster
-        pass
+    def test_write_EXT(self, u, outfile):
+        # Use the `extended` keyword to force the EXT format
+        u.atoms.write(outfile, extended=True)
+
+        with open(outfile, 'r') as inf:
+            format_line = inf.readlines()[2]
+            assert 'EXT' in format_line, "EXT format expected"
+
```
orbeckst: Add a test `test_write_EXT_read`  that MDAnalysis can *read* the file, too.

orbeckst (changes requested):
Looks really good! I have a minor doc comment and I'd like to see a second test that shows that we can *read* the extended file that we wrote. It would look similar to the test you wrote but now you also read back the `outfile` and assert that you get the right number of atoms, residues, and perhaps the center of geometry of all positions, and the total charge of the system (or something like that — you get the idea, test in aggregate so that you see if something is not read properly).

testsuite/MDAnalysisTests/coordinates/test_crd.py
mdpoleto: I included a test to assert: 1) number of atoms, 2) number of residues, 3) number of segments and 4) center of geometry. 

Since the EXT format supports more decimals for coordinates than the standard noextended format, I used `np.around(cog,6)` for the assertion of center of geometries.

testsuite/MDAnalysisTests/coordinates/test_crd.py
```diff
@@ -40,7 +42,8 @@ def u(self):
 
     @pytest.fixture()
     def outfile(self, tmpdir):
-        return str(tmpdir) + '/out.crd'
+        #return str(tmpdir) + '/out.crd'
+        return os.path.join(str(tmpdir), 'test.crd')
```
richardjgowers: why this change?  iirc `tmpdir` is a `Path` object so should work ok as is?

mdpoleto: From what I understood from @tylerjereddy 's comment, using this construct is preferred since it does not leave artifacts behind. See full discussion at: https://github.com/MDAnalysis/mdanalysis/pull/3635#discussion_r851043069

mdpoleto:
@orbeckst at the time of my last commit all checks passed but now there seems to be a conflict with AUTHORS. Should I address it?

testsuite/MDAnalysisTests/coordinates/test_crd.py
orbeckst: This was indeed what @tylerjereddy suggested.

Maybe @richardjgowers was thinking to use
```python
return str(tmpdir / 'test.crd')
```
We do need a `str` for Universe as we cannot use py.path or pathfile objects yet.

By the way, please remove commented out code.

orbeckst:
@mdpoleto yes please resolve the conflict — just add your name at the end of the AUTHORS list. 

We just had another PR merged that collided with yours. That's not uncommon.

testsuite/MDAnalysisTests/coordinates/test_crd.py
```diff
@@ -66,11 +69,36 @@ def CRD_iter(fn):
...
-        pass
+    def test_write_EXT(self, u, outfile):
+        # Use the `extended` keyword to force the EXT format
+        u.atoms.write(outfile, extended=True)
+
+        with open(outfile, 'r') as inf:
+            format_line = inf.readlines()[2]
+            assert 'EXT' in format_line, "EXT format expected"
+
+    def test_read_EXT(self, u, outfile):
```
orbeckst: call the test `test_write_EXT_read`

package/AUTHORS
```diff
@@ -191,7 +191,9 @@ Chronological list of authors
   - Sukeerti T
   - Robot Jelly
   - Mark Verma
+  - Marcelo D. Poleto
   
+>>>>>>> c0a4f610656baf08294d01ebff2fafc859887cfc
```
orbeckst: remove conflict marker

testsuite/MDAnalysisTests/coordinates/test_crd.py
```diff
@@ -66,11 +69,36 @@ def CRD_iter(fn):
...
+        u.atoms.write(outfile, extended=True)
+
+        u2 = mda.Universe(outfile)
+
+        sel1 = u.select_atoms('all')
+        sel2 = u2.select_atoms('all')
+
+        # Rounding floats since EXT format support more decimals
+        cog1 = np.around(sel1.center_of_geometry(),6)
+        cog2 = np.around(sel2.center_of_geometry(),6)
```
orbeckst: Do not round results.

Instead use `assert_allclose` or `assert_almost_equal`

testsuite/MDAnalysisTests/coordinates/test_crd.py
```diff
@@ -66,11 +69,36 @@ def CRD_iter(fn):
...
+
+        assert_equal(len(u.atoms), len(u2.atoms)), 'Equal number of '\
+                'atoms expected in both CRD formats'
+        assert_equal(len(u.atoms.residues),
+            len(u2.atoms.residues)), 'Equal number of residues expected in'\
+                        'both CRD formats'
+        assert_equal(len(u.atoms.segments), 
+            len(u2.atoms.segments)), 'Equal number of segments expected in'\
+                        'both CRD formats'
+        assert_equal(cog1, cog2), 'Same centroid expected for both CRD formats'
```
orbeckst: Use [`assert_allclose()`](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html) with proper tol=1e-6 and atol=0

orbeckst (changes requested):
please see comments, sorry to be short

mdpoleto:
@orbeckst I believe I addressed them all. Let me know if anything else is needed

testsuite/MDAnalysisTests/coordinates/test_crd.py
```diff
@@ -66,11 +67,36 @@ def CRD_iter(fn):
...
+    def test_write_EXT_read(self, u, outfile):
+        # Read EXT format and check atom positions
+        u.atoms.write(outfile, extended=True)
+
+        u2 = mda.Universe(outfile)
+
+        sel1 = u.select_atoms('all')
+        sel2 = u2.select_atoms('all')
+
+        # Rounding floats since EXT format support more decimals
```
orbeckst: don't need the comment
```suggestion
```

testsuite/MDAnalysisTests/coordinates/test_crd.py
```diff
@@ -66,11 +67,36 @@ def CRD_iter(fn):
...
-        # Must have *lots* of atoms, maybe fake the system
-        # to make tests faster
-        pass
+    def test_write_EXT(self, u, outfile):
+        # Use the `extended` keyword to force the EXT format
+        u.atoms.write(outfile, extended=True)
+
+        with open(outfile, 'r') as inf:
+            format_line = inf.readlines()[2]
+            assert 'EXT' in format_line, "EXT format expected"
```
orbeckst: can be outside context manager
```suggestion
        assert 'EXT' in format_line, "EXT format expected"
```

package/MDAnalysis/coordinates/CRD.py
```diff
@@ -151,11 +154,21 @@ def __init__(self, filename, **kwargs):
         ----------
         filename : str or :class:`~MDAnalysis.lib.util.NamedStream`
              name of the output file or a stream
+
+        extended : bool (optional)
+             By default, noextended CRD format is used [``False``].
+             However, extended CRD format can be forced by
+             specifying `extended` ``=True``.
```
orbeckst: Make clear that we still use EXT of necessary.
```suggestion
             specifying `extended` ``=True``. Note that the extended format
             is *always* used if the number of atoms exceeds 99,999, regardless
             of the setting of `extended`.
```

orbeckst (approved):
All good. Thank you for your contribution!

I had some minor comments that I am adding to the PR.

When the tests come back all clear, I will squash-merge the PR.

orbeckst:
Congratulations on your first contribution, @mdpoleto !

I'll send you an invite for the org so that you're also listed as a Contributor, if you accept.

mdpoleto:
Thanks for all the help and guidance @orbeckst

PR #3640, opened 2022-04-15
Bat cartesian method modifies input data

### LINKED ISSUE #3501: BAT method Cartesian modifies input data

umak1106:
Fixes #3501 

 - Use copy.deepcopy() to sort the issue caused by scope and pass-by-reference resulting addition of non-primary torsions in the input of BAT data
 - Wrote a regression test for the same

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/testsuite/MDAnalysisTests/analysis/test_bat.py
+++ b/testsuite/MDAnalysisTests/analysis/test_bat.py
@@ -131,10 +131,8 @@ class TestBAT(object):
 
     def test_bat_transformation(self, selected_residues, bat):
         R = BAT(selected_residues)
-        R.run()
-        XYZ = R.results.bat[0]
-        PQR = copy.deepcopy(XYZ)
-        R.Cartesian(R.results.bat[0])
-        assert_almost_equal(PQR, XYZ, 5,
-                            err_msg="error: Bat input data"
-                                    "modified by the Cartesian method")
+        pre_transformation = copy.deepcopy(bat[0])
+        R.Cartesian(bat[0])
+        assert_almost_equal(pre_transformation, bat[0], 5,
+                            err_msg="error: Bat.Cartesian"
+                                    "modified input data")
--- a/testsuite/MDAnalysisTests/analysis/test_bat.py
+++ b/testsuite/MDAnalysisTests/analysis/test_bat.py
@@ -133,7 +133,7 @@ class TestBAT(object):
         XYZ = R.Cartesian(bat[0])
         PQR = R.Cartesian(bat[0])
         assert_almost_equal(PQR, XYZ, 5,
-                           err_msg="error: Reconstructed Cartesian"
-                                   "coordinate after multiple "
-                                   "transformations don't match "
-                                   "original")
+                            err_msg="error: Reconstructed Cartesian"
+                                    "coordinate after multiple "
+                                    "transformations don't match "
+                                    "original")
--- a/testsuite/MDAnalysisTests/analysis/test_bat.py
+++ b/testsuite/MDAnalysisTests/analysis/test_bat.py
@@ -127,11 +127,12 @@ class TestBAT(object):
         errmsg = 'Dimensions of array in loaded file'
         with pytest.raises(ValueError, match=errmsg):
             R = BAT(selected_residues, filename=bat_npz)
-            
-    def test_bat_between_transformations(self, selected_residues, bat):
+
+    def test_bat_transformation(self, selected_residues, bat):
         R = BAT(selected_residues)
         XYZ = R.Cartesian(bat[0])
         PQR = R.Cartesian(bat[0])
         assert_almost_equal(PQR, selected_residues.positions, 5,
-            err_msg="error: Cartesian coordinates after multiple transformation" + \
-                    "don't match original")
+              err_msg="error: Reconstructed Cartesian coordinates"
+                      "after multiple transformations don't match "
+                      "original")
--- a/testsuite/MDAnalysisTests/analysis/test_bat.py
+++ b/testsuite/MDAnalysisTests/analysis/test_bat.py
@@ -127,3 +127,12 @@ class TestBAT(object):
         errmsg = 'Dimensions of array in loaded file'
         with pytest.raises(ValueError, match=errmsg):
             R = BAT(selected_residues, filename=bat_npz)
+            
+    def test_bat_between_transformations(self, selected_residues, bat):
+        R = BAT(selected_residues)
+        XYZ = R.Cartesian(bat[0])
+        PQR = R.Cartesian(bat[0])
+        assert_almost_equal(PQR, selected_residues.positions, 5,
+            err_msg="error: Reconstructed Cartesian coordinates " + \
+                    "after multiple transformations " + \
+                    "don't match original")
--- a/testsuite/MDAnalysisTests/analysis/test_bat.py
+++ b/testsuite/MDAnalysisTests/analysis/test_bat.py
@@ -132,7 +132,8 @@ class TestBAT(object):
         R = BAT(selected_residues)
         XYZ = R.Cartesian(bat[0])
         PQR = R.Cartesian(bat[0])
-        assert_almost_equal(PQR, selected_residues.positions, 5,
-              err_msg="error: Reconstructed Cartesian coordinates"
-                      "after multiple transformations don't match "
-                      "original")
+        assert_almost_equal(PQR, XYZ, 5,
+                           err_msg="error: Reconstructed Cartesian"
+                                   "coordinate after multiple "
+                                   "transformations don't match "
+                                   "original")
--- a/testsuite/MDAnalysisTests/analysis/test_bat.py
+++ b/testsuite/MDAnalysisTests/analysis/test_bat.py
@@ -132,12 +132,9 @@ class TestBAT(object):
     def test_bat_transformation(self, selected_residues, bat):
         R = BAT(selected_residues)
         R.run()
-        R.Cartesian(R.results.bat[0])
         XYZ = R.results.bat[0]
         PQR = copy.deepcopy(XYZ)
         R.Cartesian(R.results.bat[0])
         assert_almost_equal(PQR, XYZ, 5,
-                            err_msg="error: Reconstructed Cartesian"
-                                    "coordinate after multiple "
-                                    "transformations don't match "
-                                    "original")
+                            err_msg="error: Bat input data"
+                                    "modified by the Cartesian method")
--- a/testsuite/MDAnalysisTests/analysis/test_bat.py
+++ b/testsuite/MDAnalysisTests/analysis/test_bat.py
@@ -25,6 +25,7 @@ from __future__ import absolute_import
 import numpy as np
 from numpy.testing import assert_equal, assert_almost_equal
 import pytest
+import copy
 
 import MDAnalysis as mda
 from MDAnalysisTests.datafiles import (PSF, DCD, mol2_comments_header, XYZ_mini,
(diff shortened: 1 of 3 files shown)


umak1106:
Hello mentors , 
Regarding the regression test , 
The issue stated that the  co-ordinates differed from the original one  after the second transformation  
So according to the test I wrote i check if the the co-ordinates after the second transformation as same as the original one.
Because if that passes then the co-ordinates will be same even after multiple transformations . 

Incase there is something wrong with my test or if you could suggest a better alternative that would be great

jbarnoud:
I don't have the time to look at this in details right now but here are a couple of things for you to look at :
* the bot found some formatting issues with your code. It would be good to fix them. Look at the comment above.
* without your fix, does the test you wrote fail? This is a check I try to apply to my own tests. If your test does not fail without your fix, it may not be testing your problem well enough.
* I see you fixed the issues with the torsions, I see very similar lines just above for angles and bonds. Should they be fixed as well?

umak1106:
Yes  @jbarnoud sir , about the formatting issues actually some of the other tests already written in the file had the same formatting issue . So I didn't change that part . But I'll make the changes . 

if we do not use copy.deepcopy() 
then the test I wrote fails . I already checked it on my local pc . 

Umm the issue mentioned that the problem was arising because of the torsions so I don't think the angles and bonds will create a problem . But still we can make a  deepcopy there too just in case that it doesn't create an issue later . Kindly tell me if I should make this change for angles and bonds part  too ..

testsuite/MDAnalysisTests/analysis/test_bat.py
```diff
@@ -127,3 +127,12 @@ def test_bat_incorrect_dims(self, bat_npz):
         errmsg = 'Dimensions of array in loaded file'
         with pytest.raises(ValueError, match=errmsg):
             R = BAT(selected_residues, filename=bat_npz)
+
+    def test_bat_transformation(self, selected_residues, bat):
+        R = BAT(selected_residues)
+        XYZ = R.Cartesian(bat[0])
```
lilyminium: You don't use this line.

lilyminium (changes requested):
Thanks for the fix @umak1106, great job on the test. Could you please update CHANGELOG as well?

testsuite/MDAnalysisTests/analysis/test_bat.py
umak1106: Umm actually  , the Cartesian method should atleast run twice before the co-ordinates are checked with the original ones , which is why I think it's necessary to put that line . 😅

lilyminium: In that case, my mistake -- this test is not actually testing the right problem. I think you are testing an indirect side effect by comparing the Cartesian transformation. It would be better to directly test the problem noticed in https://github.com/MDAnalysis/mdanalysis/issues/3501#issue-1098151358 by checking that `bat[0]` has not changed. Otherwise this test is quite confusing as written.

lilyminium (changes requested):

testsuite/MDAnalysisTests/analysis/test_bat.py
umak1106: Umm actually the issue arises while we run the cartesian method  in torsions = bat_frame[2*n_torsions + 9:] due to pass by reference. But as the cartesian method doesn't return torsions and as torsions or changes in bat[0]/bat_frame are local to the function how do i test that ?
 
The change in torsions results in a change in cartesian co-ords which are returned by the method which is why i tested that .

umak1106: Which is why i am checking the cartesian-cords after the 2nd transformation (PQR) and after the 1st transformation(XYZ) according to my latest commit.

lilyminium: @umak1106 if I understand the issue correctly, the problem is not that PQR != XYZ, but that `bat[0]` before executing `R.Cartesian` does not equal `bat[0]` afterwards.

PQR != XYZ is a side effect of that.

umak1106: Yes mam , but i tried writing this before and comparing bat[0] but the test doesn't fail without the fix and there is no way of
accessing bat_frame/torsions from the cartesian method either so am kinda stuck . Could you suggest what i should do instead ?
R = BAT(selected_residues)
        R.run()
        R.Cartesian(bat[0])
        XYZ = R.results.bat[0]
        R.Cartesian(bat[0])
        PQR = R.results.bat[0]
        assert_almost_equal(XYZ, PQR, 5,
                            err_msg="error: Reconstructed Bat"
                                    "coordinates after multiple "
                                    "transformations don't match"
                                    "original")

lilyminium: Is there any other `bat` in the test than the one you compare with XYZ and PQR?

umak1106: No I don't think so 😅.

lilyminium: What do you pass to R.Cartesian?

umak1106: Bat-frame

umak1106: Yup I figured out my mistake . Ill make the changes right now in 5 mins

umak1106:
@lilyminium 
According to my latest test what I did was run the cartesian method and store the bat co-ords in xyz . make a copy of them in pqr . Then run the cartesian method again , so according to the issue the bat co-ords ie xyz gets modified therefore the assert_almost_equal test fails in absence of the fix in bat.py.

testsuite/MDAnalysisTests/analysis/test_bat.py
```diff
@@ -127,3 +128,16 @@ def test_bat_incorrect_dims(self, bat_npz):
...
+        R.run()
+        R.Cartesian(R.results.bat[0])
+        XYZ = R.results.bat[0]
+        PQR = copy.deepcopy(XYZ)
+        R.Cartesian(R.results.bat[0])
+        assert_almost_equal(PQR, XYZ, 5,
+                            err_msg="error: Reconstructed Cartesian"
+                                    "coordinate after multiple "
+                                    "transformations don't match "
+                                    "original")
```
lilyminium: Could you please explain this test to me? Which lines test the specific problem in #3501 ? Is `R.run()` necessary if there's already a `bat` fixture? Why have you chosen to call the coordinates XYZ and PQR -- can you think of more informative variable names? Why do you run `R.Cartesian` twice? Does the error message still make sense for the assertion?

Keep in mind that the specific problem in #3501 is (quoting from the first comment):

> When calling the Cartesian method of a BAT class instance to convert BAT coordinates back to Cartesian, the input BAT data should not be modified.

lilyminium (changes requested):

testsuite/MDAnalysisTests/analysis/test_bat.py
umak1106: If we do not use R.run() then we get the errors that Results object has no attribute bat . Yes should I call XYZ as bat coords and PQR as deepcopy of  original bat co-ords?
No it wasn't necessary to run Cartesian twice. I changed the error message .

umak1106: According to my latest commit I store the original/input bat coords in XYZ and create a deepcopy of them in PQR. Then after running the cartesian method the bat coords(XYZ) get modified  and so aren't equal to PQR and so the test fails in absence of the fix

testsuite/MDAnalysisTests/analysis/test_bat.py
```diff
@@ -127,3 +128,13 @@ def test_bat_incorrect_dims(self, bat_npz):
...
+
+    def test_bat_transformation(self, selected_residues, bat):
+        R = BAT(selected_residues)
+        R.run()
+        XYZ = R.results.bat[0]
+        PQR = copy.deepcopy(XYZ)
+        R.Cartesian(R.results.bat[0])
+        assert_almost_equal(PQR, XYZ, 5,
+                            err_msg="error: Bat input data"
+                                    "modified by the Cartesian method")
```
lilyminium: How about this instead? The below suggestion does not need to `.run()` again, which saves time. It also has more informative function and variable nams.

```suggestion
    def test_Cartesian_does_not_modify_input(self, selected_residues, bat):
        R = BAT(selected_residues)
        pre_transformation = copy.deepcopy(bat[0])
        R.Cartesian(bat[0])
        assert_almost_equal(
            pre_transformation, bat[0],
            err_msg="BAT.Cartesian modified input data"
        )
```

lilyminium (changes requested):

testsuite/MDAnalysisTests/analysis/test_bat.py
```diff
@@ -127,3 +128,11 @@ def test_bat_incorrect_dims(self, bat_npz):
         errmsg = 'Dimensions of array in loaded file'
         with pytest.raises(ValueError, match=errmsg):
             R = BAT(selected_residues, filename=bat_npz)
+
+    def test_bat_transformation(self, selected_residues, bat):
+        R = BAT(selected_residues)
+        pre_transformation = copy.deepcopy(bat[0])
+        R.Cartesian(bat[0])
+        assert_almost_equal(pre_transformation, bat[0], 5,
```
lilyminium:
```suggestion
        assert_almost_equal(pre_transformation, bat[0],
```

We don't want changes of lower precision to slide without notice -- we don't want any changes to the input at all.

testsuite/MDAnalysisTests/analysis/test_bat.py
```diff
@@ -127,3 +128,11 @@ def test_bat_incorrect_dims(self, bat_npz):
...
+                            err_msg="error: Bat.Cartesian"
+                                    "modified input data")
```
lilyminium:
```suggestion
                            err_msg="error: BAT.Cartesian modified input data")
```

This is missing a space. Also, the name is BAT, not Bat. If it triggers a PEP8 error, consider using the formatting I suggested previously.

lilyminium (changes requested):

lilyminium:
And please change the name of the test to something more informative like the one I suggested above.

lilyminium (approved):
Thank you @umak1106!

PR #3642, opened 2022-04-16
Rewind singleframe before on iter call

### LINKED ISSUE #3423: Iterating SingleFrame readers does not call rewind()
Indirectly related to #3416 

## Expected behavior ##

All readers should behave similarly.

## Actual behavior ##

See: https://github.com/MDAnalysis/mdanalysis/pull/3415#issuecomment-922109775

Iterating a SingleFrame reader does not rewind the reader, which means that any modified coordinates remain the same. This differs in behaviour from the trajectory readers which do rewind (and therefore update positions).

``` python
import MDAnalysis as mda
from MDAnalysisTests.datafiles import two_water_gro, GRO, XTC

# single frame
u = mda.Universe(two_water_gro)

assert u.atoms[0].position[2] != 0

for at in u.atoms:
    at.position = at.position[0], at.position[1], 0

assert u.atoms[0].position[2] == 0

for ts in u.trajectory:
    pass

assert u.atoms[0].position[2] == 0

# trajectory
u = mda.Universe(GRO, XTC)

assert u.atoms[0].position[2] != 0

for at in u.atoms:
    at.position = at.position[0], at.position[1], 0

assert u.atoms[0].position[2] == 0

for ts in u.trajectory:
    pass

assert u.atoms[0].position[2] != 0
```

- Which operating system? All under CI

IAlibay:
Fixes #3423 

There might be a philosophical argument here as to whether rewind should be before or after we yield the timestep. I see it as we should rewind / re-read from file _before_ ts is provided, since that's closer to what happens with trajectories, but others may not agree.

 - Make it such that rewind is called _before_ yielding ts on a `__iter__()` call for SingelFrameReaderBase.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


IAlibay:
@orbeckst this should hopefully sufficiently fix things for #3415

richardjgowers (approved):
The unstated gotcha here is that rewinding clears any modifications that have been made to the single frame since loading.  But yes this does match the existing gotcha that editing an entire trajectory also has the same behaviour, so I'm in favour?

IAlibay:
Yeah maybe we should be quite loud about this - technically it's a "bug" but I'm a bit concerned that some folks might be relying on this undefined behaviour.

Any suggestions for suitably letting users know of this would be appreciated.

richardjgowers (approved):
In general we could have a flag that is triggered when positions are modified and then if the frame is iterated we warn about losing modifications?  This would have a similar feel to that pandas warning you get when modifying a subcopy of a dataframe.  But if someone has happily written out / used their modified coordinates we've warned for no reason. It's hard to determine when the user has shot themselves in the foot with this behaviour.

I think best we can do is make sure it's well documented behaviour and hope that this is read.

IAlibay:
ha we have at least two tests relying on the undefined behaviour...

orbeckst:
Quick comment (don’t have time for a proper review): rewinding before yield is better than after because then you get a chance to work with the frame even outside a loop. So I agree with your choice. 

I wouldn’t put in warnings, just text. This has just been poorly defined and now it’s getting fixed and properly described.

package/CHANGELOG
```diff
@@ -20,6 +20,9 @@ The rules for this file:
  * 2.2.0
 
 Fixes
+  * Iterating a SingleFrameReaderBase trajectory now rewinds the
+    reader, resetting any modified trajectory attributes to those
+    read from file (Issue #3423)
```
orbeckst: Add a sentence like

Fixed behavior that was previously undefined.

package/MDAnalysis/coordinates/base.py
```diff
@@ -2301,6 +2301,9 @@ class SingleFrameReaderBase(ProtoReader):
     .. versionchanged:: 0.11.0
        Added attribute "_ts_kwargs" for subclasses
        Keywords "dt" and "time_offset" read into _ts_kwargs
+    .. versionchanged:: 2.1.0
+       Calling `__iter__` now rewinds the reader before yielding a
+       :class:`Timestep` object.
```
orbeckst: Fixed behavior that was previously not well defined.

orbeckst (changes requested):
I agree with the design choices.

We now have to change the failing tests so that they follow the new corrected behavior.

Please also add some text to the "Trajectory API" in the docs (I think it still lives in the `coordinates/__init__.py` docstring).

orbeckst:
It looks as if the change here uncovered a bug that we hadn't caught before: the `copy()` method of `ReaderBase` and `SingleFrameReader` did not copy the state of `convert_units` and *always* used the default `False`. I fixed it in ecc0ab7aa8c3a5a3ee9664e0b162642dc251ee75 . 

I'd appreciate another pair of 👀 ... @jbarnoud @richardjgowers maybe?

orbeckst:
@IAlibay I _think_ I added everything that is necessary to finalize this PR but please have a look.

package/MDAnalysis/coordinates/base.py
```diff
@@ -134,6 +141,8 @@
 .. autoclass:: FrameIteratorIndices
```
orbeckst, line 143: This is a duplicate label and needs to be made unique.

orbeckst: fixed in 9c65128f6e1d6b591eb7e03b087f9c8ed18021d7

orbeckst (approved):

orbeckst:
@IAlibay please have a last look at this PR. Because I contributed to it and changed your code, I don't want to merge it without you having a last look. Thanks.

(For course, if anyone else wants to look again then that would be great, too.)

IAlibay:
Yup it's on my to-do list for today @orbeckst !

package/MDAnalysis/coordinates/base.py
```diff
@@ -2159,11 +2170,20 @@ def copy(self):
         New Reader will have its own file handle and can seek/iterate
         independently of the original.
 
-        Will also copy the current state of the Timestep held in
-        the original Reader
+        Will also copy the current state of the Timestep held in the original
+        Reader.
+
+
+        .. warning:: No special kwargs are copied, only `convert_units` is
```
IAlibay, line 2177: This seems like it's a bad issue waiting to happen. Is there any way we could cache the kwargs and re-use them instead of just doing `convert_units`? It doesn't have to be in this PR, but it would be good to fix this properly.

IAlibay (commented):
Overall lgtm, that being said I do wonder if we should just do store the rest of the kwargs too to make it a true copy instead of just dealing with the one argument?

Something like

```python
self._kwargs = kwargs
```
in `__init__`

and then:

```
new = self.__class__(self.filename, n_atoms=self.n_atoms, convert_units=self.convert_units, **self._kwargs)
```

in `copy`

package/MDAnalysis/coordinates/base.py
orbeckst: Yes, I just fixed the immediate thing (which wasn’t even part of the original issue but needed fixing). I don’t have time to write the full fix and tests. Raise issue and merge???

IAlibay: Sure, sounds good.

PR #3643, opened 2022-04-16
Improve docstring

### LINKED ISSUE #3313: Improve HELANAL docstring

Following the discussion of https://github.com/MDAnalysis/mdanalysis/pull/3308#r633977107 the docstring of the HELANAL class could be improved. First of all 

https://github.com/MDAnalysis/mdanalysis/blob/e6b3c2fe208c506b6f61b9f9d58f46be9d09e22d/package/MDAnalysis/analysis/helix_analysis.py#L309-L310

should mention that this only is of use when `len(groups) > 1`. Additionally, an example in the module docstring would be great.

Improve the docstring.

 Leave everything as it is

robotjellyzone:
Fixes #3313 

 - Improved the docstring for `split_residue_sequences` parameter of  **`HELANAL`** class

PR Checklist
------------
 - [ ] Tests?
 - [x] Docs?
 - [ ] CHANGELOG updated?


robotjellyzone:
so sorry, due to not updating my local repo , some unwanted commits are also pushed from my previous branch (which was merged) can you remove these unwanted commits while merging?

package/MDAnalysis/analysis/helix_analysis.py
```diff
@@ -307,7 +307,14 @@ class HELANAL(AnalysisBase):
...
     split_residue_sequences : bool, optional
-        Wether to split residues into a list of same or consecutive helices.
+        Whether to split residues into a list of same or consecutive helices.
+        If False, the residues will be analysed as a single helix.
+        Holds True, if more than one `select` string given Otherwise False.
+        i.e., if len(select) > 1, then split_residue_sequences = True
+        example-1: if select=['protein and name CA', 'protein and name CA']
+                   then split_residue_sequences = True
+        example-2: if select=['protein and name CA']
+                   then split_residue_sequences = False
```
lilyminium: Have you checked if this is actually true? There's an example of how to use HELANAL at the user guide here: https://userguide.mdanalysis.org/stable/examples/analysis/structure/helanal.html which you can download as a Jupyter notebook here: https://github.com/MDAnalysis/UserGuide/blob/develop/doc/source/examples/analysis/structure/helanal.ipynb

Maybe try passing in your example `select` sequence and see if it has the effect described here.

lilyminium (changes requested):

package/MDAnalysis/analysis/helix_analysis.py
robotjellyzone: > Have you checked if this is actually true? There's an example of how to use HELANAL at the user guide here: https://userguide.mdanalysis.org/stable/examples/analysis/structure/helanal.html which you can download as a Jupyter notebook here: https://github.com/MDAnalysis/UserGuide/blob/develop/doc/source/examples/analysis/structure/helanal.ipynb
> 
> Maybe try passing in your example `select` sequence and see if it has the effect described here.

Hi, I have written this actually according to test examples shared by @PicoCentauri in this  ->https://github.com/MDAnalysis/mdanalysis/issues/3313#issuecomment-1099137937

lilyminium: It's always wise to test something works instead assuming it -- even core developers are not always right, or have the perfect solution! We do not want to have incorrect documentation as that will be frustrating for users. When you write code or documentation you should always try to verify that it works for yourself, instead of just trusting that it will.

robotjellyzone: oh yeah alright! I will test it for sure!

robotjellyzone: oh yeah alright! I will test it for sure!

robotjellyzone: Oh yeah ok alright! I will test it for sure !

robotjellyzone: Oh yeah ok alright! I will test it for sure !

robotjellyzone: @lilyminium i think i need to change the statements for this `split_residues_sequences`. I ran the code in the notebook & i got to know that this `split_residues_sequences` will come into consideration only when len(groups) > 1 ie when `selection` has gaps in the residues. ie. if we have `selection = ['resid 6:50 or resid 100:130 or resid 132:148']` then only the `len(groups)>1` and then only we consider the passed value of this parameter `split_residues_sequences` . ie if the passed value for this parameter is True then it will split the residue having gaps otherwise will not split it.
as can be seen in the code  --> 
image

robotjellyzone: can you please confirm my finding so that i can make changes in the commit accordingly.

lilyminium: Yeah, that's right -- good job figuring that out! It would be ideal if in the documentation, you explained what kind of `select` string might lead to `len(groups) > 1`, as `groups` is not passed into the analysis but the user themselves.

robotjellyzone: @lilyminium Can you please have a look at new commit, I have already explained about the type of select string in that & made the changes!

package/MDAnalysis/analysis/helix_analysis.py
```diff
@@ -307,7 +307,18 @@ class HELANAL(AnalysisBase):
...
+        If True, split residues having gaps into a list of consecutive helices.
+        This will be checked when `select` is having gaps in residue.
+        example-1: if select=['resid 6:50 or resid 100:130 or resid 132:148']
+                   and split_residue_sequences = True
+                   then selection has gaps in the residues.
+                   So Splitting into 3 helices.
+        example-2: if select=['resid 6:50', 'resid 100:130', 'resid 132:148']
+                   and split_residue_sequences = True
+                   then selection has no gaps in the residues.
+                   So Splitting into 1 helix.
```
lilyminium:
```suggestion
        Whether to split the residue sequence into individual helices.
        This keyword only applies if a residue gap is present in the
        AtomGroup generated by a ``select`` string.
        If ``False``, the residues will be analysed as a single helix.
        If ``True``, each group of consecutive residues will be treated as a separate helix.
```

lilyminium: The below might be a more concise summary of this docstring.

```suggestion
        Whether to split the residue sequence into individual helices.
        This keyword only applies if a residue gap is present in the
        AtomGroup generated by a ``select`` string.
        If ``False``, the residues will be analysed as a single helix.
        If ``True``, each group of consecutive residues will be treated as a separate helix.
```

We typically do not put examples in a parameter's documentation; instead, there is usually an examples section laid out by the heading:

```
Examples
----------

```

or you could modify the module docstring as suggested by @PicoCentauri here: https://github.com/MDAnalysis/mdanalysis/blob/32de86f149561b2cea474d098b135c7ffdb5949f/package/MDAnalysis/analysis/helix_analysis.py#L54-L82

However, example 2 is misleading. Even though it's true that the selection is split, it's best to write documentation from the perspective of what a user cares about; they usually only care about how that affects the final result. The resulting analysis still has 3 helices because you've passed in 3 selection strings. Showing an example of `select='resid 6:50 or resid 100:130 or resid 132:148'` and `split_residue_sequences=True|False` would have been more informative.

IMO adding this as an example is optional, though.

lilyminium (changes requested):

package/MDAnalysis/analysis/helix_analysis.py
robotjellyzone: So, I should only add example - 1 then right? @lilyminium

robotjellyzone: > IMO adding this as an example is optional, though.

so does this means it's fine even if I don't add any example? shall I skip adding it? @lilyminium

lilyminium:
Yes let's just make that optional

robotjellyzone:
> Yes let's just make that optional

ok just adding other changes without  any examples

lilyminium (approved):

lilyminium:
Thank you @robotjellyzone! Just restarting the Azure tests which were failing.

Issue #3431, opened 2021-09-29
**kwargs in AnalysisBase does not trigger errors


lilyminium:
## Expected behavior ##

When I pass in keywords that don't exist to analysis classes, it fails.

## Actual behavior ##

`**kwargs` in classes like RMSD get passed to AnalysisBase, which then silently ignores them.

## Code to reproduce the behavior ##

``` python
R = rms.RMSD(ligand, ref_frame=0, superposition=False, center=False).run(start=0, step=25, stop=-1)
```

This is a very reasonable set of arguments to use, given that `superposition` and `center` are keywords in `rms.rmsd`. However, RMSD *actually* functions as `rms.rmsd(superposition=True, center=True)`.

Additional context here: https://groups.google.com/d/msgid/mdnalysis-discussion/a0b0d7db-079b-4a7d-8125-81bb5c90e8d5n%40googlegroups.com?utm_medium=email&utm_source=footer

IMO we could just remove `**kwargs` from the AnalysisBase function signature for an indirect trigger.

## Current version of MDAnalysis ##

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`)
- Which version of Python (`python -V`)?
- Which operating system?

orbeckst:
Note that in #1463 we decided that `verbose` would be a kwarg for both `__init__` and `run()`. So if we remove `**kwargs` (which sounds sensible) then we have to pass through `verbose=None` (or whatever the default ought to be) explicitly.

If we want to remove `verbose` from `__init__` then we can only do it for 3.0 because it would be a breaking API change.

PR #3644, opened 2022-04-16
Cleanup and restructure of AnalysisBase classes docstrings

### LINKED ISSUE #919: Where should base class API definition docs be written?
In #892 I noticed that we actually document the default API behavior of base classes like `Reader` and `AnalysisBase`. I actually never noticed this because the docs for the classes are in the `__init__.py` file and *not* where the class is defined.

This is an issue for me because when I change the API I will very likely forget to update the docs. For the very simple reason that it's not in the same file and I will only look for docs in files that actually define a class. I'm also sure I would overlook this in a PR.

In #892 I went a different way (because I didn't know any better) and used an example piece of code to show the API and an actual analysis can use `AnalysisBase`. Here the definition of the API lives together with the code. Something I like because it makes it harder for me to forget changing it.

So we have two methods for documentation right now

1. document Base Classes API in `__init__`
2. document Base Classes API in baseclass docstring.

Of course this is a objective issue and different people prefer different styles. I'm OK with both methods (I prefer 2 because I'm forgetful). I just want to know what other people think about this.

### LINKED ISSUE #3651: Deprecate `verbose` option in initialization of `AnalysisBase`

The verbose option for Analysis classes appears in the `init` function and as well in the  `run` method. The first one seems to be not necessary for instance creation.

Deprecate and remove `verbose` from the init method. I already have a suggestion in PR #3644 

Leave everything as it is.

## Additional context ##

Removing verbose from the init also would allow removing the **kwargs argument in many classes inheriting from the base class and closing #3431

PicoCentauri:
Fixes #919
Fixes #3651

* Deprecate `verbose` parameter in `AnalysisBase.__init__` (follows issue #3651)
* Moved all docstrings from `__init__` function to class docstring (follows issue #919)
* Removed doubled docs in `EinsteinMSD`
* Removed docs for non existing verbose parameter in `conformational_distance_matrix`
* Cleanup and sorted imports of all touched files

I also restructured the docs of the RMS classes a bit. If you disagree with anything please 
let me know. I checked the rendered docs and hope I did not oversaw any 
issues within the html pages. These are many changed lines, but I saw a lot of things 
while I created the decorator. Thanks for your feedback and your time to review!

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


lilyminium:
The decorator is interesting. I suppose it makes things more consistent, but it does make writing documentation a bit more annoying (having to check what string is actually replaced)... and you have the same problem where `verbose` is not in the documentation unless the author remembers to import and use it. @MDAnalysis/coredevs any thoughts? Tbh we could go all the way and write a metaclass that automatically appends base class arguments to the docs, although that sounds *really* complicated.

All the other docs look good to me, although the import sorting -- did you use `isort`? If we want to maintain alphabetical imports we could look into a more long-term solution like pre-commit hooks.

lilyminium:
Actually, do you know if the docstring decorator works well in IDEs and/or `help`?

PicoCentauri:
> The decorator is interesting. I suppose it makes things more consistent, but it does make writing documentation a bit more annoying (having to check what string is actually replaced)... and you have the same problem where verbose is not in the documentation unless the author remembers to import and use it. https://github.com/orgs/MDAnalysis/teams/coredevs any thoughts? Tbh we could go all the way and write a metaclass that automatically appends base class arguments to the docs, although that sounds really complicated.

Of course, authors have to remember but the decorator is consistent and we could add this to documentation for writing new classes based on the AnalysisBase class. A metaclass would be indeed optimal. But This would include parsing the docstring, which could be a bit painful as I learned from the cli.

> All the other docs look good to me, although the import sorting -- did you use isort? If we want to maintain alphabetical imports we could look into a more long-term solution like pre-commit hooks.

I used isort indeed. A pre-commit hook sounds like something very nice. The alphabetical order is a choice (not the only one) but gives us developers a guideline and prevents chaotic and vast import sections.

PicoCentauri:
> Actually, do you know if the docstring decorator works well in IDEs and/or `help`?

I works in `help` and with the `?` syntax. However at least vscode does not parse it 'correctly'.

jbarnoud:
While I appreciate the improved consistency, I am not sure I like the decorator as a solution. The AnalysisBase API to be adopted, needs to be as streamlined as possible. The issue I see with the decorator is that, on top of inheriting from the base class and overwriting the expected methods, a user needs to add a class decorator and the template placeholder. These two extra things are MDAnalysis-specific knowledge, which adds to the issue, I think.

What plays in favour of the solution, in my eyes, is that it may be considered for internal use only. Then, the users do not need to care. But it also means our analyses are a bit less useful as templates/examples for the users.

It makes me remember, didn't we discuss deprecating `verbose` in the `__init__` in favour of the one in `run`?

IAlibay:
So my 2 cents is that a) 90% of my docstring reading is happening when I'm reading the code itself, b) for the remaining 10%, 9% is happening in VSCode... Now I realise that both those use cases are probably "me specific", so I'm not going to say it's a blocker, but I would strongly support alternative solutions 😅 

> It makes me remember, didn't we discuss deprecating verbose in the `__init__` in favour of the one in run?

Yeah I thought that was the case too, can't find the relevant issue but I swear that was in the plans. Honestly I'd prefer we just get rid of it, `verbose` is somewhat useless for class construction...

jbarnoud:
> So my 2 cents is that a) 90% of my docstring reading is happening when I'm reading the code itself, b) for the remaining 10%, 9% is happening in VSCode...

That too!

PicoCentauri:
I just want to get rid off the 20 different phrases of describing a verbose option. I am happy with removing the verbose parameter completely for the `__init__` function. I also remember that there was a discussion about it somewhere @IAlibay  @jbarnoud 😅

I can deprecate the verbose parameter in the `AnalysisBase` and remove all verbose parameters from the classes if the @MDAnalysis/coredevs agree?

jbarnoud:
> I can deprecate the verbose parameter in the AnalysisBase and remove all verbose parameters from the classes

I would be up for it.

IAlibay:
> I just want to get rid off the 20 different phrases of describing a verbose option. I am happy with removing the verbose parameter completely for the `__init__` function. I also remember that there was a discussion about it somewhere @IAlibay @jbarnoud 😅
> 
> I can deprecate the verbose parameter in the `AnalysisBase` and remove all verbose parameters from the classes if the @MDAnalysis/coredevs agree?

I'd possibly bringing it up in a separate issue (given we all seem to have forgotten where the original issue was) and pinging the #developers channel on discord with the issue. Given how close we are to GSoC / Outreachy deadlines there's always a risk that the current discussion will get lost in the sea of notifications.

PicoCentauri:
I can create a new one but at least @orbeckst commented in #3431 that we could remove the `verbose` parameter which and maybe also remove the `**kwargs` argument...

jbarnoud:
We cannot remove kwargs here, otherwise third party code will break every time we change the signature.

jbarnoud:
I mean that for AnalysisBase. The other classes can loose the kwarg.

richardjgowers:
WRT removing verbose, I think we can have verbose in one place, but I think it'll have to be `__init__` else how will we see log stuff that happens during init?

IAlibay:
I can't think of any analysisbase method where verbosity in init is even a thing - my 2 cent is that verbose is within the realm of tqdm and the like i.e. run only. If you want to be informative on setup then logger.info might be where we want to go...

package/MDAnalysis/analysis/base.py
```diff
@@ -450,58 +461,3 @@ def __init__(self, trajectory=None, *args, **kwargs):
                                                *args, **kwargs)
 
     return WrapperClass
-
-
-def _filter_baseanalysis_kwargs(function, kwargs):
```
PicoCentauri, line 455: Due to the removal of the explicit verbose option, the tests of the _private_ method `_filter_baseanalysis_kwargs` failed. `_filter_baseanalysis_kwargs` splits and tests the keyword arguments of a function and the Base class. With the removal of the `verbose` (the only explicit keyword argument) `_filter_baseanalysis_kwargs` will fail because the `inspect.getargspec ` will return `None`. I see two options for the future of `_filter_baseanalysis_kwargs `

1. **Keeping** `_filter_baseanalysis_kwargs`: Test for `None` and return an empty dictionary and the unchanged `kwargs`.

I would go for the second option:

2. **Remove** `_filter_baseanalysis_kwargs`: With the removal of the only keyword argument of `AnalysisBase` `_filter_baseanalysis_kwargs` lost its purpose. There is no possible clash of the keyword arguments between a passed function and `AnalysisBase` anymore. Additionally, option 1 results in a lot of dead code which will only be executed if another explicit keyword argument of `AnalysisBase` will be introduced in the future.

package/MDAnalysis/analysis/align.py
```diff
@@ -663,9 +667,7 @@ def __init__(self, mobile, reference, select='all', filename=None,
 
         # do this after setting the memory reader to have a reference to the
         # right reader.
-        super(AlignTraj, self).__init__(mobile.trajectory, **kwargs)
-        if not self._verbose:
-            logging.disable(logging.WARN)
```
PicoCentauri, line 668: I had to move this to the `_prepare` method. Will this cause any issues?

package/MDAnalysis/analysis/align.py
```diff
@@ -864,8 +873,6 @@ def __init__(self, mobile, reference=None, select='all', filename=None,
         # do this after setting the memory reader to have a reference to the
         # right reader.
         super(AverageStructure, self).__init__(mobile.trajectory, **kwargs)
-        if not self._verbose:
-            logging.disable(logging.WARN)
```
PicoCentauri, line 868: Same question as above.

PicoCentauri:
I now removed the explicit `verbose` option from the module initialization of `AnalysisBase`. This brought a little cascade of issues I had to fix. The main questions are what to do with `_filter_baseanalysis_kwargs ` and is it okay to move the logging in the align classes?

_The is no rush with this PR and we can discuss this when there is more time after the deadlines of GSOC. I want to sort and document my ideas before I forget everything again._

package/MDAnalysis/analysis/align.py
orbeckst: Yes, because `get_matching_atoms()` uses logging and the output is important for debugging.

This is actually one of the cases that @richardjgowers talked about when he said "How would we know what happens in init?" (and where @IAlibay and myself said "Bah, humbug, nothing ever happens in init."

(Sorry for the *slight* misquote @IAlibay )

This is also an example where it makes sense that init does a stuff that you wouldn't necessarily want to do in `_prepare` every time. Nevertheless, it could be rewritten so that `_prepare` only does things once.

package/MDAnalysis/analysis/align.py
orbeckst: same answer ;-) — yes, here it matters for as long as `__init__` is doing work.

package/MDAnalysis/analysis/base.py
orbeckst: Getting rid of it would be good.

Add a note to AnalysisBase that the call signature `trajectory, **kwargs` is fixed and part of the API specification.

orbeckst (changes requested):
Overall this looks good to me. I am in favor of deprecating verbose in `AnalysisBase.__init__` (but I don't want to pre-empt an emerging consensus). I also like the cleanup of the imports and the consistent docs in the class definition, which we decided to do in #919. For the latter, can you add a note to the docs for AnalysisBase that doc strings go into the classdef and init remains bare (if it's not already there)?

The two cases where verbose is needed in `__init__` are indeed problematic. Along the lines of our discussion about `__init__` vs `__prepare`, one approach would be to move most of the code to `_prepare`, at the cost of having to either take a performance hit (not good) or adding code to only run once. Semantically, `__init__` is the better place because it's clear that the code is only run once. Opinions welcome...

orbeckst:
@PicoCentauri really stupid question: the current PR doesn't contain a verbosity decorator anymore, does it? It just deprecates the verbose kwarg in `AnalysisBase.__init__` and takes it from the `**kwargs`, right?

orbeckst:
If there's no decorator in there anymore then please update the PR comment and also reference issue #3651.

PicoCentauri:
> @PicoCentauri really stupid question: the current PR doesn't contain a verbosity decorator anymore, does it? It just deprecates the verbose kwarg in AnalysisBase.__init__ and takes it from the **kwargs, right?

Yes! I changed the description accordingly.

package/MDAnalysis/analysis/base.py
PicoCentauri: Good point. I added a note.

PR #3650, opened 2022-04-20
Remove None from filenames in analysis.hole.HoleAnalysis

mjtadema:
analysis.hole.HoleAnalysis keeps track of the filenames that contributed to a universe, however this breaks if any of the filenames is None. This can happen when a universe trajectory is setup using a list of coordinates (and probably any other case where the universe coordinates are set without a file, such as from a stream). I propose None be removed from the filenames list before joining.

 - Remove None from filenames in analysis.hole.HoleAnalysis

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?


lilyminium:
Thanks for opening this, @mjtadema. I just have a few questions about this:

* does having `None` break the code for you (trigger an error or other undesired behaviour)?
* where are the filenames used in the code -- just for writing output files?

Regarding the change you've made, have you considered what to do if the list does not have `None` in it?

mjtadema:
@lilyminium yes, the code breaks because it tries to join the list of filenames while join throws an exception when there are any objects that are not strings. 
From what I can see the only place where these particular filenames are used is to write the header file for hole (where they are even commented out and not used for anything besides telling the user which files were  used). I felt this would be a safe change that shouldn't negatively impact any other part of the code.

mjtadema:
> Regarding the change you've made, have you considered what to do if the list does not have `None` in it?

Ah yes my bad, I was a little hasty making the PR I suppose. I'll fix it.

mjtadema:
patch coverage probably fails because there is no test case where the universe trajectory was set without a filename. Probably the easiest fix for this would be to simply reuse one of the tests and set the filename to `None`?

lilyminium:
> Probably the easiest fix for this would be to simply reuse one of the tests and set the filename to None?

Yeah, writing a new test where the Universe has been modified to have `None` as a filename would seem like the best solution :)

package/MDAnalysis/analysis/hole2/hole.py
```diff
@@ -550,6 +550,8 @@ def __init__(self, universe,
             filenames.extend(universe.trajectory.filenames)
         except AttributeError:
             filenames.append(universe.trajectory.filename)
+        while None in filenames:
+            filenames.remove(None)
```
lilyminium: Would a list generator be more efficient here, like `filenames = [x for x in filenames if x is not None]` ?

mjtadema: yes, much more efficient :)

testsuite/MDAnalysisTests/analysis/test_hole2.py
```diff
@@ -294,6 +294,19 @@ def test_sph_process_oserror(self, universe):
...
+    def universe_none_filename(self):
+        u = mda.Universe(PDB_HOLE, in_memory=True)
+        u.trajectory.filename = None
+        return u
+
+    def test_none_filename(self, universe_none_filename, tmpdir):
+        with tmpdir.as_cwd():
+            with hole2.HoleAnalysis(universe_none_filename) as h:
+                h.run()
+
```
lilyminium: As this is just a single test of HoleAnalysis, could you please just keep it as a single function in the TestHoleAnalysis class? The `universe_none_filename` fixture is only used for this one test, so it could just get created within the function itself.

lilyminium (changes requested):
Is this also your first contribution to MDAnalysis? If so, please add yourself to AUTHORS and the CHANGELOG! There's no issue to reference but the title of this PR would work just fine in the bug fixes section, IMO.

testsuite/MDAnalysisTests/analysis/test_hole2.py
mjtadema: Yes you're right, that would be better

mjtadema:
> Is this also your first contribution to MDAnalysis? If so, please add yourself to AUTHORS and the CHANGELOG! There's no issue to reference but the title of this PR would work just fine in the bug fixes section, IMO.

I'll add it to CHANGELOG, but I am already in AUTHORS (right below you apparently ;) ).

lilyminium (approved):
Awesome, looks great @mjtadema! I just fixed a conflict with MDAnalysis that arose from another PR getting merged. Thanks for both finding the bug and fixing it!

Edit: Just waiting on CI.

lilyminium:
Hmm looks like a random error in Azure pipelines?

/azp run

PR #3653, opened 2022-04-22
Issue 2764 Setting attributes should have dtype checks

fortune-max:
---------------------------------------
 - Enforced strict adherence to the data types for `TopologyAttr` values

This pull request attempts to address issue #2764 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


jbarnoud:
Hi @fortune-max. Thank you for opening this pull request.

It is difficult to review, though, because it lacks some context. Can you edit your pull request message to follow the template, please? Especially, we need you to reference the issue you are fixing. This is important for mainly 2 reasons: 1) it lets us know what your pull request is about, and 2) it will automatically close the issue when the pull request is merged.

Something else I noticed is that you copied the file `testsuite/MDAnalysisTests/core/test_topologyattrs.py` into `package/MDAnalysis/core/test_topologyattrs.py`. Tests should stay together in the `testsuite` directory; the `package` directory contains the library itself. Also, because you added your changes into a copy of the original file, we do not get to see what you changed as everything appears as new.

Could you please fix the two points above? Without that, we will not be able to proceed with the code review.

fortune-max:
Thanks for your response @jbarnoud. I will fix the issues right away

PR #3662, opened 2022-05-11
Issue 3659 fixed loading for MOL2 and PDB edge cases

### LINKED ISSUE #3659: Inconsistent resnames when loading pdb and mol2 with repeat resnums
## Expected behavior ##

I am using a PDB file that appears to be prepared incorrectly and includes residue number 200 for both the ligand and a protein residue. When loading the PDB, it appears as if MDA retains the original ligand resname (CNA). I need the fragment functionality so i have prepared a MOL2 file for this system which has the same numbering error. I expect loading this MOL2 file would have the same behavior for the ligand resname as the PDB.

## Actual behavior ##

When I load the MOL2 file the resname for the ligand is replaced by the resname for the 200th protein residue (TRP200).

u = mda.Universe('1a26A.pdb')
u.atoms.resnames

u = mda.Universe('1a26A.mol2')
u.atoms.resnames

Version 2.1.0 on both MacOS and Linux.

[1a26.zip](https://github.com/MDAnalysis/mdanalysis/files/8635956/1a26.zip)

zwsmith200:
Fixes #3659 

 - New squash function `squash_by_attributes` that uses unique combinations of attributes to determine residues
 - `MOL2Parser` and `PDBParser` now use this function to avoid issues when loading non-unique resids and non-contiguous residues  (#3659)

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/topology/base.py
+++ b/package/MDAnalysis/topology/base.py
@@ -37,6 +37,7 @@ Classes
 
 """
 from functools import reduce
+from typing import List
 
 import itertools
 import numpy as np
@@ -142,27 +143,33 @@ def squash_by(child_parent_ids, *attributes):
     return atom_idx, unique_resids, [attr[sort_mask] for attr in attributes]
 
 
-def squash_by_attributes(squash_attributes, *other_attributes):
+def squash_by_attributes(
+    squash_attributes: List[np.ndarray], *other_attributes: np.ndarray
+):
     """Squash a child-parent relationship using combinations of attributes
     parents will retain their order of appearance
 
-    Arguments
-    ---------
-    squash_attributes - list of attribute arrays (attributes used to
-                        identify the parent)
-    *other_attributes - other arrays that need to follow the sorting of ids
+    Parameters
+    ----------
+    squash_attributes: list of numpy ndarray
+      list of attribute arrays (attributes used to identify the parent)
+      Their unique combinations will set child-parent relationships
+    *other_attributes: numpy ndarrays
+      other arrays that need to follow the sorting of ids
 
     Returns
     -------
-    child_parents_idx - an array of len(child) which points to the index of
-                        parent
-    parent_combos - len(parent) of the unique combinations
-    squashed_attrs - len(parent) of the attributes used for squashing
-    *other_attrs - len(parent) of the other attributes
+    sorted_atom_idx: numpy ndarray
+      an array of len(child) which points to the index of the parent
+    squashed_attrs: list of numpy ndarray
+      List of re-ordered squash_attributes
+    *other_attrs: list of numpy ndarray
+      List of re-ordered other_attributes
     """
     squash_array = np.column_stack(squash_attributes).astype(str)
     unique_combos, sort_mask, atom_idx = np.unique(
-        squash_array, return_index=True, return_inverse=True, axis=0)
+        squash_array, return_index=True, return_inverse=True, axis=0
+    )
 
     appearance_order = np.argsort(sort_mask)
     new_index = np.arange(0, len(appearance_order))
--- a/package/MDAnalysis/topology/base.py
+++ b/package/MDAnalysis/topology/base.py
@@ -142,6 +142,38 @@ def squash_by(child_parent_ids, *attributes):
     return atom_idx, unique_resids, [attr[sort_mask] for attr in attributes]
 
 
+def squash_by_attributes(squash_attributes, *other_attributes):
+    """Squash a child-parent relationship using combinations of attributes
+    parents will retain their order of appearance 
+
+    Arguments
+    ---------
+    squash_attributes - list of attribute arrays (attributes used to identify the parent)
+    *other_attributes - other arrays that need to follow the sorting of ids
+
+    Returns
+    -------
+    child_parents_idx - an array of len(child) which points to the index of parent
+    parent_combos - len(parent) of the unique combinations
+    squashed_attrs - len(parent) of the attributes used for squashing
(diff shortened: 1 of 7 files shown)


orbeckst:
Welcome to contributing to MDAnalysis! Thanks for the PR.

Sorry, I don't have time to review, I just wanted to ask if this PR is going to fix #3659 ? If so, please enter the issue number in the `Fixes #` line at the top of your issue report so that PRs and issues get linked.

IAlibay:
<s>So I should ask here - the fact that both squash_by and change_squash exist just next to each other makes me wonder if there is a file format specific reason as to why these behaviours are different :/ Is there anything in the PR history that explains why these exist?</s>

Nevermind, I better understand the issue now.

zwsmith200:
The scope has expanded slightly since the initial issue. Originally I was concerned that there were different behaviours when reading PDB and MOL2, namely the MOL2 format deleted the resname with a repeated resid while PDB did not. I then went to re-use `change_squash` from PDB and realized this will double the number of residues if they are not contiguous (for example openbabel adds hydrogens to the end of PDBs).

In light of those findings, I wrote a squash function that is robust to both repeated resids with different attributes and non-contiguous residues. I have applied this change to the MOL2 and PDB parsers but I suspect experts in other formats may want to take a look at whether similar unit tests are needed.

richardjgowers:
I need to double check this, hopefully in the next few days

package/MDAnalysis/topology/base.py
```diff
@@ -142,6 +142,42 @@ def squash_by(child_parent_ids, *attributes):
...
+                        identify the parent)
+    *other_attributes - other arrays that need to follow the sorting of ids
+
+    Returns
+    -------
+    child_parents_idx - an array of len(child) which points to the index of
+                        parent
+    parent_combos - len(parent) of the unique combinations
+    squashed_attrs - len(parent) of the attributes used for squashing
+    *other_attrs - len(parent) of the other attributes
```
richardjgowers: Can this follow the doc style a bit better?  Usually it's type annotation on the same line as the variable name and description on the following line

richardjgowers (changes requested):
@zwsmith200 thanks for putting this together.  I think it's an improvement for a lot of cases, but I can foresee cases where it changes behaviour, e.g. if the resids wrapped around the max value and the resname happened to be the same by bad luck then this would be considered the same residue.  I think it might be better to allow this new residue grouping strategy via a keyword argument to Universe, which is passed to the parser.

package/MDAnalysis/topology/base.py
zwsmith200: This is a good point, I have actually run into an example of this using mol2 generated from pdb in my work because it lost chain information. I’ll add a test for wrapped resids.

IAlibay:
@zwsmith200 is this still something you are looking to contribute?

zwsmith200:
Yes, I can finish cleaning this up. I got it working for my now-completed project then forgot about the PR.

zwsmith200:
> Hello @zwsmith200! Thanks for updating this PR. We checked the lines you've touched for [PEP 8](https://www.python.org/dev/peps/pep-0008) issues, and found:
> 
> * In the file [`package/MDAnalysis/topology/PDBParser.py`](https://github.com/MDAnalysis/mdanalysis/blob/1c997ef047613870491078924046b1fa98d7e50e/package/MDAnalysis/topology/PDBParser.py):
> 
> > [Line 409:80](https://github.com/MDAnalysis/mdanalysis/blob/1c997ef047613870491078924046b1fa98d7e50e/package/MDAnalysis/topology/PDBParser.py#L409): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (80 > 79 characters)
> > [Line 418:80](https://github.com/MDAnalysis/mdanalysis/blob/1c997ef047613870491078924046b1fa98d7e50e/package/MDAnalysis/topology/PDBParser.py#L418): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (81 > 79 characters)
> 
> * In the file [`testsuite/MDAnalysisTests/topology/test_mol2.py`](https://github.com/MDAnalysis/mdanalysis/blob/1c997ef047613870491078924046b1fa98d7e50e/testsuite/MDAnalysisTests/topology/test_mol2.py):
> 
> > [Line 325:80](https://github.com/MDAnalysis/mdanalysis/blob/1c997ef047613870491078924046b1fa98d7e50e/testsuite/MDAnalysisTests/topology/test_mol2.py#L325): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (86 > 79 characters)
> 
> * In the file [`testsuite/MDAnalysisTests/topology/test_pdb.py`](https://github.com/MDAnalysis/mdanalysis/blob/1c997ef047613870491078924046b1fa98d7e50e/testsuite/MDAnalysisTests/topology/test_pdb.py):
> 
> > [Line 388:80](https://github.com/MDAnalysis/mdanalysis/blob/1c997ef047613870491078924046b1fa98d7e50e/testsuite/MDAnalysisTests/topology/test_pdb.py#L388): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (87 > 79 characters)
> 
> ##### Comment last updated at 2023-11-06 01:06:50 UTC

I realize darker used max length of 88 and we use 79. It may be useful to add -l 79 [here](https://userguide.mdanalysis.org/stable/contributing_code.html#format-darker) so that copying the command matches our requirements.

zwsmith200:
@richardjgowers @IAlibay It looks like I have addressed everything I needed to, let me know if I missed something.

PR #3665, opened 2022-05-13
[proposal / MAINT] Update to python>=3.8, numpy>=1.9.0, gsd>=1.9.3, networkx>=2.0, and scipy>=1.5.0

IAlibay:
- More closely follow NEP29, scipy bump is to match ~ release of numpy 1.9.0.
- GSD 1.9.3 is now rather old, we can make it the minimum version so as not to have a difference between *nix/macOS and windows
- As requested by @RMeli networkx 1.0 is very old and so is 2.0. We should at the very least set the minimum version at 2.0.
- Various maintenance changes to track the new dependency versions we use

PR Checklist
------------
 - Tests?
 - Docs?
 - [x] CHANGELOG updated?
 - Issue raised/referenced?

### DIFF
--- a/.github/workflows/gh-ci.yaml
+++ b/.github/workflows/gh-ci.yaml
@@ -28,7 +28,7 @@ jobs:
         fail-fast: false
         matrix:
           os: [ubuntu-latest, ]
-          python-version: [3.8, 3.9, "3.10"]
+          python-version: [3.8, 3.9]
           full-deps: [true, ]
           install_hole: [true, ]
           codecov: [true, ]
--- a/.github/workflows/gh-ci.yaml
+++ b/.github/workflows/gh-ci.yaml
@@ -45,7 +45,7 @@ jobs:
               full-deps: true
               install_hole: true
               codecov: true
-            - name: macOS_catalina_py37
+            - name: macOS_catalina_py38
               os: macOS-10.15
               python-version: 3.8
               full-deps: true
--- a/.github/workflows/gh-ci.yaml
+++ b/.github/workflows/gh-ci.yaml
@@ -28,7 +28,7 @@ jobs:
         fail-fast: false
         matrix:
           os: [ubuntu-latest, ]
-          python-version: [3.8, 3.9]
+          python-version: [3.8, 3.9, "3.10"]
           full-deps: [true, ]
           install_hole: [true, ]
           codecov: [true, ]
--- a/.github/workflows/gh-ci.yaml
+++ b/.github/workflows/gh-ci.yaml
@@ -28,7 +28,7 @@ jobs:
         fail-fast: false
         matrix:
           os: [ubuntu-latest, ]
-          python-version: [3.7, 3.8, 3.9]
+          python-version: [3.8, 3.9]
           full-deps: [true, ]
           install_hole: [true, ]
           codecov: [true, ]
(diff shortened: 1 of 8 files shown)


package/pyproject.toml
```diff
@@ -5,8 +5,7 @@ requires = [
     "packaging",
     # lowest NumPy we can use for a given Python,
     # except for more exotic platforms (linux/Mac ARM flavors)
-    "numpy==1.18; python_version=='3.7' and (platform_machine!='arm64' or platform_system!='Darwin') and platform_machine!='aarch64'",
-    "numpy==1.18; python_version=='3.8' and (platform_machine!='arm64' or platform_system!='Darwin') and platform_machine!='aarch64'",
+    "numpy==1.19; python_version=='3.8' and (platform_machine!='arm64' or platform_system!='Darwin') and platform_machine!='aarch64'",
```
tylerjereddy, line 8: Any reason to bump the minimum NumPy version for `3.8`? Maybe not a big deal, though we're drifting even farther from https://github.com/scipy/oldest-supported-numpy/blob/main/setup.cfg#L53

IAlibay: NEP29, if we only support 1.19 then the minimum version for build should be that version right?

package/CHANGELOG
```diff
@@ -53,6 +53,8 @@ Fixes
   * Fixed BAT method Cartesian modifies input data. (Issue #3501)
 
 Enhancements
+  * The minimm Python, NumPy, and SciPy versions are now 3.8, 1.19, and 1.5.0,
+    GSD now also have a minimum version of 1.9.3 (Issue #3665)
```
orbeckst: Should go under Changes

package/pyproject.toml
orbeckst: In support of @IAlibay :
```
$ nep29 numpy
| version |    date    |
|---------|------------|
|  1.22.0 | 2021-12-31 |
|  1.21.0 | 2021-06-22 |
|  1.20.0 | 2021-01-30 |
|  1.19.0 | 2020-06-20 |
```

.github/workflows/gh-ci.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
         fail-fast: false
         matrix:
           os: [ubuntu-latest, ]
-          python-version: [3.7, 3.8, 3.9]
+          python-version: [3.8, 3.9]
```
orbeckst, line 31: Are we still not able to run 3.10 just normally?

orbeckst (approved):
One minor thing in CHANGELOG but otherwise I agree with your implementation of NEP29.

I'll approve so that you can move aiong.

package/CHANGELOG
```diff
@@ -71,6 +71,9 @@ Enhancements
     keyword (Issue #3605)
 
 Changes
+  * The minimm Python, NumPy, and SciPy versions (following NEP29) are now 3.8,
```
orbeckst, line 74: I edited CHANGELOG to address my comment...

IAlibay: Thanks!

.github/workflows/gh-ci.yaml
IAlibay: Last I checked we were waiting for rdkit and openmm, I know that rdkit added py310 support in 2022.03.2 ~ 20 days ago.

I'm not sure if about openmm 7.7.0, I know it got added 2 months ago but I know they then pulled some stuff. Let's try it out!

IAlibay: No good, looks like chemfiles-python is the last culprit here, I'll raise an issue upstream.

package/pyproject.toml
tylerjereddy: I guess I'm accustomed to less rigid adherence--I tend to follow the NEP29 Python schedule but maximize the NumPy support surface to prevent annoying stack rebuilds. Maybe less of a concern for MDA though, and if you can cite the NEP to avoid maintenance burden seems like a good idea.

.github/workflows/gh-ci.yaml
richardjgowers: isn't chemfiles optional? (for reasons like this...)

richardjgowers (approved):

.github/workflows/gh-ci.yaml
IAlibay: Yeah this is why the py310  minimum deps tests are fine. That matrix is the "all the deps" matrix (since it's easier to run this)

package/pyproject.toml
IAlibay: I general do favour your approach @tylerjereddy - however in this case I think bumping up to 1.19 is going to be useful to just be able to match conda-forge a little bit. If long term we'll just match whatever their numpy version defaults are for conda releases then it's best if we don't lag behind too much (since conda is the main way folks appear to install MDA).

IAlibay:
Thanks for all the input here folks, I'm going to go ahead and merge so we can speed ahead to 2.2.0.

PR #3667, opened 2022-05-15
replacing usage of lib.util.which and adding decorator

### LINKED ISSUE #3649: Replace custom lib.util.which with shutil.which
Given the newest version of MDAnalysis support Python 3.6+, is it reasonable to replace `lib.util.which` with [`shutil.which`](https://docs.python.org/3/library/shutil.html)? They seems to have identical intended uses and functionality. Using the standard library instead of custom code reduces the maintenance burden.

If this is a reasonable/desired change, I'll make a PR performing the replacement.

I realize this is a small nitpick, but I found it while reading the codebase and would like to take this opportunity to become a contributor to this project.

aditi2906:
Fixes #3649 

 - Replaced usage of lib.util.which with shutil.which 
 - added decroator warnings 

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [ ] CHANGELOG updated?


package/MDAnalysis/analysis/hole2/utils.py
```diff
@@ -514,16 +515,16 @@ def create_vmd_surface(sphpdb='hole.sph',
...
         raise OSError(errno.ENOENT, exe_err.format(name=sph_process,
                                                    kw='sph_process'))
     base_path = os.path.dirname(sph_process_path)
 
-    sos_triangle_path = util.which(sos_triangle)
+    sos_triangle_path = shutil.which(sos_triangle)
     if sos_triangle_path is None:
         path = os.path.join(base_path, sos_triangle)
-        sos_triangle_path = util.which(path)
+        sos_triangle_path = shutil.which(path)
```
IAlibay: 🤔 I thought there was a test for this already... @aditi2906 can you add one please?

package/MDAnalysis/analysis/legacy/x3dna.py
```diff
@@ -140,7 +140,7 @@
 import matplotlib.pyplot as plt
 
 from MDAnalysis import ApplicationError
-from MDAnalysis.lib.util import which, realpath, asiterable
+from MDAnalysis.lib.util import realpath, asiterable
```
IAlibay: Legacy stuff is fine being untested so this is all good.

package/MDAnalysis/lib/util.py
```diff
@@ -534,39 +535,6 @@ def isstream(obj):
     return np.any(alternative_results)
 
 
-def which(program):
```
IAlibay, line 538: So the idea here is not to remove `which` but deprecate it, since there might be downstream stuff that still uses it.

package/MDAnalysis/lib/util.py
```diff
@@ -185,6 +185,7 @@
    underlying stream and ``NamedStream.close(force=True)`` will also
    close it.
 """
+import shutil
```
IAlibay: This isn't necessary (see below).

package/MDAnalysis/lib/util.py
```diff
@@ -2417,3 +2385,26 @@ def check_box(box):
     if np.all(box[3:] == 90.):
         return 'ortho', box[:3]
     return 'tri_vecs', triclinic_vectors(box)
+
+
+@deprecate(release="2.2.0", remove="3.0.0",
+           message="lib.util.which to be remove from 3.0.0 release")
+def which(program):
```
IAlibay, line 2394: no we don't want to re-implement a new version of which, in the future folks will just use `shutil.which`. The deprecation needs to go on the old version of `lib.util.which`.

package/MDAnalysis/lib/util.py
IAlibay: Also once you re-instate this code & add the deprecation decorator, can you add a tests that checks the deprecation warnings please?

IAlibay (changes requested):

package/MDAnalysis/lib/util.py
aditi2906: > So the idea here is not to remove `which` but deprecate it, since there might be downstream stuff that still uses it.

As per @orbeckst's comment [url](https://github.com/MDAnalysis/mdanalysis/issues/3649#issuecomment-1126864675), the function body of which was supposed to contain a function call to shutil.which.

package/MDAnalysis/analysis/hole2/utils.py
aditi2906: Sure...I will add the test and update the PR

package/MDAnalysis/lib/util.py
IAlibay: No by that he means that the calls to `utils.which` in other parts of the code need to be switched to `shutil.which`, not the existing `utils.which` method

package/MDAnalysis/analysis/hole2/utils.py
aditi2906: Do I need to write the test for create_vmd_surface function ?

package/MDAnalysis/lib/util.py
aditi2906: I moved the "which" function at EOF because the deprecate function was declared in the same file below the existing "which" function declaration. So it was throwing an "undefined error".

package/MDAnalysis/analysis/hole2/utils.py
aditi2906: @IAlibay I added the required test for create_vmd_surface.

package/MDAnalysis/lib/util.py
aditi2906: @IAlibay explained below.

aditi2906: @IAlibay explained below.

package/MDAnalysis/lib/util.py
orbeckst: @IAlibay I think we should remove the old code and replace it with something like
```python
# TODO: Remove in 3.0.0
@deprecate(release="2.2.0", remove="3.0.0",
           message="lib.util.which() will be removed in 3.0.0; use shutil.which() instead")
def which(program):
   """Determine full path of executable `program` on :envvar:`PATH`."""
   return shutil.which(program)
```

testsuite/MDAnalysisTests/analysis/test_hole2.py
```diff
@@ -467,6 +469,21 @@ def test_none_filename(self, tmpdir):
...
+        errmsg = exe_err.format(name='dummy_path', kw='sos_triangle')
+        with pytest.raises(OSError, match=errmsg):
+            create_vmd_surface(sos_triangle='dummy_path')
+        errmsg = 'sph_process failed'
+        with pytest.raises(OSError, match=errmsg):
+            create_vmd_surface()
+        with tmpdir.as_cwd():
+            h = hole2.HoleAnalysis(universe)
+            h.run()
+            h.create_vmd_surface()
```
orbeckst, line 494: Why are these tests necessary? Aren't they already covered in `TestOSError` above?

orbeckst (changes requested):
Looks generally good but 
1. see comments
2. add entry to CHANGES

package/MDAnalysis/lib/util.py
IAlibay: I'm not sure I agree, in my opinion if you're doing a deprecation like this you're better off keeping the old code in case there's some very specific difference between the two `which` implementations that somehow someone relies on.

Also going with a change of the underlying method just for the sake of removing it later also just seems like unnecessary line changes. If the old one is unwanted why not just add the deprecation decorator to it and leave it? Are there any benefits here which I'm missing?

orbeckst: My thinking was that we can remove code and as long as it still passes the tests, it provides what we promised.

You're, of course, right that there could be subtle differences. In this particular case, I doubt anyone cares for such a utility function. But sure, we can just keep the old one, your call.

IAlibay: So the call I'm going to make here is that whatever we do we shouldn't just move the `which` method to the bottom of the file. The line diff here is going to be confusing if someone comes along and tries to work out the history of this file.

I have a preference for just keeping the old method for now but that's mostly for the sake of simplicity. That being said, I don't really mind if we go @orbeckst's way, I'm happy to approve either as long as we can follow the above requirement.

IAlibay: Note: if it's a case of deprecate being defined within the same file, just adding the warning manually would be a preferred option, that's fewer line changes.

orbeckst: Let's just keep the old code for `lib.util.which()` and _not_ replace it as I suggested in https://github.com/MDAnalysis/mdanalysis/pull/3667#discussion_r878671837.

@IAlibay has a strong opinion on not rearranging the file so `which` needs to go back to its original place and the deprecation message added manually.

orbeckst:
@aditi2906 will you be able to complete the PR? There's not much else to do as far as I can tell.

IAlibay:
Closing this PR as stale, please do re-open if you still wish to contribute!

Issue #3534, opened 2022-02-14
GSD: Deprecation warning (read_frame?)


IAlibay:
Running test_pickleio.py::test_GSD_pickle throws the following DeprecationWarning:

```
site-packages/gsd/hoomd.py:844: DeprecationWarning: Deprecated, trajectory[idx]
```

Looks like it's specifically down to the `gsd_io.read_frame` call on line 167.

MDA Version: 2.1.0-dev0

HenokB:
Hi, @IAlibay could you explain to me about this issue?

richardjgowers:
Hi Henok, theres a deprecation warning from the gsd library which has been
tracked down to a certain line. It’s likely we have to change how we’re
using the gsd library to avoid a future problem. To fix this issue,
investigate how/if the gsd library has changed and modify our code
accordingly.

PR #3669, opened 2022-05-16
Fix deprecation warnings from Issue #3535

### LINKED ISSUE #3535: DeprecationWarnings: numpy - elementwise comparison, calling nonzero on 0d arrays, `np.object`.

IAlibay:
Fixes #3535 #3534

 - fix GSD read_frames being deprecated (just index)
 - assert on size rather than comparing arrays for empty residuegroups in test_topologyattr
   - Note: we can probably just do an assert on the residuegroups themselves (which uses np.array_equal), but doing this probably avoids us kicking the bucket down the road.

PR Checklist
------------
 - [x] Tests?
 - Docs?
 - CHANGELOG updated?

### DIFF
--- a/testsuite/MDAnalysisTests/core/test_topologyattrs.py
+++ b/testsuite/MDAnalysisTests/core/test_topologyattrs.py
@@ -185,12 +185,17 @@ class TestAtomnames(TestAtomAttr):
         return mda.Universe(PSF, DCD)
 
     def test_prev_emptyresidue(self, u):
-        assert_equal(u.residues[[]]._get_prev_residues_by_resid(),
-                     u.residues[[]])
+        # Checking group size rather than doing
+        # array comparison on zero-sized arrays (Issue #3535)
+        groupsize = len(u.residues[[]]._get_prev_residues_by_resid().atoms)
+        assert groupsize == 0
+        assert groupsize == len(u.residues[[]].atoms)
 
     def test_next_emptyresidue(self, u):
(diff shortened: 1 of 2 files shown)


IAlibay:
(cycling to see if I can get rid of the HTTP 500 errors on accessing the runners)

richardjgowers (approved):
LGTM, checked we don't also do this inside the real reader and we're good

PR #3670, opened 2022-05-19
Fix reading/writing CONECT records in PDB files with many atoms

### LINKED ISSUE #988: Problem with reading/writing CONECT records in PDB files with many atoms
### Expected behaviour

Reading in a pdb file using the standard `MDAnalysis.Universe("file_name.pdb")`
### Actual behaviour

Fails with the following error message:

```
  File "./run_simulation.py", line 19, in <module>
    md_simulation.use_packmol()
  File "/sansom/s118/shil4444/Documents/Influenza_model/Input_files/md_simulation.py", line 405, in use_packmol
    shapes.add_rpo(ip)
  File "/sansom/s118/shil4444/Documents/Influenza_model/Input_files/shapes.py", line 462, in add_rpo
    u_vesicle = MDAnalysis.Universe(ip["filename"]+".pdb")
  File "/sansom/s118/shil4444/.local/lib/python2.7/site-packages/MDAnalysis-0.15.1.dev0-py2.7-linux-x86_64.egg/MDAnalysis/core/AtomGroup.py", line 4339, in __init__
    self._topology = p.parse()
  File "/sansom/s118/shil4444/.local/lib/python2.7/site-packages/MDAnalysis-0.15.1.dev0-py2.7-linux-x86_64.egg/MDAnalysis/topology/PDBParser.py", line 90, in parse
    bonds = self._parsebonds(atoms)
  File "/sansom/s118/shil4444/.local/lib/python2.7/site-packages/MDAnalysis-0.15.1.dev0-py2.7-linux-x86_64.egg/MDAnalysis/topology/PDBParser.py", line 181, in _parsebonds
    atom, atoms = _parse_conect(line.strip())
  File "/sansom/s118/shil4444/.local/lib/python2.7/site-packages/MDAnalysis-0.15.1.dev0-py2.7-linux-x86_64.egg/MDAnalysis/topology/PDBParser.py", line 215, in _parse_conect
    "record: {}".format(conect))
RuntimeError: Bond atoms aren't aligned proberly for CONECT record: CONECT100001100000100002100007
```

``` python
import MDAnalysis as mda

u = mda.Universe("Flu_membrane_capsule.pdb")
```

[Flu_membrane_capsule.txt](https://github.com/MDAnalysis/mdanalysis/files/474513/Flu_membrane_capsule.txt)

0.15.1-dev0

yuxuanzhuang:
Fixes #988

 - Raise `UserWarning` instead `RuntimeError` when PDB CONECT records are corrupt. The CONECT is ignored then.
 - Do not write unusable PDB CONECT records when n_atoms>100000.
(there might be corner cases when n_atoms is larger than 100000, but bonded information only spans before that?)

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


richardjgowers (changes requested):
PDBWriter will use the `.id` attribute of an atom (this is different to `.ix` which is *our* 0-based index of the atom).  So a way to trigger the 100,000+ issue is to change the `.ids` attribute to something large, rather than construct a huge AtomGroup

testsuite/MDAnalysisTests/coordinates/test_pdb.py
```diff
@@ -748,6 +752,23 @@ def test_write_bonds_partial(tmpdir):
...
+    u = mda.Universe(CONECT)
+
+    # construct a AtomGroup with more than 100000 atoms
+    ag = u.atoms
+    for i in range(52):
+        ag += u.atoms
+
+    if ag.n_atoms < 100000:
+        raise ValueError("test AtomGroup does not have"
+                         "more than 100000 atoms")
```
richardjgowers: This is a bit quis custodiet ipsos custodes?  Tests don't have checks inside, they should just work.

yuxuanzhuang:
> PDBWriter will use the `.id` attribute of an atom (this is different to `.ix` which is _our_ 0-based index of the atom). So a way to trigger the 100,000+ issue is to change the `.ids` attribute to something large, rather than construct a huge AtomGroup

I think PDBWriter will do a reindex (by default) first, so `.id` will be discarded.

Besides, I am only checking if the to-be-saved `AtomGroup` has more than 100,000 atoms instead of checking the `.id` of each atom. There definitely will be corner cases when e.g. a user tries to save an `Atomgroup` with less than 100,000 atoms but with some atoms has an `.id` of more than 100,000 while retaining its index by `ag.write('test.pdb', reindex=False)`.

Maybe add another line checking if the `.id` of every atom that has bond info is less than 100,000 when `reindex=False`?

The safest option would be to check each entry of the written CONECT, if one of them fails, discard all of them---which in my mind, is way too unnecessary.

testsuite/MDAnalysisTests/coordinates/test_pdb.py
yuxuanzhuang: I was thinking maybe one day the `CONECT` file changes and this test will fail---I will delete this check :->

richardjgowers (changes requested):
Ok so it sounds like there needs to be a check on the actual values going into each CONECT records rather than using heuristics based on the number of atoms.

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -851,6 +854,11 @@ def _write_pdb_bonds(self):
             con[a2].append(a1)
             con[a1].append(a2)
 
+        if not all(ag_ind <= 100000 for ag_ind in con.keys()):
```
yuxuanzhuang: It ensures every entry of CONECT should be valid.

richardjgowers: rather than iterating through the dictionary again, you could do this check as you populate the dictionary

yuxuanzhuang: Would you suggest move it even earlier? The majority of time spend here is constructing the `bondset`.
for a system with # Bond: 430447 # Atom: 579699
```python
%lprun -f mda.coordinates.PDB.PDBWriter._write_pdb_bonds u.atoms.write('test.pdb')
Total time: 131.148 s
File: /home/scottzhuang/mdanalysis/package/MDAnalysis/coordinates/PDB.py
Function: _write_pdb_bonds at line 808

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   808                                               def _write_pdb_bonds(self):
   809                                                   """Writes out all the bond records"""
   810         1          3.0      3.0      0.0          if self.bonds is None:
   811                                                       return
   812                                           
   813         3         11.0      3.7      0.0          if (not hasattr(self, "obj") or
   814         1          2.0      2.0      0.0                  not self.obj or
   815         1    3598068.0 3598068.0      2.7                  not hasattr(self.obj.universe, 'bonds')):
   816                                                       return
   817                                           
   818         1  127232750.0 127232750.0     97.0          bondset = set(itertools.chain(*(a.bonds for a in self.obj.atoms)))
   819         1          3.0      3.0      0.0          if self._reindex:
   820         1          1.0      1.0      0.0              index_attribute = 'index'
   821         2     308498.0 154249.0      0.2              mapping = {
   822                                                           index: i
   823         1         21.0     21.0      0.0                  for i, index in enumerate(self.obj.atoms.indices, start=1)
   824                                                       }
   825         1       5797.0   5797.0      0.0              atoms = np.sort(self.obj.atoms.indices)
   826                                                   else:
   827                                                       index_attribute = 'id'
   828                                                       mapping = {id_: id_ for id_ in self.obj.atoms.ids}
   829                                                       atoms = np.sort(self.obj.atoms.ids)
   830         1          3.0      3.0      0.0          if self.bonds == "conect":
   831                                                       # Write out only the bonds that were defined in CONECT records
   832         2          6.0      3.0      0.0              bonds = (
   833                                                           (
   834                                                               getattr(bond[0], index_attribute),
   835                                                               getattr(bond[1], index_attribute),
   836                                                           )
   837         1          1.0      1.0      0.0                  for bond in bondset if not bond.is_guessed
   838                                                       )
   839                                                   elif self.bonds == "all":
   840                                                       bonds = (
   841                                                           (
   842                                                               getattr(bond[0], index_attribute),
   843                                                               getattr(bond[1], index_attribute),
   844                                                           )
   845                                                           for bond in bondset
   846                                                       )
   847                                                   else:
   848                                                       raise ValueError("bonds has to be either None, 'conect' or 'all'")
   849                                           
   850         1          5.0      5.0      0.0          con = collections.defaultdict(list)
   851         2       2718.0   1359.0      0.0          for a1, a2 in bonds:
   852         2          5.0      2.5      0.0              if not (a1 in mapping and a2 in mapping):
   853                                                           continue
   854         2          4.0      2.0      0.0              if mapping[a1] >= 100000 or mapping[a2] >= 100000:
   855         1        130.0    130.0      0.0                  warnings.warn("Atom with index >=100000 cannot write "
   856                                                                         "bonds to PDB CONECT records.")
   857         1          1.0      1.0      0.0                  return
   858         1          3.0      3.0      0.0              con[a2].append(a1)
   859         1          1.0      1.0      0.0              con[a1].append(a2)
   860                                           
   861                                                   conect = ([mapping[a]] + sorted([mapping[at] for at in con[a]])
   862                                                             for a in atoms if a in con)
   863                                           
   864                                                   for c in conect:
   865                                                       self.CONECT(c)
```

richardjgowers: Ouch yeah that's 2 minutes making that set.  @hmacdope an example of where bond performance is terrible currently.

package/MDAnalysis/topology/PDBParser.py
```diff
@@ -203,6 +206,9 @@ def parse(self, **kwargs):
         except AttributeError:
             warnings.warn("Invalid atom serials were present, "
                           "bonds will not be parsed")
+        except RuntimeError:
+            warnings.warn("CONECT records was corrupt, "
+                          "bonds will not be parsed")
```
richardjgowers, line 216: Not sure I see where the RunTime error gets generated?

yuxuanzhuang: inside this function [def _parse_conect(conect):](https://github.com/MDAnalysis/mdanalysis/blob/c01d82bb86b480c8b2ae87f36497e7fc0a726c84/package/MDAnalysis/topology/PDBParser.py#L414)

package/MDAnalysis/coordinates/PDB.py
hmacdope: Roger roger thanks for flagging. Good to have as a bench/test case.

yuxuanzhuang: @hmacdope  If you are interested. https://github.com/yuxuanzhuang/MDA_benchmark_sys

hmacdope: Absolutely legendary @yuxuanzhuang thanks so much. :)

hmacdope (approved):
This all looks good to me @yuxuanzhuang. :)

Have your requested changes been made @richardjgowers? I'll merge if good to roll.

richardjgowers (approved):

PR #3680, opened 2022-05-26
Enable automated wheel and dist uploads on releases

### LINKED ISSUE #1300: build binary .whl packages ?

IAlibay:
Fixes #1300

 - Added a deploy action that should build wheels using cibuildwheel, and sdist and then upload them to testpypi / pypi on specific triggers
 - Action runs on every push (so we can notice if something goes wrong), but push to testpypi only happens when you create a tag, and push to pypi only happens when you create a release
 - Created a deploy environment with the necessary pypi / testpypi API token secrets
 - Added maintainer script to normalize release versions in the same way setuptools does (so we can pick it up from testpypi)
 - Standard CI has been expanded to do a full deps check of py3.10 (instead of the minimum deps check done previously)

TODO:
  - Docs

PR Checklist
------------
 - Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -129,13 +129,13 @@ jobs:
           mv dist/MDAnalysisTests-* dist-test/dist
 
       - name: upload_source
-        uses: pypa/gh-action-pypi-publish@release/v1.5.0
+        uses: pypa/gh-action-pypi-publish@v1.5.0
         with:
           user: __token__
           password: ${{ secrets.TESTPYPI_API_TOKEN_SRC }}
 
       - name: upload_tests
-        uses: pypa/gh-action-pypi-publish@release/v1.5.0
+        uses: pypa/gh-action-pypi-publish@v1.5.0
         with:
           user: __token__
           password: ${{ secrets.TESTPYPI_API_TOKEN_TEST }}
@@ -161,13 +161,13 @@ jobs:
           mv dist/MDAnalysisTests-* dist-test/dist
 
       - name: upload_source
-        uses: pypa/gh-action-pypi-publish@release/v1.5.0
+        uses: pypa/gh-action-pypi-publish@v1.5.0
         with:
           user: __token__
           password: ${{ secrets.PYPI_API_TOKEN_SRC }}
 
       - name: upload_tests
-        uses: pypa/gh-action-pypi-publish@release/v1.5.0
+        uses: pypa/gh-action-pypi-publish@v1.5.0
         with:
           user: __token__
           password: ${{ secrets.PYPI_API_TOKEN_TEST }}
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -161,7 +161,13 @@ jobs:
         exclude:
           # Multiple deps don't like windows
           - os: windows-latest
-            python-version: ['3.8', '3.9', '3.10']
+            python-version: "3.8"
+            type: "FULL"
+          - os: windows-latest
+            python-version: "3.9"
+            type: "FULL"
+          - os: windows-latest
+            python-version: "3.10"
             type: "FULL"
           # OpenMM isn't available for this yet
           - os: macos-latest
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -73,7 +73,7 @@ jobs:
   upload_testpypi:
     if: |
       github.repository == 'MDAnalysis/mdanalysis' &&
-      github.event_name == 'push' && startsWith(github.ref, 'refs/tags/'
+      (github.event_name == 'push' && startsWith(github.ref, 'refs/tags/'))
     name: testpypi upload
     environment: deploy
     runs-on: ubuntu-latest
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -125,8 +125,8 @@ jobs:
 
       - name: move_test_dist
         run: |
-          mkdir -p dist-test/dist
-          mv dist/MDAnalysisTests-* dist-test/dist
+          mkdir -p testsuite/dist
+          mv dist/MDAnalysisTests-* testsuite/dist
 
       - name: upload_source
         uses: pypa/gh-action-pypi-publish@v1.5.0
@@ -141,7 +141,7 @@ jobs:
         with:
           user: __token__
           password: ${{ secrets.TESTPYPI_API_TOKEN_TEST }}
-          packages_dir: dist-test/
+          packages_dir: testsuite/
           skip_existing: true
           repository_url: https://test.pypi.org/legacy/
  
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -120,11 +120,23 @@ jobs:
           name: artifact
           path: dist
 
-      - uses: pypa/gh-action-pypi-publish@v1.5
+      - name: move_test_dist
+        run: |
+          mkdir -p dist-test/dist
+          mv dist/MDAnalysisTests-* dist-test/dist
+
+      - name: upload_source
+        uses: pypa/gh-action-pypi-publish@v1.5
         with:
           user: __token__
-          password: ${{ secrets.TEST_PYPI_API_TOKEN }}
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy.yaml
```diff
@@ -204,8 +229,9 @@ jobs:
 
       - name: pip_install_mda
         run: |
-          pip install --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple MDAnalysis
-          pip install --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple MDAnalysisTests
+          ver=$(grep "__version__ =" package/MDAnalysis/version.py | awk '{print $3}')
```
orbeckst: If you're using awk (instead of `cut`) then you can do it in one command
```bash
awk '/__version++ =/ {print $3; quit}' package/MDAnalysis/version.py
```
(I *think* the `quit` works.)

IAlibay:
Ok this should be fine now, I've tested the testpypi portion of the release and wheels were built fine here: https://test.pypi.org/project/MDAnalysisTests/2.2.0.dev0/#modal-close

@MDAnalysis/coredevs this needs reviewing as a priority, it's necessary for the 2.2.0 release. It _should_ make our life a lot easier going forward too!

@Seanny123 - I _think_ this should fix the wheel issue, but if you can have a look at if this meets your needs that would be great.

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,246 @@
+name: Build and upload to PyPI
+
+on:
+  pull_request:
+    branches:
```
orbeckst, line 5: Will this trigger only on branches that start with package- ?

Could you add docs that describe the release workflow to make clearer how it's intended to work?

.github/workflows/deploy.yaml
IAlibay: done

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,246 @@
...
+    defaults:
+      run:
+        working-directory: ./package
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Build sdist
+        run: pipx run build --sdist
```
orbeckst, line 78: is `pipx` installed in ubuntu-latest?

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,246 @@
...
+        with:
+          name: artifact
+          path: dist
+
+      - name: move_test_dist
+        run: |
+          mkdir -p testsuite/dist
+          mv dist/MDAnalysisTests-* testsuite/dist
+
+      - name: upload_source
```
orbeckst: Does this upload wheels as well? Just asking because the name implies "source packages only".

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,246 @@
...
+          password: ${{ secrets.PYPI_API_TOKEN_SRC }}
+
+      - name: upload_tests
+        uses: pypa/gh-action-pypi-publish@v1.5.0
+        with:
+          user: __token__
+          password: ${{ secrets.PYPI_API_TOKEN_TEST }}
+          packages_dir: testsuite/dist
+
+  check_testpypi:
```
orbeckst, line 181: Really cool that we can automatically test the testpypi distribution. Nice.

orbeckst: How long does it take to complete?

orbeckst (changes requested):
This is great DevOps work, @IAlibay . I have some minor questions because I am not that well versed in all the GH-actions magic and I am happy to positively re-review soon.

The main think I am looking for is minimal maintainer docs that explain the release process and outline the thinking behind the changes, i.e., in plain language, what are the goals and gotchas?

.github/workflows/deploy.yaml
IAlibay: I've updated the wiki notes here: https://github.com/MDAnalysis/mdanalysis/wiki/PreparingReleases

We keep a copy of this in the userguide which I'll update once this PR is merged.

.github/workflows/deploy.yaml
IAlibay: Yup, https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md#package-management

Actually was quite surprising to find, I thought I would have needed to install it, but I guess not.

.github/workflows/deploy.yaml
IAlibay: Yeah it uploads both, I've updated the names for the actions to reflect this.

.github/workflows/deploy.yaml
IAlibay: ~ 20 mins per runner

It's quite handy that github actions just lets us deploy to lots of runners if the pool is empty, so an aggregate ~ 6 hours set of tests can be squashed down to 20 mins :)

IAlibay:
Ok, on top of updating the release notes in the wiki I've also opened up this PR in the userguide @orbeckst: https://github.com/MDAnalysis/UserGuide/pull/208

I would prefer not have to duplicate docs a third time, so I might suggest that from now on all maintainer docs go in the userguide.

Seanny123:
> @Seanny123 - I _think_ this should fix the wheel issue, but if you can have a look at if this meets your needs that would be great.

This definitely meets my needs. Previously, I was also posting about trying to build the wheels with minimal Numpy versions, but I am no longer concerned about that, even if I can't tell from this PR the compatibility of the wheels being built.

IAlibay:
> > @Seanny123 - I _think_ this should fix the wheel issue, but if you can have a look at if this meets your needs that would be great.
> 
> This definitely meets my needs.

Awesome. Sorry it took so long to get done :( Our release methods are a bit overly complicated.

> Previously, I was also posting about trying to build the wheels with minimal Numpy versions, but I am no longer concerned about that, even if I can't tell from this PR the compatibility of the wheels being built.

So it _should_ be pulling from our [pyproject.toml](https://github.com/MDAnalysis/mdanalysis/blob/develop/package/pyproject.toml) and using the minimum compatible versions of `numpy` for ABI compatibility purposes.

That being said, I agree that it's pretty annoying that it's not clear from the action what's going on. Let me try re-enabling it with a higher verbosity level, maybe that will tell us what it's doing.

IAlibay:
Ok yeah, it is respecting `pyproject.toml` (hence the 1.19.3 pick): https://github.com/MDAnalysis/mdanalysis/runs/6645750506?check_suite_focus=true#step:3:191

I'll leave verbosity at level 1 so we can debug better if something goes wrong :)

Seanny123:
I appreciate the increased verbosity. This will help me onboard other projects of this type (mdtraj in particular) to a CI release pipeline.

>  Sorry it took so long to get done :( Our release methods are a bit overly complicated.

Your methods are perfectly understandable given the constraints! Making Python fast is non-trivial and makes packaging significantly harder. I'm actually super impressed you got it done at all, let alone so quickly!

.github/workflows/deploy.yaml
orbeckst: Nice!

.github/workflows/deploy.yaml
orbeckst: Thank you — I agree with keeping things DRY and to eventually have docs in only one place.

orbeckst (approved):
Thank you for addressing all my questions and concerns @IAlibay ; I am approving from my end.

However, it would certainly be good if some other @MDAnalysis/coredevs could spare a look.

IAlibay:
> Thank you for addressing all my questions and concerns @IAlibay ; I am approving from my end.
> 
> However, it would certainly be good if some other @MDAnalysis/coredevs could spare a look.

@MDAnalysis/coredevs I'm going to merge this in ~ 24h unless there are any further comments here.

richardjgowers (approved):

PR #3683, opened 2022-05-31
Cythonized Timestep [Core]

### LINKED ISSUE #3709: Cythonize Timestep for greater performance

Belatedly realised that I forgot to open an issue. 

For greater performance of the core `Timestep` object and its various roles in holding coordinate, box and reader data, `Timestep` should be made a Cython Extension type.  

- The `Timestep` object should be made a Cython extension type and all attributes that can be converted to a compiled representation should be.
- No change to the API behaviour permitted
- NumPy operations should use the [NumPy C API](https://numpy.org/devdocs/reference/c-api/array.html) wherever possible
- Outdated attributes and tests should be removed or changed appropriately
- Readers that subclass `Timestep` (INPCRD) should no longer use their own subclass.
- Readers that use the `_frame` attribute should now hold it in the `ts.data` dictionary rather than as an optional attribute 
- All references to the old timestep in the docs should be updated
- A single surface where you can `cimport` the cython layer of `core` should be constructed. 

Feedback welcome on PR #3683 !

hmacdope:
# Cythonized Timestep

First large changes of CZIEOSS4 performance track.  Primarily cythonization of the Timestep class and separation from the iterators. Would be good to get some input at this point to see what people think. 

Fixes #3709

Status:

- [x] Still a few bugs to track down wrt `MemoryReader` and `NetCDF` readers 
- [x] Finalise docs
- [x] Sort out API wrt dtypes following some discussion 

Design Ideas:
- We initially trialled a fully C++ data-structure and found that moving back and forward from two possible types was difficult despite the use of templates because the Cython wrapper class had to know which template to instantiate at `__cinit__` time, making it very ugly for limited performance gain.
- Moved towards a more idiomatic Cython representation
- A big takeaway (suprise suprise) was that using the [NumPy C API](https://numpy.org/devdocs/reference/c-api/array.html) was the best way to get performance.

Aims of this PR:
- Cythonize the Timestep class with an almost exact API match
- Have a test passing and well documented implementation moving forward
- Improve set and get performance of the reader (initial test indicate ~~10x~~ 2x faster to set positions ~~without a copy~~ with a copy if no casting required)

- Timestep is now an Extension type (cdef class) with a `pxd` file to define relevant imports
- `dtype` kwarg added
- ~~`_frame ` is no longer optional as must be initialised (to -1)~~ `_frame` is no longer an attribute and has been  moved to `data` dictionary in `TRZReader`
- `typenum`, `*_dims` attributes added to store enumerated NumPy dtype and associated array dimensionality
- position, velocity and force setters split into 3 branches depending on input object and whether a cast is needed.
- position, forces and velocities are always C contiguous. 
- `__getstate__` and `__setstate__` implemented manually as cdef classes do not have the `__dict__` attribute.
- `__getnewargs_ex__` added to help `__cinit__` construction upon unpickling.
- libmda added as a folder to import all cython from

Discussion Points:
- Datatypes are a big concern here, as of current we do not allow mixed precision in the Timestep. How do we want to tackle this moving forward?
- We currently force convert position dependent data to `np.float32` but this can be changed.
- Boxes are also `np.float32` but elsewhere in the code are promoted and demoted regularly.  
- Do we want to allow a boxes to have a different datatype with an additional kwarg (`dimensions_dtype=np.float...`)?
- Mixed precision ops are slow / difficult for the compiler to vectorize. 

Looking forwards:
- What is the best way to hook C level functions to the data (especially templated functions)? Richard and I are trialling an idiom with a `cython.floating` memoryview and matching templated C function.
- Do we want to provide hooks directly into the array such as with [PyArray_Getptr](https://numpy.org/devdocs/reference/c-api/array.html#c.PyArray_GetPtr)
- What is good to expose at PXD level?

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


testsuite/MDAnalysisTests/coordinates/test_timestep_api.py
```diff
@@ -162,17 +163,17 @@ def test_coordinate_setter_shortcuts(self, ts):
...
     # all Timesteps require this attribute
     def test_n_atoms(self, ts):
-        assert_equal(ts.n_atoms, ts._n_atoms)
+        assert_equal(ts.n_atoms, self.size)
 
     def test_n_atoms_readonly(self, ts):
         with pytest.raises(AttributeError):
             ts.__setattr__('n_atoms', 20)
 
     def test_n_atoms_presence(self, ts):
```
hmacdope, line 191: These `hasattr` tests are now basically useless as the attribute will **always** be there, but we can leave them in for posterity?

package/MDAnalysis/coordinates/timestep.pyx
```diff
@@ -0,0 +1,1030 @@
...
+       :class:`Timestep` init now only accepts integer creation.
+       :attr:`n_atoms` now a read only property.
+       :attr:`frame` now 0-based instead of 1-based.
+       Attributes `status` and `step` removed.
+    .. versionchanged:: 2.0.0
+       Timestep now can be (un)pickled. Weakref for Reader
+       will be dropped.
+       Timestep now stores in to numpy array memory in 'C' order rather than
+       'F' (Fortran).
+    .. versionchanged:: 2.2.0
```
IAlibay: I'd make this 2.3.0 (and everywhere else) at the very least (I'm releasing 2.2.0 as soon as possible).

package/MDAnalysis/coordinates/base.py
```diff
@@ -232,714 +143,6 @@
 from ..lib.util import asiterable, Namespace
```
richardjgowers, line 145: Can import Timestep here to avoid some tears

package/MDAnalysis/coordinates/timestep.pyx
```diff
@@ -0,0 +1,1030 @@
...
+        A dtype for the Timestep can be specified with the `dtype` kwarg.
+        All arrays are now forced to be C contiguous using the NumPy C API.
+    """
+
+    order = 'C'
+
+    cdef uint64_t _n_atoms
+    cdef public int64_t  frame
+    # no longer optional
+    cdef public int64_t  _frame
```
richardjgowers: Doesn't make sense to have two?  If there's readers using `_frame` mush them into shape.  Extra variables should be put into `data`

package/MDAnalysis/coordinates/timestep.pyx
```diff
@@ -0,0 +1,1030 @@
...
+
+    # info for numpy C API
+    cdef int _typenum
+    cdef cnp.npy_intp _particle_dependent_dim[2]
+
+    cdef bool _has_positions
+    cdef bool _has_velocities
+    cdef bool _has_forces
```
richardjgowers: Can these be replaced by properties that check the size (`!0 === has_X`)

richardjgowers: Wait no, this is stupid

richardjgowers:
I had thought that `ts._unitcell` was float64, but apparently it's 32.  So we can avoid mixed precision Timesteps.  I think in the future there's a discussion to be had around our mixed precision distance calculations for the reasons you raise, but it's not here thankfully.

I think putting all the pxds in one place would be a nice idea, so it's obvious which components exposed this way.  I'm not sure if there's a prevalent way to do this yet, the only one I usually use is `Cython.libcpp`, so `from MDAnalysis.libmda cimport Timestep` maybe?

package/MDAnalysis/coordinates/base.py
hmacdope: its the first import :)

package/MDAnalysis/coordinates/timestep.pyx
hmacdope: Done, `TRZReader` was the issue.

package/MDAnalysis/coordinates/timestep.pyx
hmacdope: Done

hmacdope:
FYI this is ready for proper review now. :)

package/MDAnalysis/coordinates/timestep.pxd
```diff
@@ -0,0 +1,31 @@
+from libcpp cimport bool
```
richardjgowers, line 1: header junk required

package/MDAnalysis/coordinates/timestep.pxd
```diff
@@ -0,0 +1,31 @@
...
+    cdef bool _has_forces
+
+    # these have to be public for testing
+    cdef public cnp.ndarray _unitcell
+    cdef public cnp.ndarray _pos
+    cdef public cnp.ndarray _velocities
+    cdef public cnp.ndarray _forces
+
+    # python objects
+    cdef object _dtype
```
richardjgowers, line 42: is this either `np.float32` or `np.float64`?

package/MDAnalysis/coordinates/timestep.pxd
```diff
@@ -0,0 +1,31 @@
...
+
+cdef class Timestep:
+
+    cdef uint64_t _n_atoms
+    cdef public int64_t  frame
+
+
+    # info for numpy C API
+    cdef int _typenum
+    cdef cnp.npy_intp _particle_dependent_dim[2]
```
richardjgowers, line 20: is the second dimension always 3?  &if so isn't this variable always `_n_atoms, 3`?

package/MDAnalysis/coordinates/timestep.pyx
```diff
@@ -0,0 +1,982 @@
...
+        cdef cnp.npy_intp unitcell_dim_tmp[1]
+        unitcell_dim_tmp[0] = 6
+
+        # use temps so that we don't have to allocate a bunch of empty
+        # arrays of large size, eg for vel and frc
+        cdef cnp.npy_intp particle_dependent_dim_tmp[2]
+        particle_dependent_dim_tmp[0] = 0
+        particle_dependent_dim_tmp[1] = 0
+
+        # these must be initialised, we initialise small
```
richardjgowers: initialise small as in size zero?

package/MDAnalysis/coordinates/timestep.pyx
```diff
@@ -0,0 +1,982 @@
...
+        return self._has_positions
+
+    @has_positions.setter
+    def has_positions(self, val):
+        if val and not self._has_positions:
+            # Setting this will always reallocate position data
+            # ie
+            # True -> False -> True will wipe data from first True state
+            self._pos = cnp.PyArray_ZEROS(
+                2, self._particle_dependent_dim, self._typenum, 0)
```
richardjgowers: This is reallocating right?  So if we had a pathological traj which had saved positions every other frame, we'd be reallocating every other frame?  Maybe better could be another if inside, which is if the array exists, memset(0)

package/MDAnalysis/coordinates/timestep.pyx
```diff
@@ -0,0 +1,982 @@
...
+
+        return ts
+
+    def __eq__(self, other):
+        """Compare with another Timestep
+
+        .. versionadded:: 0.11.0
+        """
+        if not isinstance(other, Timestep):
+            return False
```
richardjgowers: I know this copies previous behaviour, but I think I've since learnt that the appropriate response here is to return `return NotImplemented` which iirc then checks if `other.__eq__(self)` returns anything better.  This is how an `other.__class__` could say it is equal to this, despite this not knowing of other.

package/MDAnalysis/coordinates/timestep.pyx
```diff
@@ -0,0 +1,982 @@
...
+        if not self.has_forces == other.has_forces:
+            return False
+        if self.has_forces:
+            if not (self.forces == other.forces).all():
+                return False
+
+        return True
+
+    def __ne__(self, other):
+        return not self == other
```
richardjgowers: I can't remember if this is automatically implemented in py3?

package/MDAnalysis/coordinates/timestep.pyx
```diff
@@ -0,0 +1,982 @@
...
+        else:
+            raise TypeError
+
+    def __getattr__(self, attr):
+        # special-case timestep info
+        if attr in ('velocities', 'forces', 'positions'):
+            raise NoDataError('This Timestep has no ' + attr)
+        err = "{selfcls} object has no attribute '{attr}'"
+        raise AttributeError(err.format(selfcls=type(self).__name__,
+                                        attr=attr))
```
richardjgowers, line 756: I can't remember why we have this... Is it still necessary now that `.positions` etc are properties?

package/MDAnalysis/coordinates/timestep.pyx
```diff
@@ -0,0 +1,982 @@
...
+        """
+        for i in range(self.n_atoms):
+            yield self[i]
+
+    def __repr__(self):
+        desc = "< Timestep {0}".format(self.frame)
+        try:
+            tail = " with unit cell dimensions {0} >".format(self.dimensions)
+        except NotImplementedError:
+            tail = " >"
```
richardjgowers: This might not be being caught because `.dimensions` is always present but sometimes `None`

package/MDAnalysis/coordinates/timestep.pxd
```diff
@@ -0,0 +1,31 @@
+from libcpp cimport bool
+from libc.stdint cimport int64_t, uint64_t
+cimport numpy as cnp
+cnp.import_array()
+
+
+cdef class Timestep:
```
richardjgowers, line 7: minimal docs (a few words each variable) would be nice

richardjgowers (changes requested):

richardjgowers:
Oh, and WRT dtype, we can fix this to float32 for now and sort that out later in a different PR.  I can't see anything here that prevents us switching dtype, but it'll require plumbing into other parts of the code and a boatload of extra tests, so let's keep this PR simple.

package/MDAnalysis/coordinates/timestep.pxd
hmacdope: Yeah its set to `_n_atoms, 3` on `__cinit__`

package/MDAnalysis/coordinates/timestep.pxd
hmacdope: Yep will note, its checked on `__init__`

package/MDAnalysis/coordinates/timestep.pyx
hmacdope: Depends if you want a custom error message for attribute error, but I agree that it doesn't serve its original purpose as `velocities, forces ...` are always present. 

Should I remove?

package/MDAnalysis/coordinates/timestep.pxd
hmacdope: need some of this but cleaned up a bit

hmacdope:
> Oh, and WRT dtype, we can fix this to float32 for now and sort that out later in a different PR. I can't see anything here that prevents us switching dtype, but it'll require plumbing into other parts of the code and a boatload of extra tests, so let's keep this PR simple.

With respect to this should I remove the `dtype=` kwarg for this PR?

package/MDAnalysis/coordinates/timestep.pyx
hmacdope: Defers to `__eq__` and inverts so I have removed.

package/MDAnalysis/coordinates/timestep.pyx
hmacdope: I don't think its now possible or sensible to catch anything here anyway so I will remove the `try: except: block` and replace with a test of `dimensions is not None`

package/MDAnalysis/coordinates/timestep.pyx
hmacdope: I have added a flag that tracks allocation of the array and either zero fills using `PyArray_FILLWBYTE (memset under hood)` if the array is already allocd OR allocs it to zero with `PyArray_ZEROS` with the correct shape if its the first time.

This way we can avoid lots of reallocs, without carrying around a bunch of large uninitialised arrays from `__cinit__`

hmacdope:
I am removing the `dtype` kwarg from this PR and we can discuss further on an issue. I will raise and link here.

package/MDAnalysis/coordinates/timestep.pyx
hmacdope: EDIT:~~I am going to remove this as it no longer serves a useful purpose with attributes being always present.~~

This is used in `AtomGroup` so can stay for now.

hmacdope:
@tylerjereddy would I please be able to get a review from you specifically on my use of the NumPy C API?

There are quite a few calls to the C API, but I am not aware if there are any NumPy conventions I might be ignoring and would like to be as conformant as possible. 

Particularly, initialising the `_position`, `_velocities` and `_forces` arrays to be empty with size 0 using `PyArray_EMPTY()` in `__cinit__` was something I wanted to query. 

 Is there a better way to have a C level array hook without allocating to their full size (natoms,  3)?

hmacdope:
> I am removing the `dtype` kwarg from this PR and we can discuss further on an issue. I will raise and link here.

Discussion of dtypes can take place on #3707

richardjgowers (approved):

PR #3685, opened 2022-05-31
Fix reader copying by storing init arguments

### LINKED ISSUE #3664: `ProtoReader` based classes don't use `kwargs` on `copy`
Follow-up from #3642 

The `copy` method of `ProtoReader` derived classes doesn't account for kwargs which were passed on class construction. The use of `convert_units` was fixed in #3642 but we should do it for kwargs too.

IAlibay:
Fixes #3664

This is the final PR needed for 2.2.0.

So this isn't a very pretty fix but it's the best I could come up with for now. I'm happy to go with alternatives, including deprecating / removing `copy`.

 - Added decorator for storing class initialisation arguments.
 - Fixed `copy` methods for readers, allowing for copies to get the same init arguments as the original object

Todo
 - Need to update a couple of tests for a few readers that have custom init arguments (just to check that everything works as intended since the base classes aren't used by all readers).

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


richardjgowers:
I like the decorator, but I think maybe `MemoryReader` is again going to be an exception.  If I create a `MemoryReader`, modify it's coordinates (i.e. unwrap or something) then copy it, I expect to get the same copied coordinates back.  I *think* this is what will happen with the current code (since it's all shallow copies).  One wrinkle is the docs say you'll get "Return independent copy of this Reader.
        New Reader will have its own file handle and can seek/iterate
        independently of the original".  This won't quite be true with memoryreader as the new Reader will share the same numpy array, so modifying one affects the other.  I think maybe memoryreader will have to initiate the copy with a deepcopy of its internal coordinates to behave more like other readers here.

IAlibay:
> I like the decorator, but I think maybe `MemoryReader` is again going to be an exception. If I create a `MemoryReader`, modify it's coordinates (i.e. unwrap or something) then copy it, I expect to get the same copied coordinates back. I _think_ this is what will happen with the current code (since it's all shallow copies). One wrinkle is the docs say you'll get "Return independent copy of this Reader. New Reader will have its own file handle and can seek/iterate independently of the original". This won't quite be true with memoryreader as the new Reader will share the same numpy array, so modifying one affects the other. I think maybe memoryreader will have to initiate the copy with a deepcopy of its internal coordinates to behave more like other readers here.

Thanks for not shooting down the decorator 😅 @richardjgowers (although I still think it's ugly)

re: memoryreader - doesn't it already do deepcopies for coordinates/velocities/positions/dimensions? https://github.com/MDAnalysis/mdanalysis/blob/1945878e224beb1076be258432d911d5fda5f54a/package/MDAnalysis/coordinates/memory.py#L431-L437

richardjgowers:
Ah ok cool, didn't realise MemoryReader was already clever.  So long as the input arrays aren't captured by the decorator this looks good

IAlibay:
> So long as the input arrays aren't captured by the decorator this looks good

Yeah I got caught out by that early on whilst doing this fix 🤣 hence why for the memoryreader case I just do a shallow copy of the kwargs input dictionary only (so it should only be for `dt` and `time_offset`).

package/MDAnalysis/coordinates/DCD.py
```diff
@@ -68,6 +68,7 @@
 from . import base, core
 from ..lib.formats.libdcd import DCDFile
 from ..lib.mdamath import triclinic_box
+from ..lib.util import store_init_arguments
```
lilyminium, line 72: Are there readers that would not want to store init arguments? If not, would it be easier to use the Reader metaclass to wrap `__init__`? Or would that lead to problems subclassing

IAlibay: MemoryReader is a good example of a case where we wouldn't want to store all the init arguments (you could but then it gets rather messy to extract out just the kwargs, and you also end up capturing references to all the input arrays in `self._kwargs`).

There is probably a very elegant way to do this using metaclasses, however I'm rather short on my self-allocated time to get 2.2.0 done. I gave it a quick shot and failed, so unless someone has a very specific "here is how I would do it" instruction, I might say that we kick this down the bucket as a cleanup issue for the future?

testsuite/MDAnalysisTests/coordinates/test_copying.py
```diff
@@ -37,19 +37,25 @@
     DMS,
```
IAlibay, line 37: 🤦🏽 I had completely forgot this set of tests existed (to be fair, looks like we've been forgetting about it for the last few readers too).

IAlibay (commented):
Sorry for the mess here, ended up going back and forth on tests so there's quite a few files with just pep8 / import order fixes (didn't think it worth changing back).

Overall this should be it for now. Thanks for the comments up until now, if we could get a formal yeah/nay on this that would be great @MDAnalysis/coredevs, it is one of the last things holding up 2.2.0.

package/MDAnalysis/coordinates/memory.py
```diff
@@ -302,11 +303,15 @@ def __init__(self, coordinate_array, order='fac',
         .. versionchanged:: 0.19.0
             The input to the MemoryReader now must be a np.ndarray
             Added optional velocities and forces
+        .. versionchanged:: 2.2.0
+            Input kwargs are now stored under the `_kwargs` attribute, and
```
orbeckst:
```suggestion
            Input kwargs are now stored under the :attr:`_kwargs` attribute, and
```
or use double backticks.

package/MDAnalysis/lib/util.py
```diff
@@ -203,6 +204,8 @@
 from functools import wraps
 import textwrap
 import weakref
+import copy
+import shutil
```
orbeckst: Why is `shutil` added if it's not used in your code?

package/MDAnalysis/lib/util.py
```diff
@@ -2417,3 +2420,42 @@ def check_box(box):
...
+      by the class after passing through the decorator this will be
+      reflected in the stored arguments.
+    * If not empty, ``args`` is not unpacked and stored as-is in the
+      dictionary. If no ``args`` are passed, then no ``arg`` entry will be
+      stored in the dictionary.
+
+
+    .. versionadded:: 2.2.0
+    """
+    sig = inspect.signature(func)
```
orbeckst, line 2440: Whenever I see `inspect` I shudder... :-p

testsuite/MDAnalysisTests/coordinates/test_copying.py
orbeckst: You earned the test archaeologist badge 🏅

testsuite/MDAnalysisTests/coordinates/test_copying.py
```diff
@@ -113,6 +179,37 @@ def test_reader_filename(original_and_copy):
     assert original.filename == copy.filename
 
 
+def test_reader_copied_extra_attributes(original_and_copy_extra_args):
+    # Issue #3664
+    original, copy = original_and_copy_extra_args
+
+    # memory and chair readers subclass protoreader directly and
```
orbeckst:
```suggestion
    # memory and chain readers subclass protoreader directly and
```

testsuite/MDAnalysisTests/coordinates/test_mol2.py
```diff
@@ -34,6 +34,7 @@
     mol2_zinc, mol2_comments_header, mol2_ligand, mol2_sodium_ion
 )
 from MDAnalysis import Universe
+from MDAnalysis.coordinates.MOL2 import MOL2Reader
```
orbeckst: Previously, the tests worked _without_ this line??!?

testsuite/MDAnalysisTests/coordinates/test_trj.py
```diff
@@ -29,6 +28,8 @@
     assert_almost_equal
 )
 
+import MDAnalysis as mda
+from MDAnalysis.coordinates.TRJ import TRJReader
```
orbeckst: Previously, the tests worked without this line??!?

orbeckst (approved):
That was a lot of worked for a "small" issue! Thank you for doing it all.

I understand the rationale for using the decorator approach instead of introducing yet another "base/proto/intermediate" class in the readers. Maybe in the future we have a rethink on how to organize the readers... perhaps as part of the Cythonization effort?

I have minor comments/suggestions but nothing blocking.

orbeckst:
There's an issue with 3.10, though
```
>       r = get_reader_for(filename, format=fmt_name)(filename, **extras)

/home/runner/work/mdanalysis/mdanalysis/testsuite/MDAnalysisTests/coordinates/test_copying.py:101: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/home/runner/work/mdanalysis/mdanalysis/package/MDAnalysis/lib/util.py:2460: in wrapper
    return func(self, *args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <[AttributeError("'ChemfilesReader' object has no attribute 'filename'") raised in repr()] ChemfilesReader object at 0x7fc5308527a0>
filename = '/home/runner/work/mdanalysis/mdanalysis/testsuite/MDAnalysisTests/data/lammps/datatest.data'
chemfiles_format = 'LAMMPS Data', kwargs = {}
```

orbeckst:
... ah, you were fixing it while I was complaining...

testsuite/MDAnalysisTests/coordinates/test_chemfiles.py
```diff
@@ -102,6 +103,39 @@ def ref():
         return ChemfilesXYZReference()
 
 
+    def test_copy(self, ref):
```
IAlibay, line 105: These extra tests in chemfiles and H5MD aren't super pretty, but it's the only way I could think of having them sit within import guarded regions without making `test_copying` overly complicated.

package/MDAnalysis/lib/util.py
IAlibay: oops here and below are leftover imports from me testing out various things - should be fixed now.

IAlibay:
Thanks for the review @orbeckst, hopefully the last few commits should have addressed your comments?

>  Maybe in the future we have a rethink on how to organize the readers... perhaps as part of the Cythonization effort?

Yeah 100%, I'll open an issue after this is merged with a general "re-organising readers. We probably should also one day get back to moving the reader tests to use the base classes too.

> ... ah, you were fixing it while I was complaining...

Yeah sorry I noticed that the readers with optional deps don't like being in `test_copying`. I've added individual tests for each of them, it's not pretty, but at least it should do a suitable job of checking things work for now?

testsuite/MDAnalysisTests/lib/test_util.py
```diff
@@ -2104,3 +2105,58 @@ def test_check_box_wrong_shape(self):
...
+                            d='has', e='a', f='recipe', g='allegedly')
+        assert store.a == store._kwargs['a'] == "parsnips"
+        assert store.b == store._kwargs['b'] == "roast"
+        assert store.c == store._kwargs['c'] == "richard"
+        assert store.d == store._kwargs['d'] == "has"
+        assert store.e == store._kwargs['e'] == "a"
+        assert store.kwargs['f'] == store._kwargs['f'] == "recipe"
+        assert store.kwargs['g'] == store._kwargs['g'] == "allegedly"
+        assert store.args[0] == store._kwargs['args'][0] == "honey"
+        assert store.args[1] == store._kwargs['args'][1] == "glaze"
```
richardjgowers, line 2159: They're pretty delicious

richardjgowers (approved):

PR #3691, opened 2022-06-01
Add Cython 0.29.30 generated files

IAlibay:
Fixes #

 - 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


IAlibay:
Worth noting that the sdist and wheel actions will re-create the C/C++ files, in this case it's fine since I used the same Cython version as the C files I'm adding here. But it's worth keeping in mind (actually maybe in the future we just don't want to have the C/C++ files as part of the package)?

richardjgowers (approved):

richardjgowers:
I think we discussed removing the c files from the repo before, it's probably a good idea.

PR #3694, opened 2022-06-01
switch awk call, temporarily disable guards for tests

### LINKED ISSUE #3693: awk call in deploy.yaml does not work on macos-latest
Not too problematic since it only affects testing, but this line fails on the macos-latest runners:

https://github.com/MDAnalysis/mdanalysis/blob/eab18cb8418ddb1dd72b44f474833de4a2999884/.github/workflows/deploy.yaml#L240

I suspect reverting to `ver=$(grep "__version__ =" package/MDAnalysis/version.py | awk '{print $3}')` will probably fix things.

IAlibay:
Fixes #3693 #3696 

Opening this now so that we can see the macos tests pass for testpypi's 2.2.0 build. Once we are 2.3.0 we can move towards merging this.

 - switch out the awk call

PR Checklist
------------
 - Tests?
 - Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -238,7 +238,7 @@ jobs:
 
       - name: pip_install_mda
         run: |
-          grep "__version__ =" package/MDAnalysis/version.py | awk '{print $3}' > version.dat
+          grep "__version__ =" package/MDAnalysis/version.py | awk '{print $3}' | tr -d \" > version.dat
           ver=$(python maintainer/norm_version.py --file version.dat)
           pip install --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple MDAnalysis==$ver
           pip install --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple MDAnalysisTests==$ver
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -3,6 +3,7 @@ name: Build and upload to PyPI
 on:
   pull_request:
     branches:
+      - develop
       - "package-*"
   push:
     branches:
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yaml
```diff
@@ -237,7 +238,7 @@ jobs:
 
       - name: pip_install_mda
         run: |
-          awk '/__version__ =/ {print $3; quit}' package/MDAnalysis/version.py | tr -d \" > version.dat
+          grep "__version__ =" package/MDAnalysis/version.py | awk '{print $3}' | tr -d \" > version.dat
```
orbeckst:
```suggestion
          awk '/__version__ =/ {print $3; exit}' package/MDAnalysis/version.py | tr -d \" > version.dat
```

orbeckst (commented):
exit instead of quit might do the job... but it's really not that important.

.github/workflows/deploy.yaml
```diff
@@ -237,7 +237,7 @@ jobs:
 
       - name: pip_install_mda
         run: |
-          awk '/__version__ =/ {print $3; quit}' package/MDAnalysis/version.py | tr -d \" > version.dat
+          awk '/__version__ =/ {print $3; exit}' package/MDAnalysis/version.py | tr -d \" > version.dat
```
orbeckst, line 240: In case you want to do everything inside `awk`:
```suggestion
          awk '/__version__ =/ {gsub("\"", "", $3); print $3; exit}' package/MDAnalysis/version.py > version.dat
```

IAlibay: 🤣 is this a really necessary change or are we just showing off `awk` knowledge at this point?

orbeckst: I'm just showing off my age ;-p

orbeckst: Ok, hard data:
```
time for x in seq 10000000; do awk '/__version__ =/ {gsub("\"", "", $3); print $3; exit}' package/MDAnalysis/version.py > version.dat; done

real	0m0.013s
user	0m0.002s
sys	0m0.005s
```

vs

```
time for x in seq 10000000; do awk '/__version__ =/ {print $3; exit}' package/MDAnalysis/version.py | tr -d \" > version.dat; done

real	0m0.021s
user	0m0.007s
sys	0m0.018s
```

Almost 50% speed-up ... that's worth something, right? ;-)

IAlibay: As much as I like awk (the reference book is still one of the best manuals I've read to this date), I'm going to say wontfix on this. Mostly because re-enabling all the pipelines is likely more CO2 than the difference between piping to `tr` and not 🙃 (and more effort).

orbeckst: I won't block the PR ;-)

orbeckst (approved):
lgtm

Issue #3698, opened 2022-06-03
run mypy and pyright in CI


jbarnoud:
## Is your feature request related to a problem? ##

Type hinting is an addition to modern python that helps detect errors before runtime. Type analysis must be run on the code base separately from when the code is run.

## Describe the solution you'd like ##

Run mypy and pyright as part of CI. Because type hinting of the library is not implemented yet, so the job should only be indicative and not blocking.

## Describe alternatives you've considered ##
Not running the tools as part of CI. However, it will not be sustainable as new code will break the type annotation making its maintenance impossible.

## Additional context ##
This issue is part of @umak1106's outreachy project.

IAlibay:
Since you raised the non blocking issue - as far as I'm aware you can't do that properly with GitHub actions (see: https://github.com/actions/runner/issues/2347). It'll need digging into, but my suggestion might be to make it into a separate action so at least we don't get a red badge for multiple months. Right now I can't think of a plan that isn't "we'll ignore the failure", we could make the action trigger only on specific keywords / branch names / etc... if it becomes problematic.

tylerjereddy:
I'd suggest a slightly different approach described in https://github.com/MDAnalysis/mdanalysis/issues/3699#issuecomment-1145942156

I think it might also be more satisfying/motivating for the student to have their hard work "guarded" by CI in early stages of their effort.

That said, while I know this works perfectly well for `mypy`, I can't comment on `pyright`. Maybe you could start with `mypy` in any case.

hmacdope:
is this closed by #3705?

jbarnoud:
I'd like pyright to be included as well, but it is fairly low priority.

Issue #3542, opened 2022-02-22
retire "master" branch


orbeckst:
# Background
MDAnalysis has existed for a while and so there’s a bit of history built into our development process. One of these pieces of history is that we  follow the git-flow development model http://nvie.com/posts/a-successful-git-branching-model/ (see https://github.com/MDAnalysis/mdanalysis/issues/96 ; the post from 2010 also has a 2020 update). In this model, development is carried out on the develop branch and release-ready code is merged into the master branch. Feature branches are branched off of develop.

Experience over the last 10+ years has shown that for MDAnalysis, the develop branch is stable and useable (mainly thanks to broad testing) and that the master branch is little more than a place to hold release tags. We also don’t have the capacity to support multiple releases in parallel (for instance, MDA 1.x is legacy now and we are not updating it anymore). Many other projects have adopted a rolling development model with a single main branch that is continually updated. In this model, releases are either short branches of the main branch or just tags. 

# Change: Retire master branch

**I propose that we adopt the single main-branch model for future MDAnalysis development.**

In effect, we will just retire master and **continue working on develop** as before. 

Releases will be branched off of develop. We can keep the name ‘develop’ for this branch. 

This change should simplify our release management and syncing the structure of the User Guide (which has versions for current development and published releases).

# Discussion

The proposal was discussed on the developer list https://groups.google.com/g/mdnalysis-devel/c/ncocXN3EoaE/m/YfJQwjAIFgAJ and found broad agreement.

@jbarnoud commented: 
> At some point, it was question to have one branch per major version (1.x, 2.x) for the occasional backport. Is it still the idea?

Yes. (Although we decided not to touch 1.x anymore as the CI for Python 2.7 is too complicated and fragile. The discussion for long term support/backports is also open.)

@tylerjereddy commented
> * Sounds good, there are also ethical reasons to distance a project from a branch named "master."
> * Creating small release branches off of develop would mirror NumPy/SciPy as well, if you go that route.

IAlibay:
Do we need to do anything here?

As far as I'm aware all this means is that starting next release (i.e. 2.1.0) I'll just make a tag and/or feature branch and go with that?

I guess the main decision is - are tags good enough or do we want a feature branch for each release?

jbarnoud:
[...]

PR #3705, opened 2022-06-05
Type Checking in CI

### LINKED ISSUE #3699: make type checking in CI blocking
This issue depends on #3698 which requests mypy and pyright to be run in CI to check type annotations. This issue suggests making the job non-blocking as type annotation is not yet implemented. Once the annotations are implemented to a point where the checking job passes, it should be made blocking to avoid breaking the existing annotations and to avoid introducing new broken ones.

This is part of @umak1106's outreach project.

umak1106:
Fixes #3699 #3698 

 - Added mypy.ini and mypy_requirements.txt

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


umak1106:
Hello mentors ,
I have added the mypy.ini file . Currently I have set it to ignore all the errors in the modules . As and when I'll start type hinting the module I'll make the required changes to the file .
Incase you have any suggestions regarding the improvements to be made here kindly let me know

.github/workflows/mypy.yaml
```diff
@@ -0,0 +1,45 @@
+name: Mypy CI
+on:
+  push:
+    branches:
+      - develop
+      - master
```
jbarnoud: We do not have a master branch anymore for MDAnalysis. You can read the discussion about this here: https://github.com/MDAnalysis/mdanalysis/issues/3542

.github/workflows/mypy.yaml
```diff
@@ -0,0 +1,45 @@
...
+  group: "${{ github.ref }}-${{ github.head_ref }}"
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: bash -l {0}
+
+jobs:
+  mypy_test :
+    if: "github.repository == 'MDAnalysis/mdanalysis'"
```
jbarnoud, line 21: Why do you need this conditional? Note, I am not saying you do not need it, I genuinely do not know about the possible usefulness of that line.

.github/workflows/mypy.yaml
```diff
@@ -0,0 +1,45 @@
...
+      runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: setup_miniconda
+        uses: conda-incubator/setup-miniconda@v2
+        with:
+          python-version: 3.7
```
jbarnoud: Out of curiosity, why did you choose this version of python?

jbarnoud (commented):
It looks like there is a syntax error in `mypy.yaml`. You can see the result of the workflow here: https://github.com/umak1106/mdanalysis/actions

.github/workflows/mypy.yaml
lilyminium: This limits this action only to pushes and PRs to `MDAnalysis/mdanalysis`, instead of forks. We have it throughout the rest of our CI, I assume as a courtesy to the GitHub servers / personal allotments.

IAlibay: it's a nice way to avoid forks from running CI if they don't want to (for mypy it's not that bad, but every little helps when it comes to reducing CO2 emissions) edit: Lily got there before me :P

.github/workflows/mypy.yaml
```diff
@@ -0,0 +1,45 @@
...
+      - develop
+      - master
+  pull_request:
+    branches:
+      - develop
+      - master
+
+concurrency:
+
+  group: "${{ github.ref }}-${{ github.head_ref }}"
```
IAlibay: Please use the grouping which we use in gh-ci.yaml This one is prone to collisions

.github/workflows/mypy.yaml
```diff
@@ -0,0 +1,43 @@
...
+  run:
+    shell: bash -l {0}
+
+jobs:
+  mypy :
+    if: "github.repository == 'MDAnalysis/mdanalysis'"
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
```
jbarnoud: In the main workflow, we use version 3.

jbarnoud (commented):
Good progress today! The workflow runs :smiley: 

I see a couple of things left to fix but we have a great start here. One thing to fix is that you need the workflow to install the dependencies fo MDAnalysis. In the main workflow, it is done here: https://github.com/MDAnalysis/mdanalysis/blob/ec5da1de682322c436970252f03b5a24ea673113/.github/workflows/gh-ci.yaml#L91

.github/workflows/mypy.yaml
umak1106: I removed master branch

.github/workflows/mypy.yaml
umak1106: Actually in gh-ci both pylint and pypi used python version 3.8 so i did the same here

umak1106:
@jbarnoud 
Hello mentor , I made some changes in the mypy.yaml file and am unable to figure out why some of the tests are failing . If i need to make any other changes please let me know.

jbarnoud:
The tests being broken seems unrelated. We will look at fixing this.

Regarding the install phase in the yaml file, I do not think we need the full matrix. Indeed, the type check should not change with different operating systems, for instance. This means you should be able to simplify your file by removing the matrix you copied and, instead, use values directly where you have variables. I *think* you should go with `true` for the full dependencies, `1.19.0` for the NumPy version (the oldest version we support), and I do not think you need extra pip dependencies.

As a side note, it would be good to have more informative commit messages. For the moment, all your commit messages are the same, which makes it difficult for me to know what you changed and why you changed it, and for you to look back at a specific commit. Note that we, as a project, are not very picky with commit messages; other projects, though, may use commit messages almost like cover letters where a contributor justifies why the commit should be included. As an example, your previous commit could have been titled something like "Use a dependency matrix for the mypy workflow".

.github/workflows/mypy.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+      - name: setup_miniconda
+        uses: conda-incubator/setup-miniconda@v2
+        with:
+          python-version: 3.8
+          auto-update-conda: true
+          channel-priority: flexible
+          channels: conda-forge
+          add-pip-as-python-dependency: true
+          architecture: x64
```
jbarnoud, line 36: Looks like you need `mamba-version: "*"` for mamba to be installed. See https://github.com/MDAnalysis/mdanalysis/blob/ec5da1de682322c436970252f03b5a24ea673113/.github/workflows/gh-ci.yaml#L88

.github/workflows/mypy.yaml
```diff
@@ -0,0 +1,51 @@
...
+          channels: conda-forge
+          add-pip-as-python-dependency: true
+          mamba-version: "*"
+          architecture: x64
+
+      - name: install_deps
+        uses: ./.github/actions/setup-deps
+        with:
+          mamba: true
+          full-deps: true
```
IAlibay, line 42: Given that the 2.3.0 release will likely support a minimum numpy version of 1.20 and that type annotations were added to numpy in 1.20, it would be better to set this to 1.20.0.

I'll submit a PR today or tomorrow raising our minimum numpy requirements to 1.20 for everything else.

.github/workflows/mypy.yaml
umak1106: yes  mentor , I just realized after making the commit .

.github/workflows/mypy.yaml
umak1106: ok mentor.

.github/workflows/mypy.yaml
jbarnoud: I'll wait a bit before giving away too many answers, then :wink:

umak1106: Actually am getting the error , nothing provides numpy 1.20.0 . What am I supposed to do about it ?

jbarnoud: The problem is the way you specify the numpy version. It is a bit unintuitive: the `numpy` attribute expects `numpy=1.20.0` and not just `1.20.0`.

Since, we are at it, you should also check the `channel` attribute.

jbarnoud: Oh, you already checked `channel` :smile:

umak1106: I think the workflow runs properly now 😄.

jbarnoud (approved):
Looks like it is working :tada: 

I cannot merge right now, until the rest of the CI is fixed. But we will fix that soon.

Also, @IAlibay could you have a quick look? I am not super up to date with our CI so I may be missing things.

richardjgowers (approved):
LGTM.  Since the CI entry from this isn't failing, I don't think merging this makes the current CI problems any worse...

.github/workflows/mypy.yaml
```diff
@@ -0,0 +1,51 @@
...
+          numpy: numpy=1.20.0
+
+      - name: install
+        run: |
+          python -m pip install mypy
```
IAlibay, line 47: Do you not need to install MDAnalysis?

IAlibay (commented):
Sorry only doing a very cursory look at things, overall looks good but I don't remember if you have to (or should) install the MDA package for a mypy check. If so then a `python setup.py install` call somewhere (unless I missed it), would be necessary

IAlibay:
Also I agree with @richardjgowers it's fine to go ahead with the merge, the failures are unrelated.

.github/workflows/mypy.yaml
umak1106: Umm as we already installed the dependencies should i add 
          cd package
          python setup.py develop --no-deps
instead of 
cd package
python setup.py install ?

.github/workflows/mypy.yaml
```diff
@@ -0,0 +1,51 @@
...
+          add-pip-as-python-dependency: true
+          mamba-version: "*"
+          architecture: x64
+
+      - name: install_deps
+        uses: ./.github/actions/setup-deps
+        with:
+          mamba: true
+          full-deps: true
+          numpy: numpy=1.20.0
```
tylerjereddy, line 43: For future reference, I'd suggest using even newer NumPy for the type checking because the type hint machinery available in NumPy continues to improve with each release. There are some "new in version.." statements in the docs here:

https://numpy.org/devdocs/reference/typing.html#typing-numpy-typing

mypy.ini
```diff
@@ -0,0 +1,54 @@
+[mypy]
+ignore_missing_imports = True
+exclude = doc/sphinx/source/
```
tylerjereddy, line 3: If the team decides that using NumPy `1.22.x` series is acceptable for the purposes of type checking, you can use the new NumPy plugin as well: `plugins = numpy.typing.mypy_plugin`

see: https://numpy.org/devdocs/reference/typing.html#mypy-plugin

tylerjereddy (approved):
Sounds like folks are happy here, I'll squash merge then given some of the commit messages.

I added a few comments about fancy things you could use if you bump the NumPy version a bit more, but I'll leave that up to the team/mentors.

tylerjereddy:
I created a "typing" label and applied it here, hope that's ok, feel free to change it if you prefer something else.

PR #3706, opened 2022-06-06
Remove OpenMP private clauses by narrowing variable scopes

RMeli:
I was looking around the code base and I noticed that OpenMP "private" clauses were used a lot in `calc_distances.h`. AFAIK (but correct me if I'm wrong!), it is preferred to narrow the scope of the variables (within the parallel region) instead of making them private.

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


richardjgowers (approved):
lgtm thanks @RMeli

richardjgowers:
I'm not sure which is preferred wrt private vs scope, but it seems like there's less room for trouble with smaller scoped variables anyway

IAlibay:
> I'm not sure which is preferred wrt private vs scope, but it seems like there's less room for trouble with smaller scoped variables anyway

It's been some time but it does technically speed things up iirc. In part because you don't have to create & distribute the variables to workers. In practice I doubt it has a major impact (although I've been pleasantly surprised with micro optimisations in OpenMP before!). 

That being said 💯 smaller scoped variables seem like a better idea.

The only thing I'm going to ask here @RMeli  - could you confirm from the action logs that OpenMP build are actually happening? Would be bad if we've not been testing this entire time 😨

RMeli:
On Ubuntu, OpenMP seems to be correctly detected:
```
Attempting to autodetect OpenMP support... Compiler supports OpenMP
```

On macOS, OpenMP support is not detected
```
Attempting to autodetect OpenMP support... Did not detect OpenMP support.
No openmp compatible compiler found default to serial build.
```
but this is consistent to what I see locally when using Apple's `clang`. Building on macOS locally with
```
CC=gcc-11 CXX=g++-11 pip install -e package
```
works as expected (where `gcc-11` is a `brew` installed GCC, not the default `gcc` on macOS, that is actually `clang`...).

hmacdope:
@RMeli  do we have any performance metrics to check the impact? I'd be happy with a screenshot :)

RMeli:
@hmacdope Good question. I got the following with one of the tests from the test suite.

Original implementation:
```
%timeit val = distances.distance_array(a, b, box=box, backend="OpenMP")[0, 0]
29.3 µs ± 8.48 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

This implementation:
```
%timeit val = distances.distance_array(a, b, box=box, backend="OpenMP")[0, 0]
27 µs ± 3.29 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

I also checked the assembly produced by GCC12 (using [Compiler Explorer](https://godbolt.org/)) for the following snippets and they appear to be equivalent:
```cpp
int main(){
  double r[3];
  #pragma omp parallel for private(r)
  for (int i = 0; i < 100; i++){
    r[0] = i;
    r[1] = 2 * i;
    r[2] = i * i;
  }
  return 0;
}
```
```cpp
int main(){
  #pragma omp parallel for
  for (int i = 0; i < 100; i++){
    double r[3];
    r[0] = i;
    r[1] = 2 * i;
    r[2] = i * i;
  }
  return 0;
}
```

hmacdope:
@RMeli great work! Cheers.

PR #3711, opened 2022-06-09
add h5py to setup.py's extra_requires

### LINKED ISSUE #3701: Add h5py to `extra_requires` and set minimum version to 3.0.0
Follows up from discussions here: https://github.com/conda-forge/mdanalysis-feedstock/pull/43

Since h5py can be installed via PyPI we should add it to our setup.py as an extra requirement. Also we should set the minimum version to 3.0 so we can drop the `ValueError` check which had to be included for backwards compatibility here: https://github.com/MDAnalysis/mdanalysis/pull/3025

orbeckst:
Fixes #3701

- keep minimum version of h5py at >= 2.1.0
- new 'extra_formats' target in extra_requires where we can add pip-installable packages for additional formats that are not installed by default
- deprecated 'AMBER' target in extra_requires (now part of 'extra_formats')

PR Checklist
------------
 - [ ] Tests? (not sure how to test)
 - [ ] Docs?
 - [x] CHANGELOG updated?


package/setup.py
```diff
@@ -63,15 +63,8 @@
...
 
-if sys.version_info[0] < 3:
-    import ConfigParser as configparser
-else:
-    import configparser
-
-if sys.version_info[0] >= 3:
-    from subprocess import getoutput
-else:
-    from commands import getoutput
```
orbeckst, line 74: removed Python 2 compatibility crud

package/setup.py
```diff
@@ -650,6 +643,10 @@ def long_description(readme):
               'AMBER': [
                   'netCDF4>=1.0',  # for fast AMBER writing, also needs HDF5
               ],
+              'extra_formats': [   # additional file formats
+                  'netCDF4>=1.0',  # for fast AMBER writing, also needs HDF5
+                  'h5py>=2.10',    # H5MD
+                  ],
```
orbeckst: possible add here
* gsd (unless we just install it anyway)
* chemfiles

package/setup.py
```diff
@@ -650,6 +643,10 @@ def long_description(readme):
               'AMBER': [
```
orbeckst: deprecated the AMBER target (not sure if anyone ever used it...)

package/setup.py
```diff
@@ -650,6 +643,10 @@ def long_description(readme):
...
                   'netCDF4>=1.0',  # for fast AMBER writing, also needs HDF5
               ],
+              'extra_formats': [   # additional file formats
+                  'netCDF4>=1.0',  # for fast AMBER writing, also needs HDF5
+                  'h5py>=2.10',    # H5MD
```
orbeckst, line 655: left at minimal version for right now

IAlibay: This needs to be reflected in all the CI places:
- [x] gh action
- [x] azure
- <s>[ ] Travis</s>
- [x] maintainer's yaml

package/setup.py
IAlibay: gsd is an install requires for now so it shouldn't also be going into extra requires

chemfiles might be an option if we want

IAlibay (changes requested):
oops comments weren't via review - I'll request changes over setting minimum versions for all the CI files please.

package/setup.py
orbeckst: - Travis does not install h5py. https://github.com/MDAnalysis/mdanalysis/blob/0b3ba59d7a31d461593dfa83ef50b38f171ad9eb/.travis.yml#L27-L28
- Azure does not pin or restrict packages when h5py is installed https://github.com/MDAnalysis/mdanalysis/blob/0b3ba59d7a31d461593dfa83ef50b38f171ad9eb/azure-pipelines.yml#L84-L98 although farther down, we have packages with restrictions (such as chemfiles). I am not sure what to add where... please review

IAlibay: I think it's fine, I'll give it a review once I'm back from the wilds of Edinburgh

package/setup.py
orbeckst: I added chemfiles>=0.10

package/setup.py
orbeckst: I had change the azure file to have the restriction and the tests passed, so at least it didn't immediately break anything.

orbeckst:
I installed into an empty py39 env with
```
pip install ./package[extra_formats,analysis]
```
and packages import
```
Python 3.9.0 | packaged by conda-forge | (default, Nov 26 2020, 07:54:06)
[Clang 11.0.0 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
imp>>> import h5py
>>> import chemfiles
>>> h5py.__version__
'3.7.0'
>>> chemfiles.__version__
'0.10.2'
>>> import netCDF4
>>> netCDF4.__version__
'1.5.8'
>>> import MDAnalysis
>>> MDAnalysis.__version__
'2.3.0-dev0'
```
and after installing the testsuite,  all tests pass
```
pytest  --disable-warnings  -n 4  testsuite/MDAnalysisTests
```
```
================= 17782 passed, 359 skipped, 2 xfailed, 2 xpassed, 32302 warnings in 315.77s (0:05:15) ==================
```

azure-pipelines.yml
```diff
@@ -92,7 +92,7 @@ jobs:
       pytest-xdist
       scikit-learn
       scipy
-      h5py
+      h5py>=2.10
```
IAlibay, line 95: Let's leave it as this for now, long term we might want to come back and put minimum pins on stuff here too (although to be honest we'll just pick up the maximum version here anyways, I guess I'm just being too pedantic...)

package/CHANGELOG
```diff
@@ -25,10 +25,15 @@ Enhancements
...
 Changes
+  * To add additional optional packages for different file formats when
+    using `pip` use the new "extra_formats" extras_require target as in
+    `pip install ./package[extra_formats]` (Issue #3701)
 
 Deprecations
   * Deprecate `density` parameter in favor of `norm` in InterRDF_s
     (Issue #3687)
+  * The extras_requires target "AMBER" for `pip install ./package[AMBER]`
+    will be removed in 3.0. Use "extra_formats". (Issue #3701, PR #3711)
```
```suggestion
  * The extras_requires target "AMBER" for `pip install ./package[AMBER]`
    will be removed in 3.0. Use "extra_formats". (Issue #3701, PR #3711)
  * Deprecate `density` parameter in favor of `norm` in InterRDF_s
    (Issue #3687)
```

Newest entry first :)

IAlibay: Also I guess we shouldn't be a pain to the couple of folks that might use this, but does it really need to be a major version breaking change? The number of automated workflow using this should be near 0, and there's no way they'll get a deprecatewarning about this. If we enact the change the biggest issue is that things will just break cleanly right? (or will it just ignore the keyword and go ahead)

IAlibay (changes requested):
just the one thing, otherwise lgtm, thanks @orbeckst !

IAlibay:
Whilst we're here, do the changes here indicate that this is also fixed? https://github.com/MDAnalysis/mdanalysis/issues/3406

package/CHANGELOG
orbeckst: I don’t think it really needs to be a major revision deprecation. I was just very conservative. I’m not sure how many people used it, tbh.

i haven’t tried proving missing/incorrect target names. I *assume* it will break. 

I’m happy to go with a deprecation for 2.4, too, just let me know or change directly in setup.py and CHANGELOG.

orbeckst:
Regarding #3406 I don’t know — we could add a restriction to not use h5py 3.4.0 but we would be guessing as we never pinpointed the ultimate cause. Ultimately we need someone to test if that failure is really only occurring with specific releases, then we exclude all bad ones here, and then I’d be comfortable closing the issue.

package/CHANGELOG
orbeckst: I set the deprecation for 2.4.

azure-pipelines.yml
orbeckst: I understand your comment as "h5py>=2.10 is fine for now" so I am not changing it.

package/CHANGELOG
IAlibay: I tried passing the wrong extra (as 'analyses' instead of 'analysis')
```
WARNING: mdanalysis 2.3.0.dev0 does not provide the extra 'analyses'
````

So it won't fail if you give it the wrong thing (which is rather annoying). I'm still ok with setting the deprecation for 2.4 given the fact that we can't suitably warn users anyways, it'll be a problem either way.

orbeckst:
coverage-related error makes one test fail — no idea what to do about it; is this related to the need to pin coverage again that was mentioned somewhere on discord?

orbeckst:
@IAlibay let me know if you think something else needs to be done here. No rush.

orbeckst:
GitHub hangs on "Checking for ability to merge automatically…" —weird. 

Anyway, @IAlibay is there anything else that needs doing?

IAlibay (approved):
"I'll give it a review once I'm back from the wilds of Edinburgh" oof I think I see why I didn't get around to re-reviewing this until this week. So sorry D: !

PR #3713, opened 2022-06-09
add 0.5 for correct midpoints

### LINKED ISSUE #3715: incorrect midpoints in HOLE calculation
This line is wrong:
https://github.com/MDAnalysis/mdanalysis/blob/0b3ba59d7a31d461593dfa83ef50b38f171ad9eb/package/MDAnalysis/analysis/hole2/hole.py#L1306

Therefore, any output in HOLE that uses midpoints is wrong.

(Issue raised based on PR #3713 by @miss77jun which also gives the correct expression `midpoints = 0.5*(bins[1:] + bins[:-1])`)

miss77jun:
Fixes #3715

 - As the title, the midpoints missed a 0.5.

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/analysis/hole2/hole.py
```diff
@@ -1303,7 +1303,7 @@ def plot_mean_profile(self, bins=100, range=None,
 
         binned, bins = self.bin_radii(frames=frames, bins=bins, range=range)
         mean = np.array(list(map(np.mean, binned)))
-        midpoints = 0.5 * bins[1:] + bins[:-1]
+        midpoints = 0.5 * bins[1:] + 0.5 * bins[:-1]
```
orbeckst:
```suggestion
        midpoints = 0.5 * (bins[1:] + bins[:-1])
```

orbeckst (changes requested):
Yes, this looks like a bug! Thank you for finding and fixing it. Did you raise an issue, too? EDIT: I raised #3715 for you — it's convenient to be able to refer to an issue as the bug and the PR as the fix.

We also need a test in [testsuite/MDAnalysisTests/analysis/test_hole2.py](https://github.com/MDAnalysis/mdanalysis/blob/develop/testsuite/MDAnalysisTests/analysis/test_hole2.py) that checks that the bug has been fixed (and stays fixed). The best way to do this is to write a test that **fails** with the **old code** and passes with the fixed code.

If you need help with writing tests then please ask.


Please also
- add an entry to CHANGELOG under Fixes
- add yourself to AUTHORS

package/MDAnalysis/analysis/hole2/hole.py
orbeckst: Just "commit the suggestion" in order to make the code a bit more concise.

orbeckst:
It looks as if the **tests are wrong, too**, so I think all you need to do for testing is to change the following line https://github.com/MDAnalysis/mdanalysis/blob/0b3ba59d7a31d461593dfa83ef50b38f171ad9eb/testsuite/MDAnalysisTests/analysis/test_hole2.py#L406

lilyminium:
Yikes, thanks for finding this, @miss77jun.

The line that @orbeckst pointed out (which I most likely wrote!) is a really good example of a bad test, since all it does is replicate the error. You could just fix that line, since now multiple people have had a look and think the suggested change is correct. However, a better one (if slightly overkill) might be checking the actual spirit and goal of `midpoints`, e.g. by putting in an assertion (`assert_allclose`) that checks if `midpoints` is of very similar distance from `bins[1:]` as it is from `bins[:-1]`.

orbeckst:
@miss77jun it would be really fantastic if you could complete the PR — I think we would already be happy if you could 
- [x] fix the test
- [x] accept the suggestion to make the code more compact
- [x] update CHANGELOG
- [x] add yourself to AUTHORS
- [x] introduce yourself on the developer list as a new contributor (see comment early in the thread).

It would be great to have you as a new contributor to MDAnalysis!

If you have any questions or need help then please ask. We're happy to help.

testsuite/MDAnalysisTests/analysis/test_hole2.py
```diff
@@ -403,7 +404,10 @@ def test_plot_mean_profile(self, hole, frames, profiles):
         binned, bins = hole.bin_radii(bins=100)
         mean = np.array(list(map(np.mean, binned)))
         stds = np.array(list(map(np.std, binned)))
-        midpoints = 0.5 * bins[1:] + bins[:-1]
+        midpoints = 0.5 * bins[1:] + 0.5 * bins[:-1]
+        # assume a bin of length 1.5
+        assert_allclose(midpoints, bins[1:], atol=0.75)
```
lilyminium: This is a really creative use of `assert_allclose`! However, it is usually to better to be much more direct with tests. For example, instead of allowing any value within 0.75 of a midpoint in either direction, you could directly calculate the difference between `midpoint` and `bins` and check that it is 0.

```suggestion
        difference_right = bins[1:] - midpoints
        assert_allclose(difference_right, 0.75)
```

A similar test could be done with the left-hand-side (`bins[:-1]`) if you wanted to check that the midpoint was actually midway through.

package/MDAnalysis/analysis/hole2/hole.py
lilyminium: @miss77jun are you happy with @orbeckst's suggestion? If so, there's an easy way to add it to the code with the "Commit suggestion" button. If not, it would be good to explain why so we get what you're thinking!

package/CHANGELOG
```diff
@@ -13,12 +13,13 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" http://semver.org
 
 ------------------------------------------------------------------------------
-??/??/?? IAlibay, PicoCentauri, orbeckst
+??/??/?? IAlibay, PicoCentauri, orbeckst, Tengyu Xie
```
lilyminium:
```suggestion
??/??/?? IAlibay, PicoCentauri, orbeckst, miss77jun
```

Typically we ask for the GitHub username here, just so we can track contributions being added :)

lilyminium (changes requested):
Thanks for fixing the PEP8 issues @miss77jun! Just a few minor comments

lilyminium:
@miss77jun I just took the liberty of merging the MDAnalysis/develop branch into your pull request to allow tests to run. MDAnalysis/develop got updated while this was open, which caused a conflict in the CHANGELOG file, so I resolved it. This shouldn't require anything from you, but if you want to make changes to your code and push to this branch again, you might need to `git pull origin` first just to update your local copy :)

miss77jun:
> @miss77jun I just took the liberty of merging the MDAnalysis/develop branch into your pull request to allow tests to run. MDAnalysis/develop got updated while this was open, which caused a conflict in the CHANGELOG file, so I resolved it. This shouldn't require anything from you, but if you want to make changes to your code and push to this branch again, you might need to `git pull origin` first just to update your local copy :)

Thank you for all you have done and the careful tips. Very happy. I learned a lot from my first PR.

testsuite/MDAnalysisTests/analysis/test_hole2.py
```diff
@@ -403,7 +404,11 @@ def test_plot_mean_profile(self, hole, frames, profiles):
         binned, bins = hole.bin_radii(bins=100)
         mean = np.array(list(map(np.mean, binned)))
         stds = np.array(list(map(np.std, binned)))
-        midpoints = 0.5 * bins[1:] + bins[:-1]
+        midpoints = 0.5 * bins[1:] + 0.5 * bins[:-1]
```
orbeckst:
```suggestion
        midpoints = 0.5 * (bins[1:] + bins[:-1])
```
Writing it this way, saves a whole bunch of calculations and performance is significantly better:
```python
>>> bins = np.linspace(0, 10, 80)   # typical bins

>>>  %timeit  midpoints = 0.5 * bins[1:] + 0.5 * bins[:-1]
2.15 µs ± 47.1 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

>>> %timeit  midpoints = 0.5 * (bins[1:] + bins[:-1])
1.5 µs ± 6.91 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```
In relative terms, a speed-up of 30% for such a small change is really good.

In practice, the difference between 2.15 µs and 1.5 µs does not matter for the HOLE analysis. But nevertheless we want to write the best code that we can, not the least because existing code serves as an example for other code.

testsuite/MDAnalysisTests/analysis/test_hole2.py
```diff
@@ -403,7 +404,11 @@ def test_plot_mean_profile(self, hole, frames, profiles):
...
+        # assume a bin of length 1.5
+        difference_right = bins[1:] - midpoints
+        assert_allclose(difference_right, 0.75)
```
orbeckst: This one currently fails as you can see eg in https://github.com/MDAnalysis/mdanalysis/runs/6858196294?check_suite_focus=true#step:7:520

```
        difference_right = bins[1:] - midpoints
>       assert_allclose(difference_right, 0.75)
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=0
E       
E       Mismatched elements: 100 / 100 (100%)
E       Max absolute difference: 0.54100065
E       Max relative difference: 0.7213342
E        x: array([0.208999, 0.208999, 0.208999, 0.208999, 0.208999, 0.208999,
E              0.208999, 0.208999, 0.208999, 0.208999, 0.208999, 0.208999,
E              0.208999, 0.208999, 0.208999, 0.208999, 0.208999, 0.208999,...
E        y: array(0.75)
```

Adding a comment here to document your assumption was a good idea. I don't think you can just assume that the bin size is 1.5. You could calculate the bin width as 
```python
binwidths = np.diff(bins)
binwidth = binwidths[0]
assert_allclose(binwidths, bindwidth)  # just making sure that we have equidistant bins

difference_right = bins[1:] - midpoints
assert_allclose(difference_right, binwidth/2)
```

orbeckst (changes requested):
Good progress but your new assertion failed. Please see comments.

orbeckst (approved):
Looks all good to me — thank you!

The failures that we currently see in CI do not seem to be related to your work. We'll have to figure out what it is (and I posted a question on discord) but I am approving the PR in the mean time. Well done!

orbeckst:
@miss77jun thank you for your contribution. Sorry I wasn't able to review sooner. In the future, if you haven't received a new review after, say, 3 days, please feel free to notify your reviewers by adding a comment and including their `@`-handle.

orbeckst:
@lilyminium what do you think? Have a quick look. 

Should we merge despite the CI (which seems unrelated) or do you (or anyone else) have an idea how to fix the following (copied back from Discord #developers):

I am seeing some weird CI failures such as this one https://github.com/MDAnalysis/mdanalysis/runs/6911249174?check_suite_focus=true with 
```
INTERNALERROR>     raise DataError(
INTERNALERROR> coverage.exceptions.DataError: Conflicting file tracer name for '/home/runner/work/mdanalysis/mdanalysis/package/MDAnalysis/coordinates/timestep.pyx': 'Cython.Coverage.Plugin' vs ''**
```
or https://github.com/MDAnalysis/mdanalysis/runs/7027723546?check_suite_focus=true with
```
INTERNALERROR>     raise DataError(
INTERNALERROR> coverage.exceptions.DataError: Conflicting file tracer name for '/home/runner/work/mdanalysis/mdanalysis/package/MDAnalysis/lib/formats/libmdaxdr.pyx': 'Cython.Coverage.Plugin' vs ''
```

Both are part of https://github.com/MDAnalysis/mdanalysis/pull/3713 (this PR) and make a number of Python 3.8 runs fail. Note that none of our tests fail but the testing process exits with error code 3 and CI records a failure.

Does anyone know what's happening here? Is it related to the coverage plugin?

RMeli:
@orbeckst I've seen the same error yesterday on #3728 but today I added a commit and everything came back green. Can you try to just re-run the failing job and see if that fixes it?

IAlibay:
The internal error stuff is down to when we unpinned coverage and pytest-cov, it needs repining whilst we work out why is happening (I can't reproduce it locally). Rerunning CI may clear it, it's random it seems.

lilyminium (approved):

lilyminium:
Looks great to me, thank you @miss77jun!

orbeckst:
I'll merge the PR (just had to update it).

(Thank you everyone for commenting on the CI failures.)

orbeckst:
We get the same internal failures but I'll merge anyway.

orbeckst:
Congratulations @miss77jun on your first merged PR! Well done! 🎉

PR #3718, opened 2022-06-15
[HOLD] Re-implement distance_array using a batched algorithm and add compiled iterator layer [Core, WIP]

### LINKED ISSUE #3708: Provide Ndarray or AtomGroup to distance routines

We should encourage AtomGroups  as the currency of choice for passing between routines. A major place where this is not the case is in the routines in `distances.py` which can only accept a NumPy array which is then  converted to a `float*` in  `calc_distances.h`.

 It would be fantastic to be able to leverage the power of the routines in `calc_distances.h` directly with an `AtomGroup`

Taking `distance_array` as an example it could have the type hinted signature:

```python
def distance_array(reference: Union[np.ndarray, AtomGroup], configuration: Union[np.ndarray, AtomGroup], box: Optional[np.ndarray] = None, result=None,
                   backend: str = "serial") -> None:
```

After some benchmarking in https://github.com/hmacdope/Distance_array_wrapper @richardjgowers and myself settled on using a batched algorithm with a coordinate buffer and preloading of the indices as the most performant option.

This may require partial Cythonization of the `AtomGroup` object, but due to a lot of meta-programming around the `AtomGroup` class itself it may be best to replace only the data intensive portion of the AtomGroup with an Extension type.

hmacdope:
Fixes #3708

UPDATE
------------
This PR is currently on hold pending decisions around batched distance arrays. 

**Aim:** 
- Add a batched distance array algorithm for efficient memory use.  
- Allow distance routines to accept `NumPy` arrays or `AtomGroups` usign a compiled iterator.

Second large set of changes of CZIEOSS4 performance track. I have so far only changed `_distance_array` without a box but  probably at a good stage to discuss now before moving on to change the rest of `calc_distances.h`

**Changes made in this pull request:**

-  Introduces the `Iterator` layer to homogenise behaviour between `ndarrays` and `AtomGroups` at the C++ level with a shared interface. 
- Allows distance routines to accept either `ndarray` or `AtomGroup` in any position using **templates**. This stops us having to write a huge number of overloads or use Cython polymorphism which is pretty annoying. For example `_calc_dihedrals(a,b,c,d)` would require 24 overloads for flexibility at every position. 
-  Introduces a **batched** distance  array algorithm that uses stack allocated arrays. Testing of the `C++` with `googlebench`demonstrated this to be very fast with a modest memory footprint. This concept also  links in well with the  structure in `distopia` and possible some `CUDA` code.

Status:

- [x] Finalise overhang part of batched distance array calculation (currently not working for `ncoords>batchsize`)
- [ ] Change remainder of `_calc_distances.h`
- [x] Finalise docs for iterator API
- [ ] Make sure types are consistent (`int` vs `uint64_t` etc)
- [ ] Add tests for `AtomGroup` distance calculations
- [ ] Check performance with benchmarks.

Design Ideas:

We need a `AtomGroup` to behave like a `NumPy` array at the C++ level. We achieve this using C++ classes that provide a shared interface for iteration over `float*` coordinate arrays with different implementations for their specific datastructure. 

These are then wrapped in a thin Cython class wrapper. AtomGroups or NumPy arrays are then turned into their respective iterators before being passed to the C++ level.

If people like this pattern, other groups can have iterators written very easily. They just need to share an interface with the others. 

PR Checklist
------------
 - [ ] Tests?
 - [X] Docs?
 - [ ] CHANGELOG updated?


package/MDAnalysis/lib/iterators.pxd
```diff
@@ -0,0 +1,24 @@
+from libcpp.vector cimport vector
+from libc.stdint cimport int64_t
+
+cdef extern from "iterators.h":
+    cdef cppclass _AtomGroupIterator:
+        int64_t n_atoms
+        vector[int64_t] ix
```
richardjgowers: why can't this (and below) be an `int64_t*`?

package/MDAnalysis/lib/include/batched_distances.h
```diff
@@ -0,0 +1,97 @@
...
+    // batchsize is in particles, so allocate buffers of 3x batchsize.
+    const uint64_t atom_bufsize = 3 * batchsize;
+    float ref_buffer[atom_bufsize];
+    float conf_buffer[atom_bufsize];
+
+    // slight indirection required here to homogenize interface between
+    // _AtomGroupIterator and _ArrayIterator where passing stack allocated
+    // array as float*& does not decay to a float* as desired.
+    float *ref_buffer_ = ref_buffer;
+    float *conf_buffer_ = conf_buffer;
```
richardjgowers: Could the scope of these be reduced to inside the loop if `load_into` returned a float*?

package/MDAnalysis/lib/include/iterators.h
```diff
@@ -0,0 +1,142 @@
...
+#include <cstdio>
+
+/* This header provides classes to iterate groups (eg AtomGroup) and NumPy
+   arrays using the same API. This is done by using thin wrapper classes with
+   the same interface methods:
+
+   - reset_iteration() -> rewind to start of group/array
+   - load_into_external_buffer -> pass values from group/array into an external
+     buffer (likely a stack allocated array)
+   - seek (optional/unused) -> go to a particular atom in the group/array
```
richardjgowers: yagni?

package/MDAnalysis/lib/iterators.pxd
hmacdope: Good point will change

package/MDAnalysis/lib/include/batched_distances.h
hmacdope: I will investigate

package/MDAnalysis/lib/include/iterators.h
hmacdope: Might need to fix overhanging part of array but will investigate.

package/MDAnalysis/lib/include/batched_distances.h
```diff
@@ -0,0 +1,179 @@
...
+    // using a gcd allows us to use the maximum size of memory buffer that fits
+    // remaining iteration in (gcd_ref, gcd_conf) sized tiles
+    int64_t gcd_conf = _gcd(nconf, bsize_conf);
+    int64_t gcd_ref = _gcd(nref, bsize_ref);
+
+    // note i is used as index of ref and j is used as index of conf.
+
+    for (iter_ref = 0; iter_ref < nref - ref_overhang; iter_ref += bsize_ref)
+    {
+        ref.load_into_external_buffer(ref_buffer_, bsize_ref);
```
richardjgowers: So to avoid the two loops, one for regular and one for overhang, how about the following:
```c++
for (int i=0; i<n_ref; ++i) {
    if !(i % bsize_ref) {
        ref.load_into_external_buffer(ref_buffer, i, bsize_ref);
        local_i = 0;
    }

    for (int j=0; j<nconf; ++j) {
        if !(j % bsize_conf) {load stuff}


        dx[0] = conf_buffer_[3 * local_j] - ref_buffer_[3 * local_i];
        dx[1] = conf_buffer_[3 * local_j + 1] - ref_buffer_[3 * local_i + 1];
        dx[2] = conf_buffer_[3 * local_j + 2] - ref_buffer_[3 * local_i + 2];
        rsq = (dx[0] * dx[0]) + (dx[1] * dx[1]) + (dx[2] * dx[2]);
        *(distances + nconf * i + j) = sqrt(rsq);
    }
}
```

So
- you loop over `i` and `j` as you normally would, and every `bsize` iterations you refill the buffers.
- there's a new variable `local_i` which tracks the position within a buffer.  This is identical to `i % bsize` (and if bsize_ref is a clever number this is therefore `i & 1023` or similar)
- for the above reason I'd make `bsize` a constant somewhere so the compiler does things (but maybe you know this and are optimizing this later)
- The signature to `load_into_external_buffer` changes to accept a start point for the load, which removes the need for it to track any state (this is instead done inside the function stack)

hmacdope: Just for clarity I will paraphrase a private conversation between myself and @richardjgowers  here. With my current implementation this approach doesn't allow us to utilise the so called "squareness" of the computation. i.e.

```c++
for (int i=0; i<n_ref; ++i) {
    if !(i % bsize_ref) {
        ref.load_into_external_buffer(ref_buffer, i, bsize_ref);
        local_i = 0;
    }

    for (int j=0; j<nconf; ++j) { // <- HERE j=0 and j%bsize_conf is true every i
    // so we reload the buffer way too often 
        if !(j % bsize_conf) {load stuff}

```
Which means we are refilling one buffer for every iteration on one coordinate. There is likely a way around this but it didn't jump out to me right away. 

A secondary concern is that with a single loop only and no separation for overhang you need to explicitly check that you are not overrunning the edge of the total number of indices. i.e.

```c++
    if !(i % bsize_ref) {
        ref.load_into_external_buffer(ref_buffer, i, bsize_ref); // <- this can read over the total number of idx 
        local_i = 0;
    }
```
thus requiring an explicit check in the loading function which is slow. 

As a result of this I have reverted to my previous design. I know its a bit ugly with three large nested loops, but it is as fast as `_calc_distance_array()` or faster according to my benchmarking, is very memory efficient and gives us the flexibility that we want.

package/MDAnalysis/lib/include/iterators.h
hmacdope: Needed in the end.

package/MDAnalysis/lib/iterators.pxd
hmacdope: Changed

package/MDAnalysis/lib/include/batched_distances.h
hmacdope: Done, I think not passing the pointer by reference is advisable in general as risk for modification is high.

hmacdope:
Some further benchmarking has shown that there doesn't appear a large speedup to be gained using a compiled iterator vs using a Python NumPy index call.
 
Given the large changes proposed in this PR for apparently limited gain I think I will switch tack and use a call to  `.positions`. It's a shame to invest a fair bit into this design and then change tack but I think we needed to explore it fully to make sure it wasn't the right fit. Regardless lesson learnt about the speed of NumPy indexing. 

The batching of coordinates using stack allocated memory as is done `batched_distances.h` does however hold some promise with a 2-8% speedup compared to current `develop`. It also fits more into the style we had in `distopia` but would be a large rewrite of `calc_distance.h` so perhaps best shelved for now. 
 
With that in mind would people prefer I leave this here and open another PR to address the original issue (#3708 )? @richardjgowers or should I just continue here.

hmacdope:
If there are no objections I will move the solution to the original issue (#3708) to a new PR and rename this PR accordingly? Will  do tomorrow if no objections.

hmacdope:
@richardjgowers and I moved away from this idea, I am going to close, feel free to re-open if we want to revive some aspect of this PR.

PR #3719, opened 2022-06-17
Providing type hints for core module

umak1106:
Fixes #

 - 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/core/_get_readers.py
+++ b/package/MDAnalysis/core/_get_readers.py
@@ -1,4 +1,4 @@
-# -*- Mode: python; tab-width: 4; indent-tabs-mode:nil; coding:utf-8 -*-
+#-*- Mode: python; tab-width: 4; indent-tabs-mode:nil; coding:utf-8 -*-
 # vim: tabstop=4 expandtab shiftwidth=4 softtabstop=4
 #
 # MDAnalysis --- http://www.MDAnalysis.org
@@ -22,13 +22,16 @@ circular imports.
 import copy
 import inspect
 
+import mmtf
+
 from .. import (_READERS, _READER_HINTS,
                 _PARSERS, _PARSER_HINTS,
                 _MULTIFRAME_WRITERS, _SINGLEFRAME_WRITERS, _CONVERTERS)
 from ..lib import util
+from typing import Optional, Union, Type
 
 
-def get_reader_for(filename, format=None):
+def get_reader_for(filename: str, format: Optional[Union[str, _READERS]] = None) -> _READERS:
     """Return the appropriate trajectory reader class for `filename`.
 
     Parameters
@@ -107,7 +110,7 @@ def get_reader_for(filename, format=None):
         raise ValueError(errmsg) from None
 
 
-def get_writer_for(filename, format=None, multiframe=None):
+def get_writer_for(filename: Optional[str], format: Optional[str] = None, multiframe: Optional[bool] = None) -> Union[_SINGLEFRAME_WRITERS, _MULTIFRAME_WRITERS]:
     """Return an appropriate trajectory or frame writer class for `filename`.
 
     The format is determined by the `format` argument or the extension of
@@ -202,7 +205,7 @@ def get_writer_for(filename, format=None, multiframe=None):
         raise TypeError(errmsg.format(format)) from None
 
 
-def get_parser_for(filename, format=None):
+def get_parser_for(filename: Union[str, mmtf.MMTFDecoder], format: Optional[str] = None) -> _PARSERS:
     """Return the appropriate topology parser for `filename`.
 
     Automatic detection is disabled when an explicit `format` is
@@ -257,7 +260,8 @@ def get_parser_for(filename, format=None):
         else:
             return _PARSERS['MINIMAL']
 
-def get_converter_for(format):
+
+def get_converter_for(format: str) -> Type[_CONVERTERS]:
     """Return the appropriate topology converter for ``format``.
 
     Parameters
--- a/package/MDAnalysis/core/_get_readers.py
+++ b/package/MDAnalysis/core/_get_readers.py
@@ -28,9 +28,11 @@ from .. import (_READERS, _READER_HINTS,
                 _PARSERS, _PARSER_HINTS,
                 _MULTIFRAME_WRITERS, _SINGLEFRAME_WRITERS, _CONVERTERS)
 from ..lib import util
-from typing import Optional, Union, Type, Any
-from ..coordinates.base import ProtoReader, WriterBase, ConverterBase
-from ..topology.base import TopologyReaderBase
+from typing import Optional, Union, Type, Any, TYPE_CHECKING
+if TYPE_CHECKING:
+    
+    from ..coordinates.base import ProtoReader, WriterBase, ConverterBase
+    from ..topology.base import TopologyReaderBase
 
 
 def get_reader_for(filename: Any, format: Optional[Union[str, ProtoReader]] = None) -> ProtoReader:
--- a/package/MDAnalysis/core/_get_readers.py
+++ b/package/MDAnalysis/core/_get_readers.py
@@ -30,12 +30,12 @@ from .. import (_READERS, _READER_HINTS,
 from ..lib import util
 from typing import Optional, Union, Type, Any, TYPE_CHECKING
 if TYPE_CHECKING:
-    
+
     from ..coordinates.base import ProtoReader, WriterBase, ConverterBase
     from ..topology.base import TopologyReaderBase
 
 
-def get_reader_for(filename: Any, format: Optional[Union[str, ProtoReader]] = None) -> ProtoReader:
+def get_reader_for(filename: Any, format: Optional[Union[str, Type[ProtoReader]]] = None) -> Type[ProtoReader]:
     """Return the appropriate trajectory reader class for `filename`.
 
     Parameters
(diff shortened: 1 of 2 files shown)


umak1106:
When I run mypy on this file it runs without any errors .
But I am unable to figure out what class name should be used instead of ConverterWrapper in t.Type[Converterwrapper].
Also if there are any other mistakes in this please let me know so .

jbarnoud:
I'll have a look on Monday :)

package/MDAnalysis/core/accessors.py
```diff
@@ -63,10 +63,14 @@
    :members:
 
 """
+from __future__ import absolute_import
```
tylerjereddy: What does this accomplish? According to the table in the standard library docs here: https://docs.python.org/3/library/__future__.html

`absolute_import` has been mandatory since Python `3.0`, and we have a higher minimum version of Python than that, right?

package/MDAnalysis/core/accessors.py
```diff
@@ -63,10 +63,14 @@
...
+from __future__ import annotations
 
 from functools import partial, update_wrapper
 
 from .. import _CONVERTERS
+import typing as t
```
tylerjereddy: minor, I think: I often see objects from `typing` imported directly, i.e., `from typing import Optional, Dict`. I guess this may be personal preference though.

package/MDAnalysis/core/accessors.py
```diff
@@ -63,10 +63,14 @@
...
+from .groups import AtomGroup
```
tylerjereddy: Looks like CI is complaining about a circular import here: 

```
ImportError: cannot import name 'AtomGroup' from partially initialized module 'MDAnalysis.core.groups' (most likely due to a circular import)
```

This kind of thing, avoiding circular imports with type hinting, can indeed be a genuine problem. The team may have some suggestions, I know this one looks familiar: https://stackoverflow.com/a/39757388. For example, NumPy and SciPy both use the `TYPE_CHECKING` guard in select locations. Maybe something to discuss with the team/mentor a bit.

package/MDAnalysis/core/accessors.py
umak1106: Yes , I'll remove that part .

package/MDAnalysis/core/accessors.py
umak1106: Ill make the changes.

package/MDAnalysis/core/accessors.py
hmacdope: I have already started to come across this in my work, eg on #3718 so possibly good to come to some kind of general agreement on this.

package/MDAnalysis/core/_get_readers.py
```diff
@@ -22,13 +22,16 @@
...
+
 from .. import (_READERS, _READER_HINTS,
                 _PARSERS, _PARSER_HINTS,
                 _MULTIFRAME_WRITERS, _SINGLEFRAME_WRITERS, _CONVERTERS)
 from ..lib import util
+from typing import Optional, Union, Type
 
 
-def get_reader_for(filename, format=None):
+def get_reader_for(filename: str, format: Optional[Union[str, _READERS]] = None) -> _READERS:
```
jbarnoud: If you read the docstring, you will see that `filename` is a misleading name. It can be much more than a string.  It can be a string, but it can also be:
- a NumPy array,
- an MMTF object,
- a ParmEd object,
- an MDAnalysis reader,
- an iterator of everything that is authorised.

Actually, `filename` can be anything. What is passed there will go through a series of functions that will tell if we can do something with it or not. Because users can add such functions, we have no control over what is readable or not at the end user level.

This is one of the rare places where the actual type hint is `typing.Any`.

jbarnoud: `_READERS` is not a type. It is a dictionary of classes. What these classes have in common is that they should all inherit from `mda.coordinates.base.ProtoReader`.

package/MDAnalysis/core/_get_readers.py
```diff
@@ -107,7 +110,7 @@ def get_reader_for(filename, format=None):
         raise ValueError(errmsg) from None
 
 
-def get_writer_for(filename, format=None, multiframe=None):
+def get_writer_for(filename: Optional[str], format: Optional[str] = None, multiframe: Optional[bool] = None) -> Union[_SINGLEFRAME_WRITERS, _MULTIFRAME_WRITERS]:
```
jbarnoud: Here, the input types look correct. The return type, however, has the same issue as above. Neither `_SINGLEFRAME_WRITERS` not `_MULTIFRAME_WRITERS` are types. Instead, they are dicts of classes that inherit from `mda.coordinates.base.WriterBase`.

package/MDAnalysis/core/_get_readers.py
```diff
@@ -202,7 +205,7 @@ def get_writer_for(filename, format=None, multiframe=None):
         raise TypeError(errmsg.format(format)) from None
 
 
-def get_parser_for(filename, format=None):
+def get_parser_for(filename: Union[str, mmtf.MMTFDecoder], format: Optional[str] = None) -> _PARSERS:
```
jbarnoud: Like above, `_PARSERS` is not a type.

jbarnoud: This method is an example of a method where the docstring is out of date. I expect that you will encounter many of those. Fixing them would be of great benefit.

In this method, `filename` can be anything. Like in `get_reader_for`, the `filename` argument will be passed to a series of functions to figure out what parser is appropriate. The docstring does not reflect that and should.

The `format` argument can either be a string, in which case it would be the name of a format, or a class describing a parser.

package/MDAnalysis/core/_get_readers.py
```diff
@@ -257,7 +260,8 @@ def get_parser_for(filename, format=None):
         else:
             return _PARSERS['MINIMAL']
 
-def get_converter_for(format):
+
+def get_converter_for(format: str) -> Type[_CONVERTERS]:
```
jbarnoud: `_CONVERTERS` is not a class.

package/MDAnalysis/core/accessors.py
jbarnoud: Looks like it is the right solution, indeed. I was initially thinking about forward references, where you provide the type name as a string, but I am not sure it will work for imported stuff.

lilyminium: I've typically seen it done as `if TYPE_CHECKING` too.

package/MDAnalysis/core/accessors.py
```diff
@@ -97,7 +101,7 @@ class Accessor:
     .. versionadded:: 2.0.0
     """
 
-    def __init__(self, name, accessor):
+    def __init__(self, name: str, accessor: Accessor) -> None:
```
jbarnoud: This one looks tricky to annotate. I need to give it a deeper look, but I do not have the time today. I'll go back to it either tomorrow or on Wednesday.

jbarnoud (commented):
We use a bunch of library-level dictionaries that link a format name to a class (e.g. `_READERS`, `_PARSERS`). These cannot be used as type hints as they are not classes. The fact that mypy did not complain at any point makes me think that there are issues in the way we run the type analysis. How do you run mypy to test you changes to these files? Did you remove them from the exclusions?

The modules you are working on now are defining things that are called elsewhere. You need to run the analysis on places where these methods are called. The tests are likely a good place for that, actually.

On a different note. Could you please avoid to force push? When you force the a git push, you make it more difficult for us to review you code for 2 reasons:
- if we loaded your branch on our computer, your force push can create issues when we try to upgrade to your latest changes
- we do not see the evolution of your changes, so it is very difficult to know what are the last things you changed.

umak1106:
> We use a bunch of library-level dictionaries that link a format name to a class (e.g. `_READERS`, `_PARSERS`). These cannot be used as type hints as they are not classes. The fact that mypy did not complain at any point makes me think that there are issues in the way we run the type analysis. How do you run mypy to test you changes to these files? Did you remove them from the exclusions?
> 
> The modules you are working on now are defining things that are called elsewhere. You need to run the analysis on places where these methods are called. The tests are likely a good place for that, actually.
> 
> On a different note. Could you please avoid to force push? When you force the a git push, you make it more difficult for us to review you code for 2 reasons:
> 
>     * if we loaded your branch on our computer, your force push can create issues when we try to upgrade to your latest changes
> 
>     * we do not see the evolution of your changes, so it is very difficult to know what are the last things you changed.

I removed mypy.ini and mypy.yaml from my local repo and ran this command
mypy --follow-imports=skip _get_readers.py which is why it skipped those  errors

If i set --follow-imports=normal I do get errors regarding these .
_get_readers.py:25: error: Skipping analyzing "mmtf": module is installed, but missing library stubs or py.typed marker
_get_readers.py:25: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
_get_readers.py:34: error: Variable "MDAnalysis._READERS" is not valid as a type
_get_readers.py:34: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
_get_readers.py:113: error: Variable "MDAnalysis._SINGLEFRAME_WRITERS" is not valid as a type
etc..


I used to squash merge my commits for a cleaner commit history but i'll avoid force pushing now

jbarnoud:
I would suggest you run mypy on the tests, this is where the objects get used and therefore where you get to evaluate your annotations. Also, at least when running locally, I would avoid silencing the errors too much; that requires filtering the output when reading it, but it avoids missing important messages.

package/MDAnalysis/core/_get_readers.py
umak1106: _PARSERS are from TopologyReaderbase ??

PR #3726, opened 2022-06-22
Revert breaking change to OpenMP constructs for self_distance_array OpenMP backend with triclinic boxes (Revert #3706)

hmacdope:
Fixes #3725 

This PR reverts a PR that appears to have broken the OpenMP backend for triclinic boxes. 
Pending some discussion here or on the issue. 

Changes made in this Pull Request:
 - Reverts changes made in #3706 
 - TODO: add tests for triclinic code path in `self_distance_array`

PR Checklist
------------
 - [ ] Tests?
 - [ ] CHANGELOG updated?
 - [X] Issue raised/referenced?

PR #3728, opened 2022-06-22
Fix OpenMP

### LINKED ISSUE #3725: OpenMP backend for self_distance_array is broken for tricinic boxes
## Expected behavior ##

The OpenMP distance backend for `self_distance_array` is broken for triclinic boxes on current develop.
There doesn't appear to be a problem with orthorhombic boxes (the `_calc_self_distance_array_ortho` code path) or without a box ( the `calc_self_distance_array`) code path or any of the serial code. There also doesn't appear to be an issue with `distance_array`.

 The problem seems to be limited to `calc_self_distance_array_triclinic` only and isn't picked up by the `TestTriclinicDistances::test_selfdist`  test in `test_distances.py`

## Actual behavior ##

Using the OpenMP backend for `self_distance_array` with a triclinic box segfaults or returns rubbish.  

## Possible Fix

Reverting recent changes to the scoping of OpenMP constructs in commits `4f7087f99f63b5a8f837be894a1027b051b25ca7` and `33d119cc7e0e534993c7af40537768a1df5de81c` (PR #3706 )
 seems to fix the issue.  Perhaps a good place to start is by  reverting these two commits?

I am no OpenMP expert so I couldn't pin down the exact issue. 

``` python
import numpy as np
import MDAnalysis as mda
from MDAnalysis.lib.distances import self_distance_array
from MDAnalysis.tests.datafiles import TRIC

U_tric = mda.Universe(TRIC)
pos_tric = U_tric.coord.positions
box_tric = U_tric.coord.dimensions
self_da_serial_tric = self_distance_array(pos_tric, box=box_tric, backend='serial') # all good
self_da_openmp_tric = self_distance_array(pos_tric, box=box_tric, backend='openmp') #  <-- may segfault
np.testing.assert_allclose(self_da_serial_tric, self_da_openmp_tric) # nope
```

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) 2.3.0-dev0
- Which version of Python (`python -V`)? 3.9

RMeli:
Fixes #3725 . PR #3706 introduced the bug. Feel free to close this PR if reverting is preferred (#3726).

---

Lesson learned the hard way:

> if it ain't broke, don't fix it!

PR Checklist
------------
 - [x] Tests?


testsuite/MDAnalysisTests/lib/test_distances.py
```diff
@@ -1419,3 +1419,38 @@ def test_minimize_vectors(box, shift, dtype):
 
     assert_allclose(res, vec, atol=0.00001)
     assert res.dtype == dtype
+
+
+def test_issue_3725():
```
hmacdope: Sorry to be a PITA, but could you move this up to below `class TestTriclinicDistances`? :)

testsuite/MDAnalysisTests/lib/test_distances.py
```diff
@@ -1419,3 +1419,38 @@ def test_minimize_vectors(box, shift, dtype):
...
+        pos_tric, box=box_tric, backend='serial'
+    )
+    self_da_openmp_tric = distances.self_distance_array(
+        pos_tric, box=box_tric, backend='openmp'
+    )
+
+    np.testing.assert_allclose(self_da_serial_tric, self_da_openmp_tric)
+
+
+def test_DCD_serial_vs_omp(DCD_Universe):
```
hmacdope: Likewise with this, can we move this up to below `class TestSelfDistanceArrayDCD`?

hmacdope (changes requested):
Looks good! 
Just a small nitpick, would you be able to move the testcases around a bit so that they are closer to similar tests? Sorry to be a pain.

hmacdope:
Also CHANGELOG.

testsuite/MDAnalysisTests/lib/test_distances.py
```diff
@@ -574,6 +589,26 @@ def test_pbc_wrong_wassenaar_distance(self, backend):
         assert np.linalg.norm(point_a - point_b) != dist[0, 0]
 
 
+def test_issue_3725():
+    """
+    Code from @hmacdope
+    https://github.com/MDAnalysis/mdanalysis/issues/3725
+    """
+    U_tric = MDAnalysis.Universe(TRIC)
```
richardjgowers: Can we replace this with a seeded random numpy array? File IO isn't necessary here and above.

richardjgowers (dismissed):

testsuite/MDAnalysisTests/lib/test_distances.py
RMeli: No problem at all!

testsuite/MDAnalysisTests/lib/test_distances.py
RMeli: Good point. I did that and parametrized over the box, so that now all branches in `distances.self_distance_array()` are covered.

hmacdope (approved):
LGTM :)

PR #3729, opened 2022-06-23
Type hints for lib module

umak1106:
Fixes #

 - 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


umak1106:
Regarding the stp() function if i set the arguments as arraylike i get this error
Argument 1 to "cross" has incompatible type "Union[_SupportsArray[dtype[Any]], _NestedSequence[_SupportsArray[dtype[Any]]], bool, int, float, complex, str, bytes, _NestedSequence[Union[bool, int, float, complex, str, bytes]]]"; expected "Union[_SupportsArray[dtype[bool_]], _NestedSequence[_SupportsArray[dtype[bool_]]], bool, _NestedSequence[bool]]"

And if i set it as NDArray I still get an error
Argument 1 to "stp" has incompatible type "Union[_SupportsArray[dtype[Any]], _NestedSequence[_SupportsArray[dtype[Any]]], bool, int, float, complex, str, bytes, _NestedSequence[Union[bool, int, float, complex, str, bytes]]]"; expected "ndarray[Any, dtype[Any]]"

The same error above shows  for normal()

umak1106:
actually typing module is introduced in numpy from 1.20.0 and the gh-ci.yaml file has numpy = 1.19.0 which is why i think the tests are failing

jbarnoud:
@IAlibay plans on bumping the minimal version of numpy to 1.20.0. It means we'll have to wait for the bump to happen before we merge, but you can proceed and ignore the failing pipeline.

package/MDAnalysis/lib/mdamath.py
```diff
@@ -90,7 +92,7 @@ def norm(v):
     return np.sqrt(np.dot(v, v))
 
 
-def normal(vec1, vec2):
+def normal(vec1: npt.NDArray, vec2: npt.NDArray) -> npt.NDArray:
```
jbarnoud: Here, the inputs can be `ArrayLike`, I think.

The idea is to be loose on the input to accommodate everything that is valid, but precise on the output.

package/MDAnalysis/lib/mdamath.py
```diff
@@ -110,7 +112,7 @@ def normal(vec1, vec2):
     return normal / n
 
 
-def pdot(a, b):
+def pdot(a: npt.NDArray, b: npt.NDArray) -> npt.NDArray:
```
jbarnoud: Here too, I think the input can be `ArrayLike`. It should be true for other functions as well.

jbarnoud (commented):
Do you still have the error you described with `normal` and `stp`? I did not get them when testing locally.

I only have small comments about the typing itself. In general, we want to be broad when it comes to inputs and as specific as possible for the outputs. This is because we do not know what we will get as inputs and will accept anything that can work, but (most of the time) we know exactly what we will output.

You should remove the files you annotate from those ignored by mypy so that the CI actually runs mypy on the files you edit.

package/MDAnalysis/lib/mdamath.py
umak1106: Ok I'll make the changes

package/MDAnalysis/lib/mdamath.py
umak1106: Actually the doctoring said ndarray so I did that but I'll make the changes.
After making the changes I get error on that line

jbarnoud: Automatic tools may tell you that, indeed, because it is how it is use in the examples it seen. But if you read the code, you see that `a` and `b`  are passed to the `np.einsum`function that expects an `ArrayLike` there.

hmacdope:
@umak1106 just a quick note, I am adding type hints to a lot of `distances.py` as part of #3730, you can probably leave that one to me for now.

umak1106:
ok

package/MDAnalysis/__init__.py
```diff
@@ -167,18 +167,18 @@
...
 # Metaclass magic fills these as classes are declared.
-_READERS = {}
-_READER_HINTS = {}
-_SINGLEFRAME_WRITERS = {}
-_MULTIFRAME_WRITERS = {}
-_PARSERS = {}
-_PARSER_HINTS = {}
-_SELECTION_WRITERS = {}
-_CONVERTERS = {}
+_READERS = {}  # type: dict
```
jbarnoud: Is there a reason for the different style of annotation? If not, I'd rather have the style consistent:

```python
_READERS: Dict = {}
```

Also, the annotation can be more specific:

```python
_READERS: Dict[str, ProtoReader] = {}
```

umak1106: Ok I'll make the changes

umak1106: I made the changes like you suggested . But the tests are failing

jbarnoud: I looked at it and it looks like the issue is caused by the imports in `__init__.py`.

I tried two ways of fixing the problem:
* I guarded the import with `TYPE_CHECKING`, but then I needed to reference the type with forward references (using the name of the types as strings instead of just providing the types)
* providing the types in a stub file (see <https://mypy.readthedocs.io/en/stable/stubs.html>)

Both methods seem to work in the sense that nowhere mypy complains about not having annotations for these variables. However, I know the annotations are wrong, yet I do not get any errors about them. So I would suggest keeping it simple for now and just annotating them as `Dict`. It will not be as precise as it should, but at least it won't be wrong.

umak1106:
Errors am facing :
MDAnalysis/lib/pkdtree.py:141: error: Incompatible types in assignment (expression has type "Optional[float]", variable has type "None")
MDAnalysis/lib/pkdtree.py:184: error: Unsupported operand types for > ("float" and "None")
MDAnalysis/lib/pkdtree.py:234: error: Unsupported operand types for > ("float" and "None")
MDAnalysis/lib/pkdtree.py:290: error: Unsupported operand types for > ("float" and "None")

package/MDAnalysis/lib/pkdtree.py
```diff
@@ -202,7 +204,7 @@ def search(self, centers, radius):
         self._indices = np.asarray(unique_int_1d(self._indices))
         return self._indices
 
-    def get_indices(self):
+    def get_indices(self) -> npt.NDArray:
```
umak1106: according to the docstring self._indices should be a list but doesn't np.asarray return NDArray

package/MDAnalysis/lib/pkdtree.py
```diff
@@ -245,7 +247,7 @@ def search_pairs(self, radius):
             pairs = unique_rows(pairs)
         return pairs
 
-    def search_tree(self, centers, radius):
+    def search_tree(self, centers: npt.ArrayLike, radius: float) -> npt.NDArray:
```
umak1106: Same as above

package/MDAnalysis/lib/pkdtree.py
jbarnoud: Looks like the docstring is out of date, indeed. Could you fix it?

package/MDAnalysis/lib/pkdtree.py
jbarnoud: Here the docstring and the code seem to match, don't they? `pair` is an array is I recall correctly.

package/MDAnalysis/lib/pkdtree.py
```diff
@@ -95,7 +97,7 @@ def pbc(self):
         """
         return self.box is not None
 
-    def set_coords(self, coords, cutoff=None):
+    def set_coords(self, coords: npt.ArrayLike, cutoff: Union[float, None] = None) -> None:
```
jbarnoud:
```suggestion
    def set_coords(self, coords: npt.ArrayLike, cutoff: Optional[float] = None) -> None:
```

jbarnoud (commented):
Regarding the errors you report:
* for the first one, you probably should type the instance attribute. See https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#classes
* for the other errors, we assume that cutoff is set if pbc is True. It holds with the logic of how we expect the class to be used, but it can break, which is what mypy is warning us about. You may need to add a few checks, here and there, to make sure that cutoff is not None when pbc is True.

package/MDAnalysis/lib/pkdtree.py
umak1106: Yes I misread . The doctoring mentions array .

package/MDAnalysis/lib/pkdtree.py
```diff
@@ -176,7 +178,8 @@ def search(self, centers, radius):
         centers = np.asarray(centers)
         if centers.shape == (self.dim, ):
             centers = centers.reshape((1, self.dim))
-
+        if self.cutoff is None:
```
jbarnoud: `self.cutoff` can be `None` is `self.pbc` is `False`. It is only if `self.pbc == True` that we assume we have a value for the cutoff.

package/MDAnalysis/lib/pkdtree.py
```diff
@@ -176,7 +178,8 @@ def search(self, centers, radius):
...
+            raise RuntimeError("Provide cutoff distance")
```
jbarnoud: I *think* a `ValueError` would be more appropriate here.

jbarnoud (changes requested):
Last few small things:

* there is a small misunderstanding with the test about cutoff. I explained this in a comment at one of the places you added the test but it is true for all the places where you have that test.
* there seems to be an issue with using `npt.NDArray` and numpy 1.20.0. See #3748. For the time being, can you use `np.ndarray` instead so we can move forward?

package/MDAnalysis/lib/pkdtree.py
umak1106: If i use 
if self.pbc and (self.cutoff is None) :
raise Valuerror("Provide cutoff")

mypy still raises this error 
Unsupported operand types for > ("float" and "None")
Left operand is of type "Optional[float]"

jbarnoud: Can you push the code with the conditional moved? This way I can comment on the code you have rather than the code I imagine you may have. Also, it will let us see if the tests, other than mypy, pass.

jbarnoud:
It looks like you are still using `npt.NDArray` in pkdtree.

umak1106:
> It looks like you are still using `npt.NDArray` in pkdtree.

Yes actually after i faced mypy errors there i didn't push my code

package/MDAnalysis/lib/pkdtree.py
jbarnoud: This remains outstanding. It looks like it is the latest obstacle in terms of type annotations.

jbarnoud: I did not realise you did push your latest change. Here is a possible solution. I did not test it, but it might work:

```python
if self.pbc:
    if self.cutoff is None:
        raise ValueError("Cutoff needs to be provided when working with PBC.")
    cutoff: float = self.cutoff
    # From here and until the end if the if self.pbc block, use cutoff instead of self.cutoff
```

I haven't tested that solution, but I expect that mypy will figure out that `cutoff` can only be a float.

umak1106: I tried that but it still raises the same error

jbarnoud: I am confused. I tried on a small example and it works perfectly fine:

```python
from typing import Optional

def test(pbc: bool, cutoff: Optional[float]):
    if pbc:
        #if cutoff is None:
        #    raise ValueError
        if cutoff < 3:
            print('Hello')
    print('Good bye')

test(False, 2)  # OK, cutoff is just ignored
test(False, None)  # OK
test(True, 2)  # OK
test(True, None)  # Not OK
```

With the `if cutoff is None` commented, I get the same mypy error:

```
test.py:7: error: Unsupported operand types for > ("int" and "None")
test.py:7: note: Left operand is of type "Optional[float]"
```

If it is uncommented, I do not get any errors.

umak1106: Done

package/MDAnalysis/lib/pkdtree.py
```diff
@@ -127,22 +130,22 @@ def set_coords(self, coords, cutoff=None):
...
-            raise RuntimeError('Provide a cutoff distance'
-                               ' with tree.set_coords(...)')
 
         # set coords dtype to float32
         # augment coordinates will work only with float32
         coords = np.asarray(coords, dtype=np.float32)
 
         if self.pbc:
-            self.cutoff = cutoff
+            if self.cutoff is None:
```
jbarnoud: Here, the original test is not about `self.cutoff`, it is instead about `cutoff` as provided as an argument to the function. This seems to be why the tests fail.

```suggestion
            if cutoff is None:
```

It is subtle and got me puzzled for a while!

package/MDAnalysis/lib/pkdtree.py
```diff
@@ -127,22 +130,22 @@ def set_coords(self, coords, cutoff=None):
 
         """
         # If no cutoff distance is provided but PBC aware
```
jbarnoud: Move the comment with the code. Now we have an orphan comment that will confuse future contributors and maintainers.

package/MDAnalysis/lib/pkdtree.py
```diff
@@ -127,22 +130,22 @@ def set_coords(self, coords, cutoff=None):
...
 
         # set coords dtype to float32
         # augment coordinates will work only with float32
         coords = np.asarray(coords, dtype=np.float32)
 
         if self.pbc:
-            self.cutoff = cutoff
+            if self.cutoff is None:
+                raise ValueError("Cutoff needs to be provided "
+                                 "when working with PBC.")
```
jbarnoud: In this specific instance, because you move an existing exception raising, it would be good to raise the same exception as what you removed:

```suggestion
                raise RuntimeError('Provide a cutoff distance'
                               ' with tree.set_coords(...)'
```

package/MDAnalysis/lib/pkdtree.py
```diff
@@ -127,22 +130,22 @@ def set_coords(self, coords, cutoff=None):
...
-        if self.pbc and (cutoff is None):
-            raise RuntimeError('Provide a cutoff distance'
-                               ' with tree.set_coords(...)')
 
         # set coords dtype to float32
         # augment coordinates will work only with float32
         coords = np.asarray(coords, dtype=np.float32)
 
         if self.pbc:
-            self.cutoff = cutoff
```
jbarnoud, line 139: This line is extremely important and cannot be removed! This line actually sets the cutoff in the object; without it, `self.cutoff` will always be `None`. The line needs to be run if the exception about the cutoff is not raised.

package/MDAnalysis/lib/NeighborSearch.py
```diff
@@ -102,7 +109,7 @@ def search(self, atoms, radius, level='A'):
             unique_idx = unique_int_1d(np.asarray(pairs[:, 1], dtype=np.intp))
         return self._index2level(unique_idx, level)
 
-    def _index2level(self, indices, level):
+    def _index2level(self, indices: List[int], level: str) -> None:
```
jbarnoud:
```suggestion
    def _index2level(self, indices: List[int], level: str) -> Union[AtomGroup, ResidueGroup, SegmentGroup]:
```

jbarnoud (changes requested):
I made some suggestions to fix errors that caused either the tests to fails or mypy to complain. After that, all the complaints from mypy are about `distances.py` that have wrong annotations. I'll try and fix them.

jbarnoud:
Now that #3787 is merged into develop, the develop branch should be merged in here.

jbarnoud:
Wooo! 🎉 Only one mypy error left! With the last few comment addressed we should be good to go 😁

umak1106:
Making those changes in pkdtree raises errors like unresolved reference to self

jbarnoud:
> Making those changes in pkdtree raises errors like unresolved reference to self

Can you paste the full error message please? It is difficult to figure out what can go wrong without it.

package/MDAnalysis/__init__.py
```diff
@@ -167,18 +169,18 @@
...
+_PARSER_HINTS: Dict = {}
+_SELECTION_WRITERS: Dict = {}
+_CONVERTERS: Dict = {}
 # Registry of TopologyAttributes
-_TOPOLOGY_ATTRS = {}  # {attrname: cls}
-_TOPOLOGY_TRANSPLANTS = {}   # {name: [attrname, method, transplant class]}
-_TOPOLOGY_ATTRNAMES = {}  # {lower case name w/o _ : name}
+_TOPOLOGY_ATTRS: Dict = {}   # {attrname: cls}
+_TOPOLOGY_TRANSPLANTS: Dict = {}   # {name: [attrname, method, transplant class]}
+_TOPOLOGY_ATTRNAMES: Dict = {}   # {lower case name w/o _ : name}
```
jbarnoud: @umak1106 It would be great to have these lines in a separate PR so we can work at annotating other parts of the code without triggering errors here.

umak1106: ok mentor.
 #3823 this is the pr .

package/MDAnalysis/lib/pkdtree.py
umak1106: Done

umak1106:
> > Making those changes in pkdtree raises errors like unresolved reference to self
> 
> Can you paste the full error message please? It is difficult to figure out what can go wrong without it.

It got fixed . There are no mypy errors in this module now .

jbarnoud (approved):
Youhou! :tada:

jbarnoud:
Congratulations on that PR being merged!

umak1106:
> Congratulations on that PR being merged!

Thanks

PR #3730, opened 2022-06-23
Allow distance routines to accept an NumPy Array or AtomGroup [Core]

### LINKED ISSUE #3708: Provide Ndarray or AtomGroup to distance routines

hmacdope:
Fixes #3708 
Successor to #3718 

Second attempt at addressing #3708 using a python only design. CZI_EOSS4 performance track. 

- Change `@check_coords` to take also `AtomGroups` and return positions
-  Atomgroup parsing in `@check_coords` must be enabled with argument `allow_atomgroup` with default `False` maintaining current behaviour
- Added type hinting to lots of distances module
- Change `distance_array`, `self_distance_array`, `calc_bonds`, `calc_angles`, `calc_dihedrals`, and `apply_PBC` to take an atomgroup
- Added tests for AtomGroup input to distances.  Still some code duplication here can be possibly cleaned up .

 Do we want to also change `nsgrid` functions,   and the `transform_` functions? If so is that best done here or separate PR?

PR Checklist
------------
 - [X] Tests?
 - [X] Docs?
 - [X] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/lib/distances.py
+++ b/package/MDAnalysis/lib/distances.py
@@ -73,7 +73,6 @@ Functions
 """
 import numpy as np
 from numpy.lib.utils import deprecate
-import numpy.typing as npt
 
 from typing import Union, Optional, Callable
 from typing import TYPE_CHECKING
@@ -140,7 +139,7 @@ from .c_distances import (_UINT64_MAX,
 from .c_distances_openmp import OPENMP_ENABLED as USED_OPENMP
 
 
-def _check_result_array(result: npt.NDArray, shape: tuple) -> npt.NDArray:
+def _check_result_array(result: np.ndarray, shape: tuple) -> np.ndarray:
     """Check if the result array is ok to use.
 
     The `result` array must meet the following requirements:
@@ -183,11 +182,11 @@ def _check_result_array(result: npt.NDArray, shape: tuple) -> npt.NDArray:
 
 @check_coords('reference', 'configuration', reduce_result_if_single=False,
               check_lengths_match=False, allow_atomgroup=True)
-def distance_array(reference: Union[npt.NDArray, 'AtomGroup'],
-                   configuration: Union[npt.NDArray, 'AtomGroup'],
-                   box: Optional[npt.NDArray] = None,
-                   result: Optional[npt.NDArray] = None,
-                   backend: str = "serial") -> npt.NDArray:
+def distance_array(reference: Union[np.ndarray, 'AtomGroup'],
+                   configuration: Union[np.ndarray, 'AtomGroup'],
+                   box: Optional[np.ndarray] = None,
+                   result: Optional[np.ndarray] = None,
+                   backend: str = "serial") -> np.ndarray:
     """Calculate all possible distances between a reference set and another
     configuration.
 
@@ -271,10 +270,10 @@ def distance_array(reference: Union[npt.NDArray, 'AtomGroup'],
 
 
 @check_coords('reference', reduce_result_if_single=False, allow_atomgroup=True)
-def self_distance_array(reference: Union[npt.NDArray, 'AtomGroup'],
-                        box: Optional[npt.NDArray] = None,
-                        result: Optional[npt.NDArray] = None,
-                        backend: str = "serial") -> npt.NDArray:
+def self_distance_array(reference: Union[np.ndarray, 'AtomGroup'],
+                        box: Optional[np.ndarray] = None,
+                        result: Optional[np.ndarray] = None,
+                        backend: str = "serial") -> np.ndarray:
     """Calculate all possible distances within a configuration `reference`.
 
     If the optional argument `box` is supplied, the minimum image convention is
@@ -1268,11 +1267,11 @@ def transform_StoR(coords, box, backend="serial"):
 
 
 @check_coords('coords1', 'coords2', allow_atomgroup=True)
-def calc_bonds(coords1: Union[npt.NDArray, 'AtomGroup'],
-               coords2: Union[npt.NDArray, 'AtomGroup'],
-               box: Optional[npt.NDArray] = None,
-               result: Optional[npt.NDArray] = None,
-               backend: str = "serial") -> npt.NDArray:
+def calc_bonds(coords1: Union[np.ndarray, 'AtomGroup'],
+               coords2: Union[np.ndarray, 'AtomGroup'],
+               box: Optional[np.ndarray] = None,
+               result: Optional[np.ndarray] = None,
+               backend: str = "serial") -> np.ndarray:
     """Calculates the bond lengths between pairs of atom positions from the two
     coordinate arrays `coords1` and `coords2`, which must contain the same
     number of coordinates. ``coords1[i]`` and ``coords2[i]`` represent the
@@ -1359,12 +1358,12 @@ def calc_bonds(coords1: Union[npt.NDArray, 'AtomGroup'],
 
 
 @check_coords('coords1', 'coords2', 'coords3', allow_atomgroup=True)
-def calc_angles(coords1: Union[npt.NDArray, 'AtomGroup'],
-                coords2: Union[npt.NDArray, 'AtomGroup'],
-                coords3: Union[npt.NDArray, 'AtomGroup'],
-                box: Optional[npt.NDArray] = None,
-                result: Optional[npt.NDArray] = None,
-                backend: str = "serial") -> npt.NDArray:
+def calc_angles(coords1: Union[np.ndarray, 'AtomGroup'],
+                coords2: Union[np.ndarray, 'AtomGroup'],
+                coords3: Union[np.ndarray, 'AtomGroup'],
+                box: Optional[np.ndarray] = None,
+                result: Optional[np.ndarray] = None,
+                backend: str = "serial") -> np.ndarray:
     """Calculates the angles formed between triplets of atom positions from the
     three coordinate arrays `coords1`, `coords2`, and `coords3`. All coordinate
     arrays must contain the same number of coordinates.
@@ -1461,13 +1460,13 @@ def calc_angles(coords1: Union[npt.NDArray, 'AtomGroup'],
 
 
 @check_coords('coords1', 'coords2', 'coords3', 'coords4', allow_atomgroup=True)
-def calc_dihedrals(coords1: Union[npt.NDArray, 'AtomGroup'],
-                   coords2: Union[npt.NDArray, 'AtomGroup'],
-                   coords3: Union[npt.NDArray, 'AtomGroup'],
-                   coords4: Union[npt.NDArray, 'AtomGroup'],
-                   box: Optional[npt.NDArray] = None,
-                   result: Optional[npt.NDArray] = None,
-                   backend: str = "serial") -> npt.NDArray:
+def calc_dihedrals(coords1: Union[np.ndarray, 'AtomGroup'],
+                   coords2: Union[np.ndarray, 'AtomGroup'],
+                   coords3: Union[np.ndarray, 'AtomGroup'],
+                   coords4: Union[np.ndarray, 'AtomGroup'],
+                   box: Optional[np.ndarray] = None,
+                   result: Optional[np.ndarray] = None,
+                   backend: str = "serial") -> np.ndarray:
     r"""Calculates the dihedral angles formed between quadruplets of positions
     from the four coordinate arrays `coords1`, `coords2`, `coords3`, and
     `coords4`, which must contain the same number of coordinates.
@@ -1578,9 +1577,9 @@ def calc_dihedrals(coords1: Union[npt.NDArray, 'AtomGroup'],
 
 
 @check_coords('coords', allow_atomgroup=True)
-def apply_PBC(coords: Union[npt.NDArray, 'AtomGroup'],
-              box: Optional[npt.NDArray] = None,
-              backend: str = "serial") -> npt.NDArray:
+def apply_PBC(coords: Union[np.ndarray, 'AtomGroup'],
+              box: Optional[np.ndarray] = None,
+              backend: str = "serial") -> np.ndarray:
     """Moves coordinates into the primary unit cell.
 
     Parameters
@@ -1626,7 +1625,7 @@ def apply_PBC(coords: Union[npt.NDArray, 'AtomGroup'],
 
 
 @check_coords('vectors', enforce_copy=False, enforce_dtype=False)
-def minimize_vectors(vectors: npt.NDArray, box: npt.NDArray) -> npt.NDArray:
+def minimize_vectors(vectors: np.ndarray, box: np.ndarray) -> np.ndarray:
     """Apply minimum image convention to an array of vectors
 
     This function is required for calculating the correct vectors between two
--- a/package/MDAnalysis/lib/distances.py
+++ b/package/MDAnalysis/lib/distances.py
@@ -1491,22 +1491,26 @@ def calc_dihedrals(coords1: Union[np.ndarray, AtomGroup],
 
     Parameters
     ----------
-    coords1 : numpy.ndarray
+    coords1 : numpy.ndarray or :class:`~MDAnalysis.core.groups.AtomGroup`
         Coordinate array of shape ``(3,)`` or ``(n, 3)`` containing the 1st
         positions in dihedrals (dtype is arbitrary, will be converted to
-        ``numpy.float32`` internally)
-    coords2 : numpy.ndarray
+        ``numpy.float32`` internally).  Also accepts an 
+        :class:`~MDAnalysis.core.groups.AtomGroup`.
+    coords2 : numpy.ndarray or :class:`~MDAnalysis.core.groups.AtomGroup`
         Coordinate array of shape ``(3,)`` or ``(n, 3)`` containing the 2nd
         positions in dihedrals (dtype is arbitrary, will be converted to
-        ``numpy.float32`` internally)
-    coords3 : numpy.ndarray
+        ``numpy.float32`` internally).  Also accepts an 
+        :class:`~MDAnalysis.core.groups.AtomGroup`.
+    coords3 : numpy.ndarray or :class:`~MDAnalysis.core.groups.AtomGroup`
         Coordinate array of shape ``(3,)`` or ``(n, 3)`` containing the 3rd
         positions in dihedrals (dtype is arbitrary, will be converted to
-        ``numpy.float32`` internally)
-    coords4 : numpy.ndarray
+        ``numpy.float32`` internally).  Also accepts an 
+        :class:`~MDAnalysis.core.groups.AtomGroup`.
+    coords4 : numpy.ndarray or :class:`~MDAnalysis.core.groups.AtomGroup`
         Coordinate array of shape ``(3,)`` or ``(n, 3)`` containing the 4th
         positions in dihedrals (dtype is arbitrary, will be converted to
-        ``numpy.float32`` internally)
+        ``numpy.float32`` internally).  Also accepts an 
+        :class:`~MDAnalysis.core.groups.AtomGroup`.
     box : numpy.ndarray, optional
         The unitcell dimensions of the system, which can be orthogonal or
         triclinic and must be provided in the same format as returned by
--- a/package/MDAnalysis/lib/distances.py
+++ b/package/MDAnalysis/lib/distances.py
@@ -70,7 +70,7 @@ Functions
 import numpy as np
 from numpy.lib.utils import deprecate
 
-from typing import Union, Optional
+from typing import Union, Optional, Callable
 from ..core.groups import AtomGroup
 from .util import check_coords, check_box
 from .mdamath import triclinic_vectors
@@ -93,7 +93,8 @@ except ImportError:
     pass
 del importlib
 
-def _run(funcname, args=None, kwargs=None, backend="serial"):
+def _run(funcname: Callable, args: Optional[tuple] = None,
+         kwargs: Optional[dict] = None, backend: str = "serial"):
     """Helper function to select a backend function `funcname`."""
     args = args if args is not None else tuple()
     kwargs = kwargs if kwargs is not None else dict()
@@ -131,7 +132,7 @@ from .c_distances import (_UINT64_MAX,
 from .c_distances_openmp import OPENMP_ENABLED as USED_OPENMP
 
 
-def _check_result_array(result, shape):
+def _check_result_array(result: np.ndarray, shape: tuple):
     """Check if the result array is ok to use.
 
     The `result` array must meet the following requirements:
@@ -195,10 +196,10 @@ def distance_array(reference: Union[np.ndarray, AtomGroup],
 
     Parameters
     ----------
-    reference : numpy.ndarray
+    reference :numpy.ndarray or :class:`~MDAnalysis.core.groups.AtomGroup`
         Reference coordinate array of shape ``(3,)`` or ``(n, 3)`` (dtype is
         arbitrary, will be converted to ``numpy.float32`` internally).
-    configuration : numpy.ndarray
+    configuration : numpy.ndarray or :class:`~MDAnalysis.core.groups.AtomGroup`
         Configuration coordinate array of shape ``(3,)`` or ``(m, 3)`` (dtype is
         arbitrary, will be converted to ``numpy.float32`` internally).
     box : array_like, optional
@@ -275,7 +276,7 @@ def self_distance_array(reference: Union[np.ndarray, AtomGroup],
 
     Parameters
     ----------
-    reference : numpy.ndarray
+    reference : numpy.ndarray or :class:`~MDAnalysis.core.groups.AtomGroup`
         Reference coordinate array of shape ``(3,)`` or ``(n, 3)`` (dtype is
         arbitrary, will be converted to ``numpy.float32`` internally).
     box : array_like, optional
--- a/package/MDAnalysis/lib/distances.py
+++ b/package/MDAnalysis/lib/distances.py
@@ -77,7 +77,7 @@ import numpy.typing as npt
 
 from typing import Union, Optional, Callable
 from typing import TYPE_CHECKING
-if TYPE_CHECKING:
+if TYPE_CHECKING: # pragma: no cover
     from ..core.groups import AtomGroup
 from .util import check_coords, check_box
 from .mdamath import triclinic_vectors
--- a/package/MDAnalysis/lib/distances.py
+++ b/package/MDAnalysis/lib/distances.py
@@ -176,7 +176,8 @@ def _check_result_array(result, shape):
               check_lengths_match=False)
 def distance_array(reference: Union[np.ndarray, AtomGroup],
                    configuration: Union[np.ndarray, AtomGroup],
-                   box: Optional[np.ndarray] = None, result=None,
+                   box: Optional[np.ndarray] = None,
+                   result: Optional[np.ndarray] = None,
                    backend: str = "serial") -> None:
     """Calculate all possible distances between a reference set and another
     configuration.
@@ -225,6 +226,8 @@ def distance_array(reference: Union[np.ndarray, AtomGroup],
     .. versionchanged:: 0.19.0
        Internal dtype conversion of input coordinates to ``numpy.float32``.
        Now also accepts single coordinates as input.
+    .. versionchanged:: 2.3.0
+       Now accepts AtomGroups as input and checks types using type hinting.
     """
     confnum = configuration.shape[0]
     refnum = reference.shape[0]
@@ -256,7 +259,10 @@ def distance_array(reference: Union[np.ndarray, AtomGroup],
 
 
 @check_coords('reference', reduce_result_if_single=False)
-def self_distance_array(reference, box=None, result=None, backend="serial"):
+def self_distance_array(reference: Union[np.ndarray, AtomGroup],
+                        box: Optional[np.ndarray] = None,
+                        result: Optional[np.ndarray] = None,
+                        backend: str = "serial") -> None:
     """Calculate all possible distances within a configuration `reference`.
 
     If the optional argument `box` is supplied, the minimum image convention is
@@ -302,6 +308,8 @@ def self_distance_array(reference, box=None, result=None, backend="serial"):
        Added *backend* keyword.
     .. versionchanged:: 0.19.0
        Internal dtype conversion of input coordinates to ``numpy.float32``.
+    .. versionchanged:: 2.3.0
+       Now accepts AtomGroups as input and checks types using type hinting.
     """
     refnum = reference.shape[0]
     distnum = refnum * (refnum - 1) // 2
--- a/package/MDAnalysis/lib/distances.py
+++ b/package/MDAnalysis/lib/distances.py
@@ -1255,7 +1255,11 @@ def transform_StoR(coords, box, backend="serial"):
 
 
 @check_coords('coords1', 'coords2')
-def calc_bonds(coords1, coords2, box=None, result=None, backend="serial"):
+def calc_bonds(coords1: Union[np.ndarray, AtomGroup],
+               coords2: Union[np.ndarray, AtomGroup],
+               box: Optional[np.ndarray] = None,
+               result: Optional[np.ndarray] = None,
+               backend: str = "serial") -> None:
     """Calculates the bond lengths between pairs of atom positions from the two
     coordinate arrays `coords1` and `coords2`, which must contain the same
     number of coordinates. ``coords1[i]`` and ``coords2[i]`` represent the
@@ -1337,8 +1341,12 @@ def calc_bonds(coords1, coords2, box=None, result=None, backend="serial"):
 
 
 @check_coords('coords1', 'coords2', 'coords3')
-def calc_angles(coords1, coords2, coords3, box=None, result=None,
-                backend="serial"):
+def calc_angles(coords1: Union[np.ndarray, AtomGroup],
+                coords2: Union[np.ndarray, AtomGroup],
+                coords3: Union[np.ndarray, AtomGroup],
+                box: Optional[np.ndarray] = None,
+                result: Optional[np.ndarray] = None,
+                backend: str = "serial") -> None:
     """Calculates the angles formed between triplets of atom positions from the
     three coordinate arrays `coords1`, `coords2`, and `coords3`. All coordinate
     arrays must contain the same number of coordinates.
@@ -1429,8 +1437,13 @@ def calc_angles(coords1, coords2, coords3, box=None, result=None,
 
 
 @check_coords('coords1', 'coords2', 'coords3', 'coords4')
-def calc_dihedrals(coords1, coords2, coords3, coords4, box=None, result=None,
-                   backend="serial"):
+def calc_dihedrals(coords1: Union[np.ndarray, AtomGroup],
+                   coords2: Union[np.ndarray, AtomGroup],
+                   coords3: Union[np.ndarray, AtomGroup],
+                   coords4: Union[np.ndarray, AtomGroup],
+                   box: Optional[np.ndarray] = None,
+                   result: Optional[np.ndarray] = None,
+                   backend: str = "serial") -> None:
     r"""Calculates the dihedral angles formed between quadruplets of positions
     from the four coordinate arrays `coords1`, `coords2`, `coords3`, and
     `coords4`, which must contain the same number of coordinates.
--- a/package/MDAnalysis/lib/distances.py
+++ b/package/MDAnalysis/lib/distances.py
@@ -178,7 +178,7 @@ def _check_result_array(result: np.ndarray, shape: tuple) -> np.ndarray:
 
 
 @check_coords('reference', 'configuration', reduce_result_if_single=False,
-              check_lengths_match=False)
+              check_lengths_match=False, allow_atomgroup=True)
 def distance_array(reference: Union[np.ndarray, AtomGroup],
                    configuration: Union[np.ndarray, AtomGroup],
                    box: Optional[np.ndarray] = None,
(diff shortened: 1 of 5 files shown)


testsuite/MDAnalysisTests/lib/test_distances.py
```diff
@@ -424,6 +545,51 @@ def test_periodic(self, DCD_Universe, backend):
...
+        assert_allclose(d_ag, d_arr,
+                        err_msg="AtomGroup and NumPy distances do not match")
+
+    # check triclinic box and some slices
+    @pytest.mark.parametrize("sel, np_slice", [
+                            ("index 0 to 8 ", np.s_[0:9,:]),
+                            ("index 9", np.s_[8,:])])
+    def test_atomgroup_matches_numpy_tric(self, Triclinic_Universe, backend, sel, np_slice):
+        U, _ = Triclinic_Universe
+        #BUG serial only for now as the OMP code path appears broken
```
hmacdope: Dependent on #3728

testsuite/MDAnalysisTests/lib/test_distances.py
```diff
@@ -640,6 +816,25 @@ def test_bonds(self, positions, box, backend, dtype):
         assert_almost_equal(dists_pbc[3], 3.46410072, self.prec,
                             err_msg="PBC check #w with box")
 
+    def test_bonds_atomgroup(self, positions_atomgroups, box, backend, ):
```
hmacdope: code duplication, should be done with parametrize and some overloads, will fix.

testsuite/MDAnalysisTests/lib/test_distances.py
```diff
@@ -697,6 +899,20 @@ def test_angles(self, positions, backend, dtype):
         assert_almost_equal(angles[3], 0.098174833, self.prec,
                             err_msg="Small angle failed in calc_angles")
 
+    def test_angles_atomgroup(self, positions_atomgroups, backend):
```
hmacdope: As above

testsuite/MDAnalysisTests/lib/test_distances.py
```diff
@@ -740,6 +956,17 @@ def test_dihedrals(self, positions, backend, dtype):
         assert_almost_equal(dihedrals[3], -0.50714064, self.prec,
                             err_msg="arbitrary dihedral angle failed")
 
+    def test_dihedrals_atomgroup(self, positions_atomgroups, backend):
```
hmacdope: as above

richardjgowers:
Not sure on nsgrid and transforms. Maybe keep this PR as is and think about those

hmacdope:
I think this is ready for review now that I have sorted out code duplication in the tests.  The functions changed are:

- `distance_array`
-  `self_distance_array`
-  `calc_bonds`
-  `calc_angles`
-  `calc_dihedrals`
-  `apply_PBC` 

As per Richard's comment I have left `nsgrid` and the `transform_*` functions for now. 

I have fixed the PEP8 issues that I agree with, the remaining ones are visually ugly, but if people feel strongly I can change them.

package/MDAnalysis/lib/distances.py
```diff
@@ -129,7 +137,7 @@ def _run(funcname, args=None, kwargs=None, backend="serial"):
 from .c_distances_openmp import OPENMP_ENABLED as USED_OPENMP
 
 
-def _check_result_array(result, shape):
+def _check_result_array(result: np.ndarray, shape: tuple) -> np.ndarray:
```
jbarnoud, line 143: Because the minimum version of numpy is about to bump to 1.20.0, we went for `numpy.typing` to type numpy:

```python
import numpy.typing as not
from typing import Tuple

def _check_result_array(result: npt.NDArray, shape: Tuple) -> npt.NDArray:
```

package/MDAnalysis/lib/util.py
```diff
@@ -214,6 +214,8 @@
 from .picklable_file_io import pickle_open, bz2_pickle_open, gzip_pickle_open
 
 from ..exceptions import StreamWarning, DuplicateWarning
+import MDAnalysis as mda
```
jbarnoud: Since you only use the `AtomGroup` from this import, I would suggest you only import that class. I am concerned about import conflicts down the line. We used to have rules about where you can import top-level; I don't remember them, though :roll_eyes:

orbeckst: Good point, the rule was that nothing in lib should import anything from MDA core... so in that sense the whole PR goes against that idea.

hmacdope: This is a circular import unless done at the top level. You can always use a top level import to get around the circular-ness but whether it is good practice is another question. 

The circular import occurs between `lib.util` and `core.groups`. I think this can be fixed by moving `@check_coords ` to its own module in `lib` because `groups.py` imports `lib.util` but doesn't require `check_coords`. Would people be happy with this solution?

Or we can hit it with a hammer and use a top level import. Or we can look into a different solution. Let me know what you think.

package/MDAnalysis/lib/util.py
```diff
@@ -2051,32 +2071,46 @@ def check_coords_decorator(func):
...
+                        coords = coords.astype(
+                            np.float32, order='C', copy=enforce_copy)
+                    except ValueError:
+                        errmsg = (f"{fname}(): {argname}.dtype must be"
+                                  f"convertible to float32, got"
+                                  f" {coords.dtype}.")
+                        raise TypeError(errmsg) from None
+                # coordinates should now be the right shape
+                ncoord = coords.shape[0]
+            elif isinstance(coords, mda.core.groups.AtomGroup):
```
orbeckst: If this is the only place where AtomGroup is needed then maybe duck-typing is good enough --- basically instead of doing "isinstance()" do the below in a try/except?

package/MDAnalysis/lib/util.py
orbeckst: I find the mda import here very weird and out of place, given what else is `util`. 

I'd say, if the import is required then rather move `check_coordinates()` and anything related to a separate module. (Presumably, you'd leave the old code here with a deprecation and create the new code with more functionality elsewhere until we can remove it in 3.0.). 

Although I am making a suggestion to use duck-typing in `check_coordinates()`, it might not a bad idea to just break up `util` anyway — it's become a big kitchen sink!

package/MDAnalysis/lib/util.py
jbarnoud: That would solve the problem, I think. Using `isinstance` is a bad habit anyway.

hmacdope: This seems like a good solution that keeps the separation of `lib` and `core` without moving stuff around. Thanks both.

hmacdope:
Implemented your solution @orbeckst :)

package/MDAnalysis/lib/distances.py
```diff
@@ -69,7 +73,12 @@
 """
 import numpy as np
 from numpy.lib.utils import deprecate
+import numpy.typing as npt
 
+from typing import Union, Optional, Callable
+from typing import TYPE_CHECKING
+if TYPE_CHECKING:
+    from ..core.groups import AtomGroup
```
orbeckst, line 80: Add a `#pragma: no coverage` (or whatever excludes it from coverage) as this looks as some code that does not need testing/coverage — unless you tell me that it ought to have been executed!

package/MDAnalysis/lib/distances.py
```diff
@@ -1323,9 +1358,13 @@ def calc_bonds(coords1, coords2, box=None, result=None, backend="serial"):
     return bondlengths
 
 
-@check_coords('coords1', 'coords2', 'coords3')
-def calc_angles(coords1, coords2, coords3, box=None, result=None,
-                backend="serial"):
+@check_coords('coords1', 'coords2', 'coords3', allow_atomgroup=True)
+def calc_angles(coords1: Union[npt.NDArray, 'AtomGroup'],
+                coords2: Union[npt.NDArray, 'AtomGroup'],
+                coords3: Union[npt.NDArray, 'AtomGroup'],
```
orbeckst: I don't know types well so I am asking for my own education: why is "AtomGroup" a string here? Is this how you indicate a custom class?

I read in https://docs.python.org/3/library/typing.html#typing.Union that the preferred syntax is
```reST
coords1: npt.NDArray | 'AtomGroup'
```
instead of the explicit `Union()`. Do we have a style guide (maybe as part of the Outreachy project with @umak1106 / @jbarnoud )?

package/MDAnalysis/lib/util.py
```diff
@@ -2014,12 +2024,20 @@ def check_coords(*coord_names, **options):
     >>> coordsum(np.zeros(3), np.ones(6)[::2])
     array([1., 1., 1.], dtype=float32)
     >>>
+    >>> # automatic handling of AtomGroups
+    >>> u = mda.Universe(PSF,DCD)
```
orbeckst: add space after comma

package/MDAnalysis/lib/util.py
```diff
@@ -2051,32 +2070,54 @@ def check_coords_decorator(func):
...
-                except ValueError:
-                    errmsg = (f"{fname}(): {argname}.dtype must be convertible to "
-                              f"float32, got {coords.dtype}.")
-                    raise TypeError(errmsg) from None
-            return coords, is_single
+                    coords = coords.positions  # homogenise to a numpy array
+                    ncoord = coords.shape[0]
+                    if allow_atomgroup:
+                        pass
+                    else:
```
orbeckst: Why not write as
```python
if not allow_atomgroup:
```
and save the `pass`?

package/MDAnalysis/lib/util.py
```diff
@@ -2051,32 +2070,54 @@ def check_coords_decorator(func):
...
+                        err = TypeError("AtomGroup or other class with a"
+                                        "`.positions` method supplied as an"
+                                        "argument, but allow_atomgroup is"
+                                        " False")
+                        raise err
+                except TypeError:
+                    raise err
```
orbeckst: This will not have `err` defined: if the TypeError is raised, the code will have never seen line 2107.

package/MDAnalysis/lib/util.py
```diff
@@ -2051,32 +2070,54 @@ def check_coords_decorator(func):
...
+                except:
```
orbeckst: I don't like blank exceptions — can we make this a TypeError or AttributeError or whatever you expect to raise this exception?

package/MDAnalysis/lib/util.py
orbeckst: Shouldn't this except just be merged with the subsequent one? I don't quite understand the logic here.

testsuite/MDAnalysisTests/lib/test_distances.py
```diff
@@ -20,6 +20,7 @@
 # MDAnalysis: A Toolkit for the Analysis of Molecular Dynamics Simulations.
 # J. Comput. Chem. 32 (2011), 2319--2327, doi:10.1002/jcc.21787
 #
+from turtle import position
```
orbeckst: Really, we're using `turtle` somewhere?!

orbeckst: Give that `position` is ambiguous, I suggest to just
```python
import turtle
```
and then use `turtle.position` to make a bit clearer what is being used.

testsuite/MDAnalysisTests/lib/test_distances.py
```diff
@@ -371,16 +426,71 @@ def test_periodic(self, DCD_Universe, backend):
...
+                     " Natoms entries")
         assert_almost_equal(d.min(), 0.11981228170520701, self.prec,
                             err_msg="wrong minimum distance value with PBC")
         assert_almost_equal(d.max(), 53.572192429459619, self.prec,
                             err_msg="wrong maximum distance value with PBC")
 
+    def test_atomgroup_simple(self, DCD_Universe, backend):
+        # need two copies as moving ts updates underlying array on atomgroup
+        U1, trajectory1 = DCD_Universe
+        U2 = MDAnalysis.Universe(PSF, DCD)
```
orbeckst: make a fixture for consistency

testsuite/MDAnalysisTests/lib/test_util.py
```diff
@@ -1738,6 +1737,59 @@ def func(a, b):
...
+        u = mda.Universe(PSF, DCD)
+        return u.atoms
+
+    # check atomgroup handling with every option except allow_atomgroup
+    @pytest.mark.parametrize('enforce_copy', [True, False])
+    @pytest.mark.parametrize('enforce_dtype', [True, False])
+    @pytest.mark.parametrize('allow_single', [True, False])
+    @pytest.mark.parametrize('convert_single', [True, False])
+    @pytest.mark.parametrize('reduce_result_if_single', [True, False])
+    @pytest.mark.parametrize('check_lengths_match', [True, False])
```
orbeckst, line 1751: Ah, the joys of combinatorical testing.... ;-)

orbeckst (changes requested):
Great work! Some minor comments but the one major issue is the logic in the new try/except. Can you please have a look — see comments? Thanks!

package/MDAnalysis/lib/distances.py
hmacdope: AFAIK the `|` syntax is only supported for 3.10

See the note on the page  you linked above
```
Changed in version 3.10: Unions can now be written as X | Y. See [union type expressions](https://docs.python.org/3/library/stdtypes.html#types-union).
```

hmacdope: Normally you would import the module and use the naked type `AtomGroup`, but the use of the TYPE_CHECKING guard (ie `AtomGroup` is not imported at runtime) means this needs to be a forward declaration.

 See https://docs.python.org/3/library/typing.html#constant for a admittedly brief explanation. 

@tylerjereddy suggested this as a solution to circular imports in [this](https://github.com/MDAnalysis/mdanalysis/pull/3719#discussion_r901164303) comment on #3719.

orbeckst: Thanks for educating me.

Sorry, didn't see the 3.10 — there's lots of movement in the type annotations!!

hmacdope: Yeah its taking a bit of getting used to for me also but all in all I think a good change.  My new (mid) year resolution is to try and write typed code. :laughing:

testsuite/MDAnalysisTests/lib/test_distances.py
hmacdope: I have no idea how that got in there, definatly an error, have removed.

package/MDAnalysis/lib/distances.py
hmacdope: Okay something I was just reading is that PEP563 (https://peps.python.org/pep-0563/) means we may not need the string annotations if using Python 3.7 + and we using something similar to 

```python
if typing.TYPE_CHECKING:
    import expensive_mod

def a_func(arg: expensive_mod.SomeClass) -> None:
    a_var: expensive_mod.SomeClass = arg
    ...

```

testsuite/MDAnalysisTests/lib/test_util.py
hmacdope: :stuck_out_tongue_winking_eye:

package/MDAnalysis/lib/util.py
hmacdope: I'm sure it can be done better as it definitely looks janky but I think my logic was:

If an `AtomGroup` is supplied but not allowed by `allow_atomgroup` we need to manually raise an exception inside the `try `block. 

If we get this `TypeError` then re-raise it in the `except: TypeError` block. 

Otherwise we catch an `AttributeError` and raise a TypeError that the wrong type was supplied (see updated code).

orbeckst: With the explanation it makes a bit more sense. It's just confusing (and a bit fragile) because you assume that the only place where a TypeError can originate is L2109 where you raise it yourself. Then you catch the exception to re-raise. 

Why do you have to re-raise? Can't you just delete
```python
except TypeError:
   raise err
```
and just `raise TypeError("AtomGroup or other class ...")`. The outer `except AttributeError` will not swallow the `TypeError`.

Perhaps you needed the re-raise when you had the blank `except:` but hopefully not anymore.

hmacdope: Done :)

testsuite/MDAnalysisTests/lib/test_distances.py
hmacdope: done.

hmacdope:
I should add as a note that because of the use of `numpy.typing` pending the bump of NumPy min version to 1.20 in #3737 has made this PR dependent on it, thats why a lot of the `Azure` CI is failing.

testsuite/MDAnalysisTests/lib/test_distances.py
```diff
@@ -321,11 +367,24 @@ class FakeArray(np.ndarray):
...
     universe = MDAnalysis.Universe(PSF, DCD)
     trajectory = universe.trajectory
+    return universe, trajectory
+
+# second independent universe required for 
+# TestDistanceArrayDCD_TRIC.test_atomgroup_simple
+@pytest.fixture()
+def DCD_Universe2():
+    universe = MDAnalysis.Universe(PSF, DCD)
+    trajectory = universe.trajectory
```
orbeckst, line 379: Just a comment: For all of these fixtures, we could have just written
```python
    return universe, universe.trajectory
```
although it's a bit superfluous to return `trajectory` when calling code can just get it as easily (`U.trajectory` is not much more work than `trajectory` and it's also clearer); furthermore, the fixture name is also confusing because it is not just the universe but also the trajectory.

None of this is your fault, of course, I'm just commenting while looking at the code.

orbeckst (approved):
Thank you for addressing all my comments. Excellent test coverage!

The PEP8 reports has some issues with blank spaces and lines (types E261, W293, E302, W291) and lines that are too long (E501) that would be nice to be fixed; the visual indentation is probably intentional.

We should _not_ merge until numpy 1.20 is the law of the land (PR #3737 ) so that our repo CI does not show ugly red status badges.

testsuite/MDAnalysisTests/lib/test_distances.py
```diff
@@ -596,6 +759,19 @@ def test_issue_3725(box):
...
 
+def conv_dtype_if_ndarr(a, dtype):
+    if isinstance(a, np.ndarray):
+        return a.astype(dtype)
+    else:
+        return a
+
+
+def convert_position_dtype_if_ndarray(a, b, c, d, dtype):
+    return conv_dtype_if_ndarr(a, dtype), \
```
IAlibay: Implicit continuations please

testsuite/MDAnalysisTests/lib/test_distances.py
```diff
@@ -256,6 +256,15 @@ def ref_system():
...
 
 
+@pytest.fixture()
+def ref_system_universe(ref_system):
+    box, points, ref, conf = ref_system
+    u = MDAnalysis.Universe.empty(points.shape[0], trajectory=True)
+    u.atoms.positions = points
+    u.trajectory.ts.dimensions = box
+    return box, u.atoms, u.select_atoms("index 0"), u.select_atoms("index 1 to"
+                                                                   " 3")
```
IAlibay: This is pretty unreadable, can you use implicit continuation here to make make this more human readable or just break PEP8?

package/MDAnalysis/lib/util.py
```diff
@@ -2051,32 +2070,50 @@ def check_coords_decorator(func):
...
+                        err = TypeError("AtomGroup or other class with a"
+                                        "`.positions` method supplied as an"
+                                        "argument, but allow_atomgroup is"
+                                        " False")
+                        raise err
+                except AttributeError:
+                    raise TypeError("{}(): Parameter '{}' must be a"
+                                    " numpy.ndarray or an AtomGroup,"
+                                    " got {}.".format(fname,
+                                    argname, type(coords)))
```
IAlibay: Use f strings?

testsuite/MDAnalysisTests/lib/test_distances.py
```diff
@@ -264,24 +273,61 @@ def _dist(x, ref):
...
+    def test_PBC_mixed_combinations(self, backend, ref_system, pos0, pos1,
+                                    request):
+        box, points, _, _ = ref_system
+        _, _, ref_val, _ = request.getfixturevalue(pos0)
+        _, points_val, _, _ = request.getfixturevalue(pos1)
+        d = distances.distance_array(ref_val, points_val,
+                                     box=box,
+                                     backend=backend)
+        assert_almost_equal(d, np.array([[0., 0., 0., self._dist(points[3],
+                            ref=[1, 1, 2])]]))
```
IAlibay: This would probably fit visual indentation and improve readability.

```suggestion
        assert_almost_equal(
            d, np.array([[0., 0., 0., self._dist(points[3], ref=[1, 1, 2])]]))
```

IAlibay (changes requested):
A quick review, all formatting comments.

testsuite/MDAnalysisTests/lib/test_distances.py
hmacdope: Would you like me to fix it here, raise an issue or leave it for now?

orbeckst: Up to you — it would make for an outreachy-type issue.

hmacdope:
Hopefully looking visually better.

testsuite/MDAnalysisTests/lib/test_distances.py
```diff
@@ -256,6 +256,15 @@ def ref_system():
...
 
+@pytest.fixture()
+def ref_system_universe(ref_system):
+    box, points, ref, conf = ref_system
+    u = MDAnalysis.Universe.empty(points.shape[0], trajectory=True)
+    u.atoms.positions = points
+    u.trajectory.ts.dimensions = box
+    return (box, u.atoms, u.select_atoms("index 0"),
+            u.select_atoms("index 1 to 3"))
+
```
IAlibay, line 267:
```suggestion


```

testsuite/MDAnalysisTests/lib/test_distances.py
```diff
@@ -424,6 +542,53 @@ def test_periodic(self, DCD_Universe, backend):
...
+        U, _ = Triclinic_Universe
+        x0_ag = U.select_atoms(sel)
+        x0_arr = U.atoms.positions[np_slice]
+        d_ag = distances.self_distance_array(x0_ag, box=U.coord.dimensions,
+                                             backend=backend)
+        d_arr = distances.self_distance_array(x0_arr, box=U.coord.dimensions,
+                                              backend=backend)
+        assert_allclose(d_ag, d_arr,
+                        err_msg="AtomGroup and NumPy distances do not match")
+
```
IAlibay, line 592:
```suggestion


```

testsuite/MDAnalysisTests/lib/test_distances.py
```diff
@@ -596,6 +761,19 @@ def test_issue_3725(box):
...
+    else:
+        return a
+
+
+def convert_position_dtype_if_ndarray(a, b, c, d, dtype):
+    return (conv_dtype_if_ndarr(a, dtype),
+            conv_dtype_if_ndarr(b, dtype),
+            conv_dtype_if_ndarr(c, dtype),
+            conv_dtype_if_ndarr(d, dtype))
+
```
IAlibay, line 778:
```suggestion


```

testsuite/MDAnalysisTests/lib/test_distances.py
```diff
@@ -1091,6 +1381,14 @@ def test_input_unchanged_apply_PBC(self, coords, box, backend):
...
         assert_equal(crd, ref)
 
+    @pytest.mark.parametrize('box', boxes[:2])
+    @pytest.mark.parametrize('backend', ['serial', 'openmp'])
+    def test_input_unchanged_apply_PBC_atomgroup(self, coords_atomgroups, box,
+                                                 backend):
+        crd = coords_atomgroups[0]
+        ref = crd.positions.copy()
+        res = distances.apply_PBC(crd, box, backend=backend)
+        assert_equal(crd.positions, ref)
```
IAlibay, line 1394:
```suggestion
        assert_equal(crd.positions, ref)
        
```

testsuite/MDAnalysisTests/lib/test_distances.py
```diff
@@ -371,16 +434,71 @@ def test_periodic(self, DCD_Universe, backend):
...
+        assert_almost_equal(d.min(), 0.11981228170520701, self.prec,
+                            err_msg="wrong minimum distance value")
+        assert_almost_equal(d.max(), 53.572192429459619, self.prec,
+                            err_msg="wrong maximum distance value")
+
+    # check no box and ortho box types and some slices
+    @pytest.mark.parametrize('box', [None, [50, 50, 50, 90, 90, 90]])
+    @pytest.mark.parametrize("sel, np_slice", [("all", np.s_[:, :]),
+                            ("index 0 to 8 ", np.s_[0:9, :]),
+                            ("index 9", np.s_[8, :])])
```
IAlibay:
```suggestion
    @pytest.mark.parametrize("sel, np_slice", [("all", np.s_[:, :]),
                             ("index 0 to 8 ", np.s_[0:9, :]),
                             ("index 9", np.s_[8, :])])
```

testsuite/MDAnalysisTests/lib/test_distances.py
```diff
@@ -371,16 +434,71 @@ def test_periodic(self, DCD_Universe, backend):
...
+                                        backend=backend)
+        d_arr = distances.distance_array(x0_arr, x1_arr, box=box,
+                                         backend=backend)
+        assert_allclose(d_ag, d_arr,
+                        err_msg="AtomGroup and NumPy distances do not match")
+
+    # check triclinic box and some slices
+    @pytest.mark.parametrize("sel, np_slice", [("all", np.s_[:, :]),
+                            ("index 0 to 8 ", np.s_[0:9, :]),
+                            ("index 9", np.s_[8, :])])
```
IAlibay:
```suggestion
    @pytest.mark.parametrize("sel, np_slice", [("all", np.s_[:, :]),
                             ("index 0 to 8 ", np.s_[0:9, :]),
                             ("index 9", np.s_[8, :])])
```

testsuite/MDAnalysisTests/lib/test_distances.py
```diff
@@ -424,6 +542,53 @@ def test_periodic(self, DCD_Universe, backend):
...
+        assert_almost_equal(d.min(), 0.92905562402529318, self.prec,
+                            err_msg="wrong minimum distance value")
+        assert_almost_equal(d.max(), 52.4702570624190590, self.prec,
+                            err_msg="wrong maximum distance value")
+
+    # check no box and ortho box types and some slices
+    @pytest.mark.parametrize('box', [None, [50, 50, 50, 90, 90, 90]])
+    @pytest.mark.parametrize("sel, np_slice", [("all", np.s_[:, :]),
+                            ("index 0 to 8 ", np.s_[0:9, :]),
+                            ("index 9", np.s_[8, :])])
```
IAlibay:
```suggestion
    @pytest.mark.parametrize("sel, np_slice", [("all", np.s_[:, :]),
                             ("index 0 to 8 ", np.s_[0:9, :]),
                             ("index 9", np.s_[8, :])])
```

package/MDAnalysis/lib/util.py
```diff
@@ -2051,32 +2070,49 @@ def check_coords_decorator(func):
...
-                if coords.ndim == 1:
-                    is_single = True
-                    if convert_single:
-                        coords = coords[None, :]
+            if isinstance(coords, np.ndarray):
+                if allow_single:
+                    if (coords.ndim not in (1, 2)) or (coords.shape[-1] != 3):
+                        raise ValueError("{}(): {}.shape must be (3,) or"
+                                         " (n, 3), got {}.".format(fname,
+                                         argname, coords.shape))
```
IAlibay:
```suggestion
                        errmsg = (f"{fname}(): {argname}.shape must be (3,) or "
                                  f"(n, 3), got {coords.shape})
                        raise ValueError(errmsg)
```

package/MDAnalysis/lib/util.py
```diff
@@ -2051,32 +2070,49 @@ def check_coords_decorator(func):
...
+                    if coords.ndim == 1:
+                        is_single = True
+                        if convert_single:
+                            coords = coords[None, :]
+                else:
+                    if (coords.ndim != 2) or (coords.shape[1] != 3):
+                        raise ValueError("{}(): {}.shape must be (n, 3),"
+                                         " got {}.".format(fname, argname,
+                                                          coords.shape))
```
IAlibay:
```suggestion
                        errmsg = (f"{fname}(): {argname}.shape must be (n, 3) "
                                  f"got {coords.shape})
                        raise ValueError(errmsg)
```

IAlibay (changes requested):
Not 100% sure why some of this isn't being picked up by pep8speaks, anyways just a couple of PEP8 changes for the lines that changed (I'm ignoring one of the pep8speaks one because 81 char >> readable than 79 char in that occasion).

hmacdope:
Thanks @IAlibay much appreciated.

hmacdope:
Hmmmm still getting what looks to me like NumPy versioning issues on the `Azure` pipelines despite #3737 merged?

IAlibay:
How odd, was developed merged into this branch?

hmacdope:
Yeah it looks all up to date AFAIK. Log shows #3737 as merged in my history.

IAlibay:
@hmacdope - Here we go this is the problem: https://github.com/numpy/numpy/releases/tag/v1.21.0

"A runtime-subcriptable alias has been added for ndarray"

Essentially NDArray isn't going to work in 1.20... That's a bit of a pain, I guess our options here are to either vendor the alias somewhere in our code or break NEP29. We should probably move this to a new issue since it's a pretty big decision.

orbeckst:
That's just related to typing, though, right?

It looks to me that typing will remain a moving target for some time so my view is to do the best you can within current constraints and then move on. Add a comment how to do it nicely when np 1.21 is lay of the land and then put in something ugly or less specific. 

I recognize that there are different views and mine is just the one of an outsider as far as typing goes...

IAlibay:
> That's just related to typing, though, right?
> 
> It looks to me that typing will remain a moving target for some time so my view is to do the best you can within current constraints and then move on. Add a comment how to do it nicely when np 1.21 is lay of the land and then put in something ugly or less specific.
> 
> I recognize that there are different views and mine is just the one of an outsider as far as typing goes...

Hugo will open an issue later today, but one of the things we'll need to discuss is what we do with the outreachy typing project since I think they've already started using npt.NDArray

orbeckst:
> Hugo will open an issue later today, but one of the things we'll need to discuss is what we do with the outreachy typing project since I think they've already started using npt.NDArray

Ahhh... and that's not been caught in tests with lowest supported versions? Alright, that's discussion that should happen on its own issue.

hmacdope:
Pending discussion on #3748

hmacdope:
Other than the flaky tests this should be good to go again following discussion in #3748.

hmacdope:
@MDAnalysis/coredev any chance I could get a second review here? It would be good to get this and #3751 merged soonish. :)

hmacdope:
@MDAnalysis/coredevs  any chance I could get a second review here? It would be good to get this and #3751 merged soonish. :)

IAlibay:
@hmacdope can you address whatever can be addressed from pep8speaks (all seems reasonable stuff like two lines vs 1). And get CI to run green. I'll review tomorrow once that's done.

package/MDAnalysis/lib/distances.py
```diff
@@ -128,8 +138,8 @@ def _run(funcname, args=None, kwargs=None, backend="serial"):
 
 from .c_distances_openmp import OPENMP_ENABLED as USED_OPENMP
 
-
-def _check_result_array(result, shape):
+# typing: numpy
```
IAlibay, line 142:
```suggestion

# typing: numpy
```

package/MDAnalysis/lib/distances.py
```diff
@@ -169,11 +179,14 @@ def _check_result_array(result, shape):
 #        raise ValueError("{0} is not C-contiguous.".format(desc))
     return result
 
-
+# typing: numpy
```
IAlibay, line 184:
```suggestion

# typing: numpy
```

package/MDAnalysis/lib/distances.py
```diff
@@ -250,9 +268,12 @@ def distance_array(reference, configuration, box=None, result=None,
 
     return distances
 
-
-@check_coords('reference', reduce_result_if_single=False)
-def self_distance_array(reference, box=None, result=None, backend="serial"):
+# typing: numpy
```
IAlibay, line 274:
```suggestion

# typing: numpy
```

package/MDAnalysis/lib/distances.py
```diff
@@ -1240,9 +1265,13 @@ def transform_StoR(coords, box, backend="serial"):
     _run("coord_transform", args=(coords, box), backend=backend)
     return coords
 
-
-@check_coords('coords1', 'coords2')
-def calc_bonds(coords1, coords2, box=None, result=None, backend="serial"):
+# typing: numpy
```
IAlibay, line 1272:
```suggestion

# typing: numpy
```

package/MDAnalysis/lib/distances.py
```diff
@@ -1322,10 +1356,14 @@ def calc_bonds(coords1, coords2, box=None, result=None, backend="serial"):
 
     return bondlengths
 
-
-@check_coords('coords1', 'coords2', 'coords3')
-def calc_angles(coords1, coords2, coords3, box=None, result=None,
-                backend="serial"):
+# typing: numpy
```
IAlibay, line 1364:
```suggestion

# typing: numpy
```

package/MDAnalysis/lib/distances.py
```diff
@@ -1414,10 +1458,15 @@ def calc_angles(coords1, coords2, coords3, box=None, result=None,
 
     return angles
 
-
-@check_coords('coords1', 'coords2', 'coords3', 'coords4')
-def calc_dihedrals(coords1, coords2, coords3, coords4, box=None, result=None,
-                   backend="serial"):
+# typing: numpy
```
IAlibay, line 1467:
```suggestion

# typing: numpy
```

package/MDAnalysis/lib/distances.py
```diff
@@ -1519,16 +1575,19 @@ def calc_dihedrals(coords1, coords2, coords3, coords4, box=None, result=None,
 
     return dihedrals
 
-
-@check_coords('coords')
-def apply_PBC(coords, box, backend="serial"):
+# typing: numpy
```
IAlibay, line 1585:
```suggestion

# typing: numpy
```

package/MDAnalysis/lib/distances.py
```diff
@@ -1561,9 +1623,9 @@ def apply_PBC(coords, box, backend="serial"):
 
     return coords
 
-
+# typing: numpy
```
IAlibay, line 1634:
```suggestion

# typing: numpy
```

package/MDAnalysis/lib/util.py
```diff
@@ -2051,32 +2070,49 @@ def check_coords_decorator(func):
...
-                                                      coords.shape))
-                if coords.ndim == 1:
-                    is_single = True
-                    if convert_single:
-                        coords = coords[None, :]
+            if isinstance(coords, np.ndarray):
+                if allow_single:
+                    if (coords.ndim not in (1, 2)) or (coords.shape[-1] != 3):
+                        errmsg = (f"{fname}(): {argname}.shape must be (3,) or "
+                                  f"(n, 3), got {coords.shape}")
```
IAlibay:
```suggestion
                        errmsg = (f"{fname}(): {argname}.shape must be (3,) "
                                  f"or (n, 3), got {coords.shape}")
```

testsuite/MDAnalysisTests/lib/test_distances.py
```diff
@@ -1091,6 +1384,15 @@ def test_input_unchanged_apply_PBC(self, coords, box, backend):
...
 
+    @pytest.mark.parametrize('box', boxes[:2])
+    @pytest.mark.parametrize('backend', ['serial', 'openmp'])
+    def test_input_unchanged_apply_PBC_atomgroup(self, coords_atomgroups, box,
+                                                 backend):
+        crd = coords_atomgroups[0]
+        ref = crd.positions.copy()
+        res = distances.apply_PBC(crd, box, backend=backend)
+        assert_equal(crd.positions, ref)
+
```
IAlibay: I sometimes get issues with suggestions deleting lines when I do this.. hopefully it doesn't happen here.
```suggestion

```

IAlibay (approved):
@hmacdope here are all the suggested changes that were left.

I'll eagerly approve with the hope that you'll handle these suggestions before merging this PR.

hmacdope:
Thanks @IAlibay, sorry I swear I made the changes before I pinged but I think I confused myself with #3751.

PR #3735, opened 2022-06-26
Nuclinfo Major Pair and Minor Pair overhaul

### LINKED ISSUE #3720: Re-implement nuclinfo using `AnalysisBase` style subclasses
To complete the overhaul of `nucinfo` we need to rebuild the rest of the functions in the analysis base style in the `nucleicacids` module.

Using the `NucPairDist` as a super class rebuild
```python
dists = MinorPairDist([ag1, ag2],[ag3,ag4], U)
dists.run()
```

## TODO
- [x] rebuild `minor_pair`
- [x] rebuild `major_pair`
- [ ] rebuild `phase_cp`
- [ ] rebuild `phase_as`
- [ ] rebuild `tors`

## Additional context ##
In the last update in PR #3611 the I began the process of overhauling nuclinfo, and I would like to finish that process.

ALescoulie:
Fixes #3720 

 - Add a class for Major Pair and Minor Pair distance in the new overhauled nucleic acids module.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -209,3 +218,155 @@ def __init__(self, strand1: List[Residue], strand2: List[Residue],
...
+
+class MinorPairDist(NucPairDist):
+    r"""Minor-Pair basepair distance for selected residues over a trajectory.
+
+    Takes two lists of :class:`~MDAnalysis.core.groups.Residue` objects and
+    calculates the Watson-Crick distance between them over the trajectory.
+    Bases are matched by their index in the lists given as arguments.
+
+    Parameters
+    __________
```
IAlibay: (here and everywhere else) I know these still render, but the line style is very much different from the numpy-like style guide we use. Can you instead replace this with a `------` line?

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -209,3 +218,155 @@ def __init__(self, strand1: List[Residue], strand2: List[Residue],
...
+        sel2: List[mda.AtomGroup] = []
+        strand = zip(strand1, strand2)
+
+        for s in strand:
+            if s[0].resname[0] in [c_name, t_name, u_name]:
+                a1, a2 = o2_name, c2_name
+            elif s[0].resname[0] in [a_name, g_name]:
+                a1, a2 = c2_name, o2_name
+            else:
+                raise ValueError(f"{s} are not valid nucleic acids")
```
IAlibay: Unless I'm misreading, this code block is essentially being repeated in all the child classes. Could you just use a staticmethod here that generalises this code?

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -209,3 +218,155 @@ def __init__(self, strand1: List[Residue], strand2: List[Residue],
...
+
+            sel1.append(s[0].atoms.select_atoms(f'name {a1}'))
```
IAlibay: You can double check this on your end but testing things out you'd probably get a ~ 5x speedup if you did
`s[0].atoms[np.where(s[0].atoms.names == a1)]` instead

package/MDAnalysis/analysis/nucleicacids.py
IAlibay: Also please cover these errors with tests

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -209,3 +218,155 @@ def __init__(self, strand1: List[Residue], strand2: List[Residue],
...
+    c_name: str (optional)
+        Name of Cytosine in topology
+        by default assigned to C
+    **kwargs: dict
+        arguments for :class:`~MDAnalysis.analysis.base.AnalysisBase`
+
+    Attributes
+    __________
+    results: numpy.ndarray
+    first index is selection second index is time
```
IAlibay: This is missing indentation it seems?

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -209,3 +218,155 @@ def __init__(self, strand1: List[Residue], strand2: List[Residue],
...
+        for s in strand:
+            if s[0].resname[0] in [c_name, t_name, u_name]:
+                a1, a2 = o2_name, c2_name
+            elif s[0].resname[0] in [a_name, g_name]:
+                a1, a2 = c2_name, o2_name
+            else:
+                raise ValueError(f"{s} are not valid nucleic acids")
+
+            sel1.append(s[0].atoms.select_atoms(f'name {a1}'))
+            sel2.append(s[1].atoms.select_atoms(f'name {a2}'))
```
IAlibay: What happens here if one of the selections returns an empty atomgroup and the other doesn't? (for some reason or another, maybe the wrong name selection was passed?)

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -56,6 +64,7 @@
 from .distances import calc_bonds
 from .base import AnalysisBase, Results
 from MDAnalysis.core.groups import Residue
+from .dihedrals import Dihedral
```
IAlibay: Probably missing something obvious, where is this used?

IAlibay (changes requested):
Sorry only had the time/energy for a quick review.

Couple of extra things I couldn't add comments for but worth mentioning since it seems like there's going to be further changes to this code:
1. lines 132, 139, and 142 - we already populate a times array under self.times, is this still necessary to populate? Can we just do a call to `np.array(self.times)`?
2. The Attributes for NucPairDist aren't indented properly, could you change them here?
3. Dicussion point (other @MDAnalysis/coredevs please weigh in here), the way the `results` dictionary is being used here is very different from how we do results everywhere else & imho makes it a bit hard to specifically extract the pairs of distances vs say the times entry. I realise it's a bit late since this already shipped in 2.2.0, but before we keep going, would it make sense to consider switching this to instead be `results.pair_distances` (maybe even making it a pairs x times 2D ndarray?

testsuite/MDAnalysisTests/analysis/test_nucleicacids.py
```diff
@@ -23,7 +23,8 @@
 
 import MDAnalysis as mda
 import pytest
-from MDAnalysis.analysis.nucleicacids import WatsonCrickDist
+from MDAnalysis.analysis.nucleicacids import WatsonCrickDist,\
+    MajorPairDist, MinorPairDist
```
IAlibay: It doesn't get picked up by pep8speaks, but implicit continuation is best where possible

```suggestion
from MDAnalysis.analysis.nucleicacids import (WatsonCrickDist,
                                              MajorPairDist, MinorPairDist,)
```

testsuite/MDAnalysisTests/analysis/test_nucleicacids.py
```diff
@@ -43,3 +44,27 @@ def test_wc_dist(u):
...
+
+def test_minor_dist(u):
+    strand: mda.AtomGroup = u.select_atoms("segid RNAA")
+    strand1 = [strand.residues[2], strand.residues[19]]
+    strand2 = [strand.residues[16], strand.residues[4]]
+
+    MI = MinorPairDist(strand1, strand2)
+    MI.run()
+
+    assert_allclose(MI.results[0][0], 15.06506, atol=1e-3)
```
IAlibay: another one for @MDAnalysis/coredevs to consider for our testing style going ahead - single float comparisons are ~ 20x faster (at least on my laptop) if done using `pytest.approx` instead of `assert_allclose`. It's microseconds (170 microsec vs 9 microsec), but it adds up quickly to seconds, how do we feel about making pytest.approx the assert method of choice in those cases?

package/CHANGELOG
```diff
@@ -27,6 +27,8 @@ Enhancements
   * Additional logger.info output when per-frame analysis starts (PR #3710)
   * Timestep has been converted to a Cython Extension type
     (CZI Performance track, PR #3683)
+  * Add higher performance AnalysisBase derived Major and Minor pair distances to
```
IAlibay: We usually go newest first for these entries, could you move it to the top of the enhancements list?

IAlibay (dismissed):
Couple of extra things since I was playing with pytest.approx

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: Oh that didn't get cut when I moved a commit around. I was working on the rebuilding another the torsion calculator and had imported that, I'll remove that

package/MDAnalysis/analysis/nucleicacids.py
richardjgowers: I'd be curious if this was still true after the string interning changes...

richardjgowers (commented):
Thanks for doing this!  Refactoring old code is never pleasant.

testsuite/MDAnalysisTests/analysis/test_nucleicacids.py
orbeckst: I raised #3743 for you.

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -209,3 +218,155 @@ def __init__(self, strand1: List[Residue], strand2: List[Residue],
             sel2.append(s[1].atoms.select_atoms(f'name {a2}'))
 
         super(WatsonCrickDist, self).__init__(sel1, sel2, **kwargs)
+
+
+class MinorPairDist(NucPairDist):
+    r"""Minor-Pair basepair distance for selected residues over a trajectory.
+
+    Takes two lists of :class:`~MDAnalysis.core.groups.Residue` objects and
```
orbeckst: Can these be `ResidueGroup`, too? That would tie in even better with the OO approach.

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -209,3 +218,155 @@ def __init__(self, strand1: List[Residue], strand2: List[Residue],
...
+        by default assigned to G
+    u_name: str (optional)
+        Name of Uracil in topology
+        by default assigned to U
+    t_name: str (optional)
+        Name of Thymine in topology
+        by default assigned to T
+    c_name: str (optional)
+        Name of Cytosine in topology
+        by default assigned to C
```
orbeckst: Would it be clearer to use `guanine="G"` instead of `g_name="G"` and same for everything else?

I find the meaning to be clearer when reading `guanine="G"`.

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -209,3 +218,155 @@ def __init__(self, strand1: List[Residue], strand2: List[Residue],
...
+    first index is selection second index is time
+    results.times: numpy.ndarray
+    times used in analysis
+
+    Raises
+    ______
+    ValueError
+    if the residues given are not amino acids
+    ValueError
+    if the selections given are not the same length
```
orbeckst: indentation

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -209,3 +218,155 @@ def __init__(self, strand1: List[Residue], strand2: List[Residue],
...
+
+    """
+
+    def __init__(self, strand1: List[Residue], strand2: List[Residue],
```
orbeckst: `Union(List[Residue], ResidueGroup)` ?

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -209,3 +218,155 @@ def __init__(self, strand1: List[Residue], strand2: List[Residue],
...
+                 o2_name: str = 'O2', c2_name: str = "C2",
+                 g_name: str = 'G', a_name: str = 'A', u_name: str = 'U',
+                 t_name: str = 'T', c_name: str = 'C',
+                 **kwargs) -> None:
+        sel1: List[mda.AtomGroup] = []
+        sel2: List[mda.AtomGroup] = []
+        strand = zip(strand1, strand2)
```
orbeckst: plural "strands" ?

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -209,3 +218,155 @@ def __init__(self, strand1: List[Residue], strand2: List[Residue],
...
+
+        for s in strand:
```
orbeckst: rename `s` to `pair`  to make meaning clearer?

package/MDAnalysis/analysis/nucleicacids.py
orbeckst: Or perhaps clearer, `double_strand`?

package/MDAnalysis/analysis/nucleicacids.py
orbeckst: yes, please 👍

orbeckst: And then you can test it separately.

package/MDAnalysis/analysis/nucleicacids.py
orbeckst: @richardjgowers PLEASE write a benchmark case to be run for anything that you improve

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -209,3 +218,155 @@ def __init__(self, strand1: List[Residue], strand2: List[Residue],
...
+            else:
+                raise ValueError(f"{s} are not valid nucleic acids")
+
+            sel1.append(s[0].atoms.select_atoms(f'name {a1}'))
+            sel2.append(s[1].atoms.select_atoms(f'name {a2}'))
+
+        super(MinorPairDist, self).__init__(sel1, sel2, **kwargs)
+
+
+class MajorPairDist(NucPairDist):
```
orbeckst, line 538: most of my comments from `MinorPairDist` apply here, too

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -209,3 +218,155 @@ def __init__(self, strand1: List[Residue], strand2: List[Residue],
...
+    calculates the Watson-Crick distance between them over the trajectory.
+    Bases are matched by their index in the lists given as arguments.
+
+    Parameters
+    __________
+    strand1: List[Residue]
+        First list of bases
+    strand2: List[Residue]
+        Second list of bases
+    o6_name: str (optional)
```
orbeckst, line 554: `oxygen6` ?

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -209,3 +218,155 @@ def __init__(self, strand1: List[Residue], strand2: List[Residue],
...
+        Name of Oxygen 6 of nucleic acids
+        by default assigned to N1
```
orbeckst: Why is O6 assigned to N1? Do you mean that this is the distance? Needs to be written more clearly.

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -209,3 +218,155 @@ def __init__(self, strand1: List[Residue], strand2: List[Residue],
...
+    n4_name: str (optional)
+        Name of Nitrogen 4 of nucleic acids
+        by default assigned to N3
```
orbeckst: not clear why N4 is N3 by default ... is this what it's meant here??

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -209,3 +218,155 @@ def __init__(self, strand1: List[Residue], strand2: List[Residue],
...
+    first index is selection second index is time
+    results.times: numpy.ndarray
+    times used in analysis
+
+    Raises
+    ______
+    ValueError
+    if the residues given are not amino acids
+    ValueError
+    if the selections given are not the same length
```
orbeckst: underlining and indentation

orbeckst (changes requested):
Thank you for the rewrite.

I have a bunch of readability comments (see inline).

I agree with @IAlibay that we should remove `results['times']`. I also agree that `results` needs restructuring, both for consistency but also for ease of accessing the data with e.g. mdacli. Let's raise a separate issue where others can also easily chime in. Normally we adhere to semantic versioning. However, I would consider this module still experimental and I am willing to break strict semver here and just change `results` without deprecations. The changes to results can go into a separate PR (preferrable) or this one (if a lot easier).

orbeckst:
Before going much further here we should discuss the format of `Results` — please see #3744.

IAlibay:
@ALescoulie It would be great to not let this work get forgotten, are you still looking to contribute this?

ALescoulie:
Yeah, I'll continue working on this. Sorry for abandoning it, I was busy with school and dealing with mental health issues. I'm getting the development repo set up on my desktop now.

orbeckst:
Hi @ALescoulie good to hear from you! Would be fantastic to get your work in!

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: fixed

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: I test in the refactor of the selection into `NucPairDist`

ALescoulie: refactored it into the `NucPairDist` parent class like @IAlibay recommended

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: fixed in a recent commit

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: I added a check for that in the static method in `NucPairDist`

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: removed in latest push

testsuite/MDAnalysisTests/analysis/test_nucleicacids.py
ALescoulie: fixed the issue in my latest push

testsuite/MDAnalysisTests/analysis/test_nucleicacids.py
ALescoulie: I refactored my test to use pytest approx

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: I updated it to use the full name, ex: `guanine_name=G`

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: fixed

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: I copied the doc string and forgot to update the variable description, I fixed it in the last push

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: its an error from copying the string

ALescoulie:
I pushed a larger change to the PR. It was a bit hard picking up where I left off, but I was able to jump back into developing without too many issues. 

I refactored my code so that all of the Pair distance analysis classes use a `NucPairDist` static method `select_strand_atoms` which takes in the strands, selected atoms names, and the nucleic acid names. It then returns a tuple containing the the lists of atoms for each strand. It also has checks ensuring that the strands only contain Nucleic Acids and that the selections don't return empty `AtomGroups`. I also added tests for both errors.

### TODO
- [x] Reformat `NucPairDist` to use the newer results format
- [x] Fix misc docs issues
- [x] Update the change log

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: fixed

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: I'm looking at the source and docs for `ResidueGroup`. I can see the utility in merging `NucPairDist` with it, since it would allow for easier access to the residue related data in the classes. I'm a little unclear though on how it would handle the two strands the user provides, can `GroupBase` automatically deal with that if I add a line with `super(ResidueGroup, strand1, strand2)`.

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: I set it to strands

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: good idea, make the code more readable

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: I'll add that, I looked at `ResidueGroup` and there is no reason `NucPairDist` can't accept it or `Residue`

ALescoulie: I used a type alias to do it so that the code is less verbose, what do you think about that solution? It is a bit less clear in the docs, but simplifies the code.

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: Has this been bench-marked yet? If not I can set one up and adjust the selection code accordingly.

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: I updated all the names so that they are more clear

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -140,6 +148,41 @@ def __init__(self, selection1: List[mda.AtomGroup],
             self._s1 += selection1[i]
             self._s2 += selection2[i]
 
+    @staticmethod
+    def select_strand_atoms(
```
orbeckst, line 191: As a public facing method, needs some docs.

Include a `.. versionadded:: 2.7.0`.

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -164,19 +207,19 @@ class WatsonCrickDist(NucPairDist):
         First list of bases
     strand2: List[Residue]
         Second list of bases
-    n1_name: str (optional)
+    nitrogen1_name: str (optional)
```
orbeckst: These parameter renamings are technically speaking an API break so we'd need to deprecate them, document deprecations, and test.

If you think the more explicit names are worth the effort then please do the above. Otherwise, I am also ok with reverting to the original kwargs. They are reasonably clear already and the docs say explicitly what they are.

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -210,23 +253,150 @@ class WatsonCrickDist(NucPairDist):
     """
 
     def __init__(self, strand1: List[Residue], strand2: List[Residue],
-                 n1_name: str = 'N1', n3_name: str = "N3",
-                 g_name: str = 'G', a_name: str = 'A', u_name: str = 'U',
-                 t_name: str = 'T', c_name: str = 'C',
+                 nitrogen1_name: str = 'N1', nitrogen3_name: str = "N3",
+                 guanine_name: str = 'G', adenine_name: str = 'A', uracil_name: str = 'U',
+                 tymine_name: str = 'T', cytosine_name: str = 'C',
```
orbeckst: See comment above — if you want to rename you'll also have to accept the old kwargs (possibly via `**kwargs`) for at least one minor release cycle.

package/MDAnalysis/analysis/nucleicacids.py
orbeckst: I like making it a separate method.

package/MDAnalysis/analysis/nucleicacids.py
orbeckst: I didn't think that far ahead: All I thought was that instead of `strand1: List[Residue]` use `strand1: ResidueGroup` (etc).

orbeckst: However, we then should also use ResidueGroup for WatsonCrick and then we're back to API changes... so if we use ResidueGroup here then we also need to deprecate `List[Residue]` for `WatsonCrick`.

I think I like `ResidueGroup` better because `strand = u.select_atoms("nucleic and resid 1-17").residues` looks better than `strand = list(u.select_atoms("nucleic and resid 1-17").residues)`. I _think_ one can simply convert a `res_list: List[Residue]` to `residues: [ResidueGroup]` with
```python
residues = ResidueGroup(res_list)
```
so accepting both for a transition period wouldn't be difficult.

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -210,23 +253,150 @@ class WatsonCrickDist(NucPairDist):
...
-        strand = zip(strand1, strand2)
 
-        for s in strand:
-            if s[0].resname[0] in [c_name, t_name, u_name]:
-                a1, a2 = n3_name, n1_name
-            elif s[0].resname[0] in [a_name, g_name]:
-                a1, a2 = n1_name, n3_name
-            else:
-                raise ValueError(f"{s} are not valid nucleic acids")
+        selections: Tuple[List[mda.AtomGroup], List[mda.AtomGroup]] = self.select_strand_atoms(
```
orbeckst: Instead of `selections` I would call it `strand_atomgroups` or something similar (maybe just `strands`), so that it's clear that it contains actual objects and not just "selection strings".

package/MDAnalysis/analysis/nucleicacids.py
orbeckst: I don't understand what a type alias is and where you used it — sorry, apologies for my ignorance. Can you please explain and comment on the corresponding piece of code?

orbeckst: At least from the docs it is not clear to me that the class would now also accept `strand1: ResidueGroup` — ... at least from your comment I surmise that's what _should_ be possible now.

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -210,23 +253,150 @@ class WatsonCrickDist(NucPairDist):
...
+    calculates the Watson-Crick distance between them over the trajectory.
+    Bases are matched by their index in the lists given as arguments.
+
+    Parameters
+    ----------
+    strand1: List[Residue]
+        First list of bases
+    strand2: List[Residue]
+        Second list of bases
+    oxygen2_name: str (optional)
```
orbeckst: All these names must be consistent between `WatsonCrick` and the other classes, which means either deprecating the old kwargs in `WatsonCrick` _or_ changing the ones here to the current ones.

Ultimately, that's your choice, but it needs to be **consistent across the whole module**. You're defining the API for everything else so think about it how you want it to be. I am fine with either long or short names as long as the docs are clear (which they are) and everything is consistent and as uniform as possible so that it is a smooth user experience.

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -210,23 +253,150 @@ class WatsonCrickDist(NucPairDist):
...
+        )
+
+        super(WatsonCrickDist, self).__init__(selections[0], selections[1], **kwargs)
+
+
+class MinorPairDist(NucPairDist):
+    r"""Minor-Pair basepair distance for selected residues over a trajectory.
+
+    Takes two lists of :class:`~MDAnalysis.core.groups.Residue` objects and
+    calculates the Watson-Crick distance between them over the trajectory.
```
orbeckst: How is `MinorPairDist` different from `WatsonCrickDist`? The docs do not say. Please add details to the docs to describe the difference.

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -210,23 +253,150 @@ class WatsonCrickDist(NucPairDist):
...
+        super(MinorPairDist, self).__init__(selections[0], selections[1], **kwargs)
+
+
+
+class MajorPairDist(NucPairDist):
+    r"""Minor-Pair basepair distance for selected residues over a trajectory.
+
+    Takes two lists of :class:`~MDAnalysis.core.groups.Residue` objects and
+    calculates the Watson-Crick distance between them over the trajectory.
+    Bases are matched by their index in the lists given as arguments.
```
orbeckst: How is `MajorPairDist` different from `WatsonCrickDist`? The docs do not say. Please add details to the docs to describe the difference.

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -210,23 +253,150 @@ class WatsonCrickDist(NucPairDist):
...
+
+    Parameters
+    ----------
+    strand1: List[Residue]
+        First list of bases
+    strand2: List[Residue]
+        Second list of bases
+    oxygen6_name: str (optional)
```
orbeckst: See comments above about consistent argument naming.

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -210,23 +253,150 @@ class WatsonCrickDist(NucPairDist):
...
+    cytosine_name: str (optional)
+        Name of Cytosine in topology
+        by default assigned to C
+    **kwargs: dict
+        arguments for :class:`~MDAnalysis.analysis.base.AnalysisBase`
+
+    Attributes
+    ----------
+    results: numpy.ndarray
+    first index is selection second index is time
```
orbeckst: indent (I think)

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -210,23 +253,150 @@ class WatsonCrickDist(NucPairDist):
...
+    results.times: numpy.ndarray
+    times used in analysis
```
orbeckst: indent (I think)

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -210,23 +253,150 @@ class WatsonCrickDist(NucPairDist):
...
+        Name of Cytosine in topology
+        by default assigned to C
+    **kwargs: dict
+        arguments for :class:`~MDAnalysis.analysis.base.AnalysisBase`
+
+    Attributes
+    ----------
+    results: numpy.ndarray
+    first index is selection second index is time
+    results.times: numpy.ndarray
```
orbeckst: should just be times (and not in results)

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -210,23 +253,150 @@ class WatsonCrickDist(NucPairDist):
...
+        Name of Cytosine in topology
+        by default assigned to C
+    **kwargs: dict
+        arguments for :class:`~MDAnalysis.analysis.base.AnalysisBase`
+
+    Attributes
+    ----------
+    results: numpy.ndarray
+        first index is selection second index is time
+    results.times: numpy.ndarray
```
orbeckst: should just be times (and not in results)

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -210,23 +253,150 @@ class WatsonCrickDist(NucPairDist):
...
+        by default assigned to T
+    cytosine_name: str (optional)
+        Name of Cytosine in topology
+        by default assigned to C
+    **kwargs: dict
+        arguments for :class:`~MDAnalysis.analysis.base.AnalysisBase`
+
+    Attributes
+    ----------
+    results: numpy.ndarray
```
orbeckst: use results.distances or something like that in the Results dict

orbeckst: Describe the output data structure.

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -210,23 +253,150 @@ class WatsonCrickDist(NucPairDist):
...
```
orbeckst: Should be results.distances or something like that inside Results dict.

Describe the output data structure.

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -210,23 +253,150 @@ class WatsonCrickDist(NucPairDist):
...
 
-            sel1.append(s[0].atoms.select_atoms(f'name {a1}'))
-            sel2.append(s[1].atoms.select_atoms(f'name {a2}'))
+    Raises
+    ------
+    ValueError
+        if the residues given are not amino acids
+    ValueError
+        if the selections given are not the same length
+
```
orbeckst, line 516: needs versionadded

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -210,23 +253,150 @@ class WatsonCrickDist(NucPairDist):
...
+    results.times: numpy.ndarray
+    times used in analysis
+
+    Raises
+    ------
+    ValueError
+        if the residues given are not amino acids
+    ValueError
+        if the selections given are not the same length
+
```
orbeckst, line 598: needs versionadded

orbeckst (changes requested):
I like the compact code and moving the strand selection into a separate method. 

1. The primary issue is just that renaming kwargs is an API break and because the code has been in core for a while (and not marked as "Experimental" or anything like that) we'd need to properly deprecate any API changes. The alternative is using the current (old) names also for the newly added classes.
2. In terms of API coherency, I think it were better if actual MDA objects, namely `ResidueGroup`, instead of `list[Residue]` were accepted. This would require a deprecation of the current list of residues input although that would be worth it in my opinion, so that we have a seamless OO API.
3. There are also some documentation issues but these are easy to fix.

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -140,6 +148,41 @@ def __init__(self, selection1: List[mda.AtomGroup],
...
+        sel1: List[mda.AtomGroup] = []
+        sel2: List[mda.AtomGroup] = []
+        strand: Tuple[Residue, Residue] = zip(strand1, strand2)
+
+
+        for s in strand:
+            if s[0].resname[0] in pyrimidines:
+                a1, a2 = a2_name, a1_name
+            elif s[0].resname[0] in purines:
+                a1, a2 = a1_name, a2_name
```
richardjgowers: This could probably get simplified to:
```python
if not s[0].resname[0] in pyrimidines_and_purines:
    raise ValueError
```
I'm not sure what assigning to a1 and a2 does here, so you could remove that step.

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -140,6 +148,41 @@ def __init__(self, selection1: List[mda.AtomGroup],
...
+            else:
+                raise ValueError(f"AtomGroup in {s} is not a valid nucleic acid")
+
+            ag1 = s[0].atoms.select_atoms(f'name {a1}')
+            ag2 = s[1].atoms.select_atoms(f'name {a2}')
+
+            if not all([len(ag) > 0 for ag in [ag1, ag2]]):
```
richardjgowers: the `all` function doesn't require a list, so you could reduce this to `all(len(ag) for ag in [ag1, ag2])`

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -140,6 +148,41 @@ def __init__(self, selection1: List[mda.AtomGroup],
...
+        a1_name: str, a2_name: str, guanine_name: str = 'G',
+        adenine_name: str = 'A', uracil_name: str = 'U',
+        tymine_name: str = 'T', cytosine_name: str = 'C'
+        ) -> Tuple[List[mda.AtomGroup], List[mda.AtomGroup]]:
+        pyrimidines: List[str] = [cytosine_name, tymine_name, uracil_name]
+        purines: List[str] = [adenine_name, guanine_name]
+        
+        sel1: List[mda.AtomGroup] = []
+        sel2: List[mda.AtomGroup] = []
+        strand: Tuple[Residue, Residue] = zip(strand1, strand2)
```
richardjgowers: the type annotations are a great addition.  I think this one (from `zip`) is technically a generator/iterator.  I think instead you want to annotate `s: Tuple[Residue, Residue]`

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: I'll just revert to the older version to prevent API changes.

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: I just reverted to the old kwargs

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: @orbeckst A type alias is when you give a specific name to a composite type in my case `ResidueClass` is an alias for `Union[Residue, ResidueGroup]` so my code's type signatures are a bit less verbose

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: Ok, I'll make it clear in my code that either can be used for these processes.

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: oh cool, I never knew that!

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: I just moved zip into the for look line.

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -92,8 +106,6 @@ class NucPairDist(AnalysisBase):
 
     Attributes
     ----------
-    times: numpy.ndarray
-        Simulation times for analysis.
```
orbeckst, line 89: This should remain because it's created by AnalysisBase, isn't it?

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -183,8 +229,6 @@ class WatsonCrickDist(NucPairDist):
...
```
orbeckst, line 180: Needs to stay, I think.

package/MDAnalysis/analysis/nucleicacids.py
orbeckst: Now I see how we got to the API-breaking naming... my comment from a year ago when it would have been easy to change the API without a break. As I said elsewhere, `g_name` etc work.

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -209,24 +253,147 @@ class WatsonCrickDist(NucPairDist):
...
+        Name of Thymine in topology
+        by default assigned to T
+    c_name: str (optional)
+        Name of Cytosine in topology
+        by default assigned to C
+    **kwargs: dict
+        arguments for :class:`~MDAnalysis.analysis.base.AnalysisBase`
+
+    Attributes
+    ----------
```
orbeckst, line 579: add docs for `times` (thanks to AnalysisBase)

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -209,24 +253,147 @@ class WatsonCrickDist(NucPairDist):
...
```
orbeckst, line 498: add docs for `times` thanks to AnalysisBase

ALescoulie:
> I like the compact code and moving the strand selection into a separate method.
> 
> 1. The primary issue is just that renaming kwargs is an API break and because the code has been in core for a while (and not marked as "Experimental" or anything like that) we'd need to properly deprecate any API changes. The alternative is using the current (old) names also for the newly added classes.
> 2. In terms of API coherency, I think it were better if actual MDA objects, namely `ResidueGroup`, instead of `list[Residue]` were accepted. This would require a deprecation of the current list of residues input although that would be worth it in my opinion, so that we have a seamless OO API.
> 3. There are also some documentation issues but these are easy to fix.

@orbeckst I think I have a good way to switch to using `ResidueGroup` from `List[Residue]` without API breaks. I'll make `NucPairDist` and`WCDist` accept both, but raise a `DepreciatedWarning` if you give it `List[Residue]` then just convert it internally. Than raise an issue tagged with 3,0 to remove support for residue lists when we do API changes for the major version update. What do you think of that solution?

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: done

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: updated the docs to expalin the arrays orginzation

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: updated the docs to explain the arrays orginization

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: added it on both major and minor pair

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: I added a description to the docs for both

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: I uptated the docs for all the classes

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: updated for all the classes

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: fixed in a prev commit

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: I updated the docs for all the classes describing how its a 2d numpy array with base pairs and frames as the indicies

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: done

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: done

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: done

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: done

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: This is based on how it was done in nucinfo, I'm basing my code on that originonal material. As far as I know its so that the selections are specified to one side of the pair.

orbeckst:
good idea!

ALescoulie:
@orbeckst I'll push that change today, was waiting for feedback on the idea before implementing it

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: fixed in recent versions

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: fixed in the recent versions

ALescoulie:
I think this one is pretty close to done, There might still be some misc typos in the docs, but aside from that I refactored the code based on the suggestions of @orbeckst to use `ResidueGroup` and pulled selections into their own static methods. I also have gotten full code coverage, and updated CHANGELLOG.

IAlibay:
@ALescoulie - could you please confirm that you are ok releasing this code under a LGPLv2.1+ license and that it adheres to the [developer certificate of origin](https://developercertificate.org/)?

Aside from this I will dismiss my review block and leave this to @orbeckst to review & merge.

ALescoulie:
> @ALescoulie - could you please confirm that you are ok releasing this code under a LGPLv2.1+ license and that it adheres to the [developer certificate of origin](https://developercertificate.org/)?
> 
> 
> 
> Aside from this I will dismiss my review block and leave this to @orbeckst to review & merge.

I am happy to release this under the LGPL license. I wrote all of this on my own based on the older nucinfo code, and the MDAnalysis framework. I hope that it helps people with their research.

orbeckst (changes requested):
A test fails with issues importing TypeAlias https://github.com/MDAnalysis/mdanalysis/actions/runs/6059699690/job/16443750172?pr=3735 — please fix first.

You can ignore test failures related to #4209 in `test_creating_multiple_universe_without_offset`.

ALescoulie:
I just removed the type alias import, apparently that wasn't in Python 3.9, @orbeckst is there anything else to fix on this PR

package/CHANGELOG
```diff
@@ -13,7 +13,9 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" http://semver.org
 
 ------------------------------------------------------------------------------
-??/??/?? IAlibay, ianmkenney, PicoCentauri, ALescoulie
+
+??/??/?? IAlibay, ianmkenney, PicoCentauri, pgbarletta, p-j-smith, richardjgowers, ALescoulie
```
orbeckst: line break
```suggestion
??/??/?? IAlibay, ianmkenney, PicoCentauri, pgbarletta, p-j-smith, richardjgowers,
         ALescoulie
```

package/CHANGELOG
```diff
@@ -64,6 +68,8 @@ Changes
     as DEF is deprecated as of Cython>=3.0 (Issue #4237, PR #4246)
 
 Deprecations
+  * added `DeprecationWarning` to `analysis.nucleicacids.WatsonCrickDist` when
+    `List[Residue]` is given instead of `ResidueGroup`
```
orbeckst: Write for users:
```suggestion
  * using `List[Residue]` as input for `analysis.nucleicacids.WatsonCrickDist()` is 
    deprecated and will be removed in release 3.0.0; instead use a `ResidueGroup` 
    (Issue #3720, PR #3735)
```

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -183,50 +290,213 @@ class WatsonCrickDist(NucPairDist):
...
+        and the second index selecting the frame number
+    times: numpy.ndarray
+            Simulation times for analysis.
 
     Raises
     ------
-    ValueError
-        If the residues given are not amino acids
+    DeprecationWarning
+        If a lits of :class:`~MDAnalysis.core.groups.Residue` is given for
```
orbeckst:
```suggestion
        If a list of :class:`~MDAnalysis.core.groups.Residue` is given for
```

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -183,50 +290,213 @@ class WatsonCrickDist(NucPairDist):
...
     .. versionchanged:: 2.5.0
        Accessing results by passing strand indices to :attr:`results` is
        was deprecated and is now removed as of MDAnalysis version 2.5.0. Please
        use :attr:`results.pair_distances` instead.
        The :attr:`results.times` was deprecated and is now removed as of
        MDAnalysis 2.5.0. Please use the class attribute :attr:`times` instead.
+
+    .. versionchanged:: 2.7.0
+        Deprecated the use of :attr:`List[Residue]` instead use
+        :class:`~MDAnalysis.core.groups.ResidueGroup`.
```
orbeckst: This is a bit tricky; normally I would put a `.. deprecated::` role here, but that's going to be too confusing for the casual reader. We're really only deprecating one type of input so I'd rewrite as:
```suggestion
        `strand1` and `strand2` now also accept a :class:`~MDAnalysis.core.groups.ResidueGroup`
        as input. The previous input type of ``List[Residue]`` is still supported
        but it is **deprecated** and will be removed in release 3.0.0.
```

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -183,50 +290,213 @@ class WatsonCrickDist(NucPairDist):
...
+    times: numpy.ndarray
+            Simulation times for analysis.
 
     Raises
     ------
-    ValueError
-        If the residues given are not amino acids
+    DeprecationWarning
+        If a lits of :class:`~MDAnalysis.core.groups.Residue` is given for
+        :attr:`strand1` of :attr:`strand2` instead of a
```
orbeckst: Just bare role for arguments
```suggestion
        `strand1` of `strand2` instead of a
```

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -55,20 +55,34 @@
...
 import numpy as np
 
 import MDAnalysis as mda
 from .distances import calc_bonds
 from .base import AnalysisBase, Results
-from MDAnalysis.core.groups import Residue
+from MDAnalysis.core.groups import Residue, ResidueGroup
+
 
+ResidueClass = Union[List[Residue], ResidueGroup]
```
orbeckst, line 79:
```suggestion
# Deprecation: In 3.0.0 change type to just
# ResidueClass = ResidueGroup
ResidueClass = Union[List[Residue], ResidueGroup]
```

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -55,20 +55,34 @@
...
+r"""A type alias for :code:`Union[Residue, ResidueGroup]`
```
orbeckst:
```suggestion
r"""A type alias for :code:`Union[List[Residue], ResidueGroup]`
```

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -183,50 +290,213 @@ class WatsonCrickDist(NucPairDist):
...
                  t_name: str = 'T', c_name: str = 'C',
                  **kwargs) -> None:
-        sel1: List[mda.AtomGroup] = []
-        sel2: List[mda.AtomGroup] = []
-        strand = zip(strand1, strand2)
+        
+        if isinstance(strand1, list) or isinstance(strand2, list):
+            raise DeprecationWarning("ResidueGroup should be used instead of giving a Residue list")
+            strand1: ResidueGroup = ResidueGroup(strand1)
+            strand2: ResidueGroup = ResidueGroup(strand2)
```
orbeckst: This indentation does not look right. Please check.

package/MDAnalysis/analysis/nucleicacids.py
orbeckst: Start by copying the original docs eg https://docs.mdanalysis.org/stable/documentation_pages/analysis/nuclinfo.html#MDAnalysis.analysis.nuclinfo.major_pair

Also, the docs here incorrectly say "Minor-Pair".

package/MDAnalysis/analysis/nucleicacids.py
orbeckst: Start by adding docs from orginal functions https://docs.mdanalysis.org/stable/documentation_pages/analysis/nuclinfo.html#MDAnalysis.analysis.nuclinfo.minor_pair

orbeckst: Explicitly state what distances are calculated.

package/MDAnalysis/analysis/nucleicacids.py
orbeckst: Explicitly state what distances are calculated.

orbeckst (changes requested):
There's one potential bug in the code where I noted funny indentation (code never executed after a `raise`).

Please also fix/update the docs for major and minor pair distances.

Finally, please accept my suggestions for rewriting some of the changes.

Almost there! Thanks!

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: The indentation is fine, it appears wrong due to the `DeprecationWarning` going onto the next line. There is an error I overlooked thought, I used `raise` which caused a failure. I switched it to warn so that it just throws a warring then continues running the code. I'll also make sure the code continues working correctly after the warning to be safe.

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: done

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: done

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: I went through and removed the all `attr` used for arguments

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: done

ALescoulie:
@orbeckst I believe I addressed your feedback on the documentation, and added some further clarity. I also rewrote how `WatsonCrickDist` handles `List[Resiude]` in order to catch possible errors when both a `List[Residue]` and a `ResidueGroup` are passed in and in cases where the list is not entirely `Residues`. I think I should still add one additional test to get back to 100% coverage as there is one if statement I don't hit, I also will go read the documentation one more time.

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -92,21 +108,32 @@ class NucPairDist(AnalysisBase):
...
     results.pair_distances: numpy.ndarray
         2D array of pair distances. First dimension is simulation time, second
         dimension contains the pair distances for each each entry pair in
         selection1 and selection2.
 
         .. versionadded:: 2.4.0
 
+    results.distances: numpy.ndarray
+        stored in a 2d numpy array with first index selecting the Residue pair, 
+        and the second index selecting the frame number
```
orbeckst: How is `distances` different from `pair_distances`?

orbeckst: And if it's the same then you need to (sigh) deprecate the first one as it's part of the API.

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -148,21 +257,31 @@ def _single_frame(self) -> None:
         self._res_array[self._frame_index, :] = dist
 
     def _conclude(self) -> None:
-        self.results['pair_distances'] = self._res_array
+        self.results['distances'] = self._res_array
```
orbeckst, line 291: If you want to change the name then you need to
1. deprecate pair_distances (add entry to CHANGELOG and `.. deprecated` to the entry)
2. keep an entry `self.results['pair_distances'] = self.results['distances']` until it's removed in 3.0.0

The entries in `results` are part of the API so they fall under SemVer.

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -183,50 +302,234 @@ class WatsonCrickDist(NucPairDist):
...
+        and the second index selecting the frame number
+    times: numpy.ndarray
+            Simulation times for analysis.
 
     Raises
     ------
-    ValueError
-        If the residues given are not amino acids
+    DeprecationWarning
+    .. _Deprecation Warning
```
orbeckst: Does this anchor work? Perhaps a trailing colon missing?

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -183,50 +302,234 @@ class WatsonCrickDist(NucPairDist):
...
 
     .. versionchanged:: 2.5.0
        Accessing results by passing strand indices to :attr:`results` is
        was deprecated and is now removed as of MDAnalysis version 2.5.0. Please
        use :attr:`results.pair_distances` instead.
        The :attr:`results.times` was deprecated and is now removed as of
        MDAnalysis 2.5.0. Please use the class attribute :attr:`times` instead.
+
+    .. versionchanged:: 2.7.0
+    .. _Deprecation Notice
```
orbeckst: as above : does the anchor work, did you check the generated docs?

orbeckst (changes requested):
Thanks for addressing my comments. The only real issue is that we need to deprecate `results.pair_distance` in the WatsonCrick class properly.

In the new classes there's nothing to deprecate as they appear here for the first time.

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: OK, I'll add a deprecation notice and add that to the docs and change log

ALescoulie: It's done, I added it to the docs.

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: I think the change from `pair_distances` to `distances` was a suggestion from above, I added a deprecation notice in the docs and a todo in the code to remove it.

package/MDAnalysis/analysis/nucleicacids.py
ALescoulie: neither anchor works, I'm having an issues building my docs locally that I asked about on discord. Once I sort that out I'll go over the docs again.

ALescoulie:
@orbeckst I think this PR about ready, assuming I didn't miss any typos in the docs. I also pulled most of the atom selections in the tests into a fixture so there is less repetition when running the CI.

package/MDAnalysis/analysis/nucleicacids.py
orbeckst: Thanks. I rewrote some of the additions using the [.. deprecated:: x.x](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated) role.

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -133,93 +176,428 @@ def __init__(self, selection1: List[mda.AtomGroup],
...
+        :class:`~MDAnalysis.analysis.base.AnalysisBase`
 
     Attributes
     ----------
-    times: numpy.ndarray
-        Simulation times for analysis.
+    results.distances: numpy.ndarray
+        Distances are stored in a 2d numpy array with axis 0 (first index)
+        indexing the trajectory frame and axis 1 (second index) selecting the  
+        Residue pair.
```
orbeckst: @ALescoulie please check that this rewritten description is correct. It is now consistent with the description of `results.pair_distances`.

orbeckst (approved):
Thanks @ALescoulie , looks good. 

I made some changes to the docs, mainly to use sphinx markup for deprecations, but I also included a few fixes. In particular, please have a look for yourself that I correctly defined `results.distances`.

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -133,93 +176,428 @@ def __init__(self, selection1: List[mda.AtomGroup],
...
+
+        super(WatsonCrickDist, self).__init__(
+            strand_atomgroups[0], strand_atomgroups[1], **kwargs
+            )
+
+
+
+class MinorPairDist(NucPairDist):
+    r"""Minor-Pair basepair distance for selected residues
+        over a trajectory.
```
orbeckst:
```suggestion
    r"""Minor-Pair basepair distance for selected residues over a trajectory.
```

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -133,93 +176,428 @@ def __init__(self, selection1: List[mda.AtomGroup],
...
+        super(MinorPairDist, self).__init__(
+            selections[0], selections[1], **kwargs
+            )
 
-        super(WatsonCrickDist, self).__init__(sel1, sel2, **kwargs)
+
+
+class MajorPairDist(NucPairDist):
+    r"""Minor-Pair base pair distance for 
+    selected residues over a trajectory.
```
orbeckst:
```suggestion
    r"""Minor-Pair base pair distance for selected residues over a trajectory.
```

orbeckst (changes requested):
In principle all good, but please fix PEP8 things.

ALescoulie:
@orbeckst, I'll get it done by tomorrow than I can start on the next part of #3720, the pep8 stuff will just take a few minutes.

ALescoulie:
@orbeckst I got the pep8 stuff done

orbeckst (approved):

orbeckst:
Hooray. I reopen #3720 for the other things.

PR #3736, opened 2022-06-26
 Pin back pytest-cov

IAlibay:
This is going to take a bit of time, essentially just seeing what pin will stops these conflicting tracer file errors.

 - re-pin pytest-cov but at 2.8.1 max

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/.coveragerc
+++ b/package/.coveragerc
@@ -1,2 +1,0 @@
-[run]
-plugins = Cython.Coverage
--- a/.coveragerc
+++ b/.coveragerc
@@ -7,6 +7,7 @@ omit =
     */legacy/*
     */due.py
 concurrency = thread, multiprocessing
+parallel = True
 
 [report]
 exclude_lines =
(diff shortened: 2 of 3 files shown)


IAlibay:
So thrawling through the various issues surrounding this, it looks like coverage > 5.0 is still broken for plenty of Cython related things, we'll just have to settle with these pins again, although they are getting increasingly old & outdated... I also think this isn't purely upstream, there's seems to be some deeper issues about our tests that will need digging into & addressing at some point (although I can't quite put my finger down on what they are).

package/.coveragerc
```diff
@@ -1,2 +0,0 @@
-[run]
-plugins = Cython.Coverage
```
jbarnoud, line 2: Is the cython coverage still reported?

jbarnoud (approved):
Assuming the coverage for cython is reported it looks good to me.

package/.coveragerc
IAlibay: Should do (see: https://app.codecov.io/gh/MDAnalysis/mdanalysis/compare/3736/tree/package/MDAnalysis/lib)

This file _shouldn't_ be used by coverage anyways since we force pytest to pick up the coveragerc at the top of MDA directory.

orbeckst (approved):
Let's hope for the best.

(cython coverage is reported)

PR #3737, opened 2022-06-27
bump up numpy min version to 1.20

IAlibay:
For 2.3.0 we'll likely have a minimum of 1.20 support as per NEP29

 - 

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -29,6 +29,7 @@ Enhancements
     (CZI Performance track, PR #3683)
 
 Changes
+  * Minimum supported NumPy version is now 1.20 as per NEP29 (PR #3737)
   * Narrowed variable scope to reduce use of OpenMP `private` clause (PR #3706, PR
     #3728)
 
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -29,7 +29,8 @@ Enhancements
     (CZI Performance track, PR #3683)
 
 Changes
-  * Minimum supported NumPy version is now 1.20 as per NEP29 (PR #3737)
+  * Minimum supported NumPy version is now 1.20 (1.21 for macosx-arm64)
+    as per NEP29 (PR #3737)
   * Narrowed variable scope to reduce use of OpenMP `private` clause (PR #3706, PR
     #3728)
 
--- a/package/pyproject.toml
+++ b/package/pyproject.toml
@@ -4,13 +4,10 @@ requires = [
     "Cython>=0.28",
     "packaging",
     # lowest NumPy we can use for a given Python,
-    # except for more exotic platforms (linux/Mac ARM flavors)
-    "numpy==1.19; python_version=='3.8' and (platform_machine!='arm64' or platform_system!='Darwin') and platform_machine!='aarch64'",
-    "numpy==1.19.3; python_version=='3.9' and (platform_machine!='arm64' or platform_system!='Darwin') and platform_machine!='aarch64'",
+    # arm65 on Darwin supports py3.8 and above and requires numpy>=1.20.0
+    "numpy==1.20.0; python_version=='3.8'",
+    "numpy==1.20.0; python_version=='3.9'",
     "numpy==1.21.4; python_version=='3.10' and platform_python_implementation != 'PyPy'",
     "setuptools",
     "wheel",
-    # arm64 on Darwin supports Python 3.8 and above and requires numpy>=1.20.0
-    "numpy==1.20.0; python_version=='3.8' and platform_machine=='arm64' and platform_system=='Darwin'",
-    "numpy==1.20.0; python_version=='3.9' and platform_machine=='arm64' and platform_system=='Darwin'",
 ]
--- a/package/pyproject.toml
+++ b/package/pyproject.toml
@@ -4,10 +4,16 @@ requires = [
     "Cython>=0.28",
     "packaging",
     # lowest NumPy we can use for a given Python,
-    # arm65 on Darwin supports py3.8 and above and requires numpy>=1.20.0
-    "numpy==1.20.0; python_version=='3.8'",
-    "numpy==1.20.0; python_version=='3.9'",
-    "numpy==1.21.4; python_version=='3.10' and platform_python_implementation != 'PyPy'",
+    # except for more exotic platform (Mac Arm flavors)
+    # aarch64, AIX, s390x all support < 1.20 so we can safely pin to this
+    "numpy==1.20.0; python_version=='3.8' and (platform_machine!='arm64' or platform_system!='Darwin')",
+    "numpy==1.20.0; python_version=='3.9' and (platform_machine!='arm64' or platform_system!='Darwin')",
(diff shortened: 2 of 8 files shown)


package/pyproject.toml
```diff
@@ -4,13 +4,10 @@ requires = [
     "Cython>=0.28",
     "packaging",
     # lowest NumPy we can use for a given Python,
-    # except for more exotic platforms (linux/Mac ARM flavors)
-    "numpy==1.19; python_version=='3.8' and (platform_machine!='arm64' or platform_system!='Darwin') and platform_machine!='aarch64'",
```
IAlibay, line 8: @tylerjereddy I wasn't 100% sure of the logic here, we seem to have entries specfic for "not ((arm64 or Darwin) and aarch64)" and "arm64 and Darwin" but nothing for aarch64. Is there a specific reason we were restricting aarch64 to py3.10 only?

package/pyproject.toml
```diff
@@ -4,13 +4,10 @@ requires = [
...
-    "numpy==1.19.3; python_version=='3.9' and (platform_machine!='arm64' or platform_system!='Darwin') and platform_machine!='aarch64'",
+    # arm65 on Darwin supports py3.8 and above and requires numpy>=1.20.0
```
orbeckst: Is "arm65" a thing? OR should this be "arm64"?

orbeckst (approved):
I didn't check other files where numpy might show up, but what I see in the diff makes sense to me. 

There's probably a tiny typo in a comment ("arm65" instead of "arm64") but that's not a blocker for me.

I'd like to see @tylerjereddy comment on the specific question in `pyproject.toml` but leave it to others to evaluate that part.

package/pyproject.toml
IAlibay: typo definitely

hmacdope (approved):
I am not familiar with all the places that the numpy version would need to be changed, but similar to @orbeckst what I can see from the diff makes sense. I had a poke around but couldn't find anywhere else in the docs or anything like that would need changing so I will approve. :smiley:

package/pyproject.toml
tylerjereddy: Older Python versions should be "ok" with `aarch64` I think. This is a good reference as always: https://github.com/scipy/oldest-supported-numpy/blob/main/setup.cfg#L29

Of course, there's no substitute for actually trying it with `pip install ..`, but my original draft was probably incomplete anyway, especially for the more exotic platforms.

package/pyproject.toml
```diff
@@ -4,13 +4,10 @@ requires = [
...
+    "numpy==1.20.0; python_version=='3.8'",
+    "numpy==1.20.0; python_version=='3.9'",
     "numpy==1.21.4; python_version=='3.10' and platform_python_implementation != 'PyPy'",
     "setuptools",
     "wheel",
-    # arm64 on Darwin supports Python 3.8 and above and requires numpy>=1.20.0
-    "numpy==1.20.0; python_version=='3.8' and platform_machine=='arm64' and platform_system=='Darwin'",
-    "numpy==1.20.0; python_version=='3.9' and platform_machine=='arm64' and platform_system=='Darwin'",
```
tylerjereddy, line 15: I remember this tripping up a user on Discord because it forces a from-source build of NumPy on M1 macs--perhaps we should consider doing the same thing as `oldest-supported-numpy` and just requiring `1.21.0` as described here: https://github.com/scipy/oldest-supported-numpy/blob/main/setup.cfg#L37 ?

IAlibay: Yeah reading the comment re: universal2 I'm pretty happy bumping up the darwin + arm64 requirements to 1.21.0.

orbeckst: Sounds sensible.

And thank you @tylerjereddy for linking to the setup.cfg... might be good to keep the link around somewhere.... maybe a wiki page of tips of the trade for packaging? ;-)

package/pyproject.toml
orbeckst:
```suggestion
    # arm64 on Darwin supports py3.8 and above and requires numpy>=1.21.0
```
and then change below

package/pyproject.toml
```diff
@@ -4,13 +4,10 @@ requires = [
     "Cython>=0.28",
     "packaging",
     # lowest NumPy we can use for a given Python,
-    # except for more exotic platforms (linux/Mac ARM flavors)
-    "numpy==1.19; python_version=='3.8' and (platform_machine!='arm64' or platform_system!='Darwin') and platform_machine!='aarch64'",
-    "numpy==1.19.3; python_version=='3.9' and (platform_machine!='arm64' or platform_system!='Darwin') and platform_machine!='aarch64'",
+    # arm65 on Darwin supports py3.8 and above and requires numpy>=1.20.0
+    "numpy==1.20.0; python_version=='3.8'",
```
orbeckst:
```suggestion
    "numpy==1.21.0; python_version=='3.8'",
```

package/pyproject.toml
```diff
@@ -4,13 +4,10 @@ requires = [
...
+    "numpy==1.20.0; python_version=='3.9'",
```
orbeckst:
```suggestion
    "numpy==1.21.0; python_version=='3.9'",
```

IAlibay:
Sorry @orbeckst I was pushing changes just as you were posting those suggested changes

package/pyproject.toml
```diff
@@ -4,13 +4,16 @@ requires = [
     "Cython>=0.28",
     "packaging",
     # lowest NumPy we can use for a given Python,
-    # except for more exotic platforms (linux/Mac ARM flavors)
-    "numpy==1.19; python_version=='3.8' and (platform_machine!='arm64' or platform_system!='Darwin') and platform_machine!='aarch64'",
-    "numpy==1.19.3; python_version=='3.9' and (platform_machine!='arm64' or platform_system!='Darwin') and platform_machine!='aarch64'",
-    "numpy==1.21.4; python_version=='3.10' and platform_python_implementation != 'PyPy'",
+    # except for more exotic platform (Mac Arm flavors)
+    # aarch64, AIX, s390x all support < 1.20 so we can safely pin to this
```
IAlibay, line 8: As far as I can tell this should work, there are aarch64 wheels for 1.20 for py3.9.

I've bumped up the requirements for arm64+darwin to 1.21.0 as per https://github.com/scipy/oldest-supported-numpy/blob/main/setup.cfg

I've also set the py3.10 build for everything to be 1.21.6 given recommendation from the above too.

orbeckst:
Do you want more reviews? Otherwise you could merge.

package/CHANGELOG
```diff
@@ -29,6 +29,7 @@ Enhancements
     (CZI Performance track, PR #3683)
 
 Changes
+  * Minimum supported NumPy version is now 1.20 as per NEP29 (PR #3737)
```
orbeckst: Should this say 1.21 for some archs?

PR #3746, opened 2022-07-04
type hints for Visualization module

umak1106:
Fixes #

 - 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/doc/sphinx/source/documentation_pages/overview.rst
+++ b/package/doc/sphinx/source/documentation_pages/overview.rst
@@ -32,7 +32,7 @@ of a protein and the radius of gyration of the backbone atoms are calculated::
     cterm = u.select_atoms('segid 4AKE and name C')[-1]  # ... takes the last atom named 'C'
     bb = u.select_atoms('protein and backbone')  # a selection (a AtomGroup)
     for ts in u.trajectory:  # iterate through all frames
-        r = cterm.pos - nterm.pos  # end-to-end vector from atom positions
+        r = cterm.position - nterm.position  # end-to-end vector from atom positions
         d = numpy.linalg.norm(r)   # end-to-end distance
         rgyr = bb.radius_of_gyration()  # method of a AtomGroup; updates with each frame
         print(f"frame = {ts.frame}: d = {d} Angstroem, Rgyr = {rgyr} Angstroem")


package/MDAnalysis/visualization/streamlines.py
```diff
@@ -45,7 +45,9 @@
 import multiprocessing
 
 import numpy as np
+from numpy.typing import NDArray
```
tylerjereddy: The CI failures seem to be just because the NumPy in use is too old for this, but I think the bump to `1.20.0` minimum is imminent anyway.

package/MDAnalysis/visualization/streamlines.py
```diff
@@ -61,8 +63,7 @@
 import MDAnalysis
 
 
-
-def produce_grid(tuple_of_limits, grid_spacing):
+def produce_grid(tuple_of_limits: Tuple, grid_spacing: float) -> NDArray:
```
tylerjereddy: I think here you might want to use `Sequence` instead of `Tuple`. I used to use `Tuple` myself in these situations (arguments that accept tuples) until a student pointed out this [from the docs](https://mypy.readthedocs.io/en/stable/kinds_of_types.html):

> Usually it’s a better idea to use Sequence[T] instead of tuple[T, ...], as [Sequence](https://docs.python.org/3/library/typing.html#typing.Sequence) is also compatible with lists and other non-tuple sequences.

I believe that should be safe here since other sequences would also support the unpacking safely?

package/MDAnalysis/visualization/streamlines.py
```diff
@@ -85,7 +86,7 @@ def produce_grid(tuple_of_limits, grid_spacing):
     return grid
 
 
-def split_grid(grid, num_cores):
+def split_grid(grid: NDArray, num_cores: int) -> Tuple[List[List[int]], List[List[int]], int, int]:
```
tylerjereddy: Isn't the return type a `List` instead of `Tuple` here?

package/MDAnalysis/visualization/streamlines.py
```diff
@@ -172,7 +174,7 @@ def produce_list_indices_point_in_polygon_this_frame(vertex_coord_list):
             list_indices_point_in_polygon.append(index_list_in_polygon)
         return list_indices_point_in_polygon
 
-    def produce_list_centroids_this_frame(list_indices_in_polygon):
+    def produce_list_centroids_this_frame(list_indices_in_polygon: List[int]) -> List[float]:
```
tylerjereddy: `Sequence[int]` instead of `List[int]` in the args, for same reason, maybe?

Also, for the return value, I see a comment indicating that `list_centroids_this_frame` may contain arrays instead of just floats? And looking at the control flow, `list_centroids_this_frame` could also contain the string `empty`? So, maybe a bit more complex than `List[float]`?

package/MDAnalysis/visualization/streamlines.py
umak1106: Yes I made a mistake out there

package/MDAnalysis/visualization/streamlines.py
umak1106: Sequence[Optional[float]] but Optional gives None so i dont know  if it will return empty string

umak1106:
Should the arguments for 
def update_dictionary_point_in_cube_start_frame(array_simulation_particle_coordinates,dictionary_cube_data_this_core):
be Sequence[float] and Dict[float] ?

package/MDAnalysis/visualization/streamlines.py
```diff
@@ -85,7 +86,8 @@ def produce_grid(tuple_of_limits, grid_spacing):
     return grid
 
 
-def split_grid(grid, num_cores):
+# typing: numpy
+def split_grid(grid: np.ndarray, num_cores: int) -> List[List[List[int]], List[List[int]], int, int]:
```
jbarnoud: A list does not let you annotate each value of it, only a tuple does. There is no reason for this function to return a list; returning a tuple should be OK.

package/MDAnalysis/visualization/streamlines.py
```diff
@@ -152,8 +154,10 @@ def split_grid(grid, num_cores):
     return [list_square_vertex_arrays_per_core, list_parent_index_values, current_row, current_column]
```
jbarnoud: This is where you would be returning a tuple instead of a list.
```suggestion
    return (list_square_vertex_arrays_per_core, list_parent_index_values, current_row, current_column)
```

Assuming it passes the test, it means you should also update the docstring and ass a `.. version-changed::` note.

package/MDAnalysis/visualization/streamlines.py
jbarnoud: Are the 2 first return values really lists of lists? The docstring talks about arrays of arrays, and the numpy doc let me think they may be lists of arrays (<https://numpy.org/doc/stable/reference/generated/numpy.split.html>). Anyway, the docstring and/or the annotation need fixing.

package/MDAnalysis/visualization/streamlines.py
```diff
@@ -172,7 +176,7 @@ def produce_list_indices_point_in_polygon_this_frame(vertex_coord_list):
             list_indices_point_in_polygon.append(index_list_in_polygon)
         return list_indices_point_in_polygon
 
-    def produce_list_centroids_this_frame(list_indices_in_polygon):
+    def produce_list_centroids_this_frame(list_indices_in_polygon: Sequence[int]) -> Sequence[Optional[float]]:
```
jbarnoud: You want to be broad when annotating the input so you can accept anything valid from the user, but you want to be specific for the output so the user knows exactly what the function returns. Here you seem to be returning a list of iterators.

Also, `Optional` means a value can be `None`, here it looks like it can be the string `"empty"` instead. You need a `Union[float, str]` rather than `Optional`.

package/MDAnalysis/visualization/streamlines.py
```diff
@@ -226,8 +230,10 @@ def produce_list_centroids_this_frame(list_indices_in_polygon):
     return list(zip(reconstruction_index_list, xy_deltas_to_write.tolist()))
 
 
-def generate_streamlines(topology_file_path, trajectory_file_path, grid_spacing, MDA_selection, start_frame,
-                         end_frame, xmin, xmax, ymin, ymax, maximum_delta_magnitude, num_cores='maximum'):
+def generate_streamlines(topology_file_path: str, trajectory_file_path: str, grid_spacing: float, MDA_selection: str,
+                         start_frame: int, end_frame: int,
+                         xmin: float, xmax: float, ymin: float, ymax: float, maximum_delta_magnitude: float,
+                         num_cores: int = 'maximum') -> Tuple[Sequence[float], Sequence[float], float, float]:
```
jbarnoud: Here too, the annotations for the return values need to be specific. The sequences seem to be arrays.

package/MDAnalysis/visualization/streamlines_3D.py
```diff
@@ -331,9 +339,10 @@ def produce_coordinate_arrays_single_process(topology_file_path, trajectory_file
     return (start_frame_relevant_particle_coordinate_array_xyz, end_frame_relevant_particle_coordinate_array_xyz)
 
 
-def generate_streamlines_3d(topology_file_path, trajectory_file_path, grid_spacing, MDA_selection, start_frame,
-                            end_frame, xmin, xmax, ymin, ymax, zmin, zmax, maximum_delta_magnitude=2.0,
-                            num_cores='maximum'):
+def generate_streamlines_3d(topology_file_path: str, trajectory_file_path: str, grid_spacing: float, MDA_selection: str,
+                            start_frame: int, end_frame: int, xmin: float, xmax: float, ymin: float, ymax: float,
+                            zmin: float, zmax: float, maximum_delta_magnitude: float=2.0,
+                            num_cores: int ='maximum') -> Tuple[Sequence[float], Sequence[float], Sequence[float]]:
```
jbarnoud: Specific returns.

jbarnoud (changes requested):
I will probably need another run of review for this one, but here are already some comments. The functions in these files are very long which makes the code more difficult to follow.

package/MDAnalysis/visualization/streamlines.py
```diff
@@ -61,8 +62,7 @@
 import MDAnalysis
 
 
-
-def produce_grid(tuple_of_limits, grid_spacing):
+def produce_grid(tuple_of_limits: Sequence[float], grid_spacing: float) -> np.ndarray:
```
jbarnoud:
```suggestion
def produce_grid(
    tuple_of_limits: Sequence[float],
    grid_spacing: float
) -> np.ndarray:
```

package/MDAnalysis/visualization/streamlines.py
```diff
@@ -85,7 +85,7 @@ def produce_grid(tuple_of_limits, grid_spacing):
     return grid
 
 
-def split_grid(grid, num_cores):
+def split_grid(grid: np.ndarray, num_cores: int) -> Tuple[List[np.ndarray], List[np.ndarray], int, int]:
```
jbarnoud:
```suggestion
def split_grid(
    grid: np.ndarray,
    num_cores: int
) -> Tuple[List[np.ndarray], List[np.ndarray], int, int]:
```

package/MDAnalysis/visualization/streamlines.py
```diff
@@ -149,11 +149,13 @@ def split_grid(grid, num_cores):
...
-    return [list_square_vertex_arrays_per_core, list_parent_index_values, current_row, current_column]
+    return (list_square_vertex_arrays_per_core, list_parent_index_values, current_row, current_column)
 
 
-def per_core_work(topology_file_path, trajectory_file_path, list_square_vertex_arrays_this_core, MDA_selection,
-                  start_frame, end_frame, reconstruction_index_list, maximum_delta_magnitude):
+def per_core_work(topology_file_path: str, trajectory_file_path: str,
+                  list_square_vertex_arrays_this_core: List[List[float]],
+                  MDA_selection: str, start_frame: int, end_frame: int, reconstruction_index_list: Sequence[int],
+                  maximum_delta_magnitude: float) -> None:
```
jbarnoud:
```suggestion
def per_core_work(
    topology_file_path: str, trajectory_file_path: str,
    list_square_vertex_arrays_this_core: List[List[float]],
    MDA_selection: str, start_frame: int, end_frame: int,
    reconstruction_index_list: Sequence[int],
    maximum_delta_magnitude: float,
) -> List[Tuple[int, List[float]]]:
```

I think I get the return type correct here, but the code is a bit convoluted. In any case, this function *does* return something as I wrote you in an email.

package/MDAnalysis/visualization/streamlines.py
```diff
@@ -172,7 +174,7 @@ def produce_list_indices_point_in_polygon_this_frame(vertex_coord_list):
             list_indices_point_in_polygon.append(index_list_in_polygon)
         return list_indices_point_in_polygon
 
-    def produce_list_centroids_this_frame(list_indices_in_polygon):
+    def produce_list_centroids_this_frame(list_indices_in_polygon: Sequence[int]) -> Sequence[Union[float, str]]:
```
jbarnoud:
```suggestion
    def produce_list_centroids_this_frame(
        list_indices_in_polygon: Iterable[int]
    ) -> List[Union[np.ndarray, str]]:
```

`list_indices_in_polygon` only needs to be an iterator, it does not need to be a sequence. The difference between the two is that an iterator only promises that you can run a for-loop once (which is what we do here), while a sequence promises you can run the for-loop more than once and you can index the object (which you do not need).

Regarding the return, you explicitly return a list so let's be explicit in the annotation, so a potential caller that explicitly needs a list is happy. Also, I think the values in the list are either the string "empty" or an array.

package/MDAnalysis/visualization/streamlines.py
```diff
@@ -226,8 +228,10 @@ def produce_list_centroids_this_frame(list_indices_in_polygon):
     return list(zip(reconstruction_index_list, xy_deltas_to_write.tolist()))
 
 
-def generate_streamlines(topology_file_path, trajectory_file_path, grid_spacing, MDA_selection, start_frame,
-                         end_frame, xmin, xmax, ymin, ymax, maximum_delta_magnitude, num_cores='maximum'):
+def generate_streamlines(topology_file_path: str, trajectory_file_path: str, grid_spacing: float, MDA_selection: str,
+                         start_frame: int, end_frame: int,
+                         xmin: float, xmax: float, ymin: float, ymax: float, maximum_delta_magnitude: float,
+                         num_cores: int = 'maximum') -> Tuple[np.ndarray, np.ndarray, float, float]:
```
jbarnoud:
```suggestion
def generate_streamlines(
    topology_file_path: str, trajectory_file_path: str,
    grid_spacing: float, MDA_selection: str,
    start_frame: int, end_frame: int,
    xmin: float, xmax: float, ymin: float, ymax: float,
    maximum_delta_magnitude: float,
    num_cores: int = 'maximum'
) -> Tuple[np.ndarray, np.ndarray, float, float]:
```

package/MDAnalysis/visualization/streamlines_3D.py
```diff
@@ -51,8 +51,11 @@
 import scipy.spatial.distance
 
 import MDAnalysis
+from typing import Tuple, List, Dict, Sequence, Union, Type
 
-def determine_container_limits(topology_file_path, trajectory_file_path, buffer_value):
+
+def determine_container_limits(topology_file_path: str,
+                               trajectory_file_path: str, buffer_value: float) -> Tuple[float, ...]:
```
jbarnoud:
```suggestion
def determine_container_limits(
    topology_file_path: str,
    trajectory_file_path: str,
    buffer_value: float,
) -> Tuple[float, ...]:
```

package/MDAnalysis/visualization/streamlines_3D.py
```diff
@@ -84,7 +87,8 @@ def determine_container_limits(topology_file_path, trajectory_file_path, buffer_
     return tuple_of_limits
 
 
-def produce_grid(tuple_of_limits, grid_spacing):
+# typing: numpy
+def produce_grid(tuple_of_limits: Tuple[float, ...], grid_spacing: float) -> np.ndarray:
```
jbarnoud:
```suggestion
def produce_grid(
    tuple_of_limits: Tuple[float, ...], grid_spacing: float
) -> np.ndarray:
```

package/MDAnalysis/visualization/streamlines_3D.py
```diff
@@ -108,7 +112,8 @@ def produce_grid(tuple_of_limits, grid_spacing):
     return grid
 
 
-def split_grid(grid, num_cores):
+# typing: numpy
+def split_grid(grid: np.ndarray, num_cores: int) -> Tuple[List[Dict], int, int, Tuple]:
```
jbarnoud:
```suggestion
def split_grid(
    grid: np.ndarray, num_cores: int
) -> Tuple[List[Dict], int, int, Tuple]:
```

package/MDAnalysis/visualization/streamlines_3D.py
```diff
@@ -222,8 +227,9 @@ def split_grid(grid, num_cores):
     return list_dictionaries_for_cores, total_cubes, num_sheets, delta_array_shape
 
 
-def per_core_work(start_frame_coord_array, end_frame_coord_array, dictionary_cube_data_this_core, MDA_selection,
-                  start_frame, end_frame):
+def per_core_work(start_frame_coord_array: Sequence[int], end_frame_coord_array: Sequence[int],
+                  dictionary_cube_data_this_core: Dict, MDA_selection: str,
+                  start_frame: int, end_frame: int) -> None:
```
jbarnoud:
```suggestion
def per_core_work(
    start_frame_coord_array: Sequence[int],
    end_frame_coord_array: Sequence[int],
    dictionary_cube_data_this_core: Dict,
    MDA_selection: str,
    start_frame: int,
    end_frame: int
) -> None:
```

package/MDAnalysis/visualization/streamlines_3D.py
```diff
@@ -232,7 +238,9 @@ def per_core_work(start_frame_coord_array, end_frame_coord_array, dictionary_cub
     list_previous_frame_indices = []
     # define some utility functions for trajectory iteration:
 
-    def point_in_cube(array_point_coordinates, list_cube_vertices, cube_centroid):
+    # typing: numpy
+    def point_in_cube(array_point_coordinates: Sequence[float], list_cube_vertices: Sequence[float],
+                      cube_centroid) -> np.ndarray:
```
jbarnoud:
```suggestion
    def point_in_cube(
        array_point_coordinates: Sequence[float],
        list_cube_vertices: Sequence[float],
        cube_centroid: np.ndarray,
    ) -> np.ndarray:
```

`cube_centroid` was not annotated, from what I see it is an array.

package/MDAnalysis/visualization/streamlines_3D.py
```diff
@@ -306,8 +314,8 @@ def update_dictionary_end_frame(array_simulation_particle_coordinates, dictionar
     return dictionary_cube_data_this_core
 
 
-def produce_coordinate_arrays_single_process(topology_file_path, trajectory_file_path, MDA_selection, start_frame,
-                                             end_frame):
+def produce_coordinate_arrays_single_process(topology_file_path: str, trajectory_file_path: str, MDA_selection: str,
+                                             start_frame: int, end_frame: int) -> Tuple[Sequence[int], Sequence[int]]:
```
jbarnoud:
```suggestion
def produce_coordinate_arrays_single_process(
    topology_file_path: str, trajectory_file_path: str,
    MDA_selection: str,
    start_frame: int, end_frame: int,
) -> Tuple[np.ndarray, np.ndarray]:
```

From what I can see, you got the return type wrong.  Both returned values are coordinate arrays.

package/MDAnalysis/visualization/streamlines_3D.py
jbarnoud:
```suggestion
def generate_streamlines_3d(
    topology_file_path: str, trajectory_file_path: str,
    grid_spacing: float, MDA_selection: str,
    start_frame: int, end_frame: int,
    xmin: float, xmax: float,
    ymin: float, ymax: float,
    zmin: float, zmax: float,
    maximum_delta_magnitude: float=2.0,
    num_cores: int ='maximum'
) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
```

I fixed the return annotation.

jbarnoud (changes requested):
I made some suggestions to address the too-long lines flagged in https://github.com/MDAnalysis/mdanalysis/pull/3746#issuecomment-1173715280. One issue with type annotations is that they make the function signatures much more verbose leading to very long lines that need to be split.

I also fixed a couple of annotations that looked wrong or too strict.

Did you address all the comments from my last email (from July 19th)? I spotted at least one of my comments that was not addressed (one function that should have had a return annotation) so there may be others.

PR #3752, opened 2022-07-13
Type hint for analysis module

umak1106:
Fixes #

 - 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


package/MDAnalysis/analysis/align.py
```diff
@@ -943,32 +951,33 @@ def _conclude(self):
...
         wmsg = ("The `rmsd` attribute was deprecated in MDAnalysis 2.0.0 "
                 "and will be removed in MDAnalysis 3.0.0. Please use "
                 "`results.rmsd` instead.")
         warnings.warn(wmsg, DeprecationWarning)
         return self.results.rmsd
 
 
-def sequence_alignment(mobile, reference, match_score=2, mismatch_penalty=-1,
-                       gap_penalty=-2, gapextension_penalty=-0.1):
+def sequence_alignment(mobile: Type[AtomGroup], reference: Type[AtomGroup], match_score: float = 2,
```
tylerjereddy: Sorry if I'm just out of the loop on this--doesn't `Type[AtomGroup]` accept `type(AtomGroup)` per https://docs.python.org/3/library/typing.html#typing.Type instead of an instance? It looks like it would also accept the actual *class object* rather than instance--is that what we'd want?

umak1106: Actually , even I wasn't really sure if I should Type hint it as Atomgroup or Type[Atomgroup] 😅.

jbarnoud: `Type[AtomGroup]` is the `AtomGroup` class. You may need that in some places in core, but here you need an `AtomGroup` object/instance so just `AtomGroup`.

package/MDAnalysis/analysis/align.py
```diff
@@ -207,7 +211,8 @@
 logger = logging.getLogger('MDAnalysis.analysis.align')
 
 
-def rotation_matrix(a, b, weights=None):
+def rotation_matrix(a: npt.ArrayLike, b: npt.ArrayLike,
+                    weights: Optional[npt.ArrayLike] = None) -> Tuple[np.ndarray, float]:
```
jbarnoud: Can you please add a comment so that we can come back to the annotation when we can use `npt.NDArray`.

package/MDAnalysis/analysis/align.py
```diff
@@ -287,8 +292,9 @@ def rotation_matrix(a, b, weights=None):
     return rot.reshape(3, 3), rmsd
 
 
-def _fit_to(mobile_coordinates, ref_coordinates, mobile_atoms,
-            mobile_com, ref_com, weights=None):
+def _fit_to(mobile_coordinates: np.ndarray, ref_coordinates: np.ndarray, mobile_atoms: Type[AtomGroup],
```
jbarnoud: Same as above, please add a comment so we can find the `np.ndarray` later.

package/MDAnalysis/analysis/align.py
```diff
@@ -346,9 +352,10 @@ def _fit_to(mobile_coordinates, ref_coordinates, mobile_atoms,
     return mobile_atoms, min_rmsd
 
 
-def alignto(mobile, reference, select=None, weights=None,
-            subselection=None, tol_mass=0.1, strict=False,
-            match_atoms=True):
+def alignto(mobile: Union[Type[Universe], Type[AtomGroup]], reference: Union[Type[Universe], Type[AtomGroup]],
```
jbarnoud: None of these should be `Type`. You need a universe, not the `Universe` class.

package/MDAnalysis/analysis/align.py
```diff
@@ -943,32 +951,33 @@ def _conclude(self):
             logging.disable(logging.NOTSET)
 
     @property
-    def universe(self):
+    def universe(self) -> Type[Universe]:
```
jbarnoud: Not `Type`.

package/MDAnalysis/analysis/align.py
```diff
@@ -943,32 +951,33 @@ def _conclude(self):
...
         wmsg = ("The `universe` attribute was deprecated in MDAnalysis 2.0.0 "
                 "and will be removed in MDAnalysis 3.0.0. Please use "
                 "`results.universe` instead.")
         warnings.warn(wmsg, DeprecationWarning)
         return self.results.universe
 
     @property
-    def positions(self):
+    def positions(self) -> np.ndarray:
```
jbarnoud, line 983: Please add a comment so we can find it back.

package/MDAnalysis/analysis/align.py
jbarnoud: Not `Type`s.

package/MDAnalysis/analysis/align.py
```diff
@@ -1238,7 +1248,8 @@ def resid(nseq, ipos, t=t):
     return {'reference': ref_selection, 'mobile': target_selection}
 
 
-def get_matching_atoms(ag1, ag2, tol_mass=0.1, strict=False, match_atoms=True):
+def get_matching_atoms(ag1: Type[AtomGroup], ag2: Type[AtomGroup], tol_mass: float = 0.1, strict: bool = False,
```
jbarnoud: Not `Type`s.

package/MDAnalysis/analysis/bat.py
```diff
@@ -179,16 +179,19 @@ class to calculate dihedral angles for a given set of atoms or residues
...
 from MDAnalysis.lib.mdamath import make_whole
+from typing import List, Union, Optional
 
 from ..due import due, Doi
 
 logger = logging.getLogger(__name__)
 
 
-def _sort_atoms_by_mass(atoms, reverse=False):
+def _sort_atoms_by_mass(atoms: AtomGroup, reverse: bool = False) -> List[AtomGroup]:
```
jbarnoud:
```suggestion
def _sort_atoms_by_mass(atoms: AtomGroup, reverse: bool = False) -> List[Atom]:
```

package/MDAnalysis/analysis/contacts.py
```diff
@@ -390,8 +393,10 @@ class Contacts(AnalysisBase):
        :class:`Contacts` accepts both AtomGroup and string for `select`
     """
 
-    def __init__(self, u, select, refgroup, method="hard_cut", radius=4.5,
-                 pbc=True, kwargs=None, **basekwargs):
+    def __init__(self, u: Type[Universe], select: Union[Tuple[Type[AtomGroup], Type[AtomGroup]], Tuple[str, str]],
+                 refgroup: Tuple[Type[AtomGroup], Tuple[AtomGroup]],
```
jbarnoud: Not `Type`s.

package/MDAnalysis/analysis/contacts.py
```diff
@@ -390,8 +393,10 @@ class Contacts(AnalysisBase):
...
+                 method: Union[str, Callable[[np.ndarray, np.ndarray], int]] = "hard_cut", radius: float = 4.5,
```
jbarnoud: The annotation for the callable is incorrect, but it is very difficult (or perhaps even not possible) to fix right.

The `method` argument expects a method name or a function that takes at least 2 arrays. The function can accept other arguments that will be provided using the `**kwargs` mechanism later on. The extra arguments are passed with the `kwargs` argument. The actual call is done here:

https://github.com/MDAnalysis/mdanalysis/blob/5e8d5f6f957418e3ca74812b6c48870c78e0bbd5/package/MDAnalysis/analysis/contacts.py#L505

The following signature are all correct:

```python
def a(x, y) -> out: ...  # with kwargs={}
def b(x, y, z) -> out: ...  # with kwagrs={'z': 42}
def c(x, y, z=0) -> out: # with kwargs={} or kwagrs={'z': 42}
```

The current annotation (`Callable[[np.ndarray, np.ndarray], int]`) only matches the first signature (`a`). It does not match the others. My understanding of [this StackOverflow post](https://stackoverflow.com/questions/61569324/type-annotation-for-callable-that-takes-kwargs) and [that part of the python doc](https://docs.python.org/3/library/typing.html#typing.ParamSpec) let me think it is not possible to write an annotation that really matches what we want on only that. `typing.ParamSpec` seems to be close and would let us link the annotation of `kwargs` with the annotation of `method` but still does not let us do what we want.

From there, I would suggest being too broad and taking the risk of accepting functions that do not match rather than not accepting valid ones: `Callable[..., int]`. This will accept any function as long as it returns an `int`. It is not ideal but it will avoid complaining for valid things.

package/MDAnalysis/analysis/contacts.py
```diff
@@ -516,7 +521,7 @@ def _new_selections(u_orig, selections, frame):
     return [u.select_atoms(s) for s in selections]
 
 
-def q1q2(u, select='all', radius=4.5):
+def q1q2(u: Type[Universe], select: str = 'all', radius: float = 4.5) -> Type[Contacts]:
```
jbarnoud: Not `Type`s.

package/MDAnalysis/analysis/density.py
```diff
@@ -398,10 +401,11 @@ class DensityAnalysis(AnalysisBase):
 
     """
 
-    def __init__(self, atomgroup, delta=1.0,
-                 metadata=None, padding=2.0,
-                 gridcenter=None,
-                 xdim=None, ydim=None, zdim=None):
+    # typing : numpy
+    def __init__(self, atomgroup: AtomGroup, delta: float = 1.0,
+                 metadata=None, padding: float = 2.0,
```
jbarnoud: `metadata` seems to be `Optional[dict]` but it also seems not to be used.

@MDAnalysis/coredevs does anybody knows about that argument? It appears to not be used at all. Later on it is mentioned in the doc of the `Density` class but I do not see how it would be passed on. It looks to me like `metadata` got deprecated, but I cannot see anything about it in the docstring.

package/MDAnalysis/analysis/dielectric.py
```diff
@@ -117,14 +118,14 @@ class DielectricConstant(AnalysisBase):
 
     .. versionadded:: 2.1.0
     """
-    def __init__(self, atomgroup, temperature=300, make_whole=True, **kwargs):
+    def __init__(self, atomgroup: AtomGroup, temperature: float = 300, make_whole: bool = True, **kwargs: int) -> None:
```
jbarnoud:
```suggestion
    def __init__(self, atomgroup: AtomGroup, temperature: float = 300, make_whole: bool = True, **kwargs) -> None:
```

The keyword arguments are not ints. They can be anything that AnalysisBase will accept.

package/MDAnalysis/analysis/diffusionmap.py
```diff
@@ -247,8 +249,8 @@ class DistanceMatrix(AnalysisBase):
     .. versionchanged:: 2.2.0
          :class:`DistanceMatrix` now also accepts `AtomGroup`.
     """
-    def __init__(self, universe, select='all', metric=rmsd, cutoff=1E0-5,
-                 weights=None, **kwargs):
+    def __init__(self, universe: Union[Universe, AtomGroup], select: str = 'all', metric=rmsd, cutoff: float = 1E0-5,
+                 weights: Optional[Sequence] = None, **kwargs: int) -> None:
```
jbarnoud:
```suggestion
                 weights: Optional[npt.ArrayLike] = None, **kwargs) -> None:
```

package/MDAnalysis/analysis/diffusionmap.py
```diff
@@ -247,8 +249,8 @@ class DistanceMatrix(AnalysisBase):
     .. versionchanged:: 2.2.0
          :class:`DistanceMatrix` now also accepts `AtomGroup`.
     """
-    def __init__(self, universe, select='all', metric=rmsd, cutoff=1E0-5,
-                 weights=None, **kwargs):
+    def __init__(self, universe: Union[Universe, AtomGroup], select: str = 'all', metric=rmsd, cutoff: float = 1E0-5,
```
jbarnoud:
```suggestion
    def __init__(self, universe: Union[Universe, AtomGroup], select: str = 'all', metric: Callable[..., float] = rmsd, cutoff: float = 1E0-5,
```

I tried being more precise with the annotation of `method`, but I did not figure out something that would accept anything that can be called there and still accept more aguments: https://github.com/MDAnalysis/mdanalysis/blob/5e8d5f6f957418e3ca74812b6c48870c78e0bbd5/package/MDAnalysis/analysis/diffusionmap.py#L282

package/MDAnalysis/analysis/dihedrals.py
```diff
@@ -404,8 +407,8 @@ class Ramachandran(AnalysisBase):
 
     """
 
-    def __init__(self, atomgroup, c_name='C', n_name='N', ca_name='CA',
-                 check_protein=True, **kwargs):
+    def __init__(self, atomgroup: Union[AtomGroup, ResidueGroup], c_name: str = 'C', n_name: str = 'N',
+                 ca_name: str = 'CA', check_protein: bool = True, **kwargs: int) -> None:
```
jbarnoud:
```suggestion
                 ca_name: str = 'CA', check_protein: bool = True, **kwargs) -> None:
```

package/MDAnalysis/analysis/dihedrals.py
```diff
@@ -469,10 +472,10 @@ def _single_frame(self):
         phi_psi = [(phi, psi) for phi, psi in zip(phi_angles, psi_angles)]
         self.results.angles.append(phi_psi)
 
-    def _conclude(self):
+    def _conclude(self) -> None:
         self.results.angles = np.rad2deg(np.array(self.results.angles))
 
-    def plot(self, ax=None, ref=False, **kwargs):
+    def plot(self, ax: Optional[Axes] = None, ref: Optional[bool] = False, **kwargs: Optional[int]) -> Axes:
```
jbarnoud:
```suggestion
    def plot(self, ax: Optional[Axes] = None, ref: Optional[bool] = False, **kwargs) -> Axes:
```

package/MDAnalysis/analysis/distances.py
```diff
@@ -50,15 +50,19 @@
...
 
 
 import warnings
 import logging
 logger = logging.getLogger("MDAnalysis.analysis.distances")
 
 
-def contact_matrix(coord, cutoff=15.0, returntype="numpy", box=None):
+def contact_matrix(coord: npt.ArrayLike, cutoff: float = 15.0, returntype: str = "numpy",
+                   box: Optional[npt.ArrayLike] = None) -> npt.ArrayLike:
```
jbarnoud:
```suggestion
                   box: Optional[npt.ArrayLike] = None) -> Union[np.ndarray, scipy.sparse.base.spmatrix]:
```

See https://stackoverflow.com/questions/71501140/type-hinting-for-scipy-sparse-matrices

package/MDAnalysis/analysis/distances.py
```diff
@@ -120,7 +124,7 @@ def contact_matrix(coord, cutoff=15.0, returntype="numpy", box=None):
         return sparse_contacts
 
 
-def dist(A, B, offset=0, box=None):
+def dist(A: AtomGroup, B: AtomGroup, offset: Union[int, Tuple] = 0, box=None) -> np.ndarray:
```
jbarnoud: Here, `offset` is either a single `int` or 2 `int`s. In the latter case, it is used here: https://github.com/MDAnalysis/mdanalysis/blob/5e8d5f6f957418e3ca74812b6c48870c78e0bbd5/package/MDAnalysis/analysis/distances.py#L165

The strict annotation would be `Union[int, Tuple[int, int]]`. This guarantees the use of the `offset` argument cannot fail. However, it also forbids some valid cases. While it authorises `offset=2` and `offset=(3, 4)`, it does not let a user do `offset=[3, 4]` or `offset=np.array(3, 4)` which is valid in python.

A broader annotation is `Union[int, Iterable[int]]`. This does not offer the same guarantees as the stricter annotation, but it does allow everything valid. Indeed, the syntax `a, b = c` is called unpacking, and it only requires `c` to be [iterable](https://docs.python.org/3/glossary.html#term-iterable). @fiona-naughton explained what it meant on her blog a while ago: https://fiona-naughton.github.io/blog/2016/06/08/How-to-train-your-Python.

jbarnoud (changes requested):
I did not go through the whole thing yet, but here are already some comments. Some annotations involving callable seems to not be possible,

I cannot comment on the following line because you did not edit it: https://github.com/MDAnalysis/mdanalysis/blob/5e8d5f6f957418e3ca74812b6c48870c78e0bbd5/package/MDAnalysis/analysis/contacts.py#L476

The annotation for `sel` should be `Union[str, AtomGroup]` and the return is `AtmGroup`.

jbarnoud:
Do not forget to stop ignoring the module in thee mypy config file.

package/MDAnalysis/analysis/distances.py
```diff
@@ -50,15 +50,21 @@
...
+from typing import Optional, Union, Tuple, Iterable
 
 
 import warnings
 import logging
 logger = logging.getLogger("MDAnalysis.analysis.distances")
 
 
-def contact_matrix(coord, cutoff=15.0, returntype="numpy", box=None):
+def contact_matrix(coord: npt.ArrayLike, cutoff: float = 15.0,
```
jbarnoud, line 64: It looks like `coord` cannot be `npt.ArrayLike`. It is passed to the `len` function but not everything that matches `ArrayLike` has a size. It looks like we need to tighten the annotation to `np.ndarray`.

This causes the following mypy errors:

```
package/MDAnalysis/analysis/distances.py:109: error: Argument 1 to "len" has incompatible type "Union[Union[int, float, complex, str, bytes, generic], Sequence[Union[int, float, complex, str, bytes, generic]], Sequence[Sequence[Any]], _SupportsArray]"; expected "Sized"
package/MDAnalysis/analysis/distances.py:119: error: Argument 1 to "len" has incompatible type "Union[Union[int, float, complex, str, bytes, generic], Sequence[Union[int, float, complex, str, bytes, generic]], Sequence[Sequence[Any]], _SupportsArray]"; expected "Sized"
```

package/MDAnalysis/analysis/distances.py
```diff
@@ -156,7 +163,7 @@ def dist(A, B, offset=0, box=None):
     if A.atoms.n_atoms != B.atoms.n_atoms:
         raise ValueError("AtomGroups A and B do not have the same number of atoms")
     try:
-        off_A, off_B = offset
+        off_A, off_B = int(offset)
```
jbarnoud, line 166: This is not valid. The construct `a, b = c` we have here assumes that `c` is an iterator with 2 values. Before your change, `c` was `offset` which fits the requirement.

Now, `c` is `int(offset)` which cannot be built because `int` cannot work with an iterator. This causes:

```
package/MDAnalysis/analysis/distances.py:166: error: Argument 1 to "int" has incompatible type "Union[int, Tuple[Any, ...]]"; expected "Union[str, bytes, array[Any], mmap, _CData, PickleBuffer, SupportsInt, SupportsIndex, SupportsTrunc]"
```

Assuming the conversion would work, the result of `int` is a single number that cannot be unpacked into `off_A` and `off_B`. This causes:

```
package/MDAnalysis/analysis/distances.py:166: error: "int" object is not iterable
```

package/MDAnalysis/analysis/distances.py
```diff
@@ -120,7 +126,8 @@ def contact_matrix(coord, cutoff=15.0, returntype="numpy", box=None):
         return sparse_contacts
 
 
-def dist(A, B, offset=0, box=None):
+def dist(A: AtomGroup, B: AtomGroup, offset: Union[int, Tuple] = 0,
```
jbarnoud, line 129:
```suggestion
def dist(A: AtomGroup, B: AtomGroup, offset: Union[int, Tuple[int, int]] = 0,
```

This should remove the need to explicitly convert offset to int later in the function.

package/MDAnalysis/analysis/distances.py
```diff
@@ -156,7 +163,7 @@ def dist(A, B, offset=0, box=None):
     if A.atoms.n_atoms != B.atoms.n_atoms:
         raise ValueError("AtomGroups A and B do not have the same number of atoms")
     try:
-        off_A, off_B = offset
+        off_A, off_B = int(offset)
     except (TypeError, ValueError):
         off_A = off_B = int(offset)
```
jbarnoud, line 168: Mypy sees `offset` as possibly a tuple causing:

```
package/MDAnalysis/analysis/distances.py:168: error: Argument 1 to "int" has incompatible type "Union[int, Tuple[Any, ...]]"; expected "Union[str, bytes, array[Any], mmap, _CData, PickleBuffer, SupportsInt, SupportsIndex, SupportsTrunc]"
```

We may be able just to get rid of the type conversion.
```suggestion
        off_A = off_B = offset
```

jbarnoud (commented):
Sending the comments I have while I can. More comments to come.

---

```
package/MDAnalysis/analysis/data/filenames.py:110: error: Library stubs not installed for "pkg_resources" (or incompatible with Python 3.8)
package/MDAnalysis/analysis/data/filenames.py:110: note: Hint: "python3 -m pip install types-setuptools"
package/MDAnalysis/analysis/data/filenames.py:110: note: (or run "mypy --install-types" to install all missing stub packages)
package/MDAnalysis/analysis/data/filenames.py:110: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
```

This indicates that we need to install some stubs. Stubs are external type annotations for existing modules. Here the missing ones are for the setuptools library. The `types-setuptools` package should be installed before running mypy; possibly on this line: https://github.com/MDAnalysis/mdanalysis/blob/6302b4c7d0dc3fbd57c1e85e5d6fd90efddd3fd8/.github/workflows/mypy.yaml#L47

---

```
package/MDAnalysis/analysis/encore/dimensionality_reduction/DimensionalityReductionMethod.py:40: error: Module "MDAnalysis.analysis.encore.dimensionality_reduction" has no attribute "stochasticproxembed"
package/MDAnalysis/analysis/encore/clustering/ClusteringMethod.py:41: error: Module "MDAnalysis.analysis.encore.clustering" has no attribute "affinityprop"
```

`stochasticproxembed` and `affinityprop` are modules written in cython. Mypy does not see them, hence the error. Here we need to tell mypy that everything is fine by adding `# type: ignore` at the end of the lines that trigger the error. This also warrants a comment to tell future contributors why we ask mypy to ignore the lines.

---

```
package/MDAnalysis/analysis/encore/clustering/__init__.py:27: error: Name "ClusterCollection" already defined (by an import)
```

The import on the line triggering the error overwrite an import done just a few lines above, here: https://github.com/MDAnalysis/mdanalysis/blob/6302b4c7d0dc3fbd57c1e85e5d6fd90efddd3fd8/package/MDAnalysis/analysis/encore/clustering/__init__.py#L24

It should be safe to remove the first import.

---

```
 package/MDAnalysis/__init__.py:170: error: Need type annotation for "_READERS" (hint: "_READERS: Dict[<type>, <type>] = ...")
package/MDAnalysis/__init__.py:171: error: Need type annotation for "_READER_HINTS" (hint: "_READER_HINTS: Dict[<type>, <type>] = ...")
package/MDAnalysis/__init__.py:172: error: Need type annotation for "_SINGLEFRAME_WRITERS" (hint: "_SINGLEFRAME_WRITERS: Dict[<type>, <type>] = ...")
package/MDAnalysis/__init__.py:173: error: Need type annotation for "_MULTIFRAME_WRITERS" (hint: "_MULTIFRAME_WRITERS: Dict[<type>, <type>] = ...")
package/MDAnalysis/__init__.py:174: error: Need type annotation for "_PARSERS" (hint: "_PARSERS: Dict[<type>, <type>] = ...")
package/MDAnalysis/__init__.py:175: error: Need type annotation for "_PARSER_HINTS" (hint: "_PARSER_HINTS: Dict[<type>, <type>] = ...")
package/MDAnalysis/__init__.py:176: error: Need type annotation for "_SELECTION_WRITERS" (hint: "_SELECTION_WRITERS: Dict[<type>, <type>] = ...")
package/MDAnalysis/__init__.py:177: error: Need type annotation for "_CONVERTERS" (hint: "_CONVERTERS: Dict[<type>, <type>] = ...")
package/MDAnalysis/__init__.py:179: error: Need type annotation for "_TOPOLOGY_ATTRS" (hint: "_TOPOLOGY_ATTRS: Dict[<type>, <type>] = ...")
package/MDAnalysis/__init__.py:180: error: Need type annotation for "_TOPOLOGY_TRANSPLANTS" (hint: "_TOPOLOGY_TRANSPLANTS: Dict[<type>, <type>] = ...")
package/MDAnalysis/__init__.py:181: error: Need type annotation for "_TOPOLOGY_ATTRNAMES" (hint: "_TOPOLOGY_ATTRNAMES: Dict[<type>, <type>] = ...")
package/MDAnalysis/analysis/distances.py:64: error: Missing return statement
```

These are issues you fixed in your lib PR (#3729). Once that PR is merged, the errors should go away.

---

```
package/MDAnalysis/analysis/distances.py:64: error: Missing return statement
```

The `contact_matrix` function returns something different depending on the value of the `returntype` argument. To do so, the function uses an if statement. It does something if `returntype` is `"numpy"` and something else if the `returntype` is `"sparse"`. The issue, however, is when `returntype` is any other value! Then the function just does nothing! It never reaches a `return` statement, hence the very justified mypy error.

The fix, here, is to implement the missing `else` and raise a `ValueError` complaining about the value of `returntype`. Note that making this fix also means writing the test for it: in the `TestContactMatrix` class (https://github.com/MDAnalysis/mdanalysis/blob/6302b4c7d0dc3fbd57c1e85e5d6fd90efddd3fd8/testsuite/MDAnalysisTests/analysis/test_distances.py#L37) you need to add a test calling `contact_matrix` with a `returntype` other than numpy or spase and make sure an exception is raised.

---

```
package/MDAnalysis/analysis/legacy/x3dna.py:204: error: Too many values to unpack (6 expected, 12 provided)
package/MDAnalysis/analysis/legacy/x3dna.py:213: error: Too many values to unpack (6 expected, 12 provided)
```

These errors are in an unmaintained module and seem to be caused by actual code errors. Solving the issue is out of the scope of your project. You can ignore the errors by adding `# type: ignore` at the end of the problematic lines.

jbarnoud:
```
package/MDAnalysis/analysis/polymer.py:78: error: "AtomGroup" has no attribute "__iter__" (not iterable)
package/MDAnalysis/analysis/polymer.py:87: error: "AtomGroup" has no attribute "__iter__" (not iterable) 
```

There are a bunch of errors like this one. We can iterate on groups because they have getitem and len. This make them sequences that mypy does not identify as being iterable. I do not know how to solve this issue at the moment. Any idea would be welcome.

hmacdope:
Random guesses but does using 

'''python
def __iter__():
    pass 
'''

or the forward declaration as a string

'''python
ag: 'AtomGroup'
'''

jbarnoud:
@ALescoulie these 2 mypy errors are about one of your recent changes. Maybe you are interested in having a look?

```
package/MDAnalysis/analysis/nucleicacids.py:134: error: Incompatible types in assignment (expression has type "List[<nothing>]", variable has type "Type[List[Any]]")
package/MDAnalysis/analysis/nucleicacids.py:141: error: Too few arguments for "append" of "list"
```

jbarnoud:
> Random guesses but does using
> 
> '''python def **iter**(): pass '''
> 
> or the forward declaration as a string
> 
> '''python ag: 'AtomGroup' '''

I expect redefining `__iter__` to just `pass` will simply break the ability to iterate on groups.

I do not see how a forward declaration would help here. Could you elaborate?

jbarnoud:
Here is a common pattern that is annoying mypy : an attribute is set to None in init, it is given a value in prepare, single_frame and conclude assume that the attribute has a non-None value.

The assumption is correct viewed from our use case. Indeed, single_frame and conclude are only called by run, therefore always after prepare.

What mypy sees however is that the attributes may be None and we recklessly call methods on what could be None.

I see 2 possibilities here.

Either we go for our assumption and we ignore the errors. I don't like it because it is easy to miss actual errors when we are too ready to ignore them.

Or we go for the no-assumption approach from mypy and we put guards in the code to avoid tripping onto a None. But this makes the code more complex and maybe less performant.

Maybe somebody has ideas here?

package/MDAnalysis/analysis/leaflet.py
```diff
@@ -259,8 +265,9 @@ def __repr__(self):
             len(self.components))
 
 
-def optimize_cutoff(universe, select, dmin=10.0, dmax=20.0, step=0.5,
-                    max_imbalance=0.2, **kwargs):
+def optimize_cutoff(universe: Universe, select: Optional[AtomGroup, str], dmin: float = 10.0,
```
jbarnoud, line 268:
```suggestion
def optimize_cutoff(universe: Universe, select: Optional[Union[AtomGroup, str]] , dmin: float = 10.0,
```
This should solve

```
package/MDAnalysis/analysis/leaflet.py:268: error: Optional[...] must have exactly one type argument
```

ALescoulie:
@jbarnoud I'll look when I get home. It shouldn't be too hard to fix.

package/MDAnalysis/analysis/gnm.py
```diff
@@ -109,7 +112,7 @@ def _dsq(a, b):
     return np.dot(diff, diff)
 
 
-def generate_grid(positions, cutoff):
+def generate_grid(positions: np.ndarray, cutoff: float) -> np.ndarray:
```
jbarnoud, line 115:
```
package/MDAnalysis/analysis/gnm.py:144: error: Need type annotation for "grid"
package/MDAnalysis/analysis/gnm.py:152: error: Incompatible return value type (got "List[List[List[List[Any]]]]", expected "ndarray")
```

The function does not return an array. Instead, it returns a `List[List[List[np.ndarray]]]` if I am not mistaken. 

We could fix the return annotation or make the function return an array. I do not know if the latter would be considered a breaking change. @MDAnalysis/coredevs  would it?

jbarnoud, line 115: The docstring does not have a returns section. We should fix that since we are at it.

package/MDAnalysis/analysis/gnm.py
```diff
@@ -149,7 +152,7 @@ def generate_grid(positions, cutoff):
     return grid
 
 
-def neighbour_generator(positions, cutoff):
+def neighbour_generator(positions: np.ndarray, cutoff: float) -> Tuple[int, int]:
```
jbarnoud, line 155:
```
package/MDAnalysis/analysis/gnm.py:155: error: The return type of a generator function should be "Generator" or one of its supertypes
```

Also

```
package/MDAnalysis/analysis/gnm.py:446: error: "int" object is not iterable
package/MDAnalysis/analysis/gnm.py:447: error: Cannot determine type of "j_atom"
package/MDAnalysis/analysis/gnm.py:447: error: Cannot determine type of "i_atom"
package/MDAnalysis/analysis/gnm.py:448: error: Cannot determine type of "j_atom"
package/MDAnalysis/analysis/gnm.py:449: error: Cannot determine type of "i_atom"
package/MDAnalysis/analysis/gnm.py:450: error: Cannot determine type of "j_atom"
```

---

```suggestion
def neighbour_generator(positions: np.ndarray, cutoff: float) -> Generator[Tuple[int, int], None, None]:
```

This function is a *generator*, a special type of function. A generator does not return its value. Instead, it sends a value and pauses; when we go back to the generator, it unpauses until it sends its next value or ends. The sending happens with the `yield` keyword.

My use of the name "generator" is a bit confusing here. The function is called a generator function and it returns a generator. The generator is actually the object that holds the state when the generating gets paused.

This is why the annotation should be `Generator`.

Note that generators are a more complex topic than the 2 sentences summary I give here. You can learn more [in the doc](https://docs.python.org/3/library/typing.html#typing.Generator)

package/MDAnalysis/analysis/gnm.py
```diff
@@ -412,18 +415,18 @@ class closeContactGNMAnalysis(GNMAnalysis):
...
-                 universe,
-                 select='protein',
-                 cutoff=4.5,
-                 ReportVector=None,
-                 weights="size"):
+                 universe: Universe,
+                 select: str = 'protein',
+                 cutoff: float = 4.5,
+                 ReportVector: Optional[float] = None,
+                 weights: Optional[None] = "size") -> None:
```
jbarnoud, line 422:
```
package/MDAnalysis/analysis/gnm.py:422: error: Incompatible default for argument "weights" (default has type "str", argument has type "None")
```

```suggestion
                 weights: Optional[str] = "size") -> None:
```

package/MDAnalysis/analysis/gnm.py
```diff
@@ -412,18 +415,18 @@ class closeContactGNMAnalysis(GNMAnalysis):
...
     def __init__(self,
-                 universe,
-                 select='protein',
-                 cutoff=4.5,
-                 ReportVector=None,
-                 weights="size"):
+                 universe: Universe,
+                 select: str = 'protein',
+                 cutoff: float = 4.5,
+                 ReportVector: Optional[float] = None,
```
jbarnoud, line 421:
```
package/MDAnalysis/analysis/gnm.py:426: error: Argument 4 to "__init__" of "GNMAnalysis" has incompatible type "Optional[float]"; expected "Optional[str]"
```

```suggestion
                 ReportVector: Optional[str] = None,
```

package/MDAnalysis/analysis/contacts.py
```diff
@@ -325,7 +331,9 @@ def radius_cut_q(r, r0, radius):
     return hard_cut_q(r, radius)
 
 
-def contact_matrix(d, radius, out=None):
+# typing: numpy
+def contact_matrix(d: npt.ArrayLike, radius: float,
+                   out: Optional[Sequence] = None) -> np.ndarray:
```
jbarnoud, line 336:
```
package/MDAnalysis/analysis/contacts.py:356: error: Unsupported target for indexed assignment ("Sequence[Any]")
```

and

```
package/MDAnalysis/analysis/contacts.py:359: error: Incompatible return value type (got "Optional[Sequence[Any]]", expected "ndarray")
```

and maybe

```
package/MDAnalysis/analysis/contacts.py:356: error: Unsupported operand types for >= ("float" and "complex")
package/MDAnalysis/analysis/contacts.py:356: error: Unsupported operand types for >= ("float" and "Sequence[Union[int, float, complex, str, bytes, generic]]")
package/MDAnalysis/analysis/contacts.py:356: error: Unsupported operand types for >= ("float" and "Sequence[Sequence[Any]]")
package/MDAnalysis/analysis/contacts.py:356: error: Unsupported operand types for >= ("float" and "_SupportsArray")
package/MDAnalysis/analysis/contacts.py:356: note: Left operand is of type "Union[Union[int, float, complex, str, bytes, generic], Sequence[Union[int, float, complex, str, bytes, generic]], Sequence[Sequence[Any]], _SupportsArray]"
package/MDAnalysis/analysis/contacts.py:356: error: Unsupported operand types for <= ("str" and "float")
package/MDAnalysis/analysis/contacts.py:356: error: Unsupported operand types for <= ("bytes" and "float")
package/MDAnalysis/analysis/contacts.py:358: error: Unsupported operand types for >= ("float" and "complex")
package/MDAnalysis/analysis/contacts.py:358: error: Unsupported operand types for >= ("float" and "Sequence[Union[int, float, complex, str, bytes, generic]]")
package/MDAnalysis/analysis/contacts.py:358: error: Unsupported operand types for >= ("float" and "Sequence[Sequence[Any]]")
package/MDAnalysis/analysis/contacts.py:358: error: Unsupported operand types for >= ("float" and "_SupportsArray")
package/MDAnalysis/analysis/contacts.py:358: note: Left operand is of type "Union[Union[int, float, complex, str, bytes, generic], Sequence[Union[int, float, complex, str, bytes, generic]], Sequence[Sequence[Any]], _SupportsArray]"
package/MDAnalysis/analysis/contacts.py:358: error: Incompatible types in assignment (expression has type "Union[bool, Any]", variable has type "Optional[Sequence[Any]]")
package/MDAnalysis/analysis/contacts.py:358: error: Unsupported operand types for <= ("str" and "float")
package/MDAnalysis/analysis/contacts.py:358: error: Unsupported operand types for <= ("bytes" and "float")
```

---

```suggestion
                   out: Optional[np.ndarray] = None) -> np.ndarray:
```

package/MDAnalysis/analysis/bat.py
```diff
@@ -171,16 +171,19 @@ class to calculate dihedral angles for a given set of atoms or residues
...
 from MDAnalysis.lib.mdamath import make_whole
+from typing import List, Union, Optional
 
 from ..due import due, Doi
 
 logger = logging.getLogger(__name__)
 
 
-def _sort_atoms_by_mass(atoms, reverse=False):
+def _sort_atoms_by_mass(atoms: AtomGroup, reverse: bool = False) -> List[Atom]:
```
jbarnoud, line 186:
```
 package/MDAnalysis/analysis/bat.py:227: error: Argument 1 to "_sort_atoms_by_mass" has incompatible type "Generator[Any, None, None]"; expected "AtomGroup"
package/MDAnalysis/analysis/bat.py:232: error: Argument 1 to "_sort_atoms_by_mass" has incompatible type "Generator[Any, None, None]"; expected "AtomGroup"
package/MDAnalysis/analysis/bat.py:238: error: Argument 1 to "_sort_atoms_by_mass" has incompatible type "Generator[Any, None, None]"; expected "AtomGroup"
```

```suggestion
def _sort_atoms_by_mass(atoms: Union[AtomGroup, Iterable[Atom]], reverse: bool = False) -> List[Atom]:
```

package/MDAnalysis/analysis/bat.py
```diff
@@ -263,7 +266,9 @@ class BAT(AnalysisBase):
     @due.dcite(Doi("10.1002/jcc.26036"),
                description="Bond-Angle-Torsions Coordinate Transformation",
                path="MDAnalysis.analysis.bat.BAT")
-    def __init__(self, ag, initial_atom=None, filename=None, **kwargs):
+    def __init__(self, ag: Union[AtomGroup, Universe],
+                 initial_atom: Optional[AtomGroup] = None,
```
jbarnoud, line 270:
```
package/MDAnalysis/analysis/bat.py:314: error: Incompatible types in assignment (expression has type "Atom", variable has type "Optional[AtomGroup]")
```

```suggestion
                 initial_atom: Optional[Atom] = None,
```

jbarnoud (commented):
Some more comments.

---

```
package/MDAnalysis/analysis/density.py:835: error: Cannot determine type of "edges"
```

The type of the `edge` attribute should be annotated at the top of the `mda.analysis.density.Density` class to `List[np.ndarray]`

---

```
package/MDAnalysis/analysis/bat.py:305: error: Item "Universe" of "Union[AtomGroup, Universe]" has no attribute "fragments"
```

This one is going to be fun. `Universe.fragments` is a cached property defined here:
https://github.com/MDAnalysis/mdanalysis/blob/578d28be562d141b4148bbde94938af9b9f6669b/package/MDAnalysis/core/universe.py#L1328-L1330

---

```
package/MDAnalysis/analysis/bat.py:314: error: Incompatible types in assignment (expression has type "Atom", variable has type "Optional[AtomGroup]")
```

This is about that line: https://github.com/MDAnalysis/mdanalysis/blob/c5767df4b932a91a89ecbdd0895edfb65d209c53/package/MDAnalysis/analysis/bat.py#L314

Part of the issue (`AtomGroup` instead of `Atom`) is fixed in an inline comment. The most annoying issue is the `Optional` part of the error. The issue is that `initial_atom` is an optional argument of the function; as such it has to be `Optional[Atom]`. Later on, it is treated by the function not to be `None` anymore so it can be used; so when it is used, we know it is not `None`.

One way to solve the issue is to write the function in a way that mypy can follow the type of `initial_atom`. Another, which I think is better, is to not use the same variable for the user input and for the in-function use. Then you can type the usage function as `Atom` and simplify the function.

This also addresses
```
package/MDAnalysis/analysis/bat.py:320: error: Item "None" of "Optional[AtomGroup]" has no attribute "bonded_atoms"
```

---

```
package/MDAnalysis/analysis/bat.py:325: error: Item "Universe" of "Union[AtomGroup, Universe]" has no attribute "n_atoms"
package/MDAnalysis/analysis/bat.py:339: error: Argument 2 to "_find_torsions" has incompatible type "Union[AtomGroup, Universe]"; expected "AtomGroup"
package/MDAnalysis/analysis/bat.py:345: error: Item "Universe" of "Union[AtomGroup, Universe]" has no attribute "indices"
package/MDAnalysis/analysis/bat.py:366: error: Item "Universe" of "Union[AtomGroup, Universe]" has no attribute "n_atoms"
package/MDAnalysis/analysis/bat.py:452: error: Item "Universe" of "Union[AtomGroup, Universe]" has no attribute "n_atoms"
package/MDAnalysis/analysis/bat.py:453: error: Item "Universe" of "Union[AtomGroup, Universe]" has no attribute "n_atoms"
package/MDAnalysis/analysis/bat.py:499: error: Item "Universe" of "Union[AtomGroup, Universe]" has no attribute "n_atoms"
 package/MDAnalysis/analysis/bat.py:536: error: Item "Universe" of "Union[AtomGroup, Universe]" has no attribute "n_atoms"
package/MDAnalysis/analysis/bat.py:573: error: Incompatible return value type (got "Union[AtomGroup, Universe]", expected "AtomGroup")
```

This could be solved by changing this line:  https://github.com/MDAnalysis/mdanalysis/blob/c5767df4b932a91a89ecbdd0895edfb65d209c53/package/MDAnalysis/analysis/bat.py#L300

By setting `self._ag` to `ag.atoms` we make sure we have an `AtomGroup` even if we initially provided a `Universe`.

PR #3755, opened 2022-07-17
Add formalcharge attribute and read/write them from PDB

### LINKED ISSUE #3418: PDBParser does not save charges.
I have a PDB file containing charges according to format given in the documentation of the class. 

https://github.com/MDAnalysis/mdanalysis/blob/39850215eae5a3e2c9d28a24dbc670646e305f79/package/MDAnalysis/coordinates/PDB.py#L209-L213

However, charges are not saved as the feature was removed in 0.14 

https://github.com/MDAnalysis/mdanalysis/blob/39850215eae5a3e2c9d28a24dbc670646e305f79/package/CHANGELOG#L1540

which is also correctly stated in the documentation of the parser. Most likely, there was a reason for removing the feature but I didn't found any related issue or pr...

PDBParser should save charges.

Leave as it is but make the docs clearer and put the lack of charge parsing at a more prominent position.

IAlibay:
Fixes #3418 

 - Adds the `formalcharge` attribute - note this isn't the pretties names, so I'm 100% open to suggestions for renaming this.
 - Adds the ability to read and write formal charges from PDB files.
 - Clean up docs for the XPDB parser.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/coordinates/PDB.py
```diff
@@ -1023,6 +1023,47 @@ def _deduce_PDB_atom_name(self, atomname, resname):
...
+        np.ndarray
+            NumPy array of dtype object with strings representing the
+            formal charges of the atoms being written.
+        """
+        if charges.dtype != int:
+            raise ValueError("formal charges array should be of `int` type")
+
+        outcharges = charges.astype(object)
+        outcharges[outcharges == 0] = ''  # empty strings for no charge case
+        # using np.where is more efficient than looping in sparse cases
```
IAlibay, line 1054: We're looking at ~ 4x speedup for sparse arrays (which this will always be)

richardjgowers (commented):
Does the selection work if there's a space between the sign and the value? Eg "formalcharge + 1"

IAlibay:
> Does the selection work if there's a space between the sign and the value? Eg "formalcharge + 1"

That'll throw (edit: as I would expect it to) a SelectionError, interestingly slightly different types of errors for `+ 1` and `- 1` (in the former case it's an inability to parse `+` whilst in the latter it's that `- 1` is an invalid expression).

Ranges do work here (since that's what int selections default to), so technically `formalcharge -1-1` will select all the atoms with formal charges ranging from -1 through to +1.

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -1023,6 +1023,47 @@ def _deduce_PDB_atom_name(self, atomname, resname):
...
+            NumPy array of integers representing the formal charges of
+            the atoms being written.
+
+        Returns
+        -------
+        np.ndarray
+            NumPy array of dtype object with strings representing the
+            formal charges of the atoms being written.
+        """
+        if charges.dtype != int:
```
hmacdope: Do we allow `np.int32` here? I am not 100 % sure how the format strings work and I only ask because `np.int32` != int

package/MDAnalysis/topology/PDBParser.py
```diff
@@ -186,6 +188,10 @@ class PDBParser(TopologyReaderBase):
        are now assigned (Issue #2422).
        Aliased ``bfactors`` topologyattribute to ``tempfactors``.
        ``bfactors`` is deprecated and will be removed in 3.0 (Issue #1901)
+    .. versionadded:: 2.3.0
```
hmacdope: nitpick but `versionchanged` rather than `versionadded`

package/MDAnalysis/topology/PDBParser.py
```diff
@@ -186,6 +188,10 @@ class PDBParser(TopologyReaderBase):
...
+       Formal charges are now read from PDB files if present. No formalcharge
+       attribute is create if no formal charges are present in the PDB file.
```
hmacdope:
```suggestion
       attribute is created if no formal charges are present in the PDB file.
```

package/MDAnalysis/topology/PDBParser.py
```diff
@@ -335,6 +341,30 @@ def _parseatoms(self):
...
+                            errmsg = (f"Unknown formal charge {entry} "
+                                      "encountered")
+                            raise ValueError(errmsg)
+                    else:
+                        errmsg = (f"Formal charge {entry} is unrecognized")
+                        raise ValueError(errmsg)
+                else:
+                    formalcharges[i] = 0
+            attrs.append(
+                    FormalCharges(np.array(formalcharges, dtype=np.int32)))
```
hmacdope: See above question about `int` vs `int32`

testsuite/MDAnalysisTests/topology/test_pdb.py
```diff
@@ -325,3 +326,48 @@ def test_nobonds_error():
...
+    properly given a PDB file with a valid formal charges record.
+    """
+    u = mda.Universe(PDB_charges)
+    formal_charges = np.array([0, 0, 0, 0, 0, 0, 0, -1, 0, 0, 0, 0, 0, 0, 0,
+                               0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0,
+                               0, 0, 0, 0, 0], dtype=int)
+    assert_equal(u.atoms.formalcharges, formal_charges)
+
+
+PDB_charges_nosign = """\
```
hmacdope, line 342: Could possibly be a fixture but i guess its not recycled, up to you :)

hmacdope (changes requested):
Looking good, just a few queries. :)

package/MDAnalysis/coordinates/PDB.py
IAlibay: Yeah that's my bad, getting my np.integer crossed with the plain old int. Switching to a cleaner `np.issubdtype` call.

package/MDAnalysis/topology/PDBParser.py
IAlibay: Switching to `int` here since the topologyattr code will just re-type to this anyways unless dtype is set to None.

That being said, charges really shouldn't exceed the bounds of int8, so one wonders how wasteful this is... (can't remember if int32/int64 still has the weird behaviour of ending up faster than int8/int16 in some weird cases)

testsuite/MDAnalysisTests/topology/test_pdb.py
IAlibay: I'm going to skip on that, the advantages here are limited, especially given we use declared PDB strings like this everywhere else here.

hmacdope (approved):
Thanks for addressing comments @IAlibay looks good to roll. You can merge when ready.

IAlibay:
oops I think I got my subtypes the wrong way around

Issue #3743, opened 2022-06-30
modernize testing code


orbeckst:
## Is your feature request related to a problem? ##
The testing code uses pytest and relies on numpy.testing. It contains some deprecated testing constructs such as [`assert_almost_equal()`](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_almost_equal.html), [`assert_array_almost_equal()`](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_almost_equal.html) and some sub-optimal tests (such as testing single floats with an array assertion).

## Describe the solution you'd like ##
Update the testing code, following current best practices, such as

1. use `numpy.testing.assert_allclose()` instead of `assert_almost_equal()` or `assert_array_almost_equal()` for [more consistent floating point comparisons](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_almost_equal.html)
2. replace array comparisons for floating point scalars (a single floating point number) with [`pytest.approx()`](https://docs.pytest.org/en/4.6.x/reference.html#pytest-approx) (see https://github.com/MDAnalysis/mdanalysis/pull/3735#discussion_r907848364)
3. ...

Please add additional best practices here or in the comments.

## Describe alternatives you've considered ##
We don't _have_ to do anything right now because everything is working.

## How to work on this issue

You can open issues for individual points above, you don't have to work on all of them. If you create a new issue, mention this issue number in the issue text so that it gets linked. 

Find occurrences of the constructs that you want to change in the code, e.g.,
```bash
cd testsuite/
git grep 'assert_almost_equal'
```

Figure out _how_ to change the code to get an equivalent result by learning about the different arguments that the old and the new function take and how they affect the assertion. Then pick one example and implement your solution. See if it works. Improve your example until it works. Only then start converting more of the test files. Get feedback from other developers.

utkarsh147-del:
i want to work on this issue

orbeckst:
We always welcome contributions. Just submit a PR that references this issue. We will then discuss there. When you start the work, please also introduce you on the developer mailing list. 

Btw, we don’t “reserve” work, we just look at the first PR that addresses the issue. Details of our policy are in our GSOC FAQ on our wiki.

rzhao271:
I tried converting more tests to use `assert_allclose` instead of `assert_almost_equals`, but an issue I've run into is that for scalars, `pytest.approx` doesn't work well when the original test had a custom error message.
[...]

PR #3757, opened 2022-07-20
Modernize tests under converters and visualization

rzhao271:
Affects #3743

 - Changed `assert_almost_equal` to `assert_allclose` or `approx`

This PR only affects tests under the `converters` and `visualization` directories for now.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -13,7 +13,7 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" http://semver.org
 
 ------------------------------------------------------------------------------
-??/??/?? IAlibay, PicoCentauri, orbeckst, hmacdope, rmeli, miss77jun
+??/??/?? IAlibay, PicoCentauri, orbeckst, hmacdope, rmeli, miss77jun, rzhao271
 
  * 2.3.0
 
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -13,7 +13,7 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" http://semver.org
 
 ------------------------------------------------------------------------------
-??/??/?? IAlibay, PicoCentauri, orbeckst, hmacdope, rmeli, miss77jun
+??/??/?? IAlibay, PicoCentauri, orbeckst, hmacdope, rmeli, miss77jun, rzhao271
 
  * 2.3.0
 
@@ -36,6 +36,8 @@ Changes
     as per NEP29 (PR #3737)
   * Narrowed variable scope to reduce use of OpenMP `private` clause (PR #3706, PR
     #3728)
+  * Refactored tests under `converters` and `visualization` directories
+    to avoid using `assert_almost_equal`
 
 Deprecations
   * The extras_requires target "AMBER" for `pip install ./package[AMBER]`
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -13,7 +13,7 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" http://semver.org
 
 ------------------------------------------------------------------------------
-??/??/?? IAlibay, PicoCentauri, orbeckst, hmacdope, rmeli, miss77jun, rzhao271
+??/??/?? IAlibay, PicoCentauri, orbeckst, hmacdope, rmeli, miss77jun
 
  * 2.3.0
 
@@ -36,8 +36,6 @@ Changes
     as per NEP29 (PR #3737)
   * Narrowed variable scope to reduce use of OpenMP `private` clause (PR #3706, PR
     #3728)
-  * Refactored tests under `converters` and `visualization` directories
-    to avoid using `assert_almost_equal`
 
 Deprecations
   * The extras_requires target "AMBER" for `pip install ./package[AMBER]`
--- a/testsuite/MDAnalysisTests/converters/test_openmm.py
+++ b/testsuite/MDAnalysisTests/converters/test_openmm.py
@@ -22,7 +22,7 @@
 #
 import pytest
 import numpy as np
-from numpy.testing import assert_almost_equal
+from numpy.testing import assert_allclose
 
 from MDAnalysisTests.datafiles import PDBX
 from MDAnalysisTests.coordinates.reference import RefAdKSmall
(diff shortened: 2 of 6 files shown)


package/CHANGELOG
```diff
@@ -36,6 +36,8 @@ Changes
     as per NEP29 (PR #3737)
   * Narrowed variable scope to reduce use of OpenMP `private` clause (PR #3706, PR
     #3728)
+  * Refactored tests under `converters` and `visualization` directories
+    to avoid using `assert_almost_equal`
```
orbeckst: I am not 100% sure if we need to include internal changes here as this is not particularly important for users — @IAlibay ??

In any case, add the issue number and PR as reference, please

testsuite/MDAnalysisTests/converters/test_openmm.py
```diff
@@ -65,18 +65,19 @@ def omm_sim_uni(self):
...
+            atol=1e-3,
+            err_msg="OpenMMBasicSimulationReader failed to get unitcell " +
+                    "dimensions from OpenMM Simulation Object",
         )
 
     def test_coordinates(self, omm_sim_uni):
         up = omm_sim_uni.atoms.positions
         reference = np.ones((5, 3))
-        assert_almost_equal(up, reference, decimal=3)
+        assert_allclose(up, reference, rtol=0, atol=1e-3)
```
orbeckst, line 80: That's a good equivalent test. For coordinates it makes sense to only use atol.

testsuite/MDAnalysisTests/converters/test_openmm.py
```diff
@@ -93,21 +94,21 @@ class TestOpenMMPDBFileReader(_SingleFrameReader):
     def setUp(self):
         self.universe = mda.Universe(app.PDBFile(RefAdKSmall.filename))
         self.ref = mda.Universe(RefAdKSmall.filename)
-        self.prec = 3
```
orbeckst, line 96: This is still used elsewhere. There are now tests failing because this is missing. Please check.

orbeckst (changes requested):
Thank you for tackling this issue.

I only had time for an initial review. The primary problem is that there are now failing tests (probably because of the removal of `.prec`). Please address this problem and then ping me to review again. Thank you!

testsuite/MDAnalysisTests/converters/test_openmm.py
rzhao271: Sorry, I noticed but didn't think of reverting the PR back to a draft. Will check in a few hours!

package/CHANGELOG
IAlibay: Agreed, I would advocate for not including this in entry the changelog (sorry about that), please do keep your user name under the changelog authors though!

testsuite/MDAnalysisTests/converters/test_parmed.py
```diff
@@ -187,7 +187,13 @@ def test_equivalent_atoms(self, ref, output):
             for attr in self.almost_equal_atom_attrs:
                 ra = getattr(r, attr)
                 oa = getattr(o, attr)
-                assert_almost_equal(ra, oa, decimal=2, err_msg='atom {} not almost equal for atoms {} and {}'.format(attr, r, o))
+                assert_allclose(
+                    ra,
+                    oa,
+                    rtol=0,
+                    atol=1e-2,
+                    err_msg=('atom {} not almost equal for atoms' +
```
IAlibay: I'm not super fond of the formatting given we end up using a whole line for a two character variable. I won't block on that but I would ask to use f-strings and implicit continuation here instead of `+`

IAlibay (changes requested):
just the one thing extra from what @orbeckst already mentioned

testsuite/MDAnalysisTests/converters/test_openmm.py
```diff
@@ -65,18 +65,19 @@ def omm_sim_uni(self):
...
+        assert_allclose(
             omm_sim_uni.trajectory.ts.dimensions,
             np.array([20., 20., 20., 90., 90., 90.]),
-            3,
-            "OpenMMBasicSimulationReader failed to get unitcell dimensions " +
-            "from OpenMM Simulation Object",
+            rtol=0,
+            atol=1e-3,
+            err_msg="OpenMMBasicSimulationReader failed to get unitcell " +
+                    "dimensions from OpenMM Simulation Object",
```
IAlibay: Can you use implicit continuation here? i.e.

```suggestion
            err_msg=("OpenMMBasicSimulationReader failed to get unitcell "
                     "dimensions from OpenMM Simulation Object"),
```

testsuite/MDAnalysisTests/converters/test_openmm.py
```diff
@@ -121,18 +123,19 @@ def setUp(self):
...
-        assert_almost_equal(
+        assert_allclose(
             self.universe.trajectory.ts.dimensions,
             self.ref.trajectory.ts.dimensions,
-            self.prec,
-            "OpenMMModellerReader failed to get unitcell dimensions " +
-            "from OpenMMModeller",
+            rtol=0,
+            atol=1e-3,
+            err_msg="OpenMMModellerReader failed to get unitcell dimensions " +
```
IAlibay: As above re; implicit line continuation instead of `+`

testsuite/MDAnalysisTests/converters/test_openmm.py
```diff
@@ -154,18 +157,19 @@ def setUp(self):
...
-        assert_almost_equal(
+        assert_allclose(
             self.universe.trajectory.ts.dimensions,
             self.ref.trajectory.ts.dimensions,
-            self.prec,
-            "OpenMMSimulationReader failed to get unitcell dimensions " +
-            "from OpenMMSimulation",
+            rtol=0,
+            atol=1e-3,
+            err_msg="OpenMMSimulationReader failed to get unitcell " +
```
IAlibay: As above

testsuite/MDAnalysisTests/converters/test_parmed.py
```diff
@@ -51,38 +49,40 @@ class TestParmEdReaderGRO:
...
     def test_dimensions(self):
-        assert_almost_equal(
+        assert_allclose(
             self.universe.trajectory.ts.dimensions, 
             self.ref.trajectory.ts.dimensions,
-            self.prec,
-            "ParmEdReader failed to get unitcell dimensions from ParmEd")
+            rtol=0,
+            atol=1e-3,
+            err_msg="ParmEdReader failed to get unitcell dimensions " +
```
IAlibay: As above

IAlibay (changes requested):
just a couple of pep8 things - essentially the preference is to use implicit line continuation where possible. You could also just declare the error message before the assert_allclose call so as to make for a cleaner error message definition

orbeckst (approved):
Thank you for addressing all my comments. I'll leave it to @IAlibay to put the final approval on the PR and merge. Thank you very much for your first contribution!!

orbeckst:
@IAlibay I assigned the PR to you, just so that it's clear that you're responsible for seeing it through.

IAlibay (approved):
awesome, thanks for your contribution @rzhao271 !

PR #3777, opened 2022-08-15
Add `BondTable` to `Bonds` TopologyAttr for fast search of bonds

### LINKED ISSUE #3733: Investigate making Bonds use a C++ unordered map datastructure  and  Cython

As far as I can tell the `Bonds` datastructures and topology attributes currently use an `N*M` matrix to store Bond information where `N` is the total number of bonds and `M` is the length of the bond type. 

For example for the index array `bonds` type  is of shape [`N`, 2].

This may make getting the bonds involved in a specific atom an inefficient process  Complexity(O)?  due to needing to traverse. 

This could possibly be improved by using Cython on its own and also a different representation. 

Would a data-structure of type:
```c++
std::unordered_map<int, std::vector<int>>
```
where `int` is the atom and the `vector` contains connected atoms be more efficient?

This is similar to the idea used in `topDict` which is  stores bond information in 
```python
topDict[(1,2)] =  [Bond(blah, blah), ....]
```

This is the equivalent of the C++ type 

```c++
std::unordered_map<std::tuple<int,int>, std::vector<int>>
```

Or could we do even better by assuming a small size for the vector and using a fixed size container with a linked list tacked on. For example:

```c++
std::vector<std::array<int, 8>>
```
Where the 8th item in the vector indicates (if present) which vector element the row continues on : Complexity(O)?

This is motivated by the fact that there are not many atoms with 8+ bonds. 

## Additional context ##

Still trying to get to grips with the Topology system, so this may be updated as I figure out more.

hmacdope:
Fixes #3733 

This PR introduces a BondTable Cython based representation of Bonds. 

It uses a kind of table based lookup for the bonds and their attributes in linear arrays using spans.
I'm not sure if there is a comp-sci name (perhaps @richardjgowers knows what its called) for this pattern but basically:

1. Make sure every bond is in the linear array both forwards and backwards
2. Search through the input bonds to find the unique ones, record which unique bond each bond in the input corresponds to
3. Use a span to index the resulting representation for O(1) lookup. 

Critically this drastically improves the scaling of bond generation compared to the `defaultdict` based impmentation in `_Connection`

Some very quick tests on @yuxuanzhuang 's example bond system (https://github.com/yuxuanzhuang/MDA_benchmark_sys) were 5 fold faster (41 vs ~8 seconds~ 30 seconds) using the table based version. 

 I am working on improving the performance of several parts and can hopefully push it down as there are a few bits that are a bit sloppy still. 

 -  Added a Cython based table representation  for bonds
 -  Change `bonds` TopologyAttr to query the table repr for `get_atoms`
 
There are still a few thorns to work through:

TODO
------
- [x]  Sort out adding and deleting bonds
- [x] Track down issue with sorting of Types
- [x] Add BondTable specific tests
~~- [ ] Improve Numpy returns to Numpy C API rather than `np.asarray`~~
- [x] Trim down final API following testing.

PR Checklist
------------
 - [x] Tests?
 - [X] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/core/bondtable.pyx
```diff
@@ -0,0 +1,760 @@
...
+        for i in range(targets.shape[0]):
+            if targets[i] <= self.max_index:
+                ret_struct = self._get_pair(targets[i])
+                pair_arr = ret_struct.first
+                has_bonds = ret_struct.second
+                if has_bonds:
+                    for j in range(pair_arr.size()):
+                        bonds.push_back(pair_arr[j])
+            else:
+                pass
```
tylerjereddy: There's a bunch of explicit `else: pass` blocks--are those placeholders or just a stylistic preference over not writing the `else` at all?

hmacdope: Stylistic but I can change it if people don't like it, I know it looks a bit strange.

hmacdope:
Almost there, been tracking down some issues on my end. Hopefully finalised soon.

hmacdope:
This is ready for review now. Currently about a 2x increase in Bond  speed (see graph below) with some more to be squeezed out with further use of the NumPy C API for the array construction. 

bonds_timing

One final issue is what to do about duplicates. The table doesn't allow duplicates with different values for `type`, `guessed`, `order` etc and will just use the first value for the bond found post sorting the inputs. This isn't an issue most of the time because duplicates are removed when the `TopologyGroup` is constructed, but I would like to make sure that the behaviour matches exactly before finalising.

hmacdope:
UPDATE: Not really efficient to use NumPy C API here as the returned arrays are not attributes and are temporaries created in functions so would have to do some serious lifetime management as their base pointers would otherwise go out of scope. Much simpler and less error prone to just return a Python object.

hmacdope:
See discussion on #3818 for duplicate bond behaviour, should be agreed upon before moving forward here.

hmacdope:
@MDAnalysis/coredevs any chance of a review here?

package/MDAnalysis/core/bondtable.pxd
```diff
@@ -0,0 +1,67 @@
...
+    
+    # maximum index in the table 
+    cdef public int max_index
+    # whether the table has any values in it 
+    cdef public cbool _is_empty
+    # which bond index entry in _ix_pair_array is value in input
+    cdef public vector[int] _bix
+    # span of each atom in _bix
+    cdef public vector[int] _spans_start
+    cdef public vector[int] _spans_end
```
richardjgowers, line 48: Maybe we can merge these two span arrays

package/MDAnalysis/core/bondtable.pxd
```diff
@@ -0,0 +1,67 @@
...
+cimport numpy as cnp
+
+cnp.import_array()
+
+
+cdef class BondTable:
+    # attributes are public for testing but are not intended to be accessed
+    
+    # maximum index in the table 
+    cdef public int max_index
```
richardjgowers, line 41: why is this necessary if all our containers are `vector`

richardjgowers (changes requested):
I think I can squash the span arrays down to a single one, otherwise good

package/MDAnalysis/core/bondtable.pxd
hmacdope: IIRC the reason the spans had to be de-merged was if you have atoms without bonds in the middle of the set of bonds.
If this is the case, then using a merged numbering system as a consistent span didn't work, but you may be able to do something clever with it.

hmacdope:
@richardjgowers do you think this could possibly go in and we can improve on it later on?

package/MDAnalysis/core/topologyattrs.py
```diff
@@ -2629,6 +2629,86 @@ class Bonds(_Connection):
...
+        querying the BondTable
+
+        Parameters
+        ----------
+        ag : AtomGroup
+
+        .. versionadded:: 2.3.0
+        """
+        b_idx, types, guessed, order = self._bondtable.get_b_t_g_o_slice(ag.ix)
+        return TopologyGroup(b_idx, ag.universe,
```
jaclark5, line 2654: This line reinitializes the `TopologyGroup` every time `get_atoms` is called. Shouldn't this be cached?

hmacdope:
@richardjgowers do we want to think about continuing here or should we just close?

PR #3753, opened 2022-07-15
basic guesser features

aya9aladdin:
Fixes #

Changes made in this Pull Request:
 - I have created a guesser package that will contain all upcoming the context aware guessers. it contain the following so far:
>  a core module that only contain one method that will be used by the guess_topologyAttribute method of the universe
>  a base module that contains the `BaseGusser` class that will be inherited by every guesser
> a DefaultGuesser module which will handle the currently existing guess methods (for now, I just added the mass guessing to test with one attribute guessing method)

- I made changes to the universe class to work with the new methodology
- I removed the mass guessing from the PDBParser to test guessing it at the universe level

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?
 - [x] Issue raised/referenced?

PR #3779, opened 2022-08-20
gussed_attributes and read_attributes methods unexpected behavior

aya9aladdin:
**the `guessed_attributes` and `read_attributes` methods of the `Topology` class don't behave as expected with `bonds`, `dihedrals`, `angles`, `Impropers`, and any other attribute object with a `list` values for its `is_guessed` attribute.
This error encountered me while I was using the `guessed_attributes` method in working on some unit tests for the `guess_topology_attributes` API here #3753 
Fixes #
I added an extra check to cover the case of the `is_guessed` value being a list

------------

```python

import MDAnalysis as mda
from MDAnalysisTests.datafiles import MMTF


# MMTF file contains bond data already
u = mda.Universe(MMTF)
print(hasattr(u, 'bonds'))

# print bonds a (false positive result) 
# as guessed_attributes method checks the is_guessed list as a whole
# not the values contained inside it. That's why 'x.is_guessed' 
# always gives true
for g in u._topology.guessed_attributes:
    print(g.attrname)
 
# doesn't print bonds a (false negative result) 
# as read_attributes method checks the is_guessed list as a whole
# not the values contained inside it. That's why 'not x.is_guessed'
# will always gives false
for r in u._topology.read_attributes:
    print(r.attrname)
 

```

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


jbarnoud:
It is very good you created a separate pull request for this issue. Solving this issue is its own unit, indeed.

I need more context in the pull request description, though. When you open a pull request, you need to tell the maintainers why they should merge the changes. This means you need to describe what the issue is and what you did to solve it. Most of the time, describing the issue is done in an issue ticket that you can just link and maybe summarise a little. Here there is no issue ticket so the full description must happen in the pull request description. What is the expected behaviour? What is happening instead? How can I reproduce the bug?

Answering these questions is not only beneficial for the reviewer. It is also useful for you. Indeed, you need to write tests that validate your changes. These tests make sure that your fix indeed fixes the bug but also that future changes do not reintroduce it (what we call regression).

aya9aladdin:
> It is very good you created a separate pull request for this issue. Solving this issue is its own unit, indeed.
> 
> I need more context in the pull request description, though. When you open a pull request, you need to tell the maintainers why they should merge the changes. This means you need to describe what the issue is and what you did to solve it. Most of the time, describing the issue is done in an issue ticket that you can just link and maybe summarise a little. Here there is no issue ticket so the full description must happen in the pull request description. What is the expected behaviour? What is happening instead? How can I reproduce the bug?
> 
> Answering these questions is not only beneficial for the reviewer. It is also useful for you. Indeed, you need to write tests that validate your changes. These tests make sure that your fix indeed fixes the bug but also that future changes do not reintroduce it (what we call regression).

I updated the issue description to clarify it more. Hope it's clear enough now

package/MDAnalysis/core/topology.py
```diff
@@ -518,12 +518,16 @@ def del_TopologyAttr(self, topologyattr):
     @property
     def guessed_attributes(self):
         """A list of the guessed attributes in this topology"""
-        return filter(lambda x: x.is_guessed, self.attrs)
+        return filter(lambda x: x.is_guessed
+                      if(not isinstance(x.is_guessed, list))
```
jbarnoud: Relying on the specific `list` type is a bit fragile. If we ever change `is_guessed` to return a tuple or a numpy array, then this test will fail silently.

A less fragile approach would be to test that `x.is_guessed` is an instance of `typing.Container`. `Container` is an "abstract base class" (or ABC); every object that implements the `__contains__` method (so all the objects with which you can use `in`) will appear as an instance of `Container`.

package/MDAnalysis/core/topology.py
```diff
@@ -518,12 +518,16 @@ def del_TopologyAttr(self, topologyattr):
...
+                      else True in x.is_guessed, self.attrs)
 
     @property
     def read_attributes(self):
         """A list of the attributes read from the topology"""
-        return filter(lambda x: not x.is_guessed, self.attrs)
+        return filter(lambda x: not x.is_guessed
+                      if(not isinstance(x.is_guessed, list))
```
jbarnoud: Same as above.

jbarnoud (changes requested):
Now that I understand the problem better, your fix looks very good.

On top of the small comment I made in-line, you need to add an entry in the CHANGELOG.

It would be interesting to test the case where the bonds are a mixture of guessed and read, but I can live without that test.

aya9aladdin:
> Hello @aya9aladdin! Thanks for updating this PR. We checked the lines you've touched for [PEP 8](https://www.python.org/dev/peps/pep-0008) issues, and found:
> 
> * In the file [`testsuite/MDAnalysisTests/core/test_topology.py`](https://github.com/MDAnalysis/mdanalysis/blob/7fcd05e59bee2c4a4631e5f8f9ca1e0344a2954a/testsuite/MDAnalysisTests/core/test_topology.py):
> 
> > [Line 690:16](https://github.com/MDAnalysis/mdanalysis/blob/7fcd05e59bee2c4a4631e5f8f9ca1e0344a2954a/testsuite/MDAnalysisTests/core/test_topology.py#L690): [E713](https://duckduckgo.com/?q=pep8%20E713) test for membership should be 'not in'
> 
> ##### Comment last updated at 2022-08-27 17:09:26 UTC

this pep message seem to be outdated, the line mentioned in the error is not relevant

package/CHANGELOG
```diff
@@ -13,6 +13,17 @@ The rules for this file:
...
 ------------------------------------------------------------------------------
+
+09/16/22 aya9aladdin
+
+*2.4
+
+Fixes
+  *fixes guessed_attributes and read_attributes methods of Topology class, as 
+   those methods was giving unexpected result regarding atoms attribute objects that 
+   has a list values for its is_guessed attribute (bonds, angles, etc)
```
```suggestion
  * fixes guessed_attributes and read_attributes methods of the Topology class, as 
    those methods were giving unexpected results for some topology attributes
    (e.g. bonds, angles) (PR #3779).
```

I simplified the phrasing and referenced the pull request.

package/CHANGELOG
```diff
@@ -13,6 +13,17 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" http://semver.org
 
 ------------------------------------------------------------------------------
+
+09/16/22 aya9aladdin
```
jbarnoud: Have another look at how we do the CHANGELOG. Here you created a new version block while you only need to update the existing one.

jbarnoud (changes requested):
Just a small fix to do with the CHANGELOG and we are good :)

jbarnoud:
> this pep message seem to be outdated, the line mentioned in the error is not relevant

Looks odd indeed. Just ignore that message.

package/CHANGELOG
aya9aladdin: I've modified it, sorry for that

package/CHANGELOG
```diff
@@ -13,13 +13,17 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" http://semver.org
 
 ------------------------------------------------------------------------------
-??/??/?? IAlibay, Luthaf, hmacdope, rafaelpap, jbarnoud
+09/16/22 IAlibay, Luthaf, hmacdope, rafaelpap, jbarnoud, aya9aladdin
```
```suggestion
??/??/?? IAlibay, Luthaf, hmacdope, rafaelpap, jbarnoud, aya9aladdin
```

We only set the date when we do a release.

jbarnoud: I fixed this as I resolved the merge conflict.

jbarnoud (approved):
With the date in the CHANGELOG fixed, it is good to go! :tada:

PR #3780, opened 2022-08-20
CI: native M1 testing

tylerjereddy:
* add native M1 MacOS testing to CI
with Cirrus CI:
https://cirrus-ci.org/guide/macOS/#macos-virtual-machines

* there is no concurrency so just a single
job/Python version is added here for now

* it runs in about 6 minutes on my fork test here:
https://github.com/tylerjereddy/mdanalysis/pull/13
you can click through to inspect the logs from there

* if the team is "ok" with this, we can install the marketplace
app for Cirrus CI with MDA organization (I didn't want to do that without checking first,
so scoped the access to my fork only); just close-reopen my PR here
after doing that I think

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,15 @@
+macos_instance:
+  image: ghcr.io/cirruslabs/macos-monterey-base:latest
+
+task:
+  script: |
+    brew install python@3.10
+    brew install python-tk@3.10
+    /opt/homebrew/opt/python@3.10/bin/python3 -m venv ~/py_310
+    source ~/py_310/bin/activate
+    cd package
+    python -m pip install .
+    cd ../testsuite
+    python -m pip install .
+    python -m pip install pytest
+    python -m pytest MDAnalysisTests


.cirrus.yml
```diff
@@ -0,0 +1,15 @@
...
+    brew install python@3.10
+    brew install python-tk@3.10
+    /opt/homebrew/opt/python@3.10/bin/python3 -m venv ~/py_310
+    source ~/py_310/bin/activate
+    cd package
+    python -m pip install .
+    cd ../testsuite
+    python -m pip install .
+    python -m pip install pytest
+    python -m pytest MDAnalysisTests
```
tylerjereddy, line 15: We may get access to more than 1 core, just not more than 1 "CI job" I think, not sure. May also want to install more deps for testing, but it is a start at least..

IAlibay:
Thanks @tylerjereddy - do we have any sense of what the limitations might be for the provider? (admittedly it's completely off my radar)

IAlibay:
mmm no monthly limits but one runner. that might get problematic if we get lots of competing pushes.

(I assume this is getting triggered every push?)

Is there a way to have a GitHub actions style "cancel if there's another similar triggered workflow"?

Otherwise we might want to consider this being only triggered on merge?

tylerjereddy:
@IAlibay I can show you the provider screen shot and provide the upstream discussion link I saw below:

image

https://github.com/FFY00/meson-python/issues/131

I guess my takeaway is basically "no hard limit on minutes," but probably don't go crazy, so that's why'd I'd probably go for a single job. There's a report of a bit of a delay between job launches, but at 6 minutes we could probably absorb that anyway.

IAlibay:
Thanks @tylerjereddy I'll have a read through. I suspect the only potential worry case I have is if someone does ton of micro commits and triggers the workflow endlessly.

P.S. on a separate note, the NF infrastructure committee is interested in helping out with these things, maybe we should touch base to see how best we can help get things done on a multi org level?

IAlibay:
Sorry I'm being silly, I found the right docs for auto cancellation, it's on by default: https://cirrus-ci.org/guide/writing-tasks/#auto-cancellation-of-tasks

IAlibay (approved):
LGTM!

tylerjereddy:
Worst case scenario we try it out and if it becomes a bottleneck then maybe just use it for providing actual M1 mac wheels now and then.

orbeckst (approved):
Let's try it.

Thank you for finding new resources, @tylerjereddy .

tylerjereddy:
Alright, this has two core dev approvals and I've now added the Cirrus CI GitHub app to `MDAnalysis/mdanalysis`. I'll go ahead and merge this then--please do ping me if the new service is annoying/slow or even catches new problems, I'm curious to see how this behaves before adopting in too many other places.

PR #3787, opened 2022-08-23
Fix type annotations in lib.distances

jbarnoud:
A couple of type annotations in distances.py caused mypy to complain.
Most importantly, the check_coords decorator does some type wrangling
that is opaque to mypy.

With this commit, distances.py does not yield any mypy error anymore.

Once merged, this should get rid of the errors in #3729.

### DIFF
--- a/package/MDAnalysis/lib/distances.py
+++ b/package/MDAnalysis/lib/distances.py
@@ -100,7 +100,7 @@ except ImportError:
 del importlib
 
 
-def _run(funcname: Callable, args: Optional[tuple] = None,
+def _run(funcname: str, args: Optional[tuple] = None,
          kwargs: Optional[dict] = None, backend: str = "serial") -> Callable:
     """Helper function to select a backend function `funcname`."""
     args = args if args is not None else tuple()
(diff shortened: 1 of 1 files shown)


package/MDAnalysis/lib/distances.py
```diff
@@ -450,16 +450,23 @@ def capped_distance(reference: Union[np.ndarray, 'AtomGroup'],
...
                              "dimension should be either None or of the form "
                              "[lx, ly, lz, alpha, beta, gamma]")
-    method = _determine_method(reference, configuration, max_cutoff,
-                               min_cutoff=min_cutoff, box=box, method=method)
-    return method(reference, configuration, max_cutoff, min_cutoff=min_cutoff,
-                  box=box, return_distances=return_distances)
+
+    # The check_coords decorator made sure that reference and configuration
+    # are arrays of positions. Mypy does not know about that so we have to
+    # tell it.
```
jbarnoud, line 457: It would be better to address that directly in the decorator, but I am not sure how to do that or if it is even possible.

richardjgowers (approved):

hmacdope (approved):
Thanks @jbarnoud, much appreciated

Issue #3707, opened 2022-06-07
Multiprecision in core, NumPy dtypes and Cythonization


hmacdope:
## Is your feature request related to a problem? ##

With the ongoing work on Cythonising Core as part of the CZI performance track (see #3683),  we need to start proper discussion of `dtypes` and multi-precision .  This was first broached in #3545  which highlights some inconsistent use of NumPy dtypes and various examples of willy nilly casting to and from `float32` and `float64`. 

It is important to discuss what kinds of precision we want to support as when writing a compiled layer, the design can look very different depending on if multiple datatypes are supported and if templates need to be used or not. 

While interoperability at the `NumPy` level in compiled land is not too difficult, auxiliary  C++ datastructures that own memory such as a coordinate buffer need careful thought. 

My feeling is that while multiprecision support is not essential for speed, it is important for interoperability with other packages and for consistency throughout the package. 

For me most pressing issue is to decide whether we want to support a `dtype` argument to the core data-structures and what that would mean downstream.  I started on this for `Timestep` in #3683, but removed the `dtype` kwarg pending further discussions. 

## Current state of play

- Coordinates are always a `float32`
- Unitcell is nominally a `float32 `in `Timestep` but is promoted in the routines in `calc_distances.h`
- You cannot specify what precision your coordinates are in as an argument to any of the `Core` datastructures

## Describe the solution you'd like ##

I would like to see support for double precision and a `dtype` kwarg in the core datastructures. 

For example:

```python
import MDAnalysis as mda
u = mda.Universe(traj, top, dtype=np.float64)
ts = mda.coordinaties.timestep.Timestep(n_atoms, dtype=np.float64)
```

Of course we would preserve the current defaults of `np.float32`

We could also support **mixed precision** timesteps, where the `unitcell` and particle dependent data such as `positions`  can have different datatypes, however this is not ideal due to poor performance in regular ops and compiler vectorisation and data pipe-lining. 

Changes to support multiple precision would likley need to be coupled with template-ization of `calc_distances.h` which shouldn't be too challenging. 

## Describe alternatives you've considered ##

We stick to the current setup as we move forward with CZI stuff. 

## To discuss ##
I would love feedback from anyone,  especially the @MDAnalysis/coredevs team, as this is potentially a large change.

orbeckst:
On the surface it looks nice to be able to provide `dtype=np.float64` to the `Universe`. I like that it looks like numpy. 

Under the hood it will likely be a mess. 

[...]

PR #3797, opened 2022-08-27
#3500 issue, added optional dtype arg

tanmaymunjal:
Fixes #   Add optional dtype argument to avoid forced reduction to np.float32(Issue #3500)

 - 
Added optional dtype arg to check_box() func 

PR Checklist
------------
 - [Did it ] Tests?
 - [Updated ] Docs?
 - [Yes] CHANGELOG updated?
 - [ #3500 ] Issue raised/referenced?


package/MDAnalysis/lib/util.py
```diff
@@ -2412,13 +2412,15 @@ def check_box(box):
         triclinic and must be provided in the same format as returned by
         :attr:`MDAnalysis.coordinates.timestep.Timestep.dimensions`:
         ``[lx, ly, lz, alpha, beta, gamma]``.
+    dtype: numpy.dtype
+        The data type of the inputted and returned box matrix.
```
tylerjereddy: There's a typo here (`inputted`), but more broadly I'm not so sure we should impose any restriction on the input `dtype`. For example, someone may want to downcast, and I'm not sure we should stop them--I don't think the code currently checks for that? Maybe just documenting `dtype` similarly to `triclinic_vectors` would make sense?

tylerjereddy (commented):
Do you think it would make sense to add a regression test for respecting the return type specified? For example, this picks up the issue for one of the two functions:

```diff
--- a/testsuite/MDAnalysisTests/lib/test_util.py
+++ b/testsuite/MDAnalysisTests/lib/test_util.py
@@ -326,16 +326,17 @@ class TestMatrixOperations(object):
             box = np.zeros(6, dtype=np.float32)
         return box
 
+    @pytest.mark.parametrize('dtype', [np.float32, np.float64])
     @pytest.mark.parametrize('lengths', comb_wr([-1, 0, 1, 2], 3))
     @pytest.mark.parametrize('angles',
                              comb_wr([-10, 0, 20, 70, 90, 120, 180], 3))
-    def test_triclinic_vectors(self, lengths, angles):
-        box = lengths + angles
+    def test_triclinic_vectors(self, lengths, angles, dtype):
+        box = np.asarray(lengths + angles, dtype=dtype)
         ref = self.ref_trivecs(box)
         res = mdamath.triclinic_vectors(box)
         assert_array_equal(res, ref)
-        # check for default dtype:
-        assert res.dtype == np.float32
+        # check for dtype pass-through:
+        assert res.dtype == dtype
         # belts and braces, make sure upper triangle is always zero:
         assert not(res[0, 1] or res[0, 2] or res[1, 2])

```

That said, the number of parameters and test time do increase a fair bit for me locally with that, so perhaps Richard might prefer a smaller/simpler test to deal with the type preservation separately?

hmacdope:
We also have another open PR with this fix #3544 and ongoing discussion in   #3707 #3545. Basically we are waiting on further discussion as to whether we allow np.float64 coordinates etc.

tylerjereddy:
Good point, I suggest we close this then.

It looks like for this specific set of fairly minor changes Richard and I are both already in favor--I can't see many problems arising from these changes in isolation, even if we didn't go all in on custom dtypes everywhere. It sounds like there was also demand for multi-precision at a recent conference--could that context be added to one of the discussion issues? Probably to gh-3707?

package/MDAnalysis/lib/util.py
tanmaymunjal: 1) Probably makes sense to have some tests to it, your regression test changes seem great to me

2) Will fix the typo 

3) Don't really think there are any restrictions on downcasting dtype from np.float32 that I added in

tanmaymunjal: Made a few changes in line with the points

hmacdope:
Hi @tanmaymunjal. This PR is very similar to an existing PR with this fix #3544. We will aim to merge that first and see if there is anything outstanding that you have picked up here. Thanks for your understanding.

tanmaymunjal:
Sure, @hmacdope , np. I have also been working on #2046, lib.distances.distance_array() issue, almost done. Any idea if there is any current PR or discussion around it or should I send a PR for it?

hmacdope:
> Sure, @hmacdope , np. I have also been working on #2046, lib.distances.distance_array() issue, almost done. Any idea if there is any current PR or discussion around it or should I send a PR for it?

@tanmaymunjal which subissue on the checklist are you looking to address?

tanmaymunjal:
@hmacdope, the lib,distances.distance_array() issue and maybe the false sharing one as well,15th and 16th issue I think under the code issues headline

hmacdope:
The distance array issue no one is working on as far as I know. For openMP some changes were made recently by @RMeli  in #3706 and #3728, so some of that may have been addressed but I am not certain.

tanmaymunjal:
Ok, Thanks!

orbeckst:
@hmacdope is this PR still relevant or should we close it?

hmacdope:
> @hmacdope is this PR still relevant or should we close it?

Closeable

PR #3807, opened 2022-08-30
Avoid using set operations in distances.between

### LINKED ISSUE #3794: distances.between does not return AtomGroup
## Expected behavior ##

Following the docs analysis.distances.between should return an AtomGroup.
(https://docs.mdanalysis.org/2.2.0/documentation_pages/analysis/distances.html)

## Actual behavior ##

 if there are no atoms selected the type returned is int. In te code below actual1 is
<class 'MDAnalysis.core.groups.AtomGroup'>
whereas actual2 is
<class 'int'>

``` python
import MDAnalysis as mda
from MDAnalysis.tests.datafiles import  GRO
from MDAnalysis.analysis import distances

u = mda.Universe(GRO)
group = u.atoms[40:]
ag  = u.atoms[ 1:10]
ag2 = u.atoms[11:20]
dist1 = 6.0
dist2 = 2.0

actual1 = distances.between(
            group,
            ag,
            ag2,
            dist1
        )

print(type(actual1))

actual2 = distances.between(
            group,
            ag,
            ag2,
            dist2
        )

print(type(actual2))


```

- Which version are you using? 2.2.0
- Which version of Python (`python -V`)? Python 3.9.13
- Which operating system? "openSUSE Leap 15.3"

rafaelpap:
Fixes #3794

No set operation between groups. Use of intersection directly.

PR Checklist
------------
 - [x] Tests? Already existing tests are valid
 - [x] Docs? The change is in accord with current docs.
 - [x] CHANGELOG updated?


jbarnoud:
Thank you for your contribution! The change looks good. It would be good to have the example from the issue as a test, just to make sure the problem does not reoccur. It would be good also to add a line in the changelog and your name in the AUTHORS file.

rafaelpap:
Thanks for your comments!
1. I don't know how to make a test which check that the result of distances.between is an AtomGroup.
2. I'll add the line to Changelog.
3. My name is already in the AUTHORS file.

package/CHANGELOG
```diff
@@ -13,6 +13,9 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" http://semver.org
 
 ------------------------------------------------------------------------------
+08/30/22 rafaelpap
```
hmacdope: Thanks for the fix @rafaelpap. The CHANGELOG entries need to go underneath the version number (`2.3.0`) and are ordered newest first.   A good place to put this would be under `Fixes` :)

hmacdope (changes requested):

hmacdope:
To add a test, have a look in the `testsuite` directory of the package. In particular the file `testsuite/MDAnalysisTests/analysis/test_distances.py` is the file to work with. In particular the class [TestBetween](https://github.com/MDAnalysis/mdanalysis/blob/develop/testsuite/MDAnalysisTests/analysis/test_distances.py#L180) is the one you are looking for.

package/CHANGELOG
IAlibay: Hold for 2.4.0

rafaelpap:
I know that the test for between is in the file you mentioned.
What I don't know if how to check that the returned value is an AtomGroup and not and int, for instance.
The existing test check that the index are in the right order. Albeit looking at it now I do not understand why it uses sorted.

hmacdope:
> What I don't know if how to check that the returned value is an AtomGroup and not and int, for instance.

You could do something like 

```python
res = between(A, B)
assert isinstance(res, mda.core.groups.AtomGroup) 
```

rafaelpap:
Thanks again!
I have added a couple of test to check that the type returned is AtomGroup when the result is not empty and when is empty.
I hope to have understood how assert works.

package/CHANGELOG
```diff
@@ -14,11 +14,12 @@ The rules for this file:
 
 ------------------------------------------------------------------------------
 08/29/22 IAlibay, PicoCentauri, orbeckst, hmacdope, rmeli, miss77jun, rzhao271,
-         yuxuanzhuang, hsadia538, lilyminium
+         yuxuanzhuang, hsadia538, lilyminium, rafaelpap
 
  * 2.3.0
 
 Fixes
+  * Fixes distances.between not returning AtomGroup (Issue #3794)
```
hmacdope: Sorry I mislead you, this now needs to be under 2.4.0 which will require you to merge with develop.

testsuite/MDAnalysisTests/analysis/test_distances.py
```diff
@@ -213,13 +213,39 @@ def expected(self, group, ag, ag2):
...
+        actual = MDAnalysis.analysis.distances.between(
             group,
             ag,
             ag2,
             self.distance
-        ).indices)
+        ).indices
         assert_equal(actual, expected)
+
+    distance = 5.9
```
hmacdope: Setting this should be unnecessary here. Make this and the one below two separate variables, eg `self.distance_not_empty` and `self.distance_empty=0.0` I would also make it 0 to be super clear.

testsuite/MDAnalysisTests/analysis/test_distances.py
```diff
@@ -213,13 +213,39 @@ def expected(self, group, ag, ag2):
...
+
+    def test_between_return_type_not_empty(self, group, ag, ag2):
```
hmacdope: Alternately you could combine these into one test using `@pytest.mark.parametrize`. I would probably go for this option. :)

hmacdope (changes requested):

testsuite/MDAnalysisTests/analysis/test_distances.py
```diff
@@ -213,13 +213,25 @@ def expected(self, group, ag, ag2):
...
+    def test_between_return_type(self, dists, group, ag, ag2):
+        '''Test MDAnalysis.analysis.distances.between() for
+        returned type when returned group is not empty.'''
+        actual = MDAnalysis.analysis.distances.between(
+            group,
+            ag,
+            ag2,
+            dists
+        )
+        assert(isinstance(actual, MDAnalysis.core.groups.AtomGroup))
```
orbeckst: `assert` is a statement, not a function
```suggestion
        assert isinstance(actual, MDAnalysis.core.groups.AtomGroup)
```

orbeckst (approved):
lgtm, thanks for your contribution @rafaelpap

orbeckst:
@hmacdope could you please check if all your concerns have been addressed? If yes, please squash-merge. I am assigning you to shepherd the PR to completion.

testsuite/MDAnalysisTests/analysis/test_distances.py
```diff
@@ -213,13 +213,25 @@ def expected(self, group, ag, ag2):
...
             ag2,
             self.distance
-        ).indices)
+        ).indices
         assert_equal(actual, expected)
+
+    @pytest.mark.parametrize('dists', [5.9, 0.0])
+    def test_between_return_type(self, dists, group, ag, ag2):
+        '''Test MDAnalysis.analysis.distances.between() for
+        returned type when returned group is not empty.'''
```
hmacdope: One last tiny nitpick just on this docstring as this now test both the empty and non-empty case right?

hmacdope (changes requested):
Just one final tiny nitpick @rafaelpap

hmacdope (approved):
Thanks so much for the contribution @rafaelpap! :)

orbeckst:
Thank you, @rafaelpap  🚀

PR #3808, opened 2022-08-30
Update chemfiles tests for 0.10.3

### LINKED ISSUE #3798: `Chemfiles`  change in float writing for 0.10.3 causing CI failures
There seems to be an issue with `chemfiles` that is causing the full-deps builds to fail in #3127 and #3777 (likely others as well). I have re-triggered a few other jobs to check. 

It is worth noting that `chemfiles` released a new version on [pip](https://pypi.org/project/chemfiles/0.10.3/) yesterday (0.10.3) which could be the issue [see here for the tag](https://github.com/chemfiles/chemfiles.py/releases/tag/0.10.3). 

 I could reproduce the error (which seemed to be related to finding some shared library, at least on my computer) locally with a conda install of `chemfiles` but seemed to work OK for a pip install.

Luthaf:
Fixes #3798

 - Updated the tests for chemfiles 0.10.3, skipping them for lower versions.

PR Checklist
------------
 - [X] Tests?
 - [ ] ~Docs?~
 - [ ] ~CHANGELOG updated?~


hmacdope (approved):
Looks good to me @Luthaf. I would just add a `CHANGELOG` entry.

testsuite/MDAnalysisTests/coordinates/test_chemfiles.py
```diff
@@ -236,7 +242,10 @@ def test_write_atom_group(self, tmpdir):
             check = mda.Universe(outfile)
             assert check.trajectory.ts.n_atoms == group.n_atoms
 
-    @pytest.mark.xfail  # Issue #3798 - float writing not consisten
+    @pytest.mark.skipif(
```
IAlibay: This really isn't ideal, if anything we really shouldn't have code that does an indirect explicit float comparison like this.

The main question is - was this change in writing behaviour intended?

testsuite/MDAnalysisTests/coordinates/test_chemfiles.py
```diff
@@ -40,9 +41,9 @@
 chemfiles = pytest.importorskip("chemfiles")
 
 
-@pytest.mark.parametrize('version', ['0.9.3', '0.11.0', '1.1.0'])
+@pytest.mark.parametrize("version", ["0.9.3", "0.11.0", "1.1.0"])
 def test_version_check(version, monkeypatch):
-    monkeypatch.setattr('chemfiles.__version__', version)
+    monkeypatch.setattr("chemfiles.__version__", version)
```
IAlibay, line 47: As far as I can tell (please correct me if I'm right) these ' to " don't adhere to any given convention, why are they all being changed? (i.e. is this worth the git history  footprint)

IAlibay (changes requested):
I'm going to block for details - essentially I'd like to make sure this isn't just a temporary patch to match some arbitrary change in float writing.

testsuite/MDAnalysisTests/coordinates/test_chemfiles.py
RMeli: It's `black`'s convention: https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#strings

Luthaf: Exactly, these were automated changes by black. I can revert them if you prefer.

testsuite/MDAnalysisTests/coordinates/test_chemfiles.py
Luthaf: Yes, the change was intended on chemfiles side. I can use a different format (maybe a binary format?) than LAMMS data files to protect such small changes, since the purpose of this tests it to check velocities are passed & used by chemfiles; not the actual output.

hmacdope: IIRC its closer to how ASE writes data files but not LAMMPs itself does, but moot point anyway as this is about chemfiles support.

IAlibay: As long as it's an intended change in writing behaviour, and that chemfiles can guarantee there won't be any arbitrary (i.e. non-intentional) changes in this writing behaviour, then I'm happy.

I know I'm being a pedantic here, but it's better being clear about how things are intended to work now than to have to come back here in 3-6 months.

testsuite/MDAnalysisTests/coordinates/test_chemfiles.py
IAlibay: I'm not a fan of arbitrary (read non-PEP/NEP/*EP) line changes, especially if they affect the history. However, I will leave it to you to decide on what to do here.

testsuite/MDAnalysisTests/coordinates/test_chemfiles.py
Luthaf: I totally understand wanting to be clear! We have very similar tests on the chemfiles side (e.g. https://github.com/chemfiles/chemfiles/blob/387d4920b7eb5107ccd6050da238f82dacf1c85b/tests/formats/lammps-data.cpp#L115), so there should not be any unintentional changes here.

There might still be intentional changes that would introduce extra "noise" on the MDA side (although I can be more mindful of this), so I'll still try to write a more stable test here, maybe using a different format.

testsuite/MDAnalysisTests/coordinates/test_chemfiles.py
Luthaf: I tend to treat `black` as mostly standard nowadays, especially given it is developed by the PSF, and the output is quite stable.

testsuite/MDAnalysisTests/coordinates/test_chemfiles.py
IAlibay: Extra noise isn't as much of an issue (to me at least) to be honest (although more stable tests are always appreciated, that can be future work if you want).

testsuite/MDAnalysisTests/coordinates/test_chemfiles.py
IAlibay: Actually had a long discussion yesterday with @RMeli (or more we have the same long discussion every couple of months).

My stance is a pretty simple yet strong one re: `black`'s "style": if the PSF wants to codify it as a PEP or even a BEP, then I'd be happy adopting it & evangelizing it. Otherwise it remains a set of decisions set by an arbitrary tool choice, which makes it hard for me to go "no you should do X not Y".

In any case, given you're the code author of the lines being changed, not a hill I wish to spend more time on.

IAlibay (approved):
I'll approve early here - if you could just add yourself to the changelog for the 2.4.0 release that'd be great!

testsuite/MDAnalysisTests/coordinates/test_chemfiles.py
Luthaf: I've updated the test to use a NetCDF file instead, which can easily be checked with scipy. This should be much more stable (up to the byte!) now

Luthaf:
CHANGELOG done, I tried to follow the previous release template

package/CHANGELOG
```diff
@@ -13,12 +13,14 @@ The rules for this file:
...
 
 ------------------------------------------------------------------------------
-??/??/?? IAlibay
+??/??/?? IAlibay, Luthaf
 
  * 2.4.0
 
 Fixes
 
+  * Upgrade to chemfiles 0.10.3 (Issue #3798)
```
```suggestion
  * Upgrade to chemfiles 0.10.3 (Issue #3798)
```

IAlibay (approved):
Just the one thing

PR #3824, opened 2022-09-11
Parallelized inner instead of outer loops as requested in the #2046

tanmaymunjal:
Fixes #

 - 
Parallelized inner instead of outer loops as requested in the #2046, the second last issue under the Code issues heading, and the last issue under TODO suggestions->Code

PR Checklist
------------
 - [Done ] Tests?
 - [ ] Docs?
 - [Done ] CHANGELOG updated?
 - [ #2046 ] Issue raised/referenced?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -14,7 +14,7 @@ The rules for this file:
 
 ------------------------------------------------------------------------------
 
-??/??/?? IAlibay, Luthaf, hmacdope, rafaelpap, jbarnoud, BFedder
+??/??/?? IAlibay, Luthaf, hmacdope, rafaelpap, jbarnoud, BFedder, aya9aladdin
 
  * 2.4.0
 
@@ -23,6 +23,10 @@ Fixes
     behaviour for cutoff values < -1 (PR # 3749)
   * Fixes distances.between not always returning AtomGroup (Issue #3794)
   * Upgrade to chemfiles 0.10.3 (Issue #3798)
+  * fixes guessed_attributes and read_attributes methods of the Topology class, as 
+    those methods were giving unexpected results for some topology attributes
+    (e.g. bonds, angles) (PR #3779).
+   
 
 
 Enhancements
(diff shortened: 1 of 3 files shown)


hmacdope:
Hi @tanmaymunjal would you be able to provide some timings to show that this is faster than the old setup?

tanmaymunjal:
Hi @hmacdope, will do some performance testing on the code and get back to you in a day or two

tanmaymunjal:
Hey @hmacdope, did some performance testing on the code, the newer pull request code is showing a run time of 0.177s for 10,000 function calls while the older code is showing around 2s runtime for 10,000 function calls, around about an 11-12% increase in speed 

https://imgur.com/a/dd0pBH7

orbeckst:
@tanmaymunjal thank you for the contribution. How many OMP threads were used in obtaining the benchmarks?

Additionally, your example of `ref = [1,2,3]` and `config = [10, 9, 8]` is not representative of what the function normally does. Can you rerun with a few thousand coordinates, e.g.,
```python
import MDAnalysis as mda
from MDAnalysisTests import datafiles as data
u = mda.Universe(data.PSF, data.DCD)
protein = u.select_atoms("protein and not name H*")
u.trajectory[0]
ref = p.positions
u.trajectory[-1]
config = p.positions
```

Can you run it with [timeit](https://docs.python.org/3/library/timeit.html)?

tanmaymunjal:
Hi @orbeckst! I think I used 5 OMP threads at the time I tested.

But I just checked the newer version on your ref and config variable schema and there seems to be very little improvement with these modifications if we use those ref and config variables. I am sorry if I wasted anybody's time.

Will convert the pull request to draft and check if I can make any other changes that might be helpful

hmacdope:
Never a waste of time @tanmaymunjal! Thanks for investigating.

It can often be difficult to rationalise what might be the faster loop arrangement due to the compiler being free to re-order under some circumstances. 

It would be super helpful to have the timings you did just for completeness?

Cheers

tanmaymunjal:
Yeah, sure @hmacdope

The timing for the newer code for 100 function calls came to 0.629s while the timing for the earlier code came to 0.651s for 100 function calls.

I think we should also mark the two issues away from the list in the issue #2046 to indicate them as investigated

PR #3826, opened 2022-09-12
adding element attribute to txyz parser

aya9aladdin:
Fixes #

While working on my GSoC project #3753 I encountered an issue with the `TXYZParser`. As I'm removing guessing processes from happening inside parsers and transferring it to happen through the `universe`'s new guesser API (`guess_topologyAttributes`), I found that TXYZParser has a special behavior of guessing `masses` from `names` instead of guessing it from `elements` or `AtomTypes` as per rest of the parsers. 

In the new guessing methodology,  guessing happens through various guesser classesn, one of which is the `DfaultGuesser`, which is a general-purpose guesser that mimic the current guesser behavior of the library. In the `DefaultGuesser` masses are guessed either from `elements` or `atom types`. So, in order to not break the current behavior of the `TXYZParser`, I added Elements attribute to its output to be used in `mass` guessing.
 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


jbarnoud:
There is some more context in that comment: https://github.com/MDAnalysis/mdanalysis/pull/3753#discussion_r959349765

package/MDAnalysis/topology/TXYZParser.py
```diff
@@ -117,9 +120,16 @@ def parse(self, **kwargs):
...
+        for n in names:
+             if n.capitalize() in SYMB2Z:
+                 validated_elements.append(n.capitalize())
+             else:
+                validated_elements.append('')
+
         attrs = [Atomnames(names),
                  Atomids(atomids),
                  Atomtypes(types),
+                 Elements(np.array(validated_elements, dtype=object)),
```
jbarnoud: I **think** it would be better to add the `elements` attribute only if all the names are valid element symbols. Otherwise, we take the risk of mistaking names for element symbols.

testsuite/MDAnalysisTests/topology/test_txyz.py
```diff
@@ -30,7 +30,7 @@
 class TestTXYZParser(ParserBase):
     parser = mda.topology.TXYZParser.TXYZParser
     guessed_attrs = ['masses']
-    expected_attrs = ['ids', 'names', 'bonds', 'types']
+    expected_attrs = ['ids', 'names', 'bonds', 'types', 'elements']
```
jbarnoud, line 34: If you follow the suggestion I made above, you can use the PDB parser as an example for the tests:

https://github.com/MDAnalysis/mdanalysis/blob/e7ee5a4a077c2179d6037ba634240ce7ebcb76ef/testsuite/MDAnalysisTests/topology/test_pdb.py#L266

https://github.com/MDAnalysis/mdanalysis/blob/e7ee5a4a077c2179d6037ba634240ce7ebcb76ef/testsuite/MDAnalysisTests/topology/test_pdb.py#L278

https://github.com/MDAnalysis/mdanalysis/blob/e7ee5a4a077c2179d6037ba634240ce7ebcb76ef/testsuite/MDAnalysisTests/topology/test_pdb.py#L308

jbarnoud (commented):
Thank you for the PR!

I made a couple of comments. One about the behaviour, and another one about the tests.

package/MDAnalysis/topology/TXYZParser.py
IAlibay: I can't remember where we got on this conversation, but I actually thought we had somehow agreed that `''` was a valid "invalid element was found" fallback case?

I know there's a few places where we allow for incomplete element arrays, but raise a user warning about it?

package/MDAnalysis/topology/TXYZParser.py
```diff
@@ -117,9 +120,16 @@ def parse(self, **kwargs):
         # Guessing time
         masses = guessers.guess_masses(names)
 
+        for n in names:
+             if n.capitalize() in SYMB2Z:
+                 validated_elements.append(n.capitalize())
+             else:
+                validated_elements.append('')
```
IAlibay: Do we want to raise a warning about this case? I _think_ we already do this for a few other places.

package/MDAnalysis/topology/TXYZParser.py
jbarnoud: An empty string is indeed what we agreed upon as a placeholder value. My logic, here, is that the column we use to fill the element attribute is defined as being either the atom names or the element symbols. If we have invalid element symbols it is likely that the column is used as atom names; then, we cannot assume that the names are actual element symbols so we should consider we do not know about the elements.

The relevant part of the Tinker documentation says:
> The .xyz file is the basic Tinker Cartesian coordinates file type. It contains a title line followed by one line for each atom in the structure. Each line contains: the sequential number within the structure, **an atomic symbol or name**, X-, Y-, and Z-coordinates, the force field atom type number of the atom, and a list of the atoms connected to the current atom.

So, at least, it is how **I** understand the problem.

IAlibay: Ah ok yeah that makes sense. I guess it's like a pseudo guess to work out which of the two is happening.

jbarnoud:
Beyond the comments, you will have to address the todo list from the PR template.

* Tests. I dressed them in this review comment: https://github.com/MDAnalysis/mdanalysis/pull/3826#discussion_r969197179
* Docs. You need to document your change in two ways. The first is to explain the new behaviour in the docstring of the `TXYZParser`. The second is to announce in the change in the docstring using the `.. versionchanged::` keyword. See https://github.com/MDAnalysis/mdanalysis/blob/e7ee5a4a077c2179d6037ba634240ce7ebcb76ef/package/MDAnalysis/topology/PDBParser.py#L181-L196 for instance.
* Changelog. You need to add your GitHub handle here: https://github.com/MDAnalysis/mdanalysis/blob/e7ee5a4a077c2179d6037ba634240ce7ebcb76ef/package/CHANGELOG#L36-L37 and you need to describe what you changed under that section: https://github.com/MDAnalysis/mdanalysis/blob/e7ee5a4a077c2179d6037ba634240ce7ebcb76ef/package/CHANGELOG#L60 the point is for users who update to figure out if they are affected.
* There is no issue to reference so you can skip that point.

You also need to fix the points raised in the automatic PEP8 comment: https://github.com/MDAnalysis/mdanalysis/pull/3826#issuecomment-1244519406

package/MDAnalysis/topology/TXYZParser.py
aya9aladdin: I got it but as I'm taking the guessing point of view, if the name column have invalid elements then guessing behavior will take two different patterns before and after the new guesser API as follows:
old behavior: guessing masses will be done from atom names which can lead to trashy results here and there
new behavior: guessing masses will be done from atom type, which will result in completely invalid output cause atom types are represented by numeric values
I think sticking to not guess any trashy masses is a good update, but I'm afraid it will break any current default behavior

aya9aladdin:
> Hello @aya9aladdin! Thanks for opening this PR. We checked the lines you've touched for [PEP 8](https://www.python.org/dev/peps/pep-0008) issues, and found:
> 
> * In the file [`package/MDAnalysis/topology/TXYZParser.py`](https://github.com/MDAnalysis/mdanalysis/blob/45da09036a4fa654da700c0a820bffaf0fada718/package/MDAnalysis/topology/TXYZParser.py):
> 
> > [Line 124:14](https://github.com/MDAnalysis/mdanalysis/blob/45da09036a4fa654da700c0a820bffaf0fada718/package/MDAnalysis/topology/TXYZParser.py#L124): [E111](https://duckduckgo.com/?q=pep8%20E111) indentation is not a multiple of four
> > [Line 124:14](https://github.com/MDAnalysis/mdanalysis/blob/45da09036a4fa654da700c0a820bffaf0fada718/package/MDAnalysis/topology/TXYZParser.py#L124): [E117](https://duckduckgo.com/?q=pep8%20E117) over-indented
> > [Line 125:18](https://github.com/MDAnalysis/mdanalysis/blob/45da09036a4fa654da700c0a820bffaf0fada718/package/MDAnalysis/topology/TXYZParser.py#L125): [E111](https://duckduckgo.com/?q=pep8%20E111) indentation is not a multiple of four
> > [Line 126:14](https://github.com/MDAnalysis/mdanalysis/blob/45da09036a4fa654da700c0a820bffaf0fada718/package/MDAnalysis/topology/TXYZParser.py#L126): [E111](https://duckduckgo.com/?q=pep8%20E111) indentation is not a multiple of four

I'm confused about those pep8 errors, the lines that are mentioned seems to be well indented, can't figure out where is the issue

package/CHANGELOG
```diff
@@ -13,6 +13,14 @@ The rules for this file:
...
 
 ------------------------------------------------------------------------------
+
+09/16/22 aya9aladdin
+
+* 2.4.0
+
+Changes
+  *adding element attribute to TXYZParser if all atom names are valid element symbols
+
```
jbarnoud: That is the idea. However, the only things you need to do is add your handle alongside the ones already there for version 2.4.0 and add the description line with the changes for that version. There is no need to create a new block.

package/MDAnalysis/topology/TXYZParser.py
```diff
@@ -69,8 +72,11 @@ class TXYZParser(TopologyReaderBase):
 
     - Atomnames
     - Atomtypes
+    - Elements
 
     .. versionadded:: 0.17.0
+    .. versionchanged:: 2.4
+       Adding Element attribute if all names is a valid element symbol
```
jbarnoud:
```suggestion
       Adding the `Element` attribute if all names are valid element symbols.
```

package/MDAnalysis/topology/TXYZParser.py
```diff
@@ -69,8 +72,11 @@ class TXYZParser(TopologyReaderBase):
 
     - Atomnames
     - Atomtypes
+    - Elements
 
     .. versionadded:: 0.17.0
+    .. versionchanged:: 2.4
```
jbarnoud:
```suggestion
    .. versionchanged:: 2.4.0
```

testsuite/MDAnalysisTests/topology/test_txyz.py
```diff
@@ -51,3 +52,25 @@ def test_atom_type_type(self, top):
...
         assert all(type_is_str)
+
+def test_TXYZ_elements():
+    """The test checks whether elements attribute are assigned
+    properly given a TXYZ file with valid elements record.
+    """
+    u = mda.Universe(TXYZ, format='TXYZ')
+    element_list = np.array(['C', 'H', 'H', 'O', 'H', 'C', 'H', 'H', 'H'], dtype=object)
+    assert_equal(u.atoms.elements, element_list)
+
```
jbarnoud, line 63:
```suggestion
    
    
```

jbarnoud (changes requested):
Almost there. Look at how the CHANGELOG is constructed, though. There is one block per version, and there is especially one block for the version we are currently working on (i.e. 2.4.0). This is the block you need to update rather than create a new one.

package/CHANGELOG
jbarnoud: You also need to reference the rull request. Look at other entries to see how to do it.

aya9aladdin:
all done except for the pep8 error messages that seems to be wiered

package/CHANGELOG
```diff
@@ -13,7 +13,7 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" http://semver.org
 
 ------------------------------------------------------------------------------
-??/??/?? IAlibay, Luthaf, hmacdope, rafaelpap, jbarnoud
+09/16/22 IAlibay, Luthaf, hmacdope, rafaelpap, jbarnoud, aya9aladdin
```
```suggestion
??/??/?? IAlibay, Luthaf, hmacdope, rafaelpap, jbarnoud, aya9aladdin
```

Version 2.4.0 is not yet released and we do not know when it will be. Hence the question marks instead of a date.

package/MDAnalysis/topology/TXYZParser.py
```diff
@@ -69,8 +72,11 @@ class TXYZParser(TopologyReaderBase):
 
     - Atomnames
     - Atomtypes
+    - Elements
```
jbarnoud:
```suggestion
    - Elements (if all atom names are element symbols)
```

Otherwise users will be surprised when they do not have an `elements` attribute.

jbarnoud (approved):
Great! :tada:

jbarnoud:
One more PR merged :grin: Congrats!

PR #3829, opened 2022-09-14
Rename wheel_cache_size to build_cache_size in asv.conf.json

IAlibay:
Simple enough one

See: https://github.com/airspeed-velocity/asv/blob/b54d3024f5b77ddf81f82e5929e9b97ed85f7067/asv/config.py#L66-L70

Happened ~ 4 years ago from the looks of things, so we should be good? 

PR Checklist
------------
 - Tests?
 - Docs?
 - CHANGELOG updated?
 - Issue raised/referenced?


hmacdope (approved):
LGTM it was annoying me that warning so thanks for fixing.

PR #3834, opened 2022-09-18
Ensure Timestep in ChainReader is updated for multiple trajectory files

### LINKED ISSUE #3657: Applying Transformation to Chain of Traj doesn't Apply to First Frame of 2nd, 3rd... trajectories
## Expected behavior ##

After creating a universe with multiple trajectory files, and a series of transformations are applied following the protein example. Those transformations are expected to be applied to all frames.

## Actual behavior ##

The first frame of the 2nd, 3rd, etc. trajectories are unwrapped. This is evidenced by writing out the coordinates to the screen, or writing to a pdb file.

It seems to be a discrepancy with the atoms.positions and ts._pos being synced, as the later is used by the Writer.

Run short LJ lammps simulation.
```
units           lj
atom_style      full

region          box block 0 10 0 10 0 10
create_box      2 box bond/types 1 extra/bond/per/atom 1
create_atoms    1 single 5 5 5.5
create_atoms    1 single 5 5 4.5
create_atoms    2 random 200 1234 box

pair_style      lj/cut 4.5
pair_coeff      * * 1.0 1.0
mass            * 1.0

bond_style      harmonic
bond_coeff      1 1000.0 1.0
create_bonds    single/bond 1 1 2

##########
neighbor 2.0 bin
neigh_modify delay 0 every 1 check yes

min_style cg # conjugate gradient (CG) algorithm
minimize 1.0e-4 1.0e-6 1000 10000

###########

variable TK equal 2
variable Pbar equal 0.01
variable seed equal 1234
variable freq equal 1000

velocity all create ${TK} ${seed} rot yes dist gaussian

fix             1 all nvt temp ${TK} ${TK} 0.05

timestep        0.01
reset_timestep 0
thermo          ${freq}

write_data initial.data
dump 1 all custom ${freq} dump_1.lammpstrj id mol type q xu yu zu
run             5000
undump 1

dump 1 all custom ${freq} dump_2.lammpstrj id mol type q xu yu zu
run             5000

```

Import LAMMPS trajectories, center around bonded molecule, and write to pdb.
``` python
import MDAnalysis as mda

runs = [1,2]
select = "type 1"

data_file = "run0/initial.data"
dumpfile = "run0/dump_{}.lammpstrj"
dump_file = [dumpfile.format(run) for run in runs]

universe = mda.Universe(data_file, dump_file, format="LAMMPSDUMP")
group_target = universe.select_atoms(select)
group_remaining = universe.select_atoms("all") - group_target
box_center = universe.trajectory[0].dimensions[:3] / 2
com = group_target.center_of_mass(unwrap=True)
transforms = [
    trans.unwrap(group_target),
    trans.translate(box_center-com),
    trans.wrap(group_remaining)
]
universe.trajectory.add_transformations(*transforms)

group = universe.select_atoms("all")
with mda.Writer("output_unwrapped.pdb", group.n_atoms) as W:
    for ii,ts in enumerate(universe.trajectory):
        W.write(group)

```

- MDA Version: '2.1.0-dev0'
- Python Version: 3.8.12
- MacOS

jaclark5:
Fixes #3657

 - If statement in transform added to ensure that the provided Timestep class is the same used in the Atom Group trajectory. This is important when multiple trajectory files are used, otherwise the first step of the next trajectory will not undergo a transform.

PR Checklist
------------
 - [X] Tests?
 - [NA] Docs?
 - [X] CHANGELOG updated?


testsuite/MDAnalysisTests/transformations/test_translate.py
```diff
@@ -49,6 +49,19 @@ def test_translate_coords(translate_universes):
     trans = translate(vector)(trans_u.trajectory.ts)
     assert_array_almost_equal(trans.positions, ref.positions, decimal=6)
 
+def test_wrap_id(translate_universes):
+    # If the translation is given a different timestep from what is has, the timestep will
```
tylerjereddy:
```suggestion
    # If the translation is given a different timestep from what it has, the timestep will 
```

testsuite/MDAnalysisTests/transformations/test_translate.py
```diff
@@ -49,6 +49,19 @@ def test_translate_coords(translate_universes):
...
+    # be updated to the new one given. This should only matter in a ChainReader
+    ref_u, trans_u = translate_universes
+    trans_u.dimensions = [363., 364., 365., 90., 90., 90.]
+    ref_u.dimensions = [363., 364., 365., 90., 90., 90.]
+
+    ag = trans_u.residues[24].atoms
+    trans = wrap(ag)
+    id_trans = id(trans_u.trajectory.ts)
+    trans = trans._transform(ref_u.trajectory.ts)
+    assert id_trans != id(trans_u.trajectory.ts) and id(trans_u.trajectory.ts) == id(ref_u.trajectory.ts)
```
tylerjereddy, line 63: Just a suggestion, but the first thing I wondered reading over the issue and this PR was why `wrap` might be different from the opposite operation, `unwrap`, so I wrote a short test script that seems to pass on `develop` and fail here (see below). Perhaps a simpler version of that test and a similar change to `unwrap` might be in order? You might also discuss dealing with `unwrap` in a separate PR with the team though, they might be "ok" with that, not sure.

```python
import MDAnalysis
import MDAnalysis.transformations

from numpy.testing import assert_allclose


def main():
    u1 = MDAnalysis.Universe(PSF, DCD, DCD)
    ag1 = u1.atoms
    u2 = MDAnalysis.Universe(PSF, DCD, DCD)

    dim = [2, 2, 2, 90, 90, 90]
    transform_0 = MDAnalysis.transformations.boxdimensions.set_dimensions(dim)
    transform_1 = MDAnalysis.transformations.wrap(ag1)
    transform_2 = MDAnalysis.transformations.unwrap(ag1)
    u1.trajectory.add_transformations(transform_0, transform_1, transform_2)

    for ts1, ts2 in zip(u1.trajectory, u2.trajectory):
        try:
            assert_allclose(ts1.positions, ts2.positions)
        except:
            print("ts1.frame differs:", ts1.frame)



if __name__ == "__main__":
    main()

```

jaclark5: That's a good point, it seems like it would be true for any transformation call to _transform that involves an atom group attribute. This would also include fit_rot_trans, while positionaveraging already checks and corrects the timestep.
If I check the other cases for fit_rot_trans and unwrap in the original issue, wouldn't it make more sense to change the issue title to be more broad? Especially since these are so closely related.

jaclark5:
@tylerjereddy I cleared it on the original issue, but you'll see the edits to applicable transforms with tests have been merged.

jaclark5:
The CodeCoverage stipulation is failing because of the most recent merge into the development branch related to RDKit... is there a work around for this? The code relevant to this pull request is fully covered.

tylerjereddy:
I think you're good now as far as codecov checks are concerned. I think they can take a while to properly aggregate.

richardjgowers:
I had a dig into this.  I think this fix is probably fine, but I think there's a more fundamental issue with ChainReader not respecting the Reader base class, which makes lots of annoying bugs like this appear.  I think a better fix (which I'll try and put together) is to put ChainReader under the boot of ProtoReader, rather than have it be "equal".

jaclark5:
Ok sure, let me know when you decide to go with this fix or if I should delete it.

To find this solution I followed the order of operations in updating the timestep and this cuts through the chicken and the egg cycle. I didn't see another solution, but I'm a novice with MDAnalysis' structure.

jaclark5:
@richardjgowers any thoughts on this? I think this is a glaring issue in MDAnalysis, although it seems that an analysis that relies on capped_distances won't be affected as long as a box is defined to correct it. However, if one were to attempt to make a movie like the "Centering a trajectory in the box" example for a case with multiple trajectories, they would not achieve a clean movie. The solution presented here resolves the issue simply for the time being.

PR #3839, opened 2022-09-20
Empty atomgroup

### LINKED ISSUE #3837: Obscure Eigenvalue error when calling shape_parameter for empty ag
## Expected behavior ##

When calculating the shape_parameter of an AtomGroup, provide error if the group is empty. This is best placed in `center_of_mass` and `center_of_charge` since all other functions reference these two methods first.

## Actual behavior ##

Obscure error that can eventually be determined to result from an empty AtomGroup:
`numpy.linalg.LinAlgError: Eigenvalues did not converge`

``` python
import MDAnalysis as mda
universe = mda.Universe(PDB)
ag = universe.select_atoms("type 1")
print(len(ag), ag.shape_parameter())
```

- Which version are you using? 2.4.0-dev0
- Which version of Python? 3.9.13
- Which operating system? MacOS 11.6

jaclark5:
Fixes #3837 

 - Raise error if atom group provided to `center_of_mass` or `center_of_charge` is empty. This prevents obscure eigenvalue errors when using other AtomGroup methods.

PR Checklist
------------
 - [X] Tests?
 - [NA] Docs?
 - [X] CHANGELOG updated?


package/MDAnalysis/core/topologyattrs.py
```diff
@@ -1519,6 +1519,8 @@ def center_of_mass(group, wrap=False, unwrap=False, compound='group'):
            is deprecated and will be removed in version 3.0.
         """
         atoms = group.atoms
+        if not len(atoms):
```
orbeckst: This can be simplified to
```python
if not atoms:
```
because the truthiness of an empty AtomGroup is `False`.

orbeckst (commented):
I had a quick look and I am only commenting because the bigger question of how this behavior fits into the API has not been addressed. This will require a discussion because returning (EDIT: ~0~)  empty array `[]` for an empty AtomGroup can be considered a sensible solution. There may be code out there that depends on this behavior so this change would then lead to crashing code.

In terms of technical comments, see the inline comments and that we would also need a test that checks that your code is doing what it is supposed to do. I.e., is the `ValueError` raised for an empty AtomGroup. However, don't worry about implementing the test yet. The API discussion comes first.

EDIT: I changed my mind and think it should raise an exception. See https://github.com/MDAnalysis/mdanalysis/issues/3837#issuecomment-1255535272

package/MDAnalysis/core/topologyattrs.py
jaclark5: Fair enough, the weird error come into play with the shape_parameter and asphericity methods. The center_of_mass doesn't have an issue, I just thought it was a convenient commonality. 
This could also be resolved with an if statement inside shape_parameter and asphericity with: 
```
if not group.atoms:
    return np.array([])
```

orbeckst:
I started a discussion in https://github.com/MDAnalysis/mdanalysis/issues/3837#issuecomment-1253033215 to see how to continue.

hmacdope:
> I started a discussion in [#3837 (comment)](https://github.com/MDAnalysis/mdanalysis/issues/3837#issuecomment-1253033215) to see how to continue.

I replied on the issue, but just to double up my opinion is that this should raise an exception unless it breaks heaps of tests.

package/MDAnalysis/core/topologyattrs.py
```diff
@@ -1519,6 +1519,8 @@ def center_of_mass(group, wrap=False, unwrap=False, compound='group'):
...
+            raise ValueError("There are no atoms provided group")
```
hmacdope: Assuming we are going with an error (@orbeckst?, no tests seem to complain) can you add a test for this error @jaclark5?  

Something like 

```python
with pytest.raises(ValueError, match ="there are no ..."):
    ...

```

hmacdope (changes requested):

package/MDAnalysis/core/topologyattrs.py
IAlibay: One approach to make this contribution much more powerful would be to add a simple decorator which raises an error if the atomgroup is empty.

hmacdope: This is a great idea something like `@check_atomgroup_not_empty`!

orbeckst (dismissed):
Based on the discussion in #3837 we want to **raise ValueError** when AtomGroup is trying to compute something on an empty AtomGroup. The methods include
* center_of_mass
* centroid
* total_mass
* total_charge
* ... possibly others that need to be identified

The preferred solution is a decorator that is applied to a method and raises `ValueError` if the AG is empty.

We will need tests for any methods that are to be decorated.

package/MDAnalysis/lib/util.py
```diff
@@ -2165,6 +2166,33 @@ def wrapper(*args, **kwargs):
         return wrapper
     return check_coords_decorator
```
IAlibay, line 2168:
```suggestion


```
Just a pep8 thing about double lines between methods.

testsuite/MDAnalysisTests/core/test_groups.py
```diff
@@ -213,6 +214,47 @@ def test_group_return_unsorted_sorted_unique(self, ugroup):
         assert unsorted_unique is sorted_unique
         assert unique._cache['unsorted_unique'] is sorted_unique
 
+class TestEmptyAtomGroup(object):
```
IAlibay, line 218:
```suggestion

class TestEmptyAtomGroup(object):
```
As above, slight PEP8 thing

package/MDAnalysis/lib/util.py
```diff
@@ -2165,6 +2166,33 @@ def wrapper(*args, **kwargs):
...
 
+def check_atomgroup_not_empty(groupmethod):
+    """Decorator triggering a ``ValueError`` if the underlying group is empty.
+
+    Avoids obscure computational errors on group methods.
```
hmacdope: This is a bit vague 

```suggestion
    Avoids downstream errors in computing properties of empty atomgroups. 
```

package/MDAnalysis/lib/util.py
```diff
@@ -2165,6 +2166,33 @@ def wrapper(*args, **kwargs):
...
+
+    Raises
+    ------
+    ValueError
+        If the :class:`~MDAnalysis.core.groups.AtomGroup`,
```
hmacdope: Does this work on all groups? If so we should change the name and test. Otherwise we need to change docstring.

hmacdope (changes requested):
Some quick comments.

jaclark5:
Is there a way for me to restart one of the tests? In looking at the log file, it "failed" because it took over an hour to run, but that's because it spent half an hour "receiving files" which doesn't pertain to the PR.

orbeckst:
I have very little available time over the next week — so I am likely not able to get to the PR review, sorry.

One of the other reviewers can dismiss my review if it's in the way.

hmacdope (approved):
@jaclark5 this looks good to me now. I think this decorator will be very useful.

hmacdope:
@jaclark5 sorry I was so slow to circle back and review.  One last thing could you please add a changelog entry?

IAlibay (commented):
just passing by to check I didn't have a blocking review, looks like there's two PEP8 issues to fix. I'm not going to block over them but please do fix.

jaclark5:
@hmacdope If everything is in order, @orbeckst said that his review could be dismissed by another reviewer.

> I have very little available time over the next week — so I am likely not able to get to the PR review, sorry.
> 
> One of the other reviewers can dismiss my review if it's in the way.

package/AUTHORS
```diff
@@ -82,7 +82,6 @@ Chronological list of authors
   - Pedro Reis
  2017
   - Ruggero Cortini
-  - Xiki Tempula
```
hmacdope: Please add this author back in.

package/CHANGELOG
```diff
@@ -13,23 +13,26 @@ The rules for this file:
...
 ------------------------------------------------------------------------------
-
-??/??/?? IAlibay, Luthaf, hmacdope, rafaelpap, jbarnoud, BFedder, aya9aladdin
+??/??/?? IAlibay, Luthaf, hmacdope, rafaelpap, jbarnoud, BFedder, aya9aladdin,
+         jaclark5
 
  * 2.4.0
 
 Fixes
+  * Added isolayer selection method (Issue #3845)
```
hmacdope: I'm not sure why this is in the diff?

package/CHANGELOG
```diff
@@ -53,6 +56,7 @@ Changes
   * The "AMBER" entry in setup.py's extra_requires has now been removed. Please
     use `pip install ./package[extra_formats]` instead (PR #3810)
   * `Universe.empty` emmits less warnings (PR #3814)
+  * adding element attribute to TXYZParser if all atom names are valid element symbols (PR #3826)
```
hmacdope: Same here, idk why this is in the diff.

hmacdope (changes requested):
Make sure to add the `Xtiki-tempula` back into authors.

The diff from #3846 and #3826 have somehow crept in. Im no it wizard but it should be possible to fix.

hmacdope (approved):
Thanks for fixing up @jaclark5 :)

hmacdope:
@IAlibay the Azure 32 bit build seems to not be working have we seen in other PRs?

IAlibay:
> @IAlibay the Azure 32 bit build seems to not be working have we seen in other PRs?

ha, scipy seems to have finally killed 32bit, I'll open a PR and check with Tyler.

For now you can probably ignore this tbh.

hmacdope:
@IAlibay cheers for that. :)

PR #3848, opened 2022-09-21
Hbond update

### LINKED ISSUE #3847: HBond Analysis, efficiency, usability, LAMMPS friendliness
## Expected behavior ##

- The language in the documentation specifies the cutoffs in hbond analysis to be the min or max *allowed* which implies inclusive logical statements
- LAMMPS files only have types (at the moment, it is possible that molecule numbers and types could be saved as resnames and names) making this package unusable.
- A result of nan from the function is difficult to decipher
- guess_donor function takes a long time when hydrogen atoms are already known

## Actual behavior ##

- The cutoffs in hbond analysis code use exclusive statements.
- LAMMPS files cannot be used with this function
- One must be developer savy  enough to comb through the package to find out why a value of NaN was given.
- guess_donor function could use bonding information since hydrogen types are called anyway and would be faster, instead of existing distance cutoff.

jaclark5:
Fixes #3847

 - Added warning when distance or angle cutoffs produce zero hbonds to clarify a returned value of NaN
 - Guess_donor uses bonding information when available since it calls hydrogen types anyway
 - If resnames and names are not available, atom types are used. This is inclusive to LAMMPS users.
 - Change min max constraints to match inclusive language in the documentation.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


jaclark5:
I expect these changes will spur discussion so I didn't bother with the tests yet.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -455,19 +463,30 @@ def guess_donors(self, select='all', max_charge=-0.5):
             hydrogens_sel = self.hydrogens_sel
         hydrogens_ag = self.u.select_atoms(hydrogens_sel)
 
-        ag = hydrogens_ag.residues.atoms.select_atoms(
-            "({donors_sel}) and around {d_h_cutoff} {hydrogens_sel}".format(
-                donors_sel=select,
-                d_h_cutoff=self.d_h_cutoff,
-                hydrogens_sel=hydrogens_sel
+        if hasattr(hydrogens_ag[0],"bonded_atoms") and hydrogens_ag[0].bonded_atoms:
+            donors_ag = self.u.atoms[[x.bonded_atoms[0].ix for x in hydrogens_ag]]
```
richardjgowers: Might be worth reusing this function here: https://github.com/MDAnalysis/mdanalysis/blob/develop/package/MDAnalysis/analysis/hydrogenbonds/hbond_autocorrel.py#L223

richardjgowers (commented):
I think these changes don't look that controversial.  Changing the output schema/columns might break something in the output, but I guess that will shake out when tests are added.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
jaclark5: Thanks for the heads up!

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -401,17 +401,25 @@ def guess_hydrogens(self,
         ag = self.u.select_atoms(select)
         hydrogens_ag = ag[
             np.logical_and.reduce((
-                ag.masses < max_mass,
-                ag.charges > min_charge,
-                ag.masses > min_mass,
+                ag.masses <= max_mass,
```
hmacdope: Technically an API break but I think the way it was before was confusing? I would support the change proposed here.

jaclark5:
It won't break the output. I've been using it with success for quite some time. I actually forgot I made some of these changes but I want to merge them and then stay current with the package.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -455,19 +464,31 @@ def guess_donors(self, select='all', max_charge=-0.5):
...
         hydrogens_ag = self.u.select_atoms(hydrogens_sel)
 
-        ag = hydrogens_ag.residues.atoms.select_atoms(
-            "({donors_sel}) and around {d_h_cutoff} {hydrogens_sel}".format(
-                donors_sel=select,
-                d_h_cutoff=self.d_h_cutoff,
-                hydrogens_sel=hydrogens_sel
+        if hasattr(hydrogens_ag[0],"bonded_atoms") and hydrogens_ag[0].bonded_atoms:
+            donors_ag = find_hydrogen_donors(hydrogens_ag)
+#            donors_ag = self.u.atoms[[x.bonded_atoms[0].ix for x in hydrogens_ag]]
```
hmacdope: Don't forget to remove this comment.

package/doc/sphinx/source/documentation_pages/references.rst
```diff
@@ -180,6 +180,10 @@ If you use the Mean Squared Displacement analysis code in
 
 .. _`10.21105/joss.00877`: https://doi.org/10.21105/joss.00877
 
+For additional information on non-gaussian parameter, consider [Fuchs1998]_.
```
hmacdope: What is this in relation to exactly? Is there a specific thing in H-bonds that you are referring to?

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -401,17 +402,25 @@ def guess_hydrogens(self,
...
-            ]
-        )
+        if hasattr(hydrogens_ag,"resnames") and hasattr(hydrogens_ag,"names"):
+            hydrogens_list = np.unique(
+                [
+                    '(resname {} and name {})'.format(r, p) for r, p in zip(hydrogens_ag.resnames, hydrogens_ag.names)
+                ]
+            )
+        else:
+            hydrogens_list = np.unique(
```
hmacdope: Add a note to the docstring of the class that the types can be used as well.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -401,17 +402,25 @@ def guess_hydrogens(self,
...
+                [
+                    'type {}'.format(tp) for tp in hydrogens_ag.types
```
hmacdope: You will need to update `versionchanged` for the class also.

hmacdope (changes requested):
looking really good! few comments.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
IAlibay: Confusing doesn't mean it's a bug, if the behaviour changes such that the numbers will change this is a pretty big break on a science perspective. Knowing the types of places this code gets used, I'm going to take the stance that this needs at least one version of user warning?

hmacdope: Tagging @MDAnalysis/coredevs for discussion.

orbeckst: The docs state (my emphasis)

> This function makes use of atomic masses and atomic charges to identify which atoms are hydrogen atoms that are capable of participating in hydrogen bonding. If an atom has a mass **less than** :attr:`max_mass` and an  atomic charge **greater than** :attr:`min_charge` then it is considered capable of participating in hydrogen bonds.

so the current code does not contradict the docs, does it? Therefore, what's the justification of changing the code, especially as @IAlibay notes, people may rely on the defaults.

Maybe just making the docs clearer (the description of the kwargs) would be sufficient? Given that these are floating point comparisons, the equality part really should not matter too much.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
jaclark5: Oh thanks!

package/doc/sphinx/source/documentation_pages/references.rst
jaclark5: Nope I think I was on the wrong branch, this is foreshadowing.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
jaclark5: I now see that distinction in the notes of the methods `guess_hydrogens`, but the definition under Parameters uses inclusive language:
>         max_mass: float (optional)
>             Maximum allowed mass of a hydrogen atom.
>         min_charge: float (optional)
>             Minimum allowed charge of a hydrogen atom.

If you would prefer, I can update the language under Parameters to be consistent with the Notes section for this PR.
Affected parameters: 
- hydrogen `max_mass`, `min_charge`, `min_mass`, 
- donor `max_charge`,
- acceptor `max_charge`

There are a few points I would make first (With vote to change inequalities or not):
- (Nay) In other cases in MD cutoffs are exclusive. (pair cutoffs and coordination cutoffs)
- (Yay) The distance based selection criteria available through MDAnalysis use inclusive maximum cutoffs, but with exclusive minimum, cutoffs. 
    - To stay consistent, this would mean changing only the maximum constraints

orbeckst: For 2.x just changing the docs should do it.

For 3.x we can consider changing the treatment of the boundary if people are interested in it. If so, we would
1. raise a new issue for this specific problem
2. add deprecation notes and warnings to the code
3. make the change when we release 3.0

.gitignore
```diff
@@ -21,7 +21,7 @@ examples/output.txt
 # ignore Vagrant virtual machines
 .vagrant
 # ignore coverage files
-.coverage
+.coverage*
```
IAlibay, line 24: woudln't this technically drop `.coveragerc`? (which we don't want to ignore)

package/CHANGELOG
```diff
@@ -19,6 +19,7 @@ The rules for this file:
  * 2.4.0
 
 Fixes
+  * Update hbond analysis doc string to use exclusive bound language (#3847)
```
IAlibay: If this must go in as a changelog entry, I would prefer it under enhancement than fix. This is purely on the basis that fixes is where I look for "code things that were broken".

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -407,24 +429,35 @@ def guess_hydrogens(self,
...
```
IAlibay: I'm only quickly going through this so I might be missing something obvious - doesn't this assume that all types are the same? Wouldn't this be a problem if you end up with a type where some of the atoms are outside the bounds of the allowable charges but not others?

My recommendation here would be to default to a unique attribute, like the `ix` value, instead of using something that's broad and could pick up stuff outside the intended atomgroup selection.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -455,31 +488,48 @@ def guess_donors(self, select='all', max_charge=-0.5):
...
+        if hasattr(donors_ag,"resnames") and hasattr(donors_ag,"names"):
+            donors_list = np.unique(
+                [
+                    '(resname {} and name {})'.format(r, p) for r, p in zip(donors_ag.resnames, donors_ag.names)
+                ]
+            )
+        else:
+            donors_list = np.unique(
+                [
+                    'type {}'.format(tp) for tp in donors_ag.types if tp not in hydrogens_ag.types
```
IAlibay: Same as above.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -503,11 +553,18 @@ def guess_acceptors(self, select='all', max_charge=-0.5):
 
         ag = self.u.select_atoms(select)
         acceptors_ag = ag[ag.charges < max_charge]
-        acceptors_list = np.unique(
-            [
-                '(resname {} and name {})'.format(r, p) for r, p in zip(acceptors_ag.resnames, acceptors_ag.names)
-            ]
-        )
+        if hasattr(acceptors_ag,"resnames") and hasattr(acceptors_ag,"names"):
```
IAlibay: Having seen this construct three times now, it seems like this could be a staticmethod that gets called instead?

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -540,7 +597,12 @@ def _get_dh_pairs(self):
         # Otherwise, use d_h_cutoff as a cutoff distance
         else:
 
-            hydrogens = self.u.select_atoms(self.hydrogens_sel)
+            if self.hydrogens_sel is None:
```
IAlibay: Could you give more details on why this change is necessary? It feels like this changes the default behaviour, where I would expect that if I pass something as None I should expect to get None?

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -622,6 +684,14 @@ def _single_frame(self):
             return_distances=True,
         )
 
+        if np.size(d_a_indices) == 0:
+            warnings.warn(
+                "No hydrogen bonds were found given d-a cutoff of {} between "\
```
IAlibay: please use implicit continuation and f-strings here.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -644,6 +714,14 @@ def _single_frame(self):
         )
         hbond_indices = np.where(d_h_a_angles > self.d_h_a_angle)[0]
 
+        if np.size(hbond_indices) == 0:
+            warnings.warn(
+                "No hydrogen bonds were found given angle of {} between "\
```
IAlibay: as above

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -800,8 +878,17 @@ def count_by_type(self):
         d = self.u.atoms[self.hbonds[:, 1].astype(np.intp)]
         a = self.u.atoms[self.hbonds[:, 3].astype(np.intp)]
 
-        tmp_hbonds = np.array([d.resnames, d.types, a.resnames, a.types],
-                              dtype=str).T
+        if hasattr(d,"resnames"):
```
IAlibay: I would probably say that this construct can be squashed, since they both stem from the same universe, if `d` has no resnames then `a` doesn't either. You can just do the one check?

testsuite/MDAnalysisTests/analysis/test_hydrogenbonds_analysis.py
```diff
@@ -204,6 +204,198 @@ def test_no_bond_info_exception(self, universe):
...
+        # using step 2
+        hbonds.run(step=2)
+
+        caplog.set_level(logging.WARNING)
+        hbonds.lifetime(tau_max=2, intermittency=1)
+
+        warning = \
+            "Autocorrelation: Hydrogen bonds were computed with step > 1."
+        assert any(warning in rec.getMessage() for rec in caplog.records)
+
```
IAlibay, line 277:
```suggestion


```
pep8

testsuite/MDAnalysisTests/analysis/test_hydrogenbonds_analysis.py
```diff
@@ -204,6 +204,198 @@ def test_no_bond_info_exception(self, universe):
...
+    def test_logging_step_not_1(self, universe, caplog):
+        hbonds = HydrogenBondAnalysis(universe, **self.kwargs)
+        # using step 2
+        hbonds.run(step=2)
+
+        caplog.set_level(logging.WARNING)
+        hbonds.lifetime(tau_max=2, intermittency=1)
+
+        warning = \
```
IAlibay: implicit continuation please

testsuite/MDAnalysisTests/analysis/test_hydrogenbonds_analysis.py
```diff
@@ -204,6 +204,198 @@ def test_no_bond_info_exception(self, universe):
...
+        u = universe.copy()
+        n_residues = 2
+        u.add_TopologyAttr('mass', [15.999, 1.008, 1.008] * n_residues)
+        u.add_TopologyAttr('charge', [-1.04, 0.52, 0.52] * n_residues)
+        h = HydrogenBondAnalysis(u, **kwargs)
+        donors = u.select_atoms(h.guess_donors())
+
+        assert len(donors) == 2
+
     def test_first_hbond(self, hydrogen_bonds):
```
IAlibay: This used to be part of the `TestHydrogenBondAnalysisIdeal` is there a reason for moving it to `TestHydrogenBondAnalysisNoRes` isntead?

testsuite/MDAnalysisTests/analysis/test_hydrogenbonds_analysis.py
```diff
@@ -204,6 +204,198 @@ def test_no_bond_info_exception(self, universe):
...
+        'hydrogens_sel': 'type H H',
+        'acceptors_sel': 'type O',
+        'd_h_cutoff': 1.2,
+        'd_a_cutoff': 3.0,
+        'd_h_a_angle_cutoff': 120.0
+    }
+
+    @staticmethod
+    @pytest.fixture(scope='class')
+    def universe():
```
IAlibay, line 292: no something that's necessary for this PR, but rather than duplicating this, it may make more sense to refactor this to be two child classes of a class that defines this staticmethod and then in `TestHydrogenBondAnalyssiNoRes` you just `del_TopologyAttr` the resnames.

IAlibay (changes requested):

.gitignore
jaclark5: Oh sorry! When I was debugging the coverage locally there were files name .coverageblahblahblah that I accidentally included. 
Would you prefer that I remove this edit, or I can add `!.coveragerc`

IAlibay: I'd be happy with adding a `!.coveragerc` here.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
jaclark5: @IAlibay In most cases an atom type in LAMMPS is all identical with the exception of forcefields such as ReaxFF and COMB. Although you are right that LAMMPS allows for atoms to have individual charges so this *could* be an issue as forcefields develop...

Here's the issue, these structures look for a list of atom selection strings that represent **categories**, so if this was on an index basis, the list would be very long, and the resulting output would be useless (representing a single interaction). One could have this structure instead search the atoms of each type and make a long string that specifies the specific indices that meet the charge criteria. However, this string will be unnecessary computational time in 99% of cases, and in the 1% I would argue it would skew the results as these many-body forcefields will continue to fluctuate in charge throughout the simulation, as they are supposed to. I would argue counting "low" or "high" charges in that case wouldn't disrupt the intention of the calculation, it would result in a broken hydrogen bond with a more "realistic" decay time due to local atom environments.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
jaclark5: This doesn't change the default behavior since this method is called in `_prepare` and if `hydrogens_sel==None` then `self.guess_hydrogens()` is called first anyway. Because this is a hidden method, the only place this change makes a difference is in testing, which I believe is why I did it. Additionally, I think this change better aligns with the purpose of the function. The description of this method `_get_dh_pairs` is:

> Finds donor-hydrogen pairs.
>
> Returns
> donors, hydrogens: AtomGroup, AtomGroup
>&nbsp;&nbsp;&nbsp;&nbsp;AtomGroups corresponding to all donors and all hydrogens. AtomGroups are ordered such that, if zipped, will produce a list of donor-hydrogen pairs.

jaclark5: As an aside, I would advocate that all instances of `variable is None` be changed to `variable == None` since all values of `None` don't necessarily come from the same memory location... I've been burned before.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
jaclark5: Oh good point!

testsuite/MDAnalysisTests/analysis/test_hydrogenbonds_analysis.py
jaclark5: I'm confused on why you said "used to be" since `TestHydrogenBondAnalysisIdeal` hasn't been touched. I did repeat it but you'll notice in the "add_TopologyAttr" I used types instead of resnames and names. This is also the case in the method selection criteria. Although the system in the two universes is identical, the format of the universe in `TestHydrogenBondAnalysisIdeal` represents something imported from CHARMM or AMBER etc. and the one I made in `TestHydrogenBondAnalysisNoRes` represents that system after importing from LAMMPS.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
richardjgowers: I was/am pretty sure None was a singleton in Python and always "is" itself.

IAlibay: > As an aside, I would advocate that all instances of variable is None be changed to variable == None since all values of None don't necessarily come from the same memory location... I've been burned before.

So the main reason why we use `is None` over `== None` is due to a PEP8.

As far as I understand it, although `==` and `is` are fundamentally different, due to the nature of `None`, the two end up being equivalent? At least I don't know of an edge case where the comparison does not hold up. There is this specific case I believe, but I take it as a none-issue [pun intended ;) ] http://jaredgrubb.blogspot.com/2009/04/python-is-none-vs-none.html

testsuite/MDAnalysisTests/analysis/test_hydrogenbonds_analysis.py
jaclark5: This isn't my addition, but yes I can change it. Since the method tests of the class I copied are identical it seems the `diff` has decided that these were changed and those in my new class are established.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
jaclark5: In the development of my own package I remember having versions of `None` with different memory locations. Unfortunately, I can't find my notes/emails on discussions of this topic and I can't reproduce it, so I will yield.

testsuite/MDAnalysisTests/analysis/test_hydrogenbonds_analysis.py
jaclark5: Ya I'm not interested in spending the time to do that.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -455,31 +474,35 @@ def guess_donors(self, select='all', max_charge=-0.5):
...
         hydrogens_ag = self.u.select_atoms(hydrogens_sel)
 
-        ag = hydrogens_ag.residues.atoms.select_atoms(
-            "({donors_sel}) and around {d_h_cutoff} {hydrogens_sel}".format(
-                donors_sel=select,
-                d_h_cutoff=self.d_h_cutoff,
-                hydrogens_sel=hydrogens_sel
+        # We're using u._topology.bonds rather than u.bonds as it is a million times faster to access.
+        # This is because u.bonds also calculates properties of each bond (e.g bond length).
+        # See https://github.com/MDAnalysis/mdanalysis/issues/2396#issuecomment-596251787
```
IAlibay: 👍🏽

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
IAlibay: If you do find your notes on this I would be quite interested / open to further changes in follow-up work. The behaviour of `None` is quite odd compared to other comparisons, so it wouldn't surprise me if there were edge cases that haven't been considered.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -503,13 +526,40 @@ def guess_acceptors(self, select='all', max_charge=-0.5):
...
+
+    @staticmethod
+    def __group_categories(group):
+        """ Find categories according to universe constraints
+        
+        Parameters
+        ----------
+        group : AtomGroup
+            AtomGroups corresponding to either hydrogen bond acceptors, 
+            donors, or hydrogen atoms that meant their respective charge
```
IAlibay:
```suggestion
            donors, or hydrogen atoms that meet their respective charge
```

testsuite/MDAnalysisTests/analysis/test_hydrogenbonds_analysis.py
IAlibay: I see, thanks for the explanation, I think it's the diff messing with how things look.

testsuite/MDAnalysisTests/analysis/test_hydrogenbonds_analysis.py
IAlibay: It's a pretty big maintenance thing given the confusion with the comment below. Could I ask you to at least raise an issue looking to refactor these two classes? We can probably get someone looking for an easy first contribution to deal with this.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -455,31 +474,35 @@ def guess_donors(self, select='all', max_charge=-0.5):
...
-        ag = hydrogens_ag.residues.atoms.select_atoms(
-            "({donors_sel}) and around {d_h_cutoff} {hydrogens_sel}".format(
-                donors_sel=select,
-                d_h_cutoff=self.d_h_cutoff,
-                hydrogens_sel=hydrogens_sel
+        # We're using u._topology.bonds rather than u.bonds as it is a million times faster to access.
+        # This is because u.bonds also calculates properties of each bond (e.g bond length).
+        # See https://github.com/MDAnalysis/mdanalysis/issues/2396#issuecomment-596251787
+        if (hasattr(self.u._topology, 'bonds') and len(self.u._topology.bonds.values) != 0):
+            donors_ag = find_hydrogen_donors(hydrogens_ag)
```
IAlibay, line 509: I think I'm missing something obvious here, sorry. Doesn't this ignore `select`?

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -503,13 +526,40 @@ def guess_acceptors(self, select='all', max_charge=-0.5):
...
+            String for each hydrogen bond acceptor/donor/hydrogen atom category.
+        """
+
+        if hasattr(group,"resnames") and hasattr(group,"names"):
+            group_list = np.unique(
+                [
+                    '(resname {} and name {})'.format(r, p) for r, p in zip(group.resnames, group.names)
+                ]
+            )
+        else:
```
IAlibay, line 600: I'm still rather wary of this non optional fallback and also the behaviour this might have on force fields where there are equivalent types of different charges. @MDAnalysis/coredevs can we get a quick look at this please? I'm happy going with the consensus.

IAlibay (changes requested):
Just the one thing and one issue to raise please.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
jaclark5: I'll just point out that if there are no resnames, then a scheme like lammps is most likely used where "equivalent" atoms with different charges are also different types.

IAlibay: I think my views here would be alleviated if the type option wasn't a fallback. If it was a defined input variable someone could toggle then I can see this being less likely to cause problems in weird edge cases? (or at least, it's more of a user problem for not following the docstring)

It also makes it so that LAMMPS users aren't secondary citizens when it comes to this class? (unless there's a case where the fallback is needed?)

jaclark5: I appreciate your viewpoint and look forward to others with potential fridge cases. 

I was thinking the fallback behavior allows the current users to perform calculations "business as usual" but would automatically switch to the behavior required for LAMMPS users. Thus, LAMMPS users aren't second class in that the "default" behavior will work for them, rather than requiring them to find a keyword that allows their calculation to run.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
jaclark5: Good point!

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
IAlibay: I don't want this to get too lost, so I'm going to ping the coredev team again @MDAnalysis/coredevs

package/CHANGELOG
```diff
@@ -31,6 +31,9 @@ Fixes
     (e.g. bonds, angles) (PR #3779).
 
 Enhancements
+  * Update hbond analysis doc string to use exclusive bound language (#3847)
```
orbeckst: I think that's a "Fix".

package/CHANGELOG
```diff
@@ -31,6 +31,9 @@ Fixes
...
+  * Added ability for hbond anaylsis to use types when resnames aren't there
```
orbeckst: analysis

orbeckst: reference issue or PR, please

package/CHANGELOG
```diff
@@ -31,6 +31,9 @@ Fixes
...
+  * Added explanatory warnings when hbond analysis doesn't find any hbonds
```
orbeckst, line 46: reference issue or PR, please

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -174,6 +174,17 @@
 protein-water and protein-protein hydrogen bonds will be found, but
 no water-water hydrogen bonds.
 
+One can now also define hydrogen bonds with atom types::
```
orbeckst: remove "now" — write the documentation as "timeless"; the versionchanged/versionadded notes provide historical information

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -174,6 +174,17 @@
...
+
+  import MDAnalysis
```
orbeckst: I don't think this import is needed

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -260,6 +272,9 @@ def __init__(self, universe,
         """Set up atom selections and geometric criteria for finding hydrogen
         bonds in a Universe.
 
+        Hydrogen bond selections may be achieved with either a resname, atom 
+        name combination, or when those are absent, atom types.
+
```
orbeckst: reference the `donor_sel` and `acceptors_sel` keywords
```suggestion
        Hydrogen bond selections with `donors_sel` , `hydrogens_sel`, and
        `acceptors_sel` may be achieved with either a *resname*, atom *name* 
        combination, or when those are absent, with atom *type* selections.
        
```
I also emphasized the selection keywords.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -367,14 +384,19 @@ def guess_hydrogens(self,
                         ):
         """Guesses which hydrogen atoms should be used in the analysis.
 
+        Hydrogen selections may be achieved with either a resname, atom 
+        name combination, or when those are absent, atom types.
+
```
orbeckst: Move into the Notes selection below.

Add a versionchanged 2.4.0 for the new functionality.

I'd also suggest you reference the select keyword (in backticks) and emphasize the selection keyword *type*.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -407,24 +429,21 @@ def guess_hydrogens(self,
...
-
-        return " or ".join(hydrogens_list)
+        return self.__group_categories(hydrogens_ag)
 
     def guess_donors(self, select='all', max_charge=-0.5):
         """Guesses which atoms could be considered donors in the analysis. Only use if the universe topology does not
         contain bonding information, otherwise donor-hydrogen pairs may be incorrectly assigned.
 
+        Donor selections may be achieved with either a resname, atom 
+        name combination, or when those are absent, atom types.
```
orbeckst: Move into the Notes selection below.

Add a versionchanged 2.4.0 for the new functionality.

I'd also suggest you reference the select keyword (in backticks) and emphasize the selection keyword type.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
orbeckst: Or add under the `select` keyword.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
orbeckst: Or add under the `select` keyword.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
orbeckst: thank you for the comment and the link to the issue that confirmed the outrageous statement "a million times faster"

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -455,31 +474,37 @@ def guess_donors(self, select='all', max_charge=-0.5):
...
+        donors_ag = donors_ag[donors_ag.charges < max_charge]
+
+        return self.__group_categories(donors_ag)
 
     def guess_acceptors(self, select='all', max_charge=-0.5):
         """Guesses which atoms could be considered acceptors in the analysis.
 
+        Acceptor selections may be achieved with either a resname, atom 
+        name combination, or when those are absent, atom types.
+
```
orbeckst, line 529: Move into the Notes selection below or under the docs for `select`.

Add a versionchanged 2.4.0 for the new functionality.

I'd also suggest you reference the select keyword (in backticks) and emphasize the selection keyword type.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -503,13 +528,40 @@ def guess_acceptors(self, select='all', max_charge=-0.5):
...
+        ----------
+        group : AtomGroup
+            AtomGroups corresponding to either hydrogen bond acceptors, 
+            donors, or hydrogen atoms that meet their respective charge
+            and mass constraints. 
+
+        Returns
+        -------
+        select : str
+            String for each hydrogen bond acceptor/donor/hydrogen atom category.
```
orbeckst, line 588: add versionadded 2.4.0

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -503,13 +528,40 @@ def guess_acceptors(self, select='all', max_charge=-0.5):
...
+        """
+
+        if hasattr(group,"resnames") and hasattr(group,"names"):
```
orbeckst: pep8 — space after comma

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
orbeckst: The fact that we're relying on selections here instead of ag instances is iffy. I think that needs to change eventually. Reverse engineering a compact selection from an AG is just rife with potential failures, regardless of using types or resname+name. (We could create an `index` selection but I assume that performance-wise it might be bad to have a selection string with 1M numbers.)

As such I don't think that the proposed change will make the situation worse but it will give users with type-only universes the same functionality. 

I don't think that the problem is really the addition of the alternative *type* selection here. 

----
⚠️ I wanted to suggest to make it
```python
      elif hasattr(group, "types"):
```
and add a final `else` that raises a `NoDataError` but **this point is moot** because [types is a canonical/guessed attribute](https://userguide.mdanalysis.org/stable/topology_system.html#canonical-attributes) and will always be present.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -800,8 +866,14 @@ def count_by_type(self):
         d = self.u.atoms[self.results.hbonds[:, 1].astype(np.intp)]
         a = self.u.atoms[self.results.hbonds[:, 3].astype(np.intp)]
 
-        tmp_hbonds = np.array([d.resnames, d.types, a.resnames, a.types],
-                              dtype=str).T
+        if hasattr(d,"resnames"):
```
orbeckst: pep8 spaces after comma

orbeckst (changes requested):
Minor doc changes + pep 8, please.

I am basically ok with the use of types, given that what we do is messy either way... see comment.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
lilyminium: Generally agreed with @orbeckst . From skimming the code it looks like:

* the reason we need a selection string is to make an updating atomgroup
* the selection strings output by this method are static? (i.e. name and resname, or, type)

Would it be possible / a good idea to redo the method to return an AtomGroup or UpdatingAtomGroup directly where it's called in `_prepare`, if an `acceptors_sel` is not provided?

(`ag.select_atoms("all", updating=True)`)

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
jaclark5: Is this what you were thinking?

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
jaclark5: I actually copied this from the _get_dh_pairs()

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -370,11 +387,14 @@ def guess_hydrogens(self,
         Parameters
         ----------
         select: str (optional)
-            Selection string for atom group from which hydrogens will be identified.
+            :doc:`Selection string </documentation_pages/selections>` for atom group from which hydrogens
```
orbeckst: Don't use doc links but instead `:ref:` to an anchor as these are more robust. In this case, a label is already defined in [`selections.rst`](https://raw.githubusercontent.com/MDAnalysis/mdanalysis/develop/package/doc/sphinx/source/documentation_pages/selections.rst) so use
```reST
:ref:`Selection string <selection-commands-label>`
```
(I think the version with `<label>` works but if not, just use the label in the ref.)
```

File: package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py [hunk 37]
```diff
@@ -422,9 +443,10 @@ def guess_donors(self, select='all', max_charge=-0.5):
         Parameters
         ----------
         select: str (optional)
-            Selection string for atom group from which donors will be identified.
+            :doc:`Selection string </documentation_pages/selections>` for atom group from which donors
```
orbeckst: see above, use ref please

orbeckst: please also keep within 79 char linewidth (unfortunately, PEP8 checks are currently not running #3854 )

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -455,31 +483,38 @@ def guess_donors(self, select='all', max_charge=-0.5):
...
 
-        ag = hydrogens_ag.residues.atoms.select_atoms(
-            "({donors_sel}) and around {d_h_cutoff} {hydrogens_sel}".format(
-                donors_sel=select,
-                d_h_cutoff=self.d_h_cutoff,
-                hydrogens_sel=hydrogens_sel
+        # We're using u._topology.bonds rather than u.bonds as it is a million times faster to access.
+        # This is because u.bonds also calculates properties of each bond (e.g bond length).
+        # See https://github.com/MDAnalysis/mdanalysis/issues/2396#issuecomment-596251787
+        if (hasattr(self.u._topology, 'bonds') and len(self.u._topology.bonds.values) != 0):
```
orbeckst: the outer parentheses are not needed

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -455,31 +483,38 @@ def guess_donors(self, select='all', max_charge=-0.5):
...
         """Guesses which atoms could be considered acceptors in the analysis.
 
+        Acceptor selections may be achieved with either a resname, atom 
+        name combination, or when those are absent, atom types.
+
         Parameters
         ----------
         select: str (optional)
-            Selection string for atom group from which acceptors will be identified.
+            :doc:`Selection string </documentation_pages/selections>` for atom group from which acceptors will
```
orbeckst: use ref please

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
orbeckst: remove from here as it is now in the Notes

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -310,6 +325,8 @@ def __init__(self, universe,
         information is used, as this is the only way that guarantees the
         correct identification of donor-hydrogen pairs.
 
+        .. versionchanged:: 2.4.0
+            Added use of atom types
```
orbeckst: I'd prefer indentation under "v" but given the precedence, I'm ok with indentation under "e".

I *think* we normally order the versionchanged as oldest to newest so then you'd need to switch order (but please check).

orbeckst: You need *two* empty lines above versionchanged, otherwise formatting errors can happen.

This applies to _all_ versionchanged that you added.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -393,6 +416,10 @@ def guess_hydrogens(self,
         Alternatively, this function may be used to quickly generate a :class:`str` of potential hydrogen atoms involved
         in hydrogen bonding. This str may then be modified before being used to set the attribute
         :attr:`hydrogens_sel`.
+
```
orbeckst, line 427: add newline

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -445,6 +470,9 @@ def guess_donors(self, select='all', max_charge=-0.5):
         in hydrogen bonding. This :class:`str` may then be modified before being used to set the attribute
         :attr:`donors_sel`.
 
+        .. versionchanged:: 2.4.0
```
orbeckst, line 489: add newline above

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -499,17 +537,50 @@ def guess_acceptors(self, select='all', max_charge=-0.5):
         in hydrogen bonding. This :class:`str` may then be modified before being used to set the attribute
         :attr:`acceptors_sel`.
 
+        .. versionchanged:: 2.4.0
```
orbeckst, line 564: add newline above

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -499,17 +537,50 @@ def guess_acceptors(self, select='all', max_charge=-0.5):
...
-            [
-                '(resname {} and name {})'.format(r, p) for r, p in zip(acceptors_ag.resnames, acceptors_ag.names)
-            ]
-        )
 
-        return " or ".join(acceptors_list)
+        return self.__group_categories(acceptors_ag)
+
+    @staticmethod
+    def __group_categories(group):
```
orbeckst: I don't see a need to make it double-underscore. Just make it single underscore, that's enough to mark it as "private" and not part of the public API. This would also make it easier to test this fragile method by itself.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -499,17 +537,50 @@ def guess_acceptors(self, select='all', max_charge=-0.5):
...
+        group : AtomGroup
+            AtomGroups corresponding to either hydrogen bond acceptors, 
+            donors, or hydrogen atoms that meet their respective charge
+            and mass constraints. 
+
+        Returns
+        -------
+        select : str
+            String for each hydrogen bond acceptor/donor/hydrogen atom category.
+
```
orbeckst, line 589: probably needs a second newline

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -622,6 +693,13 @@ def _single_frame(self):
             return_distances=True,
         )
 
+        if np.size(d_a_indices) == 0:
```
orbeckst, line 721: Do we need `np.size` here, can't we just do `if not d_a_indices:` and simplify the code?

(I don't really know what kind of data structure `d_a_indices` is so please explain.)

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -644,6 +722,13 @@ def _single_frame(self):
         )
         hbond_indices = np.where(d_h_a_angles > self.d_h_a_angle)[0]
 
+        if np.size(hbond_indices) == 0:
```
orbeckst, line 750: As above, is `np.size` needed?

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -800,8 +885,14 @@ def count_by_type(self):
         d = self.u.atoms[self.results.hbonds[:, 1].astype(np.intp)]
         a = self.u.atoms[self.results.hbonds[:, 3].astype(np.intp)]
 
-        tmp_hbonds = np.array([d.resnames, d.types, a.resnames, a.types],
-                              dtype=str).T
+        if hasattr(d, "resnames"):
+            d_res = d.resnames
+            a_res = a.resnames
+        else:
+            d_res = [None for x in range(len(d.types))]
```
orbeckst: simplify to
```python
d_res = len(d.types) * [None]
a_res = len(a.types) * [None]
```

orbeckst: Given that this is turned into `str`, replace `None` with `"None"`?? It's twice as fast for me
```python
In [5]: a = 1000*[None]

In [6]: %timeit np.array(a, dtype=str)
255 µs ± 2.47 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [7]: b = 1000*["None"]

In [8]: %timeit np.array(b, dtype=str)
128 µs ± 561 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```
although it's already fast so perhaps clarity of using `None` is better than introducing magic-looking strings.

However, given that these resnames are not really necessary for anything, why not just use `np.zeros(len(a.types), dtype=str)`
```python
In [10]: %timeit np.zeros(1000, dtype=str)
412 ns ± 4.31 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

orbeckst (changes requested):
Thanks for addressing my previous requests. I found a few doc things and suggestions for simplifying code.

I am waiting for @IAlibay (and anyone else) to weigh in on the `__group_categories()` function.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
jaclark5: I let that slide since the rest of the doc strings didn't adhere.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
jaclark5: I changed it to use "None", thanks for the benchmarking. I think the suggest behavior of using zero could lead a user to believe that their whole system is important in with a uniform integer as a rename. While, use of "None" is unambiguous.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
jaclark5: The variable `hbond_indices` is a np.array, so if it is not empty the error "The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()" arises.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
orbeckst: raise issue to modernize HBA to use AG as objects internally instead of selection strings>?

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
orbeckst: Thanks for explaining!

orbeckst (approved):
Thank you for addressing my requested changes. LGTM!

package/CHANGELOG
```diff
@@ -31,6 +32,10 @@ Fixes
     (e.g. bonds, angles) (PR #3779).
 
 Enhancements
+  * Added ability for hbond analysis to use types when resnames are not 
+    present (#3847)
+  * Added explanatory warnings when hbond analysis doesn't find any hbonds
+    (#3847)
```
hmacdope: PR or issue?

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -310,8 +325,12 @@ def __init__(self, universe,
         information is used, as this is the only way that guarantees the
         correct identification of donor-hydrogen pairs.
 
+
         .. versionadded:: 2.0.0
             Added `between` keyword
+        .. versionchanged:: 2.4.0
+            Added use of atom types
```
hmacdope: Could you make this a tiny bit more detailed?

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -407,117 +442,177 @@ def guess_hydrogens(self,
...
-            Maximum allowed charge of a donor atom.
+            The charge of a donor atom must be less than this value.
 
         Returns
         -------
         potential_donors: str
-            String containing the :attr:`resname` and :attr:`name` of all atoms that potentially capable of forming
-            hydrogen bonds.
+            String containing the :attr:`resname` and :attr:`name` of all atoms 
+            that potentially capable of forming hydrogen bonds.
```
hmacdope:
```suggestion
            that are potentially capable of forming hydrogen bonds.
```

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -644,6 +746,13 @@ def _single_frame(self):
         )
         hbond_indices = np.where(d_h_a_angles > self.d_h_a_angle)[0]
 
+        if np.size(hbond_indices) == 0:
+            warnings.warn(
```
hmacdope, line 751: Test this warning? :)

hmacdope (changes requested):
Just a few minor things.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
hmacdope: @jaclark5 would you be able to raise this as an issue if not already raised?

hmacdope (approved):
Other than raising the issue mentioned, all the comments I had have been addressed.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
jaclark5: I'm not sure why I'm making an issue when I don't "have" an issue, but it's done.

orbeckst: Thank you for raising #3933.

"Please raise an issue" is a common request in reviews. When we look at code as part of a review and find something needs to be changed but we do not want to overburden you (by requesting it as a change on the current PR) then we ask you to create the "TODO" instead. Otherwise this information will get lost. The issue tracker serves as our collective memory.

testsuite/MDAnalysisTests/analysis/test_hydrogenbonds_analysis.py
jaclark5: I made a child class to resolve this.

IAlibay:
@jaclark5 I've not forgotten about this, it's at the top of my review list over the next few days.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -407,117 +443,177 @@ def guess_hydrogens(self,
...
+        .. versionadded:: 2.4.0
+
+        """
+
+        if hasattr(group, "resnames") and hasattr(group, "names"):
+            group_list = np.unique([
+                '(resname {} and name {})'.format(r, 
+                    p) for r, p in zip(group.resnames, group.names)
+            ])
+        else:
```
IAlibay, line 600: I still think we should have the option to _choose_ whether or not to default to types here, however for the sake of getting this done I'm going to let it go as-is here, and raise a separate issue to deal with later.

testsuite/MDAnalysisTests/analysis/test_hydrogenbonds_analysis.py
```diff
@@ -246,11 +271,97 @@ def test_logging_step_not_1(self, universe, caplog):
...
         hbonds.lifetime(tau_max=2, intermittency=1)
 
-        warning = \
-            "Autocorrelation: Hydrogen bonds were computed with step > 1."
+        warning = ("Autocorrelation: Hydrogen bonds were computed with "
+                   "step > 1.")
         assert any(warning in rec.getMessage() for rec in caplog.records)
 
 
+class TestHydrogenBondAnalysisNoRes(TestHydrogenBondAnalysisIdeal):
```
IAlibay, line 279: Thanks for doing this, avoiding duplicate code as much as possible really helps us avoid future maintenance debt.

IAlibay (approved):
lgtm, thanks for doing this @jaclark5. Sorry it took so long to re-review, but at least it'll make it for the 2.4.0 release 🎉 !

orbeckst:
Thank you @jaclark5 , nice work! 🎉

PR #3855, opened 2022-10-02
Remove duplicate author

IAlibay:
Tiny one but this always comes up every time I have to work out our current author count.

This is just something that fell out of the testsuite authors merge in https://github.com/MDAnalysis/mdanalysis/pull/2100, turns out one of the authors was using their github handle for one and their real name for the other.

PR Checklist
------------
 - Tests?
 - Docs?
 - CHANGELOG updated?
 - Issue raised/referenced?

### DIFF
--- a/package/AUTHORS
+++ b/package/AUTHORS
@@ -82,7 +82,6 @@ Chronological list of authors
   - Pedro Reis
  2017
   - Ruggero Cortini
-  - Xiki Tempula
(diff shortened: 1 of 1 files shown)


richardjgowers (approved):

IAlibay:
Looks like we're also encountering https://github.com/actions/first-interaction/issues/101

PR #3857, opened 2022-10-10
WIP, CI: 3.11rc2 check

tylerjereddy:
* Python 3.11rc2 is guaranteed ABI stable with the final release per: https://www.python.org/downloads/release/python-3110rc2/

* the final release is about two weeks away, so let's see what goes wrong with `3.11` a bit before it hits; most of the major dependencies have releasd PyPI wheels because of the above ABI stability guarantee

* I believe the CI workflow I adjusted may use a mix of conda and pip, so I'm less confident of what will go wrong vs. pure PyPI workflow

### DIFF
--- a/.github/workflows/gh-ci.yaml
+++ b/.github/workflows/gh-ci.yaml
@@ -28,7 +28,7 @@ jobs:
         fail-fast: false
         matrix:
           os: [ubuntu-latest, ]
-          python-version: [3.8, 3.9, "3.10"]
+          python-version: [3.8, 3.9, "3.10", "3.11.0-rc.2"]
           full-deps: [true, ]
           install_hole: [true, ]
           codecov: [true, ]
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -48,6 +48,12 @@ jobs:
           BUILD_TYPE: 'wheel'
           NUMPY_MIN: '1.20.0'
           imageName: 'windows-2019'
+        Linux-Python311rc-64bit-full-wheel:
+          PYTHON_VERSION: '3.11.0-rc.2'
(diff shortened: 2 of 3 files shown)


.github/workflows/gh-ci.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
         fail-fast: false
         matrix:
           os: [ubuntu-latest, ]
-          python-version: [3.8, 3.9, "3.10"]
+          python-version: [3.8, 3.9, "3.10", "3.11.0-rc.2"]
```
IAlibay, line 31: I suspect our non minimal deps will fail here, like rdkit and OpenMM

Might need to put this as a separate entry with a minimal deps argument.

IAlibay:
Setting up a pure pypi runner for this is going to take a bit more work I'm afraid (since we rely on mamba as our base installer for our normal actions CI). Might just be easier to put this in the azure pipelines that are pure pypi?

IAlibay:
@tylerjereddy do you mind if I jump in here and test this out?

tylerjereddy:
@IAlibay yeah go for it, may tell you a bit about some useful changes that can be made soon if needed

IAlibay:
Just testing out this azp thing here now, then I'll probably open a separate PR to not spam too much. Easiest thing imho will be to just add a runner in cron CI that picks up the latest unstable version from https://github.com/actions/python-versions/blob/main/versions-manifest.json

IAlibay:
Ok MDA wheel build fine, but the presence of h5py in the azp script is messing with us. I'll switch to a new PR at this point to keep the commit noise down.

IAlibay:
Closing as superseded by #3862

Issue #3858, opened 2022-10-12
Deprecating windows 32bit support


IAlibay:
As noticed recently, the 32bit windows azure runner has been failing.

This appears to stem from a mix of the recent scipy 1.9.2 release which no longer providing win32 wheels, and azure not providing a suitable toolchain for compiling a 32bit version of scipy.

A deeper conversation on this can be seen here: https://discuss.scientific-python.org/t/releasing-or-not-32-bit-windows-wheels/282

As a consequence, it appears (and this is what I've been told on the scientific-python scipy channel - @tylerjereddy may correct me here), that scipy is going ahead with dropping win32 support completely.

Overall I see this as a good opportunity to consider also aiming to drop 32bit support completely in the near future. 

At worst we should drop support once scipy 1.9.1 is no longer supported, so ~ 2 years from now?

Pinging @tylerjereddy for further comments here on the situation re: scipy and 32bit

PR #3859, opened 2022-10-12
Pin 32bit Windows Azure runner to 1.9.1

IAlibay:
Towards #3858

 - Pins the scipy version for win32 to 1.9.1.

PR Checklist
------------
 - Tests?
 - Docs?
 - CHANGELOG updated?

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -26,7 +26,7 @@ jobs:
           PYTHON_ARCH: 'x86'
           BUILD_TYPE: 'normal'
           imageName: 'windows-2019'
-          scipyPin: '==1.9.1'
+          PIP_PRE_EXTRAS: 'scipy==1.9.1'
         Win-Python38-64bit-full:
           PYTHON_VERSION: '3.8'
           PYTHON_ARCH: 'x64'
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -80,6 +81,9 @@ jobs:
     condition: and(succeeded(), eq(variables['BUILD_TYPE'], 'wheel'))
   - script: python -m pip install --upgrade pip setuptools wheel
     displayName: 'Install tools'
+  - script: python -m pip install $(PIP_PRE_EXTRAS)
+    displayName: 'Pre-deps extras'
+    condition: ne(variables['PIP_PRE_EXTRAS'], '')
```
tylerjereddy, line 86: Won't this get overridden by the `pip install scipy` below? I suppose another approach might be to use `--only-binary="h5py,scipy"` or whatever the exact syntax for that is, to avoid trying a source install of `1.9.2`.

Apart from that, your assessment is consistent with my understanding--SciPy and scitkit-learn seem to be on a path to removing win 32-bit support, though some debate still seems to be happening re: some folks using 32-bit Python on Windows if only by some default Windows download mechansim.

tylerjereddy: ah, I guess there's no `upgrade` flag so you're ok

tylerjereddy (approved):

azure-pipelines.yml
IAlibay: o yeah that's very true only binary might be cleaner, I'm happy to go with that if you like that option too

IAlibay: too late, o well there's a couple of uses to that extra step I can think of anyways

PR #3863, opened 2022-10-14
Adding Appending Functionality method to WriterBase

### LINKED ISSUE #3861: Adding Appending Functionality to Writers
## Current Writer API ##

```python
# option 1
AtomGroup.write(filename,..., **kwargs)

# option 2
with mda.Writer(filename, n_atoms) as w:
    for ts in Universe.trajectory:
        w.write(AtomGroup)
```

## New API (`append=False` in default)
```python
# option 1
AtomGroup.write(filename,..., append=False, **kwargs)

# option 2
with mda.Writer(filename, n_atoms, append=False) as w:
    for ts in Universe.trajectory:
        w.write(AtomGroup)
```
```python
# adding new Base class for single frame writer
# and making sure all writers do not overwrite `write` function
class WriterBase(WriterBase):
    def __init__(self, filename, n_atoms, append=False):
        self.filename = filename
        self.n_atoms = n_atoms
        if append:
            self.check_appendibility()
        self.append = append
        self._n_frames_written = 0

    def write(self, obj, **kwargs):
        self._n_frames_written += 1
        return self._write_next_frame(obj, **kwargs)

class SingleFrameWriterBase(WriterBase):
    def __init__(self, filename, n_atoms, append=False):
        self.filename = filename
        self.n_atoms = n_atoms
        if append:
            raise ValueError("Single frame writers do not support appending")
        self.append = append
        self._n_frames_written = 0
    
    def write(self, obj, **kwargs):
        if self._n_frames_written >= 1:
            raise ValueError("Single frame writers can only write one frame")

        self._n_frames_written += 1
        return self._write_next_frame(obj, **kwargs)
```
I also find the current API quite messy with mixed uses of `obj` and `ag`---let alone `ts` (https://github.com/MDAnalysis/mdanalysis/issues/2757). I will just make sure all writers do not overwrite the `write` function for now...

## Note ##
- This issue also addresses a potential bug that it doesn't raise any warning/error when one tries to write multiple frames to an (implicit) single-frame writer:
```python
with mda.Writer("test.gro", u.atoms.n_atoms) as W:
    for ts in u.trajectory:
        W.write(u.atoms)
```
- This issue only concerns writing trajectories with constant `n_atoms`. (Related: https://github.com/MDAnalysis/mdanalysis/issues/3836)
- (minor) add supporting information to https://userguide.mdanalysis.org/stable/formats/index.html
- In terms of things to check before appending, For now (in my mind), the writer should check `n_atoms` matches. I will probably learn my lessons along the way :)

## Use Cases ##
- Perform some post-processing (e.g. transformations to fix PBC issues) to raw simulations and save them as new trajectory files. When the raw simulations update, with appending, not all frames need to be processed over and over again. This could be a potential selling point because I think `gmx trjconv` doesn't support appending :>)
- Do frame extraction from multiple processes. (this might need a proper file lock)
- pending...

Quoting @richardjgowers 
> The way forward to this is to flag every reader as non-append friendly, so they raise errors if mode==‘a’, then slowly unflag them as tests/functionality is rolled out in future PRs.

yuxuanzhuang:
Fixes #3861 

 - add new SingleFrameWriterBase
 - add new flag `append` for appending trajectory to WriterBase
 - add a check for the number of frames written by WriterBase.
 - make sure `write` is not overwritten by Writers (they only modified `_write_next_frame`)

TODO:
- add tests for writers that are not covered by BaseWriterTest

PR Checklist
------------
 - [x] Tests?
 - [] Docs?
 - [] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/coordinates/base.py
+++ b/package/MDAnalysis/coordinates/base.py
@@ -1475,6 +1475,25 @@ class WriterBase(IOBase, metaclass=_Writermeta):
        :func:`write` instead.
 
     """
+    def __init__(self,
+                 filename,
+                 n_atoms,
+                 append=False):
+        """Set up single frame writer.
+
+        Parameters
+        ----------
+        filename : str
+            filename of trajectory file
+        n_atoms : int
+            number of atoms in trajectory
+        append : bool (optional)
+            whether to append to an existing file or overwrite it
+        """
+        self.filename = filename
+        self.n_atoms = n_atoms  # number of atoms to be written
+        self.append = append
+        self._n_frames_written = 0
 
     def convert_dimensions_to_unitcell(self, ts, inplace=True):
         """Read dimensions from timestep *ts* and return appropriate unitcell.
@@ -1493,7 +1512,7 @@ class WriterBase(IOBase, metaclass=_Writermeta):
         lengths = self.convert_pos_to_native(lengths)
         return np.concatenate([lengths, angles])
 
-    def write(self, obj):
+    def write(self, obj, **kwargs):
         """Write current timestep, using the supplied `obj`.
 
         Parameters
@@ -1511,7 +1530,8 @@ class WriterBase(IOBase, metaclass=_Writermeta):
            Deprecated support for Timestep argument to write has now been
            removed. Use AtomGroup or Universe as an input instead.
         """
-        return self._write_next_frame(obj)
+        self._n_frames_written += 1
+        return self._write_next_frame(obj, **kwargs)
 
     def __del__(self):
         self.close()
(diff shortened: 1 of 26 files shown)


package/MDAnalysis/coordinates/base.py
```diff
@@ -1545,6 +1578,61 @@ def has_valid_coordinates(self, criteria, x):
...
+
+        Parameters
+        ----------
+        filename : str
+            filename of trajectory file
+        n_atoms : int
+            number of atoms in trajectory
+        append: bool (optional)
+            If ``True``, append to an existing file. If ``False``, overwrite
+            the file. Default is ``False``.
```
richardjgowers, line 1604: Document that this option only exists for API compatibility and will raise an issue

package/MDAnalysis/coordinates/chemfiles.py
```diff
@@ -308,8 +309,7 @@ def __init__(
             raise RuntimeError(
                 "Please install Chemfiles > {}" "".format(MIN_CHEMFILES_VERSION)
             )
-        self.filename = filename
-        self.n_atoms = n_atoms
+        super().__init__(filename, n_atoms, append)
         if mode != "a" and mode != "w":
             raise IOError("Expected 'a' or 'w' as mode in ChemfilesWriter")
```
richardjgowers, line 314: ChemFiles is possibly using `'a'` for append mode here

richardjgowers (changes requested):
ChemFiles treatment needs fixing.  I think `append = append or 'a' in mode` might work

package/MDAnalysis/coordinates/chemfiles.py
yuxuanzhuang: It's up for discussion but my idea is to focus this PR on only adding general API to WriterBase and leave specific implementations to other PRs (in a batch for similar Writers) so it doesn't get messy :)

Also after this PR is merged, the current API for `mode=a` (if anyone is really using it) should not break.

IAlibay: I support @yuxuanzhuang's approach here, keep it simple to make sure we can merge quickly.

PR #3868, opened 2022-10-18
Add rmsip docs

lilyminium:
~~Fixes #~~

 - Add an example to RMSIP function to clarify use

PR Checklist
------------
 - [ ] Tests?
 - [x] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/doc/sphinx/source/conf.py
+++ b/package/doc/sphinx/source/conf.py
@@ -39,12 +39,18 @@ sys.path.insert(0, os.path.abspath('../../..'))
 
 # Add any Sphinx extension module names here, as strings. They can be extensions
 # coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
-extensions = ['sphinx.ext.autodoc', 'sphinx.ext.intersphinx',
-              'sphinx.ext.mathjax', 'sphinx.ext.viewcode',
-              'sphinx.ext.napoleon', 'sphinx.ext.todo',
-              'sphinx_sitemap',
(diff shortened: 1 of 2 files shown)


lilyminium:
RTD is throwing an error on installing openblas, but `make html` locally produces the following:

Screen Shot 2022-10-18 at 12 35 23 pm
Screen Shot 2022-10-18 at 12 35 40 pm

package/doc/sphinx/source/conf.py
```diff
@@ -39,12 +39,18 @@
...
+    'sphinx.ext.autodoc',
+    'sphinx.ext.intersphinx',
+    'sphinx.ext.mathjax',
+    'sphinx.ext.viewcode',
+    'sphinx.ext.napoleon',
+    'sphinx.ext.todo',
+    'sphinx_sitemap',
+    'sphinx_rtd_theme',
+    'sphinxcontrib.bibtex',
+    'sphinx.ext.doctest',
```
IAlibay, line 52: new entries?

IAlibay (approved):

IAlibay:
The RTD openblas error is prevalent on more or less every conda build atm (I think mamba builds are not affected because they use use cf as the base channel), I suspect it'll get fixed in the morning. However it shouldn't be a blocker to merging this imho.

package/doc/sphinx/source/conf.py
lilyminium: I added `doctest` but everything else is just a reformat

PR #3871, opened 2022-10-20
[to be discussed] Don't limit pyproject.toml NumPy support to known versions

IAlibay:
This is something that I realised recently after looking at a few packaging things over the last few days.

The way we define the NumPy bounds in `pyproject.toml` make it such that if a new version of Python is released, then a released version of the MDA package can no longer be installed with that new Python version because there are no adequate NumPy version defined for it in `pyproject.toml`.

The proposed way around this is to just set `pyproject.toml` to have a bare `numpy` install requirement when dealing with unknowns, such as for unreleased Python versions (note: scipy does this too).

On the positive side of things, this will allow folks to use packages beyond the known dependency ranges at the time of release.

On the negative side of things, this may lead folks to use older versions of MDA with newer dependencies in a way that _could_ be broken.

**My personal view:**

I would prefer allowing installs with unknown versions of Python, in part because we happen to be upstream of a lot of packages that might want to test their code with new versions of Python.

We don't apply upper dependency pins on anything else, so it seems weird that we apply it to NumPy.

One middle ground option here is to use an upper bound on the `numpy` entry, such that it is no greater than the next semver major version (i.e. `numpy<2.0`). It doesn't really do much but at least it'll protect against really bad API changes?

PR Checklist
------------
 - Tests?
 - Docs?
 - CHANGELOG updated?
 - Issue raised/referenced?

### DIFF
--- a/package/pyproject.toml
+++ b/package/pyproject.toml
@@ -4,6 +4,7 @@ requires = [
     "Cython>=0.28",
     "packaging",
     # lowest NumPy we can use for a given Python,
+    # In part adapted from: https://github.com/scipy/oldest-supported-numpy/blob/main/setup.cfg
     # except for more exotic platform (Mac Arm flavors)
     # aarch64, AIX, s390x all support < 1.20 so we can safely pin to this
     "numpy==1.20.0; python_version=='3.8' and (platform_machine!='arm64' or platform_system!='Darwin')",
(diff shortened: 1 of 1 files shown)


IAlibay:
I've also added some PyPy entries, I suspect that doesn't even work with MDA? Are we ok just not allowing PyPy at all?

hmacdope:
Makes sense to me.

hmacdope (approved):

orbeckst:
I think your "middle ground solution" to restrict to numpy < 2 is a good solution.

orbeckst (changes requested):
Limit numpy only by next major release to allow more downstream flexibility (@IAlibay 's "middle ground solution".

IAlibay:
> Limit numpy only by next major release to allow more downstream flexibility (@IAlibay 's "middle ground solution".

I'm happy going with that, thoughts @hmacdope as having originally approved as-is?

Also any other @MDAnalysis/coredevs that want to chime in here, this will probably be our approach to upper pins.

hmacdope:
LGTM @IAlibay

richardjgowers (approved):

hmacdope:
@orbeckst I think this might be good to roll? @IAlibay?

orbeckst (approved):

PR #3886, opened 2022-10-26
Fix 32bit azure

IAlibay:
Azure 32 bit is failing again, looks like somehow pip is just ignoring the pre-installed scipy and is trying to go ahead with an install of scipy again.

I've switched things to what @tylerjereddy proposed last time (just rely on `only-binary`). Let's see what happens.

PR Checklist
------------
 - Tests?
 - Docs?
 - CHANGELOG updated?
 - Issue raised/referenced?


IAlibay:
That seems to do the trick

richardjgowers (approved):

PR #3888, opened 2022-10-27
Improve libdcd cython

### LINKED ISSUE #3882: Improve DCD reader typing and performance.
The Cython DCD reader could do with a little bit of an update. There was a large regression in its performance compared to the other readers as measured by the ASV benchmarks around 6ae0eb12  See attached screen-cap of the ASV benchmarks. 

There is room for improvement in the typing used in the cython. Cython can be very particular about the use of types wherever possible. Inspection of the annotated html indicates there is room for improvement that will hopefully help with performance. 

Screenshot from 2022-10-26 15-35-43

### LINKED ISSUE #3923: Improve deprecation warning for DCDReader copying timestep
Issue for proper paper trail

Add a deprecation warning for the DCDReader copying timestep unlike the other readers. Behaviour documented in #3889.

paraphrasing:

### Expected behavior ###

The readers should all have the same behaviour with respect to iterating through the trajectory and changing the base values in the `position` attribute. 

### Actual behavior ###

The DCD reader creates a new **copy** of the current timestep which allows the coordinates in each frame to vary independently.

hmacdope:
Fixes #3882 #3923

Improves the cython in the DCD reader to be more compiled.  

 - Improves the cython in the DCD reader. 

PR Checklist
------------
 - [X] Tests?
 - [X]  Docs?
 - [X] CHANGELOG updated?


package/MDAnalysis/coordinates/DCD.py
```diff
@@ -191,7 +191,7 @@ def _read_next_timestep(self, ts=None):
         if ts is None:
             # use a copy to avoid that ts always points to the same reference
             # removing this breaks lammps reader
-            ts = self.ts.copy()
+            ts = self.ts.copy() # why is this copy required ??
```
hmacdope: @orbeckst @MDAnalysis/coredevs does anyone know specifically why this copy is required? None of the other readers use this pattern and it is ~30% overhead but removing it breaks a bunch of tests. I will keep working on it but if anyone has a quick answer would be appreciated.

hmacdope: I will keep working on it in the meantime but the behaviour seems to be inconsistent with the other readers.

hmacdope: I am going to move discussion to an issue.

hmacdope: Moved to #3889

package/MDAnalysis/lib/formats/libdcd.pyx
```diff
@@ -88,18 +89,23 @@ cdef extern from 'sys/types.h':
...
 cdef enum:
     FIO_READ = 0x01
     FIO_WRITE = 0x02
 
+# exportable versions in python
 DCD_IS_CHARMM       = 0x01
 DCD_HAS_4DIMS       = 0x02
 DCD_HAS_EXTRA_BLOCK = 0x04
+# module level C decls 
+cdef  int DCD_IS_CHARMM_       = 0x01
```
richardjgowers: Maybe the compiler is consting this out, but iirc there is a version of #Define in Cython

package/MDAnalysis/coordinates/DCD.py
```diff
@@ -191,11 +191,10 @@ def _read_next_timestep(self, ts=None):
         if ts is None:
             # use a copy to avoid that ts always points to the same reference
             # removing this breaks lammps reader
-            ts = self.ts.copy() # why is this copy required ??
+            ts = self.ts # why is this copy required ??
         frame = self._file.read()
         self._frame += 1
-        ts = self._frame_to_ts(frame, ts)
-        self.ts = ts
```
orbeckst, line 197: Do the other readers not set self.ts?

orbeckst (changes requested):
Looking fine. Just minor question and 
- docs (versionchanged)
- Changelog update 

thanks.

package/CHANGELOG
```diff
@@ -30,6 +32,7 @@ Fixes
     (e.g. bonds, angles) (PR #3779).
 
 Enhancements
+  * Quality of life updates to libdcd Cython (Issue #3882, PR #3888)
```
orbeckst: Be more specific.

package/MDAnalysis/coordinates/DCD.py
```diff
@@ -191,11 +191,10 @@ def _read_next_timestep(self, ts=None):
         if ts is None:
             # use a copy to avoid that ts always points to the same reference
```
orbeckst: Can this comment be removed?

package/MDAnalysis/coordinates/DCD.py
```diff
@@ -191,11 +191,10 @@ def _read_next_timestep(self, ts=None):
...
             # removing this breaks lammps reader
-            ts = self.ts.copy()
+            ts = self.ts # why is this copy required ??
```
orbeckst: I don’t understand the comment — remove?

package/MDAnalysis/lib/formats/libdcd.pyx
orbeckst: Do you have to ~~repay~~ set the values explicitly twice? Can’t you only define the ~~Constanta~~ constants once?

orbeckst: Constants

(typing on crappy device)

package/CHANGELOG
orbeckst: If it improves performance then say that. If it had any user facing consequences then definitely say so. If it’s code modernization then we can list it (but we don’t always have been consistent iirc - @IAlibay ?.

orbeckst:
@hmacdope for my first review I only saw one set of changes. You clearly have a change log entry. For other things I just added comments. (Sorry for bad typing, reviewing from phone is not ideal)

package/CHANGELOG
IAlibay: Pretty much, we list it if there's a direct user impact (we have historically taken a poor stance towards acknowledging QoL / maintenance work unfortunately, that's something to be revisted)

package/setup.cfg
```diff
@@ -6,6 +6,8 @@
 ## Uncomment the following two lines to get the _release_ behavior.
 #use_cython=False
 #keep_cythonized=True
+## Uncomment the below line to generate the annotated cython HTML output
+# annotate_cython=True
```
IAlibay: Didn't we merge this in already?

testsuite/MDAnalysisTests/coordinates/base.py
```diff
@@ -444,6 +444,17 @@ def test_pickle_reader(self, reader):
         assert_equal(len(reader), len(reader_p))
         assert_equal(reader.ts, reader_p.ts,
                      "Timestep is changed after pickling")
+    
+    def test_frame_collect_all_same(self, reader):
+        # check that the timestep resets so that the base pointer is the same 
+        # for all timesteps in a collection witht eh exception of memoryreader
+        if isinstance(reader, mda.coordinates.memory.MemoryReader):
+            pytest.xfail()
```
IAlibay: give xfail a reason?

package/MDAnalysis/coordinates/DCD.py
kain88-de: This copy used to be required for the lammps reader 5 years ago as the comment says. This might have changed due to refactorings in the lammps reader since then.

However, the changes you had to make in the tests mean reaming the copy will break user code. Pretty sure a good number of phd students using only dcd files never noticed that `ts.positions` is only a reference. Maybe think if it is worth it to keep the copy for compatibility.

package/MDAnalysis/lib/formats/libdcd.pyx
hmacdope: It's to avoid cython having to juggle between the two. The C level constants can be baked in and the Python constants are used where the pythonic behaviour was required eg behaviour wrt &.

package/MDAnalysis/coordinates/DCD.py
hmacdope: Yes sorry should have removed.

package/setup.cfg
hmacdope: Yes idk why it's showing up in diff.

package/MDAnalysis/coordinates/DCD.py
hmacdope: @kain88-de i agree that this is a dangerously non-backwards compatible change. We can go with a deprecate warning here and then move to 3.0? What do people think @MDAnalysis/coredevs?

package/MDAnalysis/lib/formats/libdcd.pyx
orbeckst: Ok, I believe you. Was just wondering if you couldn't do something like
```python
def  int DCD_IS_CHARMM_       = 0x01

DCD_IS_CHARMM      =  DCD_IS_CHARMM_ 
```

package/MDAnalysis/coordinates/DCD.py
orbeckst: If it's a 30% performance increase and if it hasn't actually been documented then I am actually tending towards getting it fixed sooner than later. Maybe deprecate in 2.4 and change in 2.5 if there are strong feelings about its impact.

IAlibay: If it's likely to affect user behaviour then imho coordinates is one of the places we want to be the most cautious with. DCD is an unnaturally popular format, my 2 cents is that I'd prefer bump up 3.0's timeline than make a major break here.

package/setup.cfg
IAlibay: Might want to try rebasing, I guess we'll squash merge anyways, but it might get annoying for you if you get too out of synch :/

package/MDAnalysis/lib/formats/libdcd.pyx
hmacdope: I will investigate.

package/MDAnalysis/coordinates/DCD.py
hmacdope: They other readers don't set self.ts as ts as ts is modified inplace in `_frame_to_ts`

package/MDAnalysis/coordinates/DCD.py
hmacdope: I'll move to 3.0

richardjgowers: I think this has come up a few times, in general we don't specify one way or another if `Timestep` is the same but there's a definite performance improvement on sharing the same one.

package/MDAnalysis/lib/formats/libdcd.pyx
hmacdope: I know its a bit ugly but problem  is we also need an exportable version at the python level.

package/MDAnalysis/lib/formats/libdcd.pxd
```diff
@@ -0,0 +1,147 @@
+# -*- Mode: python; tab-width: 4; indent-tabs-mode:nil; coding:utf-8 -*-
```
richardjgowers, line 1: since this is a glorified `.h` file, would it make more sense to place it with the other one we have?

hmacdope: As in with the actual format stuff in `/src` or in `libmdanalysis`?

package/CHANGELOG
```diff
@@ -19,9 +19,14 @@ The rules for this file:
  * 2.4.0
 
 Fixes
+<<<<<<< HEAD
```
orbeckst: fix bad merge conflict in CHANGELOG

package/MDAnalysis/coordinates/DCD.py
```diff
@@ -188,13 +188,10 @@ def _read_next_timestep(self, ts=None):
         if self._frame == self.n_frames - 1:
             raise IOError('trying to go over trajectory limit')
         if ts is None:
-            # use a copy to avoid that ts always points to the same reference
-            # removing this breaks lammps reader
-            ts = self.ts.copy()
+            ts = self.ts
```
orbeckst: My understanding was that the majority of devs here did _not_ want to change the behavior of the ts being a copy but instead deprecate the current behavior and change it in 3.0, or did I misunderstand the discussion??

testsuite/MDAnalysisTests/analysis/test_pca.py
```diff
@@ -196,7 +196,7 @@ def test_project_compare_projections(u_fresh):
     project1 = pca.project_single_frame(1)
 
     u_fresh.trajectory[0]
-    coord0 = project0(u_fresh.trajectory.ts).positions
+    coord0 = project0(u_fresh.trajectory.ts).positions.copy()
```
orbeckst, line 199: Even if we do not change the ts.copy() in the DCDReader for now, I'd like to keep the copy() here for best practice demonstration and so that the tests could function with any reader and also for 3.0.

(Side note: it's confusing that `ts.positions` as a raw array behaves different from `ag.positions` as a view... I think the only (lame) excuse is that users aren't really supposed to play with `ts.positions` directly.)

testsuite/MDAnalysisTests/coordinates/base.py
```diff
@@ -443,6 +443,17 @@ def test_pickle_reader(self, reader):
         assert_equal(len(reader), len(reader_p))
         assert_equal(reader.ts, reader_p.ts,
                      "Timestep is changed after pickling")
+    
+    def test_frame_collect_all_same(self, reader):
```
orbeckst, line 447: Does this check that we are _not_ providing a copy of the ts?

I like the test. We just need to make an exception (or xfail) for DCDReader if we keep the current behavior until 3.0.

orbeckst (changes requested):
Overall, I am happy with the changes.

The main question is about `ts.copy()` in the DCDReader: I _thought_ the majority of devs wanted to keep current behavior until 3.0 and deprecate the copy-behavior. Could you please comment?

(Parenthetically, I'd say that rewriting in more cython reduces the readability but that's fine if performance increases significantly. I'm just a bit sad that the 30% increase in speed will have to wait for 3.0.)

package/MDAnalysis/coordinates/DCD.py
hmacdope: Ah sorry I had misinterpreted and thought this whole PR was shifted to 3.0, but what you propose makes a lot more sense.  Deprecate and change for 3.0.

testsuite/MDAnalysisTests/coordinates/base.py
hmacdope: Yep it checks that ts.positions is always the same.

testsuite/MDAnalysisTests/analysis/test_pca.py
hmacdope: Roger that.

package/MDAnalysis/coordinates/DCD.py
orbeckst: You, know, doing it for 3.0 would be a possibility. But I suggest we do most of the changes right now _except removing the copy_ for two reasons
1. We need to put in a deprecation note (docs and CHANGELOG). Better sooner than later.
2. If time-to-close-PR is a metric for CZI & friends then we rather close/merge PRs sooner ;-)

I don't know if we should issue a DeprecationWarning — do people have opinions? If yes, the I think it should be on creating a DCDReader _or_ in `Timestep.__init__` _if_ we only rarely create new Timestep instances. If we generate a new ts often as part of DCD reading (I can't remember...) then better only once in the DCDReader init. Open for ideas...

hmacdope: I have put it in DCDReader `init`

package/MDAnalysis/lib/formats/libdcd.pyx
```diff
@@ -590,14 +539,22 @@ cdef class DCDFile:
...
         if self.n_frames == 0:
             raise IOError("opened empty file. No frames are saved")
-
-        cdef np.ndarray xyz = np.empty((self.natoms, self.ndims), dtype=FLOAT, order='F')
-        cdef np.ndarray unitcell = np.empty(6, dtype=DOUBLE)
+        cdef np.npy_intp[2] dims 
+        dims[0] = self.natoms
+        dims[1] = self.ndims
+        # note use of fortran flag (1)
+        cdef np.ndarray[FLOAT_T, ndim=2] xyz = np.PyArray_EMPTY(2, dims, np.NPY_FLOAT32, 1)
```
richardjgowers: allocate this once per reader as a private scratch buffer

package/MDAnalysis/lib/formats/libdcd.pyx
```diff
@@ -660,58 +616,87 @@ cdef class DCDFile:
...
 
-        cdef np.ndarray[FLOAT_T, ndim=3] xyz = np.empty(shape, dtype=FLOAT)
-        cdef np.ndarray[DOUBLE_T, ndim=2] box = np.empty((n, 6))
-
-
-        cdef np.ndarray xyz_tmp = np.empty((self.natoms, self.ndims), dtype=FLOAT, order='F')
-        cdef int ok, i
-
-        if start == 0 and step == 1 and stop == self.n_frames:
+        cdef np.ndarray[FLOAT_T, ndim=3] xyz = np.PyArray_EMPTY(3, dims, np.NPY_FLOAT32, 0)
```
richardjgowers: can reuse the scratch buffer here?

package/MDAnalysis/lib/formats/libdcd.pyx
```diff
@@ -609,7 +566,6 @@ cdef class DCDFile:
         self.current_frame += 1
         return DCDFrame(xyz, unitcell)
 
-
     def readframes(self, start=None, stop=None, step=None, order='fac', indices=None):
```
richardjgowers, line 581: maybe deprecate this method for 3.0?  Or if we want to keep it, seems like we could handle this sort of behaviour in a parent class if we really thought it was useful

richardjgowers (changes requested):
@hmacdope can probably use a shared scratch buffer.  You can raise an issue about deprecating that timeseries helper method

package/MDAnalysis/lib/formats/libdcd.pyx
hmacdope: Issue raised #3930

package/CHANGELOG
```diff
@@ -19,6 +19,8 @@ The rules for this file:
  * 2.4.0
 
 Fixes
+  * DCD reader no longer copies timestep and instead updates it in line with other readers
```
orbeckst: update CHANGES: we are _deprecating_ the copy for 3.0

package/MDAnalysis/coordinates/DCD.py
```diff
@@ -130,6 +138,8 @@ def __init__(self, filename, convert_units=True, dt=None, **kwargs):
 
         .. versionchanged:: 0.17.0
            Changed to use libdcd.pyx library and removed the correl function
+        .. versionchanged:: 2.4.0
```
orbeckst: I don't think that adding a deprecation warning requires a versionchanged. The deprecated in the class docs is sufficient.

orbeckst (approved):
only minimal doc/changelog updates — please do them but I'll approve as to not hold up

package/MDAnalysis/lib/formats/libdcd.pyx
hmacdope: If we are deprecating, I might just leave it alone for now. Main reading methods use buffer now.

hmacdope:
> @hmacdope can probably use a shared scratch buffer. You can raise an issue about deprecating that timeseries helper method

Performance improvement of ~ 12% with use of shared scratch buffer 

image

package/MDAnalysis/lib/formats/libdcd.pyx
hmacdope: Done

hmacdope:
Windows C++ compiler seems to be complaining about libdcd, I'll revert it back to compiling as C sometime tommorow.

richardjgowers (approved):

hmacdope:
@MDAnalysis/coredevs  I think we are good to roll here?

Issue #2948, opened 2020-09-18
'PDBReader' object has no attribute 'timeseries'


sujigeo:
Hi,
I am using back-mapped trajectory from a coarse grain simulation. I want to evaluate the trajectory convergence in doing so i am using the following script 
ces_conv = encore.ces_convergence(u,  # universe
                                  10,  # window size
                                  select="name CA")  # default
I am getting the following error 
 'PDBReader' object has no attribute 'timeseries'

In the backmapped trajectory file has the frame no instead of the time. Further in the example at "  https://userguide.mdanalysis.org/1.0.0/examples/analysis/trajectory_similarity/convergence.html#Using-default-arguments-with-dimension-reduction-ensemble-similarity" wrongly uses selection="name CA" instead of select="name CA". Any help would be greatly appreciated

Thank you

orbeckst:
@sujigeo could you please show the code that you used to run your analysis (starting at when you make the Universe). The copy and paste the complete error output. Thank you.

hmacdope:
Fixed in #3890

PR #3890, opened 2022-10-28
Add timeseries for all readers

### LINKED ISSUE #2578: Can only run encore convergence on DCDReader/MemoryReader trajectories

### LINKED ISSUE #3761: Remove use of `trajectory.timeseries` from `ENCORE`
## Expected behavior ##

The `ENCORE` module should work with all kinds of readers.  As revealed by the use of the DUMPReader in [this](https://groups.google.com/g/mdnalysis-discussion/c/PEqfPk_aVvo ) mailing list question this is not the case.

## Actual behavior ##

The use of `trajectory.timestep` in `ENCORE` in many places means it cannot use readers that do not have the `timeseries` attribute. Currently only `DCDReader` and `MemoryReader` expose the `timeseries` attribute in the correct way to allow ENCORE to work.

``` python
import MDAnalysis as mda
from MDAnalysis.analysis import encore
from MDAnalysisTests.datafiles import LAMMPSDUMP
u = mda.Universe(LAMMPSDUMP, format='LAMMPSDUMP',atom_style="id type x y z")

es_conv = encore.ces_convergence(u,  50,  select='index 1') 

```

The current workaround is to load your trajectory into the `MemoryReader` using `AnalysisFromFunction`

```python
import MDAnalysis as mda
from MDAnalysis.coordinates.memory import MemoryReader
from MDAnalysis.analysis.base import AnalysisFromFunction
from MDAnalysisTests.datafiles import PDB, XTC
from MDAnalysis.analysis import encore

u = mda.Universe(PDB, XTC)
coordinates = AnalysisFromFunction(lambda ag: ag.positions.copy(), u.atoms).run().results
u2 = mda.Universe(PDB, coordinates['timeseries'], format=MemoryReader)
es_conv = encore.ces_convergence(u2,  50, select='index 1')
```

## Solution ##

We should change to using `AtomGroup.positions`.

### LINKED ISSUE #3881: Revive  having `timeseries` for all readers and /or having a `transfer_to_memory` function at the reader level.

MDAnalysis/mdanalysis#2271 and MDAnalysis/mdanalysis#1400 first raised the idea of having a timeseries attribute for all readers. Currently timeseries is available  only for `DCDReader` and `MemoryReader` causing all kinds of problems particularly with ENCORE. See MDAnalysis/mdanalysis#3761  MDAnalysis/mdaencore#30 MDAnalysis/mdanalysis#2948 MDAnalysis/mdanalysis#2578 and others. 
 

IMO the simplest way to do this is with a method attached to `ProtoReader` that returns the full in-memory representation of the trajectory. This can then be overridden if there is a faster way, eg with a Cythonised reader etc. 

This can also then tie in with the memory-reader nicely if possible. I am no memory-reader expert so would love some guidance on how this might tie in @richardjgowers.

hmacdope:
Fixes MDAnalysis/mdanalysis#3881 MDAnalysis/mdanalysis#2271 MDAnalysis/mdanalysis#3898 MDAnalysis/mdanalysis#2948 MDAnalysis/mdaencore#30

 - Add timeseries attribute to all readers as a method of `ReaderBase`
 - Make MemoryReader raise a NoDataError on empty atomgroup. 

@MDAnalysis/coredevs one issue to be discussed. The MemoryReader has **inclusive** start, stop, step which is not in line with python `range` while this has PR been set up to have non-inclusive start-stop-step. Which would we prefer? One of the two implementations will need to change for consistency. 

PR Checklist
------------
 - [X] Tests?
 - [X] Docs?
 - [X] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -19,8 +19,9 @@ The rules for this file:
  * 2.4.0
 
 Fixes
-  * NoDataErrors raised on empty atomgroup provided to timeseries in DCDReader and MemoryReader
-    were changed to ValueError as NoDataError is intended for TopologyAttrs (cf #3901)
+  * NoDataError raised on empty atomgroup provided to `DCDReader.timeseries` 
+    was changed to ValueError (see #3901). A matching ValueError was added to
+    `MemoryReader.timeseries`(PR #3890). 
   * LAMMPSDump Reader translates the box to the origin (#3741)
   * hbond analysis: access hbonds results through new results member in count_by_ids() and count_by_type()
   * Added isolayer selection method (Issue #3845)
@@ -33,8 +34,8 @@ Fixes
     (e.g. bonds, angles) (PR #3779).
 
 Enhancements
-  * The timeseries method for dumping coordinates to a NumPy array was added
-    to ProtoReader (PR #3890)
+  * The timeseries method for exporting coordinates from multiple frames to a
+    NumPy array was added to ProtoReader (PR #3890)
   * LAMMPSDump Reader optionally unwraps trajectories with image flags upon 
     loading (#3843
   * LAMMPSDump Reader now imports velocities and forces (#3843)
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -33,7 +33,8 @@ Fixes
     (e.g. bonds, angles) (PR #3779).
 
 Enhancements
-  * The timeseries method for dumping coordinates to a NumPy array was added to ProtoReader
+  * The timeseries method for dumping coordinates to a NumPy array was added
+    to ProtoReader (PR #3890)
   * LAMMPSDump Reader optionally unwraps trajectories with image flags upon 
     loading (#3843
   * LAMMPSDump Reader now imports velocities and forces (#3843)
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -19,6 +19,8 @@ The rules for this file:
  * 2.4.0
 
 Fixes
+  * NoDataErrors raised on empty atomgroup provided to timeseries in DCDReader and MemoryReader
+    were changed to ValueError as NoDataError is intended for TopologyAttrs (cf #3901)
   * LAMMPSDump Reader translates the box to the origin (#3741)
   * hbond analysis: access hbonds results through new results member in count_by_ids() and count_by_type()
   * Added isolayer selection method (Issue #3845)
@@ -31,6 +33,7 @@ Fixes
     (e.g. bonds, angles) (PR #3779).
 
 Enhancements
+  * The timeseries method for dumping coordinates to a NumPy array was added to ProtoReader
   * LAMMPSDump Reader optionally unwraps trajectories with image flags upon 
     loading (#3843
   * LAMMPSDump Reader now imports velocities and forces (#3843)
--- a/package/MDAnalysis/coordinates/base.py
+++ b/package/MDAnalysis/coordinates/base.py
@@ -1039,12 +1039,13 @@ class ProtoReader(IOBase, metaclass=_Readermeta):
         if order != default_order:
             try:
                 newidx = [default_order.index(i) for i in order]
+                print(newidx)
             except ValueError:
                 raise ValueError(f"Unrecognized order key in {order}, "
                                  "must be permutation of 'fac'")
 
             try:
-                coordinates = np.moveaxis(coordinates, [0, 1, 2], newidx)
+                coordinates = np.moveaxis(coordinates, newidx, [0, 1, 2])
             except ValueError:
                 errmsg = ("Repeated or missing keys passed to argument "
                           f"`order`: {order}, each key must be used once")
--- a/package/MDAnalysis/coordinates/base.py
+++ b/package/MDAnalysis/coordinates/base.py
@@ -986,6 +986,7 @@ class ProtoReader(IOBase, metaclass=_Readermeta):
                    step: Optional[int]=None,
                    order: Optional[str]='fac') -> np.ndarray:
         """Return a subset of coordinate data for an AtomGroup
+        
         Parameters
         ----------
         asel : AtomGroup (optional)
@@ -993,15 +994,15 @@ class ProtoReader(IOBase, metaclass=_Readermeta):
             coordinates from. Defaults to ``None``, in which case the full set of
             coordinate data is returned.
         start :  int (optional)
-             Begin reading the trajectory at frame index `start` (where 0 is the index
-             of the first frame in the trajectory); the default ``None`` starts
-             at the beginning.
+            Begin reading the trajectory at frame index `start` (where 0 is the index
+            of the first frame in the trajectory); the default ``None`` starts
+            at the beginning.
         stop : int (optional)
-             End reading the trajectory at frame index `stop`-1, i.e, `stop` is excluded.
-             The trajectory is read to the end with the default ``None``.
+            End reading the trajectory at frame index `stop`-1, i.e, `stop` is excluded.
+            The trajectory is read to the end with the default ``None``.
         step : int (optional)
-             Step size for reading; the default ``None`` is equivalent to 1 and means to
-             read every frame.
+            Step size for reading; the default ``None`` is equivalent to 1 and means to
+            read every frame.
         order : str (optional)
             the order/shape of the return data array, corresponding
             to (a)tom, (f)rame, (c)oordinates all six combinations
--- a/package/MDAnalysis/coordinates/base.py
+++ b/package/MDAnalysis/coordinates/base.py
@@ -131,7 +131,6 @@ from typing import Union, Optional, List, Dict
 
 from .timestep import Timestep
 from . import core
-from .. import NoDataError
 from .. import (
     _READERS, _READER_HINTS,
     _SINGLEFRAME_WRITERS,
@@ -1023,7 +1022,7 @@ class ProtoReader(IOBase, metaclass=_Readermeta):
 
         if asel is not None:
             if len(asel) == 0:
-                raise NoDataError(
+                raise ValueError(
                     "Timeseries requires at least one atom to analyze")
             atom_numbers = asel.indices
             natoms = len(atom_numbers)
--- a/package/MDAnalysis/coordinates/base.py
+++ b/package/MDAnalysis/coordinates/base.py
@@ -1029,7 +1029,7 @@ class ProtoReader(IOBase, metaclass=_Readermeta):
             natoms = self.n_atoms
             atom_numbers = np.arange(natoms)
 
-        # allocate output array in 'fac' (_default_timeseries_order) order
+        # allocate output array in 'fac' order
         coordinates = np.empty((nframes, natoms, 3), dtype=np.float32)
         for i, ts in enumerate(self[start:stop:step]):
             coordinates[i, :] = ts.positions[atom_numbers]
@@ -1050,7 +1050,6 @@ class ProtoReader(IOBase, metaclass=_Readermeta):
                           f"`order`: {order}, each key must be used once")
                 raise ValueError(errmsg)
         return coordinates
-
 
 # TODO: Change order of aux_spec and auxdata for 3.0 release, cf. Issue #3811
     def add_auxiliary(self,
--- a/package/MDAnalysis/coordinates/base.py
+++ b/package/MDAnalysis/coordinates/base.py
@@ -986,23 +986,24 @@ class ProtoReader(IOBase, metaclass=_Readermeta):
                    step: Optional[int]=None,
                    order: Optional[str]='fac') -> np.ndarray:
         """Return a subset of coordinate data for an AtomGroup
-        
+
         Parameters
         ----------
         asel : AtomGroup (optional)
             The :class:`~MDAnalysis.core.groups.AtomGroup` to read the
-            coordinates from. Defaults to ``None``, in which case the full set of
-            coordinate data is returned.
+            coordinates from. Defaults to ``None``, in which case the full set
+            of coordinate data is returned.
         start :  int (optional)
-            Begin reading the trajectory at frame index `start` (where 0 is the index
-            of the first frame in the trajectory); the default ``None`` starts
-            at the beginning.
+            Begin reading the trajectory at frame index `start` (where 0 is the
+            index of the first frame in the trajectory); the default
+            ``None`` starts at the beginning.
         stop : int (optional)
-            End reading the trajectory at frame index `stop`-1, i.e, `stop` is excluded.
-            The trajectory is read to the end with the default ``None``.
+            End reading the trajectory at frame index `stop`-1, i.e, `stop` is
+            excluded. The trajectory is read to the end with the default
+            ``None``.
         step : int (optional)
-            Step size for reading; the default ``None`` is equivalent to 1 and means to
-            read every frame.
+            Step size for reading; the default ``None`` is equivalent to 1 and
+            means to read every frame.
         order : str (optional)
             the order/shape of the return data array, corresponding
             to (a)tom, (f)rame, (c)oordinates all six combinations
--- a/package/MDAnalysis/coordinates/base.py
+++ b/package/MDAnalysis/coordinates/base.py
@@ -1039,7 +1039,6 @@ class ProtoReader(IOBase, metaclass=_Readermeta):
         if order != default_order:
             try:
                 newidx = [default_order.index(i) for i in order]
-                print(newidx)
             except ValueError:
                 raise ValueError(f"Unrecognized order key in {order}, "
                                  "must be permutation of 'fac'")
--- a/package/MDAnalysis/coordinates/base.py
+++ b/package/MDAnalysis/coordinates/base.py
@@ -1018,6 +1018,9 @@ class ProtoReader(IOBase, metaclass=_Readermeta):
         start, stop, step = self.check_slice_indices(start, stop, step)
         nframes = len(range(start, stop, step))
 
+        if sorted(order) != sorted(self._default_timeseries_order):
+            raise ValueError("The order of the returned coordinate array must"
+                            " be a permutation of `fac`.")
         if asel is not None:
             if len(asel) == 0:
                 raise NoDataError(
@@ -1049,8 +1052,7 @@ class ProtoReader(IOBase, metaclass=_Readermeta):
             coordinates = np.swapaxes(coordinates, 2, 0)
             coordinates = np.swapaxes(coordinates, 1, 2)
         else:
-            raise ValueError("The order of the returned coordinate array must"
-                            " be a permutation of `fac`.")
+            pass
         return coordinates
 
 
--- a/package/MDAnalysis/coordinates/base.py
+++ b/package/MDAnalysis/coordinates/base.py
@@ -664,8 +664,6 @@ class ProtoReader(IOBase, metaclass=_Readermeta):
     #: The appropriate Timestep class, e.g.
     #: :class:`MDAnalysis.coordinates.xdrfile.XTC.Timestep` for XTC.
     _Timestep = Timestep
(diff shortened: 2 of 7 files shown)


orbeckst:
Let’s be consistent with Python range.

hmacdope:
> Let’s be consistent with Python range.

@orbeckst to be clear this will mean making API break for memoryreader. Are we happy with that?

package/MDAnalysis/coordinates/base.py
```diff
@@ -980,6 +982,79 @@ def __repr__(self):
...
+        order : str (optional)
+            the order/shape of the return data array, corresponding
+            to (a)tom, (f)rame, (c)oordinates all six combinations
+            of 'a', 'f', 'c' are allowed ie "fac" - return array
+            where the shape is (frame, number of atoms,
+            coordinates)
+        See Also
+        --------
+        MDAnalysis.coordinates.memory
+        .. versionadded:: 2.4.0
```
IAlibay:
```suggestion

        See Also
        --------
        :class:`MDAnalysis.coordinates.memory`


        .. versionadded:: 2.4.0
```
Should fix docs rendering

package/MDAnalysis/coordinates/base.py
```diff
@@ -980,6 +982,79 @@ def __repr__(self):
...
+        """
+        start, stop, step = self.check_slice_indices(start, stop, step)
+        nframes = len(range(start, stop, step))
+
+        if sorted(order) != sorted(self._default_timeseries_order):
+            raise ValueError("The order of the returned coordinate array must"
+                            " be a permutation of `fac`.")
```
IAlibay:
```suggestion
```
Can use the moveaxis behaviour instead of directly testing here (see suggestion below).

package/MDAnalysis/coordinates/base.py
```diff
@@ -664,6 +664,8 @@ class ProtoReader(IOBase, metaclass=_Readermeta):
     #: The appropriate Timestep class, e.g.
     #: :class:`MDAnalysis.coordinates.xdrfile.XTC.Timestep` for XTC.
     _Timestep = Timestep
+    # the default coordinate layout for timeseries
+    _default_timeseries_order = 'fac'
```
IAlibay: why make this a class attribute if it's only used once?
```suggestion
```

package/MDAnalysis/coordinates/base.py
```diff
@@ -980,6 +982,79 @@ def __repr__(self):
...
+        elif order[2] == self._default_timeseries_order[2]:
+            coordinates = np.swapaxes(coordinates, 0, 1)
+        elif order[0] == self._default_timeseries_order[1]:
+            coordinates = np.swapaxes(coordinates, 1, 0)
+            coordinates = np.swapaxes(coordinates, 1, 2)
+        elif order[0] == self._default_timeseries_order[2]:
+            coordinates = np.swapaxes(coordinates, 2, 0)
+            coordinates = np.swapaxes(coordinates, 1, 2)
+        else:
+            pass
```
IAlibay:
```suggestion
        default_order = 'fac'
        if order != default_order:
            try:
                newidx = [default_order.index(i) for i in order]
            except ValueError:
                raise ValueError(f"Unrecognized order key in {order}, "
                                 "must be permutation of 'fac'")

            try:
                coordinates = np.moveaxis(coordinates, [0, 1, 2], newidx)
            except ValueError:
                errmsg = ("Repeated or missing keys passed to argument "
                          f"`order`: {order}, each key must be used once")
                raise ValueError(errmsg)
```
That should take care of order checking + also avoids the whole if/else tree.

testsuite/MDAnalysisTests/coordinates/base.py
```diff
@@ -21,6 +21,7 @@
 # J. Comput. Chem. 32 (2011), 2319--2327, doi:10.1002/jcc.21787
 #
 import itertools
+from os import terminal_size
```
IAlibay:
```suggestion
```
not sure why this is here?

testsuite/MDAnalysisTests/coordinates/base.py
```diff
@@ -445,6 +446,29 @@ def test_pickle_reader(self, reader):
         assert_equal(reader.ts, reader_p.ts,
                      "Timestep is changed after pickling")
 
+    
+    @pytest.mark.parametrize('order', ('fac', 'fca', 'afc', 'acf', 'caf', 'cfa'))
+    def test_timeseries_shape(self, reader, order):
+        timeseries = reader.timeseries(order=order)
+        a_index = order.find('a')
+        f_index = order.find('f')
+        c_index = order.find('c')
```
IAlibay:
```suggestion
        a_index = order.index('a')
        f_index = order.index('f')
        c_index = order.index('c')
```
It's a trivial thing, but I'd recommend index over find because the latter returns -1 on failure instead of a ValueError, which can lead to some unintended behaviour.

testsuite/MDAnalysisTests/coordinates/base.py
```diff
@@ -445,6 +446,29 @@ def test_pickle_reader(self, reader):
...
+        assert(timeseries.shape[a_index] == reader.n_atoms)
+        assert(timeseries.shape[f_index] == len(reader))
+        assert(timeseries.shape[c_index] == 3)
+
+    @pytest.mark.parametrize('slice', ([0,2,1], [0,10,2], [0,10,3]))
+    def test_timeseries_values(self, reader, slice):
+        ts_positions = []
+        if slice[1] > len(reader):
+            pytest.skip()
```
IAlibay:
```suggestion
            pytest.skip("too few frames in reader")
```

testsuite/MDAnalysisTests/coordinates/base.py
```diff
@@ -445,6 +446,29 @@ def test_pickle_reader(self, reader):
...
+        for i in range(slice[0], slice[1], slice[2]):
+            ts = reader[i]
+            ts_positions.append(ts.positions.copy())
+        positions = np.asarray(ts_positions)
+        timeseries = reader.timeseries(start=slice[0], stop=slice[1], step=slice[2], order='fac')
+        np.testing.assert_allclose(timeseries, positions)
```
IAlibay:
```suggestion
        assert_allclose(timeseries, positions)
```
already imported

testsuite/MDAnalysisTests/coordinates/base.py
```diff
@@ -445,6 +446,29 @@ def test_pickle_reader(self, reader):
...
+    @pytest.mark.parametrize('slice', ([0,2,1], [0,10,2], [0,10,3]))
+    def test_timeseries_values(self, reader, slice):
+        ts_positions = []
+        if slice[1] > len(reader):
+            pytest.skip()
+        for i in range(slice[0], slice[1], slice[2]):
+            ts = reader[i]
+            ts_positions.append(ts.positions.copy())
+        positions = np.asarray(ts_positions)
+        timeseries = reader.timeseries(start=slice[0], stop=slice[1], step=slice[2], order='fac')
```
IAlibay:
```suggestion
        timeseries = reader.timeseries(start=slice[0], stop=slice[1],
                                       step=slice[2], order='fac')
```
>79 char fix

testsuite/MDAnalysisTests/coordinates/test_reader_api.py
```diff
@@ -133,6 +133,10 @@ def test_raises_StopIteration(self, reader):
         with pytest.raises(StopIteration):
             next(reader)
 
+    @pytest.mark.parametrize('order', ['turnip', 'abc', ''])
+    def test_timeseries_raises_incorrect_order(self, reader, order):
+        with pytest.raises(ValueError, match="must be a permutation of `fac`"):
+            reader.timeseries(order=order)
```
IAlibay:
```suggestion
    @pytest.mark.parametrize('order', ['turnip', 'abc'])
    def test_timeseries_raises_unknown_order_key(self, reader, order):
        with pytest.raises(ValueError, match="Unrecognized order key"):
            reader.timeseries(order=order)

    @pytest.mark.parametrize('order', ['faac', 'affc', 'afcc', ''])
    def test_timeseries_raises_incorrect_order_key(self, reader, order):
        with pytest.raises(ValueError, match="Repeated or missing keys"):
            reader.timeseries(order=order)
```

IAlibay (changes requested):
mostly reformatting to avoid the if/else tree

IAlibay:
I'd be accepting of calling the MemoryReader weird indexing a bug rather than a feature, unless it's documented as inclusive. If it is documented to be a given way, then we'll need to do the whole deprecation warning + allow for both options specifically for the MemoryReader.

orbeckst:
Unfortunately, MemoryReader.timeseries is documented as start and stop as inclusive https://docs.mdanalysis.org/stable/documentation_pages/coordinates/memory.html?highlight=memoryreader#MDAnalysis.coordinates.memory.MemoryReader.timeseries , this is probably a remnant of the very earliest versions of timeseries that was linked closely to the old DCDReader (even though DCDReader.timeseries now contains exclusive stop). 

Let’s deprecate the behavior for MemoryReader and change in 3.0. Make everything else consistent and range-like.

orbeckst (changes requested):
Let’s deprecate the behavior for MemoryReader and change in 3.0. Make everything else consistent and range-like.

hmacdope:
I will PR deprecating the indexing of the memory reader separately.

hmacdope:
> Let’s deprecate the behavior for MemoryReader and change in 3.0. Make everything else consistent and range-like.

Issue raised MDAnalysis/mdanalysis#3893

package/MDAnalysis/coordinates/base.py
```diff
@@ -980,6 +980,77 @@ def __repr__(self):
...
+                                 "must be permutation of 'fac'")
+
+            try:
+                coordinates = np.moveaxis(coordinates, newidx, [0, 1, 2])
+            except ValueError:
+                errmsg = ("Repeated or missing keys passed to argument "
+                          f"`order`: {order}, each key must be used once")
+                raise ValueError(errmsg)
+        return coordinates
+
```
IAlibay, line 1054:
```suggestion
```
next thing along is another class method right? should only be one empty line if so

testsuite/MDAnalysisTests/coordinates/base.py
```diff
@@ -445,6 +445,33 @@ def test_pickle_reader(self, reader):
         assert_equal(reader.ts, reader_p.ts,
                      "Timestep is changed after pickling")
 
+
```
IAlibay:
```suggestion
```
As above, one empty line between class methods

testsuite/MDAnalysisTests/coordinates/base.py
```diff
@@ -445,6 +445,33 @@ def test_pickle_reader(self, reader):
...
+        if slice[1] > len(reader):
+            pytest.skip("too few frames in reader")
+        for i in range(slice[0], slice[1], slice[2]):
+            ts = reader[i]
+            ts_positions.append(ts.positions.copy())
+        positions = np.asarray(ts_positions)
+        timeseries = reader.timeseries(start=slice[0], stop=slice[1],
+                                       step=slice[2], order='fac')
+        assert_allclose(timeseries, positions)
+
```
IAlibay, line 472:
```suggestion
```

testsuite/MDAnalysisTests/coordinates/base.py
```diff
@@ -445,6 +445,33 @@ def test_pickle_reader(self, reader):
...
+        c_index = order.index('c')
+        assert(timeseries.shape[a_index] == reader.n_atoms)
+        assert(timeseries.shape[f_index] == len(reader))
+        assert(timeseries.shape[c_index] == 3)
+
+    @pytest.mark.parametrize('slice', ([0,2,1], [0,10,2], [0,10,3]))
+    def test_timeseries_values(self, reader, slice):
+        ts_positions = []
+        if isinstance(reader, mda.coordinates.memory.MemoryReader):
+            pytest.xfail("MemoryReader uses deprecated stop inclusive"
```
IAlibay, line 461: Raise an issue and tie it to the 3.0 milestone?

package/MDAnalysis/coordinates/base.py
```diff
@@ -980,6 +980,77 @@ def __repr__(self):
...
+            if len(asel) == 0:
+                raise NoDataError(
+                    "Timeseries requires at least one atom to analyze")
+            atom_numbers = asel.indices
+            natoms = len(atom_numbers)
+        else:
+            natoms = self.n_atoms
+            atom_numbers = np.arange(natoms)
+
+        # allocate output array in 'fac' (_default_timeseries_order) order
```
IAlibay:
```suggestion
        # allocate output array in 'fac' order
```
Since the global attribute is no longer there

IAlibay (changes requested):
couple of small things

testsuite/MDAnalysisTests/coordinates/base.py
hmacdope: happy with #3893 to cover it?

hmacdope:
I have added an exception on empty atomgroup for `MemoryReader.timeseries(asel)` in line with the DCDReader. Feel free to discuss on MDAnalysis/mdanalysis#3898.

package/MDAnalysis/coordinates/base.py
```diff
@@ -980,6 +980,78 @@ def __repr__(self):
...
+        """
+        start, stop, step = self.check_slice_indices(start, stop, step)
+        nframes = len(range(start, stop, step))
+
+        if asel is not None:
+            if len(asel) == 0:
+                raise NoDataError(
+                    "Timeseries requires at least one atom to analyze")
+            atom_numbers = asel.indices
+            natoms = len(atom_numbers)
```
IAlibay, line 1028: add tests?

package/MDAnalysis/coordinates/base.py
```diff
@@ -980,6 +980,78 @@ def __repr__(self):
...
+
+
+        .. versionadded:: 2.4.0
+        """
+        start, stop, step = self.check_slice_indices(start, stop, step)
+        nframes = len(range(start, stop, step))
+
+        if asel is not None:
+            if len(asel) == 0:
+                raise NoDataError(
```
IAlibay: As per #3898, I believe both @orbeckst and @richardjgowers were keen to enforce that `NoDataError` was a topology attrs thing rather than zero width atomgroups (fwiw I also thought it was meant for these cases, but I'm happy going with the alternative).

Would we be ok just raising a ValueError here and then switching the DCDReader to ValueError? I don't necessarily think we need to semver this change, I doubt anyone is specifically looking for this specific edge case anywhere?

package/MDAnalysis/coordinates/memory.py
```diff
@@ -559,6 +560,9 @@ def timeseries(self, asel=None, start=0, stop=-1, step=1, order='afc'):
         if (asel is None or asel is asel.universe.atoms):
             return array
         else:
+            if len(asel) == 0:
+                raise NoDataError("Timeseries requires at least one atom "
```
IAlibay: Same discussion as above.

IAlibay (changes requested):
Just temp blocking as a follow up from #3898, the folks that have the stronger stance on the use of `NoDataError` probably will have strong views.

package/MDAnalysis/coordinates/base.py
orbeckst: I agree with making it `ValueError`.

package/MDAnalysis/coordinates/memory.py
orbeckst: Make it `ValueError`.

orbeckst: You'd still want to document it with a versionchanged and list it under fixes in CHANGELOG (we're fixing the meaning) — or listing it under Changes and pointing out that NoDataError was derived from ValueError??

hmacdope: Thanks all, I will change ASAP, travelling this weekend

package/CHANGELOG
```diff
@@ -19,6 +19,8 @@ The rules for this file:
  * 2.4.0
 
 Fixes
+  * NoDataErrors raised on empty atomgroup provided to timeseries in DCDReader and MemoryReader
```
IAlibay: Is this correct, I thought that MemoryReader just didn't raise anything before?

package/CHANGELOG
```diff
@@ -31,6 +33,7 @@ Fixes
     (e.g. bonds, angles) (PR #3779).
 
 Enhancements
+  * The timeseries method for dumping coordinates to a NumPy array was added to ProtoReader
```
```suggestion
  * The timeseries method for dumping coordinates to a NumPy array was added
    to ProtoReader (PR #3890)
```
PR number for release sanity + 79 char limit

package/MDAnalysis/coordinates/base.py
```diff
@@ -980,6 +979,78 @@ def __repr__(self):
...
+    
+    def timeseries(self, asel: Optional['AtomGroup']=None,
+                   start: Optional[int]=None, stop: Optional[int]=None,
+                   step: Optional[int]=None,
+                   order: Optional[str]='fac') -> np.ndarray:
+        """Return a subset of coordinate data for an AtomGroup
+
+        Parameters
+        ----------
+        asel : AtomGroup (optional)
```
IAlibay, line 991: I know we're just doing the same as what memoryreader does, but it just struck me that `asel` is such a weird parameter name for an atomgroup. Not much we can do here, but if I'm not the only one maybe we can consider changing this for 3.0 (in a different PR, let's get this merged!).

IAlibay (approved):
Just two changelog things - I'm going to approve so that this can be merged once those comments are addressed.

package/MDAnalysis/coordinates/base.py
hmacdope: I'll raise an issue

hmacdope: Raised in #3911

package/CHANGELOG
hmacdope: You are right, will change.

hmacdope:
@orbeckst this should be good to roll if you have a second to review? :smiley_cat:

orbeckst (approved):
lgtm

PR #3892, opened 2022-10-28
Improve xdr cython

### LINKED ISSUE #3891: Improve XDR typing and performance
Similar to work on the DCD reader in #3882  The XDR readers could do with a little bit of love and possibly some better typing on the cython side.

hmacdope:
Fixes #3891 

 - Improves cython in libmdaxdr

PR Checklist
------------
 - [X] Tests?
 - [X] Docs?
 - [X] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -34,6 +34,10 @@ Fixes
     (e.g. bonds, angles) (PR #3779).
 
 Enhancements
+  * Improve C content of libxdr Cython, add `read_direct` methods to read 
+    coordinates, velocities and forces directly into memoryviews of `Timestep`
+    attributes, make `TRR` timestep have positions, velocities and forces on
+    `__init__`. (Issue #3891 PR  #3892).
   * The timeseries method for exporting coordinates from multiple frames to a
     NumPy array was added to ProtoReader (PR #3890)
   * MDAnalysis now officially supports py3.11 (Issue #3878)
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -14,14 +14,17 @@ The rules for this file:
 
 ------------------------------------------------------------------------------
 ??/??/?? IAlibay, Luthaf, hmacdope, rafaelpap, jbarnoud, BFedder, aya9aladdin,
-         jaclark5, jfennick
+         jaclark5, jfennick, lilyminium
 
  * 2.4.0
 
 Fixes
+  * NoDataError raised on empty atomgroup provided to `DCDReader.timeseries` 
+    was changed to ValueError (see #3901). A matching ValueError was added to
+    `MemoryReader.timeseries`(PR #3890). 
+  * Removes ``pbc`` kwarg from ``AtomGroup.wrap`` (Issue #3909)
   * LAMMPSDump Reader translates the box to the origin (#3741)
   * hbond analysis: access hbonds results through new results member in count_by_ids() and count_by_type()
-  * Added isolayer selection method (Issue #3845)
   * Auxiliary; determination of representative frames: Removed undefined
     behaviour for cutoff values < -1 (PR # 3749)
   * Fixes distances.between not always returning AtomGroup (Issue #3794)
@@ -31,6 +34,9 @@ Fixes
     (e.g. bonds, angles) (PR #3779).
 
 Enhancements
+  * The timeseries method for exporting coordinates from multiple frames to a
+    NumPy array was added to ProtoReader (PR #3890)
+  * MDAnalysis now officially supports py3.11 (Issue #3878)
   * LAMMPSDump Reader optionally unwraps trajectories with image flags upon 
     loading (#3843
   * LAMMPSDump Reader now imports velocities and forces (#3843)
@@ -45,8 +51,10 @@ Enhancements
   * A new AuxReader for the GROMACS EDR file format was implemented.
     (Issue # 3629, PR # 3749)
   * Added benchmarks for bonds topology attribute (Issue #3785, PR #3806)
+  * Added isolayer selection method (Issue #3845)
 
 Changes
+ * Moved `libmdanalysis` cython header to `lib`(Issue #3912, PR #3913)
  * Auxiliary; determination of representative frames: The default value for
    cutoff is now None, not -1. Negative values are now turned
    to None. (PR #3749)
@@ -62,6 +70,8 @@ Changes
   * adding element attribute to TXYZParser if all atom names are valid element symbols (PR #3826)
 
 Deprecations
+  * Add deprecation warning for inclusive stop indexing in
+    MemoryReader.timeseries (#PR 3894)
 
 
 08/29/22 IAlibay, PicoCentauri, orbeckst, hmacdope, rmeli, miss77jun, rzhao271,
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -19,8 +19,12 @@ The rules for this file:
  * 2.4.0
 
 Fixes
+<<<<<<< HEAD
   * XTC and TRR readers now fail with IOError when a status except EOK (=0) is
     detected on reading a frame instead of silently continuing
+=======
+  * Consolidate license files across MDAnalysis library packages (PR #3939)
+>>>>>>> 9e1953359 (Consolidate licenses across top directory, testsuite and package (#3939))
   * Kwargs passed through to `MemoryReader` when using `transfer_to_memory()`
     and `in_memory=True`.
   * NoDataError raised on empty atomgroup provided to `DCDReader.timeseries` 
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -19,12 +19,9 @@ The rules for this file:
  * 2.4.0
 
 Fixes
-<<<<<<< HEAD
   * XTC and TRR readers now fail with IOError when a status except EOK (=0) is
     detected on reading a frame instead of silently continuing
-=======
   * Consolidate license files across MDAnalysis library packages (PR #3939)
->>>>>>> 9e1953359 (Consolidate licenses across top directory, testsuite and package (#3939))
   * Kwargs passed through to `MemoryReader` when using `transfer_to_memory()`
     and `in_memory=True`.
   * NoDataError raised on empty atomgroup provided to `DCDReader.timeseries` 
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -21,6 +21,8 @@ The rules for this file:
 Fixes
   * XTC and TRR readers now fail with IOError when a status except EOK (=0) is
     detected on reading a frame instead of silently continuing
+  * Kwargs passed through to `MemoryReader` when using `transfer_to_memory()`
+    and `in_memory=True`.
   * NoDataError raised on empty atomgroup provided to `DCDReader.timeseries` 
     was changed to ValueError (see #3901). A matching ValueError was added to
     `MemoryReader.timeseries`(PR #3890). 
@@ -40,6 +42,7 @@ Enhancements
     coordinates, velocities and forces directly into memoryviews of `Timestep`
     attributes, make `TRR` timestep have positions, velocities and forces on
     `__init__`. (Issue #3891 PR  #3892).
+  * Improve C content of libdcd Cython (Issue #3882, PR #3888)
   * The timeseries method for exporting coordinates from multiple frames to a
     NumPy array was added to ProtoReader (PR #3890)
   * MDAnalysis now officially supports py3.11 (Issue #3878)
@@ -76,6 +79,8 @@ Changes
   * adding element attribute to TXYZParser if all atom names are valid element symbols (PR #3826)
 
 Deprecations
+  * Add deprecation warning for `timestep` copying in DCDReader
+    (Issue #3889, PR #3888)
   * Add deprecation warning for inclusive stop indexing in
     MemoryReader.timeseries (#PR 3894)
 
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -19,6 +19,8 @@ The rules for this file:
  * 2.4.0
 
 Fixes
+  * XTC and TRR readers now fail with IOError when a status except EOK (=0) is
+    detected on reading a frame instead of silently continuing
   * NoDataError raised on empty atomgroup provided to `DCDReader.timeseries` 
     was changed to ValueError (see #3901). A matching ValueError was added to
     `MemoryReader.timeseries`(PR #3890). 
--- a/package/MDAnalysis/coordinates/TRR.py
+++ b/package/MDAnalysis/coordinates/TRR.py
@@ -158,6 +158,8 @@ class TRRReader(XDRBaseReader):
             raise IOError(errno.EIO, 'trying to go over trajectory limit')
         if ts is None:
             ts = self.ts
+        # allocate arrays to read into, will set to proper values
+        # in _frame_to_ts
         ts.has_positions = True
         ts.has_velocities = True
         ts.has_forces = True
--- a/package/MDAnalysis/coordinates/TRR.py
+++ b/package/MDAnalysis/coordinates/TRR.py
@@ -163,7 +163,7 @@ class TRRReader(XDRBaseReader):
         ts.has_positions = True
         ts.has_velocities = True
         ts.has_forces = True
-        frame = self._xdr.read_direct_pvf(ts.positions, ts.velocities, ts.forces)
+        frame = self._xdr.read_direct_xvf(ts.positions, ts.velocities, ts.forces)
         self._frame += 1
         self._frame_to_ts(frame, ts)
         return ts
--- a/package/MDAnalysis/coordinates/TRR.py
+++ b/package/MDAnalysis/coordinates/TRR.py
@@ -156,8 +156,9 @@ class TRRReader(XDRBaseReader):
         """copy next frame into timestep
         
         versionadded:: 2.4.0
-            TRRReader now implements this method so that it can use full
-            `xvf` read from file. 
+            TRRReader implements this method so that it can use
+            read_direct_xvf to read the data directly into the timestep
+            rather than copying it from a temporary array.
         """
         if self._frame == self.n_frames - 1:
             raise IOError(errno.EIO, 'trying to go over trajectory limit')
--- a/package/MDAnalysis/coordinates/TRR.py
+++ b/package/MDAnalysis/coordinates/TRR.py
@@ -153,7 +153,12 @@ class TRRReader(XDRBaseReader):
     _file = TRRFile
 
     def _read_next_timestep(self, ts=None):
-        """copy next frame into timestep"""
+        """copy next frame into timestep
+        
+        versionadded:: 2.4.0
+            TRRReader now implements this method so that it can use full
+            `xvf` read from file. 
+        """
         if self._frame == self.n_frames - 1:
             raise IOError(errno.EIO, 'trying to go over trajectory limit')
         if ts is None:
--- a/package/MDAnalysis/coordinates/TRR.py
+++ b/package/MDAnalysis/coordinates/TRR.py
@@ -31,6 +31,7 @@ See Also
 MDAnalysis.coordinates.XTC: Read and write GROMACS XTC trajectory files.
 MDAnalysis.coordinates.XDR: BaseReader/Writer for XDR based formats
 """
+import errno
 from . import base
 from .XDR import XDRBaseReader, XDRBaseWriter
 from ..lib.formats.libmdaxdr import TRRFile
@@ -150,6 +151,20 @@ class TRRReader(XDRBaseReader):
              'force': 'kJ/(mol*nm)'}
     _writer = TRRWriter
     _file = TRRFile
+
+    def _read_next_timestep(self, ts=None):
+        """copy next frame into timestep"""
+        if self._frame == self.n_frames - 1:
+            raise IOError(errno.EIO, 'trying to go over trajectory limit')
+        if ts is None:
+            ts = self.ts
+        ts.has_positions = True
+        ts.has_velocities = True
+        ts.has_forces = True
+        frame = self._xdr.read_direct_pvf(ts.positions, ts.velocities, ts.forces)
+        self._frame += 1
+        self._frame_to_ts(frame, ts)
+        return ts
 
     def _frame_to_ts(self, frame, ts):
         """convert a trr-frame to a mda TimeStep"""
--- a/package/MDAnalysis/coordinates/XDR.py
+++ b/package/MDAnalysis/coordinates/XDR.py
@@ -294,9 +294,12 @@ class XDRBaseReader(base.ReaderBase):
             raise IOError(errno.EIO, 'trying to go over trajectory limit')
         if ts is None:
             ts = self.ts
-        frame = self._xdr.read()
+        if ts.has_positions:
+            self._xdr.read_direct(ts.positions)
+        else:
+            frame = self._xdr.read()
+            self._frame_to_ts(frame, ts)
         self._frame += 1
-        self._frame_to_ts(frame, ts)
         return ts
 
     def Writer(self, filename, n_atoms=None, **kwargs):
--- a/package/MDAnalysis/coordinates/XDR.py
+++ b/package/MDAnalysis/coordinates/XDR.py
@@ -290,6 +290,7 @@ class XDRBaseReader(base.ReaderBase):
         return timestep
 
     def _read_next_timestep(self, ts=None):
+        # NOTE: TRR implements its own version
         """copy next frame into timestep"""
         if self._frame == self.n_frames - 1:
             raise IOError(errno.EIO, 'trying to go over trajectory limit')
--- a/package/MDAnalysis/coordinates/XDR.py
+++ b/package/MDAnalysis/coordinates/XDR.py
@@ -295,11 +295,11 @@ class XDRBaseReader(base.ReaderBase):
         if ts is None:
             ts = self.ts
         if ts.has_positions:
-            self._xdr.read_direct(ts.positions)
+            frame = self._xdr.read_direct(ts.positions)
         else:
             frame = self._xdr.read()
-            self._frame_to_ts(frame, ts)
         self._frame += 1
+        self._frame_to_ts(frame, ts)
         return ts
 
     def Writer(self, filename, n_atoms=None, **kwargs):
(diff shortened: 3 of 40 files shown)


package/MDAnalysis/lib/formats/libmdaxdr.pyx
```diff
@@ -674,32 +701,40 @@ cdef class XTCFile(_XDRFile):
...
+    @cython.boundscheck(False)
+    @cython.wraparound(False)
     def _calc_natoms(self, fname):
         cdef int n_atoms
-        return_code = read_xtc_natoms(fname, &n_atoms)
+        cdef int return_code = read_xtc_natoms(fname, &n_atoms)
         return return_code, n_atoms
 
-
+    @cython.boundscheck(False)
```
richardjgowers: IIRC you can set this with a pragma like statement in a comment at the top of the file. We can probably always default to fast and dangerous.

hmacdope: Yeah for sure can set with `# cython: boundscheck=False` but was just playing with the various functions. Will fix.

richardjgowers:
@hmacdope does this get rid of the yellow lines from the annotation?  One lazy way of smoking out Python bits is to put code blocks into `with nogil:` statements, as Cython throws a tantrum when there's Python (slow) sections inside nogil blocks.

hmacdope:
This is ready for a first review. 

Reading directly into a memoryview of `ts.positions` in the `read_direct` functions is a modest improvement (5% or so.)

As I see it the main remaining problem is in the `TRReader` which allocates enough memory for positions, velocities and forces every frame despite them frequently being absent. 

This is a bit of a chicken and eg problem though and  I don't see an easy way around this without the ability to "read" from a TRR frame whether there are forces, velocities present etc.  

I had a browse of the XDR source but couldn't see a flag that readily indicates presence or absence of data, `do_trn`  (`xdrfile_trr.c`) only checks that the input memory is not `NULL` or that the effective size of the array is 0. 

Query  @tylerjereddy @kain88-de as to whether it may be possible to read some kind of flagged state from the file? 

Cheers

kain88-de:
There is no flag known to me, the format is really basic. However, we already generated a map of the different frames to support random seeks. Nothing is stopping us from extending that map to also include information about the velocities and forces for each frame.

hmacdope:
> There is no flag known to me, the format is really basic. However, we already generated a map of the different frames to support random seeks. Nothing is stopping us from extending that map to also include information about the velocities and forces for each frame.

Okay, thanks for that. That was my understanding also that there were no flags. Thanks for confirming.

hmacdope:
This now has a "fat timestep" with positions, velocities and forces at every frame, as the arrays are allocated anyway in reading.  These are then read using memoryviews. 

Its about 5-8% more efficient in my testing.

package/MDAnalysis/lib/formats/libmdaxdr.pyx
```diff
@@ -549,6 +520,166 @@ cdef class TRRFile(_XDRFile):
...
+        if return_code == EOK:
+            self.current_frame += 1
+
+        has_x = bool(has_prop & HASX)
+        has_v = bool(has_prop & HASV)
+        has_f = bool(has_prop & HASF)
+        return TRRFrame(positions, velocity, forces, box, step, time, lmbda,
+                        has_x, has_v, has_f)
+
+    def read_direct_pvf(self, np.float32_t[:, ::1] positions,
```
orbeckst: Minor comment — Have we used *pvf* before? Otherwise I'd use *xvf* which looks  more like "positions, velocities, forces" to me. Or maybe that's just me.

hmacdope: fixed now

package/MDAnalysis/coordinates/XDR.py
```diff
@@ -294,7 +295,10 @@ def _read_next_timestep(self, ts=None):
             raise IOError(errno.EIO, 'trying to go over trajectory limit')
         if ts is None:
             ts = self.ts
-        frame = self._xdr.read()
+        if ts.has_positions:
+            frame = self._xdr.read_direct(ts.positions)
+        else:
+            frame = self._xdr.read()
```
orbeckst: When will the else clause be executed? codecov says it is not covered.

package/MDAnalysis/lib/formats/libmdaxdr.pyx
```diff
@@ -515,10 +477,19 @@ cdef class TRRFile(_XDRFile):
 
         # Use this instead of memviews here to make sure that references are
         # counted correctly
```
orbeckst: Does the comment still apply after switching to `np.PyArray_EMPTY()`?

package/MDAnalysis/lib/formats/libmdaxdr.pyx
```diff
@@ -549,6 +520,166 @@ cdef class TRRFile(_XDRFile):
...
+        See Also
+        --------
+        TRRFrame
+        XTCFile
+
+        Raises
+        ------
+        IOError
+        """
+        if self.reached_eof:
```
orbeckst: Have you figured out why none of the lines in this function appear in the code coverage?

package/MDAnalysis/lib/formats/libmdaxdr.pyx
```diff
@@ -549,6 +520,166 @@ cdef class TRRFile(_XDRFile):
...
+
+        cdef np.npy_intp[2] dim
+        dim[0] = self.n_atoms
+        dim[1] = DIMS
+
+        cdef np.npy_intp[2] unitcell_dim
+        unitcell_dim[0] = DIMS
+        unitcell_dim[1] = DIMS
+
+        cdef np.ndarray[np.float32_t, ndim=2] velocity = np.PyArray_EMPTY(2, dim, np.NPY_FLOAT32, 0)
```
orbeckst: Why are we only reading positions directly and not also velcities and forces?

Is this related to the question that you asked in the PR?

orbeckst (changes requested):
Half-baked review, sorry didn't get to everything.

package/MDAnalysis/coordinates/XDR.py
hmacdope: If a timestep is provided to `read_next_timestep` as an argument its not possible to guarantee it has `ts.positions` allocated.  I'll add a test

package/MDAnalysis/lib/formats/libmdaxdr.pyx
hmacdope: We now read positions, velocities and forces in the `read_direct_xvf()` function which is the only one called in TRR.py with the overload of `read_next_frame` but I kept this `read_direct()` method so that the base class implementation in XDR.py could theoretically still function.  I am happy to remove it if we want to YAGNI it away. 

Even if we are only reading one attribute you still need to allocate the memory for the others.

package/MDAnalysis/lib/formats/libmdaxdr.pyx
orbeckst: seem to be ok now

package/MDAnalysis/lib/formats/libmdaxdr.pyx
```diff
@@ -549,6 +517,174 @@ cdef class TRRFile(_XDRFile):
...
+
+        # In a trr the integer error seems to indicate that the file is ending.
+        # There might be corrupted files where this is a legitimate error. But
+        # then we just can't read it and stop there which is not too bad.
+        if return_code == EENDOFFILE or return_code == EINTEGER:
+            self.reached_eof = True
+            raise StopIteration
+
+        if return_code == EOK:
+            self.current_frame += 1
```
orbeckst: Just curious: under which circumstances do we want to continue with processing the frame if NOT `EOK`? I.e., shouldn't we simply do something like
```python
if return_code == EENDOFFILE or return_code == EINTEGER:
   ...
   raise StopIteration
elif return_code != EOK:
   raise IOError(f"TRR error {return_code}")

self.current_frame += 1
...
```

I assume that this code existed like this before.

package/MDAnalysis/lib/formats/libmdaxdr.pyx
```diff
@@ -549,6 +517,174 @@ cdef class TRRFile(_XDRFile):
...
+        """
+        if self.reached_eof:
+            raise EOFError('Reached last frame in TRR, seek to 0')
+        if not self.is_open:
+            raise IOError('No file opened')
+        if self.mode != 'r':
+            raise IOError('File opened in mode: {}. Reading only allow '
+                               'in mode "r"'.format('self.mode'))
+
+        return_code = 1
```
orbeckst: Why not `cdef int` the return_code?  (just curious, I know little about cython)

package/MDAnalysis/lib/formats/libmdaxdr.pyx
```diff
@@ -549,6 +517,174 @@ cdef class TRRFile(_XDRFile):
...
+
+        # In a trr the integer error seems to indicate that the file is ending.
+        # There might be corrupted files where this is a legitimate error. But
+        # then we just can't read it and stop there which is not too bad.
+        if return_code == EENDOFFILE or return_code == EINTEGER:
+            self.reached_eof = True
+            raise StopIteration
+
+        if return_code == EOK:
+            self.current_frame += 1
```
orbeckst: same question as for `read_direct`: should anything BUT `EOK` be a reason to fail?

package/MDAnalysis/lib/formats/libmdaxdr.pyx
orbeckst: Is the reason to keep `read_direct()` because it's part of the API? Can we deprecate it and remove in 3.0?

If it's not needed for what we're doing then we should consider removing it.

package/MDAnalysis/lib/formats/libmdaxdr.pyx
```diff
@@ -747,6 +892,67 @@ cdef class XTCFile(_XDRFile):
...
+        if return_code != EOK and return_code != EENDOFFILE:
+            raise IOError('XTC read error = {}'.format(
+                error_message[return_code]))
+
+        if return_code == EENDOFFILE:
+            self.reached_eof = True
+            raise StopIteration
+
+        if return_code == EOK:
+            self.current_frame += 1
```
orbeckst: same question as for read_direct above: should anything BUT `EOK` be a reason to fail?

orbeckst (changes requested):
In principle I would approve the PR. I am blocking it because of a few questions but I am ready to green-light it — just ping me.

It would be good if someone else with more cython experience than I could also have a look at it.

package/MDAnalysis/lib/formats/libmdaxdr.pyx
hmacdope: good point

package/MDAnalysis/lib/formats/libmdaxdr.pyx
hmacdope: I think its because the `return code` enum has other options that are not nessecarily errors but indicate some kind of file state. 

```cython
cdef enum:
    EOK = 0
    EHEADER = 1
    ESTRING = 2
    EDOUBLE = 3
    EINTEGER = 4
    EFLOAT = 5
    EUNSIGNEDINTEGER = 6
    ECOMPRESSION = 7
    ECLOSE = 8
    EMAGIC = 9
    EMEMORY = 10
    EENDOFFILE = 11
    ENOTFOUND = 12

```

hmacdope: So checking exclusively for EOK may break this ? Im not super sure (query @kain88-de @jbarnoud?)

package/MDAnalysis/lib/formats/libmdaxdr.pyx
hmacdope: See answer below.

package/MDAnalysis/lib/formats/libmdaxdr.pyx
hmacdope: This is copied from above so I was under the impression there was some logic to it. See query below for discussion.

hmacdope:
> Is the reason to keep read_direct() because it's part of the API? Can we deprecate it and remove in 3.0?
> If it's not needed for what we're doing then we should consider removing it.

@orbeckst 
`read_direct` was added in this PR so no need to deprecate if we don't want it **for TRR**. State of play in this PR is:

- `XDRReader._read_next_timestep()` base class calls `read()` or `read_direct(ts.positions)` depending if positions are already allocated in the `Timestep`
- `TRRReader` now implements its own `read_next_timestep` that calls `read_direct_xvf`, because it guarantees that `x`, `v` and `f` are always allocated.
- `XTC` does need an implementation of `read_direct()` as it uses the base class `_read_next_timestep()`
 - Not having an implementation of `read_direct` for `TRR` would be fine as only `read_direct_xvf` is called. 
- Only question is if we want it so that the **Base class API** is consistent (ie both XTC and TRR implement read_direct)

I know this is a bit convoluted, let me know if anything is unclear.

package/MDAnalysis/coordinates/XDR.py
```diff
@@ -289,12 +290,16 @@ def _read_frame(self, i):
...
         """copy next frame into timestep"""
         if self._frame == self.n_frames - 1:
             raise IOError(errno.EIO, 'trying to go over trajectory limit')
         if ts is None:
             ts = self.ts
-        frame = self._xdr.read()
+        if ts.has_positions:
+            frame = self._xdr.read_direct(ts.positions)
+        else:
+            frame = self._xdr.read()
```
orbeckst: If TRR implements its own then why have a generic method here that only works for XTC?

Let's remove `_read_next_timestep` from the base class. Then we can also cleanly remove `read_direct` from TRR.

package/MDAnalysis/coordinates/XDR.py
```diff
@@ -120,7 +120,8 @@ class XDRBaseReader(base.ReaderBase):
        XDR offsets read from trajectory if offsets file read-in fails
     .. versionchanged:: 2.0.0
        Add a InterProcessLock when generating offsets
-
+    .. versionchanged:: 2.4.0
+       Use a direct read into ts attributes
```
orbeckst, line 124: Although this is simply in the TRR and XTC xdr reader classes, it concur that we should keep a note here in the user-visible docs.

package/MDAnalysis/lib/formats/libmdaxdr.pxd
```diff
@@ -0,0 +1,110 @@
...
+cdef class _XDRFile:
+    """Base python wrapper for gromacs-xdr formats
+
+    This class can be similar to the normal file objects in python. The read()
+    function will return a frame and all information in it instead of a single
+    line. Additionally the context-manager protocoll is supported as well.
+
+    Note
+    ----
+    This class can't be initialized use one of the subclasses XTCFile, TRRFile
```
orbeckst:
```suggestion
    This class can't be initialized; use one of the subclasses :class:`XTCFile` or :class:`TRRFile`.
```

package/MDAnalysis/lib/formats/libmdaxdr.pyx
```diff
@@ -454,30 +410,36 @@ cdef class TRRFile(_XDRFile):
     -----
     This class can be pickled. The pickle will store filename, mode, current
     frame and offsets
+
+
+    .. versionchanged:: 2.4.0
+       Added read_direct and read_direct_xvf method to read into an existing positions array
```
orbeckst: Only add read_direct_xvf

package/MDAnalysis/lib/formats/libmdaxdr.pyx
```diff
@@ -549,6 +517,174 @@ cdef class TRRFile(_XDRFile):
         has_f = bool(has_prop & HASF)
         return TRRFrame(xyz, velocity, forces, box, step, time, lmbda,
                         has_x, has_v, has_f)
+    
+    def read_direct(self, np.float32_t[:, ::1] positions):
```
orbeckst: I'd drop it because it's not needed for the TRRFile class.

package/MDAnalysis/lib/formats/libmdaxdr.pyx
orbeckst: But we don't even return the state so it looks as if our reader may end up in a situation where we read a frame back but didn't increment the frame number but for the user still everything looks just fine. 

I suggest we fail harder here and wait for feedback from our tests and the community... I consider this fixing a behavior that can lead to silent errors.

package/MDAnalysis/coordinates/XDR.py
```diff
@@ -289,12 +290,16 @@ def _read_frame(self, i):
         return timestep
 
     def _read_next_timestep(self, ts=None):
```
orbeckst: ditch and implement individually in XTC and TRR

orbeckst (changes requested):
Thanks for the explanation, @hmacdope .

1. My take is we ditch the `XDRBaseReader._read_next_frame()` method and implement it individually in XTC and TRR — we only have two formats so why are we trying to be overly general, now that TRR needs its own anyway? Then we definitely don't need `TRRReader._read_direct()` and don't need to consider including it because it's clearly superfluous.
2. I would like the XDR readers to fail reading a frame unless the `return_code == EOK`, at least in the absence of any evidence that there's a good reason to continue the current behavior. We do not propagate the `return_code` so calling code cannot check the status and will always assume that a frame was read correctly. The only thing we do is to not increment the frame number. This all looks iffy to me for the potential to silently generate incorrect data. Should be fixed, in my opinion.

package/MDAnalysis/lib/formats/libmdaxdr.pyx
hmacdope: FIxed.

hmacdope:
@orbeckst  I have added an `IOError` raised on `retcode != EOK` and no errors seem to come up in tests, so seems relatively harmless. 

 I may have to `#pragma nocover` it perhaps, unless someone knows how to make a botched file, which seems a bit excessive.

package/MDAnalysis/coordinates/TRR.py
```diff
@@ -151,6 +152,28 @@ class TRRReader(XDRBaseReader):
...
 
+    def _read_next_timestep(self, ts=None):
+        """copy next frame into timestep
+        
+        versionadded:: 2.4.0
+            TRRReader implements this method so that it can use
+            read_direct_xvf to read the data directly into the timestep
+            rather than copying it from a temporary array.
+        """
+        if self._frame == self.n_frames - 1:
```
richardjgowers, line 163: I'm not sure if this should get caught in `ProtoReader`... I'd be curious if it has coverage

package/MDAnalysis/coordinates/TRR.py
```diff
@@ -151,6 +152,28 @@ class TRRReader(XDRBaseReader):
...
+            raise IOError(errno.EIO, 'trying to go over trajectory limit')
+        if ts is None:
+            ts = self.ts
```
richardjgowers, line 166: Can we raise an issue about this `ts=None` thing and remove it for 3.0.. it's not clear when this is ever used.  I imagine this is related to the copy behaviour confusion too. I think a Reader should hold a `Timestep` which is reused consistently, then people can **manually** copy if they want to.

package/MDAnalysis/lib/formats/libmdaxdr.pyx
```diff
@@ -513,12 +475,18 @@ cdef class TRRFile(_XDRFile):
...
-        cdef np.ndarray box = np.empty((DIMS, DIMS), dtype=DTYPE)
+        cdef np.npy_intp[2] dim
+        dim[0] = self.n_atoms
+        dim[1] = DIMS
+
+        cdef np.npy_intp[2] unitcell_dim
+        unitcell_dim[0] = DIMS
+        unitcell_dim[1] = DIMS
+
+        cdef np.ndarray[np.float32_t, ndim=2] xyz = np.PyArray_EMPTY(2, dim, np.NPY_FLOAT32, 0)
```
richardjgowers, line 486: can likely allocate this once per Reader, unless natoms is changing?

package/MDAnalysis/lib/formats/libmdaxdr.pyx
```diff
@@ -543,13 +511,99 @@ cdef class TRRFile(_XDRFile):
...
+            raise IOError('TRR read error = {}'.format(
+                error_message[return_code]))
 
         has_x = bool(has_prop & HASX)
         has_v = bool(has_prop & HASV)
         has_f = bool(has_prop & HASF)
         return TRRFrame(xyz, velocity, forces, box, step, time, lmbda,
                         has_x, has_v, has_f)
 
+    def read_direct_xvf(self, np.float32_t[:, ::1] positions,
```
richardjgowers, line 519: can we remove `.read()` and in 3.0 make direct how we say that the XDRFile should be used, it's the much faster pattern

richardjgowers (commented):
Looks good, but I'd make the direct versions the only implementations, but this would be. a break so it'll have to be a 3.0 thing

package/MDAnalysis/coordinates/TRR.py
hmacdope: Issue raised #3928

package/MDAnalysis/lib/formats/libmdaxdr.pyx
hmacdope: Issue raised #3929

package/CHANGELOG
```diff
@@ -34,6 +34,10 @@ Fixes
     (e.g. bonds, angles) (PR #3779).
 
 Enhancements
+  * Improve C content of libxdr Cython, add `read_direct` methods to read 
+    coordinates, velocities and forces directly into memoryviews of `Timestep`
+    attributes, make `TRR` timestep have positions, velocities and forces on
+    `__init__`. (Issue #3891 PR  #3892).
```
orbeckst, line 45: add under Fixes: XTC and TRR readers now fail with IOError when a status except EOK (=0) is detected on reading a frame instead of silently continuing

package/MDAnalysis/lib/formats/libmdaxdr.pyx
```diff
@@ -543,13 +511,99 @@ cdef class TRRFile(_XDRFile):
 
         if return_code == EOK:
             self.current_frame += 1
+        else:
+            raise IOError('TRR read error = {}'.format(
+                error_message[return_code]))
```
orbeckst: I hate to be a stickler but this can now be simplified to
```python
if return_code != EOK:
   raise IOError
self.current_frame += 1
```

package/MDAnalysis/lib/formats/libmdaxdr.pyx
```diff
@@ -543,13 +511,99 @@ cdef class TRRFile(_XDRFile):
...
+
+        # In a trr the integer error seems to indicate that the file is ending.
+        # There might be corrupted files where this is a legitimate error. But
+        # then we just can't read it and stop there which is not too bad.
+        if return_code == EENDOFFILE or return_code == EINTEGER:
+            self.reached_eof = True
+            raise StopIteration
+
+        if return_code == EOK:
```
orbeckst: write more compactly with single `if`

package/MDAnalysis/lib/formats/libmdaxdr.pyx
orbeckst: Is it possible that the weird original error handling is due to the need to be able to read through for building the offsets/getting total number of frames?? Or is this using other code paths?

package/MDAnalysis/lib/formats/libmdaxdr.pyx
```diff
@@ -745,8 +811,75 @@ cdef class XTCFile(_XDRFile):
 
         if return_code == EOK:
             self.current_frame += 1
+        else:
```
orbeckst: write more compact with single `if status != EOK`

package/MDAnalysis/lib/formats/libmdaxdr.pyx
```diff
@@ -745,8 +811,75 @@ cdef class XTCFile(_XDRFile):
...
+                                      <rvec*>&positions[0,0], <float*> &prec)
+        if return_code != EOK and return_code != EENDOFFILE:
+            raise IOError('XTC read error = {}'.format(
+                error_message[return_code]))
+
+        if return_code == EENDOFFILE:
+            self.reached_eof = True
+            raise StopIteration
+
+        if return_code == EOK:
```
orbeckst: write more compactly

orbeckst (changes requested):
Minor things + note that TRR and XTC reader may now file more frequently (CHANGELOG and add versionchanged where user visible)

package/MDAnalysis/lib/formats/libmdaxdr.pyx
hmacdope: It didn't seem to make any issues changing it but famous last words.

package/CHANGELOG
hmacdope: Some readers had enough redundancy built in to avoid not continuing with  `retcode != EOK` but at least its now done consistently.

hmacdope: This was the pattern that protected some branches 

```python
        if return_code != EOK and return_code != EENDOFFILE \
           and return_code != EINTEGER:
            raise IOError('TRR read error = {}'.format(
                **error_message[return_code]))**

        if return_code == EENDOFFILE or return_code == EINTEGER:
            self.reached_eof = True
            raise StopIteration

# can only be EOK at this point but it was a bit opaque. 

```

package/MDAnalysis/lib/formats/libmdaxdr.pyx
hmacdope: I tried and broke a bunch of tests and couldn't figure out why. I will come back to it.

hmacdope: If we are deprecating anyway should we even bother?

hmacdope:
@richardjgowers I'm going to make the executive decision to not implement buffers, because they are only relevant for methods that we plan to deprecate? Thoughts? LMK if you would prefer to do it.

richardjgowers:
@hmacdope yeah, because the buffers we're using are already the Timestep ones right? That sounds fine

richardjgowers (approved):

package/CHANGELOG
orbeckst: Ok, good to know that we probably do not change a lot of behavior. Nevertheless, the current approach is cleaner.

(Btw, while somewhere I came across my own explanation of why TRR stops with EINTEGER and I think it was related to not being able to read the magic number.)

orbeckst (approved):

orbeckst:
Should be good to go for a squash-merge.

hmacdope:
With two approvals I will go ahead and merge. :)

Issue #3893, opened 2022-10-29
Deprecate inclusive range in MemoryReader


hmacdope:
## Expected behavior ##

The MemoryReader should use non-inclusive indexing like Python's range and Numpy. 

## Actual behavior ##

The memoryreader is documented as having inclusive ranges 

Following discussion in #3890 we would like to deprecate this behaviour and change in 3.0.

PR #3894, opened 2022-10-29
Add deprecation warning for memoryreader inclusive indexing in timeseries. 

hmacdope:
Related to #3893 
 - Adds deprecation warning for memoryreader. 

PR Checklist
------------
 - [X] Tests?
 - [X] Docs?
 - [X] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -62,6 +62,7 @@ Changes
   * adding element attribute to TXYZParser if all atom names are valid element symbols (PR #3826)
 
 Deprecations
+  * Add deprecation warning for inclusive start, stop, step indexing in MemoryReader.timeseries 
 
 
 08/29/22 IAlibay, PicoCentauri, orbeckst, hmacdope, rmeli, miss77jun, rzhao271,
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -62,7 +62,8 @@ Changes
   * adding element attribute to TXYZParser if all atom names are valid element symbols (PR #3826)
 
 Deprecations
-  * Add deprecation warning for inclusive start, stop, step indexing in MemoryReader.timeseries 
+  * Add deprecation warning for inclusive start, stop, step indexing
+    in MemoryReader.timeseries (#PR 3894)  
 
 
 08/29/22 IAlibay, PicoCentauri, orbeckst, hmacdope, rmeli, miss77jun, rzhao271,
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -62,8 +62,8 @@ Changes
   * adding element attribute to TXYZParser if all atom names are valid element symbols (PR #3826)
 
 Deprecations
-  * Add deprecation warning for inclusive start, stop, step indexing
-    in MemoryReader.timeseries (#PR 3894)  
+  * Add deprecation warning for inclusive stop indexing in
+    MemoryReader.timeseries (#PR 3894)
 
 
 08/29/22 IAlibay, PicoCentauri, orbeckst, hmacdope, rmeli, miss77jun, rzhao271,
--- a/package/MDAnalysis/coordinates/memory.py
+++ b/package/MDAnalysis/coordinates/memory.py
@@ -515,6 +515,12 @@ class MemoryReader(base.ProtoReader):
         .. versionchanged:: 1.0.0
            Deprecated `format` keyword has been removed. Use `order` instead.
         """
+        warnings.warn(
+            "Inclusive `start` and `stop` indexing will be removed in 3.0 in" 
+            " favour of exclusive indexing"
+            category=DeprecationWarning
+        )
+
         array = self.get_array()
         if order == self.stored_order:
             pass
--- a/package/MDAnalysis/coordinates/memory.py
+++ b/package/MDAnalysis/coordinates/memory.py
@@ -501,11 +501,14 @@ class MemoryReader(base.ProtoReader):
             of the underlying numpy array is returned, while a copy of the
             data is returned whenever `asel` is different from ``None``.
         start : int (optional)
+            the start trajectory frame
         stop : int (optional)
+            the end trajectory frame
             .. deprecated:: 2.4.0
                 Note that `stop` is currently *inclusive* but will be
                 changed in favour of being *exclusive* in version 3.0.  
         step : int (optional)
+            the number of trajectory frames to skip
         order : {"afc", "acf", "caf", "fac", "fca", "cfa"} (optional)
             the order/shape of the return data array, corresponding
             to (a)tom, (f)rame, (c)oordinates all six combinations
--- a/package/MDAnalysis/coordinates/memory.py
+++ b/package/MDAnalysis/coordinates/memory.py
@@ -504,9 +504,11 @@ class MemoryReader(base.ProtoReader):
             the start trajectory frame
         stop : int (optional)
             the end trajectory frame
+
             .. deprecated:: 2.4.0
-            Note that `stop` is currently *inclusive* but will be
-            changed in favour of being *exclusive* in version 3.0.  
+               Note that `stop` is currently *inclusive* but will be
+               changed in favour of being *exclusive* in version 3.0.  
+
         step : int (optional)
             the number of trajectory frames to skip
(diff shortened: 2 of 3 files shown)


package/MDAnalysis/coordinates/memory.py
```diff
@@ -515,6 +515,10 @@ def timeseries(self, asel=None, start=0, stop=-1, step=1, order='afc'):
         .. versionchanged:: 1.0.0
            Deprecated `format` keyword has been removed. Use `order` instead.
         """
+        warnings.warn("MemoryReader.timeseries inclusive `start` and `stop`"
```
orbeckst: Can  we do the warning only if `stop` is changed and not the default? 

The warning should only be about `start` becoming exclusive instead of inclusive.

orbeckst (changes requested):
I suggest to make the warning more specific and we need
- test
- entry in CHANGELOG under Deprecations
- doc entry for deprecation (under `stop` keyword)

package/MDAnalysis/coordinates/memory.py
hmacdope: @orbeckst can warn only if stop is set. 

I assume you mean stop rather than start in "The warning should only be about start becoming exclusive instead of inclusive"?

package/MDAnalysis/coordinates/memory.py
```diff
@@ -514,7 +518,14 @@ def timeseries(self, asel=None, start=0, stop=-1, step=1, order='afc'):
 
         .. versionchanged:: 1.0.0
            Deprecated `format` keyword has been removed. Use `order` instead.
+        .. versionchanged:: 2.4.0
```
IAlibay: use a `.. deprecated` and add more details on what's deprecated?

IAlibay (changes requested):

package/MDAnalysis/coordinates/memory.py
orbeckst: Yes, you’re right, of course. Sorry for the typo.

package/CHANGELOG
```diff
@@ -62,6 +62,8 @@ Changes
   * adding element attribute to TXYZParser if all atom names are valid element symbols (PR #3826)
 
 Deprecations
+  * Add deprecation warning for inclusive start, stop, step indexing
```
orbeckst: The only change is just how stop is handled.

package/MDAnalysis/coordinates/memory.py
```diff
@@ -502,8 +502,12 @@ def timeseries(self, asel=None, start=0, stop=-1, step=1, order='afc'):
             data is returned whenever `asel` is different from ``None``.
         start : int (optional)
         stop : int (optional)
+            Note that `stop` is currently *inclusive* but will be deprecated in
+            favour of being *exclusive* in 3.0.
```
orbeckst: Add a deprecation note (rest) right here under the keyword.

package/MDAnalysis/coordinates/memory.py
```diff
@@ -502,8 +502,12 @@ def timeseries(self, asel=None, start=0, stop=-1, step=1, order='afc'):
             data is returned whenever `asel` is different from ``None``.
         start : int (optional)
         stop : int (optional)
+            Note that `stop` is currently *inclusive* but will be deprecated in
+            favour of being *exclusive* in 3.0.  
         step : int (optional)
-            range of trajectory to access, `start` and `stop` are *inclusive*
+            range of trajectory to access, `start` and `stop` are
```
orbeckst: Just remove the inclusive Blabla from here, it never made sense under step anyway. The only place that changes is stop.

package/MDAnalysis/coordinates/memory.py
orbeckst: I probably wouldn’t add a versionchanged here and only have the .. deprecated under the stop kwarg.

package/MDAnalysis/coordinates/memory.py
```diff
@@ -514,7 +518,14 @@ def timeseries(self, asel=None, start=0, stop=-1, step=1, order='afc'):
 
         .. versionchanged:: 1.0.0
            Deprecated `format` keyword has been removed. Use `order` instead.
+        .. versionchanged:: 2.4.0
+           Deprecation warning 
         """
+        if stop != -1:
+            warnings.warn("MemoryReader.timeseries inclusive `start` and `stop`"
```
orbeckst: Only stop. Start is always inclusive. 

If we mention start here then users might be confused if the meaning of start also changes.

orbeckst (changes requested):

package/MDAnalysis/coordinates/memory.py
```diff
@@ -502,8 +502,10 @@ def timeseries(self, asel=None, start=0, stop=-1, step=1, order='afc'):
             data is returned whenever `asel` is different from ``None``.
         start : int (optional)
         stop : int (optional)
+            .. deprecated:: 3.0
+                Note that `stop` is currently *inclusive* but will be
+                deprecated in favour of being *exclusive* in 3.0.  
         step : int (optional)
-            range of trajectory to access, `start` and `stop` are *inclusive*
```
IAlibay, line 506: Might be missing something but diff indicates no replacement for this? You probably should have done kind of "this is what step does"?

package/MDAnalysis/coordinates/memory.py
```diff
@@ -502,8 +502,10 @@ def timeseries(self, asel=None, start=0, stop=-1, step=1, order='afc'):
             data is returned whenever `asel` is different from ``None``.
         start : int (optional)
         stop : int (optional)
+            .. deprecated:: 3.0
+                Note that `stop` is currently *inclusive* but will be
+                deprecated in favour of being *exclusive* in 3.0.
```
IAlibay:
```suggestion
            .. deprecated:: 2.4.0
                Note that `stop` is currently *inclusive* but will be
                changed in favour of being *exclusive* in version 3.0.  
```
essentially deprecation happens now, change happens in the future

IAlibay (changes requested):

package/MDAnalysis/coordinates/memory.py
```diff
@@ -501,9 +501,14 @@ def timeseries(self, asel=None, start=0, stop=-1, step=1, order='afc'):
             of the underlying numpy array is returned, while a copy of the
             data is returned whenever `asel` is different from ``None``.
         start : int (optional)
+            the start trajectory frame
         stop : int (optional)
+            the end trajectory frame
+            .. deprecated:: 2.4.0
+            Note that `stop` is currently *inclusive* but will be
+            changed in favour of being *exclusive* in version 3.0.
```
IAlibay:
```suggestion

            .. deprecated:: 2.4.0
               Note that `stop` is currently *inclusive* but will be
               changed in favour of being *exclusive* in version 3.0.  

```
Reformatting broke docs, see: https://mdanalysis--3894.org.readthedocs.build/en/3894/documentation_pages/coordinates/memory.html

IAlibay (changes requested):
See edit: oops I started writing here then wrote the specific comment

package/MDAnalysis/coordinates/memory.py
orbeckst: Yes, you need an empty line above reST paragraph constructs.

package/MDAnalysis/coordinates/memory.py
```diff
@@ -515,6 +520,11 @@ def timeseries(self, asel=None, start=0, stop=-1, step=1, order='afc'):
         .. versionchanged:: 1.0.0
            Deprecated `format` keyword has been removed. Use `order` instead.
         """
+        if stop != -1:
+            warnings.warn("MemoryReader.timeseries inclusive `stop` "
+                      "indexing will be removed in 3.0 in favour of exclusive "
+                      "indexing",category=DeprecationWarning)
```
orbeckst: pep8
```suggestion
                      "indexing", category=DeprecationWarning)
```

orbeckst (approved):
lgtm

(just applying suggestions)

IAlibay (approved):
lgtm assuming CI comes back green

PR #3899, opened 2022-11-01
Bump up codecov action, add token secret

IAlibay:
The codecov reports here are a little bit over the place, and I'm not really sure that this is the best way forward, but at the same time if it doesn't prevent PRs coming from forks then there's no major cost.

We should keep an eye out to see if the relevant issues, e.g. https://github.com/codecov/feedback/issues/126 are fixed & things can be reverted back.

PR Checklist
------------
 - Tests?
 - Docs?
 - CHANGELOG updated?
 - Issue raised/referenced?

### DIFF
--- a/.github/workflows/gh-ci.yaml
+++ b/.github/workflows/gh-ci.yaml
@@ -122,8 +122,9 @@ jobs:
 
     - name: codecov
       if: matrix.codecov
-      uses: codecov/codecov-action@v2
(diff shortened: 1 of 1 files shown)


orbeckst (approved):
Let's try it.

(I saw that you added `CODECOV_TOKEN` to the repo secrets.)

Issue #3878, opened 2022-10-25
Python 3.11


IAlibay:
Python 3.11 is now out 🚀 

As usual, making sure CI is fully 3.11 compliant is going to take a long while, mostly for conda-forge to catch up, and then for all our optional upstream deps to release 3.11 builds.

We've been testing against 3.11-rc.2 and everything seems fine, so we can probably start with pure python builds in the immediate term.

This issue exists to keep track of everything we need to do:

- [x] Update the twice weekly pure Python cron jobs to include 3.11 instead of 3.11-rc.2
- [x] Update the setup classifiers to include 3.11
- [x] Update main CI (conda-forge / mamba installs) to include 3.11 as a minimum deps 3.11 install
- [x] Update main CI to have a full deps 3.11 install
- [x] Update CD to include cp3.11 wheel building: https://github.com/MDAnalysis/mdanalysis/blob/develop/.github/workflows/deploy.yaml#L39
- [x] Update CD testpypi checks to include 3.11 https://github.com/MDAnalysis/mdanalysis/blob/develop/.github/workflows/deploy.yaml#L192
- [x] Add 3.11 to azure pipelines yaml call
- [x] Update cirrus build to use 3.11
- [x] Bump up asv benchmarks to also be py3.11

orbeckst:
Given that 3.11 is supposedly increasing performance by 10-30% we should add it to the ASV benchmarks, too. (Not in this issue…more as a reminder to myself and anyone interested.)

orbeckst:
Can we close this issue? Aren't we fully supporting 3.11 by now?

IAlibay:
> Can we close this issue? Aren't we fully supporting 3.11 by now?

Please see the two missing tasks.

Most of our optional deps (including h5py which is required for azure pipelines) do not support py3.11 yet.

edit: by support I mean wheels here, technically h5py probably works fine with py311, but we can't build the wheels natively in azure pipelines without a lot of extra work.

PR #3903, opened 2022-11-04
bump up azure and gh CI to py3.11

IAlibay:
Towards #3878

 - bump up azure to py3.11
 - bump up gh actions CI to py3.11 for a "minimal deps" build (I know that openmm and rdkit don't have 3.11 builds yet, so we're blocked on a full deps build for now).
 - Add py3.11 to CD workflow

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -47,7 +47,7 @@ jobs:
           fetch-depth: 0
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2
+        uses: pypa/cibuildwheel@v2.11
         with:
           package-dir: package
         env:
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -4,7 +4,6 @@ on:
   pull_request:
     branches:
       - "package-*"
-      - "develop"
   push:
     branches:
       - "package-*"
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -189,9 +189,12 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ["3.8", "3.9", "3.10"]
+        python-version: ["3.8", "3.9", "3.10", "3.11"]
         type: ["FULL", "MIN"]
         exclude:
+          # Py311 doesn't support all optional deps yet
+          - python-version: "3.11"
+            type: "FULL"
           # Multiple deps don't like windows
           - os: windows-latest
             python-version: "3.8"
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -4,6 +4,7 @@ on:
   pull_request:
     branches:
       - "package-*"
+      - "develop"
   push:
     branches:
       - "package-*"
@@ -33,10 +34,10 @@ jobs:
       fail-fast: false
       matrix:
         buildplat:
-          - [ubuntu-18.04, manylinux_x86_64]
-          - [macos-10.15, macosx_*]
+          - [ubuntu-20.04, manylinux_x86_64]
+          - [macos-11, macosx_*]
           - [windows-2019, win_amd64]
-        python: ["cp38", "cp39", "cp310"]
+        python: ["cp38", "cp39", "cp310", "cp311"]
     defaults:
       run:
         working-directory: ./package
@@ -46,7 +47,7 @@ jobs:
           fetch-depth: 0
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.6.0
+        uses: pypa/cibuildwheel@v2
         with:
           package-dir: package
         env:
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -47,7 +47,7 @@ jobs:
           fetch-depth: 0
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.11
+        uses: pypa/cibuildwheel@v2.11.2
         with:
           package-dir: package
         env:
--- a/.github/workflows/gh-ci.yaml
+++ b/.github/workflows/gh-ci.yaml
@@ -28,11 +28,17 @@ jobs:
         fail-fast: false
         matrix:
           os: [ubuntu-latest, ]
-          python-version: [3.8, 3.9, "3.10"]
+          python-version: ["3.8", "3.9", "3.10"]
           full-deps: [true, ]
           install_hole: [true, ]
           codecov: [true, ]
           include:
+            - name: py311-ubuntu-min
+              os: ubuntu-latest
+              python-version: "3.11"
+              full-deps: false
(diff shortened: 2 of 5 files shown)


IAlibay:
ok a minimal build works fine using `conda`.

IAlibay:
Stuck waiting on the next h5py release for now (should be a week or so)

azure-pipelines.yml
```diff
@@ -36,23 +36,23 @@ jobs:
           PYTHON_ARCH: 'x64'
           BUILD_TYPE: 'normal'
           imageName: 'windows-2019'
-        Win-Python39-64bit-full-wheel:
-          PYTHON_VERSION: '3.9'
+        Win-Python310-64bit-full-wheel:
```
IAlibay, line 39: lxml missing wheels are a Cython level issue for win64. We need to get wheel generation via cibuildwheels up by the 1st, so let's just add py3.11 windows checking at a later date (probably when lxml does its next release).

azure-pipelines.yml
```diff
@@ -36,23 +36,23 @@ jobs:
...
+          PYTHON_VERSION: '3.10'
           PYTHON_ARCH: 'x64'
           BUILD_TYPE: 'wheel'
-          NUMPY_MIN: '1.20.0'
+          NUMPY_MIN: '1.21.6'
```
IAlibay: @tylerjereddy I was hoping I might be able to pick your brains on this one.

Essentially for the python 3.10 "full wheels" runners (i.e. no deps pip install of MDA, then install deps, then install the minimum version of numpy), the linux runner works perfectly fine when setting `NUMPY_MIN` to `1.21.6` (the version required to build MDAnalysis for python 3.10). However, the windows one ends up with a ton of ABI issues when running tests. 

I can't seem to work out why this is happening. My understanding is that since `NUMPY_MIN` is equivalent to the numpy version used to build MDAnalysis then there should be no ABI compatibility issues?

IAlibay:
This is a major blocker for 2.4.0 by the way - we can't move forward if build isolation is broken (we're relying on this behaviour for deployment, etc...)

IAlibay:
@tylerjereddy can I ping you again on this py3.10 wheel failure? I have yet to be able to find a solution on why this is happening.

azure-pipelines.yml
tylerjereddy: @IAlibay I think because we depend on SciPy (wheels), which have a strict Windows Python 3.10 requirement for newer NumPy because of https://github.com/scipy/scipy/blob/main/pyproject.toml#L38.

```
# On Windows we need to avoid 1.21.6, 1.22.0 and 1.22.1 because they were
# built with vc142. 1.22.3 is the first version that has 32-bit Windows
# wheels *and* was built with vc141. So use that:
"numpy==1.22.3; python_version=='3.10' and platform_system=='Windows' and platform_python_implementation != 'PyPy'",
```

So the transitive dependency on newer NumPy creates an ABI mismatch when using SciPy functionality. i.e.,

```
<snip>
C:\hostedtoolcache\windows\Python\3.10.8\x64\lib\site-packages\scipy\spatial\_kdtree.py:5: in <module>
    from ._ckdtree import cKDTree, cKDTreeNode
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   ???
E   ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject
```

tylerjereddy: Hopefully the MDA team can live with bumping up the min to accommodate that?

tylerjereddy: Alternatively, I suppose pin to an older SciPy, not sure how much luck you'd have with that, but suppose it may be an option perhaps.

IAlibay: Aha! Thanks so much @tylerjereddy!

@MDAnalysis/coredevs would you be ok with us just adding an extra pyproject.toml line to bump up the windows numpy requirements by one minor version?

hmacdope: Sounds good to me 🤗

IAlibay:
@MDAnalysis/coredevs this one should be good to review now (it's required for the 2.4.0 release)

.github/workflows/deploy.yaml
```diff
@@ -33,10 +33,10 @@ jobs:
       fail-fast: false
       matrix:
         buildplat:
-          - [ubuntu-18.04, manylinux_x86_64]
-          - [macos-10.15, macosx_*]
+          - [ubuntu-20.04, manylinux_x86_64]
+          - [macos-11, macosx_*]
```
orbeckst, line 37: fyi, I think they are switching to macos-12 soon, I remember seeing notices in Actions .... or is that only relevant for tests and not deploy?

.github/workflows/gh-ci.yaml
```diff
@@ -28,11 +28,17 @@ jobs:
...
           full-deps: [true, ]
           install_hole: [true, ]
           codecov: [true, ]
           include:
+            - name: py311-ubuntu-min
+              os: ubuntu-latest
+              python-version: "3.11"
+              full-deps: false
+              install_hole: true
+              codecov: false
```
orbeckst, line 41: Generic question: what's the advantage of `codecov: false`? For a minimal build, wouldn't we also want to check that our code paths with missing dependencies are exercised? I don't remember if we have tests that only trigger when a dependency is missing.

azure-pipelines.yml
```diff
@@ -104,7 +104,6 @@ jobs:
       python -m pip install numpy==$(NUMPY_MIN)
     displayName: 'pin to older NumPy (wheel test)'
     condition: and(succeeded(), eq(variables['BUILD_TYPE'], 'wheel'))
-  # TODO: recent rdkit is not on PyPI
```
orbeckst, line 107: hooray for being able to remove the comment

orbeckst (approved):
This is a lot of nitty-gritty, "just" to build nice distributions. Thank you for digging into it.

.github/workflows/deploy.yaml
IAlibay: So my understanding is that macos-latest is switching to macos-12 (from 11) soon. I've been trying to deploy on the older version of the OS with the idea that it'll _maybe_ be forward compatible but maybe not as backwards compatible. I'm not sure it really matters too much but happy to bump it up/down as necessary.

.github/workflows/gh-ci.yaml
IAlibay: speed mostly (roughly 2x in some cases). We have a `minimal-ubuntu` runner that just does lowest python with no optional deps so I've been relying on that one being the one to pick up any "raise error if missing dependency" things.

IAlibay:
> This is a lot of nitty-gritty, "just" to build nice distributions.

Yeah I always despair when I see a diff of a couple dozen lines and realise I spent many days hunting some obscure things across github 😅

PR #3906, opened 2022-11-05
ChainReader now subclasses BaseReader

### LINKED ISSUE #3657: Applying Transformation to Chain of Traj doesn't Apply to First Frame of 2nd, 3rd... trajectories

### LINKED ISSUE #4008: Some frames (the ones at the trajectory boundaries) are not transformed when using transformations on universes from chainreaders
I am not sure if this is related (i.e. the same underlying cause) as #3343 (there it seems to happen more often than just at the trajectory boundaries, but for the DCD traj below I also have more issues than just at the traj boundaries).
Please let me know if you need any additional info.

## Expected behavior ##

All frames in a trajectory are transformed according to the defined on-the-fly transformations.

## Actual behavior ##

The first frame (frame 0) of all trajectories except the first trajectory one are not transformed (if my transformation depends on an atom group).
Although this seems to be True only for XTC and TRR. The behavior I get If I do the same with the DCD from the testfiles more closely resembles #3343 i.e. every second value is wrong (independent of the trajectory boundaries). See the code example below.

The below code will produce three plots, all of them should be a flat line (we substract the Center of Mass of all atoms in the trafo and plot the center of mass of all atoms). However for XTC and TRR we have one jump at frame 10 (which is the first frame of the second trajectory) and for DCD we have strange oscillations. Adding a third TRR/XTC results in another jump at frame 20. I dont see any correlation for the DCD with the trajectory boundary (at frame 98) except maybe that the oscillattions of the y-coordinate seem even stronger.

``` python
import numpy as np
import MDAnalysis as mda
from MDAnalysis.tests.datafiles import PSF, DCD,  TPR, XTC, TRR

# define our two trafos, the first one works, the second one does not work on all frames
def shift_up(shift=10):
    def wrapped(ts):
        ts.positions[:, 2] += shift
        return ts
    return wrapped

def shift_by_group(ag):
    # this just shifts all atoms in the box by their center of mass,
    # i.e. afterwards the center of mass of all atoms should be [0,0,0]
    def wrapped(ts):
        shift = ag.center_of_mass()
        ts.positions[:] -= shift
        return ts
    return wrapped

# the lists make sure we get the chainreader
universes = [mda.Universe(PSF, [DCD, DCD]), mda.Universe(TPR, [XTC, XTC]), mda.Universe(TPR, [TRR, TRR])]
traj_formats = ["DCD", "XTC", "TRR"]
fig, axs = plt.subplots(3)

for ax, u, name in zip(axs, universes, traj_formats):
    com = np.empty((len(u.trajectory), 3))
    ag = u.atoms
    trafos = (shift_by_group(ag),)
    u.trajectory.add_transformations(*trafos)
    for i, ts in enumerate(u.trajectory):
         com[i] = ag.center_of_mass()
    ax.plot(com, 'o-')
    ax.set_title(name)
ax.set_xlabel("frames")
fig.tight_layout()
```

- Which version are you using? 2.4.2
- Which version of Python (`python -V`)? Python 3.9.7
- Which operating system? linux

EDIT: nicer plot produced.

richardjgowers:
Fixes #4008
Fixes #3657

 - 

ChainReader currently subclasses `ProtoReader` rather than `ReaderBase`, then attempts to implement many things regarding auxiliaries and transformations itself.  I think it's getting many subtle things wrong with aux and transformations.

[This PR makes `ChainReader` just another subclass of `ReaderBase`, which ultimately simplifies it.](https://github.com/MDAnalysis/mdanalysis/pull/3906/files#r1205817202)

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/coordinates/base.py
+++ b/package/MDAnalysis/coordinates/base.py
@@ -621,7 +621,7 @@ class _Readermeta(abc.ABCMeta):
     """
     # Auto register upon class creation
     def __init__(cls, name, bases, classdict):
-        type.__init__(type, name, bases, classdict)
+        type.__init__(type, name, bases, classdict)  # pylint: disable=non-parent-init-called
         try:
             fmt = asiterable(classdict['format'])
         except KeyError:
--- a/package/MDAnalysis/coordinates/base.py
+++ b/package/MDAnalysis/coordinates/base.py
@@ -122,12 +122,11 @@ writers share.
    :members:
 
 """
+import abc
 import numpy as np
 import numbers
-import copy
 import warnings
-import weakref
-from typing import Union, Optional, List, Dict
+from typing import Any, Union, Optional, List, Dict
 
 from .timestep import Timestep
 from . import core
@@ -614,7 +613,7 @@ class IOBase(object):
         return False  # do not suppress exceptions
 
 
-class _Readermeta(type):
+class _Readermeta(abc.ABCMeta):
     """Automatic Reader registration metaclass
 
     .. versionchanged:: 1.0.0
@@ -664,6 +663,10 @@ class ProtoReader(IOBase, metaclass=_Readermeta):
     #: The appropriate Timestep class, e.g.
     #: :class:`MDAnalysis.coordinates.xdrfile.XTC.Timestep` for XTC.
     _Timestep = Timestep
+    _transformations: list
+    _auxs: dict
+    _filename: Any
+    n_frames: int
 
     def __init__(self):
         # initialise list to store added auxiliary readers in
@@ -671,7 +674,7 @@ class ProtoReader(IOBase, metaclass=_Readermeta):
         self._auxs = {}
         self._transformations=[]
 
-    def __len__(self):
+    def __len__(self) -> int:
         return self.n_frames
 
     @classmethod
@@ -691,7 +694,7 @@ class ProtoReader(IOBase, metaclass=_Readermeta):
         raise NotImplementedError("{} cannot deduce the number of atoms"
                                   "".format(cls.__name__))
 
-    def next(self):
+    def next(self) -> Timestep:
         """Forward one step to next frame."""
         try:
             ts = self._read_next_timestep()
@@ -706,22 +709,22 @@ class ProtoReader(IOBase, metaclass=_Readermeta):
 
         return ts
 
-    def __next__(self):
+    def __next__(self) -> Timestep:
         """Forward one step to next frame when using the `next` builtin."""
         return self.next()
 
-    def rewind(self):
+    def rewind(self) -> Timestep:
         """Position at beginning of trajectory"""
         self._reopen()
         self.next()
 
     @property
-    def dt(self):
+    def dt(self) -> float:
         """Time between two trajectory frames in picoseconds."""
         return self.ts.dt
 
     @property
-    def totaltime(self):
+    def totaltime(self) -> float:
         """Total length of the trajectory
 
         The time is calculated as ``(n_frames - 1) * dt``, i.e., we assume that
@@ -733,7 +736,7 @@ class ProtoReader(IOBase, metaclass=_Readermeta):
         return (self.n_frames - 1) * self.dt
 
     @property
-    def frame(self):
+    def frame(self) -> int:
         """Frame number of the current time step.
 
         This is a simple short cut to :attr:`Timestep.frame`.
@@ -781,20 +784,21 @@ class ProtoReader(IOBase, metaclass=_Readermeta):
             pass
         return core.writer(filename, **kwargs)
 
-    def _read_next_timestep(self, ts=None):  # pragma: no cover
+    @abc.abstractmethod
+    def _read_next_timestep(self, ts=None):
         # Example from DCDReader:
         #     if ts is None:
         #         ts = self.ts
         #     ts.frame = self._read_next_frame(etc)
         #     return ts
-        raise NotImplementedError(
-            "BUG: Override _read_next_timestep() in the trajectory reader!")
+        ...
 
     def __iter__(self):
         """ Iterate over trajectory frames. """
         self._reopen()
         return self
 
+    @abc.abstractmethod
     def _reopen(self):
         """Should position Reader to just before first frame
 
@@ -980,7 +984,7 @@ class ProtoReader(IOBase, metaclass=_Readermeta):
             nframes=self.n_frames,
             natoms=self.n_atoms
         ))
-    
+
     def timeseries(self, asel: Optional['AtomGroup']=None,
                    start: Optional[int]=None, stop: Optional[int]=None,
                    step: Optional[int]=None,
@@ -1703,6 +1707,9 @@ class SingleFrameReaderBase(ProtoReader):
 
     def next(self):
         raise StopIteration(self._err.format(self.__class__.__name__))
+
+    def _read_next_timestep(self, ts=None):
+        raise NotImplementedError(self._err.format(self.__class__.__name__))
 
     def __iter__(self):
         self.rewind()
--- a/package/MDAnalysis/coordinates/chain.py
+++ b/package/MDAnalysis/coordinates/chain.py
@@ -44,8 +44,7 @@ import warnings
 
 import os.path
 import bisect
-import copy
-
+from typing import Tuple
 import numpy as np
 
 from ..lib import util
@@ -159,7 +158,7 @@ def check_allowed_filetypes(readers, allowed):
                                   "supported for formats: {}".format(allowed))
 
 
-class ChainReader(base.ProtoReader):
+class ChainReader(base.ReaderBase):
     """Reader that concatenates multiple trajectories on the fly.
 
     The :class:`ChainReader` is used by MDAnalysis internally to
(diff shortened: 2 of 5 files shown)


package/MDAnalysis/coordinates/base.py
```diff
@@ -664,12 +663,27 @@ class ProtoReader(IOBase, metaclass=_Readermeta):
     #: The appropriate Timestep class, e.g.
     #: :class:`MDAnalysis.coordinates.xdrfile.XTC.Timestep` for XTC.
     _Timestep = Timestep
+    filename: Any
+    convert_units: bool
+    _transformations: list
+    _auxs: dict
 
-    def __init__(self):
+    def __init__(self, filename, convert_units):
```
hmacdope: Does convert units have an accepted default?

package/MDAnalysis/coordinates/base.py
```diff
@@ -781,20 +795,21 @@ def OtherWriter(self, filename, **kwargs):
...
+    @abc.abstractmethod
+    def _read_next_timestep(self, ts=None):
         # Example from DCDReader:
         #     if ts is None:
         #         ts = self.ts
         #     ts.frame = self._read_next_frame(etc)
         #     return ts
-        raise NotImplementedError(
-            "BUG: Override _read_next_timestep() in the trajectory reader!")
+        raise NotImplementedError()
```
hmacdope: Leave the original error message?

package/MDAnalysis/coordinates/base.py
```diff
@@ -1613,7 +1625,11 @@ def copy(self):
 
         return new
 
-    def _read_first_frame(self):  # pragma: no cover
+    def _read_next_timestep(self, ts=None):
+        raise NotImplementedError("SingleFrameReader can't read next frame")
```
hmacdope: Add a test.

package/MDAnalysis/coordinates/base.py
```diff
@@ -1613,7 +1625,11 @@ def copy(self):
...
+
+    @abc.abstractmethod
+    def _read_first_frame(self):
         # Override this in subclasses to create and fill a Timestep
```
hmacdope: I assume  abc.abstractmethod takes care of raising error

package/MDAnalysis/coordinates/chain.py
```diff
@@ -217,9 +217,11 @@ class ChainReader(base.ProtoReader):
 
     """
     format = 'CHAIN'
+    _n_frames: int
```
hmacdope: I'm missing why this is a class attr rather than attribute?

package/MDAnalysis/coordinates/chain.py
```diff
@@ -450,15 +456,6 @@ def convert_pos_from_native(self, x):
     def convert_pos_to_native(self, x):
         return self.active_reader.convert_pos_to_native(x)
 
-    def copy(self):
```
hmacdope, line 453: Why remove copy?

hmacdope (changes requested):
Few quick comments. :)

package/MDAnalysis/coordinates/chain.py
richardjgowers: this is just a type annotation

package/MDAnalysis/coordinates/chain.py
richardjgowers: it's in the base class, this is removing a faulty version of copy - it won't copy auxiliaries correctly

hmacdope: Roger that.

package/MDAnalysis/coordinates/chain.py
```diff
@@ -621,57 +614,3 @@ def __repr__(self):
...
-            list of all the transformations that will be applied to the coordinates
-
-        See Also
-        --------
-        :mod:`MDAnalysis.transformations`
-        """
-        #Overrides :meth:`~MDAnalysis.coordinates.base.ProtoReader.add_transformations`
-        #to avoid unintended behaviour where the coordinates of each frame are transformed
-        #multiple times when iterating over the trajectory.
-        #In this method, the trajectory is modified all at once and once only.
```
orbeckst, line 655: This comment suggests that there was a reason for applying transformations to individual trajectories instead of to the ChainReader itself. Did you look into this reason more closely?

package/MDAnalysis/coordinates/base.py
richardjgowers: I changed the method to be an `abc.abstractmethod` so it's impossible to reach this code now

orbeckst:
Is it realistic to have this in 2.5 or can we move it to 2.6?

IAlibay:
Moving to a 2.6.0 target

package/MDAnalysis/coordinates/chain.py
richardjgowers: test coverage for this sort of behaviour seems good, see: https://github.com/MDAnalysis/mdanalysis/blob/develop/testsuite/MDAnalysisTests/coordinates/test_chainreader.py#L138

package/MDAnalysis/coordinates/chain.py
```diff
@@ -159,7 +158,7 @@ def check_allowed_filetypes(readers, allowed):
                                   "supported for formats: {}".format(allowed))
 
 
-class ChainReader(base.ProtoReader):
+class ChainReader(base.ReaderBase):
```
richardjgowers, line 161: this line change is what's happening in this PR.  Previously by subclassing ProtoReader (which is a more distant ancestor) you had to also reimplement how transformations worked.  

Instead ChainReader now subclasses `ReaderBase` (like a "normal" reader) and now only implements `_read_next_timestep` (like a "normal" Reader).  This means that methods `__next__`, `copy`, `add_transformations` are all done by the base class.

tl;dr less code = less bugs

richardjgowers:
I'm not sure what the pylint failure means, it's not a line I've touched so I'm not going to try and understand it

package/MDAnalysis/coordinates/base.py
```diff
@@ -614,7 +613,7 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         return False  # do not suppress exceptions
 
 
-class _Readermeta(type):
+class _Readermeta(abc.ABCMeta):
```
IAlibay, line 616: re: pylint failure, do you not need to do a super.__init__ call or something on line 625?

IAlibay (dismissed):
Unfortunately pylint errors are not on diff - we'll need to fix it (or mute it) _before_ merge.

package/MDAnalysis/coordinates/base.py
richardjgowers: Honestly code like this is very much in the "if it ain't broke don't fix it" category for me

IAlibay: 🤷🏽 then mute it I guess?

richardjgowers: So remove the linter?

IAlibay: no there's a # comment you can use - it's in the pylint docs iirc

hmacdope (approved):
Great work, thanks @richardjgowers!

PR #3914, opened 2022-11-10
Add distopia!

### LINKED ISSUE #3783: Integrate the result of the Distopia project into MDAnalysis
Over in the `Distopia` project https://github.com/MDAnalysis/distopia we have created a bunch of code for calculating distances efficiently under periodic boundary conditions using SIMD.  We should take the fruits of our labour and incorporate it into MDA! 

As such this issue is a long term reminder to integrate with the result. My current thoughts are to have `distopia` as an optional dependency for MDAnalysis as a drop in monkey-patch for some of the distance code.  

This is also a good place to discuss requirements for distopia integration on both the MDA and distopia side. Currently distopia uses some C++ 17 features which I think we will have to trim. It also requires a CMake build system. I am not so sure how well that integrates with Python packaging but perhaps `scikit-build` can help with this aspect.  I also suspect we will only be able to support  x86_64 SIMD at least in the short term 

@richardjgowers and myself have also been discussing a potential re-work or partial fork of distopia to encourage simplicity and reduce the overall size and complexity of the package. I would love to hear peoples opinions and will keep this thread updated with any important discussion points.

hmacdope:
Fixes #3783 

Following bump of conda-forge feedstock of `distopia` to 0.2.0, including the inplace API we can now integrate distopia into MDAnalysis. 

- Note that currently only calc-bonds is implemented. 
- Note that a kludge is required to get the input array to the right datatype.  

 - Adds distopia to MDA distances library, 

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


orbeckst:
Do we have ASV benchmarks that could show the before/after?

hmacdope:
> Do we have ASV benchmarks that could show the before/after?

I'll run them today. :)

orbeckst:
I was more thinking about having a benchmark for `calc_bonds` and friends added to `benchmarks` in a separate PR. This this PR can extend them.

IAlibay (changes requested):
I assume this is in the works since it's WIP, but since discussion is happening:

Needs:
  - A CI runner (can just use extra conda deps argument to add a py3.10 runner with distopia)
    - My view is that we don't need to do a full sweep here, distopia already tests relevant OS/Python versions, so it shouldn't be needed here.
  - Docs

.github/actions/setup-deps/action.yaml
```diff
@@ -111,6 +113,7 @@ runs:
         CONDA_OPT_DEPS: |
           ${{ inputs.chemfiles-python }}
           ${{ inputs.clustalw }}
+          ${{ inputs.distopia }}
```
IAlibay, line 120: As per my previous comment, can we not do this please? Distopia is already tested upstream so we'd want to keep testing our full deps CI to be what's in core's lib for now. Please just add a separate runner with the extra conda dep defined

azure-pipelines.yml
```diff
@@ -109,6 +109,7 @@ jobs:
       python -m pip install
       biopython
       chemfiles>=0.10
+      distopia>=0.2.0
```
IAlibay: As above, until distopia becomes a core requirement it shouldn't be in our base Windows tests otherwise you'll end up not testing what's currently in core against windows

IAlibay (changes requested):

.github/actions/setup-deps/action.yaml
hmacdope: Sorry misinterpreted your CI comment.

maintainer/conda/environment.yml
```diff
@@ -6,6 +6,7 @@ dependencies:
   - chemfiles>=0.10
   - codecov
   - cython
+  - distopia>=0.2.0
```
IAlibay: Is distopia actually required for docs?

IAlibay (changes requested):

hmacdope:
> I was more thinking about having a benchmark for `calc_bonds` and friends added to `benchmarks` in a separate PR. This this PR can extend them.

@orbeckst there is a benchmark for calc_bonds, but only handles the `no-box` case. I think there is an open PR to add PBC based distance benchmarks in #3475.

hmacdope:
See discussion  [on the distopia issue](https://github.com/MDAnalysis/mdanalysis/issues/3783#issuecomment-1312842770) about whether it should be  drop in replacement or selectable backend.

.github/actions/setup-deps/action.yaml
hmacdope: @IAlibay sorry I am a little confused. Is this not in the right spot under `CONDA_OPT_DEPS`?

hmacdope:
@richardjgowers, @orbeckst I have run some timings comparing distopia, MDA bindings to distopia and also to  MDTraj/freud. 

First thing to clear up is Richard and I had a discussion about checking the overhead of the `_run()` backend selector function and it had almost no effect from my benchmarks, so the rest of the benchmarks were run this way. 

TLDR of this whole thing:

1. Our C++ layer is inherently slow probably due to problem 2
2. Our use of **mixed precision** and various casting that needs to take place makes everything sluggish. 

Here is a comparison showing the time required to calculate N distances (x axis), fully complying to our API which takes `float` coordinates and returns `double`.  As distopia doesn't allow mixed precision, this requires we cast the input datatype to `float32` and then upcast to the result to `float64` with something that looks like this

```python
                    # need explicit branch on backend to prepare input types correctly
                    # must assign to result to get correct reference on memview
                    bondlengths = _run("calc_bonds_ortho_float",
                         args=(coords1, coords2, box[:3]),
                         kwargs={'results': bondlengths.astype(np.float32)},
                         backend=backend)
                    # upcast is currently required
                    bondlengths = bondlengths.astype(np.float64)
```

compare_dists-incast-outcast


we can see that the MDA distopia bindings (purple) are far off its theoretical peak (blue) primarily due to the input and output casting required. 

What happens if we remove the output cast which is an API break for MDA (guarantees `double` output)


compare_dists-incast-only

We can see that the performance is a lot better. 

While we can remove the input cast due to distopia requiring consistent precision, we can subtract its approximate overhead
compare_dists-simulated_zerocast

With overhead of casting approximately removed (brown line) we are very close to the distopia-only maximum speed.


Hopefully this all makes sense. 

 My main proposal is **we remove the guarantee that results are returned as `doubles` in 3.0**. If people are amenable to this we can continue to dicsuss in a new issue?

richardjgowers (commented):
LGTM.  Looking forward to going single precision in 3.0 for these reasons

hmacdope:
> LGTM. Looking forward to going single precision in 3.0 for these reasons

Issue raised #3927

.github/actions/setup-deps/action.yaml
hmacdope: Can I query this again @IAlibay? If it needs to go somewhere else I'll move it, but I followed the same process in #3765. Perhaps I need to change both.

IAlibay: Sorry I know there's a few things in my review pile, I'll try to deal with them over the next few days (i.e. before the 2.4.0 deadline)

IAlibay (changes requested):
> https://github.com/MDAnalysis/mdanalysis/actions/runs/3597359163/jobs/6059031158#step:4:129

Something's going wrong here, there's no reason there should be an incompatibility with that numpy=1.20 pin... I'm digging into it.

hmacdope:
@IAlibay @richardjgowers and I were looking into this and I'm fairly sure it's a distopia issue somewhere on the conda recipe independent of MDA.

hmacdope:
See the open issue on the feedstock.

IAlibay:
Install works fine with 1.20.3 though, let me have a look at your recipe

testsuite/MDAnalysisTests/lib/test_distances.py
```diff
@@ -825,7 +826,14 @@ def convert_position_dtype_if_ndarray(a, b, c, d, dtype):
...
 
-@pytest.mark.parametrize('backend', ['serial', 'openmp'])
+def distopia_conditional_backend():
+    # functions that allow distopia acceleration need to be tested with 
+    # distopia backend argument but distopia is an optional dep. 
+    if HAS_DISTOPIA:
+        return ['serial', 'openmp', 'distopia']
+    else:
+        return  ['serial', 'openmp']
+
```
IAlibay, line 837:
```suggestion


```

package/MDAnalysis/lib/distances.py
```diff
@@ -1369,7 +1419,7 @@ def calc_bonds(coords1: Union[np.ndarray, 'AtomGroup'],
                coords2: Union[np.ndarray, 'AtomGroup'],
                box: Optional[np.ndarray] = None,
                result: Optional[np.ndarray] = None,
-               backend: str = "serial") -> np.ndarray:
+               backend: str = "distopia" if HAS_DISTOPIA else "serial" ) -> np.ndarray:
```
IAlibay: I'm really not sure how I feel about this on a brand new addition, changing defaults based on an optional library can be quite an issue when dealing with reproducibility. For example you can write a script, give it to two folks and one of them doesn't install distopia, then you end up spending ages trying to debug why the two things don't match.

package/MDAnalysis/lib/distances.py
```diff
@@ -85,6 +124,13 @@
 from .c_distances import _minimize_vectors_ortho, _minimize_vectors_triclinic
 
 
+try:
+    import distopia
```
IAlibay: As far as I can tell you don't actually use this import directly anywhere right? Can you not just move this try/except down to 146 and just feed the _distances dictionary directly? (edit: feed not feel, you can also feel the dictionary but that might be more of a philosophy thing).

package/MDAnalysis/lib/distances.py
```diff
@@ -97,7 +143,11 @@
                                           package="MDAnalysis.lib")
 except ImportError:
     pass
-del importlib
```
IAlibay, line 100: why is this going?

package/MDAnalysis/lib/distances.py
```diff
@@ -1440,17 +1493,42 @@ def calc_bonds(coords1: Union[np.ndarray, 'AtomGroup'],
...
+                    # upcast is currently required, change for 3.0
+                    bondlengths = bondlengths.astype(np.float64)
+                else:
+                    _run("calc_bond_distance_ortho",
+                         args=(coords1, coords2, box, bondlengths),
+                         backend=backend)
             else:
+                if backend == 'distopia':
+                    # distopia does not support triclinic boxes
+                    backend_ = 'serial'
```
IAlibay: Is this documented somewhere? (I'm not seeing anything). We probably don't want to raise a warning every time it happens, but folks should be aware.

.github/actions/setup-deps/action.yaml
IAlibay: Because it's not an optional backend and you're not preventing non-distopia things to run at the same time, this isn't a problem anymore. If it wasn't a non-optional backend, then there's a specific way to create matrix additions and pass extra things that need to get installed to the "extra-conda-deps" input (which is what I was referring to).

IAlibay (changes requested):
Couple of small things, also I'm kinda shaky on the idea of optionally changing defaults. I get it, but if we do this we should make sure users are actively taking the step to install distopia (i.e. it can't be in the base MDA feedstock).

package/MDAnalysis/lib/distances.py
orbeckst: For 2.x distopia should be opt-in, just to be on the safe side.

hmacdope: Okie dokie makes sense. Thoughts on making it the default for 3.0 once it's a bit more battle tested ? I suppose we can decide later :)

IAlibay: The over-cautious person in me agrees with @orbeckst but I also know that most of our methods don't offer optional backends which means you end up using the slow stuff by default. Hence being really unsure about what the best option is :/

I'll take a passive stance on this one and go with concensus. If we go for non-default then I'm also going to propose that we check all the analysis methods and check if we can add a keyword arguments to select backends?

orbeckst: I'd be cautious with anything that can change results. I totally get (and want!) faster distance calculations for everyone. Being able to set the backend in analysis tools is a good idea. It will allow us to directly run analysis with either and compare. If we find that for typical cases the difference is small or machine-eps-like then we can consider changing the default earlier. Otherwise we have a safe and clean upgrade path.

Propagating the backend keyword would also be important for folks who run analysis on different archs — some that are supported and others that are not. They might want to ensure that exactly the same analysis backend is used everywhere.

orbeckst: I raised #3946

package/MDAnalysis/lib/distances.py
hmacdope: The aim was to have the `HAS_DISTOPIA` attr available for backend selection so I wouldn't mind keeping it in? On the other hand we are removing the `backend: str = "distopia" if HAS_DISTOPIA else "serial"` but. IDK if people feel strongly about YAGNI  I can pull it out.

package/MDAnalysis/lib/distances.py
hmacdope: mb

package/MDAnalysis/lib/distances.py
hmacdope: I added a `warning` block.

package/MDAnalysis/lib/distances.py
hmacdope: Its also needed for the tests as module level import to avoid repetition.

package/MDAnalysis/lib/distances.py
hmacdope: I reverted to default = `serial`. We can decide what we want to do for defaults with a bit more maturity in `distopia`.

hmacdope:
Not really sure how to satisfy MyPy here, tempted to ignore it.

package/MDAnalysis/lib/distances.py
```diff
@@ -96,8 +149,15 @@
     _distances['openmp'] = importlib.import_module(".c_distances_openmp",
                                           package="MDAnalysis.lib")
 except ImportError:
-    pass
-del importlib
+    del importlib
+
+if HAS_DISTOPIA:
+    import importlib
```
IAlibay: I think I'm missing something obvious, why is importlib being moved inside of the try/except here? So it only now dels it _if_ there's an import error?

Wouldn't it make more sense to do like we did before, del at the end after the attempt to import everything?

package/MDAnalysis/lib/distances.py
IAlibay: I don't understand why you need two try/excepts (which aren't exactly cheap if we assume the base case is that you'll hit the error most of the time).

Why can't we move the `HAS_DISTOPIA` assignment under with the same imporlib section on line 156? i.e.

```python
try:
    _distances["distopia"] = importlib.import_module("._distopia", package="distopia")
except:
    HAS_DISTOPIA = False
else:
    HAS_DISTOPIA = True
```

package/MDAnalysis/lib/distances.py
```diff
@@ -50,10 +50,56 @@
...
+
+MDAnalysis has developed a standalone library, `distopia`_ for accelerating
+the distance functions in this module using explicit SIMD vectorisation.
+This can provide many-fold speedups in calculating distances. Distopia is
+under active development and as such only a selection of functions in this
+module are covered. Consult the following table to see if the function
+you wish to use is covered by distopia. For more information see the
+`distopia documentation`_.
+
+.. Table:: Functions available using the `distopia`_ backend.
```
IAlibay, line 64: The table arrangement here is a bit off (it's left justified, https://mdanalysis--3914.org.readthedocs.build/en/3914/documentation_pages/lib/distances.html#id2), is there any way to pass `:align: center` to the table directive?

package/MDAnalysis/lib/distances.py
```diff
@@ -50,10 +50,56 @@
...
+
+    +-------------------------------------+
+    | Functions                           |
```
IAlibay: If a summary is being provided, I would have a second column for extra details, like the fact that it doesn't work for triclinic boxes.

package/MDAnalysis/lib/distances.py
```diff
@@ -50,10 +50,56 @@
...
+    +=====================================+
+    | MDAnalysis.lib.distances.calc_bonds |
+    +-------------------------------------+
+
+If `distopia`_ is installed, the functions in this table will accept the key
+'distopia' for the `backend` keyword argument. If the distopia backend is
+selected the `distopia` library will be used to calculate the distances. Note
+that for functions listed in this table **distopia is the default backend if it
+is available**.
```
IAlibay: Am I correct in understanding that this is no longer true?

package/MDAnalysis/lib/distances.py
```diff
@@ -1412,8 +1472,9 @@ def calc_bonds(coords1: Union[np.ndarray, 'AtomGroup'],
         Preallocated result array of dtype ``numpy.float64`` and shape ``(n,)``
         (for ``n`` coordinate pairs). Avoids recreating the array in repeated
         function calls.
-    backend : {'serial', 'OpenMP'}, optional
-        Keyword selecting the type of acceleration.
+    backend : {'serial', 'OpenMP', 'distopia'}, optional
+        Keyword selecting the type of acceleration. Distopia acceleration is
+        the default if available otherwise serial acceleration is used.
```
IAlibay:
```suggestion
        Keyword selecting the type of acceleration. Defaults to 'serial'.
```

package/MDAnalysis/lib/distances.py
```diff
@@ -1432,25 +1493,62 @@ def calc_bonds(coords1: Union[np.ndarray, 'AtomGroup'],
...
+                        "calc_bond_distance_ortho",
+                        args=(coords1, coords2, box, bondlengths),
+                        backend=backend,
+                    )
             else:
-                _run("calc_bond_distance_triclinic",
-                     args=(coords1, coords2, box, bondlengths),
-                     backend=backend)
+                if backend == "distopia":
+                    # distopia does not support triclinic boxes
```
IAlibay: Since it's currently an active choice, should we pass a warning here? (I'd be against a warning if it was the default choice, but I think if folks are trying it out / benchmarking they might want to know that they shouldn't be expecting faster performance for this case? I'm happy if the answer is no here.

IAlibay (changes requested):
We're going to have to make mypy happy otherwise it'll complain forever, I'll have a stab.

package/MDAnalysis/lib/distances.py
```diff
@@ -1432,25 +1493,62 @@ def calc_bonds(coords1: Union[np.ndarray, 'AtomGroup'],
...
+                    # need explicit branch on backend to prepare input types correctly
+                    # must assign to result to get correct reference on memview
+                    bondlengths = _run(
+                        "calc_bonds_ortho_float",
+                        args=(coords1, coords2, box[:3]),
+                        kwargs={"results": bondlengths.astype(np.float32)},
+                        backend=backend,
+                    )
+                    # upcast is currently required, change for 3.0
+                    bondlengths = bondlengths.astype(np.float64)
```
richardjgowers: rather than have extra branches here, would it make more sense to have a stub `._distopia.py` which does the logic in this branch.

this stub could also redirect calls that aren't currently supported in distopia (anything but calc_bonds) back to the "serial" backend

IAlibay: Yeah I'm +1ing this - your mypy issues are mostly down to trying to shoehorn in a Callable into an ndarray. It would be a lot better if the behaviour of the `_run` calls was the same across all the backends, rather than have some that return things and some that do inplace changes to bondlengths.

package/MDAnalysis/lib/distances.py
hmacdope: Yep i'll add a warning.

package/MDAnalysis/lib/distances.py
hmacdope: Done now with a stub

hmacdope:
Hang on some issues with passing through args and returning the right thing

package/MDAnalysis/lib/distances.py
```diff
@@ -97,8 +145,15 @@
                                           package="MDAnalysis.lib")
 except ImportError:
     pass
-del importlib
 
+if HAS_DISTOPIA:
+    try:
+        _distances["distopia"] = importlib.import_module(
```
IAlibay: If you've already verified that the module is available you don't need to wrap it around a try/except?

package/MDAnalysis/lib/distances.py
```diff
@@ -1432,25 +1489,33 @@ def calc_bonds(coords1: Union[np.ndarray, 'AtomGroup'],
     .. versionchanged:: 2.3.0
        Can now accept an :class:`~MDAnalysis.core.groups.AtomGroup` as an
        argument in any position and checks inputs using type hinting.
+    .. versionchanged:: 2.4.0
```
IAlibay:
```suggestion
    .. versionchanged:: 2.5.0
```

package/MDAnalysis/lib/distances.py
```diff
@@ -50,10 +50,57 @@
...
+.. _ISA: https://en.wikipedia.org/wiki/Instruction_set_architecture
+.. _VCL2: https://github.com/vectorclass/version2
+.. _relevant distopia issue: https://github.com/MDAnalysis/mdanalysis/issues/3915
+
 .. versionadded:: 0.13.0
 .. versionchanged:: 2.3.0
    Distance functions can now accept an
    :class:`~MDAnalysis.core.groups.AtomGroup` or an :class:`np.ndarray`
+.. versionchanged:: 2.5.0
+   Interface to the `distopia` package added.
```
IAlibay:
```suggestion
   Interface to the `distopia`_ package added.
```

Use the link directly here?

testsuite/MDAnalysisTests/lib/test_distances.py
```diff
@@ -1022,23 +1052,58 @@ def test_dihedrals_wronglength(self, positions, wronglength, backend):
...
+                0.25 * np.pi,
+            ),  # 45 degree
+            (
+                np.array(
+                    [[1, 2, 1], [1, 1, 1], [2, 1, 1], [2, 0, 2]], dtype=np.float32
+                ),
+                0.75 * np.pi,
+            ),  # 135
+        ],
+    )
```
IAlibay, line 1103: Is there actually any change here except from conforming to `black`? this 100% makes readability worse, I'd ask that we go back to the previous formatting even style if `black` complains.

testsuite/MDAnalysisTests/lib/test_distances.py
```diff
@@ -963,19 +977,33 @@ def test_angles(self, backend, dtype, pos, request):
...
+            (
+                np.array([[1, 1, 1], [1, 2, 1], [1, 3, 1]], dtype=np.float32),
+                np.pi,
+            ),  # straight line / 180.
+            (
+                np.array([[1, 1, 1], [1, 2, 1], [2, 1, 1]], dtype=np.float32),
+                0.25 * np.pi,
+            ),  # 45
+        ],
+    )
```
IAlibay, line 1003: Here again, another example of `black` making it much harder to read stuff

IAlibay (changes requested):
Some initial comments.

Btw, we don't have to follow the `black` part of darken, it's more of an aid than anything else.

package/MDAnalysis/lib/distances.py
```diff
@@ -1432,26 +1485,39 @@ def calc_bonds(coords1: Union[np.ndarray, 'AtomGroup'],
...
 
     if numatom > 0:
         if box is not None:
             boxtype, box = check_box(box)
-            if boxtype == 'ortho':
-                _run("calc_bond_distance_ortho",
-                     args=(coords1, coords2, box, bondlengths),
-                     backend=backend)
+            if boxtype == "ortho":
+                if backend == 'distopia':
```
hmacdope, line 1492: I know this is still messy @IAlibay @richardjgowers and "should" be pushable to the stub. I'm probably missing something super obvious but IMO there is a reason it couldn't be in there. 
- All the **other** distance functions use a pass by reference idiom without assignment and this is reflected in the `run` call as they don't return anything.
-  any use of `.astype()` on the `results` arg forces a copy in the stub function and then we need to return the result. 
- Then the pattern used in `run()` becomes non-consistent with some functions return values while others modify them inplace. making life hell.

What am I missing? As I see it there isn't a way to push the casting into the stub function??? Am I stoopid?

hmacdope: Possibly ping @lilyminium @orbeckst if anyone has any ideas. It could just be too late at night.

hmacdope:
I think this should be good for re-review

package/MDAnalysis/lib/distances.py
IAlibay: Was this addressed in the end?

IAlibay (approved):
aside from the one thing that I'm not sure if it got resolved - lgtm! (sorry about the delay here)

Issue #2497, opened 2020-02-03
Using pathlib Path object


mimischi:
**Expected behavior**

Providing a `PosixPath` from `pathlib` to `mda.Universe` should create the Universe. Instead, we currently have to wrap the `PosixPath` object in a `str()` call.

**Actual behavior**

Code throws an `AttributeError`:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-154-4c04e03d6c7d> in <module>
      2 gro = base_path / "ref.gro"
      3 xtc = base_path / "trjcat_p1p10.mc.xtc"
----> 4 u = mda.Universe(gro, xtc)

~/.conda/envs/mda020/lib/python3.7/site-packages/MDAnalysis/core/universe.py in __init__(self, *args, **kwargs)
    335                 coordinatefile = None
    336 
--> 337             self.load_new(coordinatefile, **kwargs)
    338             # parse transformations
    339             trans_arg = kwargs.pop('transformations', None)

~/.conda/envs/mda020/lib/python3.7/site-packages/MDAnalysis/core/universe.py in load_new(self, filename, format, in_memory, **kwargs)
    602         kwargs['n_atoms'] = self.atoms.n_atoms
    603 
--> 604         self.trajectory = reader(filename, **kwargs)
    605         if self.trajectory.n_atoms != len(self.atoms):
    606             raise ValueError("The topology and {form} trajectory files don't"

~/.conda/envs/mda020/lib/python3.7/site-packages/MDAnalysis/coordinates/XDR.py in __init__(self, filename, convert_units, sub, refresh_offsets, **kwargs)
    134                                             convert_units=convert_units,
    135                                             **kwargs)
--> 136         self._xdr = self._file(self.filename)
    137 
    138         self._sub = sub

MDAnalysis/lib/formats/libmdaxdr.pyx in MDAnalysis.lib.formats.libmdaxdr._XDRFile.__cinit__()

AttributeError: 'PosixPath' object has no attribute 'encode'
```

**Code to reproduce the behavior**

```
from pathlib import Path 
import MDAnalysis as mda

base_path = Path(".")
gro = base_path / "example.gro"

u = mda.Universe(gro)
```

**Currently version of MDAnalysis**

- Which version are you using? 0.20.1
- Which version of Python (`python -V`)? 3.7.3
- Which operating system? Scientific Linux

orbeckst:
[...]

PR #3935, opened 2022-11-22
ENH: support pathlib in ReaderBase

tylerjereddy:
* the creation of a `Universe` object involving GROMACS formats should now accept `pathlib`
objects as requested in gh-2497; resolving that issue entirely is a separate matter, this branch focuses only on the GROMACS format and only if you proceed via the public `Universe` creation API

* probably could make it work in Cython directly, though I'm not convinced we need to do that for this part of the fix at least, and we probably should still support strings as well of course

### DIFF
--- a/package/MDAnalysis/coordinates/XDR.py
+++ b/package/MDAnalysis/coordinates/XDR.py
@@ -145,7 +145,7 @@ class XDRBaseReader(base.ReaderBase):
         super(XDRBaseReader, self).__init__(filename,
                                             convert_units=convert_units,
                                             **kwargs)
-        self._xdr = self._file(str(self.filename))
+        self._xdr = self._file(self.filename)
 
         self._sub = sub
         if self._sub is not None:
--- a/package/MDAnalysis/coordinates/XDR.py
+++ b/package/MDAnalysis/coordinates/XDR.py
@@ -145,7 +145,7 @@ class XDRBaseReader(base.ReaderBase):
         super(XDRBaseReader, self).__init__(filename,
                                             convert_units=convert_units,
                                             **kwargs)
-        self._xdr = self._file(self.filename)
+        self._xdr = self._file(str(self.filename))
 
         self._sub = sub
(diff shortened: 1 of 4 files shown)


IAlibay (commented):
Shouldn't this be a ReaderBase thing instead to store self.filename properly?

richardjgowers (approved):

hmacdope:
> Shouldn't this be a ReaderBase thing instead to store self.filename properly?

Seconding this, can this not be dealt with at `ReaderBase` level to fix it for all readers?

tylerjereddy:
Ok, I moved the patch to `ReaderBase` and added a test for one more format (`DCD`) as a sniff test for generalization.

I think that's probably good enough for now, but note that:

1) This still doesn't cover all readers because `SingleFrameReaderBase` has its own `self.filename` -- I'd be happy to delay patching/testing that for now.
2) The testing could probably be centralized/parametrized somewhere, covering a bunch of formats rather than testing the same pass-through for just two formats. Maybe good enough for now though..

IAlibay (approved):
Assuming this passes tests I'm happy for this PR to go through. @tylerjereddy could you raise an issue about the missing bits? Maybe we should move stuff to ProtoReader in the future but it seems fine for this current issue?

tylerjereddy:
Eh, the failure is related--this is the problem with inheritance... I'd just as soon scope this back to GROMACS and move on.

I'll see if I can add a shim to fix it though.

tylerjereddy:
Ok, I pushed in a commit to shim around our `NamedStream` stuff in the base class, and this seems to allow the full suite to pass for me locally at least.

gh-3937 contains the the follow-up reminders I noted above.

hmacdope:
LGTM @tylerjereddy

hmacdope (approved):

tylerjereddy:
Ok, I have approvals from 3 core developers and CI is passing, in it goes.

Issue #3947, opened 2022-12-04
Add support for pickling of ResidueGroup and SegmentGroup


pgbarletta:
## Is your feature request related to a problem? ##

Currently, `AtomGroup` can be pickled but `ResidueGroup` and `SegmentGroup` can't.

```
import pickle
import MDAnalysis as mda
from MDAnalysis.tests.datafiles import TPR, XTC

u = mda.Universe(TPR, XTC)

with open('atoms.pkl', 'wb') as f:        # this works
    pickle.dump(u.atoms, f)
with open('residues.pkl', 'wb') as f:        # this doesn't
    pickle.dump(u.residues, f)
with open('segments.pkl', 'wb') as f:        # neither does this
    pickle.dump(u.segments, f)
```

## Describe the solution you'd like ##

Add a general pickling support to `BaseGroup`.

## Describe alternatives you've considered ##

Add pickling support to `ResidueGroup` and `SegmentGroup` separately as `AtomGroup` has, [here](https://github.com/MDAnalysis/mdanalysis/blob/develop/package/MDAnalysis/core/groups.py#L118), and [here](https://github.com/MDAnalysis/mdanalysis/blob/develop/package/MDAnalysis/core/groups.py#L2528).

orbeckst:
This issue was closed with PR #3953 .

PR #3948, opened 2022-12-06
Add ResidueGroup and SegmentGroup pickling.

pgbarletta:
From #3947

Maybe the tests could be better? I'll update the changelog if these changes are ok.

- Add __reduce__() method to both ResidueGroup and SegmentGroup
- Change _unpickle() so it can create any kind of group from its indices. 
- Add one test for each group in the parallelism section.

PR Checklist
------------
 - [X] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/.github/actions/setup-deps/action.yaml
+++ b/.github/actions/setup-deps/action.yaml
@@ -71,9 +71,9 @@ inputs:
     default: 'tidynamics>=1.0.0'
   # pip-installed min dependencies
   coverage:
-    default: 'coverage<5.0'
+    default: 'coverage'
   pytest-cov:
-    default: 'pytest-cov<=2.10.1'
+    default: 'pytest-cov'
   pytest-xdist:
     default: 'pytest-xdist'
   # pip-install optional dependencies


IAlibay:
Huh that's a new codecov issue, hopefully it's transient

richardjgowers (commented):
this is very cool, can we also include `Residue` `Atom` and `Segment` (since they would work in the same way)?

testsuite/MDAnalysisTests/parallelism/test_pickle_transformation.py
```diff
@@ -151,3 +151,17 @@ def test_add_unwrap_pickle(unwrap_transformation, u):
...
+    assert_almost_equal(u.atoms.positions, atm.positions)
+
+
+def test_residuegroup_pickle(u):
+    res = pickle.loads(pickle.dumps(u.residues[10:20]))
+    assert_almost_equal(u.residues[10:20].universe.atoms.positions, res.universe.atoms.positions)
+
+def test_segmentgroup_pickle(u):
+    seg = pickle.loads(pickle.dumps(u.segments[0:]))
+    assert_almost_equal(u.segments[0:].universe.atoms.positions, seg.universe.atoms.positions)
```
richardjgowers: I'm not sure this is the right file, if there's not an existing suitable file you could always make `test_core_pickling.py`

package/MDAnalysis/core/groups.py
```diff
@@ -115,8 +115,8 @@
 from ._get_readers import get_writer_for, get_converter_for
 
 
-def _unpickle(u, ix):
-    return u.atoms[ix]
+def _unpickle(cls, ix, u):
+    return cls(ix, u)
```
orbeckst: Does this change affect unpickling of pickles that were created with earlier versions?

(We don't encourage pickling universes to save state but it's allowed. If it breaks old pickles then it would be a *backward-incompatible* change.)

orbeckst: It's neat.

testsuite/MDAnalysisTests/parallelism/test_pickle_transformation.py
```diff
@@ -151,3 +151,17 @@ def test_add_unwrap_pickle(unwrap_transformation, u):
     u.trajectory[0]
     for u_ts, u_p_ts in zip(u.trajectory[:5], u_p.trajectory[:5]):
         assert_almost_equal(u_ts.positions, u_p_ts.positions)
+
+
+def test_atomgroup_pickle(u):
```
orbeckst: move to core/test_atomgroup.py

testsuite/MDAnalysisTests/parallelism/test_pickle_transformation.py
```diff
@@ -151,3 +151,17 @@ def test_add_unwrap_pickle(unwrap_transformation, u):
...
+    atm = pickle.loads(pickle.dumps(u.atoms))
+    assert_almost_equal(u.atoms.positions, atm.positions)
+
+
+def test_residuegroup_pickle(u):
```
orbeckst: move to core/test_resduegroup.py

testsuite/MDAnalysisTests/parallelism/test_pickle_transformation.py
orbeckst: This should go into core/test_segmentgroup.py

orbeckst (changes requested):
Neat idea but see comments inline.

Please also update CHANGELOG and AUTHORS.

IAlibay:
? what happened here @pgbarletta

pgbarletta:
Sorry, I messed up. I think I ran `fetch` `rebase` and `reset` in the wrong order so for a moment my PR was 0 commits ahead of `develop` and that triggered the PR closure. I reopened with a better name, though. [Here](https://github.com/MDAnalysis/mdanalysis/pull/3953).

testsuite/MDAnalysisTests/parallelism/test_pickle_transformation.py
pgbarletta: Fixed on #3953.

testsuite/MDAnalysisTests/parallelism/test_pickle_transformation.py
pgbarletta: Fixed on #3953.

package/MDAnalysis/core/groups.py
pgbarletta: You're absolutely right. I rewrote it to make it backwards compatible with a warning.
Fixed on #3953.

testsuite/MDAnalysisTests/parallelism/test_pickle_transformation.py
pgbarletta: Fixed on #3953.

PR #3952, opened 2022-12-07
Replace deprecated pytest.warns(None)

IAlibay:
That was the only other deprecation warning being thrown (at least from my local checks)

 - replace pytest.warns(None) calls with the recommended approach from https://docs.pytest.org/en/latest/how-to/capture-warnings.html#additional-use-cases-of-warnings-in-tests

PR Checklist
------------
 - [x] Tests?
 - Docs?
 - CHANGELOG updated?
 - Issue raised/referenced?

### DIFF
--- a/testsuite/MDAnalysisTests/converters/test_rdkit.py
+++ b/testsuite/MDAnalysisTests/converters/test_rdkit.py
@@ -398,7 +398,8 @@ class TestRDKitConverter(object):
         with pytest.warns(UserWarning, match="Could not sanitize molecule"):
             u.atoms.convert_to.rdkit(NoImplicit=False)
         with warnings.catch_warnings():
-            warnings.simplefilter("error")
+            warnings.filterwarnings("error", "Could not sanitize molecule")
+            warnings.warn("Could not sanitize molecule: failed during step")
             u.atoms.convert_to.rdkit()
 
 
--- a/testsuite/MDAnalysisTests/converters/test_rdkit.py
+++ b/testsuite/MDAnalysisTests/converters/test_rdkit.py
@@ -399,7 +399,6 @@ class TestRDKitConverter(object):
             u.atoms.convert_to.rdkit(NoImplicit=False)
         with warnings.catch_warnings():
             warnings.filterwarnings("error", "Could not sanitize molecule")
-            warnings.warn("Could not sanitize molecule: failed during step")
             u.atoms.convert_to.rdkit()
 
 
--- a/testsuite/MDAnalysisTests/converters/test_rdkit.py
+++ b/testsuite/MDAnalysisTests/converters/test_rdkit.py
@@ -23,6 +23,7 @@
 
 import copy
 from io import StringIO
+import warnings
 import pytest
 import MDAnalysis as mda
 from MDAnalysis.topology.guessers import guess_atom_element
--- a/testsuite/MDAnalysisTests/converters/test_rdkit.py
+++ b/testsuite/MDAnalysisTests/converters/test_rdkit.py
@@ -400,9 +400,6 @@ class TestRDKitConverter(object):
         with warnings.catch_warnings():
             warnings.simplefilter("error")
             u.atoms.convert_to.rdkit()
-        if record:
-            assert all("Could not sanitize molecule" not in str(w.message)
-                       for w in record.list)
 
 
 @requires_rdkit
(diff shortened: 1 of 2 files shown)


testsuite/MDAnalysisTests/converters/test_rdkit.py
```diff
@@ -255,9 +255,9 @@ def test_error_no_hydrogen(self, uo2):
             uo2.atoms.convert_to("RDKIT")
 
     def test_error_no_hydrogen_implicit(self, uo2):
-        with pytest.warns(None) as record:
+        with warnings.catch_warnings():
+            warnings.simplefilter("error")
             uo2.atoms.convert_to.rdkit(NoImplicit=False)
-        assert len(record) == 0
```
orbeckst, line 260: It's a bit odd not to assert anything here but I also don't know what should be asserted if we just try to get through the code path.

orbeckst (commented):
lgtm

orbeckst (approved):

orbeckst (changes requested):
Well, I should have waited for the tests... something is not right,
```
FAILED testsuite/MDAnalysisTests/converters/test_rdkit.py::TestRDKitConverter::test_error_no_hydrogen_implicit - NameError: name 'warnings' is not defined
```

orbeckst:
Argh, why is there now an RDKit failure? [TestRDKitConverter.test_sanitize_fail_warning](https://github.com/MDAnalysis/mdanalysis/actions/runs/3649646773/jobs/6164593327#step:7:521):
```
FAILED testsuite/MDAnalysisTests/converters/test_rdkit.py::TestRDKitConverter::test_sanitize_fail_warning - UserWarning: NaN detected in coordinates, the output molecule will not have 3D coordinates assigned
```

testsuite/MDAnalysisTests/converters/test_rdkit.py
```diff
@@ -396,7 +397,8 @@ def test_sanitize_fail_warning(self):
         u = mda.Universe(mol)
         with pytest.warns(UserWarning, match="Could not sanitize molecule"):
             u.atoms.convert_to.rdkit(NoImplicit=False)
-        with pytest.warns(None) as record:
+        with warnings.catch_warnings():
+            warnings.simplefilter("error")
             u.atoms.convert_to.rdkit()
         if record:
             assert all("Could not sanitize molecule" not in str(w.message)
```
orbeckst: This needs to go, too, there's no `record` in the new code.

IAlibay:
> Argh, why is there now an RDKit failure? [TestRDKitConverter.test_sanitize_fail_warning](https://github.com/MDAnalysis/mdanalysis/actions/runs/3649646773/jobs/6164593327#step:7:521):
> 
> ```
> FAILED testsuite/MDAnalysisTests/converters/test_rdkit.py::TestRDKitConverter::test_sanitize_fail_warning - UserWarning: NaN detected in coordinates, the output molecule will not have 3D coordinates assigned
> ```

I'll have a look in a couple hours, I suspect it's likely my fault 😰

testsuite/MDAnalysisTests/converters/test_rdkit.py
```diff
@@ -396,7 +397,8 @@ def test_sanitize_fail_warning(self):
         u = mda.Universe(mol)
         with pytest.warns(UserWarning, match="Could not sanitize molecule"):
             u.atoms.convert_to.rdkit(NoImplicit=False)
-        with pytest.warns(None) as record:
+        with warnings.catch_warnings():
+            warnings.simplefilter("error")
```
orbeckst: Is the idea that this block should pass without a warning but if it does then the test fails?

Well, the test fails, and I don't know why — this seems to be behavior that we masked previously.

orbeckst (changes requested):
An unexpected failure of an RDKit test that should have passed without a warning.

testsuite/MDAnalysisTests/converters/test_rdkit.py
IAlibay: So here I see it as the simplefilter is a type of assert (i.e. it asserts that there are no warnings)

testsuite/MDAnalysisTests/converters/test_rdkit.py
IAlibay: It's because we are expecting _some_ warnings just not that specific warning we're checking the record for. I've changed the simplefilter to a filterwarnings that should do the trick instead

orbeckst (approved):

PR #3953, opened 2022-12-08
Add pickling of Atom, Residue, Segment and groups.

pgbarletta:
From #3947

- Add reduce() method to both ResidueGroup and SegmentGroup 
- Change _unpickle() so it can create any kind of group from its indices. 
- Add one test for each group in the parallelism section.

PR Checklist
------------
 - [X] Tests?
 - [x] Docs?
 - [X] CHANGELOG updated?

PD: sorry for the new PR. I synced up my fork from develop and that somehow closed the previous PR.

### DIFF
--- a/package/AUTHORS
+++ b/package/AUTHORS
@@ -200,6 +200,7 @@ Chronological list of authors
   - Jennifer A Clark
   - Jake Fennick
   - Utsav Khatu
+  - Patricio Barletta
 
 
 External code
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -13,13 +13,16 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" http://semver.org
 
 ------------------------------------------------------------------------------
-??/??/?? IAlibay
+??/??/?? IAlibay, pgbarletta
 
  * 2.5.0
 
 Fixes
 
 Enhancements
+
+* Add pickling support for Atom, Residue, Segment, ResidueGroup 
+    and SegmentGroup. (PR #3953)
 
 Changes
 
--- a/package/MDAnalysis/core/groups.py
+++ b/package/MDAnalysis/core/groups.py
@@ -119,6 +119,11 @@ def _unpickle(u, ix):
     return u.atoms[ix]
 
 
+# TODO 3.0: deprecate _unpickle in favor of _unpickle2.
+def _unpickle2(u, ix, cls):
+    return cls(ix, u)
+
+
 def _unpickle_uag(basepickle, selections, selstrs):
     bfunc, bargs = basepickle[0], basepickle[1:][0]
     basegroup = bfunc(*bargs)
@@ -2525,9 +2530,6 @@ class AtomGroup(GroupBase):
        Indexing an AtomGroup with ``None`` raises a ``TypeError``.
     """
 
-    def __reduce__(self):
-        return (_unpickle, (self.universe, self.ix))
-
     def __getattr__(self, attr):
         # special-case timestep info
         if attr in ('velocities', 'forces'):
@@ -2535,6 +2537,9 @@ class AtomGroup(GroupBase):
         elif attr == 'positions':
             raise NoDataError('This Universe has no coordinates')
         return super(AtomGroup, self).__getattr__(attr)
+
+    def __reduce__(self):
+        return (_unpickle, (self.universe, self.ix))
 
     @property
     def atoms(self):
@@ -3655,6 +3660,9 @@ class ResidueGroup(GroupBase):
        Indexing an ResidueGroup with ``None`` raises a ``TypeError``.
     """
 
+    def __reduce__(self):
+        return (_unpickle2, (self.universe, self.ix, ResidueGroup))
+
     @property
     def atoms(self):
         """An :class:`AtomGroup` of :class:`Atoms<Atom>` present in this
@@ -3847,6 +3855,9 @@ class SegmentGroup(GroupBase):
     .. versionchanged:: 2.1.0
        Indexing an SegmentGroup with ``None`` raises a ``TypeError``.
     """
+
+    def __reduce__(self):
+        return (_unpickle2, (self.universe, self.ix, SegmentGroup))
 
     @property
     def atoms(self):
@@ -4140,6 +4151,9 @@ class Atom(ComponentBase):
             me += ' and altLoc {}'.format(self.altLoc)
         return me + '>'
 
+    def __reduce__(self):
+        return (_unpickle2, (self.universe, self.ix, Atom))
+
     def __getattr__(self, attr):
         # special-case timestep info
         ts = {'velocity': 'velocities', 'force': 'forces'}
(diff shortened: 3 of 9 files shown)


package/MDAnalysis/core/groups.py
```diff
@@ -115,8 +115,15 @@
...
-    return u.atoms[ix]
+def _unpickle(u, ix, cls=None):
+    if cls:
+        group = cls(ix, u)
+    else:
+        group = u.atoms[ix]
+        warnings.warn("You may be trying to read an AtomGroup that was pickled "
+        "with a pre 2.4 MDAnalysis version. If so, rewrite the pickle file to "
+        "silence this warning. More info at issue #3947 and PR #3953.", DeprecationWarning)
+    return group
```
richardjgowers: so a different solution to this would be to keep AtomGroup pickling going through this function (and not break anything) and create a new function (`_unpickle2`) which has the new signature.  This can then get tidied in 3.0 (which we *can* break compatibility with).

then you won't have the untested branch in the code

testsuite/MDAnalysisTests/core/test_segmentgroup.py
```diff
@@ -109,3 +111,16 @@ def test_set_segids_many():
 def test_atom_order(universe):
     assert_equal(universe.segments.atoms.indices,
                  sorted(universe.segments.atoms.indices))
+
+class TestSegmentGroupPickle(object):
+    """Test SegmentGroup pickling support."""
+    @pytest.fixture()
+    def universe(self):
+        return mda.Universe(TPR, XTC)
```
richardjgowers: minor quibble, but if you use `Universe.empty()` it'll make the tests a little faster.  it's not a big deal, but over 1,000 tests it adds up slowly

richardjgowers (changes requested):
can we use empty Universes?

testsuite/MDAnalysisTests/core/test_segmentgroup.py
pgbarletta: Sorry, don't get it. Say I do:
```
seg_group = mda.Universe.empty(10).segments
seg = pickle.loads(pickle.dumps(seg_group))
```

how do I compare them? And would that be a useful test?

package/MDAnalysis/core/groups.py
pgbarletta: Done.

testsuite/MDAnalysisTests/core/test_segmentgroup.py
richardjgowers: you could check that the new segments have the same universe as the old one:
```python
u = mda.Universe.empty()

segs = u.segments

new = pickle.loads(...)

assert new.universe is u
```

I think maybe you're right, a better test would have dummy data inside the Universe...
```python
u = mda.Universe.empty()
u.add_TopologyAttr('segids', values=['A', 'B', etc]

... some pickling

assert new.segids == ['A', 'B', etc]
```

The idea is you can create a Universe without using files, and it'll make the tests faster if we don't (and also makes the tests not reliant on TPRParser working)

pgbarletta: thanks. I went with the 2nd choice:
```
def test_segmentgroup_pickle():
    u = mda.Universe.empty(10)
    u.add_Segment(segid="X")
    u.add_Segment(segid="Y")
    u.add_Segment(segid="Z")
    segids = ["A", "X", "Y", "Z"]
    u.add_TopologyAttr("segids", values=["A", "X", "Y", "Z"])
    seg_group = mda.SegmentGroup((1, 3), u)
    seg = pickle.loads(pickle.dumps(seg_group))
    assert_equal(seg.universe.segments.segids, segids)
```

richardjgowers (approved):

package/MDAnalysis/core/groups.py
```diff
@@ -119,6 +119,10 @@ def _unpickle(u, ix):
     return u.atoms[ix]
 
 
+def _unpickle2(u, ix, cls):
```
hmacdope, line 123: Just one little nitpick, would you be able to document in a quick comment why the two are necessary for someone following along later?

package/CHANGELOG
```diff
@@ -39,6 +39,8 @@ Fixes
   * fixes guessed_attributes and read_attributes methods of the Topology class, as 
     those methods were giving unexpected results for some topology attributes
     (e.g. bonds, angles) (PR #3779).
+  * Add pickling support for Atom, Residue, Segment, ResidueGroup
```
hmacdope: Entries newest first and this needs to be put under 2.5.0. Merging with upstream/develop also required. :)

hmacdope (changes requested):
Two little things and then I think good to go.

package/AUTHORS
```diff
@@ -199,6 +199,7 @@ Chronological list of authors
   - Haleema Khan
   - Jennifer A Clark
   - Jake Fennick
+  - Patricio Barletta
```
hmacdope: Nitpick but you should be last

hmacdope (changes requested):
Two tiny nitpicks sorry.

package/CHANGELOG
```diff
@@ -21,6 +21,9 @@ Fixes
 
 Enhancements
 
+* Add pickling support for Atom, Residue, Segment, ResidueGroup
```
hmacdope, line 24: Add your self to 2.5.0 authors to the right of IAlibay. :)

pgbarletta:
I'll be out for a ~week. I'll rebase once the numpy deprecations issues are fixed.

hmacdope (approved):
LGTM thanks @pgbarletta

package/CHANGELOG
```diff
@@ -21,12 +21,15 @@ Fixes
...
+* Add pickling support for Atom, Residue, Segment, ResidueGroup 
+    and SegmentGroup. (PR #3953)
+
 Changes
 
 Deprecations
 
 
-12/21/22 IAlibay
+12/21/22 IAlibay, pgbarletta
```
```suggestion
12/21/22 IAlibay
```
Not a 2.4.2 change? (edit: I think the intention here was to add this to line 16 instead?)

IAlibay (changes requested):

package/CHANGELOG
pgbarletta: Fixed.

package/MDAnalysis/core/groups.py
```diff
@@ -119,6 +119,11 @@ def _unpickle(u, ix):
     return u.atoms[ix]
 
 
+# TODO 3.0: deprecate _unpickle in favor of _unpickle2.
```
IAlibay, line 122: Just a comment here, unless it changes user-facing behaviour there is no need to deprecate a private method like this. We can just throw it out right now and none would be the wiser.

Please raise a relevant issue regarding this though.

IAlibay (approved):
I'll approve although please do raise the issue for the follow-up 3.0 work, otherwise it'll get lost.

package/MDAnalysis/core/groups.py
pgbarletta: Done.

PR #3957, opened 2022-12-13
Add more detail to DCD timestep deprecation message

### LINKED ISSUE #3923: Improve deprecation warning for DCDReader copying timestep

hmacdope:
Fixes #3923 

 - Adds more detail to deprecation warning. 

PR Checklist
------------
 - [X] Tests?
 - [X] Docs?
 - [X] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/coordinates/DCD.py
+++ b/package/MDAnalysis/coordinates/DCD.py
@@ -164,8 +164,14 @@ class DCDReader(base.ReaderBase):
         self.ts.dt = dt
         warnings.warn("DCDReader currently makes independent timesteps"
                       " by copying self.ts while other readers update"
-                      " self.ts inplace. This behaviour will be changed in"
-                      " 3.0 to be the same as other readers",
+                      " self.ts inplace. This behavior will be changed in"
+                      " 3.0 to be the same as other readers. In practice this"
+                      " means that you will no longer be able to store independent"
+                      " references to the Timestep object associated with the"
+                      " DCDReader. Read more at"
+                      " https://github.com/MDAnalysis/mdanalysis/issues/3899"
+                      " and https://github.com/MDAnalysis/mdanalysis/issues/3923"
+                      " to learn if this change in behavior might affect you.",
                        category=DeprecationWarning)
 
     @staticmethod
--- a/package/MDAnalysis/coordinates/DCD.py
+++ b/package/MDAnalysis/coordinates/DCD.py
@@ -165,12 +165,8 @@ class DCDReader(base.ReaderBase):
         warnings.warn("DCDReader currently makes independent timesteps"
                       " by copying self.ts while other readers update"
                       " self.ts inplace. This behavior will be changed in"
-                      " 3.0 to be the same as other readers. In practice this"
-                      " means that you will no longer be able to store independent"
-                      " references to the Timestep object associated with the"
-                      " DCDReader. Read more at"
-                      " https://github.com/MDAnalysis/mdanalysis/issues/3899"
-                      " and https://github.com/MDAnalysis/mdanalysis/issues/3923"
+                      " 3.0 to be the same as other readers. Read more at"
+                      " https://github.com/MDAnalysis/mdanalysis/issues/3889"
                       " to learn if this change in behavior might affect you.",
                        category=DeprecationWarning)
 


hmacdope:
Related #3889

package/MDAnalysis/coordinates/DCD.py
```diff
@@ -164,8 +164,14 @@ def __init__(self, filename, convert_units=True, dt=None, **kwargs):
...
         warnings.warn("DCDReader currently makes independent timesteps"
                       " by copying self.ts while other readers update"
-                      " self.ts inplace. This behaviour will be changed in"
-                      " 3.0 to be the same as other readers",
+                      " self.ts inplace. This behavior will be changed in"
+                      " 3.0 to be the same as other readers. In practice this"
+                      " means that you will no longer be able to store independent"
+                      " references to the Timestep object associated with the"
+                      " DCDReader. Read more at"
+                      " https://github.com/MDAnalysis/mdanalysis/issues/3899"
```
orbeckst: This is about *Bump up codecov action, add token secret* — probably wrong URL.

orbeckst (changes requested):
Sorry, quick review. 

As a user I'd still not understand what's changing. I'd want to see an example that demonstrates the code pattern that will fail and how to change it. Either add and example to #3923 or create an entry in our [WIki FAQ](https://github.com/MDAnalysis/mdanalysis/wiki/FAQ) (the [User Guide FAQ](https://userguide.mdanalysis.org/stable/faq.html) would also be a reasonable place).

package/MDAnalysis/coordinates/DCD.py
hmacdope: Whoops sorry.

package/MDAnalysis/coordinates/DCD.py
```diff
@@ -164,8 +164,14 @@ def __init__(self, filename, convert_units=True, dt=None, **kwargs):
         self.ts.dt = dt
         warnings.warn("DCDReader currently makes independent timesteps"
                       " by copying self.ts while other readers update"
-                      " self.ts inplace. This behaviour will be changed in"
-                      " 3.0 to be the same as other readers",
+                      " self.ts inplace. This behavior will be changed in"
+                      " 3.0 to be the same as other readers. In practice this"
+                      " means that you will no longer be able to store independent"
+                      " references to the Timestep object associated with the"
```
IAlibay: I'm going to +1 @orbeckst here, "references to Timestep" is _a lot_ for most users. I'd suggest probably putting up some code examples of how the behaviour is going to change.

package/MDAnalysis/coordinates/DCD.py
```diff
@@ -164,8 +164,14 @@ def __init__(self, filename, convert_units=True, dt=None, **kwargs):
...
+                      " DCDReader. Read more at"
+                      " https://github.com/MDAnalysis/mdanalysis/issues/3899"
+                      " and https://github.com/MDAnalysis/mdanalysis/issues/3923"
+                      " to learn if this change in behavior might affect you.",
```
IAlibay:
```suggestion
                      " self.ts inplace. This behavior will be changed in"
                      " 3.0 to be the same as other readers. Read more at"
                      " https://github.com/MDAnalysis/mdanalysis/issues/3889"
                      " to learn if this change in behavior might affect you.",
```
For the sake of not making this take too long, could we be happy with this and we can workshop the exact details on what needs to be communicated there?

IAlibay (changes requested):

package/MDAnalysis/coordinates/DCD.py
orbeckst: sounds good to me

IAlibay (approved):

hmacdope:
👍 Cheers @IAlibay

orbeckst (approved):

PR #3958, opened 2022-12-15
Deprecate nucleicacids pair distances results in favour of results.pair_distances [Issue 3744]

### LINKED ISSUE #3744: change nucleicacids results dict
The new experimental `nucleicacids` analysis module contains the https://github.com/MDAnalysis/mdanalysis/blob/c01d82bb86b480c8b2ae87f36497e7fc0a726c84/package/MDAnalysis/analysis/nucleicacids.py#L61 class as its base class. It presents results in a `Results` dict. The content is different from what other AnalysisBase-based classes use in that 
* it contains "times" which duplicates the `.times` attribute coming from AnalysisBase
* has integers as additional keys where "first index is selection, second index is time"

Specifically, for each of the selections, a dict is stored https://github.com/MDAnalysis/mdanalysis/blob/c01d82bb86b480c8b2ae87f36497e7fc0a726c84/package/MDAnalysis/analysis/nucleicacids.py#L134-L135 and res_dict comes [from https://github.com/MDAnalysis/mdanalysis/blob/c01d82bb86b480c8b2ae87f36497e7fc0a726c84/package/MDAnalysis/analysis/nucleicacids.py#L128-L129](https://github.com/MDAnalysis/mdanalysis/blob/c01d82bb86b480c8b2ae87f36497e7fc0a726c84/package/MDAnalysis/analysis/nucleicacids.py#L126-L129) where `dist` is an array of distances.

The "selections" are the individual base pairs so in order to query the distance of base pair `j` at time frame `t` I would use `results[j][t]`.

Please correct me, @ALescoulie , if I misrepresented anything here.

@IAlibay 's suggestion (from https://github.com/MDAnalysis/mdanalysis/pull/3735#pullrequestreview-1020755173) :
> would it make sense to consider switching this to instead be `results.pair_distances` (maybe even making it a *pairs x times* 2D ndarray?

Additionally, I  suggest to transpose so that the first axis corresponds to time frame and the second axis to pair index:
```python
results.pair_distances[:, j]    # time series of pair j
results.pair_distances[-1]      # all distances at the last time frame
```
The "one observation per row" layout is canonical and also used e.g. in the RMSD results.

Don't change anything... but this makes this module behave and feel different from the others and tools such as mdacli might have a hard time working with it — comments by @PicoCentauri  and @joaomcteixeira especially welcome on how to best structure `results`.

## Additional context ##

* https://github.com/MDAnalysis/mdanalysis/pull/3735#pullrequestreview-1020755173
* PR #3611 where nucleicacids was implemented
* *I am willing to ignore semver here* and change Results and anything else that needs changing in a breaking manner between 2.2 and 2.3, based on the fact that the module is in its infancy and is effectively experimental and likely has not had much uptake yet, so we're not going to hurt many users. It's much more important to define the API well (and not be bogged down with deprecations and crud code).

IAlibay:
Fixes #3744

A bit of a hack on Results, but thankfully we were only deprecating the one thing.

 - Directly indexing `results` is deprecated in favor of `results.pair_distances`. To be removed in v2.5.0
 - (being a bit eager on the 2.5.0 removal, can shift to 3.0.0 if it's really needed)

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/analysis/nucleicacids.py
+++ b/package/MDAnalysis/analysis/nucleicacids.py
@@ -45,7 +45,7 @@ For applications see :cite:p:`b-Denning2011,b-Denning2012`.
     Denning2012
 
 Distances
-_________
+---------
 
 .. autoclass:: NucPairDist
     :members:
@@ -60,6 +60,7 @@ _________
 """
 
 from typing import List, Dict
+import warnings
 
 import numpy as np
 
@@ -67,6 +68,28 @@ import MDAnalysis as mda
 from .distances import calc_bonds
 from .base import AnalysisBase, Results
 from MDAnalysis.core.groups import Residue
+
+
+# Remove in 2.5.0
+class DeprecatedResults(Results):
+    def _try_deprecated(self, attr):
+        """
+        If key is int, try to replicate deprecated results storage behaviour.
+        """
+        if isinstance(attr, int):
+            wmsg = ("Accessing results via selection indices is deprecated "
+                    "and will be removed in MDAnalysis 2.5.0")
+            warnings.warn(wmsg)
+            return self['pair_distances'][:, attr]
+        else:
+            return self[attr]
+
+    def __getattr__(self, attr):
+        try:
+            return _try_deprecated(attr)
+        except KeyError as err:
+            raise AttributeError("'Results' object has no "
+                                 f"attribute '{attr}'") from err
 
 
 class NucPairDist(AnalysisBase):
@@ -79,7 +102,7 @@ class NucPairDist(AnalysisBase):
     :class:`~MDAnalysis.core.groups.AtomGroup`.
 
     Parameters
-    __________
+    ----------
     selection1: List[AtomGroup]
         list of :class:`~MDAnalysis.core.groups.AtomGroup` containing an atom
         of each nucleic acid being analyzed.
@@ -90,14 +113,14 @@ class NucPairDist(AnalysisBase):
         arguments for :class:`~MDAnalysis.analysis.base.AnalysisBase`
 
     Attributes
-    __________
+    ----------
         results: numpy.ndarray
             first index is selection second index is time
         results.times: numpy.ndarray
             times used in analysis
 
     Raises
-    ______
+    ------
 
     ValueError
         if the selections given are not the same length
@@ -127,7 +150,7 @@ class NucPairDist(AnalysisBase):
             self._s1 += selection1[i]
             self._s2 += selection2[i]
 
-        self.results = Results()
+        self.results = DeprecatedResults()
 
     def _prepare(self) -> None:
         self._res_dict = {k: [] for k in range(self._n_sel)}
@@ -154,7 +177,7 @@ class WatsonCrickDist(NucPairDist):
     their index in the lists given as arguments.
 
     Parameters
-    __________
+    ----------
     strand1: List[Residue]
         First list of bases
     strand2: List[Residue]
@@ -184,14 +207,14 @@ class WatsonCrickDist(NucPairDist):
         arguments for :class:`~MDAnalysis.analysis.base.AnalysisBase`
 
     Attributes
-    __________
+    ----------
         results: numpy.ndarray
             first index is selection second index is time
         results.times: numpy.ndarray
             times used in analysis
 
     Raises
-    ______
+    ------
     ValueError
         if the residues given are not amino acids
     ValueError
--- a/package/MDAnalysis/analysis/nucleicacids.py
+++ b/package/MDAnalysis/analysis/nucleicacids.py
@@ -85,7 +85,7 @@ class DeprecatedResults(Results):
             else:
                 wmsg = ("Accessing results via selection indices is "
                         "deprecated and will be removed in MDAnalysis 2.5.0")
-                warnings.warn(wmsg)
+                warnings.warn(wmsg, DeprecationWarning)
                 return item
         raise KeyError(key)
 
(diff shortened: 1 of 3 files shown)


IAlibay:
@MDAnalysis/coredevs - last one on the todo list for 2.4.0

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -69,6 +70,26 @@
 from MDAnalysis.core.groups import Residue
 
 
+# Remove in 2.5.0
+class DeprecatedResults(Results):
```
orbeckst, line 74: So hacky... how convenient that you can fake anything in python.

testsuite/MDAnalysisTests/analysis/test_nucleicacids.py
```diff
@@ -41,5 +41,12 @@ def test_wc_dist(u):
...
 
-    assert_allclose(WC.results[0][0], 4.3874702, atol=1e-3)
-    assert_allclose(WC.results[1][0], 4.1716404, atol=1e-3)
+    assert_allclose(WC.results.pair_distances[0, 0], 4.3874702, atol=1e-3)
+    assert_allclose(WC.results.pair_distances[0, 1], 4.1716404, atol=1e-3)
+
+    wmsg = ("Accessing results via selection indices is "
+            "deprecated and will be removed in MDAnalysis 2.5.0")
+
+    with pytest.warns(match=wmsg):
```
orbeckst: MAybe use https://docs.pytest.org/en/6.2.x/reference.html#pytest.deprecated_call
```python
with pytest.deprecated_call(match=wmsg):
```

orbeckst (approved):
lgtm, thanks for the hack

testsuite/MDAnalysisTests/analysis/test_nucleicacids.py
IAlibay: That's a new one to me, sounds fun.

Good side-effect catch, I'm emitting the wrong warning type too.

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -79,30 +100,47 @@ class NucPairDist(AnalysisBase):
...
+    ----------
+    results: numpy.ndarray
+        Array of pair distances. First index is selection, second index is time.
+
+        .. deprecated:: 2.4.0
+           Will be removed in MDAnalysis 2.5.0. Please use
+           :attr:`results.pair_distances` instead.
+
+    results.times: numpy.ndarray
+        Simulation times used in analysis
```
orbeckst, line 129: I don't think we want `results.time` because we have the `times` attr in the class itself https://docs.mdanalysis.org/stable/documentation_pages/analysis/base.html#MDAnalysis.analysis.base.AnalysisBase.times

orbeckst (changes requested):
sorry, times should not be in results, I think

There's also low test coverage but I let this go because we remove it soon.

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -154,48 +187,54 @@ class WatsonCrickDist(NucPairDist):
...
-        by default assigned to N3
+        Name of Nitrogen 3 of nucleic acids, by default assigned to N3
     g_name: str (optional)
-        Name of Guanine in topology
-        by default assigned to G
+        Name of Guanine in topology, by default assigned to G
     a_name: str (optional)
-        Name of Adenine in topology
-        by default assigned to G
+        Name of Adenine in topology, by default assigned to G
```
fiona-naughton: As a minor typo comment, I would assume this is meant to read "...assigned to A" (but otherwise lgtm :) )

package/MDAnalysis/analysis/nucleicacids.py
IAlibay: Deprecating multiple things will get a little bit messy, one sec.

package/MDAnalysis/analysis/nucleicacids.py
IAlibay: Should be addressed now thanks!

package/MDAnalysis/analysis/nucleicacids.py
IAlibay: Done, we should think of a cleaner way to deprecate things in the future, maybe we can just bake it into the base class?

IAlibay:
> There's also low test coverage but I let this go because we remove it soon.

Always good to ask for more tests, turns out there were 3 cases I had overlooked!

package/MDAnalysis/analysis/nucleicacids.py
orbeckst: would be good

thanks for fixing this one

package/MDAnalysis/analysis/nucleicacids.py
```diff
@@ -127,23 +178,18 @@ def __init__(self, selection1: List[mda.AtomGroup],
...
-        for i in range(self._n_sel):
-            self._res_dict[i].append(dist[i])
-            self._times.append(self._ts.time)
+        self._res_array[self._frame_index, :] = dist
 
     def _conclude(self) -> None:
-        self.results['times'] = np.array(self._times)
-        for i in range(self._n_sel):
-            self.results[i] = np.array(self._res_dict[i])
+        self.results['times'] = np.array(self.times)
```
orbeckst, line 192: To be removed in 2.5.0

orbeckst (approved):
lgtm, thanks for the updates

PR #3970, opened 2022-12-16
Add Cython 0.29.32 generated files

IAlibay:
Third time is the charm?
(after this we'll have to do a 2.4.1 release if things go wrong I think...)


richardjgowers (approved):
I've totally read all 240k lines

PR #3976, opened 2022-12-20
histogramdd normed argument replaced by density

### LINKED ISSUE #3975: [upstream change] histogramdd normed has been removed
normed has been removed in favour of density in numpy 1.24. Which is making CI fail.

Strangely enough this was done 7 months ago and we didn't pick this up at all in our Cron jobs using nightly builds until yesterday, so we need to look at why this is happening (it's possible the install order is wrong).

IAlibay:
Fixes #3975 

 - 

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -18,14 +18,21 @@ The rules for this file:
  * 2.5.0
 
 Fixes
-  * np.histogramdd calls in :class:`DensityAnalysis` now pass the `density`
-    argument rather than the numpy 1.24 removed `density` (PR #3976)
 
 Enhancements
 
 Changes
 
 Deprecations
+
+
+12/21/22 IAlibay
+
+ * 2.4.2
+
+Fixes
+  * np.histogramdd calls in :class:`DensityAnalysis` now pass the `density`
+    argument rather than the NumPy 1.24 removed `density` (PR #3976)
 
 
 12/17/22 IAlibay
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -32,7 +32,7 @@ Deprecations
 
 Fixes
   * np.histogramdd calls in :class:`DensityAnalysis` now pass the `density`
-    argument rather than the NumPy 1.24 removed `density` (PR #3976)
+    argument rather than the NumPy 1.24 removed `normed` (PR #3976)
 
 
 12/17/22 IAlibay
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -18,6 +18,8 @@ The rules for this file:
  * 2.5.0
 
 Fixes
+  * np.histogramdd calls in :class:`DensityAnalysis` now pass the `density`
+    argument rather than the numpy 1.24 removed `density` (PR #3976)
 
 Enhancements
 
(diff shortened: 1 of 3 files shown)


pgbarletta:
I think you're just missing this [normed](https://github.com/MDAnalysis/mdanalysis/blob/123dabbe10764bfd6431aea445c3e96069c10041/package/MDAnalysis/analysis/density.py#L476). And [this one](https://github.com/MDAnalysis/mdanalysis/blob/123dabbe10764bfd6431aea445c3e96069c10041/package/MDAnalysis/analysis/density.py#L647) on the docs.

package/CHANGELOG
```diff
@@ -18,6 +18,8 @@ The rules for this file:
  * 2.5.0
 
 Fixes
+  * np.histogramdd calls in :class:`DensityAnalysis` now pass the `density`
```
IAlibay: @MDAnalysis/coredevs please look at discord for this one - in my opinion we probably should do a 2.4.2 release since this has effectively killed the DensityAnalysis code. However, _how_ we do bugfix releases really should be discussed.

TLDR; I'm pushing for us to be able to do bugfixes from a separate feature branch than `develop`. That way PRs can keep rolling into `develop` without breaking semver if we need to do a bugfix release.

IAlibay:
So the consensus (from @orbeckst and @richardjgowers that spoke up on this), is that we'll do separate bugfix branches based off the `release-2.4.x` tags where we cherry-pick commits from develop (where possible) and do release from them (please correct me if I misinterpreted things!).

As such, I've moved the changelog entry here to be 2.4.2, with the aim that I'll just cherry-pick the squash-merged commit out of here and re-deploy (yay..).

@MDAnalysis/coredevs please review this as a priority - it's the main cause of current CI failures, and it's blocking the 2.4.x announcement (we can't announce it if it's failing as soon as it's released).

orbeckst (approved):
thank you, lgtm

hmacdope (approved):
LGTM

hmacdope:
@IAlibay only concern is the 32 bit build? Causing some failures on the Azure CI

package/CHANGELOG
```diff
@@ -26,6 +26,15 @@ Changes
 Deprecations
 
 
+12/21/22 IAlibay
+
+ * 2.4.2
+
+Fixes
+  * np.histogramdd calls in :class:`DensityAnalysis` now pass the `density`
+    argument rather than the NumPy 1.24 removed `density` (PR #3976)
```
tylerjereddy: you may have meant `normed` here at the end?

tylerjereddy (commented):
CI failure looks like `h5py` not supporting NumPy `1.24.0` just yet on that platform.

32-bit Windows binaries seem to be getting dropped upstream a lot lately.

Releasing off of maintenance branches is what I do for SciPy and what Chuck does for NumPy, so +1 for that.

IAlibay:
we've had this particular h5py issue for some while (pre numpy 1.24 release), but the fact that there are more now might indicate that it's either related or it has accentuated the problem. I'm going to try switching the pipeline up a little bit to limit things for 32bit.

IAlibay:
Only extra thing I've done is slightly change the azure pipelines to pin numpy for 32bit (hopefully this fixes it 🤞🏽 thanks @tylerjereddy !), so with two approvals I'm going to go ahead with the merge.

PR #3977, opened 2022-12-20
Fix nightly NumPy cron job, add numpy 1.25-dev0 fixes

IAlibay:
Towards #3980 

 - Fixes cron job against nightly wheels
 - Fixes array comparison in streamlines_3D

PR Checklist
------------
 - [x] Tests?
 - Docs?
 - [x] CHANGELOG updated?
 - Issue raised/referenced?

### DIFF
--- a/.github/workflows/gh-ci-cron.yaml
+++ b/.github/workflows/gh-ci-cron.yaml
@@ -38,29 +38,16 @@ jobs:
       with:
         os-type: "ubuntu"
 
-    - name: setup_miniconda
-      uses: conda-incubator/setup-miniconda@v2
+    - uses: actions/setup-python@v4
       with:
-        python-version: 3.9
-        auto-update-conda: true
-        channel-priority: flexible
-        channels: conda-forge, bioconda
-        mamba-version: "*"
-        add-pip-as-python-dependency: true
-        architecture: x64
+        python-version: ${{ matrix.python-version }}
 
     - name: install_dev_versions
       run: |
         python -m pip install -i https://pypi.anaconda.org/scipy-wheels-nightly/simple numpy
         python -m pip install -i https://pypi.anaconda.org/scipy-wheels-nightly/simple scipy
 
-    - name: install_other_deps
-      uses: ./.github/actions/setup-deps
-      with:
-        mamba: true
-        numpy: ""
-        scipy: ""
-
+      # Intentionally going with setup.py builds so we can build with latest
     - name: build_srcs
       uses: ./.github/actions/build-src
       with:
@@ -68,10 +55,10 @@ jobs:
         build-tests: true
         build-docs: false
 
-    - name: check_install
+    - name: check_and_postinstall
       run: |
+        python -m pip install pytest-xdist
         python -m pip list
-        mamba list
 
     - name: run_tests
       run: |
--- a/.github/workflows/gh-ci-cron.yaml
+++ b/.github/workflows/gh-ci-cron.yaml
@@ -32,7 +32,8 @@ jobs:
       with:
         os-type: "ubuntu"
 
-    - uses: conda-incubator/setup-miniconda@v2
+    - name: setup_miniconda
+      uses: conda-incubator/setup-miniconda@v2
       with:
         python-version: "3.10"
         miniforge-variant: Mambaforge
@@ -54,7 +55,7 @@ jobs:
         pip install --pre -U -i https://pypi.anaconda.org/scipy-wheels-nightly/simple scipy numpy h5py matplotlib
 
 
-    - name: check_deps
+    - name: list_deps
       run: |
         mamba list
         pip list
--- a/.github/workflows/gh-ci-cron.yaml
+++ b/.github/workflows/gh-ci-cron.yaml
@@ -1,5 +1,9 @@
 name: GH Actions Cron CI
 on:
+  pull_request:
+    branches:
+      - develop
+      - master
   schedule:
     # 3 am Tuesdays and Fridays
     - cron: "0 3 * * 2,5"
@@ -45,8 +49,8 @@ jobs:
 
     - name: install_dev_versions
       run: |
-        pip install -i https://pypi.anaconda.org/scipy-wheels-nightly/simple numpy
-        pip install -i https://pypi.anaconda.org/scipy-wheels-nightly/simple scipy
+        python -m pip install -i https://pypi.anaconda.org/scipy-wheels-nightly/simple numpy
+        python -m pip install -i https://pypi.anaconda.org/scipy-wheels-nightly/simple scipy
 
     - name: install_other_deps
       uses: ./.github/actions/setup-deps
@@ -61,6 +65,11 @@ jobs:
         build-hole: true
         build-tests: true
         build-docs: false
+
+    - name: check_install
+      run: |
+        python -m pip list
+        mamba list
 
     - name: run_tests
       run: |
--- a/.github/workflows/gh-ci-cron.yaml
+++ b/.github/workflows/gh-ci-cron.yaml
@@ -42,10 +42,11 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
 
-    - name: install_dev_versions
+    - name: install_deps
       run: |
         python -m pip install -i https://pypi.anaconda.org/scipy-wheels-nightly/simple numpy
         python -m pip install -i https://pypi.anaconda.org/scipy-wheels-nightly/simple scipy
+        python -m pip install cython packaging setuptools
 
       # Intentionally going with setup.py builds so we can build with latest
     - name: build_srcs
--- a/.github/workflows/gh-ci-cron.yaml
+++ b/.github/workflows/gh-ci-cron.yaml
@@ -57,7 +57,8 @@ jobs:
       # overwrite installs by picking up nightly wheels
     - name: nightly_wheels
       run: |
-        pip install --pre -U -i https://pypi.anaconda.org/scipy-wheels-nightly/simple scipy numpy h5py matplotlib
+        pip install --pre -U -i https://pypi.anaconda.org/scipy-wheels-nightly/simple scipy h5py matplotlib
+        #pip install --pre -U -i https://pypi.anaconda.org/scipy-wheels-nightly/simple scipy numpy h5py matplotlib
 
 
     - name: check_deps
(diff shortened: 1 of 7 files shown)


IAlibay:
Looks like we were having numpy conflicts between mamba and pip https://github.com/MDAnalysis/mdanalysis/actions/runs/3739353709/jobs/6346461409#step:6:195

IAlibay:
right it's fixed, and we've actually picked up a breaking change with the nightly dev versions!

package/MDAnalysis/visualization/streamlines_3D.py
```diff
@@ -281,15 +281,15 @@ def update_dictionary_point_in_cube_start_frame(array_simulation_particle_coordi
...
+                cube['centroid_of_particles_first_frame'] = None
             cube_counter += 1
 
     def update_dictionary_end_frame(array_simulation_particle_coordinates, dictionary_cube_data_this_core):
         """Update the cube dictionary objects again as appropriate for the second and final frame."""
         cube_counter = 0
         for key, cube in dictionary_cube_data_this_core.items():
             # if there were no particles in the cube in the first frame, then set dx,dy,dz each to 0
-            if cube['centroid_of_particles_first_frame'] == 'empty':
+            if cube['centroid_of_particles_first_frame'] == None:
```
IAlibay: That took me far too long to narrow down, looks like the issue here was that we are at times doing an array vs str comparison, which until 1.24 led to a scalar return, but now returns an array of bools instead, that in turns messes up this if/else call.

Doing a comparison against `None` is probably just safer & more pythonic here?

P.S. Weirdly enough I think this should have thrown an error rather than silently failing i.e. it _should_ have complained about the truth of an array being ambiguous.. but it didn't 😕

IAlibay: See https://github.com/MDAnalysis/mdanalysis/actions/runs/3824585204/jobs/6506856469 for a successful run.

IAlibay:
@MDAnalysis/coredevs this should be good to go I think. Once it's merged I'll cherry pick the changes here and create a 2.4.2 bugfix release.

.github/workflows/gh-ci-cron.yaml
```diff
@@ -32,29 +32,34 @@ jobs:
       with:
         os-type: "ubuntu"
 
-    - name: setup_miniconda
```
orbeckst, line 35: You don't like naming steps anymore?

.github/workflows/gh-ci-cron.yaml
```diff
@@ -32,29 +32,34 @@ jobs:
...
       with:
         mamba: true
-        numpy: ""
-        scipy: ""
+        full-deps: true
+
+      # overwrite installs by picking up nightly wheels
+    - name: nightly_wheels
+      run: |
+        pip install --pre -U -i https://pypi.anaconda.org/scipy-wheels-nightly/simple scipy numpy h5py matplotlib
```
orbeckst, line 55: comment: there's a lot of  .\\\\agic hidden in these action files... I would have no idea where to look for these wheels.

.github/workflows/gh-ci-cron.yaml
```diff
@@ -32,29 +32,34 @@ jobs:
...
+
+
+    - name: check_deps
```
orbeckst: probably more accurately named "list deps"

package/MDAnalysis/visualization/streamlines_3D.py
```diff
@@ -281,15 +281,15 @@ def update_dictionary_point_in_cube_start_frame(array_simulation_particle_coordi
                                                            axis=0)
                 cube['centroid_of_particles_first_frame'] = centroid_particles_in_cube
             else:  # empty cube
-                cube['centroid_of_particles_first_frame'] = 'empty'
+                cube['centroid_of_particles_first_frame'] = None
```
orbeckst, line 286: apparently not covered by tests... not sure if this is incorrect or if we _must_ or _can_ cover it but I'll mention it

package/MDAnalysis/visualization/streamlines_3D.py
```diff
@@ -281,15 +281,15 @@ def update_dictionary_point_in_cube_start_frame(array_simulation_particle_coordi
...
             cube_counter += 1
 
     def update_dictionary_end_frame(array_simulation_particle_coordinates, dictionary_cube_data_this_core):
         """Update the cube dictionary objects again as appropriate for the second and final frame."""
         cube_counter = 0
         for key, cube in dictionary_cube_data_this_core.items():
             # if there were no particles in the cube in the first frame, then set dx,dy,dz each to 0
-            if cube['centroid_of_particles_first_frame'] == 'empty':
+            if cube['centroid_of_particles_first_frame'] is None:
```
orbeckst, line 294: comment: Looked innocently enough...

orbeckst (approved):
Looks fine to me, thank you for tracking down on all the little things to make it work!

(Minor _optional_ comments below.)

.github/workflows/gh-ci-cron.yaml
IAlibay: oops that was accidentally deleted

.github/workflows/gh-ci-cron.yaml
IAlibay: I think one of the next major tasks here will be to a) document the CI actions, b) clean them up (it's grown a lot beyond what I initially planned back at the start of last year when I squashed things into using re-usable actions).

package/MDAnalysis/visualization/streamlines_3D.py
IAlibay: O that is a bit problematic, we really must cover this. It also means we're getting 'empty' entries being fed somewhere else 🙀 Could you place a blocking review here?

orbeckst (changes requested):
Test assignment of `None` instead of "empty", please.

package/MDAnalysis/visualization/streamlines_3D.py
orbeckst: done

package/MDAnalysis/visualization/streamlines_3D.py
IAlibay: I'm surprised it's the only failure we have! 1.25 is going to cause a lot of trouble for legacy downstream codes.

package/MDAnalysis/visualization/streamlines_3D.py
IAlibay: aaaaahhh I was wondering why I was getting to print in that branch but CI wasn't picking up forced errors.. we're using multiprocessing 🤦🏽 

I'll have a quick look around but it looks like:
a) Errors aren't being propagated from child workers (hence why it didn't raise the relevant "truth of an array" error),
b) Codecov isn't picking up on coverage since it's not being properly logged by the child workers... we do have the multiprocessing thing set in coveragerc, but it's obviously not working

package/MDAnalysis/visualization/streamlines_3D.py
IAlibay: there's a really similar 'empty' check in https://github.com/MDAnalysis/mdanalysis/blob/0c4f823a457fd9ca91c25e41b318b3b2d0c582d8/package/MDAnalysis/visualization/streamlines.py#L204

Those probably also need to get changed...

.github/workflows/gh-ci-cron.yaml
IAlibay: Just to add here before I forget - there is a draft SPEC for nightly wheels: https://scientific-python.org/specs/spec-0004/

It needs a lot of fleshing out though...

IAlibay:
I'm a little bit confused as to why this specific branch doesn't get covered, it's definitely being entered by some of the workers, and both coverage.py and pytest-cov claim that it _should_ work...

I wonder if there's a bad termination happening somewhere else.

package/MDAnalysis/visualization/streamlines_3D.py
IAlibay: Can confirm the same FutureWarning is happening in streamlines :/ the test isn't good enough to pick it up though.

.github/workflows/gh-ci-cron.yaml
```diff
@@ -64,7 +76,7 @@ jobs:
 
     - name: run_tests
       run: |
-        pytest -n $numprocs testsuite/MDAnalysisTests --disable-pytest-warnings --durations=50
+        pytest -n $numprocs testsuite/MDAnalysisTests --durations=50 -W error::FutureWarning
```
IAlibay, line 73: Ideally we'd error on DeprecationWarnings too, but we emit too many of them internally for this to be viable. At least this way we can maybe pick up potential issues?

package/MDAnalysis/visualization/streamlines.py
```diff
@@ -176,7 +176,7 @@ def produce_list_centroids_this_frame(list_indices_in_polygon):
         list_centroids_this_frame = []
         for indices in list_indices_in_polygon:
             if not indices[0].size > 0:  # if there are no particles of interest in this particular square
-                list_centroids_this_frame.append('empty')
```
IAlibay, line 179: Same issue as per streamlines_3D, also can't get coverage over it (even though the new test assigns None to everything (hence the zeroed array return)

IAlibay:
Ok @orbeckst I think I'm at a point here where codecov is just not playing ball and there's not much else I can think of / find in the docs / issues. I don't understand the problem enough to be able to do a proper bug report either.

Short story is that those code branches are being covered, but they aren't being reported for some reason or another.

My suggestion at this point is to go with this knowing the above, and not delay 2.4.2 any further.

orbeckst (approved):
Thanks for all the digging.

_Perhaps_ add CHANGELOG entry for the internal fix in streamlines and streamlines2D? At a later point in time, this will be an actual error so I think it qualifies for a mentioning — it's a fix for a future error that would have happened.

IAlibay:
> Perhaps add CHANGELOG entry for the internal fix in streamlines and streamlines2D? At a later point in time, this will be an actual error so I think it qualifies for a mentioning — it's a fix for a future error that would have happened.

I might not be following what you mean here @orbeckst, do you mean something else than what has already been added to the 2.4.2 entry for these?

orbeckst:
Ah, sorry — the CHANGELOG checkmark was missing in the top comment (EDIT: I checked it now) and the _first_ file entry was _not_ CHANGELOG (as I was used to) so I wrongly assumed that there was no CHANGELOG entry. My bad. The one you have is perfect.

IAlibay:
Ok now to cherry pick everything into a 2.4.2 release.

PR #3941, opened 2022-11-30
[WIP] Change constant generation to use `pint` and reflect 2019 SI units. 

hmacdope:
Fixes #3924 

Currently very WIP, need to iron out the fact that some unit factors are 1/x in `speedfactor` and `forcefactor` but I have put this up so that people can see what a solution to #3924 would look like. 

Changes made in this Pull Request:
 - Changes constant generation for `units.py` to use `pint`

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?
 - [X] Issue raised/referenced?

PR #3981, opened 2023-01-01
Fixes #3944: Adding mass unit in "Base units"

### LINKED ISSUE #3944: mass unit not listed in 'Base units' table
It seems that the unit for the mass in MDAnalysis is missing. Should be unified atomic mass unit.

```rst
=========== ============== ===============================================
quantity    unit            SI units
=========== ============== ===============================================
mass         u              :math:`1.660538921 \times 10^{-27}` kg
```

1e9abhi1e10:
Fixes #3944 

 - Adding mass unit in "Base units"

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/units.py
+++ b/package/MDAnalysis/units.py
@@ -43,6 +43,7 @@ table on :ref:`table-baseunits`.
    =========== ============== ===============================================
    length       Å              :math:`10^{-10}` m
    time         ps             :math:`10^{-12}` s
+   mass         u              :math:`1.660538921 \times 10^{-27}` kg
    energy       kJ/mol         :math:`1.66053892103219 \times 10^{-21}` J
    charge       :math:`e`      :math:`1.602176565 \times 10^{-19}` As
    force        kJ/(mol·Å)     :math:`1.66053892103219 \times 10^{-11}` J/m
@@ -327,6 +328,7 @@ chargeUnit_factor = {
 conversion_factor = {
     'length': lengthUnit_factor,
     'density': densityUnit_factor,
+    'mass': massUnit_factor,
     'time': timeUnit_factor,
     'charge': chargeUnit_factor,
     'speed': speedUnit_factor,
@@ -346,6 +348,7 @@ for utype, ufactor in conversion_factor.items():
 #: Lookup table for base units in MDAnalysis by unit type.
 MDANALYSIS_BASE_UNITS = {"length": "A",
                          "time": "ps",
+                         "mass": "u",
                          "energy": "kJ/mol",
                          "charge": "e",
                          "force": "kJ/(mol*A)",
--- a/package/AUTHORS
+++ b/package/AUTHORS
@@ -200,6 +200,8 @@ Chronological list of authors
   - Jennifer A Clark
   - Jake Fennick
   - Utsav Khatu
+2023
+  - Abhishek Patidar
 
 
 External code
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -26,6 +26,14 @@ Changes
 Deprecations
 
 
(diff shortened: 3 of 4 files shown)


package/MDAnalysis/units.py
```diff
@@ -327,6 +328,7 @@
 conversion_factor = {
     'length': lengthUnit_factor,
     'density': densityUnit_factor,
+    'mass': massUnit_factor,
```
orbeckst, line 331: massUnit_factor is not defined anywhere

orbeckst (dismissed):
Thank you for your PR. Sorry that nobody had a look at it in a while. I only could do a brief review. The main thing seems to be that it is currently not possible to do anything with the defined mass unit. There should be some test that make use of it.

I am asking @PicoCentauri as the original author of issue #3944 to also review as he should have a clearer idea of what should be present.

package/MDAnalysis/units.py
```diff
@@ -43,6 +43,7 @@
    =========== ============== ===============================================
    length       Å              :math:`10^{-10}` m
    time         ps             :math:`10^{-12}` s
+   mass         u              :math:`1.660538921 \times 10^{-27}` kg
```
PicoCentauri, line 46: This where is belongs to. Thanks!

 However, I am not sure about the exact value of u. There was had a [comment](https://github.com/MDAnalysis/mdanalysis/issues/3944#issuecomment-1351805481) about a more accurate value. @v-parmar can you give a reference where we are using your value?

package/MDAnalysis/units.py
PicoCentauri: Yes, there could should be dictionary similar to the other. This would also define how we define our mass. Units could be `kg`, `g` and `u`


```python
massUnit_factor = {
    'u': 1.0,
    'kg': XXX
    'g': XXX,
}
```

orbeckst:
I dismissed my own review. @PicoCentauri please decide how to proceed with this PR. Please either do a "Request Changes" with actionable questions or todos or if you're happy submit an "Approval" and merge. I think at the moment it is not clear to @1e9abhi1e10 what they need to do here.  Thanks!

package/MDAnalysis/units.py
PicoCentauri, line 46: We should the use the latest value of the mass here

```suggestion
   mass         u              :math:`1.66053906660 \times 10^{-27}` kg
```

package/MDAnalysis/units.py
PicoCentauri, line 331: I think this conversion is useful but you should add a dictionary as given below.

```python
massUnit_factor = {
    'u': 1.0,
    'kg': 1.6605390666e-27
    'g': 1.6605390666e-24,
}
```

PicoCentauri (changes requested):
Thanks again @1e9abhi1e10 for this useful PR. In order to get this merged we have do to some minor changes. If you do not know how to implement them just let me know.

package/MDAnalysis/units.py
hmacdope: This is important change but tbh should be reflected in all units otherwise our unit system is inconsistent, see discussions about 2019 changes in #3941. Perhaps this is ok as a temporary fix?? What does @MDAnalysis/coredevs think?

hmacdope: I am going to block on this issue.  so people have a day or two to discuss.

hmacdope (changes requested):
Block on unit consistency.

orbeckst:
Are we continuing on this PR or should we close it? I am not sure what actually needs doing.

Issue #3748, opened 2022-07-06
Use of NumPy 1.21 style `npt.NDArray` typing hints


hmacdope:
## Problem

With the merge of #3737 NumPy 1.20 is now our minimum supported NumPy version as per [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html).
 However some open PRs (eg @hmacdope #3730 and @umak1106 #3746) are using the NumPy 1.21 style typing:

```python
import numpy.typing as npt
def blah(arr: npt.NDArray)
```

This is not going to play nice with our CI or advertised supported versions so we will have to decide how to proceed. 

See the [release notes](https://github.com/numpy/numpy/releases/tag/v1.21.0) for NumPy 1.21 to see what `npt.NDArray` actually is, but long and the short of it is that it is an alias for 

```
np.ndarray[Any, np.dtype[~Scalar]]
```

from looking at the NEP29 table the important dates are:

Jun 21, 2022 | 3.8+ | 1.20+
-- | -- | --
Jan 31, 2023 | 3.8+ | 1.21+
-- | -- | --

Some possible ideas to consider:

- Roll back some of the typing to the plain old ```arr: np.ndarray```
- Embed the alias ourselves until we have NumPy 1.21 (bit hacky)
- Make typing a future branch and postpone merging it until we have NumPy 1.21 underneath
- postpone a release until we have NumPy 1.21 underneath  (6 Months)
- Break NEP29
- Anything else?

Tagging @MDAnalysis/coredevs for input? This is probably also relevant to all GSOC and Outreachy people @MDAnalysis/gsoc-mentors (particularly for @umak1106's project).

tylerjereddy:
For one data point, I grepped around the SciPy codebase a bit and found that we indeed don't use `NDArray` in any `.py` files, opting for `np.ndarray` as you note above, but we do use the fancy new stuff in the static type stubs like `.pyi` files. The latter must be safe then, since it would only be used by `mypy` and you can just require a newer NumPy for the static checks that leverage the stub files in CI.

I guess it depends what mentors think, but maybe we could adopt a similar compromise for the next 6 months or so.

tylerjereddy:
[...]

PR #3983, opened 2023-01-02
Increase minimum numpy version to 1.21

IAlibay:
As per NEP29, the 2.5.0 release of MDA will happen ~ March 2023 and will be Python 3.8+ & NumPy 1.21+

See: https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table

 - Raised the minimum NumPy version to 1.21.0 
 - Change some uses of np.ndarray to npt.NDArray for typing

PR Checklist
------------
 - [x] Tests?
 - Docs?
 - [x] CHANGELOG updated?
 - Issue raised/referenced?


IAlibay:
I think that _might_ be fine now, I'm not a typing expert though so I'm heavily relying on mypy to tell me if I'm wrong (also I've only made changes where it was indicated in `lib`).

package/MDAnalysis/lib/distances.py
```diff
@@ -455,8 +452,8 @@ def capped_distance(reference: Union[np.ndarray, 'AtomGroup'],
     # The check_coords decorator made sure that reference and configuration
     # are arrays of positions. Mypy does not know about that so we have to
     # tell it.
-    reference_positions: np.ndarray = reference  # type: ignore
-    configuration_positions: np.ndarray = configuration  # type: ignore
+    reference_positions: npt.NDArray = reference  # type: ignore
+    configuration_positions: npt.NDArray = configuration  # type: ignore
```
hmacdope, line 456: Do we want to have subtyping on eg `npt.NDArray[np.float32]` ?

hmacdope (commented):
Quick question.

package/MDAnalysis/lib/distances.py
IAlibay: Unsure, but I'd probably ask that this get punted to a separate issue (unless these changes lessen the quality of typing from what it used to be)?

I'd prefer not move the goalpost here from "bumping up our numpy requirements" to "incidentally dealing with a ton of typing issues"

hmacdope: Yep totally fair. I'll raise an issue once merged. The typing here is improvement for sure, it can just be more stringent but agree its a separate issue.

hmacdope (approved):
No expert on all the places CI needs changing but the use of the new typing looks good.

hmacdope:
Note fixes #3748

PR #3988, opened 2023-01-04
taking a swing at #3825

### LINKED ISSUE #3825: occupancy and b-factors not updated when importing multi-frame PBD
MDAnalysis uses the occupancies/b-factors of the first frame for all subsequent frames when importing a multi-frame PDB.

```
import MDAnalysis

u   = MDAnalysis.Universe('traj.pdb')
sel = u.select_atoms("resid 178 and segid A and name HD2")

for _ in u.trajectory:
   print(sel.positions)

for _ in u.trajectory:
   print(list(sel.occupancies))

for _ in u.trajectory:
   print(list(sel.tempfactors))
```

Also, 

```
grep 'HD2 ASPTA 178' traj.pdb                                
ATOM   2808  HD2 ASPTA 178      81.480  35.990  49.120  0.00  0.00           H
ATOM   2808  HD2 ASPTA 178      83.700  34.570  50.290  1.00  1.00           H
ATOM   2808  HD2 ASPTA 178      80.540  40.020  49.060  0.06  0.06           H
ATOM   2808  HD2 ASPTA 178      81.100  37.920  49.720  0.20  0.20           H
ATOM   2808  HD2 ASPTA 178      84.130  38.260  50.810  0.09  0.09           H
ATOM   2808  HD2 ASPTA 178      81.790  39.980  49.520  0.04  0.04           H
```

And [link to traj.pdb](https://drive.google.com/file/d/1IMJJkRI9M1uzaZU08O6q4RC0P6jrd1tA/view?usp=sharing).

## Expected behavior ##

```
[[81.48 35.99 49.12]]
[[83.7  34.57 50.29]]
[[80.54 40.02 49.06]]
[[81.1  37.92 49.72]]
[[84.13 38.26 50.81]]
[[81.79 39.98 49.52]]
[0.00]
[1.00]
[0.06]
[0.20]
[0.09]
[0.04]
[0.00]
[1.00]
[0.06]
[0.20]
[0.09]
[0.04]
```

## Actual behavior ##

```
[[81.48 35.99 49.12]]
[[83.7  34.57 50.29]]
[[80.54 40.02 49.06]]
[[81.1  37.92 49.72]]
[[84.13 38.26 50.81]]
[[81.79 39.98 49.52]]
[0.0]
[0.0]
[0.0]
[0.0]
[0.0]
[0.0]
[0.0]
[0.0]
[0.0]
[0.0]
[0.0]
[0.0]
```

- Which version are you using? 2.3.0
- Which version of Python ? 3.10.6

richardjgowers:
Fixes #3825

 - ts.data['tempfactor'] is now populated each frame (from PDB format files)

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


richardjgowers:
@bradyajohnston this will pull out tempfactor (bfactor) and occupancies columns into the ts.data dict.  While we're here, is there any other column that you regularly see abused?  I think it's probably just those two right?

package/CHANGELOG
```diff
@@ -23,6 +23,9 @@ Enhancements
 
 * Add pickling support for Atom, Residue, Segment, ResidueGroup 
     and SegmentGroup. (PR #3953)
+* PDBReader now populates ts.data['tempfactor'] with the tempfactor for
```
IAlibay: Newest entry first 🙃

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -414,6 +418,10 @@ def _read_frame(self, frame):
                 except ValueError:
                     # Be tolerant for ill-formated or empty occupancies
                     pass
+                try:
```
IAlibay, line 430: This isn't the prettiest looking code 🤣 - if it's just going to be the one addition then fair, if we envision expanding this ever then maybe we should use a nicer looking loop or something

testsuite/MDAnalysisTests/coordinates/test_pdb.py
```diff
@@ -965,6 +965,31 @@ def test_first_residue(self, universe):
...
+    def u():
+        return mda.Universe(PDB_varying)
+
+    def test_ts_data_keys(self, u):
+        ts = u.trajectory.ts
+
+        assert 'occupancy' in ts.data
+        assert 'tempfactor' in ts.data
+
+    def test_varying_occupancy(self, u):
```
IAlibay: Could you:

a) parametrize these two into one test (seems easy enough with two keywords)
b) maybe do a more traditional trajectory iteration and rewind just to check everything works as expected?

IAlibay (changes requested):
Just a small update to tests and I think it's good to go :)

IAlibay (changes requested):
O actually the more tedious thing - could you add a section in the docs about this? We should just document this properly if we're going to (even temporarily) encourage this use case.

package/CHANGELOG
richardjgowers: this is a daft rule that doesn't improve anything, but ok

IAlibay: I wasn't going to block over it to be fair 🤣 

All jokes aside, it does help a lot when trying to work out the history of changes when doing progressive updates towards a milestone - it's a tiny convenience that can go a really long way.

IAlibay (approved):
lgtm!

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -397,6 +404,7 @@ def _read_frame(self, frame):
 
         pos = 0
         occupancy = np.ones(self.n_atoms)
+        tempfactor = np.zeros(self.n_atoms)
```
IAlibay: So a question here - shouldn't the default here be an array of `NaNs` or `None` rather than letting this fail silently if we can't read this column in?

(I realise we didn't use to do this, but I would suspect using ones or zeros to be a pain for someone downstream to distinguish being intended 0 and unintended 0?)

richardjgowers: Yeah I did wonder that for a bit. I think occupancies defaulting to 1.0 is ok, but maybe tempfactors could be more obviously not present rather than 0?

IAlibay: I think for both cases, since folks are using the fields to feed arbitrary data, we probably can't assert that a given default value is ok :/

IAlibay: @richardjgowers I should probably say that occupancies is something we should deal with separately, that'll need deprecating etc...

hmacdope (approved):
LGTM.

IAlibay (changes requested):
Just going to put a slight halt on this so we can discuss what a reasonable default should be (that way we can get it done with minimal further changes) - feel free to discard when needed @richardjgowers

IAlibay:
@richardjgowers what about just not populating ts.data if there's no information? That seems like the cleanest answer here, if folks get a KeyError then they know that frame doesn't have data?

richardjgowers:
Yes arguably better but, we already started always populating occupancies
so this would technically be a break. We could do that for tempfactors, but
then it’s inconsistent, so pick your poison

IAlibay:
@richardjgowers - how about 50% poison - don't populate for tempfactors, raise a warning for occupancies and say we'll change it in 2.6.0?

IAlibay:
@MDAnalysis/coredevs any views here? It'd be nice to fix this sooner than later if it's being actively needed downstream.

orbeckst:
I like the approach to "not populate if we don't read it" and working towards making this the default.

richardjgowers:
ok, now only populates `ts.data['tempfactors']` if values were seen, otherwise the key doesn't exist

package/MDAnalysis/coordinates/PDB.py
IAlibay: I think this is the only blocker to merging this PR. If the record isn't full, this by default will just set unseen records to `0` right? Are we ok with that? It seems like a recipe for disaster given that 0.0 has a meaning with tempfactors (so do occupancies, but I guess we can't fix that without deprecating unless we think it's a bug).

IAlibay: En lieu of further views here, please make the default value here `''` as per what we do for incomplete topology attr fields.

i.e. the cases should be:
- Field is seen but not incomplete -> mix of floats & '' for incomplete entries
- Field is fully seen -> floats
- Fied is not seen -> nothing

IAlibay (dismissed):
Let's fix this "partly seen" case please - I'd like to get this merged by end of week for the v2.5 release.

orbeckst:
@richardjgowers would you be able to do the tiny last bit from the review so that this can get into 2.5?

package/MDAnalysis/coordinates/PDB.py
```diff
@@ -397,6 +409,9 @@ def _read_frame(self, frame):
 
         pos = 0
         occupancy = np.ones(self.n_atoms)
+        tempfactor = np.empty(self.n_atoms)
+        tempfactor.fill(np.nan)  # Default value for unseen entry is a NaN
```
IAlibay: Ideally I would want to do the same thing for occupancy, but I guess that's a regression? (unless we think it's a bug), how do we want to handle this?

IAlibay:
@lilyminium @orbeckst or @jbarnoud - you three have some of the stronger views on what PDBs & attributes should do. Can one of you review this please? I self authored this now so I can't really push a merge.

package/MDAnalysis/coordinates/PDB.py
orbeckst: I'd leave occupancy as is (default 1) and only change it for 3.0 (if we want to) because some tools may do weird things with occupancy 0 atoms.  I understand that this is only in `ts.data` so you must already be looking for it there but at least by keeping it this way, it remains consistent with `AtomGroup.occupancy`, right?

orbeckst (approved):
looks good, I would leave occupancy as is for right now (see comment)

package/MDAnalysis/coordinates/PDB.py
IAlibay: Erm... no.

Going by https://github.com/MDAnalysis/mdanalysis/blob/5ab95ec55d2ff85c677827b4fc8a304957ac881a/package/MDAnalysis/topology/PDBParser.py#L297-L298

Occupancy defaults to 0 (not 1!), and tempfactors 1.0.

Should we just align things to that there?

orbeckst: For example, [PLUMED RMSD](https://www.plumed.org/doc-v2.8/user-doc/html/_r_m_s_d.html) uses occupancy = 1 to select atoms included in the calculation and uses occupancy = 0 to exclude. This may be a niche application. However, semantically it makes more sense to me to say "if we see an atom in a file and we don't have any other specific information, let's assume it exists (as opposed that it doesn't)". Thus I'd be in favor of **defaulting occupancy to 1**.

orbeckst: If defaulting occupancy everywhere to 1 is a breaking change then I suppose it's better to make `ts.data['occupancy']` consistent with the current "default 0" and raise an issue to change it for 3.0.

IAlibay: I'll be honest I'm kinda mindlessly trying to get 2.5.0 out - what's our easiest solution here? Set both of the ts.data default values to whatever gets populated in the topology attrs? I'd be willing to say that's a bug correction.

IAlibay (dismissed):
I'm going to put a hold here for now on the following question: "what should the defaults be, and should we just use the ones populated into topology attrs?"

package/MDAnalysis/coordinates/PDB.py
orbeckst: Follow the topology as default is a safe choice and I would not find it controversial. At least it follows the principle of "minimum surprise".

Then we can decide afterwards if the current defaults make sense.

orbeckst (changes requested):
After discussion, I am changing to "make ts.data[*] consistent with the defaults in the topology, even if these are questionable"
- default for occupancy = 0
- default for tempfactor = 1

And then we can decide in a separate issue how to change everything (probably for 3.0) but at least the above approach maintains internal consistency.

orbeckst:
@IAlibay I'll be 🏕️ in the 🌵 ⛰️ 🌵 so please dismiss my review at any point if you need to as I can't guarantee that I can look at anything over the next few days.

IAlibay:
> @IAlibay I'll be camping in the cactus mountain cactus so please dismiss my review at any point if you need to as I can't guarantee that I can look at anything over the next few days.

@orbeckst I think I've got this fixed with the state we want it in if you have time for a re-review. Otherwise some other @MDAnalysis/coredevs please review.

orbeckst (approved):

IAlibay:
I'm merging on the basis that the linux arm64 runner is being silly slow and it's not a big deal right now.

Issue #3925, opened 2022-11-16
[DOC] doctest MEGA-issue


hmacdope:
## Expected behavior ##

All the code in our docstring examples should be self-contained and run on its own. They should:

- import the correct modules
- demonstrate the desired behaviour
- run correctly

## Actual behavior ##

A lot of the examples in our docs do not run correctly as they are missing imports, steps, some code logic or all of the above. 

By my initial count there are ~350 failing examples and only a handful that work.

We would love this to be fixed! **This is a great opportunity for community members and budding open source contributors to get involved as there are lots of docstrings to tackle.**

Luckily there is a way to test which doc examples are failing  programatically!

1. Navigate to the `/package/doc/sphinx`  directory
2. run `make doctest`

This will indicate which doc examples are failing. 

If you want to tackle multiple doc changes at once  editing your makefile command on L130 to include `--keep-going`  will allow you to see the full set of errors. 

eg 
```
#L130
	$(SPHINXBUILD) -b doctest $(ALLSPHINXOPTS) $(BUILDDIR)/doctest --keep-going
```

**Please contribute your changes in a new PR!**

If you are new to making PRs and github in general, try our handy [contributing guide](https://userguide.mdanalysis.org/stable/contributing.html)! Feel free to ping our team if unsure. :)

richardjgowers:
Another idea would be to add the doctest in the CI and allow it to fail for now, then once we're passing make it another row in the CI.

sahilempire:
Hey maintainers, I want to work on this issue but I don't have enough knowledge, so can you please assign me and guide me through it?

hmacdope:
Hi @sahilempire sorry for the slow response. To address this issue you will need to run `doctests` that is navigate to `package/doc/sphinx` and run `make doctest`.

This will show you a bunch of failing doctests which we would like to fix up. Submit a PR fixing one or more! See our [contributing guide](https://userguide.mdanalysis.org/stable/contributing.html) for more info.

chrispfae:
Hi, I would also like to work on this issue. @sahilempire Have you already resolved some of the errors? Maybe we can split the work as there are a lot of failing examples?

sahilempire:
@chrispfae I haven't done anything in this yet and if you want to do something then you can do it.

IAlibay:
[...]

PR #3997, opened 2023-01-18
Add doctest step to docs build

IAlibay:
Enables #3925 

 - Adds a doctest CI step that will allow us to actually see progress towards #3925 
 - Moves linters out of the main CI file so it's a bit easier for us to see
 - Make it so that darker doesn't run on merge (that way we don't get failing badges)

PR Checklist
------------
 - [x] Tests?
 - Docs?
 - CHANGELOG updated?


IAlibay:
well darker lint shouldn't be skipping, something went wrong here...

IAlibay:
Also need to fix the doctest call, keep-going isn't propagating properly

hmacdope:
Great idea @IAlibay  and also @richardjgowers who may have suggested

IAlibay:
Ok that should be all addressed - that's 372 test failures (out of 388 it seems...). It won't fail but at least it'll give us a place to look at how things have improved / worsened based on a given PR.

.github/workflows/gh-ci.yaml
```diff
@@ -171,6 +171,12 @@ jobs:
         build-tests: false
         build-docs: true
 
+    - name: doctests
+      if: github.event_name == 'pull_request'
+      continue-on-error: true
+      run: |
+        cd package && python setup.py build_sphinx -b doctest --keep-going
```
orbeckst: I think `python setup.py build_sphinx` has been deprecated because setup.py does not want to be a build system (or something like that). Is there an alternative for invoking doctests??

orbeckst (commented):
I like the modularization and the use of a separate badge.

I could give it a "LGTM" if necessary but would prefer other folks with stronger opinions on the matter than myself  (@hmacdope @jbarnoud @richardjgowers ...) to weigh in.

.github/workflows/linters.yaml
```diff
@@ -0,0 +1,113 @@
...
+
+      - name: setup_miniconda
+        uses: conda-incubator/setup-miniconda@v2
+        with:
+          python-version: 3.8
+          auto-update-conda: true
+          channel-priority: flexible
+          channels: conda-forge, bioconda
+          add-pip-as-python-dependency: true
+          uses-mamba: true
```
orbeckst: Error/warning message

> Unexpected input(s) 'uses-mamba', valid inputs are ['installer-url', 'miniconda-version', 'miniforge-variant', 'miniforge-version', 'conda-version', 'conda-build-version', 'environment-file', 'activate-environment', 'python-version', 'add-anaconda-token', 'add-pip-as-python-dependency', 'allow-softlinks', 'auto-activate-base', 'auto-update-conda', 'condarc-file', 'channel-alias', 'channel-priority', 'channels', 'show-channel-urls', 'use-only-tar-bz2', 'remove-profiles', 'mamba-version', 'use-mamba', 'architecture', 'clean-patched-environment-file', 'run-post']

IAlibay: > use-mamba
I think I've been making a typo everywhere 😅

IAlibay: Should be fixed now.

.github/workflows/gh-ci.yaml
IAlibay: I _think_ setuptools isn't as harsh on the deprecation here, but I'll try switching to sphinx-build instead and see what happens.

IAlibay: That should do the trick. I'd probably say we should do the docs building & deployment elsewhere? I'm a bit apprehensive of changing that right now and then unleashing a whole other set of problems.

hmacdope (approved):
LGTM @IAlibay, thanks for the addition. Will be a great help for #3925 and by extension GSOC szn.

hmacdope (changes requested):
Not sure what is going on with Azure  CI

hmacdope:
Seems ok now.

hmacdope (approved):

Issue #3818, opened 2022-09-09
Discuss/document bond duplicate behaviour 


hmacdope:
In the course of moving forward on #3777 the question of what to do with duplicate bond entries has come up. 

The key problem is that this is a valid input for adding bonds:

```python
u.add_bonds([[0,1], [1,0]], types=[1, 2], order=['single', 'double'])
```
Where you can have a bond of the same index but different auxiliary values (types, orders, guesses).  Only the **first** bond entry is stored as far as I can tell  in the process of de-duplication when the `TopologyGroup` is constructed.

@richardjgowers and I discussed raising an error on duplicate entry in one of our meetings but this would be an API break from the current behaviour which is to remove duplicate entries preserving the first entry.  This behaviour has a test `core/test_universe.py::TestAddTopologyObjects::test_add_reversed_duplicates` and about 50 other tests fail if you enforce duplication==error. 

Currently the approach taken in #3777 matches this behaviour but as far as I can see there is not clear documentation of this behaviour (which entry will be removed). 

The docs currently  have the rather cryptic "Only one of the two permutations is stored, typically the one with the lower atom number first."  in the developer notes and the following in the `TopologyGroup` docstring "Duplicate entries are automatically removed upon creation and combination of different Dicts. This means a bond between atoms 1 and 2 will only ever appear once in a dict despite both atoms 1 and 2 having the bond in their bond attribute."
 At least that was all that I could see about duplicate bonds. 

My main points/questions are.

-  Do we want the current behaviour to maintained in the short term (#3777)?
-  Do we want to preserve this behaviour long term (think 3.0)?
- Regardless we should document more clearly that this is what happens when duplicate entries are added IMO?

My instinct is to move forward with the current system to avoid and API break and potentially change behaviour if desired in 3.0?

## Code to reproduce the behavior ##

Minimal example

``` python
import MDAnalysis as mda
from MDAnalysisTests import make_Universe
u = make_Universe()
u.add_bonds([[0, 1], [1, 0]], ['0', '1'], [False, True], ['do', 're'])
tg = u.bonds
tg._bondtypes # equal to 0
tg._guessed # equal to False
tg._order # equal to 'do'

# ... the same can be done with equivalent indices eg [[0,1],[0,1]]
# ... or with the ordering reversed [[1,0], [0,1]]
```

orbeckst:
I would have thought that adding a duplicate bond should raise an exception.

[...]

PR #4003, opened 2023-01-20
Resolve wrongly merging bond/angle/dihedral types

### LINKED ISSUE #4001: LAMMPS bond/angle/dihedral types are wrongly deleted
## Expected behavior ##

Correct number of bond/angle/dihedral types are written to lammps data file.

## Actual behavior ##

Method, core.topologyobjects._removeDupes(), wrongly removes bond/angle/dihedral types. This function is written for cases where the bond/angle/dihedral types are defined as a tuple of atom types, so that if both bond (1,2) and (2,1) are present, the two will be merged. However, in lammps the types are defined as integers in string format, so if there are more than 20 types, then bond 12 and 21 will be merged.

jaclark5:
Fixes #4001

 - if statement before calling _removeDupes() to make sure the bond/angle/dihedral type names are tuples, otherwise this step is skipped.

PR Checklist
------------
 - [X] Tests?
 - [NA] Docs?
 - [X] CHANGELOG updated?


hmacdope (changes requested):
@jaclark5 sorry I have been so slow to review. Would you be able to add a test to check the new behaviour is correct?

jaclark5:
I could use some help understanding why the test I wrote isn't working. Basically I import a LAMMPS data file with 22 bond types. 

1. The first assertion confirms that there are indeed 22 bond types since the bond types aren't defined with tuples an _removeDupes is not called. 
2. I then call `universe.atoms.bonds.topDict._removeDupes()` which will combine bond type 12 and bond type 21 to return 21 bond types. I then assert that the number of bonds does not equal 22.

However, it seems that the `TopologyDict` attribute `dict` doesn't actually update to reflect the changes made in step 2.
I added some print statements to the TopologyDict to try to debug this, here is the output from `pytest test_topologyobjects.py::TestTopologyGroup::test_bond_no_reversal -vvv`
```
Creating a TopologyDict Instance
Initiating TopologyDict
topDict.keys:         topDict id: 4589672672, topDict.dict id: 4590454144, len(topDict.dict.keys()): 22
Creating a TopologyDict Instance
Initiating TopologyDict
topDict._removeDupes: topDict id: 4589672672, topDict.dict id: 4589549056, len(topDict.dict.keys()): 21
Creating a TopologyDict Instance
Initiating TopologyDict
topDict.keys:         topDict id: 4589672672, topDict.dict id: 4590379904, len(topDict.dict.keys()): 22
```

Edit: I figured it out, it appears that the class is reinitiated every time the bond types are called. Probably because `TopologyGroup.topDict` returns `TopologyDict(self)` instead of checking if it exists. This is a separate issue, should I raise it?

hmacdope:
I think `bondDict` and its cacheing is due to be removed with #3777, but I agree this is perhaps a bit of a "dark-arts" part of MDAnalysis. Well done for figuring out your issue.

There is also the related issue of the current system of bonds and even the solution in #3777 only retaining one bond with identical  indices (eg only one of 1-2 or 2-1 can exist). This is documented in #3818  and I have been meaning to get on with it as it is definately an issue. I think the semi-conclusion  we came to last time was that this is a topology parser issue, IE the parser should sort out the bonds so that there are no duplicates. 

Sorry for the info dump but thought I would help out with some context.

jaclark5:
I'm not sure it will since that PR doesn't touch topologyobjects.py. The issue is that `TopologyGroup.topDict` is supposed call `TopologyDict` and then cache it so that it doesn't keep reinitiating a new instance. This could be resolved by setting something like `self._topDict = None` in `TopologyGroup.__init__` and then changing `TopologyGroup.topDict` from:
```
@property
@cached('dict')
def topDict(self):
    return TopologyDict(self)
```
to
```
@property
def topDict(self):
    if self._topDict is None:
        self._topDict = TopologyDict(self)
    return self._topDict
```

or just initiate topDict in `TopologyGroup.__init__`... but I haven't looked through to understand the order of operations that would justify the current pattern with a cached method acting as an attribute instead of defining an attribute to begin with.

package/MDAnalysis/core/topologyobjects.py
```diff
@@ -462,6 +464,7 @@ def _removeDupes(self):
                 newdict[k[::-1]] += self.dict[k]
 
         self.dict = newdict
+        print("topDict._removeDupes: topDict id: {}, topDict.dict id: {}, len(topDict.dict.keys()): {}".format(id(self), id(self.dict), len(self.dict)))
```
hmacdope: remove

package/MDAnalysis/core/topologyobjects.py
```diff
@@ -473,6 +476,7 @@ def __len__(self):
 
     def keys(self):
         """Returns a list of the different types of available bonds"""
+        print("topDict.keys:         topDict id: {}, topDict.dict id: {}, len(topDict.dict.keys()): {}".format(id(self), id(self.dict), len(self.dict)))
```
hmacdope: Remove

package/MDAnalysis/core/topologyobjects.py
```diff
@@ -656,6 +660,7 @@ def topDict(self):
 
         .. versionadded 0.9.0
         """
+        print("Creating a TopologyDict Instance")
```
hmacdope: Remove

hmacdope (changes requested):
I think this seems reasonable approach for now, I guess you found that the original setup is error prone and likely still is, but this is a step in the right direction

jaclark5:
Hey @hmacdope, I was wrong, the issue isn't that the `TopologyDict` isn't caching correctly in the `TopologyGroup`, but instead that `topologyattrs._Connection.get_atoms` reinitializes the `TopologyGroup` every time you call the `AtomGroup.bonds` and this is not resolved in #3777 but it could be. Once that is resolved and pushed my tests will pass.

hmacdope:
@jaclark5 do we still need #3777 or is this now ready for a seperate review? I think the approach taken here can stand on its own. :)

package/MDAnalysis/core/topologyobjects.py
```diff
@@ -445,7 +445,8 @@ def __init__(self, topologygroup):
             except KeyError:
                 self.dict[btype] = [b]
 
-        self._removeDupes()
+        if self.dict and isinstance(list(self.dict.keys())[0], tuple):
```
hmacdope, line 455: Comment here as to why the `if` is needed would be good.

testsuite/MDAnalysisTests/core/test_topologyobjects.py
```diff
@@ -300,6 +301,12 @@ def test_bond_reversal(self, PSFDCD, b_td):
         tg2 = b_td[b[::-1]]
         assert tg1 == tg2
 
+    def test_bond_no_reversal(self):
```
hmacdope, line 307: Again a quick comment on why this tests what it tests, eg `21` vs `12`.

hmacdope (changes requested):
Two small comments but other than that I am happy.

hmacdope (approved):
Thanks @jaclark5 for the fantastic work!

hmacdope:
@MDAnalysis/coredevs any chance anyone can take a quick peek here also?

jaclark5:
> @jaclark5 do we still need #3777 or is this now ready for a seperate review? I think the approach taken here can stand on its own. :)

@hmacdope Not anymore. I had another test that I removed that also removed that dependency. This other test would call `universe.atoms.bonds.topdict._removeDupes()` and then again call `universe.atoms.bonds.topdict.types()` to show that bond types 12 and 21 were combined. However it failed since calling `universe.atoms.bonds.topdict._removeDupes()` reinitialized the `topdict` and wouldn't retain the effect of calling `universe.atoms.bonds.topdict._removeDupes()`. 

I think this is still something to think about, since it would slow MDAnalysis down to reinitialize the bonds topdict every time it's called. Should I make it another issue or would you want to incorporate that into #3777 ?

hmacdope:
> > @jaclark5 do we still need #3777 or is this now ready for a seperate review? I think the approach taken here can stand on its own. :)
> 
> @hmacdope Not anymore. I had another test that I removed that also removed that dependency. This other test would call `universe.atoms.bonds.topdict._removeDupes()` and then again call `universe.atoms.bonds.topdict.types()` to show that bond types 12 and 21 were combined. However it failed since calling `universe.atoms.bonds.topdict._removeDupes()` reinitialized the `topdict` and wouldn't retain the effect of calling `universe.atoms.bonds.topdict._removeDupes()`.
> 
> I think this is still something to think about, since it would slow MDAnalysis down to reinitialize the bonds topdict every time it's called. Should I make it another issue or would you want to incorporate that into #3777 ?

@jaclark5 raise an issue, I think we can just move forward with this as it is :)

jaclark5:
@MDAnalysis/coredevs I would appreciate another review :)

hmacdope:
@MDAnalysis/coredevs can I get another set of eyes here? If no action in a week I will just go ahead. Sorry we have been a bit slow @jaclark5.

hmacdope:
Cycling CI for aarch.

package/CHANGELOG
```diff
@@ -35,6 +35,7 @@ Deprecations
  * 2.5.0
 
 Fixes
+  * Fix deletion of bond/angle/dihedral types (PR #4003, Issue #4001)
```
hmacdope: @jaclark5 can you move this to 2.6.0?

hmacdope (changes requested):
@jaclark5 just need to move some CHANGELOG entry to 2.6.0

hmacdope:
@IAlibay can I safely ignore CI failure here?

IAlibay:
> @IAlibay can I safely ignore CI failure here?

Ahh sorry for the delay, I thought I had seen a notification come through 😅 

Yeahhh ignore it here, it's not relatetd to this code. It's because of the 1.1 release of contourpy from ~ 3 days ago and the fact that we annoyingly still support 32bit... I'll open a PR.

hmacdope (approved):
Thanks so much @jaclark5! Great work as always. Great to finally get it in.

PR #4006, opened 2023-01-23
Solve3993

### LINKED ISSUE #3993: [ENH]: Change MSD code to accumulate values using `.timeseries`.

The MSD code currently uses a custom loop to collect positions  in `_run()` and then calculates the actual MSD in `_conclude`. The loop over ts.positions is the same as that handled by  `.timeseries` so we could probably streamline it. 

This could be made more streamlined using `.timeseries` to replace the _run() loop.

MohitKumar020291:
Fixes #3993 

-I created a new function named `_single_frame_byReader` in msd.py and deleting `_single_frame` because to remove `results.timeseries()` from msd.py and adding `reader.timeseries()` as mentioned in issue #3993. 
  
 

PR Checklist
------------
 - [ ] Tests?
 - [x] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/benchmarks/benchmarks/analysis/rdf.py
+++ b/benchmarks/benchmarks/analysis/rdf.py
@@ -45,4 +45,4 @@ class SimpleRdfBench(object):
         """Benchmark a full trajectory parse
         by MDAnalysis.analysis.rdf.InterRDF
         """
-        self.rdf.timeseries()
+        self.rdf.run()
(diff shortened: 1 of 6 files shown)


hmacdope (changes requested):
Hi @MohitKumar020291, I think before I review again, you should have a look at the way 'AnalysisBase' works, the way .timeseries works and how to run the test suite. Is your function working as expected? I think possibly not because the tests are failing. Perhaps try to go through those and see what isn't working and go from there.

hmacdope:
Also @MohitKumar020291 we only allow each contributor to have one PR open per issue as this avoids wasteful duplication of effort. I am closing this so we can continue discussion on #4004

PR #4010, opened 2023-01-26
Filtering pytng reader warnings (resolves #3986)

ooprathamm:
Fixes #3986

Changes made in this Pull Request:
 -  added filter for repeated warning in test_tng.py

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?
 - [x] Issue raised/referenced?

PR #4011, opened 2023-01-26
[doc] Resolving doc index issue.

### LINKED ISSUE #3998: incorrect subheadings in docs overview pages
## Expected behavior ##

Docs for classes and functions do not show up in table of content lists at the module level at 
* https://docs.mdanalysis.org/stable/documentation_pages/coordinates_modules.html
* https://docs.mdanalysis.org/stable/documentation_pages/topology_modules.html
* https://docs.mdanalysis.org/stable/documentation_pages/lib_modules.html
* https://docs.mdanalysis.org/stable/documentation_pages/analysis_modules.html

## Actual behavior ##
For a number of  modules, the classes appear in the list, e.g.,

* [6.2. CRD structure files in MDAnalysis — MDAnalysis.coordinates.CRD](https://docs.mdanalysis.org/stable/documentation_pages/coordinates/CRD.html)
* [CRDReader](https://docs.mdanalysis.org/stable/documentation_pages/coordinates/CRD.html#MDAnalysis.coordinates.CRD.CRDReader)
* [CRDWriter](https://docs.mdanalysis.org/stable/documentation_pages/coordinates/CRD.html#MDAnalysis.coordinates.CRD.CRDWriter)

See more in the screen shots below and on the pages listed under **Expected behavior** 

image
image

Other pages where the same thing happens are listed above under *Expected behavior*.

Also note https://github.com/MDAnalysis/mdanalysis/pull/4011#pullrequestreview-1294850027 — the [2.3.0 docs](https://docs.mdanalysis.org/2.3.0/documentation_pages/coordinates_modules.html) still look correct, starting with the [2.4.0 docs](https://docs.mdanalysis.org/2.4.0/documentation_pages/coordinates_modules.html) the layout problems start. **We need to understand why the sphinx output changed.**

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) docs 2.4.1 (stable)

ooprathamm:
Fixes #3998 

 - 

PR Checklist
------------
 - [ ] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


orbeckst:
Closing the PR because another GSoC PR #4010 is already open (see https://github.com/MDAnalysis/mdanalysis/pull/4013#issuecomment-1407900084 for explanation).

ooprathamm:
Reopening

package/doc/sphinx/source/documentation_pages/coordinates_modules.rst
```diff
@@ -15,36 +15,227 @@ provide the format in the keyword argument *format* to
 
 .. rubric:: Coordinate formats
 
+Trajectory readers and writers
+==============================
 .. toctree::
```
orbeckst: It should not be necessary to manually add headings and add toctrees.

orbeckst (changes requested):
Thanks for the work. But I am not convinced that adding headings and separate toc trees really solves the problem. It was working in our 2.3.0 docs https://docs.mdanalysis.org/2.3.0/documentation_pages/coordinates_modules.html but then the problems appeared with https://docs.mdanalysis.org/2.4.0/documentation_pages/coordinates_modules.html . I don't know what changed – maybe the version of sphinx? 

What is needed is figuring our where the problem comes from. Once we know this we can make changes.

But just changing the text in the way that you suggest is not sustainable. 

I am looking for a solution where the files remain similar to how they were before but where we change some build options or how we generate tocs. Perhaps we need to be more careful in how we indicate sections. But in any case, we **first need to understand why the output changed**.

ooprathamm:
@orbeckst After comparing the release 2.3 with the release 2.4, I came to the conclusion that sphinx version is updated and referring sphinx-doc/sphinx#10804 seems that the how sphinx build automodule members has changed after 5.2.0

so, this solution is the most appropriate fix I could find without changing the .rst files.

orbeckst (approved):
Thanks for digging into the changes in sphinx and identifying the key change to toctrees.

I looked at https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-toctree and noticed that the `:titlesonly:` flag to switch off the new behavior on a case-by-case basis.

I am ok with the global settings change in the config file to restore our previous behavior but we should keep in mind that we could also use the new behavior to document content in more detail.

You are already in CHANGELOG so that's all good.

PR #4016, opened 2023-01-30
Set expected list of elements for Parmed PSF test

### LINKED ISSUE #4015: New parmed release causes PSF tests to fail
## Expected behavior ##

All tests pass

## Actual behavior ##

See: https://github.com/MDAnalysis/mdanalysis/actions/runs/4039771911/jobs/6944817173#step:7:517

Looks like the latest release of parmed can now parse elements.

IAlibay:
Fixes #4015 

 - Add a list of expected elements rather than just empty strings

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -18,6 +18,7 @@ The rules for this file:
  * 2.5.0
 
 Fixes
+  * Fix element parsing from PSF files read via Parmed (Issue #4015)
   * Fix test clobbering in lib/test_util.py (PR #4000)
   * Fix MSD docs to use the correct error metric in example (Issue #3991)
   * Add 'PairIJ Coeffs' to the list of sections in LAMMPSParser.py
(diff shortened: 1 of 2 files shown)


IAlibay:
@MDAnalysis/coredevs can I get a quick check here - this issue is making CI fail across the board.

orbeckst (approved):
lgtm

Is there anything in the docs that we need to add about new capabilities?

Do we need to restrict to newer versions of parmed or is the new capability just incidental and it doesn't really matter to us?

IAlibay:
> lgtm
> 
> Is there anything in the docs that we need to add about new capabilities?
> 
> Do we need to restrict to newer versions of parmed or is the new capability just incidental and it doesn't really matter to us?

@orbeckst It's purely incidental & only affects tests - i.e. upstream now supports it so if you use the latest parmed you'll get elements instead of empty strings. Otherwise you'll get empty strings (i.e. the expected behaviour before the previous release).

I've clarified this in the changelog.

PR #4017, opened 2023-01-30
Add an `AnalysisCollection` class

### LINKED ISSUE #3569: Collective analysis of different analysis classes on the same trajectory
_Disclaimer: I thought there was an issue in the past with this idea but I was not able to find it. Sorry if the following idea is 
a duplication_

If one wants to run several analyses on the same trajectory one has to loop through the trajectory several times. For example, one naive way of analyzing two RDFs is the following:

```python
import MDAnalysis as mda
from MDAnalysis.analysis.rdf import InterRDF
from MDAnalysis.analysis.base import AnalysisCollection

# Load simulation results
u = mda.Universe('topol.tpr','traj.trr')

# Select atoms
O = u.select_atoms('name O')
H = u.select_atoms('name H')

# Create individual analysis objects
rdf_OO = InterRDF(O,O)
rdf_OH = InterRDF(O,H)

# Run the individual analysis
rdf_OO.run(start=0, end=100, step=10)
rdf_OH.run(start=0, end=100, step=10)

# Results are stored in indivial objects
print(rdf_OO.results)
print(rdf_OH.results)
```

I know that there is the `InterRDF_s` class which exactly does this job. But, the problem of a collective analysis loop holds for any analysis class and I only chose the RDF since it a straightforward example. In addition, a collected Analysis could make `InterRDF_s` superfluous.

I would introduce an `AnalysisCollection(*analysis_object)` class. When creating an object one can register several 
`analysis_object`s as arguments. The run method of `AnalysisCollection` loops through the 
trajectory and calls the `single_frame` method of each `analysis_object` in every frame. Results are stored in each method. 
The code from above will simplify to

```python
import MDAnalysis as mda
from MDAnalysis.analysis.rdf import InterRDF
from MDAnalysis.analysis.base import AnalysisCollection

# Load simulation results
u = mda.Universe('topol.tpr','traj.trr')

# Select atoms
O = u.select_atoms('name O')
H = u.select_atoms('name H')

# Create individual analysis objects
rdf_OO = InterRDF(O,O)
rdf_OH = InterRDF(O,H)

# Create collection for common trajectory loop
collection = AnalysisCollection(rdf_OO, rdf_OH)

# Run the collected analysis
collection.run(start=0, end=100, step=10)

# Results are stored in indivial objects
print(rdf_OO.results)
print(rdf_OH.results)
```

I think this is approach is fairly easy to implement since the managing of the data remains in the name space of each `analysis_object `.

Keep everything as it is.

PicoCentauri:
Fixes #3569 

The `AnalysisCollection ` follows the discussion of #3569 resetting the timestep object for each analysis by default. If requested the user can set the `reset_timestep` variable to `False` allowing altering the timestep object.

Within this PR I moved the run method from `AnalysisBase` to `AnalysisCollection` to avoid code a lot of duplication. These changes interface of Github might make this a bit difficult to follow. The core changes the `run` method only that instead of running the `_prepare`, `_single_frame` and `_conclude` methods only once it is looped over all provided analysis instances.

 - Add an `AnalysisCollection` class to perform multiple analyses on the same trajectory
 - Use `assert_equal` in `test_base.py`
 
PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

_Disclaimer: This PR was written with inspiration of OpenAI's ChatGPT._


package/MDAnalysis/analysis/base.py
```diff
@@ -220,7 +220,165 @@ def __setstate__(self, state):
...
+                ts_original = ts.copy()
+
+            for analysis_object in self._analysis_objects:
+                analysis_object._frame_index = i
+                analysis_object._ts = ts
+                analysis_object.frames[i] = ts.frame
+                analysis_object.times[i] = ts.time
+                analysis_object._single_frame()
+
+                if reset_timestep:
```
richardjgowers: Ts is never unassigned, so I'm not sure this is necessary here

package/MDAnalysis/analysis/base.py
```diff
@@ -220,7 +220,165 @@ def __setstate__(self, state):
...
+        )
+
+        for i, ts in enumerate(
+            ProgressBar(self._analysis_objects[0]._sliced_trajectory, verbose=verbose)
+        ):
+            if reset_timestep:
+                ts_original = ts.copy()
+
+            for analysis_object in self._analysis_objects:
+                analysis_object._frame_index = i
```
richardjgowers: This is messy but understandable. In hindsight, maybe passing these variables through the single frame method would have been cleaner.

richardjgowers (approved):
I like this. Can I be cheeky and inquire what sort of speed up you get running something like 2 separate rdfs in tandem?

package/MDAnalysis/analysis/base.py
PicoCentauri: But, if an instance is changing the ts object we have to restore it from the stored one.

package/MDAnalysis/analysis/base.py
PicoCentauri: Yes, maybe be, but changing this might break a lot. I can put a comment so that these setters might become clearer.

PicoCentauri:
If I use the code given in the example of class I got a speedup of ~30% compared to running each class individually. Quite good for such a simple change  🙂

orbeckst:
@richardjgowers I made you the responsible adult in the room ;-) — please shepherd the PR to completion

yuxuanzhuang:
Up for discussion, would it be a good idea to inherit `abc.Collection` class so that one can do e.g.
```python
for analysis in collection:
    print(analysis.results)
```

Besides, it feels a bit weird to me that `AnalysisBase` is inherited from `AnalysisCollection` which is a collection of `AnalysisBase`s...

package/MDAnalysis/analysis/base.py
```diff
@@ -220,7 +224,176 @@ def __setstate__(self, state):
...
+
+        logger.info("Finishing up")
+
+        for analysis_object in self._analysis_instances:
+            analysis_object._conclude()
+
+        return self
+
+
+class AnalysisBase(AnalysisCollection):
```
orbeckst: I agree with @yuxuanzhuang : it feels very weird and circular to have AnalysisBase inherited from AnalysisCollection. It just does not make sense semantically.

Can this be changed?

orbeckst (changes requested):
I do not like turning the core of AnalysisBase inside out and into AnalysisCollection (ie `run()`). 

It's semantically confusing and it also makes it even harder to understand how to write your own analysis.

I think the logic needs to be rethought.

PicoCentauri:
We could wither duplicate the logic, which I don't like. Or, create en external class that both the Base and the Collection use.

yuxuanzhuang:
> We could wither duplicate the logic, which I don't like. Or, create en external class that both the Base and the Collection use. 

Would it be a good idea to create an external `Runner` class for the `run` task and have e.g. `SingleAnalysisRunner`, `CollectionAnalysisRunner` (I'm not totally sure what the best design pattern is here) ? I'm also thinking about if we plan to implement parallel analysis based on it, it can be simply creating a new `ParallelAnalysisRunner` (or even `ParallelCollectionAnalysisRunner`) instead of doing complex inheritance.

PicoCentauri:
> Would it be a good idea to create an external Runner class for the run task and have e.g. SingleAnalysisRunner, CollectionAnalysisRunner (I'm not totally sure what the best design pattern is here) ? I'm also thinking about if we plan to implement parallel analysis based on it, it can be simply creating a new ParallelAnalysisRunner (or even ParallelCollectionAnalysisRunner) instead of doing complex inheritance.

I am fine with creating these two classes. To avoid code duplication the `SingleAnalysisRunner` will inherit from the `CollectionAnalysisRunner` and just call the run method for a single instance.

Also, should they rather be private classes? I think users will not use them because these runners will require an already prepared analysis class.

hmacdope:
> We could wither duplicate the logic, which I don't like. Or, create en external class that both the Base and the Collection use.

I think this solution is the cleanest and doesn't require even more classes which I don't like the idea of. IMO there is already enough of a barrier to entry without every new analysis having to handle 2x new API points. 

Personally I would just have an `abc` abstact base class that they can both inherent from.

hmacdope (changes requested):
Sorry should have left a review.

PicoCentauri:
Yes, we can do an ABC class if the others also agree.

package/MDAnalysis/analysis/base.py
```diff
@@ -220,7 +224,176 @@ def __setstate__(self, state):
...
+            only be used *instead* of `start`, `stop`, and `step`. Setting
+            *both* `frames` and at least one of `start`, `stop`, `step` to a
+            non-default value will raise a :exc:`ValueError`.
+        verbose : bool, optional
+            Turn on verbosity
+        reset_timestep : bool, optional
+            Reset the timestep object after for each ``analysis_object``.
+            Setting this to ``False`` can be useful if an ``analysis_object``
+            is performing a trajectory manipulation which is also useful for the
+            subsequent ``analysis_instances`` e.g. unwrapping of molecules.
```
richardjgowers: on the basis that there should be one way to do anything, I don't like the suggestion that you could use an analysis class to transform data for downstream classes (use a transformation instead).  Instead make `reset_timestep=True` the only option and remove this kwarg

package/MDAnalysis/analysis/base.py
```diff
@@ -220,7 +224,176 @@ def __setstate__(self, state):
...
+        """
+
+        # Ensure compatibility with API of version 0.15.0
+        if not hasattr(self, "_analysis_instances"):
+            self._analysis_instances = (self,)
```
richardjgowers: this is confusing, instead this is making a single analysis class work as a subclass of a `Collection` right?

package/MDAnalysis/analysis/base.py
```diff
@@ -316,6 +489,7 @@ def __init__(self, trajectory, verbose=False, **kwargs):
         self._trajectory = trajectory
         self._verbose = verbose
         self.results = Results()
+        super(AnalysisBase, self).__init__(self)
```
richardjgowers: this ugly super call signature was Python 2.x, I think we can use `super().__init__(self)` now

package/MDAnalysis/analysis/base.py
```diff
@@ -220,7 +224,176 @@ def __setstate__(self, state):
         self.data = state
 
 
-class AnalysisBase(object):
+class AnalysisCollection(object):
```
richardjgowers: I don't think we need to inherit from object any more either (it's implicit)

package/CHANGELOG
```diff
@@ -26,6 +25,8 @@ Fixes
     (Issue #3336)
 
 Enhancements
+  * Add an `AnalayisCollection` class to perform multiple analysis on the same
```
richardjgowers: typo on Analysis

richardjgowers (commented):
looking at it, it initially seems strange that Single inherits from Collection, but for the code it does make sense.

@PicoCentauri have you thought about how results could be accessed from the `AnalysisCollection` object?  Or is the intent to always access results from the individual classes again?

package/MDAnalysis/analysis/base.py
PicoCentauri: Yes, I agree and I will change this.

package/MDAnalysis/analysis/base.py
PicoCentauri: Agree

package/MDAnalysis/analysis/base.py
PicoCentauri: agree

package/CHANGELOG
PicoCentauri: cheers

package/MDAnalysis/analysis/base.py
PicoCentauri: No, this is just a leftover from the code before. We can also remove it I think, if we do not support 0.15.0 anymore.

PicoCentauri:
Thanks to all of you for your comments and suggestions and sorry for my later answer.

I agree that the current design is a bit confusing. I thought about the `ABC` abstract base class. I think this does not really fit here. The reason is that `AnalysisCollection` works with and on `AnalysisBase` instances.

A cleaner way for reading the code would be the idea that @yuxuanzhuang suggested that we provide a (private ?) `_Runner` class that basically is the code of the `AnalysisCollection` and the `setup_frames`. This `_Runner` will analyze one or more `AnalysisBase` instances. What do you think?

Regarding some questions and comments:

> I think this solution is the cleanest and doesn't require even more classes which I don't like the idea of. IMO there is already enough of a barrier to entry without every new analysis having to handle 2x new API points.

This was never the idea. There will only be one API point for new Analysis classes.

> @PicoCentauri have you thought about how results could be accessed from the AnalysisCollection object? Or is the intent to always access results from the individual classes again?

The latter. I think accessing the results from the individual instances is already handy. An additional way would be just confusing and blowing up the code.

package/MDAnalysis/analysis/base.py
PicoCentauri: The only thing why I did this to do not copy code of the `run` method between `AnalysisBase` and `AnalysisCollection`. We can remove the inheritance by explicitly leaving the run method in `AnalysisBase` untouched. This would be fine for me!

PicoCentauri:
@orbeckst @hmacdope I would still like to get this feature into main. Before I rebase again I would like to ask for your opinion again. I think the [biggest issue](https://github.com/MDAnalysis/mdanalysis/pull/4017#discussion_r1307329706) to be discussed is the inheritance of `AnalysisBase` and `AnalysisCollection`.

orbeckst:
@PicoCentauri sorry for the following short comment... I've run out of "open source time" for the day: There's a longish discussion on parallel analysis in #4158 and it would be good to hear your opinion how the proposed feature could/should work with an AnalysisBase that would try to parallelize with split-apply-combine (see PMDA https://doi.org/10.25080/Majora-7ddc1dd1-013 for the simple idea).

marinegor:
Hi @PicoCentauri , thanks so much for the PR, it's a neat idea!

Regarding of being compatible with #4158, I feel like you'd need to change the `_compute` method of your `AnalysisCollection` class instead of `run`, and the rest will be more or less the same.

And I'd probably indeed subclass `AnalysisBase` in this case -- you'd need to re-implement `__init__`, `_single_frame`, `_prepare` and `_conclude`, and `run` will be then implemented automatically, with an option to be executed on multiple backends too.

PicoCentauri:
Thanks @orbeckst and @marinegor for the comments.

> Regarding of being compatible with https://github.com/MDAnalysis/mdanalysis/issues/4158, I feel like you'd need to change the _compute method of your AnalysisCollection class instead of run, and the rest will be more or less the same.

Changing to `_compute` is an easy fix and I could do it..

> And I'd probably indeed subclass AnalysisBase in this case -- you'd need to re-implement __init__, _single_frame, _prepare and _conclude, and run will be then implemented automatically, with an option to be executed on multiple backends too.

I am not sure If I get your comment correctly. You would  subclass `AnalysisBase` ? But this means we don't have reimplement the methods. Maybe you meant to NOT subclass `AnalysisBase`?

marinegor:
@PicoCentauri I indeed mean "subclass `AnalysisBase`" here, and re-implement the methods that need implementation in subclasses -- otherwise, you need to manually add `run` method to it, and make sure its semantics is maintained the same with `AnalysisBase` (I assume this was your intention). 

But if you subclass `AnalysisBase` here, you only need to add `_single_frame`, `_prepare` and `_conclude` methods.
`_single_frame` would run `_single_frame` of all analysis objects from the collection, `_prepare` would initialize their `_prepare` methods, and `_conclude` would aggregate the computed results back.

PicoCentauri:
Sorry for the lack of updates for a while. I addressed your concerns and moved the implementation of the `run` method into a standalone function that is used by the `AnalysisBase` as well as the `AnalysisCollection`.

marinegor:
hey @PicoCentauri , the parallelization PR that we've mentioned earlier has been merged into develop (https://github.com/MDAnalysis/mdanalysis/pull/4162), and your `AnalysisCollection` would be extremely helpful here. Reason is, parallel analysis doesn't provide any real speedup unless processing of a single frame is noticeably slower that reading it from the disk -- but if we're running an `AnalysisCollection`, it's almost guaranteed!

Could you perhaps have a look at your code after you merge `develop` into it? I can't yet see how your approach would interact with the current parallelization protocol (you can look at the diagram here: https://docs.mdanalysis.org/dev/documentation_pages/analysis/parallelization.html#split-apply-combine), but it might actually work well.

If this doesn't work, perhaps I'd suggest implementing `AnalysisCollection` as `AnalysisBase` subclass, and in its `_single_frame` running each subclass's `_single_frame()` and manually updating the trajectory -- this way, you will likely get the parallelization with multiprocessing as soon as you figure out how to aggregate the results. But it's more tedious, and I'm happy to discuss it in more details if you want.

marinegor:
@PicoCentauri just wondering, are you planning to keep working on this PR, and do you perhaps need any help with that?

Reason I'm writing is, I keep coming back to this PRs idea, and actually find it really cool, especially in line with streaming approaches and basically all scenarios when reading a trajectory is expensive, but you still want to run many different analysis classes :)

PicoCentauri:
Hey, sorry. I kinda forgot about it.

I think, I can do another round of updates based on my other implementation that I am successfully using in [MAICoS](https://github.com/maicos-devel/maicos/blob/e26c10fb2f0cd31dd6ba2ac9216fe411acb2ab55/src/maicos/core/base.py#L646)

PicoCentauri:
@marinegor I rebased the code against main. I have to test if the logic still works in general. The current issue is that with the current code we will not get a performance boost as the loop over trajectory happens inside.

This most likely requires more refactoring...

marinegor:
> This most likely requires more refactoring...

I agree. I'm also not sure why do you need a separate `_Runner` private class (and I'm not a big fan of it tbh). 

Could you perhaps elaborate why is that a problem to inherit from `AnalysisBase`, in `init` listing eg all functions that will initialize subruns, and in it's `_single_frame` calling all subruns `_single_frame` manually? And same with `_conclude`.

PicoCentauri:
> I agree. I'm also not sure why you need a separate _Runner private class (and I'm not a big fan of it tbh).

I see. You like more a pure function? I think a class is cleaner. You directly see when you inspect the class where stuff comes from. But one can also just do a private function. Take a look at the long discussion above. We had a lot of intense discussions where to place this piece of code.

> Could you perhaps elaborate why that is a problem to inherit from AnalysisBase, in init listing eg all functions that will initialise subruns, and in its _single_frame calling all subruns _single_frame manually? And the same with _conclude.

I may take another look. I am quite busy before Christmas, but I may find the time. And tbh, after two years, and a lot of rounds of rebases and review that lead to nothing, my motivation to invest a lot of additional time is very limited...

marinegor:
>I see. You like more a pure function? 

I mean, I don't like `AnalysisBase` inheriting some other class, and would prefer a flat inheritance structure rather than hierarchical that you're proposing. Also making `AnalysisBase` subclass of something else is a substantial change to the library's (internal) API, which definitely requires more consideration.

>I may take another look. I am quite busy before Christmas, but I may find the time. And tbh, after two years, and a lot of rounds of rebases and review that lead to nothing, my motivation to invest a lot of additional time is very limited...

I understand that completely, thanks for saying that. I'll also take a look myself and (hopefully) materialize my words in some code :) Happy holidays!

PicoCentauri:
> I mean, I don't like AnalysisBase inheriting from some other class, and would prefer a flat inheritance structure rather than the hierarchical one you're proposing. Also, making AnalysisBase a subclass of something else is a substantial change to the library's (internal) API, which definitely requires more consideration.

Sure, we can change the design if you are not happy. My idea (which was the result of the discussions above) is that having a runner class instead of a single function might be cleaner if one in the future wants to extend the running methods as they are shared by the `AnalysisBase` and the `AnalysisCollection`. In the beginning, I had the code inside the `AnalysisBase`, but then the inheritance structure becomes very opaque as the running code should also be used by the `AnalysisCollection`... 

> I understand that completely, thanks for saying that. I'll also take a look myself and (hopefully) materialise my words in some code :) Happy holidays!

Thanks, holidays some time to go :-). I am still happy to work on this project if there is a common interest to get this merged.

PR #4028, opened 2023-02-15
change _single_frame by using timeseries

### LINKED ISSUE #3993: [ENH]: Change MSD code to accumulate values using `.timeseries`.

SophiaRuan:
Fixes #3993

 - changed _single_frame() function, using .timeseries instead of atom group positions.

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/analysis/msd.py
+++ b/package/MDAnalysis/analysis/msd.py
@@ -350,7 +350,7 @@ class EinsteinMSD(AnalysisBase):
         start, stop, step = self._trajectory.check_slice_indices(
             self.start, self.stop, self.step)
         self._position_array = self._trajectory.timeseries(
-            start=start, stop=stop, step=step, order='fac')
+            start=start, stop=stop+1, step=step, order='fac')
         self._position_array = \
             self._position_array[:, :self.n_particles, :self.dim_fac]
 
--- a/package/MDAnalysis/analysis/msd.py
+++ b/package/MDAnalysis/analysis/msd.py
@@ -348,9 +348,7 @@ class EinsteinMSD(AnalysisBase):
         #     (self.n_frames, self.n_particles, self.dim_fac))
         # self.results.timeseries not set here
         start, stop, step = self._trajectory.check_slice_indices(self.start, self.stop, self.step)
-        self._position_array = self._trajectory.timeseries()
-        self._position_array = self._position_array[start: stop: step]
-        self._position_array = np.transpose(self._position_array, (1, 0, 2))
+        self._position_array = self._trajectory.timeseries(start=start, stop=stop+1, step=step, order='fac')
         self.results.timeseries = self._position_array
 
     def _parse_msd_type(self):
--- a/package/MDAnalysis/analysis/msd.py
+++ b/package/MDAnalysis/analysis/msd.py
@@ -347,9 +347,12 @@ class EinsteinMSD(AnalysisBase):
         # self._position_array = np.zeros(
         #     (self.n_frames, self.n_particles, self.dim_fac))
         # self.results.timeseries not set here
-        start, stop, step = self._trajectory.check_slice_indices(self.start, self.stop, self.step)
-        self._position_array = self._trajectory.timeseries(start=start, stop=stop, step=step, order='fac')
-        self._position_array = self._position_array[:, :self.n_particles, :self.dim_fac]
+        start, stop, step = self._trajectory.check_slice_indices(
+            self.start, self.stop, self.step)
+        self._position_array = self._trajectory.timeseries(
+            start=start, stop=stop, step=step, order='fac')
+        self._position_array = \
+            self._position_array[:, :self.n_particles, :self.dim_fac]
 
     def _parse_msd_type(self):
         """ Sets up the desired dimensionality of the MSD.
--- a/package/MDAnalysis/analysis/msd.py
+++ b/package/MDAnalysis/analysis/msd.py
@@ -374,7 +374,7 @@ class EinsteinMSD(AnalysisBase):
         """
         # shape of position array set here, use span in last dimension
         # from this point on
-        self._position_array[self._frame_index] = (
+        self._position_array[:, self._frame_index] = (
                  self.ag.positions[:, self._dim])
 
     def _conclude(self):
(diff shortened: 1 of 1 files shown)


hmacdope (changes requested):
The `_single_frame` method is called repeatedly over the trajectory by `run()` to access the values in each frame. The equivalent is done in `timeseries()` so you don't need to call timeseries every frame, only once. `_prepare()` would be a good place to do this. Have a look at the `AnalysisBase` class for hints and check the results of any failing tests for more clues.

SophiaRuan:
> The `_single_frame` method is called repeatedly over the trajectory by `run()` to access the values in each frame. The equivalent is done in `timeseries()` so you don't need to call timeseries every frame, only once. `_prepare()` would be a good place to do this. Have a look at the `AnalysisBase` class for hints and check the results of any failing tests for more clues.

Thank you for the hints! I submitted another change to this pull request.

SophiaRuan:
Hi, I tried a few times to solve the array shape issue in `_prepare()`. I tried to make the `_position_array` to be exactly the same shape as the original one. However, I still got errors after a few attempts. Is the shape of the array the only issue that causes the error?

SophiaRuan:
@hmacdope I have tried to change the` _prepare()` by using `_trajectory.timeseries()`. I also tried to add the `start`, `stop`, `step` to solve the array shape issue. Is there any other thing that I didn't notice? Thank you for your help!

orbeckst:
@SophiaRuan , since PR #4004 was closed, your PR is now considered the active one addressing issue #3993 .

package/MDAnalysis/analysis/msd.py
```diff
@@ -344,9 +344,14 @@ def _prepare(self):
         # these need to be zeroed prior to each run() call
         self.results.msds_by_particle = np.zeros((self.n_frames,
                                                   self.n_particles))
-        self._position_array = np.zeros(
-            (self.n_frames, self.n_particles, self.dim_fac))
+        # self._position_array = np.zeros(
+        #     (self.n_frames, self.n_particles, self.dim_fac))
         # self.results.timeseries not set here
+        start, stop, step = self._trajectory.check_slice_indices(self.start, self.stop, self.step)
+        self._position_array = self._trajectory.timeseries()
```
hmacdope: you can use the arguments to `timeseries` to control start, stop and step.

hmacdope: Be careful of the inclusive range bug (#3893)

package/MDAnalysis/analysis/msd.py
```diff
@@ -344,9 +344,14 @@ def _prepare(self):
...
+        self._position_array = self._position_array[start: stop: step]
+        self._position_array = np.transpose(self._position_array, (1, 0, 2))
```
hmacdope: You can use the `order` argument to `timeseries` to control the output order so that you don't need this transpose.

hmacdope (changes requested):

package/MDAnalysis/analysis/msd.py
SophiaRuan: Thank you for your suggestions! I will try to fix them.

SophiaRuan:
Hi @hmacdope @orbeckst, I already looked into the #3893 and added the `timeseries ` arguments `start`,` stop`, `step` and `order`. However, the shape is still not right. What else did I ignore?

hmacdope:
I think it's to do with the timeseries indexing being inclusive. See my other comment. You may need to add one to the index

hmacdope:
Sorry I meant remove one.

SophiaRuan:
@hmacdope I've tried to remove 1 from `stop` but there's even 1 more test that failed in `msd_test.py`.

SophiaRuan:
Hi @hmacdope and @orbeckst , I made some changes by slicing the array with `n_particles` and `dim_fac`. There are more tests passed, but still, some checks were not successful. Could you take a look? Thank you very much!

MohitKumar020291:
@SophiaRuan As an advice I will say try removing `_single_frame` as it is one of the objective of issue , give start , stop , step in `__init__` . After defining `_position_array` into `_prepare` , you may need give these coordinates into `conclude` . Please keep it as beginner advice. Thanks.

SophiaRuan:
> @SophiaRuan As an advice I will say try removing `_single_frame` as it is one of the objective of issue , give start , stop , step in `__init__` . After defining `_position_array` into `_prepare` , you may need give these coordinates into `conclude` . Please keep it as beginner advice. Thanks.

Hi @MohitKumar020291, thank you for your suggestion! I'll look into that!

SophiaRuan:
> @SophiaRuan As an advice I will say try removing `_single_frame` as it is one of the objective of issue , give start , stop , step in `__init__` . After defining `_position_array` into `_prepare` , you may need give these coordinates into `conclude` . Please keep it as beginner advice. Thanks.

Hi @MohitKumar020291 @hmacdope, I am not sure if I need to remove `_single_frame` as I thought this function is a standard function defined in `Analysisbase` for collecting data across different frames of the trajectory.

MohitKumar020291:
@SophiaRuan There is something like that the issue want to insert `Reader.timeseries` but `_single_frame` uses `Results.timeseries`. I am also bit confused that `run` uses `_single_frame` so , how can we remove that? But I think (not sure) that it may solved by doing variations in `run` . Please check closed [https://github.com/MDAnalysis/mdanalysis/pull/4004](url) . Although it is closed but I had good advices from mentors. Thanks.

orbeckst:
@SophiaRuan I am going to close your PR, following the discussion in #3993. Sorry that we asked you to do all the work. I really appreciated how you asked important questions that ultimately convinced us that we had not thought the suggested approach through.

Please ping @hmacdope and myself once you have a new PR up so that we can give you feedback.

SophiaRuan:
> [ENH]: Add a progress bar to postprocessing in `_conclude()` functions in MSD.py #4070

No problem! I learned a lot from this issue as well! I will try the following issue #4070.

PR #4030, opened 2023-02-18
Fix: EDRReader fails when reading only one step

### LINKED ISSUE #3999: EDR reader fails when only one step is present
## Expected behavior ##

It does not fail, reads the EDR without problems

## Actual behavior ##

The following error:
```
StopIteration                             Traceback (most recent call last)
Cell In [2], line 1
----> 1 a=EDRReader('/home/ebosoni/Downloads/system_npt-skPu.edr', convert_units=False)

File /shared/envs/base_nextmol_func/lib/python3.10/site-packages/MDAnalysis/auxiliary/EDR.py:324, in EDRReader.__init__(self, filename, convert_units, **kwargs)
    322 # attribute to communicate found energy terms to user
    323 self.terms = list(self.data_dict.keys())
--> 324 super(EDRReader, self).__init__(**kwargs)

File /shared/envs/base_nextmol_func/lib/python3.10/site-packages/MDAnalysis/auxiliary/base.py:321, in AuxReader.__init__(self, represent_ts_as, auxname, cutoff, **kwargs)
    319 if self.time_selector is not None and self.constant_dt:
    320     self.auxstep._initial_time = self.time
--> 321     self._read_next_step()
    322     self.auxstep._dt = self.time - self.initial_time
    323     self.rewind()

File /shared/envs/base_nextmol_func/lib/python3.10/site-packages/MDAnalysis/auxiliary/EDR.py:374, in EDRReader._read_next_step(self)
    372 else:
    373     self.rewind()
--> 374     raise StopIteration

StopIteration: 
```

Unfortunately I can not upload here an .edr, if you need it I can send it through other menas. 

## WORKAROUND

The crash actually happens in the `if self.time_selector is not None and self.constant_dt`, therefore I found that setting `constant_dt=False` avoids the error. However I'm not sure you want to recommend this as a solution...also because I struggled quite a bit to understand the meaning of  `constant_dt` from the documentation and I'm still failing to understand the implications of this parameter. For instance, for the `EDR reader` we have the  `time_selector` set, isn't it enough to catch possible cases of not constant dt?

- Which version are you using? Version '2.4.1'
- Which version of Python? 3.10.6

BFedder:
Fixes #3999

 - `the _read_next_step()` method now checks if only one frame is present. If so, the `StopIteration` exception is not raised.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/auxiliary/EDR.py
```diff
@@ -362,6 +362,7 @@ def _read_next_step(self) -> EDRStep:
         StopIteration
             When end of auxiliary data set is reached.
         """
+        SINGLE_FRAME = True if self.n_steps == 1 else False
```
RMeli:
```suggestion
        SINGLE_FRAME = self.n_steps == 1
```

BFedder: Thanks! I considered doing that, but then decided against it because "explicit is better than implicit". I only have a mild preference here though, happy to change it if you prefer the shorter version

RMeli: What about "Beautiful is better than ugly." or "Simple is better than complex."? ;) Jokes aside, some linters complain about [simplifiable-if-expression](https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/simplifiable-if-expression.html).

BFedder: Haha, fair enough - thanks :)

hmacdope (approved):
LGTM :)

package/MDAnalysis/auxiliary/EDR.py
```diff
@@ -371,7 +372,8 @@ def _read_next_step(self) -> EDRStep:
             return auxstep
         else:
             self.rewind()
-            raise StopIteration
+            if not SINGLE_FRAME:
+                raise StopIteration
```
IAlibay:
```suggestion
            if self.n_steps > 1:
                raise StopIteration
```

package/MDAnalysis/auxiliary/EDR.py
```diff
@@ -362,6 +362,7 @@ def _read_next_step(self) -> EDRStep:
         StopIteration
             When end of auxiliary data set is reached.
         """
+        SINGLE_FRAME = self.n_steps == 1
```
IAlibay:
```suggestion
```

IAlibay (changes requested):
If n_steps is defined _a priori_ then you don't need to also define SINGLE_FRAME?

BFedder:
Good point, cheers!

IAlibay (approved):
CI failure is fixed in develop, I'm happy with things as-is.

PR #4031, opened 2023-02-19
Nojump

### LINKED ISSUE #3703: Nojump transform using the “hybrid unwrap” scheme

MDAnalysis does not have a transformation to create “unwrapped” no-jump trajectories where particles are never put back in the primary unit cell. This is a problem for MSD calculations #3179 and analysis/visualization of oligomeric assemblies.

The preprint https://arxiv.org/abs/2111.12052 *A Reversible Unwrapping Algorithm for Constant Pressure Molecular Dynamics Simulations* by [Martin Kulke](https://arxiv.org/search/physics?searchtype=author&query=Kulke%2C+M), [Josh V Vermaas](https://arxiv.org/search/physics?searchtype=author&query=Vermaas%2C+J+V) shows a simple algorithm for orthorhombic boxes that accurately unwrap long NPT trajectories. It improves over work by von Bülow et al by also maintaining correct bond distances in all instances. It is implemented in various PBC packages for VMD. 

The authors do not provide an algorithm for triclinic boxes in the paper but maybe @jvermaas has some ideas that didn’t make it into the paper…?

Do not do anything and have users use native tools such as `gmx trajectory -nojump` or VMD with pbctools/fastpbc/qwrap

jvermaas:
Fixes #3703 

 - Defines a "NoJump" transform, using the same logic as the position averaging transformation. Rather than adding frames together, the nojump transformation looks "back" to extrapolate what the new unwrapped coordinates should be. Uses the algorithm from [Kulke & Vermaas](https://doi.org/10.1021/acs.jctc.2c00327)

PR Checklist
------------
 - [x] Tests? Yup. It works with non-orthogonal boxes.
 - [x] Docs?
 - [x] CHANGELOG updated?


package/MDAnalysis/transformations/nojump.py
```diff
@@ -0,0 +1,115 @@
...
+# MDAnalysis: A Toolkit for the Analysis of Molecular Dynamics Simulations.
+# J. Comput. Chem. 32 (2011), 2319--2327, doi:10.1002/jcc.21787
+#
+
+"""\
+No Jump Trajectory Unwrapping --- :mod:`MDAnalysis.transformations.nojump` 
+=======================================================================================
+
+Unwraps the trajectory such that atoms never move more than half a periodic box length.
+The algorithm used is based on Kulke & Vermaas 2022, 10.1021/acs.jctc.2c00327
```
orbeckst: Cite  using the bookstore Tex extension (sorry, search for examples)

package/MDAnalysis/transformations/nojump.py
```diff
@@ -0,0 +1,115 @@
...
+.. autoclass:: NoJump
+
+"""
+import numpy as np
+import warnings
+
+from .base import TransformationBase
+
+
+class NoJump(TransformationBase):
```
orbeckst, line 46: Add a due credit decorators with your DOI, please look for `@due`

package/MDAnalysis/transformations/nojump.py
```diff
@@ -0,0 +1,115 @@
...
+        if self.check_c and np.abs(self.old_frame - ts.frame) != 1:
+            warnings.warn(
+                "NoJump transform is only accurate when positions"
+                "do not move by more than half a box length."
+                "Currently jumping between frames with a step of more than 1."
+                "This might be fine, but depending on the trajectory stride,"
+                "this might be inaccurate.",
+                Warning,
+            )
+        # Remember that @ is a shorthand for matrix multiplication.
```
orbeckst: No need for the comment, by now devs have caught up on modern Python ;-)

testsuite/MDAnalysisTests/transformations/test_nojump.py
```diff
@@ -0,0 +1,49 @@
+###
+
+import numpy as np
+import pytest
+from numpy.testing import assert_array_almost_equal
+
+import MDAnalysis as md
```
orbeckst: We traditionally import as mda, please change for consistency

testsuite/MDAnalysisTests/transformations/test_nojump.py
```diff
@@ -0,0 +1,49 @@
...
+    transformation = NoJump(check_continuity=False)
+    u.trajectory.add_transformations(transformation)
+    return u
+
+def test_nojump_fwd(nojump_universes):
+    """
+    Test if the nojump transform is returning the correct
+    values when iterating forwards over the sample trajectory.
+    """
+    ref_matrix_fwd1 = np.asarray([-3.42261, 11.28495, -1.37211])
```
orbeckst: What are the red values and how did you obtain them? Add a comment for provenance.

orbeckst (changes requested):
Thank you! Please see inline for quick drive-by comments. (Literally while on shuttle to conference center…)

package/MDAnalysis/transformations/nojump.py
jvermaas: Yeah, that comment was for me. :D

package/MDAnalysis/transformations/nojump.py
jvermaas: I think I have set this up correctly now.

testsuite/MDAnalysisTests/transformations/test_nojump.py
jvermaas: Done

testsuite/MDAnalysisTests/transformations/test_nojump.py
jvermaas: My workflow for testing this was probably non-ideal. I basically made water orthogonal and non-orthogonal water boxes, checked that the results were equivalent to what I got from VMD/fastpbc. I then ran the algorithm on existing triclinic systems, and did some sanity checks. These are two atomic positions that happened to move the most. Comments are now included to highlight this.

hmacdope: @jvermaas  May I suggest making a test case with a single particle that travels at a constant speed through the box? That way it should be easy to verify the correcness of the unwrapping. You can do this directly  in MDAnalysis using

```python
import MDAnalysis 
import numpy as np

N = 1
u = mda.Universe.empty(N, trajectory=True)
coordinates = np.empty((1000,  # number of frames
                        u.atoms.n_atoms,
                        3))
coordinates = ... # generate wrapped coordinates for a particle moving at a constant speed in the box. 
# I can help you with this if you need. 
u.load_new(coordinates, order='fac')
```
You can then make this universe a test fixture.

package/MDAnalysis/transformations/nojump.py
```diff
@@ -0,0 +1,127 @@
...
+        :filter: False
+        :style: MDA
+
+        Kulke2022
+
+    """
+
+    @due.dcite(Doi("10.1021/acs.jctc.2c00327"),
+               description="Works through the orthogonal case, "
+               "and proposes the non-orthogonal approach.", path=__name__)
```
orbeckst: I can't quite remember how the description shows up. The best thing for you would be to try it out https://docs.mdanalysis.org/stable/documentation_pages/references.html#citations-using-duecredit and see if you're happy with how your paper is presented to users.

package/MDAnalysis/transformations/nojump.py
orbeckst: Just remove the comment to keep the code concise.

package/MDAnalysis/transformations/nojump.py
```diff
@@ -0,0 +1,127 @@
...
+            self.old_frame = ts.frame
+            return ts
+        if self.check_c and np.abs(self.old_frame - ts.frame) != 1:
+            warnings.warn(
+                "NoJump transform is only accurate when positions"
+                "do not move by more than half a box length."
+                "Currently jumping between frames with a step of more than 1."
+                "This might be fine, but depending on the trajectory stride,"
+                "this might be inaccurate.",
+                Warning,
```
orbeckst: We typically use UserWarning instead of bare Warning.

(We have a few dedicated warning classes in https://docs.mdanalysis.org/stable/documentation_pages/exceptions.html but we try to not proliferate more)

testsuite/MDAnalysisTests/transformations/test_nojump.py
jvermaas: Oooh... That is cleverer @hmacdope . I'll let you know how I make out. I'll probably need to look at how to add PBC information to a universe, but there is probably documentation somewhere.

package/MDAnalysis/transformations/nojump.py
```diff
@@ -0,0 +1,127 @@
...
+        super().__init__(max_threads=max_threads, parallelizable=parallelizable)
+        self.prev = None
+        self.old_frame = 0
+        self.check_c = check_continuity
+
+    def _transform(self, ts):
+        if self.prev is None:
+            self.prev = ts.positions @ np.linalg.inv(ts.triclinic_dimensions)
+            self.old_frame = ts.frame
+            return ts
```
orbeckst: How does this work when I go through the trajectory a second time:
```python
u.add_transformation(NoJump) 
[ts for ts in u.trajectory]          
[ts for ts in u.trajectory[2:]]     # once the trajectory resets, will prev point to the correct starting value?
```
I think there should be a test that it's still correct the second time around. Keeping state is tricky business and our readers' behavior is not always obvious (for instance, they should rewind when you start iteration).

testsuite/MDAnalysisTests/transformations/test_nojump.py
```diff
@@ -0,0 +1,57 @@
+###
+
```
orbeckst: remove

testsuite/MDAnalysisTests/transformations/test_nojump.py
```diff
@@ -0,0 +1,57 @@
...
+import numpy as np
+import pytest
+from numpy.testing import assert_array_almost_equal
```
orbeckst: We use [`assert_allclose`](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html) nowadays but haven't updated everything.

testsuite/MDAnalysisTests/transformations/test_nojump.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def test_nojump_fwd(nojump_universes):
+    """
+    Test if the nojump transform is returning the correct
+    values when iterating forwards over the sample trajectory.
+    """
+    # These were determined based on determining what the TRICLINIC system
+    # would return on a validated version that worked on triclinic and
+    # orthogonal systems for a small water trajectory.
```
orbeckst: mention VMD and fastpbc explicitly (include version numbers if you have!); we have other cases that we checked against a VMD implementation. We don't consider it shameful to give credit to other software ;-). More importantly, the more detail is provided for validation, the more confidence someone has in the correctness.

testsuite/MDAnalysisTests/transformations/test_nojump.py
hmacdope: I am going to make this comment a blocking review @jvermaas, let me know if you need any help, feel free to just ping me.

hmacdope (changes requested):
Blocking on a constant velocity test trajectory.

testsuite/MDAnalysisTests/transformations/test_nojump.py
orbeckst: I really like @hmacdope 's idea. 

$$
\mathbf{r}(t) = \mathbf{v}_0 t + \mathbf{r}_0
$$

with some odd angle of the initial velocity and a smallish box and then wrap the straight shot trajectory with the  `wrap()` transform.

testsuite/MDAnalysisTests/transformations/test_nojump.py
```diff
@@ -0,0 +1,57 @@
...
+    # These are specific to atoms 166 and 362, which moved the most in the
+    # trajectory.
+    ref_matrix_fwd1 = np.asarray([-3.42261, 11.28495, -1.37211])
```
orbeckst: Instead of calling these variables "matrix", call them something like `ref_last_pos_166` or something that makes clearer what they are.

testsuite/MDAnalysisTests/transformations/test_nojump.py
hmacdope: RE setting PBC the following should work to give you a 10x10x10 orthorhombic box using the `dimensions` argument. 
```python
import MDAnalysis 
import numpy as np

N = 1
u = mda.Universe.empty(N, trajectory=True)
coordinates = np.empty((1000,  # number of frames
                        u.atoms.n_atoms,
                        3))

u.load_new(coordinates, order='fac', in_memory=True, dimensions=[10,10,10, 90.0, 90.0,90.0])
```

testsuite/MDAnalysisTests/transformations/test_nojump.py
```diff
@@ -0,0 +1,57 @@
...
+    ref_matrix_fwd2 = np.asarray([3.674243, -8.725193, -0.07884017])
+    size = (
+        nojump_universes.trajectory.ts.positions.shape[0],
+        nojump_universes.trajectory.ts.positions.shape[1],
+        len(nojump_universes.trajectory),
+    )
+    parr = np.empty(size)
+    for ts in nojump_universes.trajectory:
+        parr[..., ts.frame] = ts.positions.copy()
```
orbeckst: I think you can replace the explicit loop with
```python
transformed_coordinates = u.trajectory.timeseries()
```
By default, it has shape `(n_frames, n_atoms, 3)` so you have to change the lines below.

orbeckst (changes requested):
Thanks for the fixes. 

- Integrate into docs by adding to [package/doc/sphinx/source/documentation_pages/trajectory_transformations.rst ](https://github.com/MDAnalysis/mdanalysis/blob/develop/package/doc/sphinx/source/documentation_pages/trajectory_transformations.rst) and [package/doc/sphinx/source/documentation_pages/transformations](https://github.com/MDAnalysis/mdanalysis/tree/develop/package/doc/sphinx/source/documentation_pages/transformations). Then check that docs look ok https://mdanalysis--4031.org.readthedocs.build/en/4031/index.html
- I think @hmacdope 's idea of creating a simple trajectory as a test case is really good. Can we do that?
- As with any transformations that keep state we need to tell the user under which circumstances it should work and under which it will fail. FOr example, [positionaveraging](https://docs.mdanalysis.org/stable/documentation_pages/transformations/positionaveraging.html) has the `check_reset` flag and explains different cases. We need to have some documentation that outlines when/how to use it. If something does not make sense (see below) then we should explicitly say so, *especially* of we're not able to warn or raise an error.
- Other comments inline.


As a side note:  The user can almost certainly break the algorithm by breaking the frame sequence
```python
[ts for ts in u.trajectory[[0, 5, 2, 10, 10, 3]] ]
```
but I don't know what we could do about it. We could check that ts.frame is monotonically increasing (or decreasing — `u.trajectory[::-1]` should work thanks to time reversibility... NPT MD is technically reversible, isn't it?) Could we check and then raise an error??

testsuite/MDAnalysisTests/transformations/test_nojump.py
jvermaas: I scrapped the old benchmark and replaced it with a synthetic one. I think its still intelligible, and I can explain why the non-orthogonal number is what it is.

package/MDAnalysis/transformations/nojump.py
jvermaas: It makes ZERO sense to unwrap an already unwrapped trajectory, and so ideally the transformation should only ever act once... I think. I could adopt something similar to what positionaveraging does, and try and remember what direction the steps are going in. However, in principle, unwrapping an already unwrapped trajectory is basically a no-op, since the round term will always be zero. I suppose I'd just need to enact a reset.

testsuite/MDAnalysisTests/transformations/test_nojump.py
jvermaas: Yeah, the TRICLINIC system just was too ugly. The synthetic system is easier to understand.

package/MDAnalysis/transformations/nojump.py
orbeckst: Note that the trajectory will not stay unwrapped. It’s not in memory, the coordinates are read from disk again. (Sorry if I misunderstood your comment… typing quickly before they really force me to switch to airplane mode)

orbeckst: Only if the user is using the MemoryReader then changes are permanent but that’s handled automatically with transformations.

jvermaas: Hmm, ok. Now it'll throw a warning that you are going in the opposite direction (or with an unequal stride). When it encounters this, it will reset the unwrapping procedure.

testsuite/MDAnalysisTests/transformations/test_nojump.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+    u = nojump_universe
+    dim = np.asarray([1, 1, 1, 90, 60, 90], np.float32)
+    workflow = [mda.transformations.boxdimensions.set_dimensions(dim), NoJump()]
+    u.add_transformations(*workflow)
+    transformed_coordinates = u.trajectory.timeseries()[0]
+    # Step is 1 unit every 3 steps. After 99 steps from the origin,
+    # we'll end up at 33. However, since the unit cell is non-orthogonal,
+    # we'll end up at a distorted place.
+    assert_allclose(transformed_coordinates[99], 33 * np.array([0.5, 1, np.sqrt(3)/2]))
```
hmacdope: Same here, match the whole array.
- Do we know the exact for of the skew factor `33 * np.array([0.5, 1, np.sqrt(3)/2]` is correct? If so I would love to see the working in the code.

testsuite/MDAnalysisTests/transformations/test_nojump.py
```diff
@@ -0,0 +1,54 @@
...
+    values when iterating forwards over the sample trajectory.
+    """
+    u = nojump_universe
+    dim = np.asarray([1, 1, 1, 90, 90, 90], np.float32)
+    workflow = [mda.transformations.boxdimensions.set_dimensions(dim), NoJump()]
+    u.add_transformations(*workflow)
+    transformed_coordinates = u.trajectory.timeseries()[0]
+    # Step is 1 unit every 3 steps. After 99 steps from the origin,
+    # we'll end up at 33.
+    assert_allclose(transformed_coordinates[99], 33 * np.ones(3))
```
hmacdope: Make this test cover the whole array, matching along its whole length.

hmacdope (changes requested):

testsuite/MDAnalysisTests/transformations/test_nojump.py
jvermaas: Oof. This is a pain to do for how I constructed the non-orthogonal case. The z position shows this the most clearly.

```
[[ 0.          0.          0.        ]
 [ 0.33333334  0.33333334  0.33333334]
 [ 0.6666667   0.6666667   0.6666667 ]
 [ 0.5         1.          0.8660254 ]
 [ 0.8333334   1.3333334   1.1993587 ]
 [ 1.1666667   1.6666667   1.5326921 ]
 [ 1.          2.          1.7320508 ]
...snip...
```

The jumps aren't all the same size, since the skewed box means that a "constant" displacement in wrapped space meets up with the next box over in a weird place. Its too late at night to think about how this works.

hmacdope: Feel free to change your box vectors freely to make the working easier if it helps. 😀

jvermaas: Yeah, except that you REALLY want to test on non-orthogonal boxes, since its harder. The path I chose is just inherently jagged. I've made another test based on a straight line.

package/MDAnalysis/transformations/nojump.py
```diff
@@ -0,0 +1,148 @@
...
+        if (
+            self.check_c
+            and self.older_frame != "A"
+            and (self.old_frame - self.older_frame) != (ts.frame - self.old_frame)
+        ):
+            warnings.warn(
```
hmacdope, line 135: Test this warning

package/MDAnalysis/transformations/nojump.py
```diff
@@ -0,0 +1,148 @@
...
+                "NoJump detected that the interval between frames is unequal."
+                "We are resetting the reference frame to the current timestep.",
+                UserWarning,
+            )
+            self.prev = ts.positions @ np.linalg.inv(ts.triclinic_dimensions)
+            self.old_frame = ts.frame
+            self.older_frame = "A"
+            return ts
+        if self.check_c and np.abs(self.old_frame - ts.frame) != 1:
```
hmacdope, line 144: test this warning.

testsuite/MDAnalysisTests/transformations/test_nojump.py
```diff
@@ -0,0 +1,95 @@
...
+    dim = np.asarray([1, 1, 1, 90, 60, 90], np.float32)
+    workflow = [mda.transformations.boxdimensions.set_dimensions(dim), NoJump()]
+    u.trajectory.add_transformations(*workflow)
+    transformed_coordinates = u.trajectory.timeseries()[0]
+    # Step is 1 unit every 3 steps. After 99 steps from the origin,
+    # we'll end up at 33. However, since the unit cell is non-orthogonal,
+    # we'll end up at a distorted place.
+    assert_allclose(
+        transformed_coordinates[::3],
+        np.outer(np.arange(33.5), np.array([0.5, 1, np.sqrt(3) / 2])),
```
hmacdope, line 88: Do you have working for the `np.sqrt(3) / 2` part?

hmacdope (changes requested):
Query, this only works for orthogonal boxes right? 

If so need to raise an exception on non-orthogonal box. `mda.lib.util.check_box()` should be able to help you identify the box type.

testsuite/MDAnalysisTests/transformations/test_nojump.py
jvermaas: Again, this is as close as you'll get with the test I originally designed for non-orthogonal boxes. That's why I added the wrapped version from a line, since it can test every value easily.

jvermaas:
This code will work for orthogonal OR non-orthogonal boxes. That's why I insisted on testing with non-orthogonal boxes.

jvermaas:
Ok, Codecov is now failing, but I'm covering all the branches by triggering the warnings. I think the service might be temporarily down or something.

orbeckst:
Codecov has lapses, it often does not like it when a lot of requests are submitted. But in general it is pretty random. Sigh.

jvermaas:
The one other thing I'd like some assistance with, is how we should test if the dimensions don't exist for a particular timestep within a trajectory. Right now, this will bail since you can't invert a matrix that doesn't exist, but I assume users might prefer a friendlier error message.

orbeckst:
Failing is ok and generally better than guessing. Just fail with a `ValueError` and state the timestep frame number that does not have the right data (ie misses a time step).

orbeckst:
Or we could actually use our `mda.exceptions.MissingDataError`. That's probably more consistent with other parts of the code.

jvermaas:
The one other thing I'd like some assistance with, is how we should test if the dimensions don't exist for a particular timestep within a trajectory. Right now, this will bail since you can't invert a matrix that doesn't exist, but I assume users might prefer a friendlier error message.

jvermaas:
So, this is fun. I cannot trigger the warnings. In my testing, which is a script like this:

```python
import MDAnalysis as mda
from MDAnalysis.transformations import NoJump, wrap
import numpy as np
from numpy.testing import assert_allclose

Natom = 1
Nframe = 100
coordinates = np.empty((Nframe,  # number of frames
                        Natom,
                        3))
coordinates[:,0,0] = np.linspace(0,45,Nframe)
coordinates[:,0,1] = np.linspace(0,15,Nframe)
coordinates[:,0,2] = np.linspace(0,10,Nframe)
reference = mda.Universe.empty(Natom, trajectory=True)
reference.load_new(coordinates, order='fac')
wrapped = mda.Universe.empty(Natom, trajectory=True)
wrapped.load_new(coordinates, order='fac')
dim = np.asarray([5, 5, 5, 54, 60, 90], np.float32)
print(wrapped.trajectory.ts.triclinic_dimensions)

wrapped.trajectory.add_transformations(*[mda.transformations.boxdimensions.set_dimensions(dim),wrap(wrapped.atoms),
        NoJump()])
#print(wrapped.trajectory.timeseries()[0])
for ts in wrapped.trajectory[[0,1,2,3,4,6,4,2]]:
        print(ts.frame)
```

Internally, I was printing the frames over which the trajectory was having its transformations applied, and it was applying the transformation on every frame, from 0 to 99. I don't actually understand why this is happening, and would like some explanation of *why* this could possibly be happening, or how I can actually get the warning to be emitted. Do transformations apply over a whole trajectory or something?

orbeckst:
When you use `Universe.empty()`, you will generate a Universe that will hold a trajectory in memory (via the [MemoryReader](https://docs.mdanalysis.org/stable/documentation_pages/coordinates/memory.html)). Transformations behave differently for on-disk trajectories (normal) and in-memory (special):

* on-disk: transformation is called every time the data are read from disk and transformed on the fly
* in-memory: transformations are applied *once* and all the data in memory are stored transformed

I assume that this leads to inconsistencies.

(ALso, I am not sure if this behavior is properly documented anywhere. The MemoryReader is a swiss-army knife but really quite special in many ways.)

jvermaas:
Perfect! Hopefully this one will work!

orbeckst:
@hmacdope could you please be the person in charge here? — this can include telling me to review again etc...

orbeckst:
At the moment a whole bunch of new tests are failing @jvermaas , something about `data` not defined.

jvermaas:
Yeah, I'm on it. I have no idea what the "normal" way of testing this before hand is supposed to be. So far the workflow is to try something, and look at the failure. Weirdly, the exception for missing a box *also* isn't being caught properly.

jvermaas:
Ok, I feel dumb now. I found the documentation that describes this exactly. This is much easier than waiting 5 minutes for the automated process to fail.

orbeckst:
Just for posterity who want to know how to do testing https://github.com/MDAnalysis/mdanalysis/pull/4031#issuecomment-1444597048 , see  https://userguide.mdanalysis.org/stable/testing.html

testsuite/MDAnalysisTests/transformations/test_nojump.py
hmacdope: @jvermaas I would still love to see the working for the non-orthogonal case. why do we end up with `np.sqrt(3)` etc etc. Just so others can follow along (myself included.)

package/MDAnalysis/transformations/nojump.py
```diff
@@ -0,0 +1,158 @@
...
+    ):
+        super().__init__(max_threads=max_threads, parallelizable=parallelizable)
+        self.prev = None
+        self.old_frame = 0
+        self.older_frame = "A"
+        self.check_c = check_continuity
+
+    def _transform(self, ts):
+        if self.prev is None:
+            try:
```
hmacdope: I think this should be seperated into two checks, one that checks `ts.data != None` and another that checks for invertibility as this will be easier for users.

hmacdope (changes requested):
Looking good @jvermaas, just a few nitpicks.

package/MDAnalysis/transformations/nojump.py
jvermaas: I've worked out what the transform should be for the [1::3] and [2::3] cases, and added more text to try and explain this. The "correct" answer is that the sqrt(3)/2 comes from the 60 degree angle set in the dimension. This makes a hexagonal unit cell, which has as a consequence that the origin for adjacent PBC images shifts by an unusual amount. Since in this particular test, I'm transforming a line *in unit cell coordinates*, the actual unwrapped trajectory is zig-zagging a bit along the hexagonal lattice.

@hmacdope , do we actually need to check that `ts.data != None`? The check you highlighted is actually testing if this is the first frame to be unwrapped or not. You can't unwrap the first frame. I have however noticed that the L and Linverse computations can be done together, before this test.

hmacdope:
Sorry for the reviewing delay @jvermaas Could you fix up the failing tests and conflicts and then we will go from there? Hopefully I'll get to it this week.

jvermaas:
No worries. @hmacdope . I must have been too hasty trying to come up with a non-invertible test-case. This one ought to work, and I *think* I fixed the conflicts.

jvermaas:
@hmacdope , just fixed the new conflicts in the authorlist and the changelog. If you get a chance, I think we are good to go now.

hmacdope:
Can you fix the unused import? See the output of black linter. I'll try and review ASAP.

hmacdope:
OMG im so sorry, that import should be there, apologies. Just revert that change.

jvermaas:
From context I thought it was fine. Linter just doesn't like it. :D

amangupta201:
Hey can you please assign me this issue?

hmacdope:
> Hey can you please assign me this issue?

No this is a current in progress PR. Try and look for another issue on the issue tracker.

package/MDAnalysis/transformations/nojump.py
```diff
@@ -0,0 +1,154 @@
...
+        self.check_c = check_continuity
+
+    def _transform(self, ts):
+        L = ts.triclinic_dimensions
+        if L is None:
+            msg = "Periodic box dimensions not provided at step %d." % ts.frame
+            raise NoDataError(msg)
+        try:
+            Linverse = np.linalg.inv(L)
+        except np.linalg.LinAlgError:
```
hmacdope, line 123: Needs test coverage.

testsuite/MDAnalysisTests/transformations/test_nojump.py
```diff
@@ -0,0 +1,181 @@
...
+    Natom = 1
+    Nframe = 100
+    coordinates = np.empty((Nframe, Natom, 3))  # number of frames
+    coordinates[:, 0, 0] = np.linspace(0, 45, Nframe)
+    coordinates[:, 0, 1] = np.linspace(0, 15, Nframe)
+    coordinates[:, 0, 2] = np.linspace(0, 10, Nframe)
+    reference = mda.Universe.empty(Natom, trajectory=True)
+    reference.load_new(coordinates, order="fac")
+    towrap = mda.Universe.empty(Natom, trajectory=True)
+    towrap.load_new(coordinates, order="fac")
```
hmacdope: `reference` and `towrap` are the same? Can't you just use `copy` in that case?

testsuite/MDAnalysisTests/transformations/test_nojump.py
```diff
@@ -0,0 +1,181 @@
...
+        rtol=1.2e-7
+    )
+    assert_allclose(
+        transformed_coordinates[2::3],
+        np.outer(np.arange(32.5), np.array([0.5, 1, np.sqrt(3) / 2])) + 2 * np.ones(3) / 3,
+        rtol=1.2e-7
+    )
+
+
+def test_nojump_constantvel(nojump_constantvel_universe):
```
hmacdope, line 102: You should do this for both an orthogonal and nonorthogonal box

hmacdope (changes requested):
Looking really good, almost there :)

package/MDAnalysis/transformations/nojump.py
jvermaas: Sorry, what? I'm not sure what part of this isn't being tested. __init__ is init, and I do have enough test cases to cover what happens if no periodic box dimensions are provided or they aren't invertible for some reason.

package/MDAnalysis/transformations/nojump.py
jvermaas: This is meant to be tested by test_nojump_constantvel_jumparound. Logically, I think it should be, since I'm going from frame 0 to 1 to 9. This SHOULD emit a warning.

package/MDAnalysis/transformations/nojump.py
jvermaas: Should be in test_nojump_constantvel_skip. Like, I test my warnings. I am unable to understand what you are asking me to do.

package/MDAnalysis/transformations/nojump.py
hmacdope: Codecov report says L109-111 is not covered.

package/MDAnalysis/transformations/nojump.py
hmacdope: No thats okay I'm happy with your warning tests

testsuite/MDAnalysisTests/transformations/test_nojump.py
jvermaas: Since I'll be wrapping and unwrapping `towrap`, and need these to be distinct memory locations so I only mutate one set. My understanding of numpy copy semantics is that you aren't actually sure that you create distinct memory locations for a "trajectory".

testsuite/MDAnalysisTests/transformations/test_nojump.py
jvermaas: Yeah, this is done in `test_nojump_orthogonal_fwd`. Its just a single allclose call in that case, since I can use a single linspace to do the matrix outer product.

package/MDAnalysis/transformations/nojump.py
jvermaas: Ok, I the one that is supposed to cover this is `test_notinvertible`. I'll take a closer look, since that *should* raise a linalgerror if you try and invert a null matrix.

jvermaas: Well I feel dumb now. Apparently mdanalysis is smart enough to say that no dimensions are provided if you pass every box dimension as zero. But if you pass at least one non-zero box dimension and make sure the matrix still isn't invertible, you'll make it to the LinAlgError.

jvermaas:
Ok, @hmacdope , I think we are good now. I used copy to clone the universe so that the wrapping/unwrapping is shown to be reversible against the wrapping algorithm used, and now we actually have a test that will yield a non-invertible box dimension, although codecov isn't being rerun for some reason. I'll fix the authors again, but hopefully for the last time.

hmacdope (approved):
Thanks for all the hard work @jvermaas! Looks great to me. Thank you so much as this is something we have been needing for a long time! @orbeckst have all your concerns been addressed?

package/MDAnalysis/transformations/nojump.py
orbeckst: yes,... we adopted all zeros to mean "no box"... and there was a reasonably good reason that we couldn't always rely on `None`, at least that's what I _want_ to say.

package/MDAnalysis/transformations/nojump.py
```diff
@@ -0,0 +1,154 @@
...
+        self.prev = None
+        self.old_frame = 0
+        self.older_frame = "A"
+        self.check_c = check_continuity
+
+    def _transform(self, ts):
+        L = ts.triclinic_dimensions
+        if L is None:
+            msg = "Periodic box dimensions not provided at step %d." % ts.frame
+            raise NoDataError(msg)
```
orbeckst: Modern-ish Python
```suggestion
            msg = f"Periodic box dimensions not provided at step {ts.frame}"
            raise NoDataError(msg)
```

package/MDAnalysis/transformations/nojump.py
```diff
@@ -0,0 +1,154 @@
...
+        try:
+            Linverse = np.linalg.inv(L)
+        except np.linalg.LinAlgError:
+            msg = "Periodic box dimensions are not invertible at step %d." % ts.frame
```
orbeckst:
```suggestion
            msg = f"Periodic box dimensions are not invertible at step {ts.frame}"
```

orbeckst (changes requested):
Looks really good, juts minor nit-picks
- use modern f-strings
- please add a sentence or two to the doc string about how the algorithm behaves when there are jumps in the trajectory or under which situations it fails; also mention that it works best when using every frame

jvermaas:
@orbeckst , I think we are good now. I've been using C-style formatting for so long ("%03d" is so powerful for keeping NAMD trajectories sorted) that I forget that f-strings exist.

orbeckst (approved):

orbeckst:
Thank you @jvermaas , a very valuable contribution.

@hmacdope all yours now.

PR #4034, opened 2023-02-20
[CI] Update main ci components to use py311 at upper limit

IAlibay:
Most if not all of upstream is not py311 compliant, let's see if this works.

 - 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -41,7 +41,7 @@ jobs:
           PYTHON_VERSION: '3.11'
           PYTHON_ARCH: 'x64'
           BUILD_TYPE: 'wheel'
-          NUMPY_MIN: '1.22.3'
+          NUMPY_MIN: '1.23.2'
           imageName: 'windows-2019'
         Win-Python38-64bit-full-wheel:
           PYTHON_VERSION: '3.8'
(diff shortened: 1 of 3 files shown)


azure-pipelines.yml
```diff
@@ -49,8 +49,8 @@ jobs:
           BUILD_TYPE: 'wheel'
           NUMPY_MIN: '1.21.0'
           imageName: 'windows-2019'
-        Linux-Python310-64bit-full-wheel:
-          PYTHON_VERSION: '3.10'
+        Linux-Python311-64bit-full-wheel:
+          PYTHON_VERSION: '3.11'
```
tylerjereddy, line 53: From a quick glance at the CI, maybe we need to bump `NUMPY_MIN` below to `1.23.2` to match what `pyproject.toml` does at `3.11` (which does look correct).

Probably the same adjustment for the Windows "wheel" job above, which also uses `NUMPY_MIN` to enforce portability check on the built wheel.

IAlibay: oops yeah sorry completely forgot to update those pins

orbeckst:
@tylerjereddy can I leave the PR in your hands, i.e., you'll eventually give an approval (not sure if @IAlibay wants this merged asap or holding off)

IAlibay:
Review would be great

tylerjereddy (approved):
LGTM now, we still have at least one job left at each supported Python version.

The cron job change seems harder to judge--given the use of conda (?) ecosystem there, I wouldn't be surprised if that needs another shim we only discover after merging and having the cron run.

PR #4043, opened 2023-02-26
fix: get_writer errors if filename is none

jandom:
Fixes a small bug

 - catch a case when filename is truthy and format is uninitialized 

PR Checklist
------------
 - [x] Tests? Added a simple unit test, maybe too much mocking, please review
 - [x] Docs? Not applicable
 - [x] CHANGELOG updated? Yes!
 - [x] Issue raised/referenced? Nope


package/CHANGELOG
```diff
@@ -19,6 +19,7 @@ The rules for this file:
  * 2.5.0
 
 Fixes
+  * Fix get_writer format uninitialized (PR #4043)
```
orbeckst: Also add your GH handle to the list of contributors.

orbeckst (commented):
What's the bug, i.e., do you have example code that triggers it?

jandom:
It's a mini-bug that doesn't actually happen in the way the function is called. 

However, if you call the function with just the format argument, it'll error

```
selections.get_writer(None, "blah") 
```

This will error

richardjgowers:
7cpwc6

package/CHANGELOG
orbeckst: still todo

orbeckst (changes requested):
lgtm, just add your GH handle to CHANGELOG

package/CHANGELOG
IAlibay: I'm not going to make this a blocking request - but for the few folks that read the changelog to work out what changed in the package, could I ask for a little bit more information? Something like "Fix uninitialized `format` variable issue when calling `selections.get_writer directly`" would probably go a long way.

orbeckst: Good point, especially when we don't have an issue to point to.

jandom: Gotcha, on it

orbeckst (dismissed):
Just a more explicit CHANGELOG entry, please, see https://github.com/MDAnalysis/mdanalysis/pull/4043/files#r1122512703

orbeckst:
Not sure why CI is not running.

richardjgowers (approved):

jandom:
Thanks for the merge, it's a small contribution but I hope to send more in

PR #4048, opened 2023-03-01
Issue 4039 large dcd

### LINKED ISSUE #4039: Error when loading multiple large DCD trajectories
## Expected behavior ##

I create a Universe with several large trajectory files (size > 2 GB) and use the transfer_to_memory function to load them.

## Actual behavior ##

The transfer_to_memory function fails after loading 53458/117600 frames with the error "OSError: DCD seek failed with DCD error=Normal EOF".

```
Loading frames:  45%|████████████████████████████████████████████████████████▎                                                                   | 53458/117600 [00:07<00:09, 7050.72it/s]
  File "/Users/vladimir/test.py", line 15, in <module>
    u.transfer_to_memory(verbose=True)
  File "/opt/anaconda3/envs/work/lib/python3.9/site-packages/MDAnalysis/core/universe.py", line 626, in transfer_to_memory
    for i, ts in enumerate(ProgressBar(self.trajectory[start:stop:step],
  File "/opt/anaconda3/envs/work/lib/python3.9/site-packages/tqdm/std.py", line 1195, in __iter__
    for obj in iterable:
  File "/opt/anaconda3/envs/work/lib/python3.9/site-packages/MDAnalysis/coordinates/chain.py", line 672, in __next__
    self.ts = self.active_reader[f]
  File "/opt/anaconda3/envs/work/lib/python3.9/site-packages/MDAnalysis/coordinates/base.py", line 828, in __getitem__
    return self._read_frame_with_aux(frame)
  File "/opt/anaconda3/envs/work/lib/python3.9/site-packages/MDAnalysis/coordinates/base.py", line 860, in _read_frame_with_aux
    ts = self._read_frame(frame)  # pylint: disable=assignment-from-no-return
  File "/opt/anaconda3/envs/work/lib/python3.9/site-packages/MDAnalysis/coordinates/DCD.py", line 198, in _read_frame
    self._file.seek(i)
  File "MDAnalysis/lib/formats/libdcd.pyx", line 398, in MDAnalysis.lib.formats.libdcd.DCDFile.seek
OSError: DCD seek failed with DCD error=Normal EOF
```
(53458 frames take up about 2 GB of memory)

``` python
import MDAnalysis as mda

u = mda.Universe(PSF, DCD)

# creating a big trajectory (2.2 GB)
ag = u.select_atoms("all")
with mda.Writer('test.dcd', ag.n_atoms) as w:
    for _ in range(600):
        for ts in u.trajectory:
            w.write(ag)


u = mda.Universe(PSF, 'test.dcd', 'test.dcd')
u.transfer_to_memory(verbose=True)

```

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`): 2.4.2
- Which version of Python (`python -V`)?: Python 3.9.15
- Which operating system?: MacOS Ventura 13.1 (22C65)

richardjgowers:
Fixes #4039

 - use `fio_size_t` for variables related to filesize

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


IAlibay:
yup this works re: bugfix, let's just make sure we squash merge

testsuite/MDAnalysisTests/coordinates/test_dcd.py
```diff
@@ -436,3 +437,39 @@ def test_pathlib():
...
+
+    with mda.Writer(newf, n_atoms=len(u.atoms)) as w:
+        for _ in range(nreps_reqs):
+            for ts in u.trajectory:
+                w.write(u.atoms)
+
+    yield newf, nreps_reqs
+
+
+@pytest.mark.skipif(not os.environ.get('LARGEDCD', False),
```
IAlibay: That's kinda confusing logic, and looks undocumented, are we really expecting to use it?

IAlibay (dismissed):
Do we actually care about that skipif so much?

testsuite/MDAnalysisTests/coordinates/test_dcd.py
richardjgowers: I can just remove the test if you like, it was handy while I was fixing the bug

IAlibay:
FYI lint failure is optional, I might make it print comments instead when I get time to play with the GH API again

testsuite/MDAnalysisTests/coordinates/test_dcd.py
IAlibay: test would be good to keep, I just don't really know why you'd need a skipif that isn't really documented. Did we not already have a high memory flag from the EDR tests? Can we just use that instead?

testsuite/MDAnalysisTests/coordinates/test_dcd.py
```diff
@@ -436,3 +437,38 @@ def test_pathlib():
     # we really only care that pathlib
     # object handling worked
     assert u.atoms.n_atoms == 3341
+
+
+@pytest.fixture
```
orbeckst, line 442: Make a module level fixture so that it really only runs once? Unfortunately will need to use the tmpdir factory

package/MDAnalysis/lib/formats/libdcd.pyx
```diff
@@ -391,7 +391,9 @@ cdef class DCDFile:
         if frame == 0:
             offset = self._header_size
         else:
-            offset = self._header_size + self._firstframesize + self._framesize * (frame - 1)
+            offset = self._header_size
+            offset += self._firstframesize
```
orbeckst, line 395: Does this ensure that the overflow cannot happen? 

Btw, frames was declared as int in the methods signature. Should that be changed, too, or is that a Python int with infinite size?

testsuite/MDAnalysisTests/coordinates/test_dcd.py
```diff
@@ -436,3 +437,38 @@ def test_pathlib():
...
+def large_dcdfile(tmpdir):
+    # creates a >2Gb DCD file
+    fsize = 3.8  # mb
+    nreps_reqs = int(2100 // fsize)  # times to duplicate traj to hit 2.1Gb
+
+    newf = str(tmpdir / "jabba.dcd")
```
orbeckst, line 448: Name approved!

testsuite/MDAnalysisTests/coordinates/test_dcd.py
```diff
@@ -436,3 +437,38 @@ def test_pathlib():
     # we really only care that pathlib
     # object handling worked
     assert u.atoms.n_atoms == 3341
+
+
+@pytest.fixture
+def large_dcdfile(tmpdir):
+    # creates a >2Gb DCD file
+    fsize = 3.8  # mb
```
orbeckst, line 445: To be super-flexible, get the size from DCD itself. Totally optional

testsuite/MDAnalysisTests/coordinates/test_dcd.py
```diff
@@ -436,3 +437,38 @@ def test_pathlib():
...
+    with mda.Writer(newf, n_atoms=len(u.atoms)) as w:
+        for _ in range(nreps_reqs):
+            for ts in u.trajectory:
+                w.write(u.atoms)
+
+    yield newf, nreps_reqs
+
+
+@pytest.mark.skipif(
+    not os.environ.get("LARGEDCD", False), reason="Skipping large file test"
```
orbeckst, line 461: If the env bar is supposed to skip the test then better call it SKIPLATGEFILETESTS or something like that. In any case, update CI so that it runs somewhere.

orbeckst: Env bar = env var… sorry typing from mobile

orbeckst (changes requested):
Minor comments

testsuite/MDAnalysisTests/coordinates/test_dcd.py
orbeckst: We should run the test in at least one runner every time. 

And as I said in the original issue, eventually every reader should be tested with a large trajectory so that we have a better chance catching these kind of issues.

orbeckst:
@IAlibay if you don't want to be in charge please assign someone else, but given that this is related to releases etc I thought you'd be the best person.

IAlibay:
> @IAlibay if you don't want to be in charge please assign someone else, but given that this is related to releases etc I thought you'd be the best person.
> 

no worries, I'm happy to be in charge of merging, will make sure I don't forget to release

IAlibay:
So one thing to be aware of here - there are only 14 GB disk space available on GitHub runners. We'll need to make sure we're 100% sure with clearing up space, especially when dealing with pytest-xdist.

package/MDAnalysis/lib/formats/libdcd.pyx
richardjgowers: I've repro'd the exact bug (with the contentious test) and this fixes it.  I've not looked at the raw c and followed all the types.. but by eye promoting some variables to the correct datatype seemed to jiggle it into place

richardjgowers:
I'm not sure it's a good idea to run a large file test for every format for every run.  They're slow to create for one (2Gb of I/O) and it's probably not necessary.

I'm (we're) not going to have time to solve the entire issue of testing large files I/O in this PR, but I might suggest that we take this patch and fix a popular format in a bugfix release

IAlibay:
> I'm not sure it's a good idea to run a large file test for every format for every run.  They're slow to create for one (2Gb of I/O) and it's probably not necessary.
> 
> I'm (we're) not going to have time to solve the entire issue of testing large files I/O in this PR, but I might suggest that we take this patch and fix a popular format in a bugfix release

fair, do you want to just raise an issue with the current state of things?

orbeckst:
I don’t suggest that this PR should solve the testing big files for every format. 

But I think the PR should make some changes to the CI files that ensure that this test is run — either in at least one runner or at an absolute minimum in the cron. @IAlibay might have a better idea of when we should run it. But I’d want to avoid seeing such a bad regression again.

IAlibay:
May I counter @orbeckst and ask that we don't deal with CI here? This is a good cherry pickable PR. Add CI and it's going to be a pain (changes a decent chunk between releases).

IAlibay:
I'm happy to take on the responsibility of fixing up a CI entry for this if @richardjgowers would prefer not opening a second PR.

IAlibay:
Also note that I approve but leave red so that I can just fixup stuff here directly so we don't need a second pre-2.4.3 PR.

hmacdope:
> I'm not sure it's a good idea to run a large file test for every format for every run. They're slow to create for one (2Gb of I/O) and it's probably not necessary.
> 
> I'm (we're) not going to have time to solve the entire issue of testing large files I/O in this PR, but I might suggest that we take this patch and fix a popular format in a bugfix release

I agree, I more was thinking that this is a potential class of bug we should investigate, especially with the XDR reader that I cythonised in #3892. We (I) can raise an issue and we can work from there?

package/MDAnalysis/lib/formats/libdcd.pyx
hmacdope: @orbeckst the `def(blah: int )` syntax in cython allows it to switch between `int` or `PyInt` depending on how much it knows about types. I think @richardjgowers approach of changing the size of the directly declared C types is the correct one.

hmacdope (approved):
Thanks @richardjgowers!

orbeckst (approved):
Given that @IAlibay prefers the PR in this form for easier handling and given that he also committed to getting the test to run on CI somehow, I have no further objections.

orbeckst:
@IAlibay can we merge this into develop and you cherry-pick once you are ready for a hotfix 2.4.2 #4061 ?

I'd be keen to close #4048 and #4039 .

IAlibay:
> @IAlibay can we merge this into develop and you cherry-pick once you are ready for a hotfix 2.4.2 #4061 ?
> 
> I'd be keen to close #4048 and #4039 . 

I still need to make a couple of changes, I'll try to deal with this tomorrow sorry about the delay.

orbeckst:
Progress?

IAlibay:
> Progress?

Sorry I've been swamped lately and I mostly need a good empty half day to do the release, depending on how metrics generation & fixing darker lint goes I'll try to do so today or tomorrow.

orbeckst:
I understand that the release requires a solid chunk of time. My (poorly worded) question was more along the lines of what needs to be done to be able to merge the PR into develop — when you said

>> I still need to make a couple of changes

Once it's merged then we can offer at least a working development version and you can cherry-pick from develop when you can fit it in. At least that's how I understood your comment for the process.

IAlibay:
> Once it's merged then we can offer at least a working development version and you can cherry-pick from develop when you can fit it in. At least that's how I understood your comment for the process.

Yeah sorry, there's a significant element of "I don't remember fully what I need fixed to cherry pick easily" (I have the medium term memory of a goldfish lately...), so I was trying to get a bit of time to review what I needed before having to make a mess out of this.

I've booked off the evening for this, so let's try to get this done now.

PR #4051, opened 2023-03-02
Updates testsuite/MDAnalysisTests/analysis/test_nucleicacids.py

ooprathamm:
Adds -
1.  Check wc_dist_shape
2.  Check if pair_distances is in wc_dist_results_keys
3.  Tests if ValueError is thrown if the residues given are not amino acids
4.  Tests if ValueError is thrown If the selections given are not the same length

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?


ooprathamm:
@orbeckst Please have a look at this by sparing some from your precious time :)
Let me know if you any questions.

testsuite/MDAnalysisTests/analysis/test_nucleicacids.py
```diff
@@ -44,11 +44,37 @@ def wc_rna(u):
...
+    assert "pair_distances" in wc_rna.results.keys()
+
+
 def test_wc_dist(wc_rna):
     assert_allclose(wc_rna.results.pair_distances[0, 0], 4.3874702, atol=1e-3)
     assert_allclose(wc_rna.results.pair_distances[0, 1], 4.1716404, atol=1e-3)
 
 
+def test_wc_dist_invalid_residue_types(u):
+    with pytest.raises(ValueError, match=r"are not valid nucleic acids"):
```
richardjgowers: When writing tests to catch exceptions, you as few lines as possible inside the "with" block, ideally 1. This makes it more precise in what it catches and clearer what is getting tested

testsuite/MDAnalysisTests/analysis/test_nucleicacids.py
```diff
@@ -44,11 +44,37 @@ def wc_rna(u):
     return WC
 
 
+def test_wc_dist_shape(wc_rna):
+    assert wc_rna.results.pair_distances.shape == (1, 2)
+
+
+def test_wc_dist_results_keys(wc_rna):
+    assert "pair_distances" in wc_rna.results.keys()
```
richardjgowers: Don't need the .keys for a dict "in" check

richardjgowers (changes requested):
This looks good, couple fixes and we're good to go.

testsuite/MDAnalysisTests/analysis/test_nucleicacids.py
RMeli:
```suggestion
    with pytest.raises(ValueError, match="are not valid nucleic acids"):
```

ooprathamm:
As guided to have a single line in 'with' block.
test_wc_dist_invalid_residue_types throws flake error "local variable is assigned to but never used."

solution to this -
1. Add inline comment to ignore this error (done in the latest commit")
2. Add unnecessary .run()

@richardjgowers please guide which is a better solution.

ooprathamm:
@richardjgowers Pardon couldn't see the simple approach.

ooprathamm:
@richardjgowers Could you please look at the latest changes I made by sparing some from your important time. :)

richardjgowers (approved):

PR #4056, opened 2023-03-06
Switching to importlib resources

### LINKED ISSUE #4055: Move away from pkg_resources to importlib

We use `pkg_resources` to get filenames from the test suite. This may be going away in the future, as use of pkg_resources is now deprecated: https://github.com/pypa/setuptools/pull/3843

Setuptools encourages moving towards `importlib`.

IAlibay:
Fixes #4055 

Still some work to be done here (mostly getting rid of `__all__` and transferring the information across), but I thought I'd get this in now and let folks comment early since this one is likely to be contentious.

Notes:
- Because the old pkg_resources approach we took generated strings, we can't just use the Path objects directly, hence all the `str` calls (there's probably a nicer way to do this, I spent all of ~ 10 mins on this).

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?


mattwthompson:
Following out of self-interest as I need to make parallel changes in a few places

> hence all the `str` calls

There's also `Path.as_posix()` which I use because it feels like it might be safer, but I can't find any evidence of that. It's been synonymous with `str(Path)` in every case I've run into on Unix-like systems ... not sure how much Windows support complicates things as I don't make an effort to do that.

https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.as_posix

IAlibay:
So this code won't work with py3.8. There are ways to make it backwards compatible with py3.8, but honestly given that NEP29 means we ditch py3.8 on the 14th of April, I would suggest just pushing the 2.5.0 release until then and using the modern API for dealing with resources?

richardjgowers (approved):

IAlibay:
Spurious codecov failure aside - this seems to be working fine now.

@MDAnalysis/coredevs can I get a couple of extra eyes on a final review here please? This is critical for 2.5.0.

orbeckst (approved):
Thanks to you and `sed`!

PR #4059, opened 2023-03-09
Issue  #2395 AtomGroup.guess_bonds

### LINKED ISSUE #2395: AtomGroup.guess_bonds misleading documentation, could pass in fudge_factor
**Expected behavior**

https://github.com/MDAnalysis/mdanalysis/blob/971248a1cf099f26db899b899a8e10d4e96ac3cd/package/MDAnalysis/core/groups.py#L2918-L2926

AtomGroup.guess_bonds guesses bonds, only.

**Actual behavior**

It guesses bonds, angles, and dihedrals.

If the method is going to guess angles, it could guess improper dihedrals too from the angles.

https://github.com/MDAnalysis/mdanalysis/blob/971248a1cf099f26db899b899a8e10d4e96ac3cd/package/MDAnalysis/core/groups.py#L2937-L2961

- Which version are you using? current develop branch

**Also**

It would be useful to be able to pass in `fudge_factor` and `lower_bound` kwargs to this method (to pass onto `guessers.guess_bonds`)

AHMED-salah00:
Fixes #2395 

 - I have added fudge_factor and lower_bound to AtomGorup.guess_bonds() method. Also added test_universe_guess_bonds_arguments() test function https://github.com/MDAnalysis/mdanalysis/issues/2395#issuecomment-1438669606

PR Checklist
------------
 - [x] Docs?
 - [x] CHANGELOG updated?


orbeckst:
@ianmkenney could you please look after this PR? Thanks!

package/MDAnalysis/core/groups.py
```diff
@@ -3270,20 +3270,29 @@ def split(self, level):
         return [self[levelindices == index] for index in
                 unique_int_1d(levelindices)]
 
-    def guess_bonds(self, vdwradii=None):
-        """Guess bonds that exist within this :class:`AtomGroup` and add them to
-        the underlying :attr:`~AtomGroup.universe`.
+    def guess_bonds(self, vdwradii=None, fudge_factor=0.55, lower_bound=0.1, **kwargs):
+        """Guess bonds, angles and dihedrals between the atoms in this :class:`AtomGroup` and
+        add them to the underlying :attr:`~AtomGroup.universe`.
```
ianmkenney: Since the rest of the file tries to keep lines below the 79 column limit, might as well be consistent. Also, add a comma after angles.

```suggestion
    def guess_bonds(self, vdwradii=None, fudge_factor=0.55, lower_bound=0.1,
                                   **kwargs):
        """Guess bonds, angles, and dihedrals between the atoms in this
        :class:`AtomGroup` and add them to the underlying
        :attr:`~AtomGroup.universe`.
```

package/MDAnalysis/core/groups.py
```diff
@@ -3270,20 +3270,29 @@ def split(self, level):
...
             Dict relating atom types: vdw radii
 
+        fudge_factor : float, optional
+            The factor by which atoms must overlap eachother to be considered a
+            bond.  Larger values will increase the number of bonds found. [0.55]
+        lower_bound : float, optional
+            The minimum bond length. All bonds found shorter than this length will
+            be ignored. This is useful for parsing PDB with altloc records where
+            atoms with altloc A and B maybe very close together and there should be
+            no chemical bond between them. [0.1]
```
ianmkenney: Once again, should just try and be consistent with the column limits the best we can. "eachother" -> "each other", "maybe" -> "may be"

```suggestion
        fudge_factor : float, optional
            The factor by which atoms must overlap each other to be considered
            a bond.  Larger values will increase the number of bonds found. [0.55]
        lower_bound : float, optional
            The minimum bond length. All bonds found shorter than this length
            will be ignored. This is useful for parsing PDB with altloc records
            where atoms with altloc A and B may be very close together and
            there should be no chemical bond between them. [0.1]
```

ianmkenney (changes requested):
Apart from minor formatting issues, this looks good! @AHMED-salah00 can you also update the CHANGELOG and add yourself the AUTHORS file if you haven't contributed before?

AHMED-salah00:
@ianmkenney thanks for your review, I will!

AHMED-salah00:
I am not sure what is causing these failing checks.

package/MDAnalysis/core/groups.py
```diff
@@ -3270,20 +3270,31 @@ def split(self, level):
...
+        """Guess bonds, angles, and dihedrals between the atoms in this
+        :class:`AtomGroup` and add them to the underlying
+        :attr:`~AtomGroup.universe`.
 
         Parameters
         ----------
         vdwradii : dict, optional
             Dict relating atom types: vdw radii
 
+        fudge_factor : float, optional
```
IAlibay, line 3289: Could you add some tests that check that these new parameters are being passed properly through?

package/MDAnalysis/core/groups.py
```diff
@@ -3270,20 +3270,31 @@ def split(self, level):
...
 
         See Also
         --------
         :func:`MDAnalysis.topology.guessers.guess_bonds`
-
+        :func:`MDAnalysis.topology.guessers.guess_angles`
+        :func:`MDAnalysis.topology.guessers.guess_dihedrals`
 
         .. versionadded:: 0.10.0
         .. versionchanged:: 0.20.2
```
IAlibay, line 3306: Could you add a new versionchanged entry pleae?

IAlibay (changes requested):
Thanks, this is a good start. I would like to see maybe one additional test to ensure the parameters are passing through fine, if possible.

package/MDAnalysis/core/groups.py
AHMED-salah00: should I make a `TestGuessBonds` class and test `guess_bonds` for different values for fudge_factor, lower_bond etc?

package/MDAnalysis/core/groups.py
AHMED-salah00: Sure, I will. I am sorry i forgot that

package/MDAnalysis/core/groups.py
AHMED-salah00: Oh, found TestGuessBonds class I will add these tests to it 

    def test_fudge_factor(self):
        u = mda.Universe(two_water_gro, guess_bonds=True, fudge_factor=0.5)
        self._check_universe(u)
        assert u.kwargs['guess_bonds']
        assert_equal(0.5, u.kwargs['fudge_factor'])

    def test_lower_bound(self):
        u = mda.Universe(two_water_gro, guess_bonds=True, lower_bound=0.8)
        self._check_universe(u)
        assert u.kwargs['guess_bonds']
        assert_equal(0.8, u.kwargs['lower_bound'])

So, I should add these attributes to Universe class too...

package/MDAnalysis/core/groups.py
AHMED-salah00: Done.

package/MDAnalysis/core/universe.py
```diff
@@ -316,7 +321,8 @@ class Universe(object):
     """
     def __init__(self, topology=None, *coordinates, all_coordinates=False,
                  format=None, topology_format=None, transformations=None,
-                 guess_bonds=False, vdwradii=None, in_memory=False,
+                 guess_bonds=False, vdwradii=None,fudge_factor=0.55,
+                 lower_bound=0.1, in_memory=False,
```
ianmkenney, line 325:
```suggestion
                 guess_bonds=False, vdwradii=None, fudge_factor=0.55,
                 lower_bound=0.1, in_memory=False,
```
Missing a space here

ianmkenney:
```suggestion
                 guess_bonds=False, vdwradii=None, fudge_factor=0.55,
                 lower_bound=0.1, in_memory=False,
```
Missing a space here

package/MDAnalysis/core/groups.py
```diff
@@ -3302,8 +3314,13 @@ def get_TopAttr(u, name, cls):
...
         # indices of bonds
-        b = guess_bonds(self.atoms, self.atoms.positions,
-                        vdwradii=vdwradii, box=self.dimensions)
+        b = guess_bonds(self.atoms,
+                        self.atoms.positions,
+                        vdwradii=vdwradii,
+                        box=self.dimensions,
+                        fudge_factor=fudge_factor,
+                        lower_bound=lower_bound,
+                        **kwargs)
```
ianmkenney: I think the question of if parameters are passed correctly mainly refers to these lines? Correct me if I'm wrong @IAlibay.

package/MDAnalysis/core/universe.py
AHMED-salah00: How did one space make another two tests fail :smiling_face_with_tear:

package/MDAnalysis/core/groups.py
ianmkenney: https://github.com/MDAnalysis/mdanalysis/blob/develop/testsuite/MDAnalysisTests/core/test_universe.py#L461 might be a good place to put tests for this. Ideally, we'd like to run with defaults, then check those results (for instance, the number of bonds determined) against different inputs.

AHMED-salah00: >  we'd like to run with defaults, then check those results (for instance, the number of bonds determined) against different inputs.

I thought about that too at first but I didn't know which way is the best suit for this case.

package/MDAnalysis/core/universe.py
```diff
@@ -312,11 +317,14 @@ class Universe(object):
     .. versionchanged:: 2.0.0
         Universe now can be (un)pickled.
         ``topology`` and ``trajectory`` are reserved upon unpickle.
-
+    .. versionchanged:: 2.4.0
```
AHMED-salah00: will fix to 2.5.0 in the next commit.

package/MDAnalysis/core/groups.py
```diff
@@ -3270,24 +3270,38 @@ def split(self, level):
...
         --------
         :func:`MDAnalysis.topology.guessers.guess_bonds`
-
+        :func:`MDAnalysis.topology.guessers.guess_angles`
+        :func:`MDAnalysis.topology.guessers.guess_dihedrals`
 
         .. versionadded:: 0.10.0
         .. versionchanged:: 0.20.2
            Now applies periodic boundary conditions when guessing bonds.
+        .. versionchanged:: 2.4.2
```
IAlibay: Sorry should have mentioned earlier, this should be 2.5.0

AHMED-salah00: Oh, sure! thanks.

package/MDAnalysis/core/groups.py
AHMED-salah00: Can I go now for the modifications that Ian mentioned above @IAlibay ? I found that `TestGuessBonds` already exists and have multiple tests for the parameters so I added additional params for the new ones to check if they passed correctly.

package/MDAnalysis/core/universe.py
```diff
@@ -312,11 +317,14 @@ class Universe(object):
     .. versionchanged:: 2.0.0
         Universe now can be (un)pickled.
         ``topology`` and ``trajectory`` are reserved upon unpickle.
-
+    .. versionchanged:: 2.5.0
+        Added fudge_factor and lower_bound parameters for use with
+        **guess_bonds**.
```
ianmkenney: Single asterisk to be consistent with rest of docs.
 
```suggestion
        *guess_bonds*.
```

testsuite/MDAnalysisTests/core/test_universe.py
```diff
@@ -460,11 +474,17 @@ def _check_atomgroup(self, ag, u):
...
+            u = mda.Universe(two_water_gro)
+            ag = u.atoms[:3]
+            ag.guess_bonds(fudge_factor=ff, lower_bound=lb)
+            assert_equal(len(ag.bonds), nbonds)
+
+        # We run here multiple tests with different values for
+        # fudge_factor and lower_bound
+        multi_parameters_atomgroup(0.55, 0.1, 2)
+        multi_parameters_atomgroup(0.9, 1.6, 1)
+        multi_parameters_atomgroup(0.5, 0.2, 2)
```
ianmkenney: You should be able to do what `multi_parameters_atomgroup` does with [pytest fixtures](https://docs.pytest.org/en/6.2.x/parametrize.html). It will create separate tests that can fail individually.

ianmkenney (changes requested):
Simple formatting change and use fixtures for testing different parameter sets.

package/MDAnalysis/core/groups.py
```diff
@@ -3276,24 +3276,38 @@ def split(self, level):
...
+            will be ignored. This is useful for parsing PDB with altloc records
+            where atoms with altloc A and B may be very close together and
+            there should be no chemical bond between them. [0.1]
 
         See Also
         --------
         :func:`MDAnalysis.topology.guessers.guess_bonds`
-
+        :func:`MDAnalysis.topology.guessers.guess_angles`
+        :func:`MDAnalysis.topology.guessers.guess_dihedrals`
```
IAlibay, line 3303:
```suggestion


```
There's a need for two empty lines here otherwise the docs render weirdly, see: https://mdanalysis--4059.org.readthedocs.build/en/4059/documentation_pages/core/groups.html

package/MDAnalysis/core/groups.py
```diff
@@ -3276,24 +3276,38 @@ def split(self, level):
...
 
         .. versionadded:: 0.10.0
         .. versionchanged:: 0.20.2
            Now applies periodic boundary conditions when guessing bonds.
+        .. versionchanged:: 2.5.0
+           corrected misleading docs ,and now you can pass fudge factor
+           and lower bound.
```
IAlibay:
```suggestion
           Corrected misleading docs, and now allows passing of `fudge_factor`
           and `lower_bound` arguments.
```
Slight improvement in wording.

package/MDAnalysis/core/groups.py
```diff
@@ -3308,9 +3322,16 @@ def get_TopAttr(u, name, cls):
...
-                        vdwradii=vdwradii, box=self.dimensions)
-        bondattr = get_TopAttr(self.universe, 'bonds', Bonds)
+        b = guess_bonds(
+            self.atoms,
+            self.atoms.positions,
+            vdwradii=vdwradii,
+            box=self.dimensions,
+            fudge_factor=fudge_factor,
+            lower_bound=lower_bound,
+            **kwargs
```
IAlibay: What additional kwargs are you anticipating passing here?

testsuite/MDAnalysisTests/core/test_universe.py
```diff
@@ -460,11 +474,17 @@ def _check_atomgroup(self, ag, u):
...
-        u = mda.Universe(two_water_gro)
-
-        ag = u.atoms[:3]
-        ag.guess_bonds()
-        self._check_atomgroup(ag, u)
+        def multi_parameters_atomgroup(ff=0.55, lb=0.1, nbonds=2):
+            u = mda.Universe(two_water_gro)
+            ag = u.atoms[:3]
+            ag.guess_bonds(fudge_factor=ff, lower_bound=lb)
+            assert_equal(len(ag.bonds), nbonds)
```
IAlibay: in addition to @ianmkenney's comment about using some kind of pytest fixture to do this (probably pytest.parametrize), please don't use `assert_equal` for comparing single ints. Just use a simple `assert`, `assert_equal` for single items is about 10x slower.

testsuite/MDAnalysisTests/core/test_universe.py
```diff
@@ -441,6 +441,20 @@ def test_universe_guess_bonds_off(self):
...
+        self._check_universe(u)
+        assert u.kwargs["guess_bonds"]
+        assert_equal(0.5, u.kwargs["fudge_factor"])
+
+    def test_lower_bound(self):
+        """Test if lower_bound parameter is being passed correctly"""
+        u = mda.Universe(two_water_gro, guess_bonds=True, lower_bound=0.8)
+        self._check_universe(u)
+        assert u.kwargs["guess_bonds"]
+        assert_equal(0.8, u.kwargs["lower_bound"])
```
IAlibay: There's really no point of separating this out into two separate tests, I would just do the one test that checks that kwargs is populated fine.

Also here, due to `assert_equal`'s very large cost on single values, I would encourage you to use a standard assert instead, you can use pytest.approx for comparing floats, see: https://github.com/MDAnalysis/mdanalysis/blob/a05a5245042361fd4fa27f03993f34cee2cda901/testsuite/MDAnalysisTests/analysis/test_align.py#L569

package/MDAnalysis/core/groups.py
IAlibay: Sorry about the delay, yes @ianmkenney is correct. What you want to do is check the effect of passing in the keywords instead of solely looking that they got assigned to the right things. That way if the guesser.guess_bonds changes, then you can catch that and know that something using it broke.

IAlibay (changes requested):
Just a couple of extra review comments on top of the things @ianmkenney raised.

testsuite/MDAnalysisTests/core/test_universe.py
```diff
@@ -458,13 +469,20 @@ def _check_atomgroup(self, ag, u):
...
+            (0.5, 0.2, 2), (0.1, 0.1, 0)
+        ]
+    )
+    def test_atomgroup_guess_bonds(self, ff, lb, nbonds):
         """Test an atomgroup doing guess bonds"""
         u = mda.Universe(two_water_gro)
 
         ag = u.atoms[:3]
-        ag.guess_bonds()
-        self._check_atomgroup(ag, u)
```
IAlibay, line 467: Why is this being removed?

AHMED-salah00: As it raises an assertion error each time a different bond number than the default value(2 bonds) was guessed.Also the number of angles, and dihedrals.

AHMED-salah00: we can solve this by passing these parameters to _check_atomgroup, but would it be useful then?

ianmkenney: Let's leave the check_universe in, but only call it if the default parameters are used.

AHMED-salah00: Sure @ianmkenney ! It's actually a good idea.

AHMED-salah00:
I've made some progress @ianmkenney,  Would you take a look at it and let me know if anything is missing?

testsuite/MDAnalysisTests/core/test_universe.py
```diff
@@ -458,13 +469,22 @@ def _check_atomgroup(self, ag, u):
...
+        ag.guess_bonds(fudge_factor=ff, lower_bound=lb)
+        if ff == 0.55 and lb == 0.1:
```
ianmkenney, line 488: Could you add a comment why this `if` statement was added? Future developers won't necessarily understand the context.

ianmkenney (changes requested):
I think this looks pretty good, just need to add a comment for the `if` statement. @IAlibay, any final thoughts?

testsuite/MDAnalysisTests/core/test_universe.py
```diff
@@ -398,7 +398,7 @@ class TestGuessBonds(object):
      - fail properly if not
      - work again if vdwradii are passed.
     """
-    @pytest.fixture()
+    @pytest.fixture(scope='module')
```
IAlibay, line 401: why is scope changing?

AHMED-salah00: When we run the tests we don't need to call `vdw` each time we need it. We can only call it once when first called as it's values does not change across the tests. Although it does a simple task, I guess it's better to be done once, rather than repeatedly for each test function.

orbeckst: Makes sense to me to make "read only" fixtures to be module-scoped  — even though the speed-up is almost certainly negligible for something simple like a dict.

ianmkenney (changes requested):
LGTM. Can you resolve the merge conflicts with `develop`?

IAlibay (approved):

ianmkenney (approved):

PR #4062, opened 2023-03-10
Fix doctest for MDAnalysis.core.groups (#3925)

xhgchen:
* Towards #3925
* Add imports to asunique() examples
* Fix code logic for ag*.ix in unique() and asunique(): change expected array to specify dtype=int64
* Fixes 102 failing tests (344 failures -> 242 failures)

Thank you @hmacdope for helping me get started on this PR! This is just a simple fix to help me get more familiar with the docs, and I will continue tackling coding issues once I have made some more headway on my proposal.

### DIFF
--- a/package/MDAnalysis/core/groups.py
+++ b/package/MDAnalysis/core/groups.py
@@ -2672,6 +2672,9 @@ class AtomGroup(GroupBase):
         Examples
         --------
 
+           >>> import MDAnalysis as mda
+           >>> from MDAnalysis.tests.datafiles import PSF, DCD
+           >>> u = mda.Universe(PSF, DCD)
            >>> ag = u.atoms[[2, 1, 2, 2, 1, 0]]
            >>> ag
            <AtomGroup with 6 atoms>
--- a/package/MDAnalysis/core/groups.py
+++ b/package/MDAnalysis/core/groups.py
@@ -2676,12 +2676,12 @@ class AtomGroup(GroupBase):
            >>> ag
            <AtomGroup with 6 atoms>
            >>> ag.ix
-           array([2, 1, 2, 2, 1, 0])
+           array([2, 1, 2, 2, 1, 0], dtype=int64)
            >>> ag2 = ag.unique
            >>> ag2
            <AtomGroup with 3 atoms>
            >>> ag2.ix
-           array([0, 1, 2])
+           array([0, 1, 2], dtype=int64)
            >>> ag2.unique is ag2
            False
 
@@ -2725,19 +2725,22 @@ class AtomGroup(GroupBase):
         Examples
         --------
 
+           >>> import MDAnalysis as mda
+           >>> from MDAnalysis.tests.datafiles import PSF, DCD
+           >>> u = mda.Universe(PSF, DCD)
            >>> ag = u.atoms[[2, 1, 0]]
            >>> ag2 = ag.asunique(sorted=False)
            >>> ag2 is ag
            True
            >>> ag2.ix
-           array([2, 1, 0])
+           array([2, 1, 0], dtype=int64)
            >>> ag3 = ag.asunique(sorted=True)
            >>> ag3 is ag
            False
            >>> ag3.ix
-           array([0, 1, 2])
+           array([0, 1, 2], dtype=int64)
            >>> u.atoms[[2, 1, 1, 0, 1]].asunique(sorted=False).ix
-           array([2, 1, 0])
+           array([2, 1, 0], dtype=int64)
 
 
         .. versionadded:: 2.0.0


orbeckst:
@orionarcher could you look after this issue? (You can always ping someone else such as @hmacdope ;-) for a second pair of 👀 )

xhgchen:
It seems like some of the checks are failing because of an error uploading to Codecov. Is there anything I can do to help?

orbeckst:
no, there's nothing you can do, codecov is glitchy and fails like this (too often... :-( ). We can rerun the CI or it will rerun on your next commit.

orbeckst:
I just restarted all failing jobs.

package/MDAnalysis/core/groups.py
```diff
@@ -2725,19 +2725,22 @@ def asunique(self, sorted=False):
         Examples
         --------
 
+           >>> import MDAnalysis as mda
+           >>> from MDAnalysis.tests.datafiles import PSF, DCD
```
orionarcher, line 2732: Since you are importing MDAnalysis as mda, you should use `mda` rather than `MDAnalysis` going forward.

package/MDAnalysis/core/groups.py
```diff
@@ -2676,12 +2676,12 @@ def unique(self):
            >>> ag
```
orionarcher, line 2679: Maybe add imports on this docstring too?

package/MDAnalysis/core/groups.py
```diff
@@ -2676,12 +2676,12 @@ def unique(self):
...
            <AtomGroup with 6 atoms>
            >>> ag.ix
-           array([2, 1, 2, 2, 1, 0])
+           array([2, 1, 2, 2, 1, 0], dtype=int64)
```
orionarcher, line 2682: When I return a numpy array in the terminal, the dtype usually doesn't appear, can you explain why you added the dtype?

orionarcher (changes requested):

package/MDAnalysis/core/groups.py
xhgchen: Thanks for the prompt review and feedback @orionarcher! I just tried using `mda` in my terminal and this was the result:
```
$ python
Python 3.11.0 | packaged by conda-forge | (main, Oct 25 2022, 06:12:32) [MSC v.1929 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import MDAnalysis as mda
C:\Users\Xu Hong Chen\Projects\mdanalysis\package\MDAnalysis\topology\TPRParser.py:161: DeprecationWarning: 'xdrlib' is depreca
ted and slated for removal in Python 3.13
  import xdrlib
>>> from mda.tests.datafiles import PSF, DCD
ModuleNotFoundError: No module named 'mda'
```
How might you go about doing this? It does work when I use `MDAnalysis` instead.

package/MDAnalysis/core/groups.py
xhgchen: When I run `make doctest` (from #3925), no failing tests related to imports appear for that docstring. Would it still be better to add imports in this case? The page itself is: https://docs.mdanalysis.org/stable/documentation_pages/core/groups.html#MDAnalysis.core.groups.AtomGroup.unique

Edit: I think you're right. Will add those imports in the next commit, after I hear your thoughts on everything.

package/MDAnalysis/core/groups.py
xhgchen: After adding the imports, the file output.txt from `make doctest` showed:
```
File "..\..\..\MDAnalysis\core\groups.py", line ?, in default
Failed example:
    ag2.ix
Expected:
    array([2, 1, 0])
Got:
    array([2, 1, 0], dtype=int64)
```

When I ran the commands in my terminal (Git Bash, Windows 10, Python 3.11.0, miniconda), I got a similar result:
```
>>> ag = u.atoms[[2, 1, 0]]
>>> ag2 = ag.asunique(sorted=False)
>>> ag2 is ag
True
>>> ag2.ix
array([2, 1, 0], dtype=int64)
```

So I thought it made sense to add the dtype to resolve those failing tests. Do you think it might be because I am on Windows? I have been meaning to switch to Linux but used Windows for this PR because I had the dev environment set up already.

Edit: I just found a similar issue here: https://stackoverflow.com/questions/46285518/why-is-the-dtype-shown-even-if-its-the-native-one-when-using-floor-division-w and will continue reading about this.

Here is another thread: https://stackoverflow.com/questions/10288757/remove-dtype-at-the-end-of-numpy-array

Looking forward to hearing your input on this :)

package/MDAnalysis/core/groups.py
orionarcher: The original PR states:

> All the code in our docstring examples should be self-contained and run on its own. They should:
> import the correct modules
> demonstrate the desired behaviour
> run correctly

So I do believe we want the imports on every doc example

package/MDAnalysis/core/groups.py
orionarcher: Ahh, nvm, I may be wrong here.

package/MDAnalysis/core/groups.py
orionarcher: odd, not sure why my terminal is showing something different. If you need the dtype to pass the test then feel free to continue including it!

package/MDAnalysis/core/groups.py
xhgchen: Right, thank you for confirming! I added the imports and they actually (unexpectedly?) resolved some more (~19) failing tests.

package/MDAnalysis/core/groups.py
xhgchen: Do you think it would be worthwhile to ask a third person to try these commands in their terminal to check whether it makes sense to add the dtype or not?

package/MDAnalysis/core/groups.py
xhgchen: All good; appreciate you taking the time to address everything!

package/MDAnalysis/core/groups.py
orionarcher: Haha that's great, I'll admit I'm a little unclear on how the doctesting framework works.

package/MDAnalysis/core/groups.py
orionarcher: @hmacdope are y'all trying to standardize this across the docs in one direction or the other?

hmacdope: Whatever makes the doctest work. :)

xhgchen: Sounds good! @orionarcher is there anything else I should add or edit in this PR? The tip about imports was quite helpful, and I will keep it in mind if I work on this issue again in the future.

orionarcher (approved):
Code here all LGTM! Maybe @hmacdope can contextualize why some tests are failing? Is this good to merge?

IAlibay (approved):
lgtm, azure CI failures are transient and related to some 32bit troubles with h5py, probably will go away on re-run.

xhgchen:
Thanks for looking this over, everyone :)

IAlibay:
Happy to go ahead and merge this, the issue with 32bit windows isn't related to this PR.

PR #4063, opened 2023-03-12
modernize testing code #3743

ShubhamZoro:
Fixes #

use numpy.testing.assert_allclose() instead of assert_almost_equal() or assert_array_almost_equal() and removed these deprecated method. There is still some library like Xdrlib which is deprecated and some library or method(like warning in test_contacts.py) which has been imported but never used. But I change the assert_almost_equal and array_almost_equal to all_close

### DIFF
--- a/package/MDAnalysis/analysis/dielectric.py
+++ b/package/MDAnalysis/analysis/dielectric.py
@@ -129,7 +129,7 @@ class DielectricConstant(AnalysisBase):
             raise NoDataError("No charges defined given atomgroup.")
 
         if not np.allclose(self.atomgroup.total_charge(compound='fragments'),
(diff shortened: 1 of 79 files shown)


IAlibay:
@ShubhamZoro This issue was never meant to be tackled across all the test files in one go. Given the large amount of test failures you're seeing, may I suggest scaling this PR back to just one or two test files? This will not only make it easier to review & faster to merge, but will also be an easier code problem to solve.

IAlibay:
Also I would heavily suggest not pushing code from "develop" in your own fork but rather a feature branch. Using "develop" often leads to merge issues after a while because we squash merge our PRs by default.

PR #4066, opened 2023-03-12
Fix the case where distance_type = 'heavy' in water bridge analysis

### LINKED ISSUE #4040: WaterBridgeAnalysis can't seem to do (Acceptor <-> WAT <-> WAT <-> Acceptor) water bridges?
## Expected behavior ##

I want to quantify a second order water bridge (two waters between a protein residue (resid 43 and name OD2) and a DNA nucleotide (resid 230 and name P). I would like a time series of the presence of these bonds.

resid 43   = resname ASP
resid 230 = resname DA 

## Troubleshooting thus far ## 
Notably, selection two is a phosphorus atom. (not a usual hydrogen bond acceptor atom) I know this 'P' isn't in any standard library so I added it with the "acceptors" keyword in addition to forcing selection2_type to be an acceptor. I tried troubleshooting by enlarging the acceptable ranges for angle and distance, and even manually specifying every donor and acceptor used in the interaction in-case any were not in the default atom mask libraries. I also played with using resid, resname or simply atom masks to see if the selection was the issue. I tried on a full trajectory and truncated trajectories where I've manually extracted the frames with water bridges present. 

I've verified visually in pymol that the water bridge is indeed present and within spec for the identities, angles and distances I used. 

## Actual behavior ##

```
MDAnalysis.analysis.WaterBridgeAnalysis: INFO     WaterBridgeAnalysis: selection = 'resname DA and name P' (update: True)
MDAnalysis.analysis.WaterBridgeAnalysis: INFO     WaterBridgeAnalysis: water selection = 'resname WAT' (update: True)
MDAnalysis.analysis.WaterBridgeAnalysis: INFO     WaterBridgeAnalysis: criterion: donor heavy atom and acceptor atom distance <= 4.000 A
MDAnalysis.analysis.WaterBridgeAnalysis: INFO     WaterBridgeAnalysis: criterion: angle D-H-A >= 120.000 degrees
MDAnalysis.analysis.WaterBridgeAnalysis: INFO     WaterBridgeAnalysis: force field CHARMM27 to guess donor and         acceptor names
MDAnalysis.analysis.base: INFO     Choosing frames to analyze
MDAnalysis.analysis.base: INFO     Starting preparation
MDAnalysis.analysis.WaterBridgeAnalysis: INFO     WaterBridgeAnalysis: initial checks passed.
MDAnalysis.analysis.WaterBridgeAnalysis: INFO     WaterBridgeAnalysis: starting
MDAnalysis.analysis.WaterBridgeAnalysis: INFO     Starting analysis (frame index start=0 stop=10, step=1)
MDAnalysis.analysis.base: INFO     Starting analysis loop over 10 trajectory frames
100%|██████████| 10/10 [00:00<00:00, 21.44it/s]
MDAnalysis.analysis.base: INFO     Finishing up
[[], [], [], [], [], [], [], [], [], []]
```

you can see on the last line no hydrogen bonds are identified.
I've opened up the 'w' dataframe and confirmed that the universe was created successfully and that the correct acceptors and donors are listed. 

``` 
import MDAnalysis as mda
import MDAnalysis.analysis.hydrogenbonds
import numpy.linalg
import numpy as np
np.set_printoptions(linewidth=100)

#%%
#Set working directory
import os
os.chdir('c:\\Users\\makay.murray\\Documents\\GitHub\\md_water_bridge\\analysis_dir')
[MDAnalysis.log](https://github.com/MDAnalysis/mdanalysis/files/10827603/MDAnalysis.log)

#Create Universe
parm = "../trajectory_data/2w35-DL-HID.prmtop"
traj = "../trajectory_data/2w35-DL-HID-rep2-500ns-10frames.mdcrd"

u = mda.Universe(parm,traj)  # always start with a Universe

w = MDAnalysis.analysis.hydrogenbonds.WaterBridgeAnalysis(u, 
                                                          'resname ASP and name OD2', 
                                                          'resname DA and name P', 
                                                          update_selection=True,
                                                          selection1_type='acceptor',
                                                          selection2_type='acceptor',
                                                          water_selection='resname WAT', 
                                                          order=2, 
                                                          update_water_selection=True,
                                                          distance=4.0, angle=120.0,
                                                          distance_type = 'heavy',
                                                          debug=True,
                                                          verbose=True,
                                                          acceptors=('P', 'OD1', 'OD2', 'O'),
                                                          donors=('H1', 'H2'),
                                                          filter_first=True
                                                          )
w.run()

print(w.timeseries)
....

```

- MDAnalaysis  Version 2.4.2
- Python 3.9.16
- OS, Windows 10 Education
- IDE Spyder 5.4.2

Attached files:
[2w35-DL-HID-rep2-500ns-10frames.zip](https://github.com/MDAnalysis/mdanalysis/files/10827332/2w35-DL-HID-rep2-500ns-10frames.zip)
[2w35-DL-HID.zip](https://github.com/MDAnalysis/mdanalysis/files/10827334/2w35-DL-HID.zip)
Water Bridge Image black
[MDAnalysis.log](https://github.com/MDAnalysis/mdanalysis/files/10827608/MDAnalysis.log)

### LINKED ISSUE #4136: WaterBridge: IndexError if distance_type = "heavy"
https://github.com/MDAnalysis/mdanalysis/blob/6c50334239c2fe14b86caea16c1fe46462b70ae0/package/MDAnalysis/analysis/hydrogenbonds/wbridge_analysis.py#L1198

Hello everyone, 

I was calculating water-bridges and encountered the following issue:
If I use _**distance_type = "heavy"**_, I get an IndexError:

File ~/miniconda3/envs/MD_simulation/lib/python3.10/site-packages/MDAnalysis/analysis/hydrogenbonds/wbridge_analysis.py:1265, in WaterBridgeAnalysis._donor2acceptor(self, donors, h_donors, acceptor)
     1237     d = tmp_donors[index]
     1238     a = tmp_acceptors[index]
     1239     result.append((h, d, a, self._expand_index(h),
     1240                    self._expand_index(a),
-> 1241                    distances[index], angles[index]))
     1242 return result

I had a quick look at the code and it seems like there is a logic error:

First, we are looking for valid distances: (which is fine)
https://github.com/MDAnalysis/mdanalysis/blob/6c50334239c2fe14b86caea16c1fe46462b70ae0/package/MDAnalysis/analysis/hydrogenbonds/wbridge_analysis.py#L1207-L1211

If we now want to check respective angles, we are creating a list of temporary donors, since a heavy donor atom can have multiple hydrogens: (which is also fine)
https://github.com/MDAnalysis/mdanalysis/blob/6c50334239c2fe14b86caea16c1fe46462b70ae0/package/MDAnalysis/analysis/hydrogenbonds/wbridge_analysis.py#L1216-L1224

Afterwards, we are considering H-Bonds as valid, if respective angle was within a certain cutoff, however the indices of _angles_ and _distances_ do not match anymore... (since we had to consider multiple hydrogens attached to the donor atoms)
https://github.com/MDAnalysis/mdanalysis/blob/6c50334239c2fe14b86caea16c1fe46462b70ae0/package/MDAnalysis/analysis/hydrogenbonds/wbridge_analysis.py#L1234-L1242

Edit: For now, I do not return respective distances, since I am not interested in them... :)

At least that is what I understood...
Best wishes,
Nico

xiki-tempula:
Fixes #4040

In water bridge analysis, in the `distance_type = 'hydrogen'` mode, I get the pair index (**hydrogen bond `hydrogen` atom** and acceptor) and the pair distance from the distance search.
Then for each hydrogen atom bonded to the donor heavy atom, I check the angle between the hydrogen bond donor heavy atom, the hydrogen atom and the acceptor and map the index to the pair distance. This works because the distance is computed based on the hydrogen atom and we map the hydrogen atom index back to the distance.

In the `distance_type = 'heavy'` mode, In the original code, I get the pair index (**hydrogen bond donor heavy atom** and acceptor) and the pair distance from the distance search. Then for each hydrogen atom bonded to the donor heavy atom, I check the angle between the hydrogen bond donor heavy atom, the hydrogen atom and the acceptor. In this case, I cannot map back to the pair distance, as the distance search is done on the heavy atom and we are mapping the index of the hydrogen atom back to the distance.

This PR fixes the issue.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


xiki-tempula:
Got complain from code formatting so took the change to black the code.

RMeli:
> Got complain from code formatting so took the change to black the code.

This is the way! =)

However, the changes are probably too many to review. This is yet another example of why IMHO we should take the plunge and format the whole codebase with `black`, which is now the de-facto standard.

IAlibay:
1. The linter CI is informative not meant to be a direction, as long as you don't trigger the flake8 warnings you're fine
2. There are plenty of cases where `black` causes code that folks that don't agree with, it's popped up a few times, so I'm not in agreement that we should just apply it to the whole code base.

IAlibay (changes requested):
Can you revert the whole file `black` please? The diff is now making it impossible for us to review the real changes you introduced. See my comment about linting being informative.

RMeli:
> There are plenty of cases where black causes code that folks that don't agree with

That's the whole point of adopting `black`. The agreement is to adopt `black`, and this will end any further discussion. =)

But here we are yet again loosing precious developer time on silly formatting arguments. (There were several GSoC PR suffering of this waste too.)

IAlibay:
> > There are plenty of cases where black causes code that folks that don't agree with
> 
> That's the whole point of adopting `black`. The agreement is to adopt `black`, and this will end any further discussion. =)
> 
> But here we are yet again loosing precious developer time on silly formatting arguments. (There were several GSoC PR suffering of this waste too.)

Adopting a single approach that significantly worsens readability (and yes, in some cases the whole thing was completely unreadable), is not the answer here either.

W.r.t. GSoC PRs issues - please bring them up on discord, I have yet to see anything on my end that seemed to be a big deal, but maybe I missed something important.

I would ask that we move this discussion out of this PR. My point in this particularly PR is that diffs are important, and darken is informational, so let's keep the diff clean so we can review this quickly.

xiki-tempula:
@IAlibay I have reverted the PR back to the state before BLACK.
I guess you could review it and if you are happy. I will black the `wbridge_analysis` and `test_wbridge` files and this PR could be merged. I'm fine with black my code. If you read enough black code, you adapt to black code. (Similar to all other tools, one could disable black for a certain part of the code as well)

If you look at the CI, https://github.com/MDAnalysis/mdanalysis/actions/runs/4398889837/jobs/7703050316
In the darker main code part, the process will quit without any flake warning.

IAlibay:
I would prefer to keep commits clean since we squash merge, for here let's just address the issue that needs addressing. Whole file `black` applications can be discussed / tackled elsewhere.

xiki-tempula:
@IAlibay The darker failed with no flake error.

IAlibay:
> @IAlibay The darker failed with no flake error.

That's safe to ignore then, I wouldn't worry about it. I think discussions from this have made it clear that it's an annoying failure to see (even when it's meant to be "ok"), so I've opened #4067 with the aim of making it not fail and just be informative in the future.

xiki-tempula:
@IAlibay If you are happy with the linter failing, I don't really mind.

package/MDAnalysis/analysis/hydrogenbonds/wbridge_analysis.py
```diff
@@ -866,8 +865,8 @@ def __init__(self, universe, selection1='protein',
             :attr:`~DEFAULT_ACCEPTORS` values.
             ["CHARMM27"]
         donors : sequence (optional)
-            Extra H donor atom types (in addition to those in
-            :attr:`~DEFAULT_DONORS`), must be a sequence.
+            Extra H donor atom types (The heavy atom not the hydrogen)
+            (in addition to those in :attr:`~DEFAULT_DONORS`), must be a sequence.
```
IAlibay: I'll be honest I have trouble understanding what this means, in part because there's two sets of text in brackets one after the other. Is there any way to make this a bit clearer?

Possibly linked to this, would it be worth adding something in the docstring examples or notes about it?

package/MDAnalysis/analysis/hydrogenbonds/wbridge_analysis.py
```diff
@@ -1210,18 +1209,21 @@ def _donor2acceptor(self, donors, h_donors, acceptor):
...
             tmp_donors = [h_donors[donors_idx[idx]] for idx in pairs[:, 0]]
             tmp_hydrogens = [donors_idx[idx] for idx in pairs[:, 0]]
             tmp_acceptors = [acceptor[idx] for idx in pairs[:, 1]]
         else:
             tmp_donors = []
             tmp_hydrogens = []
             tmp_acceptors = []
-            for idx in range(len(pairs[:, 0])):
+            tmp_distances = []
+            for idx, distance in enumerate(distances):
```
IAlibay, line 1229: I think this makes sense, so we don't get lost with this again, could you just add maybe an inline comment about what this is doing here / the importance of this case?

IAlibay (changes requested):
Thanks for this fix @xiki-tempula.

I _think_ this mostly makes sense, although I will say the complexity of `donors` is a bit high (maybe it's just because it's a Sunday evening), it would be good to maybe add an example case or maybe just expand on the docstring a little bit if possible.

package/CHANGELOG
```diff
@@ -19,6 +19,7 @@ The rules for this file:
  * 2.5.0
 
 Fixes
+  * Fix the heavy distance_type for water bridge analysis (Issue #4040, PR #4066).
```
IAlibay: I forgot, please add yourself above on the 2.5.0 author list.

xiki-tempula:
@IAlibay Thanks for the review. I have added some more explanations and added myself to the authors.

package/AUTHORS
```diff
@@ -209,6 +209,7 @@ Chronological list of authors
   - Vishal Parmar
   - Moritz Schaeffler
   - Xu Hong Chen
+  - Zhiyi Wu
```
RMeli:
```suggestion
```

@xiki-tempula, seem to be already under authors in 2017 (assuming it is you, since you have been contributing for a while).

RMeli: I think @IAlibay was about adding yourself to the list of authors in the `CHANGELOG`, for the `2.5.0` release.

xiki-tempula: @RMeli  Yes, that is me. Shall I remove it?

xiki-tempula: I see. I have changed it.

package/MDAnalysis/analysis/hydrogenbonds/wbridge_analysis.py
```diff
@@ -866,8 +865,9 @@ def __init__(self, universe, selection1='protein',
             :attr:`~DEFAULT_ACCEPTORS` values.
             ["CHARMM27"]
         donors : sequence (optional)
-            Extra H donor atom types (in addition to those in
-            :attr:`~DEFAULT_DONORS`), must be a sequence.
+            Extra H donor atom types (in addition to those in :attr:`~DEFAULT_DONORS`).
+            This shall be the name of the heavy atom that bonded to the hydrogen.
```
IAlibay:
```suggestion
            This shall be the name of the heavy atom that is bonded to the hydrogen.
```

package/MDAnalysis/analysis/hydrogenbonds/wbridge_analysis.py
```diff
@@ -866,8 +865,9 @@ def __init__(self, universe, selection1='protein',
...
+            For example, the oxygen ('O') in the hydroxyl group, must be a sequence.
```
IAlibay:
```suggestion
            For example, the oxygen ('O') in the hydroxyl group. Must be a sequence.
```
My understanding of that sentence is that the intent was for a full stop?

IAlibay (changes requested):
Just a couple of small things otherwise lgtm.

Could one of the other @MDAnalysis/coredevs also have a quick look here please? WaterBridgeAnalysis is reasonably complex, so having at least two pairs of eyes on the review would be appreciated.

IAlibay:
@MDAnalysis/coredevs - pinging again, could I get a second review here please?

IAlibay (approved):
I'm going to tentatively approve this, but can I please ask for a second review by one of the other @MDAnalysis/coredevs ? (asking for the third time)

package/MDAnalysis/analysis/hydrogenbonds/wbridge_analysis.py
```diff
@@ -769,7 +768,7 @@ class WaterBridgeAnalysis(AnalysisBase):
     #: N, O, P, and S. Any other heavy atoms are assumed to have hydrogens
     #: covalently bound at a maximum distance of 1.5 Å.
     r_cov = defaultdict(lambda: 1.5,  # default value
-                        N=1.31, O=1.31, P=1.58, S=1.55)
+                        N=1.31, O=1.31, P=1.58, S=1.55) # noqa: E741
```
IAlibay:
```suggestion
                        N=1.31, O=1.31, P=1.58, S=1.55)  # noqa: E741
```
With the linters working again, pep8 issues like this are a lot easier to work out!

IAlibay:
@MDAnalysis/coredevs I would like a second reviewer here please

orbeckst:
I’ll have a look at it. You can also request me as a reviewer, I can’t see how to do this action from mobile.

orbeckst:
Ah, I found it… scroll down , d’0h.

package/MDAnalysis/analysis/hydrogenbonds/wbridge_analysis.py
```diff
@@ -1210,18 +1210,24 @@ def _donor2acceptor(self, donors, h_donors, acceptor):
                                            box=self.box,
                                            return_distances=True)
         if self.distance_type != 'heavy':
```
orbeckst: Can we make this clearer:
```suggestion
        if self.distance_type == 'hydrogen':
```

orbeckst: There's a comment in `__init__` "everything except 'heavy' will give the same behavior. This is not documented in the doc string which only lists "hydrogen" and "heavy". I'd like to make it so that only two documented options work. Anything else should give a ValueError.

If you don't want to add this to the PR please raise an issue.

package/MDAnalysis/analysis/hydrogenbonds/wbridge_analysis.py
```diff
@@ -1210,18 +1210,24 @@ def _donor2acceptor(self, donors, h_donors, acceptor):
...
+            tmp_distances = distances
             tmp_donors = [h_donors[donors_idx[idx]] for idx in pairs[:, 0]]
             tmp_hydrogens = [donors_idx[idx] for idx in pairs[:, 0]]
             tmp_acceptors = [acceptor[idx] for idx in pairs[:, 1]]
         else:
+            # To make sure that for the same index i, the donor (tmp_donors[i]),
```
orbeckst, line 1222: The `else` should be `elif self.distance_type == 'heavy':` followed by an `else: raise ValueError` (and a removal of the comment in `__init__` on distance_type.

package/CHANGELOG
```diff
@@ -17,11 +17,12 @@ The rules for this file:
 ??/??/?? IAlibay, pgbarletta, mglagolev, hmacdope, manuel.nuno.melo, chrispfae, 
          ooprathamm, MeetB7, BFedder, v-parmar, MoSchaeffler, jbarnoud, jandom,
          xhgchen, jaclark5, DrDomenicoMarson, AHMED-salah00, schlaicha,
-         SophiaRuan, g2707
+         SophiaRuan, g2707, xiki-tempula
 
  * 2.5.0
 
 Fixes
+  * Fix the heavy distance_type for water bridge analysis (Issue #4040, PR #4066).
```
orbeckst: Please be more explicit in what was fixed (what gave wrong results)?

orbeckst (changes requested):
@xiki-tempula , sorry, I have a number of big picture questions here; my current actual comments in the review are minimal:

Do I understand correctly that this PR fixes the calculation of water bridges based on the "heavy" keyword by _now_ correctly calculating distances between donor heavy atom and acceptor whereas before it still calculated between the donor hydrogen  and acceptor atom? 

What needs to be done (if any) with regards to the original issue #4040 that one cannot specific a water bridge between acceptors (`selection1_type='acceptor') because `selection2_type='acceptor'` does not exist (and worse: is silently ignored?!?). Am I right in saying that this PR does _not_ completely address the original issue?

orbeckst:
@xiki-tempula did you have a chance to look at my questions?

xiki-tempula:
@orbeckst Sorry I was busy preparing for the Free energy workshop and OMSF. Will come back to this after the conference.

xiki-tempula:
> Do I understand correctly that this PR fixes the calculation of water bridges based on the "heavy" keyword by now correctly calculating distances between donor heavy atom and acceptor whereas before it still calculated between the donor hydrogen and acceptor atom?

So for non "heavy" keyword, I compute the **hydrogen** to acceptor distance, then I compute the heavy atom-hydrogen-acceptor angle and the distance array and the angle array has a 1 to 1 matching.

For "heavy" keyword, I compute the **heavy atom** to acceptor distance, then I compute the heavy atom-hydrogen-acceptor angle and I still assume that they have a 1 to 1 matching. This is wrong in that one heavy atom can have more than one hydrogen attached to them which will give more than one angle for this heavy atom, which is the bug that is being reported. 

The fix is that when computing the angle, I generate a new distance array so that it has the same length as the angle array.

> What needs to be done (if any) with regards to the original issue https://github.com/MDAnalysis/mdanalysis/issues/4040 that one cannot specific a water bridge between acceptors (selection1_type='acceptor') because selection2_type='acceptor'` does not exist (and worse: is silently ignored?!?). Am I right in saying that this PR does not completely address the original issue?

This is a bit tricky issue, as this behaviour is being documented but I do agree that this is a bit counterintuitive.

> selection1_type ({"donor", "acceptor", "both"} (optional)) – Selection 1 can be ‘donor’, ‘acceptor’ or ‘both’. Note that the value for selection1_type automatically determines how selection2 handles donors and acceptors: If selection1 contains ‘both’ then selection2 will also contain ‘both’. If selection1 is set to ‘donor’ then selection2 is ‘acceptor’ (and vice versa). [‘both’].

package/MDAnalysis/analysis/hydrogenbonds/wbridge_analysis.py
orbeckst: With the test in `__init__`, this is ok now.

orbeckst (approved):
Thank you for explanations & code clean-up (and tests!!). Looks good to me now.

package/MDAnalysis/analysis/hydrogenbonds/wbridge_analysis.py
xiki-tempula: @orbeckst Thanks for the review. Given that I have already a check in `__init__`, I think it might not be good to do this any more as this is checked several times through out the code and it would be nice if they all use the same if statement. Instead of having `if self.distance_type != 'heavy':` in line 1205 and if `self.distance_type == 'hydrogen':` in line 1216

orbeckst:
Thank you @xiki-tempula !!

PR #4067, opened 2023-03-12
Move darker information to a PR comment instead

IAlibay:
Hopefully this will at least temporarily clarify the way darker works and should prevent further misunderstandings.

## Changes made in this Pull Request:
 - Darker lint now always returns green.
 - Instead we use the github api (partly PyGithub, partly RESTful), so extract & post the outcomes as a comment.

## Unknowns
  - Right now I have no idea how well this will survive in the wild. It might need tweaking in a further iteration.
  - One of the things that I _think_ might be problematic is when branches come from forks. The best I can do for this is open a separate PR once this is fixed and see if everything behaves as expected.

### DIFF
--- a/.github/workflows/linters.yaml
+++ b/.github/workflows/linters.yaml
@@ -67,8 +67,8 @@ jobs:
       env:
         GITHUB_TOKEN: ${{ github.token }}
       run: |
-        python maintainer/ci/darker-outcomes.py  --main_stat ${{ steps.darker-main-code.outcome }} \
-                                                 --test_stat ${{ steps.darker-test-code.outcome }}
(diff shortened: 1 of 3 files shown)


IAlibay:
test

IAlibay:
Current issues:
- [x] Comments are not being overwritten
- [x] Because of continue-on-fail, either one thing passes or all things fail...

IAlibay:
What happens if I double up on comments with the words "Linter Bot Results:"?

IAlibay:
Ok this should be good to review.

See the last two entries of history of this comment for how things will be updated when there is & isn't a failure: https://github.com/MDAnalysis/mdanalysis/pull/4067#issuecomment-1469802613

richardjgowers (approved):

PR #4068, opened 2023-03-13
Add xfail on randomly failing 32bit-windows h5md test

IAlibay:
This is now happening too often. With 32bit support deprecated across the board, it's probably not worth trying to work out the right combination of pins that won't have the transient failure.

Changes
------------
- Add xfail on randomly failing test
- Remove duplicate test_slice definition (flake8 actually works)

PR Checklist
------------
 - [x] Tests?


hmacdope (changes requested):
Any reason we lost the `slice` test?

IAlibay:
> Any reason we lost the `slice` test?

@hmacdope see a couple lines below, it was a duplicated test.

hmacdope (approved):
Ah ok, LGTM then.

IAlibay:
🤞🏽 this fixes our problems!

PR #4069, opened 2023-03-14
Add readthedocs preview action

IAlibay:
This should fix the problem mentioned on discord whereby it becomes difficult to know the right link to where docs are being built through the rtd action.

Because this is a `pull_request_target` action, I don't think it'll work properly until after it gets merged.

### DIFF
--- a/.github/workflows/rtd.yaml
+++ b/.github/workflows/rtd.yaml
@@ -1,0 +1,18 @@
+name: readthedocs/preview
+
+on:
+  pull_request_target:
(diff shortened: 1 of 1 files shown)


IAlibay:
Can I get a review here @MDAnalysis/coredevs ? This isn't going to do anything unless we actually merge it (i.e. the real work doesn't happen until then).

hmacdope (approved):
I suppose there is only one way to know.

Issue #3996, opened 2023-01-17
RDKitConverter improvements


cbouy:
Running the `RDKitConverter` is slow and its accuracy could always be improved.

Here are a few ideas for improvements:

| Category | Description | Status |
| --- | --- | --- |
| Speed | Use RDKit's `RunReactantInPlace` instead of using `RunReactants` while transferring atom properties from the reactant mol to the product which causes a bottleneck. | Merged in PR #4082 |
| Both | Sorting atoms by their number of heavy atom neighbors (ascending order) before running the inferring code should help in not having to patch conjugated systems in some cases. | Merged in PR #4082 |
| Both | [This line](https://github.com/MDAnalysis/mdanalysis/blob/develop/package/MDAnalysis/converters/RDKit.py#L854) should probably exclude this case: `[O-]-C(=O)` | |
| Accuracy | Analyse the failing molecules from the [RDKitConverterBenchmark ](https://github.com/MDAnalysis/RDKitConverter-benchmark/blob/main/results/failed_molecules.smi) to identify patterns and handle them with bespoke reactions. Related issue: #3339 | |
| Both | RDKit `2022.09.1` added a C++ XYZParser which also includes code to infer bond orders and charges from the topology which could be reused here as an alternative implementation. The only downside is that it requires to know in advance the total charge of the molecule. But if the total charge is known, it should be much faster than our current code. Also adding a simple wrapper around `AssignBondOrdersFromTemplate` would be very useful. | Merged in PR #4305 |

orbeckst:
@cbouy if you want others to tackle issues, may I suggest creating sub-issues linked to this parent issue?

PR #4082, opened 2023-03-19
RDKitConverter improvements

cbouy:
Part of #3996 

 - moved some variables (`MONATOMIC_CATION_CHARGES `and `STANDARDIZATION_REACTIONS`) out of the related functions to allow users fine tuning them if necessary.
 - changed the sorting of heavy atoms when inferring bond orders and charges: previously only based on the number of unpaired electrons, now based on this and the number of heavy atom neighbors.
 - use RDKit's `RunReactantInPlace` for the standardization reactions, which should result in a significant speed improvement as we don't need to use bespoke code to transfer atomic properties from the non-standardized mol to the standardized one.
 - added more test cases for bond order and charges inferring for molecules that failed in the benchmark previously.

Ideally I need to run the benchmark again to make sure that their are no regression from these changes, and I'd also like to make a comparison between the develop branch and these changes as there should be a significant speed improvement from using `RunReactantInPlace`.

PR Checklist
------------
 - [x] Tests?
 - [x] CHANGELOG updated?

----
:books: Documentation preview :books:: https://readthedocs-preview--4082.org.readthedocs.build/en/4082/


richardjgowers (approved):
thanks @cbouy ! these changes look sane to me.  I'd be curious how this approach compares to the new rdDetermineBond functions, but I think that's a long conversation to be had elsewhere

cbouy:
Yes it would be nice to have it as an alternative bond-inferring solution as well, and try it on the benchmark that I had set up some time ago. Might give it a go someday

PR #4089, opened 2023-03-23
performance improvements

### LINKED ISSUE #2063: Numpy Performance Improvements
I recently came across two small numpy tricks for better performance. I think we use similar constructs in a lot of places in the code.

**Faster norm2**
Given 2 vectors we tend to use `np.sum(a * b)` in most places. I found that `np.einsum('i,i', a, b)` is about twice as fast on a recent numpy version (1.14.5, conda-forge). For 2 random vectors with shape (3, )

`np.sum(a,b)`: 3.3 micro-seconds
`np.einsum('i,i', a, b)`: 1.5 micro-seconds

**Faster Copying of Arrays**
A search shows that we use `a.copy()` around 180 times in the code. Often we have already allocated an array with the appropriate dimensions and would like to ideally copy directly into that skipping the reallocation. This happens often when we update `TimeStep` objects.

To illustrate take the following simplified example.

```python
# preallocate
a = np.empty(b.shape)
print(id(a))
# copy
a = b.copy()
print(id(a))
```

Here we preallocate a and copy b into it. In the second line python creates a new object that overwrites `a` and therefore reallocates. This can be seen by the changing id. This code therefore does **2** allocations. Instead look at

```python
a = np.empty(b.shape)
print(id(a))
np.copyto(a, b)
print(id(a))
```
The id of `a` did not change and no extra allocation occured.

shaivimalik:
Fixes #2063

 - improves performance as discussed in Numpy Performance Improvements [#2063](https://github.com/MDAnalysis/mdanalysis/issues/2063)

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

----
:books: Documentation preview :books:: https://readthedocs-preview--4089.org.readthedocs.build/en/4089/

### DIFF
--- a/package/AUTHORS
+++ b/package/AUTHORS
@@ -203,6 +203,7 @@ Chronological list of authors
   - Patricio Barletta
   - Mikhail Glagolev
 2023
+  - Shaivi Malik
   - Christian Pfaendner
   - Pratham Chauhan
   - Meet Brijwani
--- a/package/AUTHORS
+++ b/package/AUTHORS
@@ -203,7 +203,6 @@ Chronological list of authors
   - Patricio Barletta
   - Mikhail Glagolev
 2023
-  - Shaivi Malik
   - Christian Pfaendner
   - Pratham Chauhan
   - Meet Brijwani
@@ -216,6 +215,7 @@ Chronological list of authors
   - Josh Vermaas
   - Xiaoxu Ruan
   - Egor Marin
+  - Shaivi Malik
 
 External code
 -------------
--- a/package/MDAnalysis/analysis/bat.py
+++ b/package/MDAnalysis/analysis/bat.py
@@ -378,7 +378,7 @@ class BAT(AnalysisBase):
         # Distance between second two root atoms
         # Angle between root atoms
         a012 = np.arccos(max(-1.,min(1.,np.einsum('i,i->',v01,v21)/\
-                             np.sqrt(np.einsum('i,i->',v01,v01)*np.einsum('i,i->',v21,v21)))))
+                              np.sqrt(np.einsum('i,i->',v01,v01)*np.einsum('i,i->',v21,v21)))))
         # External coordinates
         e = v01 / r01
         phi = np.arctan2(e[1], e[0])  # Polar angle
--- a/package/MDAnalysis/analysis/bat.py
+++ b/package/MDAnalysis/analysis/bat.py
@@ -378,8 +378,7 @@ class BAT(AnalysisBase):
         # Distance between second two root atoms
         # Angle between root atoms
         a012 = np.arccos(max(-1.,min(1.,np.einsum('i,i->',v01,v21)/\
-                             np.sqrt(np.einsum('i,i->',v01,v01)*
-                                     np.einsum('i,i->',v21,v21)))))
+                             np.sqrt(np.einsum('i,i->',v01,v01)*np.einsum('i,i->',v21,v21)))))
         # External coordinates
         e = v01 / r01
         phi = np.arctan2(e[1], e[0])  # Polar angle
--- a/package/MDAnalysis/analysis/bat.py
+++ b/package/MDAnalysis/analysis/bat.py
@@ -372,13 +372,14 @@ class BAT(AnalysisBase):
         v01 = p1 - p0
         v21 = p1 - p2
         # Internal coordinates
-        r01 = np.sqrt(np.sum(v01 *
-                             v01))  # Distance between first two root atoms
-        r12 = np.sqrt(np.sum(v21 *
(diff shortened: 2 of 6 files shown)


shaivimalik:
@richardjgowers @orbeckst Hello! Can anybody review my PR?

hmacdope (changes requested):
Seems reasonable! Would you be able to see if this makes a change to the benchmarks? 

See: https://github.com/MDAnalysis/mdanalysis/blob/develop/benchmarks/README.rst

shaivimalik:
@hmacdope Yes. I will get back to you with the result.

shaivimalik:
@hmacdope I ran the benchmarks 3 times and got the following results:
Screenshot 2023-03-27 at 12 27 06 AM
Screenshot 2023-03-26 at 10 32 19 PM
Screenshot 2023-03-27 at 5 28 30 AM

shaivimalik:
Hi, @hmacdope I made changes to more files where using np.einsum in place of np.sum seemed logical on a different branch and got the following benchmark results:
Screenshot 2023-03-27 at 10 57 50 PM
Screenshot 2023-03-28 at 12 37 24 AM
If these results are up to the mark, I will make these changes to the current PR branch.

hmacdope:
The decrease in performance in some cases is unexpected. Did you have stuff running in the background at the same time  @g2707 ?

shaivimalik:
@hmacdope No, I did not start any new applications during the benchmark.

hmacdope:
@g2707 I can try on my computer tonight

shaivimalik:
@hmacdope Should I add the new file changes before tonight?

hmacdope:
Yep!

shaivimalik:
Hi! @hmacdope Did you get a chance to run the benchmarks?

hmacdope:
@g2707 yes I did thanks! Performance improvements look good!  2x runs shown below!

Screenshot from 2023-03-30 10-37-09
Screenshot from 2023-03-30 11-41-02

package/MDAnalysis/core/topologyattrs.py
```diff
@@ -3305,4 +3304,4 @@ class CMaps(_Connection):
     attrname = 'cmaps'
     singular = 'cmaps'
     transplants = defaultdict(list)
-    _n_atoms = 5
+    _n_atoms = 5
```
hmacdope: Change this back to the way it was.

package/MDAnalysis/lib/transformations.py
```diff
@@ -1827,4 +1825,4 @@ def rotaxis(a, b):
 _import_module('_transformations')
 
 # Documentation in HTML format can be generated with Epydoc
-__docformat__ = "restructuredtext en"
+__docformat__ = "restructuredtext en"
```
hmacdope: Same here

package/MDAnalysis/analysis/bat.py
```diff
@@ -563,4 +564,4 @@ def Cartesian(self, bat_frame):
     @property
     def atoms(self):
         """The atomgroup for which BAT are computed (read-only property)"""
-        return self._ag
+        return self._ag
```
hmacdope: Same here

hmacdope (changes requested):
You will need a `CHANGELOG` entry (newest first), to add yourself to `AUTHORS` and as I think you are a new contributor we ask you to introduce yourself on the developer mailing list. ! Other than that looks good and all tests pass, just a few nitpicks.

package/MDAnalysis/analysis/bat.py
shaivimalik: Is it alright now?

package/AUTHORS
```diff
@@ -203,6 +203,7 @@ Chronological list of authors
   - Patricio Barletta
   - Mikhail Glagolev
 2023
+  - Shaivi Malik
```
hmacdope: Authors is last first.

hmacdope (changes requested):

package/AUTHORS
shaivimalik: Done!

package/MDAnalysis/analysis/bat.py
hmacdope: No still needs to be changed.

shaivimalik: should be fine now.

hmacdope (approved):
Looks good to me! Introduce yourself on the developer mailing list and then I can merge. :)

https://groups.google.com/g/mdnalysis-devel

shaivimalik:
@hmacdope Done!

PR #4092, opened 2023-03-26
Ensure 'HydrogenBondAnalysis' returns the correct distances when using the 'between' keyword

### LINKED ISSUE #4091: HydrogenBondAnalysis 'between' keyword returns incorrect donor-acceptor distances
Hi, I saw a [question](https://groups.google.com/g/mdnalysis-discussion/c/NNBDSUXz_MU) on the discussion forum about using the `between` keyword of `HydrogenBondAnalysis`.

After looking into it I've found an issue whereby using the `between` keyword leads to the incorrect donor-atom distances being returned. This is because when the hydrogen bonds are filtered to remove those formed by atoms not in the `between` selection, [the donors, hydrogens, and acceptors are filtered](https://github.com/MDAnalysis/mdanalysis/blob/develop/package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py#L685) but not the distances.

The donor-hydrogen-acceptor angles are not affected because these are calculated _after_ the filtering has been done.

``` python
import numpy as np

import MDAnalysis as mda
import MDAnalysisData

np.set_printoptions(suppress=True, precision=3)

data = MDAnalysisData.ifabp_water.fetch_ifabp_water()
u = mda.Universe(data.topology, data.trajectory)

    universe=u,
    hydrogens_sel='index 1',
    acceptors_sel='index 2188',
    d_a_cutoff=3.30,
    d_h_a_angle_cutoff=150,
)
hbonds.run(stop=1)
print(f"{hbonds.results.hbonds=}")

hbonds_between = HydrogenBondAnalysis(
    universe=u,
    hydrogens_sel='resname TIP3 or protein',
    acceptors_sel='resname TIP3 or protein',
    d_a_cutoff=3.30,
    d_h_a_angle_cutoff=150,
    between=["index 0", "index 2188"],  # between uses donor atoms, not hydrogens
)
hbonds_between.run(stop=1)
print(f"{hbonds_between.results.hbonds=}")
```
Output:
```
hbonds.results.hbonds=array([[   0.   ,    0.   ,    1.   , 2188.   ,    2.741,  165.581]])
hbonds_between.results.hbonds=array([[   0.   ,    0.   ,    1.   , 2188.   ,    2.981,  165.581]])
```
The timestep, donor, hydrogen, and acceptor atoms, and the angle are all correct.
However, the donor-atom distance (5th column) is wrong.

I'll try to find some time in the next couple of days to submit a fix.

2.5.0-dev0
Python 3.10.8
- Which operating system?
macOS 12.6.3

p-j-smith:
Fixes #4091

 - if the `between` keyword is used in `HydrogenBondAnalysis`, filter the donor-acceptor distance array to remove hydrogen bonds formed by atoms not in the `between` selection
 - added tests to check the distances are correct when using `between`

Previously, the donor, hydrogen, and acceptor atom groups were filtered inside the `HydrogenBondAnalysis._filter_atoms` method. I've changed the `_filter_atoms`  method to instead return a mask that can be used for filtering the atom groups as well as the distance array. This seemed nicer than having to also pass in the distances to `_filter_atoms`, but let me know if you'd prefer it done differently.

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?

----
:books: Documentation preview :books:: https://readthedocs-preview--4092.org.readthedocs.build/en/4092/


package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -733,8 +734,11 @@ def _single_frame(self):
         # Remove donor-acceptor pairs between pairs of AtomGroups we are not
         # interested in
         if self.between_ags is not None:
-            tmp_donors, tmp_hydrogens, tmp_acceptors = \
-                self._filter_atoms(tmp_donors, tmp_hydrogens, tmp_acceptors)
+            between_mask = self._filter_atoms(tmp_donors, tmp_acceptors)
+            tmp_donors = tmp_donors[between_mask]
+            tmp_hydrogens = tmp_hydrogens[between_mask]
+            tmp_acceptors = tmp_acceptors[between_mask]
+            d_a_distances = d_a_distances[between_mask]
```
richardjgowers, line 747: a little out of scope, but would it make more sense to filter before the call to `capped_distances`?

p-j-smith: good point - that would definitely be better, but it's probably not a straightforward change and I imagine there are lots of ways to do it. One way would be to iterate over the pairs of atom selections passed to `between` and calculate donor-acceptor distances for each pair. So e.g. if `between=[["protein", "SOL"], ["protein", "protein"]]`, we would calculate:
- protein donor to SOL acceptor distances
- protein acceptor to SOL donor distances  
- protein donor to protein acceptor distances

and concatenate them. But maybe that should be a separate pr after this fix is in?

orbeckst:
@richardjgowers can you assign the PR to yourself to shepherd it to completion or otherwise find someone to do it? This seems straightforward enough (at least the original fix) so I wouldn't want the patch to sit here forever, just because it slipped through. Thanks!

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
orbeckst: Separate PR is better — I'd prefer fixing things and knowing that it works and _then_ going for optimizing. I'd be more than happy if someone raises an issue for a performance improvement with a link to this discussion.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -651,16 +651,18 @@ def _get_dh_pairs(self):
...
+        This can be used to consider only hydrogen bonds between two or more
+        specified groups.
 
         Groups are specified with the `between` keyword when creating the
         HydrogenBondAnalysis object.
 
            Returns
            -------
-           donors, hydrogens, acceptors: Filtered AtomGroups
+           mask: np.ndarray
```
orbeckst, line 665: This is a private method so there's no problem changing the return value when we need to. Given that you already updated the docs, let's also do a versionchanged, just for internal use:
```suggestion
           mask: np.ndarray


        .. versionchanged:: 2.5.0
           Change return value to a mask instead of separate AtomGroups.
``

orbeckst (approved):
Thanks for the fix @p-j-smith , sorry it took such a long time to review.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
orbeckst: I raised #4130 — please comment there if you have any follow-ups.

PR #4097, opened 2023-03-27
[darker lint] try again to trigger the linter

IAlibay:
It's possible that this doesn't work on pull_target_request... in which case 🤦🏽‍♂️ 

 - 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

----
:books: Documentation preview :books:: https://readthedocs-preview--4097.org.readthedocs.build/en/4097/

### DIFF
--- a/package/MDAnalysis/coordinates/base.py
+++ b/package/MDAnalysis/coordinates/base.py
@@ -145,6 +145,18 @@ from ..auxiliary import _AUXREADERS
 from ..lib.util import asiterable, Namespace, store_init_arguments
 from ..lib.util import NamedStream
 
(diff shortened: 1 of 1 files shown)


IAlibay:
Well after all this, it is indeed the fact that you can't lint on pull_request_target, because well it's not the merge code you're checking out.

Options are:
  - Just revert back to having everything run on pull_request, darker lint fails all the time and we live with it.
  - Somehow we pass the output values of a pull_request workflow to a pull_request_target workflow, and then have the latter write to the PR.

IAlibay:
✅ Give the right message and the right link
✅ Works on PRs from forks

Seems like we're done here.

Issue #4058, opened 2023-03-07
ENH: do_mtop performance on large (tpr) topologies


tylerjereddy:
While working on a seminar slide to demonstrate the usage of [`darshan`](https://github.com/darshan-hpc/darshan) to profile analysis IO behavior, I had a hypothesis that when reading in a 25-million particle GROMACS topology/trajectory, and only doing a simple coordinate query (i.e, find the centroid of the system), that most time would be spent on IO. It appears, surprisingly, that reading the bytes off the disk is lightning-fast, and the real bottleneck (~99 % time) is classic Python performance issues like appending to lists over and over in our topology construction, and in particular `do_mtop` in `topology/tpr/utils.py`.

First, the surprising IO "heatmaps" that had me scratching my head--almost no time spent doing genuine IO during the course of the analysis:

![image](https://user-images.githubusercontent.com/7903078/223550353-80455872-fc9c-4739-a4f9-91a61fd38caa.png)

Most of the 75+ seconds are spent doing Python appends according to `line_profiler`. We end up doing more than a quarter of a billion appends in pure Python here:

<details>

```
Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   288                                           @profile
   289                                           def do_mtop(data, fver, tpr_resid_from_one=False):
   290                                               # mtop: the topology of the whole system
   291         1        602.1    602.1      0.0      symtab = do_symtab(data)
   292         1          2.4      2.4      0.0      do_symstr(data, symtab)  # system_name
   293         1       4901.7   4901.7      0.0      do_ffparams(data, fver)  # params
   294                                           
   295         1          1.1      1.1      0.0      nmoltype = data.unpack_int()
   296         1          0.4      0.4      0.0      moltypes = []  # non-gromacs
   297        27          5.3      0.2      0.0      for i in range(nmoltype):
   298        27      40770.1   1510.0      0.0          moltype = do_moltype(data, symtab, fver)
   299        27          8.8      0.3      0.0          moltypes.append(moltype)
   300                                           
   301         1          1.1      1.1      0.0      nmolblock = data.unpack_int()
   302                                           
   303         1          2.8      2.8      0.0      mtop = obj.Mtop(nmoltype, moltypes, nmolblock)
   304                                           
   305         1          0.1      0.1      0.0      bonds = []
   306         1          0.1      0.1      0.0      angles = []
   307         1          0.1      0.1      0.0      dihedrals = []
[...]

PR #4098, opened 2023-03-28
WIP, ENH: faster TPR topology building for large systems

tylerjereddy:
Related to gh-4058
 
- for the same scenario reported there, this drops the read-in time from 76 seconds to 54 seconds; there are some pretty serious caveats though:
  - I'm not convinced the testsuite, which passes locally, is actually sufficient to catch mistakes here--consider for example that I don't add atom indices to `impropers`, `dihedrals`, and `angles` since the testsuite only caught me not doing it for `bonds` (if this is a real issue, this should be useful information for the team re: test coverage)
  - some data types may have changed, and in general I wasn't particularly sympathetic to preservation of types or data structures if the test suite didn't catch me
  - I didn't check if any of the standard `asv` benchmarks are sensitive to these changes, nor did I investigate whether the large system-tailored changes here add a large constant overhead for smaller systems, or systems with different distrubutions of topological entities (i.e., larger num molecules to total atom ratio, etc.)

Still a lot of work to do before we're IO-bound though:
image

----
:books: Documentation preview :books:: https://readthedocs-preview--4098.org.readthedocs.build/en/4098/

### DIFF
--- a/package/MDAnalysis/topology/tpr/utils.py
+++ b/package/MDAnalysis/topology/tpr/utils.py
@@ -329,6 +329,8 @@ def do_mtop(data, fver, tpr_resid_from_one=False):
         mt = mtop.moltypes[mb.molb_type]  # mt: molecule type
         molblock = mt.name.decode('utf-8')
         segid = f"seg_{i}_{molblock}"
+        # NOTE: many calculations below appear to be incorrect,
+        # but the test suite says otherwise, or at least is not sensitive...
         if mt.bonds is not None:
             curr_bonds = np.tile(mt.bonds, mb.molb_nmol).reshape((mb.molb_nmol * len(mt.bonds), 2))
             adder = np.repeat(np.arange(0, mt.number_of_atoms() * mb.molb_nmol, mt.number_of_atoms()), 2 * len(mt.bonds)).reshape(-1, 2) + atom_start_ndx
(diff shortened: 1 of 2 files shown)


package/MDAnalysis/topology/tpr/utils.py
```diff
@@ -326,32 +326,63 @@ def do_mtop(data, fver, tpr_resid_from_one=False):
...
+            bonds.extend(curr_bonds)
+        if mt.angles is not None:
+            curr_angles = np.tile(mt.angles, mb.molb_nmol).reshape((mb.molb_nmol * len(mt.angles), 3))
+            angles.extend(curr_angles)
+        if mt.dihe is not None:
+            curr_dihedrals = np.tile(mt.dihe, mb.molb_nmol).reshape((mb.molb_nmol * len(mt.dihe), 4))
+            dihedrals.extend(curr_dihedrals)
+        if mt.impr is not None:
+            curr_impropers = np.tile(mt.impr, mb.molb_nmol).reshape((mb.molb_nmol * len(mt.impr), 4))
+            impropers.extend(curr_impropers)
```
tylerjereddy, line 347: If you look at how i.e., `mt.remap_impr` works, I think it may be surprising that I can get away with this re: test coverage.

package/MDAnalysis/topology/tpr/utils.py
```diff
@@ -326,32 +326,63 @@ def do_mtop(data, fver, tpr_resid_from_one=False):
...
-            dihedrals.extend(mt.remap_dihe(atom_start_ndx))
-            impropers.extend(mt.remap_impr(atom_start_ndx))
-
-            atom_start_ndx += mt.number_of_atoms()
-            res_start_ndx += mt.number_of_residues()
+        # NOTE: many calculations below appear to be incorrect,
+        # but the test suite says otherwise, or at least is not sensitive...
+        if mt.bonds is not None:
+            curr_bonds = np.tile(mt.bonds, mb.molb_nmol).reshape((mb.molb_nmol * len(mt.bonds), 2))
+            adder = np.repeat(np.arange(0, mt.number_of_atoms() * mb.molb_nmol, mt.number_of_atoms()), 2 * len(mt.bonds)).reshape(-1, 2) + atom_start_ndx
```
tylerjereddy, line 336: good that the testsuite forced me to think carefully here, but not so for the quantities below?

package/MDAnalysis/topology/tpr/utils.py
```diff
@@ -326,32 +326,63 @@ def do_mtop(data, fver, tpr_resid_from_one=False):
...
+        adder = np.repeat(np.arange(0, mt.number_of_residues() * mb.molb_nmol, mt.number_of_residues()), mt.number_of_atoms())
+        resids_mol += adder
+        resids.extend(resids_mol)
+        resnames.extend(np.tile(resnames_mol, mb.molb_nmol))
+        atomnames.extend(np.tile(atomnames_mol, mb.molb_nmol))
+        atomtypes.extend(np.tile(atomtypes_mol, mb.molb_nmol))
+        moltypes.extend([molblock] * (mt.number_of_atoms() * mb.molb_nmol))
+        charges.extend(np.tile(charges_mol, mb.molb_nmol))
+        masses.extend(np.tile(masses_mol, mb.molb_nmol))
+        elements.extend(np.tile(elements_mol, mb.molb_nmol))
```
tylerjereddy, line 377: also, as a general comment--this is kind of partial adoption of NumPy, since we could try to embrace NumPy data structures more completely at the higher level rather than extending, but this is where I landed based on the initial bottlenecks from `line_profiler`

jbarnoud:
Do you have tests to suggest that would reassure you?

package/MDAnalysis/topology/tpr/utils.py
richardjgowers: I think ultimately we need to move TPR parsing into being compiled.  I was holding off on this on the hope that the TNG format (which was meant to hold topology info etc) might be a good replacement, but I don't think that's going to happen.  So I'd probably take these nice improvements and ultimately we might have to cythonize this module.

tylerjereddy:
> Do you have tests to suggest that would reassure you?

I guess the easiest thing might be to assume that we're correct at the moment and add regression tests to preserve current data structures that don't look fully tested like `dihedrals`? I'm happy to be wrong here though if the feeling is that we're well-covered already.

One other thing I should note for Richard before I forget is that I noticed something like 20 seconds was spent remapping string indices and stuff in `_StringInternerMixin`. I didn't get anywhere performance-wise with just minor changes, but I'll paste the idea I was playing around with below the fold. I think there were 93 million pure-Python loop hits on some of the lines for my ~25 M particle system.

<details>

```diff
--- a/package/MDAnalysis/core/topologyattrs.py
+++ b/package/MDAnalysis/core/topologyattrs.py
@@ -696,6 +696,7 @@ class _StringInternerMixin:
        Mashed together the different implementations to keep it DRY.
     """
 
+    #@profile
     def __init__(self, vals, guessed=False):
         self._guessed = guessed
 
@@ -705,16 +706,14 @@ class _StringInternerMixin:
 
         self.nmidx = np.zeros_like(vals, dtype=int)  # the lookup for each atom
         # eg Atom 5 is 'C', so nmidx[5] = 7, where name_lookup[7] = 'C'
-
-        for i, val in enumerate(vals):
-            try:
-                self.nmidx[i] = self.namedict[val]
-            except KeyError:
-                nextidx = len(self.namedict)
-                self.namedict[val] = nextidx
-                name_lookup.append(val)
-
-                self.nmidx[i] = nextidx
+        unique_val_strings = set(vals)
+        for idx, val in enumerate(unique_val_strings):
+            self.namedict[val] = idx
+            name_lookup.append(val)
+
+        for unique_val in unique_val_strings:
+            mask = (vals == unique_val)
+            self.nmidx[mask] = self.namedict[unique_val]
 
         self.name_lookup = np.array(name_lookup, dtype=object)
         self.values = self.name_lookup[self.nmidx]
diff --git a/package/MDAnalysis/topology/tpr/utils.py b/package/MDAnalysis/topology/tpr/utils.py
index a72fb5dce..90d27b939 100644
--- a/package/MDAnalysis/topology/tpr/utils.py
+++ b/package/MDAnalysis/topology/tpr/utils.py
@@ -354,8 +354,8 @@ def do_mtop(data, fver, tpr_resid_from_one=False):
             res_start_ndx += mt.number_of_residues()
 
     atomids = Atomids(np.array(atomids, dtype=np.int32))
-    atomnames = Atomnames(np.array(atomnames, dtype=object))
-    atomtypes = Atomtypes(np.array(atomtypes, dtype=object))
+    atomnames = Atomnames(np.array(atomnames, dtype="U"))
+    atomtypes = Atomtypes(np.array(atomtypes, dtype="U"))
     charges = Charges(np.array(charges, dtype=np.float32))
     masses = Masses(np.array(masses, dtype=np.float32))

```

</details>

There are some comments around the module like "woe betide anyone" who changes this, so probably some messy choices to think about. I wonder if we'd be more free to choose much faster data structure options if we were allowed to break back compat--maybe something to think about longer-term.

richardjgowers:
@tylerjereddy thanks for the tip on the string slowness.  I think the string interning makes selection much faster, but I'll admit that current construction is probably lacking.  I'll have a think if there's a way to have best of both worlds here, I think your example code is in the right direction.

orbeckst:
@richardjgowers @jbarnoud could one of you review this PR and just be generally responsible for it?

orbeckst:
@jbarnoud are you still able to look after this PR?

jbarnoud:
I have been very optimistically keeping the GitHub notification on my list for many month but it is very unlikely I'll find time to look at it.

orbeckst:
@richardjgowers might you be able to take over from @jbarnoud to look after the PR, please?

PR #4099, opened 2023-03-29
[CI] Finally fix darker lint, probably.

IAlibay:
Recapping how we got here:

1. The old darker linting action wasn't seen as acceptable because it would fail and confuse everybody, which led to chaos across our pull requests.
2. The proposed solution was to make the action never fail, but instead write out a comment that would point out that folks might want to look at their code and see if there's anything failing that they should be concerned about. This would then make it the responsibility of both the reviewers and the code author to look at it and go "nah it's just an overzealous `black` or yes these are legit flake8 issues and we need to fix thatt".
3. Because of security reasons, github rightly makes is such that `pull_request` events don't have write permissions, so you can't use them to write out a comment inside pull requests.
4. Because of securitty reasons, `pull_request_target` events don't clone the PR code, but instead the current state of the default / protected branch, which means that you can't have that do the linting either.
5. The solution here is to have linting be done in a `pull_request` event, but do the PR comment be raised in a write-allowed `workflow_run` event that tis triggered by the completion of the `pull_request` event.
6. To do this, in the `pull_request` event store a small json file (order of a few bytes) as an artifact, upload it to the github servers, and then download it back in the `workflow_run` event and process that.
7. Security wise, this should be reasonable fine unless there's a security exploit with reading json files in Python. If there was, I'd have a long list of things I'd be more concerned about first.

As far as I can tell (tested on my own fork), this _should_ work. We won't know until it goes into production though, because `workflow_run` won't work unless it's already present on the default branch.

PR Checklist
------------
 - Tests?
 - Docs?
 - CHANGELOG updated?
 - Issue raised/referenced?

----
:books: Documentation preview :books:: https://readthedocs-preview--4099.org.readthedocs.build/en/4099/

### DIFF
--- a/.github/workflows/linters.yaml
+++ b/.github/workflows/linters.yaml
@@ -33,7 +33,7 @@ permissions:
 
 jobs:
   darker_lint:
-    if: "github.repository == 'MDAnalysis/mdanalysis'"
+    #if: "github.repository == 'MDAnalysis/mdanalysis'"
     runs-on: ubuntu-latest
     permissions:
       pull-requests: write
--- a/.github/workflows/linters.yaml
+++ b/.github/workflows/linters.yaml
@@ -84,7 +84,7 @@ jobs:
     - name: upload-status
       uses: actions/upload-artifact@v3
       with:
-        name: darker_${{ env.PULL_NUMBER }}
+        name: darkerlint
         path: darker_results/
         retention-days: 1
 
--- a/.github/workflows/linters.yaml
+++ b/.github/workflows/linters.yaml
@@ -3,7 +3,6 @@ on:
   pull_request:
     branches:
       - develop
-      - main
 
 concurrency:
   # Probably overly cautious group naming.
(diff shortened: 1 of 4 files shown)


.github/workflows/linters.yaml
```diff
@@ -1,9 +1,6 @@
 name: "linters"
 on:
-  push:
```
IAlibay, line 3: No longer doing this on push to avoid `workflow_run` being activated when there's no PR. Hopefully we shouldn't ever be in a position where we merge when important linters failed.

IAlibay:
> ### Linter Bot Results:
> 
> Hi @IAlibay! Thanks for making this PR. We linted your code and found the following: 
> 
> Some issues were found with the formatting of your code.
> | Code Location | Outcome |
> | --- | --- |
> | main package | ⚠️  Possible failure|
> | testsuite | ⚠️  Possible failure|
> 
> Please have a look at the `darker-main-code` and `darker-test-code` steps here for more details: https://github.com/MDAnalysis/mdanalysis/actions/runs/4549636191/jobs/8021892068
> 
> ---
> _**Please note:** The `black` linter is purely informational, you can safely ignore these outcomes if there are no flake8 failures!_

Don't trust this, it's behaving this way because it's linting the latest commit on develop instead. That being said, this does highlight the fact that we've been letting flake8 breaking stuff through.

hmacdope (approved):

PR #4101, opened 2023-03-29
changing np to cnp

### LINKED ISSUE #3908: change `cimport numpy as np` to `cimport numpy as cnp` for consistency

We should be consistent with the way we cimport numpy in the codebase. 

Lots of places currently use
```cython
 cimport numpy as np
```

While the more modern approach and the one taken in some PRs #3777 , `timestep.pyx` and a few random other existing places is to

```cython 
 cimport numpy as cnp
```

Personally I prefer the second one (and its the one used by pandas) as it is much clearer when the C API is being used vs python API. 
@tylerjereddy  do you know if there is a standard for scipy?

Regardless of which way we go we should be consistent if possible.

Let me know what you think on this issue and if we come to a conclusion I can raise a seperate issue for taking some action. 
Might be a nice easy `good-first-issue` etc.

HunterZ17:
Fixes #3908 

 - changed `cimport numpy as np` to `cimport numpy as cnp` and it's related uses of `np` to `cnp`

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

----
:books: Documentation preview :books:: https://readthedocs-preview--4101.org.readthedocs.build/en/4101/

### DIFF
--- a/package/MDAnalysis/lib/_augment.pyx
+++ b/package/MDAnalysis/lib/_augment.pyx
@@ -25,13 +25,13 @@
 import cython
 import numpy as np
 from .mdamath import triclinic_vectors
-cimport numpy as np
+cimport numpy as cnp
 cimport MDAnalysis.lib._cutil
 from MDAnalysis.lib._cutil cimport _dot, _norm, _cross
 
 from libcpp.vector cimport vector
 
-np.import_array()
+cnp.import_array()
 
 
 __all__ = ['augment_coordinates', 'undo_augment']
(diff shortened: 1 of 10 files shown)


richardjgowers (changes requested):
@HunterZ17 thanks for starting this.  It looks like there's some test failures from where you can't use the `cnp` Cython module but instead should use the `np` (Python) import

HunterZ17:
@richardjgowers So, Should I run tests locally and replace the needed `cnp` with `np`.

RMeli:
@HunterZ17 Yes, you should understand which imports should be `np` and which one should be `cnp`. Testing thing locally will allow you to iterate much more quickly and push to this PR when things are working.

richardjgowers:
Yes that should make it work

package/MDAnalysis/lib/_augment.pyx
```diff
@@ -296,12 +296,12 @@ def augment_coordinates(float[:, ::1] coordinates, float[:] box, float r):
                 output.push_back(coord[j] - shiftZ[j])
             indices.push_back(i)
     n = indices.size()
-    return np.asarray(output, dtype=np.float32).reshape(n, 3), np.asarray(indices, dtype=np.intp)
+    return np.asarray(output, dtype=cnp.float32).reshape(n, 3), np.asarray(indices, dtype=cnp.intp)
```
Telomelonia: ig `float32` and `intp` are of numpy module rather than `cnp`

HunterZ17:
After testing locally, all `.pyx` files were running smoothly but all `.pxd` files were showing errors, always some methods are not defined as errors. Is there a different method to test `.pxd` file?

hmacdope:
@HunterZ17  do you plan on continuing to work on this? If not we will close.

HunterZ17:
@hmacdope Yes, you can close this.

PR #4102, opened 2023-03-29
Add necessary commits for 2.4.3

IAlibay:
Cherry picked commits for 2.4.3:
* Fixes DCD seeking for large (2Gb+) files.

Commits to prepare for 2.4.3:
* Update version numbers

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -13,6 +13,15 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" http://semver.org
 
 ------------------------------------------------------------------------------
+03/29/23 richardjgowers
+
+ * 2.4.3
+
+Fixes
+  * Fixed DCD reading for large (>2Gb) files (Issue #4039).  This was broken
+    for versions 2.4.0, 2.4.1 and 2.4.2
+
+
 01/04/23 IAlibay
 
  * 2.4.2
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -13,13 +13,14 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" http://semver.org
 
 ------------------------------------------------------------------------------
-03/29/23 richardjgowers
+03/29/23 richardjgowers, IAlibay
 
  * 2.4.3
 
 Fixes
   * Fixed DCD reading for large (>2Gb) files (Issue #4039).  This was broken
     for versions 2.4.0, 2.4.1 and 2.4.2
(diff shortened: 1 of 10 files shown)


IAlibay:
cirrus failure is expected - I don't see a need to cherry pick the fix here (since we don't use it for release), unless folks think otherwise

IAlibay:
I guess we'll need to cherry pick the parmed fix here.

IAlibay:
:/ this is a bit of a difficult one, if we cherry-pick the parmed fix, do we then change the changelog on develop to reflect that it went in on 2.4.3? Or do we just do it here and leave it as-is?

The alternative option is to pin the upper value of Parmed in tests & conda-forge?

IAlibay:
Note - it's a fix on a failing test, but it didn't require any change to the Parmed parser codes.

IAlibay:
Pinging @MDAnalysis/coredevs it'd be good to get a quick decision on the path forward here.

richardjgowers:
if the parmed fix is possible via a conda-forge pin it might make sense to do 2.4.3 that way?

package/CHANGELOG
```diff
@@ -13,6 +13,15 @@ The rules for this file:
...
 
 ------------------------------------------------------------------------------
+03/29/23 richardjgowers
+
+ * 2.4.3
+
+Fixes
+  * Fixed DCD reading for large (>2Gb) files (Issue #4039).  This was broken
+    for versions 2.4.0, 2.4.1 and 2.4.2
+
```
orbeckst: Just as a reminder. We have to retroactively change the main develop CHANGELOG to include 2.4.3 and reshuffle entries.

orbeckst (approved):
If the parmed fix thing does not require an increase in minor version then you can put it in here without problems (especially if it's just something in the tests). 

As with anything else, we just need to retroactively change develop's CHANGELOG.

EDIT: and add yourself to the 2.4.3 authors list @IAlibay

package/CHANGELOG
orbeckst: I just saw that you had already snuck it into PR #4048 ...

IAlibay:
Ok, given we need to update the author list anyways, I'll just cherry-pick the parmed test fix and separately change the changelog for `develop` (these bugfix changes don't guarantee a oorresponding history with `develop` anyways, so it being a bit out of synch doesn't matter).

RMeli:
Spack updated: https://github.com/spack/spack/pull/36541

PR #4105, opened 2023-03-30
Add simple atomic distance analysis (#3654)

### LINKED ISSUE #3654: add simple atomic distance analysis to analysis.distances

We have many analysis tools for complicated problems, all using the consistent AnalysisBase approach. However, there's no functionality for fairly trivial but nevertheless commonly used analysis tasks, namely **calculate the distance (under PBC) between two atom groups, atom by atom**. 

Namely, for AtomGroups `g1` and `g2`, I would like to calculate the distances `|g1[i] - g2[i]|` for all  `i` with a class such as
```python
d12 = MDAnalysis.analysis.distances.DistanceAnalysis(g1, g2).run()
plt.histogram(d12.results.distances)
```

The constructor could take `pbc=True | False` as kwarg, in case someone wants to also calculate distances without PBC.

Using `calc_bonds()` in the actual implementation should produce good performance.

This is a simple problem so one can always write one's own's code. However, providing well-tested building blocks aids in reproducibility and reduces code duplication. It also makes standard analysis available in **mdacli**.

Or use the GROMACS tool `gmx distance` :-p.

## Additional context ##

Code could be similar to https://github.com/MDAnalysis/mdanalysis/issues/3310#issuecomment-1086964719 or NucPairDist https://github.com/MDAnalysis/mdanalysis/blob/3a03af01886977aa53d2a43086cae86c45038738/package/MDAnalysis/analysis/nucleicacids.py#L61

xhgchen:
* Resolves #3654
* Add class `AtomicDistances` to `MDAnalysis.analysis.distances` to calculate the distances
  `|ag1[i] - ag2[i]|` for all `i` from `0` to `n_atoms - 1` for each frame over the trajectory
* Allow periodic boundary conditions to be considered or ignored in class `AtomicDistances` by setting kwarg `pbc` to `True` or `False`
* Add unit tests for class `AtomicDistances` to `test_distances.py`

The class itself is inspired by [this comment in #3310](https://github.com/MDAnalysis/mdanalysis/issues/3310#issuecomment-1086964719) mentioned in #3654. How would I go about crediting them if this PR gets approved? I changed the list to a NumPy array to improve performance and added the `pbc` kwarg, as outlined above.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

----
:books: Documentation preview :books:: https://readthedocs-preview--4105.org.readthedocs.build/en/4105/


package/MDAnalysis/analysis/distances.py
```diff
@@ -202,3 +204,65 @@ def between(group, A, B, distance):
...
+
+    Parameters
+    ----------
+    ag1, ag2 : AtomGroup
+        :class:`~MDAnalysis.core.groups.AtomGroup` with the
+        same number of atoms
+    pbc : bool, optional
+        If ``True``, calculates atomic distances with periodic boundary
+        conditions (PBCs). The unitcell dimensions of the system must be
+        orthogonal or triclinic for the calculations with PBCs to work.
```
richardjgowers: We don't need to mention what the unitcell must be here as we've previously enforced this

package/MDAnalysis/analysis/distances.py
```diff
@@ -55,6 +55,8 @@
 
 import warnings
 import logging
+from .base import AnalysisBase
```
richardjgowers: It might be cleaner to create a new file for this analysis class.  It feels a little weird to import `base` into `distances`.  Once you get circular imports things get confusing fast.

package/MDAnalysis/analysis/distances.py
```diff
@@ -202,3 +204,65 @@ def between(group, A, B, distance):
...
+        self._ag1 = ag1
+        self._ag2 = ag2
+        self._pbc = pbc
+
+    def _prepare(self):
+        # initialize NumPy array of frames x distances for results
+        self.results = np.zeros((self.n_frames, self._ag1.atoms.n_atoms))
+
+    def _single_frame(self):
+        if (self._pbc):
```
richardjgowers: a fun hack is to set a variable (`box`) to either `.dimensions` or `None` depending on `self._pbc`, then the call to `calc_bonds` can remain the same, and it's clearer what's happening (imho)

e.g.
```python
box = self._ag1.dimensions if self._pbc else None

self.results = calc_bonds(self._ag1.positions, self._ag2.positions, box=box)
```

package/MDAnalysis/analysis/distances.py
```diff
@@ -202,3 +204,65 @@ def between(group, A, B, distance):
...
+
+    Attributes
+    ----------
+    results : :class:`numpy.ndarray`
+        The distances `|ag1[i] - ag2[i]|` for all `i` from `0` to
+        `n_atoms - 1` for each frame over the trajectory.
+    n_frames : int
+        Number of frames included in the analysis.
+    n_atoms : int
+        Number of atoms in each atom group.
```
richardjgowers: I don't see where these variables are set?

testsuite/MDAnalysisTests/analysis/test_distances.py
```diff
@@ -235,3 +235,136 @@ def test_between_return_type(self, dists, group, ag, ag2):
             dists
         )
         assert isinstance(actual, MDAnalysis.core.groups.AtomGroup)
+
+
+class TestAtomicDistances(object):
+    @staticmethod
+    @pytest.fixture()
+    def ad_u():
+        return MDAnalysis.Universe(GRO, XTC)
```
richardjgowers: can we use a `mda.Universe.empty()` here to avoid having to read files for a test?

For its positions, you could either use a seeded random number, or just something like `np.arange(natoms * 3).reshape(natoms, -1)` to generate arbitrary coordinates

testsuite/MDAnalysisTests/analysis/test_distances.py
```diff
@@ -235,3 +235,136 @@ def test_between_return_type(self, dists, group, ag, ag2):
...
+    @staticmethod
+    @pytest.fixture()
+    def ad_ag4():
+        u = MDAnalysis.Universe(GRO)
+        # different trajectory, no XTC (exception)
+        return u.atoms[-5:]
+
+    @staticmethod
+    @pytest.fixture()
+    def expected_scipy(ad_ag1, ad_ag2):
```
richardjgowers: I'm not sure we need a check against scipy, the check against `dist` is fine

testsuite/MDAnalysisTests/analysis/test_distances.py
```diff
@@ -235,3 +235,136 @@ def test_between_return_type(self, dists, group, ag, ag2):
...
+
+            # box size (lx, ly, lz)
+            box = ad_ag1.dimensions[:3]
+
+            # apply minimum image convention i.e. take box - dist
+            # when dist > box / 2
+            dist = np.where(dist > box / 2, box - dist, dist)
+
+            # desired dist = sqrt((x2 - x1)^2 + (y2 - y1)^2 + (z2 - z1)^2)
+            expected[i] = np.sqrt(np.square(dist).sum(axis=1))
```
richardjgowers: again just use `dist` here.  we're not testing that our distance calculations work here (we do that elsewhere).  We're testing that **this class** is correctly applying those functions.

richardjgowers (changes requested):
this is a good start!

package/MDAnalysis/analysis/distances.py
xhgchen: Thanks for all the feedback! I learned a lot working through it, and I have pushed some more commits.

I added those variables because other classes seemed to have them (in the docstring) and I used those as reference. 

`results` refers to `self.results` in `def _prepare(self)` and `def _single_frame(self)`
`n_frames` is used to make the NumPy array in `def _prepare(self)`
`n_atoms` is used to check that `ag1` and `ag2` have the same number of atoms and in `def _prepare(self)` to make the NumPy array

Would it be better to remove some/all of them (from the docstring)?

testsuite/MDAnalysisTests/analysis/test_distances.py
xhgchen: So in order to do this, I had to change the `dist()` tests to use `calc_bonds()` instead because the Universe does not have resids

xhgchen:
Now that all the checks have passed, @richardjgowers could you take a look at the new files when you get the chance? I would appreciate input on the file name as well. I was torn between `atomic_distance_analysis.py` and `atomicdistances.py` but I went with the latter because the first option seemed like a mouthful.

xhgchen:
Just updating the docs so they work properly :)

This is the first time I have ever written docs, but it was fun.

Edit: The 2 failing checks both seem to be from an issue uploading to Codecov. I would expect them to work with a re-run. Here is a direct link to the preview docs page for the new module: https://mdanalysis--4105.org.readthedocs.build/en/4105/documentation_pages/analysis/atomicdistances.html

Edit 2: Very minor wording fix. Hopefully the checks will run correctly this time.

xhgchen:
Since it has been a while since this was last reviewed, I was wondering when it might be looked at again? I realize that everyone is likely busy reviewing GSoC applications in addition to their other commitments and I am not in a hurry; it would just help me follow up more promptly when the time comes.

Also, the last push was just to resolve conflicts.

richardjgowers (approved):

PR #4114, opened 2023-04-04
Add writing u.trajectory.ts.data['molecule_tag'] as molecule_tag atom attribute to LAMMPS datafile

### LINKED ISSUE #3548: LAMMPS.DATAWriter should allow to write residue ID
Dear all, I am currently working on a loop where each step consists in a short MD simulation, followed by an update of atoms charges, which are used in the MD simulation in the next step. MDAnalysis has proven very helpful to process and extract the MD data, so thank you for your hard work in developing this tool! 
Unfortunately, it appears that I have a problem when using LAMMPS: I managed to read the output and update the charges, but when I try to write a new topology file, it seems that all the information about residues is lost , even though I do have it in the data I read, i.e. trj.atoms.resids. 

Browsing the issues and the documentation (https://docs.mdanalysis.org/1.0.0/_modules/MDAnalysis/coordinates/LAMMPS.html#DATAWriter) it seems that this would be the right behaviour, i.e. do not write the residue information. Is it possible to implement a feature/flag to allow the molecule/residue IDs to be written in the new data file? Actually, I could try to modify and fork the current version myself, but before I do, I would like to be sure that there is currently no possibility to write the  molecule/residue IDs using LAMMPS.DATAWriter or others. Indeed, it seems to me very strange that, even though I have this information about the residues I cannot write it. Am I missing something?

Thank you very much for your help.

I report here a very short sample of the code I am using:

```
trj = mda.Universe("datafile.data", "trajectory.lammpstrj",format = "LAMMPSDUMP")
trj.trajectory[-1]
## Update charges##
with mda.coordinates.LAMMPS.DATAWriter(newtopfile) as f:
        f.write(trj)
```

mglagolev:
Fixes #3548

Being able to write the molecule id in a LAMMPS data file is crucial for using the MDAnalysis as a generator of LAMMPS initial configurations, for example, in machine learning workflows.
Currently, the molecule_tag field reads as atoms.resids, but when writing a data file, zero values are hardcoded.
Following the discussion in #3548, I've implemented writing ts.data['molecule_tag'] into this field. 

Added data as input parameter for DATAWriter._write_atoms
If data['molecule_tag'] is present, _write_atoms will write its values as the molecule_tag attribute of the Atoms section of the data file. Otherwise, it will fill molecule_tag with zeroes, consistent with the previous behavior.

Added special read-write-read function to the tests, which sets data['molecule_tag'] values from atoms.resids after reading the data file. Added the class to check the resids after using the aforementioned function.

PR Checklist
------------
 - [V] Tests?
 - [X] Docs?
 - [V] CHANGELOG updated?
 - [V] Issue raised/referenced?

----
:books: Documentation preview :books:: https://readthedocs-preview--4114.org.readthedocs.build/en/4114/

### DIFF
--- a/package/MDAnalysis/coordinates/LAMMPS.py
+++ b/package/MDAnalysis/coordinates/LAMMPS.py
@@ -288,11 +288,7 @@ class DATAWriter(base.WriterBase):
         indices = atoms.indices + 1
         types = atoms.types.astype(np.int32)
 
-        try:
-            moltags = data['molecule_tag']
-        except KeyError:
-            moltags = [0] * len(atoms)
-
+        moltags = data.get("molecule_tag", np.zeros(len(atoms), dtype=int))
 
         if self.convert_units:
             coordinates = self.convert_pos_to_native(atoms.positions, inplace=False)
--- a/package/MDAnalysis/coordinates/LAMMPS.py
+++ b/package/MDAnalysis/coordinates/LAMMPS.py
@@ -273,7 +273,7 @@ class DATAWriter(base.WriterBase):
         self.units['velocity'] = kwargs.pop('velocityunit',
                                  self.units['length']+'/'+self.units['time'])
 
-    def _write_atoms(self, atoms):
+    def _write_atoms(self, atoms, data):
         self.f.write('\n')
         self.f.write('Atoms\n')
         self.f.write('\n')
@@ -288,20 +288,27 @@ class DATAWriter(base.WriterBase):
         indices = atoms.indices + 1
         types = atoms.types.astype(np.int32)
 
+        try:
+            moltags = data['molecule_tag']
+        except KeyError:
+            moltags = [0] * len(atoms)
+
+
         if self.convert_units:
             coordinates = self.convert_pos_to_native(atoms.positions, inplace=False)
 
         if has_charges:
-            for index, atype, charge, coords in zip(indices, types, charges,
-                    coordinates):
-                self.f.write('{i:d} 0 {t:d} {c:f} {x:f} {y:f} {z:f}\n'.format(
-                             i=index, t=atype, c=charge, x=coords[0],
+            for index, moltag, atype, charge, coords in zip(indices, moltags,
+                    types, charges, coordinates):
+                self.f.write('{i:d} {m:d} {t:d} {c:f} {x:f} {y:f} {z:f}\n'.format(
+                             i=index, m=moltag, t=atype, c=charge, x=coords[0],
                              y=coords[1], z=coords[2]))
         else:
-            for index, atype, coords in zip(indices, types, coordinates):
-                self.f.write('{i:d} 0 {t:d} {x:f} {y:f} {z:f}\n'.format(
-                             i=index, t=atype, x=coords[0], y=coords[1],
-                             z=coords[2]))
+            for index, moltag, atype, coords in zip(indices, moltags, types,
+                    coordinates):
+                self.f.write('{i:d} {m:d} {t:d} {x:f} {y:f} {z:f}\n'.format(
+                             i=index, m=moltag, t=atype, x=coords[0],
+                             y=coords[1], z=coords[2]))
 
     def _write_velocities(self, atoms):
         self.f.write('\n')
@@ -397,6 +404,8 @@ class DATAWriter(base.WriterBase):
         else:
             frame = u.trajectory.ts.frame
 
+        data = u.trajectory.ts.data
+
         # make sure to use atoms (Issue 46)
         atoms = selection.atoms
 
@@ -441,7 +450,7 @@ class DATAWriter(base.WriterBase):
             self._write_dimensions(atoms.dimensions)
 
             self._write_masses(atoms)
-            self._write_atoms(atoms)
+            self._write_atoms(atoms, data)
             for attr in features.values():
                 if attr is None or len(attr) == 0:
                     continue
--- a/package/MDAnalysis/coordinates/LAMMPS.py
+++ b/package/MDAnalysis/coordinates/LAMMPS.py
@@ -296,15 +296,15 @@ class DATAWriter(base.WriterBase):
         if has_charges:
             for index, moltag, atype, charge, coords in zip(indices, moltags,
                     types, charges, coordinates):
-                self.f.write('{i:d} {m:d} {t:d} {c:f} {x:f} {y:f} {z:f}\n'.format(
-                             i=index, m=moltag, t=atype, c=charge, x=coords[0],
-                             y=coords[1], z=coords[2]))
+                x, y, z = coords
+                self.f.write(f"{index:d} {moltag:d} {atype:d} {charge:f}"
+                             f" {x:f} {y:f} {z:f}\n")
         else:
             for index, moltag, atype, coords in zip(indices, moltags, types,
                     coordinates):
-                self.f.write('{i:d} {m:d} {t:d} {x:f} {y:f} {z:f}\n'.format(
-                             i=index, m=moltag, t=atype, x=coords[0],
-                             y=coords[1], z=coords[2]))
+                x, y, z = coords
+                self.f.write(f"{index:d} {moltag:d} {atype:d}"
+                             f" {x:f} {y:f} {z:f}\n")
 
     def _write_velocities(self, atoms):
         self.f.write('\n')
@@ -400,8 +400,6 @@ class DATAWriter(base.WriterBase):
         else:
             frame = u.trajectory.ts.frame
 
-        data = u.trajectory.ts.data
-
         # make sure to use atoms (Issue 46)
         atoms = selection.atoms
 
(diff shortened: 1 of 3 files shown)


orbeckst:
@hmacdope do you have expertise with LAMMPS to review this PR, or can you suggest someone else?

package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -288,20 +288,27 @@ def _write_atoms(self, atoms):
...
 
         if has_charges:
-            for index, atype, charge, coords in zip(indices, types, charges,
-                    coordinates):
-                self.f.write('{i:d} 0 {t:d} {c:f} {x:f} {y:f} {z:f}\n'.format(
-                             i=index, t=atype, c=charge, x=coords[0],
+            for index, moltag, atype, charge, coords in zip(indices, moltags,
+                    types, charges, coordinates):
+                self.f.write('{i:d} {m:d} {t:d} {c:f} {x:f} {y:f} {z:f}\n'.format(
+                             i=index, m=moltag, t=atype, c=charge, x=coords[0],
```
RMeli: I'm wondering if with this many fields, switching to f-strings would be cleaner?

```python
'{i:d} {m:d} {t:d} {c:f} {x:f} {y:f} {z:f}\n'.format(i=index, m=moltag, t=atype, c=charge, x=coords[0], y=coords[1], z=coords[2])
```
vs
```python
x, y, z = coords
f"{index:d} {moltag:d} {atype:d} {charge:f} {x:f} {y:f} {z:f}\n"

File: package/MDAnalysis/coordinates/LAMMPS.py [hunk 2]
```diff
@@ -288,20 +288,27 @@ def _write_atoms(self, atoms):
         indices = atoms.indices + 1
         types = atoms.types.astype(np.int32)
 
+        try:
+            moltags = data['molecule_tag']
+        except KeyError:
+            moltags = [0] * len(atoms)
+
```
RMeli:
```suggestion
        moltags = data.get("molecule_tag", np.zeros(len(atoms), dtype=int))
```

Allocating a `numpy` array should be considerably faster. `get` allows to avoid the try-except logic.

```
$ N = 10000
$ %timeit [0] * N
23.5 µs ± 359 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

$ %timeit np.zeros(N, dtype=int)
2.85 µs ± 231 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

package/MDAnalysis/coordinates/LAMMPS.py
```diff
@@ -397,6 +404,8 @@ def write(self, selection, frame=None):
         else:
             frame = u.trajectory.ts.frame
 
+        data = u.trajectory.ts.data
```
RMeli: Is this necessary since it `data` is only used one time?

testsuite/MDAnalysisTests/coordinates/test_lammps.py
```diff
@@ -115,6 +115,29 @@ def LAMMPSDATAWriter(request, tmpdir_factory):
     return u, u_new
 
 
+@pytest.fixture(params=[
+    LAMMPSdata,
+    LAMMPSdata_mini,
+    LAMMPScnt,
+    LAMMPShyd,
```
RMeli: From the code coverage comment above, it seems that none of those covers the case where there is no charge. Can we add one such case?

RMeli (commented):
A few small comments. I'll leave it to LAMMPS experts for a more in-depth review.

package/MDAnalysis/coordinates/LAMMPS.py
mglagolev: Thanks for this suggestion! This is clearly a better way.

mglagolev:
@hmacdope  Shall I improve something in this PR?

hmacdope (changes requested):
@mglagolev sorry for the delay, looks good, if you could address @RMeli's last comments about f-strings and a test case coverage for a case with no carge we should be good to go

mglagolev:
Thank you @hmacdope and, once again, @RMeli!
I've added the @RMeli's improvements, except running a test on an uncharged system.
There is an uncharged datafile, which I added with my previous commit, but it is compressed.
When adding it to the test fixture, I run into an error, apparently because the filename upon writing is changed from .data.bz2 to .data.data

Here's a simple script to test this behavior:

```
import MDAnalysis as mda
import numpy as np
u = mda.Universe.empty(1, trajectory = True)
u.add_TopologyAttr('type')
u.add_TopologyAttr('mass')
atom = mda.core.groups.Atom(u = u, ix = 0)
atom.position = np.array([0., 0., 0.])
u.atoms = mda.AtomGroup([atom,])
u.add_bonds([])
u.add_angles([])
u.add_dihedrals([])
u.add_impropers([])
u.atoms.types = ['1',]
u.atoms.masses = [1.,]
u.dimensions = [1, 1, 1, 90, 90, 90]
with mda.Writer("test.data.bz2", n_atoms = 1) as w:
	w.write(u.atoms)
```
All the attributes are set just to get the simplest datafile written without an error.
In my environments, using both the development branch and the version from pip, the datafiles are written under the name "test.data.data"
Am I missing something? Is it a bug or a feature?

mglagolev:
Maybe @fenilsuchak can help?

RMeli:
Sounds odd, I'll have to double check this behaviour. Thanks for checking and reporting.

A possible workaround/alternative to get this PR done is to use `del_TopologyAttr` on the universe obtained from one of the current files to remove the `charge` attribute.

```python
In [1]: from MDAnalysisTests.datafiles import LAMMPSdata

In [2]: import MDAnalysis as mda

In [3]: u = mda.Universe(LAMMPSdata)

In [4]: u.atoms.charges
Out[4]: array([0., 0., 0., ..., 0., 0., 0.])

In [5]: u.del_TopologyAttr('charges')

In [6]: u.atoms.charges
NoDataError                               Traceback (most recent call last)
Cell In[6], line 1
----> 1 u.atoms.charges

File ~/Documents/git/mdanalysis/package/MDAnalysis/core/groups.py:2539, in AtomGroup.__getattr__(self, attr)
   2537 elif attr == 'positions':
   2538     raise NoDataError('This Universe has no coordinates')
-> 2539 return super(AtomGroup, self).__getattr__(attr)

File ~/Documents/git/mdanalysis/package/MDAnalysis/core/groups.py:614, in GroupBase.__getattr__(self, attr)
    612     else:
    613         err = 'This Universe does not contain {singular} information'
--> 614         raise NoDataError(err.format(singular=cls.singular))
    615 else:
    616     return super(GroupBase, self).__getattr__(attr)

NoDataError: This Universe does not contain charge information
```

mglagolev:
@RMeli, thanks for the idea!
The test is passed without an error.

testsuite/MDAnalysisTests/coordinates/test_lammps.py
```diff
@@ -115,6 +115,32 @@ def LAMMPSDATAWriter(request, tmpdir_factory):
...
+    [LAMMPSdata, True],
+    [LAMMPSdata_mini, True],
+    [LAMMPScnt, True],
+    [LAMMPShyd, True],
+    [LAMMPSdata, False]
+], scope='module')
+def LAMMPSDATAWriter_molecule_tag(request, tmpdir_factory):
+    filename, charges = request.param
+    u = mda.Universe(filename)
+    if charges == False:
```
RMeli:
```suggestion
    if not charges:
```

RMeli (approved):
Thanks @mglagolev. Small nitpick (you can safely ignore it), otherwise LGTM.

testsuite/MDAnalysisTests/coordinates/test_lammps.py
IAlibay: Please review https://github.com/MDAnalysis/mdanalysis/actions/runs/4859346505/jobs/8661858282 this was highlighted by the darker linting bot as a flake8 failure (see: https://github.com/MDAnalysis/mdanalysis/pull/4114#issuecomment-1496674113)

mglagolev: Oh, thanks again to @RMeli. Flake8 should pass fine now.

hmacdope (approved):

mglagolev:
@IAlibay, sorry for bothering, is anything else required to merge this?

RMeli:
Thanks for the contribution @mglagolev, and sorry if it took quite long to get it merged.

mglagolev:
Great! Thanks to helpful people here for making this small contribution possible:-)

By the way, @RMeli  did you reproduce the issue with compressed filenames?

RMeli:
@mglagolev unfortunately no, I forgot. Thanks for reminding me! If you don't mind, could you please open an issue with your observation? Otherwise I'll open one.

mglagolev:
@rmeli Sure!
https://github.com/MDAnalysis/mdanalysis/issues/4159

PR #4121, opened 2023-04-09
Speed up build time using micromamba

IAlibay:
Let's see how much time we can save (particularly on macos) by switching to micromamba.

In this PR:
- Improved build speeds by ~ 50% (although in practice only saves ~ 30-50s per runner total).
- Removed unecessary runners / moved them to Cron CI
- Add old ubuntu + macos scan to cron CI.

PR Checklist
------------
 - Tests?
 - Docs?
 - CHANGELOG updated?
 - Issue raised/referenced?

----
:books: Documentation preview :books:: https://readthedocs-preview--4121.org.readthedocs.build/en/4121/

### DIFF
--- a/.github/workflows/gh-ci-cron.yaml
+++ b/.github/workflows/gh-ci-cron.yaml
@@ -40,7 +40,10 @@ jobs:
         extra-specs: |
           python=="3.11"
           pip
-        channels: conda-forge, bioconda
+        channels: |
+          jaimergp/label/unsupported-cudatoolkit-shim
+          conda-forge
+          bioconda
 
     - name: install_deps
       uses: ./.github/actions/setup-deps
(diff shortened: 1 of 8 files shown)


.github/workflows/gh-ci-cron.yaml
```diff
@@ -32,21 +32,20 @@ jobs:
...
-        channel-priority: strict
-        channels: conda-forge, bioconda
-        add-pip-as-python-dependency: true
-        architecture: x64
+        environment-file: False
+        environment-name: mda
+        extra-specs: |
+          python=="3.11"
+          pip
+        channels: jaimergp/label/unsupported-cudatoolkit-shim, conda-forge, bioconda
```
hmacdope, line 43: Question what needs `jaimergp/label/unsupported-cudatoolkit-shim`?

hmacdope (approved):
One question but other than that looks good and everything green.

.github/workflows/gh-ci-cron.yaml
IAlibay: OpenMM otherwise we're pulling down cuda everywhere. - it's quite widely used everywhere (in CI) nowadays

hmacdope: Cool beans

PR #4122, opened 2023-04-09
Fix permission errors w/ readonly tmpdirs and pytest 7.3

IAlibay:
Azure pipelines has been failing recently (see #4121 ), due to the latest pytest 7.3 release.

The source of the error seems to be a mixture of windows, python 3.9, pytest and pathlib incompatibilities when trying to clean up temporary directories that have been set to read-only.

To fix this, here we just conclude the one affected test by re-enabling write permissions to the directory and removing it ahead of the test completing.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

----
:books: Documentation preview :books:: https://readthedocs-preview--4122.org.readthedocs.build/en/4122/

### DIFF
--- a/testsuite/MDAnalysisTests/coordinates/test_xdr.py
+++ b/testsuite/MDAnalysisTests/coordinates/test_xdr.py
@@ -25,7 +25,6 @@ from unittest.mock import patch
 
 import os
 import shutil
-import stat
 import subprocess
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


IAlibay:
@MDAnalysis/coredevs this will need to take priority for merging so that we can get CI working again.

testsuite/MDAnalysisTests/coordinates/test_xdr.py
```diff
@@ -890,6 +889,15 @@ def test_persistent_offsets_readonly(self, tmpdir):
         assert_equal(os.path.exists(XDR.offsets_filename(filename,
                                                     ending='.lock')), False)
 
+        # pre-teardown permission fix - leaving permission blocked dir
```
orbeckst, line 891: add reference to issue number, please --- so that we can maybe remove again when it works again

orbeckst (approved):
ugly but ok, please see comment.

Happy Easter!

PR #4126, opened 2023-04-11
Fix rtd link in PR comments

IAlibay:
Given that we didn't notice it was failing, I do wonder how useful it actually is... :/

 - Fix project_slug entry to point to mdanalysis

PR Checklist
------------
 - Tests?
 - Docs?
 - CHANGELOG updated?
 - Issue raised/referenced?

----
:books: Documentation preview :books:: https://readthedocs-preview--4126.org.readthedocs.build/en/4126/

### DIFF
--- a/.github/workflows/rtd.yaml
+++ b/.github/workflows/rtd.yaml
@@ -15,4 +15,4 @@ jobs:
     steps:
     - uses: readthedocs/actions/preview@v1
       with:
-        project-slug: "readthedocs-preview"
+        project-slug: "mdanalysis"


.github/workflows/rtd.yaml
```diff
@@ -15,4 +15,4 @@ jobs:
     steps:
     - uses: readthedocs/actions/preview@v1
       with:
-        project-slug: "readthedocs-preview"
+        project-slug: "mdanalysis"
```
tylerjereddy, line 18: FWIW, in case it isn't obvious, the actual link should be this I think, instead of what we currently see here: https://mdanalysis--4126.org.readthedocs.build/en/4126/

The configuration looks correct here though when I look at the ref config: https://github.com/readthedocs/actions/tree/v1/preview

I searched around and found a project with an example PR/config where it is working: 

https://github.com/ploomber/ploomber/pull/1096

Their config looks fairly similar: https://github.com/ploomber/ploomber/blob/master/.github/workflows/rtd.yml

Do we just need to merge this then to see the change instead of it showing up here? If so, I'd just merge it.

tylerjereddy (approved):

tylerjereddy:
Seems obviously correct based on the other working project, unless there's some other subtle issue, but I've just merged to see if this work now..

PR #4133, opened 2023-04-28
PDB: don't populate formalcharges if bad entry is encountered

### LINKED ISSUE #4027: Formal charge 1 is unrecognized
The charge from some PDB files is not recognised. Testing with PDB files from [CHARMM-GUI](https://charmm-gui.org/?doc=archive&lib=covid19) that have their charges specified in a slightly different format cannot be parsed. 

While I believe the 'official' way to specify charges is `1+` or `1-` (which MDAnalysis reads), these PDB files have either ` 1` or `-1` which gets the error :
```python
ValueError: Formal charge 1 is unrecognized
```

### Minimum file that DOESN'T work:
```
ATOM      1  N   MET A   1     173.150 193.330  29.890  0.00  0.00           N 1
ATOM      2  HT1 MET A   1     172.490 193.030  29.140  0.00  0.00           H-1
```

### Minimum file that DOES work:
```
ATOM      1  N   MET A   1     173.150 193.330  29.890  0.00  0.00           N1+
ATOM      2  HT1 MET A   1     172.490 193.030  29.140  0.00  0.00           H1-
```

This breaks at the parsing stage for `mda.Universe(file)`.

Both of the files can be parsed successfully in version `2.2.0`, but `2.3.0`+ introduced the formal charge parsing for the `.pdb` files which now breaks the import.

I am unsure if this is something that MDAnalysis would formally support, but if not, any suggestions for a fix that I could include in [Molecular Nodes](https://github.com/BradyAJohnston/MolecularNodes) so I can bump the MDAnalysis version from `2.2.0` would be very much appreciated.

IAlibay:
Fixes #4027 

 - We assume here that a bad entry means that the PDB is formatted in an unknown way so we just abstain from interpreting the formal charges completely.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

----
:books: Documentation preview :books:: https://mdanalysis--4133.org.readthedocs.build/en/4133/

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -22,6 +22,8 @@ The rules for this file:
  * 2.5.0
 
 Fixes
+  * PDB topology parser no longer fails when encountering unknown formal
+    charges and instead simply does not populate attribute (Issue #4027)
   * Added support for Cython 3.0.0b2 (PR #4129)
   * Ensure 'HydrogenBondAnalysis' returns the correct distances when using the
     'between' keyword (PR #4092)
--- a/package/MDAnalysis/topology/PDBParser.py
+++ b/package/MDAnalysis/topology/PDBParser.py
@@ -347,28 +347,29 @@ class PDBParser(TopologyReaderBase):
             attrs.append(Elements(np.array(validated_elements, dtype=object)))
 
         if any(formalcharges):
-            for i, entry in enumerate(formalcharges):
-                if not entry == '':
-                    if entry == '0':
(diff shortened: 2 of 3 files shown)


tylerjereddy (approved):
The diff looks sensible to me, CI is passing, and the design decision seems useful to downstream folks based on the discussion in the matching issue.

It would be nice if we could rely on strict adherence to the format, but then there's reality. At least we're still not explicitly supporting the population of attributes via non-standard formatting, and we're still at least providing a warning here.

PR #4153, opened 2023-05-26
Fix future warning from GSDReader, go cutting edge

### LINKED ISSUE #4152: gsd FutureWarning
We're getting some `FutureWarning` from our GSD reader, e.g.:

https://github.com/MDAnalysis/mdanalysis/actions/runs/5086781161/jobs/9141560005#step:9:3296

It looks like there was a deprecation,  see: https://github.com/glotzerlab/gsd/blob/15b0eef7f30f594b6e5937236ad95d5d0e06e1ef/gsd/fl.pyx#L323

richardjgowers:
Fixes #4152

 - 

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

----
:books: Documentation preview :books:: https://mdanalysis--4153.org.readthedocs.build/en/4153/


IAlibay:
What's the regression like here? i.e. what minimum version of gsd will work here?

IAlibay (dismissed):
The minimum numpy failure highlights the issue here, we can't make this change within the current NEP29 window.

IAlibay:
Next NEP29 window is June 23 - let's try this after 2.5 gets released

IAlibay:
So the 'r' writing mode was only introduced in the very latest version of GSD (released like 2 weeks ago). I don't know what the best approach is here - do we force folks to install the latest GSD, or do we leave it for now and accept that we are going to get a futurewarning for some cases? Or we support both the old and new version of gsd and switch between the two modes as necessary.

IAlibay:
@richardjgowers I'mk going to use this PR as the place to test GSD 3.0.1. My view here is that now that it's not a core dependency we can just go cutting edge (if CI works)?

IAlibay:
Of course, NEP29...

orbeckst (approved):
Generally ok with the approach.

If we normally announce version bumps in deps then we should add a note to CHANGELOG but I leave this to you.

orbeckst:
Failures such as 
```
E       AttributeError: module 'gsd' has no attribute '__version__'
```

I am going to revoke my unconditional "approve" until these issues are addressed.

orbeckst (changes requested):
I am cool with gsd > 3.0.0 but the test failures need to be fixed first.

orbeckst (approved):

PR #4160, opened 2023-06-01
NEP29, update to Numpy 1.22[.3]

IAlibay:
We won't be making a release in the next month, and June 23rd is when we go to Numpy 1.22+

Here I'm actually pinning builds to 1.22.3 because it makes things a bit cleaner (there's issues on a couple of archs and windows which makes life just eaiser to pin there). If we think we should pin as much as we can to 1.22.0, I can do that too if necessary.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

----
:books: Documentation preview :books:: https://mdanalysis--4160.org.readthedocs.build/en/4160/


IAlibay:
This mypy error makes absolutely no sense to me - the numpy docs specifically say that input is expected to be array_like..

RMeli:
Odd... From a quick glance at the source code it seems that it requires arrays though? Explicitly converting (if needed) seems to solve the issue:

```diff
diff --git a/package/MDAnalysis/lib/mdamath.py b/package/MDAnalysis/lib/mdamath.py
index 260bf0b87..d89805dca 100644
--- a/package/MDAnalysis/lib/mdamath.py
+++ b/package/MDAnalysis/lib/mdamath.py
@@ -104,7 +104,7 @@ def normal(vec1: npt.ArrayLike, vec2: npt.ArrayLike) -> npt.NDArray:
     .. versionchanged:: 0.11.0
        Moved into lib.mdamath
     """
-    normal: npt.NDArray = np.cross(vec1, vec2)
+    normal: npt.NDArray = np.cross(np.asarray(vec1), np.asarray(vec2))
     n = norm(normal)
     if n == 0.0:
         return normal  # returns [0,0,0] instead of [nan,nan,nan]
@@ -171,7 +171,7 @@ def stp(vec1: npt.ArrayLike, vec2: npt.ArrayLike, vec3: npt.ArrayLike) -> float:
     .. versionchanged:: 0.11.0
        Moved into lib.mdamath
     """
-    return np.dot(vec3, np.cross(vec1, vec2))
+    return np.dot(vec3, np.cross(np.asarray(vec1), np.asarray(vec2)))


 def dihedral(ab: npt.ArrayLike, bc: npt.ArrayLike, cd: npt.ArrayLike) -> float:
```

Proof: https://github.com/RMeli/mdanalysis/pull/3

IAlibay:
@tylerjereddy as the numpy expert here, do you have any ideas why the array_like type check would be falling here?

tylerjereddy:
I'm on vacation with just my phone until Sunday, but is there any code path where the arguments might be object arrays? I think static analysis prohibits that while runtime allows it, and in 1.22 release notes typing improvements are noted. @BvB93 implemented many of the NumPy static analysis improvements so they may know more.

tylerjereddy:
Seems to expect bool type, I wonder why, can't check till Monday though

BvB93:
Looking at the errors this seems to be another case of https://github.com/python/mypy/issues/11347, an upstream bug wherein mypy fails to properly deal with overload ambiguity (due to a statically unknown dtype). There was a PR from ~1.5 years ago to fix the issue, but it seems to have gotten stuck in limbo.

IAlibay:
Thanks so much for looking at this @BvB93 ! 

I'm going to treat this for what it is, and just type ignore. Hopefully that upstream issue will get addressed at some point!

package/CHANGELOG
```diff
@@ -25,6 +25,9 @@ Fixes
 Enhancements
 
 Changes
+  * As per NEP29 tthe minimum supported version has been raised to 1.22. We
```
```suggestion
  * As per NEP29 the minimum supported version has been raised to 1.22. We
```

RMeli (approved):

orbeckst (approved):

PR #4161, opened 2023-06-01
add `ignore_same_residue` kwarg to `InterRDF`

orionarcher:

This adds the ability to include the same residues in both g1 and g2 without blowing up the RDF calculation. Specifically, it adds a `ignore_same_residue` kwarg that excludes pairs of atoms with the same `resindice` from the distance calculation. This should be useful when a user wants to understand the correlation of a species with itself (as I do now).

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?
 - [x] Issue raised/referenced? (N/A)

----
:books: Documentation preview :books:: https://mdanalysis--4161.org.readthedocs.build/en/4161/


package/MDAnalysis/analysis/rdf.py
```diff
@@ -261,6 +265,13 @@ def _single_frame(self):
             mask = np.where(idxA != idxB)[0]
             dist = dist[mask]
 
+        if self.ignore_same_residue:
```
hmacdope: The exclusion_block argument achieves something similar right? I like the idea of a way to easily exclude based on same residue but is there a way to make it more general? eg

```
exclude_same_attr="resindices"
```

or failing that perhaps use the same code path as `_exclusion_block`? You could figure out the exclusion block from the indices possibly. 

What does @orbeckst think?

orionarcher: tbh I am not sure exactly what _exclusion_block is doing, nor when it would be used. I'm confused by the code and the docs. 

Happy to generalize to other properties, my only fear is that it makes it easier to mis-use. We might want to have a short list of allowed options. What else do you think would be worth including?

richardjgowers: It's been a while, but I think `exclusion_block` is used to tile out self contributions from molecules/residues of the same size.  

This then won't work if your residues are differently sized, so this approach here is preferable.

orionarcher: Got it. In that case I can push @hmacdope's general purpose exclusion idea. Exclude is definitely a better verb than ignore.

orionarcher:
Iterated slightly on this. Now users can select "residue", "segment", or "chain" to exclude. It will automatically map to the `resindices`, `segindices`, and `chainID` attributes (though I'm not sure `chainid` is the right option?). I feel this is a bit friendlier to new users than forcing them to understand the intricacies of 'resid` vs `resindices` vs `resname`, and covers the vast majority of use cases. Still happy to change it up if other implementations are preferred.

package/MDAnalysis/analysis/rdf.py
```diff
@@ -229,6 +234,12 @@ def __init__(self,
         self.rdf_settings = {'bins': nbins,
                              'range': range}
         self._exclusion_block = exclusion_block
+        assert exclude_same is None or exclude_same in ['residue', 'segment', 'chain'], (
```
hmacdope: Better to throw an exception than assert

package/MDAnalysis/analysis/rdf.py
```diff
@@ -261,6 +272,13 @@ def _single_frame(self):
             mask = np.where(idxA != idxB)[0]
             dist = dist[mask]
 
+        if self.exclude_same:
+            # Ignore distances between atoms in the same residue
+            resixA = self.g1.resindices[pairs[:, 0]]
```
hmacdope: You need to change the attr hit here as currently only `resindices`?

hmacdope (changes requested):
Quick comments.

package/MDAnalysis/analysis/rdf.py
orionarcher: Good catch

package/MDAnalysis/analysis/rdf.py
orionarcher: Changed this

orionarcher:
Fixed @hmacdope's requested. Unless there are further requests for changes, this is ready to merge.

package/MDAnalysis/analysis/rdf.py
```diff
@@ -261,6 +273,13 @@ def _single_frame(self):
             mask = np.where(idxA != idxB)[0]
             dist = dist[mask]
 
+        if self.exclude_same:
+            # Ignore distances between atoms in the same residue
```
hmacdope:
```suggestion
            # Ignore distances between atoms in the same attribute
```

package/MDAnalysis/analysis/rdf.py
```diff
@@ -261,6 +273,13 @@ def _single_frame(self):
...
+            resixA = getattr(self.g1, self.exclude_same)[pairs[:, 0]]
```
hmacdope: Nitpick but 

```suggestion 
            attr_ix_A = getattr(self.g1, self.exclude_same)[pairs[:, 0]]
```

and same with below as these are no longer resix? I could be  wrong?

package/MDAnalysis/analysis/rdf.py
```diff
@@ -229,6 +234,13 @@ def __init__(self,
         self.rdf_settings = {'bins': nbins,
                              'range': range}
         self._exclusion_block = exclusion_block
+        if exclude_same is not None and exclude_same not in ['residue', 'segment', 'chain']:
+            raise ValueError(
+                "The exclude_same argument to InterRDF must be None, 'residue', 'segment' "
+                "or 'chain'."
+            )
+        name_to_attr = {'residue': 'resindices', 'segment': 'segindices', 'chain': 'chainids'}
```
hmacdope: Does this have to be something else? chainids didnt seem to work for me but could be messing something up.

package/MDAnalysis/analysis/rdf.py
```diff
@@ -229,6 +234,13 @@ def __init__(self,
         self.rdf_settings = {'bins': nbins,
                              'range': range}
         self._exclusion_block = exclusion_block
+        if exclude_same is not None and exclude_same not in ['residue', 'segment', 'chain']:
```
hmacdope, line 237: Can you use `exclusion_block` and `exclude same` and get something that makes sense? If not throw an error. if you try to use both.

testsuite/MDAnalysisTests/analysis/test_rdf.py
```diff
@@ -93,6 +93,12 @@ def test_exclusion(sels):
     rdf = InterRDF(s1, s2, exclusion_block=(1, 2)).run()
     assert rdf.results.count.sum() == 4
 
+def test_ignore_same_residues(sels):
```
hmacdope: Sorry to be a pain but we should test all covered attributes?  ie resid, ....

hmacdope (changes requested):
Few changes and a tiny bit more testing but almost there.

package/CHANGELOG
```diff
@@ -23,6 +23,7 @@ Fixes
     (PR #4163, Issue #4159)
 
 Enhancements
+  * Added a ignore_same_residue kwarg to InterRDF (PR #4161)
```
```suggestion
  * Added `exclude_same` kwarg to InterRDF (PR #4161)
```

package/MDAnalysis/analysis/rdf.py
```diff
@@ -261,6 +273,13 @@ def _single_frame(self):
             mask = np.where(idxA != idxB)[0]
             dist = dist[mask]
 
+        if self.exclude_same:
```
RMeli:
```suggestion
        if self.exclude_same is not None:
```
`get()` can return `None`, maybe an explicit check is better because narrower?

testsuite/MDAnalysisTests/analysis/test_rdf.py
RMeli: And maybe test an unsupported argument as well, to solve the codecov warning above?

orbeckst:
@hmacdope or @RMeli — could one of you assign the issue to yourself, please, just so that it's clear that someone is in charge following up on comments and eventually merging? Thanks.

testsuite/MDAnalysisTests/analysis/test_rdf.py
orionarcher: Could you point me to some appropriate test data for segindices and chainids? In the example I am using the segindices is identical for all residues and chainids aren't present.

package/MDAnalysis/analysis/rdf.py
orionarcher: I think it's actually chainIDs, not chainids, fixed in my most recent commits.

package/MDAnalysis/analysis/rdf.py
orionarcher: Dissallowed this since it is quite confusing even if it technically works

testsuite/MDAnalysisTests/analysis/test_rdf.py
```diff
@@ -94,6 +94,21 @@ def test_exclusion(sels):
     assert rdf.results.count.sum() == 4
 
 
+@pytest.mark.parametrize("attr", ("residue"))
```
RMeli:
```suggestion
@pytest.mark.parametrize("attr", ("residue",))
```
A one-element tuple needs the final `,`. But probably best to check all the attributes as suggested in https://github.com/MDAnalysis/mdanalysis/pull/4161#discussion_r1237891018. @hmacdope do you know of appropriate existing test data for segindices and chainids? Otherwise it should be possible to create something *ad-hoc* within the test.

testsuite/MDAnalysisTests/analysis/test_rdf.py
```diff
@@ -94,6 +94,21 @@ def test_exclusion(sels):
...
+def test_ignore_same_residues(sels, attr):
+    # should see two distances with 4 counts each
+    s1, s2 = sels
+    rdf = InterRDF(s2, s2, exclude_same=attr).run()
+    assert rdf.rdf[0] == 0
+    assert rdf.results.count.sum() == 8
+
+
+def test_ignore_same_residues_fails(sels):
+    # should see two distances with 4 counts each
```
RMeli:
```suggestion
    # should see two distances with 4 counts each
```

testsuite/MDAnalysisTests/analysis/test_rdf.py
```diff
@@ -94,6 +94,21 @@ def test_exclusion(sels):
...
+    s1, s2 = sels
+    with pytest.raises(ValueError):
```
RMeli:
```suggestion
    with pytest.raises(ValueError, match="The exclude_same argument to InterRDF must be"):
```

orionarcher:
I added new tests for segment and chain and added @RMeli cleanup suggestions. Thanks for all the feedback.

testsuite/MDAnalysisTests/analysis/test_rdf.py
```diff
@@ -94,6 +96,24 @@ def test_exclusion(sels):
...
+def test_ignore_same_residues(sels, attr, count):
+    # should see two distances with 4 counts each
+    s1, s2 = sels
+    rdf = InterRDF(s2, s2, exclude_same=attr).run()
+    assert rdf.rdf[0] == 0
+    assert rdf.results.count.sum() == count
+
+
+def test_ignore_same_residues_fails(sels):
+    # should see two distances with 4 counts each
```
RMeli:
```suggestion
```
Sorry, the previous suggestion didn't work as intended.

testsuite/MDAnalysisTests/analysis/test_rdf.py
```diff
@@ -94,6 +96,24 @@ def test_exclusion(sels):
...
+    s1, s2 = sels
+    with pytest.raises(ValueError, match="The exclude_same argument to InterRDF must be"):
+        rdf = InterRDF(s2, s2, exclude_same="unsupported").run()
+
```
RMeli, line 115:
```suggestion

    with pytest.raises(ValueError, match="The exclude_same argument to InterRDF cannot be used with"):
        rdf = InterRDF(s2, s2, exclude_same="unsupported", exclude_block=tuple()).run()
        
        
```
Last nitpick, something like this should get ride of code coverage complaints.

RMeli (commented):
Thanks @orionarcher. A few last nitpicks (sorry), otherwise LGTM.

testsuite/MDAnalysisTests/analysis/test_rdf.py
RMeli: @orionarcher Sorry, this was mainly an (untested) suggestion. `exclude_same` needs to be changed to a valid entry in order not to trigger the previous exception.

orionarcher: Sorry about that, fixed now! Hopefully testing passes.

RMeli:
LGTM, thanks @orionarcher. I'll try to re-run the failed actions before approving, it failed because of time limits: https://dev.azure.com/mdanalysis/mdanalysis/_build/results?buildId=5981&view=logs&j=7a9c7d2c-346e-591b-84bb-5490fd358201

RMeli (approved):
Thanks @orionarcher. @hmacdope would you mind having the last look and merge? Thanks.

hmacdope (approved):
Looks great @orionarcher thanks so much. This is a great contribution and will be super useful to many.

orionarcher:
Yay thanks y'all!

PR #4171, opened 2023-06-16
Fix contourpy install for 32bit azure runner and pin GSD

IAlibay:
1.1.0 of contourpy was released 3 days ago and it's making the 32 bit runner fail (they switched over to meson builds, so no more x86).

We're really pushing it in terms of how many bandaids we can put on to test 32bit support.... @MDAnalysis/coredevs we need to start discussing if we want to drop 32bit ahead of the 3.0 release.

 - Hacky fix for contourpy insttall for x86
 - Max pin GSD at <3.0 - issues with pytest-cov and pytest-xdist which may not get resolved upstream

PR Checklist
------------
 - [x] Tests?

----
:books: Documentation preview :books:: https://mdanalysis--4171.org.readthedocs.build/en/4171/


orbeckst:
Do we know how much 32 bit is actually being used?

What is the position of numpy/scipy on 32 bit support, @tylerjereddy ?

Ending support for 32 bit for 3.0 sounds sensible. I think one thing we should discuss at our next meeting is a tentative roadmap.

orbeckst:
Also, where do we use contourpy? Is that a secondary dependency?

IAlibay:
> Also, where do we use contourpy? Is that a secondary dependency?

It's a dependency for matplotlib.

IAlibay:
🤦🏽 of course.. of course GSD releases a new version when I'm trying to debug something else

tylerjereddy:
SciPy has stopped providing 32-bit Windows wheels on PyPI (we also switched to `meson`, facing the limitation Irfan alludes to above). We do still have other routes to get 32-bit Win binaries I think because of https://www.lfd.uci.edu/~gohlke/pythonlibs/#scipy and maybe `conda` (which has a second release manager), but my overall impression is that 32-bit on both Linux and Windows is getting pretty heavily de-prioritized apart from maybe a CI job for the Linux one (and some desire to have 32-bit Windows CI job, but I don't think we can do it yet...).

There's some discussion here:
https://github.com/scipy/scipy/issues/15836
https://github.com/scipy/scipy/issues/16286

So, current SciPy stable releases have neither Windows nor Linux 32-bit binaries. Current NumPy stable release has no 32-bit binaries for Linux, but does have 32-bit binaries for Windows because they haven't swapped to `meson` 100 % yet (it is planned for a point release after the current stable release in a few weeks). I think NumPy couldn't switch to `meson` as fast as SciPy because they have tons of SIMD stuff, special OpenBLAS 64-bit integer requirements, and some other stuff requiring upstream `meson` improvements.

Anyway, I guess that's a long way of saying that it seems like support has decreased for 32-bit binaries, though we do still at least try to test in CI when it is possible, but the priority overall seems to have dropped a bit I'd say (I think Debian and/or other Linux distros dropping official 32-bit support also forced NumPy's hand a bit on Linux, which then forced our hand too I think..).

hmacdope:
Might just be time to banish 32 but ASAP.

IAlibay:
@MDAnalysis/coredevs this is good to go - it'll unblock CI for now.

richardjgowers (approved):

Issue #4172, opened 2023-06-16
GSD 3.0 fails w/ pytest-xdist/cov


IAlibay:
See: https://dev.azure.com/mdanalysis/mdanalysis/_build/results?buildId=5846&view=logs&j=20ededa3-d725-5c28-882d-f9e714f63884&t=ee6fadee-24b3-58c2-f7a0-e9750bae5c01

IAlibay:
Upstream discussion: https://github.com/glotzerlab/gsd/issues/255#issuecomment-1595031852

IAlibay:
Only reason the Windows and MacOS runners are affected right now is because the conda-forge release isn't up yet...

joaander:
GSD 3.0.1 fixes this error. In my testing on https://github.com/glotzerlab/gsd/issues/257 you can also work around this by importing gsd (`import gsd`) in the main thread prior to launching child threads.

IAlibay:
Fixed, gsd is now an optional dependency and 3.0.1+ only.

PR #4174, opened 2023-06-17
Make GSD optional

### LINKED ISSUE #3819: Make GSD an optional requirement
As per other file format specific dependencies, it would make more sense for GSD to not be a base dependency.

IAlibay:
Fixes #3819 

 - Make GSD an optional requirement

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

----
:books: Documentation preview :books:: https://mdanalysis--4174.org.readthedocs.build/en/4174/

### DIFF
--- a/.github/actions/setup-deps/action.yaml
+++ b/.github/actions/setup-deps/action.yaml
@@ -107,7 +107,6 @@ runs:
           ${{ inputs.cython }}
           ${{ inputs.fasteners }}
           ${{ inputs.griddataformats }}
-          ${{ inputs.gsd }}
           ${{ inputs.hypothesis }}
           ${{ inputs.matplotlib }}
           ${{ inputs.mmtf-python }}
@@ -123,6 +122,7 @@ runs:
           ${{ inputs.chemfiles-python }}
           ${{ inputs.clustalw }}
           ${{ inputs.distopia }}
+          ${{ inputst.gsd }}
           ${{ inputs.h5py }}
           ${{ inputs.hole2 }}
           ${{ inputs.joblib }}
--- a/.github/actions/setup-deps/action.yaml
+++ b/.github/actions/setup-deps/action.yaml
@@ -122,7 +122,7 @@ runs:
           ${{ inputs.chemfiles-python }}
           ${{ inputs.clustalw }}
           ${{ inputs.distopia }}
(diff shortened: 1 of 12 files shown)


IAlibay:
Opening this early so I can get some feedback from @MDAnalysis/coredevs - this is probably quite an abrupt change, do we need to have some kind of deprecation window?

richardjgowers (approved):

RMeli:
I don't think so IMO. It might surprise some users, but the error message is very clear and the solution a pip install away.

IAlibay:
Ok this should be good to go I think (unless I forgot to fix some test somewhere). Since it's a core dependency change, I'm going to leave this here for a week for other @MDAnalysis/coredevs to chime in, and if there are no objections it'll get merged.

Specifically pinging @orbeckst here, since we've spent a lot of time pondering in the past on how best to do these types of changes.

RMeli (approved):

hmacdope (approved):
LGTM thanks @IAlibay

orbeckst:
For reference: conda-forge feedstock was already updated https://github.com/conda-forge/mdanalysis-feedstock/pull/57

orbeckst (approved):
I agree.

Eventually it would be good to also allow gsd > 3.0 but we have to be able to test it. In the short term, the restriction is sensible.

orbeckst:
@IAlibay I'll let you do the merging (unless you want someone else to do it, then change the assignee). Thanks for working out the problem and the quick-fix.

joaander:
> I agree.
> 
> Eventually it would be good to also allow gsd > 3.0 but we have to be able to test it. In the short term, the restriction is sensible.

Please test GSD 3.0.1.

orbeckst:
@joaander thanks, we'll check that #4172 is solved with gsd 3.0.1. Will just take a few days (people on holidays etc).

If it works we could blacklist 3.0.0 and otherwise allow `gsd!=3.0.0`. We'll probably still make it optional as we have been trying to slim down dependencies.

joaander:
@orbeckst understood. If 3.0.1 passes your tests, I expect 3.0.0 will work for your users. I find that most users follow PEP8 guidelines and issue imports at the top of the module, not inside child threads.

Issue #2980, opened 2020-10-13
Tests are generating too many warnings


jbarnoud:
## Expected behavior ##

Running the tests issues no or only a few warnings that can be displayed and, ideally, addressed.

## Actual behavior ##

Running the tests issues more than **30,000 warnings** that are untracktable to look at, meaning we can very easily miss important warnings.

## Code to reproduce the behavior ##

Just run the tests

## Current version of MDAnalysis ##

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`)  2.0.0-dev0
- Which version of Python (`python -V`)? Python 3.9.0
- Which operating system? Ubuntu 20.04 on WSL

orbeckst:
@tylerjereddy has been stalwartly working on reducing the warnings.

Do we have a statistic of which ones are the most frequent warnings so that we can make a targeted attempt to address the worst offenders first, e.g., anything from the PDBParser regarding atom types not found? We can then either address the root problems or include warnings filters to mute any warnings during a test that we know are harmless.

richardjgowers:
Would be nice if warnings meant something in tests.  One idea is to change the common warnings we emit (i.e. `MDAnalysis.GuessingWarning`) then have a pytest fixture applied to all tests that silences these warnings.  I think this might address ~80% at least

jbarnoud:
I just had a very quick look and it seems that there are limitations to what types of warning you can ignore globally. From what I understand, there is a fixed list of warning types that can be ignored. For the other types, you would have to capture them on a per-test basis, which would probably be the cleanest but would not be tractable in some cases.

Another option is to ignore individual warnings by string. This would go in a config file, so it is doable.

Finally, it seems that more than half of the warnings are not about guessing, but about invalid values in numpy. By ignoring the `RuntimeWarnings` I fall from ~30,000 to ~13,000 warnings.

orbeckst:
Is there anything here that we could let prospective GSoC students do? 

Ignoring numpy RuntimeWarnings would be a start.

IAlibay:
So as may have been obvious from all the recent issues, I'm currently looking at the deprecation warnings so we don't lag too behind for the next release.

I've found that I can reduce the number of emitted warnings to a more manageable 382 with the `setup.cfg` contents shown below.

The first ~ 3 `RuntimeWarnings` (`double_scalars`, `true_divide`, and `sqrt`) account for ~ 22k warnings being emitted.
[...]

PR #4175, opened 2023-06-18
MAINT, TST: suppress np warns

tylerjereddy:
* while checking our compatibility with just-released NumPy `1.25.0` (seems "ok" on M2 mac), I went ahead and suppressed about 17 % of our test suite warnings (5384 / 32502) with a few NumPy floating point error handlers

* related to gh-2980

* I doubt there's any harm in doing this--if we have a real algorithm issue it should be enforced by our testsuite anyway, rather than relying on fp warnings

----
:books: Documentation preview :books:: https://mdanalysis--4175.org.readthedocs.build/en/4175/


hmacdope (approved):
Seems reasonable to me! Thanks @tylerjereddy :)

IAlibay:
Thanks @tylerjereddy !

PR #4176, opened 2023-06-23
Drop 32bit from azure CI

### LINKED ISSUE #3858: Deprecating windows 32bit support

IAlibay:
Fixes #3858 

Am I correct in u ndrestanding this is a thing now?

----
:books: Documentation preview :books:: https://mdanalysis--4176.org.readthedocs.build/en/4176/

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -21,12 +21,6 @@ jobs:
     MPLBACKEND: agg
   strategy:
     matrix:
-        Win-Python39-32bit-full:
-          PYTHON_VERSION: '3.9'
-          PYTHON_ARCH: 'x86'
-          BUILD_TYPE: 'normal'
-          NUMPY_MIN: '1.21.0'
-          imageName: 'windows-2019'
         Win-Python39-64bit-full:
           PYTHON_VERSION: '3.9'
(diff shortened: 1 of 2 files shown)


RMeli (approved):
🪓

orbeckst (changes requested):
need a note in CHANGELOG 🪓

orbeckst (approved):

PR #4198, opened 2023-07-14
Build wheels with numpy 1.25+

### LINKED ISSUE #4183: [CI/CD] Switch wheel deployment pins to 1.25+
Following up from @tylerjereddy's comment on discord regarding numpy 1.25 and backwards compatibility (see: https://numpy.org/doc/1.25/dev/depending_on_numpy.html#adding-a-dependency-on-numpy). We should be able to switch our pins out to >1.25 for wheel building.

This will require:
- [ ] Updating the necessary wheel building entry points (both standard building + deployment pipelines)
- [ ] Adding necessary roundtrip tests to ensure it's working properly (we can probably just swap around some of the azure pipelines builds).
- [ ] Updating our conda-forge recipee as necessary.

IAlibay:
Fixes #4183 

This will likely (silently) break all of CI since the order of things changes.

Things to check
 - [x] azure pipelines
 - [x] gh actions
 - [x] cirrus (all good we don't have a specific numpy install) 
 - [x] deployment

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

----
:books: Documentation preview :books:: https://mdanalysis--4198.org.readthedocs.build/en/4198/


hmacdope (approved):
Looks good to me @IAlibay

IAlibay:
> Looks good to me @IAlibay 

I mean it still needs a lot of changes / work.. I wouldn't prematurely approve it.

orbeckst (commented):
I don't feel knowledgeable to approve here — all I would do is look if the CI is still green. I assume that NEP29 allows us to depend on `numpy >= 1.25`?

IAlibay:
> I don't feel knowledgeable to approve here — all I would do is look if the CI is still green. I assume that NEP29 allows us to depend on `numpy >= 1.25`?

numpy 1.25 offers backwards compatibility all the way through to numpy 1.19 (which more than covers nep29).

What's left to do here in this PR is to go back and rework CI to do the right install order for wheel generation where necessary - i.e. install with 1.25 and then downgrade to 1.22 and see if it still works. I'll have a stab at it later this week probably.

IAlibay:
Cron CI issue is elsewhere see #4238

.github/actions/build-src/action.yaml
```diff
@@ -9,6 +9,10 @@ inputs:
     description: 'build MDA docs'
     required: true
     default: false
+  isolation:
```
IAlibay, line 12: We're adding an option to trigger the pip --no-build-isolation flag. This will allow us to avoid `pip` trying to have its own isolated build & overidding already installed dependencies.

.github/workflows/deploy.yaml
```diff
@@ -241,18 +241,18 @@ jobs:
...
         run: |
           awk '/__version__ =/ {print $3; exit}' package/MDAnalysis/version.py | tr -d \" > version.dat
           ver=$(python maintainer/norm_version.py --file version.dat)
           pip install --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple MDAnalysis==$ver
           pip install --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple MDAnalysisTests==$ver
 
+      - name: install_min_deps
+        if: "matrix.type == 'MIN'"
+        run: |
+          pip install pytest-xdist pytest-timeout
```
IAlibay, line 254: Moving this here because pip installing MDAnalysisTests should have already gotten pytest, you just need ot add pytest-xdist and pytest-timeout

.github/workflows/gh-ci-cron.yaml
```diff
@@ -3,6 +3,9 @@ on:
   schedule:
     # 3 am Tuesdays and Fridays
     - cron: "0 3 * * 2,5"
+  pull_request:
```
IAlibay: remove before merge

.github/workflows/gh-ci-cron.yaml
```diff
@@ -69,6 +72,7 @@ jobs:
       with:
         build-tests: true
         build-docs: false
+        isolation: false
```
IAlibay, line 72: turn off isolation here because we want to build with the nightly wheels

.github/workflows/gh-ci-cron.yaml
```diff
@@ -112,11 +116,13 @@ jobs:
       run: |
         sed -i "s/#extra_cflags =/extra_cflags = -march=native -mtune=native/g" package/setup.cfg
         cat package/setup.cfg
+
     - name: build_srcs
       uses: ./.github/actions/build-src
       with:
         build-tests: true
         build-docs: false
+        isolation: false
```
IAlibay, line 122: build with the deps you have installed

.github/workflows/gh-ci-cron.yaml
```diff
@@ -164,6 +170,7 @@ jobs:
...
```
IAlibay, line 170: As above

.github/workflows/gh-ci.yaml
```diff
@@ -88,6 +88,12 @@ jobs:
       with:
         build-tests: true
         build-docs: false
+        isolation: ${{ matrix.numpy != '' }}
```
IAlibay, line 94: if we define an old numpy value then we isolate (that way we can build with the correct numpy version and then test with a runtime version of numpy that's much older)

.github/workflows/gh-ci.yaml
```diff
@@ -152,6 +158,7 @@ jobs:
       with:
         build-tests: true
         build-docs: true
+        isolation: false
```
IAlibay, line 164: no need for isolation if we have all the dependencies here

azure-pipelines.yml
```diff
@@ -120,14 +120,12 @@ jobs:
...
     displayName: 'List of installed dependencies'
   - powershell: |
-      cd package
-      python setup.py install
-      cd ..
-      cd testsuite
-      python setup.py install
-      cd ..
+      python -m pip install ./package
+      python -m pip install ./testsuite
```
IAlibay, line 124: switch over to pip install for modern-ness

azure-pipelines.yml
```diff
@@ -120,14 +120,12 @@ jobs:
...
     displayName: 'Build MDAnalysis'
     condition: and(succeeded(), eq(variables['BUILD_TYPE'], 'normal'))
+  - powershell: pip list
+    displayName: 'Check installed packages'
```
IAlibay, line 128: Keeping this around so we can get an idea of what's in the environment ahead of testing

package/pyproject.toml
```diff
@@ -3,20 +3,11 @@
...
-    "numpy==1.22.3; python_version=='3.10' and platform_python_implementation != 'PyPy'",
-    "numpy==1.23.2; python_version=='3.11' and platform_python_implementation != 'PyPy'",
-    # For unreleased versions of Python there is currently no known supported
-    # NumPy version. In that case we just let it be a bare NumPy install
-    "numpy<2.0; python_version>='3.12'",
+    # NumPy 1.25+ offers backwards compatibility that tracks with greater
+    # than NEP29 (see: https://numpy.org/doc/stable/dev/depending_on_numpy.html#adding-a-dependency-on-numpy)
+    # We pin our wheel installation to that and a maximum of numpy 2.0 since
+    # this will likely involve several breaking changes
+    "numpy>=1.25,<2.0; python_version>='3.9'",
```
IAlibay, line 10: NumPy 1.25+ is backwards compatible with its C/C++ API, so (as shown in the CI tests) we can build with it and then use an older version of NumPy at runtime and things just work.

IAlibay (commented):
Some comments for when folks get to review so it makes more sense what I'm doing here.

IAlibay:
Alright can I ask for a priority re-review & merge here @MDAnalysis/coredevs ? I would like to get this merged to get the 2.6.0 release out.

.github/workflows/gh-ci.yaml
lilyminium: Can you please comment this in the file for later perusal?

.github/actions/build-src/action.yaml
lilyminium: It took me a couple readings of this comment and the pip documentation to understand (if I have this correctly) that:
- default (i.e. isolate) installs new build-time dependencies that can be separate from runtime dependencies or previously installed packages
- `--no-build-isolation` does not override previously installed packages (I think?) but makes use of what is currently installed, which is why we added this flag here to test different build package versions

If that's right, could you please add that to the description of this flag, so later maintainers understand the use of the flag in our CI?

lilyminium (approved):
LGTM but IMO more documentation would be a huge help for future maintenance. Thanks for this @IAlibay

.github/actions/build-src/action.yaml
IAlibay: Honestly I'm not really 100% sure I know exactly how build isolation works, my take is that it creates an isolated environment with brand new everything to create wheels and then it gives you the wheels back in the world you first started in.

I'll try to add something sensible here.

IAlibay: Ok I've added a longer explanation and justification for how build isolation works, why we need it, etc...

I'm going to go ahead with the merge, but if you have the time to review what I wrote here and call out anything that's unclear in an issue that'd be great.

IAlibay:
Thanks for the quick review @lilyminium ! I'll go ahead and merge when CI returns green so we can move ahead with the release.

PR #4210, opened 2023-07-25
[CI] Add post deployment conda and pypi cron tests

### LINKED ISSUE #4208: [CI] conda & pypi post deployment tests
We don't have CI infrastructure in place to regularly check that our deployments actually work post-release.

Generally we tend to pick stuff up from the `develop` branch and then push it back as a bugfix, but we probably should also regularly check that things still work / we haven't forgotten to push a bugfix.

It's a simple enough fix, we just need to add stuff to our cron runner.

IAlibay:
Fixes #4208 

 - 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

----
:books: Documentation preview :books:: https://mdanalysis--4210.org.readthedocs.build/en/4210/

### DIFF
--- a/.github/workflows/gh-ci-cron.yaml
+++ b/.github/workflows/gh-ci-cron.yaml
@@ -3,9 +3,6 @@ on:
   schedule:
     # 3 am Tuesdays and Fridays
     - cron: "0 3 * * 2,5"
-  pull_request:
-    branches:
-      - develop
 
 concurrency:
   # Probably overly cautious group naming.
--- a/.github/workflows/gh-ci-cron.yaml
+++ b/.github/workflows/gh-ci-cron.yaml
@@ -258,9 +258,11 @@ jobs:
         python-version: ${{ matrix.python-version }}
 
     - name: install_mdanalysis
+      shell: bash
       run: |
           pip install mdanalysis mdanalysistests pytest-xdist
 
     - name: run_tests
+      shell: bash
       run: |
           pytest -n auto --pyargs MDAnalysisTests
--- a/.github/workflows/gh-ci-cron.yaml
+++ b/.github/workflows/gh-ci-cron.yaml
@@ -260,7 +260,7 @@ jobs:
     - name: install_mdanalysis
       shell: bash
       run: |
-          pip install mdanalysis mdanalysistests pytest-xdist
+          pip install mdanalysis mdanalysistests pytest-xdist "gsd<3.0"
 
     - name: run_tests
       shell: bash
--- a/.github/workflows/gh-ci-cron.yaml
+++ b/.github/workflows/gh-ci-cron.yaml
@@ -204,11 +204,12 @@ jobs:
   conda-latest-release:
     # A set of runner to check that the latest conda release works as expected
     if: "github.repository == 'MDAnalysis/mdanalysis'"
-    runs-on: [ubuntu-latest, macos-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
     timeout-minutes: 60
     strategy:
         fail-fast: false
         matrix:
+          os: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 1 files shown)


.github/workflows/gh-ci-cron.yaml
```diff
@@ -196,3 +199,70 @@ jobs:
...
+          channels:
+            - jaimergp/label/unsupported-cudatoolkit-shim
+            - conda-forge
+            - bioconda
+
+    - name: install_deps
+      uses: ./.github/actions/setup-deps
+      with:
+        micromamba: true
+        full-deps: true
```
IAlibay, line 232: note to awake self - this doesn't work because this action already installs GSD which then clashes with the <3.0 requirement

IAlibay:
Ok current CI test failuires are down to known issues so I won't fix them here:
  - numpy_and_scipy_dev will be fixed by #4203 
  - macos conda release checks are broken because of #4211 

@MDAnalysis/coredevs if I can get a review / merge here we can move forward towards fixing #4211

.github/workflows/gh-ci-cron.yaml
```diff
@@ -196,3 +196,71 @@ jobs:
...
+    if: "github.repository == 'MDAnalysis/mdanalysis'"
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 60
+    strategy:
+        fail-fast: false
+        matrix:
+          os: [ubuntu-latest, macos-latest]
+          python-version: ["3.9", "3.10", "3.11"]
+    steps:
+    - uses: actions/checkout@v3
```
orbeckst, line 212: Do we need the checkout action? We don’t use the repo, do we?

.github/workflows/gh-ci-cron.yaml
```diff
@@ -196,3 +196,71 @@ jobs:
...
+  # Issue 4208
+  conda-latest-release:
+    # A set of runner to check that the latest conda release works as expected
+    if: "github.repository == 'MDAnalysis/mdanalysis'"
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 60
+    strategy:
+        fail-fast: false
+        matrix:
+          os: [ubuntu-latest, macos-latest]
```
orbeckst, line 209: Why do we not run the conda based installation on windows? (Just out of curiosity.)

orbeckst (approved):
This looks good as far as I can tell. I don’t know how you’ve tested that the cron job ran but I trust your report. 

Thank you for adding these tests!

.github/workflows/gh-ci-cron.yaml
IAlibay: Yeah we need it to pick up our custom actions (setup-deps in this case).

.github/workflows/gh-ci-cron.yaml
IAlibay: Because the conda check if a "full dependency" check it installs things that aren't windows compatible (TNG, etc...). 

My thought here is that we could have a Windows runner that just doesn't do a full dependency install, but that essentially amounts to the pip check? Maybe I'm missing something important though.

IAlibay:
Thanks! Cron jobs were just enabled to run on this PR until ~ 2 commits ago.

.github/workflows/gh-ci-cron.yaml
orbeckst: ok!

.github/workflows/gh-ci-cron.yaml
orbeckst: Thanks for the explanation. Does not seem to be necessary to do more right now — the pip run is functionally equivalent.

orbeckst:
Please merge at your leisure.

Issue #4211, opened 2023-07-26
[2.5.0] Center of charge test failures on macOS runners


IAlibay:
## Expected behavior ##

All tests on our latest production release should pass across expected machine types (especially ubuntu, and macos).

## Actual behavior ##

As reported by @orbeckst [here](https://groups.google.com/g/mdnalysis-devel/c/jTgO2eXmMKg/m/NifnvlsxBwAJ?utm_medium=email&utm_source=footer) (and reproduced in #4210), tests for CoC unwrapping are failing.

This isn't seen on develop.

## Code to reproduce the behavior ##

Run tests on the 2.5.0 on an intel macOS machine

orbeckst:
Do you even have an idea what could cause the issue?

Is it specific to Intel Mac??

IAlibay:
Yeah Intel Mac only. It's odd because the tests haven't changed, which makes me think that the base numpy version we use to cythonize might be to blame here. I want to check that once the new CI checks are in place.

IAlibay:
Actually I don't know if it's specific to x86_64 intel - can _anyone_ with a macos-arm64 confirm that this happens if they pull from conda-forge?

Note: this is only happening on conda-forge. I'm currently thinking it might be down to clang 15.

IAlibay:
I will add here that the osx-arm64 builds on the feedstock aren't facing the issue, so maybe it is just x86_64: https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=753485&view=logs&j=9c5ef928-2cd6-52e5-dbe6-9d173a7d951b&t=c66c4975-bbef-551b-2ee6-85005a231b12

IAlibay:
tagging @RMeli who has both an intel and arm macos machine

IAlibay:
Where I'm at on all this:

1. Conda-forge builds have the above failure, but not PyPi ones
2. Building from PyPi source is also fine
3. All other platforms but x86_64 macos seem fine
4. The relevant code in 2.6.0 does not seem to have changed.
5. The previous conda-forge builds of 2.5.0 also fail (so it's unlikely to be something related to building with Cython 3)
6. Building with Cython 3 using PyPi does not seem to cause any issues
7. Using clang 14 (as is used by default on gh actions) instead of clang 15 (conda-forge default) does not change the issue

IAlibay:
[...]

PR #4212, opened 2023-07-28
[WIP] investigate Issue #4211

IAlibay:
I'm checking if Cython 3.0 is the cause of all this.

Thought here is that pypi isn't failing but conda is - both having the same numpy build pin, the main thing that has changed is Cython though. It is very likely that the build version of NumPy that we do use just isn't compatible with Cython 3.0 or something along those lines.

 - 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

----
:books: Documentation preview :books:: https://mdanalysis--4212.org.readthedocs.build/en/4212/

### DIFF
--- a/.github/workflows/gh-ci-cron.yaml
+++ b/.github/workflows/gh-ci-cron.yaml
@@ -240,6 +240,11 @@ jobs:
       run: |
           micromamba install mdanalysis mdanalysistests
 
+    - name: list_environment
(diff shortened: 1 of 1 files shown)


IAlibay:
I have no futher options here, someone else take over fixing this please.

Issue #4209, opened 2023-07-25
[CI] Tests timing out occasionally


IAlibay:
For some reason some of the tests are every so often timing out. It's been mostly the ubuntu full deps py3.10 gh actions tests, but we've seen it in the macos and azure runners too.

It's rather hard to debug and I'm currenttly cycling CI with pytest-timeout in #4197

IAlibay:
From a look at the top 50 slowest, there seems to be some kind of systematic increase in test run times in the failing runners. 

Note: test_streamplot_2D is > 20s even on a good day!

IAlibay:
@MDAnalysis/coredevs this is actually a pretty big issue, if anyone has any ideas here that'd be appreciated.

IAlibay:
My initial suspicion is that this might be happening: https://github.com/pytest-dev/pytest/issues/11174

Need to gather a bit more evidence though (which annoyingly means re-running the tests until we can gather more random failures).

A potential short term solution is to just stick on pytest-timeout, but it really only shifts the problem from a job not finishing to one failing (so really it counts as a failure both ways and CI remains red).

IAlibay:
Unfortunately I've been unable to narrow down where this is happening. From runs on https://github.com/MDAnalysis/mdanalysis/pull/4197 my current conclusions are:

1. The frequency at which these stalling runs occus is somewhat low. It affects nearly every PR at some point, but we're looking at maybe a ~ 1 / 20 chance?
2. There are no specific tests which time out (once it was a multiprocessing one, another was nsgrid..)
3. It is not isolated to github actions (we saw it happen on azure pipelines).

My current completely random guess is that it's something to do with the latest 7.4.0 release and _maybe_ coverage? (I haven't seen it happen on runner not trying to use pytest-cov functionality).

I've ran out of ideas for now and I don't think I have the time to debug any furtther. So I'm going to propose the following:

1. We add [`pytest-timeout`](https://github.com/pytest-dev/pytest-timeout) to our CI runs. At least runners will complete even though they will fail. That will save a few trees and allow us to just ignore timeout issues if we don't think they are relevant.
2. We keep this issue ope and re-asses over time.

hmacdope:
Sounds good to me @IAlibay, thanks for putting in a great effort  trying to figure it out.

RMeli:
Sounds very sensible. Thanks for all the work!

drew-parsons:
[...]

PR #4215, opened 2023-07-30
Add pytest-timeout to CI

IAlibay:
Temporary patch for #4209 

 - Add pytest-timeout to gh actions, azure and Cirrus 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

----
:books: Documentation preview :books:: https://mdanalysis--4215.org.readthedocs.build/en/4215/

### DIFF
--- a/.github/actions/setup-deps/action.yaml
+++ b/.github/actions/setup-deps/action.yaml
@@ -83,6 +83,8 @@ inputs:
     default: 'coverage'
   pytest-cov:
     default: 'pytest-cov'
+  pytest-timeout:
+    default: 'pytest-timeout'
   pytest-xdist:
     default: 'pytest-xdist'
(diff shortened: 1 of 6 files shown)


tylerjereddy (approved):
LGTM, and FWIW SciPy also has `pytest-timeout` in some of our CI runs and `pyproject.toml` optional deps. Looks like threshold is `60` s over there.

We should probably check this more closely as noted in the matching issue, but core devs seem in favor of this temporary solution over there in any case.

`test_streamplot_2D` mentioned over there is probably for me to investigate, I wrote that with Matthieu.

The only failures I saw in this PR when CI last flushed were timeouts at `> 100 s`, which is actually pretty wild for a single test!

PR #4221, opened 2023-08-01
TST: test_streamplot_2D faster

tylerjereddy:
* speed up `test_streamplot_2D()` because it has been reported to take ~20 seconds on a regular basis
in CI in gh-4209

* we don't really need to plot the output, which was taking most of the time, and instead we can just check the data structures that MDAnalysis returns (this may be a better test by some definitions anyway...); I suppose we could also spot check a few values in the arrays if we wanted as well

* locally, that single test seems to run in 0.39 s on this branch vs. 4.7 s on `develop`

----
:books: Documentation preview :books:: https://mdanalysis--4221.org.readthedocs.build/en/4221/


testsuite/MDAnalysisTests/visualization/test_streamlines.py
```diff
@@ -67,22 +67,10 @@ def test_streamplot_2D(membrane_xtc, univ, tmpdir):
...
-    ax = fig.add_subplot(111, aspect='equal')
-    ax.streamplot(x, y, u1, v1, density=(10,10), color=speed, linewidth=3*speed/speed.max(),
-                  cmap=plt.cm.viridis)
-    plot_outpath = os.path.join(str(tmpdir), 'testing_streamline.png')
-    fig.savefig(plot_outpath, dpi=300)
-
-    with open(plot_outpath, 'rb'):
-        pass
+    assert u1.shape == (5, 5)
+    assert v1.shape == (5, 5)
```
tylerjereddy: maybe spot checking values here might make sense--to be fair, we weren't doing that before either though

testsuite/MDAnalysisTests/visualization/test_streamlines.py
```diff
@@ -67,22 +67,10 @@ def test_streamplot_2D(membrane_xtc, univ, tmpdir):
...
+    assert_allclose(avg, 0.965194167)
```
IAlibay: Are these single value checks? If so could we use pytest.approx instead (it is much faster last I checked)?

IAlibay (changes requested):
Thanks! If you want to add a few value spot checks that'd be great, otherwise it at least doesn't regress the amount of testing we do.

Just the one think about using pytest.approx for single float value checks.

tylerjereddy:
I revised to use `pytest.approx` for the single value checks (didn't see a perf change either way), and use `assert_allclose` to just check all the values in `u1`, `v1`, since they're fairly small anyway (still no perf change for me locally).

IAlibay (approved):
Thanks!

PR #4231, opened 2023-08-10
fix error message for XDR readers

orbeckst:
Fixes #

- error message for not being able to write lockfile did not insert the filename into message: fixed by using proper f string
- changed all .format to modern f string in XDR.py
- add explicit test for offsets_filename()
- add additional pytest.warns() to tests to catch warnings (reduces warnings output and checks that all expected warnings are raised with proper messages)

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

----
:books: Documentation preview :books:: https://mdanalysis--4231.org.readthedocs.build/en/4231/

### DIFF
--- a/package/MDAnalysis/coordinates/XDR.py
+++ b/package/MDAnalysis/coordinates/XDR.py
@@ -63,8 +63,7 @@ def offsets_filename(filename, ending='npz'):
 
     """
     head, tail = split(filename)
-    return join(head, '.{tail}_offsets.{ending}'.format(tail=tail,
-                                                        ending=ending))
+    return join(head, f'.{tail}_offsets.{ending}')
 
 
 def read_numpy_offsets(filename):
@@ -88,7 +87,7 @@ def read_numpy_offsets(filename):
 
     #  `ValueError` is encountered when the offset file is corrupted.
     except (ValueError, IOError):
-        warnings.warn("Failed to load offsets file {}\n".format(filename))
+        warnings.warn(f"Failed to load offsets file {filename}\n")
         return False
 
 class XDRBaseReader(base.ReaderBase):
@@ -201,7 +200,7 @@ class XDRBaseReader(base.ReaderBase):
         except OSError as e:
             if isinstance(e, PermissionError) or e.errno == errno.EROFS:
                 warnings.warn(f"Cannot write lock/offset file in same location as "
-                               "{self.filename}. Using slow offset calculation.")
+                              f"{self.filename}. Using slow offset calculation.")
                 self._read_offsets(store=True)
                 return
             else:
@@ -220,10 +219,10 @@ class XDRBaseReader(base.ReaderBase):
             # refer to Issue #1893
             data = read_numpy_offsets(fname)
             if not data:
-                warnings.warn("Reading offsets from {} failed, "
-                              "reading offsets from trajectory instead.\n"
-                              "Consider setting 'refresh_offsets=True' "
-                              "when loading your Universe.".format(fname))
+                warnings.warn(f"Reading offsets from {fname} failed, "
+                              f"reading offsets from trajectory instead.\n"
+                              f"Consider setting 'refresh_offsets=True' "
+                              f"when loading your Universe.")
                 self._read_offsets(store=True)
                 return
 
@@ -256,7 +255,7 @@ class XDRBaseReader(base.ReaderBase):
                          offsets=offsets, size=size, ctime=ctime,
                          n_atoms=self._xdr.n_atoms)
             except Exception as e:
-                warnings.warn("Couldn't save offsets because: {}".format(e))
+                warnings.warn(f"Couldn't save offsets because: {e}")
 
     @property
     def n_frames(self):
--- a/package/MDAnalysis/coordinates/XDR.py
+++ b/package/MDAnalysis/coordinates/XDR.py
@@ -220,9 +220,9 @@ class XDRBaseReader(base.ReaderBase):
             data = read_numpy_offsets(fname)
             if not data:
                 warnings.warn(f"Reading offsets from {fname} failed, "
-                              f"reading offsets from trajectory instead.\n"
-                              f"Consider setting 'refresh_offsets=True' "
-                              f"when loading your Universe.")
+                              "reading offsets from trajectory instead.\n"
+                              "Consider setting 'refresh_offsets=True' "
+                              "when loading your Universe.")
                 self._read_offsets(store=True)
                 return
 
--- a/testsuite/MDAnalysisTests/coordinates/test_xdr.py
+++ b/testsuite/MDAnalysisTests/coordinates/test_xdr.py
@@ -553,11 +553,9 @@ class _GromacsWriterIssue117(object):
 
     @pytest.fixture()
     def universe(self):
-        with pytest.warns(UserWarning,
-                          match="ATOMIC_NUMBER record not found"):
-            u = mda.Universe(PRMncdf, NCDF)
-        return u
+        return mda.Universe(PRMncdf, NCDF)
 
+    @pytest.mark.filterwarnings("ignore: ATOMIC_NUMBER record not found")
     def test_write_trajectory(self, universe, tmpdir):
         """Test writing Gromacs trajectories from AMBER NCDF (Issue 117)"""
         outfile = str(tmpdir.join('xdr-writer-issue117' + self.ext))
@@ -565,9 +563,7 @@ class _GromacsWriterIssue117(object):
             for ts in universe.trajectory:
                 W.write(universe)
 
-        with pytest.warns(UserWarning,
-                          match="ATOMIC_NUMBER record not found"):
-            uw = mda.Universe(PRMncdf, outfile)
+        uw = mda.Universe(PRMncdf, outfile)
 
         # check that the coordinates are identical for each time step
         for orig_ts, written_ts in zip(universe.trajectory, uw.trajectory):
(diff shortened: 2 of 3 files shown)


orbeckst:
FYI, this is what the (small) error looks like in practice
```
>>> u = mda.Universe("md-01.tpr", "md-01.part0001.xtc")
~/miniconda3/envs/mda25/lib/python3.10/site-packages/MDAnalysis/coordinates/XDR.py:203: UserWarning: Cannot write lock/offset file in same location as {self.filename}. Using slow offset calculation.
  warnings.warn(f"Cannot write lock/offset file in same location as "
~/miniconda3/envs/mda25/lib/python3.10/site-packages/MDAnalysis/coordinates/XDR.py:259: UserWarning: Couldn't save offsets because: [Errno 13] Permission denied: '.md-01.part0001.xtc_offsets.npz'
  warnings.warn("Couldn't save offsets because: {}".format(e))
```

Note how `{self.filename}` is not replaced.

testsuite/MDAnalysisTests/coordinates/test_xdr.py
```diff
@@ -46,6 +47,14 @@
 from MDAnalysis.coordinates import XDR
 from MDAnalysisTests.util import get_userid
 
+@pytest.mark.parametrize("filename,kwargs,reference", [
```
IAlibay, line 51:
```suggestion

@pytest.mark.parametrize("filename,kwargs,reference", [
```

testsuite/MDAnalysisTests/coordinates/test_xdr.py
```diff
@@ -46,6 +47,14 @@
...
+    ("foo.xtc", {}, ".foo.xtc_offsets.npz"),
+    ("foo.xtc", {"ending": "npz"}, ".foo.xtc_offsets.npz"),
+    ("bar.0001.trr", {"ending": "npzzzz"}, ".bar.0001.trr_offsets.npzzzz"),
+])
+def test_offsets_filename(filename, kwargs, reference):
+    fn = XDR.offsets_filename(filename, **kwargs)
+    assert fn == reference
```
IAlibay, line 58:
```suggestion
    assert fn == reference

```

testsuite/MDAnalysisTests/coordinates/test_xdr.py
```diff
@@ -755,25 +764,34 @@ def test_nonexistent_offsets_file(self, traj):
...
     def test_reload_offsets_if_offsets_readin_io_fails(self, trajectory):
         # force the np.load call that is called in read_numpy_offsets
         # during _load_offsets to give an IOError
         # ensure that offsets are then read-in from the trajectory
         with patch.object(np, "load") as np_load_mock:
             np_load_mock.side_effect = IOError
-            trajectory._load_offsets()
+            with pytest.warns(UserWarning) as record:
+                trajectory._load_offsets()
+            assert len(record) == 2
```
IAlibay: checking the length of the record is very flaky, can we instead just check that the two warnings you're looking for are in there? "X in Y" type of thing

IAlibay (changes requested):
Couple of small things, otherwise lgtm

testsuite/MDAnalysisTests/coordinates/test_xdr.py
orbeckst: ok, added in [4f18afb](https://github.com/MDAnalysis/mdanalysis/pull/4231/commits/4f18afbb8e423386daf600cd6b13e56158e91eeb)

IAlibay (approved):
cheers, lgtm!

testsuite/MDAnalysisTests/coordinates/test_xdr.py
```diff
@@ -755,25 +766,43 @@ def test_nonexistent_offsets_file(self, traj):
...
             np_load_mock.side_effect = IOError
-            saved_offsets = XDR.read_numpy_offsets(outfile_offsets)
-            assert_equal(saved_offsets, False)
+            with pytest.warns(UserWarning, match=re.escape(
+                    f"Failed to load offsets file {outfile_offsets}")):
+                saved_offsets = XDR.read_numpy_offsets(outfile_offsets)
+        assert saved_offsets == False
 
-    def test_nonexistent_offsets_file(self, traj):
+    def test_corrupted_offsets_file(self, traj):
```
orbeckst, line 774: btw, this shadowed the preceeding test; renaming increased our tests

testsuite/MDAnalysisTests/coordinates/test_xdr.py
```diff
@@ -755,25 +766,43 @@ def test_nonexistent_offsets_file(self, traj):
...
             np_load_mock.side_effect = IOError
-            trajectory._load_offsets()
+            with pytest.warns(UserWarning) as record:
+                trajectory._load_offsets()
+
+            # check that all warning_messages have shown up
+            assert len(record) >= len(warning_messages)
+            found_messages = [str(r.message) for r in record]
+            # find at least 2 occurences (>= instead of == for robustness)
+            assert sum([ref in msg for msg in found_messages
```
orbeckst: I looked through the other tests in the file and found the following approach that looks a bit cleaner where we stack pytest.warns:
```python
       with (pytest.warns(UserWarning, match="Couldn't save offsets") and
              pytest.warns(UserWarning, match="Cannot write")):
            self._reader(filename)
```

I'll rewrite to use that idiom.

orbeckst:
Thanks for the quick review @IAlibay . I apologize, I tacked on a few little things that I couldn't let go when I looked at the code again.

testsuite/MDAnalysisTests/coordinates/test_xdr.py
```diff
@@ -542,7 +553,10 @@ class _GromacsWriterIssue117(object):
 
     @pytest.fixture()
     def universe(self):
-        return mda.Universe(PRMncdf, NCDF)
+        with pytest.warns(UserWarning,
+                          match="ATOMIC_NUMBER record not found"):
```
IAlibay: I'm so confused, why are testing a NetCDF file warning here? Should we use pytest.filterwarning if it's not something we need to explicitly test?

IAlibay (changes requested):

testsuite/MDAnalysisTests/coordinates/test_xdr.py
orbeckst: My thinking was that it filters the warning and alerts us if anything is changing. Given how poor our handling of warning in tests is generally, that seemed reasonable.

However, I also see that it can be confusing to test something not directly related to the code in question (here: xdr).

What's the appropriate `filterwarnings` to use here?

IAlibay: > it filters the warning and alerts us if anything is changing

We should already be testing that warning in the netcdf tests (I'm pretty sure we do because I tried to bring testing to 100% via test_netcdf.py last I touched it). I'll leave it to your decision here (feel free to dismiss my block, everything else looks good).

> What's the appropriate filterwarnings to use here?

Adding the following to the tests calling the fixture I think should do the trick:
`@pytest.mark.filterwarnings("ignore: ATOMIC_NUMBER record not found")`

package/MDAnalysis/coordinates/XDR.py
```diff
@@ -220,10 +219,10 @@ def _load_offsets(self):
...
             data = read_numpy_offsets(fname)
             if not data:
-                warnings.warn("Reading offsets from {} failed, "
-                              "reading offsets from trajectory instead.\n"
-                              "Consider setting 'refresh_offsets=True' "
-                              "when loading your Universe.".format(fname))
+                warnings.warn(f"Reading offsets from {fname} failed, "
+                              f"reading offsets from trajectory instead.\n"
+                              f"Consider setting 'refresh_offsets=True' "
+                              f"when loading your Universe.")
```
RMeli:
```suggestion
                              "reading offsets from trajectory instead.\n"
                              "Consider setting 'refresh_offsets=True' "
                              "when loading your Universe.")
```

RMeli: No need for f-strings here? (Unless there it's some black/formatting thing I'm not aware of...) IIRC some linters might complain.

testsuite/MDAnalysisTests/coordinates/test_xdr.py
```diff
@@ -551,18 +565,19 @@ def test_write_trajectory(self, universe, tmpdir):
...
         for orig_ts, written_ts in zip(universe.trajectory, uw.trajectory):
             assert_almost_equal(
                 written_ts._pos,
                 orig_ts._pos,
                 self.prec,
-                err_msg="coordinate mismatch "
-                "between original and written "
-                "trajectory at frame %d (orig) vs %d "
-                "(written)" % (orig_ts.frame, written_ts.frame))
+                err_msg=(f"coordinate mismatch between original and written "
```
RMeli:
```suggestion
                err_msg=("coordinate mismatch between original and written "
```

testsuite/MDAnalysisTests/coordinates/test_xdr.py
orbeckst: Thanks. Changed.

orbeckst:
One of the runners timed out so I restarted it.

orbeckst:
@RMeli or @IAlibay could one of you please volunteer to captain the PR and eventually merge it? I can do it if you prefer but it's better style (in my opinion) to not be the one merging your own PR (unless it's urgent....).

I think it's good for the project when there's visibly someone in charge of a PR.

IAlibay (approved):

IAlibay:
lgtm!

PR #4249, opened 2023-08-17
[WIP] Modify charge neutrality verification in DielectricConstant

orionarcher:
This PR allows DielectricConstant to check charge neutrality at a customizable granularity.

I work with electrolyte systems where `fragments` are molecules in the system. In the current scheme, DielectricConstant will fail on my systems because ions have non-zero charge. This adds a keyword to modify the level at which charge neutrality is enforced.

I think changing the hard-coded option to "group" to simplify the API would also be a good fix, it's unclear to me why "fragments" is desirable.

Fixes #

 - add new `check_neutrality_of` kwarg to DielectricConstant

PR Checklist
------------
 - [ ] Tests?
 - [x] Docs?
 - [ ] CHANGELOG updated?

----
:books: Documentation preview :books:: https://mdanalysis--4249.org.readthedocs.build/en/4249/

### DIFF
--- a/package/MDAnalysis/analysis/dielectric.py
+++ b/package/MDAnalysis/analysis/dielectric.py
@@ -74,6 +74,10 @@ class DielectricConstant(AnalysisBase):
     make_whole : bool
       Make molecules whole; If the input already contains whole molecules
       this can be disabled to gain speedup
+    check_neutrality_of: str
+        Modify the level at which charge neutrality is enforced. Options are
+        {'group', 'segments', 'residues', 'molecules', 'fragments'}. Default
+        is 'fragments' and the most permissive is 'group'.
     verbose : bool
       Show detailed progress of the calculation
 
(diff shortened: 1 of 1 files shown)


hmacdope:
Seems reasonable to me, would need some tests ofc, but understand its WIP.

orionarcher:
Great thank you, wanted to get a second opinion before updating changelog and writing tests.

PicoCentauri:
@orionarcher if your system has free charges/ions you can not use this module because the equation does not hold if you have free charges.

orionarcher:
Got it. Thanks for the clarification @PicoCentauri.

PicoCentauri:
Yes, you are not the first one that run into these problems. It might be worth to explain this in the docstring.

orionarcher:
How challenging do you think it would be to implement a dielectric calculation that could handle free charges / ions?

PicoCentauri:
What you have to do is to calculate the frequency dependent dielectric spectrum and take the zero frequency limit to get the dielectric constant. I did the [implementation once](https://gitlab.com/maicos-devel/maicos/-/merge_requests/52), but the code never made it into the main branch, because I branched off too far. 😅 It would be wonderful too have this in mda.

Details on the theory are i.e. explained [here](https://doi.org/10.1021/acs.jpcb.1c05303). See equation (10) and section S6 in the SI.

orionarcher:
Thanks for sharing! I'll need to implement this myself in the next couple months. I'll ping you when/if I pick it back up.

PicoCentauri:
Sure! You can also always ask me on Discord!

PR #4253, opened 2023-08-21
Move ConverterBase out of coordinates/base.py

### LINKED ISSUE #3404: Move ConverterBase to the `converters.base`?
Just having a look at things, and I realise that ConverterBase essentially is just a small subclass of IOBase, but I wonder if it might make more sense to have that in `converters.base` so that it's more logically situated?

ianmkenney:
Fixes #3404 

 - Moved ConverterBase into converters/base.py

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

### DIFF
--- a/package/MDAnalysis/converters/base.py
+++ b/package/MDAnalysis/converters/base.py
@@ -1,0 +1,62 @@
+# -*- Mode: python; tab-width: 4; indent-tabs-mode:nil; coding:utf-8 -*-
+# vim: tabstop=4 expandtab shiftwidth=4 softtabstop=4
+#
+# MDAnalysis --- https://www.mdanalysis.org
+# Copyright (c) 2006-2017 The MDAnalysis Development Team and contributors
+# (see the file AUTHORS for the full list of names)
+#
+# Released under the GNU Public Licence, v2 or any higher version
+#
+# Please cite your use of MDAnalysis in published work:
+#
+# R. J. Gowers, M. Linke, J. Barnoud, T. J. E. Reddy, M. N. Melo, S. L. Seyler,
+# D. L. Dotson, J. Domanski, S. Buchoux, I. M. Kenney, and O. Beckstein.
+# MDAnalysis: A Python package for the rapid analysis of molecular dynamics
+# simulations. In S. Benthall and S. Rostrup editors, Proceedings of the 15th
+# Python in Science Conference, pages 102-109, Austin, TX, 2016. SciPy.
+# doi: 10.25080/majora-629e541a-00e
+#
+# N. Michaud-Agrawal, E. J. Denning, T. B. Woolf, and O. Beckstein.
+# MDAnalysis: A Toolkit for the Analysis of Molecular Dynamics Simulations.
+# J. Comput. Chem. 32 (2011), 2319--2327, doi:10.1002/jcc.21787
+#
+
+"""\
+Converters
+----------
+
+Converters output information to other libraries.
+
+.. autoclass:: ConverterBase
+   :members:
+   :inherited-members:
+"""
+
+from .. import _CONVERTERS
+from ..coordinates.base import IOBase, SingleFrameReaderBase
+from ..lib.util import asiterable
+
+
+class _Convertermeta(type):
+    # Auto register upon class creation
+    def __init__(cls, name, bases, classdict):
+        type.__init__(type, name, bases, classdict)
+        try:
+            fmt = asiterable(classdict['lib'])
+        except KeyError:
+            pass
+        else:
+            for f in fmt:
+                f = f.upper()
+                _CONVERTERS[f] = cls
+
+
+class ConverterBase(IOBase, metaclass=_Convertermeta):
+    """Base class for converting to other libraries.
+    """
+
+    def __repr__(self):
+        return "<{cls}>".format(cls=self.__class__.__name__)
+
+    def convert(self, obj):
+        raise NotImplementedError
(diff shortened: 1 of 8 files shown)


package/MDAnalysis/coordinates/base.py
```diff
@@ -1782,31 +1770,3 @@ def range_length(start, stop, step):
...
-        try:
-            fmt = asiterable(classdict['lib'])
-        except KeyError:
-            pass
-        else:
-            for f in fmt:
-                f = f.upper()
-                _CONVERTERS[f] = cls
-
-class ConverterBase(IOBase, metaclass=_Convertermeta):
```
IAlibay, line 1800: We can't just purely move this out of here without some kind of deprecation window.

Could you maybe subclass the ConverterBase here and add a warning in `__init_subclass__`?

IAlibay (changes requested):

package/MDAnalysis/coordinates/base.py
ianmkenney: Unfortunately that leads to a circular reference. Instead I include the code in both `base.py` files but have a deprecation warning when the `ConverterBase` from `coordinates` is subclassed.

package/MDAnalysis/coordinates/base.py
```diff
@@ -101,17 +101,18 @@
...
 
-
 Converters
 ----------
+Converters output information to other libraries. 
 
-Converters output information to other libraries.
+.. deprecated:: 2.7.0
+    All converter code has been moved to :mod:`MDAnalysis.converters` and will
+    be removed in 3.0.0.
```
orbeckst: Make clearer that converters still exist but will be removed from `coordinates`.
```suggestion
    be removed from the :mod:`MDAnalysis.coordinates.base` module in 3.0.0.
```
If we actually manage to move converters to kits for 3.0.0 then we can still rewrite this message.

package/MDAnalysis/coordinates/base.py
orbeckst: This solution works for me and the comments make clear what to do. I also really don't have a more elegant solution.

package/MDAnalysis/coordinates/base.py
```diff
@@ -1797,16 +1801,29 @@ def __init__(cls, name, bases, classdict):
...
-    See Also
-    --------
-    :mod:`MDAnalysis.converters`
+    .. deprecated:: 2.7.0
+        This class has been moved to 
+        :class:`MDAnalysis.converters.base.ConverterBase` and will be removed
+        from :mod:`MDAnalysis.coordinates.base` in 3.0.0.
     """
 
+    def __init_subclass__(cls):
```
orbeckst, line 1818: The deprecation must be tested.

orbeckst (changes requested):
Looks good but needs a test for the deprecation. See other tests that check for the `DeprecationWarning` and the specific message.

package/MDAnalysis/coordinates/base.py
IAlibay: I think there's maybe a misconception here - we aren't moving the ConverterBase machinery out of core.

We still want to provide the necessary API for folks to create converters, it's just that new converter implementations (and maybe some old ones) will have to move out of the library to avoid library bloat.

package/MDAnalysis/coordinates/base.py
IAlibay: Ok that makes sense - we definitely need tests to check that each respective subclass does / doesn't raise the warning.

package/MDAnalysis/converters/base.py
```diff
@@ -0,0 +1,61 @@
...
+            fmt = asiterable(classdict['lib'])
+        except KeyError:
+            pass
+        else:
+            for f in fmt:
+                f = f.upper()
+                _CONVERTERS[f] = cls
+
+
+class ConverterBase(IOBase, metaclass=_Convertermeta):
```
IAlibay, line 53: We need to also switch over our existing converters to using ConverterBase from here - otherwise we'll have a ton of user warnings.

IAlibay (changes requested):

package/MDAnalysis/coordinates/base.py
orbeckst: Thanks for the clarification.

orbeckst (approved):
Thanks for adding the tests. You addressed everything that I had seen. I _think_ you also addressed all of @IAlibay 's comments but I leave it to him to judge.

I checked the docs and they look fine. I am not 100% convinced that we need a `versionadded 2.7.0` for the `MDAnalysis.converters.base` docs because it _might_ be a bit confusing as it really had been present since 2.0.0, just not in this place, but I could be convinced either way and leave this to you if you want to add it.

IAlibay:
> I checked the docs and they look fine. I am not 100% convinced that we need a `versionadded 2.7.0` for the `MDAnalysis.converters.base` docs because it _might_ be a bit confusing as it really had been present since 2.0.0, just not in this place, but I could be convinced either way and leave this to you if you want to add it.

I would assume a `versionchanged` rather than added (if you were to add the directive at all)?

IAlibay:
@ianmkenney @orbeckst - yes my comments have been addressed, but I'm going to keep the block for now because I think we're back in another merge freeze (for non bugfixes).

IAlibay (approved):
lgtm - unblocking now the release is done

Issue #3901, opened 2022-11-04
Document NoDataError better and ensure it's properly applied everywhere


IAlibay:
As per #3898 and #3837 before it, the current interpretation of `NoDataError` seems to be that it exists within the realm of TopologyAttrs. If this is the interpretation we are going for, we should do the following:

1. Clarify the documentation for this: "Raised when empty input is not allowed or required data are missing." is probably too non-specific.
2. Make sure that the codebase reflects this. In cases where `NoDataError` is used for zero-width arrays or empty AtomGroups, then we should make sure we raise a ValueError. I'm not sure how many cases of this there are, but #3898 does point out that it's probably happening in quite a few places.

Nazi-pikachu:
Hi @IAlibay I am new to open source and excited to contribute to this library. Can i work on this issue?

hmacdope:
Hi @nazi-pikachu. MDAnalysis has a diverse range of contributors and values a respectful, tolerant, and open community. We would love to have you contribute as a member of our community, however unfortunately your name "nazi-pikachu" violates our [Code of Conduct](https://www.mdanalysis.org/pages/conduct/), specifically Section 5 which says:

 """
Be careful in the words that you choose. Be kind to others. Do not insult or put down other community members. Harassment and other exclusionary behavior are not acceptable. This includes, but is not limited to:

- threats or violent language directed against another person
- discriminatory jokes and language

"""
We would allow you to contribute under another github username, but cannot accept your contributions under your current moniker.

Nazi-pikachu:
Hi @hmacdope I regret that my github username is a violation of code of conduct. Thank for pointing it out , I was not aware of it. Now i will look for altenative to creating a seperate account  as all my existing contribution is from this account, any suggestion from you will be helpful.

hmacdope:
I would suggest either using a new account or changing your name, whichever you prefer.

chandra20500:
Hi, I would like to work on this Issue. it would be great could you please assign this issue to me

Adarshsantrapatro:
hii i would like to work on this issue can you assign me this

hmacdope:
Hi @chandra20500 and @Adarshsantrapatro, we do not assign people issues. Just make a PR and we will focus on the first PR made on that issue.

rajeck1234:
hlw i want to contribute on this issue

RMeli:
[...]

PR #4258, opened 2023-08-26
Ensure NoJump handles jumps that occur on the second frame in NPT trajectories

### LINKED ISSUE #4257: `NoJump` transformation ignores jumps across boundaries on the second frame for NPT trajectories
## Expected behavior ##

It's very cool the MDAnalysis now has a nojump transformation 🎉 

After it was released I decided it would make sense to [deprecate the nojump transformation I had written for lipyphilic](https://github.com/p-j-smith/lipyphilic/pull/124), expecting the MDAnalysis transformation had been tested with an NPT trajectory.

## Actual behavior ##

Looking at the tests, it seems that [both trajectories used for testing](https://github.com/MDAnalysis/mdanalysis/blob/8923c4a8203e755939ad9fc3f2242a837a493c28/testsuite/MDAnalysisTests/transformations/test_nojump.py#L21-L48) are NVT.

I've written a couple of tests with dummy Universes and NPT trajectories and I think there might be issue with incorrect unwrapping, but I may be misunderstanding something.

I've written two tests to check the unwrapping in the following situations:

- an atom remains stationary but crosses a periodic boundary due to decreasing box size
- an atom moves at a constant velocity and crosses a periodic boundary both due to it's velocity and the decreasing box size

First write the Universes and trajectories to file:
```python
import MDAnalysis as mda


def write_nojump_universe_stationary_npt():
    """
    Create a Universe in which a single atom remains stationary but box size
    decreases by 5 Angstrom at each frame.
    
    This initial box dimensions are [100, 100, 100, 90, 90, 90]
    The trajectory consists of three frames.

    Unwrapped coordinates: atom remains stationary
    coordinates = [
        [97.5, 50, 50]
        [97.5, 50, 50]
        [97.5, 50, 50]
    ]

    Wrapped coordinates: box dimensions decrease 5 Angstrom each frame, so atom is wrapped in x-dimension at 2nd frame
    coordinates = [
        [97.5, 50, 50]
        [2.5, 50, 50]
        [2.5, 50, 50]
    ]

    """
    n_atoms = 1
    n_frames = 3
    u = mda.Universe.empty(n_atoms, trajectory=True)

    coordinates = np.empty((n_frames, u.atoms.n_atoms, 3))
    coordinates[0] = [97.5, 50.0, 50.0]
    coordinates[1] = [2.5, 50.0, 50.0]
    coordinates[2] = [2.5, 50.0, 50.0]
    u.load_new(coordinates, order="fac")

    box = np.array([100, 100, 100, 90, 90, 90])
    u.dimensions = box
    u.atoms.write("nojump_stationary_npt.pdb")

    with mda.Writer("nojump_stationary_npt.xtc") as f:
        for ts in u.trajectory:
            box[:3] -= 5 * ts.frame  # shrink size by 5 Ansgtroms
            u.dimensions = box
            f.write(u.atoms)

def write_nojump_universe_constvel_npt():
    """
    Create a Universe in which a single atom move 5 Angstrom in the positive
    x directaion at each frame AND the box size decreases by 5 Angstrom at each frame.

    This initial box dimensions are [100, 100, 100, 90, 90, 90]
    The trajectory consists of three frames.

    Unwrapped coordinates: atom move 5 Angstrom in the positive x direction each frame
    coordinates = [
        [97.5, 50, 50]
        [102.5, 50, 50]
        [107.5, 50, 50]
    ]

    Wrapped coordinates: box dimensions decrease 5 Angstrom each frame and particle moves 5 A each frame in positive x-dimension
    coordinates = [
        [97.5, 50, 50]
        [7.5, 50, 50]
        [12.5, 50, 50]
    ]

    """
    n_atoms = 1
    n_frames = 3
    u = mda.Universe.empty(n_atoms, trajectory=True)

    coordinates = np.empty((n_frames, u.atoms.n_atoms, 3))
    coordinates[0] = [97.5, 50.0, 50.0]
    coordinates[1] = [7.5, 50.0, 50.0]
    coordinates[2] = [12.5, 50.0, 50.0]
    u.load_new(coordinates, order="fac")

    box = np.array([100, 100, 100, 90, 90, 90])
    u.dimensions = box
    u.atoms.write("nojump_constvel_npt.pdb")

    with mda.Writer("nojump_constvel_npt.xtc") as f:
        for ts in u.trajectory:
            box[:3] -= 5 * ts.frame  # shrink size by 5 Ansgtroms
            u.dimensions = box
            f.write(u.atoms)

write_nojump_universe_stationary_npt()
write_nojump_universe_constvel_npt()
```

Now run the tests:
```python
import pytest
import numpy as np
import MDAnalysis as mda

@pytest.fixture
def nojump_universe_stationary_npt():
    return mda.Universe("nojump_stationary_npt.pdb", "nojump_stationary_npt.xtc")

@pytest.fixture
def nojump_universe_constvel_npt():
    return mda.Universe("nojump_constvel_npt.pdb", "nojump_constvel_npt.xtc")

def test_nojump_stationary_npt(nojump_universe_stationary_npt):
    u = nojump_universe_stationary_npt
    u.trajectory.add_transformations(mda.transformations.NoJump())
    x_position = 97.5
    np.testing.assert_allclose(u.trajectory.timeseries()[:, 0, 0], x_position)

def test_nojump_constvel_npt(nojump_universe_constvel_npt):
    u = nojump_universe_constvel_npt
    u.trajectory.add_transformations(mda.transformations.NoJump())
    x_positions = [97.5, 102.5, 107.5]
    np.testing.assert_allclose(u.trajectory.timeseries()[:, 0, 0], x_positions)

```

The first test fails with the error message:
```
./tests/lipyphilic/analysis/transformations/test_transformations.py::test_nojump_stationary_npt Failed: [undefined]AssertionError: 
Not equal to tolerance rtol=1e-07, atol=0

Mismatched elements: 2 / 3 (66.7%)
Max absolute difference: 95.
Max relative difference: 0.97435897
 x: array([97.5,  2.5,  2.5], dtype=float32)
 y: array(97.5)
nojump_universe_stationary_npt = <Universe with 1 atoms>

    def test_nojump_stationary_npt(nojump_universe_stationary_npt):
        u = nojump_universe_stationary_npt
        u.trajectory.add_transformations(mda.transformations.NoJump())
        x_position = 97.5
>       np.testing.assert_allclose(u.trajectory.timeseries()[:, 0, 0], x_position)
```

And the second test fails with the error message:
```
./tests/lipyphilic/analysis/transformations/test_transformations.py::test_nojump_constvel_npt Failed: [undefined]AssertionError: 
Not equal to tolerance rtol=1e-07, atol=0

Mismatched elements: 2 / 3 (66.7%)
Max absolute difference: 95.
Max relative difference: 0.92682927
 x: array([97.5     ,  7.5     , 12.500001], dtype=float32)
 y: array([ 97.5, 102.5, 107.5])
nojump_universe_constvel_npt = <Universe with 1 atoms>

    def test_nojump_constvel_npt(nojump_universe_constvel_npt):
        u = nojump_universe_constvel_npt
        u.trajectory.add_transformations(mda.transformations.NoJump())
        x_positions = [97.5, 102.5, 107.5]
>       np.testing.assert_allclose(u.trajectory.timeseries()[:, 0, 0], x_positions)
```

So in both cases it seems that the atom is not being unwrapped in the x-dimension.

Note, both of these tests pass when using the lipyphilic nojump transformation:

```python
import pytest
import numpy as np
import MDAnalysis as mda
import lipyphilic as lpp

def test_nojump_stationary_npt_lpp(nojump_universe_stationary_npt):
    u = nojump_universe_stationary_npt
    u.trajectory.add_transformations(lpp.transformations.nojump(ag=u.atoms, nojump_z=True))
    x_position = 97.5
    np.testing.assert_allclose(u.trajectory.timeseries()[:, 0, 0], x_position)

def test_nojump_constvel_npt_lpp(nojump_universe_constvel_npt):
    u = nojump_universe_constvel_npt
    u.trajectory.add_transformations(lpp.transformations.nojump(ag=u.atoms, nojump_z=True))
    x_positions = [97.5, 102.5, 107.5]
    np.testing.assert_allclose(u.trajectory.timeseries()[:, 0, 0], x_positions)

```

but this may be because I have misunderstood something about nojump unwrapping!

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) 2.7.0-dev0
- Which version of Python (`python -V`)? Python 3.11.4
- Which operating system? macOS Monterey

p-j-smith:
Fixes #4257 

 - ensure jumps across periodic boundaries that occur at the second frame are accounted for by `mda.transformations.NoJump`
 - added two tests:
   - `test_nojump_2nd_frame`.- tests that jumps that occur on the second frame are handled correctly
   - `test_nojump_3rd_frame`.- tests that jumps that occur on the third frame are handled correctly
- the `test_nojump_3rd_frame` test isn't necessary and can probably be removed. However, I've included it here to illustrate the the fix in this pr is necessary. Before this pr, `test_nojump_3rd_frame` would pass but `test_nojump_2nd_frame` would fail, and the only difference between the trajectories is the frame at which the jump across a boundary occurs

- the fix requires that we know the atom positions and box size at the first frame of the trajectory, and so I've added an `ag` argument that takes an AtomGroup to be unwrapped. This however would be a **breaking change** to the api. Perhaps a better way would be to add an if-statement to `mda.transformations.NoJump._transform` that checks if we're on the first frame of the trajectory, and if so then sets the appropriate values for `NoJump.prev`, `NoJump.old_frame`, and `NoJump.older_frame`?

- updated all other tests for nojump to use the required `ag` argument

~~Note, I've temporarily put the test datasets in the testsuite itself so the tests can run before the datasets are added to `MDAnalysisData`~~

Edit: I was running into some issues with CI failing because the test datafiles I added weren't present in `MDAnalysisData`. So I've added a tranformation `mda.transformations.boxdimensions.set_variable_dimensions` to set different box dimensions at each frame (so we can create an NPT trajectory to go with `mda.Universe.empty`).

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?


p-j-smith:
> the fix requires that we know the atom positions and box size at the first frame of the trajectory, and so I've added an ag argument that takes an AtomGroup to be unwrapped. This however would be a breaking change to the api. Perhaps a better way would be to add an if-statement to mda.transformations.NoJump._transform that checks if we're on the first frame of the trajectory, and if so then sets the appropriate values for NoJump.prev, NoJump.old_frame, and NoJump.older_frame?

I've removed the `ag` argument I added and gone with using an if-statement to check it we're on the first frame. This both avoids a breaking change and ensures the trajectory can be iterated over multiple times and unwrapping still be performed correctly. I've also added a test to check that coordinates are always unwrapped correctly when iterating multiple times.

package/MDAnalysis/transformations/boxdimensions.py
```diff
@@ -25,15 +25,16 @@
 Set box dimensions --- :mod:`MDAnalysis.transformations.boxdimensions`
 =======================================================================
 
-Set dimensions of the simulation box to a constant vector across all timesteps.
-
+Set dimensions of the simulation box, eiterh to a constant vector across
```
hmacdope:
```suggestion
Set dimensions of the simulation box, either to a constant vector across
```

package/MDAnalysis/transformations/boxdimensions.py
```diff
@@ -92,3 +93,72 @@ def __init__(self,
...
+    def __init__(self,
+                 dimensions,
+                 max_threads=None,
+                 parallelizable=True):
+        super().__init__(max_threads=max_threads,
+                         parallelizable=parallelizable)
+        self.dimensions = dimensions
+
+        try:
+            self.dimensions = np.asarray(self.dimensions, np.float32)
```
hmacdope, line 97: I thought box was `np.float64`?

testsuite/MDAnalysisTests/transformations/test_nojump.py
```diff
@@ -121,6 +203,84 @@ def test_nojump_constantvel(nojump_constantvel_universe):
...
+    Wrapped coordinates are:
+    coordinates = [
+        [97.5, 50.0, 50.0],
+        [2.5, 50.0, 50.0],
+        [2.5, 50.0, 50.0],
+        [2.5, 50.0, 50.0],
+    ]
+    where each row is a different timestep.
+
+    Unwarpped coordinates are the same at each frame:
```
hmacdope:
```suggestion
    Unwrapped coordinates are the same at each frame:
```

hmacdope (changes requested):
Tiny little nitpicks, but other than that looking great. Don't forget CHANGELOG ;)

package/MDAnalysis/transformations/boxdimensions.py
p-j-smith: I copied that from [the existing `set_dimensions`](https://github.com/MDAnalysis/mdanalysis/blob/d5a30e9ad868b3b25eb27dd1df0e70cda983d8bd/package/MDAnalysis/transformations/boxdimensions.py#L77), and I just checked and the dimensions are indeed `float32`. But I think you're right - they probably should be `np.float64`?

p-j-smith:
thanks for the speedy review @hmacdope! And sorry for mixing a fix and a feature in the same pr - I probably should have added `set_variable_dimensions` separately :)

package/MDAnalysis/transformations/boxdimensions.py
```diff
@@ -92,3 +93,74 @@ def __init__(self,
     def _transform(self, ts):
         ts.dimensions = self.dimensions
         return ts
+
+
+class set_variable_dimensions(TransformationBase):
```
IAlibay: Could we just move this new functionality as a new option of the existing `set_dimension`? The order of how things are done is very similar and it would be a much better user experience to not have to deal with (choose between) a lot of classes that do very similar things.

IAlibay (changes requested):
One initial blocking request - I'll follow up with smaller review things.

package/CHANGELOG
```diff
@@ -18,9 +18,12 @@ The rules for this file:
  * 2.7.0
 
 Fixes
+  * Fix `NoJump`` unwrapping for jumps on the second frame in a
```
IAlibay: Please add your handle to the 2.7.0 changelog entry on line 16.

package/MDAnalysis/transformations/boxdimensions.py
p-j-smith: Yep I think that would make sense

package/MDAnalysis/transformations/boxdimensions.py
```diff
@@ -92,3 +93,74 @@ def __init__(self,
...
+                       'given by an float array of shape (N_frames, 6), '
+                       ' containing 3 lengths and 3 angles at each frame: '
+                       '[a, b, c, alpha, beta, gamma]')
+            raise ValueError(errmsg)
+
+    def _transform(self, ts):
+        try:
+            ts.dimensions = self.dimensions[ts.frame]
+        except IndexError as e:
+            raise NoDataError(f"Dimensions array has no data for frame {ts.frame}") from e
```
IAlibay: As per https://github.com/MDAnalysis/mdanalysis/issues/3901, `NoDataError` is really only meant to exist within the realm of `TopologyAttr`. A `ValueError` probably should be used here.

package/MDAnalysis/transformations/boxdimensions.py
richardjgowers: I think they're stored as 32s, but promoted for distance functions to 64, it's messy.

testsuite/MDAnalysisTests/transformations/test_boxdimensions.py
```diff
@@ -81,3 +93,74 @@ def test_dimensions_transformations_api(boxdimensions_universe):
...
+        [2, 2, 2, 90, 90, 90],
+        [4, 4, 4, 90, 90, 90],
+        [8, 8, 8, 90, 90, 90],
+    ])
+    transform = set_variable_dimensions(new_dims)
+    variable_boxdimensions_universe.trajectory.add_transformations(transform)
+    for ts in variable_boxdimensions_universe.trajectory:
+        assert_array_almost_equal(variable_boxdimensions_universe.dimensions,
+                                  new_dims[ts.frame], decimal=6)
+
```
IAlibay, line 113:
```suggestion


```

testsuite/MDAnalysisTests/transformations/test_boxdimensions.py
```diff
@@ -81,3 +93,74 @@ def test_dimensions_transformations_api(boxdimensions_universe):
...
+)
+def test_variable_dimensions_vector(
+    variable_boxdimensions_universe,
+    dim_vector_shapes,
+):
+    # wrong box dimension vector shape
+    ts = variable_boxdimensions_universe.trajectory.ts
+    with pytest.raises(ValueError, match='valid box dimension shape'):
+        set_variable_dimensions(dim_vector_shapes)(ts)
+
```
IAlibay:
```suggestion


```

testsuite/MDAnalysisTests/transformations/test_nojump.py
```diff
@@ -47,6 +47,88 @@ def nojump_constantvel_universe():
     reference.load_new(coordinates, order="fac")
     return reference
```
IAlibay, line 49:
```suggestion


```

testsuite/MDAnalysisTests/transformations/test_nojump.py
```diff
@@ -47,6 +47,88 @@ def nojump_constantvel_universe():
...
+    n_frames = 4
+    u = mda.Universe.empty(n_atoms, trajectory=True)
+    coordinates = np.empty((n_frames, u.atoms.n_atoms, 3))
+    coordinates[0] = [97.5, 50.0, 50.0]
+    coordinates[1] = [2.5, 50.0, 50.0]
+    coordinates[2] = [2.5, 50.0, 50.0]
+    coordinates[3] = [2.5, 50.0, 50.0]
+    u.load_new(coordinates, order="fac")
+    return u
+
```
IAlibay, line 69:
```suggestion


```

testsuite/MDAnalysisTests/transformations/test_nojump.py
```diff
@@ -121,6 +203,84 @@ def test_nojump_constantvel(nojump_constantvel_universe):
         atol=5e-06,
     )
```
IAlibay, line 207:
```suggestion


```

testsuite/MDAnalysisTests/transformations/test_nojump.py
```diff
@@ -121,6 +203,84 @@ def test_nojump_constantvel(nojump_constantvel_universe):
...
+        [95, 100, 100, 90, 90, 90],
+    ])
+    workflow = [
+        mda.transformations.boxdimensions.set_variable_dimensions(dim),
+        NoJump(),
+    ]
+    u.trajectory.add_transformations(*workflow)
+    x_position = 97.5
+    np.testing.assert_allclose(u.trajectory.timeseries()[:, 0, 0], x_position)
+
```
IAlibay, line 241:
```suggestion


```

package/MDAnalysis/transformations/boxdimensions.py
```diff
@@ -92,3 +93,74 @@ def __init__(self,
...
+    and [*2a*, *2b*, *2c*, *alpha*, *beta*, *gamma*] at the second frame.
+    Lengths *a*, *b*, *c* are in the MDAnalysis length unit (Å), and angles
+    are in degrees.
+
+    .. code-block:: python
+
+        dim = [
+            [2, 2, 2, 90, 90, 90],
+            [4, 4, 4, 90, 90, 90],
+        ]
```
richardjgowers: I'd use a numpy array in the example if possible

IAlibay (changes requested):
the "small changes" review

package/MDAnalysis/transformations/nojump.py
```diff
@@ -123,6 +123,11 @@ def _transform(self, ts):
         except np.linalg.LinAlgError:
             msg = f"Periodic box dimensions are not invertible at step {ts.frame}"
             raise NoDataError(msg)
+        if ts.frame == 0:
```
IAlibay, line 126: Would it be worth adding a small entry in the docs above to detail this case? Maybe just some inline comment of why this is needed would work - I will admit I wasn't able to quickly read through the issues and follow exactly why this was happening.

package/MDAnalysis/transformations/boxdimensions.py
p-j-smith: ah okay, so is it okay to leave it as 32 here?

package/CHANGELOG
p-j-smith: well spotted!

package/MDAnalysis/transformations/boxdimensions.py
p-j-smith: ah okay, that's good to know

package/MDAnalysis/transformations/nojump.py
p-j-smith: yeah good point, it's not immediately obvious. I've added a comment to try to explain what's going on, let me know if it's not clear

p-j-smith:
Thanks for the reviews @richardjgowers and @IAlibay ! I think I've addressed all your comments, and I've just rebased on to the latest develop

package/MDAnalysis/transformations/nojump.py
IAlibay: Thanks!

package/MDAnalysis/transformations/nojump.py
```diff
@@ -123,6 +123,17 @@ def _transform(self, ts):
         except np.linalg.LinAlgError:
             msg = f"Periodic box dimensions are not invertible at step {ts.frame}"
             raise NoDataError(msg)
```
IAlibay, line 125: Somehow I'm glad I'm not the only one that totally used `NoDataError` the wrong way around for ages - no need to fix here but we should consider fixing this later as part of #3901

package/MDAnalysis/transformations/boxdimensions.py
```diff
@@ -80,15 +99,20 @@ def __init__(self,
...
+                       'or (N, 6) where N is the number of frames in the '
+                       'trajectory and the dimension vector(s) containing '
+                       '3 lengths and 3 angles: '
                        '[a, b, c, alpha, beta, gamma]')
             raise ValueError(errmsg)
 
     def _transform(self, ts):
-        ts.dimensions = self.dimensions
+        try:
+            ts.dimensions = self.dimensions[0] if self.dimensions.shape[0] == 1 else self.dimensions[ts.frame]
```
IAlibay: This is both a long line and also a bit hard to read (at least to me), splitting this over a few lines probably would be worth it?

package/MDAnalysis/transformations/boxdimensions.py
```diff
@@ -80,15 +99,20 @@ def __init__(self,
...
+        except IndexError as e:
+            raise ValueError(f"Dimensions array has no data for frame {ts.frame}") from e
```
IAlibay:
```suggestion
            errmsg = f"Dimensions array has no data for frame {ts.frame}"
            raise ValueError(errmsg) from e
```
I know it's annoying but pep8 - note we don't care about the `black` linter stuff, but the flake8 errors from here probably should be addressed: https://github.com/MDAnalysis/mdanalysis/actions/runs/5997923144/job/16265248319

package/MDAnalysis/transformations/boxdimensions.py
```diff
@@ -25,44 +25,63 @@
...
+        Expected shapes are (6, 0) or (1, 6) or (N, 6), where
+        N is the number of frames in the trajectory. If shape is (6, 0) or
+        (1, 6), the same dimensions will be used at every frame in the
+        trajectory.
 
     Returns
     -------
     :class:`~MDAnalysis.coordinates.timestep.Timestep` object
+
+    .. versionchanged:: 2.7.0
```
IAlibay, line 84:
```suggestion

    .. versionchanged:: 2.7.0
```
This is one of those annoying ones where docs don't render well if you don't have enough empty lines, see: https://mdanalysis--4258.org.readthedocs.build/en/4258/documentation_pages/transformations/boxdimensions.html

IAlibay (changes requested):
Couple more small things, otherwise lgtm

package/MDAnalysis/transformations/nojump.py
richardjgowers: Yeah the first usage a few lines before looks ok, but this is arguably ValueError or something, including the values of the dimensions would also help debug (I'm assuming if it's not invertible it's all zeros or something)

p-j-smith: Yeah I've definitely misappropriated `NoDataError` outside of MDAnalysis too in other code I've written...

package/MDAnalysis/transformations/boxdimensions.py
p-j-smith: ah sorry I forgot to run flake8 locally, fixed now! Perhaps adding a pre-commit config for people to optionally use would be useful?

IAlibay: > pre-commit config for people to optionally use would be useful?

It's.. complicated, I'll leave it as that for now.

p-j-smith:
I just realised that with the changes in this pr there are now some parts of `NoJump` that are not tested (because of the early return statement). I'll add tests for them tomorrow

package/MDAnalysis/transformations/boxdimensions.py
p-j-smith: ha! okay :)

orbeckst:
@hmacdope could you be the one looking after this PR, chasing reviews if necessary, and merging, i.e., assign it to you?

package/MDAnalysis/transformations/boxdimensions.py
IAlibay: This still needs addressing from the looks of things.

p-j-smith: Woops, sorry I missed that

hmacdope:
> @hmacdope could you be the one looking after this PR, chasing reviews if necessary, and merging, i.e., assign it to you?

Yep, ill re-review ASAP.

IAlibay (approved):
lgtm thanks!

hmacdope (approved):
Looks good to me @p-j-smith. Thank you so much for the detailed bug report, timely fix and detailed testing! Its really appreciated.

p-j-smith:
Thanks for the speedy reviews everyone!

orbeckst:
Thank you for the contribution, Paul. Really good work!Am 9/3/23 um 09:39 schrieb Paul Smith ***@***.***>:﻿
Thanks for the speedy reviews everyone!

PR #4261, opened 2023-08-28
Revert numpy build pinning changes

### LINKED ISSUE #4260: Dealing with NumPy build time requirements for the 2.6.x release
## Some background

As detailed in #4183 the NumPy C API became backwards compatible as part of NumPy 1.25+. Prior to that point, we used to build with the "oldest supported NumPy version" (currently 1.22.3), which guaranteed forward compatibility.

## The issue

In 2.6.0 (specifically #4198) we switch out our pyproject.toml build section to just use numpy >=1.25. This worked great for pip installs.

**However** what we hadn't accounted for is the fact that conda-forge does not currently have the adequate amount of machinery to handle this.

Everything in conda-forge is set up to deal with a build where you use an "oldest supported NumPy version", so if you try to build with v1.25, you end up with an install pin range of "numpy >=1.25,<2.0". This is problematic since we need to be able to support old versions of NumPy and conda-forge is the most popular way folks install MDAnalysis.

## The status of conda-forge support & how other packages are doing it

My understanding is that this won't be fully resolved until the 2.0 release of NumPy (see: https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/4816).

Indeed scipy eventually decided to punt the issue to at least the next minor release.

## The solutions

There are two possible solution here:

1. We revert the change introduced in 2.6.0, cherry pick that change and put out a 2.6.1 release. This would ensure that there is no discrepancies in the build-time NumPy versions between our deployments and also that we don't have to deal with further headaches here until _after_ NumPy and SciPy lead the way for us.
2. We ignore the difference between PyPi and conda-forge builds, build with new numpy in one and old numpy in another.

## Which one would create less work

_A priori_ option number two would seem like less work, but honestly I think number one would allow us to bury our heads in the sand for a bit longer - also we've _just_ done a release, so we know all the machinery works right now. It would be less painful to do a bugfix release now than to do it later.

My personal preference is just to do option one and be done with it.

IAlibay:
Fixes #4260 

This reverts the numpy build pinning and would allow us to move forward with a 2.6.1 release.

 - Reverts numpy build pinning in pyproject.toml
 - Switches build isolation back on for everything but nightly wheel checks

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

----
:books: Documentation preview :books:: https://mdanalysis--4261.org.readthedocs.build/en/4261/

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -23,6 +23,8 @@ Fixes
 Enhancements
 
 Changes
+  * Reverts PR #4108, builds are now again made using the oldest
+    supported Python version.
   * NumPy `in1d` replaced with `isin` in preparation for NumPy
     `2.0` (PR #4255)
   * Update documentation for SurvivalProbabilty to be more clear
--- a/.github/actions/build-src/action.yaml
+++ b/.github/actions/build-src/action.yaml
@@ -42,7 +42,7 @@ inputs:
     #   example, when we use nightly wheels of NumPy.
     descriptions: 'Use build isolation for pip installs, if false `--no-build-isolation` is used. '
     required: true
-    default: false
+    default: true
 
 
 runs:
--- a/.github/workflows/gh-ci-cron.yaml
+++ b/.github/workflows/gh-ci-cron.yaml
@@ -69,6 +69,8 @@ jobs:
       with:
         build-tests: true
         build-docs: false
+        # We don't use build isolation because we want to ensure that we
+        # test building with brand new versions of NumPy here.
         isolation: false
 
     - name: run_tests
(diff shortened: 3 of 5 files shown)


package/CHANGELOG
```diff
@@ -23,6 +23,8 @@ Fixes
 Enhancements
 
 Changes
+  * Reverts PR #4108, builds are now again made using the oldest
```
IAlibay, line 26: I would just cut a bugfix release out of the current state of 2.7.0, rather than cherry pick - everything here seems like sensible bug fixes imho.

IAlibay:
Seems like this is the preferred path (at least for now).

I'm going to call a merge freeze on non-bugfix things. @MDAnalysis/coredevs can I have a priority review here please?

richardjgowers (approved):

PR #4263, opened 2023-08-28
Docs warning for charge in DielectricConstant

### LINKED ISSUE #4262: [Docs] document limits of DielectricConstant
Follow-up from https://github.com/MDAnalysis/mdanalysis/pull/4249#issuecomment-1696077930

As per: https://github.com/MDAnalysis/mdanalysis/pull/4249#issuecomment-1696077930

The limits of DielectricConstant, especially regarding free charge, should be better documented.

PicoCentauri:
Fixes #4262

Add a warning box to the documentation that only charge neutral systems can be handled by the current implementation.
See also #4249 for details on why this is important.

PR Checklist
------------
 - [x] Tests (we have a test for the error raise already.)?
 - [x] Docs?
 - [x] CHANGELOG updated?

----
:books: Documentation preview :books:: https://mdanalysis--4263.org.readthedocs.build/en/4263/

### DIFF
--- a/package/MDAnalysis/analysis/dielectric.py
+++ b/package/MDAnalysis/analysis/dielectric.py
@@ -65,6 +65,10 @@ class DielectricConstant(AnalysisBase):
     the usual case if electrostatics are handled with a Ewald summation
     technique. See [Neumann1983]_ for details on the derivation.
 
+    .. warning::
+      Applying this class requires that no free charges, such as ions or
+      charged fragments, are present in the simulation.
+
     Parameters
     ----------
     atomgroup : MDAnalysis.core.groups.AtomGroup
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -13,7 +13,7 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" http://semver.org
 
 ------------------------------------------------------------------------------
-??/??/?? IAlibay, hmacdope, pillose, jaclark5, tylerjereddy
+??/??/?? IAlibay, hmacdope, pillose, jaclark5, tylerjereddy, PicoCentauri
 
  * 2.7.0
 
@@ -21,6 +21,8 @@ Fixes
   * Fix Atom type guessing error (PR #4168, Issue #4167)
 
 Enhancements
+  * Added a warning about charge neutrality to the documentation of
+    ``DielectricConstant`` (Issue #4262, PR #4263)
 
 Changes
   * NumPy `in1d` replaced with `isin` in preparation for NumPy
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -22,7 +22,7 @@ Fixes
 
 Enhancements
   * Added a warning about charge neutrality to the documentation of
-    ``DielectricConstant`` (Issue #4262, PR #4263)
+    `DielectricConstant` (Issue #4262, PR #4263)
 
 Changes
   * NumPy `in1d` replaced with `isin` in preparation for NumPy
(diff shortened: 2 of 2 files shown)


package/MDAnalysis/analysis/dielectric.py
```diff
@@ -65,6 +65,10 @@ class DielectricConstant(AnalysisBase):
     the usual case if electrostatics are handled with a Ewald summation
     technique. See [Neumann1983]_ for details on the derivation.
 
+    .. warning::
+      Applying this class requires that no free charges, such as ions or
```
IAlibay, line 69: Maybe a stupid question, if this is relying on an Ewald style summation, how well does it handle a system with a net charge?

PicoCentauri: No, totally valid question. The answer will be a bit longer. sorry.

Ewald can handle systems with a net charge. You introduce a homogenous background charge that neutralizes the system, which will lead to an additional force an all atoms. This is unproblematic for homogenous systems but [lead to severe artifacts in inhomogenous systems like membranes etc](http://dx.doi.org/10.1021/ct400626b).

The equation we use here is derived for an Ewald boundary condition, meaning that the dielectric constant of sorrounding medium is a metal ($\varepsilon=\infty$).  If you have other boundary conditions like when using reaction field or cutoff electrostatics the equation will change as described in the paper we cite. We use eq 26.4.3D because the derivation for an Ewald electrostatic leads to the same results as a reaction field with an dielectric constant of $\infty$.

Screenshot 2023-08-28 at 21 39 54

Now, why are charges a problem? The issue is that in the derivation you demand that Maxwell's equation for electric field

$$
\nabla \vec E = \frac{\rho}{\varepsilon_0}
$$

where $\rho$ is the charge density is equal to zero: $\nabla \vec E = 0$. This means that no free charges are present. If you do not do this you will not end up with the equation we use.

Does this answer your question?

hmacdope: Do you think you could transfer this very insightful piece of writing to an issue for future reference. :)

package/MDAnalysis/analysis/dielectric.py
```diff
@@ -65,6 +65,10 @@ class DielectricConstant(AnalysisBase):
...
+      charged fragments, are present in the simulation.
```
hmacdope, line 70: @PicoCentauri  perhaps I am misunderstanding but you can have no  charged anything or is it that you need overall neutrality. Apologoies if silly q.

package/MDAnalysis/analysis/dielectric.py
PicoCentauri: Yes we sure. If people find it there.

package/MDAnalysis/analysis/dielectric.py
PicoCentauri: Nono it is fine. This is really a bit tricky. There are even publications where people do this wrong.

As we do it in the assertion in the class' prepare function you need charge neutral molecules. In a classical MD world they can have of course have partial charges. In fact, they have to because otherwise the system's total dipole moment would be zero and with it the dielectric constant.

What you can not have are ions such as Na+ Cl- or other charged fragments i.e negative carboxylic acid groups COO- or positive amine groups H3N+.

hmacdope (approved):
Im happy with the warning as is, once we resolve conflicts I think we can merge and address better documentation in #4262.

IAlibay (approved):
LGTM. Apologies I jumped in here and fixed the changelog conflicts. (no need to wait for CI again, this is a docs change anyways)

PR #4265, opened 2023-08-28
Try 3.12.0-rc1

IAlibay:
Fixes #

 - 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

----
:books: Documentation preview :books:: https://mdanalysis--4265.org.readthedocs.build/en/4265/


IAlibay:
Ok we're blocked by NumPy funnily enough, they have a pin on a too old version of setuptools for 1.25.2 (it's a weird world).

package/pyproject.toml
```diff
@@ -17,7 +17,7 @@ requires = [
     # For unreleased versions of Python there is currently no known supported
     # NumPy version. In that case we just let it be a bare NumPy install
     "numpy<2.0; python_version>='3.12'",
-    "setuptools",
+    "setuptools>=68.0",
```
tylerjereddy, line 21: > Ok we're blocked by NumPy funnily enough, they have a pin on a too old version of setuptools for 1.25.2 (it's a weird world).

That doesn't sound right, the first NumPy version to support Python `3.12.0rc1` is the beta release `1.26.0b1` I used to build SciPy wheels, which should drop `setuptools` for `meson`.

Two tricky things to consider that should help out:

1. `3.12` won't pickup the release candidate in `pyproject.toml` logic so you'll need: https://github.com/numpy/numpy/issues/23808#issuecomment-1678677094
2. You may also need `PIP_PRE` per: https://github.com/numpy/numpy/issues/23808#issuecomment-1677596759

Anyway, hope that helps, but NumPy shouldn't be holding you back, should just be a matter of the tweaks mentioned there.

tylerjereddy: That's what I did for example: https://github.com/scipy/scipy/blob/maintenance/1.11.x/pyproject.toml#L50

(actually I'm not sure I should have hard pinned like that, but anyway it worked)

tylerjereddy: And `PIP_PRE` is set here: https://github.com/scipy/scipy/blob/maintenance/1.11.x/.github/workflows/wheels.yml#L136

(using that extra index shouldn't be needed anymore though I don't think since the NumPy beta is on PyPI)

IAlibay: Thanks @tylerjereddy that's super useful info!

I think here I was mostly looking to see "what bits of upstream are we waiting on for py3.12", rather than adding actual shims to get py3.12 to work.

What I might do is stick the py3.12rc under the nightly wheels cron checks (which turns off build isolation) and see if it works there. If so, we can revisit adding it to normal CI checks when numpy 1.26 comes out.

Edit: nevermind, the nightly wheels cron checks are a full dependency check, there's no way that would work, not with rdkit and openmm in that list.

IAlibay:
Cycling for CI now that 1.26 is out.

.github/workflows/gh-ci-cron.yaml
```diff
@@ -184,7 +187,7 @@ jobs:
     strategy:
         fail-fast: false
         matrix:
-          python-version: ["3.10", "3.11"]
+          python-version: ["3.10", "3.11", "3.12.0-rc.1"]
```
IAlibay:
```suggestion
          python-version: ["3.10", "3.11", "3.12.0-rc.3"]
```

package/pyproject.toml
```diff
@@ -17,7 +17,7 @@ requires = [
     # For unreleased versions of Python there is currently no known supported
     # NumPy version. In that case we just let it be a bare NumPy install
     "numpy<2.0; python_version>='3.12'",
```
IAlibay:
```suggestion
     "numpy<2.0; python_version>='3.12'",
```

```suggestion
     "numpy<2.0; python_version>='3.13'",
```

package/pyproject.toml
```diff
@@ -17,7 +17,7 @@ requires = [
     # For unreleased versions of Python there is currently no known supported
```
IAlibay:
```suggestion
    "numpy==1.26.0; python_version=='3.12' and platform_python_implementation != 'PyPy'"
     # For unreleased versions of Python there is currently no known supported
```

package/pyproject.toml
```diff
@@ -14,10 +14,11 @@ requires = [
     "numpy==1.22.3; python_version=='3.9' and platform_python_implementation != 'PyPy'",
     "numpy==1.22.3; python_version=='3.10' and platform_python_implementation != 'PyPy'",
     "numpy==1.23.2; python_version=='3.11' and platform_python_implementation != 'PyPy'",
-    # For unreleased versions of Python there is currently no known supported
+    "numpy==1.26.0; python_version=='3.12' and platform_python_implementation != 'PyPy'"
```
IAlibay:
```suggestion
    "numpy==1.26.0; python_version=='3.12' and platform_python_implementation != 'PyPy'",
```

IAlibay:
Ok there's a lot of issues to digest and deal with, Py312 is going to be interesting...

tylerjereddy:
If there is anything upstream that causes friction feel free to ping me to look at it.

IAlibay:
> If there is anything upstream that causes friction feel free to ping me to look at it.

Thanks @tylerjereddy - I think for the most part we're only just struggling with some cpython changes that are catching out a few things, primarily `pickle`.

Issue #3927, opened 2022-11-21
Return distance results as `float` rather than `double` for 3.0 or make return type flexible.


hmacdope:
## Is your feature request related to a problem? ##

 
As detailed in #3914 and observed other places, the need to mix precision when computing distances makes the `mda.lib.distances` python and C++ layers slow. It also requires an ugly shim to get distopia to work properly and IMO does not make heaps of sense with `float` coordinates to start with. 

## Describe the solution you'd like ##

We could return `float` results or have the return type be flexible, Ie make no promises about the return type. 

## Describe alternatives you've considered ##

 Leave everything the way it is. 

Leave your thoughts @MDAnalysis/coredevs as would be a breaking change.

richardjgowers:
Ideally a `Universe` has a dtype, and all coordinates are of that type.  Currently we're float/single for coordinates, boxes are double and then distances are returned as doubles, it's a bit inconsistent.

orbeckst:
I'd say a dtype makes sense for a homogenous data structure like an array. For something heterogenous like a Universe it's a "nice to have". I think it's more important to think about loss of precision when you're working with the data than to make promises up-front. For instance, it may make a lot of sense to use double accumulator variables when aggregating float coordinates.

orbeckst:
The performance improvements that you show in https://github.com/MDAnalysis/mdanalysis/pull/3914#issuecomment-1321395720 make a strong case to give up the guarantee of double precision output.

Admittedly, after reading the other thread I understand better why @richardjgowers suggests
> Ideally a Universe has a dtype, and all coordinates are of that type. Currently we're float/single for coordinates, boxes are double and then distances are returned as doubles, it's a bit inconsistent.

as this situation would avoid any casting in between. 

I assume there will still be situations where we would want/need to use mixed precision. 

Do we make boxes double because that's what we get from trajectories or because we need it for, e.g., PBC stuff?

hmacdope:
My understanding is that the **actual** precision of coordinates as stored in the trajectory files varies wildly ie XTC it is less accurate than a `float` due to compression while a DCD I think reads coords as `float` and a box as `double` from what I can see. 

There was also some arguments made that box ops should be done in double precision (I remember seeing these, I'll try and find them again). But I think this is a question of accuracy, i.e  my tests  done in distopia indicate you can get 5 sig fig accuracy with  all floats anyway.

PR #4273, opened 2023-09-02
update qcprot to cython.floating types

richardjgowers:
use memoryviews throughout for potential performance

allows either float32 or float64 inputs

Related #3927 

 - qcprot functions take either float32 or float64 inputs
 - modernised qcprot to use memoryviews

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?

----
:books: Documentation preview :books:: https://mdanalysis--4273.org.readthedocs.build/en/4273/


richardjgowers:
Still need to do some tests here.  `rms.RMSD` looks like it's converting/copying everything up to float64, but I imagine that not forcing this would be some sort of precision-regression.  Instead I can probably write some tests that using `lib.qcprot` directly with floats/doubles work as expected and we can have some future discussion about precision and dtypes in the future.

IAlibay (approved):
This looks reasonable to me. Is there an associated issue this is working towards? (edit: I know there were a few out there that had been opened about fp precision).

richardjgowers:
@IAlibay I've tagged a related issue, I think we should be moving towards numpy-like behaviour where output dtype matches input dtype and either float or double are allowed.  A future discussion is the next major release of MDA will probably have a single precision **option**, since we're able to get 2x performance & 1/2 memory footprint for this.

The memoryview change is just a best-practice and we should gradually modernise to this as it's documented as the recommended route for best performance.

package/CHANGELOG
```diff
@@ -13,7 +13,7 @@ The rules for this file:
   * release numbers follow "Semantic Versioning" http://semver.org
 
 ------------------------------------------------------------------------------
-??/??/?? IAlibay, ianmkenney, PicoCentauri
+??/??/?? IAlibay, ianmkenney, PicoCentauri, richardjgowers
```
IAlibay: Odd, I thought you had conttributed in 2.6 for the GSD stuff, maybe that was 2.5... time flies by.

package/MDAnalysis/lib/qcprot.pyx
```diff
@@ -494,7 +509,7 @@ def FastCalcRMSDAndRotation(np.ndarray[np.float64_t, ndim=1] rot,
                     rot[0] = rot[4] = rot[8] = 1.0
                     rot[1] = rot[2] = rot[3] = rot[5] = rot[6] = rot[7] = 0.0
 
-                    return
```
richardjgowers, line 497: this branch looks like it was a bug.  at this point the RMSD has been calculated and the code has failed to determine a suitable rotation matrix, returning None doesn't seem to make sense

richardjgowers: e.g. compared to this implementation in BioPython: https://github.com/biopython/biopython/blob/master/Bio/PDB/qcprot.py#L190

package/MDAnalysis/lib/qcprot.pyx
```diff
@@ -174,27 +176,25 @@ cdef extern from "math.h":
...
-def InnerProduct(np.ndarray[np.float64_t, ndim=1] A,
-                 np.ndarray[np.float64_t, ndim=2] coords1,
-                 np.ndarray[np.float64_t, ndim=2] coords2,
-                 int N,
-                 np.ndarray[np.float64_t, ndim=1] weight):
+cpdef cython.floating InnerProduct(cython.floating[:] A,
+                                   cython.floating[:, :] coords1,
+                                   cython.floating[:, :] coords2,
+                                   int N,
+                                   cython.floating[:] weight):
```
tylerjereddy, line 183: I remember that a while back I tried to improve performance in SciPy by switching from the old NumPy buffer syntax, to memoryviews, and it actually caused a performance regression: https://github.com/scipy/scipy/issues/12379#issuecomment-645934588

Anyway, you may want to measure that you remain at least performance neutral, my experience with following that best practice hasn't historically been great, but hopefully the situation is better now.

testsuite/MDAnalysisTests/lib/test_qcprot.py
```diff
@@ -0,0 +1,67 @@
...
+        r = qcprot.CalcRMSDRotationalMatrix(p1, p2, len(prot), rot, weights)
+
+        rot_ref = np.array([0.999998, 0.001696, 0.001004,
+                            -0.001698,  0.999998,  0.001373,
+                            -0.001002, -0.001375,  0.999999],
+                           dtype=dtype)
+
+        err = 0.001 if dtype is np.float32 else 0.000001
+        assert r == pytest.approx(0.6057544485785074, abs=err)
+        assert_array_almost_equal(rot, rot_ref)
```
tylerjereddy: Maybe use `assert_allclose` for new test code per the `Note` at https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_almost_equal.html ?

orbeckst (changes requested):
The RMSD functions are quite time-critical for many applications so if there's any chance for a performance impact then please provide some benchmark data before merge. Thanks!

orbeckst:
About float vs double. I remember vaguely that during one of the REU projects we played around with single vs double precision and arrived at the conclusion that float32 is insufficient for the algorithm to properly converge. Of course, I don't remember any details and have no idea if there are any notes anywhere. Primarily I want to say that it simply might not be as easy to say that we just run the algorithm at whatever dtype the rest of the code wants to run and that this will require careful testing.

richardjgowers:
@orbeckst benchmarks
```python
import MDAnalysis as mda
from MDAnalysis.lib import qcprot
import numpy as np

u = mda.Universe(PSF, DCD)

prot = u.select_atoms('protein')

w = (prot.masses / np.mean(prot.masses)).astype(np.float64)

ref = prot.positions.astype(np.float64)
u.trajectory[1]
conf = prot.positions.astype(np.float64)

rot = np.zeros(9, dtype=np.float64)
A = np.zeros(9)

N = len(prot)
```
then:
```python
%%timeit

qcprot.CalcRMSDRotationalMatrix(ref, conf, len(prot), rot, w)
```
old code:
`10.1 µs ± 46.4 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)`
new code:
`9.97 µs ± 29.1 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)`

The number fluctuate a bit, I think there's no performance gain/loss.

I was able to reproduce the memoryview regression that @tylerjereddy flagged up, but I think it's an order of magnitude smaller than the numbers measured here, so it won't show up.  It's something to keep in mind if we're writing code that has a lot of function calls (there's 3 here).

re: precision, yes I had to make the qcp function use doubles throughout to keep it working.  The `InnerProduct` can use either float or double now.  Strangely these take the same amount of time, which means the compiler isn't able to do anything clever here (where it ought to be) so probably some future digging.  This does some ground work at least.

orbeckst (approved):

PR #4276, opened 2023-09-04
Switch docs theme

### LINKED ISSUE #33: Start using MDA theme in official repos
I expect this to require changes to the theme -- opened #32.

Repos:
- [x] core: https://github.com/lilyminium/mdanalysis/tree/switch-docs-theme
- [x] user guide: https://github.com/MDAnalysis/UserGuide/tree/switch-docs-theme
- [x] mdakits registry
- [x] mdakit cookiecutter https://github.com/MDAnalysis/cookiecutter-mdakit/pull/85
- mdacli: https://github.com/MDAnalysis/mdacli/tree/switch-docs-theme
- transport-analysis
- solvation-analysis
- [x] griddataformats: https://github.com/MDAnalysis/GridDataFormats/tree/switch-docs-theme
- [x] hole2-mdakit
- [ ] distopia https://github.com/MDAnalysis/distopia/pull/146
- [x] pytng https://github.com/MDAnalysis/pytng/pull/109
- [x] membrane-curvature https://github.com/MDAnalysis/membrane-curvature/pull/122
- mdanalysisdata
- pmda

### LINKED ISSUE #4199: Docs need updating to modern docutils
## Expected behavior ##

We should aspire for our documentation to be built with the latest, or near latest, versions of packages.

In #4177 and #4194 the documentation appears to have broken due to our `docutils` updating from 0.17.1 to 0.18.1. #4196 pins docutils<0.18 but this should be a temporary solution at best. docutils 0.18 actually came out at the end of 2021 and the package is now working on 0.21, so we should look at upgrading.

https://github.com/MDAnalysis/mdanalysis/commit/a963edc5f403d1a8244f2c7f8b913f06fcaf951f#diff-e0e12b88089dbc2e97e54366222048d8d05e4a7ef5efe67f0203a2560f301a72 is a diff of the differences -- it looks like docutils adds more classes and punctuation. The JS has also changed.

We have custom punctuation appended to fields in `custom.css`, so that's a good place to start fixing things.

lilyminium:
Fixes https://github.com/MDAnalysis/mdanalysis-sphinx-theme/issues/33
Fixes #4199 

 - 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

----
:books: Documentation preview :books:: https://mdanalysis--4276.org.readthedocs.build/en/4276/


.github/workflows/gh-ci.yaml
```diff
@@ -155,7 +155,7 @@ jobs:
       with:
         micromamba: true
         full-deps: true
-        extra-pip-deps: "docutils<0.18 sphinx<7.0 sphinx-sitemap sphinx_rtd_theme msmb_theme==1.2.0 sphinxcontrib-bibtex pybtex pybtex-docutils"
+        extra-pip-deps: "docutils<0.18 sphinx<7.0 sphinx-sitemap mdanalysis-sphinx-theme sphinxcontrib-bibtex pybtex pybtex-docutils"
```
IAlibay: Can we remove the docutils & sphinx pins now?

IAlibay (commented):
Thanks @lilyminium, awesome work!

.github/workflows/gh-ci.yaml
IAlibay: I guess I commented too quickly :P

lilyminium:
On ReadTheDocs, we have some remnant blue boxes:

Screenshot 2023-09-04 at 9 29 36 pm


On my machine (TM), we do not:
Screenshot 2023-09-04 at 9 30 15 pm

Moving to draft status while I figure it out...

Edit: Ok, I can reproduce this if I use the same bewilderingly old version of sphinx_rtd_theme (0.4.3) that RTD is using.

IAlibay:
> Ok, I can reproduce this if I use the same bewilderingly old version of sphinx_rtd_theme (0.4.3) that RTD is using.

Is this one of those where we can just say we don't care and it'll go away soon or is that being too hopeful?

lilyminium:
> Is this one of those where we can just say we don't care and it'll go away soon or is that being too hopeful?

I could maybe pin sphinx_rtd_theme >=1 with mdanalysis_sphinx_theme to resolve this? (or would RTD just ignore that version and go for the more permissive variant?) It is a very weirdly old version (released 2019, sphinx_rtd_theme is on 2.0.0rc2 atm) and I'm curious why RTD has picked it. It'd be a bit annoying if we couldn't preview our docs properly there.

User guide is picking up the same 0.4.3...

IAlibay:
> > Is this one of those where we can just say we don't care and it'll go away soon or is that being too hopeful?
> 
> I could maybe pin sphinx_rtd_theme >=1 with mdanalysis_sphinx_theme to resolve this? (or would RTD just ignore that version and go for the more permissive variant?) It is a very weirdly old version (released 2019, sphinx_rtd_theme is on 2.0.0rc2 atm) and I'm curious why RTD has picked it. It'd be a bit annoying if we couldn't preview our docs properly there.

I guess you could try just override the version in the yaml file here for now and see what happens? If that doesn't work then they are doing some annoying black magic in the background that is unlikely get fixed with pinning upstream.

IAlibay:
```
 + sphinx_rtd_theme                      1.3.0  pyha770c72_0        conda-forge/noarch         2 MB
```
Seems to have done a thing?

TIL RTD actually advocates for pinning the sphinx_rtd_theme version: https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html#id2

lilyminium:
Oh nice! I was just tracking stuff down myself, since everywhere seemed to be picking up the 0.4.3 version. I *think* it might be preferring anaconda/noarch (https://anaconda.org/anaconda/sphinx_rtd_theme), which tops out at 0.4.3, over conda-forge (https://anaconda.org/conda-forge/sphinx_rtd_theme).
So we could either pin the lower bound of sphinx_rtd_theme, or keep pinning every version as they release (a minor pain...), and I assume as long as it's above a certain version it should pick up the conda-forge one? Do you have any preference -- I know you generally prefer as few pins as possible?

Edit: whoops, I should have let RTD finish building before pushing the rebase :/ sorry for the wait

IAlibay:
> So we could either pin the lower bound of sphinx_rtd_theme, or keep pinning every version as they release (a minor pain...), and I assume as long as it's above a certain version it should pick up the conda-forge one? Do you have any preference -- I know you generally prefer as few pins as possible?

Less work is better work. I would just go with the lower bound pin? Would it make sense to just have the pin upstream in the new theme? It's annoying to add a dependency, but it would probably save you some headaches later?

lilyminium:
Yeah I'll pin upstream.

orbeckst (commented):
This looks super-nice already on RTD. Great work! I look forward to seeing it merged.

lilyminium:
Thanks! I think this is ready for review now. The theme fixes a few small annoying issues, plus the docutils incompatibility. There are also minor docs fixes I noticed from previewing docs.

**Old**
Screenshot 2023-09-05 at 11 00 14 am

**New**
Screenshot 2023-09-05 at 11 01 01 am

package/MDAnalysis/analysis/bat.py
```diff
@@ -31,11 +31,11 @@
 
 This module contains classes for interconverting between Cartesian and an
 internal coordinate system, Bond-Angle-Torsion (BAT) coordinates
-:cite:p:`Chang2003`, for a given set of atoms or residues. This coordinate
+:footcite:p:`Chang2003`, for a given set of atoms or residues. This coordinate
```
IAlibay, line 34: Do we document how we do citations anywhere in our docs? We probably should tell folks that cite -> footcite now

IAlibay: Are there any peculiarities to footbibliography that should be stated somewhere? (I don't know how/if it differs from the old approach)

IAlibay (approved):
Could you tick the lgplv2 box?

Otherwise, lgtm! This is really awesome, thanks @lilyminium ! 🎉

package/MDAnalysis/analysis/bat.py
lilyminium: Ticked :)

I don't think so, no more than `bibliography` is. The footbib is used for local bibliographies (on the same page as the reference), so we can remove stuff like the `a-` and `b-` prefixes we had before. The normal bibliography is for global bibliographies, so one reference list for the entire repo. This is all readily available on the sphinx bibtex extension docs though. I was planning to add in some notes in the mdanalysis-sphinx-theme -- https://github.com/MDAnalysis/mdanalysis-sphinx-theme/issues/26

The one catch is that there's meant to be one footbib per reference per page, so if footbibliographies are created multiple times for the same reference, then only the first one will ahve the reference and all links will go there. Again I got this from the official docs.

IAlibay: Could you maybe raise an issue in the userguide for docs on this? This is the kind of info any contributor new or old should be told to look at so imho, even if it's just a reference to the theme docs, your above explanation should exist there in some way.

lilyminium: Sure, done

PR #4278, opened 2023-09-04
Switch away from build_sphinx

### LINKED ISSUE #4277: Move on from build_sphinx
## Problem ##

I think sphinx and setuptools integration is gone from 7 onwards -- see failures in https://github.com/MDAnalysis/mdanalysis/pull/4276

I didn't think sphinx_rtd_theme supported 7.0 anyway but they should do from the next release, so let's fix this now than later.

lilyminium:
Fixes #4277 

 - change `python setup.py build_sphinx` to just run sphinx straight. This is less elegant since we *have* to specify the source and output directories.

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

----
:books: Documentation preview :books:: https://mdanalysis--4278.org.readthedocs.build/en/4278/

### DIFF
--- a/.github/actions/build-src/action.yaml
+++ b/.github/actions/build-src/action.yaml
@@ -88,4 +88,6 @@ runs:
       if: ${{ inputs.build-docs == 'true' }}
       shell: bash -l {0}
       run: |
-        cd package/doc && python -m sphinx -T -E -b html -D language=en sphinx/source html
+        SOURCEDIR="sphinx/source"
+        BUILDDIR="html"
+        cd package/doc && python -m sphinx -T -E -b html -D language=en ${SOURCEDIR} ${BUILDDIR}
--- a/.github/actions/build-src/action.yaml
+++ b/.github/actions/build-src/action.yaml
@@ -90,4 +90,4 @@ runs:
       run: |
         SOURCEDIR="sphinx/source"
         BUILDDIR="html"
-        cd package/doc && python -m sphinx -T -E -b html -D language=en ${SOURCEDIR} ${BUILDDIR}
+        cd package/doc && python -m sphinx -T -E --keep-going -b html -D language=en ${SOURCEDIR} ${BUILDDIR}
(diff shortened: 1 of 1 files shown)


lilyminium:
I could be very much convinced to use `make html` instead.

.github/actions/build-src/action.yaml
```diff
@@ -88,4 +88,6 @@ runs:
       if: ${{ inputs.build-docs == 'true' }}
       shell: bash -l {0}
       run: |
-        cd package && python setup.py build_sphinx -E --keep-going
+        SOURCEDIR="sphinx/source"
+        BUILDDIR="html"
+        cd package/doc && python -m sphinx -T -E -b html -D language=en ${SOURCEDIR} ${BUILDDIR}
```
IAlibay: do we not want to keep `--keep-going`?

lilyminium: Sure, added it back in.

IAlibay (commented):
I think we're now edit: finding a bit of fragmentation across the package in terms of "which command do we use to build doc things" - https://github.com/MDAnalysis/mdanalysis/blob/develop/.github/workflows/gh-ci.yaml#L171 ?

Should one or the other be updated?

lilyminium:
I think `sphinx` is just a wrapper around `sphinx-build`, can update to that if it's clearer. I'd keep the options though, they give full tracebacks (-T) and always read all files (-E).

IAlibay:
> I think `sphinx` is just a wrapper around `sphinx-build`, can update to that if it's clearer. I'd keep the options though, they give full tracebacks (-T) and always read all files (-E).

Yeah options are fine, I just want to make sure folks have a "single command of truth" if that makes sense - i.e. we just have to document the one method.

IAlibay (approved):

IAlibay:
No need to wait for rest of CI - if it builds it builds

PR #4283, opened 2023-09-06
Fix charge reading from PDBQTParser

### LINKED ISSUE #4282: PDBQTParser.parse() assigns wrong field to charges
Related to #4207 .

On page 27 from the [PDBQT reference](https://autodock.scripps.edu/wp-content/uploads/sites/56/2021/10/AutoDock4.2.6_UserGuide.pdf) (the most recent reference I could find for the PDBQT format), the PDBQT ATOM register is specified with this C-style string:

```
“ATOM%5d%-4s%1s%-3s%1s%4d%1s%8.3f%8.3f%8.3f%6.2f%6.2f%4s%6.3f%2s \n",
atom_serial_num, atom_name, alt_loc, res_name, chain_id, res_num, ins_code, x, y, z,
occupancy, temp_factor, footnote, partial_charge, atom_type
```
Going backwards from the end, we see that the partial charge takes up 6 columns and then there're 4 columns for a string called 'footnote'. The [current implementation](https://github.com/MDAnalysis/mdanalysis/blob/a1aff6638184a90ae36b879720b2fa78dee1d1ee/package/MDAnalysis/topology/PDBQTParser.py#L143C1-L144C1) of `PDBQTParser.parse()` tries to cast those 10 columns as a float, which may end up giving errors if someone decided to use those columns.

pgbarletta:
Fixes #4282

 - Columns 67 to 70 from PDBQT files are ignored instead of trying to parse them as atom charges.

PR Checklist
------------
 - [x] Tests?
 - [X] Docs?
 - [X] CHANGELOG updated?

----
:books: Documentation preview :books:: https://mdanalysis--4283.org.readthedocs.build/en/4283/


richardjgowers (approved):
thanks @pgbarletta this looks like a good catch.  If you have time adding a test (or updating whatever pdbqt file we currently have) to include a remark in cols 67-70 would make sure this behaviour is tested

orbeckst (changes requested):
Please add a test that fails without the fix and passes with it.

Please also include a link to the format document in the docs. The more we can say very clearly which format we are implementing the better.

Finally, add a `versionchanged` (take into account the "footnote" field in coulmns ....) because it's possible that the standard changed without us knowing and then we want to have a record of when we changed.

package/MDAnalysis/topology/PDBQTParser.py
```diff
@@ -98,11 +98,16 @@ class PDBQTParser(TopologyReaderBase):
...
      - masses
 
     .. versionchanged:: 0.18.0
        Added parsing of Record types
+    .. versionchanged:: 2.7.0
+       Columns 67 - 70 in ATOM records, corresponding to the field *footnote*,
+       are now ignored. See Autodock's `reference`_.
+
+       .. _reference: https://autodock.scripps.edu/wp-content/uploads/sites/\
+       56/2021/10/AutoDock4.2.6_UserGuide.pdf
```
orbeckst: It's ok to sometimes let long URLs hang over — that's cleaner than backslash-breaking
```suggestion
       .. _reference: 
          https://autodock.scripps.edu/wp-content/uploads/sites/56/2021/10/AutoDock4.2.6_UserGuide.pdf
```

orbeckst (approved):
Thanks for your changes. All looking good (will just make a minor reformatting change).

I am very sorry, the review slipped through. As always, please feel free to ping me if there's no action in a few days.

PR #4284, opened 2023-09-06
Read chainID from PDBQT

### LINKED ISSUE #4207: Error in PDBQT to PDB conversion
## Expected behavior ##

I am trying to convert a PDBQT file of a prepared protein to the PDB format using MDAnalysis. I expect that the conversion process will retain the correct atom types without any issues, ensuring that the resulting PDB file represents the original structure accurately.

## Actual behavior ##

During the conversion from PDBQT to PDB format, I encountered an unexpected behaviour. The atom types are being incorrectly assigned as the chain ID instead of retaining their actual atom types. As a result, the resulting PDB file does not accurately represent the original structure, leading to errors in further downstream processing, such as docking simulations using AutoDock Vina.

``` python
import MDAnalysis as mda

def pdbqt_to_pdb(pdbqt_file, pdb_file):
    # Load the .pdbqt file
    u = mda.Universe(pdbqt_file)

    # Save the structure in .pdb format
    u.atoms.write(pdb_file)

```

- MDAnalysis version: 2.3.0
- Python version: 3.9.4
- Operating system: Linux Ubuntu 22.04.2 LTS

## Additional Notes ## 

I have consulted the MDAnalysis documentation in an attempt to find a solution, but I could not identify any direct method or tutorial addressing this specific conversion issue. The incorrect assignment of atom types to the chain ID in the resulting PDB file hinders my ability to proceed with my research on docking simulations.

I would greatly appreciate your help in resolving this bug or providing some guidance on how to correctly perform the PDBQT to PDB conversion using MDAnalysis. Your expertise and support would be invaluable in ensuring the accurate representation of molecular structures for further analysis.

Thank you for your attention to this matter, and I look forward to hearing back from you.

Sincerely,

Rita 
@RortegaPQ

pgbarletta:
Fixes #4207

The user also complained about lacking elements, but given that the PDBQT format replaced the `element` field from the PDB `ATOM` record, I guess is fair that we don't try to create the from the `atomtype`. In any case, the solution given to the user could be made a FAQ (if it's indeed a FAQ).

 - Use the chain ID read from the atoms to create the proper `chainIDs` attribute instead of just discarding them.

PR Checklist
------------
 - [x] Tests?
 - [X] Docs?
 - [X] CHANGELOG updated?

----
:books: Documentation preview :books:: https://mdanalysis--4284.org.readthedocs.build/en/4284/

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -33,6 +33,7 @@ Enhancements
   * Added a warning about charge neutrality to the documentation of
     `DielectricConstant` (Issue #4262, PR #4263)
   * Add support for reading chainID info from prmtop amber topologies (PR #4007)
+  * Add support for reading chainID info from Autodock PDBQT files (PR #4284)
 
 Changes
   * The `mda-xdrlib` module is now a core dependency of MDAnalysis
--- a/package/MDAnalysis/topology/PDBQTParser.py
+++ b/package/MDAnalysis/topology/PDBQTParser.py
@@ -75,6 +75,7 @@ from ..core.topologyattrs import (
     Resnums,
     Resnames,
     Segids,
+    ChainIDs,
     Tempfactors,
 )
 
@@ -88,7 +89,7 @@ class PDBQTParser(TopologyReaderBase):
      - atom names
      - altLocs
      - resnames
-     - chainIDs (becomes segid)
+     - chainIDs (assigned to segid as well)
      - resids
      - record_types (ATOM/HETATM)
      - icodes
@@ -165,6 +166,8 @@ class PDBQTParser(TopologyReaderBase):
         icodes = np.array(icodes, dtype=object)
         resnames = np.array(resnames, dtype=object)
         chainids = np.array(chainids, dtype=object)
+
+        attrs.append(ChainIDs(chainids))
 
         residx, (resids, icodes, resnames, chainids) = change_squash(
             (resids, icodes), (resids, icodes, resnames, chainids))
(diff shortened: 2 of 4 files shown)


lilyminium (changes requested):
Thanks @pgbarletta for opening this PR! I assume the PDBQT format doesn't include a segment column, or else that would be preferenced for segids? I was worried about what the PDBQT writer would do if the chainIDs and segids conflict -- looks like it tries to write chainIDs first, and then uses the last (not first...) letter of the segids.

There's a test for the segid attribute:
https://github.com/MDAnalysis/mdanalysis/blob/16647cb4934950d01c7a67b70c99858da48a9237/testsuite/MDAnalysisTests/coordinates/test_pdbqt.py#L41-L45

Would it be possible to add one for chainIDs too? Otherwise LGTM.

pgbarletta:
>  I assume the PDBQT format doesn't include a segment column, or else that would be preferenced for segids? 

PDBQT has the concept of `chainID`, but not of segments. Taken from the [PDBQT reference](https://autodock.scripps.edu/wp-content/uploads/sites/56/2021/10/AutoDock4.2.6_UserGuide.pdf):

```
ATOM%5d%-4s%1s%-3s%1s%4d%1s%8.3f%8.3f%8.3f%6.2f%6.2f%4s%6.3f%2s \n",
atom_serial_num, atom_name, alt_loc, res_name, chain_id, res_num, ins_code, x, y, z,
occupancy, temp_factor, footnote, partial_charge, atom_type
```

I've seen the `occupancy` and `temp_factor` columns used for `vdW` and `Elec`, but never the `footnote` for segments, though I don't have that much experience with this format.

> Would it be possible to add one for chainIDs too?

Done. Didn't add it at first since both `segid` and `chainID` read the same info.

lilyminium (approved):
Thanks @pgbarletta! Everything LGTM!

orbeckst:
ALong the lines of https://github.com/MDAnalysis/UserGuide/issues/312 — please (1) raise an issue to document in the User Guide and then (2) do the documenting.

orbeckst:
@lilyminium I tentatively assigned you to merge because you reviewed but please feel free to punt.

PR #4300, opened 2023-09-24
MAINT: Python 3.12 support improve

### LINKED ISSUE #4299: Python 3.12 cannot pickle PickleBufferIOPicklable instances
This is one of the biggest issues we're facing with Py3.12:

https://github.com/MDAnalysis/mdanalysis/actions/runs/6277097391/job/17048178820?pr=4265#step:8:375

As far as I can tell there's isn't anything super obvious in the changelog that would cause this.

tylerjereddy:
* Fixes gh-4299, and allows the testsuite to have 12 failures instead of 56 with Python `3.12.0rc3` for me locally on x86_64 Linux; full suite still passing for Python `3.11.x`

* the main idea is to circumvent more aggressive CPython blocks on serialization by overriding the highest-priority pickling method available, `__reduce_ex__`, which CPython had locked down, making it such that our old `__getstate__` pickling shims were ignored as indicators of pickling safety

* there is probably a slightly simpler diff that allows this to work, but this took a few hours to draft so feel free to push in simplifications if you're sure you've tested on 3.11.x and 3.12 RC

* also, I believe this does slightly increase the chance of hanging the testsuite at very high levels of concurrency, but I needed more than 30 processes to repro that, so it may just be a feature of our testsuite in general (it was fine on 12 cores at least, but on 32 or so file locks had issues I think)

[skip cirrus]

----
:books: Documentation preview :books:: https://mdanalysis--4300.org.readthedocs.build/en/4300/

### DIFF
--- a/package/MDAnalysis/lib/picklable_file_io.py
+++ b/package/MDAnalysis/lib/picklable_file_io.py
@@ -109,16 +109,23 @@ class FileIOPicklable(io.FileIO):
         self._mode = mode
         super().__init__(name, mode)
 
-    def __getstate__(self):
+
+    def __setstate__(self, state):
+        name = state["name_val"]
+        super().__init__(name, mode='r')
+        try:
+            self.seek(state["tell_val"])
+        except KeyError:
+            pass
+
+    def __reduce_ex__(self, prot):
         if self._mode != 'r':
             raise RuntimeError("Can only pickle files that were opened "
                                "in read mode, not {}".format(self._mode))
-        return self.name, self.tell()
-
-    def __setstate__(self, args):
-        name = args[0]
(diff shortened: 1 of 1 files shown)


IAlibay:
Thanks @tylerjereddy - should we temporarily merge my Py3.12 PR knowing it fails and then just turn off the cron shim to test here?

tylerjereddy:
We could I suppose. Do we even need a cron guard? The final CPython `3.12` release is about a week or two away at most right, so maybe not too long to ignore a failing entry while we try to fix it up? I guess we may not plan support quite that fast though, in which case `cron` could be fine yeah.

richardjgowers (approved):

PR #4309, opened 2023-10-02
MAINT: mkdtemp 3.12 support

tylerjereddy:
* force the use of a relative path for `mkdtemp` in our source code because, from the Python `3.12` release notes:

> tempfile.mkdtemp() now always returns an absolute path, even if the
argument provided to the dir parameter is a relative path.

* with this patch, the full testsuite now passes on (x86_64 Linux) Python `3.12.0` (which is now out) when combined with gh-4301 and gh-4300 (though see my cautions about thread safety/concurrency for the latter)

* this fixes the 1 remaining failure with 3.12: `MDAnalysisTests/analysis/test_hole2.py::TestCheckAndFixLongFilename::test_symlink_dir`

[skip cirrus]

----
:books: Documentation preview :books:: https://mdanalysis--4309.org.readthedocs.build/en/4309/

### DIFF
--- a/package/MDAnalysis/analysis/hole2/utils.py
+++ b/package/MDAnalysis/analysis/hole2/utils.py
@@ -125,7 +125,7 @@ def check_and_fix_long_filename(filename, tmpdir=os.path.curdir,
     if make_symlink:
         # shorten path by creating a symlink inside a safe temp dir
         _, ext = os.path.splitext(filename)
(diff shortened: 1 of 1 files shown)


IAlibay:
Thanks @tylerjereddy !

I'll try to get the py3.12 CI PR updated tomorrow and we can start merging in all your PRs.

richardjgowers (approved):

IAlibay (approved):

Issue #1954, opened 2018-06-23
Topology readers are not aware of virtual sites


jbarnoud:
Virtual sites are atoms the coordinates of which are built frame by frame from the coordinates of other atoms. MDAnalysis reads the coordinates of these atoms as for any other ones. Yet, it does not read the connection between the virtual sites and the atoms they are built from. Therefore, the virtual sites appear as not connected to the rest of the molecule which breaks the expected behaviour of `ag.fragments` and `make_whole`.

A solution would be to have a `VirtualSites` topology attribute that connects the relevant atoms. With the relationship described, the appropriate edges can be added to the graph used by `ag.fragments` and `make_whole` can translate the virtual sites with the building atoms. The difficulty is that virtual sites can be built from any number of atoms, yet from my tests, `TopologyGroups` store the indices as an array of ints. I do not know if it is possible to make a `TopologyGroup` with an array of lists (`dtype=object`) instead.

Note that virtual sites are especially used by gromacs, and I have the TPR reader in mind, but they seem to exist in other simulation engines.

See #1949 and #588.

orbeckst:
The issue that TIP4P is not recognized as a fragment (details in #1949) makes the "make whole" transformation PR #2038  not correctly wrap typical OPLS-AA simulation systems (such as our TPR/XTC test trajectory).

orbeckst:
Let me summarize the discussion on #3168 (based @BartBruininks initial implementation of virtual sites as additional bonded interactions in PR #4318 and on @jbarnoud 's  https://github.com/MDAnalysis/mdanalysis/issues/3168#issuecomment-1755942135 , @hejamu 's https://github.com/MDAnalysis/mdanalysis/issues/3168#issuecomment-1757282196 , and @mattwthompson 'https://github.com/MDAnalysis/mdanalysis/issues/3168#issuecomment-1763222187 )

* Reading vsites and adding the dummy particles as bonded particles to the topology is straightforward and works, PR #4318.
* However, there's broad consensus that it's important to not confuse dummies/vsites with real atoms and chemical bonds because 
  1. it makes the actual chemistry confusing,
  2. implementations of vsites differ between codes so a flexible/customizable solution is needed

**How to implement vsites?** — essentially what @jbarnoud said at the top of this issue:

[...]

PR #4318, opened 2023-10-10
adds vsites as bonds between base particles and vsite

BartBruininks:
Partially addresses #1954

 - ITP reads vsites

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

----
:books: Documentation preview :books:: https://mdanalysis--4318.org.readthedocs.build/en/4318/

### DIFF
--- a/package/MDAnalysis/topology/ITPParser.py
+++ b/package/MDAnalysis/topology/ITPParser.py
@@ -309,6 +309,11 @@ class Molecule:
             'angles': self.parse_angles,
             'dihedrals': self.parse_dihedrals,
             'constraints': self.parse_constraints,
+            'virtual_sites1' : self.parse_virtual_sites1, #BMHB
+            'virtual_sites2' : self.parse_virtual_sites2, #BMHB
+            'virtual_sites3' : self.parse_virtual_sites3, #BMHB
+            'virtual_sites4' : self.parse_virtual_sites4, #BMHB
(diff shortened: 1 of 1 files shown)


orbeckst:
@BartBruininks said in https://github.com/MDAnalysis/mdanalysis/issues/3168#issuecomment-1755858191
> More on the topic of this FIX, if we can agree that this is as intended (vsite interactions are added to bonds), I could also implement it for the TPR reader although I did not have a look at its complexity (I guess it is similar). The TPR reading still does not handle the vsites correctly.

Adding them to bonds makes some sense to me but I'd appreciate the opinion of  @richardjgowers @jbarnoud here.

Once people agree, the next step would be to add tests. 

After we have tests, we can think about the TPR parser.

jbarnoud:
I commented in https://github.com/MDAnalysis/mdanalysis/issues/3168#issuecomment-1755942135

orbeckst (changes requested):
I summarized extensive discussion on the underlying implementation of vsites/dummies at https://github.com/MDAnalysis/mdanalysis/issues/1954#issuecomment-1763439788 .

We first need to properly define how vsites ought to be handled and then proceed with this PR. I will refer to discussions on #1954 as to what the data structure for vsites really should look like. It almost certainly will require work on implementing a new TopologyAttr. This could become part of this PR, given that this PR already solves the problem of reading the information from the TPR.

PR #4324, opened 2023-10-25
feat: add `zone` selection to support a box like selection from given selection

### LINKED ISSUE #4323: Add a geometry selection supporting selecting an zone shaped like a box
I am now working on an system only z axis periodic, and I need to pick the atoms contained within certain distance in only z axis from a selected group. I tried to use `cyzone` to implement, but it could not include all atoms because of the shape of selection is a circle in xy plane and could not include all atoms in xy. And the `prop` selection could not read a dynamic updating selction.

I would like to add a new geometry selection (named as `box`, for example), that could satisfy the requirement.
It would create a zone from center of geometry of given selection to `x`, `y`, or `z` direction in a box.
image

#4324

Cloudac7:
Description:

It introduces a new geometry selection, as described in #4323.
A new selection language is defined in format:
```
box dimension(s) d1_min d1_max (d2_min d2_max) (d3_min d3_max) selection
```
It could pick a zone from the center of geometry from selection.

 - Add new geometry selection: box

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

----
:books: Documentation preview :books:: https://mdanalysis--4324.org.readthedocs.build/en/4324/


testsuite/MDAnalysisTests/core/test_atomselections.py
```diff
@@ -1270,16 +1322,39 @@ def test_similarity_selection_icodes(u_pdb_icodes, selection, n_atoms):
...
-    'name O', 'name N*', 'resname stuff', 'resname ALA', 'type O',
-    'index 0', 'index 1', 'bynum 1-10',
-    'segid SYSTEM', 'resid 163', 'resid 1-10', 'resnum 2',
-    'around 10 resid 1', 'point 0 0 0 10', 'sphzone 10 resid 1',
-    'sphlayer 0 10 index 1', 'cyzone 15 4 -8 index 0',
-    'cylayer 5 10 10 -8 index 1', 'prop abs z <= 100',
-    'byres index 0', 'same resid as index 0',
-])
+
+@pytest.mark.parametrize(
```
IAlibay: We generally ask folks to not reformat large bits of code during their PRs as it makes it very hard to review what actually changed. Could you point out which bits you've added to here?

package/doc/sphinx/source/conf.py
```diff
@@ -14,6 +14,8 @@
 import sys
 import os
 import datetime
+sys.path.insert(0, os.path.abspath('../../..'))
```
IAlibay: Could you explain why this is being added?

package/doc/sphinx/source/documentation_pages/selections.rst
```diff
@@ -277,6 +277,23 @@ cyzone *externalRadius* *zMax* *zMin* *selection*
        relative to the COG of *selection*, instead of absolute z-values
        in the box.
 
+box *dimension(s)* *d1_Max* *d1_Min* (*d2_Max* *d2_Min*) (*d3_Max* *d3_Min*) *selection*
```
IAlibay: I'm very likely missing something, but I don't know if I fully understand how this differs from using `prop` - could you maybe explain the benefits of this new selection over using a succession of `prop`?

package/CHANGELOG
```diff
@@ -42,6 +42,7 @@ Enhancements
     (Issue #3994, PR #4281)  
   * Add support for reading chainID info from Autodock PDBQT files (Issue #4207,
     PR #4284)
+  * Add a new geometric selection: box (Issue #4323, PR #4324)
```
IAlibay: We typically go new entries first, could you add this on line 33 instead?

Also please add yourself to the list of authors for 2.7.0 on line 18.

IAlibay (commented):
Apologies for the slow reviews @Cloudac7

I've added some initial comments.

@MDAnalysis/coredevs could someone please take on the responsibility of reviewing this PR?

package/doc/sphinx/source/conf.py
Cloudac7: Sorry for forgetting remove the line. 
I am developing on MacOS platform and thus the compiling of documentation with original code would raise an ModuleNotFoundError: `No module named 'MDAnalysis'`. My practice followed the instruction of latest documentation for developing, and no solution found. After adding the line in more upper position, `sphinx-build` could work.

testsuite/MDAnalysisTests/core/test_atomselections.py
Cloudac7: It is an reformat came out directly from execution `darker -r upstream/develop testsuite/MDAnalysisTests -L flake8`, after I added a test script `box xz 3 2 4 -5 index 0` into the empty group test. 
Sorry for not aware of the unexpected modification done by darker.

package/doc/sphinx/source/documentation_pages/selections.rst
Cloudac7: Well, `prop` is used to define a static zone in each axes, in my understanding. The new defined `box` selection do similar thing like which `sphzone`, `cyzone` or `cylayer` do, thus could define a zone according to another selection which could also be dynamical when `updating=True`.

orbeckst: See @Cloudac7 's description in https://github.com/MDAnalysis/mdanalysis/issues/4323#issuecomment-1782184664 , which looks sensible to me. If we have cyzone then we can also have box.

package/MDAnalysis/core/selection.py
```diff
@@ -537,6 +537,101 @@ def _apply(self, group):
...
+        "zxy",
+        "zyx",
+    ]
+    axis_map = ["x", "y", "z"]
+
+    def __init__(self, parser, tokens):
+        super().__init__(parser, tokens)
+        self.periodic = parser.periodic
+        self.direction = tokens.popleft()
+        if self.direction not in self.combination:
```
richardjgowers: can we set `xmin` `xmax` etc here to `None` then optionally set them to something else in the `else:` branch below?  I think this would simplify the `__getattribute__` usage below

package/MDAnalysis/core/selection.py
```diff
@@ -537,6 +537,101 @@ def _apply(self, group):
...
+
+            if np.all(group.dimensions[3:] == 90.0):
+                # Orthogonal version
+                vecs -= box[:3] * np.rint(vecs / box[:3])
+            else:
+                # Triclinic version
+                tribox = group.universe.trajectory.ts.triclinic_dimensions
+                vecs -= tribox[2] * np.rint(vecs[:, 2] / tribox[2][2])[:, None]
+                vecs -= tribox[1] * np.rint(vecs[:, 1] / tribox[1][1])[:, None]
+                vecs -= tribox[0] * np.rint(vecs[:, 0] / tribox[0][0])[:, None]
```
richardjgowers: this isn't technically correct for some triclinic boxes with some shifts.

package/doc/sphinx/source/documentation_pages/selections.rst
richardjgowers: this would make more sense to me as `min` to `max` values rather than the opposite.  E.g. `box z 0 8 bilayer` to select a slice of water above a bilayer and `box z -8 0` to get the other half.

package/MDAnalysis/core/selection.py
richardjgowers: I think you'll want this function: https://github.com/MDAnalysis/mdanalysis/blob/c36313e4e849bc37033a142454e9e593db2b3272/package/MDAnalysis/lib/distances.py#L1791

But it won't handle some, but not all, box dimensions being given (i.e. semi periodic calculation)

richardjgowers (changes requested):

package/MDAnalysis/core/groups.py
```diff
@@ -3111,6 +3111,26 @@ def select_atoms(self, sel, *othersel, periodic=True, rtol=1e-05,
...
                 cylinder extends from 10 above the COG to 8 below. Positive
                 values for *zMin*, or negative ones for *zMax*, are allowed.
+            box *dimensions* *dN_min* *dN_max* *selection*
+                Select all atoms within a box region centered
+                on the center of geometry (COG) of a given selection.
+                *dimensions* Specifies which dimension(s) to apply
+                the box selection on. Can be ``x``, ``y``, ``z``,
+                or any combination like ``xy``, ``yz``, ``zx``, ``xyz``
+                (up to 3 characters). *dN_min*, *dN_max* are the minimum and
+                maximum bounds along the first specified dimension.
```
RMeli:
```suggestion
                (up to 3 characters). *dN_min*, *dN_max* are the minimum and
                maximum bounds along each specified dimension.
```
?

package/MDAnalysis/core/groups.py
```diff
@@ -3111,6 +3111,26 @@ def select_atoms(self, sel, *othersel, periodic=True, rtol=1e-05,
                 radius 5, external radius 10 centered on the COG. In z, the
                 cylinder extends from 10 above the COG to 8 below. Positive
                 values for *zMin*, or negative ones for *zMax*, are allowed.
+            box *dimensions* *dN_min* *dN_max* *selection*
```
RMeli:
```suggestion
            box *dimensions* *dN_min* *dN_max* [*dN_min* *dN_max*] [*dN_min* *dN_max*] *selection*
```
Specify also the optional arguments (depending on `dimension`)?

package/MDAnalysis/core/selection.py
```diff
@@ -537,6 +537,103 @@ def _apply(self, group):
...
+        "yx",
+        "zx",
+        "zy",
+        "xyz",
+        "xzy",
+        "yxz",
+        "yzx",
+        "zxy",
+        "zyx",
+    ]
```
RMeli:
```suggestion
    combination = [''.join(s) for i in range(1, 4) for s in itertools.permutations("xyz", i)]
```
Just for fun. Probably less readable than what you wrote. 

But the question is, do we really need both `xy` and `yx`? What's the use case of having all permutations as opposed to only combinations?

```suggestion
    combination = [''.join(s) for i in range(1, 4) for s in itertools.combinations("xyz", i)]
```

Nitpick: the variable is called "combination", so the name might have to change to "permutation" ("permutations"?) if that's the actual use case.

package/MDAnalysis/core/selection.py
```diff
@@ -537,6 +537,103 @@ def _apply(self, group):
...
+        super().__init__(parser, tokens)
+        self.periodic = parser.periodic
+        self.direction = tokens.popleft()
+        self.xmin, self.xmax = None, None
+        self.ymin, self.ymax = None, None
+        self.zmin, self.zmax = None, None
+        if self.direction not in self.combination:
+            raise ValueError(
+                "The direction '{}' is not valid. Must be combination of {}"
+                "".format(self.direction, ["x", "y", "z"])
```
RMeli:
```suggestion
                f"The direction '{self.direction}' is not valid. Must be combination of {axis_map}"
```

package/MDAnalysis/core/selection.py
Cloudac7: The extensive list of all permutations of `x`, `y`, and `z` is designed to avoid the need to calculate the O(3!) = O(1) permutations each time a selection is executed. This has been reformatted by `darker` to comply with PEP8 standards. It's a minor optimization that can be overlooked, though. 🤣

Regarding the latter point, my original intention was to enhance the flexibility of the selection by making both `yx` and `xy` acceptable. However, this inadvertently added extra complexity. You're correct in suggesting that it would be better to have users write in the sequence of `x -> y -> z`. This approach would eliminate the need for additional processing.

package/doc/sphinx/source/documentation_pages/selections.rst
orbeckst: @richardjgowers I had the same thoughts but unfortunately [cyzone and friends](https://docs.mdanalysis.org/stable/documentation_pages/selections.html#geometric) started this inverse "max min" syntax. I feel consistency across the `*zone` and `*layer` selections is more important here than purity. 

(I mean I am _really_ looking for someone to convince me we should do `min max` here ;-) because I know that I am going to mistype it...)

orbeckst:
@richardjgowers I am tentatively assigning the PR to you for shepherding it through. If it's too much then please unassign yourself and ping me. Thanks.

orbeckst:
@richardjgowers could you have a look if @Cloudac7 addressed the reviewers' comments? Thanks!

orbeckst:
@richardjgowers could you please have another look?

today on discord someone asked for this feature so I was reminded of this PR…

orbeckst:
@Cloudac7 could you please resolve the conflicts?

Sorry that this has been very slow from our end.

package/MDAnalysis/core/selection.py
```diff
@@ -537,6 +537,96 @@ def _apply(self, group):
...
+        # Calculate vectors between point of interest and our group
+        vecs = group.positions - sel.center_of_geometry()
+        range_map = {
+            0: (self.xmin, self.xmax),
+            1: (self.ymin, self.ymax),
+            2: (self.zmin, self.zmax),
+        }
+
+        if self.periodic and group.dimensions is not None:
+            box = group.dimensions[:3]
```
richardjgowers: I'm not sure this is correct for triclinic boxes, I think instead you want the trace of the 3x3 vector representation

package/MDAnalysis/core/groups.py
```diff
@@ -3271,6 +3271,25 @@ def select_atoms(self, sel, *othersel, periodic=True, rtol=1e-05,
                 radius 5, external radius 10 centered on the COG. In z, the
                 cylinder extends from 10 above the COG to 8 below. Positive
                 values for *zMin*, or negative ones for *zMax*, are allowed.
+            box *dimensions* *dN_min* *dN_max* [*dN_min* *dN_max*] [*dN_min* *dN_max*] *selection*
```
richardjgowers: We already have sphzone so maybe this could be called zone?

richardjgowers (commented):
Looks good, couple tweaks to go

package/MDAnalysis/core/selection.py
Cloudac7: In fact, I'm not sure if the check necessary. In my opinion, it might be nature to select a zone a little larger than the unit cell, yielding all of the atoms in the unit cell on the corresponding direction. Although it would be misleading because of PBC.

PR #4325, opened 2023-10-27
Update deploy things for py312 + OpenID Connect publishing

IAlibay:
Towards #4326 

Title says it all

 - cibuildwheel runners are all updated
 - updates various other actions being called

----
:books: Documentation preview :books:: https://mdanalysis--4325.org.readthedocs.build/en/4325/

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -4,6 +4,7 @@ on:
   pull_request:
     branches:
       - "package-*"
+      - develop
   push:
     branches:
       - "package-*"
@@ -15,9 +16,6 @@ on:
   schedule:
     # 3 am Tuesdays and Fridays
     - cron: "0 3 * * 2,5"
-  pull_request:
-    branches:
-      - develop
 
 
 concurrency:
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -15,6 +15,9 @@ on:
   schedule:
     # 3 am Tuesdays and Fridays
     - cron: "0 3 * * 2,5"
+  pull_request:
+    branches:
+      - develop
 
 
 concurrency:
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -37,31 +37,32 @@ jobs:
       fail-fast: false
       matrix:
         buildplat:
-          - [ubuntu-20.04, manylinux_x86_64]
-          - [macos-11, macosx_*]
-          - [windows-2019, win_amd64]
-        python: ["cp39", "cp310", "cp311"]
+          - [ubuntu-22.04, manylinux_x86_64, x86_64]
+          - [macos-11, macosx_*, x86_64]
+          - [windows-2019, win_amd64, AMD64]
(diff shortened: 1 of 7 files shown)


IAlibay:
OpenID Connect trusted publisher for test and normal PyPi have now been set for MDAnalysis and MDAnalysisTests

IAlibay:
For the sake of security, all old API tokens have also been deleted.

.github/workflows/deploy.yaml
```diff
@@ -4,6 +4,7 @@ on:
   pull_request:
     branches:
       - "package-*"
+      - develop
```
orbeckst: What is now happening (compared to previously) when we get a pull request against develop?

.github/workflows/deploy.yaml
```diff
@@ -172,16 +170,11 @@ jobs:
           mv dist/MDAnalysisTests-* testsuite/dist
 
       - name: upload_source_and_wheels
-        uses: pypa/gh-action-pypi-publish@v1.5.0
-        with:
-          user: __token__
-          password: ${{ secrets.PYPI_API_TOKEN_SRC }}
+        uses: pypa/gh-action-pypi-publish@v1.18.10
```
orbeckst, line 172: Is the MDA deployment so complicated that we can't use our action https://github.com/MDAnalysis/pypi-deployment ?

orbeckst (approved):
I hope it works! Thanks for the update.

(Small question about including "develop" but I assume that this fine.)

.github/workflows/deploy.yaml
IAlibay: cibuildwheels deployments don't work via pypi-deployment right now - that's a bit more complicated so I've not added that branch yet (it's a bit more messy because you usually want to parallelise at the level of runners for each flavour you build wheels for - you don't fully have to though, you can just trust the docker container to get it right).

One day I'll get that done, it's just not been a burning priority (although the amount of cython packages we're maintinaing is starting to grow).

What I can't stuff in a reusable action though is the arm builds (using Cirrus), those will need to exist everywhere we need those builds (currently just here).

.github/workflows/deploy.yaml
IAlibay: This is just a test thing for this PR, I'm removing it before merge.

.github/workflows/deploy.yaml
orbeckst: Thanks for the explanations!

IAlibay: Extra info for anyone reading this and going "but why cibuildwheels here and not everywhere else" (it's a recurring question we should document) - we use cibuildwheels to build anything that can't do a pure Python wheel (i.e. uses Cython).

orbeckst: Do we have a "maintainer FAQ" somewhere — wiki or NF drive?

IAlibay: Unfortunately not, we have barbones deployment docs on the userguide but yeah there's just not been enough of my time to add docs for everything - that's what we applied for that SDG for though so fingers crossed!

PR #4328, opened 2023-10-27
Fix deploy to use the right environment & token permissions

IAlibay:
Follows up from #4325 

I didn't see what we needed to set the environment url & token permissions, so doing so now.

This does load to some annoying proliferation of action steps, I'll probably create a reusable action at some point, but for now this will do.

----
:books: Documentation preview :books:: https://mdanalysis--4328.org.readthedocs.build/en/4328/

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -117,12 +117,16 @@ jobs:
           path: testsuite/dist/*.tar.gz
           retention-days: 7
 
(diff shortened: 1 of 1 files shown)


orbeckst (approved):
Thanks, lgtm as far as I can tell.

(I agree with you: it would be nice to reduce code duplication eventually for the [package, tests] x [testpypi, pypi] deployment matrix.)

PR #4331, opened 2023-11-03
Make networkx optional

### LINKED ISSUE #3822: Make networkx an optional dep?
Networkx is solely used by a single analysis module (leaflet.py). It may make sense to make this an optional dependency?

IAlibay:
Fixes #3822 

 - NX is now optional

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

----
:books: Documentation preview :books:: https://mdanalysis--4331.org.readthedocs.build/en/4331/

### DIFF
--- a/package/MDAnalysis/analysis/leaflet.py
+++ b/package/MDAnalysis/analysis/leaflet.py
@@ -71,13 +71,21 @@ Classes and Functions
 import warnings
 
 import numpy as np
-import networkx as NX
 
 from .. import core
 from . import distances
 from .. import selections
-
 from ..due import due, Doi
+
+
+# networkx is an optional import
+try:
+    import networkx as NX
+except ImportError:
+    HAS_NX = False
+else:
+    HAS_NX = True
+
 
 due.cite(Doi("10.1002/jcc.21787"),
          description="LeafletFinder algorithm",
@@ -111,6 +119,12 @@ class LeafletFinder(object):
         use fast :func:`~MDAnalysis.lib.distances.distance_array`
         implementation [``None``].
 
+    Raises
+    ------
+    ImportError
+      This class requires the optional package `networkx`. If not found
+      an ImportError is raised.
+
     Example
     -------
     The components of the graph are stored in the list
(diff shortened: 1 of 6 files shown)


package/MDAnalysis/analysis/leaflet.py
```diff
@@ -71,14 +71,22 @@
...
 from . import distances
 from .. import selections
-
 from ..due import due, Doi
 
+
+# networkx is an optional import
+try:
+    import networkx as NX
+except ImportError:
```
IAlibay, line 84: not sure why these aren't getting tested - it's possible networkx is getting pulled in by something else, test works fine locally

package/MDAnalysis/analysis/leaflet.py
```diff
@@ -71,14 +71,22 @@
...
 from .. import core
 from . import distances
 from .. import selections
-
 from ..due import due, Doi
 
+
+# networkx is an optional import
+try:
+    import networkx as NX
```
RMeli, line 83:
```suggestion
    import networkx as nx
```
Any reason for this to be capitalised? Networkx documentation seems to use all lowercase.

package/setup.py
```diff
@@ -665,6 +664,7 @@ def long_description(readme):
                   'scikit-learn',  # For clustering and dimensionality
                                    # reduction functionality in encore
                   'tidynamics>=1.0.0', # For MSD analysis method
+                  'networkx>=2.0',  # For LeafletFinder
```
RMeli, line 667: Note to self: this will have to be reflected in the Spack package.

package/MDAnalysis/analysis/leaflet.py
IAlibay: history, other parts of the code use NX, I'm mostly doing a minimal footprint thing, imho we should just use the full name if that's what we want to do (edit: and I'm happy to do so if that's the request!)

RMeli: Fair. I was asking mostly out of curiosity, no need to change in my view.

package/MDAnalysis/analysis/leaflet.py
RMeli: Is this a blocker?

IAlibay: Unclear, here and for the biopython removal there seems to be an issue with coverage, we have at least one runner without `networkx` (or `biopython` in the other PR) and both are uploading to codecov, but somehow this returns as uncovered.

IAlibay: You can see here in the coverage report - those lines are indeed covered: https://app.codecov.io/gh/MDAnalysis/mdanalysis/pull/4331/blob/package/MDAnalysis/analysis/leaflet.py

RMeli: Thanks for checking @IAlibay

RMeli (approved):
Thanks @IAlibay

PR #4340, opened 2023-11-05
Deprecate lib.util.which

### LINKED ISSUE #3649: Replace custom lib.util.which with shutil.which

IAlibay:
Fixes #3649

Notes:
- Not switching over x3dna because that code doesn't even work.
- Separate PR has been opened in hole2-mdakit

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

----
:books: Documentation preview :books:: https://mdanalysis--4340.org.readthedocs.build/en/4340/

### DIFF
--- a/package/MDAnalysis/analysis/hole2/hole.py
+++ b/package/MDAnalysis/analysis/hole2/hole.py
@@ -22,6 +22,7 @@
 #
 import os
 import errno
+import shutil
 import tempfile
 import textwrap
 import logging
@@ -228,7 +229,7 @@ def hole(pdbfile,
         warnings.warn(msg.format(output_level))
 
     # get executable
-    exe = util.which(executable)
+    exe = shutil.which(executable)
     if exe is None:
         raise OSError(errno.ENOENT, exe_err.format(name=executable,
                                                    kw='executable'))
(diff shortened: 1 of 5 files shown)


package/MDAnalysis/analysis/hole2/hole.py
```diff
@@ -228,7 +229,7 @@ def hole(pdbfile,
         warnings.warn(msg.format(output_level))
 
     # get executable
-    exe = util.which(executable)
+    exe = shutil.which(executable)
```
IAlibay, line 232: not looking to add tests to cover these if the migration is also being done in hole2-mdakit

hmacdope (approved):
LGTM,  thanks @IAlibay

PR #4345, opened 2023-11-14
Towards a better optional package system

IAlibay:
This was brought up a while ago by @orbeckst - how do we deal with all these optional dependencies when we need version ranges.

This aim of this PR is eventually to clean up the optional dependency approach by centralising version checks in one location.

 - Adds an optional_import method that either returns None or the package
 - Adds an optional "check the version range"

TODO:
- Switch everything over to using this new method

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

----
:books: Documentation preview :books:: https://mdanalysis--4345.org.readthedocs.build/en/4345/

### DIFF
--- a/package/MDAnalysis/analysis/align.py
+++ b/package/MDAnalysis/analysis/align.py
@@ -196,24 +196,23 @@ import collections
 
 import numpy as np
 
-try:
-    import Bio.AlignIO
-    import Bio.Align
-    import Bio.Align.Applications
-except ImportError:
-    HAS_BIOPYTHON = False
-else:
-    HAS_BIOPYTHON = True
-
 import MDAnalysis as mda
 import MDAnalysis.lib.qcprot as qcp
 from MDAnalysis.exceptions import SelectionError, SelectionWarning
 import MDAnalysis.analysis.rms as rms
 from MDAnalysis.coordinates.memory import MemoryReader
-from MDAnalysis.lib.util import get_weights
+from MDAnalysis.lib.util import get_weights, optional_import
 from MDAnalysis.lib.util import deprecate   # remove 3.0
 
 from .base import AnalysisBase
+
+biopython = optional_import('Bio', min_version='1.80')
+
+if biopython is not None:
+    import Bio.AlignIO
+    import Bio.Align
+    import Bio.Align.Applications
+
 
 logger = logging.getLogger('MDAnalysis.analysis.align')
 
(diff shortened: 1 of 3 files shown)


orbeckst:
@IAlibay please ping me when you need a review.

IAlibay:
@orbeckst first impressions would be good - is this too complicated, too unpythonic, etc?

package/MDAnalysis/analysis/align.py
```diff
@@ -1070,7 +1069,7 @@ def sequence_alignment(mobile, reference, match_score=2, mismatch_penalty=-1,
        Biopython is now an optional dependency which this method requires.
 
     """
-    if not HAS_BIOPYTHON:
+    if biopython is None:
```
orbeckst, line 1072: The semantics of `if not HAS_BIOPYTHON` are clearer.

The following is extra code
```python
def has_optional_dependency(mod):
    return mod is not None
```
but may be more readable:
```python
from MDAnalysis.lib.util import optional_import, has_optional_dependency

if not has_optional_dependency(biopython):
   # do the thing when we don't have biopython
```

package/MDAnalysis/lib/util.py
```diff
@@ -2552,3 +2556,59 @@ def wrapper(self, *args, **kwargs):
                         self._kwargs[key] = arg
         return func(self, *args, **kwargs)
     return wrapper
+
+
+def optional_import(
```
orbeckst, line 2561: How about a separate `lib.optional` module? It's a well-characterized problem so I'd prefer than throwing even more into util. We can then also have better/more focused docs to go along with it.

orbeckst (commented):
Really good start, see inline comments.

orbeckst:
As long as this is in "WIP" I won't be assigning anyone to the PR. Let me know if you need another set of eyes.

RMeli:
> As long as this is in "WIP" I won't be assigning anyone to the PR.

Removed the `WIP` label and converted to draft, to avoid confusion.

IAlibay:
@RMeli iirc not all of CI is triggered in draft PRs, need to double check but enforcing draft PR might be a bit of an issue (especially here where I need to ensure we get triggers across all OS types)

RMeli:
Ah. Fair point. For organisational purposes, I would be in favour of enabling all CI on drafts PR, to encourage their use.

IAlibay:
Iirc the issue is with the azure trigger, or maybe it was cirrus - definitely something that was out of our control at the time. Generally if gh actions runs that ok for drafts, but here + py3.12 that's a problem.

IAlibay:
I can have a proper check on what's up after I'm done with other internal MDA tasks - maybe it's not a thing anymore (especially if it's cirrus), but for now I would suggest not being too stringent unless we really have to (especially if I end up having to turn this back to WIP).

IAlibay:
This might be what I'm remembering 😅https://github.com/MicrosoftDocs/azure-devops-docs/issues/13175

RMeli:
Yep, sorry about that. I was trying to avoid #4361 from repeating.

PR #4348, opened 2023-11-26
Adding a ref point support for atom selection

### LINKED ISSUE #4344: Add reference point to geometric selections

We currently have a bunch of selections that are `select_atoms('sphzone 3.0 protein')`, for a full list: 

https://docs.mdanalysis.org/stable/documentation_pages/selections.html#geometric

This takes the COG of a selection to use as a reference point.  Sometimes it would be nice to manually specify the reference point

To select a spherical zone around point x=1.1 y=1.2 z=1.3, something like:

`select_atoms('sphzone 3.0 point 1.1 1.2 1.3')`

abdulasiraj:
Fixes #4344

 - 

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

----
:books: Documentation preview :books:: https://mdanalysis--4348.org.readthedocs.build/en/4348/


richardjgowers (changes requested):
This looks like a good start, this is trying to make `refpoint x y z` a possible selection as part of a command like `sphlayer` in the selection language.

The return from `_apply` here needs to usually be an AtomGroup, as other selections will expected it to provide things like center_of_geometry: https://github.com/MDAnalysis/mdanalysis/blob/develop/package/MDAnalysis/core/selection.py#L348

There's probably two ways to make this work:
1. you could make the `_apply` method here return a AtomGroup-like object, so define a PseudoAtomGroup class here that duck types (imitates) all the required methods (like center_of_geometry etc).
2. instead, make all geometric selections that require a selection, instead have a branch where they can instead parse the `refpoint` here.  This would change the signature of `sphlayer 1.0 2.0 selection OR refpoint`

Option 2 here is probably more lines of code and seems less elegant, but option 1 seems a little fragile as the pseudo AtomGroup class would require a lot of maintenance to keep it behaving like a real AtomGroup

abdulasiraj:
Hi @richardjgowers,
I'm going with 2nd option as it's more robust but I've few questions regarding the new signature:

1. If user pass both `selection` and `refPoint`, which should we prefer?
2. Aren't we making compulsory for users to provide something about selection either `None` or something else to avoid grammar errors?

orbeckst:
Thanks for the contribution @abdulasiraj and thanks for the initial round of feedback @richardjgowers .

As there's still a WIP-label in the title I assume that you still want to do a fair amount of work by yourself @abdulasiraj . Once you think it's really ready for review, please change the name and then we can have closer look.

RMeli:
@abdulasiraj I removed the `WIP` label to the title and converted this PR into a draft, to avoid confusion. Please feel free to mark it as "Ready to review" when it's no longer WIP.

abdulasiraj:
That's make more sense. I'll update for sure. Thanks @RMeli

orbeckst:
@abdulasiraj are you still working on this PR? Otherwise we will close it.

abdulasiraj:
Hi @orbeckst, I'll complete it in few days hopefully. just got time to work on it. Thanks!

PR #4360, opened 2023-12-10
BENCH: fix time_geometric_selections

### LINKED ISSUE #3519: Benchmarks expect flags
## Expected behavior ##

The benchmarks run without error.

## Actual behavior ##

Some benchmarks fail because they expect flags that were removed in version 2.0.0. Here is a sampling of the errors:

```
For parameters: 'prop abs z <= 5.0', False, [True, False]
 File "/home/benchy/mambaforge/lib/python3.9/site-packages/asv/benchmark.py", line 1184, in main_run_server
   main_run(run_args)
 File "/home/benchy/mambaforge/lib/python3.9/site-packages/asv/benchmark.py", line 1058, in main_run
   result = benchmark.do_run()
 File "/home/benchy/mambaforge/lib/python3.9/site-packages/asv/benchmark.py", line 537, in do_run
   return self.run(*self._current_params)
 File "/home/benchy/mambaforge/lib/python3.9/site-packages/asv/benchmark.py", line 627, in run
   samples, number = self.benchmark_timing(timer, min_repeat, max_repeat,
 File "/home/benchy/mambaforge/lib/python3.9/site-packages/asv/benchmark.py", line 694, in benchmark_timing
   timing = timer.timeit(number)
 File "/home/benchy/mambaforge/lib/python3.9/timeit.py", line 177, in timeit
   timing = self.inner(it, self.timer)
 File "<timeit-src>", line 6, in inner
 File "/home/benchy/mambaforge/lib/python3.9/site-packages/asv/benchmark.py", line 599, in <lambda>
   func = lambda: self.func(*param)
 File "/home/benchy/mdanalysis/benchmarks/benchmarks/selections.py", line 76, in time_geometric_selections
   MDAnalysis.core.flags['use_periodic_selections'] = periodic_selection[0]
AttributeError: module 'MDAnalysis.core' has no attribute 'flags'
```

tylerjereddy:
* Fixes gh-3519

* There is no good reason to keep the flags in the benchmark--this benchmark has been broken for over a year so we should just use the appropriate kwargs where possible and leave a `TODO` comment for core flags that we don't have a substitute for

* better to have this running than hard failing all the time, even if we've lost one of the parametrization dimensions

* if someone comes along with the confidence to say we can delete the `TODO` comment, then we can also delete some of the extra parametrization complexity of course

[skip cirrus]

----
:books: Documentation preview :books:: https://mdanalysis--4360.org.readthedocs.build/en/4360/

### DIFF
--- a/benchmarks/benchmarks/selections.py
+++ b/benchmarks/benchmarks/selections.py
@@ -72,11 +72,9 @@ class GeoSelectionBench(object):
                                   dynamic_selection,
                                   periodic_selection):
 
-        # set core flags for PBC accounting
-        MDAnalysis.core.flags['use_periodic_selections'] = periodic_selection[0]
-        MDAnalysis.core.flags['use_KDTree_routines'] = periodic_selection[1]
-
-        if hasattr(MDAnalysis.Universe, 'select_atoms'):
-            self.u.select_atoms(selection_string, updating=dynamic_selection)
-        else:
-            self.u.selectAtoms(selection_string, updating=dynamic_selection)
+        # TODO: Do we need a kwarg similar to old `use_KDTree_routines`
+        # flag? We used to benchmark that.
+        self.u.select_atoms(selection_string,
+                            updating=dynamic_selection,
+                            periodic=periodic_selection[0],
+                            )


RMeli (approved):
Sounds reasonable to me.

hmacdope (approved):
LGTM @tylerjereddy

PR #4367, opened 2023-12-15
TST, MAINT: test_frame_bool_fail compat

tylerjereddy:
* adjust `test_frame_bool_fail` such that it passes with both NumPy `1.26.2` and NumPy `main`; the error message string changed on the latter, so a small shim was needed

[skip cirrus]

----
📚 Documentation preview 📚: https://mdanalysis--4367.org.readthedocs.build/en/4367/


IAlibay (approved):
Thanks @tylerjereddy !

PR #4361, opened 2023-12-11
[WIP] Fix authors & changelog

IAlibay:
Various fixes as I cross-check through the history of our contributors.

## Developers certificate of origin
- [x] I certify that this contribution is covered by the LGPLv2.1+ license as defined in our [LICENSE](https://github.com/MDAnalysis/mdanalysis/blob/develop/LICENSE) and adheres to the [**Developer Certificate of Origin**](https://developercertificate.org/).

----
:books: Documentation preview :books:: https://mdanalysis--4361.org.readthedocs.build/en/4361/

PR #4368, opened 2023-12-15
DOC: CHANGELOG MM/DD/YY

tylerjereddy:
* I merged gh-4361 a bit early, so I should clean up the mess; I think this fixes the two cases where the datetime string was not formatted to match our documented convention

[skip cirrus]

----
📚 Documentation preview 📚: https://mdanalysis--4368.org.readthedocs.build/en/4368/

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -80,7 +80,7 @@ Deprecations
     PR #3735)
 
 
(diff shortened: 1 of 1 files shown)


IAlibay (approved):
Thanks! (also I would say it's mostly my fault for messing up in the first place 😅)

PR #4369, opened 2023-12-15
MAINT: encore Windows type fixes

tylerjereddy:
* in NumPy `2.0.0`, the default integer size on Windows will be 64-bit; we have 10 related test failures on Windows alongside NumPy `main` at the moment, and this patch fixes those failures by forcing a reversion to the old behavior, effectively kicking the can down the road

* I think we're going to proceed with a similar solution upstream for now (i.e., https://github.com/scipy/scipy/issues/19605#issuecomment-1852698295);

* I haven't kept up with whether this should actually be targeted at the encore "subproject" now with the reorg happening; feel free to move it if so...

[skip cirrus]

----
📚 Documentation preview 📚: https://mdanalysis--4369.org.readthedocs.build/en/4369/

### DIFF
--- a/package/MDAnalysis/analysis/encore/clustering/affinityprop.pyx
+++ b/package/MDAnalysis/analysis/encore/clustering/affinityprop.pyx
@@ -99,7 +99,7 @@ def AffinityPropagation(s, preference, float lam, int max_iterations, int conver
 
     # Prepare input and ouput arrays
     cdef numpy.ndarray[numpy.float32_t,  ndim=1] matndarray = numpy.ascontiguousarray(s._elements, dtype=numpy.float32)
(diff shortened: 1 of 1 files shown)


IAlibay:
With encore moving to MDAEncore, I'm not sure if this fix should go here or there or both :/

tylerjereddy:
Ok, let me know what you want me to do. You have a PR to improve optional dep resolution right? Does that need to go in soon/first?

IAlibay (approved):
Given it's a single line fix, I'm tempted to say let's merge this and we can port it over (if that's ok with you @tylerjereddy) -  does this sound reasonable with you @orbeckst ?

IAlibay:
We do need to deprecate Encore properly, I'll raise an issue in the morning.

orbeckst (approved):
Thanks. Fix here and let’s port to MDAKit encore.

PR #4375, opened 2023-12-21
Update README to reflect move to GitHub Discussions

RMeli:
----
📚 Documentation preview 📚: https://mdanalysis--4375.org.readthedocs.build/en/4375/


orbeckst (changes requested):
This looks fine. _Maybe_ generate a banner with shield.io for GH Discussions in their place?

Please check other places such as https://github.com/MDAnalysis/mdanalysis/blob/develop/package/pyproject.toml (and setup.py?) ... I'd `git grep` the whole source tree.

Thanks for the update... sorry for brevity.

RMeli:
@orbeckst, very good point. I didn't look past the readme. I now added the shield and removed all occurrences. I also toned down the "question" issue template, since it would be trivial now to migrate an issue to a discussion.

orbeckst (approved):
Thank you for the thorough scouring of the docs — there were more places than I thought. LGTM

Other places:
* Could you please also look at the UG? I assume we have similar occurrences there. 
* Did you change the website?
* Teaching materials? — These can probably be updated as we go along.
* MDAKits??

orbeckst:
@RMeli please squash-merge when you're happy with it.

RMeli:
Yes, I'll look at the other places in the coming days, thanks for bringing this up!

PR #4383, opened 2023-12-23
Try to fix numpy dev0 install in nightly wheels

IAlibay:
Fixes #

 - 

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [ ] CHANGELOG updated?

----
📚 Documentation preview 📚: https://mdanalysis--4383.org.readthedocs.build/en/4383/

### DIFF
--- a/.github/workflows/gh-ci-cron.yaml
+++ b/.github/workflows/gh-ci-cron.yaml
@@ -54,7 +54,7 @@ jobs:
     # Note: need to install contourpy first before matplotlib
     - name: nightly_wheels
       run: |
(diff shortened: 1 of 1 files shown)


IAlibay:
Let's see how CI looks, but I believe this should be good now. <s>I swapped the order around in the hope that mamba would be more efficient at recognising the pre-installed (via pip) modules.</s>

Actually on second thought, maybe the old approach was better so we could know if the dev0 build isn't picked up.

From the output of pip & mamba list it seems to be good.

IAlibay:
Supplemented by https://github.com/MDAnalysis/mdanalysis/pull/4502

PR #4402, opened 2023-12-29
Update minimum numpy support to match 1.23[.2]

IAlibay:
Fixes #4401 

Changes made in this Pull Request:
 - Bump everything up to 1.23.2

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?
 - [x] Issue raised/referenced?

## Developers certificate of origin
- [x] I certify that this contribution is covered by the LGPLv2.1+ license as defined in our [LICENSE](https://github.com/MDAnalysis/mdanalysis/blob/develop/LICENSE) and adheres to the [**Developer Certificate of Origin**](https://developercertificate.org/).

----
📚 Documentation preview 📚: https://mdanalysis--4402.org.readthedocs.build/en/4402/

PR #4407, opened 2024-01-02
ChainedReader __repr__ fix for MemoryReader sub-readers

### LINKED ISSUE #3349: ChainedReader's __repr__ fails with MemoryReader sub-readers.
## Expected behavior ##
Printing a ChainReader object yields some summarized info about the sub-readers, regardless of their type.

## Actual behavior ##
If one of the sub-readers is a MemoryReader (as discussed in #3344), `ChainReader.__repr__` fails because it tries to pass the respective filename to `os.path.basename` for prettier printing.

`os.path.basename` fails because the MemoryReader filename stored in `ChainReader.filenames` is the full coordinate array (!!). The reason is that the ChainReader saves under `ChainReader.filenames` the arguments it was passed to instantiate the sub-readers. Normally, these are filename strings, but for the MemoryReader it's a coordinate array.

``` python
import MDAnalysis as mda
from MDAnalysis.tests.datafiles import PSF
import numpy as np

cdx = np.random.random(3341*3*10).reshape(-1, 3341, 3).astype(np.float32) * 10
u = mda.Universe(PSF, (cdx, cdx))
print(u.trajectory)
```

## Fix
The fix is twofold: first, the ChainReader should no longer store the MemoryReader coordinate array as the filename. It should instead set that to `None`, which is what `MemoryReader.filename` always reports.

Second, `ChainReader.__repr__` must fallback to some default when the filename is `None` or empty.

2.0.0-dev0

ljwoods2:
Fixes #3349 

 - If ChainedReader is initialized with filenames that are trajectory ndarrays rather than files, the stored filenames will be "None" for each ndarray rather than a filename
 - The __repr__ function will check if the filenames are ndarrays before calling os.path.basename

If there is a use case for a ChainedReader with both MemoryReader sub-readers and other sub-reader types, this solution won't work because it depends on all the filenames being the same type. Let me know and I can fix it!

PR Checklist
------------
 - [X] Tests?
 - [X] Docs?
 - [X] CHANGELOG updated?

----
📚 Documentation preview 📚: https://mdanalysis--4407.org.readthedocs.build/en/4407/

### DIFF
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -25,7 +25,7 @@ Fixes
    `DistanceMatrix.run(frames=...)` (PR #4433)
  * Fix doctest errors of lib/picklable_file_io.py (Issue #3925, PR #4371)
  * Fix deploy action to use the correct version of the pypi upload action.
- * Fix ChainReader repr() method when sub-reader is MemoryReader 
+ * Fix ChainReader `__repr__()` method when sub-reader is MemoryReader 
    (Issue #3349, PR #4407)
 
 Enhancements
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -14,7 +14,7 @@ The rules for this file:
 
 
 -------------------------------------------------------------------------------
-??/??/?? IAlibay, HeetVekariya, marinegor
+??/??/?? IAlibay, HeetVekariya, marinegor, ljwoods2
 
  * 2.8.0
 
@@ -25,6 +25,8 @@ Fixes
    `DistanceMatrix.run(frames=...)` (PR #4433)
  * Fix doctest errors of lib/picklable_file_io.py (Issue #3925, PR #4371)
  * Fix deploy action to use the correct version of the pypi upload action.
+ * Fix ChainReader repr() method when sub-reader is MemoryReader 
+   (Issue #3349, PR #4407)
 
 Enhancements
 
--- a/package/MDAnalysis/coordinates/chain.py
+++ b/package/MDAnalysis/coordinates/chain.py
@@ -269,7 +269,7 @@ class ChainReader(base.ReaderBase):
             kwargs['dt'] = dt
         self.readers = [core.reader(filename, convert_units=convert_units, **kwargs)
                         for filename in filenames]
-        # Iterate through
+        # Iterate through all filenames, appending NoneType None for ndarrays
         self.filenames = []
         for fn in filenames:
             if isinstance(fn, np.ndarray):
@@ -610,7 +610,7 @@ class ChainReader(base.ReaderBase):
             fname = (os.path.basename(self.filenames[0])
                      if self.filenames[0] else "numpy.ndarray")
             fnames = "{fname} and {nfnames} more".format(
-                
+
                     fname=fname,
                     nfnames=len(self.filenames) - 1)
         else:
--- a/package/MDAnalysis/coordinates/chain.py
+++ b/package/MDAnalysis/coordinates/chain.py
@@ -271,6 +271,13 @@ class ChainReader(base.ReaderBase):
                         for filename in filenames]
         self.filenames = np.array([fn[0] if isinstance(fn, tuple) else fn
                                                         for fn in filenames])
+        # Set filenames array to array of "None" if chainreader initialized with
+        # numpy arrays
+        if isinstance(filenames[0], np.ndarray):
+            self.filenames = np.array(["None" for _ in range(len(filenames))])
+        else:
+            self.filenames = np.array([fn[0] if isinstance(fn, tuple) else fn
+                                      for fn in filenames])
         # pointer to "active" trajectory index into self.readers
         self.__active_reader_index = 0
 
@@ -597,7 +604,9 @@ class ChainReader(base.ReaderBase):
         self._apply('close')
 
     def __repr__(self):
-        if len(self.filenames) > 3:
+        if self.filenames[0] == "None":
+            fnames = "NumPy array"
+        elif len(self.filenames) > 3:
             fnames = "{fname} and {nfanmes} more".format(
                     fname=os.path.basename(self.filenames[0]),
                     nfanmes=len(self.filenames) - 1)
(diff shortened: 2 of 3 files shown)


RMeli:
@ljwoods2 thank you for the PR. From a quick look, it looks like this PR contains un-related changes from #4402. Could you please fix this before we review? Thanks.

ljwoods2:
> @ljwoods2 thank you for the PR. From a quick look, it looks like this PR contains un-related changes from #4402. Could you please fix this before we review? Thanks.

Done, not sure how I managed to do that hahaha. Just reset the branch entirely

RMeli:
Thanks @ljwoods2. I just had a look at the original issue #3349 and realised there is an open PR #3344 which includes a fix for the same issue. Unfortunately that PR seems to have stalled. 

@mnmelo, would you be OK to address #3349 here, or do you prefer to keep it as part of #3344 (which we can hopefully get going again)?

mnmelo:
Hi,

Definitely address it here. At least that much gets fixed, even if I don't get the time to advance #3344.

package/MDAnalysis/coordinates/chain.py
```diff
@@ -271,6 +271,13 @@ def __init__(self, filenames, skip=1, dt=None, continuous=False,
                         for filename in filenames]
         self.filenames = np.array([fn[0] if isinstance(fn, tuple) else fn
                                                         for fn in filenames])
+        # Set filenames array to array of "None" if chainreader initialized with
+        # numpy arrays
+        if isinstance(filenames[0], np.ndarray):
+            self.filenames = np.array(["None" for _ in range(len(filenames))])
```
lilyminium: > If there is a use case for a ChainedReader with both MemoryReader sub-readers and other sub-reader types, this solution won't work because it depends on all the filenames being the same type.

I'm not sure this would be a *common* use-case, but I can definitely concoct a non-implausible scenario where a user wants to pre-pend some frames onto an existing trajectory! Would it be a lot of work to iterate through `filenames` and check if each is a `np.ndarray` or `tuple`, to have it match more closely to the actual underlying readers and be a bit more robust?

Separately, it was a good idea to keep the types the same no matter what, so that `self.filenames` is always an array of strings. However, it might be more natural to use the `NoneType` `None` value here instead, just to be clear that there is no filename vs. potentially having a file named `"None"` 😅 It also helps with simple boolean `if filenames[0]` checks, as `"None"` will evaluate truthily due to be a non-empty string.

package/MDAnalysis/coordinates/chain.py
```diff
@@ -271,6 +271,13 @@ def __init__(self, filenames, skip=1, dt=None, continuous=False,
                         for filename in filenames]
         self.filenames = np.array([fn[0] if isinstance(fn, tuple) else fn
                                                         for fn in filenames])
+        # Set filenames array to array of "None" if chainreader initialized with
```
lilyminium: Also, it seems to me that the if/else you've added wholly replaces lines 272-273 above :-)

lilyminium (changes requested):
Thanks for making this much-needed fix, @ljwoods2! I left a couple comments -- IMO it'd be better to check all filenames than simply the first one.

The pep8 bot also has flagged a couple lines as not following PEP8 -- could you please fix those?

ljwoods2:
@lilyminium Thank you for the feedback, I appreciate the pointers! Just revised based on your tips and changed repr to make more sense with the potential case of prepending an ndarray to a trajectory, let me know your thoughts.

lilyminium (changes requested):
Thanks for making these changes, they look really good @ljwoods2! This is a bit nitpicky but would you also mind adding a direct check of `self.filenames` (maybe as part of `test_issue_3349`)? You can access the universe's coordinate reader through `u.trajectory[.filenames]`. After that I reckon this would be 100% good to go!

package/CHANGELOG
```diff
@@ -25,6 +25,8 @@ Fixes
    `DistanceMatrix.run(frames=...)` (PR #4433)
  * Fix doctest errors of lib/picklable_file_io.py (Issue #3925, PR #4371)
  * Fix deploy action to use the correct version of the pypi upload action.
+ * Fix ChainReader repr() method when sub-reader is MemoryReader
```
```suggestion
 * Fix ChainReader `__repr__()` method when sub-reader is MemoryReader 
```

RMeli (commented):
Nitpick, otherwise LGTM (besides @lilyminium outstanding requested change).

ljwoods2:
@lilyminium Done, let me know what you think

testsuite/MDAnalysisTests/coordinates/test_chainreader.py
```diff
@@ -455,3 +455,23 @@ def wrapped(ts):
...
     assert np.abs(com).max() < 1e-6
+
+
+def test_issue_3349():
+    # test for repr of a chainreader when the subreaders are memoryreaders
+    cdx = np.random.random(3341 * 3 * 10)
+    cdx = cdx.reshape(-1, 3341, 3).astype(np.float32) * 10
+    u = mda.Universe(PSF, (cdx, DCD))
+    u2 = mda.Universe(PSF, (cdx, cdx, cdx, DCD))
+    print(u.trajectory.filenames)
```
lilyminium:
```suggestion
```
Thanks @ljwoods2, looks great! Just one final thing, perhaps let's not keep the print statement?

lilyminium (changes requested):

package/CHANGELOG
```diff
@@ -25,6 +25,8 @@ Fixes
    `DistanceMatrix.run(frames=...)` (PR #4433)
  * Fix doctest errors of lib/picklable_file_io.py (Issue #3925, PR #4371)
  * Fix deploy action to use the correct version of the pypi upload action.
+ * Fix ChainReader `__repr__()` method when sub-reader is MemoryReader
```
hmacdope: Newest first. :)

orbeckst:
@lilyminium could you please push this PR over the finish line as you have the only outstanding review and it seems to be almost done? Thanks!

lilyminium (approved):
LGTM -- thank you for the fix @ljwoods2!

PR #4415, opened 2024-01-07
Feat: Add `n_initial_contacts` attribute and document ```Contacts``` class attributes

### LINKED ISSUE #2604: Document Contacts attributes and add n_initial_contacts
``Contacts`` gives the fraction of initial contacts in ``timeseries``, but I think most people would also like to know how many initial contacts there actually are. 
 
You can calculate this with ``Contacts.initial_contacts[0].sum()``, but it would be nice to just have it there as an attribute e.g. `n_initial_contacts`. Especially since the documentation doesn't list ``initial_contacts`` or ``r0`` as attributes, so many users will not know it's there. So, desired solution:

1. Have `n_initial_contacts`
2. Document the (important) attributes of the class

## Additional context ##

[Example notebook of using the class, and calculating initial contacts, in this user guide notebook.](https://www.mdanalysis.org/UserGuide/examples/analysis/distances_and_contacts/contacts_native_fraction.html#Calculating-fraction-of-native-contacts,-with-a-distance-lower-than-or-equal-to-the-reference-structure)

HeetVekariya:
Fixes #2604 

 - Added ```n_initial_contacts``` attribute to ```Contacts``` class.
 - Documented ```Contacts``` class attributes to its doc string.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

----
📚 Documentation preview 📚: https://mdanalysis--4415.org.readthedocs.build/en/4415/


HeetVekariya:
- I haven't added the testing code, which is needed to be added in the ```test_contacts.py``` ([here](https://github.com/MDAnalysis/mdanalysis/blob/develop/testsuite/MDAnalysisTests/analysis/test_contacts.py)).


## Questions:
1) Do i need to just compare single values for multiple examples using ```assert``` ?
2) How can i test if my "Testing code" is working, in the local ?
3) Can any of the maintainer provide me the possible cases which should be tested with expected values as i don't about it ?
like below,
```python
@pytest.mark.parametrize('pbc', (True, False))
@pytest.mark.parametrize('name, compound', (('molnums', 'molecules'),
                                            ('fragindices', 'fragments')))
# fragment is a fixture defined earlier
def test_center_of_mass_compounds_special(self, fragment,
                                          pbc, name, compound):
    ref = [a.center_of_mass() for a in fragment.groupby(name).values()]
    com = fragment.center_of_mass(pbc=pbc, compound=compound)
    assert_almost_equal(com, ref, decimal=5)
```
4) Is there any change required in the doc string for `version changed` ?

HeetVekariya:
- @lilyminium can you give me examples which i have to check in the test code, like i have shown in my above message.
- Currently for trial purpose i am using below code which gives 0 as output, which i think may be not good for testing.

```python 
import numpy as np
import MDAnalysis as mda
from MDAnalysisTests.datafiles import PSF, DCD  # Adjust as per available test files
from MDAnalysis.analysis import contacts  # Adjust the import according to your project structure

u = mda.Universe(PSF, DCD)  # Load test universe
select = ('protein', 'not protein')  # Define atom selections
refgroup = (u.select_atoms('protein'), u.select_atoms('not protein'))  # Reference groups

# Instantiate Contacts object
conts = contacts.Contacts(u, select, refgroup)

# Expected number of initial contacts (this should be determined or calculated for your test case)
print(conts.n_initial_contacts)
```

package/MDAnalysis/analysis/contacts.py
```diff
@@ -405,6 +405,13 @@ def __init__(self, u, select, refgroup, method="hard_cut", radius=4.5,
              Show detailed progress of the calculation if set to ``True``; the
              default is ``False``.
 
+        Attributes
+        ----------
+        n_initial_contacts : int
+             Total number of initial contacts.
+        r0 : list
```
lilyminium:
```suggestion
        r0 : list[numpy.ndarray]
```

In Python typing, you can use `[]` to specify what type is inside the container too!

lilyminium (changes requested):
Sorry it's taken us some time to get to reviewing this @HeetVekariya, and thank you very much for being patient.

1. Yes, I think two fairly simple tests just checking the value of `n_initial_contacts` with `assert` would do here! I say two because there are two cases you've specified here -- one if `len(self.initial_contacts) > 0` (lines 465-466), and one if not (lines 467-468).
2. If you [create a development environment](https://userguide.mdanalysis.org/dev/contributing_code.html#creating-a-development-environment), you can additionally install [pytest](https://docs.pytest.org/en/7.4.x/). From there, if you navigate to `mdanalysis/testsuite/MDAnalysisTests/`, you can run `pytest .` and that will run all the tests. As the test suite is quite large, this can take quite a while, so it might be faster to navigate to `mdanalysis/testsuite/MDAnalysisTests/analysis` and run `pytest test_contacts.py` to specifically run the Contacts tests.
3. You might need to play around with some examples yourself, as I don't know any expected values off the top of your head. The example code you've pasted isn't a very good test case as the PSF/DCD files *only* contain proteins:
```
import numpy as np
import MDAnalysis as mda
from MDAnalysisTests.datafiles import PSF, DCD  # Adjust as per available test files
from MDAnalysis.analysis import contacts  # Adjust the import according to your project structure

u = mda.Universe(PSF, DCD)  # Load test universe
select = ('protein', 'not protein')  # Define atom selections
refgroup = (u.select_atoms('protein'), u.select_atoms('not protein'))
refgroup[1]
```
```
<AtomGroup with 0 atoms>
```

It would be better to use two groups that contain atoms in both, such as different amino acids (e.g. `resname ALA` or `resname CYS`). Even better would be to use a system like the TPR/XTC files (`from MDAnalysisTests.datafiles import TPR, XTC` and `u = mda.Universe(TPR, XTC)`), which contain both protein and solvent.

4. I think this change is small enough that it doesn't warrant a `versionchanged`.

HeetVekariya:
- @lilyminium Thank you for reviewing.
- I have added the tests, can you please review it whenever you are available.

package/MDAnalysis/analysis/contacts.py
```diff
@@ -455,6 +462,11 @@ def __init__(self, u, select, refgroup, method="hard_cut", radius=4.5,
                                                 box=self._get_box(refA.universe)))
                 self.initial_contacts.append(contact_matrix(self.r0[-1], radius))
 
+        if len(self.initial_contacts) > 0:
+            self.n_initial_contacts = self.initial_contacts[0].sum()
+        else:
+            self.n_initial_contacts = 0
```
lilyminium: Could you please add a test for this `else` condition? It looks like you can trigger this condition if `refgroup=[]` is an empty list.

lilyminium (changes requested):
Thanks for adding the tests @HeetVekariya, they look great. Could you please also test if `len(self.initial_contact) == 0`? Then I think this will be good to go!

Also, please add this to CHANGELOG! It's probably an Enhancement :-)

package/MDAnalysis/analysis/contacts.py
HeetVekariya: - @lilyminium I noticed something that, my **```if```** condition checks for **```initial_contacts```**, which can never be of length 0, as it is always ```[array([], shape=(3341, 0), dtype=bool)]``` because of initialisation. So my **```else```** case may never get executed or might be i am wrong.
- So what can be done further from these options?
  1) If my explanation is correct then we can remove the **```else```** case from the **```contacts.py```** file.
  2) You can give me examples which initialises **```initial_contacts```** list as an empty list.

lilyminium: Ok yeah, looking again at the code I don't think refgroup can be an empty list without running into errors earlier. In that case I agree, the else condition isn't necessary.

HeetVekariya:
@lilyminium Thank you for guiding me. I have made changes as requested.

RMeli:
Restarted the failing test because of timeout.

package/MDAnalysis/analysis/contacts.py
```diff
@@ -455,6 +462,10 @@ def __init__(self, u, select, refgroup, method="hard_cut", radius=4.5,
                                                 box=self._get_box(refA.universe)))
                 self.initial_contacts.append(contact_matrix(self.r0[-1], radius))
 
+        if len(self.initial_contacts) > 0:
+            self.n_initial_contacts = self.initial_contacts[0].sum()
```
lilyminium: I think you pointed out that `self.initial_contacts` must always be longer than 0 -- do you still need the `if` here?

lilyminium (changes requested):
Thanks for making the change @HeetVekariya, looking pretty good! I just have one comment, and the pep8 bot has noticed some trailing whitespace -- could you just fix that up please? Then I think it's good to go!

package/MDAnalysis/analysis/contacts.py
HeetVekariya: Fair, i remove the ```if``` statement.

lilyminium (approved):
Thanks @HeetVekariya! Everything LGTM pending tests working -- thank you very much for contributing this fix!

HeetVekariya:
- @lilyminium Thank you for guiding consistentently 🙏
- Is there anything left from me on this PR ?

RMeli:
Looks like we had another timeout. I restarted the failed job.

PR #4025, opened 2023-02-13
Solvant Accesible surface area

pegerto:
Fixes #2439

Changes made in this Pull Request:
 -  Add an implementation of Shrake-Rupley algorithm to calculate SASA per atom.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?
 - [x] Issue raised/referenced?

PR #4162, opened 2023-06-05
[GSoC] Parallelisation of AnalysisBase with multiprocessing and dask

marinegor:
Fixes #4158
~~Also fixes # 4259 as a check in `AnalysisBase.run()`~~ no it doesn't, I'll do another PR that does that.

~~Related to #4158, does not help f-i-x-i-n-g (cudos to github bot) the issue per se but paves the way towards that.~~

Changes made in this Pull Request:
- several methods added `analysis.base.AnalysisBase`, implementing backend configuration, splitting of the frames for analysis, computation, and results aggregation
- module `analysis.backends` introduces `BackendBase` class, as well as built-in backends `BackendMultiprocessing`, `BackendSerial` and `BackendDask`, implementing the `apply` method for computations using various backends
- module `analysis.results` introduces `ResultsGroup` class that allows for merging of multiple uniform `Results` objects from the same module, given appropriate aggregation functions

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?
 - [x] Issue raised/referenced?

----
:books: Documentation preview :books:: https://mdanalysis--4162.org.readthedocs.build/en/4162/

PR #4417, opened 2024-01-08
Accessible Surface Area calculations

### LINKED ISSUE #2439: Molecular volume and (solvent-accessible) surface area
Hi, I would like to suggest two features. I couldn't find how to calculate the _molecular volume_ and SASA (_solvent-accessible surface area_) for a trajectory. Do you think those would be useful additions?

JureCerar:
Fixes #2439

 - Added calculation of the accessible surface area using Shrake-Rupley algorithm (modified #4025).
 - Added calculation of relative accessible surface area.
  
PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

----
📚 Documentation preview 📚: https://mdanalysis--4417.org.readthedocs.build/en/4417/


IAlibay (changes requested):
I'm going to put a blocker here because I'm not really sure how to best go about this one - given that this is acknowledged as a modification of #4025 and that @pegerto has their own MDAKit for this now - https://github.com/pegerto/mdakit_sasa/ (which I assume is also based off their own work in #4025).

Thoughts everyone? Particularly pinging @MDAnalysis/coredevs

RMeli:
It is unfortunate that the MDAKit was not more widely publicized, leading to duplication of effort. @JureCerar, would it make sense for you to contribute the calculation of relative accessible surface area to the MDAKit (assuming it has not yet been implemented)?

orbeckst (changes requested):
@JureCerar this is an impressively complete contribution with documentation and extensive testing.

You mentioned that you based your work on @pegerto's PR #4025. Could you please comment on which parts of PR #4025 you used? This is important because some parts of PR #4025 may have been inspired by other work and we need to understand where code comes from. 

Have you compared output and performance of the code here to [mdakit-sasa](https://mdakits.mdanalysis.org/mdakit-sasa.html#mdakit-sasa)?

If we proceed with the PR then we would also add @pegerto as an author to give proper credit. We would also add a note to the docs that [mdakit-sasa](https://mdakits.mdanalysis.org/mdakit-sasa.html#mdakit-sasa) exists as an alternative.

JureCerar:
@orbeckst I used and modified the main surface calculation code. As far as I know, the code from #4025 is copied from [BioPython/SASA](https://github.com/biopython/biopython/blob/master/Bio/PDB/SASA.py) which is under BSD 3 license. Only `_get_sphere` and `_single_frame` are based of BioPython’s implementation. I changed the code it to fit MDAnalysis `AnalysisBase` class, added some tweaks, input values checks, and comments. Everything else is my own code: Relative SASA, tests, documentation, etc.

orbeckst:
Thank you for the details. BSD 3 would be ok.

Have you compared output and performance of the code here to [mdakit-sasa](https://mdakits.mdanalysis.org/mdakit-sasa.html#mdakit-sasa)?

What does your code and mdakit-sasa have in common, where do they differ?

JureCerar:
I checked the code. The main difference is mdakit-sasa is a wrapper for [FreeSASA](https://freesasa.github.io/) package. So the underlying algorithm is different. FreeSASA uses [Lee-Richards](https://freesasa.github.io/doxygen/Geometry.html) algorithm where as this code uses [Shrake-Rupley](https://en.wikipedia.org/wiki/Accessible_surface_area) algorithm. 

Performance wise I did not test it. But I figure FreeSASA (mdkit-sasa) is faster, as it's implemented in C? It's hard to make head-to-head comparison as the algorithm is different. This implementation finishes a 10 frame trajectory of a ~400 residue protein in about a minute or two, which I think is a reasonable speed. In any case, precision can be lowered if speed is needed.

Output wise, the result (i.e. area) is the same regardless of the method or package used.

Here it's also implemented the [Relative Surface Area](https://en.wikipedia.org/wiki/Relative_accessible_surface_area) calculation which is a very useful to have when calculating protein surface properties. I guess it could also be implemented in mdkit-sasa?

Just as a side note. I similarly tried writing a wrapper for `BioPython/SASA` but it was very messy and I could not get it to work properly without writing a lot of temporary files.

pegerto:
Hi All.

As mention by @JureCerar  mda_kit wraps the implementation [FreeSASA](https://freesasa.github.io/) in the BaseAnalysis class, and this kit is very simple as all  the heavy lifting is done by FreeSASA: 

* Free Sasa provides both [Lee-Richards](https://freesasa.github.io/doxygen/Geometry.html) and [Shrake-Rupley](https://en.wikipedia.org/wiki/Accessible_surface_area) defaulting to the[ first one](https://github.com/freesasa/freesasa-python/blob/master/src/parameters.pyx#L51-L52)
* Includes a classifier that include a more comprehensive radii that mimics [NACESS](https://biopython.org/docs/dev/api/Bio.PDB.NACCESS.html)

Regarding performance, perforce is heavily driven by parametrisation, in the case of the Shanke-Rupley the number of points of the spheres are a main parameter if you use Gromacs SASA calculation the default parameters use very few points, FreeSASA have a nThread implementation builtin, but the kit do not implement parallelisation over multiple frames at the moment.

The reason for switching the PR to a kit initially was to separate FreeSASA dependency from core. Let me knot if there is something I can help with.

Regards.

orbeckst:
Thanks @pegerto and @JureCerar ! Some of the developers are currently discussing how to best move forward. We'll keep you updated. Thank you for your patience!

orbeckst:
>  kit do not implement parallelisation over multiple frames at the moment.

This might be very easy once we merge PR #4162 .

orbeckst:
Please see https://github.com/MDAnalysis/mdanalysis/issues/2439#issuecomment-2397437111

JureCerar:
Improvements in this version:
* `SASA` and `RSASA` calculation now relies on `radii` topology attribute instead of passing `radii_dict`  
* Improved documentation  
* Added parallelization support  
* Added testing against results from other software (FreeSASA)

Question to @orbeckst: As SASA calculations depend on radii topology information, would it be worth implementing radii guesser to `guess_TopologyAttrs`? It's a nice information to have (also for other calculations). Currently it can be easily done, for example:

```python
from MDAnalysis.guesser.tables import vdwradii
default_radius = 2.0
if not hasattr(u, 'radii'):
    u.add_TopologyAttr('radii')
    u.atoms.radii = [vdwradii.get(e, default_radius) for e in u.atoms.elements]
```

All the tools are there (and if I understood it correctly, bond guesser anyways needs radii information to find bonds). It would just be very convenient for users to simply type:

```python
u.guess_TopologyAttrs(to_guess=('radii',))
```

orbeckst:
Radius guessing makes sense for this application.

I think it would go into the default guesser.

Once we get guessers for eg coarse grained MARTINI simulations, that guesser could guess sensible radii for coarse grained particles.

PR #4422, opened 2024-01-11
Used between to update donors_sel and acceptors_sel

### LINKED ISSUE #4130: potential performance improvements for H-bond analysis 'between'

In the review of PR #4092 , @richardjgowers observed that there's an opportunity for improving the performance for the "between" keyword of HydrogenBondAnalysis  https://github.com/MDAnalysis/mdanalysis/pull/4092/#discussion_r1152956537 , namely to filter atomgroups for "between" before `capped_distances()` is called to calculate distances.

Follow up on the suggestion and implement the suggested improvement and benchmark the old code vs the updated one.

Do nothing — that's fine, the update is not mission-critical.

## Additional context
Discussion https://github.com/MDAnalysis/mdanalysis/pull/4092/#discussion_r1152956537 around https://github.com/MDAnalysis/mdanalysis/blob/3ebd5ecfa097f296c493d5b75fdab280fe1510c8/package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py#L735

* @richardjgowers : [...] would it make more sense to filter before the call to capped_distances? https://github.com/MDAnalysis/mdanalysis/blob/3ebd5ecfa097f296c493d5b75fdab280fe1510c8/package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py#L712
* @p-j-smith : good point - that would definitely be better, but it's probably not a straightforward change and I imagine there are lots of ways to do it. One way would be to iterate over the pairs of atom selections passed to `between` and calculate donor-acceptor distances for each pair. So e.g. if `between=[["protein", "SOL"], ["protein", "protein"]]`, we would calculate:

   - protein donor to SOL acceptor distances
   - protein acceptor to SOL donor distances  
   - protein donor to protein acceptor distances
   
  and concatenate them.

Sumit112192:
Fixes #4130 

 - I used the `between` keyword to update the `self.donors_sel` and `self.acceptors_sel` keywords to reduce the search space for the capped_distance function.
 - I want to know if my logic is correct so that I can think more in that direction.

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

## Developer certificate of origin

----
📚 Documentation preview 📚: https://mdanalysis--4422.org.readthedocs.build/en/4422/


Sumit112192:
In the hydrogen  bond analysis, if the user passed donor_sel, it might cause a pair of results where both hydrogen and donor have the same index, essentially the same hydrogen atom, since we were not putting any min_cutoff constraint in the `capped_distance` function of  _get_dh_pairs() function

Sumit112192:
Can someone guide me if I am going in the right direction here?

orbeckst:
@richardjgowers could you please shepherd this PR? You're the one who originally made the performance improvement suggestion in #4130. Thanks!

PR #4423, opened 2024-01-12
Fix bug in PCA trajectory iteration -- avoid explicit usage of self.start

### LINKED ISSUE #4425: analysis.pca.PCA can't be run with `frames=...`
## Expected behavior ##

I want to be able to run `analysis.pca.PCA` specifying certain frames:

```python
import MDAnalysis as mda
from MDAnalysis.analysis.pca import PCA

u = mda.Universe(PSF, DCD)
PCA(u, select='backbone').run(frames=[0, 1])



## Actual behavior ##

It gives:

```
Cell In[3], line 1
----> 1 PCA(u, select='backbone').run(frames=[0,1])

File /nvmeraid/github/mdanalysis/package/MDAnalysis/analysis/base.py:436, in AnalysisBase.run(self, start, stop, step, frames, verbose, progressbar_kwargs)
    433 self._setup_frames(self._trajectory, start=start, stop=stop,
    434                    step=step, frames=frames)
    435 logger.info("Starting preparation")
--> 436 self._prepare()
    437 logger.info("Starting analysis loop over %d trajectory frames",
    438             self.n_frames)
    440 for i, ts in enumerate(ProgressBar(
    441         self._sliced_trajectory,
    442         verbose=verbose,
    443         **progressbar_kwargs)):

File /nvmeraid/github/mdanalysis/package/MDAnalysis/analysis/pca.py:250, in PCA._prepare(self)
    248 def _prepare(self):
    249     # access start index
--> 250     self._u.trajectory[self.start]
    251     # reference will be start index
    252     self._reference = self._u.select_atoms(self._select)

File /nvmeraid/github/mdanalysis/package/MDAnalysis/coordinates/base.py:849, in ProtoReader.__getitem__(self, frame)
    847         return FrameIteratorSliced(self, frame)
    848 else:
--> 849     raise TypeError("Trajectories must be an indexed using an integer,"
    850                     " slice or list of indices")

TypeError: Trajectories must be an indexed using an integer, slice or list of indices
```


See above


- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) --> 2.8.0-dev0 (commit 8ff8acc3bd)
- Which version of Python (`python -V`)? --> 3.10
- Which operating system? --> Ubuntu 22.04.3 LTS

### PR DESCRIPTION
**Pull request by marinegor**
This fixes a bug which made impossible the 
Fixes #4425 

 - access first frame in trajectory with `self._sliced_trajectory[0]` instead of `self._trajectory[self.start]`, since `self.start` is `None` if run is initiated with `a.run(frames=[1,2])`.
 

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

----
📚 Documentation preview 📚: https://mdanalysis--4423.org.readthedocs.build/en/4423/

### CODE REVIEW

File: package/MDAnalysis/analysis/pca.py [hunk 1]
```diff
@@ -247,7 +247,8 @@ def __init__(self, universe, select='all', align=False, mean=None,
 
     def _prepare(self):
         # access start index
-        self._u.trajectory[self.start]
+        # self._u.trajectory[self.start]
+        self._sliced_trajectory[0]
```
hmacdope, line 257: I could be going insane but do we not need to pull the first index from `frames` or is that already handled? @orbeckst ?

IAlibay: Not sure I fully understand your question I think I had a similar thought but unsure - do you mean this? https://github.com/MDAnalysis/mdanalysis/blob/846131cb4a16cea48f3ccc56dce9a82995c8be27/package/MDAnalysis/analysis/base.py#L357-L366

hmacdope: Yes sorry, poorly worded by me but you read my mind. Indeed it is safe to take the 0th index as the trajectory is pre-sliced on L366.

hmacdope: Thanks @IAlibay!

hmacdope:
Could you also raise an issue @marinegor just so we have a record of what went wrong?

package/MDAnalysis/analysis/pca.py
IAlibay: So my question here is whether or not this behaviour is defined anywhere. There's a versionchanged somewhere but quickly scanning the docstring there's no info on this right? (maybe I'm missing it somewhere)

IAlibay: This behaviour == the behaviour in this analysis class to take the first frame for the mean selections.

marinegor:
hi @hmacdope , I raised the issue -- should I do anything else?
I'm not sure if the discussion above is regarding this PR, especially since I feel I can't contribute much to it😁

package/MDAnalysis/analysis/pca.py
hmacdope: @IAlibay I could be interpreting this wrong but it seems like this fixes a bug to make things behave the way they should no? It is just enforcing consistent slicing behaviour between the different ways of slicing the trajectory?

IAlibay: Please don't take my comment here as "this isn't the right fix", but rather "should there be better documentation" (which doesn't necessarily have to be handled here but if it's a few lines it could?)

To explain how I got here - I started reviewing, got to the code and went "ok makes sense", then looked around for docs to make sure it was the right behaviour and couldn't find it.

hmacdope: Ah I see what you mean. Perhaps adding the below to the docs would make it clearer? 
```
When using mean selections, the first frame of the selected trajectory slice is used as a reference.
```

IAlibay: Yeah I think so? I'm probably missing something really obvious somewhere though.

Also as mentioned above, this can just be a separate issue, we don't have to fix it here if @marinegor doesn't feel like it.

marinegor:
So anyway, I added `versionchanged` to the class documentation describing what's going on -- hopefully that'll be enough for the PR!

package/MDAnalysis/analysis/pca.py
```diff
@@ -230,6 +230,12 @@ class PCA(AnalysisBase):
        ``mean`` input now accepts coordinate arrays instead of atomgroup.
        :attr:`p_components`, :attr:`variance` and :attr:`cumulated_variance`
        are now stored in a :class:`MDAnalysis.analysis.base.Results` instance.
+    .. versionchanged:: 2.8.0
+       ``self.run()`` can now appropriately use ``frames`` parameter (bug
+       described by #4425 and fixed by #4423). Previously, behaviour was to
+       manually iterate through ``self._trajectory``, which starting #3415 that
+       introduced ``frames`` is wrong, and should use ``self._sliced_trajectory``.
+       It is now fixed.
```
hmacdope:
```suggestion
```

hmacdope (changes requested):
@marinegor could you add 

```
When using mean selections, the first frame of the selected trajectory slice is used as a reference.
```

To the main docstring of `PCA` class? other than that looks good.

marinegor:
@hmacdope done both (applied suggestion in my commit)

hmacdope (approved):
I will approve but would love a review from @IAlibay before going ahead and merging. Well done @marinegor!

IAlibay:
@hmacdope happy to have a quick look tomorrow, could you do a review request for me please? (sorry I'm on phone)

hmacdope:
> @hmacdope happy to have a quick look tomorrow, could you do a review request for me please? (sorry I'm on phone)

I think you are already requested.

package/CHANGELOG
```diff
@@ -19,6 +19,7 @@ The rules for this file:
  * 2.8.0
 
 Fixes
+ * Fix bug in PCA -- wasn't possible to use `frames=...` syntax before.
```
```suggestion
 * Fix bug in PCA preventing use of `frames=...` syntax (PR #4423)
```
We've not been amazing on PR numbers but this is actually super useful when doing the release notes! (mostly so I don't have to edit things afterwards)

package/MDAnalysis/analysis/pca.py
```diff
@@ -142,7 +142,8 @@ class PCA(AnalysisBase):
     generates the principal components of the backbone of the atomgroup and
     then transforms those atomgroup coordinates by the direction of those
     variances. Please refer to the :ref:`PCA-tutorial` for more detailed
-    instructions.
+    instructions. When using mean selections, the first frame of the selected
```
IAlibay, line 145: this is great thanks!

package/MDAnalysis/analysis/pca.py
```diff
@@ -230,6 +231,11 @@ class PCA(AnalysisBase):
        ``mean`` input now accepts coordinate arrays instead of atomgroup.
        :attr:`p_components`, :attr:`variance` and :attr:`cumulated_variance`
        are now stored in a :class:`MDAnalysis.analysis.base.Results` instance.
+    .. versionchanged:: 2.8.0
+       ``self.run()`` can now appropriately use ``frames`` parameter (bug
+       described by #4425 and fixed by #4423). Previously, behaviour was to
+       manually iterate through ``self._trajectory``, which starting #3415 that
+       introduced ``frames`` is wrong, and should use ``self._sliced_trajectory``.
```
IAlibay:
```suggestion
       manually iterate through ``self._trajectory``, which would
       incorrectly handle cases where the ``frame`` argument
       was passed.
```
Just a slight personal preference edit - I always see the versionchanged as intended for users, so what they want to know is "what changed that could have affected my results" rather than "what internal code changes were necessary".

testsuite/MDAnalysisTests/analysis/test_pca.py
```diff
@@ -133,6 +133,12 @@ def test_no_frames(u):
         PCA(u, select=SELECTION).run(stop=1)
 
 
+def test_can_run_frames(u):
+    atoms = u.select_atoms(SELECTION)
+    u.transfer_to_memory()
+    PCA(u, select=SELECTION).run(frames=[0, 1])
```
IAlibay: Since it's a bit of a smoke test of "will this work", and we don't assert on anything, could you add `mean=True` explicitly here? This makes sure we don't accidentally not run the new changes if something happens.
```suggestion
    PCA(u, select=SELECTION, mean=True).run(frames=[0, 1])
```

IAlibay (approved):
Couple of small change suggestions, otherwise lgtm!

marinegor:
@IAlibay 

>Since it's a bit of a smoke test of "will this work", and we don't assert on anything, could you add mean=True explicitly here?

Actually, `mean=None` (=the default option) ensures I run this part of the code afaik (I checked it with `assert False on L286). So I reverted back to my version of the test.

IAlibay:
> @IAlibay
> 
> > Since it's a bit of a smoke test of "will this work", and we don't assert on anything, could you add mean=True explicitly here?
> 
> Actually, `mean=None` (=the default option) ensures I run this part of the code afaik (I checked it with `assert False on L286). So I reverted back to my version of the test.

Sorry maybe I didn't explain my point well enough - `None` currently has that behaviour, but given we want to ensure it's always going to be called forever into the future, then setting `True` may be better.

That being said - I'm not particularly too fussed here, it's just nice to enforce a branch decision rather than rely on defaults.

marinegor:
@IAlibay I see what you mean now -- added explicit `frames=None` here.
Btw, I firmly believe all the tests are ok but something wrong with the CI💔

RMeli:
>  Btw, I firmly believe all the tests are ok but something wrong with the CI💔

Looked like a timeout. Re-running Azure Pipelines completed successfully.

PR #4424, opened 2024-01-14
Cleaning up packaging for `package`

IAlibay:
I expect a lot to break whilst it's being worked on but should show what the redudant bits are at least.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

----
📚 Documentation preview 📚: https://mdanalysis--4424.org.readthedocs.build/en/4424/

### DIFF
--- a/package/pyproject.toml
+++ b/package/pyproject.toml
@@ -12,9 +12,11 @@ requires = [
     # For unreleased versions of Python there is currently no known supported
     # NumPy version. In that case we just let it be a bare NumPy install
     "numpy<2.0; python_version>='3.13'",
-    "setuptools",
+    # Set to minimum version of setuptools that allows pyproject.toml
+    "setuptools >= 40.9.0",
     "wheel",
 ]
+build-backend = "setuptools.build_meta"
 
 [project]
 name = "MDAnalysis"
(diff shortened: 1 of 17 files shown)


package/pyproject.toml
```diff
@@ -95,13 +95,27 @@ doc = [
...
-
 [project.urls]
-Documentation = 'https://docs.mdanalysis.org/'
-"User Guide" = 'https://userguide.mdanalysis.org/'
-"Issue Tracker" = 'https://github.com/mdanalysis/mdanalysis/issues'
-"User Group" = 'https://groups.google.com/g/mdnalysis-discussion/'
-Discord = 'https://discord.com/channels/807348386012987462/'
-Blog = 'https://www.mdanalysis.org/blog/'
+"Project URL" = "https://www.mdanalysis.org"
+"Download URL" = "https://github.com/MDAnalysis/mdanalysis/releases"
```
RMeli:
```suggestion
Homepage = "https://www.mdanalysis.org"
Downloads = "https://github.com/MDAnalysis/mdanalysis/releases"
```
? Just a suggestion, but "URL" feels a repetition since the field is already `[project.urls]`.

RMeli (approved):

package/setup.py
```diff
@@ -567,112 +560,20 @@ def long_description(readme):
...
-          'mda-xdrlib',
-          'waterdynamics',
-          'pathsimanalysis',
-    ]
-
     setup(name='MDAnalysis',
           version=RELEASE,
-          description='An object-oriented toolkit to analyze molecular dynamics trajectories.',
           long_description=LONG_DESCRIPTION,
           long_description_content_type='text/x-rst',
```
RMeli, line 569: Should this be removed in favour of
```
readme = {file = "README", content-type = "text/x-rst"}
```

I haven't double checked the documentation... I'm just comparing with https://github.com/MDAnalysis/mdanalysis/pull/4463, so feel free to discard if this is a stupid question.

IAlibay: The issue here is that the readme exists outside of the package, so the pyproject.toml approach won't work - i.e. when you have an sdist your readme target will be a non existent file.

The current "trick", for this is to dynamically set the description based on what you can find - since it's only really needed for pypi anyways.

We could long term fix this with some symlinks or just a pre packaging copy step, but probably best not to do so here.

package/pyproject.toml
IAlibay: Looks good, I went with this but with a non-plural Download instead.

IAlibay:
Thanks for the review @RMeli - I've updated stuff based on comments both here and in #4463. I know you've already approved, but would you like to have another look ahead of merge?

RMeli (approved):
LGTM, thanks @IAlibay

PR #4442, opened 2024-01-30
Add new M1 macOS GitHub runner to CI

RMeli:
Add [new M1 macOS runner](https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/) to CI.

----
📚 Documentation preview 📚: https://mdanalysis--4442.org.readthedocs.build/en/4442/


.github/actions/build-src/action.yaml
```diff
@@ -63,8 +63,8 @@ runs:
         which python
         which pip
         pip list
-        conda info
-        conda list
+        # conda info
+        # conda list
```
RMeli: Is this assuming that `conda` is always installed on the base image?

IAlibay: It should - that action is only ever used on workflows that use some kind of mamba/conda. Is it not working?

.github/workflows/gh-ci.yaml
```diff
@@ -37,6 +37,11 @@ jobs:
               python-version: "3.11"
               full-deps: true
               codecov: true
+            - name: macOS_M1
+              os: macOS-14
+              python-version: "3.11"
```
IAlibay: why not 3.12?

.github/workflows/gh-ci-cron.yaml
```diff
@@ -223,7 +223,7 @@ jobs:
     strategy:
         fail-fast: false
         matrix:
-          os: [ubuntu, macos]
+          os: [ubuntu, macos, macos-14]
```
IAlibay: This won't work I think - you'd need to temporarily enable cron CI to double check, but this is doing a full deps install via conda and it'll fail because there are many missing deps for arm64-osx. Just doing the pypi-only check would be fine here.

.github/workflows/gh-ci-cron.yaml
```diff
@@ -269,7 +269,7 @@ jobs:
     strategy:
         fail-fast: false
         matrix:
-          os: [ubuntu, macos, windows]
+          os: [ubuntu, macos, macos-14, windows]
           python-version: ["3.9", "3.10", "3.11"]
```
IAlibay: I've opened a separate PR, that matrix needs bumping up

.github/workflows/gh-ci-cron.yaml
```diff
@@ -269,7 +269,7 @@ jobs:
...
           wheels: ['true', 'false']
```
IAlibay, line 274: would be worth enabling cron CI - I can't remember if non-wheel build will work here, you might need to exclude macos-14 & wheels=false

.github/workflows/gh-ci-cron.yaml
RMeli: Yeah, I added those changes but then realized they are not tested... 

Full install does not work indeed (`hole2`, `distopia`, ...). I agree that PyPI check would be enough here, and minimizes the chances of errors.

IAlibay: you can just change the header of the file to trigger on pull_request and it'll run here

.github/actions/build-src/action.yaml
RMeli: No, it's not working. I think we are using `micromamba`, and `conda` is not there. See [failure](https://github.com/MDAnalysis/mdanalysis/actions/runs/7714658774/job/21027430191#step:6:138).

.github/workflows/gh-ci.yaml
RMeli: Used the same version above. Is 3.12 preferred?

.github/actions/build-src/action.yaml
IAlibay: Ah yes I remember now - on some versions of micromamba `conda` isn't aliased. Maybe switch all these to micromamba calls instead? alternatively we can just guard them with a "ignore failure" on that action step?

.github/workflows/gh-ci.yaml
IAlibay: Yeah - maybe bump both up? I think we held off on updating because there were missing py3,12 "full deps" in macos space, but that _should_ be fine now?

.github/actions/build-src/action.yaml
RMeli: Thanks for the info. I think it's good to know what's up, so I'll go with `micromamba` instead.

.github/workflows/gh-ci-cron.yaml
RMeli: Smart! =)

RMeli: I forgot about this trick, thanks.

.github/workflows/gh-ci-cron.yaml
RMeli: https://github.com/MDAnalysis/mdanalysis/pull/4442/#discussion_r1471995977

RMeli:
`macos-14` jobs are still queuing... Maybe it's work testing just 3.12 in the cron job too?

IAlibay:
@RMeli I think there's a runner limit, I might be the one using that up on my other PR. I can definitely run 4 concurrent runners there.

IAlibay:
Actually I didn't see, these have been queing for 20h?! That looks like a problem with github actions itself. Wouldn't surprise me given the beta nature of these runners. Maybe we're jumping the gun too early?

.github/workflows/gh-ci-cron.yaml
```diff
@@ -269,7 +272,7 @@ jobs:
     strategy:
         fail-fast: false
         matrix:
-          os: [ubuntu, macos, windows]
+          os: [ubuntu, macos, macos-14, windows]
```
IAlibay: @RMeli the reason this isn't running is beacuse it's doing "${os}-latest", so it's trying "macos-14-latest".

The short term fix is just to remove `latest` on 270 and then do:

```suggestion
          os: [ubuntu-latest, macos-latest, macos-14, windows-latest]
```

IAlibay (changes requested):

RMeli:
> That looks like a problem with github actions itself. Wouldn't surprise me given the beta nature of these runners. Maybe we're jumping the gun too early?

Possibly a bit unstable, but I encountered not such problem in https://github.com/RMeli/spyrmsd/pull/102. Maybe it's just that everyone rushed to test them as we did?

.github/workflows/gh-ci-cron.yaml
```diff
@@ -264,14 +267,16 @@ jobs:
...
+    runs-on: ${{ matrix.os }}
     timeout-minutes: 60
     strategy:
         fail-fast: false
         matrix:
-          os: [ubuntu, macos, windows]
+          os: [ubuntu-latest, macos-latest, macos-14, windows-latest]
           python-version: ["3.9", "3.10", "3.11", "3.12"]
           wheels: ['true', 'false']
+          exclide:
```
RMeli:
```suggestion
          exclude:
```

RMeli:
All green, at last.

.github/workflows/gh-ci-cron.yaml
```diff
@@ -1,8 +1,11 @@
...
 on:
-  schedule:
-    # 3 am Tuesdays and Fridays
-    - cron: "0 3 * * 2,5"
+  # schedule:
+  #   # 3 am Tuesdays and Fridays
+  #   - cron: "0 3 * * 2,5"
+  pull_request:
+    branches:
+      - develop
```
RMeli:
```suggestion
  schedule:
    # 3 am Tuesdays and Fridays
    - cron: "0 3 * * 2,5"
```

.github/workflows/gh-ci-cron.yaml
```diff
@@ -264,14 +264,16 @@ jobs:
...
     timeout-minutes: 60
     strategy:
         fail-fast: false
         matrix:
-          os: [ubuntu, macos, windows]
+          os: [ubuntu-latest, macos-latest, macos-14, windows-latest]
           python-version: ["3.9", "3.10", "3.11", "3.12"]
           wheels: ['true', 'false']
+          exclude:
+          - {os: "macos-14", python-version: "3.9"}
```
IAlibay: why is this being disabled? as show in #4445 we should be able to build on 3.9 with macos-14

IAlibay (changes requested):
One blocking question.

P.S. the runner failure you were saying is down to the `runs-on` being incorrect - if it isn't recognise gh actions doesn' t fail but just goes on indeterminate hold.

.github/workflows/gh-ci-cron.yaml
RMeli: >   Version 3.9 was not found in the local cache
>  Error: The version '3.9' with architecture 'arm64' was not found for macOS 14.2.1.
>  The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json

I got this failure. Looking at the `versions-manifest.json` as suggested, it looks like all `3.9` is not provided with `arch: arm64`.

IAlibay:
```suggestion
          - os: "macos-14"
            python-version: "3.9"
```
If it's ok, can we change the style to match how we use it elsewhere?

IAlibay: That's.. odd. I've not glanced upstream, but we should probably keep an eye on updates - it's very possible they couldn't backport stuff properly until they had their own CI runners for this.

.github/workflows/gh-ci.yaml
```diff
@@ -34,9 +34,14 @@ jobs:
           include:
             - name: macOS_monterey_py311
               os: macOS-12
-              python-version: "3.11"
+              python-version: "3.12"
               full-deps: true
               codecov: true
+            - name: macOS_M1
```
IAlibay:
```suggestion
            - name: macOS_14_arm64_py312
```
Just adding a bit more detail for ease of recognition

IAlibay (changes requested):
Couple of extra things. Also please do update the changelog with your handle.

IAlibay (approved):
cheers, lgtm!

PR #4463, opened 2024-02-21
[PEP518] Improve testsuite packaging

IAlibay:
 - Adds pyproject.toml to testsuite
 - Clears up ambigous packaging
 - Keeps setup.py for dynamic things only
 - Update license file
 - Update links info
 - Update some setup.py docstring text
 - Adds various missing init files

PR Checklist
------------
 - Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?
 - Issue raised/referenced?

----
📚 Documentation preview 📚: https://mdanalysis--4463.org.readthedocs.build/en/4463/


package/pyproject.toml
```diff
@@ -97,11 +97,13 @@ doc = [
...
 
 [project.urls]
-Documentation = 'https://docs.mdanalysis.org/'
-"User Guide" = 'https://userguide.mdanalysis.org/'
-"Issue Tracker" = 'https://github.com/mdanalysis/mdanalysis/issues'
-"User Group" = 'https://groups.google.com/g/mdnalysis-discussion/'
-Discord = 'https://discord.com/channels/807348386012987462/'
-Blog = 'https://www.mdanalysis.org/blog/'
+"Project URL" = "https://www.mdanalysis.org"
+"Download URL" = "https://github.com/MDAnalysis/mdanalysis/releases"
```
RMeli: See https://github.com/MDAnalysis/mdanalysis/pull/4424#discussion_r1499912898

package/pyproject.toml
```diff
@@ -97,11 +97,13 @@ doc = [
...
+Documentation = "https://docs.mdanalysis.org/"
+"User Guide" = "https://userguide.mdanalysis.org/"
+"Issue Tracker" = "https://github.com/mdanalysis/mdanalysis/issues"
+"Github Discussions" = "https://github.com/MDAnalysis/mdanalysis/discussions"
+Discord = "https://discord.com/channels/807348386012987462/"
+Blog = "https://www.mdanalysis.org/blog/"
 Twitter = 'https://twitter.com/mdanalysis'
```
RMeli: Add LinkedIn? And other platforms?

Also, Twitter is no more... Should this be changed to X?

testsuite/MDAnalysisTests/parallelism/__init__.py
RMeli, line 1: Nitpick: the other `__init__.py` files all contain one (empty) line, while this is empty. The ones in https://github.com/MDAnalysis/mdanalysis/pull/4424 are all empty.

Just pointing this out because if we will enable `pre-commit` hooks one day, the empty files will be changed, I think.

testsuite/pyproject.toml
```diff
@@ -0,0 +1,170 @@
+[build-system]
+requires = [
+  "setuptools >= 40.9.0",
```
RMeli, line 3: Is it correct that MDAnalysis itself (https://github.com/MDAnalysis/mdanalysis/pull/4424) doesn't require a minimum version?

testsuite/setup.py
```diff
@@ -77,122 +72,16 @@ def run(self):
...
-          zip_safe=False,
-          cmdclass={'sdist': MDA_SDist},
-          )
+    setup(
+        version=RELEASE,
+        install_requires=[
+            'MDAnalysis=={0!s}'.format(RELEASE),  # same as this release!
+            'pytest>=3.3.0', # Raised to 3.3.0 due to Issue 2329
+            'hypothesis',
+        ],
```
RMeli, line 85: Does this remains here because of the requirement that the version match MDAnalysis?

IAlibay: Yes, anything with dynamic matching must be in setup.py

testsuite/MDAnalysisTests/parallelism/__init__.py
IAlibay: I think according to pep8 they should be empty files. I'll go ahead and update everything to match.

testsuite/pyproject.toml
IAlibay: It probably does (because older versions don't support a lot of toml stuff), but we never set it. I might just set it to this value for package too - it's a good "you'll probably have trouble if you go below this" value.

package/pyproject.toml
IAlibay: Would it be better if we just dropped socials completely? It feels a bit silly to have to keep this up to date for just the pypi metadata

testsuite/pyproject.toml
IAlibay: Updated in other PR - now we have a bare minimum pin of 40.9 which is the release that started supporting pyproject.toml.

testsuite/MDAnalysisTests/parallelism/__init__.py
IAlibay: done

package/pyproject.toml
IAlibay: I removed twitter both here and the other PR, I'm hoping the rest should be stable enough.

package/pyproject.toml
IAlibay: Updated

IAlibay:
Thanks for the review @RMeli - I think I've addressed most comments, the social media stuff might need more discussion. I'm of the opinion that the pyproject.toml url links barely get used so we could probably just not think about them too much.

package/pyproject.toml
RMeli: Yes, makes sense to remove socials!

RMeli (approved):
Thanks Irfan, LGTM

PR #4464, opened 2024-02-23
Update for external mdahole2 mdakit

### LINKED ISSUE #4414: Replace hole2 imports with mdakit-hole2 imports
Once we have packages for mdakit-hole2 we should replace the hole2 code in the MDAnalysis library with that of the MDAKit, this will significantly reduce maintenance burdens.

IAlibay:
Fixes #4414 

 - Switches all imports in the hole2 analysis module to use the mdakit mdahole2.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

----
📚 Documentation preview 📚: https://mdanalysis--4464.org.readthedocs.build/en/4464/


package/MDAnalysis/analysis/hole2/__init__.py
```diff
@@ -34,280 +34,14 @@
...
-         description="HOLE trajectory analysis with orderparameters",
-         path="MDAnalysis.analysis.hole2",
-         cite_module=True)
-del Doi
+import warnings
+from mdahole2.analysis.hole import hole, HoleAnalysis
+from mdahole2.analysis import utils, templates
+from mdahole2.analysis.utils import create_vmd_surface
+
+wmsg = ("Deprecation in version 2.8.0\n"
```
RMeli:
```suggestion
wmsg = ("Deprecated in version 2.8.0\n"
```
?

package/doc/sphinx/source/documentation_pages/analysis/hole2.rst
```diff
@@ -18,6 +18,18 @@ from http://www.holeprogram.org/ or as source from
 https://github.com/osmart/hole2. (HOLE is open source and available
 under the Apache v2.0 license.)
 
+
+.. deprecated:: 2.8.0
+  This module is deprecated in favour of the mdakit
+  `mdahole2 <https://www.mdanalysis.org/mdahole2/>`_ and
+  will be removed in MDAnalysis 3.0.0.
```
RMeli:
```suggestion
   This module is deprecated in favour of the mdakit
   `mdahole2 <https://www.mdanalysis.org/mdahole2/>`_ and
   will be removed in MDAnalysis 3.0.0.
```
Is this indentation correct?

package/pyproject.toml
```diff
@@ -44,6 +44,7 @@ dependencies = [
     'mda-xdrlib',
     'waterdynamics',
     'pathsimanalysis',
+    'mdahole2',
```
RMeli, line 47: This is going to be annoying to package upstream, but it's a "me" problem.

RMeli (approved):
LGTM. The only (non-blocking) question I have is: it is really worth it to add an additional external dependency for something that is deprecated?

IAlibay:
Thanks @RMeli 

Re: extra dependencies - unfortunately this is a temporary pain to avoid another one. The reasons for importing things directly rather than keeping the old code around are 1) to avoid having to back port a bunch of stuff (happened a few times already for hole2), 2) to get folks used to the idea of the mdakit - i.e. "you've been using them all along", is better than "ok starting from now you have to trust this instead".

From a packaging perspective, I have only been thinking about pypi and conda-forge, sorry :( How much work is it elsewhere? I would have hoped that folks were just pulling from pypi for pure python projects.

IAlibay:
I'm going to cc @orbeckst @fiona-naughton @lilyminium into this discussion.

My opinion here is very likely biased since I've spent a decent chunk of time on the mdakit migrations. However, we should make sure that everyone is happy with the plan to add in these temporary core dependencies.

RMeli:
If the module is deprecated we don't need to back port stuff IMO, we can just point to the MDAKit if users have problems (which the code itself already does).

Packaging upstream is not too hard, but you still need to write the package file etc. I should do it anyways for the MDAKits we own.

However, my concern here is adding an external dependency when we know we will remove it. It looks like the worst option, the other two being (1) deprecate but leave the old code until removal (without backporting fixes and improvements) or (2) leave the MDAKit as an external dependency, i.e. use the MDAKit code but don't deprecate the module.

IAlibay:
> If the module is deprecated we don't need to back port stuff IMO, we can just point to the MDAKit if users have problems (which the code itself already does).


I disagree, it's deprecated but we still need to ensure that folks get the right outcomes, i.e. it's deprecated for us not for them. The code here pointing to the new kit is actually a mistake - when we originally created mdahole2 we hadn't packaged it, so we included a deprecation notice but didn't replace the imports.

> Packaging upstream is not too hard, but you still need to write the package file etc. I should do it anyways for the MDAKits we own.

That would be useful if you want to do it, but it could also fall under the scope of someone working in EOSS4 if it's time you'd rather not spend on this task.

> However, my concern here is adding an external dependency when we know we will remove it. It looks like the worst option, the other two being (1) deprecate but leave the old code until removal (without backporting fixes and improvements) or (2) leave the MDAKit as an external dependency, i.e. use the MDAKit code but don't deprecate the module.


I think the issue here is that you end up with a mess of fixes that are done mostly here and need to get moved to mdahole2. Then we keep facing having to tell folks "no actually do this change over there", to which folks will go "but I'm not using that thing I'm using the core library" or if it's ourselves "but not fixing this here means that we don't get to support Y in the core library".

A good example of this is doing numpy 2.0 fixes. MDA 3.0 will likely come out after numpy 2.0. If we want to support it then we have to make relevant fixes here and then also in the kit. It's not excessively problematic, but it does have a time cost.

w.r.t. not deprecating - that's in the territory of roadmap 3.0 stuff and I'm not sure I have a strong enough opinion.

package/CHANGELOG
```diff
@@ -34,6 +34,9 @@ Enhancements
    `n_initial_contacts` attribute, with documentation. (Issue #2604, PR #4415)
 
 Changes
+ * MDAnalysis.analysis.hole2 is now directly imported from the mdakit
+   mdahole2. This module is deprecated and will be full removed in
```
orbeckst:
```suggestion
   mdahole2. This module is deprecated and will be fully removed in
```

maintainer/conda/environment.yml
```diff
@@ -29,7 +29,11 @@ dependencies:
   - tidynamics>=1.0.0
   - tqdm>=4.43.0
   - sphinxcontrib-bibtex
+  - mdaencore
+  - waterdynamics
   - pip:
+      - mdahole2
+      - pathsimanalysis
```
orbeckst, line 36: Thanks for adding all these, looks as if they should have been there all along.

Quick question: where do we use conda/environment.yml ?

package/pyproject.toml
orbeckst: (EDIT: This comment refers to @RMeli 's comment of downstream pain in packaging)

Because you need to make hole2 available in spack?

orbeckst (approved):
Thanks for doing all the work @IAlibay .

I agree with https://github.com/MDAnalysis/mdanalysis/pull/4464#issuecomment-1962931649 that we need to bite the bullet and increase dependencies while we're still shipping a thin version of mdahole2 — it's on us to make it work while we're going through deprecation, not on our users. Sorry @RMeli — I understand the desire to make clean breaks but that seems too user-unfriendly.

RMeli:
> deprecate but leave the old code until removal (without backporting fixes and improvements)

To clarify my comment, which was poorly phrased, I was not advocating to leave the code go bad. I was thinking of just not backporting fixes or enhancement from the MDAKit to change the behaviour (i.e. if there was a bug pre-deprecation, it would stay during deprecation). As a user I'd expect deprecated code to work as at time of deprecation (bugs included). 

However, I was not aware of big changes that would mean a lot of work, which would incur in high maintenance costs (and are therefore not viable/acceptable).

> w.r.t. not deprecating - that's in the territory of roadmap 3.0 stuff and I'm not sure I have a strong enough opinion.

Yeah, I'm just wondering if it's really worth removing a thin wrapper to an MDAKit once it's already in core.

package/pyproject.toml
RMeli: I thought so, but looking at [mdahole2](https://github.com/MDAnalysis/mdahole2)'s dependencies it does not look like hole2 is provided, is it? So packaging mdahole2 would be easier, although I don't think it would make much sense to provide a downstream Spack package without hole2 TBH.

orbeckst: We have a hole2 cf package https://anaconda.org/conda-forge/hole2 for Linux and osx — just not for Windows.

orbeckst: We didn't include it as a dependency because ... well, because we only got the conda-forge package recently and used the reasoning for the pip package (see also https://github.com/MDAnalysis/mdahole2/issues/25 ) that people should be able to use their own hole2 installation (as it worked previously).

orbeckst: See https://github.com/MDAnalysis/mdahole2/issues/55

package/doc/sphinx/source/documentation_pages/analysis/hole2.rst
IAlibay: I think it doesn't matter all that too much with sphinx, but I've added it, thanks!

package/pyproject.toml
IAlibay: Just to clarify the current situation here:

1. PyPi mdahole2 / pure pip install - has no hole2 dependency. This is because we can't distribute hole2 over pypi at the moment.
2. conda-forge mdahole2-base: has no hole2 dependency - this exists for all the os & architecture types where hole2 can't be used. The combination of MDAnalysis & mdahole2-base effectively replicates the current status of the MDA core library where we ask folks to install hole2 themselves.
3. conda-forge mdahole2: for x86_64 linux and macos this is mdahole2-base + hole2

package/CHANGELOG
IAlibay: Cheers, added.

maintainer/conda/environment.yml
IAlibay: This is only used for RTD docs building.

IAlibay:
> However, I was not aware of big changes that would mean a lot of work, which would incur in high maintenance costs (and are therefore not viable/acceptable).


I wouldn't say "high maintenance cost" - porting fixes across the core library & kits isn't a big thing in isolation (couple of lines here and there), but it becomes an extra thing to think about / do when you already have a tiny number of maintainers. A couple half hours here and there and the cost starts really adding up.

IAlibay:
I'll leave this PR up for a little while longer so that folks get to discuss a bit more - we can merge when everyone is happy.

maintainer/conda/environment.yml
orbeckst: thanks

package/pyproject.toml
orbeckst: Thanks for explaining. I won't resolve the comment, just so that it remains visible.

PR #4502, opened 2024-03-10
[CI] MDAnalysis self-dependency build fix

### LINKED ISSUE #4492: [CI] Dependency resolution issues MDAnalysis installs
_I'll update as I get more information, but here is a very quick overview of the issue:_

Due to the use of non-optional mdakits as dependencies, the MDAnalysis package now technically depends on itself. This is causing issues, particularly in CI, where we are installing MDA 2.7.0 via conda-forge and then pip installing develop on top of it. Because we are building without isolation, this is causing environment clash issues.

## Solution?

A quick solution here is maybe to pip install the kits with --no-deps ahead of the MDA install? It's not pretty but it could be an easy short term solution.

IAlibay:
Fixes #4492

Work in progress:
- Currently checking if we can just pick up the build via the src build directly

TODO:
- [x] Check package building
- [x] Check cron builds

 - MDAKits are no longer in the direct pre-build install lists, they get picked up during the src build.

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

----
📚 Documentation preview 📚: https://mdanalysis--4502.org.readthedocs.build/en/4502/


IAlibay:
Cron failures are good - they are pointing out to an existing issue that needs fixing in our conda-forge package.

IAlibay:
conda-forge tests are now just related to rdkit - this is a known failure (and fixed in develop), nothing else to do here.

yuxuanzhuang (approved):
lgtm!

.github/workflows/gh-ci-cron.yaml
```diff
@@ -4,6 +4,11 @@ on:
     # 3 am Tuesdays and Fridays
     - cron: "0 3 * * 2,5"
   workflow_dispatch:
+  # Uncomment when you need to test on a PR
+  pull_request:
```
yuxuanzhuang, line 8: comment out :)

IAlibay: Ah completely forgot to do this!

PR #4533, opened 2024-03-27
Fix: modernise HBA to use AG as objects internally instead of selection strings - code only PR (Issue #3933)

### LINKED ISSUE #3933: modernize HBA to use AG as objects internally instead of selection strings
modernize HBA (HydrogenBondAnalysis) to use AG (AtomGroups) as objects internally instead of selection strings to avoid the need for HBA analysis to discern the relevant atoms.

AG as objects internally instead of selection strings 

## Additional context ##
See PR Hbond update #3848 (in particular https://github.com/MDAnalysis/mdanalysis/pull/3848#discussion_r1014548556)

## Guidance for new contributors
This is a reasonably complicated issue to solve for a beginner. We recommend you first learn about the problem. Ask questions here. The more specific your question, then better we can help.

As the first steps you should:

- Familiarize yourself with [AtomGroups](https://docs.mdanalysis.org/stable/documentation_pages/core/groups.html#MDAnalysis.core.groups.AtomGroup) and [UpdatingAtomGroups](https://docs.mdanalysis.org/stable/documentation_pages/core/groups.html#MDAnalysis.core.groups.UpdatingAtomGroup). Understand the difference between the two.
- Understand the difference between *using [selection strings](https://docs.mdanalysis.org/stable/documentation_pages/selections.html)* to describe groups of atoms and using AtomGroups for the same purpose.
- Read the code for [HydrogenBondAnalysis](https://docs.mdanalysis.org/stable/documentation_pages/analysis/hydrogenbonds.html) in [package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py](https://github.com/MDAnalysis/mdanalysis/blob/develop/package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py). Be able to identify the pieces of code that use selection strings.

EDITS
- made issue clearer for new contributors (@orbeckst  )

lunamorrow:
Fixes #3933 

 - Updated `guess_acceptors`, `guess_hydrogens`, `guess_donators`, `group_categories` and `get_dh_pairs` to remove reliance of HydrogenBondAnalysis on selection strings being passed internally

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

----
📚 Documentation preview 📚: https://mdanalysis--4533.org.readthedocs.build/en/4533/


lunamorrow:
Just leaving some comments here for whoever is assigned to check this PR:

1. I will be submitting a second PR to depreciate the API once I receive any feedback on this one. I want to make sure I have made the appropriate changes here, and any adjustments necessary, before I make that second PR.
2. A block of code is commented out in `_get_dh_pairs` as I was unsure whether the check on line 627 had to be repeated (also in `guess_donors`). In #3933 I have asked about this double up, but decided to proceed with what looked like the best, most robust option. I am also unsure if there is a reason that `self.donors_sel` is not set in `_prepare` the same way that `self.hydrogens_sel` and `self.acceptors_sel` are. I have added this functionality, which should make the check in `_get_dh_pairs` redundant.
3. I have run the current test cases for `test_hydrogenbonds_analysis.py` and have 13 failed, 26 passed, 14 warnings and 5 errors. All of the errors are because `This Universe does not contain charge information`, which I do not think is due to my changes? Of the failures, 6 are from `Can't perform '__eq__' between objects: 'AtomGroup' and 'str'`, so these will require for me to change the test cases and update this PR yes? The remaining failures are due to `hydrogens_ag not being a recognised attribute` (I am unsure how to fix this?), 2x `Assertion Error` (Also unsure on best way to address this) or double ups with the charge information errors.

lunamorrow:
Update: That big block of code that I commented out was necessary. Turns out that its removal was causing 2 failures. I am adding it back in now, and attempting to tackle those 2 assertion errors.

Question: I have added a commit to my branch, and I suspect I can then push updated code into this PR (i.e., as I fix things that were erroring and adjust/add tests)? Is this correct?

orbeckst:
@yuxuanzhuang can you please look after this PR or assign to someone else, please?

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -699,15 +711,25 @@ def _filter_atoms(self, donors, acceptors):
     def _prepare(self):
         self.results.hbonds = [[], [], [], [], [], []]
 
+        # Set unpaired acceptor and hydrogen AtomGroups
+        # self.acceptors_ag = self.guess_acceptors() #NOTE: do I need this?
+        # self.hydrogens_ag = self.guess_hydrogens()
+        # self.donors_ag = self.guess_donors()
+
         # Set atom selections if they have not been provided
         if self.acceptors_sel is None:
```
yuxuanzhuang: You need to make sure acceptors_ag, hydrogens_ag, etc. always exist.
```python
else:
    self.acceptors_ag = self.u.select_atoms(self.acceptors_sel)
````

File: package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py [hunk 2]
```diff
@@ -699,15 +711,25 @@ def _filter_atoms(self, donors, acceptors):
...
+            self.hydrogens_sel = self._group_categories(self.hydrogens_ag)
+        #NOTE: add in self.donors_sel? Is this not done for a reason?
+        if self.donors_sel is None:
+            self.donors_ag = self.guess_donors()
+            self.donors_sel = self._group_categories(self.donors_ag)
 
         # Select atom groups
-        self._acceptors = self.u.select_atoms(self.acceptors_sel,
-                                              updating=self.update_selections)
+        self._acceptors = self.guess_acceptors()
```
yuxuanzhuang: You already have `self.acceptors_ag` above so you don't need to guess it again. Simply parse `self._acceptors = self.acceptors_ag`

yuxuanzhuang (changes requested):
See the in-line review that will fix some of your failed tests. Additionally, please fix other failed tests inside the `test_hydrogenbonds_analysis.py`, mostly related to the switching output of `guess_hydrogens`

yuxuanzhuang:
I also don't have experience handling depreciation before, so I might need help from @hmacdope :)

yuxuanzhuang:
> Question: I have added a commit to my branch, and I suspect I can then push updated code into this PR (i.e., as I fix things that were erroring and adjust/add tests)? Is this correct?

Yes, that's correct! After you've added a commit to your branch, you can push the updated code to the same branch that is associated with this PR. Any new commits pushed to the branch will automatically appear in the PR (as you might already saw).

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
lunamorrow: Oops I missed this one, thank you for picking it up :)

lunamorrow:
Thanks @yuxuanzhuang I am working through your comments now and will aim to push the updated branch shortly!

Quick question about this comment:

> You need to make sure acceptors_ag, hydrogens_ag, etc. always exist.
> 
> ```python
> else:
>     self.acceptors_ag = self.u.select_atoms(self.acceptors_sel)
> ```

I need to avoid using `select_atoms` (using `guess_acceptors` instead), but could you please elaborate on what you mean? I have defined them in the initialiser as None, so that they are always there, but I cannot figure out how your code piece suggestion integrates. 

Also, do I need to worry about the "This Universe does not contain charge information" errors? I do not think they have anything to do with my changes? I am worried they could be "hiding" other possible errors further down that aren't being reached before this error is thrown though...

yuxuanzhuang:
Regarding depreciation, current provided code examples will break e.g.
```python
u = MDAnalysis.Universe(PSF, DCD)

hbonds = HBA(universe=u)
hbonds.hydrogens_sel = hbonds.guess_hydrogens("protein") # this will not be selection string.
hbonds.acceptors_sel = hbonds.guess_acceptors("protein")
hbonds.run()
```

Should we consider accepting `AtomGroup` for `hbonds.hydrogens_sel` and other inputs as well?

yuxuanzhuang:
> I need to avoid using select_atoms (using guess_acceptors instead), but could you please elaborate on what you mean? I have defined them in the initialiser as None, so that they are always there, but I cannot figure out how your code piece suggestion integrates.

The logic behind `guess_acceptors` and the provision of an implicit `acceptors_sel` IMO is that with the latter, you are certain of the atoms you want to use for calculating hydrogen bonds (so you only need to `select_atoms`. In `guess_acceptors`, the Universe needs to have enough information, such as charges and masses, to make guesses; however, it's not always the case.

lunamorrow:
> I also don't have experience handling depreciation before, so I might need help from @hmacdope :)

Of course, all good! If you can just help me get this PR sorted and ready to go I'll be so grateful! The depreciation can be sorted later :)



> > I need to avoid using select_atoms (using guess_acceptors instead), but could you please elaborate on what you mean? I have defined them in the initialiser as None, so that they are always there, but I cannot figure out how your code piece suggestion integrates.
> 
> The logic behind `guess_acceptors` and the provision of an implicit `acceptors_sel` IMO is that with the latter, you are certain of the atoms you want to use for calculating hydrogen bonds (so you only need to `select_atoms`. In `guess_acceptors`, the Universe needs to have enough information, such as charges and masses, to make guesses; however, it's not always the case.

Oh ok yes, I understand what you mean now! Yeah I have been a bit unsure on how to tackle this in the "best" way. The `guess_acceptors` works the same as it used to, but I could add a layer of function that will consider the user input `acceptors_sel` if it is not None, instead of guessing based on Atom properties? Then change this line in `_prepare` (see below; add an `else` block) and pass a user defined `self.acceptors_sel` (if it exists) to define the Atoms in `self.acceptors_ag`? I am thinking I can set a selection string as an optional arg for `guess_acceptors` (and repeat this process for `guess_hydrogens` and `guess_donors` too)? Does this could like a good option?

```
        if self.acceptors_sel is None:
            self.acceptors_ag = self.guess_acceptors()
            self.acceptors_sel = self._group_categories(self.acceptors_ag) #NOTE: do I need this?
```

lunamorrow:
> Regarding depreciation, current provided code examples will break e.g.
> 
> ```python
> u = MDAnalysis.Universe(PSF, DCD)
> 
> hbonds = HBA(universe=u)
> hbonds.hydrogens_sel = hbonds.guess_hydrogens("protein") # this will not be selection string.
> hbonds.acceptors_sel = hbonds.guess_acceptors("protein")
> hbonds.run()
> ```
> 
> Should we consider accepting `AtomGroup` for `hbonds.hydrogens_sel` and other inputs as well?

Yes I agree. I am thinking a good work-around would be for users to define `hbonds.hydrogens_sel`, `hbonds.donors_sel` or `hbonds.acceptors_sel` when they initialise the object? Otherwise they could manually manipulate it before calling run() and it should all flow well (once I make the adjustments to `guess_acceptors`, `guess_hydrogens` and `guess_donors` as per above). I would like to keep `hydrogens_sel` for selection strings and `hydrogens_ag` for the AtomGroups separated? I think this will provide the best functionality/robustness overall?

lunamorrow:
@yuxuanzhuang I have made some changes, as I proposed above, which I will push shortly. I have pretty much eliminated the errors/failures bar these two.

1.
`MDAnalysis.exceptions.NoDataError: This Universe does not contain charge information`

I suspect the charge one is due to your comment? Is there anything I can do to fix this?

> In guess_acceptors, the Universe needs to have enough information, such as charges and masses, to make guesses; however, it's not always the case.

2.
`AttributeError: AtomGroup has no attribute replace.`

I cannot find a call to `replace` in the tests throwing that error. Where can I find the root cause of this?

yuxuanzhuang:
I think there may be differences in the results when using `guess_donors` compared to `_get_dh_pairs`, because `guess_donors` includes a check for charge limitations. Note `guess_donors` is not used at all in the original implementation.

```python
import MDAnalysis
from MDAnalysis.analysis.hydrogenbonds.hbond_analysis import HydrogenBondAnalysis as HBA

u = MDAnalysis.Universe(PSF, DCD)

hbonds = HBA(universe=u)
hbonds.run()
print(hbonds.results.hbonds.shape)
> (10858, 6) # old results
> (4237, 6) # new results
hbonds._donors.n_atoms
> 375 # old results
> 74 # new results

```
will end up with a different result from the current implementation.

To facilitate testing, you can copy old hbond_analysis.py as hbond_analysis_old.py and `from MDAnalysis.analysis.hydrogenbonds.hbond_analysis_old import HydrogenBondAnalysis as HBA_old`.


I think this is currently not captured by the tests.

Apologies that I don't really have enough time during the Easter to make thorough suggestions on what should be done about it right now.

However, I would suggest to make sure you
1) when doing `select_atoms`, also add `updating=update_selections` to make sure it will be an `UpdatingAtomGroup`
2) make sure the things you parse around are not a string but an `AtomGroup`. `AttributeError: AtomGroup has no attribute replace.` this error is linked to putting an `AtomGroup` into `select_atoms`.
3) you can do selections with group syntax e.g. `donors_ag = hydrogens_ag.residues.atoms.select_atoms(
                f"({select}) and around {self.d_h_cutoff} group hydrogens_ag",
                hydrogens_ag=hydrogens_ag, updating=self.update_selections
                )`.

lunamorrow:
Thanks @yuxuanzhuang, and I completely understand. I appreciate you taking time out of your Easter weekend to give me a hand. Unfortunately I am really struggling with this and I think I might need to take it back to the start and re-implement changes again from scratch. 

I have spent a good few hours reading over and tweaking code (including adjusting for your suggested changes 1 and 2), and it's still not working nicely with the tests. I have a number of tests that are getting empty AtomGroups, when the selection provided should return something, which is causing a plethora of problems. 

I'm getting a bit worried about finalising this fix before GSoC applications close, but I will keep this PR updated with any more questions as I go.

hmacdope:
> I'm getting a bit worried about finalising this fix before GSoC applications close, but I will keep this PR updated with any more questions as I go.

Don't stress about the timeline, there is no requirement that your PR is merged before the application deadline, only that we have had a chance to work with you and go back and forth a bit.  :)

hmacdope:
> I also don't have experience handling depreciation before, so I might need help from @hmacdope :)

We can move this to a seperate PR when the time comes. :)

lunamorrow:
Ok great, that is such a relief to hear @hmacdope. I have been panicking a bit! I'll focus on my written application and keep chugging away at this fix while I wait to hear if I get accepted :) 

> We can move this to a seperate PR when the time comes. :)

Perfect, I'll touch base with you before starting that to get some more clear direction!

hmacdope:
> Yes I agree. I am thinking a good work-around would be for users to define `hbonds.hydrogens_sel`, `hbonds.donors_sel` or `hbonds.acceptors_sel` when they initialise the object?

Much preferred to setting them by hand.

yuxuanzhuang:
Side note while testing: slicing UpdatingAtomGroup will result in a static AtomGroup. -> need a workaround to return an `UpdatingAtomGroup` in e.g. `guess_donors` because `donors_ag` will be sliced by `donors_ag[donors_ag.charges < max_charge]`

```python
u.select_atoms('protein', updating=True).__class__
> MDAnalysis.core.groups.UpdatingAtomGroup
u.select_atoms('protein', updating=True)[2:].__class__
> MDAnalysis.core.groups.AtomGroup

**Comment by orbeckst** [comment]
If you write this as as selection instead of slicing, can you maintain the updating group
```python
donors_ag.select_atoms(f"prop charge < {max_charge}", updating=True)
```

lunamorrow:
> Side note while testing: slicing UpdatingAtomGroup will result in a static AtomGroup. -> need a workaround to return an `UpdatingAtomGroup` in e.g. `guess_donors` because `donors_ag` will be sliced by `donors_ag[donors_ag.charges < max_charge]`

> If you write this as as selection instead of slicing, can you maintain the updating group
> 
> ```python
> donors_ag.select_atoms(f"prop charge < {max_charge}", updating=True)
> ```

This is great to know, thanks @orbeckst and @yuxuanzhuang! I'll make that change to `guess_donors`, which should help with some of the testing issues. I just checked `guess_acceptors` and `guess_hydrogens`, and the same is happening with the AtomGroup selections in these methods, so I will workshop some solutions for them. I'll get these changes done and a few more, then I'll push a new version and share any testing errors/failures for it that I cannot solve.

Just one question: will the use of `select_atoms` in this solution be appropriate in the method, since it will still be returning/passing an AtomGroup?

lunamorrow:
Hi @yuxuanzhuang @orbeckst @hmacdope. I'm jumping back on this PR as I have just finished off a bunch of university assessment. A couple weeks away from it was really beneficial for me to gain a fresh perspective and make fixes/changes relatively quickly. I have just committed updated versions of `hbond_analysis` and `test_hydrogenbonds_analysis` and I am pretty confident that I have implemented the changes as we discussed and the tests are running nicely. The code still needs some "cleaning" as I have left comments and some random prints here and there for while I was making changes, and some of the docstrings need a bit of TLC too. If everything looks good I will clean these up and commit a fresh version to be merged with version 3. Thanks! 😊

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -700,14 +718,17 @@ def _prepare(self):
...
-            self.hydrogens_sel = self.guess_hydrogens()
+        # if self.acceptors_sel is None:
+        #     self.acceptors_sel = self.guess_acceptors()
+        # if self.hydrogens_sel is None:
+        #     self.hydrogens_sel = self.guess_hydrogens()
 
         # Select atom groups
-        self._acceptors = self.u.select_atoms(self.acceptors_sel,
-                                              updating=self.update_selections)
+        if self.acceptors_sel == None:
```
yuxuanzhuang:
```suggestion
        if self.acceptors_sel is None:
```

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -720,6 +741,8 @@ def _single_frame(self):
 
         # find D and A within cutoff distance of one another
         # min_cutoff = 1.0 as an atom cannot form a hydrogen bond with itself
+        print(f"D POS: {self._donors.positions}")
```
yuxuanzhuang: You can leave the debugging as it is but please put it with a logger. Check https://github.com/MDAnalysis/mdanalysis/blob/804b6073fdfaee0c5be1e2be659c2280ee4d0cda/package/MDAnalysis/analysis/align.py#L221
as an example

yuxuanzhuang (changes requested):
I'll need a bit more time to test on the new implementation, but I want to highlight an issue with the code examples in the current documentation. Currently, one can use:

```python
hbonds = HydrogenBondAnalysis(universe=u)
hbonds.hydrogens_sel = hbonds.guess_hydrogens("resname ARG HIS LYS")
hbonds.acceptors_sel = hbonds.guess_acceptors("resname ARG HIS LYS")
```

However, this will fail with the current implementation because `hbonds.acceptors_sel` requires a string, whereas `guess_acceptors` returns an AtomGroup.

While it's acceptable to have a breaking change in the API, it would be beneficial to ensure that most of the existing code remains functional.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
```diff
@@ -720,6 +741,8 @@ def _single_frame(self):
```
yuxuanzhuang: I find it somewhat inelegant that `_get_dh_pairs`, a frame-wise operation, performs atom selection to create new donors and hydrogens objects. Ideally, given that `updating=self.update_selections` applies to both hydrogens and donors, this should not be necessary.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
orbeckst: Actually, if it's purely debugging output that you used during development, remove it. Even if you include it at level DEBUG, it might slow down people who need to enable a debug logger for other pieces of code.

lunamorrow: > You can leave the debugging as it is but please put it with a logger. 

I am already planning to remove all these debugging and commented regions for the final code I commit, I just wanted to check my solution was appropriate before doing so (incase I still needed it).

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
lunamorrow: Got it! :)

lunamorrow:
Thanks @yuxuanzhuang and @orbeckst. 

> Currently, one can use:
> 
> ```python
> hbonds = HydrogenBondAnalysis(universe=u)
> hbonds.hydrogens_sel = hbonds.guess_hydrogens("resname ARG HIS LYS")
> hbonds.acceptors_sel = hbonds.guess_acceptors("resname ARG HIS LYS")
> ```
> 
> However, this will fail with the current implementation because `hbonds.acceptors_sel` requires a string, whereas `guess_acceptors` returns an AtomGroup.
> 
> While it's acceptable to have a breaking change in the API, it would be beneficial to ensure that most of the existing code remains functional.

I understand what you mean. Just to make sure we are on the same page, I believe once these changes are implemented that I should be doing a new PR per @hmacdope's guidance that will issue a depreciation warning for some of these features and how the Class is currently used. If I ensure that the API/docstring examples are updated appropriately will that be ok? I cannot see a clean way to implement this change without making all the "guess" methods return AtomGroups instead of selection string.

There are some other examples which will lose functionality. Like being able to guess hydrogens or acceptors from a certain group, like a protein (see below). Since my implementation only uses `guess_hydrogens`, `guess_acceptors` and  `guess_donors` if their respective selection string is None. Would it be suitable to make the resulting `self._hydrogens`, `self._acceptors` and `self._donors` mutable through a new method before `run` is called? Or I could implement a check for if any of `self._hydrogens`, `self._acceptors` and `self._donors` are empty AtomGroups, the respective `guess_hydrogens`, `guess_acceptors` or `guess_donors` could be called with the selection string passed in. Its a bit of a clunky fix though... Any suggestions on this would be really appreciated!

```python
  import MDAnalysis
  from MDAnalysis.analysis.hydrogenbonds.hbond_analysis import HydrogenBondAnalysis as HBA

  u = MDAnalysis.Universe(psf, trajectory)

  hbonds = HBA(universe=u)
  hbonds.hydrogens_sel = hbonds.guess_hydrogens("protein")
  hbonds.acceptors_sel = hbonds.guess_acceptors("protein")
  hbonds.run()
 ```

I have gone through the rest of your comments and change requests, and made those changes to the code. I will clean up the comments and docstrings now, then push a new version for your review. Thank you :)

**Comment by yuxuanzhuang** [comment]
Just a heads up, it's often necessary to rebase your code and resolve any conflicts with the latest develop branch. Here’s a helpful guide on how to do that: [Rebasing Your Code](https://userguide.mdanalysis.org/stable/contributing_code.html#rebasing-your-code).

Last time, I performed the rebase directly on GitHub, which might have caused some issues for you (apologies!) You could `git pull` the modification, if there's any on GitHub, before you push.

**Comment by lunamorrow** [comment]
Thanks @yuxuanzhuang. I realised I had forgotten to rebase before pushing the previous code, when I saw you had done the rebase for me. I made sure to do so for the most recent one. That's ok, I was a bit crazy on my end for the most recent push but I think I have it all sorted now! I'll just need to push a new branch to this PR somehow for further changes.

**Comment by lunamorrow** [comment]
Ok finally rebased and fixed up now. Everything should be sorted :)

File: package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py [hunk 6]
```diff
@@ -699,27 +701,18 @@ def _filter_atoms(self, donors, acceptors):
...
+                                                updating=self.update_selections)
         self._donors, self._hydrogens = self._get_dh_pairs()
 
     def _single_frame(self):
 
         box = self._ts.dimensions
 
-        # Update donor-hydrogen pairs if necessary
-        if self.update_selections:
-            self._donors, self._hydrogens = self._get_dh_pairs()
```
yuxuanzhuang, line 719: I think this is still needed for `_single_frame` because `self._donors` will be updated every frame with the change of `self._hydrogens`. You need to separate the part that creates `self._hydrogens` from it

yuxuanzhuang (changes requested):
See the inline comment and the code below that gets different results after the fix.

```python
import MDAnalysis
from MDAnalysis.analysis.hydrogenbonds.hbond_analysis import HydrogenBondAnalysis as HBA
from MDAnalysis.analysis.hydrogenbonds.hbond_analysis_old import HydrogenBondAnalysis as HBA_old

u = MDAnalysis.Universe(PSF, DCD)

hbonds_old = HBA_old(universe=u,
             hydrogens_sel='around 10 resid 5 and name H*',
             acceptors_sel='resid 5 and name O',
             update_selections=True)
hbonds_old.run()
print(f'old results: {hbonds_old.results.hbonds.shape}')

hbonds = HBA(universe=u,
             hydrogens_sel='around 10 resid 5 and name H*',
             acceptors_sel='resid 5 and name O',
             update_selections=True)
hbonds.run()
print(f'new results: {hbonds.results.hbonds.shape}')
> old results: (21, 6)
> new results: (6, 6)
```
Given all of your tests pass, it might be worth adding it to your test as well and making sure it also passes.

I am actually not a daily user of HBA (for a reason!) and I don't really know how people normally use it, so feel free to pitch in! @MDAnalysis/coredevs 

It's worth noting that it runs so so so much faster!!!

```python
hbonds = HBA(universe=u)
%timeit -r 2 -n 5 hbonds.run()
> old 2.85 s ± 8.15 ms per loop (mean ± std. dev. of 2 runs, 5 loops each)
> new 168 ms ± 6.62 ms per loop (mean ± std. dev. of 2 runs, 5 loops each)

lunamorrow:
Thanks for your quick feedback @yuxuanzhuang! :)

> See the inline comment and the code below that gets different results after the fix.

That is really weird as far as the results differing between the old and updated HBA. I will have a dig around on my end this week and fix it up.


> Given all of your tests pass, it might be worth adding it to your test as well and making sure it also passes.

Will do!


> It's worth noting that it runs so so so much faster!!!
> 
> ```python
> hbonds = HBA(universe=u)
> %timeit -r 2 -n 5 hbonds.run()
> > old 2.85 s ± 8.15 ms per loop (mean ± std. dev. of 2 runs, 5 loops each)
> > new 168 ms ± 6.62 ms per loop (mean ± std. dev. of 2 runs, 5 loops each)
> ```

Wow that is huge, and will have a really big impact for users performing HBA on large systems! That is so amazing to see the impact that this one change has had. 

I'll wait to hear back from the core devs as far as the loss of some functionality.

orbeckst:
You're asking for other @MDAnalysis/coredevs to jump in and comment "as far as the loss of some functionality.". Can you please summarize the changes so we don't have to read everything in the PR?
- What was changed?
- What is behaving differently now (API, results, ...)?

Thanks.

yuxuanzhuang:
I'm seeking insights on typical uses of HBA prior to and following the changes introduced by this PR. Below are some user scenarios categorized into two main aspects: 1) selections made for analysis and 2) information available in the Universe (omitted for now)

## Selection
1) Conduct HBA on the entire Universe (and perform post-analysis):
   ```
   # before and after
   hbonds = HBA(universe=u)
   ```

2) Select static parts of the system:
   ```
   # before
   hbonds = HBA(universe=u)
   hbonds.hydrogens_sel = hbonds.guess_hydrogens("protein")
   hbonds.acceptors_sel = hbonds.guess_acceptors("protein")
   ```
   ```
   # after
   # the previous code is no longer functional
   ???
   ```
**I think it's an important user case and needs workaround, e.g. allow `hydrogens_sel` and `acceptors_sel` accept both `string` and `AtomGroup`**

3) Dynamically select parts of the system:
   ```
   # before and after
   hbonds = HBA(universe=u,
                hydrogens_sel='around 10 resid 5 and name H*',
                acceptors_sel='resid 5 and name O',
                update_selections=True)
   ```
**[Currently, this results in divergent results](https://github.com/MDAnalysis/mdanalysis/pull/4533#pullrequestreview-2026367147) .**

4) Select between two parts of the system:
   ```
   # before and after
   hbonds = HBA(universe=u, between=['protein', 'water'])
   ```
sidenote: they are non-updating atom groups

## Information contained in the Universe (omitted for now)

1) Charges, masses, and bonds
2) Charges and masses
3) Bonds
4) No details on any of these

Let me know if there are additional scenarios or details that might be relevant.

package/MDAnalysis/analysis/hydrogenbonds/hbond_analysis.py
yuxuanzhuang: @lunamorrow The reason you get different results is that in the old code, `self._donors` is updated every frame based on `self._hydrogens`. However, if you don't execute `_get_dh_pairs` every frame, `self._donors` will become a non-updating AtomGroup based on the first frame.

IAlibay:
> I think it's an important user case and needs workaround, e.g. allow hydrogens_sel and acceptors_sel accept both string and AtomGroup


Just a very brief 2 cent here - the current functionality cannot change or stop working until the 3.0 release. The approach we have usually taken is to temporarily support both input types (and make sure they give the same answer).

lunamorrow:
> Just a very brief 2 cent here - the current functionality cannot change or stop working until the 3.0 release. The approach we have usually taken is to temporarily support both input types (and make sure they give the same answer).

I agree. The plan I made with @hmacdope is to roll these changes out with 3.0 and create a second PR after this one is done to issue a depreciation warning. Would the support of both types be to integrate now, and then remove this support with 3.0? I can make this happen if this is the route we want to take @orbeckst @yuxuanzhuang @IAlibay.


> 2. Select static parts of the system:
>    ```
>    # before
>    hbonds = HBA(universe=u)
>    hbonds.hydrogens_sel = hbonds.guess_hydrogens("protein")
>    hbonds.acceptors_sel = hbonds.guess_acceptors("protein")
>    ```
>    
>      
>    ```
>    # after
>    # the previous code is no longer functional
>    ???
>    ```

As @yuxuanzhuang said, the main issue is that `guess_hydrogens`, `guess_acceptors` and `guess_donors` are only called if their respective selection string is None, which means a selection like "protein" won't make it to them. This prevents Atoms for HBA being selected based on what structure they are a part of which is a pretty critical functionality. I will start on an elegant work around for this. 

@MDAnalysis/coredevs would you prefer this work around to include enabling `guess_hydrogens`, `guess_acceptors` and `guess_donors` to accept strings again? Or am I ok to create another method to enable less specific selection strings like "protein" to be passed to these guesser methods? I am worried that enabling `guess_hydrogens`, `guess_acceptors` and `guess_donors` to accept strings again will make any speed improvements I have made thus far null.

yuxuanzhuang:
> prefer this work around to include enabling guess_hydrogens, guess_acceptors and guess_donors to accept strings again?

I'm considering to allow `hydrogens_sel` and `acceptors_sel` can accept both strings and AtomGroups instead. This would maintain the functionality of commands like `hbonds.hydrogens_sel = hbonds.guess_hydrogens("protein")`. I don't even think you need to depreciate it after 3.0.0

lunamorrow:
> I'm considering to allow `hydrogens_sel` and `acceptors_sel` can accept both strings and AtomGroups instead. This would maintain the functionality of commands like `hbonds.hydrogens_sel = hbonds.guess_hydrogens("protein")`. I don't even think you need to depreciate it after 3.0.0

Ok easy, I can get started on that then. I have a crazy few days coming up, but I'll aim to push a revision with this functionality added over the weekend :)

richardjgowers:
just chiming in that accepting either AG or string as suggested here: https://github.com/MDAnalysis/mdanalysis/pull/4533#issuecomment-2087797633 is a good idea;  strings are a nice way to start, but if you've got a really odd system or want to do quite specific analysis then being able to just hand the analysis an AG to work on directly is heaps easier than finding the magic string that recreates this AG

orbeckst:
Hello everyone here. Are you still working on this PR @lunamorrow ?

PR #4535, opened 2024-03-27
Pathlib object handling for Universe,  SingleFrameReaderBase and Toplogy parsers (Issue #3937)

### LINKED ISSUE #3937: MAINT, ENH: more pathlib support tasks
Some suggested follow-ups from gh-3935:

1. Add support and testing for `pathlib` object handling to `SingleFrameReaderBase` and/or upstream this to the proto infrastructure. The `pathlib` handling/testing in the connected PR is restricted to `ReaderBase` for now.
2. The testing added in the above PR might eventually benefit from being consolidated to i.e., universe creation over a parametrized set of trajectory formats instead of the copy-paste sampling I did for GRO and DCD formats.

talagayev:
Partially Fixes #3937. The issue mentioned the addition of support and testing for `pathlib` objects for `SingleFrameReaderBase`. 

Currently the `SingleFrameReaderBase` is able to handle both `pathlib` and `str` as input for `SingleFrameReaderBase` to display this, this PR is focusing on tests that display the handling of `pathlib` and `str` as input for `SingleFrameReaderBase` . 

 - Addition of tests for `pathlib` object and `str` input for `SingleFrameReaderBase` in `test_gro.py` and `test_lammps.py` 

Currently the tests are as mentioned for `GRO` and `LAMMPS` cases. `SingleFrameReaderBase` also recognizes `INPCRD`, `CRD`, `NAMDBIN` and `DMS` if given as a single input, so tests for these cases could also be added if required.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

----
📚 Documentation preview 📚: https://mdanalysis--4535.org.readthedocs.build/en/4535/


orbeckst:
@hmacdope would you be able to look at this PR or assign to someone else, please?

hmacdope:
Apologies for the delay @orbeckst @talagayev, was away over easter. Reviewing now.

hmacdope (changes requested):
@talagayev you have the right idea of what to test, and your test implementations are good! 

However to achieve better coverage and to make your tests more powerful you can instead add your tests to `BaseReaderTest` and `_SingleFrameReader` that cover the base API functionality for **all** relevant trajectory types.  This is in `estsuite/MDAnalysisTests/coordinates/base.py`.  :)

talagayev:
@hmacdope 

> @talagayev you have the right idea of what to test, and your test implementations are good!
> 
> However to achieve better coverage and to make your tests more powerful you can instead add your tests to `BaseReaderTest` and `_SingleFrameReader` that cover the base API functionality for **all** relevant trajectory types. This is in `estsuite/MDAnalysisTests/coordinates/base.py`. :)

Ah, I think i see :) i would than adjust it so that the tests are in `BaseReaderTest` and `_SingleFrameReader`.

Does it than make also sense to move the `ReaderBase` tests from #3935 also in the same way in a separate PR? since currently from the tests they cover two cases, one in the `testsuite/MDAnalysisTests/coordinates/test_dcd.py` for PSF&DCD and `testsuite/MDAnalysisTests/coordinates/test_xdr.py` for GRO&XTC ?

talagayev:
@hmacdope
I managed to get the test locally in `_SingleFrameReader` and it covers the cases, but trying to add it into `BaseReaderTest` leads to errors such as:

`AttributeError: 'PosixPath' object has no attribute 'encode'` 

in the case of `test_dcd.py`. Here it has problems with the `TestDCDReader` with also having the same errors with the `TRRReader` and `XTCReader` ,  whereas it is fine in `test_gro.py` with the `TestGROReader`.  Is it possible that `TestDCDReader` also does access `BaseReaderTest`, which leads to the error, since it possibly needs a different case of handling than `TestGROReader`. Does it make sense to leave the test for the `SingleFrameReaderBase` in `_SingleFrameReader` and make one for `MultiframeReaderTest` so that both cases are divided and tested separatly in their classes?

talagayev:
@hmacdope 
Hey Hugo, I adjusted now the tests and moved them to `base.py` as you suggested. As I mentioned there were the problems with the cases when i applied the tests for the single frame pathway to formats such as `DCD`, `XTC` so the ones that are not `SingleFrameReader` formats and thus I had to set exceptions for these cases.
Could you take a look if everything is fine now? For the checks most of them failed on the `codecov` stage in the end, so i am not sure if it is an error from the codecov site. Thanks in advance :)

hmacdope:
@talagayev sorry for the delay here. I will review ASAP.

talagayev:
@hmacdope all good, no worries :)

hmacdope:
@talagayev I have pushed some changes to your branch that fix the `encode` issue, so we shouldn't need to exclude some readers,  and also added support for `Pathlib.path` in `Universe` and the topology parsers.  I have adjusted CHANGELOG and PR title to match also. 

As we are now co-authors on this I can no longer fairly review, so I will defer to one of the other @MDAnalysis/coredevs. Perhaps @tylerjereddy as the author of previous Pathlib support  issues and PRs #3937 (if you have some spare cycles).

hmacdope:
Ok looks like I messed some stuff up, Ill fix.

talagayev:
@hmacdope  Hm strange error, I try to find the error going through the pytests and running the developer version of MDAnalysis with this version and can't track down the error currently.

package/MDAnalysis/coordinates/base.py
```diff
@@ -1663,7 +1663,11 @@ class SingleFrameReaderBase(ProtoReader):
     def __init__(self, filename, convert_units=True, n_atoms=None, **kwargs):
         super(SingleFrameReaderBase, self).__init__()
 
-        self.filename = filename
+        if isinstance(filename, NamedStream):
+            self.filename = filename
+        else:
+            self.filename = str(filename)
```
talagayev: Here errors if we use `self.filename = str(filename)` in the `test_mmtf.py`
The lead to errors similar to this one:

`FileNotFoundError: [Errno 2] No such file or directory: '<mmtf.api.mmtf_reader.MMTFDecoder object at 0x135416780>'`

I managed to fix the error by adding:
`else:`
`    self.filename = str(filename)`
`    if "MMTF" in self.filename:`
`        self.filename = filename`

basically putting a specific condition for MMTF cases, not the cleanest approach, but maybe this helps @hmacdope fixing the error in a better way :)

testsuite/MDAnalysisTests/coordinates/base.py
```diff
@@ -521,6 +529,16 @@ def test_timeseries_atomgroup_asel_mutex(self, reader):
         atoms = mda.Universe(reader.filename).select_atoms("index 1")
         with pytest.raises(ValueError, match="Cannot provide both"):
             timeseries = reader.timeseries(atomgroup=atoms, asel=atoms, order='fac')
+          
+    def test_pathlib_input_base(self, reader):     
+        if isinstance(reader, MemoryReader):
+            if isinstance(reader, MemoryReader):
+                skip_reason = "MemoryReader"
+            pytest.skip(f"Skipping test for Pathlib input with reason: {skip_reason}")
+        path = Path(reader.filename)
```
talagayev, line 685: Here adding an additional Line:
`path = path.as_posix())`

Fixes the following Failed Pytest in 4 cases:
`AttributeError: 'PosixPath' object has no attribute 'encode'`

package/MDAnalysis/coordinates/base.py
talagayev: I tried to look more into it now, the `"MMTF"` is a specific case, the other errors appear mainly because of the parmed parser which has both problems with the `base.py` in the `coordinates` and `topology`. It is possible to use in the `base.py` in the `coordinates`:

`elif hasattr(filename, 'topology'):`
`    self.filename = filename`

This helps to fix the errors, although it is quite similar as to just directly using `self.filename = filename` without any `if/elif/else` conditions. For the `base.py` in `topology` I will try to figure out if there is any option to fix it, since again having there only `self.filename = filename` also fixes the errors with the `parmed` parser

richardjgowers: @talagayev I think the `FileNotFoundError` for mmtf above is because instead of a filename, sometimes a MMTFDecoder object is encountered.  I think what you've got is fixing it, but it might be clearer to have a block like:

```python
if isinstance(filename, NamedStream):
    self.filename = filename
elif 'mmtf' in str(filename.__class__):
    # mmtf case, where we've avoided a mmtf import in detection
    self.filename = filename
```

Where you've enumerated all the special cases explicitly in each `elif` branch, rather than hiding some special cases inside other branches

package/CHANGELOG
```diff
@@ -39,6 +40,7 @@ Fixes
  * Fix groups.py doctests using sphinx directives (Issue #3925, PR #4374)
 
 Enhancements
+ * Handling of Pathlib.Path in SingleFrameReaderBase, Topology and Universe (Issue #3937)
```
richardjgowers: This is a bit vague and doesn't actually tell me what is fixed or what is now possible.  e.g. Is it that `pathlib.Path`s are now correctly handled?

package/MDAnalysis/coordinates/base.py
talagayev: Ah I see, yes the option that I got worked, but was definitely not the best option, but yes was connected with mmtf and yes  `elif 'mmtf' in str(filename.__class__):` looks much better :)

package/CHANGELOG
talagayev: Correct, originally it was only targeting pathlib object handling of SingleFrameReaderBase in specific cases, but with the adjustment of @hmacdope it would cover `Pathlib.Path`s for all cases if the errros that appear currently due to the changes in the `base.py` in topology and also the `base.py` in coordinates, which could be fixed with 

`elif hasattr(filename, 'topology'):`
 `    self.filename = filename`
 
 but this would just exclude the cases of `topology` so I am not sure if this would fix it and cover then all the  `Pathlib.Path` cases

orbeckst:
@hmacdope would it be worthwhile reviewing again? Can this PR be pushed across the finish line?

package/MDAnalysis/coordinates/base.py
orbeckst: MMTF is no longer relevant, #4639 . I think you can ignore/omit anything related to it.

talagayev:
@hmacdope I would ping you here for the PR. I tried to adjust the code to work with the `pathlib` here. For the testing I made general tests in `base.py` in both `topology` and `coordinates`, but there were some edge cases with `IMD` and `ITP` and in some cases specify the `Path` to the files differently then in the `base.py` an example being `test_itp.py`

Could you go through and look if that approach works and the edge case covering is reasonable?

PR #4548, opened 2024-03-30
Convert "Quantity" to raw value for kinetic_energy and potential_energy in OpenMM reader

orionarcher:
 - added `.value` to kinetic_energy and potential_energy

This is breaking the ability of a downstream MDAkit to write H5MD files. [Issue here](https://github.com/sef43/openmm-mdanalysis-reporter/issues/5). I suspect it is more generally breaking the ability to take in OpenMM and output H5MD, but it is only in `openmm-reporter` that I came across it.

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

----
📚 Documentation preview 📚: https://mdanalysis--4548.org.readthedocs.build/en/4548/

### DIFF
--- a/package/MDAnalysis/converters/OpenMM.py
+++ b/package/MDAnalysis/converters/OpenMM.py
@@ -125,10 +125,10 @@ class OpenMMSimulationReader(base.SingleFrameReaderBase):
         ts.frame = 0
         ts.data["time"] = state.getTime()._value
         ts.data["potential_energy"] = (
-            state.getPotentialEnergy().in_units_of(u.kilojoule/u.mole)
+            state.getPotentialEnergy().in_units_of(u.kilojoule/u.mole)._value
         )
         ts.data["kinetic_energy"] = (
-            state.getKineticEnergy().in_units_of(u.kilojoule/u.mole)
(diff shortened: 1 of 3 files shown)


package/MDAnalysis/converters/OpenMM.py
```diff
@@ -125,10 +125,10 @@ def _mda_timestep_from_omm_context(self):
         ts.frame = 0
         ts.data["time"] = state.getTime()._value
         ts.data["potential_energy"] = (
-            state.getPotentialEnergy().in_units_of(u.kilojoule/u.mole)
+            state.getPotentialEnergy().in_units_of(u.kilojoule/u.mole)._value
```
IAlibay, line 128: Would it be possible to add a quick test for these?

IAlibay (changes requested):
Thanks, just the one thing please.

package/MDAnalysis/converters/OpenMM.py
orionarcher: Done

IAlibay (approved):
lgtm!

PR #4584, opened 2024-05-01
Limit threads usage in numpy during test to reduce test time

yuxuanzhuang:

 - make tests lighter

See the comparison between duration below; it takes a lot more time in test_encore for numpy calculations before.

new test duration
```bash
============================= slowest 50 durations =============================
11.82s call     MDAnalysisTests/analysis/test_contacts.py::TestContacts::test_n_initial_contacts[datafiles1-41814]
10.93s call     MDAnalysisTests/coordinates/test_gro.py::TestGROLargeWriter::test_writer_large
10.70s call     MDAnalysisTests/coordinates/test_pdb.py::test_conect_bonds_all
9.34s call     MDAnalysisTests/parallelism/test_multiprocessing.py::test_multiprocess_COG[u11]
9.02s call     MDAnalysisTests/utils/test_pickleio.py::test_fileio_pickle
8.53s call     MDAnalysisTests/analysis/test_gnm.py::test_closeContactGNMAnalysis
8.31s call     MDAnalysisTests/analysis/test_gnm.py::test_closeContactGNMAnalysis_weights_None
8.27s call     MDAnalysisTests/analysis/test_encore.py::TestEncore::test_ces_error_estimation_ensemble_bootstrap
7.62s call     MDAnalysisTests/coordinates/test_gro.py::TestGROLargeWriter::test_write_trajectory_universe
7.52s call     MDAnalysisTests/coordinates/test_gro.py::TestGROLargeWriter::test_write_trajectory_atomgroup
```

old test duration (https://github.com/MDAnalysis/mdanalysis/actions/runs/8869719078/job/24350703650)
```
20.06s call     MDAnalysisTests/analysis/test_encore.py::TestEncore::test_hes_to_self
18.93s call     MDAnalysisTests/analysis/test_encore.py::TestEncore::test_hes
13.07s call     MDAnalysisTests/analysis/test_encore.py::TestEncore::test_ces_error_estimation_ensemble_bootstrap
12.93s call     MDAnalysisTests/analysis/test_encore.py::TestEncore::test_hes_custom_weights
12.80s call     MDAnalysisTests/coordinates/test_pdb.py::test_conect_bonds_all
12.12s call     MDAnalysisTests/analysis/test_contacts.py::TestContacts::test_n_initial_contacts[datafiles1-41814]
11.40s call     MDAnalysisTests/analysis/test_gnm.py::test_closeContactGNMAnalysis
10.79s call     MDAnalysisTests/coordinates/test_gro.py::TestGROLargeWriter::test_writer_large
```

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

----
📚 Documentation preview 📚: https://mdanalysis--4584.org.readthedocs.build/en/4584/


yuxuanzhuang:
@orbeckst @IAlibay I now believe that these three adjustments can reduce the chance of timeouts:

1) Limiting the number of threads used by numpy during testing. This also decreases the duration of tests that are heavily dependent on numpy and computationally intensive, e.g. in encore.

2) Reducing the number of workers in pytest to prevent overloading the entire test runner.

3) Lightening the tests. I have already modified some, but we can certainly further optimize the other time-consuming tests.

I have re-run the GitHub CI three times consecutively, and they all pass. I have also patched it to #4162 that are more likely to time-out in https://github.com/yuxuanzhuang/mdanalysis/pull/6 and it also seems to work well.

.github/workflows/gh-ci.yaml
```diff
@@ -108,6 +108,13 @@ jobs:
     - name: run_tests
       if: contains(matrix.name, 'asv_check') != true
       run: |
+        export OPENBLAS_NUM_THREADS=1
+        export GOTO_NUM_THREADS=1
+        export OMP_NUM_THREADS=1
+        export MKL_NUM_THREADS=1
+        # limit to 2 workers to avoid overloading the CI
+        export PYTEST_XDIST_AUTO_NUM_WORKERS=2
```
orbeckst: Could this be set to *auto*, because that's what we are already using. Does *auto* not work correctly?

If *auto* does not work then I'd prefer we define a variable GITHUB_CI_MAXCORES or similar and then use it invoking `pytest -n $GITHUB_CI_MAXCORES`. I prefer commandline arguments over env vars when determining code behavior because you immediately see what affects the command itself.

azure-pipelines.yml
```diff
@@ -128,6 +128,13 @@ jobs:
     displayName: 'Check installed packages'
   - powershell: |
       cd testsuite
+      $env:OPENBLAS_NUM_THREADS=1
+      $env:GOTO_NUM_THREADS=1
+      $env:OMP_NUM_THREADS=1
+      $env:MKL_NUM_THREADS=1
+      # limit to 2 workers to avoid overloading the CI
+      $env:PYTEST_XDIST_AUTO_NUM_WORKERS=1
```
orbeckst: The comment is confusing: if you set workers to 1 but say you limit to 2 workers. Change comment, otherwise same as above: perhaps just add to commandline args?

azure-pipelines.yml
```diff
@@ -128,6 +128,13 @@ jobs:
...
+
       pytest MDAnalysisTests --disable-pytest-warnings -n auto --timeout=200 -rsx --cov=MDAnalysis
```
orbeckst: Here we already have `-n auto` – how does this interact with `PYTEST_XDIST_AUTO_NUM_WORKERS`?

testsuite/MDAnalysisTests/analysis/test_encore.py
```diff
@@ -142,12 +142,12 @@ def test_rmsd_matrix_with_superimposition(self, ens1):
         conf_dist_matrix = encore.confdistmatrix.conformational_distance_matrix(
             ens1,
             encore.confdistmatrix.set_rmsd_matrix_elements,
-            select="name CA",
+            select="name CA and resnum 1:3",
```
orbeckst: The final output is insensitive to just using resnum 1:3 as opposed to using the full protein??!?!? What are we even testing here? Are the encore tests so insensitive/random?

testsuite/MDAnalysisTests/analysis/test_encore.py
```diff
@@ -125,10 +125,10 @@ def test_triangular_matrix(self):
                        reason="Not yet supported on Windows.")
     def test_parallel_calculation(self):
 
-        arguments = [tuple([i]) for i in np.arange(0,100)]
+        arguments = [tuple([i]) for i in np.arange(0,10)]
 
         parallel_calculation = encore.utils.ParallelCalculation(function=function,
-                                                                n_jobs=4,
+                                                                n_jobs=2,
```
orbeckst: good catch!!

orbeckst: I assume that this change alone (together with restricting threads) accounts for a lot of improvement.

orbeckst (approved):
Thank you for looking into this pesky issues. If your changes improve our CI then I am all in favor.

I have one comment on how to set the number of used processes for pytest-xdist (see comments) but I am not going to block over it.

I am also shocked at how insensitive the ENCORE tests are: you changed them dramatically and they still pass without changing the reference values. Maybe a thing to raise on the [mdaencore repo](https://github.com/MDAnalysis/mdaencore)....

orbeckst:
@IAlibay would you please shepherd the PR to completion?

.github/workflows/gh-ci.yaml
```diff
@@ -108,6 +108,13 @@ jobs:
     - name: run_tests
       if: contains(matrix.name, 'asv_check') != true
       run: |
+        export OPENBLAS_NUM_THREADS=1
```
IAlibay: todo: there's another place to put this so it applies everywhere

IAlibay (changes requested):
Thanks for doing this @yuxuanzhuang 

I'm unfortunately pretty swamped this weekend so I can only very briefly look at this today but I'll try to get back to you asap with a longer review and better explanation of where I think those env exports should go so they apply globally.

.github/workflows/gh-ci.yaml
yuxuanzhuang: I was planning to find a way to find the default number of workers pytest will use and reduce it by one because the Ubuntu runner has 4 cores and Mac has 3 cores. but I ended up setting it to 2 and found the performance acceptable.

azure-pipelines.yml
yuxuanzhuang: pytest will read `PYTEST_XDIST_AUTO_NUM_WORKERS` when `-n auto`. It's equivalent to `pytest -n $PYTEST_XDIST_AUTO_NUM_WORKERS`

testsuite/MDAnalysisTests/analysis/test_encore.py
yuxuanzhuang: These tests here are not comparing to a reference array but rmsd calculated with `rms.RMSD` (see l150).

For the tests that have a set of reference results, it's indeed different if I use resnum 1:3

azure-pipelines.yml
yuxuanzhuang: I forgot to correct the comment line when I realized Azure has only 2 cores so I changed # workers to 1.

hmacdope:
@IAlibay with my PR shepherding hat on, just pinging here, no rush though :smile:

hmacdope:
@IAlibay just pinging here with my review coordination hat on.

.github/workflows/gh-ci.yaml
orbeckst: ok

IAlibay: > Ubuntu runner has 4 cores and Mac has 3 cores.

It's unclear from this conversation that altering the value of `-n` is actually resulting in any difference in the issue we're seeing with the the one multiprocessing test failing (i.e. the only cause of timeouts), could you confirm this please?

If it's not affecting things, then my suggestion is to stick with auto unless there's a substantial difference in performance that we've missed.

azure-pipelines.yml
IAlibay: As above, is this actually affecting timeouts or is this some separate optimization?

IAlibay (changes requested):
Based on the PR main comment and the follow-up comment, it's unclear to me how this PR specifically fixes #4209

The timeout issues are specifically related to the multiprocessing test going over the time limit, whilst the optimizations to the encore tests are nice, could you please detail what is being done here that specifically changes the behaviour of the multiprocessing test?

yuxuanzhuang:
Since this PR doesn’t really **resolve** the time-out issue, I will repurpose it to focus on reducing the test duration.

orbeckst:
@yuxuanzhuang do you want to get this PR back into play?

PR #4604, opened 2024-05-22
Remove default parallelizable parameter in NoJump transformation

### LINKED ISSUE #4259: Raise an error is `parallelizable=True` is passed to the `NoJump` transformation
## Expected behavior ##

There is a comment in the `NoJump` transformation dunder init:
https://github.com/MDAnalysis/mdanalysis/blob/8923c4a8203e755939ad9fc3f2242a837a493c28/package/MDAnalysis/transformations/nojump.py#L106-L108

I would expect that this should be enforced and I should not be able to set `parallelizable=True`

## Actual behavior ##

However it's not enforced:
https://github.com/MDAnalysis/mdanalysis/blob/8923c4a8203e755939ad9fc3f2242a837a493c28/package/MDAnalysis/transformations/nojump.py#L110

and I don't think it's mentioned in the docs. I think it would be good to raise an error if a user sets `parallelizable=True`.

A more general solution would be to modify `TransformationBase` to remove the `parallelizable` argument. A `ParallelizableTransformationBase` class could then inherit from `TransformationBase`. This would allow whoever is writing the transformation to decide whether it is parallelizable, and the user calling the transformation could decide whether they want to parallelize it if possible.

``` python
import MDAnalysis as mda

u = mda.Universe(PSF, DCD)
u.trajectory.add_transformations(mda.transformations.NoJump(parallelizable=True))  # doesn't raise an error
```

- Which operating system?

marinegor:
Fixes #4259

 - remove `parallelizable` argument of `NoJump` transformation to avoid accidentally passing `parallelizable=True`

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

----
📚 Documentation preview 📚: https://mdanalysis--4604.org.readthedocs.build/en/4604/

### DIFF
--- a/package/MDAnalysis/transformations/nojump.py
+++ b/package/MDAnalysis/transformations/nojump.py
@@ -100,7 +100,9 @@ class NoJump(TransformationBase):
         check_continuity=True,
         max_threads=None,
     ):
-        super().__init__(max_threads=max_threads, parallelizable=False)
+       # NoJump transforms are inherently unparallelizable, since
+       # it depends on the previous frame's unwrapped coordinates
+       super().__init__(max_threads=max_threads, parallelizable=False)
         self.prev = None
         self.old_frame = 0
         self.older_frame = "A"
--- a/package/MDAnalysis/transformations/nojump.py
+++ b/package/MDAnalysis/transformations/nojump.py
@@ -99,11 +99,8 @@ class NoJump(TransformationBase):
         self,
         check_continuity=True,
         max_threads=None,
-        # NoJump transforms are inherently unparallelizable, since
-        # it depends on the previous frame's unwrapped coordinates
-        parallelizable=False,
     ):
-        super().__init__(max_threads=max_threads, parallelizable=parallelizable)
+        super().__init__(max_threads=max_threads, parallelizable=False)
         self.prev = None
         self.old_frame = 0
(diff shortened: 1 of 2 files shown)


package/MDAnalysis/transformations/nojump.py
```diff
@@ -99,11 +99,8 @@ def __init__(
         self,
         check_continuity=True,
         max_threads=None,
-        # NoJump transforms are inherently unparallelizable, since
-        # it depends on the previous frame's unwrapped coordinates
-        parallelizable=False,
     ):
-        super().__init__(max_threads=max_threads, parallelizable=parallelizable)
+        super().__init__(max_threads=max_threads, parallelizable=False)
```
orbeckst, line 105: Keep the comment — this is valuable information!!
```suggestion
       # NoJump transforms are inherently unparallelizable, since
       # it depends on the previous frame's unwrapped coordinates
       super().__init__(max_threads=max_threads, parallelizable=False)
```

orbeckst (approved):
Looks like a good change to me. For once I don't think we need a test here.

I will add my own code suggestion to keep the comment around.

hmacdope (approved):
LGTM

orbeckst:
oops... 

Will merge when tests are ✅

PR #4612, opened 2024-06-05
Cache creation of compound masks

PicoCentauri:
# Changes made in this Pull Request:

 Add a cache for creating the compound masks in a group. This speeds up the calculations of `accumulate` other methods that use `_split_by_compound_indices`:

```python
u = mda.Universe(TPR, XTC)
```

Running a simple timing test with caching gives on my machine `397 µs ± 1.55 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)`

```python
%%timeit 
u.atoms.accumulate("masses", compound="residues")
```

and with "disabled" cache `1.16 ms ± 11 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)`

```python
%%timeit 
u.atoms.accumulate("masses", compound="residues")
u.atoms._cache.pop("residues_masks")
```

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

----
📚 Documentation preview 📚: https://mdanalysis--4612.org.readthedocs.build/en/4612/

### DIFF
--- a/package/MDAnalysis/core/groups.py
+++ b/package/MDAnalysis/core/groups.py
@@ -936,45 +936,56 @@ class GroupBase(_MutableBase):
         n_compounds : int
             The number of individual compounds.
         """
-        # Caching would help here, especially when repeating the operation
-        # over different frames, since these masks are coordinate-independent.
-        # However, cache must be invalidated whenever new compound indices are
-        # modified, which is not yet implemented.
-        # Also, should we include here the grouping for 'group', which is
+        # Should we include here the grouping for 'group', which is
         # essentially a non-split?
 
+        cache_key = f"{compound}_masks"
         compound_indices = self._get_compound_indices(compound)
-        compound_sizes = np.bincount(compound_indices)
-        size_per_atom = compound_sizes[compound_indices]
-        compound_sizes = compound_sizes[compound_sizes != 0]
-        unique_compound_sizes = unique_int_1d(compound_sizes)
 
-        # Are we already sorted? argsorting and fancy-indexing can be expensive
-        # so we do a quick pre-check.
-        needs_sorting = np.any(np.diff(compound_indices) < 0)
-        if needs_sorting:
-            # stable sort ensures reproducibility, especially concerning who
-            # gets to be a compound's atom[0] and be a reference for unwrap.
-            if stable_sort:
-                sort_indices = np.argsort(compound_indices, kind='stable')
-            else:
-                # Quicksort
-                sort_indices = np.argsort(compound_indices)
-            # We must sort size_per_atom accordingly (Issue #3352).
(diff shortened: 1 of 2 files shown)


package/MDAnalysis/core/groups.py
```diff
@@ -936,45 +936,56 @@ def _split_by_compound_indices(self, compound, stable_sort=False):
...
+                                      .reshape(-1, compound_size))
+                else:
+                    atom_masks.append(np.where(size_per_atom 
+                                               == compound_size)[0]
+                                      .reshape(-1, compound_size))
+
+            self._cache[cache_key] = {
+                "compound_indices": compound_indices,
+                "data": (atom_masks, compound_masks, len(compound_sizes))
+            }
```
PicoCentauri, line 986: I am not sure if this is the canonical approach for caching. But the `@cached` decorator does not work here I think.

PicoCentauri:
This is related to #3169 but does different caching as proposed in #3169.

orbeckst:
@richardjgowers could you please have a look at this PR, looks like your wheelhouse.

If you don't have the bandwidth please un-assign yourself and let me know. Thanks!

orbeckst (dismissed):
The *TestUnwrapFlag.test_residues* test fails. I assume that this is actually due to the changes here. What needs to be done to make the tests pass again?

richardjgowers:
this failure here:
```
Mismatched elements: 2 / 9 (22.2%)
  Max absolute difference: 55.42303448
  Max relative difference: 26.25439738
   x: array([[21.286, 41.664, 40.465],
         [44.528, 43.426, 23.248],
         [57.534, 27.871, 53.767]])
   y: array([[21.286, 41.664, 40.465],
         [44.528, 43.426, 78.671],
         [ 2.111, 27.871, 53.767]], dtype=float32)
```
the x coordinate is off by 55.423 where the boxlength on that dimension is 55.423, so that just looks like a rounding error.  Those two x coordinates are virtually identical, and this test doesn't seem to be trying to establish which image gets picked, it's just bad luck that this regression test could fall into either.  (arguably there could be some sort of periodic assert coordinates equal function for things like this...)

The moment of inertia error is a little more tricky to pin down in a similar way, but I'd be suspicious that it is also a rounding error

I got these tests to fail locally, and then they started passing locally, so it looks a bit twitchy...

richardjgowers (commented):
barring the PBC precision fails, something I noticed is that for some reason `group.center_of_mass()` doesn't seem to create the cache entry, which is confusing as the code should be going through this new cache path.   `.accumulate()` does seem to be creating the cache entry.

PicoCentauri:
Thanks for looking into this @richardjgowers. I can also try to debug why the `.center_of_mass()` doens't seem to create the cache entry.

Issue #4598, opened 2024-05-17
`H5MD` Can not read `observables/<group>/<observable>/...`


PythonFZ:
## Expected behavior ##

The H5MD standard defines two main ways to define [observables](https://www.nongnu.org/h5md/h5md.html#observables-group). Either directly as group below the root `observables` or as a group specific observable `observables/group`.

```
observables
 \-- <observable1>
 |    \-- step: Integer[variable]
 |    \-- time: Float[variable]
 |    \-- value: <type>[variable]
 \-- <group1>
 |    \-- <observable2>
 |         \-- step: Integer[variable]
 |         \-- time: Float[variable]
 |         \-- value: <type>[variable][D][D]
```

Both should be supported.

## Actual behavior ##

In my experience only the observables after root are supported but `observables/group` are not. It leads to `KeyError: "Unable to synchronously open object (object 'value' doesn't exist)"`. This seems to be briefly discussed in https://github.com/MDAnalysis/mdanalysis/issues/4320 but I wanted to give more information here.

## Code to reproduce the behavior ##
I've attached a file [licl_100.h5.zip](https://github.com/MDAnalysis/mdanalysis/files/15345623/licl_100.h5.zip) because the `cobrotoxin.h5md` test file only contains `observables/lambda` but no group specific observables.

``` python
import MDAnalysis as mda

u = mda.Universe.empty(100, n_residues=100, atom_resindex=np.arange(100), trajectory=True)
reader = H5MDReader("licl_100.h5", convert_units=False)
u.trajectory = reader
....

```

## Current version of MDAnalysis ##

- Which version are you using? (run `python -c "import MDAnalysis as mda; print(mda.__version__)"`) 2.7.0
- Which version of Python (`python -V`)? 3.10.9 
- Which operating system? Ubuntu 22

orbeckst:
@edisj @ljwoods2 is this issue a known limitation in our H5MD implementation?

edisj:
yes, this is related to point 1) in #4561 

[...]

PR #4615, opened 2024-06-15
Skip observables contained in particle groups

PythonFZ:
This is not a fix for https://github.com/MDAnalysis/mdanalysis/issues/4598 but enables reading files that contain data in

```
observables
 \-- <group1>
 |    \-- <observable2>
 |         \-- step: Integer[variable]
 |         \-- time: Float[variable]
 |         \-- value: <type>[variable][D][D]
```

 - 

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?

----
📚 Documentation preview 📚: https://mdanalysis--4615.org.readthedocs.build/en/4615/

This file hasn't been touched in 9 months and I don't think there are some open PRs so there would be a chance to format the entire file using `ruff` which I accidently did here https://github.com/MDAnalysis/mdanalysis/commit/faafe8b671bebd954c62a1027afe7dfaae7e9bd3 - is this of interest?


orbeckst:
@edisj and @ljwoods2 could you have a quick look at this PR and comment?

orbeckst (changes requested):
I like the idea of using the logger but logging at every step is potentially a lot of output and performance hit. Furthermore, I don’t think users care. Thus I suggest you remove the logging and just `pass`.

We also need at least one test case.

orbeckst:
@hmacdope would you be able to shepherd this PR? If not switch it to me.

testsuite/MDAnalysisTests/coordinates/test_h5md.py
```diff
@@ -894,3 +895,20 @@ def test_writer_no_h5py(self, Writer, outfile):
...
                     W.write(universe)
+
+@pytest.mark.skipif(not HAS_H5PY, reason="h5py not installed")
+class TestH5MDReaderWithObservables(object):
+    """Read H5MD file with 'observables/atoms/energy'."""
+
+    prec = 3
+    ext = 'h5md'
+
+    @pytest.fixture(scope='class')
```
ljwoods2: I may be mistaken, but I think the `test_n_frames` method isn't testing the issue- if the reader was going to except out due to a key error, this would happen during the line in the `universe` fixture:
```
reader = H5MDReader(H5MD_energy, convert_units=True)
```
when the H5MDReader's `__init__` is called and the first call to `_read_next_timestep` is made. The test method doesn't call any code that reads a frame, it just ends up calling the reader's `n_frames` method. I agree with your approach to fixing the issue, just maybe have the test initialize or initialize and iterate through the reader so that the potential issue will occur in the scope of the test

ljwoods2: Also make sure to update the CHANGELOG!

PythonFZ: Should be fixed now. The CI failing seems not related to this PR?

ljwoods2: Small nitpick- can you change the name of the test to make it represent what it is testing? Also, the assert statement seems unrelated to what the test is testing. Maybe write an except and `pytest.fail` statement upon raising a KeyError but otherwise don't have an assert statement. Here's an example from `testsuite/coordinates/base.py`:
```
    def test_frame_jump_issue1942(self, ref, reader):
        """Test for issue 1942 (especially XDR on macOS)"""
        reader.rewind()
        try:
            for ii in range(ref.n_frames + 2):
                reader[0]
        except StopIteration:
            pytest.fail("Frame-seeking wrongly iterated (#1942)")
```
Otherwise LGTM

PythonFZ: I updated the test and used `pytest.fail`. I hope the test structure now is a bit cleaner. Thanks for the review.

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -231,6 +232,7 @@ class MockH5pyFile:
     HAS_H5PY = True
 
 
+
```
hmacdope: undo blank line change.

testsuite/MDAnalysisTests/coordinates/test_h5md.py
```diff
@@ -894,3 +895,21 @@ def test_writer_no_h5py(self, Writer, outfile):
...
+
+    def test_read_h5md_issue4598(self):
+        """Read a H5MD file with observables.
+        
+        The reader will ignore the 'observables/atoms/energy'.
+        """
+        try:
+            _ = H5MDReader(H5MD_energy, convert_units=True)
+        except KeyError:
+            pytest.fail("Could not read H5MD file with observables.")
```
hmacdope: You can just have

 `  _ = H5MDReader(H5MD_energy, convert_units=True)
`
As the body of the test, no need to `try, except`. Pytest can catch all the errors. 

See also comment about warnings, we should catch suggested warning.

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -673,10 +675,13 @@ def _read_frame(self, frame):
     def _copy_to_data(self):
         """assigns values to keys in data dictionary"""
 
-        if 'observables' in self._file:
-            for key in self._file['observables'].keys():
-                self.ts.data[key] = self._file['observables'][key][
-                    'value'][self._frame]
+        if "observables" in self._file:
```
hmacdope, line 677: We should issue a warning that observables of the form `observables/group1/observable1` are being ignored if present.

hmacdope (changes requested):
Great work! This is a good start until we decide what to do with the other form of observables.

hmacdope (commented):
FWIW @PythonFZ the following will successfully read into `ts.data` with the key form `group/observable`

```python
    def _copy_to_data(self):
        """assigns values to keys in data dictionary"""

        if "observables" in self._file:
            for key in self._file["observables"].keys():
                try:
                    # if has value as subkey read directly into data
                    if "value" in self._file["observables"][key]:
                        self.ts.data[key] = self._file["observables"][key][
                            "value"
                        ][self._frame]
                    # if value is not a subkey, read dict of subkeys
                    else:
                        for subkey in self._file["observables"][key].keys():
                            self.ts.data[key + "/" + subkey] = self._file["observables"][key][
                                subkey
                            ]["value"][self._frame]
                except KeyError:
                    warnings.warn(f"Could not read {key} from observables group, not a legal H5MD observable specification or data not ")

```

PythonFZ:
@hmacdope What is the expected way to auto-format the code? I'm used to running either `black` or `ruff format` but once I run that on the code base it will yield to changes in most of the files. I haven't found anything at https://userguide.mdanalysis.org/stable/contributing.html?

Maybe there are some `pre-commit` hooks I haven't found?

orbeckst:
Only format lines that you touch, otherwise the diff gets too noisy. There’s the darker linter/formatter action that tells you if you should fix any of your lines. Am 6/18/24 um 04:50 schrieb Fabian Zills ***@***.***>:﻿
@hmacdope What is the expected way to auto-format the code? I'm used to running either black or ruff format but once I run that on the code base it will yield to changes in most of the files. I haven't found anything at https://userguide.mdanalysis.org/stable/contributing.html?
Maybe there are some pre-commit hooks I haven't found?

testsuite/MDAnalysisTests/datafiles.py
```diff
@@ -372,6 +373,7 @@
 TPR_xvf = (_data_ref / 'cobrotoxin.tpr').as_posix()
 TRR_xvf = (_data_ref / 'cobrotoxin.trr').as_posix()
 H5MD_xvf = (_data_ref / 'cobrotoxin.h5md').as_posix()
+H5MD_energy = (_data_ref / 'cu.h5').as_posix()
```
edisj: is it the convention for other `h5md` files out there to use the `*.h5` extension rather than `*.h5md`?

edisj:
thanks @PythonFZ, looks good to me

testsuite/MDAnalysisTests/datafiles.py
PythonFZ: I can rename the file to `*.h5md`. I've seen both and personally prefer `*.h5` because I was lazy and did not configure the VS Code H5WEB extension to also read `*.h5md`. Afaik H5MD is the only format specification for H5 files, so there is no risk of confusion.

Edit: I haven't seen a recommendation for the `*.h5md` suffix in the H5MD definition so I would actually prefer keeping the more general `*.h5`.

hmacdope:
@PythonFZ lets go with `.h5md` as I think the missing extension may be causing some test failures. Other than that this is very close.

PythonFZ:
> @PythonFZ lets go with `.h5md` as I think the missing extension may be causing some test failures. Other than that this is very close.

I renamed the files and ensured everything is properly formatted. I hope everything is good to go?

hmacdope:
There seems to be some failing test on the `azure` runner. Can your reproduce locally @PythonFZ ?

PythonFZ:
> There seems to be some failing test on the `azure` runner. Can your reproduce locally @PythonFZ ?

All tests are passing for me locally. From what I see in the logs, it is the following test (correct me if I am wrong)
```
/home/vsts/work/1/s/testsuite/MDAnalysisTests/parallelism/test_multiprocessing.py
```

All tests from that file run locally for me. Also, these tests should not have been affected by this PR and the tests I changed also all run locally. The same seems to be true for the GH hosted runners. I'm a bit lost here on how to fix this, because I don't know the differences between the Azure runner and the GH runners.

orbeckst:
This failing test is flaky and is not related to your work. It will not block the PR. OliverAm 6/28/24 um 07:20 schrieb Fabian Zills ***@***.***>:﻿

There seems to be some failing test on the azure runner. Can your reproduce locally @PythonFZ ?

All tests are passing for me locally. From what I see in the logs, it is the following test (correct me if I am wrong)
/home/vsts/work/1/s/testsuite/MDAnalysisTests/parallelism/test_multiprocessing.py

All tests from that file run locally for me. Also, these tests should not have been affected by this PR and the tests I changed also all run locally. The same seems to be true for the GH hosted runners. I'm a bit lost here on how to fix this, because I don't know the differences between the Azure runner and the GH runners.

hmacdope:
Let me kick CI IIRC there were some failing tests on `mdanalysis-CI (Azure_Tests Linux-Python310-64bit-full-wheel)`

hmacdope (approved):
Congrats on your first contribution @PythonFZ! Much appreciated. Will be useful with some exciting H5MD developments around the corner (CC @ljwoods2).

hmacdope:
@orbeckst are you happy with changes here?

package/MDAnalysis/coordinates/H5MD.py
```diff
@@ -673,10 +674,25 @@ def _read_frame(self, frame):
...
+                            "value"
+                        ][self._frame]
+                    # if value is not a subkey, read dict of subkeys
+                    else:
+                        for subkey in self._file["observables"][key].keys():
+                            self.ts.data[key + "/" + subkey] = self._file[
+                                "observables"
+                            ][key][subkey]["value"][self._frame]
+                except KeyError:
+                    warnings.warn(
```
orbeckst: If it's not legal, shouldn't we be *failing*? Or is the reasoning that we don't overly care about observables as opposed to x/v/f?

Perhaps add a comment explaining your rationale.

Also, given that this is not tested, add a `#pragma: nocover` (or whatever you have to do to show that this code block isn't covered). It will tell anyone else reading the code that it's un-tested. 

Tested would be better, of course, but if we don't care about observables then I don't insist on a test.

package/MDAnalysis/coordinates/H5MD.py
orbeckst: Add this limitation to the documentation.

EDIT: "This" means @hmacdope 's comment that some obervables are not being read.

orbeckst (changes requested):
Looking good overall, just two docs/comments requests.

orbeckst:
@PythonFZ if you can have a quick look at my comments then we should be able to merge this PR soon.

package/MDAnalysis/coordinates/H5MD.py
PythonFZ: We should be reading all "valid" observables with the current implementation.

package/MDAnalysis/coordinates/H5MD.py
PythonFZ: I'd be in favor of raising an Error here as well. Shall I create an additional `h5md` file which contains wrong data to test against this?

orbeckst: That would be ideal!

PythonFZ:
@orbeckst I've added a file that tests for `AttributeError`. I can add another file for testing against `KeyError` but this is an unlikely case. The code will account for this though. Let me know if you prefer the additional file or are good with this.

I'm not sure how to account for
> Add this limitation to the documentation.

If you are referring to
> We should issue a warning that observables of the form observables/group1/observable1 are being ignored if present.

This should not longer be the case and all valid, non-fixed time step, H5MD datasets should work.

orbeckst (approved):

orbeckst:
lgtm, thanks for adding the test @PythonFZ !

orbeckst:
Thank you for your contribution @PythonFZ ! 🎉

PR #4618, opened 2024-06-17
Add a Dependabot config to auto-update GitHub action versions

kurtmckee:
Currently, deprecation warnings are being thrown due to actions using out-of-date Node versions ([recent example](https://github.com/MDAnalysis/mdanalysis/actions/runs/9544980203)). Rather than submitting a PR to update them manually, this PR configures Dependabot to regularly submit PRs to update GitHub action versions as needed.

If this PR merges, you can expect Dependabot to immediately submit a PR to update actions to newer versions.

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

----
📚 Documentation preview 📚: https://mdanalysis--4618.org.readthedocs.build/en/4618/

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    groups:
+      github-actions:
+        patterns:
+          - "*"
--- a/package/AUTHORS
+++ b/package/AUTHORS
@@ -240,6 +240,7 @@ Chronological list of authors
   - Sampurna Mukherjee
   - Leon Wehrhan
   - Valerij Talagayev
+  - Kurt McKee
 
 
 


.github/dependabot.yaml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
```
RMeli, line 4:
```suggestion
    directory: ".github/workflows"
```
Shouldn't this be limited to the `.github/workflows` sub-directory?

.github/dependabot.yaml
RMeli, line 1: It might be too noisy so it's up to discussion, but I think having the field `assignees` set to `MDAnalysis/coredevs` might be valuable.

RMeli (changes requested):
Thank you, this looks very useful.

.github/dependabot.yaml
kurtmckee: No, the directory is correct.

.github/dependabot.yaml
kurtmckee: I recommend introducing [a CODEOWNERS file](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) so that PR assignees are managed in one location, for all PRs.

IAlibay: We don't generally work via a codeowners style structure, but rather a defined roles structures like this: https://www.mdanalysis.org/pages/team/#roles

My 2 cent is that it's better to have staged triage based on individuals' roles.

kurtmckee: Understood re: team roles.

I can add an `assignees` field but want to confirm that @RMeli and @IAlibay, you're in agreement I should do that.

RMeli: I totally agree with @IAlibay here, in general it is not ideal if everyone gets assigned to every PR. The code base is complex and diverse, and different people often have expertise in different parts.

This is why I was suggesting to assign only dependabots PRs, which should be reasonably easy for everyone to deal with.

kurtmckee: I checked the documentation and it appears that the `assignees` field is for individual contributors, not groups.

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#assignees

However, the `reviewers` field accepts teams. Is that acceptable?

IAlibay: I think we're overthinking this - if a dependabot PR comes through, _someone_ will look at it either way, we don't need to ping folks imho.

RMeli: E-mail/GitHub notifications got out of hand for me, so having something assigned is much better IMO.

But if we can't tag a group, then there is no point. (And reviewers have a different meaning.)

RMeli:
LGTM. Please add yourself to the authors file if not there already.

.github/dependabot.yaml
IAlibay: I generally just don't rely on email notifications (even have them off), I just loop through my github inbox directly - that tends to be quite stable.

RMeli (approved):

PR #4620, opened 2024-06-18
BLD: NumPy 2 compat for wheel builds

### LINKED ISSUE #4619: CI fails when numpy 2.0.0 is installed
## Expected behavior ##

CI passes with correct code.

## Actual behavior ##

Azure CI runners such as  https://dev.azure.com/mdanalysis/mdanalysis/_build/results?buildId=7066&view=logs&jobId=86224841-1158-5011-fbb7-7c2644c2081b&j=86224841-1158-5011-fbb7-7c2644c2081b&t=44ce4b91-ecdc-5aca-e807-60d46710cf6d  in https://github.com/MDAnalysis/mdanalysis/pull/4162#issuecomment-2175553868 fail with
```
=================================== ERRORS ====================================
__________________ ERROR collecting MDAnalysisTests/analysis __________________
C:\hostedtoolcache\windows\Python\3.10.11\x64\lib\importlib\__init__.py:126: in import_module
    from .lib import log
C:\hostedtoolcache\windows\Python\3.10.11\x64\lib\site-packages\MDAnalysis\lib\__init__.py:34: in <module>
    from . import transformations
C:\hostedtoolcache\windows\Python\3.10.11\x64\lib\site-packages\MDAnalysis\lib\transformations.py:172: in <module>
    from .mdamath import angle as vecangle
C:\hostedtoolcache\windows\Python\3.10.11\x64\lib\site-packages\MDAnalysis\lib\mdamath.py:63: in <module>
    from . import util
C:\hostedtoolcache\windows\Python\3.10.11\x64\lib\site-packages\MDAnalysis\lib\util.py:228: in <module>
    from ._cutil import unique_int_1d
MDAnalysis\\lib\\_cutil.pyx:1: in init MDAnalysis.lib._cutil
    ???
E   ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject

```

The action installed [numpy 2.0.0](https://dev.azure.com/mdanalysis/mdanalysis/_build/results?buildId=7066&view=logs&j=86224841-1158-5011-fbb7-7c2644c2081b&t=44ce4b91-ecdc-5aca-e807-60d46710cf6d&l=47) via pip.

tylerjereddy:
* Related to #4619

* We should build against NumPy `2.0.0` so that we're backwards compatible to NumPy `1.x` while still supporting 2.x series. This is simpler to maintain as well. In theory, should support as far back as NumPy `1.19` if we needed it.

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?

----
📚 Documentation preview 📚: https://mdanalysis--4620.org.readthedocs.build/en/4620/

### DIFF
--- a/package/MDAnalysis/converters/ParmEd.py
+++ b/package/MDAnalysis/converters/ParmEd.py
@@ -81,6 +81,8 @@ import functools
 import itertools
 import warnings
 
+from numpy.lib import NumpyVersion
+
 from . import base
 from ..coordinates.base import SingleFrameReaderBase
 from ..topology.tables import SYMB2Z
@@ -168,11 +170,20 @@ class ParmEdConverter(base.ConverterBase):
         obj : AtomGroup or Universe or :class:`Timestep`
         """
         try:
-            import parmed as pmd
+            # TODO: remove this guard when parmed has a release
+            # that supports NumPy 2
+            if NumpyVersion(np.__version__) < "2.0.0":
+                import parmed as pmd
+            else:
+                raise ImportError
         except ImportError:
-            raise ImportError('ParmEd is required for ParmEdConverter but '
-                              'is not installed. Try installing it with \n'
-                              'pip install parmed')
+            if NumpyVersion(np.__version__) >= "2.0.0":
+                ermsg = "ParmEd is not compatible with NumPy 2.0+"
+            else:
+                ermsg = ("ParmEd is required for ParmEdConverter but is not "
+                         "installed. Try installing it with \n"
+                         "pip install parmed")
+            raise ImportError(errmsg)
         try:
             # make sure to use atoms (Issue 46)
             ag_or_ts = obj.atoms
--- a/package/MDAnalysis/converters/ParmEd.py
+++ b/package/MDAnalysis/converters/ParmEd.py
@@ -81,6 +81,7 @@ import functools
 import itertools
 import warnings
 
+import numpy as np
 from numpy.lib import NumpyVersion
 
 from . import base
--- a/package/pyproject.toml
+++ b/package/pyproject.toml
@@ -3,15 +3,13 @@
 requires = [
     "Cython>=0.28",
     "packaging",
-    # lowest NumPy we can use for a given Python,
-    # In part adapted from: https://github.com/scipy/oldest-supported-numpy/blob/main/setup.cfg
-    # As per NEP29, we set the minimum version to 1.23.2 for Python <=3.11
-    # and 1.26.0 (first to support) for Python 3.12
-    "numpy==1.23.2; python_version<='3.11' and platform_python_implementation != 'PyPy'",
-    "numpy==1.26.0; python_version=='3.12' and platform_python_implementation != 'PyPy'",
-    # For unreleased versions of Python there is currently no known supported
-    # NumPy version. In that case we just let it be a bare NumPy install
-    "numpy<2.0; python_version>='3.13'",
+    # numpy requirement for wheel builds for distribution on PyPI - building
+    # against 2.x yields wheels that are also compatible with numpy 1.x at
+    # runtime.
+    # Note that building against numpy 1.x works fine too - users and
+    # redistributors can do this by installing the numpy version they like and
+    # disabling build isolation.
+    "numpy>=2.0.0",
     # Set to minimum version of setuptools that allows pyproject.toml
     "setuptools >= 40.9.0",
     "wheel",
--- a/package/CHANGELOG
+++ b/package/CHANGELOG
@@ -56,6 +56,8 @@ Enhancements
    DOI 10.1021/acs.jpcb.7b11988. (Issue #2039, PR #4524)
 
 Changes
+ * MDAnalysis now builds against numpy 2.0 rather than the
+   minimum supported numpy version (PR #4620)
  * As per SPEC0 the minimum supported Python version has been raised
    to 3.10 (PR #4502)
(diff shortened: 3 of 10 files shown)


tylerjereddy:
some of our Windows deps may also not have NumPy 2 compatible releases yet, need to check on that

package/pyproject.toml
```diff
@@ -3,15 +3,13 @@
...
-    # In part adapted from: https://github.com/scipy/oldest-supported-numpy/blob/main/setup.cfg
-    # As per NEP29, we set the minimum version to 1.23.2 for Python <=3.11
-    # and 1.26.0 (first to support) for Python 3.12
-    "numpy==1.23.2; python_version<='3.11' and platform_python_implementation != 'PyPy'",
-    "numpy==1.26.0; python_version=='3.12' and platform_python_implementation != 'PyPy'",
-    # For unreleased versions of Python there is currently no known supported
-    # NumPy version. In that case we just let it be a bare NumPy install
-    "numpy<2.0; python_version>='3.13'",
+    # numpy requirement for wheel builds for distribution on PyPI - building
+    # against 2.x yields wheels that are also compatible with numpy 1.x at
```
IAlibay, line 7: Is the range of compatible 1.x numpy versions with 2.0 published somewhere? I.e. checking if we need to bump the minimum NumPy version anywhere.

IAlibay (commented):
Thanks @tylerjereddy - just the one question.

I have a suspicion some of our gh actions CI won't be picking this at all since we do a lot of isolation disabling here and there, but if the azure pipelines CI works then it should be good.

package/pyproject.toml
tylerjereddy: This document isn't too bad: https://numpy.org/doc/stable/dev/depending_on_numpy.html#for-downstream-package-authors. I think we may be "ok" on that front.

You can deduce the stability back to `1.25.x` and then to `1.19.x` based on that document. It looks like we have Python `3.10` and NumPy `1.23.2` as our runtime lower bounds, which is pretty close to where SciPy is (`3.10` and `1.23.5` respectively) at the moment.

tylerjereddy: Let me do a quick check of the current CI failures to see if I can make a concrete suggestion for those as well.

tylerjereddy:
For `pytng` errors, I think the cross-listed PR is roughly what we'd want to do there. Even if there's no bandwidth for a release for `pytng` (let alone MDA) for a bit, we could perhaps point to the GitHub `master` branch of `pytng` for installs in our CI here (i.e., point `pip` at the GitHub repo) to keep that testing active rather than disabling `pytng` testing here.

There may be a few other test failures here to check, but the `pytng` ones stood out first.

I've adjusted the original comment to not close the matching issue, since this may be a step in the right direction, but we may need to check other deps/things a bit still.

tylerjereddy:
I restarted the failed Azure jobs to see if the new `pytng` release helps the situation a bit.

tylerjereddy:
Looking at the logs, in terms of our dependencies, I'd say `parmed` seems to have moderate issues with NumPy 2 (they need to adjust their array copy semantics a bit like we did). `RDKit` seems to have a more serious problem, resulting in hard crashes against NumPy 2. Both projects probably need a new release to support.

IAlibay:
@tylerjereddy sorry I'm going to word this terribly - these are optional dependencies, is there a path you can see here where we stick to 1.x if you need these optional dependencies and 2+ if you don't? I guess in we'd effectively need two packages with different pins?

My view is that this doesn't seem like it would work, but I've put all of five minutes of thought into it.

tylerjereddy:
@IAlibay I think the best I can suggest is to skip the related tests/imports for `parmed` and `rdkit` under NumPy 2 until they have compatible releases available.

I pushed in a sample commit that seems to allow the testsuite to pass locally with NumPy 2. Looking at the results in CI here, the only test failure is `test_creating_multiple_universe_without_offset`, which I believe is a known issue/flake.

That doesn't mean the MDA team will like this, but I pushed it in so you can take a look. Maybe not too painful?

IAlibay:
> @IAlibay I think the best I can suggest is to skip the related tests/imports for `parmed` and `rdkit` under NumPy 2 until they have compatible releases available.
> 
> I pushed in a sample commit that seems to allow the testsuite to pass locally with NumPy 2. Looking at the results in CI here, the only test failure is `test_creating_multiple_universe_without_offset`, which I believe is a known issue/flake.
> 
> That doesn't mean the MDA team will like this, but I pushed it in so you can take a look. Maybe not too painful?

O wow thanks a bunch for digging into this @tylerjereddy !
I don't mind this solution, it's not "clean", but it's the "cleanest" I think we can do here - rdkit will update eventually, parmed might not and there's nothing we can do about that.

package/MDAnalysis/converters/ParmEd.py
```diff
@@ -168,11 +170,20 @@ def convert(self, obj):
         obj : AtomGroup or Universe or :class:`Timestep`
         """
         try:
-            import parmed as pmd
+            # TODO: remove this guard when parmed has a release
+            # that supports NumPy 2
+            if NumpyVersion(np.__version__) < "2.0.0":
```
IAlibay, line 176: I just added this extra guard here since we were doing the same for RDKit.

IAlibay (approved):
Thanks so much @tylerjereddy ! I hope you don't mind I took the liberty of pushing a few commits directly to make sure we can get this merged in asap.

IAlibay:
Thanks @tylerjereddy - going with the merge!

PR #4656, opened 2024-08-15
Fixed high dimensional GroupBase indexing.

### LINKED ISSUE #4647: GroupBase high dimensional indexing.
GroupBase does not perform enough error checking on supplied indices in the `__getitem__` method.
This allows e.g. AtomGroups to be indexed with invalid inputs which can lead to a hidden error which can cause issues further in the workflow.
Specifically higher dimensional arrays (or lists) are valid inputs currently.
This can also lead to odd behaviour when oddly shaped arrays are provided.

## Expected behavior ##
A user might expect something like: `result = [ag[ind] for ind in 2d_indices]` for this kind of higher dimensional indexing. 
The proper behaviour should probably be raising an error since the resulting shaped outputs cause lots of problems when used further i.e. in varous analysis classes.

Additionally, this can cause an issue on a valid, but oddly shaped input.

This behaviour could easily be avoided by firstly applying `np.squeeze(item)` onto the item array which would resolve issues with oddly shaped indexing arrays.
Then higher dimensional, invalid inputs could then be captured by checking that `np.squeeze(item).ndim == 1`

I'd be happy to try and implement this at some point next week if people agree this is an issue.

## Actual behavior ##
Currently no attempt to correct the input is made and no error is raised, an AtomGroup is created which will cause issues further in the program. It is not always immediate obvious where this error originated.
The heirarchy of the produced AtomGroup is not meaningful.

```
import MDAnalysis as mda

print(mda.__version__)
u = mda.Universe(PSF, DCD)
indices = [[0,1,2],[3,4,5]]

# Returns 1 AtomGroup, with 2 "Atoms", containing 3 indexes etc.
# Raise no warning/error
ag_2d = u.atoms[indices]


from MDAnalysis.analysis.rdf import InterRDF
# Still no error
rdf = InterRDF(ag_2d, u.atoms)

# Isn't until capped distance tries to run that the error is caught.
rdf.run()
```

additionally in a valid but oddly shaped array case:

```
import MDAnalysis as mda
import numpy as np

print(mda.__version__)
u = mda.Universe(PSF, DCD)

ag1 = u.atoms[np.arange(5)]
ag2 = u.atoms[np.arange(5).reshape(-1,1)]
ag3 = u.atoms[np.arange(5).reshape(1,-1)]

assert ag1.n_atoms == 5  # passes
assert ag2.n_atoms == 5  # passes
assert ag3.n_atoms == 5 # Fails. ag3.n_atoms == 1.
```

- Which version are you using? 2.8.0-dev0
- Which version of Python (`python -V`)? Python 3.12.4
- Which operating system? Reproducible on Linux and Windows

MattTDavies:
Fixes #4647 
This does not fix the underlying "Frankenatoms" issue, but stops it from occuring due to GroupBase indexing by  raising an error if improperly shaped indices are provided.

 - Added test confirming 2d index raises an IndexError.
 - Added a dimensionality check to GroupBase.\_\_getitem__ to raise an error on invalid higher dimensional inputs.

PR Checklist
------------
 - [x] Tests?
 - [ ] Docs?
 - [x] CHANGELOG updated?

----
📚 Documentation preview 📚: https://mdanalysis--4656.org.readthedocs.build/en/4656/

### DIFF
--- a/package/MDAnalysis/core/groups.py
+++ b/package/MDAnalysis/core/groups.py
@@ -597,6 +597,16 @@ class GroupBase(_MutableBase):
                 # hack to make lists into numpy arrays
                 # important for boolean slicing
                 item = np.array(item)
+            
+            # correctly flatten oddly shaped, but 1d arrays
+            # raises issues if flattening already suitable arrays
+            if isinstance(item, np.ndarray) and item.ndim > 1: # ignore slices
+                item = np.squeeze(item)
+                # disallow high dimensional indexing.
+                # this doesnt stop the underlying issue
+                if item.ndim > 1:
+                    raise IndexError('Group index must be 1d')
+
             # We specify _derived_class instead of self.__class__ to allow
             # subclasses, such as UpdatingAtomGroup, to control the class
             # resulting from slicing.
--- a/package/MDAnalysis/core/groups.py
+++ b/package/MDAnalysis/core/groups.py
@@ -597,10 +597,10 @@ class GroupBase(_MutableBase):
                 # hack to make lists into numpy arrays
                 # important for boolean slicing
                 item = np.array(item)
-            
+
             # correctly flatten oddly shaped, but 1d arrays
             # raises issues if flattening already suitable arrays
-            if isinstance(item, np.ndarray) and item.ndim > 1: # ignore slices
+            if isinstance(item, np.ndarray) and item.ndim > 1:
                 item = np.squeeze(item)
                 # disallow high dimensional indexing.
                 # this doesnt stop the underlying issue
--- a/package/MDAnalysis/core/groups.py
+++ b/package/MDAnalysis/core/groups.py
@@ -598,14 +598,10 @@ class GroupBase(_MutableBase):
                 # important for boolean slicing
                 item = np.array(item)
 
-            # correctly flatten oddly shaped, but 1d arrays
-            # raises issues if flattening already suitable arrays
             if isinstance(item, np.ndarray) and item.ndim > 1:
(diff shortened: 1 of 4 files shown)


package/MDAnalysis/core/groups.py
```diff
@@ -597,6 +597,16 @@ def __getitem__(self, item):
...
+
+            # correctly flatten oddly shaped, but 1d arrays
+            # raises issues if flattening already suitable arrays
+            if isinstance(item, np.ndarray) and item.ndim > 1:
+                item = np.squeeze(item)
+                # disallow high dimensional indexing.
+                # this doesnt stop the underlying issue
+                if item.ndim > 1:
+                    raise IndexError('Group index must be 1d')
+
```
richardjgowers: I'm not sure (?) we have to allow ndim>1, so we can probably just raise an Error inside this branch without the np.squeeze step.

Is there a code path that currently uses the `np.arange(5).reshape(-1, 1)` route?

richardjgowers (commented):
Looks good but I'm not sure we need the sanitization step

package/MDAnalysis/core/groups.py
MattTDavies: Not that I can tell. The only tests that fail with the sanitisation removed are the newly added tests, to test the sanitisation is working.

richardjgowers: Ok great then, I couldn't remember if something like Residue access was going via this route.

richardjgowers (approved):

PR #4694, opened 2024-08-30
IMDv3 Integration

ljwoods2:
Draft PR for integrating the IMD (Interactive Molecular Dynamics) new v3 protocol into MDAnalysis. 

The IMDReader hooks into the [IMDClient](https://github.com/ljwoods2/imdclient) API, which is comprised of 4 methods:

```
__init__(): Connect to the simulation engine and send it a go signal to start sending IMD data

get_imdsessioninfo(): Returns an `IMDSessionInfo` object which contains info on whether xvf/energies/dimensions are sent in the stream

get_imdframe(): Returns an `IMDFrame` or raises EOFError if the IMD stream is over

stop(): Cleanup method that ensures the `IMDClient` disconnects from the simulation server
```

There are a few different alternatives for how the IMDReader is integrated into MDAnalysis.

## IMDReader in main codebase, IMDClient in a separate (non-MDAKit) repo

The first rule of MDAKits is don't talk about MDAKits. The second first rule is that an MDAKit [has to actually use MDAnalysis](https://mdakits.mdanalysis.org/makingakit.html), so if the IMDReader is in this codebase, IMDClient can't be an MDAKit.

The advantage of splitting up the classes like this is that:

1. Users can use IMDv3 in MDAnalysis without installing additional packages
2. IMDClient remains decoupled from MDAnalysis, meaning other groups can use the client to parse IMDv3 streams
3. The IMDClient is free to do time-consuming and client-specific tasks like:
    - Test the IMDClient's ability to automatically pause a running simulation when its buffer fills, which is a complex multithreaded testcase that looks nothing like anything else in MDAnalysis. [Here's some example tests](https://github.com/ljwoods2/imdclient/blob/main/imdclient/tests/test_imdclient.py)
    - Build simulation engines that implement IMDv3 in the CI and test them against the client. These tests will potentially take several hours but add a lot of value. For example, LAMMPS offers several compiler options that impact IMD which are currently untested in the LAMMPS codebase. These options could be tested in the CI for different releases and on the develop branch. Example LAMMPS integration test class: [base class](https://github.com/ljwoods2/imdclient/blob/main/imdclient/tests/base.py), [lammps class](https://github.com/ljwoods2/imdclient/blob/main/imdclient/tests/test_lammps.py)

This is the solution I'm most excited about and that this PR illustrates, but I can see that having the IMDReader and IMDClient in separate repos could create an integration challenge if the IMDClient needs to change significantly in ways that impact the API.

## IMDReader & IMDClient in an MDAKit

This option still allows the client to be distributed separately from MDAnalysis and still allows more complex CI usage to stay out of the main codebase, but makes it so that users of MDA will have to install an additional package to use the IMDReader

## IMDReader & IMDClient in MDAnalysis main codebase

I feel strongly that this is not a good idea. The IMDClient tests would add complexity and runtime to an already complex MDA CI/CD pipeline, add a (very annoying, multithreaded) maintenance burden to core MDA developers, makes IMDClient unusable by other groups, and makes it more difficult to make rapid future changes to the IMDClient (like potentially replacing the implementation of socket interaction codes with Cython while keeping the API the same).

PR Checklist
------------
 - [ ] Tests?
 - [ ] Docs?
 - [ ] CHANGELOG updated?

----
📚 Documentation preview 📚: https://mdanalysis--4694.org.readthedocs.build/en/4694/

### DIFF
--- a/package/MDAnalysis/coordinates/IMD.py
+++ b/package/MDAnalysis/coordinates/IMD.py
@@ -1,0 +1,289 @@
+"""
+
+Example: Streaming an IMD v3 trajectory
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+To stream a trajectory from GROMACS or another simulation engine that supports 
+IMD v3, ensure that the simulation engine is running and waiting for an IMD connection.
+
+For example, in GROMACS, you can use ``gmx mdrun`` with the ``-imdwait`` flag
+to ensure that GROMACS will wait for a client before starting the simulation.
+In GROMACS, you will know that the simulation is ready and waiting for the
+MDAnalysis IMDReader client when this line is printed to the terminal:
+
+.. code-block:: none
+
+    IMD: Will wait until I have a connection and IMD_GO orders.
+
+Once the simulation is ready for a client connection, setup your :class:`Universe`
+like this: ::
+
+    import MDAnalysis as mda
+    # Pass host and port of the listening GROMACS simulation
+    # server as the trajectory argument
+    u = mda.Universe("topology.tpr", "localhost:8888")
+
+Classes
+^^^^^^^
+
+.. autoclass:: IMDReader
+   :members:
+   :inherited-members:
+
+"""
+
+import queue
+from MDAnalysis.coordinates.base import (
+    ReaderBase,
+    FrameIteratorIndices,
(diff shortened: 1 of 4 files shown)


ljwoods2:
Closing since it was decided that IMDReader & IMDClient will exist in an MDAKit

PR #4717, opened 2024-09-24
Add black configuration ignoring all files

RMeli:
Prepare work for addressing #2450, should the developers agree to move forward with adopting `black` as code formatter and incrementally format the whole code base.

This is a companion PR to https://github.com/MDAnalysis/UserGuide/pull/383, which explains the developer workflow for using `black`.

At the beginning, `black package` and `black testsuite` will do nothing (all files are ignored in this PR):

```
No Python files are present to be formatted. Nothing to do 😴
```

Files will be formatted and included incrementally.

----
📚 Documentation preview 📚: https://mdanalysis--4717.org.readthedocs.build/en/4717/

### DIFF
--- a/package/pyproject.toml
+++ b/package/pyproject.toml
@@ -118,3 +118,8 @@ find = {}
 MDAnalysis = [
     'analysis/data/*.npy',
 ]
+
+[tool.black]
+line-length = 79
+target-version = ['py39', 'py310', 'py311', 'py312']
+extend-exclude = '.'
--- a/package/pyproject.toml
+++ b/package/pyproject.toml
@@ -123,3 +123,4 @@ MDAnalysis = [
 line-length = 79
 target-version = ['py39', 'py310', 'py311', 'py312']
 extend-exclude = '.'
+required-version = '24'
--- a/package/pyproject.toml
+++ b/package/pyproject.toml
@@ -121,6 +121,6 @@ MDAnalysis = [
 
 [tool.black]
 line-length = 79
-target-version = ['py39', 'py310', 'py311', 'py312']
+target-version = ['py310', 'py311', 'py312']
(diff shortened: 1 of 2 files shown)


package/pyproject.toml
```diff
@@ -118,3 +118,8 @@ find = {}
 MDAnalysis = [
     'analysis/data/*.npy',
 ]
+
+[tool.black]
+line-length = 79
```
RMeli, line 123: We need to decide if we really want to stick to a line length of `79` as expressed in the UserGuide or use `black`'s default of `88`.

RMeli: https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length

RMeli: @IAlibay @PicoCentauri thoughts?

IAlibay: 79, let's pin to one of the standards (I think you can to the right), and otherwise lgtm!

testsuite/pyproject.toml
```diff
@@ -151,3 +151,8 @@ filterwarnings = [
     # NamedStream warnings
     "ignore:Constructed NamedStream:RuntimeWarning",
 ]
+
+[tool.black]
+line-length = 79
+target-version = ['py39', 'py310', 'py311', 'py312']
```
IAlibay: drop py39

IAlibay (approved):
Just the one thing

package/pyproject.toml
RMeli: Tested a pin to 23 with black 24:
```
⏰ 22:39:00 ☑️   black package
Oh no! 💥 💔 💥 The required version `23` does not match the running version `24.8.0`!
```

PR #4729, opened 2024-10-08
'MDAnalysis.analysis.density' parallelization 

### LINKED ISSUE #4677: `MDAnalysis.analysis.density`: Implement parallelization or mark as unparallelizable
Given a newly implemented parallelization (https://docs.mdanalysis.org/dev/documentation_pages/analysis/parallelization.html), implement it for all classes in `analysis.density` module. You may use `analysis.RMSD` in `develop` as a reference.

You need to add:

 - appropriate implementation of result aggregation for AnalysisBase subclasses in `analysis.density` module (or explicitly limit available backends if the classes aren't parallelizable)
 - a boilerplate fixture(s) to `testsuite/analysis/conftest.py`, analogous with existing ones
 - a `client_...` fixtures to all tests using in `testsuite/MDAnalysisTests/analysis/test_density.py`, and modify the way run() method is called

talagayev:
Fixes #4677

 - added backends and aggregators to `DensityAnalysis` in `analysis.density`
 - added `client_DensityAnalysis` in `conftest.py`
 - added `client_DensityAnalysis` to the tests in `test_density.py`

PR Checklist
------------
 - [x] Tests?
 - [x] Docs?
 - [x] CHANGELOG updated?

----
📚 Documentation preview 📚: https://mdanalysis--4729.org.readthedocs.build/en/4729/


yuxuanzhuang:
Three aspects in my fix.

- Attributes (private or not) that will be used later in `_conclude` should be defined in `__init__`, not in `_prepare`. @marinegor, I think it would be helpful to explicitly mention this in the documentation. Additionally, while the illustrative figure (https://docs.mdanalysis.org/dev/documentation_pages/analysis/parallelization.html#split-apply-combine) suggests that `_prepare` occurs before splitting, this is not the case.

- The `_edges` and similar attributes must be defined in` __init__` by nature because they need to be consistent across all workers and are determined based on the current positions of the atomgroup (https://github.com/MDAnalysis/mdanalysis/pull/4729/files#diff-28b7c110c6231f9f2ea864bbdecb9cb485edbed494d7e266d37896bde7b50b0dR421) .

- The results being aggregated are actually `_grid`, not `Density`—`Density` serves more as a postprocessing step that summarizes the information.

orbeckst:
@yuxuanzhuang could you please raise issues for the updates to the parallelization documentation (https://github.com/MDAnalysis/mdanalysis/pull/4729#issuecomment-2403681751) — I think that's very important to make clear, given that we are now repeatedly running in these issues. Thank you!

marinegor:
@yuxuanzhuang thanks, the illustration is indeed wrong, as well as the accompanying text. I fixed them in #4760

package/MDAnalysis/analysis/density.py
```diff
@@ -465,7 +470,7 @@ def _prepare(self):
         grid, edges = np.histogramdd(np.zeros((1, 3)), bins=bins,
                                      range=arange, density=False)
         grid *= 0.0
-        self._grid = grid
+        self.results._grid = grid
```
orbeckst, line 481: The overall grid shape needs to be determined in `__init__`, then each worker can have its own.

orbeckst (changes requested):
As we've been figuring out: data structures need to be laid out in `__init__`, thus the size of `_grid` needs to be set up before `_prepare` (which is executed in each worker process).

package/MDAnalysis/analysis/density.py
talagayev: Ah okay, yes that makes sense.

I think there was also an additional case, where I encountered this problem with the parallelization;
I will try to apply the same fix there to see if it works.

Other than that I added the Changelog and the docs, so should be ready to go :)

package/MDAnalysis/analysis/density.py
```diff
@@ -412,7 +420,6 @@ def __init__(self, atomgroup, delta=1.0,
         self._ydim = ydim
         self._zdim = zdim
 
-    def _prepare(self):
```
orbeckst, line 415: This looks like the right change.

I am noticing that the necessary changes for parallelization really shift a lot of what `_prepare()` was doing back into `__init__`. I guess that's ok and it's better than introducing yet another method like `_prepare_init()` ... We probably need to review the docs for writing analysis code with AnalysisBase.

orbeckst: Add a comment so that future readers know why we are doing everything here
and not in `_prepare()`.
```suggestion
        # The grid with its dimensions has to be set up in __init__
        # so that parallel analysis works correctly: each process
        # needs to have a `results._grid` of the same size and the
        # same `_bins`.
```

testsuite/MDAnalysisTests/analysis/test_density.py
```diff
@@ -247,125 +255,176 @@ def check_DensityAnalysis(self, ag, ref_meandensity,
...
+            delta=1.0,
+            xdim=8.0,
+            ydim=12.0,
+            zdim=17.0,
+            gridcenter=self.gridcenters["static_defined"],
+        ).run(**client_DensityAnalysis)
         assert D.results.density.grid.shape == (8, 12, 17)
 
-    def test_warn_userdefn_padding(self, universe):
+    def test_warn_userdefn_padding(self, universe, client_DensityAnalysis):
```
orbeckst, line 314: I am not 100% sure if we need to test warnings for all parallel modalities, too, but it's the safe approach and you've already converted all tests so let's leave it in. Just something to think about in the future if our test run times are too long (again...).

orbeckst (approved):
This looks fine.

I am adding a single comment, otherwise my comments on the code are just observations.

package/MDAnalysis/analysis/density.py
orbeckst: manually added to the file, suggestions don't work on deleted lines

PR #4732, opened 2024-10-13
Add py3.13 support w/ minimal deps

IAlibay:
Python 3.13 should now work with minimal deps. Optional deps like OpenMM don't have compatible releases yet.

 - Bump up non-optional deps workflows to include py3.13
 - Add py3.13 to the classifiers list

PR Checklist
------------
 - [x] Tests?
 - [x] CHANGELOG updated?

----
📚 Documentation preview 📚: https://mdanalysis--4732.org.readthedocs.build/en/4732/

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -26,13 +26,13 @@ jobs:
           PYTHON_ARCH: 'x64'
           BUILD_TYPE: 'normal'
           imageName: 'windows-2019'
-        Win-Python312-64bit-full:
-          PYTHON_VERSION: '3.12'
+        Win-Python313-64bit-full:
+          PYTHON_VERSION: '3.13'
           PYTHON_ARCH: 'x64'
           BUILD_TYPE: 'normal'
           imageName: 'windows-2019'
-        Win-Python312-64bit-full-wheel:
-          PYTHON_VERSION: '3.12'
+        Win-Python313-64bit-full-wheel:
+          PYTHON_VERSION: '3.13'
           PYTHON_ARCH: 'x64'
           BUILD_TYPE: 'wheel'
           NUMPY_MIN: '1.26.0'
@@ -43,8 +43,8 @@ jobs:
           BUILD_TYPE: 'wheel'
           NUMPY_MIN: '1.23.2'
           imageName: 'windows-2019'
-        Linux-Python312-64bit-full-wheel:
-          PYTHON_VERSION: '3.12'
+        Linux-Python313-64bit-full-wheel:
+          PYTHON_VERSION: '3.13'
           PYTHON_ARCH: 'x64'
           BUILD_TYPE: 'wheel'
           NUMPY_MIN: '1.26.0'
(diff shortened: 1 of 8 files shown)


testsuite/MDAnalysisTests/analysis/test_base.py
```diff
@@ -345,17 +345,17 @@ def test_verbose_progressbar(u, capsys):
 def test_verbose_progressbar_run(u, capsys):
     FrameAnalysis(u.trajectory).run(verbose=True)
     _, err = capsys.readouterr()
-    expected = u'100%|██████████| 98/98 [00:00<00:00, 8799.49it/s]'
+    expected = u'100%|██████████'
     actual = err.strip().split('\r')[-1]
-    assert actual[:24] == expected[:24]
+    assert expected == actual[:15]
```
orbeckst: Why did you switch actual and expected?

Doesn't really matter except for consistency throughout because in [testsuite/MDAnalysisTests/lib/test_log.py](https://github.com/MDAnalysis/mdanalysis/pull/4732/files#diff-3f7a25a1baa99374633c43432096cd1139a273ab31a5d45d05843801aca8dcff) it's still actual == expected.

testsuite/MDAnalysisTests/analysis/test_base.py
```diff
@@ -345,17 +345,17 @@ def test_verbose_progressbar(u, capsys):
 def test_verbose_progressbar_run(u, capsys):
     FrameAnalysis(u.trajectory).run(verbose=True)
     _, err = capsys.readouterr()
-    expected = u'100%|██████████| 98/98 [00:00<00:00, 8799.49it/s]'
```
orbeckst, line 348: I think the reason why the full output was there originally was to give devs an idea what it looks like even if other parts could not be reliably tested.

orbeckst (approved):
Given that this passes all tests for Python 3.13 I am approving.

🎉

testsuite/MDAnalysisTests/analysis/test_base.py
IAlibay: I had originally switched it to a `expected in actual` check, but decided to revert back to an equality check.

testsuite/MDAnalysisTests/analysis/test_base.py
IAlibay: Unfortunately the py3.13 behaviour changed here and I can't work out exactly why - I'm not sure if there's a better way to deal with this without having an explicit python version check :/

testsuite/MDAnalysisTests/analysis/test_base.py
IAlibay: Reverted the order in the latest commit.

tylerjereddy (approved):
LGTM as well, test shims seem reasonable enough.

azure-pipelines.yml
```diff
@@ -47,7 +47,7 @@ jobs:
           PYTHON_VERSION: '3.12'
           PYTHON_ARCH: 'x64'
           BUILD_TYPE: 'wheel'
-          NUMPY_MIN: '1.26.0'
+          NUMPY_MIN: '2.1.0'
```
tylerjereddy, line 50: did you mean to change to `3.13` above? I don't think that's a blocker here though

